PR #552, opened 2018-08-28
Automatically sync chart

andresmgot:
Fixes #414

Automatically release a chart in the repository `bitnami/charts` for Kubeapps. The process for updating a version is:

 1. Add a new tag for the repo (as always). This will generate all the images required for the new version.
 2. Increase the version number of the chart at `chart/kubeapps/Chart.yaml`. Once that change is in `master` CircleCI will update the status of the `bitnami/charts` repo with the files in this repository.

**IMPORTANT**: The script in 2. will automatically change the references to `tag: latest` in the file `chart/kubeapps/values.yaml` with the latest tag available (normally the tag released in 1.). 

Update example: https://github.com/andresmgot/charts-1/commit/d3c69181bf0e40428b0c83e86d66207d867a98ae

PR #554, opened 2018-08-29
Fix previous tag retrieval

andresmgot:
Discovered at https://github.com/kubeapps/kubeapps/pull/552#discussion_r213500706

The current approach is being executed when the tag is already committed so it's retrieving the latest tag instead of the previous one. If `git describe --tags` is used with the previous commit it shows a dirty tag (something like `v1.0.0-alpha.5-fdskl`). To avoid that we can use just the `git tag` command and use the second-to-last.

### DIFF
--- a/script/release_utils.sh
+++ b/script/release_utils.sh
@@ -6,8 +6,7 @@ function commit_list {
   local repo_domain=${2:?}
   local repo_name=${3:?}
   git fetch --tags
-  # Get latest two tags and filter the last one (which is the current commit)
-  local previous_tag=`git tag -l --sort=committerdate | tail -n 2 | head -n 1`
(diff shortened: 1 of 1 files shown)


script/release_utils.sh
```diff
@@ -6,7 +6,8 @@ function commit_list {
   local repo_domain=${2:?}
   local repo_name=${3:?}
   git fetch --tags
-  local previous_tag=`git describe --tags $(git rev-list --tags --max-count=1)`
+  # Get latest two tags and filter the last one (which is the current commit)
+  local previous_tag=`git tag -l --sort=committerdate | tail -n 2 | head -n 1`
```
prydonius: here is a way to do it with git
```
git describe --abbrev=0 --tags `git rev-list --tags --skip=1 --max-count=1`
```

andresmgot: true

Issue #481, opened 2018-08-14
Honor Helm CLI behavior in the UI


migmartri:
Currently, Kubeapps behaves in a different way compared to the Helm CLI itself in the following

 ### Listing charts

Helm lists by default only FAILED and DEPLOYED releases, Kubeapps list them all.

### Delete releases

Helm soft-deletes by default, Kubeapps purge them

We should probably honor the Helm behavior by 

* List by default the same way the CLI does, adding a "show all" filter in the listing
* Soft delete by default but add a checkbox in the confirm dialog to purge the release.

andresmgot:
Do we really want to do that? When listing applications it's useful to see all the applications even the ones that are deleted. If someone tries to reuse a name (because it has been deleted) and it's not listed anywhere it will be confusing.

The same applies to deleting a release, I think it's more common for people to want to delete+purge a release than just delete them. I am okay if we have a field to avoid purging the release though.

migmartri:
Thanks for the comment @andresmgot. You raise a good question. I do not know if we want to. I would be in favor of doing it but before some clarifications.

* I am not advocating for never showing deleted apps but changing the default view instead.
* Listing the deleted and failed apps should be accessible easily by toggling a filter.
* The purge functionality will be still there.

Knowing that, this is why I think we should change it.

- I would not be surprised that most Helm users don't know what delete actually means (from the implementation standpoint), they just see it disappear from the `helm list` command. And this works fine because Helm relies on autogenerated release names by default, so conflicts are prevented. This also means that potentially there will be many of these deleted releases already there (I leave them behind all the time).

- Helm deletion, by default, does not remove the configmap releases, so even in small installations the list shown in kubeapps can be, in my opinion, overwhelming and confusing.

- When we used Helm CRD we had our own reality of chart releases and our deletion by default purged, so technically this discussion did not apply by then. Now, we should assume that Kubeapps can be connected to their existing Tiller, which potentially will have deleted releases CM stored. I would expect as an user to see what I see in Helm list by default.

- I do not think that we should optimize the dashboard for the cluster operator but for the end user instead. Sure, a cluster operator might want to see by default the deleted releases to clean them up, audit them and so on, but as an end user you might just care about the apps that are deployed.
 
> If someone tries to reuse a name (because it has been deleted) and it's not listed anywhere it will be confusing.

[...]

PR #557, opened 2018-08-29
Add status query for listing releases

andresmgot:
Ref: #481

Mimic the list behavior of the Helm CLI. If no `status` is specified it will return only `DEPLOYED` and `FAILED` apps. It's possible to specify a custom query like `http://tiller-proxy.kubeapps:8080/v1/releases?statuses=deleted`.

### DIFF
--- a/cmd/tiller-proxy/internal/handler/handler.go
+++ b/cmd/tiller-proxy/internal/handler/handler.go
@@ -233,7 +233,7 @@ func (h *TillerProxy) UpgradeRelease(w http.ResponseWriter, req *http.Request, p
 
 // ListAllReleases list all releases that Tiller stores
 func (h *TillerProxy) ListAllReleases(w http.ResponseWriter, req *http.Request) {
-	apps, err := h.ProxyClient.ListReleases("", h.ListLimit)
+	apps, err := h.ProxyClient.ListReleases("", h.ListLimit, req.URL.Query().Get("status"))
 	if err != nil {
 		response.NewErrorResponse(errorCode(err), err.Error()).Write(w)
 		return
@@ -243,7 +243,7 @@ func (h *TillerProxy) ListAllReleases(w http.ResponseWriter, req *http.Request) 
 
 // ListReleases in the namespace given as Param
 func (h *TillerProxy) ListReleases(w http.ResponseWriter, req *http.Request, params Params) {
-	apps, err := h.ProxyClient.ListReleases(params["namespace"], h.ListLimit)
+	apps, err := h.ProxyClient.ListReleases(params["namespace"], h.ListLimit, req.URL.Query().Get("status"))
 	if err != nil {
 		response.NewErrorResponse(errorCode(err), err.Error()).Write(w)
 		return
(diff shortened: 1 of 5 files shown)


pkg/proxy/proxy.go
```diff
@@ -161,12 +161,45 @@ func filterList(rels []*release.Release) []*release.Release {
...
 
+func getStatuses(statusQuery string) []release.Status_Code {
+	if statusQuery == "" {
+		// Default case
+		return []release.Status_Code{
+			release.Status_DEPLOYED,
+			release.Status_FAILED,
+		}
+	} else if strings.Contains(statusQuery, "all") {
+		return releaseStatuses
```
prydonius: let's call these allReleaseStatuses and add a comment that it matches Helm list https://github.com/helm/helm/blob/master/cmd/helm/list.go#L214

cmd/tiller-proxy/internal/handler/handler.go
```diff
@@ -233,7 +233,7 @@ func (h *TillerProxy) UpgradeRelease(w http.ResponseWriter, req *http.Request, p
 
 // ListAllReleases list all releases that Tiller stores
 func (h *TillerProxy) ListAllReleases(w http.ResponseWriter, req *http.Request) {
-	apps, err := h.ProxyClient.ListReleases("", h.ListLimit)
+	apps, err := h.ProxyClient.ListReleases("", h.ListLimit, req.URL.Query().Get("status"))
```
prydonius: Should we call the parameter "statuses" since you can add multiple comma separated?

pkg/proxy/proxy_test.go
```diff
@@ -73,7 +73,7 @@ func TestListAllReleases(t *testing.T) {
 	proxy := newFakeProxy([]AppOverview{app1, app2})
 
 	// Should return all the releases if no namespace is given
-	releases, err := proxy.ListReleases("", 256)
+	releases, err := proxy.ListReleases("", 256, "")
```
migmartri: Should we have a test case that filters through some releases here?

pkg/proxy/proxy.go
```diff
@@ -161,12 +161,45 @@ func filterList(rels []*release.Release) []*release.Release {
 	return uniq
 }
 
+func getStatuses(statusQuery string) []release.Status_Code {
```
migmartri: Did we get this function from anywhere?

pkg/proxy/proxy.go
migmartri: We should maybe rename `releaseStatuses` to `allReleaseStatuses`

cmd/tiller-proxy/internal/handler/handler.go
migmartri: I agree with calling them `statuses`

pkg/proxy/proxy.go
migmartri: What will happen in the query is invalid? Will we return an empty list? An error?

andresmgot: nope, it's quite similar to what the helm CLI does, I can add the reference

andresmgot: An empty list. I think that's the most logic thing to do. Being literal any app will comply with that status. Note that the "error" will be logged.

pkg/proxy/proxy_test.go
andresmgot: I couldn't do that because the fake helm client ignores most of the filters: https://github.com/helm/helm/blob/master/pkg/helm/fake.go#L61 (that's what I tried to explain the TODO below. I couldn't find a way of at least checking that we are passing the correct parameters to helm.

PR #558, opened 2018-08-29
disable object-literal-sort-keys in dev tslint config

prydonius:
We disabled this in #544 but I think we forgot to disable it in the dev
tslint config.

cc @andresmgot


prydonius:
@andresmgot this is the default file that will be used by your editor and dev server

PR #569, opened 2018-08-30
Chart update: Tiller Proxy namespace awareness and rbac support

migmartri:
When a repository contains an authentication header, Tiller Proxy needs to read this header from the repo associate k8s secret placed in the RELEASE_NAMESPACE.

This namespace is decided via the POD_NAMESPACE env variable https://github.com/kubeapps/kubeapps/blob/master/pkg/chart/chart.go#L231-L236

This patch injects the namespace in which the pod is running via the k8s downstream API as well as extend Tiller-proxy's RBAC support (serviceAccount, role and rolebinding) to be able to read from those secrets.   

Upgrades have been tested successfully in Minikube

### DIFF
--- a/chart/kubeapps/templates/tiller-proxy-rbac.yaml
+++ b/chart/kubeapps/templates/tiller-proxy-rbac.yaml
@@ -1,0 +1,36 @@
+{{- if .Values.rbac.create -}}
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: Role
+metadata:
+  name: {{ template "kubeapps.tiller-proxy.fullname" . }}
+  labels:
+    app: {{ template "kubeapps.tiller-proxy.fullname" . }}
+    chart: {{ template "kubeapps.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - get
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: RoleBinding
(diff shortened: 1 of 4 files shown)


chart/kubeapps/templates/tiller-proxy-rbac.yaml
```diff
@@ -0,0 +1,36 @@
...
+    chart: {{ template "kubeapps.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - get
```
andresmgot: we can be careful here and only access to the secret `{{ template "kubeapps.tiller-proxy.fullname" . }}`

migmartri: Is this actually the case? 

I thought that tiller-proxy needs to access to any secret associated to every one of the repositories that have auth info no? https://github.com/kubeapps/kubeapps/blob/master/dashboard/src/actions/repos.ts#L123

prydonius: Yes, it needs to access all secrets because it will be reading the secret named after a added apprepository.

andresmgot: oh, true, I forgot about the repository secrets. I thought this was just for the TLS secret

prydonius: The TLS secret is mounted in as a volume, so an RBAC role is not required for it.

PR #574, opened 2018-08-31
[CI] Avoid getting stuck while checking if the cluster is available

### LINKED ISSUE #571: E2E tests hang
This job has been hanging for 5 hours https://circleci.com/gh/kubeapps/kubeapps/944

```
RROR: (gcloud.container.clusters.delete) Some requests did not succeed:
 - ResponseError: code=400, message=EXTERNAL: Operation operation-1535653491640-ec212c7c is currently creating cluster kubeapps-test-master-11-ci. Please wait and try again once it is done.

WARNING: Starting in 1.12, new clusters will have basic authentication disabled by default. Basic authentication can be enabled (or disabled) manually using the `--[no-]enable-basic-auth` flag.
WARNING: Starting in 1.12, new clusters will not have a client certificate issued. You can manually enable (or disable) the issuance of the client certificate using the `--[no-]issue-client-certificate` flag.
WARNING: Currently VPC-native is not the default mode during cluster creation. In the future, this will become the default mode and can be disabled using `--no-enable-ip-alias` flag. Use `--[no-]enable-ip-alias` flag to suppress this warning.
This will enable the autorepair feature for nodes. Please see
https://cloud.google.com/kubernetes-engine/docs/node-auto-repair for more
information on node autorepairs.

WARNING: Starting in Kubernetes v1.10, new clusters will no longer get compute-rw and storage-ro scopes added to what is specified in --scopes (though the latter will remain included in the default --scopes). To use these scopes, add them explicitly to --scopes. To use the new behavior, set container/new_scopes_behavior property (gcloud config set container/new_scopes_behavior true).
ERROR: (gcloud.container.clusters.create) ResponseError: code=409, message=EXTERNAL: Already exists: projects/helm-publish-ci/zones/us-east1-c/clusters/kubeapps-test-master-11-ci.
The connection to the server localhost:8080 was refused - did you specify the right host or port?
The connection to the server localhost:8080 was refused - did you specify the right host or port?
```

andresmgot:
Fixes #571

The notation `((cnt=cnt-1)) || (echo "Waited 20 seconds but cluster is not reachable" && exit 1)` was not working so I changed that with a boring `if` clause. I have also redirected the output of `kubectl get pods` to `/dev/null` so in case the condition somehow fails CircleCI will purge the job after 10 minutes without output.

I tried the changes in the CircleCI env:

```
circleci@9ed20a708281:~/project$ ./s.sh
Waiting for the cluster to respond...
Waited 20 seconds but cluster is not reachable
circleci@9ed20a708281:~/project$ echo $?
1
```

### DIFF
--- a/script/start-gke-env.sh
+++ b/script/start-gke-env.sh
@@ -29,10 +29,14 @@ fi
 
 echo "Creating cluster $CLUSTER in $ZONE (v$VERSION)"
 gcloud container clusters create --cluster-version=$VERSION --zone $ZONE $CLUSTER --num-nodes 5 --machine-type=n1-standard-2
-# Wait for the cluster to respond
+echo "Waiting for the cluster to respond..."
 cnt=20
-until kubectl get pods; do
-    ((cnt=cnt-1)) || (echo "Waited 20 seconds but cluster is not reachable" && exit 1)
+until kubectl get pods > /dev/null 2>&1; do
+    ((cnt=cnt-1))
+    if [[ "$cnt" -eq 0 ]]; then
+        echo "Waited 20 seconds but cluster is not reachable"
+        exit 1
+    fi
     sleep 1
 done
 
--- a/script/start-gke-env.sh
+++ b/script/start-gke-env.sh
@@ -34,7 +34,7 @@ cnt=20
 until kubectl get pods > /dev/null 2>&1; do
     ((cnt=cnt-1))
     if [[ "$cnt" -eq 0 ]]; then
-        echo "Waited 20 seconds but cluster is not reachable"
+        echo "Tried 20 times but the cluster is not reachable"
         exit 1
     fi
     sleep 1


script/start-gke-env.sh
```diff
@@ -29,10 +29,14 @@ fi
 
 echo "Creating cluster $CLUSTER in $ZONE (v$VERSION)"
 gcloud container clusters create --cluster-version=$VERSION --zone $ZONE $CLUSTER --num-nodes 5 --machine-type=n1-standard-2
-# Wait for the cluster to respond
+echo "Waiting for the cluster to respond..."
 cnt=20
```
migmartri: So this means that this 20 seconds check limit was not working right? So now, is it 20 seconds enough?  What's in your experience the average time that takes a cluster to be reachable?

Maybe we should play it safer? At the end of the day for example 60 seconds does not seem a big deal to me

prydonius: Yes this didn't work because as far as I can tell `((cnt=cnt-1))` will never fail, it will just keep subtracting into negative integers.

I agree that we might want to try 60 here though.

andresmgot: these are actually retries not exactly seconds (it depends in how long `kubectl get pods` takes). In my experience if after a couple of retries this doesn't work it's because there is an issue and it won't work. Note that `gcloud container clusters create` only returns control when the cluster is healthy and `.kube/config` is properly configured unless there is an error.

Let's keep this short, if we find errors with this loop after getting a success in the cluster creation we can increase it, in other case we will just be adding overhead for failed builds.

migmartri: Thanks for the explanation!

So then the concept of time here is not accurate then no? Maybe it should be called retries?

prydonius: agreed, we should rename this to retries.

script/start-gke-env.sh
```diff
@@ -29,10 +29,14 @@ fi
...
-until kubectl get pods; do
-    ((cnt=cnt-1)) || (echo "Waited 20 seconds but cluster is not reachable" && exit 1)
+until kubectl get pods > /dev/null 2>&1; do
+    ((cnt=cnt-1))
+    if [[ "$cnt" -eq 0 ]]; then
+        echo "Waited 20 seconds but cluster is not reachable"
```
prydonius: change this to "Tried 20 times but the cluster is not reachable"

PR #578, opened 2018-08-31
Allow to not purge applications when deleting them

andresmgot:
Closes: #481
Requires: #563

 - Adds a query parameter "purge" for the `DELETE` API call in tiller-proxy.
 - Show a checkbox  when deleting an app in the UI to purge or not the application.

screen shot 2018-08-31 at 18 13 57

### DIFF
--- a/dashboard/src/actions/apps.test.tsx
+++ b/dashboard/src/actions/apps.test.tsx
@@ -55,14 +55,14 @@ describe("fetches applications", () => {
 });
 
 describe("delete applications", () => {
-  const deleteApp = App.delete;
+  const deleteAppOrig = App.delete;
   let deleteAppMock: jest.Mock;
   beforeEach(() => {
-    App.delete = jest.fn(() => []);
-    deleteAppMock = App.delete as jest.Mock;
+    deleteAppMock = jest.fn(() => []);
+    App.delete = deleteAppMock;
   });
   afterEach(() => {
-    App.delete = deleteApp;
+    App.delete = deleteAppOrig;
   });
   it("delete an application", async () => {
     await store.dispatch(actions.apps.deleteApp("foo", "default", false));
--- a/dashboard/src/actions/apps.test.tsx
+++ b/dashboard/src/actions/apps.test.tsx
@@ -53,3 +53,34 @@ describe("fetches applications", () => {
     });
   });
 });
+
+describe("delete applications", () => {
+  const deleteApp = App.delete;
+  let deleteAppMock: jest.Mock;
+  beforeEach(() => {
+    App.delete = jest.fn(() => []);
+    deleteAppMock = App.delete as jest.Mock;
+  });
+  afterEach(() => {
+    App.delete = deleteApp;
+  });
+  it("delete an application", async () => {
+    await store.dispatch(actions.apps.deleteApp("foo", "default", false));
+    expect(store.getActions()).toEqual([]);
+    expect(deleteAppMock.mock.calls[0]).toEqual(["foo", "default", false]);
+  });
+  it("delete and purge an application", async () => {
+    await store.dispatch(actions.apps.deleteApp("foo", "default", true));
+    expect(store.getActions()).toEqual([]);
+    expect(deleteAppMock.mock.calls[0]).toEqual(["foo", "default", true]);
+  });
+  it("delete and throw an error", async () => {
+    const error = new Error("something went wrong!");
+    const expectedActions = [{ type: getType(actions.apps.errorDeleteApp), err: error }];
+    deleteAppMock.mockImplementation(() => {
+      throw error;
+    });
+    expect(await store.dispatch(actions.apps.deleteApp("foo", "default", true))).toBe(false);
+    expect(store.getActions()).toEqual(expectedActions);
+  });
+});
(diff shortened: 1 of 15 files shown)


dashboard/src/actions/apps.test.tsx
```diff
@@ -65,3 +65,34 @@ describe("fetches applications", () => {
     });
   });
 });
+
+describe("delete applications", () => {
+  const deleteApp = App.delete;
+  let deleteAppMock: jest.Mock;
+  beforeEach(() => {
+    App.delete = jest.fn(() => []);
```
prydonius: as suggested in https://github.com/kubeapps/kubeapps/pull/563#discussion_r214468729, I think would be more readable as:

```
deleteAppMock = jest.fn(() => []);
App.delete deleteAppMock;
```

dashboard/src/actions/apps.test.tsx
```diff
@@ -65,3 +65,34 @@ describe("fetches applications", () => {
     });
   });
 });
+
+describe("delete applications", () => {
+  const deleteApp = App.delete;
```
prydonius: as suggested in https://github.com/kubeapps/kubeapps/pull/563#discussion_r214469388, I think this could be called `deleteAppOrig` or something

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -152,7 +152,7 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
       return <div>Loading</div>;
     }
     return (
-      <section className="AppView padding-b-big">
+      <section id="appview" className="AppView padding-b-big">
```
prydonius: do we need this?

prydonius: guessing you added this for the Modal.setAppElement thing, but ultimately decided to set ariaHideApp to false?

dashboard/src/components/ConfirmDialog/index.tsx
```diff
@@ -36,6 +37,7 @@ class ConfirmDialog extends React.Component<IConfirmDialogProps, IConfirmDialogS
           isOpen={this.props.modalIsOpen}
           onRequestClose={this.closeModal}
           contentLabel="Modal"
+          ariaHideApp={false}
```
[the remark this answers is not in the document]
prydonius: I think this is fine for our usecase, wdyt @Angelmmiguel?

dashboard/src/shared/App.test.ts
```diff
@@ -57,4 +57,20 @@ describe("App", () => {
...
+      expect(await App.delete("foo", "default", false)).toBe("ok");
+    });
+    it("should delete and purge an app in a namespace", async () => {
+      moxios.stubRequest("/api/tiller-deploy/v1/namespaces/default/releases/foo?purge=true", {
+        response: "ok",
+        status: 200,
+      });
+      expect(await App.delete("foo", "default", true)).toBe("ok");
+    });
+  });
```
prydonius: just noticed we don't have any tests where responses are not 200, should we add those?

pkg/proxy/proxy_test.go
```diff
@@ -329,7 +329,8 @@ func TestHelmReleaseDeleted(t *testing.T) {
 	app := AppOverview{"foo", "1.0.0", "my_ns", "icon.png", "DEPLOYED"}
 	proxy := newFakeProxy([]AppOverview{app})
 
-	err := proxy.DeleteRelease(app.ReleaseName, app.Namespace)
+	// TODO: Add a test for a non-purged release when the fake helm cli supports it
```
prydonius: I'm confused, didn't you update fake/proxy to support it?

dashboard/src/components/ConfirmDialog/index.tsx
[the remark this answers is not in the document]
Angelmmiguel: It's not a visual change. It affects on how screen readers identify the current text in the screen :)

dashboard/src/components/AppView/AppControls.test.tsx
```diff
@@ -71,7 +71,7 @@ it("calls delete function with additional purge", () => {
   const app = new hapi.release.Release({ name, namespace });
   const deleteApp = jest.fn(() => false); // Return "false" to avoid redirect when mounting
   // mount() is necessary to render the Modal
-  const wrapper = mount(<AppControls app={app} deleteApp={deleteApp} />);
+  const wrapper = mount(<AppControls app={app} deleteApp={deleteApp} ariaHideApp={false} />);
```
Angelmmiguel: I believe we should fix the app element issue in the test instead of allowing developers to disable accessibility properties in the component. Also, we're adding some logic around the final component implementation that only will be used in tests.

The `appElement` is the root node where all the content belongs to. A possible solution is to create a `div` directly in all the tests that uses `ReactModal`:

```
ReactModal.setAppElement(document.createElement('div'));
```

You've another workarounds in https://github.com/reactjs/react-modal/issues/632

andresmgot: that worked! thanks @Angelmmiguel

PR #579, opened 2018-08-31
add SASS preprocessing

### LINKED ISSUE #549: Add support for SCSS compilation
We should add support for SCSS preprocessing for CSS in the dashboard. This will allow us to have more flexibility with variables and also support theming.

Hex theming: https://design.bitnami.com/category/Theming/

We should also add SCSS linting to the build pipeline.

prydonius:
Follows the directions described in
https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#adding-a-css-preprocessor-sass-less-etc
to enable SASS preprocessing support to the React app.

Renames css files to scss.

fixes #549


andresmgot:
Thanks for the changes! I guess you need to change the `test_dashboard` job to run the `build` step before the tests.

Apart from that: does this change the development experience in something? Do we need to run any additional command? If so, should we update the dev docs?

migmartri:
> Apart from that: does this change the development experience in something? Do we need to run any additional command? If so, should we update the dev docs?

In theory it should not. Something that I have noticed in the past though is that I could import both the css and the scss files from my module when what we want in this case is to import the css file right?

prydonius:
> In theory it should not. Something that I have noticed in the past though is that I could import both the css and the scss files from my module when what we want in this case is to import the css file right?

Yes, we should be careful to import the css into the react components instead of the scss.

prydonius:
> Apart from that: does this change the development experience in something? Do we need to run any additional command? If so, should we update the dev docs?

Good question! I don't think so, as the start, build and test commands have been changed to also precompile the CSS, and I we're always using those commands.

PR #583, opened 2018-09-04
Speed up CI builds

### LINKED ISSUE #522: Speed up image builds
The current (default) Makefile tasks for building the images that Kubeapps is composed of (based on Go) have a couple of points of improvement regarding performance:

 - It builds the image from the root directory. This causes that Docker copies the whole repository to the docker environment.
 - The Dockerfiles copies the whole root directory to the GOPATH inside the container.
 - The compilation takes place within the container. If Docker is virtualized in the host running the build the CPU will be limited to the CPU of the virtual machine.

As an alternative we can compile the binary in the host machine and just copy the binary to the container (the tiller-proxy image uses this approach). The result is significantly faster in the CI (1min vs 2min) and way faster in a local environment (for a rebuild 8s vs 2min).

andresmgot:
Closes #522

Speed up test and build jobs for Kubeapps micro services.
Time improvement: from 7 to 3 minutes. The compute time (addition of the time of all the jobs) has been reduced from ~11 to ~8 minutes.

Changes:
 - Build go binaries outside docker containers. We don't need to copy the context and we can reuse the cached files.
 - Build all go-based images in the same job. We can reuse docker and golang caching systems. We are also reducing the amount of "machine" jobs that are more difficult to obtain from CircleCI. 
 - Don't wait for the unit tests to build the images. We can run all the jobs in parallel.

Compared builds:
  - Existing solution: https://circleci.com/workflow-run/6637524d-19d6-49a4-8b34-553abc0e520e
  - With these changes: https://circleci.com/workflow-run/15e0a05b-ddaa-4a63-9380-6892e9668f75


migmartri:
The issue with this approach is that now we depend on the version of go existing in the executor no? Also in order to reproduce the build locally we need to be sure we have the same version, flags and so on than the CI no?

What about a middle ground? How long does it take to build all images inside the same machine executor the old way (Building inside a container?), that should reduce the machine provisioning time, pull time for `quay.io/deis/go-dev`, potentially we will get some layer caching too (context copying does not seem a big deal to me since our codebase seems small)

I am worried that for a speed optimization we trade build reproducibility.

prydonius:
@migmartri that still has the problem of having to upload the working directory to the docker context each time, which is quite slow. What if we used `quay.io/deis/go-dev:v1.8.2` as the executor container image? I'm not sure what the requirements are for CircleCI.

We currently all use Go locally in our dev environments, so even though that doesn't affect real builds, we still have the potential problem of mismatching Go versions today. Perhaps the way to solve this is to just document which version of Go is required?

andresmgot:
as @prydonius points out, building the binaries within a container has the downside of having to copy the context (which in my local environment takes a lot if with the `vendor/` and `node_modules/` folder). That's the main thing I want to avoid (almost 2 minutes in my case). 

> I am worried that for a speed optimization we trade build reproducibility.

I am not that worried about the reproducibility. Our coverage in go files is quite good, the images that the CI generates are being pushed (added in #584) so you can pull them and in the last case you can still open a container using the image that CircleCI uses and build the binary on your own. I don't think that's going to be a problem.

> What if we used quay.io/deis/go-dev:v1.8.2 as the executor container image?

I would stick to CircleCI images (rolling tag) just to not fall into possible issues in the documentation (that may only apply to official images). It's true that I used Go 1.9 in the images (because it was the one that tiller-proxy was using). We should unify that, do you think all of the images should use Go 1.8?

> Perhaps the way to solve this is to just document which version of Go is required?

Agree on documenting which major version the CI will use once we clarify if we are going to use 1.8 or 1.9 (or 1.10?).

prydonius:
@migmartri and I were looking into this and made some observations:

- previously, our builds were uploading a context of 84.11 MB
- with this change, the biggest context being uploaded is the tiller-proxy binary which comes to 39.72 MB
- vendor/ is only 39MB, which is roughly the same as the tiller-proxy binary
- dashboard/node_modules will be empty for the Go builds in the CI and therefore does not contribute to the 84.11 MB, it will be bigger in development though but we can exclude it
- .git/ is 38 MB and can be excluded
- docs/ is 5.5 MB and can be excluded
- the dashboard Dockerfile will run `yarn install` so the node_modules do not need to be there for the build to succeed (though it will make it slower to build in development since we won't copy them in, the CI always installs though)
- Adding a Dockerignore with .git, docs and dashboard/node_modules (for development) excluded will bring the whole context down to about 40.61 MB which again is roughly the same as the tiller-proxy binary
- We also save time in the CI by using a single executor for all the Go builds, and if we move to the remote docker instance

Given all this, I think we should add a Dockerignore file here and continue using the multi-stage builds we have today.

I also think, though this should be a separate change, we should move away from `quay.io/deis/go-dev:v1.8.2` and use `golang:1.10` instead. That image has more stuff that we probably don't need, and also the version doesn't actually map to the Go version which is confusing (the v1.8.2 tag uses Go 1.9).

andresmgot:
It's true that that will provide a similar result for a single image build in the CI but that's still painfully slow in my local development. If I compile within a container:

 - I only use 2 cores (instead of 4) because of the docker virtualization
 - The Go cache is not reused

That causes that a rebuild takes 2 minutes instead of 11 seconds (I get that numbers using `tiller-proxy` as example, excluding docs, .git and node_modules).

andresmgot:
Current CI time for compiling within containers: 4 min

prydonius:
Okay it sounds like this is now more of an issue of the development workflow for tiller-proxy and other Go components. I would say we should focus on improving the CI builds in this PR, and then follow up on the development environment later. I've created #589 to discuss the development environment.

WDYT @andresmgot @migmartri?

migmartri:
I agree with @prydonius let's not mix up CI/prod and development. For development in the meantime I'd ok if we add a local target or similar in our makefiles.

PR #584, opened 2018-09-04
[CI] Push and use temporary images

### LINKED ISSUE #537: Tagged and latest image builds should be pushed after the integration tests
Currently CircleCI pushes images before running integration tests or before other images are successfully built. This could have potential side effects where we release images that are in a broken state.

We should make sure we only push the latest and versioned image tags at the end of the pipeline (e.g. the release job). To do so, we may need to see if we can share the docker context in CircleCI, or we may need to push an intermediary tag (e.g. the commit sha), and the re-tag with latest (and the version for tag builds) at the end of the pipeline.

andresmgot:
Fixes #537

For the end-to-end tests:

 - Build and push images using as tag `build-<git commit sha1>`
 - Test the helm chart and set the tag to the images just built
 - Ensure that we are using the correct tag
 - If the e2e tests success push the images with another tag (`latest` or `v.*` for tagged builds)

Test: https://circleci.com/workflow-run/c369a345-eefb-46d5-a5ee-bc2e42c3fa45

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -70,6 +70,11 @@ workflows:
           requires:
             - GKE_1_9
             - GKE_1_10
+      - push_images:
+          <<: *build_on_master
+          requires:
+            - GKE_1_9
+            - GKE_1_10
       - release:
           <<: *build_on_tag
           requires:
@@ -89,7 +94,8 @@ exports: &exports
   run: |
     # It is not possible to resolve env vars in the environment section:
     # https://circleci.com/docs/2.0/env-vars/#using-bash_env-to-set-environment-variables
-    echo "export VERSION=${CIRCLE_TAG:-build-$CIRCLE_WORKFLOW_ID}" >> $BASH_ENV
+    echo "export DEV_TAG=build-${CIRCLE_SHA1}" >> $BASH_ENV
+    echo "export PROD_TAG=${CIRCLE_TAG:-latest}" >> $BASH_ENV
 build_image: &build_image
   machine: true
   working_directory: /home/circleci/.go_workspace/src/github.com/kubeapps/kubeapps
@@ -98,17 +104,11 @@ build_image: &build_image
   steps:
     - checkout
     - <<: *exports
-    - run: make VERSION="$VERSION" $IMAGE
+    - run: make VERSION="$DEV_TAG" $IMAGE
     - run: |
-        if [[ -z "$CIRCLE_PULL_REQUEST" ]]; then
+        if [[ -z "$CIRCLE_PULL_REQUEST" && -n "$DOCKER_USERNAME" && -n "$DOCKER_PASSWORD" ]]; then
           docker login -u="${DOCKER_USERNAME}" -p="${DOCKER_PASSWORD}"
-          if [[ -n "$CIRCLE_TAG" ]]; then 
-            docker push $IMAGE:$VERSION
-          fi
-          if [[ "$CIRCLE_BRANCH" == "master"  ]]; then
-            docker tag $IMAGE:$VERSION $IMAGE:latest
-            docker push $IMAGE:latest
-          fi
+          docker push $IMAGE:$DEV_TAG
         fi
 gke_test: &gke_test
   docker:
@@ -142,7 +142,7 @@ gke_test: &gke_test
         tar zxf helm-$HELM_VERSION-linux-amd64.tar.gz
         sudo mv linux-amd64/helm /usr/local/bin/
         helm init
-    - run: ./script/e2e-test.sh
+    - run: ./script/e2e-test.sh $DEV_TAG
     - run:
         name: Cleanup GKE Cluster
         command: gcloud container clusters delete --async --zone $GKE_ZONE $ESCAPED_GKE_CLUSTER
@@ -219,3 +219,16 @@ jobs:
             ssh-add ~/.ssh/id_rsa_* ~/.ssh/id_rsa
           fi
           ./script/chart_sync.sh kubernetes-bitnami kubernetes@bitnami.com
+  push_images:
+    machine: true
+    steps:
+      - <<: *exports
+      - run: |
+          if [[ -z "$CIRCLE_PULL_REQUEST" && -n "$DOCKER_USERNAME" && -n "$DOCKER_PASSWORD" ]]; then
+            docker login -u="${DOCKER_USERNAME}" -p="${DOCKER_PASSWORD}"
+            for IMAGE in kubeapps/chart-repo kubeapps/chartsvc kubeapps/apprepository-controller kubeapps/dashboard kubeapps/tiller-proxy; do
+              docker pull $IMAGE:$DEV_TAG
+              docker tag $IMAGE:$DEV_TAG $IMAGE:$PROD_TAG
+              docker push $IMAGE:$PROD_TAG
+            done
+          fi
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,7 +71,7 @@ workflows:
             - GKE_1_9
             - GKE_1_10
       - push_images:
-          <<: *build_on_master
+          <<: *build_always
           requires:
             - GKE_1_9
             - GKE_1_10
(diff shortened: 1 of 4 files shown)


script/libtest.sh
```diff
@@ -45,11 +45,20 @@ k8s_wait_for_pod_completed() {
...
+    echo "Checking image for $deployment"
+    kubectl get deployment -n $namespace $deployment -o jsonpath="$jsonpath" | grep $expectedPattern
+}
+
 ## helm specific helper functions
 wait_for_tiller() {
     echo "Waiting for Tiller to be ready ... "
     local -i cnt=${TEST_MAX_WAIT_SEC:?}
-    until helm version "${@}"; do
+    until helm version --tiller-connection-timeout 5 "${@}"; do
```
andresmgot: This is just something I found to speed the e2e tests. By default `helm version` waits for 5 minutes if tiller is not yet ready.

.circleci/config.yml
```diff
@@ -89,7 +94,8 @@ exports: &exports
   run: |
     # It is not possible to resolve env vars in the environment section:
     # https://circleci.com/docs/2.0/env-vars/#using-bash_env-to-set-environment-variables
-    echo "export VERSION=${CIRCLE_TAG:-build-$CIRCLE_WORKFLOW_ID}" >> $BASH_ENV
+    echo "export DEV_TAG=build-${CIRCLE_SHA1}" >> $BASH_ENV
+    echo "export PROD_TAG=${CIRCLE_TAG:-latest}" >> $BASH_ENV
```
migmartri: If you could add a comment on the top of these two lines saying that these are the tags for the resulting docker images it would be great.

.circleci/config.yml
```diff
@@ -98,17 +104,11 @@ build_image: &build_image
...
-    - run: make VERSION="$VERSION" $IMAGE
+    - run: make VERSION="$DEV_TAG" $IMAGE
     - run: |
-        if [[ -z "$CIRCLE_PULL_REQUEST" ]]; then
+        if [[ -z "$CIRCLE_PULL_REQUEST" && -n "$DOCKER_USERNAME" && -n "$DOCKER_PASSWORD" ]]; then
           docker login -u="${DOCKER_USERNAME}" -p="${DOCKER_PASSWORD}"
-          if [[ -n "$CIRCLE_TAG" ]]; then 
-            docker push $IMAGE:$VERSION
-          fi
-          if [[ "$CIRCLE_BRANCH" == "master"  ]]; then
```
migmartri: Great :)

.circleci/config.yml
```diff
@@ -98,17 +104,11 @@ build_image: &build_image
   steps:
     - checkout
     - <<: *exports
-    - run: make VERSION="$VERSION" $IMAGE
+    - run: make VERSION="$DEV_TAG" $IMAGE
     - run: |
-        if [[ -z "$CIRCLE_PULL_REQUEST" ]]; then
+        if [[ -z "$CIRCLE_PULL_REQUEST" && -n "$DOCKER_USERNAME" && -n "$DOCKER_PASSWORD" ]]; then
```
migmartri: Why of the removal of the variable expansion {}?

BTW, a bash style guide that I try without much success to follow is this one https://google.github.io/styleguide/shell.xml#Variable_expansion

.circleci/config.yml
```diff
@@ -219,3 +219,16 @@ jobs:
             ssh-add ~/.ssh/id_rsa_* ~/.ssh/id_rsa
           fi
           ./script/chart_sync.sh kubernetes-bitnami kubernetes@bitnami.com
+  push_images:
+    machine: true
```
migmartri: Out of curiosity, why do we need to set machine: true here? Is it so we have docker installed?

Also, can this step be skipped all together even before being provisioned (if it is not being done already)

What I mean is that this step should be only provisioned and run if -z "$CIRCLE_PULL_REQUEST"

script/e2e-test.sh
```diff
@@ -29,9 +30,19 @@ k8s_wait_for_pod_ready kube-system app=helm,name=tiller
 wait_for_tiller
 
 # Install Kubeapps
-# TODO: Use just built images
```
migmartri: :1st_place_medal:

script/libtest.sh
```diff
@@ -45,11 +45,20 @@ k8s_wait_for_pod_completed() {
     k8s_wait_for_pod $namespace $labelSelector Completed
 }
 
+k8s_ensure_image() {
+    namespace=${1:?}
+    deployment=${2:?}
+    expectedPattern=${3:?}
+    jsonpath=${4:-'{.spec.template.spec.containers[0].image}'}
+    echo "Checking image for $deployment"
```
migmartri: Could we make this more descriptive.

Something like, "checking that deployment %s is using image tagged with %s"

It would be ideal to get a message when it is not the case as well but anyways, not important.

.circleci/config.yml
```diff
@@ -219,3 +219,16 @@ jobs:
...
+    steps:
+      - <<: *exports
+      - run: |
+          if [[ -z "$CIRCLE_PULL_REQUEST" && -n "$DOCKER_USERNAME" && -n "$DOCKER_PASSWORD" ]]; then
+            docker login -u="${DOCKER_USERNAME}" -p="${DOCKER_PASSWORD}"
+            for IMAGE in kubeapps/chart-repo kubeapps/chartsvc kubeapps/apprepository-controller kubeapps/dashboard kubeapps/tiller-proxy; do
+              docker pull $IMAGE:$DEV_TAG
+              docker tag $IMAGE:$DEV_TAG $IMAGE:$PROD_TAG
```
migmartri: BTW, since these tags are public, we might want to `untag` the temporary build tags once the promotion is finished? Don't know if that's the right thing to do just because of noise. WDYT? cc/ @prydonius @andresmgot  

selection_676

script/libtest.sh
andresmgot: note that the `jsonpath` is an optional parameter, I thought that this can be reused for a different thing, not only the image (unless we are not probably going to use it with something different)

script/libtest.sh
andresmgot: I didn't remember that, thanks!

.circleci/config.yml
migmartri: > I'd rather not delete tags, they may be useful to reproduce a build

This can be achieved referencing the image via the @sha256 no? We could output it in the logs and in case you want to pull it exactly you just reference the sha.

> also it's a bit risky in cause something unexpected happens

What do you mean, anything in mind?

I think that probably we should not worry about this for now since that image is not useful by itself, although I'd rename the `build-abc` tag for something like `ci-build-xx`.

andresmgot: > This can be achieved referencing the image via the @sha256 no? 

I don't know if deleting the tag would cause that the build is not reachable even using the sha256. I would expect that docker hub garbage collector removes it.

>> also it's a bit risky in cause something unexpected happens

> What do you mean, anything in mind?

Yes, I was thinking on deleting `latest` or a versioned tag by accident (although that would be improbable).

Let me use a different image name, something like `kubeapps/tiller-proxy-ci:build-ABC`. I think that's a neat solution that avoid polluting the production image with tags and also prevents from unexpected deletions.

andresmgot:
@prydonius @prydonius I applied your requested changes (build: https://circleci.com/workflow-run/c369a345-eefb-46d5-a5ee-bc2e42c3fa45) PTAL

.circleci/config.yml
```diff
@@ -89,26 +94,25 @@ exports: &exports
   run: |
     # It is not possible to resolve env vars in the environment section:
     # https://circleci.com/docs/2.0/env-vars/#using-bash_env-to-set-environment-variables
-    echo "export VERSION=${CIRCLE_TAG:-build-$CIRCLE_WORKFLOW_ID}" >> $BASH_ENV
+    # DEV_TAG and PROD_TAG are the tags used for the Kubeapps docker images
+    echo "export DEV_TAG=build-${CIRCLE_SHA1}" >> $BASH_ENV
+    echo "export PROD_TAG=${CIRCLE_TAG:-latest}" >> $BASH_ENV
+    # Apart than using a DEV_TAG we use a different image ID to avoid polluting the tag
```
prydonius: Apart than -> Apart from

Personally, I don't really care about polluting the tags, but I'll leave that for you to decide.

.circleci/config.yml
migmartri: > Let me use a different image name, something like kubeapps/tiller-proxy-ci:build-ABC. I think that's a neat solution that avoid polluting the production image with tags and also prevents from unexpected deletions.

Well, I am not sure if that's better since the image will appear in search results and so on. In any case I think that it does not really matter. Choose whatever solution you think it's best. Thanks!

.circleci/config.yml
andresmgot: I didn't thought about that but it's true that it helps to navigate through available tags. It also closer to official docker images (they don't pollute that with tags) and it's easy to maintain.

PR #585, opened 2018-09-05
add sass-lint to project

prydonius:
- adds sass-lint and sass-lint-auto-fix
- adds package.json scripts for linting sass and autofix
- updates husky precommit hooks to run linter and autofix
- updates existing scss files to match linter

### DIFF
--- a/dashboard/.sass-lint.yml
+++ b/dashboard/.sass-lint.yml
@@ -1,3 +1,4 @@
+# Based off sass-lint YAML sample: https://github.com/sasstools/sass-lint/blob/master/docs/sass-lint.yml
 # Linter Options
 options:
   # Don't merge default rules
--- a/dashboard/.sass-lint.yml
+++ b/dashboard/.sass-lint.yml
@@ -1,0 +1,43 @@
+# Linter Options
+options:
+  # Don't merge default rules
+  merge-default-rules: false
+  # Raise an error if more than 50 warnings are generated
+  max-warnings: 50
+# File Options
+files:
+  include: "src/**/*.s+(a|c)ss"
+  ignore:
+    - "node_modules/**/*.*"
+# Rule Configuration
+rules:
+  extends-before-mixins: 2
+  extends-before-declarations: 2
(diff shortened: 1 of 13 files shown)


dashboard/.sass-lint.yml
```diff
@@ -0,0 +1,43 @@
...
+    - 2
+    - size: 2
+  property-sort-order:
+    - 1
+    - order:
+        - display
+        - margin
+      ignore-custom-properties: true
+  variable-for-property:
+    - 1
```
prydonius: @Angelmmiguel this config is copied from the sass-lint sample (https://github.com/sasstools/sass-lint/blob/master/docs/sass-lint.yml), but I changed this from an error to a warning since we are using values for margin. What do you think about this config? Is there a preferred config we should be using? Should we disable this setting entirely?

dashboard/.sass-lint.yml
```diff
@@ -0,0 +1,43 @@
+# Linter Options
```
migmartri: Could you add a comment about from where you got this file from? Thanks!

prydonius:
Yes, I ran sass-lint-auto-fix and it automatically fixed those issues. The only manual change I made was disabling the no-important rule for that one instance.

> Could it be possible that we are making code not linter approved because we modify it after the lint check has passed?

Hmm, potentially, but prettier cares more about styling (e.g. it prefers new lines for curly brackets) rather than semantic issues like how much to nest rules like sass-lint does. Though, there is some overlap (see https://prettier.io/docs/en/comparison.html), but we would notice as the CI `yarn run lint` step would fail if prettier was conflicting with some rule.

dashboard/.sass-lint.yml
Angelmmiguel: I think it's fine for now @prydonius. We don't have any official scss-lint config, so let's use this. I'll create a PR when we define the config in HEx :)

prydonius:
Thanks @Angelmmiguel @migmartri!

Issue #577, opened 2018-08-31
Error log in pod kubeapps-apprepository-controller


obeyler:
kubernetes 1.11.2

Hello,
Do you know what can be this trouble ? 

E0831 16:04:40.671935       1 reflector.go:205] github.com/kubeapps/kubeapps/cmd/apprepository-controller/pkg/client/informers/externalversions/factory.go:74: Failed to list *v1alpha1.AppRepository: the server could not find the requested resource (get apprepositories.kubeapps.com)

UI doesn't show any cartrige anymore and "only loading" is present on home page
![image](https://user-images.githubusercontent.com/23224162/44936695-0c5d7e00-ad76-11e8-9113-e9e54832eb9b.png)

prydonius:
@obeyler are you able to get apprepositories using the CLI (`kubectl get apprepos`)? Also what is the output of `kubectl get crds`?

obeyler:
$ kubectl get all -n kubeapps
NAME                                                    READY     STATUS    RESTARTS   AGE
pod/apprepo-cleanup-bitnami-vdwtq-qg8ml                 1/1       Running   0          5h
pod/apprepo-cleanup-incubator-qr848-7bhns               1/1       Running   0          5h
pod/apprepo-cleanup-stable-vl8r4-wcdf8                  1/1       Running   0          5h
pod/apprepo-cleanup-svc-cat-d6z47-4fmfw                 1/1       Running   0          5h
pod/kubeapps-667545b66c-4hq95                           1/1       Running   0          5h
pod/kubeapps-apprepository-controller-f56d8c55d-j8jp8   1/1       Running   0          5h
pod/kubeapps-chartsvc-64f9879fb9-xhd8l                  1/1       Running   0          5h
pod/kubeapps-dashboard-579787f774-f9fgh                 1/1       Running   0          5h
pod/kubeapps-tiller-proxy-66bd48f7ff-xwlck              1/1       Running   0          5h

NAME                            TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)     AGE
service/kubeapps                ClusterIP   10.100.200.16    <none>        80/TCP      5h
service/kubeapps-chartsvc       ClusterIP   10.100.200.238   <none>        8080/TCP    5h
service/kubeapps-dashboard      ClusterIP   10.100.200.236   <none>        8080/TCP    5h
service/kubeapps-mongodb        ClusterIP   10.100.200.74    <none>        27017/TCP   5h
[...]

PR #586, opened 2018-09-05
Update mongodb chart dependency

andresmgot:
Ref: #577


chart/kubeapps/Chart.yaml
```diff
@@ -1,19 +1,19 @@
...
+version: 0.3.3
 appVersion: DEVEL
 description: Kubeapps is a dashboard for your Kubernetes cluster that makes it easy to deploy and manage applications in your cluster using Helm
 icon: https://raw.githubusercontent.com/kubeapps/kubeapps/master/docs/img/logo.png
 keywords:
-- helm
-- dashboard
-- service catalog
-- deployment
+  - helm
```
andresmgot: this new indentation was automatically generated by the vscode editor

prydonius: It doesn't do that for me, do you have some sort of YAML lint extension or something? The rest of the chart (e.g. values.yaml) doesn't indent lists twice so we should keep it consistent with that for now.

chart/kubeapps/Chart.yaml
```diff
@@ -1,19 +1,19 @@
 apiVersion: v1
 name: kubeapps
-version: 0.3.2
+version: 0.3.3
```
prydonius: should it be a minor version bump since MongoDB is a minor version bump?

chart/kubeapps/Chart.yaml
andresmgot: I don't, it's the normal `editor.formatOnSave` feature. I will disable it 😞

chart/kubeapps/Chart.yaml
andresmgot: I don't know the criteria in charts but for example with npm modules if a dependency gets updated it doesn't affect to the main version. Following the semver notation this is a "bug fix" so it should be a patch version bump.

PR #587, opened 2018-09-06
[Chart] frontend service annotation

migmartri:
It sdds support for metadata annotations to the frontend service.

This is useful if the user opts for exposing kubeapps via a LoadBalancer but want to tweak it by making it for example internal `'service.beta.kubernetes.io/aws-load-balancer-internal: 0.0.0.0/0'`

I have not modified the Readme because in my opinion, this is a power user feature, disabled by default, and not commonly required.

### DIFF
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -92,6 +92,7 @@ frontend:
   service:
     port: 80
     type: ClusterIP
+    # annotations: {}
   livenessProbe:
     httpGet:
       path: /healthz
--- a/chart/kubeapps/Chart.yaml
+++ b/chart/kubeapps/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kubeapps
-version: 0.3.2
+version: 0.3.3
 appVersion: DEVEL
 description: Kubeapps is a dashboard for your Kubernetes cluster that makes it easy to deploy and manage applications in your cluster using Helm
 icon: https://raw.githubusercontent.com/kubeapps/kubeapps/master/docs/img/logo.png
--- a/chart/kubeapps/templates/kubeapps-frontend-service.yaml
+++ b/chart/kubeapps/templates/kubeapps-frontend-service.yaml
@@ -7,6 +7,10 @@ metadata:
     chart: {{ template "kubeapps.chart" . }}
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
+{{- with .Values.frontend.service.annotations }}
+  annotations:
+{{ toYaml . | indent 4 }}
+{{- end }}
 spec:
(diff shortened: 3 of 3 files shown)


chart/kubeapps/values.yaml
```diff
@@ -92,6 +92,7 @@ frontend:
   service:
     port: 80
     type: ClusterIP
+    # annotations: {}
```
arapulido: Why commenting it out? if it is empty, it doesn't need to be commented out. And the rest of optional values (like tolerations) are not commented out. Is this a special case?

migmartri: Thanks for catching the inconsistency, no, it is not a special case, I think I just left it like that by mistake because I was testing the differences between `range` used upstream https://github.com/helm/charts/blob/master/stable/wordpress/templates/ingress.yaml#L16 and `with values` that we use in this chart. 

My bad, sorry.

migmartri: Actually, it seems that it is consistent with the way we use annotations in the ingress. 

https://github.com/kubeapps/kubeapps/blob/1af290624f2fb6e0ca1d5972a665d2d389d69888/chart/kubeapps/values.yaml#L143-L145

migmartri: Actually commented out vs empty object in the values.yaml file does not seem to make any difference, so I am wondering why we commented it out in the ingress in the first place Maybe @prydonius could shed some light? 

With the current commented out value, no annotation added

```bash
helm template . -x templates/kubeapps-frontend-service.yaml 
---
# Source: kubeapps/templates/kubeapps-frontend-service.yaml
apiVersion: v1
kind: Service
metadata:
  name: RELEASE-NAME-kubeapps
  labels:
    app: kubeapps
    chart: kubeapps-0.4.0
    release: RELEASE-NAME
    heritage: Tiller
spec:
```
Using the empty object approach instead, no annotation added either

```bash
git diff
diff --git a/chart/kubeapps/values.yaml b/chart/kubeapps/values.yaml
index fa97086c..7b65f76c 100644
@@ -92,7 +92,7 @@ frontend:
   service:
     port: 80
     type: ClusterIP
-    # annotations: {}
+    annotations: {}
   livenessProbe:

 helm template . -x templates/kubeapps-frontend-service.yaml 
---
# Source: kubeapps/templates/kubeapps-frontend-service.yaml
apiVersion: v1
kind: Service
metadata:
  name: RELEASE-NAME-kubeapps
  labels:
    app: kubeapps
    chart: kubeapps-0.4.0
    release: RELEASE-NAME
    heritage: Tiller
spec:
```

migmartri: In case that we decide to change it, I went ahead and created a patch for it https://github.com/kubeapps/kubeapps/pull/599

Issue #566, opened 2018-08-30
Error deploying an app if `all namespaces` is selected


migmartri:
Once you select all namespaces in the UI and click on deploy an app. An error is raised because the namespace is invalid.

![selection_669](https://user-images.githubusercontent.com/24523/44881379-36268000-ac64-11e8-983a-3ac7e2b25715.png)

We need to figure out a way to get user input about in which namespace she wants to deploy the app.

prydonius:
Perhaps the simplest thing to do for now would be to switch the default namespace when going to a deployment view if namespace === "_all". @migmartri do you want to fix this this iteration, just wondering why you added the important-soon priority?

migmartri:
I think that we should fix it soon but it does not mean to be done this iteration. I am still confused with those priority labels. To me important-soon does not seem to convey this iteration and important-long-term feels like it pretty much never will be done :)

Will update the tag

migmartri:
I personally would like to be able to choose the namespace during deployment, which means that we would 1: Make choosing a namespace required, 2: When in a namespace, this value will be selected, 3: When you are in all-namespaces this value is blank and needs to be set.

That said, I think that the approach has the drawback that it can be confusing that you have one namespace selected at the top and another one in the form. So for now I think that It would be better to just do what @prydonius suggested.

PR #595, opened 2018-09-07
Chart deploy. Show an error message if namespace is not selected

migmartri:
**DEPRECATED**, see updated section below

It sets the deploy chart link to include the default namespace in the route instead of the _all placeholder. It also contains some refactoring to start storing some hardcoded defined namespaces into a shared module.

The meat itself is in here https://github.com/kubeapps/kubeapps/pull/595/files#diff-4

NOTE: I decided to set modify the link instead of redirect within the NewApp container to not to add extra code to handle this behavior. At the end of the day it seems that today we only access to deploy app from that button, if that changes we can revisit it.

Also, I did not add a regexp in the route `/apps/new` that excludes `/ns/_all` because it seems that so far we are being permissive in that route so I thought that it was not worth it.

Fixes https://github.com/kubeapps/kubeapps/issues/566

**UPDATE**

After a round of feedback, the implementation of this feature has changed from doing a redirect when the user clicked deploy to show an error message during that edge case and not to make assumptions about where to redirect the user. More info here https://github.com/kubeapps/kubeapps/pull/595#issuecomment-419499734

selection_273

It also contains:

* Raise of UnprocessableEntity if the namespace is not valid. This matches usual API responses during validation errors.
* Refactor of minor namespace definitions
* Extension of the `UnexpectedErrorPage` to support an optional title since "Something went wrong" did not match a 422 error in my opinion.
* I do not like that I am using the "raw" mode which shows a terminal but it is the way the component is bundled and will require more work to untangle it. I think it's not worth it since we might eventually move to another notification method.

### DIFF
--- a/dashboard/src/actions/apps.test.tsx
+++ b/dashboard/src/actions/apps.test.tsx
@@ -4,7 +4,7 @@ import { getType } from "typesafe-actions";
 
 import actions from ".";
 import { App } from "../shared/App";
-import { IAppState } from "../shared/types";
+import { IAppState, UnprocessableEntity } from "../shared/types";
 
 const mockStore = configureMockStore([thunk]);
 
@@ -19,6 +19,9 @@ beforeEach(() => {
   store = mockStore({
     apps: {
       state,
+    },
+    config: {
+      namespace: "kubeapps-ns",
     },
   });
 });
@@ -82,5 +85,36 @@ describe("delete applications", () => {
     });
     expect(await store.dispatch(actions.apps.deleteApp("foo", "default", true))).toBe(false);
     expect(store.getActions()).toEqual(expectedActions);
+  });
+});
+
+describe("deploy chart", () => {
+  beforeEach(() => {
+    App.create = jest.fn();
+  });
+
+  it("returns true if namespace is correct and deployment is successful", async () => {
+    const res = await store.dispatch(
+      actions.apps.deployChart("my-version" as any, "my-release", "default"),
+    );
+    expect(res).toBe(true);
+    expect(App.create).toHaveBeenCalledWith(
+      "my-release",
+      "default",
+      "kubeapps-ns",
+      "my-version",
+      undefined,
+    );
+    expect(store.getActions().length).toBe(0);
+  });
+
+  it("returns false and dispatches UnprocessableEntity if the namespace is _all", async () => {
+    const res = await store.dispatch(
+      actions.apps.deployChart("my-version" as any, "my-release", "_all"),
+    );
+    expect(res).toBe(false);
+    expect(store.getActions().length).toBe(1);
+    expect(store.getActions()[0].type).toEqual(getType(actions.apps.errorApps));
+    expect(store.getActions()[0].err.constructor).toBe(UnprocessableEntity);
   });
 });
--- a/dashboard/src/actions/apps.test.tsx
+++ b/dashboard/src/actions/apps.test.tsx
@@ -4,6 +4,7 @@ import { getType } from "typesafe-actions";
 
 import actions from ".";
 import { App } from "../shared/App";
+import { definedNamespaces } from "../shared/Namespace";
 import { IAppState, UnprocessableEntity } from "../shared/types";
 
 const mockStore = configureMockStore([thunk]);
@@ -110,7 +111,7 @@ describe("deploy chart", () => {
 
   it("returns false and dispatches UnprocessableEntity if the namespace is _all", async () => {
     const res = await store.dispatch(
-      actions.apps.deployChart("my-version" as any, "my-release", "_all"),
+      actions.apps.deployChart("my-version" as any, "my-release", definedNamespaces.all),
     );
     expect(res).toBe(false);
     expect(store.getActions().length).toBe(1);
--- a/dashboard/src/actions/apps.ts
+++ b/dashboard/src/actions/apps.ts
@@ -97,9 +97,10 @@ export function deployChart(
 ) {
   return async (dispatch: Dispatch<IStoreState>, getState: () => IStoreState): Promise<boolean> => {
     try {
+      // You can not deploy applications unless the namespace is set
       if (namespace === definedNamespaces.all) {
         throw new UnprocessableEntity(
-          "Namespace not selected. You need to select a namespace using the selector on the top right corner.",
+          "Namespace not selected. Please select a namespace using the selector on the top right corner.",
         );
       }
 
--- a/dashboard/src/actions/apps.ts
+++ b/dashboard/src/actions/apps.ts
@@ -1,8 +1,8 @@
 import { Dispatch } from "redux";
 import { createAction, getReturnOfExpression } from "typesafe-actions";
-
 import { App } from "../shared/App";
 import { hapi } from "../shared/hapi/release";
+import { definedNamespaces } from "../shared/Namespace";
 import { IAppOverview, IChartVersion, IStoreState } from "../shared/types";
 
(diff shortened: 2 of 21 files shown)


andresmgot:
I am probably late for this but maybe we are assuming too much if we say that the user wants to deploy in the `default` namespace when "All Namespaces" is selected. What if the `default` namespace doesn't exist or the user is only allowed to write in a specific namespace?

I would just throw an error if `this.props.namespace === definedNamespaces.all` in the DeploymentForm view clearly specifying that a specific namespace should be selected in the box above. I think that is needed even with the current solution since you are able to reselect "All Namespaces" in the deployment form and you will end with the same error than in the description of the issue.

dashboard/src/shared/Namespace.ts
```diff
@@ -11,3 +11,9 @@ export default class Namespace {
   private static APIBase: string = "/api/kube";
   private static APIEndpoint: string = `${Namespace.APIBase}/api/v1/namespaces`;
 }
+
+// Set of namespaces used accross the applications as default and "all ns" placeholders
+export const definedNamespaces = {
```
andresmgot: thanks for these definitions!

arapulido:
I am wondering if the problem is that the namespace dropdown shouldn't be a common element. It doesn't make sense in most views.

I think we should have it inside the Apps view, for that particular view only.

And, once you want to deploy an app, you get a dropdown at the same level as "Name" and "Version".

migmartri:
Thank you both!

> I am probably late for this but maybe we are assuming too much if we say that the user wants to deploy in the default namespace when "All Namespaces" is selected. What if the default namespace doesn't exist or the user is only allowed to write in a specific namespace?

No you are not late, just in time :)

This is something that we briefly discussed, but in the end it seems that we are already making this assumption (that the default namespace is there) for example when you go to /apps (without namespace) so this seemed like a simple extension of that possible flawed decision.

> I would just throw an error if this.props.namespace === definedNamespaces.all in the DeploymentForm view clearly specifying that a specific namespace should be selected in the box above. I think that is needed even with the current solution since you are able to reselect "All Namespaces" in the deployment form and you will end with the same error than in the description of the issue.

That's a good point @andresmgot. Ideally I was thinking https://github.com/kubeapps/kubeapps/issues/566#issuecomment-419252647 of showing the namespace selector when deploying the app (I think similarly to what @arapulido suggested), since it is not even clear where it is going to be deployed but I discarded that solution for now because as @arapulido hinted this might mean that the current allNamespaces option should not be there in the first place. I have created another issue to tackle it and give it some thoughts https://github.com/kubeapps/kubeapps/issues/598

That's said, @andresmgot suggestion seems to align with a more future proof and advanced feature that includes this namespace selector in the form, removes the `default` redirect assumption and prevents the user to run into the error if re-selects the namespace. So I am going to pivot this patch towards that solution. Thanks!

dashboard/src/actions/apps.test.tsx
```diff
@@ -84,3 +87,34 @@ describe("delete applications", () => {
...
+      "kubeapps-ns",
+      "my-version",
+      undefined,
+    );
+    expect(store.getActions().length).toBe(0);
+  });
+
+  it("returns false and dispatches UnprocessableEntity if the namespace is _all", async () => {
+    const res = await store.dispatch(
+      actions.apps.deployChart("my-version" as any, "my-release", "_all"),
```
prydonius: couldn't this use `definedNamespaces.all`?

dashboard/src/actions/apps.ts
```diff
@@ -97,6 +97,13 @@ export function deployChart(
 ) {
   return async (dispatch: Dispatch<IStoreState>, getState: () => IStoreState): Promise<boolean> => {
     try {
+      // You can not deploy applications unless the namespace is set
+      if (namespace === definedNamespaces.all) {
+        throw new UnprocessableEntity(
+          "Namespace not selected. Please select a namespace using the selector on the top right corner.",
```
prydonius: on the top right corner -> in the top right corner

dashboard/src/components/ChartView/ChartDeployButton.test.tsx
```diff
@@ -28,17 +28,30 @@ it("renders a button to deploy the chart version", () => {
   expect(button.text()).toBe("Deploy using Helm");
 });
 
-it("renders a redirect when the button is clicked", () => {
-  const wrapper = shallow(<ChartDeployButton version={testChartVersion} namespace="test" />);
-  const button = wrapper.find("button");
-  expect(button.exists()).toBe(true);
-  expect(wrapper.find(Redirect).exists()).toBe(false);
+it("renders a redirect with the correct URL when the button is clicked", () => {
```
prydonius: maybe we can revert this change since we don't have the special case for the _all namespace?

dashboard/src/components/DeploymentForm/DeploymentErrors.tsx
```diff
@@ -55,7 +55,13 @@ class DeploymentErrors extends React.Component<IDeploymentErrorProps> {
           <NotFoundErrorAlert resource={`Application "${releaseName}"`} namespace={namespace} />
         );
       case UnprocessableEntity:
-        return <UnexpectedErrorAlert text={error && error.message} raw={true} />;
+        return (
+          <UnexpectedErrorAlert
+            text={error && error.message}
+            raw={true}
+            title="Invalid deployment"
```
prydonius: I'm concerned about calling this `deployment` because it could be confused with a kubernetes deployment kind. Maybe we should say `Invalid installation` or `No namespace selected`.

dashboard/src/components/ErrorAlert/UnexpectedErrorAlert.tsx
```diff
@@ -45,9 +49,12 @@ class UnexpectedErrorPage extends React.Component<IUnexpectedErrorPage> {
         message = <p>{this.props.text}</p>;
       }
     }
+    // NOTE(miguel) We are using the non-undefined "!" token in `props.title` because our current version of
```
prydonius: thanks for the note

dashboard/src/components/ChartView/ChartDeployButton.test.tsx
andresmgot: I think the test is fine. It's not related to the issue but since it's already written it's good to have it.

dashboard/src/components/DeploymentForm/DeploymentErrors.tsx
andresmgot: That's a good point. `Failed installation` seems a good alternative to me.

migmartri: Good point, will change it.

dashboard/src/components/ChartView/ChartDeployButton.test.tsx
migmartri: I actually like it because now it makes clear that the route depends on the namespace and the chart version.

dashboard/src/components/DeploymentForm/DeploymentErrors.tsx
migmartri: > Maybe we should say Invalid installation or No namespace selected

Couple of issues here. `Invalid installation` might not be accurate because this deploymentErrors component is shared also during upgrade. 

`No namespace selected` is too specific when this component is general purpose. AFAIK there is not an easy way to customize this title since we dispatch an error object.

In any case I think that Failed installation is good enough and better than adding the deployment confusion. Thanks!

PR #601, opened 2018-09-07
add stylelint

prydonius:
As per the discussion in Slack (https://kubernetes.slack.com/archives/C9D3TSUG4/p1536314108000100), this PR switches to stylelint for better compatibility with prettier.

- replaces sass-lint
- adds stylelint-config-prettier to ensure we don't conflict with
 prettier rules


dashboard/.stylelintrc.json
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": ["stylelint-config-recommended-scss", "stylelint-config-prettier"]
```
prydonius: stylelint-config-recommended-scss is what stylelint recommends to use when using SCSS + prettier. However, this only enables the "possible errors" rules listed in https://github.com/stylelint/stylelint/blob/master/docs/user-guide/rules.md#possible-errors and none of the language feature rules. These are pretty basic rules, so I'm not sure if we'd want some of the more specific rules?

dashboard/package.json
```diff
@@ -56,13 +56,13 @@
...
-    "lint-css": "sass-lint -c ./.sass-lint.yml -v",
-    "lint-css-fix": "sass-lint-auto-fix -c ./.sass-lint.yml",
+    "lint-css": "stylelint --config ./.stylelintrc.json 'src/**/*.scss'",
+    "lint-css-fix": "stylelint --config ./.stylelintrc.json --fix 'src/**/*.scss'",
     "lint": "npm-run-all lint-js lint-css"
   },
   "lint-staged": {
     "*.{ts,tsx}": ["yarn run lint-js-fix", "prettier --write", "git add"],
-    "*.scss": ["yarn run lint-css-fix", "yarn run lint-css", "prettier --write", "git add"],
+    "*.scss": ["yarn run lint-css-fix", "prettier --write", "git add"],
```
migmartri: so lint-css is not needed anymore? I thought that we wanted to add it as a guard before git add?

prydonius: sass-lint didn't have autofix so there was a separate tool for that, stylelint has it built in and will do the lint and fixing in one go, similar to tslint.

r.e. running prettier before linting, I think we should do that in a separate PR, and I'm still not 100% sure we should do that. I'd like to understand more about why it was in this order in the first place.

PR #603, opened 2018-09-08
rename kubeapps-dashboard svc to sound more internal

### LINKED ISSUE #602: Rename kubeapps-dashboard Service to avoid confusion over what Service to use
A lot of users (#382, https://github.com/kubeapps/kubeapps/issues/577#issuecomment-419392268) mistake the `kubeapps-dashboard` Service as the one to set as a LoadBalancer or proxy to using an Ingress resource, when in fact the `kubeapps` Service should be used. It doesn't help that the default Kubernetes Dashboard manifest calls their Service `kubernetes-dashboard`.

We should rename this Service to something that sounds like less of an entrypoint, for example:

- kubeapps-react
- kubeapps-static
- kubeapps-internal

This should help avoid users mistaking the correct Service in the future.

prydonius:
In order to avoid confusion around the `kubeapps-dashboard` and
`kubeapps` Service, we should rename the `kubeapps-dashboard` Service
to sound like more of an internal component rather than a possible
entrypoint to the dashboard.

fixes #602

### DIFF
--- a/chart/kubeapps/Chart.yaml
+++ b/chart/kubeapps/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kubeapps
-version: 0.4.1
+version: 0.4.2
 appVersion: DEVEL
 description: Kubeapps is a dashboard for your Kubernetes cluster that makes it easy to deploy and manage applications in your cluster using Helm
 icon: https://raw.githubusercontent.com/kubeapps/kubeapps/master/docs/img/logo.png
--- a/chart/kubeapps/templates/_helpers.tpl
+++ b/chart/kubeapps/templates/_helpers.tpl
@@ -51,42 +51,42 @@ Create chart name and version as used by the chart label.
 Create name for the apprepository-controller based on the fullname
 */}}
 {{- define "kubeapps.apprepository.fullname" -}}
-{{ template "kubeapps.fullname" . }}-apprepository-controller
+{{ template "kubeapps.fullname" . }}-internal-apprepository-controller
 {{- end -}}
 
 {{/*
 Create name for the apprepository bootstrap job
 */}}
 {{- define "kubeapps.apprepository-jobs-bootstrap.fullname" -}}
-{{ template "kubeapps.fullname" . }}-apprepository-jobs-bootstrap
+{{ template "kubeapps.fullname" . }}-internal-apprepository-jobs-bootstrap
 {{- end -}}
 
 {{/*
 Create name for the apprepository cleanup job
 */}}
 {{- define "kubeapps.apprepository-jobs-cleanup.fullname" -}}
-{{ template "kubeapps.fullname" . }}-apprepository-jobs-cleanup
+{{ template "kubeapps.fullname" . }}-internal-apprepository-jobs-cleanup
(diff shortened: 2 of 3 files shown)


chart/kubeapps/templates/_helpers.tpl
```diff
@@ -79,14 +79,14 @@ Create name for the chartsvc based on the fullname
 Create name for the dashboard based on the fullname
 */}}
 {{- define "kubeapps.dashboard.fullname" -}}
```
prydonius: the fact that the values and the templates are still called dashboard might be confusing, so we might want to change these too? What do you think?

chart/kubeapps/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kubeapps
-version: 0.4.0
+version: 0.4.1
```
arapulido: You will need to rebase before merging, as there is another pending update with that version number.

chart/kubeapps/templates/_helpers.tpl
```diff
@@ -79,14 +79,14 @@ Create name for the chartsvc based on the fullname
...
-{{ template "kubeapps.fullname" . }}-dashboard
+{{ template "kubeapps.fullname" . }}-react
 {{- end -}}
 
 {{/*
 Create name for the dashboard config based on the fullname
 */}}
 {{- define "kubeapps.dashboard-config.fullname" -}}
-{{ template "kubeapps.fullname" . }}-dashboard-config
+{{ template "kubeapps.fullname" . }}-react-config
```
arapulido: mmm, not a fan of calling it with just the technology. maybe `reactapp` ?

chart/kubeapps/templates/_helpers.tpl
andresmgot: If we rename it to `kubeapps-dashboard-internal` the templates would still make sense. I wouldn't call it `react` to not make it depend on the underlying technology.

chart/kubeapps/templates/_helpers.tpl
migmartri: Yeah, me neither, naming is hard though because I can not think in a good name for this.

prydonius:
Discussed with @migmartri and he suggested to prefix all the names with `internal` except the main proxy Service, which should hopefully give a better indication of which service to use.

I will also modify the values.yaml file to put more emphasis on some of the options (e.g. frontend) over the others.

prydonius:
@migmartri @andresmgot  I've made the described changes, PTAL!

screen shot 2018-09-11 at 14 04 42

screen shot 2018-09-11 at 14 05 43

prydonius:
I tested it by upgrading my existing installation, so upgrades seem to work.

migmartri:
> I would still write a short document or a section in the getting-started or dashboard documents to explain how to expose Kubeapps.

We already have how to port forward both in the readme and getting started, or you mean how to expose it via loadbalancer, ingress and so on?

prydonius:
I think for now, the configuration options should just be in the chart readme. Once we have a docs site, we can rethink this and have more pages for configuration like this.

Issue #546, opened 2018-08-24
Improve API error handling


migmartri:
For example, if tiller-proxy is down we get stuck in a "loading" state.

We should show a message and maybe encourage people to open an issue in GH.

![selection_661](https://user-images.githubusercontent.com/24523/44600298-b729d700-a78d-11e8-8629-8e5d4074fd63.png)

prydonius:
Hmm, yeah the "Sorry! Something went wrong" message should be shown here. Seems like a regression?

migmartri:
FWIW. React 16 added the error boundary functionality that we might be able to use for generic not handled errors https://reactjs.org/docs/error-boundaries.html

prydonius:
@migmartri I tested that before and it only worked on uncaught errors in render, and won't catch errors that happen out of band (e.g. thunk, componentDidMount)

PR #611, opened 2018-09-11
Always show filter options

### LINKED ISSUE #593: Show deleted apps filter should be shown even if there are no apps
For example, in my case I do not have any non-deleted app but I do have deleted apps. The problem with the current approach is that I can not see those deleted apps until I have a non-deleted one.

We should always show the filter even if the filter says no apps.

selection_683

andresmgot:
Fixes #593

- Always show filter options even if the current list is empty
- Always show the `Deploy` button in the top-right corner.
- Remove `Deploy` button from welcome message to avoid duplicity

gif_2018-09-11_13 33 1536665629


prydonius:
Merging this so @migmartri can build on top of this for #546

PR #612, opened 2018-09-11
[chart] Edit NOTES to use more compatible command

andresmgot:
Fixes https://kubernetes.slack.com/archives/C9D3TSUG4/p1536328485000100

`kubectl port-forward` to a service is only available with `kubectl` +1.11. Use the previous approach to ensure that users won't see an error when trying those commands.

### DIFF
--- a/chart/kubeapps/templates/NOTES.txt
+++ b/chart/kubeapps/templates/NOTES.txt
@@ -41,7 +41,8 @@ To access Kubeapps from outside your K8s cluster, follow the steps below:
 {{- else if contains "ClusterIP" .Values.frontend.service.type }}
 
    echo "Kubeapps URL: http://127.0.0.1:8080"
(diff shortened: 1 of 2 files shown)


chart/kubeapps/templates/NOTES.txt
```diff
@@ -41,7 +41,8 @@ To access Kubeapps from outside your K8s cluster, follow the steps below:
 {{- else if contains "ClusterIP" .Values.frontend.service.type }}
 
    echo "Kubeapps URL: http://127.0.0.1:8080"
-   kubectl port-forward --namespace {{ .Release.Namespace }} svc/{{ template "kubeapps.fullname" . }} 8080:{{ .Values.frontend.service.port }}
+   export POD_NAME=$(kubectl get pods --namespace {{ .Release.Namespace }} -l "app={{ template "kubeapps.fullname" . }}" -o jsonpath="{.items[0].metadata.name}")
```
prydonius: `-o name` instead of the jsonpath, and tested working on kubectl v1.8.0

PR #615, opened 2018-09-11
API error handling in releases list

migmartri:
Looking into https://github.com/kubeapps/kubeapps/issues/546 I noticed that the main issue is that the releases page was not handling correctly errors from the API, regardless of being 503, 500. This is because our loading banner was being rendered as long as appOverview was empty.

This does some refactoring and tweaks in that page that:

### Loading state

* We show the "Applications" header so the viewport vertical space does not jump, this way the user always has the context.
* The deploy button nor the filtering options are shown.

selection_696

### Error state

* We show a message and like in the loading state we do not show the controls.

selection_697

I have tested it by making the tiller-proxy crash and even disabling it all together.

### Apps loaded state

* The view is shown as before, nothing has changed here.

selection_698

* Note that I have added a 5 seconds to Axios to give the user quicker feedback in case that something goes wrong. 5 seconds should be plenty of time for pretty much any operation we do, I think. 

Fixes https://github.com/kubeapps/kubeapps/issues/546

### DIFF
--- a/dashboard/src/components/AppList/AppList.test.tsx
+++ b/dashboard/src/components/AppList/AppList.test.tsx
@@ -7,6 +7,7 @@ import AppList from "./AppList";
 import AppListItem from "./AppListItem";
 
 let defaultProps = {} as any;
+let props = {} as any;
 
 beforeEach(() => {
   defaultProps = {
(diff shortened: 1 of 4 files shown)


dashboard/src/shared/Auth.ts
```diff
@@ -46,4 +46,8 @@ export class Auth {
   }
 }
 
-export const axios = Axios.create();
+const requestConfig = {
+  timeout: 5000, // After the timeout the request gets canceled (default no-timeout)
```
prydonius: Installing some larger charts might take longer with Tiller Proxy, we should check to make sure that won't be an issue. cc @andresmgot

dashboard/src/components/AppList/AppList.tsx
```diff
@@ -84,6 +72,26 @@ class AppList extends React.Component<IAppListProps, IAppListState> {
     );
   }
 
+  public appListControls() {
+    const { pushSearchFilter, apps: { listingAll } } = this.props;
+    return (
+      <React.Fragment>
```
prydonius: so happy about React.Fragment!

dashboard/src/components/AppList/AppList.test.tsx
```diff
@@ -19,70 +20,82 @@ beforeEach(() => {
...
+    const wrapper = shallow(<AppList {...props} />);
+    expect(wrapper.text()).toContain("Loading");
+  });
+
+  it("renders a Application header", () => {
+    const wrapper = shallow(<AppList {...props} />);
+    expect(wrapper.find("h1").text()).toContain("Applications");
+  });
+
+  it("does not show the search filter nor deploy button", () => {
```
andresmgot: do we want to ensure this? Maybe we should show those elements even if listOverview is not yet set

dashboard/src/components/AppList/AppList.test.tsx
```diff
@@ -19,70 +20,82 @@ beforeEach(() => {
...
+    const wrapper = shallow(<AppList {...props} />);
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it("renders a welcome message", () => {
+    const wrapper = shallow(<AppList {...props} />);
+    expect(wrapper.find("MessageAlertPage")).toExist();
+  });
+
+  it("shows the search filter, show deleted checkbox and deploy button", () => {
```
andresmgot: this test is the same one than `it("renders the 'Show deleted apps' button even if the app list is empty"`. Please remove that one to avoid duplicity.

dashboard/src/components/AppList/AppList.tsx
andresmgot: Didn't know about that, cool

dashboard/src/shared/Auth.ts
andresmgot: Yes, I think that 5s is too short. Imagine kubeapps that runs a job to cleanup stuff before deleting the elements. Tiller-proxy doesn't return a response until it receives the response from Tiller (and Tiller doesn't send the response until the deletion is completed). That takes longer than 5s. We should either set a bigger timeout (around even two minutes) or modify the default for Tiller Proxy.

migmartri: Thanks for the info.

2 minutes, which chart can take that long to deploy? Isn't k8s or Nginx going to terminate the connection before then?

prydonius: I'm not sure we've tested it, Kubeapps itself is an example of a chart with hooks that will take a bit longer to install. We should probably look into doing the chart installation more asynchronously.

migmartri: I see, I have noticed that our frontend Nginx gives us only 1 minute before time out, at least connection timeouts. That's why I am going to remove the timeout configuration from this patch and revisit it later https://github.com/kubeapps/kubeapps/issues/619

dashboard/src/components/AppList/AppList.test.tsx
migmartri: This represents the first time that you load the page, not when we search or change the visibility toggle, in which case the controls are actually shown.

I believe that I did this so if the first time it loads the API fails we do not get the controls. Maybe it might make more sense to not to show them if error instead.

migmartri:
@andresmgot based [on your comment](https://github.com/kubeapps/kubeapps/pull/615#discussion_r216935982) I was thinking about the code and realized that I was doing it just so if it fails we do not show actions that are not relevant. I was doing that by checking the listOverview because in our reducer a sideeffect of error is not to set listOverview. I think that it actually was a wrong approach.

So I have changed the code to show the controls and only hid them if there is an error, I think it is simpler to understand.

I have removed the timeout set because it seems that it has bigger implications and more research https://github.com/kubeapps/kubeapps/issues/619

I have also added some tests that I was missing about the app behavior if there was an error.

dashboard/src/components/AppList/AppList.test.tsx
andresmgot: did you miss this comment?

migmartri: They are the same test but in different context no?

One of them in `when fetched but not apps available` and the other one under `while fetching apps` (note the describe and beforeAll callbacks)

prydonius: yeah, it's a different scenario, I actually thought @andresmgot comment was for an earlier duplication that you had fixed

migmartri: Ahh cool, thanks for pointing it out.

PR #618, opened 2018-09-12
Send a 404 page by default

### LINKED ISSUE #597: Implement a 404 page
Accessing to invalid paths returns an empty blank page:

screen shot 2018-09-07 at 13 23 19

I think we should show a generic 404 page for those cases.

I found this, it may be helpful: https://tylermcginnis.com/react-router-handling-404-pages/

andresmgot:
Fixes #597 

[EDITED]

The goal of this page is to show  a 404 page when the user request a route that doesn't exist. For doing that:

 - We have implemented a 404 web page in the react app.
 ~~- We need to add the expected routes in NGINX to return a 404 error if the page doesn't exists.~~

Note: We have discarded the possibility of returning a 404 in the response headers to avoid having to maintain all the possible routes in the NGINX configuration.

screen shot 2018-09-12 at 18 37 08


chart/kubeapps/templates/dashboard-config.yaml
```diff
@@ -16,9 +16,71 @@ data:
       gzip on;
       gzip_static  on;
 
-      location / {
+      # /apps/ns/:namespace
+      location ~ "/apps/ns/[^/]*/?$" {
```
migmartri: Ok so this is so the server returns a proper 404 status code and not a 200 no?

Why is that that important? If this was an API I'd think differently, but for a SPA, the fact that it returns 200 does not seem a big deal as long as the user is presented with a not found page.

I think that the tradeoff of maintaining these routes in both places is too much to ask for that 404 improvement.

dashboard/src/containers/Routes.tsx
```diff
@@ -0,0 +1,70 @@
...
+        <Route exact={true} path="/login" component={LoginFormContainer} />
+        {Object.keys(Routes.exactRoutes).map(route => (
+          <PrivateRouteContainer
+            key={route}
+            exact={true}
+            path={route}
+            component={Routes.exactRoutes[route]}
+          />
+        ))}
+        <Route path="*" component={NotFound} />
```
migmartri: Do you need to pass the `path`. The documentation seems to ignore it https://reacttraining.com/react-router/web/example/no-match

Also, you mentioned that we could redirect to /404, why do we need to redirect? Isn't better that the route stays the one that you tried to access?

chart/kubeapps/templates/dashboard-config.yaml
andresmgot: okay, I agree

dashboard/src/containers/Routes.tsx
andresmgot: > Do you need to pass the path

I don't it was just to be explicit.

> Also, you mentioned that we could redirect to /404, why do we need to redirect? Isn't better that the route stays the one that you tried to access?

That was an alternative for returning a 404 in the status of the HTTP request and not having add all the routes I did. The idea was to configure only in nginx a route `/404` to return the error page instead of adding all the possible routes. For example, a request would:

 (request: `/foo`) -> nginx -> returns 200(OK) -> the react app redirects to `/404` -> nginx -> returns 404

In any case I think that is too confusing so I'd rather just return a `200` with the "not found" page content.

dashboard/src/containers/Routes.test.tsx
```diff
@@ -0,0 +1,15 @@
+import { mount } from "enzyme";
+import * as React from "react";
+import { MemoryRouter } from "react-router";
+import NotFound from "../components/NotFound";
+import Routes from "./Routes";
+
+it("invalid path should redirect to 404", () => {
```
migmartri: s/redirect to 404/show a 404 error

dashboard/src/containers/Routes.tsx
```diff
@@ -0,0 +1,69 @@
...
+        <Route exact={true} path="/login" component={LoginFormContainer} />
+        {Object.keys(Routes.exactRoutes).map(route => (
+          <PrivateRouteContainer
+            key={route}
+            exact={true}
+            path={route}
+            component={Routes.exactRoutes[route]}
+          />
+        ))}
+        <Route component={NotFound} />
```
migmartri: Maybe add a comment mentioning that this is a 404 catchAll?

chart/kubeapps/templates/dashboard-config.yaml
```diff
@@ -19,6 +19,11 @@ data:
       location / {
         try_files $uri /index.html;
       }
+
+      error_page 404 /index.html;
```
migmartri: Why is this needed?

dashboard/src/containers/Routes.test.tsx
```diff
@@ -0,0 +1,15 @@
...
+  const wrapper = mount(
+    <MemoryRouter initialEntries={["/random"]}>
```
migmartri: In other projects, for testing, I've been using staticRouter instead of MemoryRouter, I am curious what's your take on this and why you decided to choose this one?

dashboard/src/containers/Routes.test.tsx
```diff
@@ -0,0 +1,15 @@
...
+      <Routes namespace="default" />
+    </MemoryRouter>,
+  );
+  expect(wrapper.find(NotFound).exists()).toBe(true);
```
migmartri: `expect(wrapper.find(NotFound)).toExist();`

dashboard/src/components/NotFound/404.svg
```diff
@@ -0,0 +1,57 @@
+<?xml version="1.0" standalone="no"?>
```
prydonius: We should probably actually put this in https://github.com/kubeapps/kubeapps/tree/master/dashboard/src/img.

dashboard/src/components/NotFound/NotFound.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from "react";
+import logo404 from "./404.svg";
+
+class NotFound extends React.Component {
+  public render() {
+    return (
+      <div className="text-c align-center margin-t-huge">
+        <h3> The page you are looking for can't be found. </h3>
```
prydonius: remove whitespace between tags

dashboard/src/containers/Root.tsx
```diff
@@ -1,77 +1,31 @@
...
-              <Route exact={true} path="/login" component={LoginFormContainer} />
-              {Object.keys(Root.exactRoutes).map(route => (
-                <PrivateRouteContainer
-                  key={route}
-                  exact={true}
-                  path={route}
-                  component={Root.exactRoutes[route]}
-                />
-              ))}
+              <Routes namespace={this.getCurrentNamespace()} />
```
prydonius: We should connect the Routes component to the Redux store and have it get the namespace that way, rather than pass it in as a prop.

dashboard/src/containers/Routes.test.tsx
andresmgot: it was the one I found googling about the issue. It's true that we are not changing the URL in the test so staticRouter should be enough (MemoryRouter keeps track of the URLs change)

migmartri: Thanks for the explanation.

dashboard/src/containers/Routes.tsx
```diff
@@ -0,0 +1,68 @@
...
+            component={Routes.exactRoutes[route]}
+          />
+        ))}
+        {/* If the route doesn't match any expected path redirect to a 404 page  */}
+        <Route component={NotFound} />
+      </Switch>
+    );
+  }
+  public rootNamespacedRedirect = (props: any) => {
+    const { namespace } = store.getState();
```
prydonius: hmm we shouldn't have used `store` directly here, we should have connected to the Redux store and used the `mapStateToProps` function like in the other containers.

andresmgot: yes, I just moved code. What are the consequences of doing this? Is it just for standardization or not using `mapStateToProps` has an unexpected side effect?

prydonius: We're using the connect/mapStateToProps/mapDispatchToProps for all other containers to connect components so I believe we should not mix the two different approaches. The only reason we were using store directly for this before was because we were doing it within Root.tsx which was creating the store and passing it to Provider so it was not connected.

In addition, this approach is better because:

1. Not that important for this component, but getting state in through props allows you to use a components lifecycle hooks to automatically re-render, check for things when state changes, etc.
1. It makes it easier to test a component as the component doesn't need to understand the store and just takes input props

PR #621, opened 2018-09-12
add a helm template test in CI

prydonius:
this adds a helm template test to the CircleCI configuration to test
that the chart renders okay in PRs

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -74,6 +74,11 @@ install_gcloud_sdk: &install_gcloud_sdk
       rm -rf $HOME/google-cloud-sdk;
       curl https://sdk.cloud.google.com | bash;
     fi
+install_helm_cli: &install_helm_cli
+  run: |
+    wget https://storage.googleapis.com/kubernetes-helm/helm-$HELM_VERSION-linux-amd64.tar.gz
+    tar zxf helm-$HELM_VERSION-linux-amd64.tar.gz
+    sudo mv linux-amd64/helm /usr/local/bin/
 exports: &exports
   run: |
     # It is not possible to resolve env vars in the environment section:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -154,6 +156,20 @@ jobs:
       - run: yarn install --cwd=dashboard --frozen-lockfile
       - run: yarn --cwd=dashboard run lint
       - run: yarn --cwd=dashboard run test --maxWorkers=4 --coverage
+  test_chart:
+    docker:
+      - image: circleci/golang:1.9
```
migmartri: why golang? Isn't there anything more barebone than this?

.circleci/config.yml
```diff
@@ -154,6 +156,20 @@ jobs:
...
+      HELM_VERSION: v2.9.1
+    steps:
+      - checkout
+      - run: |
+          wget https://storage.googleapis.com/kubernetes-helm/helm-$HELM_VERSION-linux-amd64.tar.gz
+          tar zxf helm-$HELM_VERSION-linux-amd64.tar.gz
+          sudo mv linux-amd64/helm /usr/local/bin/
+          helm init --client-only
+      - run: helm dep build ./chart/kubeapps/
+      - run: helm template ./chart/kubeapps/
```
migmartri: Ideally I'd like to add some combinations related to the tls configuration.

All these should pass

```
helm template --set tillerProxy.tls.key="my-key" --set tillerProxy.tls.cert="my-cert"  .
helm template --set tillerProxy.tls.verify=false --set tillerProxy.tls.key="my-key" --set tillerProxy.tls.cert="my-cert"  .
 helm template --set tillerProxy.tls.verify=true --set tillerProxy.tls.key="my-key" --set tillerProxy.tls.cert="my-cert" --set tillerProxy.tls.ca="my-ca" .
```

Does it make sense? Should I add it in another patch?

.circleci/config.yml
```diff
@@ -154,6 +156,20 @@ jobs:
       - run: yarn install --cwd=dashboard --frozen-lockfile
       - run: yarn --cwd=dashboard run lint
       - run: yarn --cwd=dashboard run test --maxWorkers=4 --coverage
+  test_chart:
```
migmartri: maybe `test_chart_rendering` to avoid confusion with the actual e2e?

.circleci/config.yml
prydonius: @migmartri yeah I was looking for one and couldn't find it: https://circleci.com/docs/2.0/circleci-images/

.circleci/config.yml
```diff
@@ -154,6 +156,20 @@ jobs:
...
+    docker:
+      - image: circleci/golang:1.9
+    environment:
+      HELM_VERSION: v2.9.1
+    steps:
+      - checkout
+      - run: |
```
andresmgot: I would move the installation of helm to a "definition" (without the `helm init` to avoid duplicity)

.circleci/config.yml
andresmgot: We can add several combinations but I would do that in a script to be easier to execute locally or migrate if needed. I don't think we need a different patch for that, we can do that here.

.circleci/config.yml
prydonius: :+1: yep was thinking about doing that

prydonius:
@migmartri @andresmgot PTAL!

PR #622, opened 2018-09-12
e2e tests: update deployment names with internal prefix

prydonius:
in #603 we added a new prefix to deployment names for internal services,
but the e2e script was not updated to reflect the new names.


migmartri:
Maybe we can try to use labels instead?

prydonius:
@migmartri I guess it can be run locally, but I haven't tried. @andresmgot?

andresmgot:
yes, this can be run locally but I would recommend to use a fresh minikube instance

migmartri:
@andresmgot do you have by any chance any notes about how to do it by any chance? If you don't that's fine I'll figure it out :)

andresmgot:
Sorry, I don't have. Honestly I haven't tried :) but to reproduce the CI environment it's just needed a fresh cluster with helm installed. Then run the script with `./script/e2e-test.sh "" latest`.

Issue #165, opened 2018-03-06
Create a loader component


prydonius:
_From @Angelmmiguel on February 20, 2018 16:3_

Currently, we don't show a loader when we're fetching data from the server. This makes the current experience odd because the elements as appearing without any feedback to the user

_Copied from original issue: kubeapps/dashboard#96_

arapulido:
@Angelmmiguel ping (see above comment)

Angelmmiguel:
Yes! @arapulido, please could you open a task in bitnami/hex and reference it from this one?

Thanks

migmartri:
These are the places where we have loading state which will need to be replaced by the new component.

- [ ] main container app load
- [ ] loading wrapper (maybe the one above?)
- [ ] applications listing
- [ ] application show
- [ ] application deployment
- [ ] application upgrade
- [ ] charts listing
- [ ] charts show
- [ ] chart readme
- [ ] select repo form

migmartri:
I believe that all the loaders have been put in place and whenever the native hex loader is upstream we just need to replace `<LoadingSpinner>` for that one in a single place.

PR #637, opened 2018-09-13
LoadingSpinner component and first applications

migmartri:
This is part of the effort listed here https://github.com/kubeapps/kubeapps/issues/165#issuecomment-420845084 which:

* Creates a LoaderSpinner component which is based on @Angelmmiguel recommendations and eventually will be replaced by Hex native component https://github.com/bitnami/hex/issues/211
* Extends the existing LoaderWrapper to use this component
* Applies this new loading spinner to the config loading as well as to the applications list.
* Adds some tests and creates a shared set of jest examples that we will use once we apply the changes to the rest of the pages.

See both the global configuration wrapper and the apps list loading in action here
peek 2018-09-13 15-18


dashboard/src/shared/specs/LoadingWrapperSharedSpecs.tsx
```diff
@@ -0,0 +1,17 @@
+import { shallow } from "enzyme";
+import * as React from "react";
+
+// Shared jest examples that checks that the provided component is rendering the Loading Wrapper
+export default (Component: any, props: any) => {
```
prydonius: nice!

dashboard/src/shared/specs/LoadingWrapperSharedSpecs.tsx
```diff
@@ -0,0 +1,17 @@
...
+  describe("loading spinner", () => {
+    it("renders", () => {
+      const wrapper = shallow(<Component {...props} />);
+      expect(wrapper.find("LoadingWrapper")).toExist();
```
prydonius: we should import the LoadingWrapper class and use the constant instead of a string

dashboard/src/components/LoaderSpinner/LoaderSpinner.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from "react";
+
+import "./LoaderSpinner.css";
+
+const LoaderSpinner: React.SFC<{}> = _ => {
```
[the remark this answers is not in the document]
migmartri: Sorry, brain fart, usually I am not sure where I am GOing :P

dashboard/src/components/LoadingSpinner/LoadingSpinner.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from "react";
+
+import "./LoadingSpinner.css";
+
+const LoaderSpinner: React.SFC<{}> = () => {
```
andresmgot: Nice find the `SFC` (apart from the ugly acronym :P )

dashboard/src/components/LoadingWrapper/LoadingWrapper.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+  });
+
+  it("matches the snapshot", () => {
+    const wrapper = renderComponent(props);
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it("renders a spinner", () => {
+    const wrapper = renderComponent(props);
+    expect(wrapper.find(LoadingSpinner)).toExist();
```
andresmgot: it is a matter of taste but I think it express more intention if you merge this test with the previous one:

```
it("renders a spinner", () => {
  const wrapper = renderComponent(props);
  expect(wrapper.find(LoadingSpinner)).toExist();
});
```

For me that means: "it should matches the snapshot but the important piece I want to ensure is that it contains a LoadingSpinner". In other case, the "matches the snapshot" is kind of meaningless: when is the snapshot wrong?

dashboard/src/components/AppList/AppList.test.tsx
```diff
@@ -25,14 +26,15 @@ describe("while fetching apps", () => {
     props = { ...defaultProps, apps: { isFetching: true } };
   });
 
-  it("matches the snapshot", () => {
-    const wrapper = shallow(<AppList {...props} />);
-    expect(wrapper).toMatchSnapshot();
+  afterAll(() => {
+    // Calling the shared specs here so they are run after all the beforeEach
+    // callbacks have been run which set the default props and prop
+    sharedSpecs.loadingSpecs(AppList, props);
```
andresmgot: I find this a bit difficult to understand:

 - What does `specs` mean here? It seems that it is some kind of meta-test that can be reusable.
 - Why using `afterAll` instead of a `describe` or a simple `it`?
 - Shouldn't you set the property `isFetching: true` so you ensure that the Loading component should appear regardless of how `props` is modified outside the scope of this test?

dashboard/src/components/LoadingSpinner/LoadingSpinner.tsx
migmartri: This can be replaced by `React.StatelessComponent` but I thought that Sevilla Futbol Cub deserved a mention :trollface:

dashboard/src/components/AppList/AppList.test.tsx
migmartri: That's a good point, I do not find it ideal either.

spec == test, and what this does is just to run a set of shared tests. In a similar way rspec shared_examples work. It seems that jest does not support it natively so I based it on https://medium.com/@walreyes/sharing-specs-in-jest-82864d4d5f9e (without adding the behaves_like)

> Why using afterAll instead of a describe or a simple it?

I need to add it inside an afterAll to have access to the properties set in the other callbacks. w.r.t `it` the linter (I think) was complaining about nested its. This can be solved by moving the props setting outside of the callback and as part of the describe. The only reason is there is because it guarantees that the vble has not been mutated,

> Shouldn't you set the property isFetching: true so you ensure that the Loading component should appear regardless of how props is modified outside the scope of this test?

I want to know more, isFetching is set to true no? https://github.com/kubeapps/kubeapps/pull/637/files/7d4aa8e593537f47c46d090c7e23dee35b0508f0#diff-bad52a69fd8cd98a31a6337371ae22b9R26

I have another patch that changes this code a bit so I'll prefer to land this one and handle this there. I might end up removing this shared example after all.

dashboard/src/components/LoadingWrapper/LoadingWrapper.test.tsx
andresmgot: yes but I think that is something difficult to review. I am not saying that having a general purpose snapshot is something bad I just want to have as many hints as possible to know if a change in the snapshot is something important or not. For example a change in a className vs an element disappearing. In any case, let's see how reviews go after we add the different snapshots.

migmartri: I agree @andresmgot, I do not have enough experience with snapshots, I think I have some ideal conception of how it could help us but not idea on how it will develop in practice. Let's see. I personally have mixed feelings about it, worrying that in the end they will get on the way.

PR #638, opened 2018-09-14
Fix components docs

arapulido:
I have updated the docs in the architecture folder that were already mentioning the installer and Kubeless. I have also removed the CLI documentation for developers, as it is no longer part of the code.


prydonius:
Thanks @arapulido, although should we just remove this article since we already have https://github.com/kubeapps/kubeapps/tree/master/docs/developer?

arapulido:
> Thanks @arapulido, although should we just remove this article since we already have https://github.com/kubeapps/kubeapps/tree/master/docs/developer?

I am happy either way, but we should improve that doc then to add a bit more of information about what components do

prydonius:
Shall we merge the descriptions from the overview doc into the developer README and remove the overview doc?

migmartri:
> Shall we merge the descriptions from the overview doc into the developer README and remove the overview doc?

That sounds good to me, assuming that the overview targets developers, which it seems it does?

prydonius:
@migmartri although the architecture might be useful to users, I do see it as more of developer documentation. Perhaps a docs site might make this more discoverable in the future.

migmartri:
What should we do with this? I am ok merging this as is and moving the info to the developer README later.

PR #616, opened 2018-09-11
Tiller TLS CA validation

migmartri:
As mentioned in the attached issue, the CA certificate is only useful if tls-verify is set. Before this patch, regardless of the value of tls-verify we were requiring it to be set.

This patch also validates that the CA is present if tls-verify is set.

Fixes https://github.com/kubeapps/kubeapps/issues/547

Examples:

### tls.verify = false, tls.ca not set

The secret does not contain the data.ca key and it is not enforced.

```bash
helm template --set tillerProxy.tls.verify=false --set tillerProxy.tls.key="my-key"  --set tillerProxy.tls.cert="my-cert" . | grep kubeapps/templates/tiller-proxy-secret.yaml -A 20
# Source: kubeapps/templates/tiller-proxy-secret.yaml
# The tls ca certificate is only required when tls.verify is set to true, we fail otherwise.apiVersion: v1
kind: Secret
metadata:
  name: RELEASE-NAME-kubeapps-tiller-proxy
  labels:
    app: RELEASE-NAME-kubeapps-tiller-proxy
    chart: kubeapps-0.4.1
    release: RELEASE-NAME
    heritage: Tiller
data:
  tls.crt: |-
    bXktY2VydA==
  tls.key: |-
    bXkta2V5
---
```

### tls.verify = true, tls.ca not set

The installation fails.

```bash
helm template --set tillerProxy.tls.verify=true --set tillerProxy.tls.key="my-key" --set tillerProxy.tls.cert="my-cert" .
Error: render error in "kubeapps/templates/tiller-proxy-secret.yaml": template: kubeapps/templates/tiller-proxy-secret.yaml:4:3: executing "kubeapps/templates/tiller-proxy-secret.yaml" at <fail "tillerProxy.tl...>: error calling fail: tillerProxy.tls.ca: A valid CA certificate needs to be provided if tls-verify is set to true.
```

### tls.verify = true, tls.ca set

The ca.crt key is set as expected.

```
 helm template --set tillerProxy.tls.verify=true --set tillerProxy.tls.key="my-key" --set tillerProxy.tls.cert="my-cert" --set tillerProxy.tls.ca="my-ca" . | grep kubeapps/templates/tiller-proxy-secret.yaml -A 20
# Source: kubeapps/templates/tiller-proxy-secret.yaml
# The tls ca certificate is only required when tls.verify is set to true, we fail otherwise.apiVersion: v1
kind: Secret
metadata:
  name: RELEASE-NAME-kubeapps-tiller-proxy
[...]

PR #642, opened 2018-09-14
[e2e test] Install helm with TLS

### LINKED ISSUE #633: Chart e2e test should leverage tillerProxy tls
Lately, we have had some issues with the modification of our Helm chart which leverages not default options like the Tiller proxy tls setup. https://github.com/kubeapps/kubeapps/pull/626 https://github.com/kubeapps/kubeapps/pull/616 https://github.com/kubeapps/kubeapps/pull/627

All these mistakes would have been detected avoiding a chart release (at least during the master promotion process) by running our e2e tests using tls enabled.

Note: There is an additional ongoing process to add template testing which would have detected some of the mistakes but not all https://github.com/kubeapps/kubeapps/pull/621

andresmgot:
Fixes #633

Generate a self-signed certificate in the e2e scenario following Helm docs and use it to install Tiller and Kubeapps.

Test in a GKE build: https://circleci.com/workflow-run/decf2ece-4d5f-46d9-ad80-b3045ed24be6


script/e2e-test.sh
```diff
@@ -17,28 +17,80 @@
...
-  --set chartsvc.image.tag=$DEV_TAG \
-  --set chartsvc.image.repository=kubeapps/chartsvc$IMG_MODIFIER \
-  --set dashboard.image.tag=$DEV_TAG \
-  --set dashboard.image.repository=kubeapps/dashboard$IMG_MODIFIER \
-  --set tillerProxy.image.tag=$DEV_TAG \
-  --set tillerProxy.image.repository=kubeapps/tiller-proxy$IMG_MODIFIER
+    `# Tiller TLS flags` \
+    ${HELM_TLS_FLAGS} \
+    `# cli TLS flags` \
+    --set tillerProxy.tls.ca="$(cat ca.cert.pem)" \
```
migmartri: I'd not add this one since it is not required unless tls.verify=true

script/e2e-test.sh
```diff
@@ -17,28 +17,80 @@
 set -e
 
 ROOT_DIR=`cd "$( dirname "${BASH_SOURCE[0]}" )/.." >/dev/null && pwd`
-IMG_MODIFIER=${1:?}
-DEV_TAG=${2:?}
+DEV_TAG=${1:?}
+IMG_MODIFIER=${2:-""}
+HELM_TLS_FLAGS="--tls --tls-ca-cert ca.cert.pem --tls-cert helm.cert.pem --tls-key helm.key.pem"
```
migmartri: maybe HELM_CLIENT_TLS_FLAGS?

Are you missing `--tls-verify`? Otherwise `tls-ca-cert` is not needed nor used.

script/e2e-test.sh
```diff
@@ -17,28 +17,80 @@
...
+## client key
+openssl genrsa -out ./helm.key.pem 4096
+
+openssl req -key tiller.key.pem -new -sha256 -out tiller.csr.pem -config tls_config
+openssl req -key helm.key.pem -new -sha256 -out helm.csr.pem -config tls_config
+
+openssl x509 -req -CA ca.cert.pem -CAkey ca.key.pem -CAcreateserial -in tiller.csr.pem -out tiller.cert.pem
+openssl x509 -req -CA ca.cert.pem -CAkey ca.key.pem -CAcreateserial -in helm.csr.pem -out helm.cert.pem
+
+# Start helm
```
migmartri: Maybe change the comment to "Install Tiller with TLS support"?

script/e2e-test.sh
```diff
@@ -17,28 +17,80 @@
...
+helm init \
+  --tiller-tls \
+  --tiller-tls-cert ./tiller.cert.pem \
+  --tiller-tls-key ./tiller.key.pem \
+  --tiller-tls-verify \
+  --tls-ca-cert ca.cert.pem
+
+# The flag --wait is not available when using TLS flags:
```
migmartri: Good find!

script/e2e-test.sh
migmartri: You are not setting tls.verify=true, I am ok if we do not do it but in that case tls.ca is not necessary https://github.com/kubeapps/kubeapps/pull/616

script/e2e-test.sh
```diff
@@ -17,28 +17,80 @@
...
+## server key
+openssl genrsa -out ./tiller.key.pem 4096
+## client key
+openssl genrsa -out ./helm.key.pem 4096
+
+openssl req -key tiller.key.pem -new -sha256 -out tiller.csr.pem -config tls_config
+openssl req -key helm.key.pem -new -sha256 -out helm.csr.pem -config tls_config
+
+openssl x509 -req -CA ca.cert.pem -CAkey ca.key.pem -CAcreateserial -in tiller.csr.pem -out tiller.cert.pem
+openssl x509 -req -CA ca.cert.pem -CAkey ca.key.pem -CAcreateserial -in helm.csr.pem -out helm.cert.pem
```
prydonius: what if we just generated a cert and put it in the repo or circleci env? Surely running all these commands increases the time for this test?

script/e2e-test.sh
```diff
@@ -17,28 +17,80 @@
...
-  --set apprepository.syncImage.repository=kubeapps/chart-repo$IMG_MODIFIER \
-  --set chartsvc.image.tag=$DEV_TAG \
-  --set chartsvc.image.repository=kubeapps/chartsvc$IMG_MODIFIER \
-  --set dashboard.image.tag=$DEV_TAG \
-  --set dashboard.image.repository=kubeapps/dashboard$IMG_MODIFIER \
-  --set tillerProxy.image.tag=$DEV_TAG \
-  --set tillerProxy.image.repository=kubeapps/tiller-proxy$IMG_MODIFIER
+    `# Tiller TLS flags` \
+    ${HELM_TLS_FLAGS} \
+    `# cli TLS flags` \
```
prydonius: I don't understand this comment, isn't this the tiller proxy tls flags not the cli TLS flags?

andresmgot: I wanted to say that the flags that tiller-proxy uses are the same than the CLI flags but yes, it may be confusing.

andresmgot:
@migmartri I replied in one of the inline comments but repeating it here for visibility:

I just followed the helm documentation for setting Tiller with TLS (because it's where we point people and what they are likely going to use): https://github.com/helm/helm/blob/master/docs/tiller_ssl.md

Having said that, it's true that the ca cert flag is not necessary so I am okay removing it.

script/test-certs/ca.cert.pem
```diff
@@ -0,0 +1,30 @@
+-----BEGIN CERTIFICATE-----
```
migmartri: is this file needed? If we are not configuring the client to verify we do not need to pass this file during helm init. I believe.

andresmgot: we are using `verify` for `helm init`:

```
  --tiller-tls-verify \
   --tls-ca-cert ${CERTS_DIR}/ca.cert.pem
```

I could remove that as well

migmartri:
> One concern I have here is that we are now no longer testing Kubeapps under non-TLS scenario. This might be an issue if there was a bug in tiller-proxy that caused it to only work with TLS configuration (e.g. if a bug made it so we are always passing a TLS flag when connecting to Tiller, even when TLS is disabled). I'm not sure how realistic that scenario would be, but it might be worth considering if we can test a non-TLS scenario in addition to this.

That's a good point. We could potentially deploy two tillers, and two charts, all in the same cluster for speed sake. Although if we decide to do so, this can be done in another PR in my opinion.

prydonius:
@migmartri agree that a separate non-TLS tiller in another namespace is a good idea, that way we can also test the chart's tillerHost parameter. I will create an issue for that.

prydonius:
filed #646

andresmgot:
Thanks for your comments, replied in that issue. I am merging this, let me know if there is something else to handle

script/test-certs/ca.cert.pem
migmartri: Well, the issue is that it can be confusing since it is not actually being used.

PR #644, opened 2018-09-14
Loader components on applications and charts

migmartri:
Closes https://github.com/kubeapps/kubeapps/issues/165

Adds loader components to the app view, deployment and upgrade components, as well as enhances the shared test scenarios.

NOTE: I have decided to go with <LoadingWrapper> without children because otherwise it will require quite a refactor in some cases like in the upgrade component. 

peek 2018-09-14 15-54


dashboard/src/components/AppList/AppList.test.tsx
```diff
@@ -1,36 +1,28 @@
...
+  namespace: "default",
+  pushSearchFilter: jest.fn(),
+  toggleListAll: jest.fn(),
+};
 
-describe("while fetching apps", () => {
-  beforeEach(() => {
-    props = { ...defaultProps, apps: { isFetching: true } };
-  });
+context("while fetching apps", () => {
```
prydonius: what's the difference between context and describe?

dashboard/src/shared/specs/index.tsx
```diff
@@ -1,5 +1,4 @@
-import loadingSpecs from "./LoadingWrapperSharedSpecs";
-
-export default {
-  loadingSpecs,
+export default (sharedExampleName: string, args: any) => {
```
prydonius: 🆒

dashboard/src/components/AppList/AppList.test.tsx
migmartri: It is just syntax sugar. I like using describe to mention the test subject and context to set scenarios.

It might now be worth adding the library that's why I did not load it globally, let me know if you'd prefer not to have it.

prydonius: okay, was just wondering if there was actually some difference, thanks!

migmartri: Good catch @andresmgot.

  >  Also, in my opinion, we should try to use avoid using third-party modules if possible (to avoid possible vulnerabilities or license issues).

This is tricky, because this single module is nothing compared to all the dependencies from our main components.

I do not see any license issue in this case since it is MIT.

>  In any case I don't think we should add it, as happened with @prydonius or myself, other people may be confused with the context usage.

What do you find confusing? Having two methods? This concept [comes from rspec](http://www.betterspecs.org/#contexts) in which describe and context are used for different things. [Context used to setup a context or scenario and describe to express the test subject](http://lmws.net/describe-vs-context-in-rspec).

I can remove it and use nested describes but honestly that looks odd and adding `context` helps with tests readability (at least for me). If you are strongly opinionated against it I can remove it.

andresmgot: It's a minor thing, I know that in this case it's just a minor package with a single line. I am not against it so go ahead if you want to use it. I guess I am more used to see nested `describe`s so I don't find it weird.

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -140,47 +141,57 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
...
+
+  public appInfo() {
     const { app } = this.props;
-    if (!app || !app.info) {
-      return <div>Loading</div>;
-    }
+
+    // Although LoadingWrapper checks that the app is loaded before loading this wrapper
+    // it seems that react renders it even before causing it to crash because app is null
+    // that's why we need to have an app && guard clause
```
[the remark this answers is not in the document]
migmartri: > I would find that easier to understand, what's the benefit of setting appInfo as children?

That's a good point. I can change that. There is not an actual benefit apart of you do not need to add logic to show one or the other. That said, the fact that I need to do `app` actually defuses the purpose.

Issue #502, opened 2018-08-16
Add /healthz endpoints in our services and use them as k8s probes 


migmartri:
It seems that our chart does not include `probes` for some of our services, i.e tiller-proxy or chart-svc.

We could potentially standardize on the addition of a `/healthz` endpoint in those services than then could be hit by the helm chart tests if needed. 

This issue is about reviewing that all our services have a health/status page and probes are correctly set.

arapulido:
Won't fix

PR #645, opened 2018-09-14
chart: add liveness/readiness probes to chartsvc deployment

prydonius:
chartsvc has endpoints for checking liveness and readiness but we were
not taking advantage of these in our Kubernetes configurations. This
updates the chart to configure the probes for this service.


chart/kubeapps/values.yaml
```diff
@@ -124,6 +124,18 @@ chartsvc:
     # requests:
     #  cpu: 100m
     #  memory: 128Mi
+  livenessProbe:
+    httpGet:
+      path: /live
```
migmartri: Is this route (/live) based on any common practice? I've seen the proliferation of healthz instead

migmartri:
Refs https://github.com/kubeapps/kubeapps/issues/502

chart/kubeapps/values.yaml
migmartri: I see, you are just just using them.

chart/kubeapps/values.yaml
```diff
@@ -124,6 +124,18 @@ chartsvc:
...
+      port: 8080
+    initialDelaySeconds: 60
+    timeoutSeconds: 5
+  readinessProbe:
+    httpGet:
+      path: /ready
+      port: 8080
+    initialDelaySeconds: 0
```
migmartri: What's the default? Or you want to be explicit?

chart/kubeapps/values.yaml
prydonius: @migmartri when I implemented this in chartsvc, I used the https://github.com/heptiolabs/healthcheck library that suggests using `/live` and `/ready` https://github.com/heptiolabs/healthcheck#usage. Specifically, the difference from using `/healthz` is having two separate endpoints that might have different implementations of the healthcheck (e.g. `/ready` might check that MongoDB is connectable, though not currently implemented).

chart/kubeapps/values.yaml
prydonius: Oddly, the documentation doesn't specify what the default is: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#probe-v1-core. I just copied these from the other services that have healtchecks (dashboard and frontend). I think it's good to be explicit (especially given it's unclear what the default is), and I think that it's fine for readiness to not have any delay. The initialDelaySeconds for liveness might actually be too long for a simple service like this, but this could give it more time to attempt to connect to MongoDB.

chart/kubeapps/values.yaml
migmartri: Oh nice, I was not aware of that library, thanks for the info.

> Specifically, the difference from using a single /healthz is having two separate endpoints that might have different implementations of the healthcheck (e.g. /ready might check that MongoDB is connectable, though not currently implemented).

Right, in the past I've been handling this by passing a `liveness` parameter to the /healthz endpoint.

migmartri: > @migmartri makes sense, yeah, the name of the endpoints doesn't matter much here I guess, although I agree that `healthz` is more common. `/live` and `/ready` are also possibly too specific to Kubernetes, although the concept does transfer across. Either way, probably not important enough to change.

No it's not important, thanks for the info.

Issue #551, opened 2018-08-28
Replace typesafe-actions with built-in typescript 2.8+ redux support


migmartri:
https://medium.com/@martin_hotell/improved-redux-type-safety-with-typescript-2-8-2c11a8062575

Refs https://github.com/kubeapps/kubeapps/issues/534

Issue #534, opened 2018-08-24
Look into upgrading typescript to 3.x and remove backports


migmartri:
Currently, our runtime is running typescript 2.9.x but for example vscode is running 3.x. This is causing some libraries definitions to fail at build time time but not during development.

We should look into upgrading to the latest stable version of typescript.

We should also remove any backports we have in /backports.d.ts

migmartri:
It seems that upgrading to typescript 3.x requires us to for example upgrade @types/react-redux which also requires Redux 4.x https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-redux/package.json

This means that we should probably perform a React/Redux upgrade at the same time https://github.com/kubeapps/kubeapps/issues/516

Issue #516, opened 2018-08-21
Upgrade react to latest 16.x


migmartri:
It seems that react 16.3 contains "replacements" for the deprecated containerWillReceiveProps and new lifecycle events. https://reactjs.org/blog/2018/03/29/react-v-16-3.html

We should upgrade so our new code leverages these changes.

Issue #521, opened 2018-08-22
Upgrade and unlock jest-enzyme package


migmartri:
Currently, the jest-enzyme package is locked to version 6.0.2 due to [a regression](https://github.com/FormidableLabs/enzyme-matchers/issues/252) in 6.0.3 that breaks the typescript support.

We should upgrade and unlock the package once a new release is available.

Refs https://github.com/kubeapps/kubeapps/pull/520

cc/ @prydonius

migmartri:
Closed by https://github.com/kubeapps/kubeapps/pull/650

PR #650, opened 2018-09-18
Upgrade typescript to 3.x and latest react/redux

migmartri:
NOTE: I think it is best to **not to** aim to merge this change in the next release but instead wait for next iterations release. That way we can do additional testing. In any case, feel free to review, reject or approve the patch. 

It contains a mayor library maintenance update which includes:

* Upgraded all the minor versions of all our packages, Fixes https://github.com/kubeapps/kubeapps/issues/516, https://github.com/kubeapps/kubeapps/issues/521
* Typescript upgrade to 3.x from 2.7. Fixes https://github.com/kubeapps/kubeapps/issues/534
* The backports have been removed `backport.d.ts`
* There has been many formatting changes due to `prettier` running for some reason, i.e https://github.com/kubeapps/kubeapps/pull/650/files#diff-2d289e99afa9718b5378ebedc81ac00cL25 
* As a consequence of the TS upgrade, I needed to bump some mayor versions including `react-redux`. Fixes https://github.com/kubeapps/kubeapps/issues/551

Regarding these changes, the highlights are:

NOTE: I've tried that all the changes made are complaint with this style guide https://github.com/piotrwitek/react-redux-typescript-guide

**ThunkAction**

Up until now, we have been using a mix of import for dispatch either [coming from `react-redux`](https://github.com/kubeapps/kubeapps/pull/650/files#diff-18b740e9663a3fe593db9f24c6494a31L1) or [from `redux` itself](https://github.com/kubeapps/kubeapps/pull/650/files#diff-5801116feaa1cd958f981f3cc8d6e21cL1). The former is no longer supported and the type in the later is not longer overloaded so it requires `type: xx`. That's why now we need to use the new type `ThunkAction` (from redux-thunk) for async action creators.

**typesafe-actions and utility-types**

`typesafe-actions` has been upgraded to the newest mayor version. This means that the library [has deprecated some methods](https://github.com/piotrwitek/utility-types#deprecated-api-use-at-own-risk) since there are alternatives built-in in TS 2.8+. `typesafe-actions` also adds new methods like `ActionType` which is being used as a replacement.

typesafe-actions has also deprecated the method `createAction`, having now [4 other ways to do it](https://github.com/piotrwitek/typesafe-actions#v1xx-to-v2xx) but for now I am marking them as deprecated and they will be tackled here https://github.com/kubeapps/kubeapps/issues/656

I have manually tested both in development and production mode but I'd appreciate if you could spend some time giving it a try (this docker image is built using this branch `migmartri/kubeapps-dashboard:ts-upgrade`)

**snapshot changes**

Enzyme has now fragment support in their snapshots, that's why this PR contains some snapshot updates.

### Misc notes

I have decided not to upgrade mayor versions since for example react-select has been completely rewritten. We can look into it in other PR's. These are the only upgrades pending.

```
 Major Update Potentially breaking API changes. Use caution.
❯◯ @types/json-schema   6.0.1   ❯  7.0.0    https://github.com/DefinitelyTyped/DefinitelyTyped
 ◯ @types/react-select  1.3.4   ❯  2.0.3    https://github.com/DefinitelyTyped/DefinitelyTyped
 ◯ react-ace            5.10.0  ❯  6.1.4    https://github.com/securingsincity/react-ace#readme
 ◯ react-lazylog        2.2.0   ❯  3.1.4    https://github.com/mozilla-frontend-infra/react-lazylog#readme
 ◯ react-select         1.3.0   ❯  2.0.0    https://github.com/JedWatson/react-select#readme
 ◯ ts-jest              22.4.6  ❯  23.1.4   https://github.com/kulshekhar/ts-jest#readme
 ◯ @types/jest devDep   22.2.3  ❯  23.3.2   https://github.com/DefinitelyTyped/DefinitelyTyped
 ◯ @types/node devDep   9.6.31  ❯  10.10.1  https://github.com/DefinitelyTyped/DefinitelyTyped
 ◯ lint-staged devDep   6.1.1   ❯  7.2.2    https://github.com/okonet/lint-staged#readme

```

* Axios can not be upgraded until moxios is [DefinitelyTyped/package.json at master · DefinitelyTyped/DefinitelyTyped · GitHub](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/moxios/package.json)


dashboard/package.json
```diff
@@ -29,13 +29,13 @@
...
     "react-modal": "^3.1.11",
-    "react-redux": "^5.0.6",
+    "react-redux": "5.0.6",
     "react-router-dom": "^4.2.2",
     "react-router-redux": "^5.0.0-alpha.9",
     "react-scripts-ts": "^2.17.0",
     "react-select": "^1.2.1",
     "react-test-renderer": "^16.2.0",
-    "redux": "^3.7.2",
+    "redux": "4.0.0",
```
andresmgot: don't you trust semver ranges?

dashboard/src/actions/catalog.ts
```diff
@@ -211,7 +212,7 @@ export function getPlans() {
 }
 
 export function getCatalog(ns?: string) {
-  return async (dispatch: Dispatch<IStoreState>) => {
+  return async (dispatch: ThunkDispatch<IStoreState, null, Action>) => {
```
andresmgot: I am not sure I understand `ThunkDispatch`: the other functions in this file also call several times to `dispatch`, what's the difference here?  Is it because of the nested calls?

dashboard/src/containers/AppNewContainer/AppNewContainer.tsx
```diff
@@ -31,7 +33,9 @@ function mapStateToProps(
   };
 }
 
-function mapDispatchToProps(dispatch: Dispatch<IStoreState>) {
+function mapDispatchToProps(
+  dispatch: ThunkDispatch<IStoreState, null, ChartsAction & ServiceCatalogAction>,
```
andresmgot: what does it mean `ChartsAction & ServiceCatalogAction` here?

dashboard/src/containers/AppNewContainer/AppNewContainer.tsx
```diff
@@ -45,7 +49,7 @@ function mapDispatchToProps(dispatch: Dispatch<IStoreState>) {
       dispatch(actions.charts.getChartValues(id, version)),
     getChartVersion: (id: string, version: string) =>
       dispatch(actions.charts.getChartVersion(id, version)),
-    push: (location: string) => dispatch(push(location)),
+    push: (location: string) => dispatch(push(location) as any),
```
andresmgot: why it's needed to set `push(location) as any`?

dashboard/package.json
migmartri: You were right. I had it locked because I was fighting with versions and stuff. Changed, thanks!

dashboard/src/actions/catalog.ts
migmartri: If we are dispatching a pure sync action creator, one that contains { type: foo}, you can use the `Dispatch` type (i.e dispatch(installed())), on the other hand, if you are dispatching an async reduxThunk action, i.e `getBindings` you need to use this type instead. This is because, in those cases the signature of the action contains the original dispatch and the getState method. This type signature, in theory allows you to inject those types to the action.

```
function incrementIfOdd() {
  return (dispatch, getState) => {
  ...
```

We could have avoided this by overloading Dispatch to allow not only `type` but also a function but it will defuse the purpose of being able to define what info and return types the async action has.

migmartri: Long story short, you can always use ThunkDispatch since it is an overloaded but using specifically `redux.Dispatch` when possible will allow you to find issues like this one https://github.com/kubeapps/kubeapps/pull/650/files#diff-67442faa85ea33f56c4ef70c57573277R17 in which we are not actually sending the action (note the parenthesis) and was caught thanks to the new typing.

dashboard/src/containers/AppNewContainer/AppNewContainer.tsx
migmartri: It is the only way I managed to make it work. I tried different things like `RouterAction` but no luck. It seems that push does not fit neither Dispatch nor ThunkDispatch. 

In any case, I'll look into it tomorrow, there must be a better way to do this. Thanks!

dashboard/src/actions/catalog.ts
andresmgot: got it, thanks for the detailed explanation! I agree that's better to use the restrictive `Dispatch` instead that `ThunkDispatch` always.

dashboard/src/containers/AppNewContainer/AppNewContainer.tsx
migmartri: Ok so I was wrong here.

My intention was to say that the dispatch inside the mapStateToProps was going to receive either a chart or a serviceCatalog action, but reality is that it was receiving an async action in all those cases.

Also, I was using an intersection when I actually meant union, but it was working because of the reason described above, async (ThunkActions) are already supported in the type so it does not matter what I put as last parameter that it did not change anything.

I have noticed that actually setting a specific value i.e `ServiceCatalogAction` as last type parameter is not that valuable in the mapStateToProps since it can be confusing and also, once you define the signature of the actionCreator this value can be overridden (see https://github.com/kubeapps/kubeapps/pull/650/files#diff-2fc417299c3e2ab5151030d339ab8dfcR107) so I changed it to use the global `Action` instead. 

Then, once we adapt our action creators we can use a proper type https://github.com/kubeapps/kubeapps/issues/658

dashboard/src/containers/AppNewContainer/AppNewContainer.tsx
migmartri: I fixed it. Initially was not working because I was using type intersection instead of union. But in the end I decided to use the top level redux `Action` type for simplicity. https://github.com/kubeapps/kubeapps/pull/650#discussion_r218931399

dashboard/package.json
```diff
@@ -61,9 +58,20 @@
     "lint": "npm-run-all lint-js lint-css"
   },
   "lint-staged": {
-    "*.{ts,tsx}": ["yarn run lint-js-fix", "prettier --write", "git add"],
-    "*.scss": ["yarn run lint-css-fix", "prettier --write", "git add"],
-    "*.{js,json,md}": ["prettier --write", "git add"]
+    "*.{ts,tsx}": [
```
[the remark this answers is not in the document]
migmartri: I have no idea tbh. Note that prettier has been upgraded, let me see if there is any difference there.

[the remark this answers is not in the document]
migmartri: It seems that the newer version of prettier formats differently

Before: ./node_modules/prettier/bin-prettier.js --version
1.10.2

After: ./node_modules/prettier/bin-prettier.js --version
1.14.2

Diff package.json

 ./node_modules/prettier/bin-prettier.js package.json

```diff
diff /tmp/old-version.json /tmp/new-version.json 
15d14
<     "enzyme": "^3.3.0",
17d15
<     "jest-enzyme": "6.0.2",
38c36
<     "redux": "^3.7.2",
---
>     "redux": "^4.0.0",
42c40
<     "typesafe-actions": "^1.1.2"
---
>     "typesafe-actions": "^2.0.4"
47,48c45
<     "start-js":
<       "BROWSER=none REACT_APP_KUBEAPPS_NS=${TELEPRESENCE_CONTAINER_NAMESPACE} react-scripts-ts start",
---
>     "start-js": "BROWSER=none REACT_APP_KUBEAPPS_NS=${TELEPRESENCE_CONTAINER_NAMESPACE} react-scripts-ts start",
64,66c61,74
<     "*.{ts,tsx}": ["yarn run lint-js-fix", "prettier --write", "git add"],
<     "*.scss": ["yarn run lint-css-fix", "prettier --write", "git add"],
<     "*.{js,json,md}": ["prettier --write", "git add"]
---
>     "*.{ts,tsx}": [
>       "yarn run lint-js-fix",
>       "prettier --write",
>       "git add"
>     ],
>     "*.scss": [
>       "yarn run lint-css-fix",
>       "prettier --write",
>       "git add"
>     ],
>     "*.{js,json,md}": [
>       "prettier --write",
>       "git add"
>     ]
76c84
<     "@types/react-redux": "^5.0.14",
---
>     "@types/react-redux": "^6.0.3",
81a90
>     "enzyme": "^3.6.0",
83c92
<     "jest-enzyme": "6.0.2",
---
>     "jest-enzyme": "^6.0.4",
100c109
<     "typescript": "^2.6.2"
---
>     "typescript": "^3.0.3"
103c112,115
<     "collectCoverageFrom": ["src/**/*", "!src/**/*.d.ts"]
---
>     "collectCoverageFrom": [
>       "src/**/*",
>       "!src/**/*.d.ts"
>     ]
```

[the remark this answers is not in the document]
migmartri: @prydonius here is your answer https://prettier.io/blog/2018/05/27/1.13.0.html#added-json-stringify-parser-for-jsonstringify-style-formatting-4450-by-ikatyang

PR #652, opened 2018-09-18
Move Routes to a component

andresmgot:
From: https://github.com/kubeapps/kubeapps/pull/618#discussion_r217861841

Move `Routes` to a component and get state trough `connect`. cc/ @prydonius


dashboard/src/containers/RoutesContainer/RoutesContainer.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { ClassViewContainer } from "../ClassView";
+import FunctionListContainer from "../FunctionListContainer";
+import FunctionViewContainer from "../FunctionViewContainer";
+import InstanceListViewContainer from "../InstanceListViewContainer";
+import InstanceView from "../InstanceView";
+import LoginFormContainer from "../LoginFormContainer";
+import RepoListContainer from "../RepoListContainer";
+import ServiceCatalogContainer from "../ServiceCatalogContainer";
+
+const privateRoutes: {
```
migmartri: Do these need to be in the container? Why not in the component?

dashboard/src/containers/RoutesContainer/RoutesContainer.tsx
```diff
@@ -0,0 +1,56 @@
...
+};
+
+// Public routes that don't require authentication
+const routes: {
+  [route: string]: React.ComponentType<RouteComponentProps<any>> | React.ComponentType<any>;
+} = {
+  "/login": LoginFormContainer,
+};
+
+function mapStateToProps({ namespace }: IStoreState) {
```
migmartri: In my opinion, the container should only include the mapStateToProps and the connect. 

Then the component/routes.tsx should be practically the same than it used to be except that now it will pick up the namespace from the props?

andresmgot: yes, the reason was to import containers from other containers and for the routes to be easier to discover. In any case I can move them to the component.

andresmgot:
> I might be missing the point here, but this seems to me like an unnecessary level of indirection for not using the store directly.

The reasons I understood from @prydonius is for consistency with the current structure and for making the code easier to test without having to work with the storage.

dashboard/src/components/Routes/Routes.tsx
```diff
@@ -0,0 +1,82 @@
...
+            path={route}
+            component={privateRoutes[route]}
+          />
+        ))}
+        {/* If the route doesn't match any expected path redirect to a 404 page  */}
+        <Route component={NotFound} />
+      </Switch>
+    );
+  }
+  private rootNamespacedRedirect = (props: any) => {
```
migmartri: I believe that this type can also be `IRoutesProps`

prydonius:
Hmm @migmartri raises a good point here, so rethinking this what we might want to do is just have the component class inside the container file and have the container connect to that local React component that is defined in the container file. Does that make sense? That would address @migmartri's point about it being an indirection and Routes not being a "real" component, but we still are able to connect to the Redux state correctly and pass in the props we need to the container.

prydonius:
I mean the container is just a concept, right? React doesn't actually have a container type, everything is a component, we just use the separation pattern because it sometimes makes sense to separate the presentational components (components) from components that know more about the state (containers).

The way we implement the container pattern is a bit different because of Redux's connect helpers, but outside of Redux a container would look something like this:

```
class CommentListContainer extends React.Component {
  state = { comments: [] };
  componentDidMount() {
    fetchSomeComments(comments =>
      this.setState({ comments: comments }));
  }
  render() {
    return <CommentList comments={this.state.comments} />;
  }
}
```
(taken from https://medium.com/@learnreact/container-components-c0e67432e005)

It's just a component that knows about some state and passes that state to another component as props. With Redux, we are essentially doing the same thing but using the connect and mapState/DispatchToProps helpers.

In this particular case, we don't really see Routes as a presentational component (i.e. in the same way we do AppList), but rather something that is core to configuring the React application. In that sense, I think it doesn't make sense for it to belong in the container. Root.tsx is another example of this, by the way, it's technically a Component, but we see it as a container.

Does that make more sense?

Issue #635, opened 2018-09-13
Error handling on release/view page


migmartri:
See https://github.com/kubeapps/kubeapps/issues/632#issue-359988242 for more information

migmartri:
This issue will be fixed once https://github.com/kubeapps/kubeapps/issues/657

Issue #636, opened 2018-09-13
Relax yaml parsing in releases/view page


migmartri:
@stormmore [reported](https://github.com/kubeapps/kubeapps/issues/632) that Kubeapps was failing to show the resources part of his chart. 

The [research showed](https://github.com/kubeapps/kubeapps/issues/632#issuecomment-421111993) that the root cause was a parsing error on our client.

Our yaml parser complained about a duplicated tag vs Helm does not. We should look into if there is any way to make our parser less restrictive. More taking into account that a duplicated label is not per-se forbidden or discouraged by the helm linter so it might happen again.

prydonius:
We use the safeLoad function from https://github.com/nodeca/js-yaml#safeload-string---options-, and it looks like there is an option to be compatible with the JSON.parse behaviour which explicitly says "If true, then duplicate keys in a mapping will override values rather than throwing an error." So setting this to true should fix the reported issue, but we may want to also consider the non-safe `load` function as we know that Helm & Kubernetes have pre-validated the YAML we're reading.

prydonius:
For some reason, likely due to a bug in the upstream package, yaml.safeLoadAll does not seem to correctly support the `json` option, though yaml.safeLoad and yaml.loadAll do. Once again, given that we are only going to be parsing YAML that has already been parsed by Helm and Kubernetes, I think it is okay to switch to `yaml.loadAll` here, which seems to support some special JS-related YAML tags: https://github.com/nodeca/js-yaml#load-string---options-.

Filed https://github.com/nodeca/js-yaml/issues/456 for the upstream issue.

PR #653, opened 2018-09-18
Catch error when parsing YAML manifests

andresmgot:
Fixes: #635

Add `manifestError` to the AppView state for setting it if the manifest cannot be parsed:

screen shot 2018-09-18 at 17 38 15

### DIFF
--- a/dashboard/src/components/AppView/AppView.test.tsx
+++ b/dashboard/src/components/AppView/AppView.test.tsx
@@ -7,7 +7,7 @@ import { hapi } from "../../shared/hapi/release";
 import itBehavesLike from "../../shared/specs";
 import { ForbiddenError, IResource, NotFoundError } from "../../shared/types";
 import DeploymentStatus from "../DeploymentStatus";
-import { NotFoundErrorAlert, PermissionsErrorAlert } from "../ErrorAlert";
+import { NotFoundErrorAlert, PermissionsErrorAlert, UnexpectedErrorAlert } from "../ErrorAlert";
 import AppControls from "./AppControls";
 import AppDetails from "./AppDetails";
 import AppNotes from "./AppNotes";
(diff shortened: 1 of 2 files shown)


migmartri:
I am not sure if we want to add specific error handling for this case instead of eventually a global errorBoundary. About this specific case, did you take a look at this comment? https://github.com/kubeapps/kubeapps/issues/636#issuecomment-421160771

They are orthogonal issues meaning that both make sense but once the parser is relaxed, errors in the parsing might be scarce and hence this specific error handling not needed. WDYT?

prydonius:
@migmartri I agree, and especially since we know that this blob of YAML has gone through both Helm and Kubernetes, it really _should_ be parseable (unless someone has modified the state in the ConfigMap to make it not so).

andresmgot:
I thought that "Our error handling here seems deficient and misleading, we need to improve it. #635" meant to catch the error here and that was the purpose of the issue. If the goal was to refactor how we handle errors I have already opened #654 with a more detailed description. Should we close this PR and #635?

migmartri:
> I thought that "Our error handling here seems deficient and misleading, we need to improve it. #635" meant to catch the error here and that was the purpose of the issue. If the goal was to refactor how we handle errors I have already opened #654 with a more detailed description. Should we close this PR and #635?

Sorry for the confusion @andresmgot. What I meant with that sentence is that we need to tell the user that there was an error, not showing a partially rendered error. I believe that this could be done using https://reactjs.org/docs/error-boundaries.html at a higher level.

In this patch, you solved that, you are right, but just for the case in which there was a yaml parsing error. Which it would be fine if it was not because we could potentially prevent from this error from happening in the first place. So my argument is about 1: Not handle this specific error but change the yaml parser behavior, 2: Look into a generic way of handling runtime errors via errorBoundaries.

To your question. I'd close this PR and just closed 535 since I have created a generic error handling task https://github.com/kubeapps/kubeapps/issues/657

PR #660, opened 2018-09-20
Increase number of replicas and allow its configuration

migmartri:
Closes https://github.com/kubeapps/kubeapps/issues/643

Diff of rendered template before/after the change.

```bash
 multidoc_yamldiff /tmp/before.yaml /tmp/after.yaml 
@@ .. @@
    name: "RELEASE-NAME-kubeapps",
   },
   spec: {
-   replicas: 1,
+   replicas: 2,
    selector: {
     matchLabels: {
      app: "RELEASE-NAME-kubeapps",
@@ .. @@
    name: "RELEASE-NAME-kubeapps-internal-apprepository-controller",
   },
   spec: {
-   replicas: 1,
+   replicas: 2,
    selector: {
     matchLabels: {
      app: "RELEASE-NAME-kubeapps-internal-apprepository-controller",
@@ .. @@
    name: "RELEASE-NAME-kubeapps-internal-chartsvc",
   },
   spec: {
-   replicas: 1,
+   replicas: 2,
    selector: {
     matchLabels: {
      app: "RELEASE-NAME-kubeapps-internal-chartsvc",
@@ .. @@
    name: "RELEASE-NAME-kubeapps-internal-dashboard",
   },
   spec: {
-   replicas: 1,
+   replicas: 2,
    selector: {
     matchLabels: {
      app: "RELEASE-NAME-kubeapps-internal-dashboard",
@@ .. @@
    name: "RELEASE-NAME-kubeapps-internal-tiller-proxy",
   },
   spec: {
-   replicas: 1,
+   replicas: 2,
    selector: {
     matchLabels: {
      app: "RELEASE-NAME-kubeapps-internal-tiller-proxy",
@@ .. @@
  {
   apiVersion: "v1",
   data: {
-   mongodb-root-password: "SmRMOEg4Nzg5TQ==",
+   mongodb-root-password: "ZTZyOThrQld5RQ==",
   },
   kind: "Secret",
   metadata: {
```

After upgrade

```bash
 k get deployment
NAME                                                DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE
miguel-kubeapps                                     2         2         2            2           6d
miguel-kubeapps-internal-apprepository-controller   2         2         2            2           6d
miguel-kubeapps-internal-chartsvc                   2         2         2            2           6d
miguel-kubeapps-internal-dashboard                  2         2         2            2           6d
miguel-kubeapps-internal-tiller-proxy               2         2         2            2           6d
[...]

PR #661, opened 2018-09-20
Fix port reference in tiller-proxy deployment manifest

migmartri:
No changes as expected since the port for chartsvc and tiller proxy are the same.

Bumping two versions so this PR will wait until this other one is landed https://github.com/kubeapps/kubeapps/pull/660

```
 diff /tmp/before.yaml /tmp/after.yaml 
14c14
<   mongodb-root-password: "enVCdFRiZDZYRw=="
---
>   mongodb-root-password: "TzVseUNqZjFqWg=="
```

### DIFF
--- a/chart/kubeapps/Chart.yaml
+++ b/chart/kubeapps/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kubeapps
-version: 0.4.7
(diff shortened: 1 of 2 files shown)


prydonius:
Oops, sorry - great catch!

PR #662, opened 2018-09-20
[WIP] Refactor error handling with ErrorSelector

### LINKED ISSUE #654: Refactor error rendering
We have the same method `renderError` in many components. This method takes an error as parameter and it renders a custom message based on it. These are the components that use that:

 - Config/AppRepoList/index.tsx
 - Config/AppRepoList/AppRepoButton.tsx
 - Config/ServiceBrokerList/ServiceBrokerList.tsx
 - ChartView/ChartView.tsx
 - InstanceListView/index.tsx
 - UpgradeForm/SelectRepoForm.tsx
 - AppView/AppView.tsx
 - AppList/AppList.tsx
 - ClassList/index.tsx
 - ProvisionButton/index.tsx
 - InstanceView/AddBindingButton.tsx
 - ClassView/index.tsx

We should refactor it to at least remove duplicated code. Then we can evaluate if its possible to refactor how errors are set: it may be possible to use an external library to read errors from Redux so components can delegate how the errors are rendered.

andresmgot:
Closes #654

This PR:

 - Removes duplicated code in different components that used `renderError`
 - Adds the `ErrorSelector` component that renders an error message depending on the error type.
 - Deprecates the `NotFoundErrorAlert`. It's code is now part of `UnexpectedErrorAlert`. I haven't removed it because some functions and service broker components still use it.
 - Extends `PermissionErrorAlert` from `UnexpectedErrorAlert`.

I still need to review the existing testing and add  new unit tests for the changes in the error components.


dashboard/src/components/UpgradeForm/UpgradeForm.tsx
```diff
@@ -109,11 +117,6 @@ class UpgradeForm extends React.Component<IDeploymentFormProps, IDeploymentFormS
       <div>
         <form className="container padding-b-bigger" onSubmit={this.handleDeploy}>
           <div className="row">
-            <div className="col-8">
```
andresmgot: this code was not reachable, `this.props.error` is checked before that is not set

dashboard/src/components/UpgradeForm/UpgradeForm.tsx
```diff
@@ -201,6 +204,17 @@ class UpgradeForm extends React.Component<IDeploymentFormProps, IDeploymentFormS
   public handleReselectChartRepo = () => {
     this.props.clearRepo();
   };
+
+  private requiredRBACRoles(): IRBACRole[] {
```
andresmgot: this requiredRBACRoles were previously in `DeploymentErrors`

dashboard/src/components/ErrorAlert/ErrorSelector.tsx
```diff
@@ -0,0 +1,95 @@
...
+  // Default permissions that the component should have
+  // returned in case it's a ForbiddenError and the message
+  // doesn't contain specific RBAC roles
+  defaultRequiredRBACRoles?: {
+    [action: string]: IRBACRole[];
+  };
+  // Namespace in case the error is namespaced
+  namespace?: string;
+  // Resource causing the failure (e.g. "Application my-wordpress")
+  resource?: string;
```
migmartri: It's nice that you are using optional props.

dashboard/src/components/ErrorAlert/ErrorSelector.tsx
```diff
@@ -0,0 +1,95 @@
...
+}
+
+class ErrorSelector extends React.Component<IErrorSelectorProps> {
```
migmartri: As a personal preference, I'd try to define components as Stateless functional component (SFC) when possible.

dashboard/src/components/ErrorAlert/ErrorSelector.tsx
```diff
@@ -0,0 +1,95 @@
...
+  public render() {
+    const { namespace, error, defaultRequiredRBACRoles, action, resource } = this.props;
+    switch (error.constructor) {
+      case AppConflict:
```
migmartri: It seems that we are treating `AppConflict` like a generic error type when in reality it does not seem to be. 

I am wondering if we should remove this one from here and just allow `ErrorSelector` to support a message string.

dashboard/src/components/AppUpgrade/AppUpgrade.test.tsx
```diff
@@ -58,7 +58,7 @@ it("renders the repo selection form if not introduced", () => {
     />,
   );
   expect(wrapper.find(SelectRepoForm).exists()).toBe(true);
-  expect(wrapper.find(DeploymentErrors).exists()).toBe(false);
+  expect(wrapper.find(ErrorSelector).exists()).toBe(false);
```
migmartri: not worth changing it now but here we can use `.not.toExist()`

andresmgot:
I am happy now with the test coverage. Please @migmartri take a look when you have a chance.

dashboard/src/components/AppUpgrade/AppUpgrade.test.tsx
```diff
@@ -75,12 +76,40 @@ it("renders an error if it exists", () => {
...
 });
 
+it("renders a forbidden error", () => {
+  const repo = {
+    metadata: { name: "stable" },
+  } as IAppRepository;
+  const wrapper = shallow(
+    <AppUpgrade {...defaultProps} error={new ForbiddenError()} repos={[repo]} repo={repo} />,
+  );
+  expect(wrapper.find(ErrorSelector).exists()).toBe(true);
```
migmartri: But for new test cases I'd use the other DSL `.not.toExist()`

dashboard/src/components/AppUpgrade/AppUpgrade.test.tsx
```diff
@@ -75,12 +76,40 @@ it("renders an error if it exists", () => {
...
+  expect(wrapper.find(SelectRepoForm).exists()).toBe(false);
+  expect(wrapper.find(UpgradeForm).exists()).toBe(false);
+  expect(wrapper.html()).toContain(
+    "You don&#x27;t have sufficient permissions to update foo in <span>the <code>default</code> namespace</span>",
+  );
+  expect(
+    wrapper
+      .find(ErrorSelector)
+      .shallow()
+      .find(PermissionsErrorPage)
```
migmartri: Can't you find `PermissionsErrorPage` directly from the wrapper? Or that's why you call shallow() to go one level deep?

dashboard/src/components/AppUpgrade/AppUpgrade.test.tsx
```diff
@@ -75,12 +76,40 @@ it("renders an error if it exists", () => {
...
   );
-  expect(wrapper.find(DeploymentErrors).exists()).toBe(true);
+  expect(wrapper.find(ErrorSelector).exists()).toBe(true);
   expect(wrapper.find(SelectRepoForm).exists()).toBe(false);
   expect(wrapper.find(UpgradeForm).exists()).toBe(false);
+  expect(wrapper.html()).toContain("Sorry! Something went wrong.");
   expect(wrapper).toMatchSnapshot();
 });
 
+it("renders a forbidden error", () => {
```
migmartri: Maybe you can split this to context + case name?

context("when we receive a ForbiddenError error prop")...
it("renders the error message")

or something like that.

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -148,7 +148,15 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
 
   public render() {
     if (this.props.error) {
-      return this.renderError(this.props.error);
+      return (
+        <ErrorSelector
+          error={this.props.error}
+          defaultRequiredRBACRoles={RequiredRBACRoles}
```
migmartri: Why is this called `defaultRequiredRBACRoles`? Why default? Wouldn't `requiredRBACRoles` make more sense?

dashboard/src/shared/AxiosInstance.test.ts
```diff
@@ -63,7 +63,7 @@ describe("createAxiosInterceptor", () => {
     { code: 401, errorClass: UnauthorizedError },
     { code: 403, errorClass: ForbiddenError },
     { code: 404, errorClass: NotFoundError },
-    { code: 409, errorClass: AppConflict },
+    { code: 409, errorClass: ConflictError },
```
migmartri: Nice :)

dashboard/src/components/ErrorAlert/ErrorSelector.tsx
```diff
@@ -0,0 +1,93 @@
...
+  NotFoundError,
+  UnprocessableEntity,
+} from "../../shared/types";
+import { PermissionsErrorAlert, UnexpectedErrorAlert } from "./";
+import { namespaceText } from "./helpers";
+
+interface IErrorSelectorProps {
+  error: Error;
+  // Resource causing the failure (e.g. "Application my-wordpress")
+  resource: string;
```
migmartri: I'd personally rename this prop to `resourceName`

dashboard/src/components/ErrorAlert/ErrorSelector.tsx
```diff
@@ -0,0 +1,93 @@
...
+    [action: string]: IRBACRole[];
+  };
+  // Namespace in case the error is namespaced
+  namespace?: string;
+}
+
+const ErrorSelector: React.SFC<IErrorSelectorProps> = props => {
+  const { namespace, error, defaultRequiredRBACRoles, action, resource } = props;
+  switch (error.constructor) {
+    case ConflictError:
```
migmartri: I am not convinced about the need of having this error type. It seems too specific when it could come under the umbrella of  `UnprocessableEntity`. Actually that's the error (422) that the server should return since it is a data validation error. 

I personally think that allowing the client to call ErrorSelector with a `UnprocessableEntity` error and a custom title would work best. 

In any case, if you think that it is worth maintaining this error type, go ahead.

dashboard/src/components/ErrorAlert/ErrorSelector.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+    );
+    expect(wrapper).toMatchSnapshot();
+  });
+  it("should extract the required RBAC roles from the error message", () => {
+    const role = { apiGroup: "v1", namespace: "my-ns", resource: "my-app", verbs: ["get", "list"] };
+    const message = JSON.stringify([role]);
+    const wrapper = shallow(
+      <ErrorSelector error={new ForbiddenError(message)} resource="my-app" />,
+    );
+    const items = wrapper
```
migmartri: Wouldn't make more sense to create the wrapper using `mount` in these cases so you do not need to chase the items down the line?

andresmgot:
I have updated the PR, PTAL @prydonius. This PR is quite big so I may have missed something.

BTW, regarding the `import` style: what I do is to rely in vscode automatic imports which is pretty handy for me. I don't even look to the import paths, I consider it something more for machines to read. Anyway, if you strongly think that we should properly format that maybe we should add a linter rule for that so the linter corrector would automatically correct it (I don't know if there is a rule for that). WDYT?

dashboard/src/components/DeploymentForm/DeploymentForm.test.tsx
```diff
@@ -32,7 +31,8 @@ describe("renders an error", () => {
         selected={{ error: new NotFoundError() } as IChartState["selected"]}
       />,
     );
-    expect(wrapper.find(NotFoundErrorPage).exists()).toBe(true);
+    expect(wrapper.find(ErrorSelector).exists()).toBe(true);
+    expect(wrapper.find(ErrorSelector).html()).toContain("Chart &quot;foo&quot; (1.0.0) not found");
```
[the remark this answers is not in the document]
prydonius: I would avoid using mount in case there are a lot of components to dig into, but if there is only one like here, I think it's okay to use `.html()` here. Thanks!

dashboard/src/components/DeploymentForm/DeploymentForm.tsx
```diff
@@ -114,13 +115,12 @@ class DeploymentForm extends React.Component<IDeploymentFormProps, IDeploymentFo
...
         <form className="container padding-b-bigger" onSubmit={this.handleDeploy}>
           {this.props.error && (
-            <DeploymentErrors
-              {...this.props}
-              kubeappsNamespace={kubeappsNamespace}
-              chartName={chartID.split("/")[0]}
-              releaseName={releaseName}
-              repo={chartID.split("/")[1]}
-              version={version.attributes.version}
+            <ErrorSelector
```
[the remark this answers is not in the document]
prydonius: Ah, I recalled this being an issue when we still had HelmReleases as it would not create the namespace for you, but tiller-proxy will create inexistant namespaces. You're right that this isn't an issue.

prydonius:
@andresmgot I don't know if there is a linter rule for it, but it seems a bit odd to me. In my experience, VSCode has added the import I need from the index.ts file instead of the specific file, I would err on the side of importing from index.ts instead of the file directly.

dashboard/src/components/UpgradeForm/SelectRepoForm.tsx
```diff
@@ -32,7 +32,16 @@ class SelectRepoForm extends React.Component<ISelectRepoFormProps, ISelectRepoFo
...
       <LoadingWrapper loaded={this.props.repos.length > 0}>
         <div className="container margin-normal">
-          <div className="col-8">{this.props.error && this.renderError()}</div>
+          <div className="col-8">
+            {this.props.error && (
+              <ErrorSelector
+                error={this.props.error}
+                defaultRequiredRBACRoles={{ view: [this.requiredRBACRoles()] }}
+                action="view"
+                resource={`Chart ${this.state.repo}/${this.props.chartName}`}
```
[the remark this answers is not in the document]
prydonius: @andresmgot should we create a separate issue to tackle that?

dashboard/src/components/ErrorAlert/ErrorSelector.tsx
migmartri: I understand, but what I am arguing for is to not to treat a "release name invalid or taken" as something different than a validation error 422. That way we do not need to maintain a different use case for such a specific scenario.

Looking at the [409 error description](https://httpstatuses.com/409) to me, it does not seem to fit this purpose either, since it seems to be mainly focused on resource updates.

That said, it seems that the return code is set in Tiller-proxy so it might not make sense to make any change in that regard and carry on with the Conflict error handling.

migmartri: This is what I meant with the 422 vs 409 error https://groups.google.com/d/msg/api-craft/fQWIs3_UwJE/rUqyqWaxbCkJ

dashboard/src/components/UpgradeForm/SelectRepoForm.tsx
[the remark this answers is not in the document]
andresmgot: not needed, I already fixed it in this PR

dashboard/src/components/ErrorAlert/ErrorSelector.tsx
andresmgot: oh, I didn't know about that discussion. In this case (tiller-proxy) I mimic K8s behavior. They use a 409 when the resource already exists:

```
curl --data $JSON -H "Content-Type: application/json" -X POST http://127.0.0.1:8001/api/v1/namespaces/default/pods/
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {

  },
  "status": "Failure",
  "message": "pods \"nginx\" already exists",
  "reason": "AlreadyExists",
  "details": {
    "name": "nginx",
    "kind": "pods"
  },
  "code": 409
}%
```

andresmgot:
> In my experience, VSCode has added the import I need from the index.ts file instead of the specific file

You are right, when deleting the import and using auto import it uses the `index` file. I am not sure how happened the other way 🤷‍♂️ . In any case I have updated the PR to change that.

PR #669, opened 2018-09-21
Add missing loader components

migmartri:
During the efforts to implement loader components across the app I missed a couple https://github.com/kubeapps/kubeapps/issues/165


migmartri:
Thanks @andresmgot, I missed that one too.

I followed your suggestion and added the spinner to the modal as well.

peek 2018-09-24 12-06

Issue #648, opened 2018-09-17
Replace deprecated react-router-redux 


migmartri:
https://github.com/ReactTraining/react-router/tree/master/packages/react-router-redux

Replaced by https://github.com/supasate/connected-react-router

PR #675, opened 2018-09-24
Replace deprecated react-router-redux

migmartri:
Replaced react-router-redux for connected-react-router, this also means that:

* There is a new way of extending the root reducer by using `connectRouter` on the rootReducer.
* Types are not part of the library itself.
* Pathname seems to be now on `location.pathname`.

Closes https://github.com/kubeapps/kubeapps/issues/648


dashboard/src/reducers/namespace.ts
```diff
@@ -24,7 +24,7 @@ const namespaceReducer = (
     case getType(actions.namespace.setNamespace):
       return { ...state, current: action.namespace };
     case LOCATION_CHANGE:
-      const pathname = action.payload.pathname;
+      const pathname = action.payload.location.pathname;
```
andresmgot: it's weird that this is the only breaking change, isn't it? I know it's fine but it's a good opportunity to add unit tests :)

migmartri: It seems so, the main change I can see is this one and the use of `connectRouter(history)(rootReducer)` instead of making the user add a router key in the combineReducers. I looked at the code and it does exactly that.

I have added some unit tests on the locationChange case. Testing the whole router might be too much for this patch and I am not sure how much it overlaps with future integration tests. I have done some manual testing and the app seems to work as expected.

Issue #658, opened 2018-09-19
Add typing to action creators


migmartri:
Once this is landed https://github.com/kubeapps/kubeapps/pull/650/files#diff-5801116feaa1cd958f981f3cc8d6e21cL1, we should start using `ThunkAction` generics 
 on Thunk action creators.

This is an example of what I think it's a properly typed action.

```ts
export function deployChart(
  chartVersion: IChartVersion,
  releaseName: string,
  namespace: string,
  values?: string,
): ThunkAction<Promise<boolean>, IStoreState, null, AppsAction> {
  return async (dispatch, getState) => {
    try {
      // You can not deploy applications unless the namespace is set
      if (namespace === definedNamespaces.all) {
        throw new UnprocessableEntity(
          "Namespace not selected. Please select a namespace using the selector in the top right corner.",
        );
      }

      const {
        config: { namespace: kubeappsNamespace },
      } = getState();
      await App.create(releaseName, namespace, kubeappsNamespace, chartVersion, values);
      return true;
    } catch (e) {
      dispatch(errorApps(e));
      return false;
    }
  };
}
```

being the main difference

```diff
--- a/dashboard/src/actions/apps.ts
+++ b/dashboard/src/actions/apps.ts
@@ -102,8 +102,8 @@ export function deployChart(
   releaseName: string,
   namespace: string,
   values?: string,
-) {
-  return async (dispatch: Dispatch, getState: () => IStoreState): Promise<boolean> => {
+): ThunkAction<Promise<boolean>, IStoreState, null, AppsAction> {
+  return async (dispatch, getState) => {

```

In this case, the types for dispatch, getState and the returned value will be infered from `hunkAction<Promise<boolean>, IStoreState, null, AppsAction>`

PR #677, opened 2018-09-24
Update TS ThunkAction types and returned values in action creators

migmartri:
It updates the ReduxThunk action creators signature types. It also removes unnecessary returned values since an action creator should just do that, dispatch actions, not to be consumed directly. There are still a couple of actions returning a boolean that we could change later. 

Closes https://github.com/kubeapps/kubeapps/issues/658

### DIFF
--- a/dashboard/src/actions/apps.ts
+++ b/dashboard/src/actions/apps.ts
@@ -1,4 +1,3 @@
-import { Dispatch } from "redux";
 import { ThunkAction } from "redux-thunk";
 import { ActionType, createActionDeprecated } from "typesafe-actions";
 import { App } from "../shared/App";
@@ -55,8 +54,11 @@ const allActions = [
 
 export type AppsAction = ActionType<typeof allActions[number]>;
 
-export function getApp(releaseName: string, namespace: string) {
-  return async (dispatch: Dispatch): Promise<void> => {
+export function getApp(
+  releaseName: string,
+  namespace: string,
+): ThunkAction<Promise<void>, IStoreState, null, AppsAction> {
+  return async dispatch => {
     dispatch(requestApps());
     try {
       const app = await App.getRelease(namespace, releaseName);
@@ -67,8 +69,12 @@ export function getApp(releaseName: string, namespace: string) {
   };
 }
 
-export function deleteApp(releaseName: string, namespace: string, purge: boolean) {
-  return async (dispatch: Dispatch): Promise<boolean> => {
+export function deleteApp(
+  releaseName: string,
+  namespace: string,
+  purge: boolean,
+): ThunkAction<Promise<boolean>, IStoreState, null, AppsAction> {
+  return async dispatch => {
     try {
       await App.delete(releaseName, namespace, purge);
       return true;
@@ -83,7 +89,7 @@ export function fetchApps(
   ns?: string,
   all: boolean = false,
 ): ThunkAction<Promise<void>, IStoreState, null, AppsAction> {
-  return async (dispatch: Dispatch): Promise<void> => {
+  return async dispatch => {
     if (ns && ns === definedNamespaces.all) {
       ns = undefined;
     }
@@ -97,8 +103,6 @@ export function fetchApps(
   };
 }
 
-// TODO(miguel) This action creator is properly typed. We should do the same for the rest
-// https://github.com/kubeapps/kubeapps/issues/658
 export function deployChart(
   chartVersion: IChartVersion,
   releaseName: string,
@@ -131,8 +135,8 @@ export function upgradeApp(
   releaseName: string,
   namespace: string,
   values?: string,
-) {
-  return async (dispatch: Dispatch, getState: () => IStoreState): Promise<boolean> => {
+): ThunkAction<Promise<boolean>, IStoreState, null, AppsAction> {
+  return async (dispatch, getState) => {
     try {
       const {
         config: { namespace: kubeappsNamespace },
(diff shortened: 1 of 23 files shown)


dashboard/src/actions/apps.ts
```diff
@@ -55,8 +54,11 @@ const allActions = [
 
 export type AppsAction = ActionType<typeof allActions[number]>;
 
-export function getApp(releaseName: string, namespace: string) {
-  return async (dispatch: Dispatch): Promise<void> => {
+export function getApp(
+  releaseName: string,
+  namespace: string,
+): ThunkAction<Promise<void>, IStoreState, null, AppsAction> {
```
prydonius: why don't we alias this type so it's easier to reuse? e.g. we could create a `KubeappsThunkAction<void, AppsAction>` type, the first parameter being the return of the promise and the last being the action type.

dashboard/src/actions/auth.ts
```diff
@@ -27,22 +28,24 @@ const allActions = [setAuthenticated, authenticating, authenticationError];
...
+export function authenticate(
+  token: string,
+): ThunkAction<Promise<void>, IStoreState, null, AuthAction> {
+  return async dispatch => {
     dispatch(authenticating());
     try {
       await Auth.validateToken(token);
       Auth.setAuthToken(token);
-      return dispatch(setAuthenticated(true));
+      dispatch(setAuthenticated(true));
```
prydonius: I think the reason I decided to return this is because actions created by `createAction` also returned the action, so it was to have the same behaviour, but we are not really using it so I don't mind making this change.

dashboard/src/components/AppUpgrade/AppUpgrade.tsx
```diff
@@ -29,11 +29,11 @@ interface IAppUpgradeProps {
   ) => Promise<boolean>;
   clearRepo: () => any;
   checkChart: (repo: string, chartName: string) => any;
-  fetchChartVersions: (id: string) => Promise<{}>;
-  getApp: (releaseName: string, namespace: string) => Promise<void>;
-  getBindings: () => Promise<IServiceBindingWithSecret[]>;
-  getChartVersion: (id: string, chartVersion: string) => Promise<{}>;
-  getChartValues: (id: string, chartVersion: string) => Promise<any>;
+  fetchChartVersions: (id: string) => Promise<any>;
```
prydonius: shouldn't this be void also?

andresmgot: `fetchChartVersions` returns a list of `IChartVersions` and it's used in `src/actions/charts.ts`

dashboard/src/actions/apps.ts
andresmgot: I think it's a bit easier to understand if we leave this as `ThunkAction`. 

For example, reviewing this PR I needed to investigate what the parameters of `ThunkAction` meant. If we add `KubeappsThunkAction` first I would need to check what the parameters mean there and then how it's forwarded to `ThunkAction` and finally check the implementation of that. I would avoid that extra step if it's just to avoid typing the two parameters `IStoreState, null`.

migmartri: I do not have strong opinions, but I agree with @andresmgot that this thing is convoluted and complex enough that adding another layer of indirection will not help.

dashboard/src/actions/auth.ts
migmartri: They are different things no?

createAction returns an action creator that then you need to `dispatch` no? In this case we are returning a dispatched value?

dashboard/src/components/AppUpgrade/AppUpgrade.tsx
migmartri: Avoid what? The use of any? And what's the difference of returning an empty object.

To be clear, what I am trying to do here is to be sure that we use action creations to only dispatch values, not to rely on any internal returned value and use them outside of the redux scope. 

In this specific case though I believe that I could not remove the returned value because as @andresmgot mentioned, this function is called from another action creator and changing it will require a bigger change.

So basically I've tried to not to return anything for actions that are pure but maintain for now the others.

In any case, I am going to double check this one, thanks for pointing it out.

dashboard/src/actions/apps.ts
prydonius: yeah I agree that it's not that important, thanks!

dashboard/src/components/AppUpgrade/AppUpgrade.tsx
prydonius: > Avoid what? The use of any? And what's the difference of returning an empty object.

I think you misread, I was asking why this isn't `void`, but I see now that the return value is being used and we don't want to change this.

migmartri: Ahhh, sorry for the misunderstanding. Yes. that's the reason.

migmartri:
@andresmgot, @prydonius, thank you for your reviews.

Based on your comments I have changed a couple of things:

* For the function props passed to the components, I have properly typed them (many of them had => `:any`), thanks @andresmgot for the `Promise<IChartVersion[]>;` catch.
* For the ones that the action creator returned `Promise<void>` I have changed their type in the component to `void` directly, that way we basically are saying that the consumer should not do anything (async/await) with those. Thanks @prydonius for that suggestion.
* This patch does not change any returned value for those action creations, although in future iterations we should probably start refactoring them to in general not to depend of their returned value but instead use the global redux store dispatching transient events (i.e deploying)

PR #678, opened 2018-09-25
chart: update NOTES.txt so select first pod to port-forward

prydonius:
The `-o name` output lists both pods now that we have 2 replicas, which
the `port-forward` command does not like. This commit changes the NOTES
to use a jsonpath template to get the first pod's name only.

### DIFF
--- a/chart/kubeapps/templates/NOTES.txt
+++ b/chart/kubeapps/templates/NOTES.txt
@@ -41,7 +41,7 @@ To access Kubeapps from outside your K8s cluster, follow the steps below:
 {{- else if contains "ClusterIP" .Values.frontend.service.type }}
 
    echo "Kubeapps URL: http://127.0.0.1:8080"
-   export POD_NAME=$(kubectl get pods --namespace {{ .Release.Namespace }} -l "app={{ template "kubeapps.fullname" . }}" -o name)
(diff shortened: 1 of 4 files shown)


chart/kubeapps/templates/NOTES.txt
```diff
@@ -41,7 +41,7 @@ To access Kubeapps from outside your K8s cluster, follow the steps below:
 {{- else if contains "ClusterIP" .Values.frontend.service.type }}
 
    echo "Kubeapps URL: http://127.0.0.1:8080"
-   export POD_NAME=$(kubectl get pods --namespace {{ .Release.Namespace }} -l "app={{ template "kubeapps.fullname" . }}" -o name)
+   export POD_NAME=$(kubectl get pods --namespace {{ .Release.Namespace }} -l "app={{ template "kubeapps.fullname" . }}" -o jsonpath="{.items[0].metadata.name}")
```
prydonius: This is what @andresmgot originally had it as in https://github.com/kubeapps/kubeapps/pull/612/commits/52ab277bf8bef5cfaf06cbf76ba235e9e7f38723, but switched to the shorter `-o name`. However, now that we increased the replicas, that shorthand no longer works and we must do it this way.

Issue #656, opened 2018-09-19
Migrate action creators from typesafe-actions 1 to 2


migmartri:
Once this PR is merged https://github.com/kubeapps/kubeapps/pull/650/files#diff-5801116feaa1cd958f981f3cc8d6e21cL1 we will have a set of `deprecatedAction` methods that we will need to migrate.

See the different options here:

https://github.com/piotrwitek/typesafe-actions#v1xx-to-v2xx

PR #679, opened 2018-09-26
Migrate to new typesafe-actions action creators 

migmartri:
Moved to the new typesafe-actions `action creators` which enforce the usage of the `payload:` property  

I have also taken the opportunity to add the missing tests for our actions (except functions), going from:
```
Test Suites: 55 passed, 55 total
Tests:       229 passed, 229 total
Snapshots:   44 passed, 44 total
Time:        8.399s
```

selection_278

To:

```
Test Suites: 61 passed, 61 total
Tests:       314 passed, 314 total
Snapshots:   48 passed, 48 total
Time:        6.3s, estimated 7s
```
selection_279

The new tests include testing pure simple actions as well as in checking service calls during ThunkActions. All these during regular and erroed conditions.

Closes https://github.com/kubeapps/kubeapps/issues/656


migmartri:
> @migmartri getting this build error:
> 
> ```
> /app/src/actions/catalog.test.tsx
> (1,1): 'context' is declared but its value is never read.
> ```

It should be fixed now, the pre-commit hook did not catch it for some reason and it is not the first time that happens to me. https://github.com/kubeapps/kubeapps/issues/694

dashboard/src/reducers/repos.ts
```diff
@@ -46,41 +46,40 @@ const reposReducer = (
...
       return { ...state, form: { ...state.form, show: true } };
     case getType(actions.repos.hideForm):
       return { ...state, form: { ...state.form, show: false } };
     case getType(actions.repos.redirect):
-      return { ...state, redirectTo: action.path };
+      return { ...state, redirectTo: action.payload };
     case getType(actions.repos.redirected):
       return { ...state, redirectTo: undefined };
-    case getType(actions.repos.errorChart):
+    case getType(actions.charts.errorChart):
```
andresmgot: this seems like a bug (I know it's correct because it's the same action) but is there any reason for not using `actions.repos.errorChart`?

migmartri: I believe that I removed `actions.repos.errorChart` because it was duplicated. Any reason why you would like to have errorChart inside the repoActions?

dashboard/src/actions/config.test.tsx
```diff
@@ -0,0 +1,33 @@
+import configureMockStore from "redux-mock-store";
+import thunk from "redux-thunk";
+import { getType } from "typesafe-actions";
+import actions from ".";
+import Config from "../shared/Config";
+
+const mockStore = configureMockStore([thunk]);
+
+let store: any;
+const CONFIG = "have you tried to turn it off and on again";
```
prydonius: why all caps? why not call this `testConfig` or `mockConfig`?

dashboard/src/reducers/repos.ts
prydonius: Hmm, shouldn't they be different? Doesn't this now mean that both the charts and repos reducers will pick up this action and set this? Is that what we desire?

dashboard/src/actions/config.test.tsx
migmartri: Because of a brain fart due to to many languages. In case of doubt I tend to check this https://github.com/airbnb/javascript#naming--uppercase, in which case is telling me to indeed remove the case. 

Thanks!

dashboard/src/reducers/repos.ts
migmartri: Maybe, I do not know. The only thing that I've done is to remove duplicated code.

[This one](https://github.com/kubeapps/kubeapps/pull/679/files#diff-f14bc38d0d5e1089c558504b84144608L27), although referenced by a different variable name, it sends the same action.type `ERROR_CHART`, than https://github.com/kubeapps/kubeapps/pull/679/files#diff-32e6760e59b5e0433a9fae60151e0a8dR19

Both reducers do something with this action, inside their own namespace `repos` or `charts`, I do not know if that's what we want, it is what we had.

If you want me to dig deeper into this let me know, but I think that this will be potentially solved by a general purpose error review and handling.

prydonius: I'm not sure if createAction was creating a special type that was unique even if it had the same type name though? It's possible that getType does something special to ensure that even though they are duplicates, they are different -- although I could also be making stuff up :P

migmartri: >  although I could also be making stuff up :P

You are indeed. It does not do that check afaict. Good point though.

prydonius: Thanks for checking!

PR #681, opened 2018-09-26
Refactor purge UX for deleted apps

### LINKED ISSUE #655: Purge should not show the confirm dialog
When we show the purge button, we should show a confirmation dialog but the only action should be just that, purge.

Today, when you click purge you get exactly the same behavior than a regular deletion

selection_708

selection_709
and if you do not choose `purge` it fails.

selection_710

andresmgot:
Fixes #655

Don't show the `purge` check box when the application has been already deleted. Force to purge it if the user deletes it again.

### DIFF
--- a/dashboard/src/components/AppView/AppControls.test.tsx
+++ b/dashboard/src/components/AppView/AppControls.test.tsx
@@ -107,3 +107,45 @@ context("when name or namespace do not exist", () => {
     props: { ...props, app: { ...props.app, namespace: null } },
   });
 });
+
+context("when the application has been already deleted", () => {
+  const props = {
+    app: new hapi.release.Release({ name: "name", namespace: "my-ns", info: { deleted: {} } }),
+    deleteApp: jest.fn(() => false), // Return "false" to avoid redirect when mounting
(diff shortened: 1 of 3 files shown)


dashboard/src/components/AppView/AppControls.test.tsx
```diff
@@ -107,3 +107,45 @@ context("when name or namespace do not exist", () => {
...
+    ReactModal.setAppElement(document.createElement("div"));
+    wrapper.setState({ modalIsOpen: true, purge: false });
+    wrapper.update();
+
+    const confirm = wrapper.find(ConfirmDialog);
+    expect(confirm.exists()).toBe(true);
+
+    // Check that the "purge" is forwarded to deleteApp
+    confirm.props().onConfirm(); // Simulate confirmation
+    expect(deleteApp.mock.calls[0]).toEqual([true]);
```
migmartri: Maybe you can use this more explicit helper?

Something like 
```
expect(confirm.props().deleteApp).toHaveBeenCalledWith(true)
```

Issue #668, opened 2018-09-21
Too much noise, too many pods


migmartri:
My namespace is polluted with many instances of the sync jobs. It seems that the list grows on every chart upgrade. It would be great if we could limit the number of pods kept for reference.
```
repo-sync-bitnami-tbr7j-6bvkn                                  0/1       Completed   0          21h
apprepo-sync-bitnami-tt4mj-tc9fl                                  0/1       Completed   0          8d
apprepo-sync-bitnami-vv2nl-kdq9f                                  0/1       Completed   0          1d
apprepo-sync-bitnami-z22l7-wj2xq                                  0/1       Completed   0          1d
apprepo-sync-incubator-1537552800-flk79                           0/1       Completed   0          2h
apprepo-sync-incubator-1537556400-kf5pf                           0/1       Completed   0          1h
apprepo-sync-incubator-1537560000-4jtz4                           0/1       Completed   0          29m
apprepo-sync-incubator-2tz4j-jfwrv                                0/1       Completed   0          1h
apprepo-sync-incubator-2vch8-wczsj                                0/1       Completed   0          3m
apprepo-sync-incubator-6b2d9-kqqxq                                0/1       Completed   0          21h
apprepo-sync-incubator-84lwv-9m29p                                0/1       Completed   0          22h
apprepo-sync-incubator-bm5bt-vn976                                0/1       Completed   0          1m
apprepo-sync-incubator-hgxds-54hwx                                0/1       Completed   0          8d
apprepo-sync-incubator-kzhwf-h624b                                0/1       Completed   0          8d
apprepo-sync-incubator-lk7fm-mg9x6                                0/1       Completed   0          3d
apprepo-sync-incubator-qh7nj-q8ssj                                0/1       Completed   0          21h
apprepo-sync-incubator-qj6hh-nbwgw                                0/1       Completed   0          1d
apprepo-sync-incubator-smvcv-h4zhh                                0/1       Completed   0          1d
apprepo-sync-incubator-ts5hv-zt2tr                                0/1       Completed   0          1h
apprepo-sync-incubator-xkzgn-db8fb                                0/1       Completed   0          5m
apprepo-sync-stable-1537552800-qrzhz                              0/1       Completed   0          2h
apprepo-sync-stable-1537556400-2tdkv                              0/1       Completed   0          1h
apprepo-sync-stable-1537560000-m5zsf                              0/1       Completed   0          29m
[...]

PR #691, opened 2018-09-27
Simplify upgrade process

### LINKED ISSUE #628: update repository list
step 1 deploy with helm kubeapps with a list of repo
step 2 update with helm kubeapps with same list of repo + one new

When upgrade the list of repository in chart deployment kubeapps, the list in ui is not updated.

### LINKED ISSUE #665: Improve Kubeapps chart upgrade experience
Upgrading the chart requires that you pass the previous Mongodb password, otherwise the newly upgrade release will not be able to connect to MongoDB.

We [have it documented](https://github.com/kubeapps/kubeapps/tree/master/chart/kubeapps#upgrading-kubeapps) but it seems [common for users](https://kubernetes.slack.com/archives/C9D3TSUG4/p1537474402000100?thread_ts=1537467768.000100&cid=C9D3TSUG4) to miss this additional step, probably mainly because they were not asked to do anything during install.

We should find away to avoid this pitfall

* Is there any way we could detect that the password was not passed and let the user know during upgrade?
* Maybe asking the user for the password during installation?
* Do not autogenerate a password?

andresmgot:
Fixes #665 

The main change is that the current values now include `mongodb.existingSecret` set to `kubeapps-mongodb`. This means that the mongodb secret is no longer managed by Helm templates. Instead of that, we create/delete it with Helm hooks.

When installing the chart we generate the secret using a random password. When deleting the chart we delete the secret as well. For avoiding breaking changes, we have added a check in the `post-upgrade` hook to ensure that the mongodb secret exists (which won't be the case when upgrading from a previous version). If the secret doesn't exists we create it, we restart mongodb and we re-sync the existing app-repositories.


chart/kubeapps/templates/kubeapps-jobs-upgrade.yaml
```diff
@@ -0,0 +1,61 @@
...
+            if kubectl get secrets -n {{ .Release.Namespace }} {{ template "kubeapps.mongodb-secret.fullname" . }}; then
+              # Secret exists, do nothing
+              echo "MongoDB secret found"
+            else
+              # Secret doesn't exists, create it and restart mongodb
+              kubectl create secret generic -n {{ .Release.Namespace }} {{ template "kubeapps.mongodb-secret.fullname" . }} --from-literal mongodb-root-password=`cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1`
+              kubectl delete pods -n {{ .Release.Namespace }} -l app=mongodb,release={{ .Release.Name }}
+            fi
+            # Re-sync repositories
+            kubectl apply -f /tmp/apprepositories/apprepositories.yaml
```
migmartri: I am a little bit worried that this will make this issue even worse https://github.com/kubeapps/kubeapps/issues/668. It might make sense to bump it to 1.0? cc/ @andresmgot @prydonius ?

chart/kubeapps/values.yaml
```diff
@@ -188,6 +188,8 @@ mongodb:
   # Kubeapps uses MongoDB as a cache and persistence is not required
   persistence:
     enabled: false
+  # MongoDB credentials are handled by kubeapps to facilitate upgrades
+  existingSecret: kubeapps-mongodb
```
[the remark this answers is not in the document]
prydonius: Hmm we would have to hardcode the name of the secret, I don't think there is another way... That would limit Kubeapps to being installed once per namespace.

chart/kubeapps/templates/_helpers.tpl
```diff
@@ -68,6 +68,34 @@ Create name for the apprepository cleanup job
 {{ template "kubeapps.fullname" . }}-internal-apprepository-jobs-cleanup
 {{- end -}}
 
+{{/*
+Create name for the mongodb secret name
+*/}}
+{{- define "kubeapps.mongodb-secret.fullname" -}}
+{{ .Release.Name }}-mongodb
```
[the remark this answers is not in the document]
prydonius: Actually I think this needs to be hardcoded, I think we need to use `.Chart.Name` instead of `.Release.Name`.

chart/kubeapps/README.md
```diff
@@ -50,7 +50,9 @@ The command deploys Kubeapps on the Kubernetes cluster in the `kubeapps` namespa
 
 ## Upgrading Kubeapps
 
-To upgrade Kubeapps, first ensure you have the updated your local chart repository cache:
+You can upgrade Kubeapps from the Kubeapps web interface. Select the namespace in which kubeapps is installed (`kubeapps` if you followed the instructions in this guide) and click in the "Upgrade" button. Select the new version and confirm.
```
prydonius: capitalise Kubeapps in "in which kubeapps is installed"

chart/kubeapps/README.md
```diff
@@ -50,7 +50,9 @@ The command deploys Kubeapps on the Kubernetes cluster in the `kubeapps` namespa
...
+
+It's possible to also upgra Kubeapps using the HELM CLI. For doing so first upgrade the existing repositories:
```
prydonius: I would change this to:
> You can also use the Helm CLI to upgrade Kubeapps, first ensure you have the updated your local chart repository cache:

andresmgot:
> What if we just created another migration guide to help people move the secret and bump the major version of the chart? I think it's low-impact right now due to the amount of users.

I'd rather avoid having to point users to a migration guide for a minor version. Also I like the current approach because it fixes #628.

prydonius:
As discussed at the standup:

- remove refreshing of initialRepos, and instead refresh the current apprepos in the namespace (patch to increase resyncRequests or something)
- create an issue to remove the upgrade job in a future chart release

chart/kubeapps/README.md
```diff
@@ -50,7 +50,11 @@ The command deploys Kubeapps on the Kubernetes cluster in the `kubeapps` namespa
 
 ## Upgrading Kubeapps
 
-To upgrade Kubeapps, first ensure you have the updated your local chart repository cache:
+You can upgrade Kubeapps from the Kubeapps web interface. Select the namespace in which Kubeapps is installed (`kubeapps` if you followed the instructions in this guide) and click in the "Upgrade" button. Select the new version and confirm.
```
prydonius: click _on_ the "Upgrade" button

chart/kubeapps/templates/_helpers.tpl
```diff
@@ -68,6 +68,34 @@ Create name for the apprepository cleanup job
...
+{{- end -}}
+
+{{/*
+Create name for the mongodb secret bootstrap job
+*/}}
+{{- define "kubeapps.mongodb-secret-bootstrap.fullname" -}}
```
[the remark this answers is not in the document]
prydonius: we should change the name of the template here to match as well

chart/kubeapps/templates/_helpers.tpl
```diff
@@ -68,6 +68,27 @@ Create name for the apprepository cleanup job
...
+{{ template "kubeapps.fullname" . }}-internal-mongodb-jobs-boostrap
+{{- end -}}
+
+{{/*
+Create name for the mongodb secret bootstrap job
+*/}}
+{{- define "kubeapps.mongodb-secret-cleanup.fullname" -}}
```
prydonius: mongodb-jobs-cleanup

chart/kubeapps/templates/_helpers.tpl
```diff
@@ -68,6 +68,27 @@ Create name for the apprepository cleanup job
...
+{{ template "kubeapps.fullname" . }}-internal-mongodb-jobs-cleanup
+{{- end -}}
+
+{{/*
+Create name for the kubeapps upgrade job
+*/}}
+{{- define "kubeapps.kubeapps-jobs-upgrade.fullname" -}}
+{{ template "kubeapps.fullname" . }}-internal-kubeapps-upgrade-job
```
prydonius: kubeapps-jobs-upgrade

chart/kubeapps/templates/mongodb-secret-bootstrap-rbac.yaml
```diff
@@ -0,0 +1,45 @@
+{{- if .Values.rbac.create -}}
```
prydonius: we should rename the filenames to match the naming, i.e. mongodb-jobs-bootstrap-rbac here

chart/kubeapps/templates/mongodb-secret-bootstrap-serviceaccount.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
```
prydonius: we should rename the filenames to match the naming, i.e. mongodb-jobs-bootstrap-serviceaccount here

chart/kubeapps/templates/mongodb-secret-bootstrap.yaml
```diff
@@ -0,0 +1,46 @@
+apiVersion: batch/v1
```
prydonius: we should rename the filenames to match the naming, i.e. mongodb-jobs-bootstrap here

chart/kubeapps/templates/mongodb-secret-bootstrap-rbac.yaml
```diff
@@ -0,0 +1,45 @@
...
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - create
+  - delete
```
prydonius: why does it need to delete?

chart/kubeapps/templates/mongodb-secret-bootstrap.yaml
```diff
@@ -0,0 +1,46 @@
...
+      labels:
+        app: {{ template "kubeapps.mongodb-secret-bootstrap.fullname" . }}
+        release: {{ .Release.Name }}
+    spec:
+      containers:
+      - name: kubectl
+        image: {{ template "kubeapps.image" .Values.hooks.image }}
+        command:
+        - kubectl
+        - create
```
prydonius: what happens if the secret happens to exist? this will fail?

chart/kubeapps/templates/mongodb-secret-cleanup.yaml
```diff
@@ -0,0 +1,43 @@
...
+    metadata:
+      labels:
+        app: {{ template "kubeapps.mongodb-secret-cleanup.fullname" . }}
+        release: {{ .Release.Name }}
+    spec:
+      containers:
+      - name: kubectl
+        image: {{ template "kubeapps.image" .Values.hooks.image }}
+        command:
+        - kubectl
```
prydonius: what happens if the secret somehow doesn't exist, this will fail? Should we account for that?

chart/kubeapps/templates/kubeapps-jobs-upgrade.yaml
andresmgot: I'll move the repository sync just in case the mongodb pod is restarted.

chart/kubeapps/templates/mongodb-secret-bootstrap.yaml
andresmgot: yes, if the secret already exists the job execution will fail but the hook will success because it can create the job. Because of that the job will be deleted.

chart/kubeapps/templates/mongodb-secret-cleanup.yaml
andresmgot: The same than the bootstrap. The job execution will fail but it will be cleanup anyway.

chart/kubeapps/templates/mongodb-secret-bootstrap-rbac.yaml
andresmgot: 👍 it doesn't

prydonius:
> yes, if the secret already exists the job execution will fail but the hook will success because it can create the job. Because of that the job will be deleted afterwards.

@andresmgot hmm are you sure? AFAIK Jobs are a special type of hook where Helm will wait for it to be created.

Also, now that I think about it, why not have the Secret be the actual pre-install Hook instead of creating a Job to create it? Wouldn't that be a lot easier?

andresmgot:
> AFAIK Jobs are a special type of hook where Helm will wait for it to be created.

I'd swear I've seen the job failing and being deleted afterwards but checking it you are right.

> Also, now that I think about it, why not have the Secret be the actual pre-install Hook instead of creating a Job to create it? Wouldn't that be a lot easier?

True, we need to keep the cleanup job though.

chart/kubeapps/README.md
```diff
@@ -61,12 +65,9 @@ Now upgrade Kubeapps:
 ```console
 $ export RELEASE_NAME=kubeapps
 $ export NAMESPACE=kubeapps
-$ export MONGODB_ROOT_PASSWORD=$(kubectl get secret --namespace $NAMESPACE $RELEASE_NAME-mongodb -o jsonpath="{.data.mongodb-root-password}" | base64 --decode)
```
migmartri: Nice :)

chart/kubeapps/values.yaml
```diff
@@ -87,6 +82,17 @@ apprepository:
   tolerations: []
   affinity: {}
 
+# Hooks are used to perform actions like populating
+# apprepositories or creating required resources
```
migmartri: maybe add to the comment, 

> apprepositories or creating required resources

... during installation or upgrade

chart/kubeapps/README.md
```diff
@@ -50,7 +50,11 @@ The command deploys Kubeapps on the Kubernetes cluster in the `kubeapps` namespa
 
 ## Upgrading Kubeapps
 
-To upgrade Kubeapps, first ensure you have the updated your local chart repository cache:
+You can upgrade Kubeapps from the Kubeapps web interface. Select the namespace in which Kubeapps is installed (`kubeapps` if you followed the instructions in this guide) and click on the "Upgrade" button. Select the new version and confirm.
+
+> NOTE: If the chart values were modified when deploying Kubeapps the first time, those values need to be set again when upgrading.
```
migmartri: Thank you

chart/kubeapps/templates/mongodb-secret-bootstrap.yaml
```diff
@@ -0,0 +1,14 @@
...
+  name: {{ .Values.mongodb.existingSecret }}
+  annotations:
+    helm.sh/hook: pre-install
+  labels:
+    chart: {{ template "kubeapps.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+data:
+  mongodb-root-password: {{ randAlphaNum 10 | b64enc | quote }}
+{{- end -}}
```
migmartri: add new line to the end of the file

chart/kubeapps/templates/kubeapps-jobs-upgrade.yaml
```diff
@@ -0,0 +1,52 @@
+apiVersion: batch/v1
```
migmartri: I think we should add a comment in this file with:

* An explanation on why this is needed, i.e `upgrade from a previous version that did not support existing secret`
* What it does, create secret, sync repos.
* Mention that this is temporary and add a link to a task to remove it

chart/kubeapps/templates/kubeapps-jobs-upgrade.yaml
```diff
@@ -0,0 +1,52 @@
...
+            # Recreate MongoDB secret if doesn't exists (happens for chart versions < 0.5.2)
+            if kubectl get secrets -n {{ .Release.Namespace }} {{ .Values.mongodb.existingSecret }}; then
+              # Secret exists, do nothing
+              echo "MongoDB secret found"
+            else
+              # Secret doesn't exists, create it and restart mongodb
+              kubectl create secret generic -n {{ .Release.Namespace }} {{ .Values.mongodb.existingSecret }} --from-literal mongodb-root-password=`cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1`
+              kubectl delete pods -n {{ .Release.Namespace }} -l app=mongodb,release={{ .Release.Name }}
+              # Re-sync repositories reseting resyncRequests
+              kubectl get apprepositories -o=name -n kubeapps {{ .Release.Namespace }} | xargs kubectl patch $1 -n {{ .Release.Namespace }} --type merge -p '{ "spec": { "resyncRequests": 0 } }'
```
migmartri: Does the controller retry and so on if mongodb is down? Note that a new mongodb image will need to be downloaded so from `k delete pod mongodb` to being ready it might take some time. Time that the sync could fail.

andresmgot: I added a check for the rollout status

andresmgot:
@prydonius PTAL I think all the comments have been addressed.

Issue #674, opened 2018-09-24
[Implementation] Revisit application access URL


arapulido:

migmartri:
Refs https://github.com/kubeapps/kubeapps/issues/492 https://github.com/kubeapps/kubeapps/issues/286

PR #697, opened 2018-10-02
Refactor URL handling (for services)

andresmgot:
First part of: #674

This PR refactors how we handle the URL table in the AppView only for services. A different PR will be created for adding Ingress objects.

The `ServiceTable` item that previously handled the URLs view and the service table at the bottom has been split so the URL handling happens now in a different component `AccessURLTable`.

The `ServiceItem.scss` has been renamed since it handles the rendering of the URL table.

There is a specific pain point in this PR: The state property `services` of `AppView` (as well as `deployments` and `otherResources`) is set as `Map<string,IResource>` but instead of using the native methods of a Typescript Map we were manually adding/reading elements like a Javascript object. This causes some issues like `services.size` to not work. I changed how we set/read those elements. If there is a reason for not doing so please let me know. In other case we should change `deployments` and `otherResources` too.


dashboard/src/components/AppView/AccessURLItem.scss
```diff
@@ -0,0 +1,13 @@
+.ServiceItem__url a {
```
migmartri: Maybe these two can be in the same one, `a` being inline.

dashboard/src/components/AppView/AccessURLTable.tsx
```diff
@@ -0,0 +1,43 @@
+import * as React from "react";
```
migmartri: I'd like to start actually adding these components inside their own directory, directory that will contain its own index.js as well.

So for example `AppView/AccessURLTable.tsx` will go inside `AppView/AccessURLTable/AccessURLTable.tsx`

I'd like to make it standard across all the components but for now I'll be happy if we do it at least for new components.

dashboard/src/components/AppView/AccessURLTable.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+import { IResource, IServiceSpec } from "../../shared/types";
+import AccessURLItem from "./AccessURLItem";
+
+interface IServiceTableProps {
+  services: Map<string, IResource>;
+}
+
+class AccessURLTable extends React.Component<IServiceTableProps> {
```
migmartri: Could we make this a stateless component?

dashboard/src/components/AppView/AccessURLTable.tsx
```diff
@@ -0,0 +1,43 @@
...
+          </thead>
+          <tbody>
+            {publicServices.map((k: string) => (
+              <AccessURLItem key={k} service={services.get(k) as IResource} />
+            ))}
+          </tbody>
+        </table>
+      );
+    } else {
+      return null;
```
migmartri: is this needed?

dashboard/src/components/AppView/AccessURLTable.tsx
```diff
@@ -0,0 +1,43 @@
...
+          <thead>
+            <tr>
+              <th>NAME</th>
+              <th>TYPE</th>
+              <th>URL</th>
+            </tr>
+          </thead>
+          <tbody>
+            {publicServices.map((k: string) => (
+              <AccessURLItem key={k} service={services.get(k) as IResource} />
```
migmartri: Do you need this `as IResource`, isn't supposed to be defined in the map?

dashboard/src/components/AppView/AccessURLTable.tsx
```diff
@@ -0,0 +1,43 @@
...
+  public render() {
+    const { services } = this.props;
+    const publicServices: string[] = [];
+    services.forEach((svc, key) => {
+      const spec = svc.spec as IServiceSpec;
+      if (spec.type === "LoadBalancer") {
+        publicServices.push(key);
+      }
+    });
+    if (publicServices.length > 0) {
```
migmartri: Just in case, this could be written as
```
return publicServices.length > 0 && (<table>...
```

dashboard/src/components/AppView/AppDetails.test.tsx
```diff
@@ -29,7 +29,7 @@ it("renders a deployment details", () => {
 it("renders a services details", () => {
   const services = new Map<string, IResource>();
   const svc = "foo";
-  services[svc] = {
+  services.set(svc, {
```
migmartri: What has changed here that forced you to do this?

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -136,7 +136,7 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
         this.setState({ deployments: { ...this.state.deployments, [key]: resource } });
         break;
       case "Service":
-        this.setState({ services: { ...this.state.services, [key]: resource } });
+        this.setState({ services: this.state.services.set(key, resource) });
```
migmartri: So we were transforming the map into an object? Is the change to preserve the type? Why did you change it?

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -194,7 +193,7 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
                     <AppControls app={app} deleteApp={this.deleteApp} />
                   </div>
                 </div>
-                <ServiceTable services={this.state.services} extended={false} />
+                {this.state.services.size > 0 && <AccessURLTable services={this.state.services} />}
```
migmartri: Cool

dashboard/src/components/AppView/ServiceTable.tsx
```diff
@@ -1,44 +1,31 @@
...
-
   public render() {
-    const { extended, services } = this.props;
+    const { services } = this.props;
+    const svcList: JSX.Element[] = [];
+    services.forEach((svc: IResource, k: string) =>
+      svcList.push(<ServiceItem key={k} service={svc} />),
+    );
     return (
       <table>
```
migmartri: do you want to check that `svcList` is empty before rendering the headers?

dashboard/src/components/AppView/AccessURLItem.tsx
```diff
@@ -0,0 +1,84 @@
...
+  URLs: string[];
+}
+
+class AccessURLItem extends React.Component<IAccessURLItem, IAccessURLState> {
+  public state: IAccessURLState = {
+    isLink: false,
+    URLs: [],
+  };
+
+  public componentDidMount() {
```
migmartri: I'd add a comment describing what this method does.

dashboard/src/components/AppView/AccessURLItem.tsx
```diff
@@ -0,0 +1,84 @@
...
+  service: IResource;
+}
+
+interface IAccessURLState {
+  isLink: boolean;
+  URLs: string[];
+}
+
+class AccessURLItem extends React.Component<IAccessURLItem, IAccessURLState> {
+  public state: IAccessURLState = {
```
migmartri: Do you need state? neither URL nor isLink seem to change so state is not needed no?

dashboard/src/components/AppView/AccessURLItem.tsx
migmartri: See my comment about not needing state and related to that, I do not think that you need to use componentDidMount, you can just have two methods, or two getters, or one returning a struct.

dashboard/src/components/AppView/AccessURLItem.tsx
```diff
@@ -0,0 +1,84 @@
...
+            </span>
+          ))}
+        </td>
+      </tr>
+    );
+  }
+
+  private getURL(base: string, port: number) {
+    const protocol = port === 443 ? "https" : "http";
+    const portText = port === 443 || port === 80 ? "" : `:${port}`;
```
migmartri: I'd also add a comment explaining the reasoning about showing it sometimes.

Maybe, call it portSuffix?

dashboard/src/components/AppView/AccessURLItem.tsx
```diff
@@ -0,0 +1,84 @@
...
+    return (
+      <tr>
+        <td>{service.metadata.name}</td>
+        <td>{service.spec.type}</td>
+        <td>
+          {URLs.map(l => (
+            <span
+              key={l}
+              className={`${
+                isLink ? "ServiceItem__url" : "ServiceItem__not-url"
```
migmartri: You do not need two classNames.

dashboard/src/components/AppView/AccessURLItem.scss
prydonius: +1, also I think there should be some shared properties here. Maybe this should be:

```
.ServiceItem {
  display: inline-block;
  color: #fff;
  border-radius: 16px;
  &__url {
    background-color: #1598cb;
  }
  &__not-url {
    background-color: #b3b3b3;
  }
}
```

Also just to check, are these colours from the Hex palette?

dashboard/src/components/AppView/AccessURLItem.test.tsx
```diff
@@ -0,0 +1,109 @@
...
+});
+
+describe("when the status is populated", () => {
+  interface Itest {
+    description: string;
+    ports: any[];
+    ingress: any[];
+    expectedURLs: string[];
+  }
+  const tests: Itest[] = [
```
prydonius: I love how clear this is, thanks!

dashboard/src/components/AppView/AccessURLItem.tsx
prydonius: Agreed, this can just be evaluated in render

dashboard/src/components/AppView/AccessURLItem.tsx
prydonius: @migmartri what do you mean?

dashboard/src/components/AppView/AccessURLItem.tsx
```diff
@@ -0,0 +1,84 @@
...
+              } type-small margin-r-small padding-tiny padding-h-normal`}
+            >
+              {isLink ? (
+                <a className="padding-tiny padding-h-normal" href={l} target="_blank">
```
prydonius: I don't think the `a` should be inside the `span`, I think we should either show the `span` or the `a` depending on `isLink`. That way we can use `<a className="ServiceItem ServiceItem__url"...` and `<span className="ServiceItem ServiceItem__not-url...`.

dashboard/src/components/AppView/AppDetails.test.tsx
prydonius: Hmm, I torn on this, the Map type is not a real JavaScript type (atleast in ES5) and I prefer accessing/setting as if it's a normal JavaScript object over using set/get, but that's just my preference. I don't have a strong opinion though.

andresmgot:
@migmartri @prydonius PTAL I applied your requests. BTW for you to know the CSS changes this is the result (for link, link:hover and not link):
screen shot 2018-10-03 at 17 24 31

dashboard/src/components/AppView/AccessURLTable/AccessURLItem/AccessURLItem.scss
```diff
@@ -0,0 +1,18 @@
+.ServiceItem {
+  display: inline-block;
+  color: #fff;
+  border-radius: 16px;
+  background-color: #5f6369;
+  &.with-link {
```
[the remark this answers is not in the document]
andresmgot: you are right, better to call it `--with-link` since it's a modifier

dashboard/src/components/AppView/AppView.test.tsx
```diff
@@ -137,13 +138,22 @@ describe("AppViewComponent", () => {
 
   describe("renderization", () => {
     it("renders all the elements of an application", () => {
-      const wrapper = shallow(<AppViewComponent {...validProps} />);
+      const wrapper = mount(<AppViewComponent {...validProps} />);
```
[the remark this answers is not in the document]
andresmgot: Because if not, the shallow doesn't render enough to check that there is a `ServiceTable` there.

dashboard/src/components/AppView/AccessURLTable/AccessURLItem/AccessURLItem.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { IResource, IServiceSpec, IServiceStatus } from "../../../../shared/types";
+import "./AccessURLItem.css";
+
+interface IAccessURLItem {
+  service: IResource;
+}
+
+class AccessURLItem extends React.Component<IAccessURLItem> {
+  // isLink returns true if there are any link in the Item
+  get isLink(): boolean {
```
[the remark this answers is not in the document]
migmartri: >  the Service IResource should

I do not agree with that statement, isLink is very specific to this component rendering logic so I can not see how it would fit in the model.

W.R.T the getters, I was the one that suggested it to @andresmgot and we just had a chat about it. I personally have not put much thought on it but he convinced me that actually using a function (as you also suggested) is better so it does not get mistaken by a static property.

About making them private functions, that means that we will not (and should not) test them, I guess that's ok.

dashboard/src/components/AppView/AccessURLTable/AccessURLItem/AccessURLItem.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  public render() {
+    const { service } = this.props;
+    const isLink = this.isLink;
+    return (
+      <tr>
+        <td>{service.metadata.name}</td>
+        <td>{service.spec.type}</td>
+        <td>
+          {this.URLs.map(l => (
```
migmartri: instead of `l` I'd have called it `url` to be more descriptive.

dashboard/src/components/AppView/AccessURLTable/AccessURLItem/AccessURLItem.tsx
```diff
@@ -0,0 +1,81 @@
...
+                l
+              )}
+            </span>
+          ))}
+        </td>
+      </tr>
+    );
+  }
+
+  private getURL(base: string, port: number) {
```
migmartri: Can you add a comment saying what it does?

dashboard/src/components/AppView/AccessURLTable/AccessURLTable.tsx
```diff
@@ -0,0 +1,46 @@
...
+          <thead>
+            <tr>
+              <th>NAME</th>
+              <th>TYPE</th>
+              <th>URL</th>
+            </tr>
+          </thead>
+          <tbody>
+            {this.publicServices.map((k: string) => (
+              <AccessURLItem key={k} service={services[k]} />
```
migmartri: rename `service` prop for `load-balancer-service`

dashboard/src/components/AppView/AccessURLTable/AccessURLItem/AccessURLItem.tsx
```diff
@@ -0,0 +1,81 @@
+import * as React from "react";
+
+import { IResource, IServiceSpec, IServiceStatus } from "../../../../shared/types";
+import "./AccessURLItem.css";
+
+interface IAccessURLItem {
+  service: IResource;
```
migmartri: rename to `load-balancer-service`

dashboard/src/components/AppView/AccessURLTable/AccessURLItem/AccessURLItem.tsx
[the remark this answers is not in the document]
prydonius: > I do not agree with that statement, isLink is very specific to this component rendering logic so I can not see how it would fit in the model.

Why not? Shouldn't it be possible to generally tell if an IService has a link? It is a function that acts directly on the IService spec to figure it out, the component has to use the prop to get that value. In any case, I think it's arguable but I agree and I think my bigger concern here was not using getters/setters because they look too much like static properties.

Good point about private functions.

Issue #694, opened 2018-09-27
tslint should catch errors during pre-commit


migmartri:
For some reason, errors that will cause a runtime error during `yarn run start` or a build error `yarn build` are not caught in our pre-commit hooks.

prydonius:
See also https://github.com/kubeapps/kubeapps/pull/679#issuecomment-425273321

Issue #158, opened 2018-03-06
Upgrade husky hooks


prydonius:
_From @prydonius on February 14, 2018 8:1_

We're currently locking to 0.14.3 because the hooks from newer versions are not working properly. We need to look into why that's not working and upgrade the dependency.

_Copied from original issue: kubeapps/dashboard#64_

prydonius:
cc @evanlouie @sozercan

prydonius:
@migmartri can you look into whether you updated this?

PR #700, opened 2018-10-04
Add typescript compilation check and hooks update

migmartri:
The issue with tslint is that it actually does not check that the compilation works, from [their documentation](https://palantir.github.io/tslint/usage/cli)

> Please ensure that the TypeScript source files compile correctly before running the linter.

So I've added a compilation step during the pre-hook. Please note that it was added in the husky hooks not in the `lint-staged`. This is because the compilation does not depend on the changed files (but on the tsconfig.json) so it can not be used there. More info [here](https://github.com/okonet/lint-staged/issues/468)

Closes https://github.com/kubeapps/kubeapps/issues/694 
Closes https://github.com/kubeapps/kubeapps/issues/158

### DIFF
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -58,6 +58,7 @@
   },
   "lint-staged": {
     "*.{ts,tsx}": [
+      "tsc -p tsconfig.json --noEmit",
       "yarn run lint-js-fix",
       "prettier --write",
       "git add"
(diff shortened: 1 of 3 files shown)


prydonius:
Won't this slow down the time to commit significantly?

andresmgot:
> Won't this slow down the time to commit significantly?

The command takes 2s in my local env. I think the prettier check takes more.

migmartri:
5 seconds here. I think it is acceptable. If we change our mind we can always revert it. 

```
time yarn run ts-compile-check
yarn run v1.10.1
$ tsc -p tsconfig.json --noEmit
Done in 5.50s.
```

Issue #610, opened 2018-09-10
Add Slack info to our readme


migmartri:
Kubeapps has a [public Slack channel](https://kubernetes.slack.com/messages/kubeapps
) in the Kubernetes organization that we might want to link from the readme file. In a similar way Minikube, or Helm [do it](https://github.com/kubernetes/minikube#community) 

If we want to add a badge, there does not seem to be a native way to do it but there seems to be some [workarounds](https://www.quora.com/How-can-I-create-a-markdown-badge-for-a-GitHub-repo-which-links-to-a-Slack-channel-workspace)

PR #701, opened 2018-10-04
Add Slack link to Readme

migmartri:
Followed a similar approach as seen in other projects like Minikube.

NOTE: I decided not to add this badge but let me know if you think otherwise

[Slack](https://kubernetes.slack.com/messages/C9D3TSUG4)

Closes https://github.com/kubeapps/kubeapps/issues/610

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 
 [![CircleCI](https://circleci.com/gh/kubeapps/kubeapps/tree/master.svg?style=svg)](https://circleci.com/gh/kubeapps/kubeapps/tree/master)
 
+
 Kubeapps is a web-based UI for deploying and managing applications in Kubernetes clusters. Kubeapps allows you to:
 
 - Browse and deploy [Helm](https://github.com/helm/helm) charts from chart repositories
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -94,3 +94,7 @@ For a more detailed and step-by-step introduction to Kubeapps, read our [introdu
 * [Detailed installation instructions](chart/kubeapps/README.md)
 * [Kubeapps Dashboard documentation](docs/user/dashboard.md)
 * [Kubeapps components](docs/architecture/overview.md)
+
+## Community
+
+* [#kubeapps on Kubernetes Slack](https://kubernetes.slack.com/messages/C9D3TSUG4)
```
prydonius: Helm links to `https://kubernetes.slack.com/messages/helm-users` so we should be able to use `https://kubernetes.slack.com/messages/kubeapps`. Also we should instruct people on how to join this slack (link to slack.k8s.io, as it's not obvious.

andresmgot: +1

migmartri: > Helm links to https://kubernetes.slack.com/messages/helm-users so we should be able to use https://kubernetes.slack.com/messages/kubeapps.

The reason why I used the ID is because it seems that sometimes using `/kubeapps` the link gets lost after signing in and was landing in #announcements (I reproduced it a couple of times, but is not constant). I'll change it though.

Thanks!

PR #709, opened 2018-10-08
Add "id" to headings

### LINKED ISSUE #702: Relative links don't work in the Chart view
Many charts include links to sections in the same document (e.g. in the Wordpress one there is a link to the `#configuration`). This link is not working since titles are not being generated with an `id`. There is probably an option in the markdown renderer to set that.

andresmgot:
Fixes #702

Add the "id" property to the headings when parsing markdown. The approach to do that is specified here:
https://github.com/rexxars/react-markdown/issues/69

The logic to transform the heading text into a link is the same than GitHub: https://github.com/jch/html-pipeline/blob/master/lib/html/pipeline/toc_filter.rb#L42

### DIFF
--- a/dashboard/src/components/ChartView/ChartReadme.tsx
+++ b/dashboard/src/components/ChartView/ChartReadme.tsx
@@ -3,6 +3,7 @@ import { FileText } from "react-feather";
 import * as ReactMarkdown from "react-markdown";
 
 import LoadingWrapper from "../LoadingWrapper";
+import HeadingRenderer from "./HeadingRenderer";
 
 import "./ChartReadme.css";
 
@@ -11,23 +12,6 @@ interface IChartReadmeProps {
   hasError: boolean;
   readme?: string;
   version: string;
-}
-
-// Code from https://github.com/rexxars/react-markdown/issues/69
-function flatten(text: string, child: any): any {
-  return typeof child === "string"
-    ? text + child
-    : React.Children.toArray(child.props.children).reduce(flatten, text);
-}
-
-function HeadingRenderer(props: any) {
-  const children = React.Children.toArray(props.children);
-  const text = children.reduce(flatten, "");
-  const slug = text
-    .toLowerCase()
-    .replace(/[^a-zA-Z0-9_\s]/g, "") // remove punctuation
-    .replace(/\s/g, "-"); // replace spaces with dash
-  return React.createElement("h" + props.level, { id: slug }, props.children);
 }
 
 class ChartReadme extends React.Component<IChartReadmeProps> {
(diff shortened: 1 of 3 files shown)


dashboard/src/components/ChartView/ChartReadme.tsx
```diff
@@ -13,6 +13,23 @@ interface IChartReadmeProps {
...
 }
 
+// Code from https://github.com/rexxars/react-markdown/issues/69
+function flatten(text: string, child: any): any {
+  return typeof child === "string"
+    ? text + child
+    : React.Children.toArray(child.props.children).reduce(flatten, text);
+}
+
+function HeadingRenderer(props: any) {
```
prydonius: Hmm I'm not a huge fan of having a Component inline in another component, even though this is small one. @Angelmmiguel what do you think?

dashboard/src/components/ChartView/ChartReadme.tsx
```diff
@@ -13,6 +13,23 @@ interface IChartReadmeProps {
   version: string;
 }
 
+// Code from https://github.com/rexxars/react-markdown/issues/69
+function flatten(text: string, child: any): any {
```
prydonius: If we keep this function in this class, we should probably name it better, or instead just have it as an anonymous function in the `.reduce`

dashboard/src/components/ChartView/ChartReadme.tsx
Angelmmiguel: Agree with @prydonius. If the component is specific to another one, I usually keep them in the same folder:

```
CommonComponent
  - CommonComponent.test.tsx
  - CommonComponent.tsx
  - CommonComponent.css
  - SpecificComponent
    - SpecificComponent.test.tsx
    - SpecificComponent.tsx
    - SpecificComponent.css
```

dashboard/src/components/ChartView/ChartReadme.tsx
andresmgot: I moved it to a different component file. I prefer to leave it as it is to maintain fidelity to the linked issue.

dashboard/src/components/ChartView/HeadingRenderer.tsx
```diff
@@ -0,0 +1,20 @@
+import * as React from "react";
+
+// Code from https://github.com/rexxars/react-markdown/issues/69
+function flatten(text: string, child: any): any {
+  return typeof child === "string"
+    ? text + child
+    : React.Children.toArray(child.props.children).reduce(flatten, text);
+}
+
+const HeadingRenderer: React.SFC<{}> = (props: any) => {
```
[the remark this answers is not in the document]
andresmgot: I didn't add a test for this because it's only meant to be used with the `ReactMarkdown` component so I added the test there.

PR #714, opened 2018-10-09
Class list

### LINKED ISSUE #684: Service catalog page headings don't match other page headings
screen shot 2018-09-05 at 16 48 47

They don't have separators like the other pages do.

andresmgot:
Fixes #684

Changes included in this PR:
 - Use the same heading style than the rest of views
 - Show a warning message if no classes are found
 - Show the loading spinner if classes has not been retrieved yet
 - Unit tests

As a side effect, this PR has some changes for the InstanceList and ClassView components. The idea is to refactor those in a different PR.


dashboard/src/components/ClassList/ClassList.tsx
```diff
@@ -0,0 +1,98 @@
...
+                action="list"
+                defaultRequiredRBACRoles={{ list: RequiredRBACRoles }}
+                resource="Service Classes"
+              />
+            ) : (
+              <p>Types of services available from all brokers</p>
+            )}
+            {classes.list.length === 0 ? (
+              <MessageAlert level={"warning"}>
+                <h5>Unable to find any class.</h5>
```
prydonius: I think we should change this to "Service Classes not found" and then add some explanation on what happened, e.g. maybe pointing to this documentation: https://kubernetes.io/docs/concepts/extend-kubernetes/service-catalog/#usage

dashboard/src/components/ClassView/index.tsx
```diff
@@ -83,7 +86,8 @@ export class ClassView extends React.Component<IClassViewProps> {
               <tbody>
                 {svcClass &&
                   classPlans.map(plan => {
-                    const serviceClass = classes.find(
+                    // TODO: Check classes.ifFetching
```
prydonius: isFetching

dashboard/src/components/InstanceListView/index.tsx
```diff
@@ -136,9 +139,10 @@ export class InstanceListView extends React.PureComponent<
               ) : brokers.length > 0 ? (
                 <div>
                   {instances.length > 0 ? (
+                    // TODO: Check ifFetching
```
prydonius: isFetching

dashboard/src/components/InstanceListView/index.tsx
```diff
@@ -20,7 +20,10 @@ import { InstanceCardList } from "./InstanceCardList";
 
 export interface InstanceListViewProps {
   brokers: IServiceBroker[];
-  classes: IClusterServiceClass[];
+  classes: {
```
prydonius: instead of repeating it, we can use `IServiceCatalogState["classes"]` as the type

dashboard/src/components/ClassView/index.tsx
```diff
@@ -23,7 +23,10 @@ const RequiredRBACRoles: IRBACRole[] = [
 ];
 
 interface IClassViewProps {
-  classes: IClusterServiceClass[];
+  classes: {
```
prydonius: instead of repeating it, we can use `IServiceCatalogState["classes"]` as the type

PR #715, opened 2018-10-09
e2e test: remove check for chartsvc image

prydonius:
We use the upstream built image for chartsvc and no longer build it in
the CI, so we shouldn't check for the image tag


andresmgot:
merging this to fix master builds

Issue #683, opened 2018-09-26
listing all classes, plans, instances and bindings in multiple views


prydonius:
We are requesting all the Service Catalog API endpoints on a few of the pages even when they are not required and it would be enough to retrieve just the things we need (e.g. in the case below, just the Service Instances).

![screen shot 2018-09-05 at 16 30 32](https://user-images.githubusercontent.com/1544861/46110542-cccd5a80-c198-11e8-8e5b-2c50bec15ef9.png)

The reason for this is we are using the `getCatalog` action quite a bit instead of requesting individual resources: https://github.com/kubeapps/kubeapps/blob/master/dashboard/src/actions/catalog.ts#L212. We should remove this in favour of being more explicit.

andresmgot:
The components that use the action `getCatalog` are: 

 - [ ] InstanceListView
 - [ ] ServiceList
 - [ ] InstanceView

Let's revamp each one of those in different PRs.

PR #719, opened 2018-10-10
Refactor InstanceListView

andresmgot:
Ref #683 

Refactor the InstanceListView component to:

 - Remove calls to `getCatalog`. Instead of that we are requesting specific resources (classes, brokers, instances).
 - Add property `isFetching` for `brokers` and `instances` and show the loading spinner if those resources are not loaded yet.
 - The search box and the deploy button are showed in any case to match the behavior in AppList.
 - Use `ErrorSelector`.
 - Add unit tests.


dashboard/src/containers/InstanceListViewContainer.ts
```diff
@@ -16,14 +16,14 @@ function mapStateToProps(
   const plans = catalog.plans;
   const classes = catalog.classes;
   const instances = catalog.instances;
-  const isInstalled = catalog.isInstalled;
+  const isServiceCatalogInstalled = catalog.isServiceCatalogInstalled;
```
prydonius: what do you think about using destructuring here to get these, e.g. `{ catalog: { brokers, plans, classes, instances, isServiceCatalogInstalled }, namespace }: IStoreState...`?

Alternatively, I think we can avoid the constants and just use the dot notation in the object below

dashboard/src/actions/catalog.ts
```diff
@@ -225,8 +225,8 @@ export function checkCatalogInstalled(): ThunkAction<
   ServiceCatalogAction
 > {
   return async dispatch => {
-    const isInstalled = await ServiceCatalog.isCatalogInstalled();
-    isInstalled ? dispatch(installed()) : dispatch(notInstalled());
-    return isInstalled;
+    const isServiceCatalogInstalled = await ServiceCatalog.isCatalogInstalled();
```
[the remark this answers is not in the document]
andresmgot: I got confused when I was reading the component `ServiceInstanceList`. There I didn't know what "was installed" until I read the ternary operator `isInstalled ? ... : <ServiceCatalogNotFound />`. I think it's useful to have it verbose there (and in this file it doesn't hurt either), don't you think?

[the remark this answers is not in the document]
prydonius: Fair enough, I almost wonder if it would make more sense if it was `apiExists` or something?

dashboard/src/components/ServiceInstanceList/ServiceInstanceList.test.tsx
```diff
@@ -1,27 +1,117 @@
 import { shallow } from "enzyme";
+import context from "jest-plugin-context";
 import * as React from "react";
 
-import { MessageAlert } from "../ErrorAlert";
-import ServiceInstanceList, { IServiceInstanceListProps } from "./ServiceInstanceList";
+import { IServiceBroker } from "shared/ServiceCatalog";
```
prydonius: we have a mix of relative and non-relative imports here, we should make this more consistent. In my experience, the tests failed with relative imports, but it seems it's not failing here... anyway I think we should use relative imports.

dashboard/src/components/ServiceInstanceList/ServiceInstanceList.tsx
```diff
@@ -1,37 +1,38 @@
...
 import { Link } from "react-router-dom";
 
-import { IServiceCatalogState } from "../../reducers/catalog";
-import { IServiceBroker, IServicePlan } from "../../shared/ServiceCatalog";
 import { IServiceInstance } from "../../shared/ServiceInstance";
-import { ForbiddenError, IRBACRole } from "../../shared/types";
+import { IRBACRole } from "../../shared/types";
 import { escapeRegExp } from "../../shared/utils";
+
+import { IServiceCatalogState } from "reducers/catalog";
```
prydonius: change to relative import

dashboard/src/components/Config/ServiceCatalog/ServiceCatalog.tsx
```diff
@@ -1,5 +1,6 @@
 import * as React from "react";
 
+import { IServiceCatalogState } from "reducers/catalog";
```
prydonius: change to relative import

dashboard/src/components/ServiceInstanceList/ServiceInstanceList.tsx
```diff
@@ -74,116 +63,106 @@ class ServiceInstanceList extends React.PureComponent<
...
+      isServiceCatalogInstalled,
+      brokers,
+      instances,
+      classes,
+      pushSearchFilter,
+    } = this.props;
+    const loaded = !brokers.isFetching && !instances.isFetching && !classes.isFetching;
     return (
-      <section className="ServiceInstanceList">
+      <section className="InstanceList">
```
prydonius: the class name should be ServiceInstanceList

dashboard/src/actions/catalog.ts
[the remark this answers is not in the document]
andresmgot: that would have the same problem, to be self explanatory it should be something like `serviceCatalogAPIExists`, since the action is `installed` or `notInstalled` I would leave it as it is to avoid more chained changes.

dashboard/src/components/ServiceInstanceList/ServiceInstanceList.test.tsx
andresmgot: this was the automatic import, I don't know what logic that follows.

PR #723, opened 2018-10-11
ServiceInstanceView refactor

### LINKED ISSUE #683: listing all classes, plans, instances and bindings in multiple views

andresmgot:
Requires #719 
Closes #683 

Refactor ServiceInstanceView to:
 - Include `PageHeader`.
 - Include `LoadingWrapper`.
 - Avoid `getCatalog`. This was the last method using it so it's no longer required.
 - Remove the namespace in the headings (like `default/my-instance`) to be consistent with the AppView.
 - Use `ErrorSelector`.
 - Add unit tests.


dashboard/src/components/ServiceInstanceList/ServiceInstanceList.tsx
```diff
@@ -93,6 +93,50 @@ class ServiceInstanceList extends React.PureComponent<
       pushSearchFilter,
     } = this.props;
     const loaded = !brokers.isFetching && !instances.isFetching && !classes.isFetching;
+    let body = <span />;
+    if (!isServiceCatalogInstalled) {
```
andresmgot: this change is required in #719 but I'm adding it here since the catalog reducer changes are more complete in this PR (sorry!)

andresmgot:
I think that's small enough to be part of the PR that refactors that part, we can just address that kind of things in the review. Just raise it whenever you see it in a review.

PR #728, opened 2018-10-11
Add global registry option

carrodher:
- Add `global.registry` at the top of `values.yaml` (commented out at the beginning).
- Add the required logic to use the global value if it is defined.
- Add a link to the `values.yaml` of the dependency in the main `values.yaml` chart (in the section of the dependency options).


chart/kubeapps/templates/apprepository-deployment.yaml
```diff
@@ -22,12 +22,12 @@ spec:
...
       containers:
       - name: controller
-        image: {{ template "kubeapps.image" .Values.apprepository.image }}
+        image: {{ template "apprepository.image" . }}
         command:
         - /apprepository-controller
         args:
         - --logtostderr
-        - --repo-sync-image={{ template "kubeapps.image" .Values.apprepository.syncImage }}
+        - --repo-sync-image={{ template "apprepository.image" . }}
```
andresmgot: this should be `app repository.syncImage`

chart/kubeapps/templates/_helpers.tpl
```diff
@@ -25,10 +25,93 @@ If release name contains chart name it will be used as a full name.
 {{- end -}}
 
 {{/*
-Render image reference
+Return the proper apprepository image name
 */}}
-{{- define "kubeapps.image" -}}
-{{ .registry }}/{{ .repository }}:{{ .tag }}
```
andresmgot: can't you just use a helper to return the repository? Then you can use it as:

```
{{ .registry }}/{{ template "kubeapps.repository" .repository }}:{{ .tag }}
```

with that you wouldn't need all the code below nor the changes in the templates

prydonius: +1 I'm not sure why we've extracted this to multiple helpers, when we can just use one. If the problem is that we need to pass in the global value, we can solve that by using a list. For example:

```
{{ template "kubeapps.image" (list .Values.apprepository.image .Values.global) }}
```

and then:

```
{{- define "kubeapps.image" -}}
{{- $image := index . 0 -}}
{{- $global := index . 1 -}}
{{- $registryName := $image.registry -}}
{{- if $global }}
    {{- if $global.regristry }}
        {{- $registryName := $global.registry -}}
    {{- end -}}
{{- end -}}
{{ .registryName }}/{{ .repository }}:{{ .tag }}
{{- end -}}
```

carrodher: Yes, the reason was related to passing the global value, I adapted your solution and it works fine

chart/kubeapps/templates/_helpers.tpl
```diff
@@ -28,7 +28,15 @@ If release name contains chart name it will be used as a full name.
 Render image reference
 */}}
 {{- define "kubeapps.image" -}}
-{{ .registry }}/{{ .repository }}:{{ .tag }}
+{{- $image := index . 0 -}}
+{{- $global := index . 1 -}}
+{{- $registryName := $image.registry -}}
+{{- if $global }}
```
andresmgot: just to double check, aren't you able to use `{{- if .Values.global -}}` here? We are using `.Values` in other places here (and you would be able to omit the changes in the other templates).

In any case, being explicit in the templates is not bad either so I am okay if we leave this like this.

chart/kubeapps/values.yaml
```diff
@@ -1,3 +1,9 @@
+## Global chart image registry
+## Please, note that this registry would be used for all the containers in the chart and dependencies
```
andresmgot: I would reword this a bit:

> Please, note that this will override the image registry for all the images, including dependencies, configured to use the global value.

chart/kubeapps/templates/_helpers.tpl
carrodher: Using
```yaml
{{/*
Render image reference
*/}}
{{- define "kubeapps.image" -}}
{{- $registryName := .registry -}}
{{- if .Values.global }}
    {{- if .Values.global.regristry }}
        {{- $registryName := .Values.global.registry -}}
    {{- end -}}
{{- end -}}
{{ $registryName }}/{{ .repository }}:{{ .tag }}
{{- end -}}
```
and without changes in the other templates, tests fails because of

> Error: render error in "kubeapps/templates/tiller-proxy-deployment.yaml": template: kubeapps/templates/_helpers.tpl:32:14: executing "kubeapps.image" at <.Values.global>: can't evaluate field global in type interface {}

chart/kubeapps/values.yaml
```diff
@@ -1,3 +1,9 @@
+## Global chart image registry
```
migmartri: Docker image (instead of chart image) registry?

chart/kubeapps/values.yaml
migmartri: `registry` might be too vague. It might make sense to call it dockerImageRegistry, imageRegistry or something like that.

> note that this registry would be used for all the containers in the chart and dependencies

I'd rephrase the comment by explaining what it does, the current message just says that it will affect to all the containers but not how.

chart/kubeapps/templates/_helpers.tpl
andresmgot: 👍  thanks for checking! (I don't know why that doesn't work though)

The only thing is that I think you missed the closing `-` in the if sentences:

  {{- if $global **-**}}
    {{- if $global.regristry **-**}}

carrodher:
@migmartri Yes, the idea is apply those changes to all [Bitnami Charts](https://github.com/bitnami/charts/) and [Helm Charts](https://github.com/bitnami/charts/), in fact the PR about Bitnami Charts is already sent (https://github.com/bitnami/charts/pull/866).

Helm Charts is more complex because we need to send a PR per chart, so I will send it once that we have everything discussed, because changes are more complicated if we need to update 30 PRs per change.

Maybe this variable is never used for Kubeapps, but the to apply those changes here is to maintain the same structure/sections in all Bitnami charts.

carrodher:
```bash
$ helm template . > before
$ helm template . --set global.imageRegistry=petete.io > after
$ diff before after
14c14
<   mongodb-root-password: "dnBwU2lOSEpyQg=="
---
>   mongodb-root-password: "enBLOHg1SFNTaw=="
907c907
<         image: docker.io/kubeapps/apprepository-controller:latest
---
>         image: petete.io/kubeapps/apprepository-controller:latest
912c912
<         - --repo-sync-image=docker.io/quay.io/helmpack/chart-repo:v1.0.0
---
>         - --repo-sync-image=petete.io/quay.io/helmpack/chart-repo:v1.0.0
950c950
<         image: docker.io/quay.io/helmpack/chartsvc:v1.0.0
---
>         image: petete.io/quay.io/helmpack/chartsvc:v1.0.0
1015c1015
<         image: docker.io/kubeapps/dashboard:latest
---
>         image: petete.io/kubeapps/dashboard:latest
1088c1088
<         image: docker.io/bitnami/nginx:1.14.0-r27
---
>         image: petete.io/bitnami/nginx:1.14.0-r27
1148c1148
<         image: docker.io/kubeapps/tiller-proxy:latest
---
>         image: petete.io/kubeapps/tiller-proxy:latest
1193c1193
<         image: docker.io/lachlanevenson/k8s-kubectl:v1.9.9
---
>         image: petete.io/lachlanevenson/k8s-kubectl:v1.9.9
1233c1233
<         image: docker.io/lachlanevenson/k8s-kubectl:v1.9.9
---
>         image: petete.io/lachlanevenson/k8s-kubectl:v1.9.9
1272c1272
<         image: docker.io/lachlanevenson/k8s-kubectl:v1.9.9
---
>         image: petete.io/lachlanevenson/k8s-kubectl:v1.9.9
1316c1316
<         image: docker.io/lachlanevenson/k8s-kubectl:v1.9.9
---
>         image: petete.io/lachlanevenson/k8s-kubectl:v1.9.9
```

carrodher:
First option was to use something like
```
{{- $registryName := $image.registry -}}
{{- if $global -}}
    {{- if $global.imageRegistry -}}
        {{- $registryName := $global.imageRegistry -}}
    {{- end -}}
{{- end -}}
{{ $registryName }}/{{ $image.repository }}:{{ $image.tag }}
```
but I found that the `$registryName` inside the if statement has a limited scope, this variable is "local" to the if scope, so at the end, we always are returning the "global" `$registryName`, so the if statements are not working.

Second attempt was to use
```
{{- $registryName := $image.registry -}}
{{- if $global -}}
    {{- if $global.imageRegistry -}}
        {{- $registryName = $global.imageRegistry -}}
    {{- end -}}
{{- end -}}
{{ $registryName }}/{{ $image.repository }}:{{ $image.tag }}
```
It is the same option, but inside the if statement we are not using `:=` to assign (and define) `$registryName`, using `=` we are overwriting the "global" variable. This options works fine! Well, works fine with Helm 2.11 (20days old) it doesn't work with Helm 2.9 nor 2.10.

So, after check it with @andresmgot we found this solution that works across all Helm versions:
```
{{- if $global -}}
    {{- if $global.imageRegistry -}}
        {{ $global.imageRegistry }}/{{ $image.repository }}:{{ $image.tag }}
    {{- else -}}
        {{ $image.registry }}/{{ $image.repository }}:{{ $image.tag }}
    {{- end -}}
{{- else -}}
    {{ $image.registry }}/{{ $image.repository }}:{{ $image.tag }}
{{- end -}}
```
we are duplicating code to avoid use variables.

What do you think @migmartri? Any other idea?

migmartri:
Thanks @carrodher 

> ```
> {{- if $global -}}
>     {{- if $global.imageRegistry -}}
>         {{ $global.imageRegistry }}/{{ $image.repository }}:{{ $image.tag }}
>     {{- else -}}
>         {{ $image.registry }}/{{ $image.repository }}:{{ $image.tag }}
>     {{- end -}}
> {{- else -}}
>     {{ $image.registry }}/{{ $image.repository }}:{{ $image.tag }}
> {{- end -}}
> ```
> we are duplicating code to avoid use variables.
> 
> What do you think @migmartri? Any other idea?

It makes sense to me to start with a compatible version of the code (3rd option) although there might be another way to write that code by leveraging [`and`](https://golang.org/pkg/text/template/#hdr-Functions), assuming that it `short-circuits`?

Something like:

```
{{- if and $global $global.imageRegistry -}}
    {{ $global.imageRegistry }}/{{ $image.repository }}:{{ $image.tag }}
{{- else -}}
    {{ $image.registry }}/{{ $image.repository }}:{{ $image.tag }}
{{- end -}}
```

You might want to double check it using `helm template`

carrodher:
> It makes sense to me to start with a compatible version of the code (3rd option) although there might be another way to write that code by leveraging [`and`](https://golang.org/pkg/text/template/#hdr-Functions), assuming that it `short-circuits`?
> 
> Something like:
> 
> ```
> {{- if and $global $global.imageRegistry -}}
>     {{ $global.imageRegistry }}/{{ $image.repository }}:{{ $image.tag }}
> {{- else -}}
>     {{ $image.registry }}/{{ $image.repository }}:{{ $image.tag }}
> {{- end -}}
> ```
> You might want to double check it using `helm template`

Nope, we can't use a single if because lazy evaluation (`short-circuits`) is not an option, I tried it some commits ago and it fails

migmartri:
@carrodher thanks for checking the lazy evaluation and providing with the template diff.

PR #730, opened 2018-10-15
Update certs for e2e tests

andresmgot:
Update the sample certs of the repo. The e2e tests were failing because the current cert has expired: `[debug] x509: certificate has expired or is not yet valid`


andresmgot:
_merging this to fix master_

migmartri:
Why did it expire? Was it a mistake creating them in the first place? I am asking because I can see now the instructions on how to create them but they seem just informative right?

Thanks for fixing it!

andresmgot:
> Why did it expire?

The previous cert used the default expiration date of 30 days (now I set it to 7300)

> Was it a mistake creating them in the first place?

It wasn't, we decided to commit them to save the time that it takes to generate them in the CI.

> I can see now the instructions on how to create them but they seem just informative right?

That's correct, it's just informative in case someone needs to regenerate them.

PR #732, opened 2018-10-15
Refactor Card list for service instances

### LINKED ISSUE #685: Service catalog card design doesn't match applications list
screen shot 2018-09-05 at 16 42 44

We should have a more consistent card layout

andresmgot:
Fixes #685

Use the same format for Card in the ServiceList view than in the chart and app list. This PR also refactor the code that was before in `ChartList` (`ChartListItem`) to a new component `PreformattedCard`.

screen shot 2018-10-15 at 16 09 06


dashboard/src/components/AppList/AppListItem.scss
```diff
@@ -0,0 +1,11 @@
+.DELETED {
```
migmartri: Minor comment and probably not worth changing.

Why the usage of uppercase?

Is it because it matches directly the status? Maybe we can transform it and still be consistent with other css class names?

dashboard/src/components/ChartList/ChartListItem.tsx
```diff
@@ -1,9 +1,8 @@
 import * as React from "react";
-import { Link } from "react-router-dom";
 
 import placeholder from "../../placeholder.png";
 import { IChart } from "../../shared/types";
-import Card, { CardContent, CardIcon } from "../Card";
+import { PreformattedCard } from "../Card";
```
migmartri: I am not convinced about the naming `PreformattedCard`. Why preformatted? Maybe something that represents its behavior or purpose like `genericCard`, `infoCard`?

andresmgot: I don't have a strong opinion about the name, I just wanted to emphasize that it has a specific layout. `infoCard` works for me as well.

andresmgot:
Merging this. Thanks for the review!

PR #735, opened 2018-10-16
Avoid to submit empty params

### LINKED ISSUE #411: Subscription (an optional parameter in Google PubSub) is always sent, even if empty
When creating a binding for Google PubSub, Subscription is optional, but with Kubeapps always sends it when creating a binding (even when you leave that area untouched).

This is the BindingCreate json schema:

```
    serviceBindingCreateParameterSchema:
      $schema: http://json-schema.org/draft-04/schema#
      definitions:
        PushConfig:
          description: Configuration for a push delivery endpoint.
          properties:
            attributes:
              additionalProperties:
                type: string
              description: |-
                Endpoint configuration attributes.

                Every endpoint has a set of API supported attributes that can be used to
                control different aspects of the message delivery.

                The currently supported attribute is `x-goog-version`, which you can
                use to change the format of the pushed message. This attribute
                indicates the version of the data expected by the endpoint. This
                controls the shape of the pushed message (i.e., its fields and metadata).
                The endpoint version is based on the version of the Pub/Sub API.

                If not present during the `CreateSubscription` call, it will default to
                the version of the API used to make such call. If not present during a
                `ModifyPushConfig` call, its value will not be changed. `GetSubscription`
                calls will always return a valid version, even if the subscription was
                created without this attribute.

                The possible values for this attribute are:

                * `v1beta1`: uses the push format defined in the v1beta1 Pub/Sub API.
                * `v1` or `v1beta2`: uses the push format defined in the v1 Pub/Sub API.
              type: object
            pushEndpoint:
              description: |-
                A URL locating the endpoint to which messages should be pushed.
                For example, a Webhook endpoint might use "https://example.com/push".
              type: string
          type: object
        Subscription:
          description: A subscription resource.
          properties:
            ackDeadlineSeconds:
              description: |-
                This value is the maximum time after a subscriber receives a message
                before the subscriber should acknowledge the message. After message
                delivery but before the ack deadline expires and before the message is
                acknowledged, it is an outstanding message and will not be delivered
                again during that time (on a best-effort basis).

                For pull subscriptions, this value is used as the initial value for the ack
                deadline. To override this value for a given message, call
                `ModifyAckDeadline` with the corresponding `ack_id` if using
                non-streaming pull or send the `ack_id` in a
                `StreamingModifyAckDeadlineRequest` if using streaming pull.
                The minimum custom deadline you can specify is 10 seconds.
                The maximum custom deadline you can specify is 600 seconds (10 minutes).
                If this parameter is 0, a default value of 10 seconds is used.

                For push delivery, this value is also used to set the request timeout for
                the call to the push endpoint.

                If the subscriber never acknowledges the message, the Pub/Sub
                system will eventually redeliver the message.
              maximum: 600
              minimum: 0
              type: integer
            pushConfig:
              $ref: '#/definitions/PushConfig'
              description: |-
                If push delivery is used with this subscription, this field is
                used to configure it. An empty `pushConfig` signifies that the subscriber
                will pull and ack messages using API methods.
            subscriptionId:
              description: A user-specified Pubsub subscription ID.
              maxLength: 255
              minLength: 3
              pattern: ^[A-Za-z][-_.~+%A-Za-z0-9]+$
              title: Subscription ID
              type: string
          type: object
      form:
      - roles
      - serviceAccount
      - createServiceAccount
      - key: subscription.subscriptionId
        validationMessage: The value must start with a letter, and contain only letters,
          numbers, dashes, underscores, periods, tildes, plus or percent signs. It
          must be between 3 and 255 characters.
      properties:
        createServiceAccount:
          default: false
          description: Create a new service account for Pub/Sub binding.
          title: Create service account
          type: boolean
        roles:
          description: The list of Cloud Pubsub roles for this binding. Affects the
            level of access granted to the service account.
          items:
            enum:
            - roles/pubsub.publisher
            - roles/pubsub.subscriber
            - roles/pubsub.viewer
            - roles/pubsub.editor
            - roles/pubsub.admin
            type: string
          title: Roles
          type: array
          uniqueItems: true
        serviceAccount:
          description: The GCP service account to which access will be granted. For
            example <storage-service-account> or <storage-service-account>@<project-id>.iam.gserviceaccount.com
          title: Service account ID
          type: string
        subscription:
          $ref: '#/definitions/Subscription'
      required:
      - roles
      - serviceAccount
      type: object
```
Subscription is a nested Json schema, but it always appears in the rendering of the form:

screen shot 2018-10-04 at 1 14 35 pm

If you select publisher as role and leave subscription empty, the subform will be sent anyways, and the GCP broker will complain, as the Subscription is only expected when you select subscriber as role:

```
Bind call failed: operation "projects/774136762833/operations/08ac200b-8f27-11e8-812b-0242ac110016/1532425846197657130" failed: generic::internal: operation projects/774136762833/operations/08ac200b-8f27-11e8-812b-0242ac110016/1532425846197657130 create binding deployment failed: generic::invalid_argument: Operation "operation-1532425847110-571bbb242f970-ff3e07a6-1aaa3a8f" failed with : [{"code":"MANIFEST_EXPANSION_USER_ERROR","location":"/deployments/bb4b83d93-c6e7-4531-abbf-67bda21b3035/manifests/manifest-1532425847188","message":"Manifest expansion encountered the following errors: Exception in gcp-services/composite:cloud-pubsub-binding-f3622e2\nTraceback (most recent call last):\n return constructor[m](evaluation_context)\n File \"\u003cstring\u003e\", line 85, in GenerateConfig\nException: 'roles/pubsub.subscriber' is a required role when 'subscription' is specified as one of 'parameters'.\n Resource: gcp-services/composite:cloud-pubsub-binding-f3622e2 Resource: config"}]
```

If that area is left untouched, we shouldn't be sending that parameter.

andresmgot:
Fixes #411

When creating a service instance or a binding, if the schema is provided, avoid to submit empty objects. This is only made when we are sure that the parameter is not required.


dashboard/src/actions/catalog.test.tsx
```diff
@@ -165,6 +166,47 @@ describe("provision", () => {
...
+          i: "",
+        },
+      },
+      j: {
+        k: {
+          l: "m",
+        },
+      },
+    };
+    // It omits "f" because it's empty but not "b" because it's required
```
migmartri: Thanks for explanation.

What's the difference between sending empty required parameters and not sending them at all? Wouldn't not sending the empty parameters regardless of being required or not have the same effect?

dashboard/src/actions/catalog.ts
```diff
@@ -60,16 +62,44 @@ const actions = [
 
 export type ServiceCatalogAction = ActionType<typeof actions[number]>;
 
+function isEmptyDeep(obj: any): boolean {
```
migmartri: Please add a description of what this function does. Mainly because it seems to have a recursive behavior.

dashboard/src/actions/catalog.ts
```diff
@@ -60,16 +62,44 @@ const actions = [
...
+  if (typeof obj === "number") {
```
migmartri: why is number a special case? What about string and others?

dashboard/src/actions/catalog.ts
```diff
@@ -60,16 +62,44 @@ const actions = [
...
+    // Check if nested objects are empty
+    // If some of the keys are not empty the result is not empty
+    return !Object.keys(obj).some(k => {
+      return !isEmptyDeep(obj[k]);
+    });
+  }
+  return _.isEmpty(obj);
+}
+
+function removeEmptyFields(obj: object, schema?: JSONSchema6) {
```
migmartri: maybe rename to `removeEmptyNonRequiredFields`?

dashboard/src/actions/catalog.ts
migmartri: I know that we are testing the overall behavior at the action level but could we add unit tests for this function?

dashboard/src/actions/catalog.ts
```diff
@@ -60,16 +62,44 @@ const actions = [
...
   namespace: string,
   className: string,
   planName: string,
   parameters: {},
+  schema?: JSONSchema6,
 ): ThunkAction<Promise<boolean>, IStoreState, null, ServiceCatalogAction> {
   return async dispatch => {
     try {
-      await ServiceInstance.create(releaseName, namespace, className, planName, parameters);
+      const filteredParams = removeEmptyFields(parameters, schema);
```
migmartri: maybe we do not need to pass schema all the time since it is actually optional and with default value to `JSONSchema6`? Or do we support other schemas?

dashboard/src/actions/catalog.ts
andresmgot: it's a `lodash` problem:

https://github.com/lodash/lodash/blob/4ea8c2ec249be046a0f4ae32539d652194caf74f/isEmpty.js#L33

In our case, we want to mark a number as non empty so we need this special case. It's not critical though because you cannot set a "number" in a form (AFAIK) but just in case.

dashboard/src/actions/catalog.ts
andresmgot: I don't understand what you mean here

dashboard/src/actions/catalog.ts
migmartri: I don't know. There should be a way to unit test functions without exporting them, otherwise we are not unit testing.

Can't we make it work like this?
```
wrapper = shallow("foo")
instance = wrapper.instance();
instance.isEmptyDeep(...) ?
```

The thing is that this function is complex enough too deserve its unit test. Relying on functional/integration test might not be the best option once the complexity starts growing.

dashboard/src/actions/catalog.ts
migmartri: My bad, I was understanding schema like the schema type and thought that the default value was `JSONSchema6`. When what it actually does it to receive an instance of jsonschema.

dashboard/src/actions/catalog.ts
```diff
@@ -60,16 +62,44 @@ const actions = [
...
+  return res;
+}
+
 export function provision(
   releaseName: string,
   namespace: string,
   className: string,
   planName: string,
   parameters: {},
+  schema?: JSONSchema6,
```
migmartri: Why is schema optional?

dashboard/src/actions/catalog.test.tsx
migmartri: right, my comment was related to not to send the parameters if they are empty in any case. The difference is that you still send them if they are required params but I am suggesting not to send them even in that case. I might be missing something though.

andresmgot: okay, yes, it's true that it should be safe to not send anything that is empty.

dashboard/src/actions/catalog.ts
andresmgot: moving this function to a different helper to be able to test it

dashboard/src/actions/catalog.test.tsx
```diff
@@ -165,6 +165,40 @@ describe("provision", () => {
     await store.dispatch(provisionCMD);
     expect(store.getActions()).toEqual(expectedActions);
   });
+
+  it("filters the submitted parameters if they are not required", async () => {
```
migmartri: Is that description accurate? 

filters submitted params if they are empty?

dashboard/src/actions/catalog.test.tsx
```diff
@@ -227,6 +261,41 @@ describe("addBinding", () => {
...
```
migmartri: same here

dashboard/src/actions/catalog.ts
```diff
@@ -69,7 +70,8 @@ export function provision(
 ): ThunkAction<Promise<boolean>, IStoreState, null, ServiceCatalogAction> {
   return async dispatch => {
     try {
-      await ServiceInstance.create(releaseName, namespace, className, planName, parameters);
+      const filteredParams = helpers.object.removeEmptyFields(parameters);
```
migmartri: nice.

dashboard/src/actions/helpers/object.test.tsx
```diff
@@ -0,0 +1,34 @@
+import * as object from "./object";
+
+describe("isEmptyDeep", () => {
```
migmartri: great :)

dashboard/src/actions/helpers/object.test.tsx
```diff
@@ -0,0 +1,34 @@
...
+  [
+    { object: {}, isEmpty: true },
+    { object: { a: 1 }, isEmpty: false },
+    { object: { a: { b: "" } }, isEmpty: true },
+    { object: { a: { c: { d: "" } } }, isEmpty: true },
+    { object: { a: { c: { d: "a" } } }, isEmpty: false },
```
migmartri: what about an example that contains both an empty and non-empty at the leaf?
```
    { object: { a: { c: { d: "a" }, d: "" } }, isEmpty: false },
```

dashboard/src/actions/helpers/object.ts
```diff
@@ -0,0 +1,30 @@
+import * as _ from "lodash";
+
+// Check if all the keys of an object are empty. If any value of the
```
migmartri: Thank you for the description :)

dashboard/src/actions/helpers/object.ts
```diff
@@ -0,0 +1,30 @@
...
+// object is a nested object it recursively checks if the inner object
+// is empty.
+export function isEmptyDeep(obj: any): boolean {
+  if (typeof obj === "number") {
+    // isEmpty(number) is true but it's not empty
```
migmartri: Could you mention that it is a limitation of lodash?

PR #740, opened 2018-10-17
[Kubeapps Chart] Simplify ingress configuration for cert-manager

juan131:
### What this PR does / why we need it:

This PR allows the user to configure the proper annotations when using cert-manager to handle TLS certificates in an easy way. The user just needs to setup a boolean to true to do so.

E.g.:

```
helm install --name kubeapps --namespace kubeapps bitnami/kubeapps \
  --set ingress.enabled=true \
  --set ingress.hosts[0].name=kubeapps.custom.domain \
  --set ingress.hosts[0].tlsSecret=kubeapps-tls \
  --set ingress.hosts[0].certManager=true
```

This will generate the following ingress:

```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: kubeapps
  namespace: kubeapps
  labels:
    app: kubeapps
    chart: kubeapps-0.8.0
    release: kubeapps
    heritage: Tiller
  annotations:
    kubernetes.io/tls-acme: "true"
spec:
  rules:
  - host: kubeapps.custom.domain
    http:
      paths:
      - path: /
        backend:
          serviceName: kubeapps
          servicePort: http
  tls:
  - hosts:
    - kubeapps.custom.domain 
    secretName: kubeapps-tls
```

See also https://github.com/bitnami/charts/pull/882

### DIFF
--- a/chart/kubeapps/Chart.yaml
+++ b/chart/kubeapps/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kubeapps
-version: 0.8.0
+version: 1.0.0
 appVersion: DEVEL
 description: Kubeapps is a dashboard for your Kubernetes cluster that makes it easy to deploy and manage applications in your cluster using Helm
 icon: https://raw.githubusercontent.com/kubeapps/kubeapps/master/docs/img/logo.png
--- a/chart/kubeapps/Chart.yaml
+++ b/chart/kubeapps/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: kubeapps
 version: 0.8.0
-appVersion: v1.0.0-beta.2
+appVersion: DEVEL
 description: Kubeapps is a dashboard for your Kubernetes cluster that makes it easy to deploy and manage applications in your cluster using Helm
 icon: https://raw.githubusercontent.com/kubeapps/kubeapps/master/docs/img/logo.png
 keywords:
--- a/chart/kubeapps/Chart.yaml
+++ b/chart/kubeapps/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kubeapps
-version: 1.0.0
+version: 0.9.0
 appVersion: DEVEL
 description: Kubeapps is a dashboard for your Kubernetes cluster that makes it easy to deploy and manage applications in your cluster using Helm
 icon: https://raw.githubusercontent.com/kubeapps/kubeapps/master/docs/img/logo.png
--- a/chart/kubeapps/Chart.yaml
+++ b/chart/kubeapps/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: kubeapps
-version: 0.7.0
-appVersion: DEVEL
+version: 0.8.0
+appVersion: v1.0.0-beta.2
 description: Kubeapps is a dashboard for your Kubernetes cluster that makes it easy to deploy and manage applications in your cluster using Helm
 icon: https://raw.githubusercontent.com/kubeapps/kubeapps/master/docs/img/logo.png
 keywords:
--- a/chart/kubeapps/README.md
+++ b/chart/kubeapps/README.md
@@ -171,7 +171,7 @@ To enable ingress integration, please set `ingress.enabled` to `true`
 
 ##### Hosts
 
-Most likely you will only want to have one hostname that maps to this Kubeapps installation, however, it is possible to have more than one host.  To facilitate this, the `ingress.hosts` object is an array.
+Most likely you will only want to have one hostname that maps to this Kubeapps installation, however, it is possible to have more than one host. To facilitate this, the `ingress.hosts` object is an array.
 
 ##### Annotations
 
@@ -179,7 +179,9 @@ For annotations, please see [this document](https://github.com/kubernetes/ingres
 
 ##### TLS
 
-TLS can be configured using the `ingress.tls` object in the same format that the Kubernetes Ingress requests. Please see [this example](https://github.com/kubernetes/contrib/tree/master/ingress/controllers/nginx/examples/tls) for more information.
+TLS can be configured using setting the `ingress.hosts[].tls` boolean of the corresponding hostname to true, then you can choose the TLS secret name setting `ingress.hosts[].tlsSecret`. Please see [this example](https://github.com/kubernetes/contrib/tree/master/ingress/controllers/nginx/examples/tls) for more information.
+
+If your cluster has a [cert-manager](https://github.com/jetstack/cert-manager) add-on to automate the management and issuance of TLS certificates, set `ingress.hosts[].certManager` boolean to true to enable the corresponding annotations for cert-manager. Otherwise, you can provide your own certificates using the `ingress.secrets` object.
 
 ## Troubleshooting
 
--- a/chart/kubeapps/README.md
+++ b/chart/kubeapps/README.md
@@ -181,7 +181,16 @@ For annotations, please see [this document](https://github.com/kubernetes/ingres
 
 TLS can be configured using setting the `ingress.hosts[].tls` boolean of the corresponding hostname to true, then you can choose the TLS secret name setting `ingress.hosts[].tlsSecret`. Please see [this example](https://github.com/kubernetes/contrib/tree/master/ingress/controllers/nginx/examples/tls) for more information.
 
-If your cluster has a [cert-manager](https://github.com/jetstack/cert-manager) add-on to automate the management and issuance of TLS certificates, set `ingress.hosts[].certManager` boolean to true to enable the corresponding annotations for cert-manager. Otherwise, you can provide your own certificates using the `ingress.secrets` object.
+You can provide your own certificates using the `ingress.secrets` object. If your cluster has a [cert-manager](https://github.com/jetstack/cert-manager) add-on to automate the management and issuance of TLS certificates, set `ingress.hosts[].certManager` boolean to true to enable the corresponding annotations for cert-manager as shown in the example below:
+
+```console
+helm install --name kubeapps --namespace kubeapps bitnami/kubeapps \
+  --set ingress.enabled=true \
+  --set ingress.certManager=true \
+  --set ingress.hosts[0].name=kubeapps.custom.domain \
+  --set ingress.hosts[0].tls=true \
+  --set ingress.hosts[0].tlsSecret=kubeapps-tls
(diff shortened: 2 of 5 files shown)


chart/kubeapps/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: kubeapps
-version: 0.7.0
-appVersion: DEVEL
+version: 0.8.0
+appVersion: v1.0.0-beta.2
```
andresmgot: revert this change, here it should be `DEVEL`

chart/kubeapps/values.yaml
```diff
@@ -60,7 +90,7 @@ apprepository:
   image:
     registry: docker.io
     repository: kubeapps/apprepository-controller
-    tag: latest
+    tag: v1.0.0-beta.2
```
andresmgot: the same here, it's needed for this to be `latest`

chart/kubeapps/templates/ingress.yaml
```diff
@@ -1,38 +1,36 @@
 {{- if .Values.ingress.enabled -}}
-{{- $fullName := include "kubeapps.fullname" . -}}
-{{- $ingressPath := .Values.ingress.path -}}
+{{- range .Values.ingress.hosts }}
```
migmartri: so this is creating an ingress per host? Why don't we want to use the multi-host capabilities of a single ingress?

Having different ingresses could mean that cert-manager instead of issuing a cert with multiple CN will potentially do different certs no?

chart/kubeapps/values.yaml
```diff
@@ -6,16 +6,46 @@
...
+
+    # Ingress annotations done as key:value pairs
+    # For a full list of possible ingress annotations, please see
+    # ref: https://github.com/kubernetes/ingress-nginx/blob/master/docs/annotations.md
+    #
+    # If certManager is set to true, annotation kubernetes.io/tls-acme: "true" will automatically be set
+    annotations:
+    #  kubernetes.io/ingress.class: nginx
+
+  secrets:
```
migmartri: I can not see the reference where this is being used.

chart/kubeapps/README.md
```diff
@@ -171,15 +171,17 @@ To enable ingress integration, please set `ingress.enabled` to `true`
...
 
 ##### TLS
 
-TLS can be configured using the `ingress.tls` object in the same format that the Kubernetes Ingress requests. Please see [this example](https://github.com/kubernetes/contrib/tree/master/ingress/controllers/nginx/examples/tls) for more information.
+TLS can be configured using setting the `ingress.hosts[].tls` boolean of the corresponding hostname to true, then you can choose the TLS secret name setting `ingress.hosts[].tlsSecret`. Please see [this example](https://github.com/kubernetes/contrib/tree/master/ingress/controllers/nginx/examples/tls) for more information.
+
+If your cluster has a [cert-manager](https://github.com/jetstack/cert-manager) add-on to automate the management and issuance of TLS certificates, set `ingress.hosts[].certManager` boolean to true to enable the corresponding annotations for cert-manager. Otherwise, you can provide your own certificates using the `ingress.secrets` object.
```
migmartri: Maybe you can add here the example you put in the description

```
helm install --name kubeapps --namespace kubeapps bitnami/kubeapps \
  --set ingress.enabled=true \
  --set ingress.hosts[0].name=kubeapps.custom.domain \
  --set ingress.hosts[0].tlsSecret=kubeapps-tls \
  --set ingress.hosts[0].certManager=true
```

migmartri:
Out of curiosity, could you also post what would have been the equivalent to the following command before the change? 
```
helm install --name kubeapps --namespace kubeapps bitnami/kubeapps \
  --set ingress.enabled=true \
  --set ingress.hosts[0].name=kubeapps.custom.domain \
  --set ingress.hosts[0].tlsSecret=kubeapps-tls \
  --set ingress.hosts[0].certManager=true
```

I am wondering if it would make sense to just add the `certManager` annotation helper but not the multi-host <-> multi-ingress support.

Thanks!

chart/kubeapps/README.md
juan131: 👍

chart/kubeapps/Chart.yaml
juan131: 👍

chart/kubeapps/values.yaml
juan131: 👍

chart/kubeapps/templates/ingress.yaml
juan131: Yes, I think it's better to use `multi-host` since it requires issuing many certificates

juan131:
>  In the example above, you referenced --set ingress.hosts[0].certManager=true which seems to reference to host[0] while the annotation is set at the ingress top level. Shouldn't then not be scoped inside the host?

Totally agree!! We should set it at the ingress top level.

> I am assuming that kubeapps.juan.bkpr.run is a typo and was actually meant to be kubeapps.custom.domain?

Yes, it's a typo.

> I am wondering if it would make sense to just add the certManager annotation helper but not the multi-host <-> multi-ingress support.

I'd rather user `multi-host` too, let me adapt it!

chart/kubeapps/templates/ingress.yaml
```diff
@@ -1,38 +1,38 @@
...
 metadata:
-  name: {{ $fullName }}
+  name: {{ template "kubeapps.fullname" $ }}
   labels:
-    app: {{ include "kubeapps.name" . }}
-    chart: {{ include "kubeapps.chart" . }}
-    release: {{ .Release.Name }}
-    heritage: {{ .Release.Service }}
-{{- with .Values.ingress.annotations }}
+    app: {{ template "kubeapps.name" $ }}
```
migmartri: Thanks for reverting back to template since it seems that here was the only place in which we used `include`.

migmartri: Why are you using `$` instead of current scope `.`? It seems that in the other templates we use `.`

chart/kubeapps/templates/ingress.yaml
```diff
@@ -1,38 +1,38 @@
...
+    chart: {{ template "kubeapps.chart" $ }}
+    release: {{ $.Release.Name }}
+    heritage: {{ $.Release.Service }}
   annotations:
```
migmartri: note that this key `annotations` in this new case will be always added regardless of we passing `.Values.ingress.annotations` or  `.Values.ingress.certManager ` is that ok?

chart/kubeapps/templates/ingress.yaml
```diff
@@ -1,38 +1,38 @@
...
-{{ toYaml . | indent 4 }}
-{{- end }}
+    {{- if .Values.ingress.certManager }}
+    kubernetes.io/tls-acme: "true"
+    {{- end }}
+    {{- range $key, $value := .Values.ingress.annotations }}
+    {{ $key }}: {{ $value | quote }}
```
migmartri: Why do you prefer this approach vs dumping everything like before? `{{ toYaml . | indent 4 }}`

chart/kubeapps/templates/ingress.yaml
```diff
@@ -1,38 +1,38 @@
...
+  {{- end }}
+  tls:
+  {{- range .Values.ingress.hosts }}
+  {{- if .tls }}
+  - hosts:
+    - {{ .name }}
+    secretName: {{ .tlsSecret }}
+  {{- end }}
   {{- end }}
+---
```
migmartri: is this multi-yaml new line neccessary?

chart/kubeapps/templates/tls-secrets.yaml
```diff
@@ -0,0 +1,18 @@
+{{- if .Values.ingress.enabled }}
+{{- range .Values.ingress.secrets }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ template "kubeapps.fullname" $ }}
```
migmartri: isn't the name going to collide if there are two secrets?

chart/kubeapps/templates/ingress.yaml
juan131: We can use `.`. I'll update it. The `$` was necessary when we were using `{{- range ...` at the top.

chart/kubeapps/templates/ingress.yaml
juan131: I just used because it's how we use it on others charts. Anyway, it's only adding the value of quoting the value...

I think it's irrelevant to use one or the another.

chart/kubeapps/templates/ingress.yaml
juan131: No

chart/kubeapps/templates/tls-secrets.yaml
juan131: Nice catch!! I guess we can do sth like:
```
metadata:
  name: {{ .name }}
```

chart/kubeapps/templates/ingress.yaml
juan131: It's fine. The result is: no annotations

chart/kubeapps/templates/ingress.yaml
migmartri: > I just used because it's how we use it on others charts

ok, it makes sense. thanks for the explanation

We might need to move the rest of the chart to this approach in another PR since I believe that we use `{{ toYaml . | indent 4 }}` in other places.

migmartri:
BTW, now that I think about this. I am a little bit worried about the change on the way we define the hosts since it will mean a breaking change right? Meaning that if somebody today has an automated upgrade process that relies on having ingress enable it will fail right? 

In that case, if we still decide to go ahead with the change we might need to bump the mayor version not the minor one.

what do you think @prydonius @juan131 @andresmgot ?

juan131:
I agree. It breaks backwards compatibility and, therefore, it should be considered a major change.

juan131:
PR updated bumping the major version.

juan131:
IMHO doing a major release of a chart should not be a drama. 

We can add the steps you need to perform to do an upgrade if you consider it necessary but I think it's something normal since charts are improved constantly and some of these changes can break backward compatibility.

migmartri:
Actually @andresmgot raised a good point that in semver, for 0.x versions, minor version changes are considered breaking. So feel free to revert back to 0.9 (in your case since 0.8 has already been taken).

Thanks @juan131!

Issue #724, opened 2018-10-11
Implement error boundary


andresmgot:
From the design document: https://docs.google.com/document/d/1cB_tRcTVLWi274Tdps5XacEVZJ2Xb6avH1cUQ0TLWwE/

Implement error boundary for unexpected runtime rendering errors, reusing the `UnexpectedError` component.

PR #741, opened 2018-10-18
ErrorBoundary implementation

migmartri:
Adds an errorBoundary for unexpected errors applied only to the content inside the view container to keep the view context.

Please note that in development we will still see the error overlay but it can be dismissed clicking on the `x` shown in the top right corner resulting in the following screen, which is the one that will be shown in production.

selection_724

NOTE: I have updated the component margin to affect not only the bottom but all the parts, so there is regular margin when it is shown standalone (see attached screenshot). I have not considered this change to be a big deal for other visualizations but please let me know if you think otherwise.

Closes https://github.com/kubeapps/kubeapps/issues/724


dashboard/src/components/ErrorBoundary/ErrorBoundary.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+    if (props.throwError) {
+      throw exampleError;
+    }
+    return <div className="no-error" />;
+  };
+
+  // tslint:disable:no-console
+  beforeEach(() => {
+    // To avoid polluting the logs
+    console.error = jest.fn();
```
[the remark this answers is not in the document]
migmartri: ### More Granular mocking

That's a good point. But I am not sure that change is worth doing here because:

 In the first test we need to mock console.log before we mount the wrapper (since the wrapper calls console.log if there is an error and we do not want to pollute the console).

In the second test I can do what you said.

So the code would look like something:

```diff
git diff
diff --git a/dashboard/src/components/ErrorBoundary/ErrorBoundary.test.tsx b/dashboard/src/components/ErrorBoundary/ErrorBoundary.test.tsx
index 7c46cb1f..76bd8788 100644
--- a/dashboard/src/components/ErrorBoundary/ErrorBoundary.test.tsx
+++ b/dashboard/src/components/ErrorBoundary/ErrorBoundary.test.tsx
@@ -13,12 +13,11 @@ describe("ErrorBoundary around a component", () => {
   };
 
   // tslint:disable:no-console
-  beforeEach(() => {
-    // To avoid polluting the logs
-    console.error = jest.fn();
-  });
 
   it("captures any synchronous error thrown during a descendant render", () => {
+    // To avoid polluting the logs and check that it was called
+    console.error = jest.fn();
+
     const wrapper = mount(
       <ErrorBoundary>
         <BadRenderor throwError={true} />
@@ -55,7 +54,8 @@ describe("ErrorBoundary around a component", () => {
 
     // the state is null
     expect(wrapper.state()).toEqual({ error: null, errorInfo: null });
-
+    // To avoid polluting the logs and check that it was called
+    console.error = jest.fn();
     // console.error is not called
     expect(console.error).not.toHaveBeenCalled();
   });

```

Which still means that in theory we have pretty much the bulk of a test case ran with a mocked console.log.

Also, note that even having console.log mocked, errors are still shown in the console since it uses the jest logger. 

```jest
 FAIL  src/components/ErrorBoundary/ErrorBoundary.test.tsx
  ErrorBoundary around a component
    ✕ captures any synchronous error thrown during a descendant render (55ms)
    ✓ renders only the wrapped components if no error (5ms)

  ● ErrorBoundary around a component › captures any synchronous error thrown during a descendant render

    ReferenceError: thisDoesNotExist is not defined

      35 |     const errorInfo: React.ErrorInfo = wrapper.state("errorInfo");
      36 |     expect(errorInfo.componentStack.length).not.toEqual(0);
    > 37 |     thisDoesNotExist
```

### Restoring mock

In theory, jest should not propagate mocks outside of the files, and I actually thought that it should not do it outside of the `describe` but it does. So I'll add a restore. Thanks!

andresmgot:
👍 thanks for the changes and the clarification.

dashboard/src/components/ErrorAlert/UnexpectedErrorAlert.tsx
```diff
@@ -55,7 +55,7 @@ class UnexpectedErrorPage extends React.Component<IUnexpectedErrorPage> {
     // typescript does not support react's defaultProps and we are running it in strictNullChecks mode.
     // Newer versions of it seems to support it https://github.com/Microsoft/TypeScript/wiki/Roadmap#30-july-2018
     return (
-      <div className="alert alert-error margin-t-bigger">
+      <div className="alert alert-error margin-bigger">
```
prydonius: The only concern I'd have about this is that other alerts don't have the same change, it could make things look inconsistent?

dashboard/src/components/ErrorBoundary/ErrorBoundary.tsx
```diff
@@ -0,0 +1,30 @@
+import * as React from "react";
+import { UnexpectedErrorAlert } from "../ErrorAlert";
+
+interface IErrorBoundaryProps {
+  children: React.ReactChildren | React.ReactNode | string;
+}
+
+interface IErrorBoundaryState {
+  error: Error | null;
```
prydonius: why `null` and not just falsey/undefined? Mostly curious

migmartri:
Based on the received feedback, I am going to re-evaluate the UI (padding) change to double check that its impact is minimal or look for another less intrusive way.

Thanks for the review you both!

migmartri:
@prydonius I have updated the patch to instead of applying a margin all over, it only applies it vertically. 

The point of my change was to be sure that there was always some margin between the error and the footer. I think this change is acceptable, non intrusive and much better than modifying the horizontal padding.

Before:

selection_735

After:

selection_736

Could you double check it? 

Thanks!

PR #742, opened 2018-10-18
Refactor binding list

### LINKED ISSUE #455: Display status message for Service Bindings
it seems that the binding went through two transitions

```
status:
  asyncOpInProgress: false
  conditions:
  - lastTransitionTime: 2018-08-09T03:32:33Z
    message: Orphan mitigation was completed successfully
    reason: OrphanMitigationSuccessful
    status: "False"
    type: Ready
  - lastTransitionTime: 2018-08-09T03:32:46Z
    message: 'ServiceBroker returned failure; bind operation will not be retried:
      Status: 500; ErrorMessage: <nil>; Description: <nil>; ResponseError: <nil>'
    reason: ServiceBindingReturnedFailure
    status: "True"
    type: Failed
  orphanMitigationInProgress: false
  reconciledGeneration: 1
  unbindStatus: Succeeded
```

Being the last one apparently a failed state (type: Failed,  unbindStatus: Succeeded), the error message is quite useless though. 

In any case, the UI seems to show only the first transition.

selection_633

Should it show the latest one? Did the binding actually succeed? Failed?

andresmgot:
Fixes #455

This PR refactors how the binding list is represented in the Service Instance View. The summary of the changes are:

 - The columns of the table are now: Binding | Status | Message | Details
 - Since the message can be quite big it's now displayed in its own modal if the user clicks in the button "Show message". (Recommendation made by @Angelmmiguel )
 - I removed the `Extend/Collapse` button since it was adding a new row to the table which looked confusing. Instead the "Details" column is now always present. The format of these details has been refactored as well.
 - It fixes #455 since the latest condition from the list is now picked by date instead of relying in the order of the array.

If these changes are approved we should use the same approach for ordering in `ServiceInstanceCardList` and apply the "Show message" button to the service instance message as well.

Screenshots:

screen shot 2018-10-24 at 12 34 44

When clicking in "Show message":

screen shot 2018-10-17 at 18 16 34

When clicking in the "show" link of the secret a similar modal is displayed:

screen shot 2018-10-24 at 12 34 55

For the record this is how the table looks without these changes:

screen shot 2018-10-18 at 12 43 34


dashboard/src/components/BindingList/BindingListEntry.tsx
```diff
@@ -10,57 +11,66 @@ interface IBindingListEntryProps {
...
-        <td>{currentStatus}</td>
-        <td style={{ display: "flex", justifyContent: "space-around" }}>
-          <button className={"button button-primary button-small"} onClick={this.toggleExpand}>
-            Expand/Collapse
+    return (
+      <tr className="row padding-t-small">
+        <td className="col-2">{name}</td>
+        <td className="col-2">{reason}</td>
+        <td className="col-2">
+          <button className="button" onClick={this.openModal}>
```
Angelmmiguel: I think a small button will work better :)

```suggestion
          <button className="button button-small" onClick={this.openModal}>
```

dashboard/src/components/BindingList/BindingListEntry.tsx
```diff
@@ -10,57 +11,66 @@ interface IBindingListEntryProps {
...
-      <tr key={"row"}>
-        <td>
-          {namespace}/{name}
-        </td>
-        <td>{currentStatus}</td>
-        <td style={{ display: "flex", justifyContent: "space-around" }}>
-          <button className={"button button-primary button-small"} onClick={this.toggleExpand}>
-            Expand/Collapse
+    return (
+      <tr className="row padding-t-small">
```
Angelmmiguel: I'm not a big fan of modifying default `table` behaviour with flex. Is there any reason you had to use `row` and `col` classes here?

dashboard/src/components/MessageDetails/MessageDetails.tsx
```diff
@@ -0,0 +1,54 @@
...
+              maxWidth: "80%",
+            },
+          }}
+        >
+          <div className="Terminal__Top type-small">
+            <div className="Terminal__Top__Buttons">
+              <a>
+                <span
+                  className="Terminal__Top__Button Terminal__Top__Button--red"
+                  onClick={this.props.closeModal}
```
Angelmmiguel: Love that red button actually close the modal 😂

dashboard/src/components/MessageDetails/MessageDetails.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+import "./MessageDetails.css";
+
+interface IMessageDetailsProps {
+  modalIsOpen: boolean;
+  closeModal: () => Promise<any>;
+  message: string;
+}
+
+class MessageDetails extends React.Component<IMessageDetailsProps> {
```
Angelmmiguel: Could you make this component `Pure`?

dashboard/src/components/BindingList/BindingListEntry.tsx
andresmgot: not really, the result is almost the same without it, it was just to set specific sizes for the columns

dashboard/src/components/MessageDetails/MessageDetails.tsx
andresmgot: _agreed offline to make this a Stateless component_

andresmgot:
Thanks @Angelmmiguel for taking a look? Can you do a quick review again?

prydonius:
I think we should hide by default the secret values, can we either show a modal with those values or show them inline with a click of a "show" link next to the secret name?

andresmgot:
@prydonius I have applied your suggestions (check the screenshots)

andresmgot:
cc/ @Angelmmiguel

Angelmmiguel:
Sure! I'm gonna review it @andresmgot. Btw, did you remember the conversation about function vs pure components? React v16.6 comes with a `React.memo` method to create pure components based on functions: https://reactjs.org/blog/2018/10/23/react-v-16-6.html

dashboard/src/components/MessageDetails/MessageDetails.tsx
```diff
@@ -0,0 +1,52 @@
...
+        <div className="Terminal__Top type-small">
+          <div className="Terminal__Top__Buttons">
+            <a>
+              <span
+                className="Terminal__Top__Button Terminal__Top__Button--red"
+                onClick={props.closeModal}
+              />
+            </a>
+          </div>
+          <div className="Terminal__Top__Title">Message</div>
```
Angelmmiguel: [Not a blocker] Could you change this title to something more specific? I think `Message` is to general.

prydonius: +1 I would make it a prop so that it can be set by the caller

dashboard/src/components/MessageDetails/MessageDetails.tsx
```diff
@@ -0,0 +1,52 @@
+import * as React from "react";
```
prydonius: I think this component should be renamed as it is used for the secret and the error message, so a more general name would be better. Perhaps something like TerminalModal makes more sense?

andresmgot:
> React v16.6 comes with a React.memo method to create pure components based on functions

Cool, nice to know, thanks for the heads up! (we are still in react 16.4)

I applied the rest of your comments. Thanks for the reviews!

PR #770, opened 2018-10-25
Update client-go to 9.0

### LINKED ISSUE #504: Update client-go
Now that we have removed our dependency on kubecfg (#503), we can update client-go to the latest version. 

- [x] use latest client-go
- [ ] enable namespaced watching of CronJobs in AppRepository controller

andresmgot:
Ref #504

Update client-go as the first step to close #504 (enable namespaced watching of CronJobs).

There was still a reference to a kubecfg function (in the `yaml` package) so I copied it in order to remove the dependency.

I also needed to update the Golang version of the compilation container from 1.9 to ~~1.10~~ 1.11 since that version is required for the new client-go.

The commits have been split for easier revision.


Gopkg.toml
```diff
@@ -15,17 +15,17 @@
...
-  version = "5.0.1"
+  version = "9.0.0"
 
-# Hardcode some packages to the revisions locked in client-go 5.0.1
+# Hardcode some packages to the revisions locked in client-go 9.0.0
 # See https://github.com/kubernetes/client-go/blob/master/INSTALL.md#dep-not-supported-yet
 [[constraint]]
   name = "k8s.io/apimachinery"
-  revision = "019ae5ada31de202164b118aee88ee2d14075c31"
+  version = "kubernetes-1.12.1"
```
prydonius: this is great!

PR #773, opened 2018-10-29
Update job scheduling mode

### LINKED ISSUE #434: Job pods in failed state after deploying helm chart
After deploying the kubeapps helm chart, you can see a set of pods triggered by the sync jobs that are in `error` state.

```
apprepo-sync-incubator-6l2f6-b7rq9                          0/1       Error       0          3h        apprepositories.kubeapps.com/repo-name=incubator,controller-uid=18b03ceb-9998-11e8-9ad3-06254f8710a6,job-name=apprepo-sync-incubator-6l2f6
```

Although this does not have any consequence from the functional standpoint so k8s will create another pod and wait to fulfill the job's completion, it does not offer a good user experience.

To me, once of the first things that a user will do after installing the chart is to check that everything is running, and seeing some pods in error state can be confusing.

The error specifically is about not being able to connect to mongoDB due to a deployment race condition. We might be able to solve this by using an init container in the job.pod.spec that waits for mongodb to be accessible. We should set a job deadline to avoid long infinite loops. 

```
time="2018-08-06T16:45:23Z" level=fatal msg="Can't connect to mongoDB: unable to connect to MongoDB"
```

migmartri:
Updated the pod spec to restart the container on failure. That way the pod does not get re-scheduled while waiting for the MongoDB connection.

Closes #434


andresmgot:
LGTM, you need to adapt the go unit tests

cmd/apprepository-controller/controller.go
```diff
@@ -390,8 +389,11 @@ func newCronJob(apprepo *apprepov1alpha1.AppRepository) *batchv1beta1.CronJob {
 		},
 		Spec: batchv1beta1.CronJobSpec{
 			// TODO: make schedule customisable
-			Schedule:          "0 * * * *",
-			ConcurrencyPolicy: "Forbid",
+			Schedule: "0 * * * *",
+			// Set to replace as short-circuit in k8s <1.12
+			// TODO re-evaluate ConcurrentPolicy when 1.12+ is mainstream (i.e 1.14)
+			// https://github.com/kubernetes/kubernetes/issues/54870s
```
andresmgot: you added an extra `s` at the end of the url

prydonius:
lgtm

PR #775, opened 2018-10-30
docs: add developer docs for chartsvc and chart-repo

### LINKED ISSUE #754: Developer environment documentation (now that chart services are in Monocular)

prydonius:
The sources for these tools are in the Monocular repository, this adds
documentation on how to develop these tools for Kubeapps from that
repository.

fixes #754


docs/developer/chart-repo.md
```diff
@@ -0,0 +1,62 @@
...
+
+```bash
+cd $MONOCULAR_DIR
+dep ensure
+make -C cmd/chart-repo docker-build
+```
+
+This builds the `chart-repo` Docker image.
+
+### Running in development
```
andresmgot: shouldn't we just point here to the monocular repo? We are having duplicated documentation

Also this is likely to get outdated since the source code of this tool is not in this repo.

prydonius:
> shouldn't this documentation live in the Monocular repo?

These commands are in the context of testing those tools against Kubeapps - so they reference the Deployment names of `kubeapps-internal-chartsvc`, etc. I think it's useful to have this here as it is kind of confusing otherwise

docs/developer/chart-repo.md
prydonius: See https://github.com/kubeapps/kubeapps/pull/775#issuecomment-434343709.

> Also this is likely to get outdated since the source code of this tool is not in this repo.

 But how often do we need to change the development workflow?

docs/developer/chart-repo.md
```diff
@@ -0,0 +1,62 @@
...
+
+#### Option 1: Using Telepresence (recommended)
```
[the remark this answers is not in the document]
prydonius: good catch, thanks!

docs/developer/chart-repo.md
andresmgot: I know that we don't update them very often

andresmgot:
I know, it just seemed a bit weird to have those docs here. In any case they are adapted and are easy enough so I am fine having them here.

PR #778, opened 2018-10-30
chart docs: update install troubleshooting for RBAC issues

prydonius:
changes the troubleshooting with instructions to install Tiller with a
cluster-admin service account

### DIFF
--- a/chart/kubeapps/Chart.yaml
+++ b/chart/kubeapps/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kubeapps
-version: 0.9.2
+version: 0.9.3
 appVersion: DEVEL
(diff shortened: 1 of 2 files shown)


prydonius:
@andresmgot yeah I saw that, and was also wondering if it should be included or not. I'm also worried that it will pollute the FAQ a bit, and also be confusing if users try to create it when they already have it? @migmartri what do you think?

migmartri:
> @andresmgot yeah I saw that, and was also wondering if it should be included or not. I'm also worried that it will pollute the FAQ a bit, and also be confusing if users try to create it when they already have it? @migmartri what do you think?

I think it is fine to leave it out for now.

prydonius:
@migmartri yes, good point.

Issue #765, opened 2018-10-24
Tiller proxy returning wrong release


migmartri:
It seems that Tiller-proxy instead of doing a full match with release name it is doing a sub-match.

This means that if for example I have `wordpress-miguel-1` and `wordpress-miguel-2`, when in the UI I access to `/apps/ns/miguel/wordpress-miguel-1` I will actually be retrieving wordpress-miguel-2`

andresmgot:
This is a duplicate of #756 (closing that one)

prydonius:
Closing as fixed

PR #781, opened 2018-11-01
Fix get release

migmartri:
Changes the way we retrieve a single release name. Instead of listing all the releases and retrieving the last one, we use what Helm itself uses instead.

I have also imported a helper to extract error messages from GRPC so we can imitate what helm does without the need of having extra error handling.

```json
// curl tiller-proxy...
{
  "code": 404,
  "message": "release: \"wordpress-miguel23\" not found"
}
```

Also note that this method does not require passing a namespace so changes will be made in further PRs

Refs #765

### DIFF
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -22,7 +22,6 @@ import (
 	"sync"
 
 	log "github.com/sirupsen/logrus"
-	"google.golang.org/grpc"
 	"google.golang.org/grpc/status"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/helm/pkg/helm"
@@ -60,12 +59,6 @@ type Proxy struct {
 	kubeClient kubernetes.Interface
 	helmClient helm.Interface
 	listLimit  int
-}
-
-func isNotFound(err error) bool {
-	// Ideally this would be `grpc.Code(err) == codes.NotFound`,
-	// but it seems helm doesn't return grpc codes
-	return strings.Contains(grpc.ErrorDesc(err), "not found")
 }
 
 // NewProxy creates a Proxy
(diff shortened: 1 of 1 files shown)


migmartri:
> Does the test need to be changed as the error message is different now?

No, this is because the returned error by GRPC matches the previous one.

 I needed to change the updateRelease code though to remove the specific error handling for 404, which I am not sure why it is needed though since the UI takes into account the code returned not the message.

pkg/proxy/proxy.go
```diff
@@ -85,25 +78,16 @@ type AppOverview struct {
 	Status      string `json:"status"`
 }
 
+// TODO: Rename get for getRelease
+// TODO: Remove namespace since release name is unique
 func (p *Proxy) get(name, namespace string) (*release.Release, error) {
```
andresmgot: why haven't you implemented the TODOs here?

pkg/proxy/proxy.go
```diff
@@ -85,25 +78,16 @@ type AppOverview struct {
...
-	list, err := p.helmClient.ListReleases(
-		helm.ReleaseListFilter(name),
-		helm.ReleaseListNamespace(namespace),
-		helm.ReleaseListStatuses(allReleaseStatuses),
-		// Get just the latest release
-		helm.ReleaseListLimit(1),
-		helm.ReleaseListSort(int32(services.ListSort_LAST_RELEASED)),
-		helm.ReleaseListOrder(int32(services.ListSort_DESC)),
-	)
+	release, err := p.helmClient.ReleaseContent(name)
```
andresmgot: Good find. I am checking the current `helm` code and it seems that the signature of the function is different:

https://github.com/helm/helm/blob/master/cmd/helm/get.go#L90

Is that because a newer version?

migmartri: Is it? I believe that it is the same. It seems to be a variadic function that accepts zero or more options

```
	ReleaseContent(rlsName string, opts ...ContentOption) (*rls.GetReleaseContentResponse, error)
```

In Helm's `get` CMD it passes the version but looking at the code I noticed that if you do not pass the version, the code returns the last one which is what we want https://github.com/helm/helm/blob/7cad59091a9451b2aa4f95aa882ea27e6b195f98/pkg/tiller/release_content.go#L32

andresmgot: 👍 I didn't check the function definition

PR #782, opened 2018-11-01
Remove namespace usage

migmartri:
Namespaces in Helm are global, not scoped by namespace.  Here https://github.com/kubeapps/kubeapps/pull/781 we have updated the `get release` code to not to rely on listing and filtering releases, which means that we do not need to filter per namespace anymore.

This patch renames `get` for `getRelease` and updates some of those proxy handlers to stop passing the unnecessary namespace.

NOTE: This does not change the tiller proxy API which eventually we might want to revisit

Refs #765

### DIFF
--- a/chart/kubeapps/Chart.yaml
+++ b/chart/kubeapps/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kubeapps
-version: 0.9.3
+version: 0.9.4
 appVersion: DEVEL
 description: Kubeapps is a dashboard for your Kubernetes cluster that makes it easy to deploy and manage applications in your cluster using Helm
 icon: https://raw.githubusercontent.com/kubeapps/kubeapps/master/docs/img/logo.png
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -97,7 +97,7 @@ apprepository:
   syncImage:
     registry: quay.io
     repository: helmpack/chart-repo
-    tag: v1.0.0
+    tag: v1.0.1
   initialRepos:
   - name: stable
     url: https://kubernetes-charts.storage.googleapis.com
@@ -165,7 +165,7 @@ chartsvc:
   image:
     registry: quay.io
     repository: helmpack/chartsvc
-    tag: v1.0.0
+    tag: v1.0.1
   service:
     port: 8080
   # https://github.com/kubeapps/kubeapps/issues/478#issuecomment-422979262
--- a/cmd/tiller-proxy/internal/handler/handler.go
+++ b/cmd/tiller-proxy/internal/handler/handler.go
@@ -26,13 +26,12 @@ import (
 
 	"github.com/gorilla/mux"
 	"github.com/kubeapps/common/response"
-	log "github.com/sirupsen/logrus"
-	"github.com/urfave/negroni"
-	"k8s.io/helm/pkg/proto/hapi/chart"
-
 	"github.com/kubeapps/kubeapps/pkg/auth"
 	chartUtils "github.com/kubeapps/kubeapps/pkg/chart"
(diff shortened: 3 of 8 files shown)


migmartri:
@andresmgot yes, you are right, that's why I added a comment here https://github.com/kubeapps/kubeapps/pull/782/files#diff-a8ee397eb96fc3f4801cb4e038840e1bR146

I was thinking of tackling this in a different PR

andresmgot:
Yes, what I want to clarify is that we are adding a regression (which was not included in this PR but in #781). This PR leaves the code in a better status from what is in master right now so I am okay merging it.

migmartri:
A regression? Where? The functionality is the same no? If you meant that we are not using the namespace to filter a release name you are right, but in practice I believe that it does not make any difference. 

Am I missing anything?

andresmgot:
> A regression? Where? The functionality is the same no?
> I'm also confused about the regression @andresmgot mentions, let's hold this until we find out what he meant.

Sorry, what I meant is that previously a `GET /namespaces/foo/releases/bar` returned an error if `bar` is not in `foo` while right now it doesn't. As I said, that is not related to this PR but the previous one. That's why I accepted the PR.

Since, as @prydonius comments, the `namespace` is necessary to check if the user has the necessary permissions to perform an action I think we can remove the `TODO` comment and, if we want to keep the previous behavior, we can add a simple check in `getRelease` that returns an error if the release is not a namespace.

(Unchecked) The above seems important because for example, when upgrading an application, an user could try to upgrade `/namespaces/foo/releases/bar`. If he has permissions to write in `foo` but not in the real namespace of `bar` (`bar` may belong to a different user) the proxy will do the operation while it should fail for two reasons: `bar` is not in `foo` and the user doesn't have permissions to write in the namespace of `bar`.

prydonius:
> (Unchecked) The above seems important because for example, when upgrading an application, an user could try to upgrade /namespaces/foo/releases/bar. If he has permissions to write in foo but not in the real namespace of bar (bar may belong to a different user) the proxy will do the operation while it should fail for two reasons: bar is not in foo and the user doesn't have permissions to write in the namespace of bar.

Great catch, @andresmgot - you're absolutely right that this change would leave this security hole open. We should definitely ensure we don't introduce this regression in this PR.

migmartri:
> (Unchecked) The above seems important because for example, when upgrading an application, an user could try to upgrade `/namespaces/foo/releases/bar`. If he has permissions to write in `foo` but not in the real namespace of `bar` (`bar` may belong to a different user) the proxy will do the operation while it should fail for two reasons: `bar` is not in `foo` and the user doesn't have permissions to write in the namespace of `bar`.

Thank you @andresmgot for the explanation, I agree with you that it is a regression so I've rolled back the previous implementation to just focus on fixing it.

Basically what I do is to retrieve the release with the "new way" but then confirm that the namespace of that release is the one we provided (if any) and previously verified in the handler. Let me know if it makes sense.

In any case, I have mixed feelings about the current API, for example I do not think that we should allow/require passing a `namespace` in the upgrade route but instead get the namespace from the referenced release. Mainly because now it is confusing that even though I am passing a namespace, the only purpose is to verify access but it does not actually have any effect in the upgrade process.  

In any case, this is not in my opinion worth discussion now.

Thank you for raising this issue @andresmgot!!

andresmgot:
> In any case, I have mixed feelings about the current API, for example I do not think that we should allow/require passing a namespace in the upgrade route but instead get the namespace from the referenced release

Yes, the only case in which it's necessary is when deploying a new app so I wouldn't mind changing the routes and make the namespace a field of the request body in that case.

Anyway, what I like of the current routes is that they are not helm specific (even though the proxy is specific to tiller). What I mean is that the routes would be the same if the backend is an App CRD controller for example (in case that for App CRDs names are not cluster-unique) so hypothetically we could extend this proxy to use different backends.

PR #787, opened 2018-11-05
namespace apprepo-controller kube informer

### LINKED ISSUE #504: Update client-go

prydonius:
The NewFilteredSharedInformerFactory was not available in previous versions
of client-go we were using. Now that we've upgraded, we can correctly
filter down to the namespace this controller is interested in
(the one Kubeapps is installed in).

chart: remove clusterrole for apprepository-controller

Now that we are able to watch on the specific namespace Kubeapps was
installed in, we no longer need the clusterrole to watch cluster-wide
CronJobs.

fixes #504

### DIFF
--- a/chart/kubeapps/templates/apprepository-rbac.yaml
+++ b/chart/kubeapps/templates/apprepository-rbac.yaml
@@ -1,44 +1,4 @@
 {{- if .Values.rbac.create -}}
-# Need a cluster role because client-go v5.0.1 does not support namespaced
-# informers
-# TODO: remove when we update to client-go v6.0.0
-apiVersion: rbac.authorization.k8s.io/v1beta1
-kind: ClusterRole
-metadata:
-  name: {{ template "kubeapps.apprepository.fullname" . }}
-  labels:
-    app: {{ template "kubeapps.apprepository.fullname" . }}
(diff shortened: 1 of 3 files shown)


prydonius:
@migmartri I'm not sure, the tests are currently for controller.go only, and I don't know if we can mock the shared informers?

andresmgot:
I don't think we need an unit test for this change (since the call is syntactically correct).

What I don't fully understand is why do we need to listen for CronJobs: Apparently is to be safe and recreate a CronJob that someone deletes manually, am I right? If we really care about that we could add an e2e test to check that the CronJob gets recreated but I don't think it's necessary.

migmartri:
> I don't think we need an unit test for this change (since the call is syntactically correct).
> 
> What I don't fully understand is why do we need to listen for CronJobs: Apparently is to be safe and recreate a CronJob that someone deletes manually, am I right? If we really care about that we could add an e2e test to check that the CronJob gets recreated but I don't think it's necessary.

@prydonius, would you mind adding a comment to make it more clear? Thanks!

PR #788, opened 2018-11-06
renames Chart list to Catalog list

### LINKED ISSUE #761: Rename Charts list views to Catalog
Charts is often confused with Applications, so we will rename this to Catalog to make it more clear that it is a list of available apps to install from.

prydonius:
- renames navigation link and page heading "Charts" -> "Catalog"
- renames ChartsList Container and components to CatalogList
- renames /charts and /charts/:repo routes to /catalog and /catalog/:repo
- removes ChartsListItem.scss and moves the shared CSS rules to InfoCard

fixes #761


dashboard/src/containers/RoutesContainer/Routes.tsx
```diff
@@ -26,8 +26,8 @@ const privateRoutes: {
   "/apps/ns/:namespace/:releaseName": AppViewContainer,
   "/apps/ns/:namespace/new/:repo/:id/versions/:version": AppNewContainer,
   "/apps/ns/:namespace/upgrade/:releaseName": AppUpgradeContainer,
-  "/charts": ChartListContainer,
-  "/charts/:repo": ChartListContainer,
+  "/catalog": CatalogListContainer,
+  "/catalog/:repo": CatalogListContainer,
```
prydonius: For now, I'm not changing the ChartView routes, but this looks a little bit confusing. One option is to not change this to `/catalog/:repo` and keep it as `/charts/:repo`, but keep `/catalog`, but then the container is called CatalogListContainer. I'm not sure what the best option here is.

dashboard/src/components/CatalogList/CatalogList.test.tsx
```diff
@@ -8,7 +8,7 @@ import { CardGrid } from "../Card";
 import { MessageAlert } from "../ErrorAlert";
 import PageHeader from "../PageHeader";
 import SearchFilter from "../SearchFilter";
-import ChartList from "./ChartList";
+import CatalogList from "./CatalogList";
```
migmartri: I'd rename this to actually AppsCatalog so we do not overload `Catalog` and leave the door open for other kind of catalogs :)

dashboard/src/components/InfoCard/InfoCard.scss
```diff
@@ -32,3 +32,11 @@
 .ListItem__content__info_tag-2 {
   background-color: #1598cb;
 }
+
+.ListItem__content__info_tag.stable {
```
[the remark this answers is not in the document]
prydonius: The issue was that it's also being imported by the ServiceCatalogList components, but maybe since they will eventually be combined we can keep it as it was.

dashboard/src/components/CatalogList/CatalogList.test.tsx
prydonius: @migmartri the idea is that eventually we will include the ServiceCatalog classes in the Catalog view as well, right? In that case, it won't make sense for it to be called AppsCatalog, but I don't mind calling it that for now and then changing it when we get to that point.

migmartri: That's a good point. Catalog is fine then. Thanks!

Issue #766, opened 2018-10-24
New app deployment not updating 


migmartri:
The application deployment page stays in "Deploying" state and also shows a stale view of the deployment table as shown below.

![selection_731](https://user-images.githubusercontent.com/24523/47425641-41d98300-d78b-11e8-857b-3b025753b417.png)

We need to check the websocket watchers in that and other pages.

andresmgot:
Yes, at some point the web socket that is supposed to receive updates gets stuck, we should investigate why.

migmartri:
It seems that the websockets connection closes (finishes) after 1 minute.

![selection_746](https://user-images.githubusercontent.com/24523/48102101-4c116d80-e1de-11e8-8a9e-871ed519f8e7.png)

PR #790, opened 2018-11-07
Increase websockets idle timeout

migmartri:
The main issue was that websocket connections were being closed after 60 seconds of idle time.

selection_746

This was because Nginx, which we use as a proxy has 60 seconds timeout by default. I have increased that idle timeout to 1 hour only in the `ws` location.

I have tested it in my local deployment.

Fixes https://github.com/kubeapps/kubeapps/issues/766


chart/kubeapps/templates/kubeapps-frontend-config.yaml
```diff
@@ -38,6 +38,9 @@ data:
         # Hide Www-Authenticate to prevent it triggering a basic auth prompt in
         # the browser with some clusters
         proxy_hide_header Www-Authenticate;
+        # Keep the connection open with the API server even if idle.
+        # Default 60 seconds, see https://github.com/kubeapps/kubeapps/issues/766
```
prydonius: This comment is confusing, sounds like you're saying that we're defaulting it to 60s.

```suggestion
        # nginx defaults this to 60 seconds, we override it to 1h. See https://github.com/kubeapps/kubeapps/issues/766
```

prydonius:
Great find @migmartri, thanks for fixing!

prydonius:
lgtm!

Issue #757, opened 2018-10-23
The error of conflicting release names changes dynamically as you fix the error


arapulido:
Steps to reproduce:

* Create a release called `foo`
* Try to create a second release called `foo`
* You get an error that `foo` is already taken:
<img width="821" alt="screen shot 2018-10-23 at 11 46 54 am" src="https://user-images.githubusercontent.com/925437/47352012-6533fd80-d6b9-11e8-8164-ebb35d11e499.png">

* Change the name to `bar`

Expected results:
* The error goes away or it ignores your change

Actual result:
* The error message changes to complain about `bar`:

<img width="665" alt="screen shot 2018-10-23 at 11 48 05 am" src="https://user-images.githubusercontent.com/925437/47352097-885ead00-d6b9-11e8-9001-7b2f33f4ea54.png">

PR #791, opened 2018-11-07
Show correct release name in case of error

migmartri:
This patch decouples the releaseName used in the controlled field in the form from the releaseName shown in the error message.
peek 2018-11-07 11-46

Fixes https://github.com/kubeapps/kubeapps/issues/757

### DIFF
--- a/dashboard/src/components/DeploymentForm/DeploymentForm.test.tsx
+++ b/dashboard/src/components/DeploymentForm/DeploymentForm.test.tsx
@@ -60,12 +60,35 @@ describe("renders an error", () => {
         error={new UnprocessableEntity("wrong format!")}
       />,
     );
-    wrapper.setState({ releaseName: "my-app" });
+    wrapper.setState({ latestSubmittedReleaseName: "my-app" });
     expect(wrapper.find(ErrorSelector).exists()).toBe(true);
     expect(wrapper.find(ErrorSelector).html()).toContain(
       "Sorry! Something went wrong processing my-app",
     );
     expect(wrapper.find(ErrorSelector).html()).toContain("wrong format!");
+  });
+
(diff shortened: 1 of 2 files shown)


dashboard/src/components/DeploymentForm/DeploymentForm.tsx
```diff
@@ -102,7 +107,7 @@ class DeploymentForm extends React.Component<IDeploymentFormProps, IDeploymentFo
   public render() {
     const { selected, bindingsWithSecrets, chartID, chartVersion, namespace } = this.props;
     const { version, versions } = selected;
-    const { appValues, releaseName } = this.state;
+    const { appValues } = this.state;
```
prydonius: why not grab `latestSubmittedReleaseName` here too?

dashboard/src/components/DeploymentForm/DeploymentForm.test.tsx
```diff
@@ -60,13 +60,36 @@ describe("renders an error", () => {
...
+        }
+        error={new UnprocessableEntity("wrong format!")}
+      />,
+    );
+
+    wrapper.setState({ latestSubmittedReleaseName: "my-app" });
+    expect(wrapper.find(ErrorSelector).exists()).toBe(true);
+    expect(wrapper.find(ErrorSelector).html()).toContain(expectedErrorMsg);
+    wrapper.setState({ releaseName: "my-app2" });
+    expect(wrapper.find(ErrorSelector).html()).toContain(expectedErrorMsg);
```
andresmgot: this assertion won't fail even if the bug is still present (since you are appending just a `2`), you'll need to use a completely different name for the `releaseName` or use `.not.toContain("my-app2")`.

PR #792, opened 2018-11-07
Fix catalog links

migmartri:
Fixes some missing links

Refs https://github.com/kubeapps/kubeapps/pull/788


migmartri:
> thanks - sorry I missed these!

No problem!

PR #796, opened 2018-11-08
Add component information in app and service lists

### LINKED ISSUE #764: Add application name and service binding class name to apps/instances listing
Currently, the deployed applications listing shows the name of the release and the icon, but does not include the application name. This means that the user needs to identify apps by its icon, which in many cases is the default placeholder. 

selection_729

The same concept should be applied to the service catalog listing that @andresmgot is working on. This should include the name of the service class.

In terms of its visualization, we might be able to add this app/instance class name before the chart version at the bottom left of the card.

andresmgot:
Fixes #764 

Add the chart name to the app list and the service name in the service instance list:

screenshot 2018-11-09 at 11 23 16
screenshot 2018-11-09 at 11 23 33

Note that changes in tiller-proxy were required in order to return the chart name in the list overview.

### DIFF
--- a/dashboard/src/components/AppList/AppListItem.tsx
+++ b/dashboard/src/components/AppList/AppListItem.tsx
@@ -20,7 +20,7 @@ class AppListItem extends React.Component<IAppListItemProps> {
         link={`/apps/ns/${app.namespace}/${app.releaseName}`}
         title={app.releaseName}
         icon={icon}
-        info={app.version || "-"}
+        info={`${app.chart}/v${app.version || "-"}`}
         tag1Content={app.namespace}
         tag2Content={app.status.toLocaleLowerCase()}
         tag2Class={app.status.toLocaleLowerCase()}
--- a/dashboard/src/components/AppList/AppListItem.tsx
+++ b/dashboard/src/components/AppList/AppListItem.tsx
@@ -20,7 +20,7 @@ class AppListItem extends React.Component<IAppListItemProps> {
         link={`/apps/ns/${app.namespace}/${app.releaseName}`}
         title={app.releaseName}
         icon={icon}
-        info={`${app.chart}/v${app.version || "-"}`}
+        info={`${app.chart} v${app.version || "-"}`}
         tag1Content={app.namespace}
         tag2Content={app.status.toLocaleLowerCase()}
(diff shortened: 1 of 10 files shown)


dashboard/src/components/AppList/AppListItem.tsx
```diff
@@ -20,7 +20,7 @@ class AppListItem extends React.Component<IAppListItemProps> {
         link={`/apps/ns/${app.namespace}/${app.releaseName}`}
         title={app.releaseName}
         icon={icon}
-        info={app.version || "-"}
+        info={`${app.chart}/v${app.version || "-"}`}
```
prydonius: the / version is confusing, for a second I thought it was repo/name. Maybe we could have a space between these instead?

migmartri: or maybe `colon` which seems more widespread? Alternative I'd go for the version in parenthesis

migmartri: Why do we need `-`? `${app.version || "-"}`

dashboard/src/components/ServiceInstanceList/ServiceInstanceCardList.tsx
```diff
@@ -45,6 +45,7 @@ const ServiceInstanceCardList: React.SFC<IServiceInstanceCardListProps> = props
                   namespace={instance.metadata.namespace}
                   link={link}
                   icon={icon}
+                  serviceClassName={(svcClass && svcClass.spec.externalName) || "-"}
```
migmartri: wondering if to be safe you should also check `svcClass.spec`

dashboard/src/components/AppList/AppListItem.tsx
prydonius: I think we should avoid confusing this with the chart version, the Helm CLI uses `{chart.name}-{chart.version}`. IMO a space will be the least confusing.

andresmgot: Actually I think that the separation between the two items using just a space is not very visible, let me try using two different lines

andresmgot: nvm, I don't like the result of printing them as a column:
screenshot 2018-11-09 at 12 14 17

PR #797, opened 2018-11-08
relax YAML parsing for Helm release manifests

### LINKED ISSUE #636: Relax yaml parsing in releases/view page

prydonius:
- enables JSON.parse behaviour for YAML parsing (allows manifests with
  duplicate keys)
- switches to yaml.loadAll as safeLoadAll does not seem to support the
  JSON.parse behaviour (https://github.com/kubeapps/kubeapps/issues/636#issuecomment-437197500). Ideally we can switch this back once this is
  fixed upstream.

fixes #636

### DIFF
--- a/dashboard/src/components/AppView/AppView.test.tsx
+++ b/dashboard/src/components/AppView/AppView.test.tsx
@@ -144,6 +144,7 @@ describe("AppViewComponent", () => {
       expect(sockets.length).toEqual(0);
     });
 
+    // See https://github.com/kubeapps/kubeapps/issues/632
     it("handles manifests with duplicate keys", () => {
       const wrapper = shallow(<AppViewComponent {...validProps} />);
       const manifest = `
--- a/dashboard/src/components/AppView/AppView.test.tsx
+++ b/dashboard/src/components/AppView/AppView.test.tsx
@@ -145,7 +145,7 @@ describe("AppViewComponent", () => {
     });
 
     // See https://github.com/kubeapps/kubeapps/issues/632
-    it("handles manifests with duplicate keys", () => {
+    it("supports manifests with duplicated keys", () => {
       const wrapper = shallow(<AppViewComponent {...validProps} />);
       const manifest = `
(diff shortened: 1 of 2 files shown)


dashboard/src/components/AppView/AppView.test.tsx
```diff
@@ -143,6 +143,24 @@ describe("AppViewComponent", () => {
       const sockets: WebSocket[] = wrapper.state("sockets");
       expect(sockets.length).toEqual(0);
     });
+
+    // See https://github.com/kubeapps/kubeapps/issues/632
+    it("handles manifests with duplicate keys", () => {
```
migmartri: nitpick: `"handles manifests with duplicate keys"` sounds too generic to me, I mean, handles it how? By capturing the error? By rendering it?

It might make sense to say something like `it supports manifests with duplicated keys`

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -78,7 +78,7 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
     if (!newApp) {
       return;
     }
-    let manifest: IResource[] = yaml.safeLoadAll(newApp.manifest);
+    let manifest: IResource[] = yaml.loadAll(newApp.manifest, undefined, { json: true });
```
migmartri: Can you add a TODO statement about the upgrade? Thanks!

prydonius: done thanks

PR #799, opened 2018-11-09
Remove upgrade hook

### LINKED ISSUE #699: Remove temporary upgrade hook
Follow up of https://github.com/kubeapps/kubeapps/pull/691

The upgrade hook is only needed when upgrading from a version less than 0.6 to a version greater than 0.6.0. After giving users some time to upgrade Kubeapps we should remove this temporary code.

andresmgot:
Closes #699 

When releasing the next version we should point users to `docs/user/migrating-to-v1.0.0.md` just in case.


andresmgot:
Agree, that's fine

Issue #625, opened 2018-09-12
all repositories have disapeared


obeyler:
Sorry I forgot to take immediatly the snapshot where I can see "Error"
I lost all repository ( Without any re deployent of the chart)

![image](https://user-images.githubusercontent.com/23224162/45456180-1b371f80-b6ea-11e8-9e9c-a5951a8efd16.png)

![image](https://user-images.githubusercontent.com/23224162/45456118-d9a67480-b6e9-11e8-834d-7090fa39175a.png)

prydonius:
What's the output of `kubectl get crd`?

obeyler:
@prydonius 
 kubectl get crd
NAME                           AGE
apprepositories.kubeapps.com   10h

andresmgot:
mm, this issue seems indeed related to #624 . Something has gone wrong with an uninstallation of Kubeapps and the `apprepositories.kubeapps.com` CRD is probably in `Deleting` state. What does it return `kubectl get -o yaml crd apprepositories.kubeapps.com` ?

If you had `apprepositories` and they disappeared it may be for a different side effect. If the clean up job from a previous installation didn't succeed and the job is being restarted even when a new installation takes place the clean up job may delete the new `apprepositories`. We should probably add a check to avoid that but it's something difficult to get. The clean up job may have failed if the CRD `apprepositories.kubeapps.com` has been deleted before uninstalling kubeapps.

In any case, to fix these issues I would recommend to uninstall the chart, delete the `kubeapps` namespace and ensure that everything is clean before reinstalling the chart.

obeyler:
I worried about 

> "In any case, to fix these issues I would recommend to uninstall the chart, delete the kubeapps namespace and ensure that everything is clean before reinstalling the chart."

I can do that but in our system as we have a full automatized installation for Kubernetes/ other somponents/and  all charts each part is launch automatically on any change 
so adding a manual step for specif chart...

obeyler:
```
kubectl get -o yaml crd apprepositories.kubeapps.com
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  annotations:
    helm.sh/resource-policy: keep
  creationTimestamp: 2018-09-12T21:45:51Z
  generation: 1
  labels:
    app: kubeapps-apprepository-controller
    chart: kubeapps-0.4.1
    heritage: Tiller
    release: kubeapps
  name: apprepositories.kubeapps.com
[...]

PR #801, opened 2018-11-09
Avoid CRD deletion when upgrading

### LINKED ISSUE #800: Upgrading the Kubeapps chart deletes the apprepositores CRD
Somehow the upgrade process is deleting the apprepositories CRD for me:

```
▶ helm install --name kubeapps  --namespace kubeapps bitnami/kubeapps
NAME:   kubeapps
LAST DEPLOYED: Fri Nov  9 13:18:13 2018
NAMESPACE: kubeapps
STATUS: DEPLOYED
...
▶ kubectl get crd
NAME                           CREATED AT
apprepositories.kubeapps.com   2018-11-09T12:14:36Z
▶ helm upgrade --namespace kubeapps kubeapps bitnami/kubeapps
Release "kubeapps" has been upgraded. Happy Helming!
LAST DEPLOYED: Fri Nov  9 13:18:47 2018
NAMESPACE: kubeapps
STATUS: DEPLOYED
...
▶ kubectl get crd
No resources found.
```

andresmgot:
Fixes #800


arapulido:
What versions of the chart are affected by this issue?

andresmgot:
The code related to the CRD was changed in the version `0.3.0` but I doubt this is an issue since then (Aug 23) we should have noticed before. Let's see if @migmartri or @prydonius know something else about this issue.

migmartri:
Thank you @andresmgot for noticing the issue and fixing it.

> The code related to the CRD was changed in the version 0.3.0 but I doubt this is an issue since then (Aug 23). We would have noticed before.

I am afraid that the issue might have been actually there and this is the reason I have not noticed it https://github.com/kubeapps/kubeapps/issues/800#issuecomment-437521909

There is a report of a potential instance of the error here https://github.com/kubeapps/kubeapps/issues/625 and I saw it once but could not reproduce.

andresmgot:
I understand the issue. In any case, CRDs are cluster wide so in any case, several instances of Kubeapps needs to be upgraded if we change the version of the CRD. I am landing this PR to fix the original issue. Thanks for the input!

PR #802, opened 2018-11-09
Add link for repo tags

### LINKED ISSUE #580: [Feature Request] Ability to filter charts by repository....
Hi,

On the charts page it would be great to be able to filter to only include charts from certain repositories.

image

So for instance I would probably have this set up to exclude all incubator charts.

andresmgot:
Fixes #580 

Add the repo tag as a link to filter by repository.


dashboard/src/components/Catalog/CatalogItem.tsx
```diff
@@ -4,6 +4,7 @@ import placeholder from "../../placeholder.png";
 import { IChart } from "../../shared/types";
 import InfoCard from "../InfoCard";
 
+import { Link } from "react-router-dom";
```
prydonius: really minor nitpick: can we move this import to group it with the react import? I like having the library imports grouped separately from the local imports.

Issue #767, opened 2018-10-25
Configure chart repo HTTP requests with Kubeapps-specific user-agent and version


prydonius:
We currently interact with chart repositories over HTTP in two components:

1. chart-repo
Regularly syncs chart repositories (requesting the index.yaml files). Currently uses a user-agent for the `chart-repo` tool, but this will be shared with deployments of Monocular, so we should make this configurable in `chart-repo` to set a Kubeapps specific user-agent (e.g. `chart-repo/v1.0.0 (Kubeapps v1.0.0)`.
1. tiller-proxy
On chart installs, fetches the index.yaml and tarball for the chart. Currently this doesn't set any specific user-agent (as far as I can see in https://github.com/kubeapps/kubeapps/blob/master/pkg/chart/chart.go, @andresmgot let me know if I'm missing something), so we should set the user-agent here to something specific (e.g. `tiller-proxy/v1.0.0 (Kubeapps v1.0.0)`

This will improve logging and debugging interactions with chart repositories so we can clearly see what components might be having issues with chart repos. These logs with also give chart repository owners a better indication of how many people are using Kubeapps and installing charts through it with their chart repositories.

migmartri:
@prydonius in your example

> chart-repo/v1.0.0 (Kubeapps v1.0.0).

Is any of those the chart version? How important do you consider that it is to have the chart-repo version for example?

Looking at the "standard" way of defining user agents `User-Agent: <product> / <product-version> <comment>`

Would it make sense make the product kubeapps or monocular while in the comment mentioning the source (i.e tiller-proxy?)

prydonius:
@migmartri Sorry for the confusion, also I think this was written before chart-repo was moved to the Monocular repo. The v1.0.0 after the chart-repo is the chart-repo version (currently v1.0.1), and then the version after the Kubeapps would be the Kubeapps version (currently v1.0.0-beta.4).

> Would it make sense make the product kubeapps or monocular while in the comment mentioning the source (i.e tiller-proxy?)

I think this could make sense, but there should be a default case where you run the tool standalone (e.g. it is sometimes useful to run chart-repo standalone to debug).

migmartri:
Thanks @prydonius 

In my head we have a couple of options if we follow this pattern `User-Agent: <product> / <product-version> <comment>`

|desc|option 1|option 2|option 3|
|----|----|----|---|
[...]

PR #804, opened 2018-11-10
Add user-agent-comment support in chart repo controller

migmartri:
chart-repo 1.0.2 supports a new flag that appends the provided value into its userAgent comment.

This patch adds support in the app repo controller to also receive that flag and forward it downstream to both the k8s jobs and cronjobs.

I am not making the changes in the helm chart yet because it requires that we release kubeapps first otherwise it will not find the flag.
 
Refs #767


chart/kubeapps/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kubeapps
-version: 0.9.5
+version: 0.9.6
```
prydonius: let's separate the chart changes from the code changes, we won't want to release a new chart with this yet

prydonius: also, I'm confused, you've bumped the chart but haven't added the flag?

PR #809, opened 2018-11-13
namespace apprepo-controller kube informer

prydonius:
The NewFilteredSharedInformerFactory was not available in previous versions
of client-go we were using. Now that we've upgraded, we can correctly
filter down to the namespace this controller is interested in
(the one Kubeapps is installed in).

chart: remove clusterrole for apprepository-controller

Now that we are able to watch on the specific namespace Kubeapps was
installed in, we no longer need the clusterrole to watch cluster-wide
CronJobs.

Note: we should hold this until we are ready to cut 1.0.


prydonius:
@andresmgot @migmartri there are a couple of chart changes we're going to need to merge just for the 1.0 release. This is one of them, #799 is another, and @migmartri has changes coming for the user agent flag. For each of these, we **should not** bump the chart version. When we are ready to release, we should merge each change, tag, and then bump the chart version once to sync the release. We should think about a way to handle this in the future, and add something to the release process doc.

andresmgot:
Thanks for handling this.

PR #810, opened 2018-11-13
Tiller proxy user agent support

migmartri:
User Agent support in Tiller-proxy, 

* Support to bake the tiller-proxy version during release
* Support to receive user-agent comment that will injected via the chart
* Updated circle-ci config to bake the version

See below some example of the user-agent seen at the repository end

Tiller-proxy built in the CI during 1.2 release and ran inside the version 1.1.1 of the Kubeapps chart

```
> 100.100.0.23 - - [13/Nov/2018:21:53:30 +0000] "GET /index.yaml HTTP/1.1" 200 1555 "-" "tiller-proxy/1.2 (Kubeapps/v1.1.1)
```
Tiller built locally and run outside of the scope of Kubeapps
```
> 100.117.0.48 - - [13/Nov/2018:21:56:42 +0000] "GET /index.yaml HTTP/1.1" 200 1555 "-" "tiller-proxy/8a1ed0f59f59d4599307c96c23bd1da4d833946c
```

Refs #767

### DIFF
--- a/cmd/tiller-proxy/version.go
+++ b/cmd/tiller-proxy/version.go
@@ -26,6 +26,12 @@ var (
 	userAgentComment string
 )
 
+// clientWithDefaultUserAgent implements chart.HTTPClient interface
+// and includes an override of the Do method which injects an User-Agent
+type clientWithDefaultUserAgent struct {
+	http.Client
+}
+
 // Returns the user agent to be used during calls to the chart repositories
 // Examples:
 // tiller-proxy/devel
--- a/cmd/tiller-proxy/version.go
+++ b/cmd/tiller-proxy/version.go
@@ -1,0 +1,47 @@
+/*
+Copyright (c) 2018 Bitnami
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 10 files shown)


cmd/tiller-proxy/version.go
```diff
@@ -0,0 +1,47 @@
...
+// More info here https://github.com/kubeapps/kubeapps/issues/767#issuecomment-436835938
+func userAgent() string {
+	ua := "tiller-proxy/" + version
+	if userAgentComment != "" {
+		ua = fmt.Sprintf("%s (%s)", ua, userAgentComment)
+	}
+	return ua
+}
+
+func (c *clientWithDefaultUserAgent) Do(req *http.Request) (*http.Response, error) {
```
prydonius: hmm it's a little confusing to me to not have this with the struct definition. Do you have a reason to declare this here instead of main?

cmd/tiller-proxy/version_test.go
```diff
@@ -0,0 +1,87 @@
...
+	}))
+	// Close the server when test finishes
+	defer server.Close()
+
+	serverURL, _ := url.Parse(server.URL)
+
+	netClient = &clientWithDefaultUserAgent{}
+	_, err := netClient.Do(&http.Request{
+		URL:    serverURL,
+		Header: map[string][]string{},
```
prydonius: is this line necessary?

migmartri: > 		Header: map[string][]string{},

Yes, it seems that it expects the slice to be defined otherwise it panics.

cmd/tiller-proxy/version.go
migmartri: No, no reason, good point.

Issue #806, opened 2018-11-13
Refactor ServiceInstanceView/ServiceClassView to only fetch relevant ClusterServiceClass and ClusterServicePlan


prydonius:
Currently, we are fetching all classes and plans in the ServiceInstanceView component and filtering them by the currently selected ServiceInstance:

https://github.com/kubeapps/kubeapps/blob/e85450e9695a985e97422f3c44074f057df70531/dashboard/src/components/ServiceInstanceView/ServiceInstanceView.tsx#L88-L89

https://github.com/kubeapps/kubeapps/blob/e85450e9695a985e97422f3c44074f057df70531/dashboard/src/components/ServiceInstanceView/ServiceInstanceView.tsx#L141-L148

https://github.com/kubeapps/kubeapps/blob/e85450e9695a985e97422f3c44074f057df70531/dashboard/src/components/ServiceInstanceView/ServiceInstanceView.tsx#L153-L159

If we have the clusterServiceClassRef and clusterServicePlanRef, we should just be able to make a direct call to the k8s API for the specific class and plan we need to display, reducing the scope needed for the API call and making the processing simpler on the frontend.

prydonius:
It looks like we're also fetching all instances, when we should just be fetching the requested instance.

andresmgot:
Modified the title to also include the ServiceClassView that present the same issue.

PR #815, opened 2018-11-14
reorganise service instance view layout

### LINKED ISSUE #688: Make Service instances page look more like the apps view
screen shot 2018-09-05 at 16 47 22

We should align it more with the Application view page

prydonius:
- aligns the page with the AppView layout
- creates two new components
  - ServiceInstanceInfo - displays a card with info about the Service Instance
  - ServiceInstanceStatus - renders the current status for the Service Instance
- adds a warning above the Service Instance View page to explain that
  the page needs to be refreshed for updates and that the Service
  Catalog integration is experimental.kkk

fixes #688 

screen shot 2018-11-13 at 16 15 08

screen shot 2018-11-13 at 16 16 22

screen shot 2018-11-13 at 16 14 18

screen shot 2018-11-13 at 16 17 45

### DIFF
--- a/dashboard/src/components/ServiceInstanceView/ServiceInstanceInfo.tsx
+++ b/dashboard/src/components/ServiceInstanceView/ServiceInstanceInfo.tsx
@@ -1,0 +1,77 @@
+import * as React from "react";
+
+import placeholder from "../../placeholder.png";
+import { IClusterServiceClass } from "../../shared/ClusterServiceClass";
+import { IServicePlan } from "../../shared/ServiceCatalog";
+import { IServiceInstance } from "../../shared/ServiceInstance";
+import Card, { CardContent, CardGrid, CardIcon } from "../Card";
+
+interface IServiceInstanceInfoProps {
+  instance: IServiceInstance;
+  svcClass?: IClusterServiceClass;
+  plan?: IServicePlan;
+}
+
+class ServiceInstanceInfo extends React.Component<IServiceInstanceInfoProps> {
+  public render() {
+    const { instance, svcClass, plan } = this.props;
+    const name = instance.metadata.name;
+    const externalMetadata = svcClass && svcClass.spec.externalMetadata;
+    const imageUrl = (externalMetadata && externalMetadata.imageUrl) || placeholder;
+
+    return (
+      <CardGrid className="ServiceInstanceInfo">
+        <Card>
+          <CardIcon icon={imageUrl} />
+          <CardContent>
+            <h5>{name}</h5>
+            {svcClass && this.renderSvcClassInfo(svcClass)}
+            {plan && this.renderPlanInfo(plan)}
+          </CardContent>
+        </Card>
+      </CardGrid>
+    );
+  }
+
+  private renderSvcClassInfo(svcClass: IClusterServiceClass) {
+    const { spec } = svcClass;
+    const { externalMetadata } = spec;
+    const svcName = externalMetadata ? externalMetadata.displayName : spec.externalName;
+    const description = externalMetadata ? externalMetadata.longDescription : spec.description;
+
+    return (
+      <React.Fragment>
+        <strong>Class:</strong> {svcName}
+        <p>{description}</p>
+      </React.Fragment>
+    );
+  }
+
+  private renderPlanInfo(svcPlan: IServicePlan) {
+    const { spec } = svcPlan;
+    const { externalMetadata } = spec;
+    const planName = externalMetadata ? externalMetadata.displayName : spec.externalName;
+    const description =
+      externalMetadata && externalMetadata.bullets ? externalMetadata.bullets : [spec.description];
+    const free = svcPlan.spec.free ? <span>Free ✓</span> : undefined;
+    const bullets = (
+      <div>
+        <ul>
+          {description.map(bullet => (
+            <li key={bullet}>{bullet}</li>
+          ))}
+        </ul>
+      </div>
+    );
+
+    return (
+      <React.Fragment>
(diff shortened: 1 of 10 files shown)


dashboard/src/components/ServiceInstanceView/ServiceInstanceView.tsx
```diff
@@ -198,45 +193,67 @@ class ServiceInstanceView extends React.Component<IServiceInstanceViewProps> {
...
+                  action="deprovision"
+                  defaultRequiredRBACRoles={RequiredRBACRoles}
+                />
+              )}
+              {instance && (
+                <div className="row collapse-b-tablet">
+                  <div className="col-12">
+                    <MessageAlert level="warning">
+                      <div>
+                        <div>Refresh the page to update the status of this Service Instance.</div>
```
andresmgot: Maybe this message is too visible? Is kind of common that if you don't see thing being updated you can just refresh.

dashboard/src/components/ServiceInstanceView/ServiceInstanceStatus.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { IServiceInstance } from "shared/ServiceInstance";
+import Check from "../../icons/Check";
+import Compass from "../../icons/Compass";
+import "../DeploymentStatus/DeploymentStatus.css";
+
+interface IServiceInstanceStatusProps {
+  instance: IServiceInstance;
+}
+
+class ServiceInstanceStatus extends React.Component<IServiceInstanceStatusProps> {
```
andresmgot: this can be a stateless component as well

dashboard/src/components/ServiceInstanceView/ServiceInstanceInfo.tsx
```diff
@@ -0,0 +1,77 @@
...
+            <li key={bullet}>{bullet}</li>
+          ))}
+        </ul>
+      </div>
+    );
+
+    return (
+      <React.Fragment>
+        <strong>Plan:</strong> {planName}
+        <p className="type-small margin-reset margin-b-big type-color-light-blue">{free}</p>
```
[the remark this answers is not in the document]
prydonius: Yeah the markup for this stuff is a bit weird tbh, I mostly copied this from what was there already (https://github.com/kubeapps/kubeapps/pull/815/files#diff-d051617c8977a8f2efd98f9f197042bbL327), but I'll update this to skip the paragraph if it's undefined.

dashboard/src/components/ServiceInstanceView/ServiceInstanceInfo.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { IServiceInstance } from "../../shared/ServiceInstance";
+import Card, { CardContent, CardGrid, CardIcon } from "../Card";
+
+interface IServiceInstanceInfoProps {
+  instance: IServiceInstance;
+  svcClass?: IClusterServiceClass;
+  plan?: IServicePlan;
+}
+
+class ServiceInstanceInfo extends React.Component<IServiceInstanceInfoProps> {
```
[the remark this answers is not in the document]
prydonius: :+1:

dashboard/src/components/ServiceInstanceView/ServiceInstanceView.tsx
prydonius: The reason I added it is because now we have the flashing icon when it's provisioning or deprovisioning, which could mislead people to think that it will auto-update. They also might expect it to act in the same way as the AppView page. I think it's good to have there (though we can remove the line about Service Catalog integration being experimental if you think the message should be shorter). Another option is to remove the flashing provisioning/deprovisioning icon and just make it static. WDYT?

dashboard/src/components/ServiceInstanceView/ServiceInstanceStatus.tsx
prydonius: :+1:

dashboard/src/components/ServiceInstanceView/ServiceInstanceView.tsx
andresmgot: yes, maybe a more subtle thing is to remove the "flashing" behavior. In any case, is that is somehow difficult it's okay to leave it as it is.

dashboard/src/components/ServiceInstanceView/ServiceInstanceView.tsx
```diff
@@ -138,17 +138,16 @@ class ServiceInstanceView extends React.Component<IServiceInstanceViewProps> {
       }
       body = this.renderInstance(instance);
 
-      const svcClass =
+      // TODO(prydonius): We should probably show an error if the svcClass or
+      // svcPlan cannot be found for some reason.
+      svcClass =
```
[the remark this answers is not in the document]
prydonius: yes, I wanted to be minimal with the changes here, but completely agree. I actually think that the refactor needs to be deeper, we shouldn't need to filter for the class and plan here, we should receive the correct one as a prop - see https://github.com/kubeapps/kubeapps/issues/806

dashboard/src/components/ServiceInstanceView/ServiceInstanceView.tsx
prydonius: I think we can keep this for now, and we can revisit it based on feedback from @arapulido and others.

prydonius:
> Modulo using if else in the status

Refactored and now that we're returning in each conditional block, no need for the else ifs.

PR #820, opened 2018-11-14
Create CRDs as part of the bootstrap job

### LINKED ISSUE #819: unable to decode "": no kind "CustomResourceDefinition"
After implementing CRDs installed as hooks https://github.com/kubeapps/kubeapps/pull/813, the installation in GKE (at least) seems to fail with the following error. 

> Error: unable to decode "": no kind "CustomResourceDefinition" is registered for version "apiextensions.k8s.io/v1beta1"

I have been able to reproduce it in GKE. It basically fails the first time with the attached error message but the CRD gets created. After that, if you try to install it again the installation will be successful.

$ helm install -n miguel-test --namespace miguel chart/kubeapps/
Error: unable to decode "": no kind "CustomResourceDefinition" is registered for version "apiextensions.k8s.io/v1beta1"

$ k get crd
NAME                                    CREATED AT
apprepositories.kubeapps.com            2018-11-14T18:38:32Z

$helm install -n miguel-test --namespace miguel chart/kubeapps/
=> ok
```

There seems to be known issues in Helm like this one https://github.com/helm/helm/issues/2994

migmartri:
NOTE: This is an alternative approach to https://github.com/kubeapps/kubeapps/pull/821

The hook does not seem to work well with CRDs by failing due to some kind of race condition during its creation. 

This patch is an alternative to upgrading Helm and using the `crd-install` hook. Which I am going to explore in another patch.

The issue with this approach is that the chart needs to create a clusterrole and a clusterbinding during boostrap. Approach that we are trying to avoid in latest releases. The pros of this approach is that it does not require upgrading Tiller to 2.10+

Fixes #819


prydonius:
I think #821 is a preferable approach over this.

migmartri:
Abandoned in favor of https://github.com/kubeapps/kubeapps/pull/821

Issue #825, opened 2018-11-15
Do e2e testing against the current stable version if releasing a chart


migmartri:
Currently, when releasing a chart we do e2e testing against the code in master, this means that we are not actually testing the chart with the images published in the latest release.

This would have allowed us to detect the regression we added after releasing 1.0 that @juan131 
 reported related to app-repo-controller not being able to enqueue sync jobs.

There are different solutions here but I think that the "easiest one" would be to extend our e2e testing stage in the pipeline to also include gke tests against the latest stable image. This should only happen if we are bumping the chart version.

PR #827, opened 2018-11-15
Add e2e tests with stable images if a new chart is been released

migmartri:
* Adds e2e tests support with current version images when releasing a new chart version. These parallel jobs only run if we are releasing a new chart version.

See https://circleci.com/workflow-run/0d20793b-c606-42ce-b44a-5cb5fcbc6da7 for an example. 

Closes https://github.com/kubeapps/kubeapps/issues/825

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -110,8 +110,6 @@ exports: &exports
     # Apart from using a DEV_TAG we use a different image ID to avoid polluting the tag
     # history of the production tag
     echo "export IMG_MODIFIER=-ci" >> $BASH_ENV
-    # Latest published release
-    echo "export LATEST_RELEASE_TAG=$(git describe --abbrev=0 --tags $(git rev-list --tags --max-count=1))" >> $BASH_ENV
 build_images: &build_images
   steps:
     - setup_remote_docker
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -110,13 +133,23 @@ gke_test: &gke_test
...
         # a build of a branch in the kubeapps repository
         if [[ -z "$GKE_ADMIN" ]]; then
+          echo "Step aborted, we are not in the Kubeapps repository"
+          circleci step halt
+        fi
+
+        # Cancel job if this is a test stable release job but
+        # the chart version has not been bumped
+        if [[ -n "$TEST_LATEST_RELEASE" ]] && ! changedVersion; then
+          echo Step "aborted, we are not releasing a new version of the chart"
```
prydonius:
```suggestion
          echo "Step aborted, we are not releasing a new version of the chart"
```

migmartri:
> I don't really like the idea of adding more GKE builds, that increases the possibility of finding spurious errors due to errors in the cloud. It's also a bit confusing having a build job that is rarely triggered. Finally if we add more versions the amount of builds would be too many.

> I suggest instead to reuse the current builds and check like you do know if the version of the chart has changed with changedVersion. In that case we can use the latest version instead of latest since it's a release build. WDYT?

@prydonius @andresmgot and I have a chat about it and concluded:

Andres suggested approach has the upside of have less moving pieces and a reduced cost but has the downside that it is possible in some cases to not to notice that a chart actually does not work with the code in master.

On the other hand, my approach prevents from not noticing that a chart broke master but adds more jobs (moving pieces) and cost to the equation. Which we said that it was fine because among other reasons we do don't usually run into transient GKE errors. 

When we add support to 1.11 we are going to revisit it and potentially move to other approach like the one suggested by Andres or even running multiple tests inside the same cluster.

Issue #743, opened 2018-10-18
Refactor/test remaining Service Catalog code


andresmgot:
The following code should be refactored and at least covered by some basic unit tests before 1.0:

- Actions:
  - [X] catalog
- Components:
  - Config
    - [x] ServiceBrokerList
    - [x] ServiceCatalog
  - [X] ServiceClassList
  - ServiceClassView
    - [x] ProvisionButton
    - [x] ServiceClassView
  - ServiceInstanceList
    - [X] ServiceInstanceCard
    - [X] ServiceInstanceList
  - ServiceInstanceView
    - [x] AddBindingButton
    - [x] DeprovisionButton
    - [X] ServiceInstanceView
- Containers (unit testing is not required):
   - [X] ServiceCatalogContainer
   - [X] ServiceClassListContainer
   - [x] ServiceClassViewContainer
   - [X] ServiceInstanceListContainer
   - [X] ServiceInstanceViewContainer

andresmgot:
Regarding remaining items in the list. The most important items left are `ServiceClassView` and `ServiceClassViewContainer`. For those components, apart than writing unit test we should do similar changes than in #746. We should:

 - Add the loading wrapper (to wait until the classes are loaded)
 - Use the ErrorSelector to represent errors.
 - Change the format of the cards to be consistent with the rest (using `InfoCard`)

For the rest of components:

 - `ProvisionButton` and `AddBindingButton` require to use the ErrorSelector.
 - `DeprovisionButton` only requires unit tests.

andresmgot:
All the components have now a minimum coverage and include the latest common components. Closing this.

prydonius:
woot, great work @andresmgot!

PR #829, opened 2018-11-16
Refactor ServiceClassView

andresmgot:
Ref: #743

Review `ServiceClassView`, `ProvisionButton` and `ServiceClassViewContainer`:

 - Add tests
 - Split code into a new component `ServiceClassPlan`
 - Refactor error rendering (`ErrorSelector`)
 - Add `LoadingWrapper`

### DIFF
--- a/dashboard/src/components/ServiceClassView/ProvisionButton.test.tsx
+++ b/dashboard/src/components/ServiceClassView/ProvisionButton.test.tsx
@@ -1,0 +1,123 @@
+import { shallow } from "enzyme";
+import context from "jest-plugin-context";
+import * as React from "react";
+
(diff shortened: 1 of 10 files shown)


andresmgot:
I guess you are thinking about #806, isn't it? I think we can refactor all these calls that are generic and just specify the one we are interested in. The "problem" is that is quite a bigger change. We would need to create the required methods in `catalog.ts`, change the routes, modify the redux events... That's a bit out of the scope of this PR, I think we should handle that separately (I think that has less priority).

andresmgot:
@prydonius would you be okay to land this as it is or do you strongly think we should refactor all the calls/routes?

PR #830, opened 2018-11-16
Minor refactor of AddBinding and Deprovision buttons

andresmgot:
Ref: #743 

- Add tests
- Use `ErrorSelector`

### DIFF
--- a/dashboard/src/components/ServiceInstanceView/AddBindingButton.test.tsx
+++ b/dashboard/src/components/ServiceInstanceView/AddBindingButton.test.tsx
@@ -1,0 +1,91 @@
+import { shallow } from "enzyme";
+import context from "jest-plugin-context";
+import * as React from "react";
+
+import { JSONSchema6 } from "json-schema";
+import { ErrorSelector } from "../../components/ErrorAlert";
+import SchemaForm from "../../components/SchemaForm";
+import AddBindingButton from "./AddBindingButton";
+
+const defaultName = "my-class";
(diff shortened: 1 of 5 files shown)


dashboard/src/components/ServiceInstanceView/AddBindingButton.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+      },
+      type: "object",
+    } as JSONSchema6;
+    const wrapper = shallow(<AddBindingButton {...defaultProps} bindingSchema={newSchema} />);
+    wrapper.setState({ modalIsOpen: true, displayNameForm: false });
+
+    const form = wrapper.find(SchemaForm);
+    expect(form).toExist();
+    const schema = form.prop("schema") as any;
+    expect(schema).toEqual(newSchema);
```
prydonius: can we test that `props.addBinding` receives the right object based on whether kubeappsRawParemeters is defined or if there is a schema with other parameters?

andresmgot: I've tried but I am not able to set the form content using `simulate()`: `the formData` is always empty regardless of the parameters I set in the `simulate` call. I'll need to skip that. Let me know if you know a way of doing this.

prydonius: I don't, thanks for checking

PR #833, opened 2018-11-16
K8s helpers improvements

migmartri:
We have noticed that the existing helpers used to wait for pods to be ready might not fulfill our requirements now that we have multiple replicas of every tier. 

I have changed that method to now rely on  `rollback status` which cover all replicas and also doing such check in all deployments in the namespace, not only the ones specifically set.

I have also slightly modified the job specific check to rely now in the job status flag instead of the pod Completed one.

it also fixed a bug in which the function did not seem to return when cnt was 0  

```
k8s_wait_for_pod() {
    ...
        ((cnt=cnt-1)) || return 1
        sleep 1
    done
}
```

I tried it in development:

Job that did not complete within the timeout window

```bash
$ k8s_wait_for_job_completed miguel apprepositories.kubeapps.com/repo-name=stable
Waiting for job 'apprepositories.kubeapps.com/repo-name=stable' to be completed, will retry in 5 seconds ... 
Waiting for job 'apprepositories.kubeapps.com/repo-name=stable' to be completed, will retry in 5 seconds ... 
Waiting for job 'apprepositories.kubeapps.com/repo-name=stable' to be completed, will retry in 5 seconds ... 
Job 'apprepositories.kubeapps.com/repo-name=stable' did not complete
```

Deployments that correctly finished

```bash
$ k8s_wait_for_deployments_rollout miguel
deployment "miguel-kubeapps" successfully rolled out
deployment "miguel-kubeapps-internal-apprepository-controller" successfully rolled out
deployment "miguel-kubeapps-internal-chartsvc" successfully rolled out
deployment "miguel-kubeapps-internal-dashboard" successfully rolled out
deployment "miguel-kubeapps-internal-tiller-proxy" successfully rolled out
deployment "miguel-kubeapps-mongodb" successfully rolled out
deployment "miguel-proxy-repo-nginx" successfully rolled out

```

Fixes https://github.com/kubeapps/kubeapps/issues/831

PR #840, opened 2018-11-19
Refactor e2e tests

andresmgot:
Follow up #833 

Refactor e2e tests to wait for deployments instead of pods.

Improve completed job check.

Sample build: https://circleci.com/gh/kubeapps/kubeapps/4263 (at the end fails for debugging purposes)

### DIFF
--- a/script/e2e-test.sh
+++ b/script/e2e-test.sh
@@ -67,18 +67,18 @@ k8s_ensure_image kubeapps kubeapps-ci-internal-dashboard $DEV_TAG
 k8s_ensure_image kubeapps kubeapps-ci-internal-tiller-proxy $DEV_TAG
 
 # Wait for Kubeapps Pods
-k8s_wait_for_deployment kubeapps kubeapps-ci
-echo "Deployment kubeapps-ci ready"
-k8s_wait_for_deployment kubeapps kubeapps-ci-internal-apprepository-controller
-echo "Deployment kubeapps-ci-internal-apprepository-controller ready"
-k8s_wait_for_deployment kubeapps kubeapps-ci-internal-chartsvc
-echo "Deployment kubeapps-ci-internal-chartsvc ready"
-k8s_wait_for_deployment kubeapps kubeapps-ci-internal-tiller-proxy
-echo "Deployment kubeapps-ci-internal-tiller-proxy ready"
-k8s_wait_for_deployment kubeapps kubeapps-ci-internal-dashboard
-echo "Deployment kubeapps-ci-internal-dashboard ready"
-k8s_wait_for_deployment kubeapps kubeapps-ci-mongodb
-echo "Deployment kubeapps-ci-mongodb ready"
+deployments=(
+  kubeapps-ci
+  kubeapps-ci-internal-apprepository-controller
+  kubeapps-ci-internal-chartsvc
+  kubeapps-ci-internal-tiller-proxy
+  kubeapps-ci-internal-dashboard
+  kubeapps-ci-mongodb
+)
+for dep in ${deployments[@]}; do
+  k8s_wait_for_deployment kubeapps ${dep}
+  echo "Deployment ${dep} ready"
+done
 
 # Wait for Kubeapps Jobs
 k8s_wait_for_job_completed kubeapps apprepositories.kubeapps.com/repo-name=stable
--- a/script/e2e-test.sh
+++ b/script/e2e-test.sh
@@ -67,12 +67,25 @@ k8s_ensure_image kubeapps kubeapps-ci-internal-dashboard $DEV_TAG
 k8s_ensure_image kubeapps kubeapps-ci-internal-tiller-proxy $DEV_TAG
 
 # Wait for Kubeapps Pods
-k8s_wait_for_pod_ready kubeapps app=kubeapps-ci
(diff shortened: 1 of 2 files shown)


script/e2e-test.sh
```diff
@@ -67,12 +67,25 @@ k8s_ensure_image kubeapps kubeapps-ci-internal-dashboard $DEV_TAG
 k8s_ensure_image kubeapps kubeapps-ci-internal-tiller-proxy $DEV_TAG
 
 # Wait for Kubeapps Pods
-k8s_wait_for_pod_ready kubeapps app=kubeapps-ci
-k8s_wait_for_pod_ready kubeapps app=kubeapps-ci-internal-apprepository-controller
-k8s_wait_for_pod_ready kubeapps app=kubeapps-ci-internal-chartsvc
-k8s_wait_for_pod_ready kubeapps app=kubeapps-ci-internal-tiller-proxy
-k8s_wait_for_pod_ready kubeapps app=mongodb
-k8s_wait_for_pod_completed kubeapps apprepositories.kubeapps.com/repo-name=stable
+k8s_wait_for_deployment kubeapps kubeapps-ci
```
migmartri: One of the reason why in https://github.com/kubeapps/kubeapps/pull/833 I decided to not to pass the name of the deployment was to make the script less error prone and be sure that we always test all the deployments in place. So if we add a new service, its rollout will be tested regardless if the user forgot to update the ci scripts.

That said, @andresmgot convinced me that doing it explicitly has the benefit of being sure that the expected services are there. So I am ok keeping this pattern.

What I think we should do though is to refactor this a little bit and use a for loop.

script/e2e-test.sh
```diff
@@ -67,12 +67,25 @@ k8s_ensure_image kubeapps kubeapps-ci-internal-dashboard $DEV_TAG
...
+echo "Deployment kubeapps-ci ready"
+k8s_wait_for_deployment kubeapps kubeapps-ci-internal-apprepository-controller
+echo "Deployment kubeapps-ci-internal-apprepository-controller ready"
+k8s_wait_for_deployment kubeapps kubeapps-ci-internal-chartsvc
+echo "Deployment kubeapps-ci-internal-chartsvc ready"
+k8s_wait_for_deployment kubeapps kubeapps-ci-internal-tiller-proxy
+echo "Deployment kubeapps-ci-internal-tiller-proxy ready"
```
migmartri: > echo "Deployment kubeapps-ci-internal-tiller-proxy ready"

Isn't this part of the output of kubectl rollout? I am wondering if this `echo` is redundant.

script/libtest.sh
```diff
@@ -12,37 +12,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+set -e
```
migmartri: Oops, good find

script/libtest.sh
```diff
@@ -12,37 +12,19 @@
...
-    k8s_wait_for_pod $namespace $labelSelector Running
-}
-
-k8s_wait_for_pod_completed() {
+k8s_wait_for_deployment() {
     namespace=${1:?}
-    labelSelector=${2:?}
-    k8s_wait_for_pod $namespace $labelSelector Completed
+    deployment=${2:?}
+    echo "Waiting for deployment ${deployment} to be successfully rolled out"
```
migmartri: I see that compared to #833 you have decided not to check rollout multiple times. Any reason for that change? https://github.com/kubeapps/kubeapps/pull/833/files#diff-f92c3290968ffccd70b499c19fdeaba2R37

script/libtest.sh
```diff
@@ -12,37 +12,19 @@
...
+    kubectl rollout status --namespace $namespace deployment ${deployment}
```
migmartri: Without storing the exit code, isn't this going to return right away here because of the `set -e`? Have you tried this code locally? This is the way I tested mine https://github.com/kubeapps/kubeapps/pull/833#issuecomment-439526636

script/libtest.sh
```diff
@@ -58,3 +40,35 @@ k8s_ensure_image() {
...
+k8s_wait_for_job_completed() {
+    namespace=${1:?}
+    labelSelector=${2:?}
+
+    local -i retryTimeSeconds=${TEST_MAX_WAIT_SEC:?}
+    local -i retryTimeStepSeconds=5
+
+    echo "Wait for job completion started"
+
+    set +e
```
migmartri: Could you explain why you needed to override the `set +e`?

Also, could it wrap just the `res=kubectl` stanza?

script/e2e-test.sh
andresmgot: yes, I added the log entry to ensure that there is no error between the `rollout` execution and the end of the function, just in case we introduce a typo while checking the exit code

script/libtest.sh
andresmgot: I didn't really understood why retrying on a rollout, the command will either wait (and hang) until the rollout success or fail (and won't success even if you retry) so I am not sure why we need to retry. That may be useful in the case the deployment has not been created yet but that will never be the case here since the deployment is created with `helm install`.

script/libtest.sh
andresmgot: Yes, it will. In fact the rest of the function code (printing the exit code) will only happen if the exit code is 0. I can "disable" errors to print the error code in any case but this function should fail if the deployment fails to rollout.

PR #844, opened 2018-11-20
update v1.0.0 migration guide to reflect major breaking changes in chart

prydonius:
This patch adds instructions for performing a clean install of Kubeapps
for an easier upgrade process. Given all the changes that we have made
around the AppRepository CRD, the Helm release is likely to get into a
failed state when upgrading, so we are recommending that users perform
a clean install.


prydonius:
Once this is merged, we should update the release notes to link to this.

docs/user/migrating-to-v1.0.0.md
```diff
@@ -1,3 +1,37 @@
...
+  that was introduced in the `v1.0.0-beta.4`
+
+If you have difficulty updating to the v1.0.0 release, we recommend backing up
+any AppRepository objects (custom repositories) you may have added and perform a
+clean install of Kubeapps.
+
+To backup a custom repository, run the following command for each repository:
+
+```
+kubectl get apprepsitory -o yaml <repo name> > <repo name>.yaml
```
andresmgot: `s/apprepsitory/apprepository/` but I'd rather dump the whole list

prydonius: Dumping the whole list is not possible, due to `kubectl` complaining about the `resourceVersion` being newer when re-applying stable, bitnami, svc-cat and incubator (since the chart installs them). Well, it does technically work (you'd get the error on those repos, but any other custom ones would succeed), but it would look like something went wrong.

prydonius:
> I think we are entering into too much detail?

Agreed, will make this more succinct.

prydonius:
updated, PTAL!

docs/user/migrating-to-v1.0.0.md
```diff
@@ -1,3 +1,32 @@
 # Migration to v1.0.0
 
-This release includes a breaking change **if you are upgrading from a version prior to `v1.0.0-beta.2`**. In the version `v1.0.0-beta.2` we introduced a change in MongoDB credentials generation that requires a migration hook that is only present from the version `v1.0.0-beta.2` to `v1.0.0-beta.4`. If you are upgrading Kubeapps from an older version please upgrade first to `v1.0.0-beta.4` to run the necessary migration job.
+This release includes several breaking changes that can make upgrading fairly
+difficult. Helm 2.10 and above is also now required to install Kubeapps.
+
+If you have difficulty upgrading to the v1.0.0 release, we recommend backing up
+any AppRepository objects (custom repositories) you may have added and perform a
```
migmartri: > any AppRepository objects (custom repositories)

`custom repositories` do not fully sound to me like the ones that the user created.  We might want to be more explicit by saying, `backup any AppRepository that you have created`

docs/user/migrating-to-v1.0.0.md
```diff
@@ -1,3 +1,37 @@
 # Migration to v1.0.0
 
-This release includes a breaking change **if you are upgrading from a version prior to `v1.0.0-beta.2`**. In the version `v1.0.0-beta.2` we introduced a change in MongoDB credentials generation that requires a migration hook that is only present from the version `v1.0.0-beta.2` to `v1.0.0-beta.4`. If you are upgrading Kubeapps from an older version please upgrade first to `v1.0.0-beta.4` to run the necessary migration job.
+This release includes several breaking changes that can make upgrading fairly
```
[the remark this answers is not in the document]
prydonius: yes, it's easier for me to read that way, but not that important, can undo if you prefer otherwise.

docs/user/migrating-to-v1.0.0.md
prydonius: it does say "you may have added", does that not cover it? The reason to use "custom repositories" in brackets is because users might not actually know what an AppRepository is if they've only added repos in the UI.

migmartri: Sounds good! :+1:

PR #845, opened 2018-11-20
Add builds for GKE 1.11

### LINKED ISSUE #822: Add e2e GKE test for k8s 1.11
We should test on 1.11 in addition to 1.9 and 1.10. We should also update our docs to mention what we support (1.9, 1.10, 1.11)

### LINKED ISSUE #834: Lock the version of kubectl used in our e2e tests
Currently, in our CI, we are installing the kubectl tool via `gcloud components install kubectl` which does not allow us to pin to a specific version of the tool.

We should pin this version either by finding a way of getting it via gcloud or directly downloading it to guarantee reproducible builds.

andresmgot:
Fixes #822 
Fixes #834 

Changes:
 - Add builds for GKE 1.11
 - Pin `kubectl` version to `v1.12.0`
 - Fixes issue for the 1.10 branch that was picking the wrong version (see https://github.com/kubeapps/kubeapps/issues/822#issuecomment-440330246)
 ~~- Enabled GKE builds for any build in the repository `kubeapps/kubeapps` so we can easily tests PRs like this one.~~

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,14 +36,7 @@ workflows:
       - build_dashboard:
           <<: *build_always
       - GKE_1_9_MASTER:
-          <<: *build_on_master
-          requires:
-            - test_go
-            - test_dashboard
-            - build_go_images
-            - build_dashboard
-      - GKE_1_10_MASTER:
-          <<: *build_on_master
+          <<: *build_always
           requires:
             - test_go
             - test_dashboard
@@ -56,7 +49,28 @@ workflows:
             - test_dashboard
             - build_go_images
             - build_dashboard
+      - GKE_1_10_MASTER:
+          <<: *build_always
+          requires:
+            - test_go
+            - test_dashboard
+            - build_go_images
+            - build_dashboard
       - GKE_1_10_LATEST_RELEASE:
+          <<: *build_on_master
+          requires:
+            - test_go
+            - test_dashboard
+            - build_go_images
+            - build_dashboard
+      - GKE_1_11_MASTER:
+          <<: *build_always
+          requires:
+            - test_go
+            - test_dashboard
+            - build_go_images
+            - build_dashboard
+      - GKE_1_11_LATEST_RELEASE:
           <<: *build_on_master
           requires:
             - test_go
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -35,22 +35,22 @@ workflows:
           <<: *build_always
       - build_dashboard:
           <<: *build_always
-      - GKE_1_9_MASTER:
-          <<: *build_on_master
+      - GKE_1_9:
```
prydonius: any reason you chose to remove the MASTER suffix? I like that it is more clear

migmartri: I agree

.circleci/config.yml
```diff
@@ -110,6 +129,8 @@ exports: &exports
     # Apart from using a DEV_TAG we use a different image ID to avoid polluting the tag
     # history of the production tag
     echo "export IMG_MODIFIER=-ci" >> $BASH_ENV
+    # Kubectl version, we use the same for all the builds since typically it's backwards compatible
+    echo "export KUBECTL_VERSION=v1.12.0" >> $BASH_ENV
```
migmartri: Do you need to export KUBECTL_VERSION here in the shared context? It seems to be used only once.

Actually it might make sense to bring here the HELM_VERSION instead.

.circleci/config.yml
```diff
@@ -228,28 +252,39 @@ jobs:
...
+    <<: *gke_test
+    environment:
+      HELM_VERSION: v2.11.0
+      GKE_BRANCH: "1.10"
   GKE_1_10_LATEST_RELEASE:
     <<: *gke_test
     environment:
       HELM_VERSION: v2.11.0
-      GKE_BRANCH: 1.10
+      GKE_BRANCH: "1.10"
```
migmartri: Can you write a comment explaining why you need quotes? 

I also think that it might make sense to quote them all?

.circleci/config.yml
andresmgot: I removed the `MASTER` suffix because before this was only executed on master but now it's run in any branch. I thought of using `LATEST` but it's quite similar to `LATEST_RELEASE`. I will try with different suffixes if you think this is not very clear.

.circleci/config.yml
andresmgot: I wanted to give visibility of the version just in case it needs to be reused but it's not going to be probably the case.

You are right, we can save a few lines if we set HELM_VERSION here.

.circleci/config.yml
migmartri: > I removed the MASTER suffix because before this was only executed on master but now it's run in any branch.

Why is that? Why have we changed that behavior?

> I thought of using LATEST but it's quite similar to LATEST_RELEASE. I will try with different suffixes if you think this is not very clear.

The reason of using `LATEST_RELEASE` is because in my opinion, LATEST means latest version of the code, while just RELEASE can mean testing the new release.

I see that you used RELEASED (in past tense) I think it's fine but I still prefer the more explicit way of `LATEST_RELEASE`

PR #847, opened 2018-11-21
Add show secret button

### LINKED ISSUE #795: Add the possibility of showing Secrets in the AppView
For getting for example the password of "Wordpress" without needing to open a terminal and execute the command showed in the NOTES, we could add a button to show secrets included in the Chart:

screenshot 2018-11-08 at 16 05 34

That way users can get the application password without leaving Kubeapps (the same approach we do with Service Instances).

andresmgot:
Closes #795

Add a Secrets section after the NOTES for people to directly read credentials from Kubeapps:

gif_2018-11-21_18 27 1542821269

TODO:
 - [x] Add unit tests
 - [x] Add tests for SecretTableContainer


dashboard/src/components/AppView/SecretsTable.tsx
```diff
@@ -0,0 +1,35 @@
...
+        <table>
+          <thead>
+            <tr className="flex">
+              <th className="col-2">NAME</th>
+              <th className="col-2">TYPE</th>
+              <th className="col-7">DATA</th>
+            </tr>
+          </thead>
+          <tbody>
+            {secretKeys.map(k => (
```
migmartri: Can't you iterate `secrets` directly? We might have support for Object.entries

```
const object = {'a': 1, 'b': 2, 'c' : 3};
for (const [key, value] of Object.entries(object)) {
  console.log(key, value);
}
```

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -124,7 +131,7 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
 
   public handleEvent(e: MessageEvent) {
     const msg = JSON.parse(e.data);
-    const resource: IResource = msg.object;
+    const resource: any = msg.object;
```
migmartri: Is there any other way of setting the type without setting any? i.e using type unions?

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -103,6 +106,7 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
     const deployments = manifest.filter(d => d.kind === "Deployment");
```
migmartri: I think that eventually we should refactor this, it looks quite inefficient. 

It seems that the initial filter could be done by mapping all the resources by kind and then for the sockets initialization just iterate on the map based on a whitelist of resources.

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -14,6 +14,7 @@ import AppDetails from "./AppDetails";
 import AppNotes from "./AppNotes";
 import "./AppView.css";
 import ChartInfo from "./ChartInfo";
+import SecretTable from "./SecretsTable";
```
migmartri: I think that sooner than latter (not for this PR) we should extract the table as a High Order Component where you pass the columns, the resources and potentially a custom `item` component.

dashboard/src/components/AppView/SecretItem.tsx
```diff
@@ -0,0 +1,68 @@
...
+      </tr>
+    );
+  }
+
+  private renderSecret = (name: string) => {
+    const toggle = () => this.toggleDisplay(name);
+    const text = atob(this.props.secret.data[name]);
+    return (
+      <span className="flex">
+        <a onClick={toggle}>{this.state.display[name] ? <EyeOff /> : <Eye />}</a>
```
migmartri: I would put the eye next to the data to be shown, that means on the right side. Also, instead of showing the data size, I'd  show  "hidden"

dashboard/src/components/AppView/SecretItem.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+import { ISecret } from "../../shared/types";
+import "./SecretContent.css";
+
+interface ISecretItemProps {
+  secret: ISecret;
+}
+
+interface ISecretItemState {
+  display: { [s: string]: boolean };
```
migmartri: Could you explain with a comment what this does?

dashboard/src/components/AppView/SecretItem.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+class SecretItem extends React.Component<ISecretItemProps, ISecretItemState> {
+  public state: ISecretItemState = {
+    display: {},
+  };
+
+  public componentDidMount() {
+    const display = {};
+    Object.keys(this.props.secret.data).forEach(k => (display[k] = false));
+    this.setState(display);
```
migmartri: I do not think that you need componentDidMount here, I believe that you can use the constructor or the state class propery.

```
  public state: ISecretItemState = {
    display: {},
  };
```

I believe that you can initialize there.

dashboard/src/components/AppView/SecretItem.tsx
```diff
@@ -0,0 +1,68 @@
...
+    return (
+      <tr className="flex">
+        <td className="col-2">{secret.metadata.name}</td>
+        <td className="col-2">{secret.type}</td>
+        <td className="col-7 padding-small">{data}</td>
+      </tr>
+    );
+  }
+
+  private renderSecret = (name: string) => {
```
migmartri: The term secret seems overloaded. I'd be more specific with things like renderSecretEntry? Or at least add a comment explaining what is this method rendering exactly.

dashboard/src/components/AppView/SecretItem.tsx
```diff
@@ -0,0 +1,68 @@
...
+    const display = {};
+    Object.keys(this.props.secret.data).forEach(k => (display[k] = false));
+    this.setState(display);
+  }
+
+  public render() {
+    const { secret } = this.props;
+    let data: JSX.Element[] = [];
+    Object.keys(secret.data).forEach(k => {
+      data = data.concat(this.renderSecret(k));
```
migmartri: this does not seem to be just data anymore but rendered secret entries. `data` seems too generic.

dashboard/src/components/AppView/SecretItem.tsx
```diff
@@ -0,0 +1,68 @@
...
+            <span className="margin-l-small">{text.length} bytes</span>
+          )}
+        </span>
+      </span>
+    );
+  };
+
+  private toggleDisplay = (name: string) => {
+    const { display } = this.state;
+    display[name] = !display[name];
```
migmartri: I am not sure if this is mutating the state. To be sure you might be able to do it in one line using the spread operator in this.setState

dashboard/src/components/AppView/SecretContent.scss
```diff
@@ -0,0 +1,16 @@
+.SecretContainer {
```
migmartri: Why is this file in the parent component?

migmartri:
Thanks @andresmgot, looking good so far! Just left some implementation specific comments :)

dashboard/src/components/AppView/SecretsTable/SecretsTable.test.tsx
```diff
@@ -0,0 +1,43 @@
+import { shallow } from "enzyme";
+import * as React from "react";
+
+import { IResource } from "shared/types";
+import SecretsTable from ".";
+import SecretItem from "./SecretItem";
+
+it("renders a secret ready", () => {
```
prydonius: ready? you mean item?

dashboard/src/components/AppView/SecretsTable/SecretsTable.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+      name: "foo",
+    },
+    status: {},
+  } as IResource;
+  const wrapper = shallow(<SecretsTable secrets={secrets} />);
+  expect(wrapper).toMatchSnapshot();
+  expect(wrapper.find(SecretItem).key()).toBe("foo");
+});
+
+it("renders secrets deployments", () => {
```
prydonius: renders multiple secrets?

prydonius:
Actually, why are we creating a websocket for the Secrets here? I don't think we need to track them real time, we just need to get the current latest revision at the time of loading the page.

andresmgot:
> Actually, why are we creating a websocket for the Secrets here? I don't think we need to track them real time, we just need to get the current latest revision at the time of loading the page.

I took the short path here. It's true that it's likely that the secret is not going to change and a websocket is a bit of an overkill but since we already had socket for the rest of components I thought it was something doable. We can go down the path of implementing the methods to do a GET request though.

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -100,19 +104,27 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
...
+        case "Service":
+          sockets.push(this.getSocket("services", i.apiVersion, i.metadata.name, newApp.namespace));
+          break;
+        case "Ingress":
+          sockets.push(
+            this.getSocket("ingresses", i.apiVersion, i.metadata.name, newApp.namespace),
+          );
+          break;
+        case "Secret":
+          Secret.get(i.metadata.name, newApp.namespace).then(sec => this.handleResource(sec));
```
prydonius: We shouldn't really be getting data this way, we should be getting this from the Redux state. Yes, the WebSockets are done this way, but it was done because handling persistent connections in the Redux state is more difficult to handle (need to appropriately close sockets when unmounting the component). I'm not a fan of it, and it needs to be refactored at some point (one idea is to have a component per WebSocket that we open, similar to the way https://www.npmjs.com/package/react-lazylog handles log streaming).

For this PR, we should change this to get the Secrets using a Redux action and receiving them as props, like we do with all other regular requests. This could be similar to how we get the Secrets for ServiceBindings.

prydonius:
> I took the short path here. It's true that it's likely that the secret is not going to change and a websocket is a bit of an overkill but since we already had socket for the rest of components I thought it was something doable. We can go down the path of implementing the methods to do a GET request though.

Sure, but there is a specific reason other things need a WebSocket. In fact, it was a mistake to add Ingress as a WebSocket, I'm sorry I didn't pick it up at the time. We only really care about Deployments (so we can calculate deployment status) and Services (to a lesser extent, due to the pending IP/hostname).

andresmgot:
@prydonius I have done it as an independent workflow using Redux, please take a look to the changes.

prydonius:
@andresmgot this approach is interesting, but what did you think about my idea about simply adding the secrets as part of the `IAppState["selected"]` state? Just curious what your feedback on that was, and why you opted for doing it this way?

Also, we have a Secret class (https://github.com/kubeapps/kubeapps/blob/20d5c16a1a58b04c7c069eb7c4c33d5546f1f14a/dashboard/src/shared/Secret.ts#L39), as well as other bespoke k8s resources classes, do you think we should remove them for this new generic class?

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -240,6 +273,50 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
   private deleteApp = (purge: boolean) => {
     return this.props.deleteApp(this.props.releaseName, this.props.namespace, purge);
   };
+
+  private filterByKind = (kind: string) => {
+    const kubeItems = _.pickBy(this.props.resources, r => {
```
prydonius: IMO this component shouldn't care about all the resources, the container should filter down the Secrets that are necessary for rendering here.

prydonius: any reason to use lodash here instead of `Array.filter` and `Array.forEach`?

dashboard/src/reducers/kube.ts
```diff
@@ -0,0 +1,34 @@
...
+  items: {},
+};
+
+const kubeReducer = (
+  state: IKubeState = initialState,
+  action: KubeAction | LocationChangeAction,
+): IKubeState => {
+  switch (action.type) {
+    case getType(actions.kube.requestResource):
+      return { ...state, items: _.merge({}, state.items, action.payload) };
```
prydonius: can't the merge be done by the built-in spread operator instead of using lodash?

dashboard/src/components/AppView/AppView.tsx
andresmgot: mm, why? I mean, the container doesn't need to know what the view wants to use, isn't' it? Also the method `getResource` is generic, I would see it weird if the `resources` are pre-filtered.

Also I am thinking about using this approach to potentially display information of all the resources that an application is composed of (I think that may be convenient for issues like #860).

andresmgot: `resources` is an object not an array so we would need to use `Object.keys` + `Array.filter` + populating a new object, using `pickBy` is faster/more readable.

dashboard/src/reducers/kube.ts
andresmgot: the `action.payload` is an object with a single key so we would need to do something like:

```
      const key = Object.keys(action.payload)[0];
      return { ...state, items: {...state.items, [key]: action.payload[key] } };
```

Unless there is an easier way I am unaware of it seems for me that it's more readable to use `merge`.

prydonius: I'm not sure I understand, why can't we just do 
```
{ ...state, items: {...state.items, ...action.payload } };
```
?

dashboard/src/components/AppView/AppView.tsx
prydonius: AFAIU the whole idea of the container is to pass down relevant data to the component for it to render. I don't like the idea of doing filtering inside a component when it's not based on user input, if we know what a component needs then I think it should be filtered at either the reducer or the container level.

If what the component needs to show might change based on some user action inside the component, it might make sense to do filtering here, but even in that case it should go through the Redux flow and enter as an updated prop in the component.

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -100,19 +124,27 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
...
+        case "Service":
+          sockets.push(this.getSocket("services", i.apiVersion, i.metadata.name, newApp.namespace));
+          break;
+        case "Ingress":
+          sockets.push(
+            this.getSocket("ingresses", i.apiVersion, i.metadata.name, newApp.namespace),
+          );
+          break;
+        case "Secret":
+          this.props.getResource("v1", "secrets", newApp.namespace, i.metadata.name);
```
prydonius: Shouldn't `getApp` be in charge of getting the Secrets related to that app too? I think that's what we do when we get service bindings, for example?

dashboard/src/components/AppView/AppView.tsx
migmartri: Following the same reasoning than here https://github.com/kubeapps/kubeapps/pull/855#pullrequestreview-180586454 I'd have just done the minimum to make this works for now. I agree with @andresmgot with taking the short path here mainly because we do not have a proper pattern in this section of the page to retrieve resources via redux yet.

dashboard/src/actions/kube.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+export const requestResource = createAction("REQUEST_RESOURCE", resolve => {
+  return (resource: { [s: string]: IKubeItem }) => resolve(resource);
+});
+
+export const receiveResource = createAction("RECEIVE_RESOURCE", resolve => {
+  return (resource: { [s: string]: IKubeItem }) => resolve(resource);
+});
+
+export const errorKube = createAction("ERROR_KUBE", resolve => {
```
migmartri: Maybe we should rename this to something less cryptic and more specific? `ReceiveResourceError`?

dashboard/src/actions/kube.tsx
```diff
@@ -0,0 +1,39 @@
...
+  return (resource: { [s: string]: IKubeItem }) => resolve(resource);
+});
+
+const allActions = [requestResource, receiveResource, errorKube];
+
+export type KubeAction = ActionType<typeof allActions[number]>;
+
+export function getResource(
```
migmartri: Cool!

I like the generic way of fetching resources.

dashboard/src/actions/kube.tsx
```diff
@@ -0,0 +1,39 @@
...
+  apiVersion: string,
+  resource: string,
+  namespace: string,
+  name: string,
+  query?: string,
+): ThunkAction<Promise<void>, IStoreState, null, KubeAction> {
+  return async dispatch => {
+    const key = `${apiVersion}/${namespace}/${resource}/${name}${query ? `/${query}` : ""}`;
```
migmartri: Couldn't we generate the key from `getResourceURL`?

dashboard/src/reducers/kube.test.ts
```diff
@@ -0,0 +1,39 @@
...
+    errorKube: getType(actions.kube.errorKube),
+  };
+
+  beforeEach(() => {
+    initialState = {
+      items: {},
+    };
+  });
+
+  describe("reducer actions", () => {
```
migmartri: It's missing tests for error and requests no?

dashboard/src/components/AppView/AppView.tsx
prydonius: > mainly because we do not have a proper pattern in this section of the page to retrieve resources via redux yet.

Hmm, I disagree, we actually use redux to retrieve other resources, however this is the only place where we go through the manifest.

dashboard/src/reducers/kube.ts
migmartri: Why doesn't it apply here? And yes, we can discuss it in other issue since it affects every single part of the store.

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -86,10 +92,15 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
...
     // with { kind: foo }
     manifest = manifest.filter(r => r && r.kind);
+    if (!_.isEqual(manifest, this.state.manifest)) {
+      this.setState({ manifest });
+    } else {
+      return;
+    }
 
-    const watchedKinds = ["Deployment", "Service"];
+    const displayedKinds = ["Deployment", "Service", "Secret"];
```
prydonius: Shouldn't Ingress be here too?

dashboard/src/components/AppView/SecretsTable/SecretsTable.tsx
```diff
@@ -0,0 +1,74 @@
...
+import * as React from "react";
+
+import { ErrorSelector } from "../../../components/ErrorAlert";
+import { IKubeItem, ISecret } from "../../../shared/types";
+import SecretItem from "./SecretItem";
+
+interface IServiceTableProps {
+  namespace: string;
+  secretNames: string[];
+  secrets: { [s: string]: IKubeItem };
```
prydonius: perhaps this type could be `IKubeItem<ISecret>` so we can keep the type of the k8s resource?

prydonius: :+1: I like that this component only knows about secrets!

dashboard/src/shared/Kube.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  public static async getResource(
+    apiVersion: string,
+    resource: string,
+    namespace?: string,
+    name?: string,
+    query?: string,
+  ) {
+    const { data } = await axios.get<IResource | ISecret>(
```
prydonius: hmm, are we going to have to enumerate all the types here? Is there a better way we can keep the type of the resource?

prydonius:
This is looking good to me, I just think we need a way to keep the ISecret type, we could look into using type params for it.

dashboard/src/reducers/kube.ts
```diff
@@ -0,0 +1,34 @@
...
+    case getType(actions.kube.receiveResource):
+      return { ...state, items: _.merge({}, state.items, action.payload) };
```
migmartri: It is very strange that the reducer has all the steps generic (via action.payload) instead of defining them here. It also makes them more difficult to test, or need to test it on a higher level (dispatcher level not the store level)

I'd personally set the different changes here and having as only common payload content `the resource`.

So for example, the reducer will be in charge of doing this `{ isFetching: false, error: e }` after getting the resource and the error.

BTW, this is the reason why I was confused about the tests, because all the reducer cases are the same.

dashboard/src/reducers/kube.ts
andresmgot: No, I don't mean it doesn't apply, I mean that we would need to properly think how to flatten this and others if necessary.

andresmgot: Created https://github.com/kubeapps/kubeapps/issues/865 elaborate your idea in that issue if you need to.

dashboard/src/actions/kube.tsx
```diff
@@ -0,0 +1,39 @@
...
+  namespace: string,
+  name: string,
+  query?: string,
+): ThunkAction<Promise<void>, IStoreState, null, KubeAction> {
+  return async dispatch => {
+    const key = Kube.getResourceURL(apiVersion, resource, namespace, name, query);
+    dispatch(requestResource(key));
+    try {
+      const r = await Kube.getResource(apiVersion, resource, namespace, name, query);
+      dispatch(receiveResource({ [key]: r }));
```
migmartri: In order to simplify all the transformation done here and in the reducer, wouldn't make more sense if you send `{ Key: key, resource: r }`

That will simplify the reducer that now does some transformation based on the position of the args and so on. Or there is a reason you want to send them as a map?

dashboard/src/components/AppView/SecretsTable/SecretsTable.tsx
prydonius: Sorry, my comment didn't go into enough detail here, I meant that we should make the IKubeItem type support a generic type for its `item` key. To do so, requires a change in the type definition. See my comment below.

dashboard/src/shared/types.ts
```diff
@@ -323,3 +337,13 @@ export interface IAppOverview {
   status: string;
   chart: string;
 }
+
+export interface IKubeItem {
```
prydonius:
```suggestion
export interface IKubeItem<T> {
```

dashboard/src/shared/types.ts
```diff
@@ -323,3 +337,13 @@ export interface IAppOverview {
...
+  isFetching: boolean;
+  item?: IResource | ISecret;
```
prydonius:
```suggestion
  item?: T;
```

dashboard/src/shared/types.ts
prydonius: We can make this a generic type so that we can define specifically what we're looking for whenever we use it. That way we can be explicit about when we only expect there to be `ISecret` items, and in other places we can be more generic and just have `IResource`. The current way this is done is less desirable as we'd have to enumerate each type we want to potentially support. What do you think?

dashboard/src/actions/kube.tsx
prydonius: :+1: I agree, I think that would be more clear.

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -100,21 +111,32 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
...
-      sockets.push(this.getSocket("deployments", d.apiVersion, d.metadata.name, newApp.namespace));
-    }
-    for (const svc of services) {
-      sockets.push(this.getSocket("services", svc.apiVersion, svc.metadata.name, newApp.namespace));
-    }
-    for (const i of ingresses) {
-      sockets.push(this.getSocket("ingresses", i.apiVersion, i.metadata.name, newApp.namespace));
-    }
+    let secrets = {};
+    manifest.forEach((i: IResource | ISecret) => {
```
prydonius: I don't think the type `ISecret` is needed here since we convert it later, but perhaps I am missing something?

dashboard/src/actions/kube.tsx
andresmgot: 👍

dashboard/src/shared/types.ts
andresmgot: The problem with that is when using it in `IKubeState` we will find trouble when using `IKubeState.items` to something different than `IResource`. For the moment I can avoid the issues with that but I am not sure if that will be always the case.

dashboard/src/components/AppView/AppView.tsx
andresmgot: It's needed to avoid this issue:

```
Conversion of type 'IResource' to type 'ISecret' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Property 'data' is missing in type 'IResource'.
```

dashboard/src/components/AppView/SecretsTable/SecretsTable.tsx
```diff
@@ -0,0 +1,74 @@
...
+            <thead>
+              <tr className="flex">
+                <th className="col-2">NAME</th>
+                <th className="col-2">TYPE</th>
+                <th className="col-7">DATA</th>
+              </tr>
+            </thead>
+            <tbody>
+              {secretKeys.map(
+                k => secrets[k].item && <SecretItem key={k} secret={secrets[k].item as ISecret} />,
```
prydonius: This should no longer be needed, right?
```suggestion
                k => secrets[k].item && <SecretItem key={k} secret={secrets[k].item} />,
```

dashboard/src/shared/types.ts
prydonius: Hmm, what if IKubeState was also a generic and took a type param?

dashboard/src/components/AppView/SecretsTable/SecretsTable.tsx
andresmgot: It is needed because even if we are checking that `item` is set, TS complain about that it may be undefined:

```
Type 'ISecret | undefined' is not assignable to type 'ISecret'.
  Type 'undefined' is not assignable to type 'ISecret'.
```

dashboard/src/shared/types.ts
andresmgot: yes, but that that won't work if the caller wants to use two different types of resources. In any case, that's not needed right now, let's evaluate that again if we find an issue.

andresmgot:
Merging this, thanks for the reviews!

its_done

dashboard/src/shared/types.ts
prydonius: @andresmgot if the caller wants to use two different types of resources they can pass a union in, e.g. `IKubeState<IResource  | ISecret>` right? But yes, I agree we can reevaluate when we need it.

dashboard/src/components/AppView/SecretsTable/SecretsTable.tsx
prydonius: In that case, we could use the bang operator instead, so `secrets[k].item!`, anyway not a big deal. It's weird that TypeScript isn't able to figure out that it's defined based on the prior condition. Maybe because it's not in a variable?

PR #855, opened 2018-11-27
Improve AppView loading behavior

### LINKED ISSUE #748: Improve access URL loading feedback
It seems that the section at the top of the application that shows URLs for LoadBalancers and ingresses take some time to appear. We might want to add a loading icon in that section or some kind of message saying that we are retrieving that info. 

Also, if no public access URL is provided, we can add a message saying that this can be changed by setting services as LB or adding ingress rules.

selection_727

andresmgot:
Fixes #748
Requires #847

The goal of #748 is to provide feedback to the user when the URLs table is loading. In this PR I have added the LoadingWrapper to all the tables in the AppView page (not only to the URLs one):

Summary of the changes:

 - The URLs table now have a title: "Access URLs"
 - All the tables shows the loading icon if the content is not ready.
 - In case there are no elements for a table a message like "No resources found" is shown.
 - I have removed the component AppDetails to directly use `DeploymentTable`, `ServiceTable` and `OtherResourcesTable`.
 - I have removed the title "Details" since it was splitting the Secrets and the rest of resources.

Todo:

 - [x] Unit tests to check that each table behaves as a loading component and to check the 0 resources scenario.

gif_


andresmgot:
Sorry, I forgot to change the base branch. Now it's compared with the `secretButton` branch.

andresmgot:
Thanks for the comments @prydonius. I will stop working on this until #847 is completed to avoid pointing to a moving target.

dashboard/src/components/AppView/AccessURLTable/AccessURLTable.tsx
```diff
@@ -6,44 +6,48 @@ import { GetURLItemFromIngress } from "./AccessURLItem/AccessURLIngressHelper";
 import { GetURLItemFromService } from "./AccessURLItem/AccessURLServiceHelper";
 
 interface IServiceTableProps {
-  services: { [s: string]: IResource };
-  ingresses: { [i: string]: IResource };
+  services: IResource[];
```
migmartri: Could we instead of using IResource use something more specific in reference to the Service? I am not sure if we have the types for that though.

dashboard/src/components/AppView/AppView.test.tsx
```diff
@@ -259,25 +253,42 @@ describe("AppViewComponent", () => {
...
+    it("renders the loading icon if the URLs table is not ready", () => {
+      const wrapper = mount(<AppViewComponent {...validProps} />);
+      const ingress = {
+        isFetching: true,
+      };
+      const ingresses = { foo: ingress };
+      wrapper.setState({ ingresses });
+
+      expect(wrapper.find(AccessURLTable)).not.toExist();
+      const loading = wrapper.find(LoadingWrapper);
```
migmartri: Maybe you can reuse `itBehavesLike(aLoadingComponent)`

https://github.com/kubeapps/kubeapps/blob/33d6bf4a47c3e4a3e95ab9a59d55cdb6bf8371fb/dashboard/src/components/UpgradeForm/UpgradeForm.test.tsx#L14

dashboard/src/components/AppView/AccessURLTable/AccessURLTable.tsx
```diff
@@ -6,44 +6,48 @@ import { GetURLItemFromIngress } from "./AccessURLItem/AccessURLIngressHelper";
 import { GetURLItemFromService } from "./AccessURLItem/AccessURLServiceHelper";
 
 interface IServiceTableProps {
-  services: { [s: string]: IResource };
-  ingresses: { [i: string]: IResource };
```
migmartri: Why has the format changed? Is it consistent across the other tables?

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -126,22 +126,26 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
 
     const sockets: WebSocket[] = [];
     manifest.forEach(i => {
+      const id = `${i.kind}/${i.metadata.name}`;
       switch (i.kind) {
         case "Deployment":
+          this.setState({ deployments: { ...this.state.deployments, [id]: { isFetching: true } } });
```
migmartri: Why don't you initialize the state with isFetching: true and remove these lines? 

Also, does this line ever set deployments? It seems that the isFetching thing is what is relevant isn't it?

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -160,20 +164,26 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
...
-        this.setState({ ingresses: { ...this.state.ingresses, [key]: resource } });
+        this.setState({
+          ingresses: { ...this.state.ingresses, [key]: { resource, isFetching: false } },
+        });
         break;
     }
   }
 
-  public get isLoading(): boolean {
+  public get isAppLoading(): boolean {
```
migmartri: You might want to explain in a comment what does loading app really means.

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -217,23 +233,32 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
               <div className="col-9">
                 <div className="row padding-t-bigger">
                   <div className="col-4">
-                    <DeploymentStatus deployments={this.deploymentArray()} info={app.info!} />
+                    <DeploymentStatus
+                      deployments={this.arrayFromState("deployments")}
```
migmartri: you could use `deployments` defined above no? 

```
const deployments = this.arrayFromState("deployments");
```

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -266,56 +291,77 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
...
+    const res: IResource[] = [];
+    elems.forEach(e => {
+      if (this.state[type][e].resource) {
+        res.push(this.state[type][e].resource);
+      }
+    });
+    return res;
+  }
+
+  // Retrieve is some deployments/service/ingresses is still loading
```
migmartri: > ingresses is still loading

ingresses are still loading

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -37,13 +39,12 @@ export interface IAppViewProps {
 }
 
 interface IAppViewState {
-  deployments: { [d: string]: IResource };
+  manifest: IResource[];
+  deployments: { [d: string]: { resource?: IResource; isFetching: boolean } };
```
migmartri: shouldn't isFetching be outside of the struct? I mean, loading means loading all the resources of that type but not loading a single one isn't it?

Or we are planning to load resources individually?

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -266,56 +291,77 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
...
+      if (this.state[type][e].resource) {
+        res.push(this.state[type][e].resource);
+      }
+    });
+    return res;
+  }
+
+  // Retrieve is some deployments/service/ingresses is still loading
+  private isLoading(type: string) {
+    return _.some(this.state[type], e => e.isFetching);
```
migmartri: See my comment about having isFetching at the resources level above.

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -266,56 +291,77 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
...
-    _.each(kubeItems, (i, k) => {
+    _.each(resources, (i, k) => {
       if (i.error) {
         error = { resource: k, error: i.error };
       }
     });
     return error;
   };
 
+  private findLoadingResource = (resources: { [s: string]: IKubeItem }): boolean => {
```
migmartri: Could you explain why is this function needed?

dashboard/src/components/AppView/OtherResourcesTable.tsx
```diff
@@ -0,0 +1,36 @@
+import * as React from "react";
+
+import { IResource } from "../../shared/types";
+
+interface IAppDetailsProps {
+  otherResources: { [r: string]: IResource };
```
migmartri: Shouldn't this for consistency be `IResource[];` ?

dashboard/src/components/AppView/DeploymentTable.tsx
```diff
@@ -4,29 +4,33 @@ import { IResource } from "../../shared/types";
 import DeploymentItem from "./DeploymentItem";
 
 interface IDeploymentTableProps {
-  deployments: { [d: string]: IResource };
+  deployments: IResource[];
```
migmartri: Could this be Ideployment[]?

dashboard/src/components/AppView/AppView.tsx
migmartri: My bad, I think that we actually fetch resources individually, although personally I am not sure if it makes sense to have this flag per resource in our case.

andresmgot:
I will wait for #847 to be finished before keep working on this.

dashboard/src/components/AppView/AccessURLTable/AccessURLTable.tsx
andresmgot: we don't have a type `IService` but it's not needed since it follows the generic layout (metadata, spec and status keys).

dashboard/src/components/AppView/AppView.test.tsx
andresmgot: `itBehavesLike("aLoadingComponent", ` checks if there is a LoadingWrapper within the given component. We cannot use that here because:

 - If we pass `AppView` as the component it may find LoadingWrapper not related to the component we are interested in. For example it can detect a LoadingWrapper around the ServiceTable but we are interested in the AccessURLTable.
 - If we pass the component that we want (AccessURLTable for example) it won't work because the LoadingWrapper is actually outside the component.

dashboard/src/components/AppView/DeploymentTable.tsx
andresmgot: with deployments happen the same than services, since it follows the same layout it's not necessary to define a new interface.

dashboard/src/components/AppView/AppView.test.tsx
andresmgot: After some refactoring I have moved the different LoadingWrappers so I can use now the `itBehavesLike` method.

dashboard/src/components/AppView/AppView.tsx
andresmgot: > Why don't you initialize the state with isFetching: true and remove these lines?

I am not sure if I understand this comment. Note that that the `isFetching` property depends on the independent elements of the array, I cannot initialize `isFetching` property.

> Also, does this line ever set deployments? It seems that the isFetching thing is what is relevant isn't it?

True, I am only interested in the isFetching property

dashboard/src/components/AppView/AppView.tsx
andresmgot: yes, it's per element right now.

dashboard/src/components/AppView/AppView.tsx
andresmgot: it's convenient to find a loading item from a list. I've moved this to a helper function.

andresmgot:
@prydonius @migmartri this should be ready to be reviewed again. Since the last review I have:

 - Moved each table to a folder
 - Moved the loading check to the table component (not in the AppView)
 - Standardized how deployments/secrets/ingresses are represented (using `Array<IKubeItem<IResource>>`
 - Wait for the deployments to load before rendering the `DeploymentStatus` box

andresmgot:
@prydonius @migmartri this should be ready to be reviewed again. Since the last review I have:

 - Moved each table to a folder
 - Moved the loading check to the table component (not in the AppView)
 - Standardized how deployments/secrets/ingresses are represented (using `Array<IKubeItem<IResource>>`
 - Wait for the deployments to load before rendering the `DeploymentStatus` box

dashboard/src/components/AppView/AccessURLTable/AccessURLTable.tsx
```diff
@@ -1,49 +1,76 @@
...
+  services: Array<IKubeItem<IResource>>;
+  ingresses: Array<IKubeItem<IResource>>;
 }
 
 class AccessURLTable extends React.Component<IServiceTableProps> {
   public render() {
-    const { services, ingresses } = this.props;
+    const { ingresses, services } = this.props;
+    let ingressSection = <p>The current application does not expose a public URL.</p>;
```
prydonius: whilst both services and ingresses are ingresses, the term is confusing because at first I thought we were excluding services here. maybe we should call this accessTable or something else more generic?

dashboard/src/components/AppView/AccessURLTable/AccessURLTable.tsx
```diff
@@ -1,49 +1,76 @@
...
+import { IKubeItem, IResource, IServiceSpec } from "../../../shared/types";
+import isSomeResourceLoading from "../helpers";
 import AccessURLItem from "./AccessURLItem";
 import { GetURLItemFromIngress } from "./AccessURLItem/AccessURLIngressHelper";
 import { GetURLItemFromService } from "./AccessURLItem/AccessURLServiceHelper";
 
 interface IServiceTableProps {
-  services: { [s: string]: IResource };
-  ingresses: { [i: string]: IResource };
+  services: Array<IKubeItem<IResource>>;
```
prydonius: if we have IService and IIngress, could this be `Array<IKubeItem<IService | IIngress>>`? would help us in not having to do the type conversion below.

dashboard/src/components/AppView/AppView.test.tsx
```diff
@@ -241,22 +226,26 @@ describe("AppViewComponent", () => {
...
+                  paths: [{ path: "/ready" }],
+                },
               },
-            },
-          ],
-        } as IIngressSpec,
-      } as IResource;
+            ],
+          } as IIngressSpec,
+        } as IResource,
```
prydonius: this is quite messy, ideally we should just be able to use IIngress here and not have to convert it twice 
(once for the spec, and for the full thing)

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -113,29 +115,38 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
 
     const sockets: WebSocket[] = [];
     let secrets = {};
+    let deployments = {};
```
prydonius: what are these for? aren't we already setting them in the state in `handleEvent`?

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -147,23 +158,32 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
...
+        this.setState({
+          ingresses: { ...this.state.ingresses, ...item },
+        });
         break;
     }
   }
 
-  public get isLoading(): boolean {
+  // isAppLoading checks if the given app has been collected from tiller
+  public get isAppLoading(): boolean {
```
prydonius: why do we need a getter instead of a normal function? I'm just curious mostly, but I also think explicit functions are better because I now can't tell if this is a property or a function when I see it being used. Why couldn't appInfo be a getter?

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -259,8 +274,17 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
     }
   }
 
-  private deploymentArray(): IResource[] {
-    return Object.keys(this.state.deployments).map(k => this.state.deployments[k]);
+  // Retrieve the deployments/service/ingresses if they are already loaded
+  private arrayFromState(type: string): Array<IKubeItem<any>> {
```
prydonius: isn't it going to be an `IKubeItem<IDeployment | IService | IIngress>` or if we need to be more general `IKubeItem<IResource>`. We should try to avoid using `any` if possible.

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -259,8 +274,17 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
...
+    const elems = Object.keys(this.state[type]);
+    const res: Array<IKubeItem<any>> = [];
+    elems.forEach(e => {
```
prydonius: couldn't this use `Array.filter` instead of populating a new array?

dashboard/src/components/AppView/helpers/index.ts
```diff
@@ -0,0 +1,13 @@
+import * as _ from "lodash";
+
+import { IKubeItem } from "./../../../shared/types";
+
+export default function isSomeResourceLoading(resources: Array<IKubeItem<any>>): boolean {
+  let isFetching = false;
+  _.each(resources, i => {
```
prydonius: couldn't this be simplified with an `Array.any` or `_.any`?

dashboard/src/components/AppView/DeploymentTable.tsx
migmartri: Sorry, I think that I do not understand this comment.

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -29,11 +31,11 @@ export interface IAppViewProps {
...
 
 interface IAppViewState {
-  deployments: { [d: string]: IResource };
-  otherResources: { [r: string]: IResource };
-  services: { [s: string]: IResource };
-  ingresses: { [i: string]: IResource };
+  deployments: { [d: string]: IKubeItem<IResource> };
+  services: { [s: string]: IKubeItem<IResource> };
+  ingresses: { [i: string]: IKubeItem<IResource> };
   secrets: { [s: string]: ISecret };
```
migmartri: Why some are IkubeItem and others are Iresrouce/ISecret?

dashboard/src/components/AppView/AppView.tsx
migmartri: I am assuming that it is to set an initial state based on the value in the manifest?

If that's the case, I'd add a comment.

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -147,23 +158,32 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
   public handleEvent(e: MessageEvent) {
     const msg = JSON.parse(e.data);
     const resource: IResource = msg.object;
-    const key = `${resource.kind}/${resource.metadata.name}`;
+    const item = {
+      [resource.metadata.name]: { item: resource, isFetching: false },
```
migmartri: So the key is just the resource name and not the `resourceType/resourceName` ? In that case we will not be able to flatten resources in the future no?

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -29,11 +31,11 @@ export interface IAppViewProps {
 }
 
 interface IAppViewState {
-  deployments: { [d: string]: IResource };
-  otherResources: { [r: string]: IResource };
-  services: { [s: string]: IResource };
-  ingresses: { [i: string]: IResource };
+  deployments: { [d: string]: IKubeItem<IResource> };
```
migmartri: Why do we need a map instead of an array? It seems that later we do `this.arrayFromState("services");`

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -207,26 +227,21 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
...
                   </div>
                   <div className="col-8 text-r">
                     <AppControls app={app} deleteApp={this.deleteApp} />
                   </div>
                 </div>
-                {(Object.keys(this.state.services).length > 0 ||
-                  Object.keys(this.state.ingresses).length > 0) && (
-                  <AccessURLTable services={this.state.services} ingresses={this.state.ingresses} />
-                )}
+                <AccessURLTable services={services} ingresses={ingresses} />
```
migmartri: Much cleaner, thanks!

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -207,26 +227,21 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
...
                   secretNames={Object.keys(this.state.secrets)}
                 />
-                <AppDetails
-                  deployments={this.state.deployments}
-                  services={this.state.services}
-                  otherResources={this.state.otherResources}
-                />
+                <DeploymentsTable deployments={deployments} />
+                <ServicesTable services={services} />
+                <OtherResourcesTable otherResources={_.map(this.state.otherResources, r => r)} />
```
migmartri: Why can't we use arrayFromState here?

dashboard/src/components/AppView/DeploymentsTable/DeploymentsTable.tsx
```diff
@@ -0,0 +1,46 @@
...
+            <tr>
+              <th>NAME</th>
+              <th>DESIRED</th>
+              <th>UP-TO-DATE</th>
+              <th>AVAILABLE</th>
+            </tr>
+          </thead>
+          <tbody>
+            {deployments.map(
+              d => d.item && <DeploymentItem key={d.item.metadata.name} deployment={d.item} />,
```
migmartri: too late but I'd have called it `resource` instead of `item` :(

dashboard/src/components/AppView/DeploymentsTable/DeploymentsTable.tsx
```diff
@@ -0,0 +1,46 @@
...
+            )}
+          </tbody>
+        </table>
+      );
+    }
+    return (
+      <React.Fragment>
+        <h6>Deployments</h6>
+        <LoadingWrapper loaded={!isSomeResourceLoading(deployments)} size="small">
+          {deploymentSection}
```
migmartri: another potentially cleaner way of doing this is by making deploymentSection a function. In any case, this looks quite good!

dashboard/src/components/AppView/SecretsTable/SecretsTable.test.tsx
```diff
@@ -64,3 +91,23 @@ it("renders a table with several secrets", () => {
       .key(),
   ).toBe(secret2);
 });
+
+it("renders a secret table with a secret and an error", () => {
```
migmartri: Either implement ,delete, or mark this test case as pending (using `xit`)

dashboard/src/components/DeploymentStatus/DeploymentStatus.tsx
```diff
@@ -1,25 +1,33 @@
...
+      return this.renderLoading();
+    }
     if (this.props.info && this.props.info.deleted) {
       return this.renderDeletedStatus();
     }
     return this.isReady() ? this.renderSuccessStatus() : this.renderPendingStatus();
   }
 
+  private renderLoading() {
+    return <span className="DeploymentStatus">Loading...</span>;
```
migmartri: having a function for this only seems an overkill

dashboard/src/components/DeploymentStatus/DeploymentStatus.tsx
```diff
@@ -40,8 +48,11 @@ class DeploymentStatus extends React.Component<IDeploymentStatusProps> {
     const { deployments } = this.props;
     if (deployments.length > 0) {
       return deployments.every(d => {
-        const status: IDeploymentStatus = d.status;
-        return status.availableReplicas === status.replicas;
+        if (d.item) {
+          const status: IDeploymentStatus = d.item.status;
```
migmartri: This is the first time I see the every method. Interesting. You might want to explain that `every` returns false at the moment that the callback returns falsey.

dashboard/src/containers/AppViewContainer/AppViewContainer.tsx
```diff
@@ -30,6 +31,13 @@ function mapDispatchToProps(dispatch: ThunkDispatch<IStoreState, null, Action>)
     deleteApp: (releaseName: string, ns: string, purge: boolean) =>
       dispatch(actions.apps.deleteApp(releaseName, ns, purge)),
     getApp: (releaseName: string, ns: string) => dispatch(actions.apps.getApp(releaseName, ns)),
+    getResource: (
```
migmartri: I'd have preferred to have this method, or even a specific one (see below) in the SecretsTableContainer instead.

```
# With specific method I meant this one in the SecretsTableContainer
getSecret(namespace, name)
```

migmartri: BTW, for some reason I can not find where this method is being used.

dashboard/src/components/AppView/AppView.tsx
andresmgot: this was a getter before and I didn't change it. I don't have a good reason why this is not a function. Actually I don't think this function is useful anyway.

dashboard/src/components/AppView/AppView.tsx
andresmgot: _removed this method_

dashboard/src/components/AppView/DeploymentsTable/DeploymentsTable.tsx
andresmgot: I'd avoid to keep adding changes to this PR, if you think that's needed please open a different issue.

dashboard/src/components/AppView/SecretsTable/SecretsTable.test.tsx
andresmgot: oops, I missed this

dashboard/src/components/DeploymentStatus/DeploymentStatus.tsx
andresmgot: I just followed the style of this file. This specific method may be an overkill but not doing it may be inconsistent. Anyway, I am fine changing this.

dashboard/src/components/DeploymentStatus/DeploymentStatus.tsx
andresmgot: isn't it quite self-explanatory?

dashboard/src/containers/AppViewContainer/AppViewContainer.tsx
andresmgot: I forgot to remove this. You are right, it's used as `getSecret` in the SecretTableContainer

andresmgot:
I have applied your feedback @migmartri @prydonius. Thanks for the detailed comments!

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -100,43 +104,46 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
...
-      .reduce((acc, r) => {
-        // TODO: skip list resource for now
-        if (r.kind === "List") {
-          return acc;
-        }
-        acc[`${r.kind}/${r.metadata.name}`] = r;
-        return acc;
-      }, {});
+      // TODO: skip list resource for now
+      .filter(d => kindsWithTable.indexOf(d.kind) < 0 && d.kind !== "List");
```
migmartri: It might make sense to use the switch case to also populate the otherResources?

Since the switch case below now it is also used to populate the initial state, it might make sense to use it for otherResources?

dashboard/src/components/DeploymentStatus/DeploymentStatus.tsx
migmartri: Well, it was not for me, but that can be my fault :)

dashboard/src/components/AppView/AppView.tsx
andresmgot: good point

andresmgot:
I am merging this now. Let me know if there is still something you want me to change and I will open a different PR.

PR #867, opened 2018-12-04
Add optional securityContext. Change kubectl image

### LINKED ISSUE #846: Replace the kubectl image in the Chart
It would be better to change our current image:

https://github.com/kubeapps/kubeapps/blob/master/chart/kubeapps/values.yaml#L127

For the bitnami version of it that is kept to up to date. Also we can upgrade the kubectl version to avoid possible issues in the future. The current latest image is `bitnami/kubectl:1.12.2-r13`.

### LINKED ISSUE #863: Add a pod security policy
feature request:
like we can see on several chart, put the opportunity to add a pod security policy in the helm chart.

andresmgot:
Fixes #863 
Fixes #846 

Add a default security context for Kubeapps deployments/pods. This setting is generic since I don't think there is a use case to enable or change it per deployment.

The current `kubectl` image is not compatible with the `securityContext` setting so I migrated that to the bitnami one to complete #846 as well.

Tested manually.


chart/kubeapps/values.yaml
```diff
@@ -261,6 +261,14 @@ rbac:
   # Perform creation of RBAC resources
   create: true
 
+## Pod Security Context
+## ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
+##
+securityContext:
+  enabled: false
+  fsGroup: 1001
```
migmartri: I am not sure if in our case we need fsGroup since we do not have persistence

prydonius: Maybe we should comment these values out as examples?

andresmgot: Even if there is no persistence the `fsGroup` is used to associate the volumes we use (in Nginx and tiller-proxy) but it's true that the daemons are not writing in those volumes so we can remove that if we want.

> Maybe we should comment these values out as examples?
 
I would comment those if after setting `enable: true` the parameters would be optional. In this case `runAsUser` is "mandatory" but we can comment out the `fsGroup`.

PR #868, opened 2018-12-04
Standardize capitalization in tags

### LINKED ISSUE #758: Capitalization inconsistency between apps status and instances status
An app is `deployed`, but an instance is `Provisioned` (capital `P`)

We should decided if those labels are capitalized or not.

screen shot 2018-10-23 at 12 19 56 pm
screen shot 2018-10-23 at 12 19 48 pm

andresmgot:
Fixes #758 

Set the first letter of the service instance status to lowercase. Note that there is not a standard about what type of notation this field should use or which values can be used. The only documentation says that it should be "machine readable":

https://github.com/kubernetes-incubator/service-catalog/blob/master/pkg/apis/servicecatalog/v1beta1/types.go#L1004

Also, the status (or any other tag) may be quite big, breaking the card limits (see the comment). To avoid that I have limited the tag text size to 150px.


dashboard/src/components/InfoCard/InfoCard.test.tsx
```diff
@@ -64,3 +64,19 @@ it("should parse JSX elements in the tags", () => {
   expect(tag2Elem).toExist();
   expect(tag2Elem.text()).toBe("tag2");
 });
+
+it("should trim a long tag", () => {
```
andresmgot: If we don't ensure this the tag surpasses the card limit:

screenshot 2018-12-04 at 18 24 57

With the fix:

screenshot 2018-12-04 at 18 23 58

dashboard/src/components/ServiceInstanceList/ServiceInstanceCardList.tsx
```diff
@@ -47,7 +54,7 @@ const ServiceInstanceCardList: React.SFC<IServiceInstanceCardListProps> = props
                   icon={icon}
                   serviceClassName={(svcClass && svcClass.spec.externalName) || "-"}
                   servicePlanName={instance.spec.clusterServicePlanExternalName}
-                  statusReason={status && status.reason}
+                  statusReason={status && lowerCaseFirst(status.reason)}
```
prydonius: Why not just use https://lodash.com/docs/#lowerFirst?

dashboard/src/components/InfoCard/InfoCard.tsx
```diff
@@ -16,6 +16,14 @@ export interface IServiceInstanceCardProps {
   tag2Content?: string | JSX.Element;
 }
 
+function truncateTag(tag: string | JSX.Element) {
```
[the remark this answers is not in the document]
andresmgot: good point, I didn't know about that property

Issue #870, opened 2018-12-06
CI - /bin/bash: line 3: latestReleaseTag: command not found


migmartri:
Our CI jobs that run an e2e test for the latest release seems to fail consistently. 

I couldn't spot any piece of code that could have caused a regression, It seems like circle-ci runs every step now in a new bash instance so previously sourced (in other steps) are not accessible anymore.

https://circleci.com/gh/kubeapps/kubeapps/4813

PR #871, opened 2018-12-06
source script in the specific step

migmartri:
Try if sourcing the script in the step is now required and fixes the issue shown here https://github.com/kubeapps/kubeapps/issues/870

This might not be the final fix but it might unblock our release process when we land code in master.

Closes https://github.com/kubeapps/kubeapps/issues/870

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -192,6 +192,7 @@ gke_test: &gke_test
     - run: |
         # If we want to test the latest version instead we override the image to be used
         if [[ -n "$TEST_LATEST_RELEASE" ]]; then
+          source ./script/chart_sync_utils.sh
           # latestReleaseTag was sourced in the first step via chart_sync_utils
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -192,6 +192,7 @@ gke_test: &gke_test
     - run: |
         # If we want to test the latest version instead we override the image to be used
         if [[ -n "$TEST_LATEST_RELEASE" ]]; then
+          source ./script/chart_sync_utils.sh
           # latestReleaseTag was sourced in the first step via chart_sync_utils
```
andresmgot: you forgot to remove this comment

andresmgot:
thanks for fixing this!

.circleci/config.yml
migmartri: Oops good catch!

PR #873, opened 2018-12-06
Fix k8s_wait_for_job_completed

### LINKED ISSUE #869: Spurious CI error on chart-svc test
It seems that we are receiving some errors during the CI builds related to tests in the chart-svc. This test looks for Wordpress in the API, and something that I've noticed in the screenshot is that the repo-sync for the stable repo is still running.

selection_773

We should double check if we have a false positive in the log when we get

```
Job apprepositories.kubeapps.com/repo-name=stable ready
```

https://circleci.com/gh/kubeapps/kubeapps/4815

andresmgot:
Fix #869 

`k8s_wait_for_job_completed` was exiting prematurely. Check the comment below.

Tested on https://circleci.com/gh/kubeapps/kubeapps/4881

### DIFF
--- a/script/libtest.sh
+++ b/script/libtest.sh
@@ -58,12 +58,11 @@ k8s_wait_for_job_completed() {
     while [ "$retryTimeSeconds" -gt 0 ]; do
         # Avoid to exit the function if the job is not completed yet
         set +e
-        kubectl get jobs -n $namespace -l $labelSelector \
-           -o jsonpath='{.items[*].status.conditions[?(@.type=="Complete")].status}' | grep "True"
+        kubectl get jobs -n $namespace -l $labelSelector -o jsonpath='{.items[*].status.conditions[?(@.type=="Complete")].status}' | grep "True"
         res=$?
         set -e
         # There is a job that finished
-        if [[ $res ]]; then
+        if [[ "$res" -eq "0" ]]; then
             echo "Job '${@:2}' completed"
             return 0
         fi


script/libtest.sh
```diff
@@ -58,12 +58,11 @@ k8s_wait_for_job_completed() {
     while [ "$retryTimeSeconds" -gt 0 ]; do
         # Avoid to exit the function if the job is not completed yet
         set +e
-        kubectl get jobs -n $namespace -l $labelSelector \
-           -o jsonpath='{.items[*].status.conditions[?(@.type=="Complete")].status}' | grep "True"
+        kubectl get jobs -n $namespace -l $labelSelector -o jsonpath='{.items[*].status.conditions[?(@.type=="Complete")].status}' | grep "True"
```
andresmgot: Removed the line split since I found that adding a trailing space `\ ` make it fail. It's not the cause of the error here but it's safer to have it in just one line.

script/libtest.sh
```diff
@@ -58,12 +58,11 @@ k8s_wait_for_job_completed() {
...
         res=$?
         set -e
         # There is a job that finished
-        if [[ $res ]]; then
+        if [[ "$res" -eq "0" ]]; then
```
andresmgot: This was the error:

```
▶ a=1; if [[ $a ]]; then echo foo; fi
foo

▶ a=0; if [[ $a ]]; then echo foo; fi
foo
```

script/libtest.sh
migmartri: Is it? Why? I mean we use line split everywhere. What error were you seeing?

script/libtest.sh
migmartri: Thanks @andresmgot, that makes sense. 

I believe that the underlying issue is that we overrode the `e` option around `kubectl.... | grep true` so the script did not fail right? In any case we needed the `set +e` for some reason I do not remember.

script/libtest.sh
andresmgot: I mean that while I was copying this code I accidentally added a trailing space after the `\`. The code in the repo was fine but while I was debugging I found that somehow I introduced the trailing space. To avoid some issue like that in the future I removed the splitting of the command. It's not that important though, we can keep using `\` if the command is too long.

script/libtest.sh
andresmgot: > In any case we needed the set +e for some reason I do not remember.

The command `kubectl get jobs` may fail because the job has not been created yet. That's why we need to "catch" errors there so we can retry.

PR #876, opened 2018-12-11
Show chart descriptions

### LINKED ISSUE #793: Add chart description to catalog view
Currently, our catalog view is very similar to the apps view. We plan on adding the chart name in the apps view https://github.com/kubeapps/kubeapps/issues/764. But another way we could differentiate these two views is by adding the chart description to the catalog view.

This is similar to what you can find in GCP launcher 

selection_747

This could help to differentiate different shapes of Redis for example.

selection_749

andresmgot:
Fixes #793 

This PR adds the chart descriptions to the Catalog view. Since descriptions can be quite long, descriptions over 90 characters are trimmed. Users can still see the full description clicking on "Show more". This is the justification of why I chose to do it that way:

 - The most common thing to trim text is using the CSS property `text-overflow: ellipsis`. Unfortunately that only works for text with just one line. One line in this case is not enough since it will never show a meaningful message.
 - There are several options to trim texts of more than one line (three lines usually gives a good description) as explained in this [post](http://hackingui.com/front-end/a-pure-css-solution-for-multiline-text-truncation/). The most common way of doing that is with a CSS class that defines `display: -webkit-box; -webkit-line-clamp: 3; -webkit-box-orient: vertical;`. Unfortunately again, that only works when the text has a fixed width. In our case, the width is responsive so we need to discard that option.
 - After evaluating the different options, it seems that the best approach is to trim the text using Javascript code. The minimum width a card can get is around ~200px. We need to calculate the amount of characters to show having in mind that if the card shrinks, some text may be hidden. Taking that into account I chose to use 90 characters as maximum so the view behaves properly with any size.
 - ~~To allow users to get the full meaningful description, I added a "Show more" button that displays the whole description of a chart.~~

screenshot 2018-12-14 at 10 23 33

### DIFF
--- a/dashboard/src/components/Catalog/CatalogItem.test.tsx
+++ b/dashboard/src/components/Catalog/CatalogItem.test.tsx
@@ -1,4 +1,5 @@
 import { shallow } from "enzyme";
+import context from "jest-plugin-context";
 import * as React from "react";
 
 import { IChart, IRepo } from "../../shared/types";
@@ -34,7 +35,7 @@ it("should render an item", () => {
 });
 
 it("should use the default placeholder for the icon if it doesn't exist", () => {
-  const chartWithoutIcon = { ...defaultChart };
+  const chartWithoutIcon = Object.assign({}, defaultChart);
   chartWithoutIcon.attributes.icon = undefined;
   const wrapper = shallow(<CatalogItem chart={chartWithoutIcon} />);
   // Importing an image returns "undefined"
@@ -48,7 +49,7 @@ it("should use the default placeholder for the icon if it doesn't exist", () => 
 });
 
 it("should place a dash if the version is not avaliable", () => {
-  const chartWithoutVersion = { ...defaultChart };
+  const chartWithoutVersion = Object.assign({}, defaultChart);
   chartWithoutVersion.relationships.latestChartVersion.data.app_version = "";
   const wrapper = shallow(<CatalogItem chart={chartWithoutVersion} />);
   expect(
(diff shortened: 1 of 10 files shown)


dashboard/src/components/InfoCard/InfoCard.tsx
```diff
@@ -10,46 +10,50 @@ export interface IServiceInstanceCardProps {
...
-  const { title, link, info, tag1Content, tag1Class, tag2Content, tag2Class } = props;
+  const { title, link, info, description, tag1Content, tag1Class, tag2Content, tag2Class } = props;
   const icon = props.icon ? props.icon : placeholder;
   return (
     <Card responsive={true} className="ListItem">
       <Link to={link || "#"} title={title}>
         <CardIcon icon={icon} />
-        <CardContent>
-          <div className="ListItem__content">
+      </Link>
```
andresmgot: Now the link is just for the icon and the title, the description doesn't contain a link.

andresmgot:
> I would not add it because it because 1) It messes with the layout, 2) it adds clutter and, 3) can be confused with the other action of seen the whole catalog item and 4) other Marketplace solutions do not implement it and users seem to be fine with it

That's reasonable. I wanted to avoid having meaningless messages in the view but it's true that it adds clutter.

dashboard/src/components/Catalog/CatalogItem.test.tsx
```diff
@@ -34,7 +35,7 @@ it("should render an item", () => {
 });
 
 it("should use the default placeholder for the icon if it doesn't exist", () => {
-  const chartWithoutIcon = { ...defaultChart };
```
migmartri: oh, so that does not work then?

dashboard/src/components/InfoCard/InfoCard.tsx
migmartri: Wouldn't make sense to make the whole thing linkable? Maybe wrapping Card inside the Link.

dashboard/src/components/InfoCard/InfoCard.tsx
```diff
@@ -10,46 +10,50 @@ export interface IServiceInstanceCardProps {
...
+                  className={`ListItem__content__info_tag ListItem__content__info_tag-2 type-small type-color-white padding-t-tiny padding-h-normal margin-b-normal ${tag2Class ||
+                    ""}`}
+                >
+                  {tag2Content}
+                </span>
+              )}
             </div>
           </div>
-        </CardContent>
-      </Link>
```
migmartri: Actually like this, I think it makes sense that the whole content is linkable. Why did you change it?

dashboard/src/components/Catalog/CatalogItem.tsx
```diff
@@ -11,29 +11,44 @@ interface ICatalogItemProps {
...
-        info={latestAppVersion || "-"}
-        icon={iconSrc}
-        tag1Content={repoTag}
-        tag1Class={repo.name}
-      />
-    );
+// 3 lines description max
+const MAX_DESC_LENGTH = 90;
+
+function trimDescription(desc: string): string {
```
migmartri: I do not think we need this function. We can rely on CSS instead. https://css-tricks.com/snippets/css/truncate-string-with-ellipsis/

dashboard/src/components/Catalog/CatalogItem.test.tsx
andresmgot: nope, and I just realized that `Object.assign` doesn't work either 😞. The problem is that it mutates inner objects:

```
> a = { a: { b: 1 }}
{ a: { b: 1 } }
> b = {...a}
{ a: { b: 1 } }
> b.a.b = 2
2
> a
{ a: { b: 2 } }
```

The only thing that I found that behaves as expected is lodash `cloneDeep`:

```
> b = u.cloneDeep(a)
{ a: { b: 1 } }
> b.a.b = 2
2
> a
{ a: { b: 1 } }
```

dashboard/src/components/Catalog/CatalogItem.tsx
andresmgot: Nope, as explained in the description, the ellipsis only work if it's just one line. I wanted to show at least three lines.

dashboard/src/components/InfoCard/InfoCard.tsx
andresmgot: that's how it was before, but I found weird that some piece of text (the description) is also a link:

screenshot 2018-12-17 at 10 10 04

We can change the CSS to show it as normal text but even in that way I find it weird.

dashboard/src/components/InfoCard/InfoCard.tsx
andresmgot: _replied before_

dashboard/src/components/InfoCard/InfoCard.tsx
migmartri: To me, ideally, we should make the whole card the link, similarly to https://console.cloud.google.com/marketplace but I am aware that we can not do it because we have a link to the repository. 

Taking into account that the description is most of the real state of the content of the card, I would make it part of the link too. If the issue is how it looks I would just change it via CSS as you suggested.

migmartri: > We can change the CSS to show it as normal text but even in that way I find it weird.

What do you find weird? The link is actually the wrapper div (card content) no?

andresmgot: > What do you find weird? The link is actually the wrapper div (card content) no?

Yes but it's weird for me that I need to override the default link CSS style in order to make it look like normal text while it's actually a link.

Issue #485, opened 2018-08-14
Auto generate release name


migmartri:
Currently, during chart deployment, we enforce the user to provide a release name. Release name which is optional in the Helm CLI.

![selection_644](https://user-images.githubusercontent.com/24523/44121166-d4352aaa-9fd3-11e8-9489-fb7a8440ad3e.png)

We might want to look into a way to auto-generate a release name if not provide.

Auto-generating a release name has the benefit of helping to avoid name collisions with existing charts now that we share Tiller, more so if we end up enabling deletion (not purge) by default. https://github.com/kubeapps/kubeapps/issues/481

From top of my head I can see two ways to implement this:

1 - Rely on Tiller release name generation

This will require us to make the field optional and let Tiller handle the name generation. We would need to do some tweaks in the UX to make it friendly enough (i.e field placeholder)

2 - Have our own generator

This has the benefit of being backend agnostic, if we support another backend that it's not tiller, we can still rely on this mechanism.

It also seems to have some benefits from the UX perspective since the user can tweak that value and see it in advance.

In term of effort, option 1 seems easier than option 2.

prydonius:
We actually have the capability currently to generate a name, we use the `moniker-native` module in the Function deployment view to do so: https://github.com/kubeapps/kubeapps/blob/a2f5dd4d1bbefc76e4d113babd8d4a2c6609c341/dashboard/src/components/FunctionList/FunctionDeployButton.tsx#L55

So we should be able to easily implement option 2 using the same module, also I like that this method doesn't rely on Tiller directly (e.g. if we change the way tiller proxy works, or support a different deployment tool in the future). Also, +1 for the UX benefit of being able to see the name and tweak before deploying.

andresmgot:
Agree, my +1 to option 2.

PR #877, opened 2018-12-11
Auto generate release name

andresmgot:
Fixes: #485

Auto generate a random name for the release using moniker.

screenshot 2018-12-11 at 18 31 20

### DIFF
--- a/dashboard/src/components/DeploymentForm/DeploymentForm.test.tsx
+++ b/dashboard/src/components/DeploymentForm/DeploymentForm.test.tsx
@@ -95,5 +95,15 @@ it("renders the full DeploymentForm", () => {
   const wrapper = shallow(
     <DeploymentForm {...defaultProps} selected={{ versions, version: versions[0] }} />,
   );
+  // Force static name
+  wrapper.setState({ releaseName: "bar" });
   expect(wrapper).toMatchSnapshot();
+});
+
+it("renders a release name by default", () => {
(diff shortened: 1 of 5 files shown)


dashboard/src/components/DeploymentForm/DeploymentForm.test.tsx
```diff
@@ -95,5 +95,15 @@ it("renders the full DeploymentForm", () => {
   const wrapper = shallow(
     <DeploymentForm {...defaultProps} selected={{ versions, version: versions[0] }} />,
   );
+  // Force static name
+  wrapper.setState({ releaseName: "bar" });
   expect(wrapper).toMatchSnapshot();
 });
+
+it("renders a release name by default", () => {
```
migmartri: You could potentially mock Moniker to test that its generation is the value set and changes between mounts.

andresmgot: I'd avoid all the `sinon`/`stub` integration just for that.

migmartri: Why is that? You do not need to use sinon but `jest`.

I think that the key here is if you care about testing that the value set as name changes and depends on an autogenerated value. Which is not being tested in your case.

andresmgot: ah, well, in that case it's not necessary to mock the behavior just to check that the release name has a length bigger than 0 and that it changes.

Issue #515, opened 2018-08-21
Custom CA support for Tiller and AppRepository controller


obeyler:
Hy, 
I deploy a chartmuseum helm repository with https and self-signed certificate.
Does it be possible to add ca on kubeapps ?
best regards

prydonius:
Kubeapps doesn't currently support it, but it's a feature we could certainly add. How do you use your chartmuseum repo with the Helm CLI today? Are you using the `--ca-file` flag with `helm install/fetch`?

One potential way we could support this is, similar to asking for the Authorization header when adding the repo to Kubeapps, we could have a textbox for optionally entering the CA bundle. This could then be stored in a Secret and loaded by the Tiller Proxy when fetching the chart. Does that sound like it makes sense?

obeyler:
I deploy chartmuseum without --ca-file. I expose it with an ingress rule and I've got a HaProxy in front of all my kubernetes worker then a Go Router from Cloud foundry to expose it on https
the CA certificate of https is a custom CA

prydonius:
@obeyler and what about on the client side when using the `helm` CLI? Is the custom CA already in your local certs folder, or do you use the `--ca-file` flag?

obeyler:
I don't use --ca-file flag on helm client.
When my CI use it deploy, it is not behind proxy so doesn't need the CA

prydonius:
@obeyler makes sense, thanks! If ChartMuseum is running inside the same cluster as Kubeapps, you can also use the service DNS name (e.g. chartmuseum.default.svc.cluster.local:PORT) in Kubeapps to connect to it, bypassing TLS for now.

obeyler:
The DNS name works to list the chart present into the repo but for user  who want to use the command line it offers a bad information as the helm add repo with svc.cluster.local will failed
(sorry to not anwser before I haven't seen your question before)

prydonius:
@obeyler are you thinking about Monocular here? Kubeapps doesn't show you a `helm repo add` command.

spiegela:
@prydonius I also am running into issues with this. I have chartmuseum installed as part of [Harbor](https://github.com/goharbor/harbor) (outside of Kubernetes) with a cert from a private CA.  Helm and Tiller appear to pick up the system trust chain, but the `apprepo-sync-...` jobs are failing with certificate errors.

Could we add a custom CA for the `apprepo-sync` containers?

prydonius:
[...]

PR #878, opened 2018-12-13
Allow tiller-proxy and AppRepository controller to use a custom CA

andresmgot:
Ref: #515

Allow users to sign a TLS certificate of a Helm Registry with a custom CA.

Tested manually using ChartMuseum and a self signed certificate.

Instructions of how to use the new feature included.

This PR requires https://github.com/helm/monocular/pull/578 to be released.


chart/kubeapps/templates/_helpers.tpl
```diff
@@ -131,3 +131,10 @@ Create name for the tiller-proxy based on the fullname
 {{- define "kubeapps.tiller-proxy.fullname" -}}
 {{ template "kubeapps.fullname" . }}-internal-tiller-proxy
 {{- end -}}
+
+{{/*
+Create name for the tiller-proxy based on the fullname
+*/}}
+{{- define "kubeapps.custom-helm-registry.secret" -}}
```
migmartri: I thought after reading this https://github.com/kubeapps/kubeapps/issues/515#issuecomment-415203354 that we were going to offer an option per-repository instead of a global option.

cmd/tiller-proxy/main.go
```diff
@@ -99,6 +100,37 @@ func main() {
...
+	systemCACert, err := ioutil.ReadFile("/etc/ssl/certs/ca-certificates.crt")
+	if err != nil {
+		log.Fatalf("Unable to read system CA certs: %v", err)
+	}
+	caCertPool.AppendCertsFromPEM(systemCACert)
+
+	if len(registryCaCerts) > 0 {
+		namespace := os.Getenv("POD_NAMESPACE")
+		for _, registryCaCert := range registryCaCerts {
+			caCertSecret, err := kubeClient.CoreV1().Secrets(namespace).Get(registryCaCert, metav1.GetOptions{})
```
migmartri: I am not convinced that referencing from here k8s is a good idea. Why isn't it loading it from disk or env?

chart/kubeapps/templates/_helpers.tpl
andresmgot: This is more simple for the moment. Note that having a single secret doesn't mean that we can only add one CA. We can add more files to that secret if we want to implement the UX that @prydonius mention in that comment. I don't know if that's worth it right now though. I think that users right now only have one personal registry.

prydonius:
> I thought after reading this #515 (comment) that we were going to offer an option per-repository instead of a global option.

Indeed, the UX I was thinking for #515 was to support providing a custom CA per-repository when adding an AppRepository. I don't think we're very far from that here, since the AppRepository controller just needs to mount the Secret with a name that relates to the particular repository and tiller-proxy can fetch the Secret the same way it does for the auth secret.

I think it might be more confusing to ship this as an option in the Helm chart and then later remove it from the chart and make it an input. Whilst this might be a simpler implementation, the UX is pretty bad. When installing Kubeapps, adding a repository with a custom CA could easily be an afterthought, and figuring out that I need to upgrade my Kubeapps release to add the custom CA to then add a repository though the UI is annoying.

> We can add more files to that secret if we want to implement the UX that @prydonius mention in that comment.

Hmm, what do we gain by managing it as a single Secret though? That sounds more difficult to manage, especially if we wanted to delete an AppRepository with a custom CA.

I think given we're close to it and if you agree, we can adapt this PR to implement the custom CA per-repository, but if you feel strongly about going for this approach first we can discuss it further.

andresmgot:
> I think it might be more confusing to ship this as an option in the Helm chart and then later remove it from the chart and make it an input

I thought of having both the same way you can have `initialRepos` and then add them from the UI.

> Indeed, the UX I was thinking for #515 was to support providing a custom CA per-repository when adding an AppRepository. I don't think we're very far from that here, since the AppRepository controller just needs to mount the Secret with a name that relates to the particular repository and tiller-proxy can fetch the Secret the same way it does for the auth secret.

If I understand what you mean: you are proposing to have the CA certificate secret as an optional field in the AppRepository CRD, isn't it? I am okay implementing that, it's just that is more work than what's present here. I don't think we are that close but I will work towards that.

prydonius:
@andresmgot not quite, we create Secrets when a user enters an Authorization header for the AppRepository from the UI, I was thinking we could reuse the same Secrets or create another one. I haven't really considered this as part of the initialRepos, because we currently don't support Authorization headers for those. I think it would be fine as a first step to implement it just in the tiller-proxy backend, and then the UI, and then maybe if it makes sense figure out a way to expose the CA and Auth header settings in the chart (though I would only do that last one if users ask for it). Perhaps we should discuss this during the standup since there seems to be some misunderstanding over this PR.

prydonius:
Ahh I see what you did was to put the secret name in the spec, that works actually, but maybe we should use the secretKeyRef in the same way we do for spec.auth.header under spec.caFile (I think calling it spec.customCA might be more descriptive)

andresmgot:
> Ahh I see what you did was to put the secret name in the spec, that works actually, but maybe we should use the secretKeyRef in the same way we do for spec.auth.header under spec.caFile (I think calling it spec.customCA might be more descriptive)

Indeed, it's a different secret in the spec. I don't fully understand the goal of having the `secretKeyRef`, is that to allow reusing the same secret and selecting a different key? From other comments (see https://github.com/helm/monocular/pull/578#discussion_r242250404) I understand that we want to force the same key `ca.crt` for the file. Would that be needed? The code is slightly more simple now.

Regarding the chart support, I think it's important to have an unattended mode to install these repositories (I believe that @obeyler only uses the chart to install repositories). It's already implemented and it's not a big deal so I think it's okay to have it.

andresmgot:
From an in-person discussion we agreed to:

 - Use a single secret to store the Auth info and the CA file to simplify the implementation in the Dashboard.
 - Do not add the possibility of adding CAs to repos using the chart (yet) since we don't have a good way of cleaning up the generated secrets and it's currently not supported to add auth information.

andresmgot:
This should be ready to be reviewed @prydonius @migmartri

cmd/apprepository-controller/controller_test.go
```diff
@@ -344,14 +350,106 @@ func Test_newSyncJob(t *testing.T) {
...
+					Namespace: "kubeapps",
+					Labels: map[string]string{
+						"name":       "my-charts",
+						"created-by": "kubeapps",
+					},
+				},
+				Spec: apprepov1alpha1.AppRepositorySpec{
+					Type: "helm",
+					URL:  "https://charts.acme.com/my-charts",
+					Auth: apprepov1alpha1.AppRepositoryAuth{
```
prydonius: should we add a test for the customCA and auth header? just for completeness sake?

cmd/tiller-proxy/internal/handler/handler.go
```diff
@@ -126,7 +126,11 @@ func getChart(req *http.Request, cu chartUtils.Resolver) (*chartUtils.Details, *
 	if err != nil {
 		return nil, nil, err
 	}
-	ch, err := cu.GetChart(chartDetails)
+	netClient, err := cu.InitNetClient(chartDetails.Auth.CustomCA)
```
migmartri: Does this mean that for every single request made to get the chart we are going to initialize the client and hence read a secret for Kubernetes? If that's the case it seems an unacceptable overload

migmartri: We might want to have the ca's preloaded or add an additional cache layer so it is fetched only once?

pkg/chart/chart.go
```diff
@@ -217,8 +228,41 @@ func (c *Chart) ParseDetails(data []byte) (*Details, error) {
...
+			TLSClientConfig: &tls.Config{
+				RootCAs: caCertPool,
+			},
+		},
+	}, nil
+}
+
 // GetChart retrieves and loads a Chart from a registry
-func (c *Chart) GetChart(details *Details) (*chart.Chart, error) {
+func (c *Chart) GetChart(details *Details, netClient HTTPClient) (*chart.Chart, error) {
```
migmartri: Actually GetChart will read the secret twice in the case of both header and CA are set.

Would it be possible to preload this or cache it?

cmd/tiller-proxy/internal/handler/handler.go
migmartri: Actually just noticed that we are already reading the secret in GetChart which means that the secret wiill be fetched twice in the case of both header and CA are set no?

Would it be possible to preload this or cache it?

prydonius: We shouldn't assume in tiller-proxy that the auth header and custom CA Secrets are the same, given they both each have their own secretKeyRefs and if we change the way we handle the Secrets in the future we don't need to worry about this.

I think a cache layer (that caches by key to avoid above issue) that fetches if it isn't loaded might be a good idea - but what happens if the Secret gets updated? When there's  a cache hit, we could return the cached value and then spin off a goroutine to get the updated Secret, but that would still cause it to fail the first time it tries to install a chart with an updated auth header/custom CA.

migmartri: > We shouldn't assume in tiller-proxy that the auth header and custom CA Secrets are the same, given they both each have their own secretKeyRefs and if we change the way we handle the Secrets in the future we don't need to worry about this.

That's fine as long as there is some caching involve.

> I think a cache layer (that caches by key to avoid above issue) that fetches if it isn't loaded might be a good idea - but what happens if the Secret gets updated?

Those are difficult problems indeed. What you mentioned about checking the secret maybe via the resourceID in a go-rountine could be enough in most cases. Alternatively, our tiller-proxy `details` could contain the sha or resourceID of the provided CustomCA, so we will compare it with the one cached. This will allow us to pre-load the data on boot but also re-fetch it if stale,

What I am worried here is about adding a synchronous operation to an already costly op (deploy or upgrade). Do we know how long it takes to pull the secret?

I am going to remove my "required changes" for this PR because it seems that we are just building on top of an already existing pattern and we are affected only when auth is set, but I'd love to see a follow up improvement here.

migmartri: https://github.com/kubeapps/kubeapps/issues/895

prydonius: > Alternatively, our tiller-proxy details could contain the sha or resourceID of the provided CustomCA

Interesting idea, that would require the frontend to pull down the Secret to compute the checksum though, right? In that case, wouldn't it be okay to simply the pass the header and custom CA directly in the details? Assuming we talk to tiller-proxy over TLS, it's no less secure than us already passing the Kubernetes Bearer token in every request.

migmartri: I am not completely familiar with how things work now but yes, somehow the sha would need to be generated and potentially stored in the appRepository CRD as annotation or field.

We could make the appRepository controller be in charge of populating that value too?

PR #880, opened 2018-12-13
Add Kubeapps version in the footer

### LINKED ISSUE #608: Feature request : Add version number of Kubeapps footer of  page
Give the opportunity to  the end user to say what version of Kubeapps they are using when they think they have a problem.
image

andresmgot:
Fixes #608 

Add the version of Kubeapps to the footer. This version is added in the Dashboard config map.

screenshot 2018-12-13 at 16 46 09

### DIFF
--- a/chart/kubeapps/templates/dashboard-config.yaml
+++ b/chart/kubeapps/templates/dashboard-config.yaml
@@ -23,4 +23,5 @@ data:
   config.json: |-
     {
       "namespace": "{{ .Release.Namespace }}"
+      "kubeappsVersion": "{{ .Chart.AppVersion }}"
     }
--- a/chart/kubeapps/templates/dashboard-config.yaml
+++ b/chart/kubeapps/templates/dashboard-config.yaml
@@ -23,5 +23,5 @@ data:
   config.json: |-
     {
       "namespace": "{{ .Release.Namespace }}"
-      "kubeappsVersion": "{{ .Chart.AppVersion }}"
+      "appVersion": "{{ .Chart.AppVersion }}"
     }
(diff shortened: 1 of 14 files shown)


chart/kubeapps/templates/dashboard-config.yaml
```diff
@@ -23,4 +23,5 @@ data:
   config.json: |-
     {
       "namespace": "{{ .Release.Namespace }}"
+      "kubeappsVersion": "{{ .Chart.AppVersion }}"
```
migmartri: maybe calling this `appVersion` would be enough.

dashboard/src/containers/LayoutContainer/LayoutContainer.ts
```diff
@@ -0,0 +1,12 @@
+import { connect } from "react-redux";
+
+import Layout from "../../components/Layout";
+import { IStoreState } from "../../shared/types";
+
+function mapStateToProps({ config }: IStoreState) {
+  return {
+    kubeappsVersion: config.kubeappsVersion,
```
migmartri: If this is being used only in the Footer, why didn't we create a FooterContainer instead?

migmartri: That way you do not need to pass down the props through the layout component.

chart/kubeapps/templates/dashboard-config.yaml
migmartri: You might also need to update the development version of this file.
```
src/public/config.json
```

andresmgot: what's that file for?

migmartri: If I remember correctly, this was a file that is required if you swap the dashboard with telepresence or something like that. @prydonius might have more info about it.

dashboard/src/containers/LayoutContainer/LayoutContainer.ts
migmartri: I do not think I fully understand the reasoning. Generally speaking, do you mean that you want to avoid importing a `container` from a `component`? Why is that?

PR #889, opened 2018-12-18
Refactor automatic release name test

andresmgot:
Ref: 
https://github.com/kubeapps/kubeapps/pull/877#discussion_r241444613

Ensure that the release name changes between loads. cc/ @migmartri

### DIFF
--- a/dashboard/src/components/DeploymentForm/DeploymentForm.test.tsx
+++ b/dashboard/src/components/DeploymentForm/DeploymentForm.test.tsx
@@ -102,8 +102,19 @@ it("renders the full DeploymentForm", () => {
 
 it("renders a release name by default", () => {
   const versions = [{ id: "foo", attributes: { version: "1.2.3" } }] as IChartVersion[];
-  const wrapper = shallow(
+  let wrapper = shallow(
     <DeploymentForm {...defaultProps} selected={{ versions, version: versions[0] }} />,
   );
-  expect(wrapper.state("releaseName")).toBeTruthy();
+  const name1 = wrapper.state("releaseName") as string;
+  expect(name1).toBeTruthy();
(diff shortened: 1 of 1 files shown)


dashboard/src/components/DeploymentForm/DeploymentForm.test.tsx
```diff
@@ -102,8 +102,19 @@ it("renders the full DeploymentForm", () => {
...
+
+  // When reloading the name should change
+  wrapper = shallow(
     <DeploymentForm {...defaultProps} selected={{ versions, version: versions[0] }} />,
   );
-  expect(wrapper.state("releaseName")).toBeTruthy();
+  const name2 = wrapper.state("releaseName") as string;
+  expect(name2).toBeTruthy();
+  expect(name2.length).toBeGreaterThan(1);
+  expect(name2).not.toEqual(name1);
```
prydonius: is there a chance it could generate the same name again? just wondering if this test could randomly flake one day and confuse us!

andresmgot: There is nothing preventing that to happen but I guess is highly unlikely. If we find issues with that maybe moniker is not the best library for that purpose :P

migmartri: Even though in practice the chances of collision are minimal. In theory I do not like this test case because it is not reproducible and it depends on an external factor (whatever result Moniker returns that day).

prydonius: I agree, if there was a way to mock moniker to change the dictionary between the test, that would be more reproducible.

andresmgot: okay

Issue #402, opened 2018-07-11
Support Kubernetes API Service proxy endpoint like Kubernetes Dashboard


prydonius:
As we move away from the `kubeapps` CLI, we should rethink the recommended way to access the Kubeapps Dashboard. Today, the `kubeapps dashboard` does a port-forward to the `kubeapps` pod. Without the CLI the same can be accomplished using the `kubectl port-forward ...` command. However this is a bit more complex to use as it requires figuring out the pod name.

The Kubernetes Dashboard [recommends users use `kubectl proxy`](https://github.com/kubernetes/dashboard#getting-started) to setup a secure proxy to the Kubernetes API and points users to access the known URL, `http://localhost:8001/api/v1/namespaces/kube-system/services/https:kubernetes-dashboard:/proxy/`.

We could support the same method in Kubeapps and point users to `http://localhost:8001/api/v1/namespaces/kubeapps/services/http:kubeapps:/proxy/`. This currently doesn't work correctly as the Dashboard and potentially the NGINX proxy expects to be under the root and not under a subpath[1]. We would need to make some changes to ensure Kubeapps works correctly under a subpath.

Supporting this method has the following advantages:
- familiar to Kubernetes Dashboard users
- proxies to the Service rather than the Pod, which is the correct way of accessing services within Kubernetes
- Easier than a complex `kubectl port-forward` command
- Potentially more reliable the `kubectl port-forward` (some people have reported issues of flaky connections with port forward, but it's not clear if this also affects proxy)

cc @migmartri

[1] ![screen shot 2018-07-11 at 14 35 42](https://user-images.githubusercontent.com/1544861/42600580-b82e2310-8517-11e8-91d1-239a12b9414b.png)

prydonius:
We'll need to enable hash-based routing in order to make this work

prydonius:
Note that enabling hash-based routing will also allow Kubeapps to be mounted under a subpath (e.g. internal.domain.com/kubeapps), which has been requested by some folks.

malejpavouk:
It would be really cool, as we have multiple apps running on our server and we want the users to use standard ports (80 or 443). 
And we can't simply deploy kubeapps on rootpath "/", as it would cause path collisions. Deploying it under some context (such as "/kubeapps") would really be helpful.

prydonius:
thanks for the feedback @malejpavouk

prydonius:
[...]

PR #890, opened 2018-12-18
Add custom path support

migmartri:
Add supports to deploy kubeapps under a custom prefix. 

To achieve it without requiring the user to do a custom build via webpack I needed to make the following modifications:

* In the package.json, set the `homepage` value relative so the bundled index.html contains a relative path to the assets.

```html
# Webpack generated index.html
# Before
<script type="text/javascript" src="/static/js/main.c4bfc712.js">
# after
<script type="text/javascript" src="./static/js/main.c4bfc712.js">
```

* Change React router history from push-state to hash-based. This simplified the setup since relative API calls defined in our code will ignore this hash state. This is not the case if we decided to modify the  push-state `basename` https://github.com/kubeapps/kubeapps/issues/402#issuecomment-448042009. Actually changing the basename would have been a runtime configuration that depends on the chart which seems brittle to me.  

* Updated Nginx rewrites and locations in both frontend and dashboard to handle the existence of an arbitrary namespace.

I could have done a rewrite in the Nginx target `nginx.ingress.kubernetes.io/rewrite-target` but I think that we are not requiring people to use only Nginx Ingress controller.

* Updated our dashboard code to use relative paths. 

peek 2018-12-18 17-26

### Call for testers

I have backed the dashboard in this docker image `migmartri/kubeapps-dashboard:devel16` and I'd appreciate if you could give it a try both deploying it with custom options but also tweaking the ingress path.

```
helm install --set dashboard.image.repository=migmartri/kubeapps-dashboard --set dashboard.image.tag=devel16 .
```

NOTE: 

In future PRs I will add the following annotation in the Ingress which will ensure (for users using Nginx ingress controller) that there is always a trailing slash

```
nginx.ingress.kubernetes.io/configuration-snippet:  rewrite ^(/[PREFIX])$ $1/ permanent;
```

Fixes https://github.com/kubeapps/kubeapps/issues/402

### DIFF
--- a/chart/kubeapps/templates/dashboard-config.yaml
+++ b/chart/kubeapps/templates/dashboard-config.yaml
@@ -17,6 +17,15 @@ data:
       gzip_static  on;
 
       location / {
+        # Support for ingress prefixes maintaining compatibility with the default /
+        # 1 - Exactly two fragment URLs for files existing inside of the public/ dir
+        # i.e /[prefix]/config.json => /config.json
+        rewrite ^/[^/]+/([^/]+)$ /$1 break;
+
+        # 2 - Any static files bundled by webpack
+        # i.e /[prefix]/static/main.js => static.main.js
+        rewrite ^/[^/]+/static/(.*) /static/$1 break;
+
         try_files $uri /index.html;
       }
     }
--- a/chart/kubeapps/templates/dashboard-config.yaml
+++ b/chart/kubeapps/templates/dashboard-config.yaml
@@ -22,8 +22,8 @@ data:
         # i.e /[prefix]/config.json => /config.json
         rewrite ^/[^/]+/([^/]+)$ /$1 break;
 
-        # 2 - Any static files bundled by webpack
-        # i.e /[prefix]/static/main.js => static.main.js
+        # 2 - Any static files bundled by webpack referenced by 3 or more URL segments
+        # i.e /[prefix]/static/main.js => static/main.js
         rewrite ^/[^/]+/static/(.*) /static/$1 break;
 
         try_files $uri /index.html;
--- a/dashboard/yarn.lock
+++ b/dashboard/yarn.lock
@@ -382,14 +382,6 @@ accepts@~1.3.4:
     mime-types "~2.1.16"
     negotiator "0.6.1"
 
-accepts@~1.3.5:
-  version "1.3.5"
-  resolved "https://registry.yarnpkg.com/accepts/-/accepts-1.3.5.tgz#eb777df6011723a3b14e8a72c0805c8e86746bd2"
-  integrity sha1-63d99gEXI6OxTopywIBcjoZ0a9I=
-  dependencies:
-    mime-types "~2.1.18"
-    negotiator "0.6.1"
-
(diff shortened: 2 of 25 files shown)


chart/kubeapps/templates/dashboard-config.yaml
```diff
@@ -17,6 +17,15 @@ data:
       gzip_static  on;
 
       location / {
+        # Support for ingress prefixes maintaining compatibility with the default /
+        # 1 - Exactly two fragment URLs for files existing inside of the public/ dir
+        # i.e /[prefix]/config.json => /config.json
+        rewrite ^/[^/]+/([^/]+)$ /$1 break;
+
+        # 2 - Any static files bundled by webpack
+        # i.e /[prefix]/static/main.js => static.main.js
```
andresmgot: either the comment is wrong or I am not understanding the regex, wouldn't this be:
```
         # i.e /[prefix]/static/main.js => /static/main.js
```
Also, isn't this regex already caught in the first?

dashboard/yarn.lock
```diff
@@ -382,6 +382,14 @@ accepts@~1.3.4:
     mime-types "~2.1.16"
```
andresmgot: why did you need to update the dependencies? (I am asking because the package.json remains the same)

chart/kubeapps/templates/dashboard-config.yaml
migmartri: This comment is wrong, thanks for catching it up!

dashboard/yarn.lock
migmartri: Good catch, this was because in previous versions of the patch I upgraded react-scripts-ts and for some reason yarn.lock did not downgrade. I'll put it back.

Thank you!

chart/kubeapps/templates/dashboard-config.yaml
```diff
@@ -17,6 +17,15 @@ data:
       gzip_static  on;
 
       location / {
+        # Support for ingress prefixes maintaining compatibility with the default /
+        # 1 - Exactly two fragment URLs for files existing inside of the public/ dir
+        # i.e /[prefix]/config.json => /config.json
```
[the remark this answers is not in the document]
migmartri: In kubectl proxy those two rewrites will be ignored as expected since they only match prefix defines in an ingress

In any case kubectl proxy does not work because of the header trim limitation.
 
selection_795

If you mean supporting multiple levels prefixes like /foo/kubeapps/ in the ingress, we do not support it and I think it's fine.

That scenario could be easily fixed if instead of providing all the files in `public` in the root i.e my-instance.com/favicon.png we add a prefix in a similar way we do with `static` `instance.com/public/favicon.png`

Having those two well defined directories would help with rewrites in the future.

prydonius:
> In future PRs I will add the following annotation in the Ingress which will ensure (for users using Nginx ingress controller) that there is always a trailing slash

> nginx.ingress.kubernetes.io/configuration-snippet:  rewrite ^(/[PREFIX])$ $1/ permanent;

I just ran into this issue in my test, I was confused why it wasn't working for me, but I see you have addressed this in #897.

PR #891, opened 2018-12-19
Fix config.json

migmartri:
Apps fail to deploy due to a syntax error in the config file in the chart.


migmartri:
> Thanks! (and sorry for not seeing that)

No problem.

It was not easy to find. The UI seemed to work fine until we did a deployment.

PR #893, opened 2018-12-19
Fix issues when parsing lists

andresmgot:
I found a regression: Previously we were ignoring lists in the AppView but now we are not and the application crashes when the release has a `List`. I have changed this to reiterate over List items.


migmartri:
What is a `List` resource? Which chart did you launch that made the UI crash? What error message did you get?

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -109,30 +109,36 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
...
-            this.getSocket("ingresses", i.apiVersion, i.metadata.name, newApp.namespace),
+            this.getSocket("ingresses", i.apiVersion, item.metadata.name, newApp.namespace),
           );
           break;
         case "Secret":
-          secretNames.push(i.metadata.name);
+          secretNames.push(item.metadata.name);
+          break;
+        case "List":
+          (i as IK8sList<IResource, {}>).items.forEach(listItem => otherResources.push(listItem));
```
prydonius: hmm what if the List contains a Secret or Service or Deployment? We could extract this outer forEach block to a function and call it recursively in the case of a List

dashboard/src/components/AppView/OtherResourcesTable/OtherResourcesTable.tsx
```diff
@@ -24,12 +24,21 @@ class OtherResourcesTable extends React.Component<IAppDetailsProps> {
...
+            </tr>
+          </thead>
           <tbody>
             {otherResources.map((r: IResource) => {
               return (
                 <tr key={`otherResources/${r.kind}/${r.metadata.name}`}>
                   <td>{r.kind}</td>
-                  <td>{r.metadata.namespace}</td>
+                  <td>
+                    {r.metadata.namespace || <i style={{ color: "lightgray" }}>Not Namespaced</i>}
```
prydonius: this isn't really related to this PR right?

andresmgot:
> What is a List resource? Which chart did you launch that made the UI crash?

It is possible to set a `List` kind in a chart to specify several resources. I saw it with the Service Catalog chart:

https://github.com/kubernetes-incubator/service-catalog/blob/master/charts/catalog/templates/rbac.yaml

>  What error message did you get?

Since it didn't match with any of the groups, a `List` was treated as a resource. When trying to access `item.metadata.name` the app crashed with `Unable to read name of undefined`.

dashboard/src/components/AppView/AppView.tsx
andresmgot: 👍 I thought that it was not possible to use those type of resources as `List` but I guess there is nothing that prevent that from happening.

dashboard/src/components/AppView/OtherResourcesTable/OtherResourcesTable.tsx
andresmgot: The PR description contain two points, one for fixing the `List` handling and the second to fix how `otherResources` are represented. I did that in just one PR because I realize that "Other resources" were not being properly represented while fixing the first point. Since it was small enough I didn't see the need to split in two different PRs. Anyway, if you find this conflictive I can split this in two PR.

PR #894, opened 2018-12-19
Implement dashboard changes to add a customCA to a repository

andresmgot:
Closes: #515
Continuation of #878

Frontend changes to support custom CA certificates:

screenshot 2018-12-19 at 17 11 46
screenshot 2018-12-19 at 17 10 48


dashboard/src/actions/repos.ts
```diff
@@ -119,33 +119,45 @@ export const installRepo = (
...
   authHeader: string,
+  customCA: string,
 ): ThunkAction<Promise<boolean>, IStoreState, null, AppReposAction> => {
   return async (dispatch, getState) => {
     try {
       const {
         config: { namespace },
       } = getState();
-      let auth;
+      const auth: { header?: any; customCA?: any } = {};
```
migmartri: why any and not string?

dashboard/src/actions/repos.ts
```diff
@@ -119,33 +119,45 @@ export const installRepo = (
...
+          secrets.authorizationHeader = btoa(authHeader);
+        }
+        if (customCA.length) {
+          auth.customCA = {
+            secretKeyRef: {
+              key: "ca.crt",
+              name: secretName,
+            },
+          };
+          secrets["ca.crt"] = btoa(customCA);
```
migmartri: We might want to be consistent in the way we reference and name the keys

```
secrets.authorizationHeader vs secrets["ca.crt"]
```

dashboard/src/components/Config/AppRepoList/AppRepo.scss
```diff
@@ -0,0 +1,12 @@
+.CertContainer {
+  margin: 0;
+  padding: 0;
+  margin-left: 0.2em;
+  background: inherit;
+}
+
+.CertContent {
+  // Copied from <code> style
+  font-family: "SFMono-Regular", Consolas, Liberation Mono, Menlo, Courier, monospace;
```
migmartri: We do not have any existing class in hex that could help us with that?

dashboard/src/actions/repos.ts
prydonius: the latter is needed because of the `.` in the name of the key, I think it's fine to use the hash notation in that case, but prefer dot notation when we can

dashboard/src/actions/repos.ts
andresmgot: It's not the text, is the `secretKeyRef` (I can avoid the `any` though)

dashboard/src/actions/repos.ts
andresmgot: yes, it's because of the `.`. Note that in other case the linter corrects you so it's not possible to set `secrets["authorizationHeader"]`

dashboard/src/components/Config/AppRepoList/AppRepo.scss
andresmgot: So the hex style for this is to wrap code around a block `<pre><code>...`. The problem here is that `textarea` has its own style that overwrite that and the combination doesn't look good. That's why I need this.

PR #897, opened 2018-12-19
Add Ingress trailing / rewrite

migmartri:
https://github.com/kubeapps/kubeapps/pull/890 companion patch. 

Our react code requires that the custom prefixes i.e `/kubeapps` contains a trailing `/` `/kubeapps/` and this can be guaranteed by adding Nginx ingress controller rewrite rules if a custom prefix is set.

peek 2018-12-19 17-45

### Hostname using default `/`

```yml
helm template -x templates/ingress.yaml chart/kubeapps/
---
# Source: kubeapps/templates/ingress.yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: RELEASE-NAME-kubeapps
  labels:
    app: kubeapps
    chart: kubeapps-1.1.0
    release: RELEASE-NAME
    heritage: Tiller
  annotations:
spec:
  rules:
```  

### Ingress with path to `/kubeapps`

```yml
helm template -x templates/ingress.yaml chart/kubeapps/
---
# Source: kubeapps/templates/ingress.yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: RELEASE-NAME-kubeapps
  labels:
    app: kubeapps
    chart: kubeapps-1.1.0
    release: RELEASE-NAME
    heritage: Tiller
  annotations:
    nginx.ingress.kubernetes.io/configuration-snippet: |
      rewrite ^(/kubeapps)$ $1/ permanent;
spec:

```
Refs https://github.com/kubeapps/kubeapps/issues/402


migmartri:
Set as do-not-merge-yet because its pending some manual testing

migmartri:
> Set as do-not-merge-yet because its pending some manual testing

Tested and ready to be merged

chart/kubeapps/templates/ingress.yaml
```diff
@@ -9,6 +9,11 @@ metadata:
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
   annotations:
+    {{- $rewrites := include "kubeapps.ingressRewrites" . -}}
+    {{- if ne $rewrites ""}}
+    nginx.ingress.kubernetes.io/configuration-snippet: |
+      {{ template "kubeapps.ingressRewrites" $ }}
```
prydonius: I had success with `rewrite ^([^.]*[^/])$ $1/ permanent;` as the snippet here, removing the need to cycle the paths.

prydonius: See https://stackoverflow.com/questions/645853/add-slash-to-the-end-of-every-url-need-rewrite-rule-for-nginx, this regex "[add a slash to the end] if the URI doesn't have a period and does not end with a slash"

migmartri:
Thanks for checking! It makes sense.

migmartri:
We will land this change to support the custom path but I'll look into a way to make it generic or move it to the config.

chart/kubeapps/templates/ingress.yaml
migmartri: It seems that this rewrite has consequences since it also rewrites async calls like namespaces or releases, and tiller proxy does not seem to be able to handle that call.

selection_802

PR #899, opened 2018-12-20
Improve Other Resources rendering

andresmgot:
Split from  #893

> I have also noticed that the table for other resources didn't have a header and that when the resource was not namespaces the field was blank. I've changed that as well to show more info:

screenshot 2018-12-19 at 12 25 34


prydonius:
thanks for splitting it out!

PR #902, opened 2018-12-21
Update chartsvc and chart-repo to v1.2.0

andresmgot:
The current chart-repo image cannot be pulled since `latest` was rebuilt yesterday. Apparently it's not possible to use the `sha256` of an old image (that behavior is different from the Docker Hub). I am changing the tag to use `latest` until there is a new release.

Note that Quay.io is under an incident: http://status.quay.io/incidents/72jyq4322h85 so I am not sure if the issue (not able to pull an old sha256) is actually caused by that incident.

### DIFF
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -101,7 +101,7 @@ apprepository:
     registry: quay.io
     # TODO: Update tag when a new release is available
     repository: helmpack/chart-repo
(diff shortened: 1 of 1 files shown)


prydonius:
I just released v1.2.0 in Monocular and updated this PR to use that version, with this we should be able to fix the CI. Thanks @andresmgot!

Issue #905, opened 2018-12-28
Spurious `v` in bottom version number


migmartri:
The bottom header seems to contain an extra `v` prefix. This is due that our versions already include `v` in the name as described in the release doc.

We would need to either revisit our use of `v` in the tag name or remove the hardcoded prefix from the UI.

![selection_796](https://user-images.githubusercontent.com/24523/50524919-525cf100-0a8d-11e9-9c2f-8c97f1cf2eea.png)

Refs https://github.com/kubeapps/kubeapps/issues/608

PR #907, opened 2018-12-28
Remove version prefix

migmartri:
Removes `v` prefix and adds link to the release page.

selection_801

Fixes https://github.com/kubeapps/kubeapps/issues/905

### DIFF
--- a/dashboard/src/components/Footer/Footer.tsx
+++ b/dashboard/src/components/Footer/Footer.tsx
@@ -9,6 +9,10 @@ interface IFooterProps {
 }
 
 const Footer: React.SFC<IFooterProps> = props => {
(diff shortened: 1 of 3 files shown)


dashboard/src/components/Footer/Footer.tsx
```diff
@@ -9,6 +9,10 @@ interface IFooterProps {
 }
 
 const Footer: React.SFC<IFooterProps> = props => {
+  const linkColor = {
+    color: "#FFF",
```
prydonius: Why is this necessary? Can't we just use the `type-color-white` class?

prydonius:
thanks for the test 🙏

PR #909, opened 2019-01-01
Fix comparison issue

carrodher:
Fixes https://github.com/bitnami/charts/issues/1007:

```
at <ne .path "/">: error calling ne: invalid type for comparison
```
This error appears because if a value is unset, it is returned as a nil value, so we need to replace
```diff
-{{- if ne .path "/" -}}
+{{- if (.path) and ne .path "/" -}}
```

### DIFF
--- a/chart/kubeapps/templates/_helpers.tpl
+++ b/chart/kubeapps/templates/_helpers.tpl
@@ -138,7 +138,7 @@ Trailing / is required for the react app to work
 */}}
 {{- define "kubeapps.ingressRewrites" -}}
 {{ range .Values.ingress.hosts -}}
-{{- if ne .path "/" -}}
+{{- if (.path) and ne .path "/" -}}
       rewrite ^({{ .path }})$ $1/ permanent;
 {{- end -}}
 {{- end }}
--- a/chart/kubeapps/Chart.yaml
+++ b/chart/kubeapps/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kubeapps
-version: 1.2.0
+version: 1.2.1
 appVersion: DEVEL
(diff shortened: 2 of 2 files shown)


prydonius:
Hmm, that value is set by default. What values are you using that make it unset? I'm wondering if we should be okay requiring that to be set.

carrodher:
> Hmm, that value is set by default. What values are you using that make it unset? I'm wondering if we should be okay requiring that to be set.

Using 
```
helm install --name kubeapps --namespace kubeapps bitnami/kubeapps \
   --set ingress.enabled=true \
   --set ingress.certManager=true \
   --set ingress.hosts[0].name=kubeapps.mydomain.com \
   --set ingress.hosts[0].tlsSecret=kubeapps-tls
```
as it is indicated in the [readme](https://github.com/kubeapps/kubeapps/tree/master/chart/kubeapps#tls), it doesn't work

chart/kubeapps/templates/_helpers.tpl
```diff
@@ -138,7 +138,7 @@ Trailing / is required for the react app to work
 */}}
 {{- define "kubeapps.ingressRewrites" -}}
 {{ range .Values.ingress.hosts -}}
-{{- if ne .path "/" -}}
+{{- if (.path) and ne .path "/" -}}
```
prydonius: Looks like this condition is wrong and is producing the rewrite rule even when `path = "/"`, which ends up causing a redirect loop in Kubeapps. Just checked and the correct way to do this is:

```
{{- if .path -}}
{{- if ne .path "/" -}}
      rewrite ^({{ .path }})$ $1/ permanent;
{{- end -}}
{{- end -}}
```

However with #910, we may not need to fix this condition.

Issue #882, opened 2018-12-17
Access URLs and services list stuck in loading 


migmartri:
If you deploy an application and then delete for example a service, the UI will get stuck in loading state.

This is likely due to us waiting for websockets to return state but this never arrives. 

![kubeapps dashboard](https://user-images.githubusercontent.com/24523/50106262-fbd1f600-01f4-11e9-9762-21556589fe44.png)

migmartri:
Note that `helm status` reports the service from being missing 
```
==> MISSING
KIND              NAME
{ v1 services}    turbulent-grizzly-ghost
 ```

**Comment by migmartri** [comment]
Same problem appears if you see a release that has been deleted but not purged.

**Comment by prydonius** [comment]
Hmm, good find. I'm not sure if there's a way from the WebSockets that we can tell if the resource was actually found or not (since we're watching the list endpoint and we only get an event if it exists). I guess there are two options I can think of:

- Have a timeout for the loading indicator, if after 30s we don't hear back, just show nothing
- Make a regular HTTP request for the resource before/in parallel to the WebSocket to check for a 404 response

What should the UX be if a resource isn't found? We can warn that the release and actual state is inconsistent and prompt the user to upgrade the chart, for example.

**Comment by andresmgot** [comment]
> Make a regular HTTP request for the resource before/in parallel to the WebSocket to check for a 404 response

My +1 to that. That's how the SecretsTable behaves. The problem is when more than one element is missing, in that case, right now, it will only show the latest error received.

I wouldn't do the timeout thing because for a deleted app it would be weird to wait 30 s to get the notification (which should be immediate).

Ideally I think that we should parse the elements from the release manifest, if a 404 is received when trying to get the object we could add a warning icon that the item is missing.

**Comment by migmartri** [comment]
There might be a way to sanitize the manifest before we subscribe to the different resources. With Helm status we get
 
```==> MISSING
KIND              NAME
{ v1 services}    turbulent-grizzly-ghost`
```

so there might be a way for us to remove the resources in the interception of those two sets.

migmartri:
w.r.t making a request from the frontend instead of waiting for the ws to return, I think it can make sense since at least for me, the websocket handshake and response seems quite slow compared to the http requests for the secrets.

prydonius:
> There might be a way to sanitize the manifest before we subscribe to the different resources. With Helm status we get

[...]

PR #918, opened 2019-01-09
more robust way of loading resources for AppView sections

prydonius:
- stores data retrieved from WebSocket connections in Redux store
- wraps ServiceItems in a container which gets the relevant Service from the store
- ServiceItems executes an HTTP fetch for the Service to retrieve data
  quicker, as well as discover errors such as missing resources

refs #882

screen shot 2019-01-09 at 20 13 45

kubeapps-serviceitemloading


dashboard/src/components/AppView/ServicesTable/ServiceItem.tsx
```diff
@@ -1,32 +1,73 @@
...
+  IKubeItem,
+  IResource,
+  IResourceRef,
+  IServiceSpec,
+  IServiceStatus,
+} from "../../../shared/types";
 
 interface IServiceItemProps {
-  service: IResource;
+  serviceRef: IResourceRef;
```
migmartri: Maybe serviceRef is redundant? Couldn't you get the name from `service: IKubeItem<IResource>;`?

dashboard/src/components/AppView/ServicesTable/ServiceItem.tsx
```diff
@@ -1,32 +1,73 @@
...
-        </td>
+        <td>{serviceRef.name}</td>
+        {this.renderServiceInfo(service)}
       </tr>
     );
   }
 
-  private getExternalIP(): string {
-    const { service } = this.props;
+  private renderServiceInfo(service?: IKubeItem<IResource>) {
```
migmartri: I like this code, it's flow is quite easy to understand.

dashboard/src/components/AppView/ServicesTable/ServicesTable.tsx
```diff
@@ -38,10 +33,9 @@ class ServiceTable extends React.Component<IServiceTableProps> {
             </tr>
           </thead>
           <tbody>
-            {services.map(
-              s =>
-                s.item && <ServiceItem key={`services/${s.item.metadata.name}`} service={s.item} />,
-            )}
+            {serviceRefs.map(s => (
+              <ServiceItem key={`services/${s.namespace}/${s.name}`} serviceRef={s} />
```
migmartri: `key={`services/${s.namespace}/${s.name}`}` related to my comment above. Wouldn't be great if serviceRef (s) could have a method to get this?

dashboard/src/containers/ServiceItemContainer/ServiceItemContainer.ts
```diff
@@ -0,0 +1,50 @@
...
+function mapStateToProps({ kube }: IStoreState, props: IServiceItemContainerProps) {
+  const { serviceRef } = props;
+  const serviceKey = Kube.getResourceURL(
+    serviceRef.apiVersion,
+    "services",
+    serviceRef.namespace,
+    serviceRef.name,
+  );
+  return {
+    ...props,
```
migmartri: I'd be more specific on what we pass here. Just for clarity. 

I wrote it in another comment but I am not sure why you need to pass `serviceRef` to the component.

dashboard/src/containers/AppViewContainer/AppViewContainer.tsx
```diff
@@ -31,6 +31,8 @@ function mapDispatchToProps(dispatch: ThunkDispatch<IStoreState, null, Action>)
     deleteApp: (releaseName: string, ns: string, purge: boolean) =>
       dispatch(actions.apps.deleteApp(releaseName, ns, purge)),
     getApp: (releaseName: string, ns: string) => dispatch(actions.apps.getApp(releaseName, ns)),
+    receiveResource: (payload: { key: string; resource: IResource }) =>
```
migmartri: This is temporary while we move the websockets handling outside of this component/container right? If that's the case I'd add a `TODO` note in this code and also inside the component.

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -129,23 +134,42 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
...
       return _.dropWhile(array, r => r.item && r.item.metadata.name === resource.metadata.name);
     };
+    let apiVersion = "v1";
+    let apiResource: string;
     switch (resource.kind) {
       case "Deployment":
         const newDeps = dropByName(this.state.deployments);
         newDeps.push(newItem);
         this.setState({ deployments: newDeps });
+        // We won't use resource.apiVersion as it may have a different version
```
andresmgot: I am not sure I understand this comment, why don't we want to use the `apiVersion` in the cache?

dashboard/src/components/AppView/ServicesTable/ServiceItem.tsx
andresmgot: `service` may be still empty, that's why serviceRef may be needed

dashboard/src/containers/ServiceItemContainer/ServiceItemContainer.ts
```diff
@@ -0,0 +1,50 @@
...
+import { IResourceRef, IStoreState } from "../../shared/types";
+
+interface IServiceItemContainerProps {
+  serviceRef: IResourceRef;
+}
+
+function mapStateToProps({ kube }: IStoreState, props: IServiceItemContainerProps) {
+  const { serviceRef } = props;
+  const serviceKey = Kube.getResourceURL(
+    serviceRef.apiVersion,
```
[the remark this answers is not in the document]
prydonius: Interesting idea, I think that would be quite nice actually

dashboard/src/containers/ServiceItemContainer/ServiceItemContainer.ts
prydonius: I use `serviceRef` in `ServiceItem` to display the name of the Service whilst fetching (the screenshot doesn't show it, but the name is shown right away, with a loading icon for the rest of the rows. It's not very important, but I like being able to see what resources are loading.

prydonius: screen shot 2019-01-10 at 13 09 13

prydonius: Now that I think about it, I can change this to just send a `name` prop to the ServiceItem component, since it doesn't need the full `serviceRef`!

dashboard/src/containers/AppViewContainer/AppViewContainer.tsx
prydonius: Yep, good shout - I'll add a TODO!

dashboard/src/components/AppView/AppView.tsx
prydonius: So I need to do some testing here, my concern is that the API server will convert, for example an `extensions/v1beta1` to an `apps/v1` version for Deployments if available. So a chart manifest with an older API version will lead to inconsistencies. I think we need to fix the API version for fetching Deployments to `apps/v1` regardless of what the server returns or is in the release manifest to avoid this inconsistency.

prydonius: On further testing, the API server doesn't actually change the apiVersion field on the object, so I'm able to consistently get `extensions/v1beta` back from the API. Given this, I think we can change this to use the `apiVersion` from the returned resource, though we should keep this in mind in case it can cause issues in the future.

dashboard/src/components/AppView/ServicesTable/ServicesTable.tsx
prydonius: Opted to just use the full resource URL here rather than creating another specific key, it has all we need to be unique and it doesn't matter if this key is verbose.

dashboard/src/components/AppView/ServicesTable/ServicesTable.test.tsx
```diff
@@ -34,51 +33,55 @@ it("renders a message if there are no services or ingresses", () => {
...
-      .props(),
-  ).toMatchObject({ service: service1 });
-  expect(
-    wrapper
-      .find(ServiceItem)
-      .at(1)
-      .props(),
-  ).toMatchObject({ service: service2 });
-});
+// it("renders a table with a service with a LoadBalancer", () => {
```
migmartri: An alternative to commenting out the code would be to skip it (`xit` instead of `it`). At least you will remember to implement it.

dashboard/src/containers/ServiceItemContainer/ServiceItemContainer.ts
```diff
@@ -0,0 +1,44 @@
...
+
+interface IServiceItemContainerProps {
+  serviceRef: ResourceRef;
+}
+
+function mapStateToProps({ kube }: IStoreState, props: IServiceItemContainerProps) {
+  const { serviceRef } = props;
+  return {
+    name: serviceRef.name,
+    service: kube.items[serviceRef.getResourceURL()],
```
migmartri: Much cleaner now.

dashboard/src/shared/ResourceRef.ts
```diff
@@ -0,0 +1,43 @@
+import { Kube } from "./Kube";
+import { IResource } from "./types";
+
+class ResourceRef {
+  // Creates a new ResourceRef instance from an existing IResource. Provide
+  // defaultNamespace to set if the IResource doesn't specify a namespace.
+  // TODO: add support for cluster-scoped resources, or add a ClusterResourceRef
+  // class.
+  public static newFromResource(r: IResource, defaultNamespace: string = "default") {
```
migmartri: why isn't this a proper class constructor? This sounds GO idiomatic.

dashboard/src/shared/ResourceRef.ts
```diff
@@ -0,0 +1,43 @@
+import { Kube } from "./Kube";
+import { IResource } from "./types";
+
+class ResourceRef {
```
migmartri: I like it. I'd maybe add a comment explaining what this class means.

dashboard/src/components/AppView/AppView.tsx
andresmgot: +1 thanks for checking (BTW you forgot to remove the comment?)

dashboard/src/components/AppView/ServicesTable/ServicesTable.test.tsx
andresmgot: do you plan to remove the tests?

prydonius: good point

dashboard/src/shared/ResourceRef.ts
prydonius: What if we want to create one without an IResource, in tests for example?

andresmgot: the pattern we have been following is to create a dummy object with the required elements that we reuse ass "default" props

dashboard/src/components/AppView/ServicesTable/ServicesTable.test.tsx
andresmgot: but if the tests doesn't apply anymore, shoudln't we delete them? (as far as they are applied where they belong). I am not sure if I am missing something

dashboard/src/shared/ResourceRef.ts
prydonius: I think the issue is that that approach doesn't work for classes because it requires the `getResourceURL` to be present in that dummy object, but let me try again.

dashboard/src/components/AppView/ServicesTable/ServiceItem.test.tsx
```diff
@@ -43,6 +53,9 @@ it("renders a view with IP", () => {
     spec: { ports: [{ port: 80 }], type: "LoadBalancer" },
     status: { loadBalancer: { ingress: [{ ip: "1.2.3.4" }] } },
   } as IResource;
-  const wrapper = shallow(<ServiceItem service={service} />);
+  kubeItem.item = service;
+  const wrapper = shallow(
+    <ServiceItem service={kubeItem} name={service.metadata.name} getService={jest.fn()} />,
+  );
   expect(wrapper).toMatchSnapshot();
 });
```
andresmgot: We would need to move here the tests to ensure that `"ItBehavesLike"` a loading component and that it renders an error if present.

prydonius: yep, although ItBehavesLike can't be used because we don't use `LoadingWrapper`, but rather use `LoadingSpinner` directly.

dashboard/src/components/AppView/ServicesTable/ServicesTable.test.tsx
prydonius: Yes, I just commented them out temporarily because it wasn't compiling in the dev environment

dashboard/src/components/AppView/ServicesTable/ServiceItem.test.tsx
andresmgot: ah, I didn't notice that. Could you add a comment there explaining why do you use that instead of the LoadingWrapper? (Since we usually use that by default).

Also then here we would need to create more verbose tests just checking if the LoadingSpinner exists,

prydonius: Good shout, and after discussing with @migmartri I've changed this to use LoadingWrapper. Also created https://github.com/kubeapps/kubeapps/issues/925 for ConfirmDialog which is using LoadingSpinner directly. Thanks!

dashboard/src/components/AppView/AppView.tsx
prydonius: done, thanks!

dashboard/src/shared/ResourceRef.ts
prydonius: I switched this to a proper constructor - thanks!

dashboard/src/shared/ResourceRef.ts
prydonius: done

prydonius:
@andresmgot @migmartri I believe I've addressed all the review feedback now and added some more test coverage - PTAL!

dashboard/src/shared/ResourceRef.test.ts
```diff
@@ -0,0 +1,78 @@
+import { Kube } from "./Kube";
+import ResourceRef from "./ResourceRef";
+import { IResource } from "./types";
+
+describe("ResourceRef", () => {
+  describe("constructor", () => {
+    it("it returns a ResourceRef with the correct details", () => {
+      const r = {
```
migmartri: I was not that difficult in the end wasn't it?

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -231,6 +262,7 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
           break;
         case "Service":
           result.services.push(resource);
+          result.serviceRefs.push(new ResourceRef(resource.item, namespace));
```
migmartri: So the namespace in this case might not be inside the resourceItem? What's `namespace` in this case?

dashboard/src/components/AppView/ServicesTable/ServiceItem.tsx
```diff
@@ -1,32 +1,67 @@
...
+    if (service === undefined || service.isFetching) {
+      return (
+        <td colSpan={4}>
+          <LoadingWrapper size="inline-small" />
```
migmartri: So this is going to be showing a loading message per element being loaded? It might be too much if we extrapolate this to let's say deployments no?

Do we need to show loading at all here? I thought that we already had some kind of struct that indicated that at least one resource was being loaded and we used that in the whole table?

dashboard/src/components/AppView/AppView.tsx
prydonius: The release namespace is used when the resource in the manifest doesn't have a namespace (likely since charts leave the namespace undefined so it deploys in whatever namespace the user chooses). I'll rename this to `releaseNamespace`, didn't want to before since it was already that.

dashboard/src/components/AppView/ServicesTable/ServiceItem.tsx
prydonius: Correct, this was intended, we now show the name of the resource and a loading icon next to it. Each resource is loaded independently in parallel so it's possible that some are still loading whilst others have data to show.

> It might be too much if we extrapolate this to let's say deployments no?

What will be too much? In most cases it loads pretty fast and I don't think the spinners are too annoying (that said I haven't seen what this looks like on a chart with a lot of Services).

> Do we need to show loading at all here? I thought that we already had some kind of struct that indicated that at least one resource was being loaded and we used that in the whole table?

The table used to ensure all Services were loaded before displaying anything, but now it just renders `ServiceItem` containers. We could wrap the ServiceTable into a container and have a single loader, but I actually _like_ that we show each Service loading independently, it feels more responsive to me.

andresmgot: For me it's okay to have a loading message per element. If having so many spinners is annoying we can remove it and have just a "Loading..." message.

dashboard/src/shared/ResourceRef.ts
```diff
@@ -0,0 +1,44 @@
...
+  public apiVersion: string;
+  public kind: string;
+  public name: string;
+  public namespace: string;
+
+  // Creates a new ResourceRef instance from an existing IResource. Provide
+  // defaultNamespace to set if the IResource doesn't specify a namespace.
+  // TODO: add support for cluster-scoped resources, or add a ClusterResourceRef
+  // class.
+  constructor(r: IResource, defaultNamespace: string = "default") {
```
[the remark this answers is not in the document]
prydonius: Hmm good question, the issue here is that if the IResource defines a namespace in `metadata.namespace`, that will always be used. The `defaultNamespace` is only used if `metadata.namespace` is undefined. If you know that the IResource you're passing to `ResourceRef` contains the namespace, you don't need to provide the `defaultNamespace`, if you did, you'd have to set it to `resource.metadata.namespace` which would be a bit annoying.

I do get that this is a bit confusing though, as if `metadata.namespace` is undefined, you probably will need to provide `defaultNamespace`, it's unlikely you'd want that to be "default".

Another option here is to have the signature be `defaultNamespace?: string`, and then in the code we would need to throw an error if both `metadata.namespace` and `defaultNamespace` are undefined. However, I'm not sure it's worth doing that.

dashboard/src/components/AppView/ServicesTable/ServiceItem.tsx
prydonius: This is what it looks like, I quite like it. IMO it's subtle and not that annoying.

kubeapps-serviceitemloading

migmartri: It does not look bad actually. I still have a preference for a global (per table) loading spinner but I understand that it complicates and couples the code, so I'd not push for it.

Just being picky and since asking for things is free. Something that annoys me is the alignment of the spinners. Something very slick would be the use of blurry lines a.l.a Facebook or Slack https://kyusuf.com/post/fake-it-til-you-make-it-css/

dashboard/src/shared/ResourceRef.ts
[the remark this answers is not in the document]
andresmgot: I guess it's better then to throw the error if `metadata.namespace` and `defaultNamespace` are unset. It could be quite confusing and difficult to debug if suddenly there are unexpected requests to the `default` namespace.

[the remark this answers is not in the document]
prydonius: done, and added a test - thanks!

PR #920, opened 2019-01-11
Remove Namespace column from other resources

### LINKED ISSUE #906: Namespace value missing in app view
Other resources do not seem to show the namespace the resources are in.

For example 
```
$ k get cm kubeapps-110-frontend-config -o jsonpath={.metadata.namespace}`
miguel-kubeapps-110
```

but as you can see below, the table says `non namespaced` for all the resources.

selection_798

andresmgot:
Fixes #906 

The manifest doesn't always contain the namespace of the resources it contains since, by default, it resolves to the namespace in which the application is deployed. We don't have a mechanism to differ between namespaced and non-namespaced components so we'd need to remove that column.


prydonius:
Ah yes, I see the issue. I think this is not important enough to have a workaround to try to determine the non-namespaced resources, so I'm okay with just removing this column. We can always revisit it if we want, though this page is going to be redesigned anyway.

PR #927, opened 2019-01-16
Add role for kubeapp-repositories-read. Fixes #926

### LINKED ISSUE #926: create apprepository read role with chart
It would make sense to create this role: https://raw.githubusercontent.com/kubeapps/kubeapps/master/docs/user/manifests/kubeapps-repositories-read.yaml with the chart, so that what's left is for consumers to bind groups/users to that role. Kubeapps knows best how to define the role now and in future (if it changes the resources/apis in use)

davidkarlsen:
Creates the role, so that users need only be bound to the role to be able to use kubeapps for deployments.

Signed-off-by: David J. M. Karlsen <david@davidkarlsen.com>


davidkarlsen:
@andresmgot fixed

docs/user/access-control.md
```diff
@@ -120,6 +120,9 @@ kubectl create clusterrolebinding example-kubeapps-service-catalog-admin --clust
 
 In order to list the configured App Repositories in Kubeapps, create and apply the
 `kubeapps-repositories-read` Role in the namespace Kubeapps is installed in.
+If you installed kubeapps with the helm-chart, this role will be present in the target
```
andresmgot: In our case, using the chart is the only way of getting Kubeapps so what I would say is that a role named `kubeapps-repositories-read` already exists and it's necessary to get the different repositories installed. Then you can remove the line below that applies the manifest from the URL

andresmgot: BTW, now that I take a look again to the docs, we can also create the role `*-repositories-write` since the use case is very similar to this one (and update the docs).

davidkarlsen: There are a number of references in the doc to manifest files to create resources with kubectl - I guess all should go then?

prydonius:
Thanks @davidkarlsen, I guess we should update the docs to use the installed role, we'd just need to document that the name of the role is $RELEASE_NAME-repositories-read/write.

docs/user/access-control.md
```diff
@@ -53,24 +53,12 @@ kubectl create -n default rolebinding example-view \
...
 available in most Kubernetes distributions, you can find more information about
 that role
 [here](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles).
 
-Additionally, we need to create a role to give read access to App Repositories and bind it to our service account.
-
-```
-export KUBEAPPS_NAMESPACE=kubeapps
-kubectl apply -n $KUBEAPPS_NAMESPACE -f https://raw.githubusercontent.com/kubeapps/kubeapps/master/docs/user/manifests/kubeapps-repositories-read.yaml
-kubectl create -n default rolebinding example-edit \
```
andresmgot: sorry, even though it's not necessary to create the role (line above) it's necessary to create both rolebindings below

davidkarlsen: Let's create rolebinding for the service-account in the chart - then the docs are correct, as what's left to do is to bind any users/groups according to the docs.

andresmgot: The problem with the role binding is that it depends on the namespace and the service account name that the user is going to use, in this example is `default:example` but the user will need to modify that for other users.

davidkarlsen: Not sure I follow.
The serviceaccount for kubeapps is created here: https://github.com/kubeapps/kubeapps/blob/master/chart/kubeapps/templates/apprepository-serviceaccount.yaml - and if bound to the role that one is OK.

What is left then is to bind selected users/groups to the role too, and this is a manual task, as documented.

andresmgot: It's a bit complicated. The service account your are pointing to (`apprepository-serviceaccount`) is the one used by the `apprepository-controller`. That service account is properly bound to the [role](https://github.com/kubeapps/kubeapps/blob/master/chart/kubeapps/templates/apprepository-rbac.yaml) necessary to manage apprepositories. Note that this role have more privileges than the "read" ones.

What this documentation says is that the user account created in the [first step](https://github.com/kubeapps/kubeapps/blame/master/docs/user/access-control.md#L22) should be bound to the `apprepositories-read` and `apprepositories-write` roles that you have uploaded in this PR. If the user is not using the namespace `default` or the account `example` he should adapt the command and that's why we cannot pre-create the bindings for them.

davidkarlsen: Yes -  I get that.
So to sum up.
Two roles are pre-created, it is the up to the user to bind accounts to to them with rolebindings, as described in the documentation:

```
In order to list the configured App Repositories in Kubeapps, [bind users/groups Subjects](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#command-line-utilities) to the `$RELEASE_NAME-repositories-read` role in the namespace Kubeapps was installed into by the helm chart.

#### Write access to App Repositories

In order to create and refresh App Repositories in Kubeapps, create and apply the
`kubeapps-repositories-write` Role in the namespace Kubeapps is installed in.
Likewise to the read access bind users/group Subjects to the
`$RELEASE_NAME-repositories-write` Role in the namespace Kubeapps is installed in
for users to create and refresh App Repositories in Kubeapps
```

prydonius: I think what @andresmgot is getting at here is we should add back the command instructions, rather than just have the verbal instructions, i.e.:

```
export KUBEAPPS_NAMESPACE=kubeapps
export KUBEAPPS_RELEASE_NAME=kubeapps
kubectl create -n default rolebinding example-edit \
  --clusterrole=edit \	
  --serviceaccount default:example	
kubectl create -n $KUBEAPPS_NAMESPACE rolebinding example-kubeapps-repositories-read \	
  --role=$KUBEAPPS_RELEASE_NAME-repositories-read \	
  --serviceaccount default:example
```

I've updated it to use the `$KUBEAPPS_RELEASE_NAME-repositories-read` role that is created by the chart instead of the one that was previously being created here by the `kubectl apply -f` command.

The other documentation should be updated similarly.

davidkarlsen:
OK - I think that should be it for the kubeapp-repositories.
What about the cluster-roles - should they be made by the chart too - so that there is only rolebinding to do for them too? If yes, I think we do that in a new PR.

docs/user/access-control.md
```diff
@@ -53,24 +53,12 @@ kubectl create -n default rolebinding example-view \
 #### Write access to Applications within a namespace
 
 In order to create, update and delete Applications in a namespace, apply the
-`edit` ClusterRole in the desired namespace and the `kubeapps-repositories-read`
+`edit` ClusterRole in the desired namespace and the `$RELEASE_NAME-repositories-read`
 Role in the namespace Kubeapps is installed in. The `edit` ClusterRole should be
 available in most Kubernetes distributions, you can find more information about
 that role
 [here](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles).
```
andresmgot: It's still needed to restore this block (closing the previous conversation:

```suggestion
kubectl create -n default rolebinding example-edit \
  --clusterrole=edit \
  --serviceaccount default:example
kubectl create -n $KUBEAPPS_NAMESPACE rolebinding example-kubeapps-repositories-read \
  --role=kubeapps-repositories-read \
  --serviceaccount default:example
```

andresmgot:
> What about the cluster-roles - should they be made by the chart too - so that there is only rolebinding to do for them too? If yes, I think we do that in a new PR.

What cluster roles? If you mean the `edit` and `read` ones should be already in the cluster. We should avoid creating clusterroles.

davidkarlsen:
> > What about the cluster-roles - should they be made by the chart too - so that there is only rolebinding to do for them too? If yes, I think we do that in a new PR.
> 
> What cluster roles? If you mean the `edit` and `read` ones should be already in the cluster. We should avoid creating clusterroles.

these: https://github.com/kubeapps/kubeapps/tree/master/docs/user/manifests could be created by helm as well, so people just need to bind to them for using them.

andresmgot:
> these: https://github.com/kubeapps/kubeapps/tree/master/docs/user/manifests could be created by helm as well, so people just need to bind to them for using them.

Those cluster roles are only used with the Service Catalog which is optional, it's better not to install those if it's not needed.

docs/user/access-control.md
```diff
@@ -118,24 +114,30 @@ kubectl create clusterrolebinding example-kubeapps-service-catalog-admin --clust
...
+Likewise to the read access bind users/group Subjects to the
+`$RELEASE_NAME-repositories-write` Role in the namespace Kubeapps is installed in
+for users to create and refresh App Repositories in Kubeapps
 
 ```
 export KUBEAPPS_NAMESPACE=kubeapps
-kubectl apply -n $KUBEAPPS_NAMESPACE -f https://raw.githubusercontent.com/kubeapps/kubeapps/master/docs/user/manifests/kubeapps-repositories-write.yaml
-kubectl create -n $KUBEAPPS_NAMESPACE rolebinding example-kubeapps-repositories-write --role=kubeapps-repositories-write --serviceaccount default:example
+export KUBEAPPS_RELEASE_NAME=kubeapps
+kubectl create -n default rolebinding example-edit \
```
prydonius: This edit role isn't needed, right?

andresmgot: True, it's not needed

docs/user/access-control.md
```diff
@@ -53,17 +53,13 @@ kubectl create -n default rolebinding example-view \
...
 [here](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles).
 
-Additionally, we need to create a role to give read access to App Repositories and bind it to our service account.
-
 ```
-export KUBEAPPS_NAMESPACE=kubeapps
-kubectl apply -n $KUBEAPPS_NAMESPACE -f https://raw.githubusercontent.com/kubeapps/kubeapps/master/docs/user/manifests/kubeapps-repositories-read.yaml
 kubectl create -n default rolebinding example-edit \
   --clusterrole=edit \
   --serviceaccount default:example
```
prydonius: doesn't this need to be updated to point to the role using the Kubeapps release name?

andresmgot: true, I missed that hidden part

davidkarlsen: Hm - this started clean and easy... :)

My initial changes were to define the roles in the `kubeapps` namespace for *apprepositories* - `git show b7afc95fd0673a0ade26c6d59c363f291dab14d0` removed them as manifest-files, and added to the helm template in `18a156471a60473621cd97b73f66219493c0eb30` and `4cba08837be8a25459f2607faf186ff16cced707`.

However - the relevant documentation is really from line 113 and onwards (for repositories).

The section at around line 60 is about apps.

Where do we want to go with this?
Add all roles to the chart? Add cluster roles?

andresmgot: You are right @davidkarlsen, this PR is getting longer than it should.

> The section at around line 60 is about apps.

Yes, but the thing is that for deploying an app the user needs to have permission to read repositories in the Kubeapps namespace so that's why these commands are duplicated here and in the section about AppRepositories (line 113).

> Where do we want to go with this?
Add all roles to the chart? Add cluster roles?

That's not needed, the only missing change is to modify the line 71 to use `s/kubeapps-repositories-read/$RELEASE_NAME-repositories-read/`. Let me change it myself to clarify it.

andresmgot: This should be read now if I have not missed anything.

Issue #883, opened 2018-12-17
Authentication improvements (design doc)


migmartri:
One of our main items for the next Kubeapps releases is to improve the authentication/authorization usability and supported options.

Currently, we require users to input an auth token in a login form, token that has been previously provided by a Kubeapps administrator. We want to simplify this flow by leveraging single sign on if the user has it set it up for their organization.

Roughly speaking this might mean:

Features

* Leverage existing OIDC setup to authenticate users https://github.com/kubeapps/kubeapps/issues/385
* Offer an easy way to setup their own single sign on authentication flow without requiring to modify the api-server flags. This means that it should work in any k8s cluster and pointing to any oidc provider.
* Ideally, be ready to support additional auth methods like SAML, LDAP.  

Implementation improvements

* Look into using jwt instead of bearer tokens https://github.com/kubeapps/kubeapps/issues/881 

This task is about writing a design document that outlines possible solutions to discuss.

migmartri:
I am working on this design doc, although it is not completed, feel free to skim through it https://docs.google.com/document/d/1YZzLtIbS2copQJgspFiMd0eAhpyt8u19MUSDEvH2X4g/edit?usp=sharing

migmartri:
The doc has been update and is ready for review

davidkarlsen:
Interesting issue.
We are on ICP:  IBM Cloud Private, https://www.ibm.com/support/knowledgecenter/en/SSBS6K_3.1.1/user_management/auth_onboard.html
via the cmd line tool "cloudctl" I login with my corp username and pw, and it returns an JWT (extractable at https://jwt.io/) - I can simply paste this token into the kubeapps UI and I'm in - so what I am looking for is some integration mechanism where I can do SSO/integration with ICP transparently when already logged into ICP. 

I haven't debugged kubeapps, but if I jump in with the token in a http header "Authorization: <thejwt>" - would it work as of today?

davidkarlsen:
> Hi @davidkarlsen,
> 
> First of all sorry, I just noticed that the doc was not fully public.
Thanks - I had a look at the doc - and it looks well founded and though-through!

[...]

PR #928, opened 2019-01-17
Add link to proposal

migmartri:
Adds link to the new OIDC proposal, listed alphabetically.

NOTE: I added the acronyms and extended terms to improve search hit chances.

Refs https://github.com/kubeapps/kubeapps/issues/883

### DIFF
--- a/docs/architecture/design-proposals/README.md
+++ b/docs/architecture/design-proposals/README.md
@@ -4,6 +4,8 @@ This folder contain documents/links used to design new features for Kubeapps.
 
 # Links
 
+- Authentication (AuthN) and Authorization (AuthZ)
+    - [Introduction](./authentication-and-authorization.md)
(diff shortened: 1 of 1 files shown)


migmartri:
> Unrelated to this PR I have a suggestion for the document you are linking. After all the investigation I am missing a conclusion. Based on the suggested approaches, which one is the one you would follow and why? That will help when the time for implementing the feature comes. I think you are now the one with more knowledge about the topic so we would need to know your preferred option.

Thanks @andresmgot 

I left any conclusion off the scope if this doc on purpose. I thought that it would be better to use this doc just as a base for future discussions when the implementation time comes. I would also first like to get validation that those approaches are valid and make sense.

If you find it useful could define an action plan based on my preferences but I'd prefer that we decide it together. cc/ @prydonius

prydonius:
@migmartri do we have a follow up issue to review and decide on an approach for this proposal?

migmartri:
Good point @prydonius, I just created it https://github.com/kubeapps/kubeapps/issues/931

Issue #588, opened 2018-09-06
[Feature request] Show on installed chart if a new chart version exist in repo


obeyler:
As since few days I cannot deploy the last kubeapps I don't know if this feature is present or not :-)

It may be a good idea to show to the operator that some deployed chart can or need to be updated.
either by a cartrige with a different color ( the color can change in function of the number of new release of the chart)
or juste by adding a logo on the cartrige . Need Upgrade in this case if the user click on it the Upgrade will be proposed and done if user agrees

prydonius:
Definitely agree that this would be a great feature to have!

andresmgot:
I've gathered some quick notes about this:

https://docs.google.com/document/d/1oG9nerd5CurWSIwH33kKCsOCtkSFMgcm8SuTJyuSnxs/

@prydonius @migmartri please take a look and add your thoughts if you have a different proposal.

prydonius:
@andresmgot can you open comment access to that doc? I think we can have it public for comments only.

PR #930, opened 2019-01-18
Show available updates

andresmgot:
Ref: #588 
Requires: https://github.com/helm/monocular/pull/592/

Show in the App view when there is an update available. If that's the case a message in the ChartInfo box is shown and the "Upgrade" button gets highlighted:

screenshot from 2019-01-18 17-29-04

If the mouse is over the "Upgrade" button it also shows a message with the new version found:

screenshot from 2019-01-18 17-29-51

If the application is up to date, only a message is shown in the ChartInfo box and the upgrade button is rendered as always:

screenshot from 2019-01-18 17-30-15


dashboard/src/actions/charts.test.tsx
```diff
@@ -23,10 +26,15 @@ beforeEach(() => {
...
   window.fetch = fetchMock;
+  axiosGetMock = jest.fn(() => {
+    return { data: response };
+  });
+  axios.get = axiosGetMock;
 });
 
 afterEach(() => {
   window.fetch = fetchOrig;
+  axios.get = axiosGetOrig;
```
prydonius: doesn't jest have a restore mocks method that does this for us?

dashboard/src/components/AppView/AppControls.tsx
```diff
@@ -103,6 +103,39 @@ class AppControls extends React.Component<IAppControlsProps, IAppControlsState>
   private togglePurge = () => {
     this.setState({ purge: !this.state.purge });
   };
+
+  private renderUpgradeButton = () => {
```
prydonius: perhaps we should make this a stateless component since it's getting a bit more complex?

dashboard/src/components/AppView/AppControls.tsx
```diff
@@ -103,6 +103,39 @@ class AppControls extends React.Component<IAppControlsProps, IAppControlsState>
...
+      app.chart.metadata.version !== latest;
+    let upgradeButton = null;
+    // If the app has been deleted hide the upgrade button
+    if (!deleted) {
+      upgradeButton = (
+        <button className="button" onClick={this.handleUpgradeClick}>
+          Upgrade
+        </button>
+      );
+      // If the app is outdated highligh the upgrade button
```
prydonius:
```suggestion
      // If the app is outdated highlight the upgrade button
```

dashboard/src/components/AppView/AppControls.tsx
```diff
@@ -103,6 +103,39 @@ class AppControls extends React.Component<IAppControlsProps, IAppControlsState>
...
+      if (isOutdated) {
+        upgradeButton = (
+          <div className="tooltip">
+            <button className="button upgrade-button" onClick={this.handleUpgradeClick}>
+              <span className="upgrade-text">Upgrade</span>
+              <ArrowUpCircle color="white" size={25} fill="#20cb15" className="notification" />
```
prydonius: is #20cb15 from the Hex colour palette?

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -97,6 +100,21 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
     if (!newApp) {
       return;
     }
+    if (
+      newApp.chart &&
+      newApp.chart.metadata &&
+      newApp.chart.metadata.name &&
+      newApp.chart.metadata.version &&
+      newApp.chart.metadata.appVersion &&
```
prydonius: what if the chart doesn't have an `appVersion`? We should still be able to check based on `version`.

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -289,6 +308,23 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
...
+      let latest = app.chart.metadata.version;
+      if (this.props.latest) {
+        this.props.latest.forEach(l => {
+          // semver.compare returns -1 if v2 is bigger than v1
+          if (semver.compare(latest, l.relationships.latestChartVersion.data.version) < 0) {
+            latest = l.relationships.latestChartVersion.data.version;
+          }
+        });
+      }
+      return latest;
```
prydonius: why don't we just calculate this result in the action or reducer and pass it directly to AppView? I don't think we need to calculate it here and we don't use all the info from the latest call.

dashboard/src/components/AppView/ChartInfo.tsx
```diff
@@ -1,28 +1,62 @@
...
 import { hapi } from "shared/hapi/release";
 import Card, { CardContent, CardFooter, CardGrid, CardIcon } from "../Card";
 
+import { Link } from "react-router-dom";
 import placeholder from "../../placeholder.png";
 import "./ChartInfo.css";
 
 interface IChartInfoProps {
   app: hapi.release.Release;
+  latest?: string;
```
prydonius: perhaps it should be `newVersion`?

dashboard/src/shared/Chart.ts
```diff
@@ -23,5 +24,13 @@ export default class Chart {
     return true;
   }
 
+  public static async checkUpdates(name: string, version: string, appVersion: string) {
```
prydonius: I think this method should match with the endpoint/handler on the API side. So I think this method could be called `listChartsWithFilters`.

dashboard/src/shared/types.ts
```diff
@@ -81,6 +81,7 @@ export interface IChartState {
     values?: string;
   };
   items: IChart[];
+  latests: { [chart: string]: IChart[] };
```
prydonius: why wouldn't this be part of `selected`?

dashboard/src/actions/charts.ts
```diff
@@ -156,3 +166,18 @@ export function getChartValues(
...
+
+export function getChartUpdates(
+  name: string,
+  version: string,
+  appVersion: string,
+): ThunkAction<Promise<void>, IStoreState, null, ChartsAction> {
+  return async dispatch => {
+    try {
+      const chartUpdates = await Chart.checkUpdates(name, version, appVersion);
+      dispatch(receiveChartUpdates(chartUpdates));
```
prydonius: IMO this action should just extract the newer version (if there is one) and the repo(s) it's from and send that info to the reducer. I don't think we need to store the full filtered charts response in the state and send that to AppView.

dashboard/src/components/AppView/AppControls.tsx
prydonius: We should aim to use colours from the Hex design palette: https://design.bitnami.com/category/Foundations/Variables/#Colors

dashboard/src/components/AppView/AppView.tsx
prydonius: Hmm, well can we match on the fact that `appVersion` is not defined and `version` is? i.e. we'd match with the `azure` repo because in `stable` the `appVersion` is defined. It would be a shame for this whole thing to not work for any chart that does not define the `appVersion`, given it is an optional field in the spec.

dashboard/src/actions/charts.ts
```diff
@@ -38,6 +46,14 @@ export const selectValues = createAction("SELECT_VALUES", resolve => {
   return (values: string) => resolve(values);
 });
 
+export const receiveChartUpdate = createAction("RECEIVE_CHART_UPDATES", resolve => {
+  return (chartUpdate: IChartUpdateCheck) => resolve(chartUpdate);
```
prydonius: the name of this type is confusing and it seems it's only used for this, I would just define an explicit type here of `{ name: string, update: IChartUpdate }`, or use 2 arguments.

dashboard/src/actions/charts.ts
```diff
@@ -156,3 +174,41 @@ export function getChartValues(
...
+export function getChartUpdates(
+  name: string,
+  version: string,
+  appVersion: string,
+): ThunkAction<Promise<void>, IStoreState, null, ChartsAction> {
+  return async dispatch => {
+    try {
+      const chartsInfo = await Chart.listWithFilters(name, version, appVersion);
+      let update: IChartUpdate = {
+        checked: true,
```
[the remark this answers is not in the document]
migmartri: This is because of my suggestion https://github.com/kubeapps/kubeapps/pull/930#discussion_r249965136

Basically to differentiate between the case in which we have checked for updates and when we have not.

Some thoughts on the implementation though. 

I'd not make this action the one in charge of setting the checked flag.

What I'd do is to initialize that flag in the reducer ` updates: { checked: false },` and then set it to true when it receives `actions.charts.receiveChartUpdate` set it to true along with the updates.

I'd also potentially change the redux schema to  updates: { checked: boolean, items?: []IChartUpdates},` that will simplify the merge and also will allow you to not to add this empty update your are adding now.

dashboard/src/actions/charts.ts
andresmgot: I am now using it in more places due to https://github.com/kubeapps/kubeapps/pull/930/files#r250379836

dashboard/src/actions/charts.ts
```diff
@@ -156,3 +174,41 @@ export function getChartValues(
...
+        repository: { name: "", url: "" },
+        latestVersion: "",
+      };
+      chartsInfo.forEach(c => {
+        const chartLatestVersion = c.relationships.latestChartVersion.data.version;
+        // semver.compare(v1, v2) returns -1 if v2 is bigger than v1
+        if (semver.compare(version, chartLatestVersion) < 0) {
```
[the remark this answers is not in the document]
andresmgot: true, I was used to use `compare` but `gt` is more readable

dashboard/src/actions/charts.ts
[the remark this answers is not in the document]
andresmgot: At the end I discarded the `checked` approach. It was blurring the code a lot. We can simply assume that if `update` is undefined it means that it hasn't been checked. I will add a comment to clarify it further.

dashboard/src/containers/AppViewContainer/AppViewContainer.tsx
```diff
@@ -15,14 +15,29 @@ interface IRouteProps {
   };
 }
 
-function mapStateToProps({ apps, kube }: IStoreState, { match: { params } }: IRouteProps) {
+function mapStateToProps({ apps, kube, charts }: IStoreState, { match: { params } }: IRouteProps) {
+  let update: IChartUpdate = {
```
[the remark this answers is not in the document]
andresmgot: it was not possible with the `checked` approach but I discarded that.

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -84,6 +86,16 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
     sockets: [],
   };
 
+  private getUpdates = _.memoize((app: hapi.release.Release) => {
```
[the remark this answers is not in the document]
andresmgot: I am not sure what you want to avoid, the call to `memoize` or calling to `getChartUpdates` here when the components that use it are `ChartInfo` and `AppControls`?

IMO having a component that relies on other component to update the redux store is not the best approach it may lead to a brittle structure. If the component that updates the store changes and no longer does that there is no way for the other component to catch that and avoid the bug.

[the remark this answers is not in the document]
Angelmmiguel: I believe the discussion about calling the method from here or child componentes will depend on how `ChartInfo` and `AppControls` are implemented. If they're just components that show information (presentational), they shouldn't call call methods to initialize the information. This is the job of the componentes that contains the logic of the view.

However, I believe the main issue of this code is that it's out of the React component lifecycle. This method is called in the `render` method, so any update on the component or its branch will trigger a call to `getUpdates`. `_.memoize` makes the trick here, but the call it's still present. If this code changes in the future and don't memoize the content it may cause a performance issue in this view.

Retrieving information from external services must be done during the lifecycle of the component. I recommend to do it right after the component is mount (`componentDidMount`). In this case, this is not possible because this call depends on another API call. So, you can use [`componentDidUpdate`](https://reactjs.org/docs/react-component.html#componentdidupdate) instead. You only need to check the value it's not present yet and **it's not loading at that point**. The second consideration it's very important to avoid calling `getChartUpdates` several times. 

Just to clarify why this is important, all the modification of the state and properties that affects to the component must be known by React. Extracting this logic to internal variables of the class may cause performance issues or wrong updates.

Another issue I found is the import of `lodash`. I've opened a second task for it: https://github.com/kubeapps/kubeapps/issues/939.

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -86,6 +88,26 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
     getApp(releaseName, namespace);
   }
 
+  public async componentDidUpdate(prevProps: IAppViewProps) {
```
Angelmmiguel: Please, avoid using `async` if the method doesn't use `await` internally

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -86,6 +88,26 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
...
+    if (!_.eq(this.props.app, prevProps.app)) {
```
Angelmmiguel: Please, avoid using `lodash` for trivial JS operations. This is the same as:

```
if (this.props.app !== prevProps.app) {
```

ECMAScript already includes a lot of new methods to work with Objects and Arrays so most of the lodash helpers are no longer required.

dashboard/src/actions/charts.test.tsx
```diff
@@ -8,8 +9,8 @@ import { NotFoundError } from "../shared/types";
 const mockStore = configureMockStore([thunk]);
 
 let store: any;
-const fetchOrig = window.fetch;
 let fetchMock: jest.Mock;
+let axiosGetMock: jest.Mock;
```
Angelmmiguel: Not a blocker for this PR, but you can easily mock external modules with jest directly :) https://jestjs.io/docs/en/mock-functions#mocking-modules

dashboard/src/actions/charts.ts
```diff
@@ -156,3 +167,35 @@ export function getChartValues(
     }
   };
 }
+
+export function getChartUpdates(
+  name: string,
+  currentVersion: string,
+  appVersion: string,
+): ThunkAction<Promise<void>, IStoreState, null, ChartsAction> {
+  return async dispatch => {
```
Angelmmiguel: Please, remove the `async` if you're not using `await` inside.

dashboard/src/components/AppView/AppControls/UpgradeButton.tsx
```diff
@@ -0,0 +1,51 @@
...
+interface IUpgradeButtonProps {
+  updateVersion?: string;
+  upgradeURL: string;
+}
+
+interface IUpgradeButtonState {
+  upgrade: boolean;
+}
+
+class UpgradeButton extends React.Component<IUpgradeButtonProps, IUpgradeButtonState> {
```
Angelmmiguel: `react-router` allows you to redirect users programatically. You only need to pass the `history` property from `AppView` to this component and use it to redirect the user:

```jsx
import { ArrowUpCircle } from "react-feather";

interface IUpgradeButtonProps {
  history: any, // You probably want to be more specific
  updateVersion?: string;
  upgradeURL: string;
}
const UpgradeButton = React.SFC<IUpgradeButtonProps> = ({ history, updateVersion, upgradeURL }) => {
  let upgradeButton;

  // Redirect the user
  const onClick = () => history.push(upgradeURL);

  // If the app is outdated highlight the upgrade button
  if (updateVersion) {
    upgradeButton = (
      <div className="tooltip">
        <button className="button upgrade-button" onClick={onClick}>
          <span className="upgrade-text">Upgrade</span>
          <ArrowUpCircle color="white" size={25} fill="#82C341" className="notification" />
        </button>
        <span className="tooltiptext tooltip-top">New version ({updateVersion}) found!</span>
      </div>
    );
  } else {
    upgradeButton = (
      <button className="button" onClick={onClick}>
        Upgrade
      </button>
    );
  }
  return (
    <React.Fragment>
      {upgradeButton}
    </React.Fragment>
  );
}

export default UpgradeButton;
```

dashboard/src/actions/charts.test.tsx
andresmgot: I am not sure how to use it to simplify those test. I would still need to define axios.get as a mock and specify the logic, isn't it?

Angelmmiguel: It simplifies the current code because `axios` would be a mock already:

```js
// Instead of 
let axiosGetMock: jest.Mock;
axiosGetMock = jest.fn(() => {
  return { data: response };
});
axios.get = axiosGetMock;

// Just
jest.mock('axios')
axios.get.mockResolvedValue({ data: response })
```

dashboard/src/components/AppView/AppView.tsx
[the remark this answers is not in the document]
prydonius: Thanks so much for the detailed feedback @Angelmmiguel! @andresmgot I do agree that having dependent components is troublesome, however in this case it might be possible for both components to kick off the call to `getChartUpdates` and the action can deal with avoiding fetching again if we already have it (there could be a race condition, but maybe it's not a huge issue to fetch it twice).

Anyway, I think we can keep this in AppView for now, I'm just concerned that AppView is doing a lot and there must be ways we can break it down and also rely more on `componentDidMount` in child components rather than utilising `componentDidUpdate` which comes with the issue of ensuring guards to not fetch data too much.

dashboard/src/actions/charts.ts
prydonius: Actually we need to return a promise here as it's a redux thunk action. Should the promise be explicitly defined? I like the use of `async` as it avoids the nesting.

dashboard/src/components/AppView/AppControls/AppControls.tsx
```diff
@@ -1,12 +1,16 @@
...
+import { IChartUpdate } from "shared/types";
+import { hapi } from "../../../shared/hapi/release";
+import ConfirmDialog from "../../ConfirmDialog";
+import LoadingWrapper from "../../LoadingWrapper";
+import "./AppControls.css";
+import UpgradeButton from "./UpgradeButton";
 
 interface IAppControlsProps {
   app: hapi.release.Release;
+  update?: IChartUpdate;
```
prydonius: I'm just thinking that this could be confused with an "update" action, what about if we called this and the struct `updateInfo` and `IChartUpdateInfo`?

dashboard/src/components/AppView/AppControls/UpgradeButton.tsx
prydonius: We've used the `Redirect` container in places to avoid having to pass down a prop all the way. If we do prefer using `push`, I would prefer to just pass down `push` and not the whole `history` prop.

dashboard/src/components/AppView/AppControls/AppControls.tsx
```diff
@@ -30,21 +34,20 @@ class AppControls extends React.Component<IAppControlsProps, IAppControlsState>
...
         {!deleted && (
-          <button className="button" onClick={this.handleUpgradeClick}>
-            Upgrade
-          </button>
-        )}
-        {this.state.upgrade && (
-          <Redirect push={true} to={`/apps/ns/${namespace}/upgrade/${name}`} />
+          <UpgradeButton
+            updateVersion={(update && update.latestVersion) || ""}
+            upgradeURL={`/apps/ns/${namespace}/upgrade/${name}`}
```
prydonius: this is a bit weird to me, why not just pass the namespace and name and let `UpgradeButton` construct the upgrade URL? Why would you need to use an UpgradeButton with a different upgrade URL?

dashboard/src/components/AppView/ChartInfo.tsx
```diff
@@ -1,28 +1,65 @@
...
 
 class ChartInfo extends React.Component<IChartInfoProps> {
   public render() {
-    const { app } = this.props;
+    const { app, update } = this.props;
     const name = app.name;
     const metadata = app.chart && app.chart.metadata;
     const icon = metadata && metadata.icon;
     const iconSrc = icon ? icon : placeholder;
+    let updateStatusInfo: JSX.Element | null = null;
```
prydonius: consider moving this to a separate method, it avoids the need for initialising a variable as we can just use the return value

dashboard/src/components/AppView/AppView.tsx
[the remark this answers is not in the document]
prydonius: An alternative we haven't considered is to dispatch `getChartUpdates` from `getApp` to avoid the need for waiting for the app prop to be set in the component.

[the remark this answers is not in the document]
andresmgot: I am okay moving this to the getApp action to avoid trouble here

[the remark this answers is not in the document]
andresmgot: > An alternative we haven't considered is to dispatch getChartUpdates from getApp to avoid the need for waiting for the app prop to be set in the component.

I am reluctant to change the approach again. That should be valid as well but we will be mixing chart actions with app actions, which may makes sense if we move the updates information to the `release` object but since this is working and there is no a major issue let's go for this. If we find a different issue we can re-evaluate but for the moment I think it should be fine.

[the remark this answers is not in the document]
prydonius: Yeah I think it's just something we should consider for the future. I'll be honest, I'm not a big fan of using `componentDidUpdate` just for this thing, and continue using `componentWillReceiveProps` for everything else. It might have been better to convert the whole component to use `componentWillReceiveProps`. Nevertheless, I don't think it's worth reverting that now, so I'm fine merging this.

dashboard/src/actions/charts.ts
prydonius: ah, I also missed the await haha :P

dashboard/src/components/AppView/ChartInfo.test.tsx
```diff
@@ -1,27 +1,54 @@
...
+      metadata: {
+        appVersion: "0.0.1",
+        description: "test chart",
+        icon: "icon.png",
+        version: "1.0.0",
+      },
+    },
+    name: "foo",
+  } as hapi.release.Release,
+  updateCheck: { checked: false },
```
prydonius: I guess this should be `undefined` and also be `updateInfo`? Could also just drop it.

andresmgot: true, I missed this one

PR #932, opened 2019-01-18
Adds placeholder-style loading to the ServiceItems

prydonius:
The placeholder style loading is used in some webapps to show that text
content is loading. This is a better fit for the ServiceTables compared
to the LoadingSpinner.

The implementation adds a new `type` prop to the LoadingWrapper
component, which is used to determine which type of loader to render.
This could easily be extended to different types of loaders in the
future.

### DIFF
--- a/dashboard/src/components/LoadingWrapper/LoadingPlaceholder/LoadingPlaceholder.scss
+++ b/dashboard/src/components/LoadingWrapper/LoadingPlaceholder/LoadingPlaceholder.scss
@@ -1,0 +1,25 @@
+.LoadingPlaceholder {
+  height: 1em;
+  background-color: #f1f1f1;
+  position: relative;
+  &:after {
+    content: "";
+    position: absolute;
+    width: 100%;
+    height: 1em;
+    background: linear-gradient(
+      90deg,
+      rgba(0, 0, 0, 0),
+      rgba(255, 255, 255, 25%),
+      rgba(0, 0, 0, 0)
+    );
+    transform: translateX(-30%);
+    animation: loading 1.5s infinite;
+  }
+}
+
+@keyframes loading {
+  100% {
+    transform: translateX(100%);
+  }
+}
--- a/dashboard/src/components/LoadingWrapper/LoadingPlaceholder/LoadingPlaceholder.tsx
+++ b/dashboard/src/components/LoadingWrapper/LoadingPlaceholder/LoadingPlaceholder.tsx
@@ -1,0 +1,11 @@
+import * as React from "react";
+
+import "./LoadingPlaceholder.css";
+
+// Renders a placeholder loader style
+// Based on https://kyusuf.com/post/fake-it-til-you-make-it-css/
+const LoadingPlaceholder: React.SFC = props => {
+  return <div className="LoadingPlaceholder" />;
+};
+
+export default LoadingPlaceholder;
--- a/dashboard/src/components/LoadingWrapper/LoadingWrapper.tsx
+++ b/dashboard/src/components/LoadingWrapper/LoadingWrapper.tsx
@@ -1,7 +1,15 @@
 import * as React from "react";
+
 import LoadingSpinner from "../LoadingSpinner";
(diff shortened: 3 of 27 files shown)


dashboard/src/components/LoadingWrapper/LoadingWrapper.tsx
```diff
@@ -1,7 +1,15 @@
 import * as React from "react";
+
 import LoadingSpinner from "../LoadingSpinner";
+import LoadingPlaceholder from "./LoadingPlaceholder";
+
+export enum LoaderType {
+  Spinner,
```
migmartri: Fancy enum :)

dashboard/src/components/LoadingWrapper/LoadingWrapper.tsx
```diff
@@ -1,7 +1,15 @@
...
+  Placeholder,
+}
 
 export interface ILoadingWrapperProps {
+  type?: LoaderType;
```
migmartri: I am a little bit worried about using such an overloaded term `type`. It does not seem a reserved word in JS though.

dashboard/src/components/LoadingWrapper/LoadingWrapper.tsx
```diff
@@ -1,7 +1,15 @@
 import * as React from "react";
+
 import LoadingSpinner from "../LoadingSpinner";
+import LoadingPlaceholder from "./LoadingPlaceholder";
```
migmartri: Maybe this should be a little bit more specific?

`LoadingTextPlaceholder`?

dashboard/src/components/LoadingWrapper/LoadingPlaceholder/LoadingPlaceholder.scss
```diff
@@ -0,0 +1,25 @@
+.LoadingPlaceholder {
+  height: 1em;
+  background-color: #f1f1f1;
```
migmartri: Are you sure that this is the minimum amount of CSS that we need? Do we need the background color set here?

dashboard/src/components/LoadingWrapper/LoadingPlaceholder/LoadingPlaceholder.tsx
```diff
@@ -0,0 +1,11 @@
+import * as React from "react";
+
+import "./LoadingPlaceholder.css";
+
+// Renders a placeholder loader style
+// Based on https://kyusuf.com/post/fake-it-til-you-make-it-css/
+const LoadingPlaceholder: React.SFC = props => {
```
migmartri: You are missing tests, even if it is a simple snapshot one.

dashboard/src/components/LoadingWrapper/LoadingWrapper.tsx
prydonius: I feel like the "placeholder" is a style in itself, just search for "placeholder loader" on Google and you come up with exactly what this is: https://www.google.com/search?q=placeholder+loader&oq=placeholder+loader&aqs=chrome.0.69i59j69i60j0l4.3666j1j7&sourceid=chrome&ie=UTF-8.

dashboard/src/components/LoadingWrapper/LoadingPlaceholder/LoadingPlaceholder.scss
prydonius: yes, this is the grey background colour that the lighter one animates over

dashboard/src/components/LoadingWrapper/LoadingPlaceholder/LoadingPlaceholder.tsx
prydonius: I saw that the LoadingSpinner didn't have tests, but I can add them for this.

andresmgot: Not a lot to test but I guess it's something nice to have

PR #936, opened 2019-01-23
Improves loading resources for the AccessURLTable in AppView

prydonius:
- wraps AccessURLTable in a container to load resources from Redux
- AccessURLTable initiates fetches for Ingresses (loads faster than watching)
- Errors fetching Ingresses are shown

screen shot 2019-01-22 at 17 28 26

refs #882

### DIFF
--- a/dashboard/src/components/AppView/AccessURLTable/AccessURLTable.test.tsx
+++ b/dashboard/src/components/AppView/AccessURLTable/AccessURLTable.test.tsx
@@ -9,12 +9,21 @@ import { IIngressSpec, IResource, IServiceSpec, IServiceStatus } from "../../../
 import AccessURLItem from "./AccessURLItem";
 import AccessURLTable from "./AccessURLTable";
 
+describe("componentDidMount", () => {
+  it("fetches ingresses", () => {
+    const mock = jest.fn();
+    shallow(<AccessURLTable services={[]} ingresses={[]} fetchIngresses={mock} />);
+    expect(mock).toHaveBeenCalled();
+  });
+});
+
 context("when fetching ingresses or services", () => {
   itBehavesLike("aLoadingComponent", {
     component: AccessURLTable,
     props: {
       ingresses: [{ isFetching: true }],
       services: [],
+      fetchIngresses: jest.fn(),
     },
   });
   itBehavesLike("aLoadingComponent", {
@@ -22,12 +31,15 @@ context("when fetching ingresses or services", () => {
     props: {
       ingresses: [],
       services: [{ isFetching: true }],
+      fetchIngresses: jest.fn(),
     },
   });
 });
 
 it("renders a message if there are no services or ingresses", () => {
-  const wrapper = shallow(<AccessURLTable services={[]} ingresses={[]} />);
+  const wrapper = shallow(
+    <AccessURLTable services={[]} ingresses={[]} fetchIngresses={jest.fn()} />,
+  );
   expect(
     wrapper
       .find(LoadingWrapper)
@@ -45,6 +57,7 @@ it("renders a message if there are no services or ingresses", () => {
 context("when the app contain services", () => {
   it("should omit the Service Table if there are no public services", () => {
     const service = {
+      kind: "Service",
       metadata: {
         name: "foo",
       },
@@ -57,7 +70,9 @@ context("when the app contain services", () => {
       } as IServiceStatus,
     } as IResource;
     const services = [{ isFetching: false, item: service }];
-    const wrapper = shallow(<AccessURLTable services={services} ingresses={[]} />);
+    const wrapper = shallow(
+      <AccessURLTable services={services} ingresses={[]} fetchIngresses={jest.fn()} />,
+    );
     expect(
       wrapper
         .find(LoadingWrapper)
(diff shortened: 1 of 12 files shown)


dashboard/src/components/AppView/AccessURLTable/AccessURLTable.test.tsx
```diff
@@ -141,8 +164,29 @@ context("when the app contain services and ingresses", () => {
...
-    const wrapper = shallow(<AccessURLTable services={services} ingresses={ingresses} />);
+    const wrapper = shallow(
+      <AccessURLTable services={services} ingresses={ingresses} fetchIngresses={jest.fn()} />,
+    );
     expect(wrapper.find(AccessURLItem)).toExist();
     expect(wrapper).toMatchSnapshot();
   });
 });
+
+context("when the contains resources with errors", () => {
```
andresmgot: when the *table* contains?

dashboard/src/components/AppView/AccessURLTable/AccessURLTable.test.tsx
```diff
@@ -141,8 +164,29 @@ context("when the app contain services and ingresses", () => {
...
+    const wrapper = shallow(
+      <AccessURLTable services={services} ingresses={ingresses} fetchIngresses={jest.fn()} />,
+    );
+    expect(wrapper.find(AccessURLItem)).not.toExist();
+    expect(wrapper.find("table").text()).toContain("could not find Ingress");
+
+    // The Service error is not shown, as it is filtered out because without the
+    // resource we can't determine whether it is a public LoadBalancer Service
+    // or not. The Service error will be shown in the Services table anyway.
+    expect(wrapper.find("table").text()).not.toContain("could not find Service");
```
andresmgot: do we need this `assert` then? It seems that we are ensuring that a side effect exists. I would just remove this assert and leave the comment which explains the situation very well.

prydonius:
Yeah, it's tricky for the Services because we don't actually know if we need to render them or not (need to check if they are LoadBalancer Services). That said, it would be possible to split this up in the same way ServiceTable works, but perhaps something for future refactoring. Thanks!

dashboard/src/shared/ResourceRef.ts
```diff
@@ -39,6 +39,8 @@ class ResourceRef {
     switch (this.kind) {
```
migmartri: Nitpick but it might make sense to implement this as a map so extending it is simpler.

dashboard/src/containers/helpers/index.ts
```diff
@@ -0,0 +1,8 @@
+import * as _ from "lodash";
+
+import ResourceRef from "../../shared/ResourceRef";
+import { IKubeState } from "../../shared/types";
+
+export function filterByResourceRefs(refs: ResourceRef[], resources: IKubeState["items"]) {
```
migmartri: please comment what this does.

dashboard/src/containers/AccessURLTableContainer/AccessURLTableContainer.tsx
```diff
@@ -0,0 +1,45 @@
...
+interface IAccessURLTableContainerProps {
+  serviceRefs: ResourceRef[];
+  ingressRefs: ResourceRef[];
+}
+
+function mapStateToProps({ kube }: IStoreState, props: IAccessURLTableContainerProps) {
+  // Extract the Services and Ingresses form the Redux state using the keys for
+  // each ResourceRef.
+  return {
+    services: filterByResourceRefs(props.serviceRefs, kube.items),
```
migmartri: I like it!

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -39,6 +39,7 @@ interface IAppViewState {
   services: Array<IKubeItem<IResource>>;
   serviceRefs: ResourceRef[];
   ingresses: Array<IKubeItem<IResource>>;
```
migmartri: Do we need ingresses anymore? And services?

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -269,6 +266,7 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
           break;
         case "Ingress":
           result.ingresses.push(resource);
```
migmartri: Do we need to push the ingress here or we could wait for the proper one is pulled via the reference?

dashboard/src/components/AppView/AccessURLTable/AccessURLTable.test.tsx
prydonius: @andresmgot good point, you're right - I'll remove it!

dashboard/src/components/AppView/AppView.tsx
prydonius: Yep, we can remove these now.

dashboard/src/shared/ResourceRef.ts
prydonius: Ah good shout, I agree that would be much easier.

dashboard/src/components/AppView/AppView.tsx
prydonius: We can just remove these.

PR #940, opened 2019-01-24
Improves loading Deployment resources in AppView

prydonius:
- wraps DeploymentItems in containers to load resources from Redux
- DeploymentItem initiates fetches for Deployments (loads faster than watching)
- Errors fetching Deployments are shown

refs #882

### DIFF
--- a/dashboard/src/components/AppView/AppView.tsx
+++ b/dashboard/src/components/AppView/AppView.tsx
@@ -36,6 +36,7 @@ export interface IAppViewProps {
 
 interface IAppViewState {
   deployments: Array<IKubeItem<IResource>>;
+  deployRefs: ResourceRef[];
   serviceRefs: ResourceRef[];
   ingressRefs: ResourceRef[];
   // Other resources are not IKubeItems because
@@ -48,6 +49,7 @@ interface IAppViewState {
 
 interface IPartialAppViewState {
   deployments: Array<IKubeItem<IResource>>;
+  deployRefs: ResourceRef[];
   serviceRefs: ResourceRef[];
   ingressRefs: ResourceRef[];
   otherResources: IResource[];
@@ -75,6 +77,7 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
     manifest: [],
     deployments: [],
     ingressRefs: [],
+    deployRefs: [],
     otherResources: [],
     serviceRefs: [],
     secretNames: [],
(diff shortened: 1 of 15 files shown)


dashboard/src/containers/DeploymentItemContainer/DeploymentItemContainer.ts
```diff
@@ -0,0 +1,44 @@
...
+  dispatch: ThunkDispatch<IStoreState, null, Action>,
+  props: IDeploymentItemContainerProps,
+) {
+  const { deployRef } = props;
+  return {
+    getDeployment: () =>
+      dispatch(
+        actions.kube.getResource(
+          deployRef.apiVersion,
+          "deployments",
```
prydonius: rather than hardcoding this, we should use the `deployRef.resourcePlural()` and make that a public method - I'll make this change in a future PR, once the WebSockets have moved to redux as it applies to them too.

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -35,7 +35,7 @@ export interface IAppViewProps {
 }
 
 interface IAppViewState {
-  deployments: Array<IKubeItem<IResource>>;
+  deployRefs: ResourceRef[];
```
prydonius: I think in the future we can combine all the refs and then filter by kind, though not sure if that's important.

dashboard/src/containers/DeploymentItemContainer/DeploymentItemContainer.ts
andresmgot: why don't you add it now? It seems trivial since you already have `ResourceKindToPlural` but I may be missing something.

dashboard/src/containers/DeploymentStatusContainer/DeploymentStatusContainer.ts
```diff
@@ -0,0 +1,23 @@
...
+
+interface IDeploymentStatusContainerProps {
+  deployRefs: ResourceRef[];
+  info?: hapi.release.IInfo;
+}
+
+function mapStateToProps({ kube }: IStoreState, props: IDeploymentStatusContainerProps) {
+  const { deployRefs, info } = props;
+  return {
+    deployments: filterByResourceRefs(deployRefs, kube.items),
```
andresmgot: we should add a comment here explaining that we rely on DeploymentItems for fetching deployments

dashboard/src/containers/DeploymentItemContainer/DeploymentItemContainer.ts
prydonius: I don't want to change it for all of the containers in this PR, would rather do that in a separate smaller PR once this is landed.

dashboard/src/containers/DeploymentStatusContainer/DeploymentStatusContainer.ts
prydonius: good shout.

PR #941, opened 2019-01-25
get plural from ResourceRef rather than hardcoding

prydonius:
refs #882

### DIFF
--- a/dashboard/src/shared/ResourceRef.ts
+++ b/dashboard/src/shared/ResourceRef.ts
@@ -41,7 +41,7 @@ class ResourceRef {
   }
 
   // Gets the plural form of the resource Kind for use in the resource path
-  private resourcePlural() {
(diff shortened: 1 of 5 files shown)


dashboard/src/shared/ResourceRef.ts
```diff
@@ -41,7 +41,7 @@ class ResourceRef {
   }
 
   // Gets the plural form of the resource Kind for use in the resource path
-  private resourcePlural() {
+  public resourcePlural() {
```
andresmgot: since this is now public I would add a test for checking the supported plurals so we make sure we don't break the function

prydonius: sure, I decided not to since it was already being tested in `getResourceURL()`, but will add a more specific one here.

PR #942, opened 2019-01-26
refactor AppView Secret rendering

prydonius:
- refactors SecretsTable, SecretItem to make use of the new ResourceRefs
- SecretsTableContainer is removed and SecretItemContainer is linked to
  the Redux store
- SecretItem now handles fetching of the Secret resource
- SecretItemDatum is added to handle the presentational concerns of a
  single key/value pair in a Secret, avoiding the need for complicated
  state management in SecretItem


dashboard/src/containers/SecretItemContainer/SecretItemContainer.ts
```diff
@@ -0,0 +1,45 @@
...
+  dispatch: ThunkDispatch<IStoreState, null, Action>,
+  props: ISecretItemContainerProps,
+) {
+  const { secretRef } = props;
+  return {
+    getSecret: () =>
+      dispatch(
+        actions.kube.getResource(
+          secretRef.apiVersion,
+          "secrets",
```
prydonius: will need to change this once #941 lands

dashboard/src/components/AppView/SecretsTable/SecretItem.tsx
```diff
@@ -1,71 +1,70 @@
...
+        </td>
+      );
+    }
+    if (secret.item) {
+      const item = secret.item;
+      return (
+        <React.Fragment>
+          <td className="col-2">{item.type}</td>
+          <td className="col-7 padding-small">
+            {item.data ? (
```
andresmgot: we should also check if `data` is an empty object (that's why `isEmpty` was being used)

prydonius: good point, however I noticed that `data` is never defined in the API response from an empty Secret. Just in case though, I'll use `_.isEmpty` here and add a comment

andresmgot: I know, this is just in case that behavior changes. Thanks

andresmgot:
New commits LGTM as well

PR #943, opened 2019-01-28
Show updates available when listing applications

andresmgot:
Closes: #588 

Show in the list view if there is an update available for some of the apps:

screenshot from 2019-01-28 16-19-13

This PR also includes some (minimal) required changes in tiller-proxy to return Chart metadata when listing applications.


dashboard/src/components/AppList/AppList.tsx
```diff
@@ -30,19 +33,38 @@ class AppList extends React.Component<IAppListProps, IAppListState> {
...
     } = this.props;
     // refetch if new namespace or error removed due to location change
-    if (nextProps.namespace !== namespace || (error && !nextProps.apps.error)) {
-      fetchApps(nextProps.namespace, listingAll);
+    if (prevProps.namespace !== namespace || (error && !prevProps.apps.error)) {
+      fetchApps(namespace, listingAll);
     }
-    if (nextProps.filter !== filter) {
-      this.setState({ filter: nextProps.filter });
+    if (prevProps.filter !== filter) {
```
prydonius: We might actually want to refactor this to use the `defaultFilter` approach described in https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#alternative-2-reset-uncontrolled-component-with-an-instance-method, or try to make it fully-controlled.

It's things like this that worry me about changing this to `componentDidUpdate` in this PR, as we haven't fully thought-out what we should replace `componentWillReceiveProps` with and it's not relevant to this PR. I would consider reverting this to continue using `componentWillRecieveProps` and leave a comment to look into this later. It's not like we can't use `componentWillRecieveProps` so switching it to `componentDidUpdate` doesn't really give us anything, we still need to come back here at some point and figure out how to fix this.

dashboard/src/components/AppList/AppList.tsx
```diff
@@ -30,19 +33,38 @@ class AppList extends React.Component<IAppListProps, IAppListState> {
...
+        if (
+          !this.props.updatesInfo[
+            getChartUpdatesKey(
+              app.chartMetadata.name,
+              app.chartMetadata.version,
+              app.chartMetadata.appVersion,
+            )
+          ]
+        ) {
+          this.props.getChartUpdates(
```
prydonius: these dependent calls are difficult inside a component, I think it would be better to create a new action, such as `fetchAppsWithUpdatesInfo` that is able to dispatch and await on the apps and then dispatch the `getChartUpdates` action. This approach has the benefit of keeping the actions decoupled (`fetchApps` and `getChartUpdates` are still separate, but `fetchAppsWithUpdatesInfo` is able to handle the dependency between them) and avoid the need for this complicated, potentially buggy (as per issues stated in https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html), logic in the component.

dashboard/src/components/AppList/AppList.tsx
```diff
@@ -126,7 +148,17 @@ class AppList extends React.Component<IAppListProps, IAppListState> {
           <div>
             <CardGrid>
               {filteredItems.map(r => {
-                return <AppListItem key={r.releaseName} app={r} />;
+                const appVersion = r.chartMetadata.appVersion || "";
+                const updateKey = `${r.chartMetadata.name}/${
```
prydonius: shouldn't `getChartUpdatesKey` be used here?

dashboard/src/components/AppList/AppListItem.test.tsx
```diff
@@ -18,6 +18,7 @@ it("renders an app item", () => {
           chart: "myapp",
         } as IAppOverview
       }
+      updateInfo={undefined}
```
prydonius: isn't this unnecessary?

arapulido:
@earnubs Can you have a look to the solution from a UI point of view, and let us know what you think, please? Thanks!

andresmgot:
> I'm not sure about the banner, it looks a little strange to me. Could you provide more context around why you decided on that?

I checked with @Angelmmiguel, it was his suggestions and I think that looks better than the other option I had. This solution maintains the size of the cards and it is very prompting to act. The other option was using a different pill to show the version but then the size of the cards changed:

screenshot from 2019-01-25 17-09-43

I also thought on adding the new version to the current info (something like `wordpress v5.1.1 - v5.1.2 available`) but that text was too long.

dashboard/src/components/AppList/AppList.tsx
prydonius: @andresmgot yes, we need to remove it when we update React, however we're still currently using React 16, so we can still use this. My main concern is not falling into the trap of thinking this is fixed when we should actually go back and refactor this. That said, it's not a huge issue to change this to `componentDidUpdate` as long as we add a TODO here.

dashboard/src/actions/apps.ts
```diff
@@ -86,6 +146,18 @@ export function fetchApps(
     try {
       const apps = await App.listApps(ns, all);
       dispatch(receiveAppList(apps));
+      const appsWithUpdateInfo = await Promise.all(
+        apps.map(
+          async (app): Promise<IAppOverviewWithUpdateInfo> => {
+            const name = app.chartMetadata.name;
+            const currentVersion = app.chartMetadata.version;
+            const appVersion = app.chartMetadata.appVersion;
+            const updateInfo = await getChartUpdates(name, currentVersion, appVersion);
```
prydonius: It's a much better idea to have `getChartUpdates` be an action that we simply dispatch and no longer care about. The `getChartUpdates` action can then be responsible for updating the entry in the `IAppState["listOverview"]` and `IAppState["selected"]` if it matches (we may want to change listOverview to the byID pattern to make it easier to reference). This way, we don't need to create a bunch of Promises and await on `Promise.all`.

This is important, because it will give us a separate Redux action for each returned app update rather than using the same one (e.g. we will get a `RECEIVE_APP_UPDATE_INFO` per app update retrieved). It's also not performant to update the whole array in one go, rather than items in the array. This will likely force a React render all the way up the tree. Also, `getChartUpdates` might need to control it's own lifecycle, e.g. if we needed a separate `isFetching` for the updatesInfo, we would want to dispatch the states from there.

We might want to do this in a separate PR, after this is merged, given there might be a lot of changes if we need to change this to the byID pattern.

dashboard/src/actions/apps.ts
```diff
@@ -58,6 +67,57 @@ export function getApp(
   };
 }
 
+async function getChartUpdates(name: string, currentVersion: string, appVersion: string) {
```
prydonius: I would consider renaming this to `getAppUpdates` now that we've moved it here.

dashboard/src/components/AppList/AppList.tsx
```diff
@@ -25,21 +25,21 @@ interface IAppListState {
...
       apps: { error, listingAll },
-      fetchApps,
+      fetchAppsWithUpdatesInfo,
       filter,
       namespace,
     } = this.props;
     // refetch if new namespace or error removed due to location change
     if (nextProps.namespace !== namespace || (error && !nextProps.apps.error)) {
-      fetchApps(nextProps.namespace, listingAll);
+      fetchAppsWithUpdatesInfo(nextProps.namespace, listingAll);
```
prydonius: this is so much nicer now! 😃

dashboard/src/shared/types.ts
```diff
@@ -187,15 +184,23 @@ export interface IDeploymentStatus {
   availableReplicas: number;
 }
 
+export interface IReleaseWithUpdateInfo extends hapi.release.Release {
```
prydonius: I would just call this our `IRelease`, the `updateInfo` is optional anyway.

dashboard/src/shared/types.ts
```diff
@@ -187,15 +184,23 @@ export interface IDeploymentStatus {
...
+  updateInfo?: IChartUpdateInfo;
+}
+
+export interface IAppOverviewWithUpdateInfo extends IAppOverview {
```
prydonius: as above, I don't think this is necessary over just having the property as an optional one in `IAppOverview`.

dashboard/src/components/AppView/ChartInfo.tsx
```diff
@@ -49,11 +47,11 @@ class ChartInfo extends React.Component<IChartInfoProps> {
   }
 
   private updateStatusInfo() {
-    const { app, updateInfo } = this.props;
+    const { app } = this.props;
     // If update is not set yet we cannot know if there is
     // an update available or not
-    if (updateInfo) {
-      if (!updateInfo.latestVersion) {
+    if (app.updateInfo) {
```
prydonius: I really like having it as part of the app object, it feels a lot cleaner 👍

dashboard/src/actions/apps.ts
```diff
@@ -74,7 +134,7 @@ export function deleteApp(
   };
 }
 
-export function fetchApps(
+export function fetchAppsWithUpdatesInfo(
```
prydonius: Although we don't use it separately, there is now no way to just `fetchApps` by themselves. I would keep the `fetchApps` action and change it to return `dispatch(receiveAppList(apps));`, and then dispatch `fetchApps` from `fetchAppsWithUpdatesInfo` and await on it. See the composition examples in https://github.com/reduxjs/redux-thunk#composition, the actions are designed to return the data that is also dispatched, so that other actions can be built on top of them.

dashboard/src/actions/apps.ts
andresmgot: I get your point. I think this only applies to `IAppState["listOverview"]` since when getting a single app we do it one by one. Since we already have an issue for refactoring the byID schedule.

If we avoid making `getChartUpdates` an action, the changes to update info per app would be minimal and will fit in this PR.

dashboard/src/reducers/apps.ts
```diff
@@ -30,6 +30,14 @@ const appsReducer = (
       return { ...state, isFetching: true, listingAll: action.payload };
     case getType(actions.apps.receiveAppList):
       return { ...state, isFetching: false, listOverview: action.payload };
+    case getType(actions.apps.updateAppListItem):
+      if (state.listOverview) {
```
prydonius: for also updating `selected`, we have the option here to check if `selected.releaseName === payload.releaseName` or we can pass down a parameter in the payload to know if this is coming from a `getAppWithUpdateInfo` or a `fetchAppsWithUpdateInfo`.

dashboard/src/reducers/apps.ts
```diff
@@ -30,6 +30,14 @@ const appsReducer = (
...
+        const appOverviewIndex = state.listOverview.findIndex(
+          a => a.releaseName === action.payload.releaseName,
+        );
+        state.listOverview[appOverviewIndex] = action.payload;
```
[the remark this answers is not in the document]
andresmgot: I think it's easier (and more readable) to copy the state into a different variable and mutate that variable

dashboard/src/reducers/apps.ts
```diff
@@ -30,6 +31,22 @@ const appsReducer = (
       return { ...state, isFetching: true, listingAll: action.payload };
     case getType(actions.apps.receiveAppList):
       return { ...state, isFetching: false, listOverview: action.payload };
+    case getType(actions.apps.receiveAppUpdateInfo):
+      const stateCopy = cloneDeep(state);
```
prydonius: Deep cloning is expensive, and seems pretty unnecessary here __just__ to update one app's `updateInfo` field. See https://medium.freecodecamp.org/handling-state-in-react-four-immutable-approaches-to-consider-d1f5c00249d5. I would try to use the previously suggested approaches (https://gist.github.com/gorangajic/e902c2ee994260b3348d#update-array-with-objects-using-arrayprototypeslice), but we could discuss whether it's worth adding an immutability helper.

@Angelmmiguel @earnubs any thoughts on this?

dashboard/src/actions/apps.test.tsx
```diff
@@ -51,11 +51,75 @@ describe("fetches applications", () => {
       { type: getType(actions.apps.listApps), payload: false },
       { type: getType(actions.apps.receiveAppList), payload: [] },
     ];
-    return store.dispatch(actions.apps.fetchApps("default", false)).then(() => {
+    return store.dispatch(actions.apps.fetchAppsWithUpdateInfo("default", false)).then(() => {
```
prydonius: why don't we use `await` here and other places in this file?

dashboard/src/actions/apps.test.tsx
```diff
@@ -51,11 +51,75 @@ describe("fetches applications", () => {
...
+        { type: getType(actions.apps.receiveAppList), payload: appsResponse },
+        {
+          type: getType(actions.apps.receiveAppUpdateInfo),
+          payload: {
+            releaseName: "foobar",
+            updateInfo: { latestVersion: "1.1.0", repository: { name: "bar" } },
+          },
+        },
+      ];
+      return store.dispatch(actions.apps.fetchAppsWithUpdateInfo("default", false)).then(() => {
```
prydonius: await?

dashboard/src/actions/apps.test.tsx
```diff
@@ -51,11 +51,75 @@ describe("fetches applications", () => {
...
+        { type: getType(actions.apps.receiveAppList), payload: appsResponse },
+        {
+          type: getType(actions.apps.receiveAppUpdateInfo),
+          payload: {
+            releaseName: "foobar",
+            updateInfo: { latestVersion: "", repository: { name: "", url: "" } },
+          },
+        },
+      ];
+      return store.dispatch(actions.apps.fetchAppsWithUpdateInfo("default", false)).then(() => {
```
prydonius: await?

dashboard/src/actions/apps.ts
```diff
@@ -19,6 +28,10 @@ export const receiveAppList = createAction("RECEIVE_APP_LIST", resolve => {
   return (apps: IAppOverview[]) => resolve(apps);
 });
 
+export const receiveAppUpdateInfo = createAction("RECEIVE_APP_UPDATE_INFO", resolve => {
+  return (partialApp: { releaseName: string; updateInfo: IChartUpdateInfo }) => resolve(partialApp);
```
prydonius: `partialApp` is confusing, why not just `updateInfo`, or `payload` to keep it simple?

dashboard/src/actions/apps.ts
```diff
@@ -58,6 +72,57 @@ export function getApp(
...
+}
+
+export function getAppWithUpdateInfo(
+  releaseName: string,
+  namespace: string,
+): ThunkAction<Promise<void>, IStoreState, null, AppsAction> {
+  return async dispatch => {
+    dispatch(requestApps());
+    try {
+      const app = await App.getRelease(namespace, releaseName);
```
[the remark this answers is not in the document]
prydonius: Are you still planning to apply this suggestion?

dashboard/src/actions/apps.ts
```diff
@@ -74,10 +148,11 @@ export function deleteApp(
   };
 }
 
+// fetchApps returns a list apps for other actions to compose on top of it
```
prydonius: list *of* apps?

dashboard/src/actions/apps.ts
```diff
@@ -86,12 +161,33 @@ export function fetchApps(
...
 }
 
+export function fetchAppsWithUpdateInfo(
+  ns?: string,
+  all: boolean = false,
+): ThunkAction<Promise<void>, IStoreState, null, AppsAction> {
+  return async dispatch => {
+    const apps = await dispatch(fetchApps(ns, all));
+    apps.forEach(app =>
+      dispatch(
```
prydonius: 👍

dashboard/src/components/AppView/AppControls/AppControls.tsx
```diff
@@ -47,7 +45,7 @@ class AppControls extends React.Component<IAppControlsProps, IAppControlsState>
         {/* If the app has been deleted hide the upgrade button */}
         {!deleted && (
           <UpgradeButton
-            updateVersion={(updateInfo && updateInfo.latestVersion) || ""}
+            updateVersion={(app.updateInfo && app.updateInfo.latestVersion) || ""}
```
prydonius: empty string? why not undefined? `UpgradeButton` can handle whether it is set or not

dashboard/src/reducers/apps.ts
prydonius: Just did a bit more digging on this, Redux also has some excellent documentation on different approaches to dealing with this problem: https://redux.js.org/recipes/structuring-reducers/immutable-update-patterns#inserting-and-removing-items-in-arrays. For example, one option for the `listOverview` is use `Array.map` to update the one we want: https://redux.js.org/recipes/structuring-reducers/immutable-update-patterns#updating-an-item-in-an-array.

Another way to help make this a little clearer would be to add a new reducer that just acts on the `listOverview`, and perhaps another one that acts on `selected`.

andresmgot: okay, honestly I don't think that using `cloneDeep` would create a bottleneck but it's okay for me to sacrifice the readability/simplicity.

Note, due to the type of `state.selected` I am not able to do:

```
selected = {...state.selected, updateInfo};
```

It fails with:

```
Type 'IRelease | { updateInfo: IChartUpdateInfo; name: string; info?: IInfo | null | undefined; chart?: IChart | null | undefined; config?: IConfig | null | undefined; manifest: string; hooks: IHook[]; version: number; namespace: string; } | undefined' is not assignable to type 'IRelease | undefined'.
  Property 'toJSON' is missing in type '{ updateInfo: IChartUpdateInfo; name: string; info?: IInfo | null | undefined; chart?: IChart | null | undefined; config?: IConfig | null | undefined; manifest: string; hooks: IHook[]; version: number; namespace: string; }' but required in type 'IRelease'.
```

I guess the operator `...` does not play good with methods of a class

dashboard/src/actions/apps.ts
prydonius: Hmm I don't think I'm overthinking it. To me, `partialApp` sounds like it would have the type of `Partial<IAppOverview>`. In this we have a specific struct for the payload that includes just the updateInfo and releaseName. Thanks for changing it.

dashboard/src/actions/apps.ts
```diff
@@ -46,12 +60,74 @@ export type AppsAction = ActionType<typeof allActions[number]>;
...
+): ThunkAction<Promise<hapi.release.Release | undefined>, IStoreState, null, AppsAction> {
   return async dispatch => {
     dispatch(requestApps());
     try {
       const app = await App.getRelease(namespace, releaseName);
       dispatch(selectApp(app));
+      return app;
+    } catch (e) {
+      dispatch(errorApps(e));
+      return;
```
prydonius: is this return necessary?

dashboard/src/actions/apps.ts
```diff
@@ -46,12 +60,74 @@ export type AppsAction = ActionType<typeof allActions[number]>;
 export function getApp(
   releaseName: string,
   namespace: string,
-): ThunkAction<Promise<void>, IStoreState, null, AppsAction> {
+): ThunkAction<Promise<hapi.release.Release | undefined>, IStoreState, null, AppsAction> {
```
prydonius: should this return an `IRelease`?

dashboard/src/reducers/apps.ts
prydonius: Perhaps not now or in your testing, but what happens on a cluster where there are a lot of apps? What happens when we later expand AppOverview to hold more data. It seems like a bad practice and the alternatives are not necessarily that much worse.

earnubs:
How important do we think it is to know that an application has an update?

dashboard/src/reducers/apps.ts
andresmgot: Yes, I know, thanks for the suggestion!

earnubs:
I will try and come back with some feedback by EOD.

dashboard/src/actions/apps.ts
andresmgot: It's necessary to avoid a linter error: `Not all the paths return a value`

dashboard/src/actions/apps.ts
andresmgot: Not in the case it throws an error.

earnubs:
> I will try and come back with some feedback by EOD.

I'm not going to have anything for today. Suggest not blocking this PR and I'll pick up the notification UI as a new branch.

dashboard/src/actions/apps.ts
prydonius: What I meant is, shouldn't `App.getRelease` return an `IRelease` so that this can return an `IRelease`?

andresmgot: ah, I don't think so because that would mean that the returned object would contain the key `updateInfo` which will be always empty. To get an `IRelease` the method to be called would be `getAppWithUpdateInfo` (even though that method is now written to not return anything).

arapulido:
Sounds good. Let's not block this on design, let's merge it, and @earnubs can work later on if he thinks the design needs improvement

PR #951, opened 2019-02-01
moving websockets to redux store

prydonius:
todo:
- add event handler for sockets to push data to store
- socket close actions


dashboard/src/actions/kube.tsx
```diff
@@ -15,25 +15,45 @@ export const receiveResourceError = createAction("RECEIVE_RESOURCE_ERROR", resol
...
+  return async (dispatch, getState) => {
+    const key = ref.getResourceURL();
+
+    // Multiple requests for a resource may happen at the same time (e.g. when
+    // loading different sections of a view). This guard attempts to prevent
+    // multiple requests for a resource that is already being fetched. Since
+    // this action is asynchronous, it's possible that multiple requests are let
+    // through, but this is not a huge concern.
+    const existing = getState().kube.items[key];
+    if (existing && existing.isFetching) {
```
andresmgot: I think we should only check if `existing` is set. If multiple requests happens sequentially we don't want to repeat the request either, isn't it? If we want to get updates fro a resource it should use a socket?

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -118,7 +116,7 @@ class AppView extends React.Component<IAppViewProps, IAppViewState> {
     }
     if (nextProps.error) {
       // close any existing sockets
-      this.closeSockets();
+      // this.closeSockets();
```
andresmgot: where do you plan to close sockets? I don't see a clear "easy" way, maybe this component is the best place to do it.

dashboard/src/shared/ResourceRef.ts
```diff
@@ -41,6 +41,22 @@ class ResourceRef {
     return Kube.getResourceURL(this.apiVersion, this.resourcePlural(), this.namespace, this.name);
   }
 
+  public watchResourceURL() {
```
andresmgot: this methods are no longer related to a "resource reference", since they are wrapping `Kube` methods, should we just use those? Maybe is `Kube` who should call methods from here when it needs to resolve URLs or plurals

dashboard/src/actions/kube.tsx
prydonius: I considered it, but I think it's helpful to refresh the data on a new page/component load, and this is the same as what happens today. I think the main concern is not repeating requests around the same time, not preventing requests that we might want to happen at different times.

dashboard/src/components/AppView/AppView.tsx
prydonius: I think each component that opens a watch in componentDidMount should close the socket in componentWillUnmount. If this a close action gets executed multiple times by different components, the reducer will just ignore if the socket is already closed/non-existent.

This obviously has the caveat you mentioned in our call that a socket might be closed by a component but still in use by another component. This is not a problem today, but one way to fix this would be to use a counter to track uses of a socket and only close it once it reaches 0. This is something that we can add in the future when necessary.

dashboard/src/shared/ResourceRef.ts
prydonius: Yeah this is why I was thinking about moving around the methods. However I ultimately decided against merging these as we will need a way to make arbitrary requests to the Kubernetes API and Kube.ts allows us to do that right now. These are kind of just convenience wrappers of those calls to make it easier if you are using ResourceRef. That's the thinking, but if you have any other ideas about how to structure this, I'm certainly open to it!

I'm not sure what you mean by they are not related to the ResourceRef, they just handle taking the parameters in the ResourceRef object and passing them appropriately to Kube.ts to avoid having to type out each parameter every time.

dashboard/src/actions/kube.tsx
andresmgot: Yes, I am thinking about requests around the same time. Depending on the performance both components (e.g. ServiceTable and AccessURLTable) may request a service with enough delay for the first request to be completely fetched before the second request. The problem in that case is that the second component will show the loading icon even if the resource has just been collected.

I think that the UX would be a bit better if the second time you visit an app the resources are already loaded. If there are updates the sockets will handle those changes eventually.

dashboard/src/components/AppView/AppView.tsx
andresmgot: :+1:  sounds good

dashboard/src/shared/ResourceRef.ts
andresmgot: My concern about the naming is small but what I mean is that if I read `reference.get()` I assume that it will return the `reference` object, not the resource, that's why I think this could be called only `Resource` since `Resource.get` or `Resource.watch` really returns the resource.

Apart from that, I think that for make it more clear that `Kube` is the low level library for this (and AFAIK now it should not be used directly) we can add it inside this component: `shared/Resource/Kube`.

It's also minor but I would move the `resourcePlural` method to `Kube`, it shouldn't be use externally, isn't it?

dashboard/src/actions/kube.tsx
prydonius: Okay, I see what you're saying. I think that's fairly unlikely to happen, given how fast dispatching actions are compared to API requests. We should keep an eye on this though in case that does happen.

I understand your point about UX, but I don't think it's a big concern. Fetching the resources outside of the WebSockets has already improved the loading speed substantially. If we were to do this, we would need to either create another action of have a `skipExisting` parameter in this action as not all resources will open a WebSocket to get the latest data (e.g. Secrets).

dashboard/src/shared/ResourceRef.ts
prydonius: > My concern about the naming is small but what I mean is that if I read reference.get() I assume that it will return the reference object, not the resource, that's why I think this could be called only Resource since Resource.get or Resource.watch really returns the resource.

I think I understand what you're saying here, but the problem is we don't have a `Resource` type here, we have a reference that is used to get the resource. We could have a `Resource` type that holds partial data about the GVK, namespace and name and implement `get()` to fetch the resource from the server, but it will be a huge change to use everywhere and I'm not sure that is valuable over what we currently have. IMO, `ResourceRef.getResource()` is fairly clear that it is to get the referenced resource.

> Apart from that, I think that for make it more clear that Kube is the low level library for this (and AFAIK now it should not be used directly) we can add it inside this component: shared/Resource/Kube.

I disagree that Kube should be a low-level library that is not directly used, as previously stated there will be times when we need to use Kube directly (e.g. to make a List call), when we don't have a ResourceRef. I think we will also want to eventually add support for POST/PUT/DELETE so we can use Kube.ts as the library for the calls made from AppRepository.ts, for example, to help us reduce the duplication.

> It's also minor but I would move the resourcePlural method to Kube since it shouldn't be use externally, isn't it?

Good point, I do think this might make sense to move to Kube, especially since we might need to use it for List and other calls. Assuming ResourceRef is going to call `resourcePlural` though, it will need to be public.

prydonius:
@andresmgot this is now updated and ready to review. I have force-pushed changes as I undid quite a few changes to make sure we only open sockets in redux for Services right now.

dashboard/src/reducers/kube.test.ts
```diff
@@ -44,5 +57,84 @@ describe("authReducer", () => {
...
+        const mock = jest.fn();
+        const newState = kubeReducer(undefined, {
+          type: actionTypes.openWatchResource,
+          payload: {
+            ref,
+            handler: mock,
+          },
+        });
+        const socket = newState.sockets[ref.watchResourceURL()];
+        const listener = (socket as any).listeners.message[0];
```
andresmgot: this is a bit weird, how is a `listener` a message of the listener? I guess that's how the library has it defined but can you add a comment explaining this?

dashboard/src/reducers/kube.ts
```diff
@@ -28,6 +30,37 @@ const kubeReducer = (
...
+          [key]: socket,
+        },
+      };
+    // TODO(adnan): this won't handle cases where one component closes a socket
+    // another one is using. Whilst not a problem today, a reference counter
+    // approach could be used here to enable this in the future.
+    case getType(actions.kube.closeWatchResource):
+      key = action.payload.watchResourceURL();
+      const { sockets } = state;
+      const { [key]: foundSocket, ...restSockets } = sockets;
```
andresmgot: :ok_hand: that's neat

Minor thing, I would change `restSockets` for `otherSockets` just to make it clear it's not a RESTful thing

dashboard/src/shared/Kube.test.ts
```diff
@@ -53,8 +84,19 @@ describe("App", () => {
...
+  });
+
+  describe("watchResource", () => {
+    it("should open a socket", async () => {
+      const socket = Kube.watchResource("v1", "pods", "default", "foo");
+      expect(socket.url).toBe(
+        `${WebSocketAPIBase}${APIBase}/api/v1/namespaces/default/pods?watch=true&fieldSelector=metadata.name%3Dfoo`,
       );
+      // it's a mock socket, so doesn't actually need to be closed
+      socket.close();
```
andresmgot: where are you mocking WebSocket? I cannot find that

dashboard/src/shared/ResourceRef.ts
andresmgot: > I think I understand what you're saying here, but the problem is we don't have a `Resource` type here, we have a reference that is used to get the resource. We could have a `Resource` type that holds partial data about the GVK, namespace and name and implement `get()` to fetch the resource from the server, but it will be a huge change to use everywhere and I'm not sure that is valuable over what we currently have. IMO, `ResourceRef.getResource()` is fairly clear that it is to get the referenced resource.

What I meant was simply renaming `ResourceRef` for `Resource` so `s/ResourceRef.getResource/Resource.get/`. It's not that `ResourceRef.getResource` is not clear what's doing it's just that I think the naming is a bit more natural the other way. In any case if you have a strong preference you can keep this naming.

> I disagree that Kube should be a low-level library that is not directly used, as previously stated there will be times when we need to use Kube directly (e.g. to make a List call), when we don't have a ResourceRef. I think we will also want to eventually add support for POST/PUT/DELETE so we can use Kube.ts as the library for the calls made from AppRepository.ts, for example, to help us reduce the duplication.

I thought that `Kube` was no longer used externally, if it's still used we should keep it accessible. My concern is how we make clear that when creating a new component (for example something like `JobsTable`) it should use this `ResourceRef` instead of directly calling the methods of `Kube`.

> Good point, I do think this might make sense to move to Kube, especially since we might need to use it for List and other calls. Assuming ResourceRef is going to call `resourcePlural` though, it will need to be public.

:+1:

dashboard/src/actions/kube.tsx
prydonius: Just to clarify, you're okay leaving this as-is for now? I'm happy to revisit caching this if we see performance problems, but I would prefer we don't prematurely optimise here.

dashboard/src/reducers/kube.test.ts
prydonius: yes good point, as far as I can tell, `listener` comes from `mock-socket` (https://github.com/thoov/mock-socket/blob/master/index.d.ts#L7), but TS doesn't use the types from that library (I think it just uses the default WebSocket type) and I can't figure out how to use it. I'll add a comment about that.

dashboard/src/reducers/kube.ts
prydonius: indeed!

I don't have a preference so happy to change it, but just for your reference, the term rest here comes from this being the rest operator. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#Rest_syntax_(parameters) and https://underscorejs.org/#rest.

dashboard/src/shared/Kube.test.ts
prydonius: It is mocked in `setupTests`: https://github.com/kubeapps/kubeapps/blob/f21c9840458fb9eea8e1ac7c9b7d866206e99309/dashboard/src/setupTests.ts#L6. This is necessary because `WebSocket` won't be defined in the test environment without it. So all WebSockets must be mocks.

dashboard/src/shared/ResourceRef.ts
prydonius: > What I meant was simply renaming `ResourceRef` for `Resource` so `s/ResourceRef.getResource/Resource.get/`. It's not that `ResourceRef.getResource` is not clear what's doing it's just that I think the naming is a bit more natural the other way. In any case if you have a strong preference you can keep this naming.

The problem with this is that `Resource` should have the schema of `IResource`. We could do that, but it'd be changing a lot of things if we wanted to use `Resource` everywhere instead of `IResource`. I like having a smaller structure of `ResourceRef` being used to get a full `IResource`, but I do also agree that having a single `Resource` object that might be able to get itself could be a cleaner approach overall. That said, I think it's out of the scope of this PR.

> I thought that `Kube` was no longer used externally, if it's still used we should keep it accessible. My concern is how we make clear that when creating a new component (for example something like `JobsTable`) it should use this `ResourceRef` instead of directly calling the methods of `Kube`.

It isn't currently used outside of `ResourceRef`, but my point is that it could easily be due to `ResourceRef`'s limitations (i.e. only pointing to a single resource). You have a good point about making it more clear that we should use ResourceRef instead where we can. I'll add a comment to Kube.ts to reflect this.

dashboard/src/actions/kube.tsx
andresmgot: Yes, I mean that you can go ahead with this code.

dashboard/src/shared/ResourceRef.ts
andresmgot: > The problem with this is that `Resource` should have the schema of `IResource`. We could do that, but it'd be changing a lot of things if we wanted to use `Resource` everywhere instead of `IResource`.

That's not exactly what I was thinking. I'd be okay having the interface `IResource` and the class `Resource`, needing the constructor of `Resource` an `IResource` (just like the `ResourceRef` now).

> It isn't currently used outside of `ResourceRef`, but my point is that it could easily be due to `ResourceRef`'s limitations (i.e. only pointing to a single resource)

I thought that in that case we should extend `ResourceRef` instead of using directly `Kube` to support list/other resources, isn't it?

dashboard/src/reducers/kube.ts
andresmgot: okay, it's just the first thing I thought when I read `restSockets`

dashboard/src/shared/ResourceRef.ts
prydonius: As discussed in the standup, I don't have a strong preference either way on this, I would say that a "Resource" class feels like it should have the properties of a resource, and not just be used to get an "IResource" - but perhaps this comes from my Rails models days ;).

> I thought that in that case we should extend `ResourceRef` instead of using directly `Kube` to support list/other resources, isn't it?

Potentially, but maybe it's simpler to keep ResourceRef to just reference a single resource? I'm not sure, but I'll add the comment to Kube.ts for now and we can decide when we need this.

prydonius:
@andresmgot addressed all your feedback, thanks - PTAL!

PR #952, opened 2019-02-01
Fix User-Agent header for Tiller-proxy

andresmgot:
Since v1.1.0 the `netClient` is created in the `chartUtils` pkg instead of the main one. This caused that the User-Agent that we were setting in the main package was not being used.

### DIFF
--- a/cmd/tiller-proxy/main.go
+++ b/cmd/tiller-proxy/main.go
@@ -45,7 +45,6 @@ var (
 	settings    environment.EnvSettings
 	proxy       *tillerProxy.Proxy
 	kubeClient  kubernetes.Interface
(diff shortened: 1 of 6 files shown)


andresmgot:
> is there a test we can do in the main package to ensure this doesn't accidentally get dropped in the other package again?

So the problem was that there was a test in the main package that checked that the "global" variable there could be used to set the proper proxy-agent. Since Golang doesn't complain about unused global variables we didn't notice that it was not being used. Now there is a test that check that a header is set when using the `GetChart` method. That should be enough for us to notice if something is wrong.

prydonius:
Sounds good, thanks for the explanation

PR #955, opened 2019-02-04
Allow to use a custom CA certificate or Authorization header in the chart

### LINKED ISSUE #919: configure custom CA cert and authorization header for initialRepos in chart
image
private certificate can be set for repository by the UI but not by the value.xml file.

andresmgot:
Fixes  #919 

Add the possibility of adding custom CA certificates and Authorization header to the initial repositories. There is a caveat that if the app repository is deleted with the Dashboard UI, the associated secrets won't be deleted until the chart is deleted.

cc/ @obeyler


chart/kubeapps/templates/NOTES.txt
```diff
@@ -8,6 +8,21 @@ Kubeapps can be accessed via port {{ .Values.frontend.service.port }} on the fol
 
    {{ template "kubeapps.fullname" . }}.{{ .Release.Namespace }}.svc.cluster.local
 
+{{- $orphanSecrets := false -}}
+{{- range .Values.apprepository.initialRepos }}
+{{- if or .caCert .authorizationHeader }}
+{{- $orphanSecrets = true -}}
```
andresmgot: this gets rejected by the template-test but I am not sure why. This is used to print the caveat only if there is a `.caCert` or `.authorizationHeader` set.

prydonius: So this works with `helm install` but not with `helm template`?

chart/kubeapps/templates/apprepositories-secret.yaml
```diff
@@ -0,0 +1,23 @@
+{{- range .Values.apprepository.initialRepos }}
+{{- if or .caCert .authorizationHeader }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ template "kubeapps.fullname" $ }}-{{ .name }}
```
prydonius: The name we create for the Secrets follows the format: `apprepo- {{ .name }}-secrets`. Please be consistent with that format: https://github.com/kubeapps/kubeapps/blob/4b21ca69f9c5817815f793b8ff6438128145c148/dashboard/src/actions/repos.ts#L138

chart/kubeapps/templates/apprepositories-secret.yaml
```diff
@@ -0,0 +1,23 @@
...
+  labels:
+    app: {{ template "kubeapps.name" $ }}
+    chart: {{ template "kubeapps.chart" $ }}
+    release: {{ $.Release.Name }}
+    heritage: {{ $.Release.Service }}
+data:
+{{- if .caCert }}
+  ca.crt: |-
+{{- .caCert | b64enc | nindent 4 }}
```
prydonius: the `nindent` shouldn't be needed if the `{{-` was `{{` right?

chart/kubeapps/templates/apprepositories-secret.yaml
```diff
@@ -0,0 +1,23 @@
...
+{{- end }}
+{{- if .authorizationHeader }}
+  authorizationHeader: |-
+{{- .authorizationHeader | b64enc | nindent 4 }}
+{{- end }}
+---
+{{- end }}
+{{ end -}}
```
prydonius: newline?

chart/kubeapps/values.yaml
```diff
@@ -111,6 +111,15 @@ apprepository:
     url: https://svc-catalog-charts.storage.googleapis.com
   - name: bitnami
     url: https://charts.bitnami.com/bitnami
+  # Additional repositories
+  # - name: chartmuseum
+  #   url: https://chartmuseum.default:8080
+  #   # Specify an Authorization Header if you are using an authentication method
+  #   authorizationHeader: "Bearer xrxNC..."
+  #   # If you're providing your own certificates, please use this to add the certificates as secrets
```
prydonius: full stop? I initially thought the next sentence was continuing on from this

chart/kubeapps/values.yaml
```diff
@@ -111,6 +111,15 @@ apprepository:
...
+  #   # key and certificate should start with -----BEGIN CERTIFICATE----- or
```
prydonius: capitalise Key?

andresmgot:
> The AppRepositories created in the chart are installed as hooks, but the Secrets you're creating are not. If I initially installed the chart with an AppRepository with a custom CA, and then were to upgrade the chart and not list it at all, the repository will still exist (by design, we don't touch initial repositories after creating them to ensure full control from the UI -- main reason, deleting a repo from the UI should not bring it back on upgrade), but the Secret now won't. This will cause the repository to no longer work. The Secret needs to be unmanaged too, to ensure this won't be an issue.

That's not what I am seeing. If I remove the repository with the secrets from the values the `apprepository` object gets deleted as well. See the next example, `chartmuseum` is the app repository with the custom CA:

```
▶ helm install ./chart/kubeapps --namespace kubeapps --name kubeapps        
NAME:   kubeapps
LAST DEPLOYED: Tue Feb  5 11:14:37 2019
NAMESPACE: kubeapps
STATUS: DEPLOYED
...

▶ kubectl get apprepositories.kubeapps.com -n kubeapps
NAME          AGE
bitnami       19s
chartmuseum   19s
incubator     19s
stable        19s
svc-cat       19s

# Now I delete chartmuseum from the values file
▶ helm upgrade kubeapps ./chart/kubeapps --namespace kubeapps                
Release "kubeapps" has been upgraded. Happy Helming!
LAST DEPLOYED: Tue Feb  5 11:15:36 2019
NAMESPACE: kubeapps
STATUS: DEPLOYED
...

▶ kubectl get apprepositories.kubeapps.com -n kubeapps              
NAME        AGE
bitnami     1m
incubator   1m
stable      1m
svc-cat     1m
```

I assume this is this way because if the `initialRepo` gets deleted the resource with the hook doesn't get rendered.

chart/kubeapps/templates/NOTES.txt
prydonius: I don't think we should require 2.12.0 just for some documentation, I would try to dig more into Go templates here to find another solution.

andresmgot:
I have found the reason, the thing is that the indentation of the annotation is wrong so it's not being used:

https://github.com/kubeapps/kubeapps/blob/master/chart/kubeapps/templates/apprepositories.yaml#L11

I think that in fact we don't need to install apprepositories in a hook so the behavior I comment in the previous message is fine. I will remove that annotation in this PR to avoid confusion.

chart/kubeapps/templates/NOTES.txt
andresmgot: I have been checking the documentation and I didn't find anything useful. The only thing I think  we can do is to print a line per app repository that contains a secret.

chart/kubeapps/templates/NOTES.txt
```diff
@@ -8,6 +8,14 @@ Kubeapps can be accessed via port {{ .Values.frontend.service.port }} on the fol
 
    {{ template "kubeapps.fullname" . }}.{{ .Release.Namespace }}.svc.cluster.local
 
+{{- range .Values.apprepository.initialRepos }}
```
prydonius: we can live with this, but one thing you could try: move this to a template and call the template using `include` (https://github.com/helm/helm/blob/master/docs/charts_tips_and_tricks.md#using-the-include-function) to get a string back, check for the string to not be empty and print if not empty.

andresmgot: I tried with `template` but not with `include`, that did the trick!

prydonius: nice 🎉!!!

chart/kubeapps/templates/apprepository-jobs-cleanup.yaml
```diff
@@ -22,12 +22,10 @@ spec:
...
-        - kubectl
-        - delete
-        - apprepositories.kubeapps.com
-        - -n
-        - {{ .Release.Namespace }}
-        - --all
+        - /bin/sh
+        args:
+        - -c
+        - kubectl delete apprepositories.kubeapps.com -n {{ .Release.Namespace }} --all; kubectl delete secrets -n {{ .Release.Namespace }} -l app={{ template "kubeapps.name" $ }},release={{ $.Release.Name }}
```
prydonius: I think this would be more readable and extendable if we formatted it as a multi-line string like in https://github.com/helm/charts/blob/master/stable/consul/templates/consul-statefulset.yaml#L109-L113.

chart/kubeapps/templates/NOTES.txt
```diff
@@ -8,6 +8,16 @@ Kubeapps can be accessed via port {{ .Values.frontend.service.port }} on the fol
 
    {{ template "kubeapps.fullname" . }}.{{ .Release.Namespace }}.svc.cluster.local
 
+{{- $reposWithOrphanSecrets := include "kubeapps.repos-with-orphan-secrets" . }}
+{{- if ne $reposWithOrphanSecrets "" }}
+
+CAVEAT:
+  Some App Repositories has been installed with a custom CA or authorization header.
+This generates secrets that won't be cleaned up if the repository is deleted through the Web application.
```
prydonius: can we indent these two lines too? It looks a little weird to me:

screen shot 2019-02-06 at 11 34 09

andresmgot:
> After testing this, it seems that neither a `pre-install` or `post-install` hook for the AppRepository CRs work, Helm will complain with:

```
Error: unable to decode "": no kind "AppRepository" is registered for version "kubeapps.com/v1alpha1"
```

I think that you hit the issue we have seen sometimes in the CI when Helm is trying to create CRD objects before the CRD is properly processed. I am not able to reproduce that issue, is it reproducible for you?

I can do your suggested change (even though I am reluctant to use the `crd-install` hook for something other than installing CRDs) but since the order is the same (install CRD and then the objects) I think that the issue will still be there. I think that the issue is that Tiller doesn't wait for CRDs to be ready:

```
7. Tiller waits until the hook is "Ready" (except for CRDs)
```

Probably this is the related issue: https://github.com/helm/helm/issues/2994

prydonius:
I tried several times, it was reproducible for me on Helm 2.11.0, but I haven't checked other versions. As soon as I switched the hook to `crd-install` and added the weight, I never saw that issue (though it's possible if an API Server is slower to pick up new CRDs, the race condition could occur again, but it didn't for me). Note that without the weight, I also reproducibly get that error (as expected).

> I think that the issue is that Tiller doesn't wait for CRDs to be ready

That would not really explain why it works in the `crd-install` hook and not `pre-install`. It's really weird though, I'm wondering if Helm 2.11.0 does any pre-validation on hooks before `crd-install`? I looked at the changelogs to see if anything related to this was fixed in Helm, but couldn't find anything.

What version of Helm are you using?

andresmgot:
> That would not really explain why it works in the `crd-install` hook and not `pre-install`. It's really weird though, I'm wondering if Helm 2.11.0 does any pre-validation on hooks before `crd-install`? I looked at the changelogs to see if anything related to this was fixed in Helm, but couldn't find anything.
What version of Helm are you using?

Indeed it's an issue with Helm 2.11 (I was using 2.12) and I can reproduce it with helm 2.12...

prydonius:
> I can reproduce it with helm 2.12...

so it's an issue in 2.12 too or did you mean 2.11? We should try to support 2.11 so I think best thing for now would be to use `crd-install` with the weight as that will work everywhere.

chart/kubeapps/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kubeapps
-version: 1.2.3
+version: 1.3.0
```
prydonius: this should now be bumped to 1.4.0

andresmgot:
> so it's an issue in 2.12 too or did you mean 2.11?

Sorry, yes, I meant 2.11

PR #957, opened 2019-02-05
Show errors when parsing Config

### LINKED ISSUE #892: config.json parsing issues seem to be swallowed
https://github.com/kubeapps/kubeapps/pull/891 fixed a syntax error in the config.json file that was causing deployments to fail, but we didn't easily figure out that it was due to the syntax error in that file. We should easily be able to detect parsing errors for that file and bubble them up so they are more noticeable. See also https://kubernetes.slack.com/archives/C9D3TSUG4/p1545178666065400.

Looks to me like axios is swallowing the error up: https://github.com/kubeapps/kubeapps/blob/490b316eb9df7d070fef7442fc5736d60d018175/dashboard/src/shared/Config.ts#L12

andresmgot:
Fixes #892

This PR adds a new component `ConfigLoader` to render an error if it's not possible to parse the configuration.

screenshot from 2019-02-05 17-01-58


dashboard/src/components/ConfigLoader/ConfigLoader.test.tsx
```diff
@@ -0,0 +1,33 @@
...
+  it("renders the error details", () => {
+    const wrapper = shallow(
+      <ConfigLoader error={new Error("Wrong config!")} getConfig={jest.fn()} />,
+    );
+    expect(wrapper.find(UnexpectedErrorPage)).toExist();
+    expect(wrapper).toMatchSnapshot();
+  });
+});
+
+it("calls getConfig", () => {
```
prydonius: small thing, but I like putting this in a `describe("componentDidMount")`

PR #959, opened 2019-02-07
move remaining sockets to redux, remove socket handling from AppView

### LINKED ISSUE #882: Access URLs and services list stuck in loading

prydonius:
This PR completes #882 by moving the rest of the socket resources to the
Redux state. AppView no longer handles the lifecycle of any sockets.

Note that this PR also removes watching Ingresses, which is not needed.
It is sufficient to retrieve data about Ingresses from a single GET.

closes #882 

Follow up issues I would like to create based on this:
- moar refactoring of AppView
  - I noticed a few strange things, like keeping the manifest in the state, and the use of IPartialAppViewState. I think this component can be further refactored to improve readability and become more functional.
- AccessURLTable can be split into items that can fetch themselves (rather than having the table fetch all items), following the pattern of the other Table/Item components and containers
- Improved handling of WebSockets, as discussed in https://github.com/kubeapps/kubeapps/issues/882#issuecomment-450185239, we are currently opening a socket for each individual resource of a kind, even though we can just have one for each kind (e.g. 1 socket for watching N deployments in a namespace, instead of N sockets)


dashboard/src/components/AppView/DeploymentsTable/DeploymentItem.test.tsx
```diff
@@ -24,13 +35,18 @@ context("when fetching deployments", () => {
...
         deployment,
-        getDeployment: jest.fn(),
+        watchDeployment: jest.fn(),
       },
     });
 
     it("displays the name of the deployment", () => {
       const wrapper = shallow(
-        <DeploymentItem deployment={deployment} name="foo" getDeployment={jest.fn()} />,
+        <DeploymentItem
```
andresmgot: we can save a few lines of code if we store this props as `defaultProps` and we use that around.

dashboard/src/components/DeploymentStatus/DeploymentStatus.test.tsx
```diff
@@ -4,13 +4,38 @@ import * as React from "react";
...
 it("renders a loading status", () => {
   const deployments = [
     {
       isFetching: true,
     },
   ];
-  const wrapper = shallow(<DeploymentStatus deployments={deployments} />);
+  const wrapper = shallow(
+    <DeploymentStatus
+      deployments={deployments}
```
andresmgot: the same here about `defaultProps`

dashboard/src/components/AppView/DeploymentsTable/DeploymentItem.test.tsx
prydonius: good shout :+1:

Issue #949, opened 2019-01-30
Add support for configuring an OIDC proxy for Kubeapps in the chart


prydonius:
Related to #385, #931 and dependent on #945 and #947. We can help automate the setup of #947 by adding configurable options in the chart to set your OIDC URL, client ID and secret. This will then setup a proxy in front of Kubeapps capable of performing the OIDC dance and returning a token that Kubeapps can use to authenticate users.

stormmore:
I currently do this using a kubectl plugin to login and a kubectl command to get the token out of my .kube/config. This would be super useful

andresmgot:
Once we implement this, we need to add to Nginx:

For keycloak-gatekeeper, to allow the dashboard to access the Authorization header:

```
`add_header Authorization $http_authorization;`
```

For oauth2_proxy, to be able to include the required headers:

```
    nginx.ingress.kubernetes.io/proxy-buffer-size: 256k
    nginx.ingress.kubernetes.io/proxy-buffering: "on"
```

PR #961, opened 2019-02-11
Allow to receive an OIDC token

### LINKED ISSUE #385: Support OIDC tokens from a third party provider
Allow for the support of OpenID tokens to be given to authenticate users to Kubeapps in a manner similiar to Kubernetes Dashboard.  

The administrator should be able to place a reverse proxy in front of Kubeapps which will redirect users to an Identity Provider (IdP).  User then authenticates and a bearer token is given to the proxy.  This bearer token is then injected transparently into subsequent requests by the user.

The OIDC process is documented here for Kubernetes: https://kubernetes.io/docs/reference/access-authn-authz/authentication/

It should be possible to tie a group provided in the Json Web Token (JWT) to a ClusterRoleBinding within Kubernetes.  

`kind: ClusterRoleBinding
metadata:
  name: keycloak-admin-group
roleRef:
  kind: ClusterRole
  name: cluster-admin
subjects:
- kind: Group
  name: /admin`

### LINKED ISSUE #945: Support reading auth bearer token from headers
Related to #385 and #931 

We want to be able to read and use an auth bearer token from the headers of the request to the dashboard to allow a proxy (like keycloak-gateway) to automatically set a valid bearer token before forwarding to Kubeapps.

See the changes @migmartri made for his PoC of this feature: https://github.com/migmartri/kubeapps/commit/a784c53d20d4df4a9437a3fed319e50502835272.

andresmgot:
Ref #385 
Closes #945 

This PR enables for a third-party OIDC proxy to set the headers `X-Auth-Token-ID` or `Authorization` to automatically get authenticated in Kubeapps without the need to get the token of a service account. This PR has been tested using Keycloak as IdP and oauth2_proxy as the proxy to bypass the authentication. Other solution like `keycloak-gatekeeper` should be supported as well.

Since the login process is no longer handled by Kubeapps, the "Logout" button gets disabled if this authorization mechanism is used.

### DIFF
--- a/dashboard/src/actions/auth.ts
+++ b/dashboard/src/actions/auth.ts
@@ -5,7 +5,7 @@ import { Auth } from "../shared/Auth";
 import { IStoreState } from "../shared/types";
 
 export const setAuthenticated = createAction("SET_AUTHENTICATED", resolve => {
-  return (authenticated: boolean, withToken?: boolean) => resolve({ authenticated, withToken });
+  return (authenticated: boolean, oidc?: boolean) => resolve({ authenticated, oidc });
 });
 
 export const authenticating = createAction("AUTHENTICATING");
@@ -14,30 +14,22 @@ export const authenticationError = createAction("AUTHENTICATION_ERROR", resolve 
   return (errorMsg: string) => resolve(errorMsg);
 });
 
-export const checkingOIDCToken = createAction("CHECKING_OIDC_TOKEN");
-
-const allActions = [setAuthenticated, authenticating, authenticationError, checkingOIDCToken];
+const allActions = [setAuthenticated, authenticating, authenticationError];
 
 export type AuthAction = ActionType<typeof allActions[number]>;
 
 export function authenticate(
-  token?: string,
+  token: string,
+  oidc?: boolean,
 ): ThunkAction<Promise<void>, IStoreState, null, AuthAction> {
   return async dispatch => {
     dispatch(authenticating());
     try {
-      await Auth.validate(token);
-      if (token) {
-        Auth.setAuthToken(token);
-      }
-      dispatch(setAuthenticated(true, !!token));
+      await Auth.validateToken(token);
+      Auth.setAuthToken(token, oidc);
+      dispatch(setAuthenticated(true, oidc));
     } catch (e) {
-      if (token) {
-        // The error will only be meaningful if a token was given
-        dispatch(authenticationError(e.toString()));
-      } else {
-        dispatch(setAuthenticated(false));
-      }
+      dispatch(authenticationError(e.toString()));
     }
   };
 }
@@ -49,9 +41,19 @@ export function logout(): ThunkAction<Promise<void>, IStoreState, null, AuthActi
   };
 }
 
-export function tryToAutoAuthenticate(): ThunkAction<Promise<void>, IStoreState, null, AuthAction> {
+export function tryToAuthenticateWithOIDC(): ThunkAction<
+  Promise<void>,
+  IStoreState,
+  null,
+  AuthAction
+> {
   return async dispatch => {
-    dispatch(checkingOIDCToken());
-    dispatch(authenticate(undefined));
+    dispatch(authenticating());
+    const token = await Auth.fetchOIDCToken();
+    if (token) {
+      dispatch(authenticate(token, true));
+    } else {
+      dispatch(setAuthenticated(false));
+    }
   };
 }
--- a/dashboard/src/actions/auth.ts
+++ b/dashboard/src/actions/auth.ts
@@ -14,7 +14,19 @@ export const authenticationError = createAction("AUTHENTICATION_ERROR", resolve 
   return (errorMsg: string) => resolve(errorMsg);
 });
 
-const allActions = [setAuthenticated, authenticating, authenticationError];
+export const autoAuthenticating = createAction("AUTO_AUTHENTICATING");
+
+export const setAutoAuthenticated = createAction("SET_AUTO_AUTHENTICATED", resolve => {
+  return (authenticated: boolean) => resolve(authenticated);
+});
+
+const allActions = [
+  setAuthenticated,
+  authenticating,
+  authenticationError,
+  autoAuthenticating,
+  setAutoAuthenticated,
+];
 
 export type AuthAction = ActionType<typeof allActions[number]>;
 
@@ -37,5 +49,22 @@ export function logout(): ThunkAction<Promise<void>, IStoreState, null, AuthActi
   return async dispatch => {
     Auth.unsetAuthToken();
     dispatch(setAuthenticated(false));
+  };
+}
+
+export function tryToAutoAuthenticate(): ThunkAction<Promise<void>, IStoreState, null, AuthAction> {
+  return async dispatch => {
+    dispatch(autoAuthenticating());
+    try {
+      const token = await Auth.fetchToken();
+      if (token) {
+        Auth.setAuthToken(token);
+        dispatch(setAutoAuthenticated(true));
+      } else {
+        dispatch(setAutoAuthenticated(false));
+      }
+    } catch (e) {
+      dispatch(authenticationError(e.toString()));
+    }
   };
 }
--- a/dashboard/src/actions/auth.ts
+++ b/dashboard/src/actions/auth.ts
@@ -5,7 +5,7 @@ import { Auth } from "../shared/Auth";
 import { IStoreState } from "../shared/types";
 
 export const setAuthenticated = createAction("SET_AUTHENTICATED", resolve => {
-  return (authenticated: boolean, oidc?: boolean) => resolve({ authenticated, oidc });
+  return (authenticated: boolean, oidc: boolean) => resolve({ authenticated, oidc });
 });
 
 export const authenticating = createAction("AUTHENTICATING");
@@ -20,7 +20,7 @@ export type AuthAction = ActionType<typeof allActions[number]>;
 
 export function authenticate(
   token: string,
-  oidc?: boolean,
+  oidc: boolean,
 ): ThunkAction<Promise<void>, IStoreState, null, AuthAction> {
   return async dispatch => {
     dispatch(authenticating());
@@ -37,7 +37,7 @@ export function authenticate(
 export function logout(): ThunkAction<Promise<void>, IStoreState, null, AuthAction> {
   return async dispatch => {
     Auth.unsetAuthToken();
-    dispatch(setAuthenticated(false));
+    dispatch(setAuthenticated(false, false));
   };
 }
 
@@ -53,7 +53,7 @@ export function tryToAuthenticateWithOIDC(): ThunkAction<
     if (token) {
       dispatch(authenticate(token, true));
     } else {
-      dispatch(setAuthenticated(false));
+      dispatch(setAuthenticated(false, false));
     }
   };
 }
--- a/dashboard/src/actions/auth.ts
+++ b/dashboard/src/actions/auth.ts
@@ -5,7 +5,7 @@ import { Auth } from "../shared/Auth";
 import { IStoreState } from "../shared/types";
 
 export const setAuthenticated = createAction("SET_AUTHENTICATED", resolve => {
-  return (authenticated: boolean) => resolve(authenticated);
+  return (authenticated: boolean, withToken?: boolean) => resolve({ authenticated, withToken });
 });
 
 export const authenticating = createAction("AUTHENTICATING");
@@ -14,33 +14,30 @@ export const authenticationError = createAction("AUTHENTICATION_ERROR", resolve 
   return (errorMsg: string) => resolve(errorMsg);
 });
 
-export const autoAuthenticating = createAction("AUTO_AUTHENTICATING");
+export const checkingOIDCToken = createAction("CHECKING_OIDC_TOKEN");
 
-export const setAutoAuthenticated = createAction("SET_AUTO_AUTHENTICATED", resolve => {
-  return (authenticated: boolean) => resolve(authenticated);
-});
-
-const allActions = [
-  setAuthenticated,
-  authenticating,
-  authenticationError,
-  autoAuthenticating,
-  setAutoAuthenticated,
-];
+const allActions = [setAuthenticated, authenticating, authenticationError, checkingOIDCToken];
 
 export type AuthAction = ActionType<typeof allActions[number]>;
 
 export function authenticate(
-  token: string,
+  token?: string,
 ): ThunkAction<Promise<void>, IStoreState, null, AuthAction> {
   return async dispatch => {
     dispatch(authenticating());
     try {
-      await Auth.validateToken(token);
-      Auth.setAuthToken(token);
-      dispatch(setAuthenticated(true));
+      await Auth.validate(token);
+      if (token) {
+        Auth.setAuthToken(token);
+      }
+      dispatch(setAuthenticated(true, !!token));
     } catch (e) {
-      dispatch(authenticationError(e.toString()));
+      if (token) {
+        // The error will only be meaningful if a token was given
+        dispatch(authenticationError(e.toString()));
+      } else {
+        dispatch(setAuthenticated(false));
+      }
     }
   };
 }
@@ -54,17 +51,7 @@ export function logout(): ThunkAction<Promise<void>, IStoreState, null, AuthActi
 
 export function tryToAutoAuthenticate(): ThunkAction<Promise<void>, IStoreState, null, AuthAction> {
   return async dispatch => {
-    dispatch(autoAuthenticating());
-    try {
-      const token = await Auth.fetchToken();
-      if (token) {
-        Auth.setAuthToken(token);
-        dispatch(setAutoAuthenticated(true));
-      } else {
-        dispatch(setAutoAuthenticated(false));
-      }
-    } catch (e) {
-      dispatch(authenticationError(e.toString()));
-    }
+    dispatch(checkingOIDCToken());
+    dispatch(authenticate(undefined));
   };
 }
--- a/dashboard/src/components/LoginForm/LoginForm.tsx
+++ b/dashboard/src/components/LoginForm/LoginForm.tsx
@@ -9,10 +9,9 @@ import "./LoginForm.css";
 interface ILoginFormProps {
   authenticated: boolean;
   authenticating: boolean;
-  checkingOIDCToken: boolean;
   authenticationError: string | undefined;
   authenticate: (token: string) => any;
-  tryToAutoAuthenticate: () => void;
+  tryToAuthenticateWithOIDC: () => void;
   location: Location;
 }
 
@@ -24,11 +23,11 @@ class LoginForm extends React.Component<ILoginFormProps, ILoginFormState> {
   public state: ILoginFormState = { token: "" };
 
   public componentDidMount() {
-    this.props.tryToAutoAuthenticate();
+    this.props.tryToAuthenticateWithOIDC();
   }
 
   public render() {
-    if (this.props.checkingOIDCToken) {
+    if (this.props.authenticating) {
       return <LoadingWrapper />;
     }
     if (this.props.authenticated) {
@@ -76,11 +75,7 @@ class LoginForm extends React.Component<ILoginFormProps, ILoginFormState> {
                     />
                   </div>
                   <p>
-                    <button
-                      type="submit"
-                      className="button button-accent"
-                      disabled={this.props.authenticating}
-                    >
+                    <button type="submit" className="button button-accent">
                       Login
                     </button>
                   </p>
--- a/dashboard/src/components/LoginForm/LoginForm.tsx
+++ b/dashboard/src/components/LoginForm/LoginForm.tsx
@@ -3,13 +3,16 @@ import * as React from "react";
 import { Lock } from "react-feather";
 import { Redirect } from "react-router";
 
+import LoadingWrapper from "../../components/LoadingWrapper";
 import "./LoginForm.css";
 
 interface ILoginFormProps {
   authenticated: boolean;
   authenticating: boolean;
+  autoAuthenticating: boolean;
   authenticationError: string | undefined;
   authenticate: (token: string) => any;
+  tryToAutoAuthenticate: () => void;
   location: Location;
 }
 
@@ -19,7 +22,15 @@ interface ILoginFormState {
 
 class LoginForm extends React.Component<ILoginFormProps, ILoginFormState> {
   public state: ILoginFormState = { token: "" };
+
+  public componentDidMount() {
+    this.props.tryToAutoAuthenticate();
+  }
+
   public render() {
+    if (this.props.autoAuthenticating) {
+      return <LoadingWrapper />;
+    }
     if (this.props.authenticated) {
       const { from } = this.props.location.state || { from: { pathname: "/" } };
       return <Redirect to={from} />;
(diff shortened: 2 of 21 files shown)


dashboard/src/containers/HeaderContainer/HeaderContainer.ts
```diff
@@ -23,13 +23,15 @@ function mapStateToProps({
...
+    // If autoAuthenticated it's not possible to logout
+    disableLogout: autoAuthenticated,
   };
 }
 
 function mapDispatchToProps(dispatch: ThunkDispatch<IStoreState, null, Action>) {
   return {
     fetchNamespaces: () => dispatch(actions.namespace.fetchNamespaces()),
-    logout: (token: string) => dispatch(actions.auth.logout()),
+    logout: () => dispatch(actions.auth.logout()),
```
andresmgot: this change was unrelated to the PR but the linter was complaining about the unused parameter

dashboard/src/shared/Auth.ts
```diff
@@ -44,6 +44,16 @@ export class Auth {
       }
     }
   }
+
+  public static async fetchToken(): Promise<string | null> {
+    try {
+      const { headers } = await Axios.get("/");
+      // oauth2_proxy set the header "authorization" while keycloak-proxy uses "x-auth-token-id"
```
migmartri: ` keycloak-proxy` is deprecated in favor of `keycloak-gatekeeper`

migmartri: Why did you choose get instead of options? https://github.com/migmartri/kubeapps/commit/a784c53d20d4df4a9437a3fed319e50502835272#diff-6715d504703bc9764780bca91a104960R27

Also, note that Keycloak gatekeeper [will return](https://github.com/keycloak/keycloak-gatekeeper/blob/3da1a34db16261f3e58cbc99b1076c10131ead71/session_test.go#L38) the Authentication Header already formed, does that mean that will get stored as token?

dashboard/src/actions/auth.ts
```diff
@@ -39,3 +51,20 @@ export function logout(): ThunkAction<Promise<void>, IStoreState, null, AuthActi
...
   };
 }
+
+export function tryToAutoAuthenticate(): ThunkAction<Promise<void>, IStoreState, null, AuthAction> {
+  return async dispatch => {
+    dispatch(autoAuthenticating());
+    try {
+      const token = await Auth.fetchToken();
+      if (token) {
+        Auth.setAuthToken(token);
```
migmartri: Are we validating this token against the API server? See `Auth.validateToken(token);`

Note that a lot of the logic is shared with `authenticate`

dashboard/src/shared/Auth.ts
prydonius: I tested a setup with Keycloak gatekeeper and it will indeed return a fully formed Authorization header. It looks like oauth2_proxy does the same (https://github.com/pusher/oauth2_proxy/blob/fb13ee87c8a301724e45e8056cb6c8b2913cd894/oauthproxy.go#L888). We prepend `Bearer` to the create the header in https://github.com/kubeapps/kubeapps/blob/499ed26514fd515787ad0cb32cec7edd24098f92/dashboard/src/shared/AxiosInstance.ts#L21. Given this, are we sending out malformed Authorization header?

One thought I have is that this might have worked for you because the proxy resets the Authorization header when forwarding the request to the API server. If this is the case, we actually might be better of not storing the token at all and delegating the handling of that to the proxy (it should be able to deal with refresh tokens too).

If we do need to store the token (because my assumption is wrong or for some other use), I would suggest that we change the logic here to only check the Authorization header and strip the `Bearer` prefix. Given both keycloak-gatekeeper and oauth2_proxy use the Authorization header, I don't think there is a need to handle the `x-auth-token-id` header.

dashboard/src/shared/Auth.ts
```diff
@@ -44,6 +44,16 @@ export class Auth {
       }
     }
   }
+
+  public static async fetchToken(): Promise<string | null> {
```
prydonius: I would call this function `fetchOIDCToken` and add a comment explaining that this will make a request to see if an Authorization header is already present

dashboard/src/shared/Auth.ts
```diff
@@ -33,13 +33,17 @@ export class Auth {
     };
   }
 
-  // Throws an error if the token is invalid
-  public static async validateToken(token: string) {
+  // Throws an error if accessing the Kubernetes API returns an Unauthorized error
+  public static async validate(token?: string) {
```
andresmgot: this is the main change, in the case of having an available OIDC proxy the token will be optional so a request to `/api/kube` will success without us having to worry about the token.

dashboard/src/shared/Auth.ts
migmartri: There seems to be an options alias https://github.com/axios/axios#axiosoptionsurl-config

prydonius: > Because in the TODO said that it should use axios instead of fetch and axios doesn't have an options method as far as I can see.

A quick look at the documentation shows that it is there, or you can also use `axios(url, config)` to set the method explicitly: https://github.com/axios/axios#axiosconfig.

Looks like it was added 2 years ago https://github.com/axios/axios/commit/2e0adc1cae76eb214800ce64cfb5d27c032c8556 but not added to the typings (which is why I guess you thought it was not there): https://github.com/axios/axios/blob/master/index.d.ts#L133-L138.

I would look for a way to add it to the typings if we can extend them, or just use `axios(url, config)` and set the method that way.

Now that I think about this, though, it might actually make more sense to use HEAD here. Responses in https://stackoverflow.com/questions/220231/accessing-the-web-pages-http-headers-in-javascript mention HEAD and GET, and not OPTIONS. HEAD also has the benefit over OPTIONS that the response can be cached (https://stackoverflow.com/questions/6660019/restful-api-methods-head-options), though not sure if that is important/useful in our usecase. Axios also has `.head` defined in the typings, so that's another plus.

> I see, I didn't check `keycloak-gatekeeper` I just picked that code from @migmartri POC. When I try gatekeeper it's only setting the header `x-auth-token-id`. Also, a difference from oauth2_proxy is that it doesn't automatically populate the Authorization header so we would need this manual token pick that is not necessary with oauth2_proxy. At least I couldn't find out how I can do that with keycloak-gatekeeper but it should be something doable, I will keep checking.

Sounds good, I can confirm that keycloak-gatekeeper is setting the auth header when testing with echo headers in the cluster @migmartri has configured, so we should check the configuration there. Actually, I just checked the documentation (https://www.keycloak.org/docs/latest/securing_apps/index.html#upstream-headers) and it looks like there is a flag `--enable-authorization-header` and it should be set by default.

dashboard/src/components/LoginForm/LoginForm.tsx
```diff
@@ -19,7 +22,15 @@ interface ILoginFormState {
 
 class LoginForm extends React.Component<ILoginFormProps, ILoginFormState> {
   public state: ILoginFormState = { token: "" };
+
+  public componentDidMount() {
+    this.props.tryToAutoAuthenticate();
+  }
+
   public render() {
+    if (this.props.autoAuthenticating) {
```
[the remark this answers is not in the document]
prydonius: Sorry, I wasn't very clear, I was actually suggesting that we reuse `this.props.authenticating` and show a loading screen instead of the form. This does mean that when logging in with an invalid token, it'll flash to a loading screen and then back to the form, but I think that's not too bad and most users entering tokens correctly won't see any issues.

prydonius:
I spent some more time looking into this today and I was able to get keycloak-gatekeeper setup on my local Kubernetes environment. The key thing here was to disable secure cookies in the proxy (which requires HTTPS) and this stopped the redirect loop. Here is the config I'm using for keycloak-gatekeeper:

```
- --client-id=kubernetes
- --client-secret=not-needed
- --discovery-url=<redacted>
- --listen=0.0.0.0:3000
- --verbose
- --enable-default-deny=true
- --upstream-url=http://kubeapps
- --secure-cookie=false
```

With this, I was able to access the dashboard (without your changes) and see the login screen after the OIDC dance.

Some further testing made me realise that it isn't sufficient to assume you're logged in if you are able to make a successful API request to `/api/kube/` because:
- If no authorization token is set, Kubernetes will put you in the "system:anonymous" group and `/api/kube/` will return a 403 forbidden
- If an auth header is set (by proxy or otherwise), it is valid, and your user **doesn't** have permission to access `/api/kube/`, the API will return a 403 forbidden
- If an auth header is set (by proxy or otherwise), it is valid, and your user **does** have permission to access `/api/kube/`, the API will return a 200
- If a token is set and it is invalid, the API will return 401

So assuming 403 for `/` means an invalid token is actually incorrect, it can mean that a user

Before this PR, we would show a login form asking for manual token input. We would then use this token to make a request against the API `/api/kube/` with the following cases:
- 401 - invalid token, show login form
- 200/403/anything else - token is valid, logged in

With this PR, we now always make a request to the API to attempt the OIDC login at `/api/kube/` with the following cases:
- 401/403 - invalid token or forbidden access to `/` on apiserver, show login form
- 200 - logged in

The side effect of this is that we now **require** the user to have necessary RBAC permission to access `/` on the API server. This resource is not required for normal operation of the cluster. It may not be a huge deal, but we'd have to document and explain it. In conservative environments where RBAC is more restricted though, this could be seen as unnecessarily annoying. By the way, this is easily testable by creating a RoleBinding in a namespace to `cluster-admin` for a user. I did this and a request to `/` returns 403, but a request for pods in the namespace is successful (`/api/v1/namespaces/default/pods`). I would need a ClusterRoleBinding with a nonResourceURL rule for `/` to give the user this permission.

img_20190212_174714

I think we're better off avoiding this requirement by going back to checking if there actually is an `Authorization` header set from the OIDC proxy (sorry 😬). This will allow us to guarantee that there is a something intentionally setting the header, and we can then treat a 403 response to `/api/kube/` as a successful login (as we always have). To clarify, this is because when we know a token is present in the request, the API server will return a 401 if that token is invalid. On the flip side, if no `Authorization` header is set from a proxy, we will just show the login form as we do today and it will work as it does today. I see the flow working as follows:

download-1

FYI, I'm being intentionally verbose here as it helps me wrap my head around all of this. I know the above is effectively what you had before we changed this to not store the token, but it helps to see the response status codes in this flow too. Also, we still don't need to store the token, I think it will be sufficient to just retrieve it and then validate it.

By the way, with keycloak-gatekeeper I needed to set some Nginx configuration to get the `Authorization` header back in the response. This was pointed out by @migmartri in the design doc (links to https://github.com/migmartri/k8s-authentication-playground/blob/master/kubeapps/templates/kubeapps-frontend-config.yaml#L69), but I also needed `add_header Authorization    $http_authorization;` to get the `Authorization` header specifically. Did you do this also, or does oauth2_proxy not require this? Perhaps this is why you weren't seeing the `Authorization` header with keycloak-gatekeeper before?

dashboard/src/reducers/auth.ts
```diff
@@ -6,25 +6,39 @@ import { AuthAction } from "../actions/auth";
 export interface IAuthState {
   authenticated: boolean;
   authenticating: boolean;
+  checkingOIDCToken: boolean;
+  autoAuthenticated: boolean;
```
prydonius: `autoAuthenticated` could be `oidc: boolean` or `oidcAuthenticated: boolean` to make it more clear what the intended usecase is

dashboard/src/actions/auth.ts
```diff
@@ -39,3 +48,10 @@ export function logout(): ThunkAction<Promise<void>, IStoreState, null, AuthActi
     dispatch(setAuthenticated(false));
   };
 }
+
+export function tryToAutoAuthenticate(): ThunkAction<Promise<void>, IStoreState, null, AuthAction> {
```
prydonius: I think we should replace "auto" with "oidc" everywhere to be more clear.

andresmgot:
Thanks for the detailed notes. That indeed makes sense, let me apply the suggestions.

andresmgot:
One caveat for this:

> This will allow us to guarantee that there is a something intentionally setting the header, and we can then treat a 403 response to `/api/kube/` as a successful login (as we always have).

Note that an empty token returns a 403 or it's even allowed to get a response so potentially we have the risk of assuming an empty token as logged in:

```
▶ curl -sk -H "Authorization: Bearer " https://192.168.99.100:8443/api/v1/      
{
  "kind": "APIResourceList",
  "groupVersion": "v1",
  "resources": [
    {
...
```

I think we are fine since we force people to people to put something in the `token` field but I found it interesting.

dashboard/src/shared/Auth.ts
andresmgot: > Now that I think about this, though, it might actually make more sense to use HEAD here

Okay, I will use HEAD then.

dashboard/src/components/LoginForm/LoginForm.tsx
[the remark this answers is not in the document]
andresmgot: ah, okay, that makes more sense

prydonius:
> One caveat for this:
> 
> > This will allow us to guarantee that there is a something intentionally setting the header, and we can then treat a 403 response to `/api/kube/` as a successful login (as we always have).
> 
> Note that an empty token returns a 403 or it's even allowed to get a response so potentially we have the risk of assuming an empty token as logged in:
> 
> ```
> ▶ curl -sk -H "Authorization: Bearer " https://192.168.99.100:8443/api/v1/      
> {
>   "kind": "APIResourceList",
>   "groupVersion": "v1",
>   "resources": [
>     {
> ...
> ```
> I think we are fine since we force people to add something in the `token` field but I found it interesting.

That is interesting, I guess the API server skips using the token in this case because the header is in an invalid format? We're following the same assumption as the API server so we'll show the login form in this case, so as you say I think we're fine.

prydonius:
> > but I also needed `add_header Authorization $http_authorization;`
> 
> Those two things did the trick! I got this working with both oauth2_proxy and keycloak-gatekeeper.

Do we know why it just worked with oauth2_proxy and why keycloak-gatekeeper needs this? Would be good to understand this.

prydonius:
Thanks for updating this, barring some minor changes and working out the issues below, we're getting close to having this working!

> Those two things did the trick!

Shouldn't we add the directive in this PR to the chart? Or would you rather do it in a separate PR?

I tested this today with keycloak-gatekeeper where the auto login worked perfectly! However, I couldn't get this working with oauth2_proxy, here is my configuration:

```
        - --client-id=kubernetes
        - --client-secret=not-needed
        - --cookie-secret=<redacted>
        - --email-domain=*
        - --cookie-secure=false
        - --provider=oidc
        - --http-address=0.0.0.0:3000
        - --oidc-issuer-url=<redacted>
        - --pass-authorization-header=true
        - --upstream=http://kubeapps
```

Specifically, with the `--pass-authorization-header` arg I see the following error in the Ingress:

```
2019/02/13 22:15:11 [error] 282#282: *63488 upstream sent too big header while reading response header from upstream
```

I'm guessing I don't actually need to increase the header size, so I must be doing something wrong in my setup. Would you be able to share your configuration?

Also, I ran into an issue with keycloak-gatekeeper and WebSockets not working consistently:

screen shot 2019-02-13 at 11 49 01
screen shot 2019-02-13 at 11 49 14

If keycloak-gatekeeper doesn't handle WebSocket upgrades properly, we might need to just whitelist `/api/kube` and rely on sending the `Authorization` token to the API server ourselves (which we already do, so no changes code-wise required).

I'm testing with `keycloak-gatekeeper version 4.6.0 (git+sha: 82f61a2-dirty, built: 14-11-2018)` so the https://github.com/keycloak/keycloak-gatekeeper/releases/tag/v2.1.1 release which apparently fixes some WebSocket issues should be included...

Can you check if you're seeing the same issue?

Also, I tried to see what would happen if the tokens expired. I set the token expiration in Keycloak to 1 minute and saw that expired tokens don't reinitiate the OAuth flow when clicking around the app, AJAX requests are getting 307 redirects to start the OAuth flow, but it is blocked by CORS. Despite this, resources seem to load... Not sure what to do here or if we need to do anything, can you investigate?

dashboard/src/reducers/auth.ts
```diff
@@ -6,25 +6,33 @@ import { AuthAction } from "../actions/auth";
...
 };
 
 const authReducer = (state: IAuthState = initialState, action: AuthAction): IAuthState => {
   switch (action.type) {
     case getType(actions.auth.setAuthenticated):
-      return { ...state, authenticated: action.payload, authenticating: false };
+      return {
+        ...state,
+        authenticated: action.payload.authenticated,
+        oidcAuthenticated: !!action.payload.oidc,
```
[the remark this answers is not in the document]
andresmgot: yes it's a boolean but it's optional, so that's why I needed the conversion. In any case there is not a requirement for this to be optional so we can avoid this.

andresmgot:
>> but I also needed `add_header Authorization $http_authorization;`
>
> Do we know why it just worked with oauth2_proxy and why keycloak-gatekeeper needs this? Would be good to understand this.


> Shouldn't we add the directive in this PR to the chart? Or would you rather do it in a separate PR?

My plan was to add that in a different PR (#949) so we don't extend this PR further.

> I tested this today with keycloak-gatekeeper where the auto login worked perfectly! However, I couldn't get this working with oauth2_proxy, here is my configuration:
Specifically, with the `--pass-authorization-header` arg I see the following error in the Ingress:
>```
>2019/02/13 22:15:11 [error] 282#282: *63488 upstream sent too big header while reading response >header from upstream
>```
>I'm guessing I don't actually need to increase the header size, so I must be doing something wrong in my setup. Would you be able to share your configuration?

Yes, I needed to modify the ingress object and add some annotations in order to avoid that issue:

```
    nginx.ingress.kubernetes.io/connection-proxy-header: keep-alive
    nginx.ingress.kubernetes.io/proxy-buffer-size: 256k
    nginx.ingress.kubernetes.io/proxy-buffering: "on"
```

This is the configuration I am using (I haven't rechecked the flags yet so some of them may not be necessary):

```
        - --email-domain=*
        - --cookie-secure=false
        - --cookie-secret=AhW93OvPPMOF+3xS
        - --client-secret=02836ab4-c4b2-49c8-adda-53fa4128eb76
        - --client-id=oidckube
        - --provider=oidc
        - --oidc-issuer-url=https://keycloak.devlocal/auth/realms/k8s
        - --ssl-insecure-skip-verify
        - --http-address=0.0.0.0:4180
        - --upstream=http://kubeapps/
        - --pass-authorization-header
        - --set-authorization-header
```

> Also, I ran into an issue with keycloak-gatekeeper and WebSockets not working consistently:
> Can you check if you're seeing the same issue?

Yes, I am seeing the same issue. As you say, if we whitelist the resource it's fine since we are already adding the token ourselves. Whitelisting works for me for gatekeeper. 

What concerns me is that `oauth2_proxy` is not able to handle the connection. If I skip the authentication (`--skip-auth-regex=/api/kube/.*`) it receives a 403 Forbidden and if I don't do that it never upgrades the connection.

> Also, I tried to see what would happen if the tokens expired. I set the token expiration in Keycloak to 1 minute and saw that expired tokens don't reinitiate the OAuth flow when clicking around the app, AJAX requests are getting 307 redirects to start the OAuth flow, but it is blocked by CORS. Despite this, resources seem to load... Not sure what to do here or if we need to do anything, can you investigate?

I have reduced the expiration time and I have tried 3 things:
 - Wait until the token expires: When doing this, when changing the location (for example from the app list to a specific app) I get momentarily redirected to the `/login` page and the token gets refreshed and I am redirected again to the page I was meant to visit. I think that's the expected behavor, I am not seeing those redirects:
screenshot from 2019-02-14 16-31-40
 - Close the session of the user from keycloak: I can navigate the application until the token expires, then I start receiving 403 errors from the application but I don't get redirected. The application claims that I don't have sufficient permissions to request the resources. When reloading the page I get redirected again to the login page of keycloak:
screenshot from 2019-02-14 16-33-23
 - Delete the user from keycloak: The same behavior than above.

prydonius:
> My plan was to add that in a different PR (#949) so we don't extend this PR further.

Can you add a comment in #949 so we don't forget it? I would mention the nginx annotations needed for oauth2_proxy as well as the set header thing.

> * Wait until the token expires: When doing this, when changing the location (for example from the app list to a specific app) I get momentarily redirected to the `/login` page and the token gets refreshed and I am redirected again to the page I was meant to visit. I think that's the expected behavior:

Sounds good, I was seeing CORS issues with this because I'm not running keycloak on the same domain.

I'm still not able to get oauth2_proxy to work for some reason (not a reason to hold this back). I am getting a "This page isn’t working If the problem continues, contact the site owner.
HTTP ERROR 400" error, and I can't see anything in the logs. Any ideas?

dashboard/src/shared/Auth.ts
```diff
@@ -44,6 +46,19 @@ export class Auth {
       }
     }
   }
+
+  // fetchOIDCToken does a HEAD request to collect the Bearer token
+  // from the authorization header if exists
+  public static async fetchOIDCToken(): Promise<string | null> {
+    const { headers } = await Axios.head("/");
+    if (headers && headers.authorization) {
+      const tokenMatch = (headers.authorization as string).match(/Bearer\s([a-zA-Z0-9\._-]*)/);
```
prydonius: Could you explain the change from `.*` to `[a-zA-Z0-9\._-]*`? Isn't it safer to use `.*` here, just in case we're missing permitting a character that a provider might use? For example, what if there is a padding and we need `=`? What if the token is not in JWT format (not sure if that's possible with OIDC, but I could, instead of using OIDC, also hardcode the Ingress with a specific bearer token my apiserver accepts to bypass all auth).

andresmgot:
> Can you add a comment in #949 so we don't forget it? I would mention the nginx annotations needed for oauth2_proxy as well as the set header thing.

Sure, it's difficult to forget because it won't work otherwise :P https://github.com/kubeapps/kubeapps/issues/949#issuecomment-463957173

> I am getting a "This page isn’t working If the problem continues, contact the site owner.
HTTP ERROR 400" error, and I can't see anything in the logs. Any ideas?

That was what I was getting without the flag `--cookie-secure=false` (the thing that caused the redirect loop). If you are adding the browser may be caching stuff, I need to work in a incognito window every time I switch context.

dashboard/src/shared/Auth.ts
andresmgot: Yes, I forgot the comment here. The problem is that when using the directive `add_header Authorization $http_authorization` with oauth2_proxy it's adding the Bearer token twice so the header looks like `Bearer ABC...XYZ, Bearer ABC...XYZ`. We can make it work just not adding that directive for oauth2_proxy but since it was simple enough I added the regex. In any case, it's true that it's not the safest regex, I can look for `,` instead.

andresmgot:
> but I also needed `add_header Authorization $http_authorization;`
> Do we know why it just worked with oauth2_proxy and why keycloak-gatekeeper needs this? Would be good to understand this.

Sorry I forgot to answer that. I was using the flag `--set-authorization-header` that has the same effect than setting `add_header Authorization`. This actually caused the effect of adding the Bearer token twice in the Authorization header if both settings are set. If I remove that flag the behavior of gatekeeper and oauth2_proxy is the same.

dashboard/src/shared/Auth.ts
prydonius: @andresmgot is it possible to disable that in oauth2_proxy by removing pass/set-authorization-token?

prydonius:
> If I remove that flag the behavior of gatekeeper and oauth2_proxy is the same.

If this is the case, we don't need the special regex, right? Should we remove it?

PR #965, opened 2019-02-12
Bump nginx image version for security vulns

prydonius:
It appears the older image contains some security vulnerabilities that have been fixed in more recent Debian versions. This PR bumps the nginx image used in the dashboard to upgrade the underlying Debian OS.


prydonius:
FYI, I think we may be able to wait until next week when we cut a release to also release this, in the meantime people wanting to use a more up-to-date version of nginx can use the latest image. Does that sound fine?

PR #967, opened 2019-02-20
Document how to use OIDC authentication

### LINKED ISSUE #947: Documentation on how to setup an OIDC proxy with Kubeapps
Related to #385, #931 and dependent on #945. We should add documentation on how someone can make use of #945 by configuring a compliant OIDC proxy to authenticate with Kubeapps.

We can assume that the user already has OIDC configured with the Kubernetes apiserver, and just needs to configure a proxy with the same client id/secret.

### LINKED ISSUE #948: Test OIDC support against different providers
Related to #385, #931 and dependent on #945. We should test a set of popular Kubernetes OIDC provider configurations and document what we know to work. For example, we should test:

- Dex
- Keycloak
- ActiveDirectory with AKS

I've only listed ones that are specific to people setting up against their Kubernetes clusters, feel free to add any others that make sense.

andresmgot:
Closes #947 
Closes #948


docs/user/login-alternatives.md
```diff
@@ -0,0 +1,36 @@
+# Kubeapps Login Form
+
+By default, when first visiting Kubeapps, a login form is shown for the user to introduce a personal token for a specific service account:
```
prydonius: we can just say a Kubernetes API token, it doesn't have to be a service account token

docs/user/login-alternatives.md
```diff
@@ -0,0 +1,36 @@
...
+
+![Dashboard Login](../img/dashboard-login.png)
+
+You can find more information about access control in Kubeapps in this [document](./access-control.md).
```
prydonius: This line doesn't feel very relevant here and it's already being linked to from other places, so I think we can skip it?

docs/user/login-alternatives.md
```diff
@@ -0,0 +1,36 @@
...
+
+However, it's possible to disable this form or delegate the authentication to an OIDC provider so Kubeapps users don't need to introduce a token in the login form.
+
+## Bypassing authentication
+
+Kubeapps expects an `Authorization` header that will be used to validate operations agains the Kubernetes API. This is usually set with the login form but if Kubeapps is exposed with an Ingress object, it's possible to hardcode a valid token in the Ingress configuration and automatically include it in all the requests.
```
prydonius:
```suggestion
Kubeapps expects an `Authorization` header that will be used to validate operations against the Kubernetes API. This is usually set with the login form but if Kubeapps is exposed with an Ingress object, it's possible to hardcode a valid token in the Ingress configuration and automatically include it in all the requests.
```

docs/user/login-alternatives.md
```diff
@@ -0,0 +1,36 @@
...
+
+**NOTE**: This is not suitable for production since anyone with access to Kubeapps would be granted with the permissions associated with the hardcoded token.
```
prydonius:
```suggestion
**NOTE**: This is not recommended in production since anyone with access to Kubeapps would be granted with the permissions associated with the hardcoded token.
```

I think "recommended" might be a better word to use here.

docs/user/login-alternatives.md
prydonius: We could also rephrase this a bit, to make it more succinct:

```suggestion
Kubeapps will skip the login form if it detects that an `Authorization` header is already being set by a reverse-proxy, such as an Ingress. An Ingress object can be configured to inject a hardcoded valid Kubernetes API token on every request to Kubeapps to force the login form to be skipped and enable all API requests to use the hardcoded token.
```

docs/user/login-alternatives.md
```diff
@@ -0,0 +1,36 @@
...
+    nginx.ingress.kubernetes.io/proxy-read-timeout: "600"
+    nginx.ingress.kubernetes.io/configuration-snippet: |
+      add_header Authorization "Bearer TOKEN";
+```
+
+You just need to substitute TOKEN with the actual value of the token. The above assumes an Nginx [Ingress Controller](https://kubernetes.io/docs/concepts/services-networking/ingress/#ingress-controllers), in case a different controller is being used that annotation will need to be adapted.
+
+# Using an OIDC provider
+
+In case you want to use OAuth 2.0 to authenticate Kubeapps users follow this [guide](./using-an-OIDC-provider.md).
```
prydonius:
```suggestion
In case you want to use OpenID Connect to authenticate Kubeapps users, and your Kubernetes API server is configured to use the same OIDC provider, follow this [guide](./using-an-OIDC-provider.md).
```

docs/user/login-alternatives.md
```diff
@@ -0,0 +1,36 @@
+# Kubeapps Login Form
```
prydonius: perhaps this file should be `login-bypass.md`?

docs/user/using-an-OIDC-provider.md
```diff
@@ -0,0 +1,162 @@
+# Using an OIDC Provider with Kubeapps
+
+OpenID Connect (OIDC) is a simple identity layer on top of the OAuth 2.0 protocol, which allows computing clients to verify the identity of an end-user based on the authentication performed by an authorization server, as well as to obtain basic profile information about the end-user.
```
prydonius: computing clients?

prydonius: simpler:
an end-user -> a user

docs/user/using-an-OIDC-provider.md
```diff
@@ -0,0 +1,162 @@
...
+
+For Kubeapps to use an Identity Provider it's necessary to configure at least the following parameters:
+
+- Client ID: Client ID of the IdP.
+- Client Secret: (If configured) Secret used to validate the Client ID.
+- OIDC Issuer URL (or discovery URL): URL of the OIDC issuer.
+
+**Note**: Depending on the configuration of the Identity Provider more parameters may be needed.
+
+In the next sections we will explain how you can find the parameters above for some of the identity providers tested.
```
prydonius: Is this necessary? We're assuming they've already configured their Kubernetes clusters correctly, and so should already know these values.

docs/user/using-an-OIDC-provider.md
```diff
@@ -0,0 +1,162 @@
+# Using an OIDC Provider with Kubeapps
+
+OpenID Connect (OIDC) is a simple identity layer on top of the OAuth 2.0 protocol, which allows computing clients to verify the identity of an end-user based on the authentication performed by an authorization server, as well as to obtain basic profile information about the end-user.
+
+It is possible to configure your Kubernetes cluster to use an OIDC provider in order to manage accounts, groups and roles with a single application. This guide will explain how you can use an existing OIDC provider to authenticate users within Kubeapps.
+
+## Pre-requisites
+
+For this guide we assume that you have a Kubernetes cluster that is properly configured to use an Identity Provider to handle the authorization of your cluster. You can find more information about how Kubernetes uses OIDC tokens [here](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#openid-connect-tokens). This means that the Kubernetes API server should be configured to use that OIDC provider.
```
prydonius:
```suggestion
For this guide we assume that you have a Kubernetes cluster that is properly configured to use an Identity Provider (IdP) to handle the authorization of your cluster. You can find more information about how Kubernetes uses OIDC tokens [here](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#openid-connect-tokens). This means that the Kubernetes API server should be configured to use that OIDC provider.
```

since we use the shorthand below, it's good to point it out here

docs/user/using-an-OIDC-provider.md
```diff
@@ -0,0 +1,162 @@
...
+      labels:
+        name: kubeapps-auth-proxy
+    spec:
+      containers:
+      - args:
+        - --client-id=$AUTH_PROXY_CLIENT_ID
+        - --client-secret=$AUTH_PROXY_CLIENT_SECRET
+        - --discovery-url=$AUTH_PROXY_DISCOVERY_URL
+        - --skip-openid-provider-tls-verify
+        - --secure-cookie=false
```
prydonius: I guess, by default, this should be true?

docs/user/using-an-OIDC-provider.md
```diff
@@ -0,0 +1,162 @@
...
+    metadata:
+      labels:
+        name: kubeapps-auth-proxy
+    spec:
+      containers:
+      - args:
+        - --client-id=$AUTH_PROXY_CLIENT_ID
+        - --client-secret=$AUTH_PROXY_CLIENT_SECRET
+        - --discovery-url=$AUTH_PROXY_DISCOVERY_URL
+        - --skip-openid-provider-tls-verify
```
prydonius: Likewise as below, let's not optimise for non-production environments.

docs/user/using-an-OIDC-provider.md
```diff
@@ -0,0 +1,162 @@
...
+    name: kubeapps-auth-proxy
+  sessionAffinity: None
+  type: ClusterIP
+EOF
+```
+
+The above is a sample deployment, depending on the configuration of the Identity Provider those flags may vary. For this example we use:
+
+- `--client-id`, `--client-secret` and `--discovery-url`: Client ID, Secret and IdP URL as stated in the section above.
+- `--skip-openid-provider-tls-verify`, `--secure-cookie=false`: If the `discovery-url` is served through HTTPS with a self-signed certificate (i.e. Keycloak or Dex), those flags are necessary to avoid errors while validating the TLS certificate.
```
prydonius: I don't think this is important enough for us to mention, we can point to the Keycloak gatekeeper docs for more configuration options.

docs/user/using-an-OIDC-provider.md
```diff
@@ -0,0 +1,162 @@
...
+- `--upstream-url`: Internal URL for the `kubeapps` service.
+- `--resources=uri=/api/kube/*|white-listed=true`: In order to use WebSockets we need to bypass the authentication for those request. In that case, Kubeapps will inject the `Bearer` token manually.
```
prydonius:
```suggestion
- `--resources=uri=/api/kube/*|white-listed=true`: This setting enables WebSockets to work correctly, which Kubeapps relies on to show up-to-date information. Kubeapps handles the injection of the OIDC token into every Kubernetes API request, so it is not necessary for Keycloak Gatekeeper to do it.
```

docs/user/using-an-OIDC-provider.md
```diff
@@ -0,0 +1,162 @@
...
+- `listen=0.0.0.0:3000`: Listen in all the interfaces.
+
+## Exposing the proxy
+
+Once the proxy is in place and it's able to connect to the IdP we will need to expose it to access it as the main endpoint for Kubeapps (instead of the `kubeapps` service). We can do that with an Ingress object. Note that for doing so an [Ingress Controller](https://kubernetes.io/docs/concepts/services-networking/ingress/#ingress-controllers) is needed. There are also other methods to expose the `kubeapps-auth-proxy` service, for example using `LoadBalancer` as type in a cloud environment. In case an Ingress is used, remember to modify the host `kubeapps.local` for the value that you want to use as hostname for your application:
```
prydonius:
```suggestion
Once the proxy is in place and it's able to connect to the IdP we will need to expose it to access it as the main endpoint for Kubeapps (instead of the `kubeapps` service). We can do that with an Ingress object. Note that for doing so an [Ingress Controller](https://kubernetes.io/docs/concepts/services-networking/ingress/#ingress-controllers) is needed. There are also other methods to expose the `kubeapps-auth-proxy` service, for example using `LoadBalancer` as type in a cloud environment. In case an Ingress is used, remember to modify the host `kubeapps.local` for the value that you want to use as a hostname for Kubeapps:
```

docs/user/using-an-OIDC-provider.md
```diff
@@ -0,0 +1,162 @@
...
+    http:
+      paths:
+      - backend:
+          serviceName: kubeapps-auth-proxy
+          servicePort: 3000
+        path: /
+EOF
+```
+
+Once you do that you can access the application, you will be prompted to introduce your credentials:
```
prydonius: I would make this more clear by saying something like, once you access the proxy you will be redirected to login with your Identity Provider. Once successfully logged in, you will be redirected to Kubeapps.

prydonius:
Thanks a lot for this, and +1 for adding the documentation about a hardcoded token! I mostly have some rephrasing suggestions.

docs/user/using-an-OIDC-provider.md
andresmgot: simplified to `clients`

docs/user/using-an-OIDC-provider.md
andresmgot: (replied in the comment below)

andresmgot:
I think I have addressed your comments @prydonius, let me know if I missed something

docs/user/using-an-OIDC-provider.md
prydonius: > Do you mean that you wouldn't show the explanation of these 2 flags or all the flags?

These two flags.

> It's true that we should not include it by default though so I will move the explanation of these flags to a note.

That sounds okay, but honestly the less we document the better, it's less work to maintain. We could just as easily point to the keycloak-gateway docs and hopefully someone finds it there.

docs/user/using-an-OIDC-provider.md
```diff
@@ -0,0 +1,161 @@
...
+```
+
+The above is a sample deployment, depending on the configuration of the Identity Provider those flags may vary. For this example we use:
+
+- `--client-id`, `--client-secret` and `--discovery-url`: Client ID, Secret and IdP URL as stated in the section above.
+- `--upstream-url`: Internal URL for the `kubeapps` service.
+- `--resources=uri=/api/kube/*|white-listed=true`: This setting enables WebSockets to work correctly, which Kubeapps relies on to show up-to-date information. Kubeapps handles the injection of the OIDC token into every Kubernetes API request, so it is not necessary for Keycloak Gatekeeper to do it.
+- `listen=0.0.0.0:3000`: Listen in all the interfaces.
+
+**NOTE**: If the identity provider is deployed with a self-signed certificate (which may be the case for Keycloak or Dex) you will need to disable the TLS and cookie verification. For doing so you can add the flags `--skip-openid-provider-tls-verify` and `--secure-cookie=false` to the deployment above. You can find more options for the `keycloak-gatekeeper` proxy [here](https://www.keycloak.org/docs/latest/securing_apps/index.html#_keycloak_generic_adapter).
```
prydonius: great, thanks 👍

docs/user/using-an-OIDC-provider.md
```diff
@@ -0,0 +1,161 @@
...
+          servicePort: 3000
+        path: /
+EOF
+```
+
+Once you create the ingress rule and you access the proxy, you will be redirected to your IdP. There, you will be prompted to introduce your credentials:
+
+![Proxy Login](../img/auth-proxy-login.png)
+
+After successfully introduce the credentials, you will be finally redirected to Kubeapps.
```
prydonius:
```suggestion
After successfully logging in with your IdP, you will be redirected to Kubeapps and be authenticated with your user's OIDC token.
```

PR #976, opened 2019-02-25
Use bitnami images before updating bitnami chart

andresmgot:
It's now a requirement to use bitnami images wherever is possible for charts published at https://github.com/bitnami/charts.

Note that bitnami images are mirrored versions from the ones we are currently using but their tags doesn't contain the prefix `v`: `v1.3.0 -> 1.3.0`. They are hosted in the docker hub and have the prefix `kubeapps-<name of the image>`.

This is an example of changes performed from the dev version of the chart in this repository to the one that will be submitted to the bitnami repository:

```patch
@@ -94,13 +94,13 @@ apprepository:
   replicaCount: 1
   image:
     registry: docker.io
-    repository: kubeapps/apprepository-controller
-    tag: latest
+    repository: bitnami/kubeapps-apprepository-controller
+    tag: 1.3.0
   # Image used to perform chart repository syncs
   syncImage:
-    registry: quay.io
-    repository: helmpack/chart-repo
-    tag: v1.4.0
+    registry: docker.io
+    repository: bitnami/kubeapps-chart-repo
+    tag: 1.4.0
   initialRepos:
   - name: stable
     url: https://kubernetes-charts.storage.googleapis.com
@@ -149,8 +149,8 @@ tillerProxy:
   replicaCount: 2
   image:
     registry: docker.io
-    repository: kubeapps/tiller-proxy
-    tag: latest
+    repository: bitnami/kubeapps-tiller-proxy
+    tag: 1.3.0
   service:
     port: 8080
   host: tiller-deploy.kube-system:44134
@@ -175,9 +175,9 @@ tillerProxy:
 chartsvc:
   replicaCount: 2
   image:
-    registry: quay.io
-    repository: helmpack/chartsvc
-    tag: v1.4.0
+    registry: docker.io
+    repository: bitnami/kubeapps-chartsvc
+    tag: 1.4.0
   service:
     port: 8080
   # https://github.com/kubeapps/kubeapps/issues/478#issuecomment-422979262
@@ -211,8 +211,8 @@ dashboard:
   replicaCount: 2
   image:
     registry: docker.io
-    repository: kubeapps/dashboard
-    tag: latest
+    repository: bitnami/kubeapps-dashboard
+    tag: 1.3.0
   service:
     port: 8080
   livenessProbe:
```

I have manually tested the changes and I didn't found any issue with the mirrored images.

### DIFF
--- a/script/chart_sync_utils.sh
+++ b/script/chart_sync_utils.sh
@@ -58,6 +58,13 @@ updateRepo() {
     rm "${targetChartPath}/Chart.yaml.bk"
     # DANGER: This replaces any tag marked as latest in the values.yaml
     sed -i.bk 's/tag: latest/tag: '"${targetTag}"'/g' "${targetChartPath}/values.yaml"
+    # Replace quay.io images for docker.io
+    sed -i.bk 's/registry: quay.io/registry: docker.io/g' "${targetChartPath}/values.yaml"
+    # Remove v prefix from tags
+    sed -i.bk 's/tag: v\(.*\)/tag: \1/g' "${targetChartPath}/values.yaml"
+    # Use bitnami images
+    sed -i.bk 's/repository: helmpack\/\(.*\)/repository: bitnami\/kubeapps-\1/g' "${targetChartPath}/values.yaml"
+    sed -i.bk 's/repository: kubeapps\/\(.*\)/repository: bitnami\/kubeapps-\1/g' "${targetChartPath}/values.yaml"
     rm "${targetChartPath}/values.yaml.bk"
 }
 
--- a/script/chart_sync_utils.sh
+++ b/script/chart_sync_utils.sh
@@ -57,13 +57,9 @@ updateRepo() {
     sed -i.bk 's/appVersion: DEVEL/appVersion: '"${targetTag}"'/g' "${chartYaml}"
     rm "${targetChartPath}/Chart.yaml.bk"
     # DANGER: This replaces any tag marked as latest in the values.yaml
-    sed -i.bk 's/tag: latest/tag: '"${targetTag}"'/g' "${targetChartPath}/values.yaml"
-    # Replace quay.io images for docker.io
-    sed -i.bk 's/registry: quay.io/registry: docker.io/g' "${targetChartPath}/values.yaml"
-    # Remove v prefix from tags
-    sed -i.bk 's/tag: v\(.*\)/tag: \1/g' "${targetChartPath}/values.yaml"
+    local tagWithoutV=$(echo $targetTag | tr -d v)
+    sed -i.bk 's/tag: latest/tag: '"${tagWithoutV}"'/g' "${targetChartPath}/values.yaml"
     # Use bitnami images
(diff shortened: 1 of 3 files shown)


script/chart_sync_utils.sh
```diff
@@ -58,6 +58,13 @@ updateRepo() {
     rm "${targetChartPath}/Chart.yaml.bk"
     # DANGER: This replaces any tag marked as latest in the values.yaml
     sed -i.bk 's/tag: latest/tag: '"${targetTag}"'/g' "${targetChartPath}/values.yaml"
+    # Replace quay.io images for docker.io
+    sed -i.bk 's/registry: quay.io/registry: docker.io/g' "${targetChartPath}/values.yaml"
+    # Remove v prefix from tags
+    sed -i.bk 's/tag: v\(.*\)/tag: \1/g' "${targetChartPath}/values.yaml"
```
prydonius: We might want to be more careful here and only change the tag of images that match the `repository: (helmpack|kubeapps)\/` on the previous line. See https://stackoverflow.com/questions/18620153/find-matching-text-and-replace-next-line for an example of how to do this with sed.

That said, it might not be necessary since we expect to do this for all images once keycloak-gatekeeper has a corresponding bitnami image.

script/chart_sync_utils.sh
```diff
@@ -58,6 +58,13 @@ updateRepo() {
...
+    # Use bitnami images
+    sed -i.bk 's/repository: helmpack\/\(.*\)/repository: bitnami\/kubeapps-\1/g' "${targetChartPath}/values.yaml"
```
prydonius: we could combine these with `...repository: (helmpack|kubeapps)\/...`, but not that important as again, we will probably just apply this chart-wide once we have the bitnami version of keycloak-gatekeeper.

script/chart_sync_utils.sh
andresmgot: the problem is that the regex gets a bit complicated:
```
    sed -i.bk '/repository: \(helmpack\|kubeapps\)\/\(.*\)/{n;s/tag: v\(.*\)/tag: \1/g}' "${targetChartPath}/values.yaml"
```

Actually it will be simple if we directly use bitnami images for `helmpack` images. Since we are not publishing directly in that repo we can just use the bitnami one in master.

script/chart_sync_utils.sh
```diff
@@ -57,7 +57,10 @@ updateRepo() {
     sed -i.bk 's/appVersion: DEVEL/appVersion: '"${targetTag}"'/g' "${chartYaml}"
     rm "${targetChartPath}/Chart.yaml.bk"
     # DANGER: This replaces any tag marked as latest in the values.yaml
-    sed -i.bk 's/tag: latest/tag: '"${targetTag}"'/g' "${targetChartPath}/values.yaml"
+    local tagWithoutV=$(echo $targetTag | tr -d v)
+    sed -i.bk 's/tag: latest/tag: '"${tagWithoutV}"'/g' "${targetChartPath}/values.yaml"
```
prydonius: nice solution 👌

PR #978, opened 2019-02-28
corrects names of chart AppRepository roles to match documentation

prydonius:
also fixes an issue where the write role was being created even if
rbac.create=false

### DIFF
--- a/chart/kubeapps/Chart.yaml
+++ b/chart/kubeapps/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kubeapps
-version: 1.5.1
+version: 2.0.0
 appVersion: DEVEL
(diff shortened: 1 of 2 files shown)


chart/kubeapps/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kubeapps
-version: 1.5.1
+version: 2.0.0
```
prydonius: this is going to be a breaking change for anyone using these in RoleBindings, so bumping the major version as per semver

andresmgot: Since the name is not usable right now, can we consider this a bug fix? People is going to be confused if we bump the major version because of this.

prydonius:
Since this change was introduced fairly recently, I agree that a patch bump makes sense.

PR #980, opened 2019-03-05
adds a note about access URLs taking time and minikube instructions

### LINKED ISSUE #953: Improve UX/docs for why Services might be stuck in pending state
Related to #946.

Sometimes, cloud provisioners can take a long time to provision and IP address for a Service LoadBalancer. Even when an application appears to be deployed (pods are running), the Service can stay in a pending state for a while and this causes confusion for Kubeapps users.

Additionally, for those testing Kubeapps on Minikube, the provisioner will never assign an IP address unless `minikube tunnel` is running.

We can look at improving the UX here by either adding an explanatory tooltip next to the pending state explaining that LoadBalancer Services can take some time to provision and to enable the Minikube tunnel if using Minikube.

prydonius:
also updates the WordPress search image to a smaller image and displays
the search feature.

fixes #953


docs/user/getting-started.md
```diff
@@ -101,7 +101,7 @@ Once you have the Kubeapps Dashboard up and running, you can start deploying app
 
 To obtain the WordPress username and password, refer to the "Notes" section of the deployment page, which contains the commands you will need to run to obtain the credentials for the deployment.
 
-You can also use the URLs shown to directly access the application.
+You can also use the URLs shown to directly access the application. Note that, depending on your cloud provider of choice, it may take some time for an access URL to be available for the application. If using Minikube, you will need to run `minikube tunnel` in your terminal in order for an IP address to be assigned to your application.
```
andresmgot: I would emphasize that the URL field will remain "Pending", hopefully an user googling that will reach this point.

prydonius:
@andresmgot updated to specifically mention the Pending state.

PR #982, opened 2019-03-05
Use Bitnami image for keycloak-gatekeeper

andresmgot:
Related to https://github.com/kubeapps/kubeapps/pull/976


chart/kubeapps/values.yaml
```diff
@@ -292,8 +292,8 @@ authProxy:
   # Image used for the proxy
   image:
     registry: docker.io
-    repository: keycloak/keycloak-gatekeeper
-    tag: 4.6.0.Final
+    repository: bitnami/keycloak-gatekeeper
+    tag: 2.3.0-r0
```
prydonius: why the version difference?

andresmgot: Because for the images they were using this versioning (I guess it's the version of Keycloak itself):
https://github.com/jboss-dockerfiles/keycloak/releases

But the latest release for keycloak-gatekeeper is that one: 
https://github.com/keycloak/keycloak-gatekeeper/releases

prydonius: Weird, thanks!

prydonius:
@andresmgot I'm tempted to wait until the `scratch` issue gets resolved, right now this image is not better or at least parity with the upstream image.

andresmgot:
Agree, I think there is no urge for us to merge this.

chart/kubeapps/templates/kubeapps-frontend-deployment.yaml
```diff
@@ -39,6 +39,7 @@ spec:
     {{- if .Values.authProxy.enabled }}
       - name: auth-proxy
         args:
+        - keycloak-proxy
```
juan131: You don't need this change using `bitnami/keycloak-gatekeeper:2.3.0-scratch-r1` on

andresmgot: actually it's needed. And now the path of the binary `/keycloack-proxy` is not in the path so I needed to specify the full path. Note that for a pod to inherit a "command" you need to set it as `ENTRYPOINT` of the image (not `CMD` what may seems the logic approach).

chart/kubeapps/values.yaml
```diff
@@ -292,8 +292,8 @@ authProxy:
   # Image used for the proxy
   image:
     registry: docker.io
-    repository: keycloak/keycloak-gatekeeper
-    tag: 4.6.0.Final
+    repository: bitnami/keycloak-gatekeeper
+    tag: 2.3.0-scratch-r1
```
prydonius: is `scratch-r1` necessary, looks like `2.3.0-r1` should just point to the scratch-based image, right?

andresmgot: that's right, that tag is the one that @juan131 pointed me to but it's true that `2.3.0-r1` also uses from scratch.

PR #985, opened 2019-03-08
switch to using LoadingWrapper instead of LoadingSpinner

### LINKED ISSUE #925: ConfirmDialog should use LoadingWrapper instead of LoadingSpinner
https://github.com/kubeapps/kubeapps/blob/master/dashboard/src/components/ConfirmDialog/ConfirmDialog.tsx#L50

We should use LoadingWrapper everywhere as the implementation of LoadingSpinner could change (e.g. might move to Hex React lib), so using LoadingWrapper allows us to update the underlying spinner in a single place.

prydonius:
fixes #925

### DIFF
--- a/dashboard/src/components/ConfirmDialog/ConfirmDialog.test.tsx
+++ b/dashboard/src/components/ConfirmDialog/ConfirmDialog.test.tsx
@@ -1,0 +1,12 @@
+import context from "jest-plugin-context";
+
+import ConfirmDialog from ".";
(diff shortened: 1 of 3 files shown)


andresmgot:
Can we add at least a snapshot test? Using the `itBehavesLike("aLoadingComponent"` formula should be easy.

prydonius:
Yeah, I should have - was being lazy. Done now, thanks!

PR #986, opened 2019-03-11
Refactor upgrade flow

### LINKED ISSUE #969: Choosing the wrong repo can cause a crash in the upgrade flow
When upgrading an application, if a repository that doesn't contain the version of the current application the Dashboard shows a wrong error. Steps to reproduce it:

 - Deploy latest Wordpress chart from stable repository
 - Add azure repository
 - Click on upgrade for Wordpress
 - Choose the azure repository

screenshot from 2019-02-21 16-47-47

andresmgot:
Fixes #969 

The goal of this PR is to avoid the issue described in #969. Now, when clicking in the "Upgrade" button, we check if the application has `updateInfo`. In that case we skip the `SelectRepoForm` and we automatically choose the repo with the latest version.

### DIFF
--- a/dashboard/src/actions/apps.ts
+++ b/dashboard/src/actions/apps.ts
@@ -81,25 +81,36 @@ function getAppUpdateInfo(
   appVersion: string,
 ): ThunkAction<Promise<void>, IStoreState, null, AppsAction> {
   return async dispatch => {
-    const chartsInfo = await Chart.listWithFilters(chartName, currentVersion, appVersion);
-    let updateInfo: IChartUpdateInfo = {
-      repository: { name: "", url: "" },
-      latestVersion: "",
-    };
-    chartsInfo.forEach(c => {
-      const chartLatestVersion = c.relationships.latestChartVersion.data.version;
-      if (semver.gt(chartLatestVersion, currentVersion)) {
-        if (updateInfo.latestVersion && semver.gt(updateInfo.latestVersion, chartLatestVersion)) {
-          // The current update is newer than the chart version, do nothing
-        } else {
-          updateInfo = {
-            latestVersion: chartLatestVersion,
-            repository: c.attributes.repo,
-          };
-        }
+    dispatch(requestApps());
+    try {
+      const chartsInfo = await Chart.listWithFilters(chartName, currentVersion, appVersion);
+      let updateInfo: IChartUpdateInfo = {
+        upToDate: true,
+        repository: { name: "", url: "" },
+        latestVersion: "",
+      };
+      if (chartsInfo.length > 0) {
+        // Initialize updateInfo with the first chart found
+        updateInfo = {
+          upToDate: true,
+          latestVersion: chartsInfo[0].relationships.latestChartVersion.data.version,
+          repository: chartsInfo[0].attributes.repo,
+        };
+        chartsInfo.forEach(c => {
+          const chartLatestVersion = c.relationships.latestChartVersion.data.version;
+          if (semver.gt(chartLatestVersion, updateInfo.latestVersion)) {
+            updateInfo.latestVersion = chartLatestVersion;
+            updateInfo.repository = c.attributes.repo;
+          }
+          if (semver.gt(chartLatestVersion, currentVersion)) {
+            updateInfo.upToDate = false;
+          }
+        });
       }
-    });
-    dispatch(receiveAppUpdateInfo({ releaseName, updateInfo }));
+      dispatch(receiveAppUpdateInfo({ releaseName, updateInfo }));
+    } catch (e) {
+      dispatch(errorApps(e));
+    }
   };
 }
 
--- a/dashboard/src/actions/apps.ts
+++ b/dashboard/src/actions/apps.ts
@@ -28,6 +28,8 @@ export const receiveAppList = createAction("RECEIVE_APP_LIST", resolve => {
   return (apps: IAppOverview[]) => resolve(apps);
 });
 
+export const requestAppUpdateInfo = createAction("REQUEST_APP_UPDATE_INFO");
+
 export const receiveAppUpdateInfo = createAction("RECEIVE_APP_UPDATE_INFO", resolve => {
   return (payload: { releaseName: string; updateInfo: IChartUpdateInfo }) => resolve(payload);
 });
@@ -49,6 +51,7 @@ const allActions = [
   requestApps,
   receiveApps,
   receiveAppList,
+  requestAppUpdateInfo,
   receiveAppUpdateInfo,
   errorApps,
   errorDeleteApp,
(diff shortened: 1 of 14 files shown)


dashboard/src/actions/apps.ts
```diff
@@ -81,25 +81,36 @@ function getAppUpdateInfo(
...
-          updateInfo = {
-            latestVersion: chartLatestVersion,
-            repository: c.attributes.repo,
-          };
-        }
+    dispatch(requestApps());
+    try {
+      const chartsInfo = await Chart.listWithFilters(chartName, currentVersion, appVersion);
+      let updateInfo: IChartUpdateInfo = {
+        upToDate: true,
```
andresmgot: previously, the condition to discern if an app has an update was to check if `latestVersion` wasn't empty. Now there is an explicit parameter for that since `latestVersion` will be populated in most of the cases, even if it's equal to the current version.

dashboard/src/components/AppUpgrade/AppUpgrade.tsx
```diff
@@ -41,13 +47,17 @@ interface IAppUpgradeState {
...
+    const { releaseName, getAppWithUpdateInfo, namespace, fetchRepositories } = this.props;
+    getAppWithUpdateInfo(releaseName, namespace);
     fetchRepositories();
   }
 
   public render() {
-    const { app, repos, error, namespace, releaseName, repoError } = this.props;
+    const { app, repos, error, namespace, releaseName, repoError, isFetching } = this.props;
+    let { repo } = this.props;
+    if (isFetching) {
```
andresmgot: it's necessary to check if we are still fetching things to avoid the `SelectRepoForm` to be momentarily shown.

dashboard/src/components/AppUpgrade/AppUpgrade.tsx
```diff
@@ -77,11 +87,31 @@ class AppUpgrade extends React.Component<IAppUpgradeProps, IAppUpgradeState> {
...
         );
+      } else if (repos.length === 0) {
+        return (
+          <ErrorSelector
+            error={new Error("Unable to any repo to upgrade from")}
+            resource={releaseName}
+          />
+        );
       } else {
-        return <LoadingWrapper />;
```
andresmgot: This call would show the loading wrapper forever since we have already fetched all the elements but we still don't have the info needed so I have added an error case (respos === []) and an unexpected error by default.

dashboard/src/actions/apps.ts
```diff
@@ -81,25 +84,36 @@ function getAppUpdateInfo(
...
+        latestVersion: "",
+      };
+      if (chartsInfo.length > 0) {
+        // Initialize updateInfo with the first chart found
+        updateInfo = {
+          upToDate: true,
+          latestVersion: chartsInfo[0].relationships.latestChartVersion.data.version,
+          repository: chartsInfo[0].attributes.repo,
+        };
+        chartsInfo.forEach(c => {
```
prydonius: The first run through of this is not going to be useful since we already have `updateInfo` set to it, so it might make sense to drop the first element (e.g. with a for-loop `for (let i=1; i < chartsInfo.length; i++) {...}`).

That said, a cleaner way to do this might be to sort the array by the latest version. I haven't tried it, but something like this should work:

```
chartsInfo.sort((a, b) => semver.compare(a.relationships.latestChartVersion.data.version, b.relationships.latestChartVersion.data.version))
```

and then initialise updateInfo to the first element of the sorted array. 

Note that `semver.compare` above follows the conditions that the [`Array.sort`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#Description) expects for the comparator function:

> compare(v1, v2): Return 0 if v1 == v2, or 1 if v1 is greater, or -1 if v2 is greater. Sorts in ascending order if passed to Array.sort().

dashboard/src/components/AppUpgrade/AppUpgrade.tsx
prydonius: Is the last case (unexpected error by default) at all possible? It seems to me like we'd never reach this code? If we will, do we have a test case for it?

dashboard/src/components/AppUpgrade/AppUpgrade.tsx
```diff
@@ -77,11 +87,33 @@ class AppUpgrade extends React.Component<IAppUpgradeProps, IAppUpgradeState> {
             resource="App Repositories"
           />
         );
+      } else if (repos.length === 0) {
```
prydonius: I couldn't see a test case for when repos are empty, can we add that if it isn't there?

dashboard/src/components/AppUpgrade/AppUpgrade.tsx
```diff
@@ -77,11 +87,33 @@ class AppUpgrade extends React.Component<IAppUpgradeProps, IAppUpgradeState> {
...
+        return (
+          <ErrorSelector
+            error={new Error("Unable to any repo to upgrade from")}
```
prydonius: We should use an error similar to https://github.com/kubeapps/kubeapps/blob/714fff14ef4fe58358b51add4310ceb9a95f66fc/dashboard/src/components/Catalog/Catalog.tsx#L51-L60 in this case, so we can prompt users to adding repos.

dashboard/src/components/AppView/AppControls/AppControls.test.tsx
```diff
@@ -137,3 +139,21 @@ context("when the application has been already deleted", () => {
...
+
+context("when there is a new version available", () => {
+  it("should forward the latest version", () => {
+    const name = "foo";
+    const namespace = "bar";
+    const app = {
+      name,
+      namespace,
+      updateInfo: {
+        upToDate: false,
```
prydonius: can we add a test for when this is `true`?

dashboard/src/components/AppUpgrade/AppUpgrade.tsx
prydonius: Hmm, I almost wonder if it would be better to just render an Unexpected error in that case? I guess there would be an issue with Tiller or a bug in Kubeapps if we get to this point, and I'm not sure that the proposed error message "Unable to obtain the required information to upgrade" is going to be very useful to the user. The Unexpected error directs users to file an issue in GitHub if they think they encountered a bug.

andresmgot: Yes, the default case for `ErrorSelector` is to use the `UnexpectedError` component. Here, since the error type is generic, it shows the generic message but we include a little bit more of info:

Screenshot from 2019-03-13 10-55-17

andresmgot:
Merging this. I am using the Unexpected Error but using the ErrorSelector interface.

dashboard/src/components/AppUpgrade/AppUpgrade.tsx
prydonius: Ah I see, thanks for the explanation!

PR #987, opened 2019-03-11
Revamp deploy button

### LINKED ISSUE #984: Change "Deploy" button text
The current text states "Deploy with Helm" and depending on the width of the browser the text gets collapsed to "Deploy ...". We can just replace that text with "Deploy" since we don't support any different deployment method.

andresmgot:
Fixes #984 

I suggest to (apart than changing the text of the button) move the Deploy button to the title just like the AppList and ServiceList views have it:

Screenshot from 2019-03-12 15-18-38

### DIFF
--- a/dashboard/src/components/ChartView/ChartHeader.scss
+++ b/dashboard/src/components/ChartView/ChartHeader.scss
@@ -1,10 +1,10 @@
-.ChartHeader_img {
+.ChartHeader__icon {
   display: flex;
   align-items: center;
   justify-content: center;
 }
 
-.ChartHeader_button {
+.ChartHeader__button {
   display: flex;
   flex-direction: column;
   justify-content: flex-end;
--- a/dashboard/src/components/ChartView/ChartHeader.scss
+++ b/dashboard/src/components/ChartView/ChartHeader.scss
@@ -1,0 +1,11 @@
+.ChartHeader_img {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+
+.ChartHeader_button {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+}
--- a/dashboard/src/components/ChartView/ChartHeader.scss
+++ b/dashboard/src/components/ChartView/ChartHeader.scss
@@ -7,5 +7,5 @@
 .ChartHeader_button {
   display: flex;
   flex-direction: column;
-  justify-content: center;
+  justify-content: flex-end;
 }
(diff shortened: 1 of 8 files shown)


prydonius:
Thanks @andresmgot, I like the concept, but I'd be worried that that button is now harder to see. It might be better to position it next to the repo/name of the chart, or right under the description.

@earnubs any thoughts on this?

andresmgot:
For comparison, this is how it looked before:

Screenshot from 2019-03-12 13-42-22

earnubs:
I don't really have any thoughts other that to try aligning the button to the bottom of the containing element, not the middle; then the eye will scan along from the last sentence in the title to the button.

andresmgot:
I have updated the screenshot to move the button

prydonius:
That looks a tad better to me, thanks for the suggestion @earnubs. @Angelmmiguel I know we discussed this before as well (how it's not clear on this page that the "Deploy" is the next action we expect a user to make), so maybe you might have some input here too.

dashboard/src/components/ChartView/ChartHeader.scss
```diff
@@ -0,0 +1,11 @@
+.ChartHeader_img {
```
prydonius:
```suggestion
.ChartHeader__img {
```

dashboard/src/components/ChartView/ChartHeader.scss
```diff
@@ -0,0 +1,11 @@
...
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+
+.ChartHeader_button {
```
prydonius:
```suggestion
.ChartHeader__button {
```

dashboard/src/components/ChartView/ChartHeader.tsx
```diff
@@ -1,30 +1,42 @@
...
-          <ChartIcon icon={icon} />
-          <div className="title margin-l-small">
-            <h1 className="margin-t-reset">{id}</h1>
-            <h5 className="subtitle margin-b-normal">
-              {appVersion && <span>{appVersion} - </span>}
-              <Link to={`/catalog/${repo}`}>{repo}</Link>
-            </h5>
-            <h5 className="subtitle margin-b-reset">{description}</h5>
+        <div className="ChartView__heading margin-normal row">
+          <div className="col-1 ChartHeader_img">
```
prydonius: since we use the ChartIcon component, I would use `ChartHeader__icon` as the class name.

Angelmmiguel:
> That looks a tad better to me, thanks for the suggestion @earnubs. @Angelmmiguel I know we discussed this before as well (how it's not clear on this page that the "Deploy" is the next action we expect a user to make), so maybe you might have some input here too.

I believe we can have a meeting during the all-hands to work on this view. I remember we discussed about it doesn't look like a launch view. Also, it's not trivial to know you can change the version and then, deploy it.

Anyway, :+1: for this change! It makes much clear the CTA of this page :)

PR #989, opened 2019-03-12
Add global imagePullSecrets to overwrite any other existing one

carrodher:
Signed-off-by: Carlos Rodriguez Hernandez <crhernandez@bitnami.com>

#### What this PR does/why we need it:
This PR allows using a global parameter to indicate the array of secret names to use as Image Registry Secret.

Use:
```
$ kubectl create secret XXXX ...
$ helm install bitnami/kubeapps --name kubeapps --set global.imagePullSecrets=XXXX ...
```

Specify a secret globally for a specific registry, in this way, we are able to modify the registry and the secret for pulling images for all Charts globally without having to specify this per image


carrodher:
I just updated the PR fixing the issue in the tests, but it seems there are issues in GitHub:
Screen Shot 2019-03-12 at 16 07 02

The commit is: https://github.com/carrodher/kubeapps/commit/325f3734984b48406681667d0988833bd497235a

🤞 I hope above commit appears in the PR ...

chart/kubeapps/templates/_helpers.tpl
```diff
@@ -160,3 +160,74 @@ Frontend service port number
...
+but Helm 2.9 and 2.10 does not support it, so we need to implement this if-else logic.
+Also, we can not use a single if because lazy evaluation is not an option
+*/}}
+{{- if .Values.global }}
+{{- if .Values.global.imagePullSecrets }}
+imagePullSecrets:
+{{- range .Values.global.imagePullSecrets }}
+  - name: {{ . }}
+{{- end }}
+{{- else if or .Values.frontend.image.pullSecrets .Values.apprepository.image.pullSecrets .Values.apprepository.syncImage.pullSecrets .Values.hooks.image.pullSecrets .Values.tillerProxy.image.pullSecrets .Values.chartsvc.image.pullSecrets .Values.dashboard.image.pullSecrets .Values.authProxy.image.pullSecrets }}
```
andresmgot: I think we don't need to add support for a different `pullSecrets` per image if we don't have a specific use case for it. The code will get simplified a lot if we omit this part.

chart/kubeapps/values.yaml
```diff
@@ -58,6 +62,12 @@ frontend:
     registry: docker.io
     repository: bitnami/nginx
     tag: 1.14.0-r27
+    ## Optionally specify an array of imagePullSecrets.
```
andresmgot: the same here, we don't need all these repeated comments if we support only the global setting for the moment

chart/kubeapps/templates/_helpers.tpl
```diff
@@ -160,3 +160,22 @@ Frontend service port number
 8080
 {{- end -}}
 {{- end -}}
+
+{{/*
+Return the proper Docker Image Registry Secret Names
+*/}}
+{{- define "kubeapps.imagePullSecrets" -}}
+{{/*
+Helm 2.11 supports the assignment of a value to a variable defined in a different scope,
```
andresmgot: this comment doesn't apply now, isn't it?

andresmgot:
Thanks for the PR @carrodher

PR #995, opened 2019-03-18
Reload OIDC token when needed

### LINKED ISSUE #606: Standardize in one way to make async API calls
Currently we are using axios for k8s API calls but native fetch to talk to some of our services. 

I am assuming that we are using axios for API calls because we leverage its interceptor feature but we should look into ways to share its code.

That way we will be able to take advantage of the error handling that we have already in place.

### LINKED ISSUE #974: Auto-refresh OIDC token
Follow up from https://github.com/kubeapps/kubeapps/pull/961#issuecomment-463675313:

> Also, I tried to see what would happen if the tokens expired. I set the token expiration in Keycloak to 1 minute and saw that expired tokens don't reinitiate the OAuth flow when clicking around the app, AJAX requests are getting 307 redirects to start the OAuth flow, but it is blocked by CORS. Despite this, resources seem to load... Not sure what to do here or if we need to do anything, can you investigate?
>>
>>I have reduced the expiration time and I have tried 3 things:
>> - Wait until the token expires: When doing this, when changing the location (for example from the app list to a specific app) I get momentarily redirected to the `/login` page and the token gets refreshed and I am redirected again to the page I was meant to visit. I think that's the expected behavior:
>>screenshot from 2019-02-14 16-31-40
>> - Close the session of the user from keycloak: I can navigate the application until the token expires, then I start receiving 403 errors from the application but I don't get redirected. The application claims that I don't have sufficient permissions to request the resources. When reloading the page I get redirected again to the login page of keycloak:
>>screenshot from 2019-02-14 16-33-23
>> - Delete the user from keycloak: The same behavior than above.

I am no longer seeing the behavior above (I am not sure what changed). When a token expires I see the following error in the console:

```
Access to XMLHttpRequest at 'https://login.windows.net/9c701005-1e88-41d9-bc41-44f1c95615bd/oauth2/authorize?client_id=1f255319-3791-486b-bfbf-edc2c341bab9&redirect_uri=https%3A%2F%2Fkubeapps.local%2Foauth%2Fcallback&response_type=code&scope=openid+email+profile&state=42688e6d-0340-46d0-9f12-fd4b8450f8fc' (redirected from 'https://kubeapps.local/api/tiller-deploy/v1/namespaces/default/releases') from origin 'https://kubeapps.local' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.
```

And the application crashes when trying to load most of the resources. Refreshing the page or going to `/login` fixes the issue but we should improve the UX there.

andresmgot:
Fix #974 

With this PR we catch errors potentially thrown because the OIDC token is no longer valid. In that case we show a modal for the users  to notice that the application needs to reload. I used the Modal to avoid the user seing the application reloading without apparent reason.

opengl-rotating-triangle

Note that we are only checking request made by axios so to avoid issues I have replaced the current `fetch` calls with `axios.get` (so this PR also fixes #606).

### DIFF
--- a/dashboard/src/actions/auth.ts
+++ b/dashboard/src/actions/auth.ts
@@ -34,9 +34,7 @@ export function authenticate(
   };
 }
 
-export function logout(
-  needRefresh?: boolean,
-): ThunkAction<Promise<void>, IStoreState, null, AuthAction> {
+export function logout(): ThunkAction<Promise<void>, IStoreState, null, AuthAction> {
   return async dispatch => {
     Auth.unsetAuthToken();
     dispatch(setAuthenticated(false, false));
--- a/dashboard/src/actions/auth.ts
+++ b/dashboard/src/actions/auth.ts
@@ -14,16 +14,11 @@ export const authenticationError = createAction("AUTHENTICATION_ERROR", resolve 
   return (errorMsg: string) => resolve(errorMsg);
 });
 
-export const setSessionExpired = createAction("SET_AUTHENTICATION_SESSION_EXPIRED");
-export const unsetSessionExpired = createAction("UNSET_AUTHENTICATION_SESSION_EXPIRED");
+export const setSessionExpired = createAction("SET_AUTHENTICATION_SESSION_EXPIRED", resolve => {
+  return (sessionExpired: boolean) => resolve({ sessionExpired });
+});
 
-const allActions = [
-  setAuthenticated,
-  authenticating,
-  authenticationError,
-  setSessionExpired,
-  unsetSessionExpired,
-];
+const allActions = [setAuthenticated, authenticating, authenticationError, setSessionExpired];
 
 export type AuthAction = ActionType<typeof allActions[number]>;
 
@@ -38,7 +33,7 @@ export function authenticate(
       Auth.setAuthToken(token, oidc);
       dispatch(setAuthenticated(true, oidc));
       if (oidc) {
-        dispatch(unsetSessionExpired());
+        dispatch(setSessionExpired(false));
       }
     } catch (e) {
       dispatch(authenticationError(e.toString()));
@@ -56,7 +51,7 @@ export function logout(): ThunkAction<Promise<void>, IStoreState, null, AuthActi
 export function expireSession(): ThunkAction<Promise<void>, IStoreState, null, AuthAction> {
   return async dispatch => {
     if (Auth.usingOIDCToken()) {
-      dispatch(setSessionExpired());
+      dispatch(setSessionExpired(true));
     }
     dispatch(logout());
   };
--- a/dashboard/src/actions/auth.ts
+++ b/dashboard/src/actions/auth.ts
@@ -14,7 +14,9 @@ export const authenticationError = createAction("AUTHENTICATION_ERROR", resolve 
   return (errorMsg: string) => resolve(errorMsg);
 });
 
-const allActions = [setAuthenticated, authenticating, authenticationError];
+export const setExpiredSession = createAction("AUTHENTICATION_EXPIRED");
+
+const allActions = [setAuthenticated, authenticating, authenticationError, setExpiredSession];
 
 export type AuthAction = ActionType<typeof allActions[number]>;
 
@@ -37,6 +39,8 @@ export function authenticate(
 export function logout(): ThunkAction<Promise<void>, IStoreState, null, AuthAction> {
   return async dispatch => {
     Auth.unsetAuthToken();
+    // Expire the session if we are using OIDC tokens
+    dispatch(setExpiredSession());
     dispatch(setAuthenticated(false, false));
   };
 }
--- a/dashboard/src/actions/auth.ts
+++ b/dashboard/src/actions/auth.ts
@@ -14,9 +14,16 @@ export const authenticationError = createAction("AUTHENTICATION_ERROR", resolve 
   return (errorMsg: string) => resolve(errorMsg);
 });
 
-export const setExpiredSession = createAction("AUTHENTICATION_EXPIRED");
+export const setSessionExpired = createAction("SET_AUTHENTICATION_SESSION_EXPIRED");
+export const unsetSessionExpired = createAction("UNSET_AUTHENTICATION_SESSION_EXPIRED");
 
-const allActions = [setAuthenticated, authenticating, authenticationError, setExpiredSession];
+const allActions = [
+  setAuthenticated,
+  authenticating,
+  authenticationError,
+  setSessionExpired,
+  unsetSessionExpired,
+];
 
 export type AuthAction = ActionType<typeof allActions[number]>;
 
@@ -30,6 +37,9 @@ export function authenticate(
       await Auth.validateToken(token);
       Auth.setAuthToken(token, oidc);
       dispatch(setAuthenticated(true, oidc));
+      if (oidc) {
+        dispatch(unsetSessionExpired());
+      }
     } catch (e) {
       dispatch(authenticationError(e.toString()));
     }
@@ -39,9 +49,16 @@ export function authenticate(
 export function logout(): ThunkAction<Promise<void>, IStoreState, null, AuthAction> {
   return async dispatch => {
     Auth.unsetAuthToken();
-    // Expire the session if we are using OIDC tokens
-    dispatch(setExpiredSession());
     dispatch(setAuthenticated(false, false));
+  };
+}
+
+export function expireSession(): ThunkAction<Promise<void>, IStoreState, null, AuthAction> {
+  return async dispatch => {
+    if (Auth.usingOIDCToken()) {
+      dispatch(setSessionExpired());
+    }
+    dispatch(logout());
   };
 }
 
--- a/dashboard/src/actions/auth.ts
+++ b/dashboard/src/actions/auth.ts
@@ -34,7 +34,9 @@ export function authenticate(
   };
 }
 
-export function logout(): ThunkAction<Promise<void>, IStoreState, null, AuthAction> {
+export function logout(
+  needRefresh?: boolean,
+): ThunkAction<Promise<void>, IStoreState, null, AuthAction> {
   return async dispatch => {
     Auth.unsetAuthToken();
     dispatch(setAuthenticated(false, false));
--- a/dashboard/src/actions/charts.ts
+++ b/dashboard/src/actions/charts.ts
@@ -56,6 +56,7 @@ export type ChartsAction = ActionType<typeof allActions[number]>;
 async function httpGet(dispatch: Dispatch, targetURL: string): Promise<any> {
   try {
     const response = await axios.get(targetURL);
+    // If non-2XX response (not ok)
     if (response.status < 200 || response.status > 299) {
       const error = response.data || response.statusText;
       if (response.status === 404) {
--- a/dashboard/src/actions/charts.ts
+++ b/dashboard/src/actions/charts.ts
@@ -56,7 +56,7 @@ export type ChartsAction = ActionType<typeof allActions[number]>;
 async function httpGet(dispatch: Dispatch, targetURL: string): Promise<any> {
   try {
     const response = await axios.get(targetURL);
-    if (response.status !== 200) {
+    if (response.status < 200 || response.status > 299) {
       const error = response.data || response.statusText;
       if (response.status === 404) {
         dispatch(errorChart(new NotFoundError(error)));
--- a/dashboard/src/actions/charts.ts
+++ b/dashboard/src/actions/charts.ts
@@ -55,8 +55,7 @@ export type ChartsAction = ActionType<typeof allActions[number]>;
 
 async function httpGet(dispatch: Dispatch, targetURL: string): Promise<any> {
   try {
-    const response: any = await axios.get(targetURL);
-    // const json = await response.json();
+    const response = await axios.get(targetURL);
     if (response.status !== 200) {
       const error = response.data || response.statusText;
       if (response.status === 404) {
--- a/dashboard/src/actions/charts.ts
+++ b/dashboard/src/actions/charts.ts
@@ -2,7 +2,7 @@ import { Dispatch } from "redux";
 import { ThunkAction } from "redux-thunk";
 import { ActionType, createAction } from "typesafe-actions";
 
-import { axios } from "../shared/Auth";
+import { axios } from "../shared/AxiosInstance";
 import Chart from "../shared/Chart";
 import { IChart, IChartVersion, IStoreState, NotFoundError } from "../shared/types";
 import * as url from "../shared/url";
--- a/dashboard/src/actions/charts.ts
+++ b/dashboard/src/actions/charts.ts
@@ -2,6 +2,7 @@ import { Dispatch } from "redux";
 import { ThunkAction } from "redux-thunk";
 import { ActionType, createAction } from "typesafe-actions";
 
+import { axios } from "../shared/Auth";
 import Chart from "../shared/Chart";
 import { IChart, IChartVersion, IStoreState, NotFoundError } from "../shared/types";
 import * as url from "../shared/url";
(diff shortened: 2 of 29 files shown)


dashboard/src/components/PrivateRoute/PrivateRoute.tsx
```diff
@@ -1,26 +1,70 @@
...
+            },
+          }}
+          isOpen={true}
+        >
+          <div>
+            <div className="margin-b-normal">
+              Your session has expired, please reload to refresh your credentials
+            </div>
+            <div className="flex text-c">
+              <button className="button" onClick={this.reload}>
```
andresmgot: something interesting I found is that with this modal, when clicking in reload, the actual page gets reloaded while if I just do:

```js
    if (!authenticated && this.state.previouslyAuthenticated && Auth.usingOIDCToken()) {
      this.reload()
    }
```

it gets redirected to the root path. I am not sure the reason of this behavior. Any ideas?

prydonius: We render multiple PrivateRoutes per private route (https://github.com/kubeapps/kubeapps/blob/f4ee15254e0d96fd4ab747a01ad783618da81edb/dashboard/src/containers/RoutesContainer/Routes.tsx#L25) so we might be rendering this modal multiple times. Can you check the DOM to see if that's the case? If so, we need to find another way to do this.

dashboard/src/shared/AxiosInstance.ts
```diff
@@ -25,15 +25,25 @@ export function createAxiosInterceptors(axios: AxiosInstance, store: Store<IStor
   axios.interceptors.response.use(
     response => response,
     e => {
+      const dispatch = store.dispatch as ThunkDispatch<IStoreState, null, Action>;
       const err: AxiosError = e;
+      if (
+        err.config.xsrfCookieName === "XSRF-TOKEN" &&
```
prydonius: this doesn't seem relevant to catching this error to me, I'm guessing axios just sets this as a default

dashboard/src/shared/AxiosInstance.ts
```diff
@@ -25,15 +25,25 @@ export function createAxiosInterceptors(axios: AxiosInstance, store: Store<IStor
...
+        err.code === undefined &&
```
prydonius: this is interesting, so the HTTP status code is `undefined` in this case? Other errors would have an actual status code, right?

What happens if you don't have a network connection, do you also hit the same error?

dashboard/src/components/PrivateRoute/PrivateRoute.tsx
```diff
@@ -1,26 +1,70 @@
...
 
 type IRouteComponentPropsAndRouteProps = RouteProps & RouteComponentProps<any>;
 
 interface IPrivateRouteProps extends IRouteComponentPropsAndRouteProps {
   authenticated: boolean;
 }
 
-class PrivateRoute extends React.Component<IPrivateRouteProps> {
+interface IPrivateRouteState {
+  previouslyAuthenticated: boolean;
```
prydonius: why can't we put this in the redux state?

dashboard/src/shared/AxiosInstance.ts
```diff
@@ -25,15 +25,25 @@ export function createAxiosInterceptors(axios: AxiosInstance, store: Store<IStor
...
+        err.message === "Network Error" &&
+        !err.response &&
+        Auth.usingOIDCToken()
+      ) {
+        // The OIDC token is no longer valid, logout
+        dispatch(actions.auth.logout());
```
prydonius: why don't we dispatch a new action called `expireSession` or something? That way we can have a specific field in the state (`sessionExpired`) that deals with this case? Then, PrivateRoute is simpler and just needs to render the modal if `sessionExpired` is true.

dashboard/src/actions/charts.ts
```diff
@@ -2,6 +2,7 @@ import { Dispatch } from "redux";
 import { ThunkAction } from "redux-thunk";
 import { ActionType, createAction } from "typesafe-actions";
 
+import { axios } from "../shared/Auth";
```
prydonius: One concern about this is that we'll be sending the Kubernetes API token to services that don't use it. We should try to avoid that if possible, to minimise the exposure of those sensitive tokens. It will also be an issue if we need a separate header for these APIs in the future.

How about we setup a different axios instance that shares the interceptor for the XHR error, but doesn't send the token?

andresmgot:
@prydonius I have applied your suggestions and added some more tests PTAL

dashboard/src/shared/AxiosInstance.ts
prydonius: Thanks for checking those cases. What happens if you load the app, and then disconnect WiFi/network connection? I'd be worried that response would be undefined in this case too.

dashboard/src/components/PrivateRoute/PrivateRoute.tsx
prydonius: Nevermind, I was wrong, the `Switch` ensures we only load one https://github.com/kubeapps/kubeapps/blob/f4ee15254e0d96fd4ab747a01ad783618da81edb/dashboard/src/containers/RoutesContainer/Routes.tsx#L55.

dashboard/src/shared/AxiosInstance.ts
prydonius: What I was thinking was having two actions, to separate the concerns of the two cases:

- actions.auth.logout - untouched, does what it does today to clear the token and set authenticated to false
- actions.auth.expireSession - sets sessionExpired to true to bring up the modal and dispatches actions.auth.logout

If we separate them in this way, we shouldn't have the issue where we show the modal if a user clicks logout, right?

dashboard/src/actions/auth.ts
```diff
@@ -37,6 +39,8 @@ export function authenticate(
 export function logout(): ThunkAction<Promise<void>, IStoreState, null, AuthAction> {
   return async dispatch => {
     Auth.unsetAuthToken();
+    // Expire the session if we are using OIDC tokens
+    dispatch(setExpiredSession());
```
prydonius: See my comment about separating these two concerns into different actions

dashboard/src/actions/charts.ts
```diff
@@ -54,17 +55,16 @@ export type ChartsAction = ActionType<typeof allActions[number]>;
 
 async function httpGet(dispatch: Dispatch, targetURL: string): Promise<any> {
   try {
-    const response = await fetch(targetURL);
-    const json = await response.json();
-    if (!response.ok) {
```
prydonius: `response.ok` is true for any status code within the range 200-299. Check if Axios has a similar method, or check for that range, as we might get a successful response that isn't 200

dashboard/src/components/PrivateRoute/PrivateRoute.tsx
```diff
@@ -14,13 +16,44 @@ class PrivateRoute extends React.Component<IPrivateRouteProps> {
...
+              right: "auto",
+              top: "50%",
+              transform: "translate(-50%, -50%)",
+            },
+          }}
+          isOpen={true}
+        >
+          <div>
+            <div className="margin-b-normal">
+              Your session has expired, please reload to refresh your credentials
```
prydonius:
```suggestion
              Your session has expired, please reload to refresh your credentials.
```

dashboard/src/reducers/auth.ts
```diff
@@ -24,17 +26,22 @@ const authReducer = (state: IAuthState = initialState, action: AuthAction): IAut
...
-        authenticated: false,
-        authenticating: false,
-        oidcAuthenticated: false,
         authenticationError: action.payload,
       };
+    case getType(actions.auth.setExpiredSession):
+      return {
+        ...state,
+        // The session can only expire if we are using OIDC tokens
+        sessionExpired: state.oidcAuthenticated ? true : state.sessionExpired,
```
prydonius: We should move this check to the action, the reducer should be simpler and just set the field to true always if the action has been dispatched.

dashboard/src/index.tsx
```diff
@@ -1,15 +1,16 @@
 import * as React from "react";
 import * as ReactDOM from "react-dom";
 import * as Modal from "react-modal";
-import { createAxiosInterceptors } from "shared/AxiosInstance";
-import { axios } from "./shared/Auth";
+import * as Axios from "shared/AxiosInstance";
```
prydonius: why are we importing everything?

dashboard/src/shared/AxiosInstance.ts
```diff
@@ -51,3 +62,18 @@ export function createAxiosInterceptors(axios: AxiosInstance, store: Store<IStor
     },
   );
 }
+// createAxiosInterceptors will configure a set of interceptors to a provided axios instance,
+// relying also on an external redux store for action dispatching
+export function createAxiosInterceptors(axiosInstance: AxiosInstance, store: Store<IStoreState>) {
```
prydonius: IMO this and `createAxiosInterceptorsWithAuth` are more difficult to read than simply exporting and calling `addErrorHandling` and `addAuthHeaders` from index https://github.com/kubeapps/kubeapps/pull/995/files#diff-ebf3c79d519804e9e573437d0db1b265R12

dashboard/src/index.tsx
```diff
@@ -1,15 +1,16 @@
...
+import * as Auth from "./shared/Auth";
 
 import Root from "./containers/Root";
 import "./index.css";
 import store from "./store";
 // import registerServiceWorker from "./registerServiceWorker";
 
-createAxiosInterceptors(axios, store);
+Axios.createAxiosInterceptors(Axios.axios, store);
```
prydonius: we should add a comment explaining what we're doing here. It's also quite hard to read since `Axios.axios` and `Auth.axios` look very similar. Can we distinguish them better somehow?

dashboard/src/shared/AxiosInstance.ts
andresmgot: I have tried that but I am using a local instance of minikube so I still have access to the services. In any case yes, it should be like the case in which I remove the ingress backend. If the response is undefined, we ask the user to reload and then the  "Connection refused" error (or similar) is shown.

dashboard/src/shared/AxiosInstance.ts
andresmgot: Yes, as far as we call `expireSession` when we receive a 401

dashboard/src/actions/charts.ts
andresmgot: it doesn't have a similar method but the range should work

dashboard/src/index.tsx
andresmgot: because there is an `axios` element in both `AxiosInstance` and `Auth` I cannot do:

```
import {
  axios,
  createAxiosInterceptors,
  createAxiosInterceptorsWithAuth,
} from "shared/AxiosInstance";
import {axios} from "./shared/Auth";
```

because `axios` is duplicated

dashboard/src/index.tsx
andresmgot: actually, I am not sure why are we doing this here? It works if I add the interceptors in `AxiosInstance`

dashboard/src/index.tsx
andresmgot: In any case, I have moved the axios intance from the Auth library to the AxiosInstance. Hopefully that's a bit more readable.

dashboard/src/shared/AxiosInstance.ts
prydonius: @andresmgot we used to show an error message on the login page when we received a 401 right? Do we still need that if we have a modal now?

dashboard/src/actions/charts.ts
prydonius: not as clean, but oh well ¯\_(ツ)_/¯

dashboard/src/index.tsx
prydonius: (haven't yet looked at the updated code yet) for future reference you can do `import { axios as authAxios } from "./shared/Auth"` (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import)

dashboard/src/actions/auth.ts
```diff
@@ -14,7 +14,16 @@ export const authenticationError = createAction("AUTHENTICATION_ERROR", resolve
   return (errorMsg: string) => resolve(errorMsg);
 });
 
-const allActions = [setAuthenticated, authenticating, authenticationError];
+export const setSessionExpired = createAction("SET_AUTHENTICATION_SESSION_EXPIRED");
+export const unsetSessionExpired = createAction("UNSET_AUTHENTICATION_SESSION_EXPIRED");
```
prydonius: why not combine this with `setSessionExpired` and give it a parameter of what to set? Similar to what we're doing for `setAuthenticated`.

dashboard/src/actions/charts.ts
```diff
@@ -54,17 +55,16 @@ export type ChartsAction = ActionType<typeof allActions[number]>;
...
-      const error = json.data || response.statusText;
+    const response = await axios.get(targetURL);
+    if (response.status < 200 || response.status > 299) {
```
prydonius: bit obvious but might be easier to read with a comment
```suggestion
    # if non-2XX response
    if (response.status < 200 || response.status > 299) {
```

dashboard/src/shared/AxiosInstance.ts
prydonius: > If the response is undefined, we ask the user to reload and then the "Connection refused" error (or similar) is shown.

If this is the case, it'll be confusing if it says my session has expired and it's just because I am no longer connected to the internet. Can you test this on a remote cluster? We might want to tweak the message we show based on this.

dashboard/src/shared/AxiosInstance.ts
```diff
@@ -51,3 +63,8 @@ export function createAxiosInterceptors(axios: AxiosInstance, store: Store<IStor
     },
   );
 }
+
+// Error handling is added with an interceptor in index.tsx
+export const axios = Axios.create();
```
prydonius: I forgot why we don't just add it here? cc @migmartri

dashboard/src/shared/ClusterServiceClass.ts
```diff
@@ -1,4 +1,4 @@
-import { axios } from "./Auth";
+import { axiosWithAuth } from "./AxiosInstance";
```
prydonius: couple of options here to avoid having to change this everywhere:

1. Keep axiosWithAuth in ./Auth and import it as `{ axios as axiosWithAuth }` in the index
2. Import this as `{ axiosWithAuth as axios }` in all these imports

That said, I like that `axiosWithAuth` is fairly descriptive, so I honestly don't have a strong opinion here.

dashboard/src/shared/AxiosInstance.ts
andresmgot: Yes, that message (and the form) is showed when not using an OIDC token

dashboard/src/shared/AxiosInstance.ts
andresmgot: I asked myself the same and I tried to add here the interceptors. What I found is that in that case we cannot mock the `store` (second parameter) in the tests so this was not testable.

dashboard/src/shared/ClusterServiceClass.ts
andresmgot: I didn't know about that notation but that may obscure a bit the origin of the `axios` instance that it's using and can be potentially confused for the one without the auth header. I'd leave this as it is to make clear the difference.

dashboard/src/shared/AxiosInstance.ts
andresmgot: >Can you test this on a remote cluster?

Yes, I tested it and without connection you see the modal. I didn't think this was very confusing since we are just asking to reload the page and then the user gets a more detailed error but it's true that it may not be accurate. I will be more explicit in the message.

dashboard/src/shared/AxiosInstance.ts
prydonius: @andresmgot ah yes, you're right - thanks

dashboard/src/shared/AxiosInstance.ts
prydonius: aha, thanks!

dashboard/src/shared/ClusterServiceClass.ts
prydonius: Agreed, sounds good to me.

dashboard/src/components/PrivateRoute/PrivateRoute.tsx
```diff
@@ -37,7 +37,8 @@ class PrivateRoute extends React.Component<IPrivateRouteProps> {
         >
           <div>
             <div className="margin-b-normal">
-              Your session has expired, please reload to refresh your credentials.
+              Your session has expired or there is a different network issue, please reload the
```
prydonius: How about "Your session has expired or the connection has been lost, please reload the page."?

PR #998, opened 2019-03-21
Add builds for GKE 1.12

### LINKED ISSUE #886: Add GKE k8s 1.12 to e2e test environments
Once GKE supports 1.12, we should add it as an e2e test environment.

andresmgot:
Fixes #886 

Example build: https://circleci.com/workflow-run/d8a54f97-f424-4d56-88d1-a1e36607bedb

I have also added the label `team=kubeapps` for accounting purposes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -242,7 +242,7 @@ jobs:
     environment:
       GKE_BRANCH: "1.11"
       TEST_LATEST_RELEASE: 1
-  GKE_1_12MASTER:
+  GKE_1_12_MASTER:
     <<: *gke_test
     environment:
       GKE_BRANCH: "1.12"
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -49,16 +49,34 @@ workflows:
...
       - push_images:
           <<: *build_always
           requires:
             - GKE_1_11_MASTER
             - GKE_1_11_LATEST_RELEASE
+            - GKE_1_12_MASTER
+            - GKE_1_12_LATEST_RELEASE
       - release:
           <<: *build_on_tag
           requires:
```
prydonius: this should be dependent on it too?

andresmgot: good catch

PR #1013, opened 2019-04-24
Enable end-to-end tests for pull requests

andresmgot:
This PR adds a new job that run our end-to-end tests (helm tests) that's run in a local Kubernetes cluster using [kind](https://github.com/kubernetes-sigs/kind). This enable us to use our end-to-end tests with untrusted code (PRs).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,65 +25,67 @@ workflows:
   version: 2
   kubeapps:
     jobs:
-      - test_go:
+      - test_e2e:
           <<: *build_always
-      - test_dashboard:
-          <<: *build_always
-      - test_chart_render:
-          <<: *build_always
-      - build_go_images:
-          <<: *build_always
-      - build_dashboard:
-          <<: *build_always
-      - GKE_1_11_MASTER:
-          <<: *build_on_master
-          requires:
-            - test_go
-            - test_dashboard
-            - build_go_images
-            - build_dashboard
-      - GKE_1_11_LATEST_RELEASE:
-          <<: *build_on_master
-          requires:
-            - test_go
-            - test_dashboard
-            - build_go_images
-            - build_dashboard
-      - GKE_1_12_MASTER:
-          <<: *build_on_master
-          requires:
-            - test_go
-            - test_dashboard
-            - build_go_images
-            - build_dashboard
-      - GKE_1_12_LATEST_RELEASE:
-          <<: *build_on_master
-          requires:
-            - test_go
-            - test_dashboard
-            - build_go_images
-            - build_dashboard
-      - sync_chart:
-          <<: *build_on_master
-          requires:
-            - GKE_1_11_MASTER
-            - GKE_1_11_LATEST_RELEASE
-            - GKE_1_12_MASTER
-            - GKE_1_12_LATEST_RELEASE
-      - push_images:
-          <<: *build_always
-          requires:
-            - GKE_1_11_MASTER
-            - GKE_1_11_LATEST_RELEASE
-            - GKE_1_12_MASTER
-            - GKE_1_12_LATEST_RELEASE
-      - release:
-          <<: *build_on_tag
-          requires:
-            - GKE_1_11_MASTER
-            - GKE_1_11_LATEST_RELEASE
-            - GKE_1_12_MASTER
-            - GKE_1_12_LATEST_RELEASE
+      # - test_go:
+      #     <<: *build_always
+      # - test_dashboard:
+      #     <<: *build_always
+      # - test_chart_render:
+      #     <<: *build_always
+      # - build_go_images:
+      #     <<: *build_always
+      # - build_dashboard:
+      #     <<: *build_always
+      # - GKE_1_11_MASTER:
+      #     <<: *build_on_master
+      #     requires:
+      #       - test_go
+      #       - test_dashboard
+      #       - build_go_images
+      #       - build_dashboard
+      # - GKE_1_11_LATEST_RELEASE:
+      #     <<: *build_on_master
+      #     requires:
+      #       - test_go
+      #       - test_dashboard
+      #       - build_go_images
+      #       - build_dashboard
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -117,17 +124,37 @@ build_images: &build_images
...
     - <<: *exports
     - run: |
+        mkdir -p images/
         read -ra IMG_ARRAY <<< "$IMAGES"
         if [[ -n "${CIRCLE_TAG}" ]]; then
           makeArgs="VERSION=${CIRCLE_TAG}"
         fi
         for IMAGE in "${IMG_ARRAY[@]}"; do
-          make IMG_MODIFIER="$IMG_MODIFIER" IMAGE_TAG="${DEV_TAG}" $makeArgs ${IMAGE}
+          make IMG_MODIFIER="$IMG_MODIFIER" IMAGE_TAG="${DEV_TAG}" $makeArgs kubeapps/${IMAGE}
```
andresmgot: Note: I have removed `kubeapps/` from `${IMAGE}` to be able store `${IMAGE}` as a file not `folder/file`. Let me know if that is readable or you prefer to keep using the previous definition of `${IMAGE}`

script/e2e-test.sh
```diff
@@ -31,7 +31,10 @@ echo "IMAGE_REPO_SUFFIX: $IMG_MODIFIER"
 kubectl version
 
 # Install Tiller with TLS support
+kubectl -n kube-system create sa tiller
```
andresmgot: we need to set the tiller service account as cluster-admin because by default, in a cluster created with `kind` the `default` service account in the `kube-system` namespace doesn't have permissions to read config maps in that namespace.

.circleci/config.yml
prydonius: The only thing that might be an issue is if we want to push these images somewhere other than Docker Hub in the future, or need to use different repos for different images, but I don't think that's the case right now.

.circleci/config.yml
```diff
@@ -117,17 +124,37 @@ build_images: &build_images
...
           if [[ -z "${CIRCLE_PULL_REQUEST}" && -n "${DOCKER_USERNAME}" && -n "${DOCKER_PASSWORD}" ]]; then
             docker login -u="${DOCKER_USERNAME}" -p="${DOCKER_PASSWORD}"
-            docker push ${IMAGE}${IMG_MODIFIER}:${DEV_TAG}
+            docker push kubeapps/${IMAGE}${IMG_MODIFIER}:${DEV_TAG}
           fi
+          docker save kubeapps/${IMAGE}${IMG_MODIFIER}:${DEV_TAG} > images/${IMAGE}${IMG_MODIFIER}:${DEV_TAG}
```
prydonius: We only need to do this if this is a pull request, right?

.circleci/config.yml
```diff
@@ -217,20 +232,44 @@ jobs:
...
+      IMAGES: "dashboard"
     <<: *build_images
   release:
     docker:
       - image: circleci/golang:1.11
     steps:
       - checkout
       - run: REPO_DOMAIN=kubeapps REPO_NAME=kubeapps ./script/create_release.sh ${CIRCLE_TAG}
+  local_e2e_tests:
+    working_directory: /home/circleci/.go_workspace/src/github.com/kubeapps/kubeapps
```
prydonius: why is this necessary?

.circleci/config.yml
```diff
@@ -217,20 +232,44 @@ jobs:
...
+    steps:
+      - checkout
+      - <<: *exports
+      # By default Go 1.9 is installed, we need 1.11 to use kind
+      - run: |
+          sudo rm -rf /usr/local/go/*
+          sudo chown $(whoami) /usr/local/go
+          curl -LO https://dl.google.com/go/go1.11.9.linux-amd64.tar.gz
+          tar -C /usr/local -xzf go1.11.9.linux-amd64.tar.gz
+      - run: go get -u sigs.k8s.io/kind
```
prydonius: Can we use a pre-built binary from a release instead? https://github.com/kubernetes-sigs/kind/releases. This should avoid the need to install Go 1.11?

script/e2e-test.sh
prydonius: Weird, I thought this was the case in GKE too - not sure how we were able to avoid it before?

.circleci/config.yml
andresmgot: Yes, we can change this later if needed.

.circleci/config.yml
andresmgot: It's mandatory when it is a PR. When running this in master we could pull the images but this is convenient because loading the images from the local file is faster.

.circleci/config.yml
andresmgot: it's probably not, we are not compiling in this job

.circleci/config.yml
andresmgot: the only release available is a prerelease. I am not sure if it will work but I can try, probably it's enough.

script/e2e-test.sh
andresmgot: tbh I am not sure. My supposition is that depending on the user that creates the cluster, GKE default permissions fallback to those, so if an user with privileges creates a cluster (like the account set here) it's not necessary to create the extra service account. I have tried with my personal account and for me it requires to use the tiller serviceaccount.

prydonius: Hmm that's pretty strange, so the service account here has admin privileges in the project?

.circleci/config.yml
prydonius: Ah I was thinking we only do the local test for PRs, but we actually run it all the time. sg.

prydonius:
Nice, looks like the prebuilt version works. I believe they're marking all 0.x releases as pre-releases in GitHub.

script/e2e-test.sh
andresmgot: yes, it has more privileges than a normal account

Issue #1025, opened 2019-05-08
Unable to install prometheus-operator


TaChao:
kubeapp version: 1.3.0
Kubernetes version: v1.12.5
helm version : v2.11.0
chart : offical/prometheus-operator5.5.1

error message: the server could not find the requested resource
recurrent : 100%

when useing helm cli chart installed fine

**useing kubeapp tiller log:**
```
2019/05/08 08:55:29 info: manifest "prometheus-operator/charts/grafana/templates/dashboards-json-configmap.yaml" is empty. Skipping.
2019/05/08 08:55:29 info: manifest "prometheus-operator/charts/prometheus-node-exporter/templates/endpoints.yaml" is empty. Skipping.
2019/05/08 08:55:29 info: manifest "prometheus-operator/templates/exporters/kube-dns/servicemonitor.yaml" is empty. Skipping.
2019/05/08 08:55:29 info: manifest "prometheus-operator/templates/exporters/kube-controller-manager/endpoints.yaml" is empty. Skipping.
2019/05/08 08:55:29 info: manifest "prometheus-operator/charts/kube-state-metrics/templates/monitor.yaml" is empty. Skipping.
2019/05/08 08:55:29 info: manifest "prometheus-operator/templates/exporters/kube-scheduler/endpoints.yaml" is empty. Skipping.
2019/05/08 08:55:29 info: manifest "prometheus-operator/templates/exporters/kube-etcd/endpoints.yaml" is empty. Skipping.
2019/05/08 08:55:29 info: manifest "prometheus-operator/templates/prometheus/podDisruptionBudget.yaml" is empty. Skipping.
2019/05/08 08:55:29 info: manifest "prometheus-operator/templates/prometheus-operator/cleanup-crds.yaml" is empty. Skipping.
2019/05/08 08:55:29 info: manifest "prometheus-operator/templates/alertmanager/podDisruptionBudget.yaml" is empty. Skipping.
2019/05/08 08:55:29 info: manifest "prometheus-operator/charts/prometheus-node-exporter/templates/monitor.yaml" is empty. Skipping.
2019/05/08 08:55:29 info: manifest "prometheus-operator/templates/prometheus/additionalAlertRelabelConfigs.yaml" is empty. Skipping.
2019/05/08 08:55:29 info: manifest "prometheus-operator/templates/grafana/configmap-dashboards.yaml" is empty. Skipping.
[...]

PR #1027, opened 2019-05-09
Bypass permission check for elements that are not found in the API

andresmgot:
Ref #1025

For charts that contains both CRDs and CRs `tiller-proxy` is not able to check if the user has permissions to create CRs because the CRD doesn't exist yet.

### DIFF
--- a/pkg/auth/auth.go
+++ b/pkg/auth/auth.go
@@ -20,14 +20,14 @@ import (
 	"fmt"
 	"strings"
 
+	yamlUtils "github.com/kubeapps/kubeapps/pkg/yaml"
 	authorizationapi "k8s.io/api/authorization/v1"
+	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	discovery "k8s.io/client-go/discovery"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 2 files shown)


pkg/auth/auth.go
```diff
@@ -170,6 +170,11 @@ func (u *UserAuth) isAllowed(verb string, itemsToCheck []resource) ([]Action, er
 	for _, i := range itemsToCheck {
 		rInfo, err := u.resolve(i.APIVersion, i.Kind)
 		if err != nil {
+			if k8sErrors.IsNotFound(err) {
+				// The resource version/kind is not registered in the k8s API so
+				// we assume it's a CRD that is going to be created with the chart
```
prydonius: There is a case here where a chart tries to install a resource that doesn't exist, I would point out in this comment that it's okay to ignore it here as the actual chart installation will fail in this case

pkg/auth/auth_test.go
```diff
@@ -180,6 +189,16 @@ kind: ClusterRoleBinding
 				{APIVersion: "rbac.authorization.k8s.io/v1", Resource: "clusterrolebindings", ClusterWide: true, Verbs: []string{"get"}},
 			},
 		},
+		// It should allow an unrecognized resource
```
prydonius:
```suggestion
		// It should allow an unrecognized resource, so that CRDs can be installed before CRs
```

Issue #970, opened 2019-02-21
Make apprepo controller spawned CronJobs configurable - nodeselector/affinity/environment 


davidkarlsen:
See thread in #966 for background.

Make the CronJobs spawned by the apprepo more configurable:
* placement of pod on node with nodeselector and/or affinity - the reason being that only a subset of nodes can access the repo
* be able to set environment vars - typically for defining proxy/no_proxy vars - in order to be able to download external resources like the icon in the Chart.yaml

PR #1028, opened 2019-05-10
[apprepository-controller] Allow to modify the repo sync job

andresmgot:
Fixes: #970

This PR allows user to specify a `podTemplateSpec` within the apprepository to modify the default behavior. See the docs added for an example.


cmd/apprepository-controller/controller.go
```diff
@@ -443,29 +443,34 @@ func syncJobSpec(apprepo *apprepov1alpha1.AppRepository) batchv1.JobSpec {
...
+	podTemplateSpec.Spec.RestartPolicy = "OnFailure"
+	// Populate container spec
+	if len(podTemplateSpec.Spec.Containers) == 0 {
+		podTemplateSpec.Spec.Containers = []corev1.Container{{}}
+	}
+	podTemplateSpec.Spec.Containers[0].Name = "sync"
+	podTemplateSpec.Spec.Containers[0].Image = repoSyncImage
+	podTemplateSpec.Spec.Containers[0].Command = []string{"/chart-repo"}
+	podTemplateSpec.Spec.Containers[0].Args = apprepoSyncJobArgs(apprepo)
+	podTemplateSpec.Spec.Containers[0].Env = append(podTemplateSpec.Spec.Containers[0].Env, apprepoSyncJobEnvVars(apprepo)...)
```
prydonius: what happens if an env var is repeated?

cmd/apprepository-controller/controller_test.go
```diff
@@ -548,6 +548,97 @@ func Test_newSyncJob(t *testing.T) {
...
+					Type: "helm",
+					URL:  "https://charts.acme.com/my-charts",
+					SyncJobPodTemplate: corev1.PodTemplateSpec{
+						ObjectMeta: metav1.ObjectMeta{
+							Labels: map[string]string{
+								"foo": "bar",
+							},
+						},
+						Spec: corev1.PodSpec{
+							Containers: []corev1.Container{
```
prydonius: we should add volumes and nodeSelector/affinity to this test too

cmd/apprepository-controller/controller_test.go
```diff
@@ -192,8 +192,8 @@ func Test_newCronJob(t *testing.T) {
 				defer func() { userAgentComment = "" }()
 			}
 			result := newCronJob(tt.apprepo)
-			if !reflect.DeepEqual(tt.expected, *result) {
-				t.Errorf("Unexpected result\nExpecting:\n %+v\nReceived:\n %+v", tt.expected, *result)
+			if diff := deep.Equal(tt.expected, *result); diff != nil {
```
prydonius: what's this change for?

docs/user/private-app-repository.md
```diff
@@ -103,3 +103,28 @@ The above command creates a token with read-only permissions. Now you can go to
 <img src="../img/jfrog-custom-repo.png" alt="JFrog custom repository" width="400px">
 
 After submitting the repository, you will be able to click on the new repository and see the chart you uploaded in the previous step.
+
+## Modifying the synchronization job
+
+Kubeapps runs a periodic job (CronJob) to populate and synchronize the charts existing in each repository. Since Kubeapps v1.4.0, it's possible to modify the spec of this job. This is useful if you need to run the pod in a certain Kubernetes node, or set some environment variables. To do so you can edit (or create) an AppRepository adding the field `syncJobPodTemplate`. For example:
```
prydonius:
```suggestion
Kubeapps runs a periodic job (CronJob) to populate and synchronize the charts existing in each repository. Since Kubeapps v1.4.0, it's possible to modify the spec of this job. This is useful if you need to run the Pod in a certain Kubernetes node, or set some environment variables. To do so you can edit (or create) an AppRepository and specify the `syncJobPodTemplate` field. For example:
```

docs/user/private-app-repository.md
```diff
@@ -103,3 +103,28 @@ The above command creates a token with read-only permissions. Now you can go to
...
+        my-repo: "isPrivate"
+    spec:
+      containers:
+        - env:
+            - name: FOO
+              value: BAR
+  url: https://my.charts.com/
+```
+
+The above will generate a pod with the label `my-repo: isPrivate` and the environment variable `FOO=BAR`.
```
prydonius:
```suggestion
The above will generate a Pod with the label `my-repo: isPrivate` and the environment variable `FOO=BAR`.
```

cmd/apprepository-controller/controller.go
andresmgot: if an env var gets repeated, the last one is the one that gets applied. This means that the env vars set by kubeapps will take precedence.

cmd/apprepository-controller/controller_test.go
andresmgot: it's for readability. It failed for me and it just printed both structs. It's' difficult to see the difference but with that `diff` it only prints what's different. For example:

```
    --- FAIL: Test_newSyncJob/my-charts_with_a_custom_pod_template (0.00s)
        /home/andres/projects/src/github.com/kubeapps/kubeapps/cmd/apprepository-controller/controller_test.go:657: [Spec.Template.Spec.Affinity: <nil pointer> != v1.Affinity]
```

prydonius: Can you try to use https://github.com/stretchr/testify since we're using that elsewhere? It also provides better error messages IMO.

andresmgot: I may be missing something but I tried `assert.Equal` from testify and it just dumps the structs as well:

```
        /home/andres/projects/src/github.com/kubeapps/kubeapps/cmd/apprepository-controller/controller_test.go:196: 
            	Error Trace:	controller_test.go:196
            	Error:      	Not equal: 
            	            	expected: v1beta1.CronJob{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ObjectMeta:v1.ObjectMeta{Name:"apprepo-sync-my-charts", GenerateName:"", Namespace:"kubeapps", SelfLink:"", UID:"", 
...
            	            	actual  : v1beta1.CronJob{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ObjectMeta:v1.ObjectMeta{Name:"apprepo-sync-my-charts", GenerateName:"", Namespace:"kubeapps", SelfLink:"", UID:"", 
```

prydonius: Hmm it looks like it's trying to show a diff but for some reason it's not able to? I've seen it show diffs like https://github.com/stretchr/testify/issues/383 in the past.

andresmgot: I don't really know, `go-test/deep` seems to do a better job. It doesn't have extra dependencies so I think it's a good addition.

prydonius: Fair enough, just thought it would be worth trying testify

PR #1032, opened 2019-05-16
Use relative path to fetch OIDC token

### LINKED ISSUE #1005: Use PathPrefix on Traefik
Hi,

I'm trying to deploy Kubeapps behind Traefik proxy.

I saw many solutions using a subdomain like kubeapps.example.com, but I'm trying to make it work as example.com/catalog.

For this, i'm using Traefik ingress with PathPrefix(Strip) solution

Here is my values.yml

```yaml
ingress:
  enabled: true
  annotations:
    kubernetes.io/ingress.class: traefik
    traefik.ingress.kubernetes.io/rule-type: PathPrefix
    traefik.ingress.kubernetes.io/auth-type: "basic"
    traefik.ingress.kubernetes.io/auth-secret: "auth-http"
  hosts:
  - name: mykube.example.com
    path: /catalog
```

With the PathPrefixStrip way, It's all blank since js & css files doesn't use the prefix `/catalog`

But with the PathPrefix way, I can see the UI, I'm being redirected to `/login` but can't see any content. Apparently there's a HEAD request triggered on `/` that returns a 404. Because it doesn't take the prefix.

Capture d’écran 2019-04-11 à 08 52 51

Is there a way to fix it ?

Thanks

### LINKED ISSUE #1024: Should fall back to manual login form if OIDC token is invalid
Currently, if the OIDC token is not accepted by the API server (returns 401), we get a loading spinner forever:

Screen Shot 2019-05-06 at 15 31 05

We should catch this and fall back to the manual login form.

andresmgot:
Fixes #1005
Fixes #1024

Use a relative path to try to fetch the OIDC token. If the calls fails, catch the error to avoid that the application hangs.


andresmgot:
Merging to include this in a new release.

PR #1034, opened 2019-05-16
Catch non semver compatibility error

### LINKED ISSUE #1011: Kubeapps crashes if the chart version is not semver compatible
I'm experiencing this when deploying a single custom chart of ours, I can't find the source of it.

It happens when viewing any namespace where the chart is deployed. Here's a screenshot of the error:
image

Looking at the Javascript console, I get this error:

`kube.ts:58 WebSocket connection to 'wss://<DOMAIN>/api/kube/api/v1/namespaces/<NAMESPACE</services?watch=true&fieldSelector=metadata.name%3Djenkins-agent' failed: WebSocket is closed before the connection is established.`

I've tried using different version numbers, no luck. This is the Chart.yml for the chart we're deploying:

```
apiVersion: v1
name: jenkins-dynamic
version: 1.1
appVersion: 2.152
description: The chart deploys a Jenkins Master and an dynamic agent nodes.
icon: https://wiki.jenkins-ci.org/download/attachments/2916393/logo.png
```

andresmgot:
Fixes #1011

If a chart has a non semver version, catch the error.


andresmgot:
Merging this to include it in the release

andresmgot:
The problem is that we cannot compare. In theory, the list of versions received from the `chartsvc` may be disordered so we would know if the last one is the latest. It's true that it may be confusing to say that it's up to date when it cannot be confirmed though. I can send a follow up PR handling that case and we can just omit that information or even better put a warning message saying that the version cannot be verified.

With this PR at least we avoid the crash.

PR #1038, opened 2019-05-21
Show a warning message for non semver versions

### LINKED ISSUE #1036: Show a warning for non-semver versions
If an application has a non-semver version (e.g. 1.0) we are not able to differ if the application is up to date but we are showing that in the UI. We should show a warning telling that it's not possible to evaluate the version.

Follow up of #1034

andresmgot:
Fixes #1036

Screenshot from 2019-05-21 10-31-33

The auto generated name of the app is well... auto generated :joy:

### DIFF
--- a/dashboard/src/components/AppView/ChartInfo.tsx
+++ b/dashboard/src/components/AppView/ChartInfo.tsx
@@ -61,7 +61,7 @@ class ChartInfo extends React.Component<IChartInfoProps> {
               size={15}
               style={{ bottom: "-0.2em" }}
             />{" "}
-            <span>Updates check failed. {app.updateInfo.error.message}</span>
+            <span>Update check failed. {app.updateInfo.error.message}</span>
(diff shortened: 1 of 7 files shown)


dashboard/src/components/AppView/ChartInfo.tsx
```diff
@@ -51,6 +51,20 @@ class ChartInfo extends React.Component<IChartInfoProps> {
...
+        return (
+          <div>
+            <AlertTriangle
+              color="white"
+              fill="#FDBA12"
+              className="icon"
+              size={15}
+              style={{ bottom: "-0.2em" }}
+            />{" "}
+            <span>Updates check failed. {app.updateInfo.error.message}</span>
```
prydonius:
```suggestion
            <span>Update check failed. {app.updateInfo.error.message}</span>
```

Might make more sense grammatically?

PR #1041, opened 2019-05-23
Catch possible errors with incomplete Ingress objects

### LINKED ISSUE #1040: Something went wrong when config ingress.
360截图-5275314

kubeapps version 1.3.2 and 1.4.0

andresmgot:
Fixes #1040

Ensure that an Ingress object has `rules`, `paths` and `tls.hosts` defined before processing them to avoid errors.

### DIFF
--- a/dashboard/src/components/AppView/AccessURLTable/AccessURLItem/AccessURLIngressHelper.test.ts
+++ b/dashboard/src/components/AppView/AccessURLTable/AccessURLItem/AccessURLIngressHelper.test.ts
@@ -4,9 +4,9 @@ import { GetURLItemFromIngress } from "./AccessURLIngressHelper";
 describe("GetURLItemFromIngress", () => {
   interface Itest {
     description: string;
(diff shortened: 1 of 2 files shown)


dashboard/src/components/AppView/AccessURLTable/AccessURLItem/AccessURLIngressHelper.test.ts
```diff
@@ -38,6 +38,27 @@ describe("GetURLItemFromIngress", () => {
       tlsHosts: ["foo.bar"],
       expectedURLs: ["https://foo.bar", "http://not-foo.bar"],
     },
+    {
+      description: "it should ignore an ingress if it has no hosts",
```
prydonius: Does this still show an entry for the Ingress in the AccessURLTable, though without any clickable links? I think we'd still want to show that an Ingress exists, but that the hosts are not correctly defined.

PR #1057, opened 2019-06-12
Use global security context for apprepositories-sync job

LP0101:
Currently, the apprepositories-sync job does not use the global securityContext defined in the values.yaml. If Kubeapps is being deployed on a cluster where this security context is required, apprepositories have to be configured individually, or else the sync job will fail.

This PR checks the value of securityContext, and will ensure the apprepository-sync job will run as the correct user.

### DIFF
--- a/chart/kubeapps/templates/apprepositories.yaml
+++ b/chart/kubeapps/templates/apprepositories.yaml
@@ -1,4 +1,3 @@
-{{- $root := . -}}
 {{- range .Values.apprepository.initialRepos }}
 apiVersion: kubeapps.com/v1alpha1
 kind: AppRepository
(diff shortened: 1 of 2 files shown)


chart/kubeapps/templates/apprepositories.yaml
```diff
@@ -25,6 +26,12 @@ metadata:
 spec:
   type: helm
   url: {{ .url }}
+  {{- if $root.Values.securityContext.enabled }}
```
prydonius: no need to define a `$root` var here, you can use `$` which points to root directly (e.g. see line 25 above)

LP0101: Neat, didn't know that worked in Go's templating engine. Fixed it

PR #1058, opened 2019-06-18
Fix memory usage for tiller-proxy

### LINKED ISSUE #1052: [tiller-proxy] Tiller-proxy pod gets killed because memory usage
Tiller-proxy reaches its memory limit when installing some apps (for example stable/linkerd). The pod gets killed:

```
  containerStatuses:
  - containerID: docker://abca9629ca890ec386bf613577bf3d5b88aa42a60d148666e5567f1a1cc790f4
    image: kubeapps/tiller-proxy:latest
    imageID: docker-pullable://kubeapps/tiller-proxy@sha256:0d2e45b0426ac7ba061c8ea6130a0087047479c4c13a078d0315df87c45bd025
    lastState:
      terminated:
        containerID: docker://a7cc49e584468031f68e55c2ad8757d9149872dedeb4ef3649cc780cadc9570c
        exitCode: 137
        finishedAt: "2019-06-07T18:40:03Z"
        reason: OOMKilled
        startedAt: "2019-06-07T18:38:08Z"
```

We should evaluate if  we should increase the maximum memory that the pod can use or if we are misusing memory in the service.

andresmgot:
Fix #1052 

Cache the result of unmarshaling the `index.yaml` file of a repository.

Increase the memory limit for Tiller-proxy.

### DIFF
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -18,7 +18,7 @@ package chart
 
 import (
 	"bytes"
-	"crypto/sha1"
+	"crypto/sha256"
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/json"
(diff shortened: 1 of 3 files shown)


pkg/chart/chart.go
```diff
@@ -145,14 +157,26 @@ func readResponseBody(res *http.Response) ([]byte, error) {
 	return body, nil
 }
 
-func parseIndex(data []byte) (*repo.IndexFile, error) {
-	index := &repo.IndexFile{}
-	err := yaml.Unmarshal(data, index)
-	if err != nil {
-		return index, err
+func checksum(data []byte) string {
+	hasher := sha1.New()
```
prydonius: any reason we want to use sha1 and not sha256?

pkg/chart/chart.go
```diff
@@ -145,14 +157,26 @@ func readResponseBody(res *http.Response) ([]byte, error) {
...
+	hasher.Write(data)
+	return string(hasher.Sum(nil))
+}
+
+func parseIndex(repoURL string, data []byte) (*repo.IndexFile, error) {
+	sha := checksum(data)
```
prydonius: I'm not sure the cache check should be part of `parseIndex`, it would make more sense if we checked the cache in a separate function before `parseIndex`.

andresmgot:
> Nice, caching the index is a great idea. Are we able to add any unit tests for this?

Added now that it has its own function

PR #1059, opened 2019-06-18
Notify app updates

### LINKED ISSUE #1039: display update of chart and update of application
New feature proposal:
Add a distinction between helm chart change and application change (and the degree of changement). I propose also that color of update ribbon notification give indication of changment level. According the classic rule 0.0.0:
First digit is major update, Second is minor update, Third is patch update 
the color could relect this level
An option could also propose to user to be only notified for application update

andresmgot:
Fixes #1039

Screenshot from 2019-07-01 17-33-18
Screenshot from 2019-07-02 12-10-00
Screenshot from 2019-07-02 12-09-54

 - If there is a new chart version and the app version changes we notify this as a new app version
 - I have removed the tooltip text box of the "Upgrade" button since it was not that helpful and it would be confusing to notify there different types of versions.

### DIFF
--- a/dashboard/src/components/AppView/AppControls/UpgradeButton.tsx
+++ b/dashboard/src/components/AppView/AppControls/UpgradeButton.tsx
@@ -3,14 +3,14 @@ import * as React from "react";
 import { ArrowUpCircle } from "react-feather";
 
 interface IUpgradeButtonProps {
-  updateVersion?: string;
+  newVersion?: boolean;
   releaseName: string;
   releaseNamespace: string;
   push: (location: string) => RouterAction;
 }
 
 const UpgradeButton: React.SFC<IUpgradeButtonProps> = props => {
-  const { updateVersion, push, releaseName, releaseNamespace } = props;
+  const { newVersion, push, releaseName, releaseNamespace } = props;
   const onClick = () => push(`/apps/ns/${releaseNamespace}/upgrade/${releaseName}`);
   let upgradeButton = (
     <button className="button" onClick={onClick}>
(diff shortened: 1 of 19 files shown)


dashboard/src/components/AppView/AppControls/UpgradeButton.tsx
```diff
@@ -3,30 +3,27 @@ import * as React from "react";
...
   // If the app is outdated highlight the upgrade button
-  if (updateVersion) {
+  if (newVersion) {
     upgradeButton = (
-      <div className="tooltip">
-        <button className="button upgrade-button" onClick={onClick}>
-          <span className="upgrade-text">Upgrade</span>
-          <ArrowUpCircle color="white" size={25} fill="#82C341" className="notification" />
-        </button>
-        <span className="tooltiptext tooltip-top">New version ({updateVersion}) found!</span>
```
andresmgot: I've removed this tooltip text because that info is already explained in the new info

Angelmmiguel: Sounds good to me. Do you think the tooltip feature may be used in another place of this app? We may want to keep it for other use cases. However, I see we're using it as an inline CSS style, so we will need to create a component. Definitively, it can be done in a separate task but I wanted to highlight it :)

dashboard/src/components/AppView/ChartInfo.scss
```diff
@@ -2,3 +2,13 @@
...
 }
+
+.ChartInfoUpdate {
+  color: white;
+  width: 100%;
+  margin-bottom: 0.5em;
+  text-align: center;
+  padding: 0.25em 0;
+  font-size: 14px;
+  background: #008145;
```
Angelmmiguel: Aren't we using HEx library in this project? I don't remember, honest. You can use the `color` method if it's available.

dashboard/src/components/AppView/ChartInfo.tsx
```diff
@@ -51,9 +52,10 @@ class ChartInfo extends React.Component<IChartInfoProps> {
     // If update is not set yet we cannot know if there is
     // an update available or not
     if (app.updateInfo) {
+      let updateContent = null;
       if (app.updateInfo.error) {
-        return (
-          <div>
+        updateContent = (
+          <React.Fragment>
```
Angelmmiguel: Nice! :)

dashboard/src/components/AppView/AppControls/UpgradeButton.tsx
andresmgot: right, for the moment nothing will use it so we don't need it but it's true that moving that to a component can be helpful

_makes a mental note_

dashboard/src/components/AppView/ChartInfo.scss
andresmgot: it seems we are not :)

PR #1060, opened 2019-06-19
Add installation values to AppView

### LINKED ISSUE #739: View values file used for application release
Hi,

When clicking on an application release, it would be useful to be able to view the current values that were used to deploy the chart.

For a user browsing a production enviornment, being able to view the values file that was used to deploy the chart would be informative for checks and debugging.

Thanks,
Dean.

andresmgot:
Fixes #739

Add the installation values used for the current release

Screenshot from 2019-06-19 12-23-06


prydonius:
Is there a discussion about the design of this? Just wondering why we chose to place this at the bottom of the AppView and not have a button or something that opens up a modal so it's less distracting?

andresmgot:
> Is there a discussion about the design of this? Just wondering why we chose to place this at the bottom of the AppView and not have a button or something that opens up a modal so it's less distracting?

There is not, this is just my proposal. I thought about that but since there is nothing below the installation values it should not really bother users. IMO there is no reason to "hide" it and make it a modal but I am open to suggestions.

andresmgot:
@Angelmmiguel what do you think?

Angelmmiguel:
> @Angelmmiguel what do you think?

I don't have a strong opinion of this one because it depends on the relevance of this information. If users are constantly looking for it I believe that showing it is the best approach. On the other hand, if it's not that important, we should hide it and show it "on demand". 

> Thanks for the explanation, I'm okay with merging this with it displayed here for now, though we may want to reconsider this when we approach the redesign

Agree with @prydonius :)

andresmgot:
Thank you both for the input!

Issue #848, opened 2018-11-22
Refactor table views


andresmgot:
We have several views in which we render tables like the Services, Deployments, Secrets, etc. of the AppView or the bindings in the ServiceView. Each of them has mostly the same component that represents the table but changing the columns content. We can create a generic component out of it.

I think we should also review the representation, I think that having a single box with different tabs for showing the different tables will provide a better UX in some cases (like the app resources).

prydonius:
I also think we may want to revisit whether we want tables in certain cases, there may be better ways to display some of the information we're showing.

andresmgot:
Yes, but probably to display better information we may need more metadata from the chart (like knowing which secrets are credentials). In any case, I think that having raw information about all the resources for an app is something useful. Advanced users/developers can use that to see what's going on.

prydonius:
> Advanced users/developers can use that to see what's going on.

Potentially, but I would see that level of detail as a cluster operator concern, not an app dev concern. I think it's useful to have it in a different tab for advanced info, but it should not be in the default view. Anyway, I think I'm getting away from the point of this issue, which is to create a common component we can reuse for similar views, which I do think is valuable for when we do want to show tabular information.

We may want to look at if there are any React components out there that make showing tabular data easier.

lindhe:
My Secrets table looks broken. Does this belong here, or should I open a new issue?

It seems to be some trouble with the flex class, but I know little about HTML/CSS...

![Screenshot from 2019-08-02 15-05-35](https://user-images.githubusercontent.com/7773090/62372194-1d3ee280-b537-11e9-8098-9dabd43a18be.png)

andresmgot:
yes @lindhe, that seems like a bug. Do you mind opening a different PR? we can fix the style there.

absoludity:
New UI has app resources in a standard table with tabs.

PR #1063, opened 2019-06-21
Refactor tables in the AppView

andresmgot:
Ref: #848

Refactor common code used by the different tables in the AppView: WorkloadTable, ServiceTable, SecretsTable and OtherResourcesTable.

Now everything is a `ResourceTable` and depending on the type of the resource (`kind`) it renders a different type of row.


dashboard/src/components/AppView/ResourceTable/ResourceItem/ResourceTableItem.tsx
```diff
@@ -0,0 +1,78 @@
...
+        case "Deployment":
+          return <DeploymentItemRow resource={resource.item} />;
+        case "StatefulSet":
+          return <StatefulSetItemRow resource={resource.item} />;
+        case "DaemonSet":
+          return <DaemonSetItemRow resource={resource.item} />;
+        case "Service":
+          return <ServiceItem resource={resource.item} />;
+        case "Secret":
+          // TODO: Set double type
```
andresmgot: I need to change the type here

dashboard/src/components/AppView/ResourceTable/ResourceTable.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { SecretColumns } from "./ResourceItem/SecretItem/SecretItem";
+import { ServiceColumns } from "./ResourceItem/ServiceItem/ServiceItem";
+import { StatefulSetColumns } from "./ResourceItem/StatefulSetItem/StatefulSetItem";
+
+interface IWorkloadItemTableProps {
+  title: string;
+  resourceRefs: ResourceRef[];
+}
+
+class WorkloadTable extends React.Component<IWorkloadItemTableProps> {
```
andresmgot: `ResourceTable`

dashboard/src/components/AppView/ResourceTable/ResourceItem/ResourceTableItem.tsx
```diff
@@ -0,0 +1,78 @@
...
+          const i = (resource.item as any) as ISecret;
```
prydonius: hmm maybe SecretItem should just accept IResource and internally change it to ISecret?

andresmgot: I would see it weird for the `SecretItem` to accept a `IResource` when it only works for an `ISecret`. It's better to do the conversion before. It's actually the generic "TableItem" the one that can receive different types of resources.

prydonius: You're right, I was just thinking this given that ServiceItem and other resources look like they accept IResources

PR #1072, opened 2019-07-05
WIP: Add proxy env for job repos sync

obeyler:
The aim of this pool request is to allow repository sync job to pull the repository even if they are located behind a proxy. By default the no_proxy=0.0.0.0/0 will disable all proxy so I keep the current behavior.
To realize that 
It creates a new configMap to set the http_proxy, https_proxy, no_proxy keys and mount them as env in sync job for repository


andresmgot:
hi @obeyler, thank you for the PR! Let me comment

> The aim of this pool request is to allow repository sync job to pull the repository even if they are located behind a proxy

Just to clarify, this is already possible, right? You can set env vars in a repository basis modifying the AppRepository object as specified at:

https://github.com/kubeapps/kubeapps/blob/master/docs/user/private-app-repository.md#modifying-the-synchronization-job

Taking that into account, what is your goal?

 1. Be able to easily set the http_proxy, https_proxy ans no_proxy env vars using the Kubeapps Helm chart

or

 2. Be able to set those env vars from a central place so regardless of how an AppRepository is created, Kubeapps adds those env vars

For 1. you can just modify the templates of the chart so if you specify `--set jobsRepositorySyncProxy,https_proxy=A.B.C.D` it gets translated to an environment variable as explained in the docs above.

For 2. I see the problem that different repositories may require to set different values so a global setting might not be suitable.

Can you clarify what's the use case you want to cover?

obeyler:
I aggree that it is able to do that manually by patching each job for each repository. But you cannot do that automatically. As I fully deploy my cluster and all helm chart I need to be able to have this env already in each job.

you mention : jobsRepositorySyncProxy but it doesn't exist into value.yaml

obeyler:
@andresmgot 
I see what you mean by  jobsRepositorySyncProxy even if it's not exist yet.
I do a new PR to go in this way 
Thanks

PR #1074, opened 2019-07-05
WIP: new version of PR to add a proxy for repo sync

obeyler:
Signed-off-by: olivier beyler <olivier.beyler@orange.com>


chart/kubeapps/values.yaml
```diff
@@ -105,6 +105,10 @@ apprepository:
     registry: docker.io
     repository: bitnami/kubeapps-chart-repo
     tag: 1.8.0-r0
+  proxy:
```
andresmgot: to differentiate this I would move this within the `initialRepos` key (to avoid confusion with `tiller-proxy`). I would also add a comment explaining what this does and a `enabled` field to enable/disable it:

```yaml
initialRepos:
  # Set env variables for HTTP proxy settings
  proxy:
    enabled: false
    http_proxy: ""
    ...
```

chart/kubeapps/templates/apprepositories.yaml
```diff
@@ -28,6 +28,14 @@ spec:
   {{- if $.Values.securityContext.enabled }}
   syncJobPodTemplate:
     spec:
+      containers:
```
andresmgot: (see my comment below)

to avoid setting this variables unnecessarily I would check if this is enabled: `{{- if $.Values.initialRepos.proxy.enabled }}`

obeyler:
@andresmgot could you help me:  I really don't see why ci/circleci failed even when I use  vi with set list to see all characters:
```
     30     spec:$
     31     {{- if .Values.apprepository.proxy.enabled }}$
     32       containers:$
     33       - env:$
     34           - name: https_proxy$
     35             value: {{ .Values.apprepository.proxy.https_proxy }}$
     36           - name: http_proxy$
     37             value: {{ .Values.apprepository.proxy.http_proxy }}$
     38           - name: no_proxy$
     39             value: {{ .Values.apprepository.proxy.no_proxy }}$
     40     {{- end }}$
     41       securityContext:$
     42         runAsUser: {{ $.Values.securityContext.runAsUser }}$
     43   {{- end }}$

```

chart/kubeapps/values.yaml
andresmgot: Please move this under `initialRepos` instead. If you are setting this within `apprepository` it seems that you are modifying the CRD while this only affects to the initial repositories. Also please add a comment explaining what this does.

obeyler: @andresmgot I can't do that as initialRepos value is an array of repo the key "proxy" will transform initialRepos as a map.

obeyler:
> 
Please move this under initialRepos instead. If you are setting this within apprepository it seems that you are modifying the CRD while this only affects to the initial repositories. Also please add a comment explaining what this does.


I can't do that @andresmgot as initialRepos is an array. if I put proxy it will convert it as a map

chart/kubeapps/templates/apprepositories.yaml
```diff
@@ -28,6 +28,16 @@ spec:
   {{- if $.Values.securityContext.enabled }}
```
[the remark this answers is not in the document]
obeyler: are you sure ?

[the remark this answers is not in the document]
andresmgot: yes I am, if not and the securityContext is not enabled it won't add the proxy settings

chart/kubeapps/values.yaml
andresmgot: you are right, yes then `initialReposProxy` should be fine

andresmgot:
Thanks for the changes @obeyler! see my replies

chart/kubeapps/templates/apprepositories.yaml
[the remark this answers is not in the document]
obeyler: I agree I've done modification to solve that

chart/kubeapps/templates/apprepositories.yaml
```diff
@@ -28,6 +28,16 @@ spec:
...
+{{- if .Values.apprepository.proxy.enabled }}
+      containers:
+      - env:
+          - name: https_proxy
+            value: {{ .Values.apprepository.proxy.https_proxy }}
+          - name: http_proxy
+            value: {{ .Values.apprepository.proxy.http_proxy }}
+          - name: no_proxy
+            value: {{ .Values.apprepository.proxy.no_proxy }}
+{{- end }}
```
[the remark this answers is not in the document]
obeyler: It was due to invisible char I found it and remove it (It can only be seen when a git clone is done on linux from scratch)

obeyler:
@andresmgot Everything should be ok now :-D

PR #1075, opened 2019-07-07
Add Doc for using Harbor as Helm chart repository

### LINKED ISSUE #1069: Document how to use Harbor with Kubeapps
Our docs to set up a private Helm registry with Kubeapps cover Chartmuseum and Artifactory. We can include how to install and use Harbor as Helm registry as well.

The source is here:

https://github.com/kubeapps/kubeapps/blob/master/docs/user/private-app-repository.md

jessehu:
Fix #1069.


docs/user/private-app-repository.md
```diff
@@ -57,6 +58,46 @@ It is possible to configure ChartMuseum to use authentication with two different
...
 
+## Harbor
+
+[Harbor](https://github.com/goharbor/harbor) is an open source trusted cloud native registry project that stores, signs, and scans content, e.g. Docker images. Harbor is hosted by the [Cloud Native Computing Foundation](https://cncf.io/). Since version 1.6.0, Harbor is a composite cloud native registry which supports both container image management and Helm chart management. Harbor integrates [ChartMuseum](https://chartmuseum.com) to provide the Helm chart repository functionality. The access to Helm Charts in a Harbor Chart Repository can be controlled via Role-Based Access Control.
+
+To use Harbor with Kubeapps, first deploy Harbor using [Harbor offline installer](https://github.com/goharbor/harbor/blob/master/docs/installation_guide.md#downloading-the-installer). Here are the minimum steps required on a Linux machine to install Harbor for serving as Helm Chart Repository.
+```
+$ wget https://storage.googleapis.com/harbor-releases/release-1.8.0/harbor-offline-installer-v1.8.1.tgz
```
andresmgot: what's the difference between using this method or using the helm chart from https://github.com/goharbor/harbor-helm?

docs/user/private-app-repository.md
```diff
@@ -57,6 +58,46 @@ It is possible to configure ChartMuseum to use authentication with two different
...
+$ tar xvf harbor-offline-installer-v1.8.1.tgz
+$ cd harbor
+$ sed -i 's/hostname: reg.mydomain.com/hostname: <Current-Machine-IP>/' harbor.yml
```
andresmgot: can you explain how you can get this `Current-Machine-IP`?

docs/user/private-app-repository.md
jessehu: Both offline-installer and Harbor helm chart work well. I just tried the offline-installer way which is easy to use and doesn't require k8s. BTW the release of Harbor helm chart will be a little behind the release of Harbor offline-installer.

docs/user/private-app-repository.md
jessehu: Since "Here are the minimum steps required on a Linux machine to install Harbor", the `Current-Machine-IP` means the IP of this Linux machine. We can run ifconfig to get its IP.

docs/user/private-app-repository.md
andresmgot: Can you mention the chart then too? It's not necessary to explain all the other steps but for example for me it's easier to deploy the chart and use the k8s Service to access the service rather than an IP. Just to let users know that's also an option.

jessehu: Sure, good suggestion. I have pushed the change which includes Harbor Helm Chart.

jessehu: We can add the detail guide for deploying Harbor Helm Chart in future.

migmartri:
It might make sense to document how to configure authentication since in other providers we do so. i.e a new section called `Harbor Authentication/Authorization`

I'd also sort the providers alphabetically. Artifactory, ChartMuseum, Harbor.

jessehu:
@migmartri Sure, I will add a new section `Harbor Authentication/Authorization`. The private project in Harbor requires authentication, while the public project doesn't.
The order of current 3 providers might be due to ChartMuseum and Harbor are both open source projects, and JFrog Artifactory Pro is a commercial product. JFrog Artifactory Open Source edition doesn't support Helm Chart repository.

migmartri:
Thanks @jessehu!

 > The order of current 3 providers might be due to ChartMuseum and Harbor are both open source projects, and JFrog Artifactory Pro is a commercial product. JFrog Artifactory Open Source edition doesn't support Helm Chart repository.

Good point, ignore my comment then :)

PR #1076, opened 2019-07-08
Revamp app repository form

### LINKED ISSUE #911: Improve AppRepository authentication input for Basic and Bearer token auth
The two most common Authorization headers to set for a repository are Basic auth and Bearer token auth. Currently we expect users to specify the whole Authorization header, including the Basic/Bearer prefix, which can be error prone (#908). We could improve the experience by having a dropdown of options and different form inputs to deal with each case:

- Basic
  - Show a username and password field
- Bearer
  - Show a field for the token
- Custom
  - Show the full custom Authorization header (e.g. as we have it today)

Selecting a different option would change the fields shown on the form. For Basic auth, the correct header will be calculated from the username and password input. For Bearer auth, the `Bearer` prefix will be prepended to the user-supplied token. For the Custom input, the full Authorization header will be sent as is (this will be useful for protocols other than Bearer and Basic).

andresmgot:
Fixes #911

Revamp the `AppRepoForm` so it's easier to add a bearer token or username/password:

Screenshot from 2019-07-08 16-02-50

I have moved the `AppRepoForm` to its own file as a React Component with state.

[TODO]

 - [ ] Adapt screenshots and docs for private repositories


dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
```diff
@@ -0,0 +1,259 @@
...
+                        type="radio"
+                        name="auth"
+                        value="HTTP Header"
+                        onChange={this.handleAuthRadioButtonChange}
+                      />
+                      Custom
+                      <br />
+                    </label>
+                  </div>
+                  <div className="col-9">
```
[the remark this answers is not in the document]
Angelmmiguel: Also, since these elements are going to appear and disappear, please add the `aria-live="polite` property to notify screen readers about these changes.

dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
```diff
@@ -0,0 +1,259 @@
...
+                        name="auth"
+                        value="None"
+                        defaultChecked={true}
+                        onChange={this.handleAuthRadioButtonChange}
+                      />
+                      None
+                      <br />
+                    </label>
+                    <label>
+                      <input
```
Angelmmiguel: Setup the `id` of the different inputs to connect them with their `label`

dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
```diff
@@ -0,0 +1,259 @@
...
+                        type="radio"
+                        name="auth"
+                        value="None"
+                        defaultChecked={true}
+                        onChange={this.handleAuthRadioButtonChange}
+                      />
+                      None
+                      <br />
+                    </label>
+                    <label>
```
Angelmmiguel: Set the value of the `htmlFor` property to the `id` of its input. This will connect both for better accessibility and allow users to change the radio when clicking on the text

dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
```diff
@@ -0,0 +1,259 @@
...
+  public render() {
+    return (
+      <form className="container padding-b-bigger" onSubmit={this.handleInstallClick}>
+        <div className="row">
+          <div className="col-12">
+            <div>
+              <h2>Add an App Repository</h2>
+            </div>
+            <div>
+              <label>
```
Angelmmiguel: Same comment about the usage of `htmlFor` and `id` on the `input`. Btw, you don't need to add the input inside the `label` for text and selectors:

`<label htmlFor="name">
  Name:
</label>
<input
  type="text"
  id="name"
  ...
/>
```

andresmgot:
@Angelmmiguel PTAL

PR #1083, opened 2019-07-11
Add doc for enabling Harbor authentication

### LINKED ISSUE #1069: Document how to use Harbor with Kubeapps

jessehu:
Fix #1069.
@migmartri @andresmgot please help review.

### DIFF
--- a/docs/user/private-app-repository.md
+++ b/docs/user/private-app-repository.md
@@ -101,6 +101,13 @@ To add Harbor as the private chart repository, go to `Configuration > App Reposi
 
 Once you create the repository you can click on the link for the specific repository and you will be able to deploy your own applications using Kubeapps.
 
+### Harbor: Authentication/Authorization
+
+It is possible to configure Harbor to use HTTP basic authentication:
+
+  - When creating a new project for serving as the helm chart repository in Harbor, set the `Access Level` of the project to non public. This enforces authentication to access the charts in the chart repository via Helm CLI or other clients.
+  - When `Adding App Repository` in Kubeapps, select `Basic Auth` for `Authorization` and specifiy the username and password for Harbor.
+
 ## Artifactory
 
 JFrog Artifactory is a Repository Manager supporting all major packaging formats, build tools and CI servers.


docs/user/private-app-repository.md
```diff
@@ -101,6 +101,14 @@ To add Harbor as the private chart repository, go to `Configuration > App Reposi
 
 Once you create the repository you can click on the link for the specific repository and you will be able to deploy your own applications using Kubeapps.
 
+### Harbor: Authentication/Authorization
+
+It is possible to configure Harbor to use authentication:
+
+- Using HTTP [basic authentication](https://github.com/goharbor/harbor/blob/master/docs/user_guide.md#working-with-helm-cli). To use this feature:
```
andresmgot: if using basic authentication is the only option you can remove the list. Also the link is pointing to `working-with-helm-cli` that doesn't seem related. Is that the link you wanted?

docs/user/private-app-repository.md
```diff
@@ -101,6 +101,14 @@ To add Harbor as the private chart repository, go to `Configuration > App Reposi
...
+  - When creating a new project for serving as the helm chart repository in Harbor, set the `Access Level` of the project to non public. This enforces authentication to access the charts in the chart repository via Helm CLI or other clients.
```
andresmgot: can you add a screenshot for this?

docs/user/private-app-repository.md
```diff
@@ -101,6 +101,14 @@ To add Harbor as the private chart repository, go to `Configuration > App Reposi
...
+  - When `Adding App Repository` in Kubeapps, set the `Authorization Header` to the `Basic` auth header using the base64 encoded string of `user:password`. For example, for the user "admin" and password "Harbor12345", use the following auth header: `Basic YWRtaW46SGFyYm9yMTIzNDU=`.
```
andresmgot: I have just merged a new feature that allows you to just select "Basic Auth" in the form and fill username and password (see other screenshots)

docs/user/private-app-repository.md
jessehu: That link points to the doc about how to use 'basic authentication' in Helm CLI to authenticate Harbor chart repository. It's similiar to `Using HTTP [basic authentication](https://chartmuseum.com/docs/#basic-auth)`. We can remove the link if not proper?
There should be other authentication ways for Harbor. We can test it and add here later. The doc above only shows Basci Auth.

docs/user/private-app-repository.md
jessehu: @andresmgot, Got it. This new feature is good. I will update the above statement.

docs/user/private-app-repository.md
jessehu: @andresmgot The screenshot is https://raw.githubusercontent.com/kubeapps/kubeapps/master/docs/img/harbor-new-project.png (attached in the above step for creating new project in Harbor) which shows 'public access'. Uncheck 'Public' means non public access.  We can attach another snapshot with 'Public' unchecked, but it seems tedious. Any thought?

docs/user/private-app-repository.md
andresmgot: I see, those docs also talks about using a SSL cert which may be confusing, I would leave the link out because it's not really relevant in this context.

For the format I would merge this line with the above:

```
It is possible to configure Harbor to use authentication using HTTP basic authentication ...
```

docs/user/private-app-repository.md
andresmgot: I see, from the wording it seemed a different thing for me, it's okay then.

docs/user/private-app-repository.md
jessehu: Thanks. This makes sense. Updated in new commit.

PR #1085, opened 2019-07-15
Add JobTemplate to web UI

### LINKED ISSUE #1053: SecurityContext not propagating to AppRepoSync Cronjob
When security concept is set in values.yaml file, it's propagated to all pods, but the apprepo-sync-chartmuseum cronjob does not inherit the runAsUser value and therefore runs as root.

```
apiVersion: batch/v1beta1
kind: CronJob
metadata:
  creationTimestamp: 2019-06-07T20:03:41Z
  name: apprepo-sync-chartmuseum
  namespace: dev-whitestar-kubeapps-dev
  ownerReferences:
  - apiVersion: kubeapps.com/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: AppRepository
    name: chartmuseum
    uid: 3f22b1e7-894f-11e9-b978-42010aeec16a
  resourceVersion: "9228438"
  selfLink: /apis/batch/v1beta1/namespaces/dev-whitestar-kubeapps-dev/cronjobs/apprepo-sync-chartmuseum
  uid: 58a3d19f-895f-11e9-b978-42010aeec16a
spec:
  concurrencyPolicy: Replace
  failedJobsHistoryLimit: 1
  jobTemplate:
    metadata:
      creationTimestamp: null
    spec:
      template:
        metadata:
          creationTimestamp: null
          labels:
            apprepositories.kubeapps.com/repo-name: chartmuseum
        spec:
          containers:
          - args:
            - sync
            - --mongo-url=kubeapps-flux-mongodb
            - --mongo-user=root
            - --user-agent-comment=kubeapps/v1.4.0
            - chartmuseum
            - http://chartmuseum-chartmuseum.apps.svc.cluster.local:8080
            command:
            - /chart-repo
            env:
            - name: MONGO_PASSWORD
              valueFrom:
                secretKeyRef:
                  key: mongodb-root-password
                  name: kubeapps-mongodb
            image: bitnami/kubeapps-chart-repo:1.8.0
            imagePullPolicy: IfNotPresent
            name: sync
            resources: {}
            terminationMessagePath: /dev/termination-log
            terminationMessagePolicy: File
          dnsPolicy: ClusterFirst
          restartPolicy: OnFailure
          schedulerName: default-scheduler
          securityContext: {}
          terminationGracePeriodSeconds: 30
  schedule: 0 * * * *
  successfulJobsHistoryLimit: 3
  suspend: false
status: {}
```

As you can see securityContext in the CronJob definition is set to {}, while every other pods are getting the defined securityContext.

andresmgot:
Closes #1053

This PR adds a new field in the AppRepository form to include a Pod Template in case it's needed as explained in https://github.com/kubeapps/kubeapps/blob/master/docs/user/private-app-repository.md#modifying-the-synchronization-job

Screenshot from 2019-08-01 17-45-05

I have also added the `Hint` component to give more information and point to the documentation:

Screenshot from 2019-08-07 11-03-26


dashboard/src/shared/App.test.ts
```diff
@@ -93,9 +93,15 @@ describe("App", () => {
     });
     it("throws an error if returns an error 404", async () => {
       moxios.stubRequest(/.*/, { status: 404 });
-      expect(App.delete("foo", "default", false)).rejects.toThrow(
-        "Request failed with status code 404",
-      );
+      let errored = false;
```
andresmgot: this change is not related to the PR, I am not sure why when I was running test, this was failing so I had to rewrite it.

dashboard/src/actions/repos.test.tsx
```diff
@@ -248,6 +252,27 @@ describe("installRepo", () => {
...
       expect(res).toBe(true);
     });
+
+    context("when a pod template is provided", () => {
+      const installRepoCMDPodTemplate = repoActions.installRepo(
+        "my-repo",
+        "http://foo.bar",
+        "",
+        "",
+        "foo: bar",
```
Angelmmiguel: This is not a real example, right? I see we've some examples in our docs so maybe we can reuse them as the content for this argument:

```
apiVersion: kubeapps.com/v1alpha1
kind: AppRepository
metadata:
  name: my-repo
  namespace: kubeapps
spec:
  syncJobPodTemplate:
    metadata:
      labels:
        my-repo: "isPrivate"
    spec:
      containers:
        - env:
            - name: FOO
              value: BAR
  url: https://my.charts.com/```

File: dashboard/src/components/Hint/Hint.scss [hunk 3]
```diff
@@ -0,0 +1,14 @@
+.extraClass {
```
Angelmmiguel: Why do you need to apply `!important` to all the CSS classes? Please, avoid the usage of `!important` when it's not required because it breaks the CSS cascade feature.

dashboard/src/components/Hint/Hint.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+export class Hint extends React.Component<IHintProps> {
+  public render() {
+    return (
+      <React.Fragment>
+        <a data-tip={true} data-for="syncJobHelp">
+          <HelpCircle className="icon" color="white" fill="#5F6369" />
+        </a>
+        <ReactTooltip
+          id="syncJobHelp"
```
Angelmmiguel: Make this `id` customizable. Identifiers must be unique per page. If we add more than one `Hint` component per page we will have duplicated identifiers

dashboard/src/components/Hint/Hint.scss
Angelmmiguel: I think so. Probably they're adding the `!important` _just in case_

dashboard/src/actions/repos.ts
```diff
@@ -165,8 +167,18 @@ export const installRepo = (
           secrets["ca.crt"] = btoa(customCA);
         }
       }
+      let syncJobPodTemplateObj = {};
+      if (syncJobPodTemplate.length) {
+        syncJobPodTemplateObj = yaml.load(syncJobPodTemplate);
```
absoludity: This needs to be `yaml.safeLoad()`. I fixed this in #1118 with a test to show why :)

PR #1086, opened 2019-07-16
Update chart repo

### LINKED ISSUE #617: chart-repo: support case insensitive readme file in charts
Do you know why some chart hasn't got any description when click on cartrige ? I suppose it's due to the chart himself but could you point what they don't have to be able to provide it ?

### LINKED ISSUE #1070: Reduce time between chart repo syncs
The default time to sync chart registries is every hour. This mean that without refreshing the repository manually, a new version/chart can be missing for this hour gap.

We can reduce the frequency of the CronJob to run it every minute but the `chart-repo` job is a bit slow (around 30s for the stable repository). To achieve that we can store the checksum of the repository in the database and only run the heavyweight logic (parse the repository, download the logos, update the database...) if the content of the repository has changed.

andresmgot:
Closes #1070
Closes #617

Updates the `chart-repo` image to the latest version. This allows us to execute as often as wanted. 

As an example, the `stable` repository takes ~35s to do a full sync while with the new changes it takes just about a second.

### DIFF
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -104,7 +104,7 @@ apprepository:
   syncImage:
     registry: docker.io
     repository: bitnami/kubeapps-chart-repo
-    tag: 1.8.1-r0
+    tag: 1.8.2-r0
   initialReposProxy:
     enabled: false
 #    http_proxy: "http://yourproxy:3128"
@@ -186,7 +186,7 @@ chartsvc:
   image:
     registry: docker.io
     repository: bitnami/kubeapps-chartsvc
-    tag: 1.8.1-r0
+    tag: 1.8.2-r0
   service:
     port: 8080
   # https://github.com/kubeapps/kubeapps/issues/478#issuecomment-422979262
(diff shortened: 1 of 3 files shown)


jessehu:
Hi @andresmgot , I'd like to share some comments:
1) Can we update the chart-repo tag to 1.8.2-r0 which was released today?
2)  Syncing the app repo every 1 second seems too frequent. The previous sync process should take several seconds or even 1 minute.

andresmgot:
> 1. Can we update the chart-repo tag to 1.8.2-r0 which was released today?

Yes, I will do that, I am just waiting for the mirror to get populated.

> Syncing the app repo every 1 second seems too frequent. The previous sync process should take several seconds or even 1 minute.

This changes the cronjob to be executed every minute (not every second). It still may be to much, we are discussing that in #1070

jessehu:
Thanks Andres. I meant to say 1 minute. Is 1 minute too frequent?

andresmgot:
> Is 1 minute too frequent?

From Kubeapps point of view it's not. A sync would take <5s if there are no updates and <40s if there are new things. From the chart repository point of view it may be, having N number of requests each minute (where N is the number of Kubeapps installations) may be a problem, that's what we are discussing in #1070

jessehu:
10 minutes interval looks good 👍

PR #1088, opened 2019-07-19
Update chart-repo.md

jessehu:


docs/developer/chart-repo.md
```diff
@@ -36,11 +36,11 @@ Kubeapps is a Kubernetes-native application. To develop and test Kubeapps compon
 
 ```bash
 cd $MONOCULAR_DIR
-dep ensure
+go mod tidy
```
andresmgot: did you need to run this when working with monocular?

jessehu: I think so. Without the packages in $GOPATH/pkg/mod/ downloaded by 'go mod tidy', monocular code cannot compile.

andresmgot: when using go modules, any command (like `go build` or `go test`) will download the required modules. `go mod tidy` is used to remove unnecessary dependencies from the `go.mod` file, that's why I found it weird that you needed that.

In any case, the `chart-repo` is meant to be built using it's Dockerfile (and it's built within a container) so running `go mod tidy` won't have any effect.

jessehu: Yes, you're right.  I referred to the monocular doc as below and 'dep ensure' will report error `could not find project Gopkg.toml`.

https://github.com/helm/monocular/blob/master/docs/development.md#developing-chartsvc
image

andresmgot: ah, I see, I don't think it's necessary but we can live it that way for consistency. Thanks!

PR #1089, opened 2019-07-22
Fix @types/json-schema version in dashboard/package.json

jessehu:
The @types/json-schema version in dashboard/package.json is out of date. So make it consistent with the version in yarn.lock.

### DIFF
--- a/dashboard/yarn.lock
+++ b/dashboard/yarn.lock
@@ -243,15 +243,10 @@
   resolved "https://registry.yarnpkg.com/@types/js-yaml/-/js-yaml-3.12.1.tgz#5c6f4a1eabca84792fbd916f0cb40847f123c656"
   integrity sha512-SGGAhXLHDx+PK4YLNcNGa6goPf9XRWQNAUUbffkwVGGXIxmDKWyGGL4inzq2sPmExu431Ekb9aEMn9BkPqEYFA==
 
-"@types/json-schema@*":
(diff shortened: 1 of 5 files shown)


andresmgot:
I think you need to run `yarn upgrade ...` and upload both the package.json and yarn.lock changes for this to work.

jessehu:
I tried 'yarn upgrade' and it didn't update package.json. So I think there is no need for 'yarn upgrade', but updating package.json separately.

dashboard/yarn.lock
```diff
@@ -5,14 +5,12 @@
 "@babel/code-frame@^7.0.0", "@babel/code-frame@^7.0.0-beta.35":
   version "7.0.0"
   resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.0.0.tgz#06e2ab19bdb535385559aabb5ba59729482800f8"
-  integrity sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==
```
andresmgot: I think  you need to upgrade `yarn`, this has deleted all the integrity info

jessehu: I used yarn 1.7.0. Now using v1.17.3 and the integrity info is kept. Please review again.

PR #1091, opened 2019-07-23
Add 'Refresh All App Repositories' button

### LINKED ISSUE #582: [Feature Request] Add refresh all repositories to the repositories page....
Hi,

Ultimately it would just do a `helm repo update` under the skin.

jessehu:
Close #582. Add a button next to 'Add App Repository'
to refresh all app repositories.

image

### DIFF
--- a/dashboard/src/actions/repos.ts
+++ b/dashboard/src/actions/repos.ts
@@ -101,6 +101,30 @@ export const resyncRepo = (
   };
 };
 
+export const resyncAllRepos = (): ThunkAction<Promise<void>, IStoreState, null, AppReposAction> => {
+  return async (dispatch, getState) => {
+    try {
+      const {
+        config: { namespace },
+      } = getState();
+      // TODO: Do something to show progress
+      const repos = await AppRepository.list(namespace);
+      if (repos.items.length > 0) {
+        repos.items.forEach(repo => {
+          if (repo && repo.spec) {
+            repo.spec.resyncRequests = repo.spec.resyncRequests || 0;
+            repo.spec.resyncRequests++;
+            const name = repo.metadata.name;
+            AppRepository.update(name, namespace, repo);
+          }
+        });
+      }
+    } catch (e) {
+      dispatch(errorRepos(e, "update"));
+    }
+  };
+};
+
 export const fetchRepos = (): ThunkAction<Promise<void>, IStoreState, null, AppReposAction> => {
   return async (dispatch, getState) => {
     dispatch(requestRepos());
--- a/dashboard/src/actions/repos.ts
+++ b/dashboard/src/actions/repos.ts
@@ -92,36 +92,19 @@ export const resyncRepo = (
       repo.spec.resyncRequests++;
       await AppRepository.update(name, namespace, repo);
       // TODO: Do something to show progress
-      dispatch(requestRepos());
-      const repos = await AppRepository.list(namespace);
-      dispatch(receiveRepos(repos.items));
     } catch (e) {
       dispatch(errorRepos(e, "update"));
     }
   };
 };
 
-export const resyncAllRepos = (): ThunkAction<Promise<void>, IStoreState, null, AppReposAction> => {
+export const resyncAllRepos = (
+  repoNames: string[],
+): ThunkAction<Promise<void>, IStoreState, null, AppReposAction> => {
   return async (dispatch, getState) => {
-    try {
-      const {
-        config: { namespace },
-      } = getState();
-      // TODO: Do something to show progress
-      const repos = await AppRepository.list(namespace);
-      if (repos.items.length > 0) {
-        repos.items.forEach(repo => {
-          if (repo && repo.spec) {
-            repo.spec.resyncRequests = repo.spec.resyncRequests || 0;
-            repo.spec.resyncRequests++;
-            const name = repo.metadata.name;
-            AppRepository.update(name, namespace, repo);
-          }
-        });
-      }
-    } catch (e) {
-      dispatch(errorRepos(e, "update"));
-    }
+    repoNames.forEach(name => {
+      dispatch(resyncRepo(name));
+    });
   };
 };
 
--- a/dashboard/src/components/Config/AppRepoList/AppRepoButton.tsx
+++ b/dashboard/src/components/Config/AppRepoList/AppRepoButton.tsx
@@ -43,7 +43,7 @@ export class AppRepoAddButton extends React.Component<
   public render() {
     const { redirectTo } = this.props;
     return (
-      <div className="AppRepoAddButton">
+      <span>
         <button className="button button-primary" onClick={this.openModal}>
           Add App Repository
         </button>
@@ -64,7 +64,7 @@ export class AppRepoAddButton extends React.Component<
           <AppRepoForm install={this.install} onAfterInstall={this.closeModal} />
         </Modal>
         {redirectTo && <Redirect to={redirectTo} />}
-      </div>
+      </span>
     );
   }
 
--- a/dashboard/src/components/Config/AppRepoList/AppRepoButton.tsx
+++ b/dashboard/src/components/Config/AppRepoList/AppRepoButton.tsx
@@ -43,7 +43,7 @@ export class AppRepoAddButton extends React.Component<
   public render() {
     const { redirectTo } = this.props;
     return (
-      <span>
+      <React.Fragment>
         <button className="button button-primary" onClick={this.openModal}>
           Add App Repository
         </button>
@@ -64,7 +64,7 @@ export class AppRepoAddButton extends React.Component<
           <AppRepoForm install={this.install} onAfterInstall={this.closeModal} />
         </Modal>
         {redirectTo && <Redirect to={redirectTo} />}
-      </span>
+      </React.Fragment>
     );
   }
 
--- a/dashboard/src/components/Config/AppRepoList/AppRepoList.tsx
+++ b/dashboard/src/components/Config/AppRepoList/AppRepoList.tsx
@@ -4,6 +4,7 @@ import { IAppRepository, IRBACRole } from "../../../shared/types";
 import ErrorSelector from "../../ErrorAlert/ErrorSelector";
 import { AppRepoAddButton } from "./AppRepoButton";
 import { AppRepoListItem } from "./AppRepoListItem";
+import { AppRepoRefreshAllButton } from "./AppRepoRefreshAllButton";
 
 export interface IAppRepoListProps {
   errors: {
@@ -16,6 +17,7 @@ export interface IAppRepoListProps {
   fetchRepos: () => void;
   deleteRepo: (name: string) => Promise<boolean>;
   resyncRepo: (name: string) => void;
+  resyncAllRepos: () => void;
   install: (name: string, url: string, authHeader: string, customCA: string) => Promise<boolean>;
   kubeappsNamespace: string;
 }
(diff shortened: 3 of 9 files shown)


dashboard/src/shared/types.ts
```diff
@@ -277,6 +277,7 @@ export interface IAppRepository
           };
         };
       };
+      resyncRequests: number;
```
jessehu: Without this definition, the code won't compile.

dashboard/src/components/Config/AppRepoList/AppRepoRefreshAllButton.tsx
```diff
@@ -0,0 +1,31 @@
...
+          title="Refresh All App Repositories"
+        >
+          Refresh All
+        </button>
+      </span>
+    );
+  }
+
+  private handleResyncAllClick() {
+    return () => {
```
jessehu: I tried to use 'disabled' property of the button to avoid multiple clicks, but got problem to set/remove 'disabled' attribute.

dashboard/src/actions/repos.ts
```diff
@@ -101,6 +101,30 @@ export const resyncRepo = (
...
+export const resyncAllRepos = (): ThunkAction<Promise<void>, IStoreState, null, AppReposAction> => {
+  return async (dispatch, getState) => {
+    try {
+      const {
+        config: { namespace },
+      } = getState();
+      // TODO: Do something to show progress
+      const repos = await AppRepository.list(namespace);
+      if (repos.items.length > 0) {
+        repos.items.forEach(repo => {
```
andresmgot: here it's better to dispatch calls to the other method: `dispatch(resyncRepo(repo.metadata.name))` to avoid repeating code.

dashboard/src/components/Config/AppRepoList/AppRepoButton.tsx
```diff
@@ -43,7 +43,7 @@ export class AppRepoAddButton extends React.Component<
   public render() {
     const { redirectTo } = this.props;
     return (
-      <div className="AppRepoAddButton">
+      <span>
```
andresmgot: why did you change this for a span?

dashboard/src/components/Config/AppRepoList/AppRepoRefreshAllButton.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from "react";
+
+import "./AppRepo.css";
+
+interface IAppRepoRefreshAllButtonProps {
+  error?: Error;
```
andresmgot: you are not using the `error` nor the `kubeappsNamespace` properties, you can skip those.

dashboard/src/components/Config/AppRepoList/AppRepoRefreshAllButton.tsx
andresmgot: let's leave it for now without showing progress

dashboard/src/actions/repos.ts
prydonius: +1, dispatching it will also make each resync request happen in parallel

dashboard/src/actions/repos.ts
```diff
@@ -101,6 +101,30 @@ export const resyncRepo = (
...
 };
 
+export const resyncAllRepos = (): ThunkAction<Promise<void>, IStoreState, null, AppReposAction> => {
+  return async (dispatch, getState) => {
+    try {
+      const {
+        config: { namespace },
+      } = getState();
+      // TODO: Do something to show progress
+      const repos = await AppRepository.list(namespace);
```
prydonius: I'm not sure we need to refresh the list of repos here, could we not just pass the list of repos we've already fetched from the component, or just get them from the state??

prydonius:
This is looking great @jessehu, just had a few thoughts

dashboard/src/actions/repos.ts
jessehu: resyncRepo(name) function will make a GET and UPDATE API call, then a call to `AppRepository.list(namespace)`, so using it here will make many duplicate `AppRepository.list` calls.  Any more idea?
```
      const repo = await AppRepository.get(name, namespace);
      repo.spec.resyncRequests = repo.spec.resyncRequests || 0;
      repo.spec.resyncRequests++;
      await AppRepository.update(name, namespace, repo);
      // TODO: Do something to show progress
      dispatch(requestRepos());
      const repos = await AppRepository.list(namespace);
```

dashboard/src/components/Config/AppRepoList/AppRepoButton.tsx
jessehu: If using <div>, the two buttons will show in two lines.

dashboard/src/components/Config/AppRepoList/AppRepoRefreshAllButton.tsx
jessehu: Yes, I will remove it.

dashboard/src/actions/repos.ts
jessehu: In case the backend repo data changed, it's meaningful to get the repos from backend again. Otherwise, for example if the backend repo has repo.spec.resyncRequests increased by 1, then the sync repo action here doesn't take effect.

However if we don't consider this case, it's OK to not to call `AppRepository.list(namespace);`. BTW could you guide me how to get it from the state please? I'm very new to ReactJS :)

dashboard/src/components/Config/AppRepoList/AppRepoButton.tsx
andresmgot: I see, in this case then it's better to use a component call `<React.Fragment>` that allows you to group elements

dashboard/src/actions/repos.ts
andresmgot: To be honest, I don't know why we are listing the repositories in the `resyncRepo` method, it's not doing anything. You can remove the lines 95-97 so we don't list them several times.

dashboard/src/actions/repos.ts
andresmgot: so, in the component `AppRepoList`, there is a property `repos`. That's already the list of repositories. You can forward that list to `AppRepoRefreshAllButton` and then make this method accept a string array so you don't need to list again.

jessehu: Nice. Let me try that.

dashboard/src/actions/repos.ts
jessehu: I guess it is to reload the updated repos list?

dashboard/src/components/Config/AppRepoList/AppRepoButton.tsx
jessehu: I found that the 'Delete' and 'Refresh' buttons in AppRepoListItem.tsx simply use two `<button>` in a `<td>`. How about using `<span>` here or remove `<span>`?

dashboard/src/actions/repos.ts
prydonius: Hmm, good catch @jessehu, we definitely don't want to be listing multiple times. I think what we could do here is remove the last few lines from resyncRepo and reuse fetchRepos when calling this action. We can update RepoListContainer to do this:

https://github.com/kubeapps/kubeapps/blob/cc186fc257eb972326752b6f1b9960e5142e0330/dashboard/src/containers/RepoListContainer/RepoListContainer.ts#L28-L30

```
    resyncRepo: async (name: string) => {
      await dispatch(actions.repos.resyncRepo(name));
      return dispatch(actions.repos.fetchRepos());
    },
```

and then do the same for resyncAllRepos. What do you think?

dashboard/src/components/Config/AppRepoList/AppRepoButton.tsx
prydonius: We should try to use something other than `span`, since `span` is an inline element and shouldn't really be wrapping buttons. You might be able to use flexbox (https://design.bitnami.com/category/Foundations/Helpers/#Flex%20Align) in the AppRepoList to make these buttons show up next to each other, but if you aren't familiar with that then I would also suggest trying to use `React.Fragment`.

We can't just use a button element here because of the Modal.

jessehu: Thanks. Let me try `React.Fragment`.

dashboard/src/actions/repos.ts
jessehu: @prydonius With the new code above, the logic will be `resyncAllRepos` -> `dispatch(resyncRepo(name));`  -> `dispatch(actions.repos.fetchRepos());`, so seems it stills calls fetchRepos() multiple times ?  For now, is fetchRepos() needed after a resyncRepo(name)?  We only show repo name and URL on the UI, so I think it won't make any changes on UI unless there is a new app repo added/deleted in the backend by another kubeapps web UI client, but this is not expected goal of refreshing a single repo.

dashboard/src/components/Config/AppRepoList/AppRepoButton.tsx
jessehu: Done. Now using `React.Fragment`.

dashboard/src/actions/repos.ts
jessehu: Done.

dashboard/src/actions/repos.ts
```diff
@@ -91,16 +91,22 @@ export const resyncRepo = (
       repo.spec.resyncRequests = repo.spec.resyncRequests || 0;
       repo.spec.resyncRequests++;
       await AppRepository.update(name, namespace, repo);
-      // TODO: Do something to show progress
```
andresmgot: please don't delete the TODO yet

dashboard/src/components/Config/AppRepoList/AppRepoButton.tsx
```diff
@@ -43,7 +43,7 @@ export class AppRepoAddButton extends React.Component<
   public render() {
     const { redirectTo } = this.props;
     return (
-      <div className="AppRepoAddButton">
+      <div>
```
andresmgot: I meant to use `React.Fragment` here (instead of div) not in the AppRepoList

dashboard/src/components/Config/AppRepoList/AppRepoList.tsx
```diff
@@ -87,11 +97,22 @@ class AppRepoList extends React.Component<IAppRepoListProps> {
             ))}
           </tbody>
         </table>
-        <AppRepoAddButton
-          error={errors.create}
-          install={install}
-          kubeappsNamespace={kubeappsNamespace}
-        />
+        <React.Fragment>
+          <td>
```
andresmgot: You should not need `React.Fragment` and `td` here.

dashboard/src/components/Config/AppRepoList/AppRepoRefreshAllButton.tsx
```diff
@@ -0,0 +1,35 @@
...
+          title="Refresh All App Repositories"
+        >
+          Refresh All
+        </button>
+      </div>
+    );
+  }
+
+  private handleResyncAllClick() {
+    return () => {
```
andresmgot: you should be able to use `onClick={this.handleResyncAllClik}` so you don't need to return a function here.

dashboard/src/actions/repos.ts
jessehu: Done

dashboard/src/components/Config/AppRepoList/AppRepoRefreshAllButton.tsx
jessehu: Done

PR #1095, opened 2019-07-29
Add dashboard/.devcontainer/ with setup for containerized developing and testing

absoludity:
Not necessarily for merging, but just to ask some questions and get a bit of feedback as to whether I should bother continuing (to get telepresence via the docker container working).

Background: I took a look at setting up for developing on KubeApps (dashboard in particular) it seemed to require quite a lot of manual setup and installation of things I didn't really want to install locally. At the same time, I'd recently been playing with VSCode's new support for remote/container development, so wanted to provide something which works by default with the normal tools (`docker-compose`) but can also be extended simply to integrate with any editor such as VSCode. This is the result.

The contents should be pretty self-explanatory, and the included `dashboard/.devcontainer/README.md` hopefully summarises the functionality (I've not advertised this functionality in the main README for now).

I'm keen to know if people are using telepresence for doing local development? Not sure yet but I assume it's necessary/encouraged because as an easy way to give the dashboard all the backend data it needs to run locally? I guess this requires each dev having their own kubeapps deployment in a cluster (or perhaps it's light enough to run via minikube? Haven't tried). Are there other options that people use or have thought about?

To be clear: this isn't a priority for me or something I'll push for, I just wanted to record what I found in my first experience setting up the dev env, before it feels normal :)

Also: I notice that just normal git interactions requires nodejs? If telepresence isn't necessary for (my own personal) local development, I might simplify this removing it but adding git client support in the container to support the nodejs hooks. Or I'll just install nodejs :) (or use dockerdev).

### DIFF
--- a/dashboard/.devcontainer/Dockerfile
+++ b/dashboard/.devcontainer/Dockerfile
@@ -1,0 +1,28 @@
+# Creates a node-8 debian image with telepresence installed and relevant tooling.
+FROM bitnami/node:8-debian-9
+
+RUN apt-get update -y \
+  # See Debian instructions at https://www.telepresence.io/reference/install#ubuntu-1604-or-later
+  && curl -sO https://packagecloud.io/install/repositories/datawireio/telepresence/script.deb.sh \
+  && env os=ubuntu dist=xenial bash script.deb.sh \
+  && apt-get install -y --no-install-recommends telepresence \
+  && rm script.deb.sh \
+  # Not sure why iptables and sudo are not pulled in as deps of telepresence, perhaps due to the above
+  # hack of pretending to be ubuntu (and telepresence not expecting to run in a docker container).
+  && apt-get install -y iptables sudo \
+  # kubectl
+  && curl -sSL -o /usr/local/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl \
+  && chmod +x /usr/local/bin/kubectl \
+  # Typescript deps
+  && npm install -g tslint typescript \
+  # Cleanup
+  && apt-get autoremove -y \
+  && apt-get clean -y \
+  && rm -rf /var/lib/apt/lists/*
+
+ENV PORT 3000
+
+COPY rootfs/ /
+ENTRYPOINT ["/entrypoint.sh"]
+
+CMD ["yarn", "start"]
--- a/dashboard/.devcontainer/README.md
+++ b/dashboard/.devcontainer/README.md
@@ -1,0 +1,40 @@
+## Running the dev server in a container
+To run the dev server in a container without telepresence or similar, run
+
+```
+docker-compose up
+```
+and then open a browser at http://localhost:3000/
+
+## Running the test runner
+You can run the frontend tests in a separate shell: 
+
+```
+docker-compose run dev-dashboard yarn test
+```
+
+## Running the dashboard via telepresence in-cluster
+To instead run the local dev server via a telepresence shell, you can:
+```
+docker-compose run dev-dashboard bash
+/telepresence-shell.sh
+```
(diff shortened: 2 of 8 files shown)


andresmgot:
hi @absoludity,

Thank you for writing this down! From the top of my mind there are three major reasons why you need to run the Dashboard in a Kubernetes environment:

 1. You need access to the Kubernetes API in order for the dashboard to work.
 2. You need access to Tiller.
 3. Even if you are able to workaround 1 and 2 to access those resource, you will never be sure if something working for you in docker-compose will work in Kubernetes (unless you are working on small changes).

I personally work with Minikube and telepresence. @prydonius (and myself in the past) used Docker for Mac and that worked as well. Any Kubernetes distribution should actually work (unless Openshift/Minishift). We are using `kind` in the CI for example.

It's true that for developing Kubeapps you need some dev tools:
 - Minikube (or any other k8s distro)
 - Telepresence (for convenience, you can check changes live)
 - Yarn/other NPM dependencies. When using telepresence you will be using your host as a K8s pod so you will need to have the required tools installed. Also, before doing a commit, it will run the linter (and pretifier) so you also need the dev-dependencies for that.

If you still have trouble setting a dev environment I am more than happy to help, we can improve the dev guide if there is something missing there.

absoludity:
Thanks Andres

> From the top of my mind there are three major reasons why you need to run the Dashboard in a Kubernetes environment:
> 
>     1. You need access to the Kubernetes API in order for the dashboard to work.
> 
>     2. You need access to Tiller.

Oh - I probably haven't dug deep enough yet... I'd thought (without yet checking) that the dashboard would be talking to the `chartsvc` and `tiller-proxy` services for that information. I'll dig a bit deeper there tomorrow. (And even that is the case, it may still be easiest to use telepresence with a full deployment - I was just keen to know as a side-note if more light-weight options had been tried or considered).

> 
>     3. Even if you are able to workaround 1 and 2 to access those resource, you will never be sure if something working for you in docker-compose will work in Kubernetes (unless you are working on small changes).

Yep, just note I'm not trying to say that docker-compose is the solution to that problem - I'm using docker-compose in this diff to create the local frontend (dashboard) development environment in a container rather than on my local machine (to automate the setup, and save me from bloating my system - but the latter is a minor point) - whether or not it is used with services from a cluster via telepresence.
 
> I personally work with Minikube and telepresence. @prydonius (and myself in the past) used Docker for Mac and that worked as well. Any Kubernetes distribution should actually work (unless Openshift/Minishift). We are using `kind` in the CI for example.

Nice - I wondered if `kind` could be used to make the dev setup more automated (and light-weight) - but probably not something worth looking at now (just thoughts while trying the existing setup).

> Also, before doing a commit, it will run the linter (and pretifier) so you also need the dev-dependencies for that.

Yeah, that caught me out (as I had nodejs with the dev server in the container, but was using git on my host machine). I can move to use git in the container too, but wonder if they need to be githooks (I can easily run those in the container without nodejs being available on my host).
 
> If you still have trouble setting a dev environment I am more than happy to help, we can improve the dev guide if there is something missing there.

Thanks! I found the dev guide pretty thorough, and it worked fine, I was just keen to try automating the dev setup in a container and avoiding requiring certain nodejs versions or telepresence on the host. Again, nothing important to prioritize (for me or anyone), just documenting my initial thoughts and finding out if anyone else had tried anything similar. Thank again for the input!

andresmgot:
>  I'd thought (without yet checking) that the dashboard would be talking to the chartsvc and tiller-proxy services for that information

That's correct, the Dashboard contacts the `chartsvc`, `tiller-proxy` and the K8s API.

>  I was just keen to try automating the dev setup in a container and avoiding requiring certain nodejs versions or telepresence on the host. 

Yeah, unfortunately I think that's not something avoidable (but I may be wrong). What we try to do is to support the latest (or latest LTS) versions of the dev-dependencies so it should not be a pain to maintain. 

Let me know if you manage to make this work, if it's useful for you it will probably be for some users.

Issue #1009, opened 2019-04-18
504 when installing slow chart


LP0101:
We have some charts that take upward of 2 minutes to deploy (ie: according to the readiness probe). When deploying them via Kubeapps, it will show the loading animation for about 60-90 seconds, then return to the deployment page with this error: `Request failed with status code 504`. However, the chart is still being deployed by helm, and the deployment will eventually be succesful.

I imagine this is caused by a timeout if a chart takes too long to deploy. Is it possible to increase this timeout to prevent this error from showing up?

andresmgot:
Hi @LP0101,

By default, the minimum timeout that we set for contacting Tiller (helm server side) is 180s:

https://github.com/kubeapps/kubeapps/blob/master/pkg/chart/chart.go#L45

Having said that, it's true that the timeout needs to be configured as well at the Ingress level if you are using an Ingress controller. By default we configure it for the Nginx Ingress controller but if you are using a different one you may need to change the annotation we set (we set it to 10 minutes):

https://github.com/kubeapps/kubeapps/blob/master/chart/kubeapps/values.yaml#L31

We also set that in the configuration for the Nginx service that is serving the Dashboard:

https://github.com/kubeapps/kubeapps/blob/master/chart/kubeapps/templates/kubeapps-frontend-config.yaml#L60

How are you accessing the application? Does the web browser console show more info about who is returning the timeout? To check if `tiller-proxy` (the tool that we use to contact tiller) is the one dropping the connection you can check the logs of that pod too.

LP0101:
Hm, the issue seems to have resolved itself by reinstalling kubeapps, unfortunately can't get any logs on what was causing it anymore.

Closing this ticket, I'll reopen it if I experience it once more.

LP0101:
Spoke too soon, apparently.

So, we're using Haproxy as an ingress. The issue doesn't seem to arise from the tiller-proxy pod, as there's no errors there, and the pod does report a successful deployment, despite the frontend showing a 504.

The chart we're deploying uses a post-install container, which takes between 80-120 seconds to execute. This delay is what's causing the problem.

Here's the log from the javascript console:
```
POST https://kubeapps.cxpt.sap.corp/api/tiller-deploy/v1/namespaces/test-whitestar-kubeapps-deploy/releases 504

  | (anonymous) | @ | xhr.js:178
-- | -- | -- | --
  | e.exports | @ | xhr.js:12
  | e.exports | @ | dispatchRequest.js:57
[...]

PR #1100, opened 2019-07-31
State not set to authenticated when validateToken fails with non-401 error

absoludity:
This is just something I noticed while investigating #1009: if `validateToken()` receives an error response other than `401`, we silently ignore it and set the user as `authenticated: true`. This is most easily seen in the dev environment (below) but would be the case if the k8s api responded with a 502 or whatever, I assume?

Here's what you see locally in the dev env without the telepresence setup after entering any token (and `validateToken()` results in a 404):
authenticatd-not-authenticated
Note that the redux state shows `authenticated: true`.

With this PR `validateToken()` always throws an error for an error response when validating the authz to the k8s api, so that (with the separate #1009), the user sees an error message:
authenticated-error
Note the redux state shows `authenticated: false`.

We could probably improve the default error displayed if we use custom `Error` types (though I haven't yet checked the error handling outside of this call-site).

### DIFF
--- a/dashboard/src/shared/Auth.test.ts
+++ b/dashboard/src/shared/Auth.test.ts
@@ -16,16 +16,40 @@ describe("Auth", () => {
   });
 
   describe("when there is an error", () => {
-    it("should throw error for 401 error codes", async () => {
-      const mock = jest.fn(() => {
-        return Promise.reject({ response: { status: 401 } });
+    [
+      {
+        name: "should throw an invalid token error for 401 responses",
+        response: { status: 401, data: "ignored anyway" },
+        expectedError: new Error("invalid token"),
+      },
+      {
+        name: "should throw a standard error for a 404 response",
+        response: { status: 404, data: "Not found" },
+        expectedError: new Error("404: Not found"),
+      },
+      {
+        name: "should throw a standard error for a 500 response",
+        response: { status: 500, data: "Server exception" },
+        expectedError: new Error("500: Server exception"),
+      },
+    ].forEach(testCase => {
+      it(testCase.name, async () => {
+        const mock = jest.fn(() => {
+          return Promise.reject({ response: testCase.response });
+        });
+        Axios.get = mock;
+        // TODO(absoludity): tried using `expect(fn()).rejects.toThrow()` but it seems we need
+        // to upgrade jest for `toThrow()` to work with async.
+        let err;
+        try {
+          await Auth.validateToken("foo");
+        } catch (e) {
+          err = e;
+        } finally {
+          expect(err).not.toBe(undefined);
+          expect(err).toEqual(testCase.expectedError);
+        }
       });
-      Axios.get = mock;
-      try {
-        await Auth.validateToken("foo");
-      } catch (e) {
-        expect(e.message).toEqual("invalid token");
-      }
     });
   });
 
--- a/dashboard/src/shared/Auth.test.ts
+++ b/dashboard/src/shared/Auth.test.ts
@@ -32,6 +32,11 @@ describe("Auth", () => {
         response: { status: 500, data: "Server exception" },
         expectedError: new Error("500: Server exception"),
       },
+      {
+        name: "should succeed for a 403 response",
+        response: { status: 403, data: "Not Allowed" },
+        expectedError: null,
+      },
     ].forEach(testCase => {
       it(testCase.name, async () => {
         const mock = jest.fn(() => {
@@ -40,13 +45,12 @@ describe("Auth", () => {
         Axios.get = mock;
         // TODO(absoludity): tried using `expect(fn()).rejects.toThrow()` but it seems we need
         // to upgrade jest for `toThrow()` to work with async.
-        let err;
+        let err = null;
         try {
           await Auth.validateToken("foo");
         } catch (e) {
           err = e;
         } finally {
-          expect(err).not.toBe(undefined);
           expect(err).toEqual(testCase.expectedError);
         }
       });
--- a/dashboard/src/shared/Auth.ts
+++ b/dashboard/src/shared/Auth.ts
@@ -50,7 +50,14 @@ export class Auth {
       if (res.status === 401) {
         throw new Error("invalid token");
       }
-      throw new Error(`${res.status}: ${res.data}`);
+      // A 403 authorization error only occurs if the token resulted in
+      // successful authentication. We don't make any assumptions over RBAC
+      // for the root "/" nonResourceURL or other required authz permissions
+      // until operations on those resources are attempted (though we may
+      // want to revisit this in the future).
+      if (res.status !== 403) {
+        throw new Error(`${res.status}: ${res.data}`);
+      }
     }
   }
 
--- a/dashboard/src/shared/Auth.ts
+++ b/dashboard/src/shared/Auth.ts
@@ -1,6 +1,7 @@
 import Axios, { AxiosResponse } from "axios";
 const AuthTokenKey = "kubeapps_auth_token";
 const AuthTokenOIDCKey = "kubeapps_auth_token_oidc";
+import { APIBase } from "./Kube";
 
 export class Auth {
   public static getAuthToken() {
@@ -43,12 +44,13 @@ export class Auth {
   // Throws an error if the token is invalid
   public static async validateToken(token: string) {
     try {
-      await Axios.get("api/kube/", { headers: { Authorization: `Bearer ${token}` } });
+      await Axios.get(APIBase + "/", { headers: { Authorization: `Bearer ${token}` } });
     } catch (e) {
       const res = e.response as AxiosResponse;
       if (res.status === 401) {
         throw new Error("invalid token");
       }
+      throw new Error(`${res.status}: ${res.data}`);
     }
   }
 


dashboard/src/shared/Auth.test.ts
```diff
@@ -16,16 +16,40 @@ describe("Auth", () => {
...
+          expect(err).not.toBe(undefined);
+          expect(err).toEqual(testCase.expectedError);
+        }
       });
-      Axios.get = mock;
-      try {
-        await Auth.validateToken("foo");
-      } catch (e) {
-        expect(e.message).toEqual("invalid token");
-      }
```
absoludity: Note that this previous test worked if an exception was thrown, but if non was thrown it also gave a false positive and passed (as the expectation was in the `catch` only). I migrated this existing test into the new test case so that the expectation is (always) handled in the `finally` clause.

prydonius:
> If for some reason that request return a 404 it could mean that it worked but the resource we are requesting is not found. In this case we are requesting the root path of the API so I think it should return a 200 in any environment.

Actually, depending on the RBAC permissions of the user, a request to root API path may return `403 Forbidden` I believe (see the 403 error here https://medium.com/@rschoening/read-only-access-to-kubernetes-cluster-fcf84670b698). The root API path might be hidden from the user, but access to other resources might still be available.

I think the assumption here is, if it's not `401 Unauthorized`, the token is valid. That said, we may want to be more specific and only assume the token is valid if we get a 2XX or 403. I'm not sure what the other cases might be, but we can always adapt it as we go?

absoludity:
Thanks for the input.

> > If for some reason that request return a 404 it could mean that it worked but the resource we are requesting is not found. In this case we are requesting the root path of the API so I think it should return a 200 in any environment.
> 
> Actually, depending on the RBAC permissions of the user, a request to root API path may return `403 Forbidden` I believe (see the 403 error here https://medium.com/@rschoening/read-only-access-to-kubernetes-cluster-fcf84670b698). The root API path might be hidden from the user, but access to other resources might still be available.

That article appears to show a 403 for an *anonymous* api call, I think?

I'll check in more detail (with older api versions), but currently I only see 401s for both anon and invalid token:

```~$ curl -Isk https://$HOST/ -H "Authorization: Bearer $TOKEN" | head -n1
HTTP/2 200 
~$ curl -Isk https://$HOST/ -H "Authorization: Bearer foo" | head -n1
HTTP/2 401 
~$ curl -Isk https://HOST/ | head -n1
HTTP/2 401 
```

prydonius:
Hmm, I'm surprised you're seeing a 401 for `/` for anon, although it looks like you might not be requesting the right host (you have `curl https://HOST` instead of `$HOST`). Can you double check that?

An anonymous API call is treated by the API server as the "system:anonymous" user, and therefore it should be possible to grant/deny that user the ability to view "/" in the same way you would grant/deny any other user that permission.

For non-anon calls with a valid token, you likely have a ClusterRole that's being applied that is granting access to the "/" `nonResourceURL`. Without this explicit permission (though it'll usually be granted in most environments), you can get a 403 response even with a valid token.

absoludity:
> Hmm, I'm surprised you're seeing a 401 for `/` for anon, although it looks like you might not be requesting the right host (you have `curl https://HOST` instead of `$HOST`). Can you double check that?

Sorry, I removed the actual (private) dns, but yes it was correct. Here's the correct paste (after actually exporting HOST:
```
~$ curl -Isk https://$HOST/ -H "Authorization: Bearer $TOKEN" | head -n1
HTTP/2 200 
~$ curl -Isk https://$HOST/ -H "Authorization: Bearer foo" | head -n1
HTTP/2 401 
~$ curl -Isk https://$HOST/ | head -n1
HTTP/2 401 
```
The point I was trying to make is that `validateToken` never does an anonymous request, so that 403 shouldn't be relevant, but you point out below that it's not just anonymous requests you were concerned about...

> An anonymous API call is treated by the API server as the "system:anonymous" user, and therefore it should be possible to grant/deny that user the ability to view "/" in the same way you would grant/deny any other user that permission.

Hmm, odd that I can impersonate "system:anonymous" for that call and get a 403:
```
~$ curl -Isk https://$HOST/ -H "Authorization: Bearer $TOKEN" -H "Impersonate-User: system:anonymous" | head -n1
HTTP/2 403
```
but get a 401 when actually trying the call anonymously. Perhaps RBAC is not explicitly allowing `system:anonymous` (and therefore denying with a 403 when impersonating) [1] but the server I'm testing with is running with `--anonymous-auth=false` so that the lack of a token does not result in authenticating as `system:anonymous`? [2]

[1] Seems to be the case: ```$ kubectl create -f - -o yaml << EOF
> 
> apiVersion: authorization.k8s.io/v1
> kind: SubjectAccessReview
> spec:
>   user: "system:anonymous"
>   groups:
>     - "system:unauthenticated"
>   nonResourceAttributes:
>     path: "/"
>     verb: get
> EOF
apiVersion: authorization.k8s.io/v1
kind: SubjectAccessReview
metadata:
  creationTimestamp: null
spec:
  groups:
  - system:unauthenticated
  nonResourceAttributes:
    path: /
    verb: get
  user: system:anonymous
status:
  allowed: false
  reason: no RBAC policy matched
```
[2] Also true (checked with `kubectl describe po $(kubectl get pods -n kube-system| grep kube-apiserver|awk '{print $1}') -n kube-system | grep "anonymous-auth"`)

> 
> For non-anon calls with a valid token, you likely have a ClusterRole that's being applied that is granting access to the "/" `nonResourceURL`. Without this explicit permission (though it'll usually be granted in most environments), you can get a 403 response even with a valid token.

Right, so with
> That said, we may want to be more specific and only assume the token is valid if we get a 2XX or 403. I'm not sure what the other cases might be, but we can always adapt it as we go?

I understand that since `validateToken` is just ensuring that the token is valid for authentication (not authorization, as we're checking the `/` non resource URL), it should raise the error for a 401 (as it does), but pass for a 403 (since the token is valid, and we don't assume specific RBAC setup for accessing `/`), and raise an error for any other error code. Great, I'll update to that. Thanks again for the background.

Thought for the future: I see that currently kubeapps displays a relevant error if a given permission isn't available at the time in the UI that we attempt the API call. I wonder if it'd be nicer to ensure all the required permissions are available right after the user authenticates (using `SelfSubjectAccessReview` or otherwise)? But not important now.

prydonius:
Thanks for the detailed investigation on this @absoludity!

> The point I was trying to make is that validateToken never does an anonymous request, so that 403 shouldn't be relevant, but you point out below that it's not just anonymous requests you were concerned about...

Yeah you're right. I honestly just found the quickest example I could find, which happened to be an anonymous user.

> but the server I'm testing with is running with --anonymous-auth=false so that the lack of a token does not result in authenticating as system:anonymous?

Very interesting, I didn't know about that flag! Just found [this in the docs](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet-authentication-authorization/#kubelet-authentication) also which confirms the behaviour you're seeing with this flag on.

> Thought for the future: I see that currently kubeapps displays a relevant error if a given permission isn't available at the time in the UI that we attempt the API call. I wonder if it'd be nicer to ensure all the required permissions are available right after the user authenticates (using SelfSubjectAccessReview or otherwise)? But not important now.

Good idea, though in certain cases, having some pages in Kubeapps inaccessible to certain users can be seen as a feature. For example, I only want admins to be able to configure the repositories Kubeapps lists, but I allow the user "absoludity" to deploy in the "absoludity" namespace. In this scenario, "absoludity" should be able to login and access Kubeapps and do things within the "absoludity" namespace, but any action performed outside of the designated namespace should result in an error.

Though I might be misunderstanding you, and what you're suggesting is that we figure out the scope of a users permissions and then use that information to hide actions in the UI based on this knowledge? That could be quite interesting, we should certainly create an issue for that.

absoludity:
> > Thought for the future: I see that currently kubeapps displays a relevant error if a given permission isn't available at the time in the UI that we attempt the API call. I wonder if it'd be nicer to ensure all the required permissions are available right after the user authenticates (using SelfSubjectAccessReview or otherwise)? But not important now.
> 
> Good idea, though in certain cases, having some pages in Kubeapps inaccessible to certain users can be seen as a feature. For example, I only want admins to be able to configure the repositories Kubeapps lists, but I allow the user "absoludity" to deploy in the "absoludity" namespace. In this scenario, "absoludity" should be able to login and access Kubeapps and do things within the "absoludity" namespace, but any action performed outside of the designated namespace should result in an error.

Yes, it doesn't need to be *all* the possible permissions that are checked, just the required permissions to use KubeApps - ah, but from your example, that would be dependent on the namespace the user has access to, which we don't know. Got it. 

> Though I might be misunderstanding you, and what you're suggesting is that we figure out the scope of a users permissions and then use that information to hide actions in the UI based on this knowledge? That could be quite interesting, we should certainly create an issue for that.

I'm not a fan of hiding features like that unless the UX is very carefully thought out (it can easily be confusing if different functionality is visually visible depending on k8s perms which you may not know - much clearer to raise that to the UI, IMO). But figuring out the scope of a users' permissions to be sure they can use KubeApps without issue was all I meant. From your example above, it sounds that all we can assume is that they can access the API and have access to create namespaces or access to create pods in one namespace, so not sure how valuable it'd be on its own. But perhaps something different.

Back to this diff, are you OK with it landing with the changes I added above from your review?

Thanks for the input!

prydonius:
> Yes, it doesn't need to be all the possible permissions that are checked, just the required permissions to use KubeApps - ah, but from your example, that would be dependent on the namespace the user has access to, which we don't know. Got it.

Yeah there isn't really a base set of permissions needed to use Kubeapps, it's all quite variable. I could allow users access to all namespaces, _but_ not allow users to create, for example, _Role_ resources. This now means a user can install any chart that doesn't include a _Role_ in any namespace. Even if we knew what permissions a user had, we wouldn't be able to exhaust the list of all resources in all charts to check if they are able to install everything.

> Back to this diff, are you OK with it landing with the changes I added above from your review?

Yep, thanks for making that change and adding the tests for this too! Will give @andresmgot a chance to re-review and merge if it looks good to him.

andresmgot:
Thank you both for the thoughtful discussion! It was indeed useful. The code now looks good to me, _merging_!

davidkarlsen:
To add to the discussion I found this experience odd:
* charts are listed to user (basically we remove all the standard ones and run with our own internal as only chart repo
* user starts the deployment-process, and first when they hit install the error-message _"You don't have sufficient permissions to create aot-ccm-be-p1 in the aot-ccm-be-p1 namespace
Ask your administrator for the following RBAC roles:

get apprepositories (kubeapps.com) in the kubeapps namespace."_ is shown.

Why is this role needed? After all they could list the charts. If the permission should be used to filter access to charts in repos, this should happen before they hit "deploy", e.g. in this case the chart-"menu" should look empty - either that or they should not need the permission.

Background in:
https://kubernetes.slack.com/archives/C9D3TSUG4/p1564747418022000

PR #1108, opened 2019-08-02
Switch to Go Modules

mkmik:


mkmik:
CC: @andresmgot, @migmartri

Issue #1111, opened 2019-08-06
Improve Kubeapps frontend proxying of k8s API


absoludity:
Certain functionality of the Kubeapps dashboard such as listing namespaces, validating login tokens and checking progress of helm deployments, currently use the Kubernetes API directly. This was done by adding a proxy rule to the frontend nginx config which [proxies requests for `/api/kube` through to `https://kubernetes.default/`](https://github.com/kubeapps/kubeapps/blob/master/chart/kubeapps/templates/kubeapps-frontend-config.yaml#L34), exposing the complete k8s API on the same network used by the frontend (still subject to the configured RBAC rules). 

Though many k8s providers configure public access to the k8s API by default, Kubeapps should document that the k8s API is also exposed through the frontend to avoid possible misunderstandings from users. In the future it may be worth replacing this proxy rule with a small backend API service limited to the Kubeapps context with its focused requirements (checking a deployment progress, obtaining a list of namespaces etc.)

absoludity:
A small backend API service would also allow the frontend client to avoid storing the k8s API token, but rather use an http-only cookie to identify itself with the backend which would include the auth header for the request to the k8s api.

absoludity:
Adding [SSO to Kubeapps, if done together with the proxy impersonation](https://www.youtube.com/watch?v=p5x1btmD2Yw) which @migmartri demonstrated, should be further investigated as a potential fix for this issue also. With that solution, all requests to the K8s API would be routed via the impersonation proxy, with the default RBAC ensuring only accounts with the correct role bindings are able to execute the request.

absoludity:
> A small backend API service would also allow the frontend client to avoid storing the k8s API token, but rather use an http-only cookie to identify itself with the backend which would include the auth header for the request to the k8s api.

The Kubernetes dashboard application solves this quite neatly in a small backend which is encrypting the user-token and returning a JWE (JSON Web Encryption) token as a cookie. This credential is for access to the backend only, and enables the backend to decrypt and access the k8s API with the users token, without requiring storing the token. The only storage required is for the backend's key (which is stored in a k8s secret and is rotatable).

absoludity:
So it turns out that this is also solved by switching to use OpenID Connect for auth (via the auth-proxy) as verified in #1199, as long as we *don't* whitelist the api endpoints.

[...]

PR #1112, opened 2019-08-06
Document Kubeapps frontend proxying to k8s api

absoludity:
See #1111 for details.

/cc @prydonius , @migmartri


andresmgot:
Can we close #1111 with this or you want to track something else?

absoludity:
> Can we close #1111 with this or you want to track something else?

I've updated the title to make it clear that the task is not just documenting. I'd be keen to leave it open as long as the proxying the complete API is in place. Thanks Andres.

PR #1113, opened 2019-08-07
Use Kube.APIBase consistently

absoludity:
The `shared/Kube` module exports the `APIBase` constant, but this was not being used in quite a bit of the application code.

The reason I bothered with this branch is a little convoluted: I want to be able to run the dashboard dev environment in a container, which works fine except I can't yet use `telepresence` within a container (due to fusefs issues).

That said, I don't need telepresence (or my own kubeapps running in k8s) to simply point the dashboard at a kubernetes host during development, I just need the dev server to proxy those requests.

All I should need is to add a [`src/setupProxy.[js|ts]`](https://facebook.github.io/create-react-app/docs/proxying-api-requests-in-development#configuring-the-proxy-manually). This would be perfect as it would not affect those using the telepresence setup (which proxies by the frontend nginx), it would only be used in dev by design, and it would allow me to proxy different `/api/*` requests as needed (supporting other proxied services like `tiller-proxy`). Alas our version of create-react-app (specifically react-scripts-ts) didn't support `setupProxy.js` at the time (you can see this in `./node_modules/react-scripts-ts/config/paths.js` where `setupTests.ts` is checked and exported but `setupProxy.[js/ts]` is not). These days `react-scripts-ts` is not used as the `--typescript` is an option to `create-react-app` itself.

But rather than shave the yak of updating/upgrading (though perhaps there's an easy way @Angelmmiguel knows?), I've found a work-around I can use right now, which is to just temporarily set the [`proxy` option in `package.json`](https://facebook.github.io/create-react-app/docs/proxying-api-requests-in-development) which is supported in `react-scripts-ts/scripts/start`, sending any unrecognised URLs through to the k8s api server I'm using. This method doesn't allow any chopping of prefixes, so I've set `APIBase = ""` temporarily. This works perfectly for the k8s requests (I won't be able to do this when I need two services proxied), is simple for me to do in the mean-time, and only requires this trivial fix of using `Kube.APIBase` consistently which probably should be done anyway :) (EDIT: it's not so useful in the end as dashboard requires both k8s api and tiller-proxy concurrently for most views)

### DIFF
--- a/dashboard/.gitignore
+++ b/dashboard/.gitignore
@@ -18,6 +18,7 @@ src/**/*.css
 .env.test.local
 .env.production.local
 .vscode
+.devcontainer
 .idea
 
 npm-debug.log*
--- a/dashboard/src/shared/AppRepository.ts
+++ b/dashboard/src/shared/AppRepository.ts
@@ -1,4 +1,5 @@
 import { axiosWithAuth } from "./AxiosInstance";
+import { APIBase } from "./Kube";
 import { IAppRepository, IAppRepositoryList } from "./types";
 
 export class AppRepository {
@@ -39,7 +40,7 @@ export class AppRepository {
     return data;
   }
 
-  private static APIBase: string = "api/kube";
+  private static APIBase: string = APIBase;
   private static APIEndpoint: string = `${AppRepository.APIBase}/apis/kubeapps.com/v1alpha1`;
   private static getResourceLink(namespace?: string): string {
     return `${AppRepository.APIEndpoint}/${
--- a/dashboard/src/shared/ClusterServiceClass.ts
+++ b/dashboard/src/shared/ClusterServiceClass.ts
@@ -1,4 +1,5 @@
 import { axiosWithAuth } from "./AxiosInstance";
+import { APIBase } from "./Kube";
 import { ServiceCatalog } from "./ServiceCatalog";
 
 export interface IClusterServiceClass {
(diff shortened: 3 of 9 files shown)


absoludity:
> LGTM @absoludity !

Thanks @Angelmmiguel , though I cc'd you more to pick your brains regarding the lack of `setupProxy` above... specifically whether you know how hard/easy it is to upgrade CRA once in use? (Latest CRA with `--typescript` includes support for `setupProxy` in the webpack dev server config).

Angelmmiguel:
> But rather than shave the yak of updating/upgrading (though perhaps there's an easy way @Angelmmiguel knows?), I've found a work-around I can use right now, which is to just temporarily set the proxy option in package.json which is supported in react-scripts-ts/scripts/start, sending any unrecognised URLs through to the k8s api server I'm using. This method doesn't allow any chopping of prefixes, so I've set APIBase = "" temporarily. This works perfectly for the k8s requests (I won't be able to do this when I need two services proxied), is simple for me to do in the mean-time, and only requires this trivial fix of using Kube.APIBase consistently which probably should be done anyway :) (EDIT: it's not so useful in the end as dashboard requires both k8s api and tiller-proxy concurrently for most views)

As you mentioned, `react-scripts-ts` is deprecated. The main contributor created a guide about how to migrate from `react-scripts-ts` to `react-scripts`: https://vincenttunru.com/migrate-create-react-app-typescript-to-create-react-app/. I have to say that I'm not a fan of this kind of abstractions. They're fine at the beginning, but they add a lot of extra features for common use cases. When you notice the limitations of the library and you `eject` the configuration, you'll get a complex webpack config with a lot of extra features you don't use (on top of the webpack complexity).

However, I still think it's a good refactor to move away from `react-scripts-ts` in favor of `react-scripts`. If I have to choose one of these abstractions I'd select `create-react-app` because is the reference for the React community 😄. I've checked the guide and it seems it's not too complex.

andresmgot:
Feel free to submit the changes you suggest :)

Issue #997, opened 2019-03-20
Support for rollback of releases


alexcurtin:
Can kubeapps do a helm rollback of an application?

andresmgot:
Hi @alexcurtin,

Right now that is not supported but it would be a nice feature. We talked in the past about storing the history of the values.yaml for the different versions (which is related to the rollback command). See this issue for more details: https://github.com/kubeapps/kubeapps/issues/614

andresmgot:
I am going to start working on this cc/ @Angelmmiguel @absoludity 

My plan is to add a new endpoint in Tiller-proxy: `POST /namespaces/{namespace}/releases/{releaseName}?action=rollback`

The body of the request should contain the version to rollback to.

Then we can just use the helmClient to call the supported method: `helmClient.RollbackRelease`

From the UX point of view we can add a new button close to the upgrade button:

![Screenshot from 2019-08-07 11-32-55](https://user-images.githubusercontent.com/4025665/62614483-c60c8980-b90b-11e9-99a4-0cae5e43f6b6.png)

Then, after clicking, we show the confirmation modal:

![Screenshot from 2019-08-07 12-02-12](https://user-images.githubusercontent.com/4025665/62614261-57c7c700-b90b-11e9-8c77-7295f4af85ba.png)

If an application doesn't have a history of releases (e.g. it has never been upgraded), we would show a message when trying to rollback. We can detect this just checking if the current version is 1:

![Screenshot from 2019-08-07 12-05-17](https://user-images.githubusercontent.com/4025665/62614431-a70df780-b90b-11e9-83f3-b0da8a5c08f2.png)

absoludity:
> My plan is to add a new endpoint in Tiller-proxy: `POST /namespaces/{namespace}/releases/{releaseName}/version/{releaseVersion}`

I'm not sure `POST` is the right verb here? We're not creating a new release revision/version, but rather modifying the revision of the existing release right? So `PUT` might be more appropriate?

I was initially wondering if a new endpoint is even required here. Currently `PUT /namespaces/{namespace}/releases/{releaseName}` is used to modify an existing release - currently limited to upgrading the release only. You can already PUT to this endpoint to downgrade (as opposed to performing a rollback) already. Since upgrading updates the release revision (version - are these synonymous) to a new value, it seems a little odd to have a separate endpoint to update the release revision to an old one. I don't feel I yet have the experience here, so see what you think... perhaps @prydonius or @migmartri too.

[...]

PR #1115, opened 2019-08-07
[tiller-proxy] Add endpoint to rollback releases

andresmgot:
As described in #997 

Added a new endpoint:
`POST /namespaces/{namespace}/releases/{releaseName}/version/{releaseVersion}`

The body of the request should contain the details of the chart we are modifying to ensure that the user has the right permissions to modify the resources.


andresmgot:
@absoludity @Angelmmiguel adding you as reviewers but maybe it's not very readable, let me know if you have any question

pkg/proxy/proxy_test.go
```diff
@@ -353,6 +353,26 @@ func TestUpdateMissingHelmRelease(t *testing.T) {
...
+		Version: "1.0.0",
+		Icon:    "icon.png",
+		Name:    "wordpress",
+	}}
+	proxy := newFakeProxy([]AppOverview{app})
+
+	_, err := proxy.RollbackRelease(rs, ns, 1)
+	if err == nil {
+		t.Error("Update should fail, there is not a release in the namespace specified")
+	}
```
absoludity: Is it not possible to test the successful path?

andresmgot: yes, forgot to add that

cmd/tiller-proxy/internal/handler/handler.go
```diff
@@ -199,6 +200,47 @@ func (h *TillerProxy) CreateRelease(w http.ResponseWriter, req *http.Request, pa
...
 
+// RollbackRelease performs an action over a release
+func (h *TillerProxy) RollbackRelease(w http.ResponseWriter, req *http.Request, params Params) {
+	log.Printf("Rolling back %s", params["releaseName"])
+	if !h.DisableAuth {
+		chartDetails, ch, err := getChart(req, h.ChartClient)
+		if err != nil {
+			response.NewErrorResponse(errorCode(err), err.Error()).Write(w)
+			return
+		}
```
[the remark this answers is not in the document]
absoludity: Yeah, I saw, but wasn't sure if the contract is that the chart is always present in the request.

cmd/tiller-proxy/internal/handler/handler.go
```diff
@@ -199,6 +200,47 @@ func (h *TillerProxy) CreateRelease(w http.ResponseWriter, req *http.Request, pa
 	response.NewDataResponse(*rel).Write(w)
 }
 
+// RollbackRelease performs an action over a release
+func (h *TillerProxy) RollbackRelease(w http.ResponseWriter, req *http.Request, params Params) {
+	log.Printf("Rolling back %s", params["releaseName"])
```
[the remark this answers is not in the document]
absoludity: Ah - of course. I'd missed that they are from the route when I was looking at the handlers. Great.

cmd/tiller-proxy/internal/handler/handler_test.go
```diff
@@ -378,6 +378,58 @@ func TestActions(t *testing.T) {
...
+			// Request params
+			RequestBody: `{"chartName": "foo", "releaseName": "foobar",	"version": "1.0.0"}`,
+			RequestQuery: "",
+			Action:       "rollback",
+			Params:       map[string]string{"namespace": "default", "releaseName": "foobar"},
+			// Expected result
+			StatusCode:        422,
+			RemainingReleases: []release.Release{},
+			ResponseBody:      "",
+		},
```
absoludity: Great, thanks for the extra tests.

pkg/proxy/proxy.go
```diff
@@ -266,6 +266,25 @@ func (p *Proxy) UpdateRelease(name, namespace string, values string, ch *chart.C
 	return res.GetRelease(), nil
 }
 
+// RollbackRelease upgrades a tiller release
```
absoludity: Copy-n-paste-o? :P

cmd/tiller-proxy/internal/handler/handler_test.go
```diff
@@ -378,6 +378,58 @@ func TestActions(t *testing.T) {
...
+		{
+			// Scenario params
+			Description:      "Rollsback a missing release",
+			ExistingReleases: []release.Release{},
+			DisableAuth:      true,
+			ForbiddenActions: []auth.Action{},
+			// Request params
+			RequestBody: `{"chartName": "foo", "releaseName": "foobar",	"version": "1.0.0"}`,
+			RequestQuery: "?version=1",
```
absoludity: I see both `version` and `revision` used for the helm release version of an installed chart, though the [helm docs seem to use `revision` consistently (or try to)](https://github.com/helm/helm/blob/master/docs/helm/helm_rollback.md) the `helm` api below uses `RollbackVersion`. Hmph. Should we just use "revision" if that's what it is going forward (and to avoid confusion when chart version is used in the same context, as here in the previous line)?

pkg/proxy/proxy_test.go
absoludity: It'd still be good to add - see how you go.

cmd/tiller-proxy/internal/handler/handler_test.go
andresmgot: that's a good point, I was using `version` because that's how they call it in the API but I guess that's because historical reasons. I am okay if we use `revision` because there are a lot of versions already (`appVersion`, `chartVersion`...)

ehsanhajian72:
how can I contact you by email? @andresmgot

andresmgot:
@ehsanhajian72 I have just set my email to public, you can find me at andresmgot@vmware.com but if you want to ask anything related to this project feel free to open an issue.

Issue #1110, opened 2019-08-03
Improve kubeapps RBAC experience


davidkarlsen:
I find this experience odd:
* charts are listed to user (basically we remove all the standard ones and run with our own internal as only chart repo
* user starts the deployment-process, and first when they hit install the error-message _"You don't have sufficient permissions to create aot-ccm-be-p1 in the aot-ccm-be-p1 namespace
Ask your administrator for the following RBAC roles:
get apprepositories (kubeapps.com) in the kubeapps namespace."_ is shown.

Why is this permission needed? After all they could list the charts. If the permission should be used to filter access to charts in repos, this should happen before they hit "deploy", e.g. in this case the chart-"menu" should look empty - either that or they should not need the permission.

Background in:
https://kubernetes.slack.com/archives/C9D3TSUG4/p1564747418022000
Related to https://github.com/kubeapps/kubeapps/pull/1100

absoludity:
From the above thread (and unrelated to the general aim of improving the general experience:

> IIRC we need to get the repository the user is deploying from to check if there are any auth credentials we need to use to access the chart tarball

So related to the specific issue which David reported, I think that ideally, like David says, it'd be great if this was not required given that it's already not required to be able to list the charts for deploying. The user/admin who adds (and has permission to add) an `AppRepository` custom resource to the Kubeapps installation may not be (by design) the target user deploying the charts.

Checking the code with Adnan's pointer, I see that when creating/installing a new chart repo, if auth was provided Kubeapps records the reference to a k8s secret on the repo CRD and creates a corresponding k8s secret. When a chart is deployed Kubeapps:

1. requests the app repo resource to obtain the auth info (k8s secret reference)
2. sends the chart info to tiller-proxy, including any auth info from 1.
3. tiller proxy parses the request details and [calls out](https://github.com/kubeapps/kubeapps/blob/master/cmd/tiller-proxy/internal/handler/handler.go#L133) to the chart utils pkg's `GetChart()`
4. that helper [requests the content of the k8s secret](https://github.com/kubeapps/kubeapps/blob/master/pkg/chart/chart.go#L341) to include in the headers when getting the chart. 

[...]

PR #1121, opened 2019-08-12
Rollback UI

andresmgot:
Closes: #997 

Add support for rolling back releases with the dashboard:

Screenshot from 2019-08-12 17-51-18

Note that the complexity of this PR comes from the fact that we need to resolve the chart of the release before doing a rollback. This is because we use the chart to determine if the user have permissions to modify the resources detailed on it. This is the same approach we use for upgrades.

For context, in order to resolve a chart we need to know from which repository it comes from. If we are not able to automatically get it, we ask the user to manually introduce the repo with `SelectRepoForm`. This will happen when the registry that contains the chart is not an `AppRepository` or if the call to get the update information fails.

TODO:
 - [x] Write tests


andresmgot:
@Angelmmiguel @absoludity this should be ready to be reviewed!

dashboard/src/actions/apps.ts
```diff
@@ -257,3 +257,25 @@ export function upgradeApp(
...
+    try {
+      const {
+        config: { namespace: kubeappsNamespace },
+      } = getState();
+      await App.rollback(releaseName, namespace, revision, kubeappsNamespace, chartVersion, values);
+      dispatch(getAppWithUpdateInfo(releaseName, namespace));
+      return true;
+    } catch (e) {
+      dispatch(errorApps(e));
+      return false;
```
Angelmmiguel: Are we adding some logic around the `booleans` that we return from thunk actions?

dashboard/src/actions/repos.ts
```diff
@@ -216,10 +216,12 @@ export function checkChart(
     try {
       await axios.get(url.api.charts.listVersions(`${repo}/${chartName}`));
       dispatch(receiveRepo(appRepository));
+      return true;
     } catch (e) {
       dispatch(
         errorChart(new NotFoundError(`Chart ${chartName} not found in the repository ${repo}.`)),
       );
+      return false;
```
Angelmmiguel: Same question

dashboard/src/components/AppView/AppControls/AppControls.tsx
```diff
@@ -44,12 +45,18 @@ class AppControls extends React.Component<IAppControlsProps, IAppControlsState>
...
+          <React.Fragment>
+            <UpgradeButton
+              newVersion={app.updateInfo && !app.updateInfo.upToDate}
+              releaseName={name}
+              releaseNamespace={namespace}
+              push={push}
+            />
+            {/* We only show the rollback button if there are versions to rollback */}
+            {app.version > 1 && (
+              <RollbackButtonContainer releaseName={name} namespace={namespace} />
```
Angelmmiguel: I believe we're not using `React.Suspense` and `lazy` in this project yet, but this is a good component to be loaded lazily in this view.

dashboard/src/components/AppView/AppControls/RollbackButton/RollbackButton.test.tsx
```diff
@@ -0,0 +1,146 @@
...
+  repos: [],
+  repo: {} as IAppRepository,
+  kubeappsNamespace: "kubeapps",
+  fetchRepositories: jest.fn(),
+  checkChart: jest.fn(),
+};
+
+it("stores the chart name and version in the state", () => {
+  const wrapper = shallow(<RollbackButton {...defaultProps} />);
+  wrapper.setProps({
```
Angelmmiguel: Calling `setProps` forces the button to update right after shallowing them. If you are not testing this update behaviour, it's better to avoid updating the component right after shallowing it. You can use the spread operator to modify the `defaultProps`. For example:

```
const appProps = {
  ...defaultProps,
  app: {
    // ...
  }
}

const wrapper = shallow(<RollbackButton {...appProps} />);
```

dashboard/src/components/AppView/AppControls/RollbackButton/RollbackButton.test.tsx
```diff
@@ -0,0 +1,146 @@
...
+      } as IRelease;
+      const wrapper = shallow(
+        <RollbackButton
+          {...defaultProps}
+          chart={undefined}
+          getChartVersion={getChartVersion}
+          app={app}
+        />,
+      );
+      wrapper.setState({ chartName: "bar", chartVersion: "1.0.0" });
```
Angelmmiguel: It's better to test how the component reacts to user interactions and lifecycle methods than forcing states manually. By forcing the state you're skipping the natural component behaviour and it may skip bugs from user interactions or other conditions.

dashboard/src/components/AppView/AppControls/RollbackButton/RollbackButton.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+class RollbackButton extends React.Component<IRollbackButtonProps> {
+  public state: IRollbackButtonState = {
+    modalIsOpen: false,
+    loading: false,
+    chartName: "",
+    chartVersion: "",
+  };
+
+  public componentWillReceiveProps(nextProps: IRollbackButtonProps) {
```
Angelmmiguel: [This method is deprecated](https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops) and it will be removed in future React versions. Since we're in `16.8.6`, we should stop adding more deprecated methods to the codebase.

For this use case, you can use the [getDerivedStateFromProps](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops) lifecycle method.

dashboard/src/components/AppView/AppControls/RollbackButton/RollbackButton.tsx
```diff
@@ -0,0 +1,157 @@
...
+        throw new Error("The current app is missing its chart information");
+      }
+    }
+  }
+
+  public render() {
+    return (
+      <React.Fragment>
+        <Modal
+          style={{
```
Angelmmiguel: Are these styles common across all the modals in the application? If it's the case, I believe it's better to create a common `Modal`. This can be done in a separate PR

dashboard/src/components/AppView/AppControls/AppControls.tsx
andresmgot: That's an interesting feature, didn't know about it. As far as I understand, this is to avoid rendering heavy components that will be used just in certain case. If I use `lazy` with this `RollbackButtonContainer` it will be always rendered isn't it? What would be the difference between using it or not?

dashboard/src/components/AppView/AppControls/RollbackButton/RollbackButton.test.tsx
andresmgot: In this case the `setState` is not needed but yes, I try to reproduce users behavior when possible. In some cases (not this one) I cannot do so because simulations like "onConfirm" don't work as expected.

dashboard/src/components/AppView/AppControls/AppControls.tsx
Angelmmiguel: Actually, it's much powerful than that :). `React.lazy` allows you to download a component lazily when the user access to the view. This approach introduces two improvements in the applications:

* The component is not added to the `main` bundle of the application. As you mentioned, this is super useful for big components that are used only in a single view
* Since you can lazy load secondary components, the application performance increases due to the reduction of components in the view and the size of the bundle

`React.Suspense` is a component that controls lazy components in the UI. Any time you use `React.lazy` to load a component, it should be added as a child of a `React.Suspense` component. Suspense receives a `fallback` component as a prop that will be displayed while the lazy component is loaded. This `fallback` could be a text, a loader component or `null` if you don't need to display anything.

Another good feature about Suspense is that you can have several components in the component hierarchy. The closest one to the lazy component is the one that will render the fallback. It means you don't need to show a loader in the entire view if you're loading a lazy component.

> If I use lazy with this RollbackButtonContainer it will be always rendered isn't it? What would be the difference between using it or not?

About the example you mentioned, `RollbackButtonContainer` won't be rendered at the beginning. If you scope this component inside a `React.Suspense` component, the `fallback` will be returned. In this case, you can even return `null` and don't display anything. 

Once the component is fully loaded, it will be mounted in the UI.

dashboard/src/components/AppView/AppControls/RollbackButton/RollbackDialog.tsx
```diff
@@ -0,0 +1,67 @@
+import * as React from "react";
+import LoadingWrapper from "../../../LoadingWrapper";
+
+import "./RollbackDialog.css";
+
+interface IRollbackDialogProps {
+  loading: boolean;
+  revision: number;
```
absoludity: You have both `props.revision` and `state.revision` which I found confusing. Perhaps the prop could be `currentRevision` and the state `targetRevision`? (would have avoided confusion for me, at least).

dashboard/src/components/AppView/AppControls/AppControls.tsx
andresmgot: I am actually undoing this change because I am not able to test this if I change it for a suspense/lazy block.

andresmgot:
> On a more general note, does the kubeapps community tend to do larger PRs like this? 

Not really, this is mostly my bad. I tend to think the changes are going to be a small PR but sometimes it ends up not being the case (like this PR). I wait until the feature is working before sending the PR because I usually don't know how to split them before hand. Even if I do, I am not sure if the code I am submitting is going to work end-to-end if the feature is not complete (so the same code may appear in several reviews).

I am aware of the pain of reviewing a big PR like this one so I will try to split them as much as possible in the future to avoid the kind of issues your mention.

dashboard/src/components/AppView/AppControls/RollbackButton/RollbackButton.test.tsx
```diff
@@ -0,0 +1,158 @@
...
+  wrapper.update();
+}
+
+context("when opening the Modal", () => {
+  context("when there is no info about the chart", () => {
+    it("should render the SelectRepoForm", () => {
+      const wrapper = shallow(<RollbackButton {...defaultProps} chart={undefined} />);
+      openModal(wrapper);
+
+      expect(wrapper.find(SelectRepoForm)).toExist();
```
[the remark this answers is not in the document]
absoludity: > > In what situations will we not have info about the chart? You mention in the description that this can happen when the repository is not an AppRepository, but it's not clear whether this means that Kubeapps does not have this info (I mean, helm must have it, right?), or the dashboard just doesn't have it (or can't currently get it), or it's one of stable etc.?
> 
> That could happen in several occasions:
> 
>     * The user installs a chart using the `helm` CLI with a repository that doesn't exists in Kubeapps. Then we are not able to see where this application comes from.

In this situation, helm/tiller then knows the repo though doesn't it? EDIT: I think not - this is the source of my confusion. Helm just finds a chart with whatever repos it has access to, and doesn't (apparently) store the repo of the chart it finds for later reference?

> 
>     * The chart doesn't have an `appVersion`. This field is optional but we use it to verify if a chart belongs to a repository. What we do is checking if `chart: wordpress, version: 1.0.0, appVersion: 1.0.1` exists in the repo `stable`. If those three parameters match, we assume that chart comes from that repository. If we don't have that information we are not confident enough to make that assumption.

And this is all info which isn't necessary to roll back, but just to verify the current user is allowed to roll back? Seems odd (but I may not have my head around it). Why can't we assume that if the user has "write" access to the namespace in which the chart is deployed, then they can roll back and just let them do it? I think on upgrade it's a bit different because it requires access to the repo, but does a rollback require this at all? (not sure if the chart versions are stored by tiller, haven't checked).

[snip]

> 
> > Isn't the repo required just to get the auth info for the request? If so, what's the implication of allowing me, the user, to select which auth to use? (security thought: I could setup my own repo which captures any auth credentials, then rollback/upgrade a release, but use the dashboard to specify another repo with secrets I want to know... won't the request then be done by tiller with the secrets from the other repo, so I can capture them?)
> 
> The repo is required to retrieve the information about the chart. This is ultimately used to verify if the user `foo` can modify the deployment `bar` that is specified in the chart. This is not perfect though since we don't have a way to verify that the chart really belongs to that repo, we just verify that there is a chart with the same version/appversion is there.

As above, why isn't it enough that the user has sufficient access to the namespace, to enable roll back?

> 
> The user doesn't interact with any other auth info so I am not sure if that's the case you are describing.
> 
> > Also, this is related to #1110 I think? (in that, we're only needing to read repositories because we need the auth... if instead we sent the repository name and tiller used that to get the auth, we'd only ever need the repo name and would not need to fetch the repo/repos). We just need to always have a repo name - interested to know why we don't know the repo name.
> 
> We would still need to list the repos if we are not able to auto-retrieve the registry that contains the chart.

Only if we required the repo for authz though right? (Unlike an upgrade, a rollback doesn't require access to the repo per-se, since it's rolling back to a previous version of the chart which it has stored?)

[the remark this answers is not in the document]
andresmgot: > Helm just finds a chart with whatever repos it has access to, and doesn't (apparently) store the repo of the chart it finds for later reference?

Right, that's the problem. The helm CLI downloads the chart locally and just send the manifests. Tiller (or the configmaps stored) has no clue of where those charts came from.

> Why can't we assume that if the user has "write" access to the namespace in which the chart is deployed, then they can roll back and just let them do it?

There is not a simple check to verify if a user has "write" access. Note that a chart can contain any resource, they can even escape from a namespace. If a chart has a ClusterRole or a CRD we need to verify that the user has permissions to modify those cluster-wide resources. Even if it's in a single namespace, a user can have permissions to modify configmaps but not secrets. We need to verify all the resources to avoid an escalation of privileges.

>  I think on upgrade it's a bit different because it requires access to the repo, but does a rollback require this at all? (not sure if the chart versions are stored by tiller, haven't checked).

The rollback operation doesn't need to access the registry but we need to do so to be able to get the resources the chart modifies. Ideally we could get this information from Tiller since those resources are specified in the configmap but for simplicity we are not doing that right now. That could be an improvement of this approach.

Issue #774, opened 2018-10-29
Remove setting of GOPATH in developer documentation


prydonius:
It's better to assume the user has set their desired GOPATH rather than documenting that developers set it to `~/gopath` (especially since this differs from the default). We can point users to https://golang.org/doc/code.html#GOPATH to learn more about the GOPATH.

Instead of instructing users to `git clone` into `$GOPATH/src/...` we can use the `go get -d` command which will clone the repositories without trying to install any packages.

PR #1122, opened 2019-08-14
Parsing of tiller request details accepts an AppRepositoryResourceName

absoludity:
Initial diff related to #1110 enabling Tiller to parse requests containing the name of an app repository custom resource. See that issue for more details.

### DIFF
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -61,7 +61,7 @@ func init() {
 type Details struct {
 	// RepoURL is the URL of the repository. Defaults to stable repo.
 	RepoURL string `json:"repoUrl,omitempty"`
-	// AppRepositoryResourceURL specifies an app repository resource to use
+	// AppRepositoryResourceName specifies an app repository resource to use
 	// for the request.
 	// TODO(absoludity): Intended to supercede RepoURL and Auth below. Remove
 	// RepoURL and Auth once #1110 complete.
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -61,6 +61,11 @@ func init() {
 type Details struct {
 	// RepoURL is the URL of the repository. Defaults to stable repo.
 	RepoURL string `json:"repoUrl,omitempty"`
+	// AppRepositoryResourceURL specifies an app repository resource to use
+	// for the request.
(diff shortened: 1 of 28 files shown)


absoludity:
Seems to be failing CI because we're doing the checkout in GOPATH and so go modules are not being used. I see there's a ticket to update the README, but who has access to the circle ci setup? @andresmgot or @prydonius ?

pkg/chart/chart.go
```diff
@@ -61,6 +61,11 @@ func init() {
 type Details struct {
 	// RepoURL is the URL of the repository. Defaults to stable repo.
 	RepoURL string `json:"repoUrl,omitempty"`
+	// AppRepositoryResourceURL specifies an app repository resource to use
+	// for the request.
```
andresmgot: I think you changed the property but not this comment.

andresmgot:
> Seems to be failing CI because we're doing the checkout in GOPATH and so go modules are not being used. I see there's a ticket to update the README, but who has access to the circle ci setup? 

We are using the vendored packages on purpose (at least for the moment). You can update it running `go mod vendor`.

I am not sure why you mention the circle ci setup, what do you need from there? what's the ticket you mention? Also, in theory you should be able to access/configure CircleCI as a member, you just need to sign-up and follow the project. Let me know if that is not the case.

absoludity:
> > Seems to be failing CI because we're doing the checkout in GOPATH and so go modules are not being used. I see there's a ticket to update the README, but who has access to the circle ci setup?
> 
> We are using the vendored packages on purpose (at least for the moment). You can update it running `go mod vendor`.
> 
> I am not sure why you mention the circle ci setup, what do you need from there?

I just saw that one of the ci/circleci tests had failed as I was leaving, quickly checked and saw it was due to the new mod (and forgot that we still vendor). Thanks, updated.

> what's the ticket you mention?

It was actually #774 that I meant about updating the README to not specify GOPATH, but checking it is actually unrelated to go mod. If it's not already, I'll update it tomorrow to reflect what it should be now (and fix it).

> Also, in theory you should be able to access/configure CircleCI as a member, you just need to sign-up and follow the project. Let me know if that is not the case.

Should be - I'll check tomorrow. Thanks Andres.

PR #1124, opened 2019-08-19
Add tests for InitNetClient. Refactor to pass chart details

absoludity:
This is the next small installment for #1110 .

I'm planning to have `InitNetClient` fetching the `AppRepository` custom resource, when present in the request, but wasn't confident to change it before adding tests. So this PR just tests the existing functionality, with a tiny refactor to pass the `ChartDetails` into the call (as this will include the auth object for the current `Details.Auth.CustomCA` secret ref, as well as the `Details.AppRepository.ResourceName` which I'll pull from k8s and parse in a later PR).

I have one other small prep branch next, before I'll add the fetch of the custom resource in `InitNetClient`.

### DIFF
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -110,7 +110,7 @@ type LoadChart func(in io.Reader) (*chart.Chart, error)
 type Resolver interface {
 	ParseDetails(data []byte) (*Details, error)
 	GetChart(details *Details, netClient HTTPClient) (*chart.Chart, error)
-	InitNetClient(customCA *CustomCA) (HTTPClient, error)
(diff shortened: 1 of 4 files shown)


pkg/chart/chart.go
```diff
@@ -311,7 +313,11 @@ func (c *Chart) InitNetClient(customCA *CustomCA) (HTTPClient, error) {
 		}
```
andresmgot: not related to your changes but re-reading this I think we should return the error (as you added) instead of logging a fatal error.

absoludity: Woah, most definitely. I'll do that and the one I actually meant (up on 301 above), in the next PR.

PR #1125, opened 2019-08-19
Handle error when specified CA cert secret is not present.

absoludity:
As mentioned on #1124, we should not be causing tiller-proxy to exit if it can't fetch a user-supplied secret.

There was another question I had, which I'll add inline.

### DIFF
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -309,7 +309,7 @@ func (c *Chart) InitNetClient(details *Details) (HTTPClient, error) {
 		namespace := os.Getenv("POD_NAMESPACE")
 		caCertSecret, err := c.kubeClient.CoreV1().Secrets(namespace).Get(customCA.SecretKeyRef.Name, metav1.GetOptions{})
 		if err != nil {
-			log.Fatalf("Unable to read the given CA cert: %v", err)
+			return nil, fmt.Errorf("unable to read secret %q: %v", customCA.SecretKeyRef.Name, err)
 		}
 
 		// Append our cert to the system pool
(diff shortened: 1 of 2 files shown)


pkg/chart/chart.go
```diff
@@ -309,7 +309,7 @@ func (c *Chart) InitNetClient(details *Details) (HTTPClient, error) {
 		namespace := os.Getenv("POD_NAMESPACE")
```
absoludity: @andresmgot So GH doesn't allow commenting on the lines not in the context, it seems. Anyway, up above on line 301, do you remember in what scenarious the call to `SystemCertPool()` would fail (that we are ignoring and just creating an empty pool)? I'm keen to add a handler for that error, but don't know if it's necessary for some situation?

andresmgot: If I remember correctly is kind of a corner case. In case there is no system certs, a user may still want to go ahead and use just their CA. `SystemCertPool` returns an error if it's not able to find any cert.

absoludity:
> If I remember correctly is kind of a corner case. In case there is no system certs, a user may still want to go ahead and use just their CA. SystemCertPool returns an error if it's not able to find any cert.

Yeah, may or may not. I think the key for me is that we silently ignore it currently. If their repository was on a server which is normally covered by a standard cert, we'd be exposing them to a mitm without them knowing, I think? (that is, we'd do the request without certs).

Do you remember where the corner case was found? (who's setup, what was different about it, was it CI, ...). If it's CI, we could easily only allow this when that env var is set.

andresmgot:
The corner case was probably me trying run this without installing the system package for the certs. In any case we are building the docker image and we include the system certs so that shouldn't be a problem.

PR #1129, opened 2019-08-20
Move fetching of Auth secret to client initialization (matching customCA)

absoludity:
Follow on from #1128 , this moves the fetching of the Auth secret to the client initialization function, to match what we already do for the customCA and also allow me to do the same (in the next PR) when fetching the `AppRepository` to get the auth details.

Ref #1110

### DIFF
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -129,7 +129,7 @@ func NewChart(kubeClient kubernetes.Interface, load LoadChart, userAgent string)
 	}
 }
 
-func getReq(rawURL, authHeader string) (*http.Request, error) {
+func getReq(rawURL string) (*http.Request, error) {
 	parsedURL, err := url.ParseRequestURI(rawURL)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


pkg/chart/chart.go
```diff
@@ -328,6 +325,22 @@ func (c *Chart) InitNetClient(details *Details) (HTTPClient, error) {
...
 	}
 
+	defaultHeaders := http.Header{"User-Agent": []string{c.userAgent}}
+	if details.Auth.Header != nil {
+		namespace := os.Getenv("POD_NAMESPACE")
+		// TODO: Check with Andres: why are we relying on the defaultNamespace here for
+		// Auth, but not above for the customCA? Should both or neither?
+		// if namespace == "" {
+		// 	namespace = defaultNamespace
+		// }
```
absoludity: @andresmgot ^^

andresmgot: yes, we should do it in both (this bug has never bitten us because we always set that env var in the chart

absoludity: Done - since both are in the one function, I moved them out so the namespace is set once and used for both.

PR #1130, opened 2019-08-20
Refactor Modals style

### LINKED ISSUE #1123: Refactor Modals style
There are several modals using the same style:

```
        <Modal
          style={{
            content: {
              bottom: "auto",
              left: "50%",
              marginRight: "-50%",
              right: "auto",
              top: "50%",
              transform: "translate(-50%, -50%)",
            },
          }}
```

We should refactor that into a common component and use that instead.

andresmgot:
Fixes #1123

Avoid to use inline style every time we use a Modal (in the case we want a small, centered Modal).

I have moved the resulting CSS to the index file so the class `centered-modal` can be used anywhere.

I have also slightly modified the `RollbackButton` modal, see inline.

### DIFF
--- a/dashboard/src/components/AppView/AppControls/RollbackButton/RollbackDialog.tsx
+++ b/dashboard/src/components/AppView/AppControls/RollbackButton/RollbackDialog.tsx
@@ -37,13 +37,15 @@ class RollbackDialog extends React.Component<IRollbackDialogProps, IRollbackDial
         <label>
           Select the revision to which you want to rollback (current: {this.props.currentRevision})
         </label>
-        <select className="margin-t-normal" onChange={this.selectRevision}>
-          {options.map(o => (
-            <option key={o} value={o}>
-              {o}
-            </option>
-          ))}
-        </select>
+        <div>
+          <select className="margin-t-normal" onChange={this.selectRevision}>
+            {options.map(o => (
+              <option key={o} value={o}>
+                {o}
+              </option>
+            ))}
+          </select>
+        </div>
         <div className="margin-t-normal button-row">
           <button className="button" type="button" onClick={this.props.closeModal}>
             Cancel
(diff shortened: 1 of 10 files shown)


dashboard/src/components/AppView/AppControls/RollbackButton/RollbackDialog.tsx
```diff
@@ -37,13 +37,15 @@ class RollbackDialog extends React.Component<IRollbackDialogProps, IRollbackDial
...
           Select the revision to which you want to rollback (current: {this.props.currentRevision})
         </label>
-        <select className="margin-t-normal" onChange={this.selectRevision}>
-          {options.map(o => (
-            <option key={o} value={o}>
-              {o}
-            </option>
-          ))}
-        </select>
+        <div>
```
andresmgot: I have moved the `select` block (inline) to its own div because `label` and `select` are inline components. That caused the modal to have the width of `select+label` but they were rendered in different lines, making the modal unnecessarily big.

absoludity: I hate fiddling with CSS like that. Angel will probably comment when back with some sage advice, I'm sure :)

dashboard/src/components/ServiceClassView/ProvisionButton.tsx
```diff
@@ -89,7 +78,7 @@ class ProvisionButton extends React.Component<IProvisionButtonProps, IProvisionB
           isOpen={this.state.modalIsOpen}
           onRequestClose={this.closeModal}
           contentLabel="Modal"
-          style={this.state.displayNameForm ? smallModalStyle : {}}
+          className={this.state.displayNameForm ? "centered-modal" : ""}
```
absoludity: Was the smallModalStyle different from centered? Or fine as is?

andresmgot: it was the same but following the inline style syntax

PR #1131, opened 2019-08-20
Add redux actions for app operations

### LINKED ISSUE #1126: [refactor] Inconsistent redux store in actions/apps.tsx
We need to revisit the current state of the redux store for operation related to applications. More context [here](https://github.com/kubeapps/kubeapps/pull/1121/files#r314130741). 

We are triggering events twice (like REQUEST_APPS) and some none (like upgrading and rolling back).

andresmgot:
Fixes #1126

Dispatch actions every time we perform an operation over an application. Avoid duplicated dispatch.


andresmgot:
I am okay with both but it's true that the naming `request/receive` is also used in other files so for consistency I would leave it as it is. Thanks for the review!

PR #1132, opened 2019-08-21
Don't ignore system cert error

absoludity:
Following up on a comment https://github.com/kubeapps/kubeapps/pull/1125/files#r315253073 this ensures that we don't swallow an error unnecessarily when getting the system certs.

So that I don't break anyone's test environment, an empty pool of certs can be explicitly allowed if necessary.

### DIFF
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -297,10 +297,14 @@ func (c *clientWithDefaultUserAgent) Do(req *http.Request) (*http.Response, erro
 // InitNetClient returns an HTTP client based on the chart details loading a
 // custom CA if provided (as a secret)
 func (c *Chart) InitNetClient(details *Details) (HTTPClient, error) {
-	// Get the SystemCertPool, continue with an empty pool on error
-	caCertPool, _ := x509.SystemCertPool()
-	if caCertPool == nil {
-		caCertPool = x509.NewCertPool()
+	caCertPool := x509.NewCertPool()
+	// Get the SystemCertPool unless the env var is explicitly set.
+	if _, ok := os.LookupEnv("TILLER_PROXY_ALLOW_EMPTY_CERT_POOL"); !ok {
+		var err error
+		caCertPool, err = x509.SystemCertPool()
+		if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/chart/chart.go
```diff
@@ -297,10 +297,14 @@ func (c *clientWithDefaultUserAgent) Do(req *http.Request) (*http.Response, erro
 // InitNetClient returns an HTTP client based on the chart details loading a
 // custom CA if provided (as a secret)
 func (c *Chart) InitNetClient(details *Details) (HTTPClient, error) {
-	// Get the SystemCertPool, continue with an empty pool on error
-	caCertPool, _ := x509.SystemCertPool()
-	if caCertPool == nil {
-		caCertPool = x509.NewCertPool()
+	caCertPool := x509.NewCertPool()
+	// Get the SystemCertPool unless the env var is explicitly set.
+	if _, ok := os.LookupEnv("TILLER_PROXY_ALLOW_EMPTY_CERT_POOL"); !ok {
```
andresmgot: I would do it a bit different, mostly because `TILLER_PROXY_ALLOW_EMPTY_CERT_POOL` suggest for me that it's going to at least try to load the cert pool:

```
	caCertPool, err := x509.SystemCertPool()
	if err != nil {
          if _, ok := os.LookupEnv("TILLER_PROXY_ALLOW_EMPTY_CERT_POOL"); !ok {
            return nil, err
          }
          caCertPool = x509.NewCertPool()
        }
```

absoludity: Yeah, I had it that way first, but then the optimizing demon in my head wondered whether `SystemCertPool()` is more expensive (ie. whether `NewCertPool()` is [just a memory allocation](https://golang.org/src/crypto/x509/cert_pool.go). But you're right in that it's premature... the code is clearer when it matches the env var. Switched.

absoludity: This required removing the tests I'd added too - as I can't force or mock `x509.SystemCertPool()` easily (`runtime.GOOS` is a const). Not sure it matters.

PR #1134, opened 2019-08-21
Remove code related to resolve chart when rolling back

andresmgot:
Frontend changes related to #1133

Since we don't require to resolve the chart here, there is a lot of code we can delete.


andresmgot:
> but that made me realise we don't have a CI that will check whether a rollback works etc. or do we?

We don't, we still don't have end-to-end tests with a browser.

PR #1136, opened 2019-08-22
Revert "Bump chart version"

andresmgot:
Reverts kubeapps/kubeapps#1135

I have to revert this until the tags are available :/

### DIFF
--- a/chart/kubeapps/Chart.yaml
+++ b/chart/kubeapps/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kubeapps
-version: 2.1.0
+version: 2.0.3
 appVersion: DEVEL
 description: Kubeapps is a dashboard for your Kubernetes cluster that makes it easy to deploy and manage applications in your cluster using Helm
 icon: https://raw.githubusercontent.com/kubeapps/kubeapps/master/docs/img/logo.png


absoludity:
A couple of questions for my own understanding regarding this:

Do you know why the Kubeapps chart does not use a specific tag for the default version of the Kubeapps services images? It's using [`latest` for dashboard](https://github.com/kubeapps/kubeapps/blob/master/chart/kubeapps/values.yaml#L224), apprepository-controller and tiller-proxy, which means, I assume, that `helm install kubeapps` by default will give you the latest images, regardless of the chart version (and whether the chart supports the related options). I took a brief look at a bunch of other charts and they all seem to specify a specific tag for the default ([airflow](https://github.com/helm/charts/blob/master/stable/airflow/values.yaml#L66), [chartmuseum](https://github.com/helm/charts/blob/master/stable/chartmuseum/values.yaml#L8), [datadog](https://github.com/helm/charts/blob/master/stable/datadog/values.yaml#L21), [mariadb](https://github.com/helm/charts/blob/master/stable/mariadb/values.yaml#L22)).

To future-proof the chart, I think it should be updated as part of the release, no?

> Reverts #1135
> 
> I have to revert this until the tags are available :/

Given the above use of `latest` (which I think we should change), why did you have to revert if the PR doesn't actually reference the tags (just out of interest)?

andresmgot:
Sorry, the release process is not documented. This is how it works:

 - We push a tag to the repository. The bitnami pipeline eventually builds the image `bitnami/kubeapps:<tag>`.
 - Once the image is available we bump the chart version.
 - In CircleCI, we check if the chart version of the Chart.yaml of this repo matches the version in the [bitnami/charts repository](https://github.com/bitnami/charts/blob/master/bitnami/kubeapps/Chart.yaml). In the case that the version is different, we trigger the `sync` process.
 - The CI [modifies the default images in the values.yaml](https://github.com/kubeapps/kubeapps/blob/master/script/chart_sync_utils.sh#L45), using the bitnami registry and the latest tag. For example `kubeapps/dashboard:latest` becomes `bitnami/dashboard:v1.5.0`.
 - Finally it pushes the changes in the bitnami/charts repository (https://github.com/bitnami/charts/commit/34c9650ed655a7daca8eff43e94936266bed0224)

The status of the chart here is for development. That's why it's using `latest` as the image tags and `DEVEL` as `appVersion`.

absoludity:
> Sorry, the release process is not documented. This is how it works:
> 
>     * We push a tag to the repository. The bitnami pipeline eventually builds the image `bitnami/kubeapps:<tag>`.
> 
>     * Once the image is available we bump the chart version.
> 
>     * In CircleCI, we check if the chart version of the Chart.yaml of this repo matches the version in the [bitnami/charts repository](https://github.com/bitnami/charts/blob/master/bitnami/kubeapps/Chart.yaml). In the case that the version is different, we trigger the `sync` process.
> 
>     * The CI [modifies the default images in the values.yaml](https://github.com/kubeapps/kubeapps/blob/master/script/chart_sync_utils.sh#L45), using the bitnami registry and the latest tag. For example `kubeapps/dashboard:latest` becomes `bitnami/kubeapps-dashboard:v1.5.0`.

Great - this is the bit I was missing. Thanks.

PR #1137, opened 2019-08-23
Fetch AppRepo to obtain auth data (so client doesn't need to send it).

absoludity:
Adds a clientset for the app revision API and uses that to get the AppRepository custom resource, if a name was specified in the request. This will enable the client (Kubeapps dashboard) to send the app repository name, rather than having to fetch the app repo (which can fail if certain RBAC is not setup).

Ref: #1110

### DIFF
--- a/cmd/tiller-proxy/main.go
+++ b/cmd/tiller-proxy/main.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/gorilla/mux"
 	"github.com/heptiolabs/healthcheck"
+	appRepo "github.com/kubeapps/kubeapps/cmd/apprepository-controller/pkg/client/clientset/versioned"
 	"github.com/kubeapps/kubeapps/cmd/tiller-proxy/internal/handler"
 	chartUtils "github.com/kubeapps/kubeapps/pkg/chart"
 	tillerProxy "github.com/kubeapps/kubeapps/pkg/proxy"
@@ -80,12 +81,17 @@ func main() {
 
 	config, err := rest.InClusterConfig()
 	if err != nil {
-		log.Fatalf("Unable to get cluter config: %v", err)
+		log.Fatalf("Unable to get cluster config: %v", err)
 	}
 
 	kubeClient, err = kubernetes.NewForConfig(config)
 	if err != nil {
 		log.Fatalf("Unable to create a kubernetes client: %v", err)
+	}
+
+	appRepoClient, err := appRepo.NewForConfig(config)
+	if err != nil {
+		log.Fatalf("Unable to create an app repository client: %v", err)
 	}
 
 	log.Printf("Using tiller host: %s", settings.TillerHost)
--- a/cmd/tiller-proxy/main.go
+++ b/cmd/tiller-proxy/main.go
@@ -125,7 +125,7 @@ func main() {
 	}
 
 	proxy = tillerProxy.NewProxy(kubeClient, helmClient)
-	chartutils := chartUtils.NewChart(kubeClient, helmChartUtil.LoadArchive, userAgent())
+	chartutils := chartUtils.NewChart(kubeClient, appRepoClient, helmChartUtil.LoadArchive, userAgent())
 
 	r := mux.NewRouter()
 
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -326,7 +326,6 @@ func (c *Chart) InitNetClient(details *Details) (HTTPClient, error) {
 	// key refs.
 	var customCASecretRef *corev1.SecretKeySelector
 	var authHeaderSecretRef *corev1.SecretKeySelector
-	defaultHeaders := http.Header{"User-Agent": []string{c.userAgent}}
 	if details.AppRepositoryResourceName != "" {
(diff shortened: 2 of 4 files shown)


pkg/chart/chart.go
```diff
@@ -309,35 +312,64 @@ func (c *Chart) InitNetClient(details *Details) (HTTPClient, error) {
...
+		}
+		if appRepo.Spec.Auth.CustomCA != nil {
+			customCASecretRef = &appRepo.Spec.Auth.CustomCA.SecretKeyRef
+		}
+		if appRepo.Spec.Auth.Header != nil {
+			authHeaderSecretRef = &appRepo.Spec.Auth.Header.SecretKeyRef
+		}
+
+	} else {
+		if details.Auth.CustomCA != nil {
```
andresmgot: I would leave a comment saying that we are leaving this for backward compatibility and it should be removed.

pkg/chart/chart.go
```diff
@@ -309,35 +312,64 @@ func (c *Chart) InitNetClient(details *Details) (HTTPClient, error) {
...
 		namespace = defaultNamespace
 	}
-	if customCA != nil {
-		caCertSecret, err := c.kubeClient.CoreV1().Secrets(namespace).Get(customCA.SecretKeyRef.Name, metav1.GetOptions{})
+
+	// If the app repo url is specified then we'll grab the customCA and
+	// defaultHeaders from there, otherwise default to the current details.Auth
+	// (until that path is removed).
+	// TODO(absoludity): Check whether there is a reason that Details.Auth is a different type
+	// to appRepo.Spec.Auth (though they have identical fields). The following would be simpler
```
andresmgot: you are right, those are duplicated. I am fine deleting the struct definition `Details.Auth`

pkg/chart/chart.go
```diff
@@ -309,35 +312,64 @@ func (c *Chart) InitNetClient(details *Details) (HTTPClient, error) {
...
-			return nil, fmt.Errorf("secret %q did not contain key %q", customCA.SecretKeyRef.Name, customCA.SecretKeyRef.Key)
+			return nil, fmt.Errorf("secret %q did not contain key %q", customCASecretRef.Name, customCASecretRef.Key)
 		}
 		if ok := caCertPool.AppendCertsFromPEM(customData); !ok {
-			return nil, fmt.Errorf("Failed to append %s to RootCAs", customCA.SecretKeyRef.Name)
+			return nil, fmt.Errorf("Failed to append %s to RootCAs", customCASecretRef.Name)
 		}
 	}
 
-	defaultHeaders := http.Header{"User-Agent": []string{c.userAgent}}
```
andresmgot: where is the `User-Agent` being set now?

pkg/chart/chart.go
absoludity: Great, I'll do that separately and get it landed, so I can clean up this PR.

pkg/chart/chart.go
absoludity: There is a comment to that effect at the start of the block above? ("otherwise default to the current details.Auth (until that path is removed)") Let me know if it's not clear enough.

pkg/chart/chart.go
absoludity: `defaultHeaders` is initialized with the `User-Agent` above.

absoludity:
Please don't land this, as I want to update after #1143 lands, as well as test the chart rbac before the next (and final?) PR updating the frontend.

cmd/tiller-proxy/main.go
```diff
@@ -80,14 +81,19 @@ func main() {
...
 	}
 
 	kubeClient, err = kubernetes.NewForConfig(config)
 	if err != nil {
 		log.Fatalf("Unable to create a kubernetes client: %v", err)
 	}
 
+	appRepoClient, err := appRepo.NewForConfig(config)
+	if err != nil {
+		log.Fatalf("Unable to create an app repository client: %v", err)
```
Angelmmiguel: Is the `NewForConfig` method trying to create a new `AppRepository` CRD? From the `tiller-proxy-rbac.yml`, it seems the proxy can only `get` the resources.

absoludity: Nope, it's a standard method which all CRDs and the core k8s api provide, to create a client for interacting with the server. The autogenerated functions enabled by the client are in `pkg/client/clientset/versioned/typed/apprepository/v1alpha1/apprepository.go` (including `Create`) :)

Angelmmiguel: Thanks for the explanation!

pkg/chart/chart.go
absoludity: Done, much cleaner :)

pkg/chart/chart.go
absoludity: Not sure if I had some in-transit code which required it earlier, but you're right, no longer there. Moved back to context. Thanks.

andresmgot:
Anything else missing in this PR @absoludity ?

absoludity:
> Anything else missing in this PR @absoludity ?

I want to QA this properly when I do the frontend change next, that's the only reason I'm holding off landing it. Is that OK? (I'm planning to do this soon, just been caught up with other non-Kubeapps work).

andresmgot:
Sure, no problem. I was just checking I was not blocking this.

PR #1147, opened 2019-09-03
Client sends chart repo resource name rather than fetching auth data.

absoludity:
Updates the frontend to send the repository resource name (and relying on the recent tiller-proxy changes to fetch the custom repo resource to get the auth), rather than fetching the resource from the client.

This removes the need for users to have read access to the app repositories CRD to simply deploy a chart, as per #1110 .

Update the docs to say the same, after QAing locally.

I temporarily removed a restriction I'd added in tiller-proxy so that the frontend can still send the repoURL as part of the request. With this PR, the task is finished, but I've got a final cleanup to followup with that will:

  * updating tiller-proxy to store the repoUrl from the fetched app repository,
  * removing the repoURL option from the `Details` struct and from the frontend request.

### DIFF
--- a/dashboard/src/shared/App.test.ts
+++ b/dashboard/src/shared/App.test.ts
@@ -1,7 +1,7 @@
 import * as moxios from "moxios";
 import { App, TILLER_PROXY_ROOT_URL } from "./App";
 import { axiosWithAuth } from "./AxiosInstance";
-import { IAppOverview } from "./types";
+import { IAppOverview, IChartVersion } from "./types";
 
 describe("App", () => {
   beforeEach(() => {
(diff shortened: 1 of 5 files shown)


absoludity:
> This code looks good to me! But there are a couple of things more to do:
> 
>     * In the `DeploymentForm`, we are setting some `requiredRBACRoles`. This is used in case we receive a ForbiddenError when rendering the form and it's specifying the need to have read access to the apprepositories so we need to remove that.
> 
>     * We would also need to follow the same approach for upgrading. We have the same unnecessary requirement to request the AppRepository just to get the `auth` info.

Great, thanks - I wasn't aware of the first and forgot the second. Added to the list.

PR #1149, opened 2019-09-04
Use repo URL from the app repository (no longer requiring Details.RepoURL)

absoludity:
When initializing the `chart.Chart` util object, store the fetched `AppRepository` custom resource, and use this as the source of the repo URL (no longer depending on the Details.RepoURL, which will be removed in the next PR).


absoludity:
Abandoning in favour of #1150 (need the branch in the kubeapps git repo so I target it for another)

PR #1151, opened 2019-09-04
App.upgrade sends app repo name

absoludity:
Removes requiredRBAC role from DeploymentForm and updates the App.upgrade method to pass the app repo name, as per Andres' comment https://github.com/kubeapps/kubeapps/pull/1147#pullrequestreview-282873375

Also updates both create and upgrade to no longer send the repoURL (in #1150 tiller-proxy is updated to use the fetched app repo's repoURL instead).

Ref #1110

### DIFF
--- a/dashboard/src/shared/App.test.ts
+++ b/dashboard/src/shared/App.test.ts
@@ -126,7 +126,7 @@ describe("App", () => {
         },
       },
     } as IChartVersion;
-    it("creates an app in a namespace", async () => {
+    it("upgrades an app in a namespace", async () => {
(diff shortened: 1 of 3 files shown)


dashboard/src/shared/App.test.ts
```diff
@@ -102,13 +102,45 @@ describe("App", () => {
...
+            name: "test",
+            repo: {
+              name: "testrepo",
+              url: "http://example.com/charts",
+            },
+          },
+        },
+      },
+    } as IChartVersion;
+    it("creates an app in a namespace", async () => {
```
andresmgot: `it upgrades an app`

PR #1152, opened 2019-09-04
Design document for deployment and upgrade improvements

### LINKED ISSUE #1144: [design] Simplify the deployment and upgrade forms
Most of the applications include a complex/difficult to read values.yaml. Also, most of users only need to modify some high-level settings like the hostname, the username or the password for the admin account.

The goal of this issue is to come up with a design document to propose a solution to simplify this process so we improve the deployment and upgrade experience for users that are not experts in each application.

andresmgot:
Closes #1144


docs/architecture/design-proposals/dashboard/deployment-improvements.md
```diff
@@ -0,0 +1,197 @@
...
+    "resources": "resources",
+    "enableMetrics": "enableMetrics",
+    "replicas": "replicas"
+  }
+}
+```
+
+This `values.schema.json` is a valid JSON schema for the `values.yaml` of WordPress. It can be more complete since it's possible to include every single parameter of the values.yaml file.
+
+In order to be able to provide a richer presentation of the form we are including the key `kubeapps`. This allows us to know which parameters we need to represent in the form and how to map them to valid values. In case there is a parameter that is not known we should be able to parse it as a generic input.
```
absoludity: So this both answers a previous question of mine (how to know which field to include) and raises another: why include the `kubeapps` key here? Won't this imply something else to the helm validation (ie. that there's a kubeapps key in the values.yaml etc. etc.) I get that it may not affect things since it won't be required, but still seems to pollute the purpose and abuse what `values.schema.yaml` is intended for.

Why not just use `values.schema.yaml` for the schema only (as intended), and include a separate `deployment-form-fields.json` or similar to represent the main fields to be presented to users when deploying? I don't see any reason for it to be kubeapps specific, but rather metadata for the chart indicating which parameters in the values.yaml are important for a simplified deploy experience.

Or another option, rather than a separate key (which has implications), check whether json schema allows keys for each field other than type, title, description etc., which could be used to tag those fields which are relevant for simple deploys. I [see in the spec](https://json-schema.org/latest/json-schema-validation.html#rfc.section.10) that "title" and "description" are standard schema annotations and that "These general-purpose annotation keywords provide commonly used information for documentation and user interface display purposes. They are not intended to form a comprehensive set of features. Rather, additional vocabularies can be defined for more complex annotation-based applications. " Might be worth investigating.

docs/architecture/design-proposals/dashboard/deployment-improvements.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Deployment Workflow
+
+This section defines how the deployment logic will work based on the JSON Schema definition above.
+
+1. The deployment component loads the details of the Chart. If the chart doesn't contain the file `values.schema.json`, continue as today; Only the advanced deployment is allowed. In other case, go to the next step.
+2. Load the `values.schema.json`. If it doesn't contain the `kubeapps` key, auto-generate the form using the tool `react-jsonschema-form`. From this point, two different tabs will be shown to the user. If the user changes tab from `basic` to `advanced` the new form will be shown but the values won't be ported. If the `kubeapps` key is defined go to the next step.
```
absoludity: As above, I'd base this on both the `values.schema.json` being present, as well as the (separate) definition of which fields are most relevant for a simple deploy.

docs/architecture/design-proposals/dashboard/deployment-improvements.md
```diff
@@ -0,0 +1,197 @@
...
+3. Kubeapps will support ceratin types of parameters (like `username` or `resources`). If those parameters are present, we will render the rich view of the parameter, for example card-choice instead of a text-box. For parameters not supported, we should be able to fallback to the automatic generation. This may not be possible so in that case the parameter will be ignored (with an error for the developer console). If a parameter has a `title` or a `description`, those should be used in the form.
+4. Once the user clicks on "submit", we can use the `kubeapps` mapping to generate a JSON object. We can also use the JSON schema to validate the the processed object complies with the definition.
```
absoludity: This could be a first step - to simply support validating the values.yaml based on the json schema.

docs/architecture/design-proposals/dashboard/deployment-improvements.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Upgrade Workflow
+
+The upgrade workflow will be similar to the deployment with some caveats:
+
+1. If the application to be upgraded doesn't contain a `values.schema.json`, behave as today. In case the current version contains the file, it should be included in new versions. If that's the case, go to the next step.
+2. If the schema contains the `kubeapps` key, use the current `values` to pre-populate the information of the form. If the schema doesn't contain that key, the renderization gets done by the automatic form generator so we may not be able to pre-fill the app information.
```
absoludity: This implies that we would be auto-generating a form even if the chart doesn't specify which fields for the simple deployment? Wouldn't that mean the form would be as large as the values.yaml? I thought the idea was that we only display a form if the metadata specifying which fields to use is present (ie. currently the `kubeapps` key, but could be a separate file as above).

docs/architecture/design-proposals/dashboard/deployment-improvements.md
```diff
@@ -0,0 +1,197 @@
...
+3. Once the user selects a new version, the new `schema` gets loaded. For parameters existing in the previous version, re-fill the form with the previous values. In case there is a new parameter in the form, leave it empty (or use its default value). If a parameter is no longer used, ignore it.
```
absoludity: Rendering/pre-populating the form seems to be happening in both 2 and 3 above. I'm assuming you mean that it'll only be rendered once the user has selected the new version and we have the new schema etc.?

Otherwise, the logic sounds sane - what I'd expect as a user (this wasn't clear earlier where I had a question about this).

docs/architecture/design-proposals/dashboard/deployment-improvements.md
```diff
@@ -0,0 +1,197 @@
...
+4. The logic to generate the final object and validate it is the same than for the deployment.
+
+## Development
+
+The amount of code/time required to implement this feature is considerably high. To avoid large PRs or temporary branches, we suggest to incrementaly implement this feature, merging changes on `master` but protecting the current status using a feature flag. If the feature flag is disabled (default behavior), the current deployment and upgrade form will be shown. For developers (or anyone that wants to test the feature), we will enable this flag to activate the new form. We can store this flag in the Kubeapps ConfigMap. Once the feature is finished, this flag will be removed from the configuration, enabling the feature for everyone.
```
absoludity: +100.
Seems like the steps could be planned too - just off the top of my head:
 * Add support for loading `values.schema.json` when present
 * Validate submitted values.json (behind flag)
 * component to parse values.yaml using schema to redux state
 * update to deploy values from the redux state directly, rather than the input field
 * add component to render form based on selected fields and json schema
 * update component so that form changes update the redux state for the values
 * update the upgrade to work similarly
 * profit!

(again, that's just me thinking out aloud - you might have a different idea - great) EDIT: I see you've got something similar down below :)

docs/architecture/design-proposals/dashboard/deployment-improvements.md
```diff
@@ -0,0 +1,197 @@
...
+  - WordPress
+  - Apache
+  - PostgreSQL
+- Create a test step to ensure that if the `values.schema.json` exists, the default `values.yaml` complies with it.
+- Create the feature flag.
+- Allow to automatically convert the `values.schema.json` to a form (ignoring the `kubeapps` key).
+- Design and create the components for the rich view, using the `kubeapps` definition.
+- Design and create generic components in case `kubeapps` is defined with unknown parameters.
+- Adapt the DeploymentForm to render the `kubeapps` form.
+- Adapt the UpgradeForm to render the `kubeapps` form.
```
absoludity: YMMV, but as above, I'd include using `values.schema.json` for exactly what it's defined for first - validation. Then also storing the validated json in the redux state (so that it is updated by either the text component, or the form). Just a thought - see what you think.

docs/architecture/design-proposals/dashboard/deployment-improvements.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Known limitations
+
+There will be cases in which the basic form won't be enough. For example, in the case of WordPress, the user will be able to set the resource requirements for the WordPress instance but not for the MariaDB dependency. The team managing the charts should decide, one by one, which parameters we should include and they should communicate with us in case a new `parameter` is needed.
```
absoludity: This is, (IMO) another reason to not make it kubeapps specific, but rather, something that communicates the main fields for a simplified deployment.

docs/architecture/design-proposals/dashboard/deployment-improvements.md
andresmgot: Yes, my idea here is that editing a form is easier than editing YAML so there would be some improvement (even if it is not a "basic" form). But it's true that the `values.schema.json` doesn't need to represent `values.yaml` file entirely. Maybe a developer is only interested on validating a specific part of the form and rendering a form from that won't be useful. I will remove this.

docs/architecture/design-proposals/dashboard/deployment-improvements.md
andresmgot: What I meant is that when opening the upgrade form, it should show the information as it were when the user clicked on "deploy" (just as we do today). Then when switching versions, port the valuest to the new version if possible.

docs/architecture/design-proposals/dashboard/deployment-improvements.md
andresmgot: Let me add the validation as a standalone task.

Regarding the redux state: because of the reason I explained in the previous comment, I think it cannot be used for both the basic and the advanced deployment. In any case, both are in the same component (`DeploymentForm`) so it's not necessary to use redux, only using the `state` of the component will be enough.

docs/architecture/design-proposals/dashboard/deployment-improvements.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Desired User Experience (UX)
+
+### Deployment and Upgrade Form
+
+When users reach the deployment form, the desired behavior is to have two different views, one for the basic form and another for an advanced deployment. That can be represented as two different tabs. The basic deployment should contain all the basic parameters while the advanced form will show the information as we do today. For the first version, changes made in the basic deployment form won't affect the advanced form and viceversa. Only when the user clicks on "submit" the current form information (either basic or advanced) will be sent to the backend.
+
+This is the proposed view for the basic deployment form:
+
+<img src="./img/deployment-form-basic.png" width="500px">
```
Angelmmiguel: What do you think about `Visual Editor` / `values.yaml` instead of `Basic` /  `Advanced`? To me, it sounds more close to the Helm terminology.

docs/architecture/design-proposals/dashboard/deployment-improvements.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Desired User Experience (UX)
+
+### Deployment and Upgrade Form
+
+When users reach the deployment form, the desired behavior is to have two different views, one for the basic form and another for an advanced deployment. That can be represented as two different tabs. The basic deployment should contain all the basic parameters while the advanced form will show the information as we do today. For the first version, changes made in the basic deployment form won't affect the advanced form and viceversa. Only when the user clicks on "submit" the current form information (either basic or advanced) will be sent to the backend.
```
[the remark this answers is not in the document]
Angelmmiguel: Agree with @absoludity this is an important feature. As AWS does with Visual / JSON editors, changes should be reflected on both options. I can start filling the basic form and them switch to the advanced one to change something.

Without entering on implementation details, I believe we can define a mapping between the basic form and the `values.yaml` keys. I agree that replacing content may introduce unexpected errors. However, by defining this mapping, we can take the value from the `vaules.yaml` and print it on the basic form and viceversa.

docs/architecture/design-proposals/dashboard/deployment-improvements.md
```diff
@@ -0,0 +1,197 @@
...
+
+<img src="./img/deployment-form-basic.png" width="500px">
+
+And for the advanced tab:
+
+<img src="./img/deployment-form-adv.png" width="500px">
+
+*Note that the design may vary, the final proposal will be submitted in a different PR
+
+Once a user has deployed the application, when trying to upgrade, the basic form should be shown. The form will be pre-filled with the values of the previous version. When switching to a new version, those basic parameters should be maintained.
```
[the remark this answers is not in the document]
Angelmmiguel: We may want to store a separate YAML file that stores the overrides to the `values.yaml` file. With this information we can prefill the basic and advanced forms.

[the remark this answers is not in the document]
andresmgot: We don't have a place to store that kind of information right know. The only information that we have is what Tiller stores. We may be able to compare the default `values` of a chart version with the current values and do that, but as I said, that can be done in a follow up feature, it's not directly related to the basic form. In any case, if you guys think that's important I can include it here and make it a different subtask.

docs/architecture/design-proposals/dashboard/deployment-improvements.md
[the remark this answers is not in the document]
andresmgot: That sounds doable but not sure if at the end it will be. I am mostly worried about order, comments and multiple spaces that's why I didn't want to add a compromise here, but if you think it's a must-have I can add and see what we can do.

docs/architecture/design-proposals/dashboard/deployment-improvements.md
andresmgot: `Visual Editor`? I am not sure what you mean with that. I am okay changing those names though.

docs/architecture/design-proposals/dashboard/deployment-improvements.md
```diff
@@ -0,0 +1,201 @@
...
+
+This JSON Schema allows us to know the structure of the `values.yaml` file but, to know which parameters should be represented in the basic form, we need more information. In particular we need to map well known parameters, like "Disk size", to parameters in the `values.yaml`. For doing so, we can include a new file: `values.form.json`. Having those two pieces of information we are able to deliver the desired experience. See the next section for an example.
```
Angelmmiguel: My concern about this is that we're forcing Helm chart maintainers to introduce a new file. Before asking people to add this file, It would be nice if we can analyze several `values.yaml` files to find common patterns. If we find them, maybe this file is not required. 

This is just a comment. I know it's quite difficult to find pattern on these files 😅

docs/architecture/design-proposals/dashboard/deployment-improvements.md
```diff
@@ -0,0 +1,201 @@
...
+  "enableMetrics": "enableMetrics",
+  "replicas": "replicas"
+}
+```
+
+## Deployment Workflow
+
+This section defines how the deployment logic will work based on the JSON Schema definition above.
+
+1. The deployment component loads the details of the Chart. If the chart doesn't contain the file `values.schema.json` or the file `values.form.json`, continue as today; Only the advanced deployment is allowed. In other case, go to the next step.
```
Angelmmiguel: Sounds reasonable for the first version. We can check how these files are embraced by the community :+1:

docs/architecture/design-proposals/dashboard/deployment-improvements.md
Angelmmiguel: Sounds good to me! :+1:

docs/architecture/design-proposals/dashboard/deployment-improvements.md
Angelmmiguel: Yes, I think for a first version it's fine. Future versions may give more flexibility to chart maintainers to define custom fields for the form that makes sense for the given application :)

docs/architecture/design-proposals/dashboard/deployment-improvements.md
andresmgot: I guess that with the new file you mean the `values.form.json` (not the `values.schema.json` that is made for Helm v3 as a "linter").

It's true that it includes some overhead for the maintainers but this file is not that big so it's relatively easy to implement. We are also not targeting all the apps in the registries, covering just a few of them will help many users.

In any case I think there is no other option. We are not going to be able to know if the username is called `wordpressUsername`, `username` or `wpUsername` and it may change in the future (we don't have control over the development of that chart). If we try to auto-guess those parameters several things are very likely to happen:

 - We will miss paramaters that should be included because of a naming convention we are not aware.
 - We may include parameters that are not relevant.
 - We may parse the wrong parameters. If we are looking for the disk size for wordpress we may end up modifying the one for mariadb.

docs/architecture/design-proposals/dashboard/deployment-improvements.md
andresmgot: > The problem is that in Kubeapps we would need to recursively walk through this JSON in order to find special annotations. Also I think it's more difficult for the maintainers to locate and update these fields.

Thinking about this twice, maybe it's better to check this with some charts maintainers. If they have a strong opinion about using a different file vs adding an extra property to some elements of the `schema` we should take that into account, mostly because they are the ones that will maintain this.

docs/architecture/design-proposals/dashboard/deployment-improvements.md
absoludity: I still think this could be part of the `values.schema.json` - just an annotation/field to identify fields which should be included. I'll comment on the other thread we had about that (which disappeared from the diff since your change).

docs/architecture/design-proposals/dashboard/deployment-improvements.md
```diff
@@ -0,0 +1,201 @@
...
+  - port: Number. Database port.
+- Enable Hostname. Object. Needs to be enabled. Hostname specifics.
+  - Hostname: String. Hostname to use.
+  - Enable TLS: Boolean. Needs to be enabled. Requires Cert Manager.
+- Disk size. Choice. Sizes to be defined per application. Small, Medium, Large or Other. In case of Other, allows to write a custom disk size.
+- Required Resources. Choice. Resources to be defined per application. Small, Medium, Large or Other. In case of Other, allows to write custom CPU an Memory requirements.
+- Enable Metrics. Boolean. Use Prometheus exporter.
+- Replicas. Number. Number of replicas.
+
+Note that we should be able to represent any parameter in a generic way. We are just defining a set of well-known parameters to be able to render a richer view.
```
absoludity: Again, if we could use the auto-generation based on the schema, for the specific fields, those fields could include an annotation for choices (which turn the field into a select). It seems from the above that you are thinking about standard parameters that we (kubeapps) defines for a richer view, but I think it could remain general (ie. not dependent on kubeapps' current supported parameters for richer views), and instead be based solely on the schema (with extra annotations to identify a field for the form, and choices to present etc.). Does that make sense? Or do you think that would be difficult?

docs/architecture/design-proposals/dashboard/deployment-improvements.md
```diff
@@ -0,0 +1,201 @@
...
+    "replicas": { "type": "integer", "title": "Number of Replicas"}
+  },
+
+  "required": ["wordpressUsername", "wordpressEmail", "persistence", "resources"],
+}
+```
+
+This `values.schema.json` is a valid JSON schema for the `values.yaml` of WordPress. It can be more complete since it's possible to include every single parameter of the values.yaml file.
+
+In order to be able to provide a richer presentation of the form we are including the file `values.form.json`. This allows us to know which parameters we need to represent in the form and how to map them to valid values. In case there is a parameter that is not known we should be able to parse it as a generic input.
```
absoludity: > allows us to know which parameters we need to represent in the form

See above comments (ie. we could do this in the schema? Not sure why we wouldn't)

> and how to map them to valid values

We'd get this for free I think? I haven't checked the auto form generation from the schema, but I assume it creates input elements with names like `mariadb.enabled` rather than needing to map to other names? But again, haven't checked.

docs/architecture/design-proposals/dashboard/deployment-improvements.md
absoludity: > I thought about that but that may be more complex. You can add arbitrary properties in JSON Schema so that's not a problem. The problem is that in Kubeapps we would need to recursively walk through this JSON in order to find special annotations.

I don't think this would be an issue? I've not looked at the form generation and whether it is pluggable, but identifying which fields in the schema have annotations should not be hard.

> Also I think it's more difficult for the maintainers to locate and update these fields.

Possibly - I can only think from my own pov. To me it seems part of the schema. I was initially thinking that it might be using the schema for something it's not intended, but when I saw the above quote "These general-purpose annotation keywords provide commonly used information for documentation and user interface display purposes." from the spec, that worry went away :)

> Thinking about this twice, maybe it's better to check this with some charts maintainers. If they have a strong opinion about using a different file vs adding an extra property to some elements of the schema we should take that into account, mostly because they are the ones that will maintain this.

That's a great idea, especially if it was on the helm list and clear that it's not kubeapps only info, but would be useful to any UI interacting with a chart. Perhaps do the POC experiment first (simple schema + extra info -> metadata for form fields).

docs/architecture/design-proposals/dashboard/deployment-improvements.md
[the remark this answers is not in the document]
andresmgot: > I haven't tried it, but from https://github.com/eemeli/yaml/ : "As a particularly distinguishing feature, yaml fully supports reading and writing comments and blank lines in YAML documents." . We could do a quick test to see if it parses Kubeapss' values.yaml and restringifies it without a diff?

Apparently, it works as advertised. The only issue I have find is that if a comment has blank lines within it, the library removes them:

```
> a
'foo: bar\n\n# THIS IS A COMMENT\n\n# Other comment\nbar:\n  foobar: barfoo\n\n'
> c = yaml.parseDocument(a)
Document {
...
> c.set("foo", "not-foo")
> c.toString()
'foo: not-foo\n\n# THIS IS A COMMENT\n# Other comment\nbar:\n  foobar: barfoo\n'
```

In any case, it's good enough so let's go with that.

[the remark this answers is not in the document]
absoludity: Excellent. I assume the parsed value easy to work with?

docs/architecture/design-proposals/dashboard/deployment-improvements.md
andresmgot: After checking with some chart maintainers, they agree on using special annotations versus the separated file. I will modify the document with that.

andresmgot:
I have addressed the feedback here. I think we are good to go! Let me know if you think otherwise @absoludity

docs/architecture/design-proposals/dashboard/deployment-improvements.md
```diff
@@ -0,0 +1,204 @@
...
+    "replicas": { "type": "integer", "title": "Number of Replicas", "form": "replicas" }
+  },
+
+  "required": ["wordpressUsername", "wordpressEmail", "persistence", "resources"],
+}
+```
+
+This `values.schema.json` is a valid JSON schema for the `values.yaml` of WordPress. It can be more complete since it's possible to include every single parameter of the values.yaml file.
+
+Note that in order to be able to provide a richer presentation of the form we are including the tags `form={}` with the set of well-known parameters. This allows us to know which parameters we need to represent in the form and how to map them to valid values. In case there is a parameter that is not known we should be able to parse it as a generic input.
```
absoludity: Great. So any field with a `form` key is to be included on the form, using input matching the schema (ie. those with `enum` will be selects, those with `string` will be input, etc). The well-known ones (like `diskSize`) can be further customised or have useful graphic labels or whatever, but the generic rendering should be a good start.

Not sure if it's also worth supporting something like `tabIndex=1` or similar to enable chart authors to define the form field layout order?

andresmgot: > Not sure if it's also worth supporting something like `tabIndex=1` or similar to enable chart authors to define the form field layout order?

That could be an interesting feature but probably it's better to keep it as simple as possible for the moment for the maintainers. If they require that property we can add it afterwards.

Issue #1146, opened 2019-09-02
[tests] Local e2e tests are flaky


andresmgot:
I have noticed that end-to-end tests (local) are failing some times in the chart-test step:

https://circleci.com/gh/kubeapps/kubeapps/8954

We should at least print the logs of the tests to know what could be happening and eventually fix the race condition.

PR #1153, opened 2019-09-04
Increase debug of helm tests

andresmgot:
Ref  #1146

Sometimes the helm tests fails but since we are not printing the logs we are not sure what could be happening. The goal of this PR is to add verbosity to the tests so we can find the cause of the issue when it appears.

### DIFF
--- a/chart/kubeapps/templates/tests/test-chartsvc.yaml
+++ b/chart/kubeapps/templates/tests/test-chartsvc.yaml
@@ -16,5 +16,5 @@ spec:
     command:
     - sh
     - -c
-    - curl -v -o /tmp/output $CHARTSVC_HOST:$CHARTSVC_PORT/v1/charts && cat /tmp/output && cat /tmp/output | grep wordpress
+    - curl -o /tmp/output $CHARTSVC_HOST:$CHARTSVC_PORT/v1/charts && cat /tmp/output && cat /tmp/output | grep wordpress
   restartPolicy: Never
--- a/chart/kubeapps/templates/tests/test-chartsvc.yaml
+++ b/chart/kubeapps/templates/tests/test-chartsvc.yaml
@@ -16,5 +16,5 @@ spec:
     command:
     - sh
     - -c
-    - curl $CHARTSVC_HOST:$CHARTSVC_PORT/v1/charts | grep wordpress
+    - curl -v -o output $CHARTSVC_HOST:$CHARTSVC_PORT/v1/charts && cat output && cat output | grep wordpress
   restartPolicy: Never
--- a/chart/kubeapps/templates/tests/test-chartsvc.yaml
+++ b/chart/kubeapps/templates/tests/test-chartsvc.yaml
@@ -16,5 +16,5 @@ spec:
     command:
     - sh
     - -c
-    - curl -v -o output $CHARTSVC_HOST:$CHARTSVC_PORT/v1/charts && cat output && cat output | grep wordpress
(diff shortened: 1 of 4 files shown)


chart/kubeapps/templates/tests/test-tiller-proxy.yaml
```diff
@@ -18,5 +18,5 @@ spec:
     command: 
     - sh
     - -c
-    - "curl -ik -H \"Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)\" $TILLER_PROXY_HOST:$TILLER_PROXY_PORT/v1/releases | grep $KUBEAPPS_RELEASE"
+    - "curl -v -o /tmp/output -ik -H \"Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)\" $TILLER_PROXY_HOST:$TILLER_PROXY_PORT/v1/releases && cat /tmp/output && cat /tmp/output | grep $KUBEAPPS_RELEASE"
```
Angelmmiguel: I guess these clusters are temporary, but here we may expose sensitive information to the CI logs

script/e2e-test.sh
```diff
@@ -97,4 +97,27 @@ echo "All deployments ready. PODs:"
...
+
+helm test ${HELM_CLIENT_TLS_FLAGS} kubeapps-ci
+code=$?
+
+set -e
+
+if [[ "$code" != 0 ]]; then
+  echo "PODS status on failure"
+  kubectl get pods -n kubeapps
+  for pod in $(kubectl get po -l release=kubeapps-ci -oname -n kubeapps); do
```
Angelmmiguel: Same here about exposing sensitive data

chart/kubeapps/templates/tests/test-tiller-proxy.yaml
andresmgot: I was going to say no because this is meant for the local cluster using `kind` but the same is executed for GKE and showing the headers would show a valid token for the cluster. Good catch! I am removing the `-v`

script/e2e-test.sh
andresmgot: in this case there is nothing sensitive for the temporary cluster

chart/kubeapps/templates/tests/test-chartsvc.yaml
```diff
@@ -16,5 +16,5 @@ spec:
     command:
     - sh
     - -c
-    - curl $CHARTSVC_HOST:$CHARTSVC_PORT/v1/charts | grep wordpress
+    - curl -o /tmp/output $CHARTSVC_HOST:$CHARTSVC_PORT/v1/charts && cat /tmp/output && cat /tmp/output | grep wordpress
```
absoludity: Why are you `cat`ing the whole /tmp/output then also redoing with the pipe? (here and everywhere).
That just means we see some of the output twice, no?

andresmgot: If we just do the `grep` pipe and the output doesn't contain what we expect it to contain (like for example an error), the actual content would be filtered out. That's why I am showing the output and then do the "grep" as the test.

andresmgot:
>  Similar to Angel's question, wondering where this output is going to be publicly visible? CircleCI? What credentials could potentially be exposed, if any?

Yes, in CircleCI. We are showing the logs of the pods of the temporary cluster. We are not showing any credentials there. You can see an example of what's being printed here:

(note that I have removed the `-v` from the `curl` calls in the tests so the headers won't be shown)

https://circleci.com/gh/kubeapps/kubeapps/9043?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

PR #1154, opened 2019-09-05
Remove no longer used code since #1151

absoludity:
Removes code no longer required after #1151 .

This required a bit of refactoring for one of the tests, now that we only support the auth from the AppRepository custom resource. Let me know if it's not clear.

NOTE: while QAing this (together with the most recent branches), I see that when upgrading an app, the dashboard is still requesting AppRepositories (a `REQUEST_REPOS` action is dispatched). It appears to be because the `SelectRepoForm` is used during upgrade, which calls `checkChart`. Thoughts @andresmgot ? Not a blocker for this PR, but I'd be keen to remove that requirement if we can.

Ref: #1110

### DIFF
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -60,12 +60,8 @@ func init() {
 
 // Details contains the information to retrieve a Chart
 type Details struct {
-	// RepoURL is the URL of the repository. Defaults to stable repo.
-	RepoURL string `json:"repoUrl,omitempty"`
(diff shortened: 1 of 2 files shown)


andresmgot:
> Thoughts @andresmgot ? Not a blocker for this PR, but I'd be keen to remove that requirement if we can.

Replied at: https://kubernetes.slack.com/archives/C9D3TSUG4/p1567671364006000

PR #1159, opened 2019-09-06
Retrieve latest tag from GitHub when syncing the chart

### LINKED ISSUE #1156: bitnami/charts syncer tool should take latest image revisions
Right now the bitnami/charts syncer script takes the latest kubeapps version and append `-r0` to form the bitnami component image tag. 
This image won't be updated again until the next version of kubeapps is released so it is possible that some CVE affects this image.

Instead of using this method to obtain the tag, we can check the latest release of the bitnami-component. For example, check https://github.com/bitnami/bitnami-docker-kubeapps-dashboard/releases to get the latest revision of the image released.

andresmgot:
Fixes #1156

Retrieve the latest tag from GitHub instead of hardcoding the tag `-r0`.

This needs to set the env var TOKEN in CircleCI to avoid hitting the GitHub API quota.

Tested manually:

```patch
diff --git a/chart/kubeapps/values.yaml b/chart/kubeapps/values.yaml
index 59742084..938095cb 100644
@@ -221,8 +221,8 @@ dashboard:
   replicaCount: 2
   image:
     registry: docker.io
-    repository: kubeapps/dashboard
-    tag: latest
+    repository: bitnami/kubeapps-dashboard
+    tag: 1.5.0-debian-9-r16
   service:
     port: 8080
   livenessProbe:
```

cc/ @tompizmor

### DIFF
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -221,11 +221,8 @@ dashboard:
   replicaCount: 2
   image:
     registry: docker.io
-    # Once issue #1156 is fixed we can continue using kubeapps repository instead of bitnami fixed tags
-    # repository: kubeapps/dashboard
-    # tag: latest
-    repository: bitnami/kubeapps-dashboard
-    tag: 1.5.0-debian-9-r14
+    repository: kubeapps/dashboard
+    tag: latest
   service:
     port: 8080
   livenessProbe:
--- a/script/chart_sync_utils.sh
+++ b/script/chart_sync_utils.sh
@@ -50,8 +50,8 @@ replaceImage() {
     local targetImageEscaped="bitnami\/kubeapps-${service}"
 
     local header=""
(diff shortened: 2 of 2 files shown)


andresmgot:
> Why does your diff from running the script not update tiller-proxy and apprepository-controller? (or perhaps you ran it before adding that?)

This diff reverts https://github.com/kubeapps/kubeapps/pull/1157 before, all of them were using `latest`. Note that the point of the script is changing `latest` -> to the latest revision (but those changes will be only pushed in the bitnami repository)

> Also, is the $TOKEN already available when the syncer script is run? Similarly, is jq available already?

The TOKEN is not yet there, I will add that to CircleCI. For `jq` I am not sure, I will check that.

andresmgot:
> The TOKEN is not yet there, I will add that to CircleCI

I noticed that we were using `ACCESS_TOKEN` before for the Github API.

PR #1167, opened 2019-09-19
Add defaultNamespace to auth state.

### LINKED ISSUE #1162: Kubeapps shows "default" namespace if user cannot list namespaces
If a service account is created for a user in a specific namespace (and without permission to list namespaces), and the associated credential is used to login, the namespace switcher shows "default" rather than the specific namespace.

Kubeapps should default to the namespace of the service account rather than the "default" namespace which to which the user may not have access.

Ideally, kubeapps would remember the selected namespace so when returning (without re-authing) it remains selected.

absoludity:
Part 1 to fix #1162. This diff adds jsonwebtoken and updates the Auth to parse this and store the `defaultNamespace` in state when logging in.

Part 2 updates the header component/container to use this.

### DIFF
--- a/dashboard/src/shared/Auth.ts
+++ b/dashboard/src/shared/Auth.ts
@@ -1,7 +1,10 @@
 import Axios, { AxiosResponse } from "axios";
+import * as jwt from "jsonwebtoken";
 const AuthTokenKey = "kubeapps_auth_token";
 const AuthTokenOIDCKey = "kubeapps_auth_token_oidc";
 import { APIBase } from "./Kube";
+
+export const DEFAULT_NAMESPACE = "default";
 
 export class Auth {
   public static getAuthToken() {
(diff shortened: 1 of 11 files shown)


dashboard/src/shared/Auth.ts
```diff
@@ -77,4 +80,18 @@ export class Auth {
...
+  // TODO(mnelson): until we call jwt.verify on the token during validateToken above
+  // we use a default namespace for both invalid tokens and tokens without the expected
+  // key.
+  public static defaultNamespaceFromToken(token: string) {
+    const payload = jwt.decode(token);
+    const namespaceKey = "kubernetes.io/serviceaccount/namespace";
+    if (!payload || !payload[namespaceKey]) {
+      return DEFAULT_NAMESPACE;
+    }
+    return payload[namespaceKey] || DEFAULT_NAMESPACE;
```
Angelmmiguel: You're checking it `payload[namespaceKey]` is defined in the previous conditional. I'd remove the `|| DEFAULT_NAMESPACE` clause from here to avoid logic duplication.

absoludity: Woops, left after a refactor. Thanks.

PR #1168, opened 2019-09-19
Use default namespace from auth state rather than 'default'.

### LINKED ISSUE #1162: Kubeapps shows "default" namespace if user cannot list namespaces

absoludity:
Follows on from #1167 to fix #1162 .

Conceptually this just removes any `default` option for the header/namespace selector, replacing it with the `state.auth.defaultNamespace` set in the previous branch. When testing I had some issues because the redirect from the root route was using `state.namespaces.current` which is no longer necessary set - eg., when first logging in after a refresh.

Also includes a drive-by to no longer ignore errors when fetching namespaces (at least storing the error in the state).

I'll follow-up with one or two small drive-by's that I noticed (one is that we retain state when logging out, for namespaces etc.)

### DIFF
--- a/dashboard/src/actions/namespace.ts
+++ b/dashboard/src/actions/namespace.ts
@@ -13,7 +13,11 @@ export const receiveNamespaces = createAction("RECEIVE_NAMESPACES", resolve => {
   return (namespaces: string[]) => resolve(namespaces);
 });
 
-const allActions = [setNamespace, receiveNamespaces];
+export const errorNamespaces = createAction("ERROR_NAMESPACES", resolve => {
+  return (err: Error, op: "list") => resolve({ err, op });
+});
+
+const allActions = [setNamespace, receiveNamespaces, errorNamespaces];
 export type NamespaceAction = ActionType<typeof allActions[number]>;
 
 export function fetchNamespaces(): ThunkAction<Promise<void>, IStoreState, null, NamespaceAction> {
@@ -23,7 +27,7 @@ export function fetchNamespaces(): ThunkAction<Promise<void>, IStoreState, null,
       const namespaceStrings = namespaces.items.map((n: IResource) => n.metadata.name);
       dispatch(receiveNamespaces(namespaceStrings));
     } catch (e) {
-      // TODO: handle namespace call error
+      dispatch(errorNamespaces(e, "list"));
       return;
     }
   };
--- a/dashboard/src/containers/RoutesContainer/RoutesContainer.tsx
+++ b/dashboard/src/containers/RoutesContainer/RoutesContainer.tsx
@@ -4,8 +4,8 @@ import { withRouter } from "react-router";
 import { IStoreState } from "../../shared/types";
 import Routes from "./Routes";
 
-function mapStateToProps({ namespace }: IStoreState) {
-  return { namespace: namespace.current };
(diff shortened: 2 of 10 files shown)


dashboard/src/containers/RoutesContainer/RoutesContainer.tsx
```diff
@@ -4,8 +4,8 @@ import { withRouter } from "react-router";
 import { IStoreState } from "../../shared/types";
 import Routes from "./Routes";
 
-function mapStateToProps({ namespace }: IStoreState) {
-  return { namespace: namespace.current };
+function mapStateToProps({ auth, namespace }: IStoreState) {
+  return { namespace: namespace.current || auth.defaultNamespace };
```
Angelmmiguel: [Not important] I understand the `defaultNamespace` is associated to the user and it's fetched during the authentication process. However, I'd expect to be defined as a property of the `namespace` redux state key.

absoludity: Yeah - I did initially think of going that way as I thought it would be natural to get it from the `state.namespace` key, but it is the responsibility of the `state.auth` (and associated reducer). So putting it in `state.namespace` would require the namespace reducer to handle the `auth.setAuthenticated` action - which is easy to do, but seemed odd to reach in like that (I see we do it for more generic things, like when the LOCATION changes which the namespace cares about).

In the end, it seemed quite natural here to use either the current namespace or the default one from the authorization data, but happy to change it if others think otherwise too.

dashboard/src/actions/namespace.ts
```diff
@@ -13,7 +13,11 @@ export const receiveNamespaces = createAction("RECEIVE_NAMESPACES", resolve => {
   return (namespaces: string[]) => resolve(namespaces);
 });
 
-const allActions = [setNamespace, receiveNamespaces];
+export const errorNamespaces = createAction("ERROR_NAMESPACES", resolve => {
+  return (err: Error, op: "list") => resolve({ err, op });
```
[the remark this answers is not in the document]
absoludity: It's probably not necessary here as we only have one operation type (to list namespaces), but I was following the precedent in the errors for other actions, in particular:

https://github.com/kubeapps/kubeapps/blob/master/dashboard/src/actions/catalog.ts#L40

where a number of set string enums are available ("fetch", "create", etc.). So here it's just saying that `op` must be exactly "list" (ie. a string with that value).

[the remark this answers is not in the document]
Angelmmiguel: Oh I got it! Thanks for the clarification

dashboard/src/containers/RoutesContainer/RoutesContainer.tsx
Angelmmiguel: No problem. As you mentioned, `auth` is the responsible of fetching that data, so it's fine to keep it in the `state.auth` object too :)

PR #1171, opened 2019-09-23
Clear namespaces on logout

absoludity:
Just two tiny clean-ups I'd noted in some earlier work.

First, ensure that we clear the namespaces in the client state when logging out: I kept finding that I would re-authenticate with a token for a user with access to only a single namespace, but find the UI still presented all namespaces, until I refreshed.

Second, avoid 302 permanent redirects when checking namespaces api endpoint.

### DIFF
--- a/dashboard/src/actions/namespace.ts
+++ b/dashboard/src/actions/namespace.ts
@@ -17,7 +17,11 @@ export const errorNamespaces = createAction("ERROR_NAMESPACES", resolve => {
   return (err: Error, op: "list") => resolve({ err, op });
 });
 
-const allActions = [setNamespace, receiveNamespaces, errorNamespaces];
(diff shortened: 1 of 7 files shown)


dashboard/src/actions/namespace.ts
```diff
@@ -17,7 +17,11 @@ export const errorNamespaces = createAction("ERROR_NAMESPACES", resolve => {
   return (err: Error, op: "list") => resolve({ err, op });
 });
 
-const allActions = [setNamespace, receiveNamespaces, errorNamespaces];
+export const clearNamespaces = createAction("CLEAR_NAMESPACES", resolve => {
```
andresmgot: You can do `export const clearNamespaces = createAction("CLEAR_NAMESPACES")` if you don't need any payload.

absoludity: Thanks.

PR #1172, opened 2019-09-23
Add tests for values.yaml state in DeploymentForm

absoludity:
I was wanting to better understand how the state is currently managed for values.yaml in charts.

Seems that we have global state in redux for the selected chart version, which we copy into the form's local state which the user can modify - which is fine, but while there, I noticed some redundant state, so added tests then removed it.

### DIFF
--- a/dashboard/src/components/DeploymentForm/DeploymentForm.tsx
+++ b/dashboard/src/components/DeploymentForm/DeploymentForm.tsx
@@ -10,7 +10,7 @@ import LoadingWrapper from "../LoadingWrapper";
 import "brace/mode/yaml";
 import "brace/theme/xcode";
 
-interface IDeploymentFormProps {
+export interface IDeploymentFormProps {
   kubeappsNamespace: string;
   chartID: string;
   chartVersion: string;
@@ -29,7 +29,7 @@ interface IDeploymentFormProps {
   namespace: string;
 }
 
-interface IDeploymentFormState {
+export interface IDeploymentFormState {
   isDeploying: boolean;
   // deployment options
   releaseName: string;
--- a/dashboard/src/components/DeploymentForm/DeploymentForm.tsx
+++ b/dashboard/src/components/DeploymentForm/DeploymentForm.tsx
@@ -39,6 +39,7 @@ export interface IDeploymentFormState {
   latestSubmittedReleaseName: string;
   namespace: string;
   appValues?: string;
+  valuesModified: boolean;
 }
 
 class DeploymentForm extends React.Component<IDeploymentFormProps, IDeploymentFormState> {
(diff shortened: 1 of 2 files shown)


dashboard/src/components/DeploymentForm/DeploymentForm.tsx
```diff
@@ -84,7 +82,8 @@ class DeploymentForm extends React.Component<IDeploymentFormProps, IDeploymentFo
       return;
     }
 
-    if (!this.state.valuesModified) {
+    const valuesModified = selected.values !== this.state.appValues;
+    if (!this.state.appValues || !valuesModified) {
```
andresmgot: mm, this would detect as "modified" if the user selects a different version (without changing any value) and the default values for both versions are different, wouldn't it? If that's the case the defaults for the new selected version won't be applied.

Also, another minor thing is that if the user deletes the values (`appValues = ""`) it would be modified but we would still replace that with the default values.

absoludity: > mm, this would detect as "modified" if the user selects a different version (without changing any value) and the default values for both versions are different, wouldn't it? If that's the case the defaults for the new selected version won't be applied.

It shouldn't as in that case it's not comparing the `values.yaml` from the newly selected version with the local state, but rather the previously selected version with the local state (`selected.values` here are not from `nextProps`, but the current props), so the condition will be true if there were no modifications from the previously selected chart's `values.yaml` and the values will be updated below... I think? But I'll be sure to double-check that when QAing.

> Also, another minor thing is that if the user deletes the values (appValues = "") it would be modified but we would still replace that with the default values.
Good catch - that was unintentional, but I'm not sure I'd expect any different as a user? (ie. if I clear the values completely, and switch versions I think I'd expect to see the default values?) Let me know what you think - either way I'll add a test to document the behaviour.

andresmgot: > selected.values here are not from nextProps, but the current props

Ah, that could be. If that's that way then it's fine, the code is a bit tricky though. I am not sure if the previous way was more readable? In any case, a comment would help.

> Good catch - that was unintentional, but I'm not sure I'd expect any different as a user? (ie. if I clear the values completely, and switch versions I think I'd expect to see the default values?) Let me know what you think - either way I'll add a test to document the behaviour.

Yes, I think that's fine. The behavior is the same if you leave the `values` empty than if you specify no values so to be explicit I am okay with this change.

absoludity: > Ah, that could be. If that's that way then it's fine, the code is a bit tricky though. I am not sure if the previous way was more readable?

Yeah - I agree. I'll revert the actual change and just leave the added tests (I mainly did this diff as an off-shoot while understanding how the state is currently handled for `values.yml`).

andresmgot:
hey @absoludity, it seems that CircleCI is unable to find your branch, maybe you deleted it? (The tests are failing because of that)

absoludity:
> hey @absoludity, it seems that CircleCI is unable to find your branch, maybe you deleted it? (The tests are failing because of that)

Not sure - the branch exists... I just clicked the "Merge master" button and it's re-triggered the tests.

PR #1175, opened 2019-09-24
Update chart to be deployable with helm 3

absoludity:
This results in Kubeapps deploying without issue in helm 3, but obviously not runnable (tiller-proxy currently errors if it can't connect to tiller).

Nonetheless, without this change the chart errors with helm 3. With this change, the chart can be installed successfully multiple times. It utilises the support for a `crds` directory in the chart, but note that these CRD files cannot be templated (nor should they need to be, since they are system-wide and not reference-counted?).

Verified that I can also install this chart successfully with helm 2.13.1 .

### DIFF
--- a/chart/kubeapps/templates/apprepositories.yaml
+++ b/chart/kubeapps/templates/apprepositories.yaml
@@ -4,19 +4,7 @@ kind: AppRepository
 metadata:
   name: {{ .name }}
   annotations:
-{{- if semverCompare ">=2.14.0" $.Capabilities.TillerVersion.SemVer }}
-    # Using the hook pre-install because since Helm v2.14 the crd-install
-    # hook should be used only for CRDs and the issues related to install
-    # Custom Resources have been already fixed
-    # https://github.com/helm/helm/issues/5756#issuecomment-493653454
     "helm.sh/hook": pre-install
-{{- else }}
-    # Using the hook crd-install to avoid issue w/Helm v1.11
-    # https://github.com/kubeapps/kubeapps/pull/955#pullrequestreview-200764260
-    "helm.sh/hook": crd-install
-    # Make sure this runs after the AppRepository CRD install
-    "helm.sh/hook-weight": "10"
-{{- end }}
   labels:
     app: {{ template "kubeapps.apprepository.fullname" $ }}
     chart: {{ template "kubeapps.chart" $ }}
--- a/chart/kubeapps/crds/apprepository-crd.yaml
+++ b/chart/kubeapps/crds/apprepository-crd.yaml
@@ -1,0 +1,13 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: apprepositories.kubeapps.com
+spec:
+  group: kubeapps.com
+  scope: Namespaced
+  names:
+    kind: AppRepository
+    plural: apprepositories
+    shortNames:
+    - apprepos
+  version: v1alpha1


chart/kubeapps/templates/apprepositories.yaml
```diff
@@ -4,19 +4,7 @@ kind: AppRepository
 metadata:
   name: {{ .name }}
   annotations:
-{{- if semverCompare ">=2.14.0" $.Capabilities.TillerVersion.SemVer }}
```
absoludity: @andresmgot Do we need to continue the pre 2.14 support? If so, I'll dig more to ensure this condition can still work if `$.Capabilities.TillerVersion` is defined and be ignored otherwise.

andresmgot: The first version of 2.14 was released on May so I guess there will be people still using 2.13. There may be a reason for people to still use 2.13 so I would maintain this if possible.

absoludity:
> I am okay with this change (better if we can maintain compatibility with Tiller <2.14)

Yep, I'll dig a bit more with that template condition then.

> but I think we should take a different approach in the mid term.
> 
> Using CRDs is a bit tricky and it's not really necessary for our use case. We would have less problems if we substitute the AppRepository CRD for secrets. WDYT? We can open a different issue in that case.

I'm not sure yet - I don't see a problem (and am probably missing the historical problems). What do you think is tricky? And by not necessary, do you mean because the app-repo controller runs on a cron anyway so it doesn't really require a k8s controller? (ie. it's not necessary to *watch* the resources for additions/modifications because we need to do updates even when the resources don't change). Perhaps it's worth opening an issue to discuss there, with the background :)

andresmgot:
> I'm not sure yet - I don't see a problem (and am probably missing the historical problems). What do you think is tricky? And by not necessary, do you mean because the app-repo controller runs on a cron anyway so it doesn't really require a k8s controller? (ie. it's not necessary to watch the resources for additions/modifications because we need to do updates even when the resources don't change). Perhaps it's worth opening an issue to discuss there, with the background :)

That's not exactly the problem (and we will still use a controller). Let me elaborate in an issue.

chart/kubeapps/templates/apprepositories.yaml
absoludity: > The first version of 2.14 was released in May so I guess there will be people still using 2.13. There may be a reason for people to still use 2.13 so I would maintain this if possible.

So i took a look at the history here. AFAICT (from https://github.com/kubeapps/kubeapps/pull/955#pullrequestreview-200764260 ), the issue is that `pre-install` didn't work with helm 2.11 (see your comment there where both you and Adnan are unable to reproduce with 2.12 but could with 2.11), which is a year old now. It was just that the work-around of using the `crd-install` hook stopped working with 2.13 for non-crd's (the app repos).

So we'd be maintaining this conditional for a year old release. I don't see any way with the built in template vars to determine whether the helm version is > 2.11 moving forward, as the capabilities are only the [kube version and api info now](https://github.com/helm/helm/blob/v3.0.0-beta.3/pkg/chartutil/capabilities.go#L37). So unless I'm missing some other built in that tells us the running tiller version, or we patch helm upstream leaving the tiller version field there but setting to zero or similar, we'd need to make the condition based on whether the field exists, which is [pretty strongly discouraged](https://stackoverflow.com/a/44676179) and requires a custom template function in the template context. All of which leaves me -1 to include the work-around, and +1 to recommend people use a version later than helm 2.11 if this is an issue. I hope that's OK (assuming so, based on your previous approval).

andresmgot: Thanks for the investigation. It's fine then

PR #1180, opened 2019-09-24
Add feature flag for basic form

andresmgot:
Based on the design document #1152 

Right now this feature flag is only usable using `telepresence`. To use it, execute:

```
▶ REACT_APP_KUBEAPPS_ENABLE_BASIC_FORM=true TELEPRESENCE_CONTAINER_NAMESPACE=kubeapps yarn run start
```

Screenshot from 2019-09-24 16-28-02

### DIFF
--- a/dashboard/src/components/DeploymentForm/DeploymentForm.tsx
+++ b/dashboard/src/components/DeploymentForm/DeploymentForm.tsx
@@ -27,6 +27,7 @@ interface IDeploymentFormProps {
   getChartVersion: (id: string, chartVersion: string) => void;
   getChartValues: (id: string, chartVersion: string) => void;
   namespace: string;
+  enableBasicForm: boolean;
 }
 
 interface IDeploymentFormState {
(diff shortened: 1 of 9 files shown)


andresmgot:
After taking a quick look, I haven't seen anything really simple (nor official) but yes, let's see if @Angelmmiguel knows something else.

dashboard/src/components/DeploymentForm/DeploymentForm.tsx
```diff
@@ -201,6 +190,28 @@ class DeploymentForm extends React.Component<IDeploymentFormProps, IDeploymentFo
   public handleValuesChange = (value: string) => {
     this.setState({ appValues: value, valuesModified: true });
   };
+
+  private renderBasicForm = () => {
```
Angelmmiguel: I'd keep both as separate components instead of methods inside the same component

Angelmmiguel:
> After taking a quick look, I haven't seen anything really simple (nor official) but yes, let's see if @Angelmmiguel knows something else.

Oh I missed these comments! To be honest, I've never used complex feature flags in React applications so I ended with custom and simple solutions. For now, if we don't plan to add much more feature flags, we can keep it as we've.

On the other hand, if we start adding more features flags we should create some kind of HOC to display a flagged component and provide a fallback in case the feature flag is not enabled. We should avoid having too many explicit conditionals inside JSX.

PR #1181, opened 2019-09-24
Deployment form tabs

andresmgot:
Based on the design document #1152

Requires #1180

Add tabs for basic and advanced deployments. I have chosen to use something a bit less bold than what is proposed in #1152. What do you think?

Basic: 

Screenshot from 2019-09-26 10-58-34

Advanced:

Screenshot from 2019-09-26 10-58-45


dashboard/src/components/DeploymentForm/DeploymentForm.test.tsx
```diff
@@ -131,15 +131,38 @@ it("renders a release name by default, relying in Monickers output", () => {
...
-  const versions = [{ id: "foo", attributes: { version: "1.2.3" } }] as IChartVersion[];
-  const wrapper = shallow(
-    <DeploymentForm
-      {...defaultProps}
-      enableBasicForm={true}
-      selected={{ versions, version: versions[0] }}
-    />,
-  );
-  expect(wrapper.find(LoadingWrapper)).not.toExist();
-  expect(wrapper.find(AceEditor)).not.toExist();
```
absoludity: Looks like we should have tests which show that the form is rendered without tabs if the new feature is not enabled?

dashboard/src/components/DeploymentForm/DeploymentForm.tsx
```diff
@@ -146,9 +150,27 @@ class DeploymentForm extends React.Component<IDeploymentFormProps, IDeploymentFo
                   ))}
                 </select>
               </div>
-              {this.props.enableBasicForm ? this.renderBasicForm() : this.renderAdvancedForm()}
+              <div className="margin-t-big" hidden={!this.props.enableBasicForm}>
```
absoludity: I'm a bit confused at this point - it looks like we're no longer using "enableBasicForm" as a feature flag... Shouldn't the snapshot below be unchanged since it's rendered without the feature flag set to true?

I expected this to do something like:
```
{this.props.enableBasicForm ? this.renderBasicAdvancedFormTabs() : this.renderAdvancedForm() }
```
Not sure if I missed it.

andresmgot: it seems it's not really readable so I will rewrite this. Yes, that's how it works but instead of using the conditional here, I am using the `hidden` property. Then, below is where I am rendering the basic or the advanced form depending on the state (that btw, `showBasicForm` will always be `false` if `enableBasicForm` is `false` as well).

andresmgot:
>  Otherwise I like the simplified tabs, but I wonder if we should start using clarity components

The last thing I heard is to not do that for the moment

dashboard/src/components/DeploymentForm/DeploymentForm.tsx
```diff
@@ -146,9 +150,27 @@ class DeploymentForm extends React.Component<IDeploymentFormProps, IDeploymentFo
...
+                  <div
+                    className={`Tabs__Tab ${this.state.showBasicForm ? "" : "Tabs__Tab-active"}`}
+                  >
+                    <button type="button" onClick={this.setBasicForm(false)}>
+                      Advanced
+                    </button>
+                  </div>
+                </div>
+              </div>
+              {this.state.showBasicForm ? this.renderBasicForm() : this.renderAdvancedForm()}
```
absoludity: But this is inside the div which is hidden isn't it? (ie. it appears that (1) the normal advanced form is only rendered within the new tabs which should only appear if the feature is on, and (2) the surrounding div is hidden if the feature is not enabled - but I could have missed something - anyway, +1 for better readability)

andresmgot: ah, I think that you think that the `Tabs` include the form within but those are separate elements. In any case, now it should be more readable.

dashboard/src/components/DeploymentForm/DeploymentForm.scss
```diff
@@ -0,0 +1,40 @@
+.Tabs {
+  display: flex;
+  flex-wrap: wrap;
+  position: relative;
+  margin-bottom: 10px;
+
+  &:after {
+    position: absolute;
+    content: "";
+    border-bottom: 1px solid#C7C9C8;
```
Angelmmiguel: Add a space between `solid` and `#C7C9C8`

dashboard/src/components/DeploymentForm/DeploymentForm.tsx
```diff
@@ -212,6 +217,32 @@ class DeploymentForm extends React.Component<IDeploymentFormProps, IDeploymentFo
...
   };
+
+  private setBasicForm = (enable: boolean) => {
+    // OnClick requires to return a function
+    return () => {
+      this.setState({ showBasicForm: enable });
+    };
+  };
+
+  private renderTabs = () => {
```
Angelmmiguel: Since we're going to use it only in this component, I'm fine with having the tabs defined inside it. However, if we plan to repeat this pattern in other places of the app, we should make it an independent component.

dashboard/src/components/DeploymentForm/DeploymentForm.tsx
Angelmmiguel: I agree with @absoludity. This is a potential accessibility issues because the content of the tabs is fully disconnected from the tabs navigation. These elements should be scoped under a section element and connect the tabs triggers with the content.

You have an example of accessible tabs on https://dev.to/link2twenty/accessibility-first-tabs-ken. For now, I believe you can ignore advanced features like arrow keys navigation. Using the tab key and the enter/space should be enough for the navigation.

andresmgot:
> Regarding adding and maintaining a tabs component - is normal to write and maintain and test stuff like this, rather than use something like https://github.com/reactjs/react-tabs ?

Much better and cleaner using that library, I should have checked that before

PR #1187, opened 2019-09-26
Proxy chartsvc requests

andresmgot:
Use `tiller-proxy` to proxy traffic to the chartsvc (instead of direct usage).

Regarding the `Dashboard` I have moved the `chart` functions to the `shared` folder and use `axiosWithAuth` when needed.

Tested manually.

### DIFF
--- a/cmd/tiller-proxy/main.go
+++ b/cmd/tiller-proxy/main.go
@@ -187,14 +187,15 @@ func main() {
 	))
 
 	// Chartsvc reverse proxy
-	chartsvc := r.PathPrefix("/chartsvc").Subrouter()
+	chartsvcPrefix := "/chartsvc"
+	chartsvcRouter := r.PathPrefix(chartsvcPrefix).Subrouter()
 	// Logos don't require authentication so bypass that step
-	chartsvc.Methods("GET").Path("/v1/assets/{repo}/{id}/logo").Handler(negroni.New(
-		negroni.Wrap(handler.WithoutParams(h.ProxyChartSVC)),
+	chartsvcRouter.Methods("GET").Path("/v1/assets/{repo}/{id}/logo").Handler(negroni.New(
+		negroni.Wrap(http.StripPrefix(chartsvcPrefix, h.ChartsvcProxy)),
 	))
-	chartsvc.Methods("GET").Handler(negroni.New(
+	chartsvcRouter.Methods("GET").Handler(negroni.New(
 		authGate,
-		negroni.Wrap(handler.WithoutParams(h.ProxyChartSVC)),
+		negroni.Wrap(http.StripPrefix(chartsvcPrefix, h.ChartsvcProxy)),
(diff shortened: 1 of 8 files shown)


cmd/tiller-proxy/main.go
```diff
@@ -173,6 +186,17 @@ func main() {
...
+	chartsvc := r.PathPrefix("/chartsvc").Subrouter()
+	// Logos don't require authentication so bypass that step
+	chartsvc.Methods("GET").Path("/v1/assets/{repo}/{id}/logo").Handler(negroni.New(
+		negroni.Wrap(handler.WithoutParams(h.ProxyChartSVC)),
+	))
+	chartsvc.Methods("GET").Handler(negroni.New(
+		authGate,
+		negroni.Wrap(handler.WithoutParams(h.ProxyChartSVC)),
+	))
+
```
absoludity: So, after reading the docs and example for `gorilla/mux` I see that if we treated all traffic equally here, we can replace the above 10 lines with:
```
const chartsvcPrefix = "/chartsvc"
r.PathPrefix(chartsvcPrefix).Handler(http.StripPrefix(chartsvcPrefix, h.ChartsvcProxy)
```
and no longer need to define an actual handler, I think? Less code to test and maintain :)

dashboard/src/components/DeploymentForm/BasicDeploymentForm/BasicDeploymentForm.tsx
```diff
@@ -0,0 +1,9 @@
+import * as React from "react";
```
absoludity: Looks like this slipped in from another branch :)

dashboard/src/actions/charts.ts
```diff
@@ -103,9 +99,13 @@ export function getChartVersion(
...
-    const response = await httpGet(dispatch, url.api.charts.getVersion(id, version));
-    if (response) {
-      dispatch(selectChartVersion(response));
+    try {
+      const chartVersion = await Chart.getChartVersion(id, version);
+      if (chartVersion) {
+        dispatch(selectChartVersion(chartVersion));
+      }
+    } catch (e) {
+      dispatchError(dispatch, e);
```
absoludity: Thanks for adding the error handler here.

dashboard/src/actions/charts.ts
```diff
@@ -77,23 +64,32 @@ export function fetchCharts(
...
+      const versions = await Chart.fetchChartVersions(id);
+      if (versions) {
+        dispatch(receiveChartVersions(versions));
+      }
+      return versions;
+    } catch (e) {
+      dispatchError(dispatch, e);
     }
-    return response;
+    return null;
```
absoludity: Is this required here? The `try` clause always returns (even if `!versions`), so it's just the `catch` clause which doesn't explicitly return?

PR #1190, opened 2019-09-27
Use auth default namespace in namespace reducer

andresmgot:
I'm still getting some 404s since some components don't try to retrieve the default namespace from the `auth` state (only the root container does that):

```
▶ ag 'namespace.current' dashboard/src   
dashboard/src/components/Header/Header.tsx
105:                      <HeaderLink {...link} currentNamespace={namespace.current} />

dashboard/src/containers/AppListContainer/AppListContainer.tsx
18:    namespace: namespace.current,

dashboard/src/containers/ServiceInstanceListContainer/ServiceInstanceListContainer.ts
23:    namespace: namespace.current,

dashboard/src/containers/AppNewContainer/AppNewContainer.tsx
30:    namespace: namespace.current,

dashboard/src/containers/ChartViewContainer/ChartViewContainer.tsx
23:    namespace: namespace.current,

dashboard/src/containers/RoutesContainer/RoutesContainer.tsx
8:  return { namespace: namespace.current || auth.defaultNamespace };

dashboard/src/containers/ServiceClassViewContainer/ServiceClassViewContainer.ts
26:    namespace: namespace.current,
```

So I have changed the default namespace in the `namespace` reducer to follow the same approach than in the `auth` reducer.

PR #1192, opened 2019-09-30
Fix currentNamespace

andresmgot:
Found a couple of issues while doing some QA:

 - After https://github.com/kubeapps/kubeapps/pull/1190, the `namespace` in `Routes.tsx` is always defined (and is pointing to `default` since the token has not been read yet). So, regardless of the content of the token once logged in, it was redirecting me to the `default` namespace. To avoid that I have added the additional check to only redirect to `/apps/ns/{ns}` when accessing the root path (`/`) if the user is already authenticated.
 - After login out the `current` namespace was cleared to the default value (`default`). This caused that after login out and in again the namespace changed from whatever it was to `default`. To avoid this, we should not remove the `current` namespace from the state.

### DIFF
--- a/dashboard/src/reducers/namespace.ts
+++ b/dashboard/src/reducers/namespace.ts
@@ -32,7 +32,8 @@ const namespaceReducer = (
     case getType(actions.namespace.errorNamespaces):
       return { ...state, errorMsg: action.payload.err.message };
     case getType(actions.namespace.clearNamespaces):
-      return { ...initialState };
+      // Clear namespaces info but keep "current" to avoid unexpected redirections
+      return { ...initialState, current: state.current };
     case LOCATION_CHANGE:
       const pathname = action.payload.location.pathname;
       // looks for /ns/:namespace in URL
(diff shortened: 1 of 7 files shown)


dashboard/src/reducers/namespace.ts
```diff
@@ -32,7 +32,8 @@ const namespaceReducer = (
     case getType(actions.namespace.errorNamespaces):
       return { ...state, errorMsg: action.payload.err.message };
     case getType(actions.namespace.clearNamespaces):
-      return { ...initialState };
+      // Clear namespaces info but keep "current" to avoid unexpected redirections
+      return { ...initialState, current: state.current };
```
absoludity: > After login out the current namespace was cleared to the default value (default). This caused that after login out and in again the namespace changed from whatever it was to default.

Isn't that the expected behaviour - that if I log out there is no state, so that when I log back in, I'll be put on the default namespace? I'm not sure what I'm missing on this one, but +1'ing so we can start those release images :)

andresmgot: not really, the state it's better cleaned up when reloading the page.

Let me put an example of the issue.

1.  I got my token for the namespace `user-1`. I login in Kubeapps with that token and go to to my namespace. 
2. I go to the `/catalog` page, which doesn't have the namespace in the route.
3. I click on "logout" and the current namespace is set back to `default`.
4. I login again, and since I haven't changed the namespace in the selector I got again in the `/catalog` page again but this time, the selected namespace is `default`.

PR #1194, opened 2019-09-30
Revamp default namespace setup

andresmgot:
Again I found more issues (mostly related to my changes, sorry). The issue is that `current.namespace` was now always set to `default` because it was read at the very beginning when the token is not yet introduced.

To avoid all the issues I have found with my tests I have added the action `SET_DEFAULT_NAMESPACE` and store it only if the `current.namespace` has never been set.

I have added more tests, hopefully this is the last PR related to the issue :crossed_fingers: 

Sorry for the noise.

### DIFF
--- a/dashboard/src/reducers/namespace.test.ts
+++ b/dashboard/src/reducers/namespace.test.ts
@@ -53,9 +53,9 @@ describe("namespaceReducer", () => {
   });
 
   context("when CLEAR_NAMESPACES", () => {
-    it("returns to the default namespace, maintaining the current namespace", () => {
+    it("returns to the default namespace", () => {
       const clearedState = {
-        current: initialState.current,
+        current: "",
         namespaces: [],
       };
       expect(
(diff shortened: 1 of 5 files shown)


dashboard/src/reducers/namespace.test.ts
```diff
@@ -66,3 +66,31 @@ describe("namespaceReducer", () => {
     });
   });
 });
+
+context("when SET_DEFAULT_NAMESPACE", () => {
+  it("set current namespace if it's empty", () => {
```
absoludity: It's a bit confusing to read that `SET_DEFAULT_NAMESPACE` is actually setting the current namespace - I think this confusion is there because we're mixing concerns. I think what I would do personally is:

1) Have a `NAMESPACE_RECEIVED` action defined in `actions/auth.ts` (with it's corresponding `namespaceReceived()`) as the auth module the subject/cause of receiving the data, and dispatch that where you're currently dispatching the imported `SET_DEFAULT_NAMESPACE/setDefaultNamespace`.
2) Update any reducers which are interested in this - which in this case, is the namespace reducer. The name of the action does not imply what reducers will do with the info, nor restrict who it is for, each module can choose to do what they want (which in this case, will be the namespace reducer updating its own current namespace state only if it's not currently set).

This is one of the reasons for redux actions - separating concerns... all the auth module cares about here is publishing the fact that it received the info (not that it needs to be set as a default or current in a separate module).

The tests here would not change, other than the action type, just the implementation is slightly different.

See what you think.

andresmgot: what I wanted to distinguish is between a manual change of the namespace (changing it in the namespace selector) vs the automatic namespace from the token but I agree with your reasoning.

PR #1195, opened 2019-10-01
Helm chart: Add ability to set static IP addr when using load balancer service type

aldemira:
I'm using metallb in a bare-metal setup, I'd like to assign a static IP address to the frontend when installing kubeapps with helm. This PR enables that.

### DIFF
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -66,7 +66,10 @@ frontend:
   service:
     port: 80
     type: ClusterIP
-    loadBalancerIP:
+    # Set type to LoadBalancer and enable
+    # the parameter below with a value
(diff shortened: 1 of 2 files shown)


chart/kubeapps/values.yaml
```diff
@@ -66,6 +66,7 @@ frontend:
   service:
     port: 80
     type: ClusterIP
+    loadBalancerIP:
```
andresmgot: I prefer not having `null` values here (parameters without a value). I would either remove this line or comment it (adding an additional comment with its meaning).

aldemira: Thanks, I've added a short description and commented out the parameter.

Issue #1182, opened 2019-09-25
Simplify the deployment and upgrade forms


andresmgot:
After the design in #1144, this is the parent issue to track the progress in the tasks specified in that proposal:

- [x] Create a values.schema.json for the desired applications.
  - [x] WordPress
  - [x] Apache
  - [x] PostgreSQL
- [x] Create a test step to ensure that if the `values.schema.json` exists, the default `values.yaml` complies with it.
- [x] Create the feature flag.
- [x] Add tabs for basic/advanced deployment
- [x] Use the file `values.schema.json` to verify the content of the values.yaml to submit.
- [x] Change the current management of the `values.yaml` from text to a [YAML Document](https://eemeli.org/yaml/#documents).
- [x] Design and create the components for the rich view, using the `form` key definition.
  - [x] Username
  - [x] Password
  - [x] Email
  - [x] Enable External Database
  - [x] Enable Hostname
  - [x] Enable TLS
  - [x] Disk size
  - [x] Required Resources
  - [x] Enable Metrics
  - [x] Replicas
- [x] Design and create generic components in case the `form` key is defined with unknown parameters.
  - [x] Text
  ~~- [ ] Choice~~ Not for the moment
  - [x] Boolean
  - [x] Number
  - [x] Slider
- [x] Adapt the DeploymentForm to render the basic form.
- [x] Migrate back changes from the advanced deployment form to the basic.
- [x] Adapt the UpgradeForm to render the form.
~~- [ ] Improve the UpgradeForm to migrate values from one version to other.~~ Continued in #1235
- [x] Developer documentation
~~- [ ] Support for parameters that affect different sections: https://github.com/helm/charts/pull/18078#issuecomment-543212398~~ Continued in #1236
~~- [ ] Support for nested parameters that are not set~~ Continued in #1237

andresmgot:
Closing and moving this task to done since all the elements has been either implemented or migrated to its own issue.

PR #1196, opened 2019-10-01
Read schema and add username to the basic form

andresmgot:
Part 1 of 2

Ref #1182 

With this PR, the dashboard obtains the schema from the chartsvc (requires https://github.com/helm/monocular/pull/645) and pass it to the `DeploymentForm` component to parse it.

As an example, I have hardcoded the basic parameter `username` to show the proposed architecture.

I have defined a basic form parameter as:

```js
{
  name: username, // Name of the basic parameter
  path: account.username, // Path of the parameter in the JSON Object (e.g. {account: {username: "foo"} }
  value: user // Current value of the parameter
}
```

Result:

Screenshot from 2019-10-01 16-11-47

### DIFF
--- a/dashboard/src/actions/repos.test.tsx
+++ b/dashboard/src/actions/repos.test.tsx
@@ -68,7 +68,10 @@ const actionTestCases: ITestCase[] = [
 actionTestCases.forEach(tc => {
   describe(tc.name, () => {
     it("has expected structure", () => {
-      const actionResult = tc.args ? tc.action.call(null, ...tc.args) : tc.action.call(null);
+      const actionResult =
+        tc.args && tc.args.length && typeof tc.args === "object"
+          ? tc.action.call(null, ...tc.args)
+          : tc.action.call(null, tc.args);
       expect(actionResult).toEqual({
         type: getType(tc.action),
         payload: tc.payload,
--- a/dashboard/src/components/DeploymentForm/DeploymentForm.tsx
+++ b/dashboard/src/components/DeploymentForm/DeploymentForm.tsx
@@ -200,7 +200,9 @@ class DeploymentForm extends React.Component<IDeploymentFormProps, IDeploymentFo
 
   public handleChartVersionChange = (e: React.FormEvent<HTMLSelectElement>) => {
     this.props.push(
-      `/apps/ns/${this.props.namespace}/new/${this.props.chartID}/versions/${e.currentTarget.value}`,
+      `/apps/ns/${this.props.namespace}/new/${this.props.chartID}/versions/${
+        e.currentTarget.value
+      }`,
     );
   };
 
@@ -218,7 +220,7 @@ class DeploymentForm extends React.Component<IDeploymentFormProps, IDeploymentFo
           </TabList>
           <TabPanel>
             <BasicDeploymentForm
-              params={Object.values(this.state.basicFormParameters)}
+              params={this.state.basicFormParameters}
               handleBasicFormParamChange={this.handleBasicFormParamChange}
             />
           </TabPanel>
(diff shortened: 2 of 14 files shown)


dashboard/src/actions/repos.test.tsx
```diff
@@ -68,7 +68,10 @@ const actionTestCases: ITestCase[] = [
 actionTestCases.forEach(tc => {
   describe(tc.name, () => {
     it("has expected structure", () => {
-      const actionResult = tc.args ? tc.action.call(null, ...tc.args) : tc.action.call(null);
+      const actionResult =
```
andresmgot: _this code was failing in my tests but it's unrelated to this PR_

dashboard/src/components/DeploymentForm/DeploymentForm.tsx
```diff
@@ -85,14 +89,19 @@ class DeploymentForm extends React.Component<IDeploymentFormProps, IDeploymentFo
...
     if (!this.state.valuesModified) {
       if (version) {
-        this.setState({ appValues: nextProps.selected.values });
+        this.setState({ appValues: nextProps.selected.values || "" });
       }
     }
+
+    if (nextProps.selected.schema) {
+      this.setState({ basicFormParameters: retrieveBasicFormParams(nextProps.selected.schema) });
+    }
```
absoludity: What version of react are we using? This should be done in https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops , as componentWillReceiveProps is being removed.

dashboard/src/components/DeploymentForm/DeploymentForm.tsx
```diff
@@ -210,7 +217,10 @@ class DeploymentForm extends React.Component<IDeploymentFormProps, IDeploymentFo
             <Tab>Advanced</Tab>
           </TabList>
           <TabPanel>
-            <BasicDeploymentForm />
+            <BasicDeploymentForm
+              params={Object.values(this.state.basicFormParameters)}
```
absoludity: This is the conversion from map to array that I'm not sure we need, as above.

dashboard/src/reducers/charts.ts
```diff
@@ -75,6 +78,8 @@ const chartsReducer = (state: IChartState = initialState, action: ChartsAction):
       return { ...state, selected: chartsSelectedReducer(state.selected, action) };
     case getType(actions.charts.selectValues):
       return { ...state, selected: chartsSelectedReducer(state.selected, action) };
+    case getType(actions.charts.selectSchema):
+      return { ...state, selected: chartsSelectedReducer(state.selected, action) };
```
absoludity: I don't know the context here, but the above 10 lines look like they could avoid the redundancy with:
```
case getType(actions.charts.resetChartVersion):
case getType(actions.charts.selectReadme):
case getType(actions.charts.errorReadme):
case getType(actions.charts.errorChart):
case getType(actions.charts.selectValues):
case getType(actions.charts.selectSchema):
  return { ...state, selected: chartsSelectedReducer(state.selected, action) };
```

dashboard/src/shared/schema.ts
```diff
@@ -0,0 +1,17 @@
+import { IBasicFormParam } from "./types";
+
+export function retrieveBasicFormParams(schema: any) {
+  // TBD
```
absoludity: Thanks for leaving this out of the PR to make the diff more manageable. I'll have a quick break before reviewing the next :)

dashboard/src/components/DeploymentForm/DeploymentForm.tsx
andresmgot: We are using `^16.2.0` but yes, we will need to update that hook (we are using it in many places). I will create a separate issue for that.

andresmgot: https://github.com/kubeapps/kubeapps/issues/1200

PR #1197, opened 2019-10-01
Schema helper functions

andresmgot:
Follow up #1196

Actual implementation of the `schema` helper function. There are two main functions:

 - retrieveBasicFormParams: Reads the schema and return the list of basic params.
 - setValues: Reads the give values and modify a certain value.

An issue to mention is that the `yaml` package doesn't support Typescript definitions (yet), the current package has some missing definitions and it's not usable so we need to work with plan javascript.


dashboard/src/shared/schema.ts
```diff
@@ -1,17 +1,110 @@
...
+  if (schema && schema.properties) {
+    const params = lookForFormParams(defaultValues, schema.properties);
+    if (params) {
+      params.forEach(p => (basicFormParameters[p.name] = p));
+    }
+  }
+  return basicFormParameters;
+}
+
+// getDocumentElem returns the YAML element given a root element
```
absoludity: Unless I'm missing something, `rootElem` here is just an object right? In which case, we can possibly remove (and not have to maintain) this function completely in favor of https://lodash.com/docs/#get ?

dashboard/src/shared/schema.ts
```diff
@@ -1,17 +1,110 @@
...
+    return rootElem.get(path);
+  }
+  // If the path contains inner elements, iterate until finding the last one
+  let res = rootElem.get(splittedPath[0]);
+  splittedPath.slice(1, splittedPath.length).forEach(p => (res = res && res.get(p)));
+  return res;
 }
 
+// setValue modifies the current values (text) based on a path
 export function setValue(values: string, path: string, newValue: any) {
```
absoludity: Similarly, https://lodash.com/docs/#set (which also gets around your issue of setting paths which don't exist yet).

It's not clear to my why you need to set the value in a yaml string... why wouldn't you just pass the parsed object (to lodash.set, for example).

dashboard/src/shared/schema.test.ts
```diff
@@ -0,0 +1,261 @@
...
+    },
+  ].forEach(t => {
+    it(t.description, () => {
+      const doc = YAML.parseDocument(t.values);
+      expect(getDocumentElem(doc, t.path)).toEqual(t.result);
+    });
+  });
+});
+
+describe("getDefaultValue", () => {
```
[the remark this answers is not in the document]
absoludity: Never mind, I see the test cases are identical because in the actual test for `getDocumentElem` you're parsing the YAML, where as in the tests for `getDefaultValue` you're passing in the YAML to be parsed.

See below, but I think we don't need `getDocumentElem` (as we can probably use lodash get for this), in which case, I'd rename `getDefaultValue` as `getYAMLValue` if that's what it is doing (taking a yaml string and getting the value at a path).

dashboard/src/shared/schema.ts
```diff
@@ -1,17 +1,110 @@
...
+    }
+    elem.set(itemToSet, newValue);
+  } else {
+    doc.set(path, newValue);
+  }
+  return doc.toString();
+}
+
+// getDefaultValues returns the current value of an object based on YAML text and its path
+export function getDefaultValue(defaultValues: string, path: string): string | undefined {
```
absoludity: As above, this seems to be just a wrapper around `getDocumentElem` that returns the result as a string? It's not clear to me why we'd build the conversion into the function... worth checking if we can just use lodash here (or whether this extra function is needed at all.

Note, it doesn't seem to be related to `defaults` at all?

dashboard/src/shared/schema.ts
```diff
@@ -1,17 +1,110 @@
...
+  const doc = YAML.parseDocument(defaultValues);
+  if (path.includes(".")) {
+    const elem = getDocumentElem(doc, path);
+    // If the path is found, return it as a string
+    return elem && elem.toString();
+  }
+  return path.includes(".") ? getDocumentElem(doc, path).toString() : doc.get(path);
```
absoludity: This condition looks identical to the one above - with an added else? Perhaps a refactory snafu? Also, is `doc.get(path)` a string? Seems you're wanting to return a string above but don't do so in the else clause here (though as above, not clear to me why we'd do that in this function rather than explicitly at the call-site).

dashboard/src/shared/schema.ts
```diff
@@ -1,17 +1,110 @@
...
+}
+
+// lookForFormParams iterates recursively over JSON Schema properties and returns IBasicFormParams
+function lookForFormParams(
```
absoludity: https://lodash.com/docs/#find :)

dashboard/src/shared/schema.test.ts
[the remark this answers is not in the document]
andresmgot: I will rename it as `getValue` since it's actually the opposite of `setValue`

dashboard/src/shared/schema.ts
andresmgot: Yeah, unfortunately it's not. It's a `YAML.ast.Document` from https://eemeli.org/yaml/#documents so we cannot use lodash methods.

dashboard/src/shared/schema.ts
andresmgot: (replied above)

dashboard/src/shared/schema.ts
andresmgot: the `Default` naming is confusing, yes

dashboard/src/shared/schema.ts
andresmgot: ups. yes, I added the `if` above and forgot to remove this.

dashboard/src/shared/schema.ts
```diff
@@ -1,17 +1,110 @@
...
+  defaultValues: string,
+  schemaProperties: {},
+  parentPath?: string,
+  params?: IBasicFormParam[],
+) {
```
absoludity: always create a params array here, and then...

dashboard/src/shared/schema.ts
```diff
@@ -1,17 +1,110 @@
...
+    // If the property has the key "form", it's a basic parameter
+    if (schemaProperties[propertyKey].form) {
+      const newParam = {
+        // The name of the param is identified by the form key
+        name: schemaProperties[propertyKey].form,
+        path: itemPath,
+        // Use the default value either from the JSON schema or the default values
+        value: getDefaultValue(defaultValues, itemPath) || schemaProperties[propertyKey].default,
+      };
+      params = params ? params.concat(newParam) : [newParam];
```
absoludity: only ever `params.concat(newParam)` here, then...

dashboard/src/shared/schema.ts
```diff
@@ -1,17 +1,110 @@
...
+    }
+    // If the property is an object, iterate recursively
+    if (schemaProperties[propertyKey].type === "object") {
+      params = lookForFormParams(
```
absoludity: `params.concat(lookForFormParams(...))` recursively here (but without passing in the `params` arg, obviously).

Or why do you need to rely on both a return value and an array reference as a function arg?

dashboard/src/shared/schema.ts
andresmgot: oh, nice catch, I understood that differently (I thought that the `*In` methods only worked with collections/arrays) but it's true that we can use that. Thanks!

The only caveat I have found is that `setIn` returns an error if the object doesn't define the parent element:

```
> d.toString()
'a: b\n'
> d.setIn(["b", "c"], "foo")
Error: Expected YAML collection at b. Remaining path: c
```

In any case, that's the same limitation that we have now.

dashboard/src/shared/schema.ts
andresmgot: took me a while to get my head around that one :sweat_smile: 

You are right, the `params` argument is not needed.

andresmgot:
Apart from your feedback @absoludity I have added the types for `yaml` and `json-schema` so the code is easier to read and properly typed. To avoid the issue that `getIn` and `setIn` are not defined in the Typescript package, I had to convert the document (`YAML.ast.Document`) to `any`. Let me know what you think.

dashboard/src/shared/schema.ts
```diff
@@ -1,16 +1,74 @@
...
+// It uses the raw yaml to setup default values.
+// It returns a key:value map for easier handling.
+export function retrieveBasicFormParams(
+  defaultValues: string,
+  schema: jsonSchema.JSONSchema4,
+): { [key: string]: IBasicFormParam } {
+  if (schema && schema.properties) {
+    return lookForFormParams(defaultValues, schema.properties);
+  }
+  return {};
```
absoludity: This function looks like it doesn't do much now - other than verify the schema has properties... why not rename `lookForFormParams` to `retrieveBasicFormParams` and have call-sites use that (if they have properties)? Given that `lookForPormParams` no longer has the array arg (which may have been awkward for call-sites to call with an empty array), it now looks exportable?

dashboard/src/shared/schema.ts
```diff
@@ -1,16 +1,74 @@
...
+  (doc as any).setIn(splittedPath, newValue);
+  return doc.toString();
+}
+
+// getValue returns the current value of an object based on YAML text and its path
+export function getValue(values: string, path: string, defaultValue?: any) {
+  const doc = YAML.parseDocument(values);
+  const splittedPath = path.split(".");
+  return (doc as any).getIn(splittedPath) || defaultValue;
+}
```
absoludity: Great - `getValue` and `setValue` are much simpler now. I'm still not sure why we pass string values into these, rather than passing the actual document. I expected that the form state would be the document, which gets updated with data by the user... converting that to a string and back on each change seems superfluous, but maybe there is context I'm missing?

dashboard/src/shared/schema.ts
andresmgot: yes, we can do that, I wanted to split it for being a bit more readable but probably there is not much difference.

dashboard/src/shared/schema.ts
andresmgot: My reasoning is that the the `values.yaml` is received as a string and it's always treated as a normal string in the `Form` component if it doesn't have support for the basic form.

My idea is to have this abstraction layer so the `Form` always work with the string `values.yaml` regardless of the implementation of this library. We are using yaml `Document`s but we could be using regexps or `lodash` to modify those values. That way, the `Form` component doesn't depend on the implementation details here.

PR #1198, opened 2019-10-01
Update chart

andresmgot:
Adapt chart to the new version


andresmgot:
_Merging this to trigger a release_

PR #1199, opened 2019-10-02
Switch to use oauth2_proxy for oidc/oauth2 authentication

absoludity:
This supports authenticating with Kubeapps on vanilla Kubernetes and also managed GKE kubernetes (which uses the OAuth2 access_token as a bearer instead of the id_token).

I've done a gazillion deployments today verifying the combinations of GKE + oidc, GKE + oauth2, local Kind cluster + oidc, kind + oauth2. Hopefully the updated docs explain the cases - let me know.

I'm not sure if we should land this - it will be a breaking change to someone who currently expects the chart options for gatekeeper, but not difficult to update. WDYT?

Also, note that I've removed the whitelist that we were using with gatekeeper - this actually secures the k8s API, as Kubeapps never needs to see the actual credential necessarily (for example, with the GKE setup, kubeapps never sees the access_key at all - it never leaves the cluster... we could update to do the same for the oidc use-case - the frontend would never need to see the id_token, but just has the cookie (which has both the id and access token encrypted)).


docs/user/using-an-OIDC-provider.md
```diff
@@ -19,11 +20,12 @@ For Kubeapps to use an Identity Provider it's necessary to configure at least th
...
-- OIDC Issuer URL (or discovery URL): URL of the OIDC issuer.
+- Provider name (which can be oidc, in which case the OIDC Issuer URL is also required).
+- Cookie secret: a 16, 24 or 32 byte base64 encoded seed string used to encrypt sensitive data (eg. `echo "not-a-good-secret" | base64`).
 
 **Note**: Depending on the configuration of the Identity Provider more parameters may be needed.
 
-In the next sections we will explain how you can find the parameters above for some of the identity providers tested. If you have configured your cluster to use an Identity Provider you will already know some of these parameters.
+Kubeapps uses [OAuth2 Proxy](https://github.com/pusher/oauth2_proxy) to handle the OAuth2/OpenIDConnect authentication. The following sections explain how you can find the parameters above for some of the identity providers tested. If you have configured your cluster to use an Identity Provider you will already know some of these parameters. More detailed information can be found on the [OAuth2 Proxy Auth configuration page](https://pusher.github.io/oauth2_proxy/auth-configuration).
 
 ### Keycloak
```
andresmgot: Are these sections still relevant? AFAIK the OIDC Issuer URL is not a needed parameter now.

absoludity: Yep, the `-oidc-issuer-url` is needed if the `oicd` provider is used - see the example I added below (it's in the additional flags, because it's not mandatory unless you're using the `oidc` provider).

migmartri:
I thought that oauth2-proxy did not work well with websockets?

docs/user/using-an-OIDC-provider.md
```diff
@@ -59,39 +61,77 @@ In the case of Google we can use an OAuth 2.0 client ID. You can find more infor
...
+  --namespace kubeapps --name kubeapps \
+  --set authProxy.enabled=true \
+  --set authProxy.provider=oidc \
+  --set authProxy.clientID=my-client-id.apps.googleusercontent.com \
+  --set authProxy.clientSecret=my-client-secret \
+  --set authProxy.cookieSecret=$(echo "not-good-secret" | base64) \
+  --set authProxy.additionalFlags="{-cookie-secure=false,-oidc-issuer-url=https://accounts.google.com}" \
+```
+
+**Example 2: Using Google OAuth2**
```
migmartri: I find this example confusing, I mean, why would I use Google Oauth2 in the general sense? I know that for GKE it makes sense to do it like that but just in that case isn't it? 

I'd have just the generic OIDC example + the new GKE one using oauth2.

absoludity:
> I thought that oauth2-proxy did not work well with websockets?

AFAIK, it didn't work at all prior to March this year, when it was fixed by merging the websockets support from the openshift fork: https://github.com/pusher/oauth2_proxy/pull/92

I've just tested and there is one small issue on our end (in dashboard), in that even though we're logged in via oauth2-proxy (so we send the auth cookie), we're additionally setting the `wsProtocol` in the client which means oauth2-proxy appends to these, which results in a 400. If I remove these from the upgrade request from the client, then oauth2-proxy adds them and the request is upgraded as expected.

docs/user/using-an-OIDC-provider.md
absoludity: > I find this example confusing, I mean, why would I use Google Oauth2 in the general sense? I know that for GKE it makes sense to do it like that but just in that case isn't it?

Yes it's confusing, but no it makes sense to use the 'google' provider generally, it's the heading I've given that is incorrect - it's really using `oauth2-proxy`'s custom Google OIDC provider. The benefit of using the `google` provider is that it provides a richer UI experience to users (as I tried to explain below in the GKE example). One aspect of that only affects you when you're requesting extra scopes (which generally would only be on GKE)[1], another is the default button shown by `oauth2-proxy` ("Login with Google" rather than "Login with OpenID Connect") - but that is configurable too.

A few of the `oauth2-proxy` supported providers are actually using OpenID Connect and setting the IDToken in the session state. These are: gitlab, google and logingov.

Let me re-word these examples to see if I can make it much clearer.

Thanks!

[1] This might change in the future - I don't see why Google wouldn't want the same UI experience for generic OIDC logins, or it could even be a quirk of `oauth2_proxy`'s generic oidc implementation - not sure.

absoludity: Please take another look :)

absoludity:
> I did verify that I could deploy a chart, but didn't explicitly check the websocket connections. oauth2_proxy defaults to proxying them (and the creds will be set in the same way), but I'll double check tomorrow two [sic].

As above, when I checked them they are 400ing because the frontend adds the wsProtocol for the authentication first, then `oauth2-proxy` does the same (appending, I guess?) resulting in a bad request. If I remove the `Sec-WebSocket-Protocol` header from the request client request, it succeeds. I have not tested this yet in the frontend, just via cURL, so will need to do that before considering landing this..

docs/user/using-an-OIDC-provider.md
migmartri: > "Login with Google"

I thought that we were redirecting by default to the IdP?

migmartri:
> > I thought that oauth2-proxy did not work well with websockets?
> 
> AFAIK, it didn't work at all prior to March this year, when it was fixed by merging the websockets support from the openshift fork: [pusher/oauth2_proxy#92](https://github.com/pusher/oauth2_proxy/pull/92)
> 

Nice, great to hear :)

docs/user/using-an-OIDC-provider.md
```diff
@@ -59,39 +61,80 @@ In the case of Google we can use an OAuth 2.0 client ID. You can find more infor
...
 The next sections explain how you can deploy this proxy either using the Kubeapps chart or manually.
 
 ### Using the chart
 
-Kubeapps chart allows you to automatically deploy the proxy for you as a sidecar container if you specify the necessary flags. In a nutshell you need to enable the feature and set the client ID, secret and the IdP URL. This is an example for using Google as IdP, modify the flags below to adapt them:
+Kubeapps chart allows you to automatically deploy the proxy for you as a sidecar container if you specify the necessary flags. In a nutshell you need to enable the feature and set the client ID, secret and the IdP URL. The following examples use Google as the Identity Provider, modify the flags below to adapt them:
+
+**Example 1: Using the OIDC provider**
+
+This example uses `oauth2-proxy`'s generic OIDC provider with Google, but is applicable to any OIDC provider. Note that the issuer url is passed as an additional flag here, together with an option to enable the cookie being set over an insecure connection for local development only:
```
migmartri: Much clearer, 

I'd add to "but is applicable to any OIDC provider"

... such as Keycloak, Dex,  Okta, Azure Active Directory, ...

docs/user/using-an-OIDC-provider.md
```diff
@@ -59,39 +61,80 @@ In the case of Google we can use an OAuth 2.0 client ID. You can find more infor
...
+  --namespace kubeapps --name kubeapps \
+  --set authProxy.enabled=true \
+  --set authProxy.provider=oidc \
+  --set authProxy.clientID=my-client-id.apps.googleusercontent.com \
+  --set authProxy.clientSecret=my-client-secret \
+  --set authProxy.cookieSecret=$(echo "not-good-secret" | base64) \
+  --set authProxy.additionalFlags="{-cookie-secure=false,-oidc-issuer-url=https://accounts.google.com}" \
+```
+
+**Example 2: Using a custom provider**
```
migmartri: I do not understand what a custom provider is nor why I should care about them. Are these pre-configured OIDC Idp and that's why you do not need scopes or the `oidc-issuer-urls`? Where can I find more information about them? 

When you say "**Example 2: Using a custom provider** in the section header, are we still talking about  complaint OIDC Id provider?

docs/user/using-an-OIDC-provider.md
```diff
@@ -59,39 +61,80 @@ In the case of Google we can use an OAuth 2.0 client ID. You can find more infor
...
+
+**Example 3: Authentication for Kubeapps on a GKE cluster**
+
+Google Kubernetes Engine does not allow an OIDC IDToken to be used to authenticate requests to the managed API server, instead requiring the standard OAuth2 access token.
+For this reason, when deploying Kubeapps on GKE we need to ensure that
+
+* The scopes required by the user to interact with cloud platform are included, and
+* The Kubeapps frontend uses the OAuth2 `access_key` as the bearer token when communicating with the managed Kubernetes API
+
+Note that using the custom `google` provider here enables google to prompt the user for consent for the specific permissions requested in the scopes below, in a user-friendly way. You can also use the `oidc` provider but in this case the user is not prompted for the extra contsent:
```
migmartri: Typo: `contsent`

>  You can also use the `oidc` provider but in this case the user is not prompted for the extra contsent:

I couldn't follow.

docs/user/using-an-OIDC-provider.md
absoludity: > I do not understand what a custom provider is nor why I should care about them.

I've updated the example title to make it clear that they are oauth2-proxy providers (it is also mentioned in the next sentence), and I had already tried to highlight below why you should care in the following para?

> Are these pre-configured OIDC Idp and that's why you do not need scopes or the oidc-issuer-urls?

Yes. That is, oauth2-proxy provides a bunch of *oauth2* providers, a few of which are actually using openid connect so can be used here if they provide better integration (eg., example 3).

> Where can I find more information about them?

About the providers themselves? I've updated with a link to the oauth2 providers, but it doesn't include details about which providers are using openid connect.

docs/user/using-an-OIDC-provider.md
absoludity: Thanks, fixed the typo. The extra consent is:

google-signin-oauth2

If you use the `oidc` provider, you don't get this (which is odd, but verified multiple times).

absoludity:
> Nice! Thanks for applying the changes
> 
> Just one tibdit
> 
> > For this guide we assume that you have a Kubernetes cluster that is properly configured to use an Identity Provider (IdP) to handle the authorization of your cluster
> 
> " to handle the authorization of your cluster" > to handle the **authentication** of your cluster?

Updated. I was actually wondering if it's safe for me to create a rando google account and setup a provider authorising `lvh.me` only, and actually include that in the README (so the README can then clearly document how to run a kind cluster with kubeapps installed using that IdPn).

PR #1201, opened 2019-10-02
Fix missing namespace

andresmgot:
I found a 404 when the user is already logged in and it loads for the first time the application in a route that doesn't contain the namespace (e.g. `/catalog`). In that case, the `authenticate` action is not executed (because the user is already logged) and none of the components rendered try to set that current namespace.

I have added that functionality to the `NamespaceSelector` component (which is loaded in the Header). There, along with fetching the existing namespaces, the default namespace is now set (from the token).

### DIFF
--- a/dashboard/src/actions/namespace.test.tsx
+++ b/dashboard/src/actions/namespace.test.tsx
@@ -1,8 +1,16 @@
 import configureMockStore from "redux-mock-store";
 import thunk from "redux-thunk";
 import { getType } from "typesafe-actions";
+import { Auth } from "../shared/Auth";
 import Namespace from "../shared/Namespace";
-import { errorNamespaces, fetchNamespaces, receiveNamespaces, setNamespace } from "./namespace";
+import {
+  errorNamespaces,
+  fetchNamespaces,
+  namespaceReceived,
+  receiveNamespaces,
+  setDefaultNamespace,
+  setNamespace,
+} from "./namespace";
 
 const mockStore = configureMockStore([thunk]);
 
@@ -73,6 +81,25 @@ describe("fetchNamespaces", () => {
 
     await store.dispatch(fetchNamespaces());
 
+    expect(store.getActions()).toEqual(expectedActions);
+  });
+});
+
+// Async action creators
+describe("setDefaultNamespace", () => {
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+  it("dispatches namespaceReceived based on the received token", async () => {
+    Auth.getAuthToken = jest.fn(() => "token");
+    Auth.defaultNamespaceFromToken = jest.fn(() => "my-ns");
+    const expectedActions = [
+      {
+        type: getType(namespaceReceived),
+        payload: "my-ns",
+      },
+    ];
+    await store.dispatch(setDefaultNamespace());
     expect(store.getActions()).toEqual(expectedActions);
   });
 });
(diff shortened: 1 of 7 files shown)


absoludity:
Similar to a diff last week (that I didn't get back to, sorry), I'm not so keen to patch up this problem - I'd rather find out why (perhaps you know already?), but...

> In that case, the authenticate action is not executed (because the user is already logged) and none of the components rendered try to set that current namespace.

If we haven't retriggered authentication because we've determined that the user is already authenticated, then the initial namespace state should already include the default namespace (see `reducers/namespace.ts`) of either `default` or from the token. This is it's initial state, set when the application loads, so I'm keen to understand how you're ending up in a state where it's not set?

I tried to reproduce the issue running an image built from master, but I get the default namespace from the token, as expected, so I must be missing something. The steps I'm trying are:

1) Go to localhost:8080 and enter the token (in my case, I'm just using the tiller service account one in my local env)
2) Navigate to Catalog,
3) hit refresh

I see the catalog and the namespace switcher has `kube-system` selected (from the tiller JWT token) as expected? Let me know what you think - if you want to focus on the form stuff, I'm happy to look at this tomorrow.

andresmgot:
You are seeing that because the `NamespaceSelector` defaults to show the `defaultNamespace` if `current` is empty: 

https://github.com/kubeapps/kubeapps/blob/master/dashboard/src/containers/HeaderContainer/HeaderContainer.ts#L16

https://github.com/kubeapps/kubeapps/blob/master/dashboard/src/components/Header/NamespaceSelector.tsx#L32

The thing is that even if the selector shows a namespace, then if you try to deploy an application you will see the 404. To reproduce it:

1. Go to localhost:8080 and enter the token (in my case, I'm just using the tiller service account one in my local env)
2. Navigate to Catalog,
3. hit refresh
4. Select an application
5. Click deploy

That will take you to something like `http://kubeapps.local/#/apps/ns//new/stable/aerospike/versions/0.2.9` where the namespace is missing. This is because, as other components, the `ChartView` is only getting the namespace from the `current` state:

https://github.com/kubeapps/kubeapps/blob/master/dashboard/src/containers/ChartViewContainer/ChartViewContainer.tsx#L23

absoludity:
> You are seeing that because the `NamespaceSelector` defaults to show the `defaultNamespace` if `current` is empty:

Yes - I thought that was the issue though, that at that point you were seeing a 404 (as per my steps to reproduce). Thanks for the specific steps, I can now reproduce it. So...

> The thing is that even if the selector shows a namespace, then if you try to deploy an application you will see the 404. To reproduce it:
> 
>     1. Go to localhost:8080 and enter the token (in my case, I'm just using the tiller service account one in my local env)
> 
>     2. Navigate to Catalog,
> 
>     3. hit refresh
> 
>     4. Select an application
> 
>     5. Click deploy
> 
> 
> That will take you to something like `http://kubeapps.local/#/apps/ns//new/stable/aerospike/versions/0.2.9` where the namespace is missing. This is because, as other components, the `ChartView` is only getting the namespace from the `current` state:
> 
> https://github.com/kubeapps/kubeapps/blob/master/dashboard/src/containers/ChartViewContainer/ChartViewContainer.tsx#L23

Right - but this should be OK, in that the current namespace should really *never* be empty that was an issue which was fixed, but then the change reverted in #1194.

All the cases I can think of:
* When you login it should be set from the default namespace,
* When you are already authd and reload a page it should be set from the url (if present), otherwise defaulting to `getNamespaceFromToken`,
* When you switch namespaces (and the location changes) it should update
* When you logout it should be reset to the initialState of the namespace reducer.

Can you think of any more? I've got a PR that would revert #1194 and provide a different solution there that would also not result in the issue you're seeing here (by ensuring the currentNamespace is never empty, as intended), but have a typescript issue I can't yet resolve :/

andresmgot:
Abandoning this PR in favor of #1205

PR #1202, opened 2019-10-03
Add param types: password, email and string

andresmgot:
Ref: #1182 

I have added the generic parameter type `string`. With that we can have `username`, `password`, `email` or any other:

Also, I have added support for specifying a custom title or description in a parameter and I have moved the `BasicDeploymentForm` to its own folder since it's growing.

Screenshot from 2019-10-03 12-50-38


andresmgot:
> Did you git mv the files? Wonder why github doesn't see them as renamed and just show the diffs... if this is normal, it'd be great to next time do the mechanical move in one PR (as it doesn't really need checking) so just the actual changes can be reviewed.

The "issue" is that in this PR the moved file are very different from the source, that's why it's not shown as a movement. In any case, I took a mental note in case this is something that can be mitigated in the future.

PR #1205, opened 2019-10-04
Set current namespace when checking authz

absoludity:
This diff is an alternative to #1201, which I think is caused by #1194. So instead, this PR reverts #1194 (well, most of it - there is one small change we want to keep) and provides an alternative.

So the issue mentioned on #1194 was that:

> Again I found more issues (mostly related to my changes, sorry). The issue is that current.namespace was now always set to default because it was read at the very beginning when the token is not yet introduced.

I *think* this meant that if you are not authenticated, we were setting the `current` namespace to `default` when the initialState was created for the reducer, and it was never overwritten from that point. The solution in #1194 was both to not set it in the initialState and add an action to update it when not set, but I think this is not the correct solution (and leads to the current issue mentioned in #1201).

So this PR provides an alternative: that we explicitly set the current namespace when a user authenticates (`setAuthenticated` is only ever called at that one point). As mentioned on #1201, I don't think it should ever be empty and the only cases where we set it are:

* When you login it should be set from the default namespace,
* When you are already authd and reload a page it should be set from the url (if present), otherwise defaulting to getNamespaceFromToken,
* When you switch namespaces (and the location changes) it should update
* When you logout it should be reset to the initialState of the namespace reducer.

Let me know what you think.


dashboard/src/reducers/namespace.ts
```diff
@@ -31,16 +34,19 @@ const namespaceReducer = (
...
-      const currentNamespace = state.current === "" ? action.payload : state.current;
-      return { ...initialState, current: currentNamespace };
     case LOCATION_CHANGE:
       const pathname = action.payload.location.pathname;
       // looks for /ns/:namespace in URL
       const matches = pathname.match(/\/ns\/([^/]*)/);
       if (matches) {
         return { ...state, current: matches[1] };
       }
+    case getType(actions.auth.setAuthenticated):
```
andresmgot: regarding your Typescript error. The problem is that `LOCATION_CHANGE` doesn't have a `break;` at the end of the case so this one is also executed after the previous one if it matches. That's why you are seeing the error, the `payload` here can be a location change.

To fix it, either add a `break` in the `case LOCATION_CHANGE` or move this case above (but you also need to add a `break;`)

absoludity: Excellent, thanks.

absoludity:
> This also works for me. So far I don't see any issue, let's see after testing it.

Great, thanks. QA'd locally and verified that:

 * When I first authenticate the default namespace is updated from the token (ie. that it fixes the issue which the reverted #1194 solved in a different way), and
 * If I click on `catalog`, then reload the page, the default namespace is set immediately from the token as the reducers initial state (ie. fixing what #1201 wanted to fix).

PR #1206, opened 2019-10-04
Add slider component for adding the disk size

andresmgot:
Ref #1182 

This PR adds a slider to select the disk size in the simplified form.

It's possible to modify it either dragging the slider or adding a custom number in the input field close to it.

Note that apart than the component `react-compound-slider` it's necessary to add some boilerplate code to make the slider work so I have created a wrapper of that component with the code avaliable [here](https://sghall.github.io/react-compound-slider/#/slider-demos/horizontal) to expose a simplified `Slider` component. That boilerplate code is not covered by unit tests and doesn't require a thoughtful review.

Screenshot from 2019-10-04 13-19-56

### DIFF
--- a/dashboard/src/components/DeploymentForm/BasicDeploymentForm/DiskSizeParam.test.tsx
+++ b/dashboard/src/components/DeploymentForm/BasicDeploymentForm/DiskSizeParam.test.tsx
@@ -52,7 +52,7 @@ it("changes the value of the param when the slider changes", () => {
   ]);
 });
 
-it("changes only the value of the state when the slider is being updated", () => {
+it("updates state but does not change param value during slider update (only when dropped in a point)", () => {
   const handleBasicFormParamChange = jest.fn();
   const wrapper = shallow(
     <DiskSizeParam {...defaultProps} handleBasicFormParamChange={handleBasicFormParamChange} />,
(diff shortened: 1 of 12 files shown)


dashboard/src/components/DeploymentForm/BasicDeploymentForm/DiskSizeParam.test.tsx
```diff
@@ -0,0 +1,118 @@
...
+  (slider.prop("onChange") as (values: number[]) => void)([20]);
+
+  expect(param.value).toBe("20Gi");
+  expect(handleBasicFormParamChange.mock.calls[0]).toEqual([
+    "diskSize",
+    { value: "20Gi", type: "string", path: "disk" },
+  ]);
+});
+
+it("changes only the value of the state when the slider is being updated", () => {
```
absoludity: This confused me until I read the actual test details... but I can't see an obviously better test title. Perhaps: "updates state but does not change param value during slider update" .

I think was confused me was "changes only the value of the state", because it seemed odd that we'd be updating the component state but not updating handling the form param change. After reading below, I see it's because the component encapsulates both the slider and the text input - we want to update both state and handle the change for the input, but only trigger the change for the slider when it is "dropped", which makes sense.

dashboard/src/components/DeploymentForm/BasicDeploymentForm/DiskSizeParam.test.tsx
```diff
@@ -0,0 +1,118 @@
...
+  it("ignores values in the input that are not digits", () => {
+    const valueChange = jest.fn();
+    const handleBasicFormParamChange = jest.fn(() => valueChange);
+    const wrapper = shallow(
+      <DiskSizeParam {...defaultProps} handleBasicFormParamChange={handleBasicFormParamChange} />,
+    );
+    expect(wrapper.state("Gi")).toBe(10);
+
+    const input = wrapper.find("input#disk");
+    const event = { currentTarget: { value: "foo20*#@$" } } as React.FormEvent<HTMLInputElement>;
```
absoludity: Hmm... might be some corner cases... see below.

dashboard/src/components/DeploymentForm/BasicDeploymentForm/DiskSizeParam.tsx
```diff
@@ -0,0 +1,93 @@
...
+  ) => (e: React.FormEvent<HTMLInputElement>) => void;
+}
+
+export interface IDiskSizeParamState {
+  Gi: number;
+}
+
+function toNumber(value: string) {
+  // Force to return a Number from a string removing any character that is not a digit
+  return Number(value.replace(/[^\d]/g, ""));
```
absoludity: So if I enter "125.5"? Currently will this set the state to 1255 Gi?

andresmgot: True, I thought that decimals were not allowed (but they are according to https://github.com/kubernetes/community/blob/master/contributors/design-proposals/scheduling/resources.md#resource-quantities) so I will add a handle for that.

andresmgot: Caveat: Since the minimal step in the slide is 1 unit, copying `125.5` will result in `126` since the slider round it to the closest int.

PR #1210, opened 2019-10-08
Basic form improvements

andresmgot:
This PR fixes some small issues:

 - Thanks to the fix of https://github.com/eemeli/yaml/issues/125 we can know parse the values.yaml as a YAML Document without major changes. The new version 0.7.1 includes that fix.
 - In case the user changes the values in the "Advanced" form, the YAML is parsed and the basic parameters are recalculated to modify the basic form if needed.
 - We only show the basic/advanced tabs if the application has some basic parameters defined.

### DIFF
--- a/dashboard/src/components/DeploymentForm/DeploymentForm.tsx
+++ b/dashboard/src/components/DeploymentForm/DeploymentForm.tsx
@@ -221,11 +221,15 @@ class DeploymentForm extends React.Component<IDeploymentFormProps, IDeploymentFo
 
   public handleValuesChange = (value: string) => {
     this.setState({ appValues: value, valuesModified: true });
-    if (this.shouldRenderBasicForm()) {
-      this.setState({
-        basicFormParameters: retrieveBasicFormParams(value, this.props.selected.schema),
-      });
-    }
+  };
+
+  private refreshBasicParameters = () => {
+    this.setState({
+      basicFormParameters: retrieveBasicFormParams(
+        this.state.appValues,
+        this.props.selected.schema,
+      ),
+    });
   };
 
   private renderTabs = () => {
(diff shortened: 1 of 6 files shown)


dashboard/src/components/DeploymentForm/DeploymentForm.tsx
```diff
@@ -221,6 +221,11 @@ class DeploymentForm extends React.Component<IDeploymentFormProps, IDeploymentFo
 
   public handleValuesChange = (value: string) => {
     this.setState({ appValues: value, valuesModified: true });
+    if (this.shouldRenderBasicForm()) {
+      this.setState({
+        basicFormParameters: retrieveBasicFormParams(value, this.props.selected.schema),
+      });
+    }
```
absoludity: Looks like this will mean the basic form params are updated on every keystroke into the text box? Might be fine as is, I was just wondering if it was necessary (ie. could we just do this once when switching to the basic form).

Also, I've not checked, but assuming this doesn't create a loop (ie. if I update the basic form, it updates the values which triggers this method, which re-evaluates the basic parameters). From memory you're using the string value as the source of truth, so it may do (not quite a loop, but perhaps not wanted).

andresmgot: > Looks like this will mean the basic form params are updated on every keystroke into the text box? Might be fine as is, I was just wondering if it was necessary (ie. could we just do this once when switching to the basic form).

Well, yes, it may be an issue if the values is big, I can move this logic whenever the `Basic` tab is clicked.

> Also, I've not checked, but assuming this doesn't create a loop (ie. if I update the basic form, it updates the values which triggers this method, which re-evaluates the basic parameters). From memory you're using the string value as the source of truth, so it may do (not quite a loop, but perhaps not wanted).

Good point, there was indeed an unnecessary step.

Issue #1173, opened 2019-09-23
Enable third party app to trigger chart deployment UI


absoludity:
Third party apps that control chart repositories, like Harbor, should be able to integrate with a Kubeapps installation so that you can click to launch the artifact from the third party app, via the Kubeapps deployment UI.

Whether this is a POST request with the repo and chart details, redirecting the user to add the repo and install the chart (or just install the chart without adding the repo?), the integration should be simple to add and smooth to use, within the configured RBAC (for app repositories, for example).

Issue #1178, opened 2019-09-24
[proposal] Replace AppRepository CRD


andresmgot:
Right now, the usage of an AppRepository CRD have several disadvantages:

 - It requires admin privileges to be installed. This requires intervention from a cluster admin to be used.
 - Installing a CRD and custom resources with Helm is not trivial. It requires either installing that using the `crd-hook` (gone in Helm 3) or move them to a different folder.
 - The user require special RBAC permissions to read/write these resources (that are not included in the default roles). This has been mitigated in the dashboard but it's still necessary for certain operations.
 - It can cause race conditions (like https://github.com/kubeapps/kubeapps/issues/819) where Helm tries to install custom resources when the CRD is not yet ready.
 - If someone accidentally deletes the CRD all the custom resources in different possible Kubeapps installations will be gone.

If we place that information as Kubernetes `secrets` instead we should be able to get rid of those issues.

 - Any user can install Kubeapps in their namespace without any special requirements.
 - Since we will be working with secrets, we don't need the complexity of managing helm hooks.
 - We get rid of race conditions like above and accidental deletions.

This is the same approach that Helm is using in Helm 3 to store the information about the existing releases (instead of using an AppCRD).

Apart than changing the chart and required roles, to achieve this, we would need to modify the apprepository controller to `watch` secrets in the Kubeapps namespace with certain label selector (e.g. `kubeapps=repository`).

andresmgot:
cc/ @absoludity

absoludity:
> Right now, the usage of an AppRepository CRD have several disadvantages:

Just looking one-by-one:
 
>     * It requires admin privileges to be installed. This requires intervention from a cluster admin to be used.

It requires a cluster admin to be installed, not so much used, but yep, -1
 
>     * Installing a CRD and custom resources with Helm is not trivial. It requires either installing that using the `crd-hook` (gone in Helm 3) or move them to a different folder.

I don't think this is still true - at least, the separate directory used with Helm 3 seems like a good solution to ensure those resources are installed (and ready?). 0

>     * The user require special RBAC permissions to read/write these resources (that are not included in the default roles). This has been mitigated in the dashboard but it's still necessary for certain operations.

[...]

PR #1213, opened 2019-10-10
Add initial design doc for third party adding repos

absoludity:
Ref: #1173 

cc @migmartri and @prydonius in case you have time and are interested.


docs/architecture/design-proposals/third-party-add-repository.md
```diff
@@ -0,0 +1,96 @@
...
+
+Enabling these projects to integrate with Kubeapps enables users to move seamlessly from uploading a new app to a repository through to testing the deployment of that same app in their configured Kubeapps installation. Or from preparing a repository with specific apps to enabling users in Kubeapps to access those apps.
+
+## Goals and non goals
+
+* Enable 3rd party sites to provide an "Add repository to Kubeapps" action which results in the Kubeapps installation scanning and including the repository in its catalog.
+* As a consequence, the 3rd party can then link to directly deploy a chart from a repository in the Kubeapps UI.
+
+## Assumptions
+ - The same OpenID Connect identity provider must be used for both the 3rd party (such as Harbor) and Kubeapps (and by implication, the Kubernetes cluster).
```
andresmgot: AFAIK, the default scenario is to have Kubeapps/Harbor configured with username and password or the token, should we support this scenario?

docs/architecture/design-proposals/third-party-add-repository.md
```diff
@@ -0,0 +1,96 @@
...
+   - Method: POST
+   - Data
+   ```
+   syntax = "proto3";
+   message CreateAppRepositoryRequest {
+     message AppRepository {
+       string name = 1;
+       string url = 2;
+     }
+     // Later add credentials
```
andresmgot: we should probably design the scenario in which the repository require authentication (which is the default). In that case, it's necessary to specify those credentials as a secret. Who should create those secrets, the backend as part of this request or the third-party with a different request?

docs/architecture/design-proposals/third-party-add-repository.md
absoludity: AFAIUI, Harbor only has username/password (called "database"), LDAP or OpenID Connect (see [auth section of the user guide](https://github.com/goharbor/harbor/blob/master/docs/user_guide.md#managing-authentication)).

I'm keen to limit this initial integration to OpenID Connect for a couple of reasons:
 * I'm actually keen for us (Kubeapps) to move towards recommending OpenID Connect as the default rather than token, with the k8s service account token being an option for a quick setup and testing, but not recommended for production use.
 * It makes the integration trivial - since an `id_token` from harbor is valid for Kubeapps/K8s if it's issued by the same identity provider).
 * It is a common auth across all three (harbor, kubeapps and the k8s api server).

Note that it doesn't exclude the systems using separate auth systems... a cookie credential could be used from harbor to Kubeapps as mentioned below (ie. a [JS Fetch request with the `include` option for cross-origin cookies](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#Sending_a_request_with_credentials_included) will send the kubeapps cookie once a user is auth'd), but this does not work for a token in localStorage - available only to the document's origin. So, unless we were wanting users to set a K8s service account token in the 3rd party app (?!), I think we're dependent on Kubeapps auth proxy to be able to have the 3rd party do api requests?

docs/architecture/design-proposals/third-party-add-repository.md
absoludity: From the third-party POV, I don't see any reason for that to be a separate request? So yes, I was assuming the former, that the single request to create an app repository in Kubeapps would include the creds (as hinted above) so the backend would create the `AppRepository` custom resource and the associated secrets (these are both in the namespace of the Kubeapps installation). What do you think? Let me know if you have other thoughts :)

docs/architecture/design-proposals/third-party-add-repository.md
```diff
@@ -0,0 +1,96 @@
...
+## User stories
+
+* As a repository maintainer in Harbor, I want to add my new repository to Kubeapps, so that Kubeapps users can deploy charts from my repository.
+* As a contributor to a Harbor repository, I want to test the installation of a chart after uploading to Harbor, so that I can verify the user installation experience.
+* As a repository maintainer in Harbor, I want to remove my repository from Kubeapps, so that users can no longer install apps from my repository until further notice. 
+
+## Implementation
+
+### Kubeapps
+Kubeapps provides an api endpoint for app repositories:
```
prydonius: I'm assuming we want to prevent external access to the Kubernetes API server (otherwise we could just use it directly instead), is that correct? Could this be as simple as a reverse proxy that only exposes the AppRepository CRD endpoints (e.g. a reverse proxy that to something like `http://kubernetes.default/apis/kubeapps.com/v1/apprepositores`)?

docs/architecture/design-proposals/third-party-add-repository.md
andresmgot: I see. I agree with the reasoning. Maybe it's not something to discuss in this document but I wanted to clarify what happen if OIDC is not configured. I am not saying that we need to support it but at least plan to have that case covered (documentation or a proper error may be enough).

docs/architecture/design-proposals/third-party-add-repository.md
andresmgot: That sounds good to me. We just need to clarify that the user needs to have permissions enough to create secrets in the Kubeapps namespace.

docs/architecture/design-proposals/third-party-add-repository.md
absoludity: > Could this be as simple as a reverse proxy that only exposes the AppRepository CRD endpoints (e.g. a reverse proxy that to something like `http://kubernetes.default/apis/kubeapps.com/v1/apprepositores`)?

Yeah I did think along these lines initially, but decided not to head that way as it is exposing Kubeapps implementation details which may change (ie. Kubeapps doesn't need to use a CRD for app repositories - in fact, see #1178 with Andres' proposal to change it... but whether or not that goes ahead at some point, it's an implementation detail which doesn't need to be exposed).

Harbor (or other repo UI) just know that kubeapps should now include repo x using provided credentials - and shouldn't care how Kubeapps keeps that state - both the repo details and the credentials, I think? WDYT? Happy to update the proposal to include that option if you think it's worthwhile considering more widely.

absoludity:
Thanks people. I'll go ahead and land this so I can point the harbor people at it, but do let me know if you've other thoughts.

PR #1214, opened 2019-10-10
Add support for resource elements

andresmgot:
Ref #1182

Added support for the resource specification:

Screenshot from 2019-10-10 12-44-59

I had to add more keys to the definition to have it as a subsection. My plan, in the close future, is to refactor the keys supported in the schema so it's not necessary to define them.

### DIFF
--- a/dashboard/src/components/DeploymentForm/BasicDeploymentForm/Subsection.tsx
+++ b/dashboard/src/components/DeploymentForm/BasicDeploymentForm/Subsection.tsx
@@ -8,8 +8,8 @@ export interface ISubsectionProps {
   label: string;
   param: IBasicFormParam;
   name: string;
-  enablerChildrenParam: string;
-  enablerCondition: boolean;
+  enablerChildrenParam?: string;
+  enablerCondition?: boolean;
   handleValuesChange: (value: string) => void;
   renderParam: (
     name: string,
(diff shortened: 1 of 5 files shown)


dashboard/src/components/DeploymentForm/BasicDeploymentForm/Subsection.tsx
```diff
@@ -40,11 +40,21 @@ class Subsection extends React.Component<ISubsectionProps> {
...
         >
-          <div className="margin-v-normal">{label}</div>
+          <div className="margin-v-normal">
+            {label}
+            {param.description && (
+              <>
+                <br />
+                <span className="description">{param.description}</span>
+              </>
+            )}
```
absoludity: Out of interest, wondering why
```
{param.description && (
  <p className="description">{param.description}</p>
)}
```
wouldn't suffice, rather than a break + span?

andresmgot: because when using a `<p>` the margin between lines grows too much:

Screenshot from 2019-10-14 12-05-22

vs

Screenshot from 2019-10-14 12-05-58

Not a very big deal but I prefer using that than change the default margin of a paragraph

absoludity: You wouldn't need to change the default margin of a paragraph though right? Just the style for p in this context. But yeah, not a big deal - I just tend to try use semantic elements rather than elements which mix semantics and presentation (ie. `<br>`).

andresmgot: > You wouldn't need to change the default margin of a paragraph though right? Just the style for p in this context

Yes, sorry I meant that. I will probably send a follow up PR touching up style so maybe I can change this.

PR #1215, opened 2019-10-11
Fix some param bugs when the value is undefined

andresmgot:


andresmgot:
> I'm a little unsure about using min as default for the slider... is it not possible to require a value for the slider? If not, then I agree min is a good default.

Unfortunately, it's not. Neither the value in the `values.yaml` nor the default in the schema are mandatory.

PR #1217, opened 2019-10-11
Refactor getChartVersion, values and schema

andresmgot:
ref #1182 

I have found some race conditions when working with the Deployment/Upgrade forms when, depending on the order in which the component receives the `chartVersion`, its `values` and its `schema` it behave differently. The logic to manage those resources asynchronously was growing and becoming unmanageable so I have merged those in a single function. In any case, the functions `getChartValues` and `getChartSchema` were never being used standalone.


dashboard/src/actions/charts.ts
```diff
@@ -149,36 +141,43 @@ export function getChartReadme(
...
-        dispatch(selectSchema({}));
-      } else {
-        // Unexpecter error
-        dispatch(errorChart(e));
+      dispatch(requestCharts());
+      const chartVersion = await Chart.getChartVersion(id, version);
+      if (chartVersion) {
+        let values = "";
+        let schema = {};
+        try {
```
absoludity: This try and the one below it look they could be DRY'd up a bit (the code is identical for both other than the function name?)

dashboard/src/components/AppUpgrade/AppUpgrade.tsx
```diff
@@ -36,8 +36,7 @@ interface IAppUpgradeProps {
   checkChart: (repo: string, chartName: string) => void;
   fetchChartVersions: (id: string) => Promise<IChartVersion[]>;
   getAppWithUpdateInfo: (releaseName: string, namespace: string) => void;
-  getChartVersion: (id: string, chartVersion: string) => void;
-  getChartValues: (id: string, chartVersion: string) => void;
+  getChartVersionWithValuesAndSchema: (id: string, chartVersion: string) => void;
```
absoludity: Just a thought: given that you're removing `getChartVersion`, could the new function just be `getChartVersion`, or does it need to be so specific?

dashboard/src/reducers/charts.ts
```diff
@@ -72,10 +67,6 @@ const chartsReducer = (state: IChartState = initialState, action: ChartsAction):
     case getType(actions.charts.selectReadme):
     case getType(actions.charts.errorReadme):
     case getType(actions.charts.errorChart):
-    case getType(actions.charts.selectValues):
-    case getType(actions.charts.selectSchema):
-      return { ...state, selected: chartsSelectedReducer(state.selected, action) };
-    case getType(actions.charts.selectSchema):
```
absoludity: Not sure if I'm misreading, but this looks like the `errorChart` type now falls through to the return below?

andresmgot:
> In what way were they behaving differently? Interested to know what the actual cause was? I'm not against this change - it looks more straight forward to use (esp. if we rename it simply getChartVersion) but it's not clear if we first understood the problem (which leaves me less confident of the solution).
> +1'ing, but keen to read a quick summary of the different behaviour you were seeing?

Yes, sorry I didn't have the time to explain further the issue. There were several. The first one was this line in the reducer:

https://github.com/kubeapps/kubeapps/pull/1217/files#diff-c1917d832ba0d88e0285de579d4668ebL26

`getChartVersion` was setting to `undefined` the values and schema. This caused that if the request of `getChartVersion` came after the `getChartSchema`, the schema was deleted.

The other issue is here:

https://github.com/kubeapps/kubeapps/blob/master/dashboard/src/components/DeploymentForm/DeploymentForm.tsx#L105

The method `componentWillReceiveProps` was executed both when the new values and the new schema were received. In that method (when I was refactoring it) I found it difficult to handle the different cases: the values has changed but the schema is not yet defined, the schema has changed but the values are not yet there, the values has been modified but the schema was the same... I had a lot of if-else that was difficult to maintain.

dashboard/src/reducers/charts.ts
andresmgot: yes, it's the same behavior than before (I just removed the cases for `selectSchema` and `selectValues`). The error is stored withing the `selected` object in the `chartsSelectedReducer` method.

PR #1218, opened 2019-10-14
Add basic form to the upgrade flow

andresmgot:
Ref: #1182 

This PR adds the basic form to the `UpgradeForm` component. Before this PR, `DeploymentForm` and `UpgradeForm` were independent components with a bunch of duplicated code. To avoid that, I have moved the common logic to a new component `DeploymentFormBody` so both components render that one and define its own particularities. This reduces the duplicated the code considerably.

Since `DeploymentFormBody` now supports the two different cases (a new deployment and upgrading), it can be considered as a "new" component (that's why this PR is so bulky).


dashboard/src/components/UpgradeForm/UpgradeForm.tsx
```diff
@@ -24,172 +20,95 @@ interface IDeploymentFormProps {
...
+    const { namespace, releaseName, error } = this.props;
+    if (error) {
       return (
-        <ErrorSelector
-          error={this.props.error}
-          namespace={namespace}
-          defaultRequiredRBACRoles={{ update: this.requiredRBACRoles() }}
-          action="update"
-          resource={`Application ${releaseName}`}
-        />
```
absoludity: So this is why the apprepository rbac role was still required for upgrades? Or it wasn't? It's not clear why this is removed with this change.

andresmgot: Yes, after your changes, it's no longer required to have RBAC permissions to upgrade an app here. We forgot to remove this then and while I was refactoring I removed that.

andresmgot:
> It's a great change to DRY up the functionality between upgrade and install, but for the future, I think the bulk could be avoided and I could do a much better job reviewing this (ie. identifying between mechanical changes and actual reviewable changes), if it was 2 simple PRs as follows:
> Move of DeploymentForm -> DeploymentFormBody with new parent DeploymentForm - this wouldn't require any mental effort to review as it's mostly a mechanical change,
Updating the UpgradeForm to use the DeploymentFormBody

Yes, sorry for that, most of this work was figuring out what was needed and identifying what I could get out of the `DeploymentForm` and what needed to stay there.

PR #1222, opened 2019-10-15
Style changes for basic form

andresmgot:
I have added some minor changes to the style of the basic form:

 - Added margin and separators between parameters
 - Display the label and input in the same line

Before:

Screenshot from 2019-10-15 12-25-53

After:

Screenshot from 2019-10-15 12-25-23


andresmgot:
> Was this with the feedback from Angel?

He agreed with the approach.

One minor thing he suggested was to increase the "step" size of the slider (like for example 128Mi for the memory instead of one by one). I tried that but the problem is that it breaks the current implementation for the input box in the slider. In that case, if you start typing "2" (because you want to write "256"), it handles 2 as a real value but since it's not a multiple of 128 it gets automatically changed to 128 so you cannot really write the value that you want.

That's probably something fixable but not trivial so I let that as a possibility for the future. For the rest, we are good to go for a first version and start gathering feedback.

PR #1224, opened 2019-10-15
Remove hardcoded values for subsections

andresmgot:
Ref: #1182

Follow up of #1223 

This PR allow developers to use specific keys to render subsections in a generic way. If a parameter enables or disables a different subsection, that parameter is moved as a children of the subsection. Examples:

 - `mariadb.enabled` disables `externalDatabase`:

```json
    "mariadb": {
      "type": "object",
      "properties": {
        "enabled": {
          "type": "boolean",
          "form": "useSelfHostedDatabase",
          "disable": "externalDatabase"
        }
      }
    },
    "externalDatabase": {
      "type": "object",
      "title": "External Database Details",
      "form": "externalDatabase",
      "properties": {
        "host": {
          ...
```

 - `ingress.enabled` enables `ingress`:

```json
    "ingress": {
      "type": "object",
      "form": "ingress",
      "properties": {
        "enabled": {
          "type": "boolean",
          "form": "enableIngress",
          "enable": "ingress",
        },
        "hostname": {
          "type": "string",
          "form": "hostname",
          "title": "Hostname"
        }
      }
    },
```

PR #1226, opened 2019-10-16
Split types for the different params

andresmgot:
Follow up of #1224 #1223 

As discussed in the above PRs, it's possible to define different types based on the information present in the param.

Let me know what you think (or if this is what you were suggesting @absoludity). I am not sure if this improves readability because you need to cast to the different types when needed, which can be confusing.


andresmgot:
okay, in that case I am fine with it, let's merge this.

PR #1228, opened 2019-10-16
Basic form docs

andresmgot:
Ref: #1182


docs/developer/basic-form-support.md
```diff
@@ -0,0 +1,289 @@
+# Basic Form Support
+
+**NOTE:** This feature is under heavy development. Some of the described below may change in the future.
+
+Since Kubeapps 1.6.0, it's possible to attach a chart with a JSON schema that defines the structure of the `values.yaml` file. This JSON schema is used with two goals:
```
absoludity: s/attach/include/

docs/developer/basic-form-support.md
```diff
@@ -0,0 +1,289 @@
...
+
+ - Validate that the given values satisfy the schema defined. In case the submitted values are not valid, the installation or upgrade will fail.
```
absoludity: Might be worth mentioning that this is actually the helm 3 use-case.

docs/developer/basic-form-support.md
```diff
@@ -0,0 +1,289 @@
...
+ - Present the user with a simple (basic) form so the chart is easier to deploy and configure.
```
absoludity: "with a simpler form" rather than "with a simple (basic) form"

docs/developer/basic-form-support.md
```diff
@@ -0,0 +1,289 @@
...
+
+The goal of the basic form is to define the most common parameters that users typically modify before deploying a chart (like username and password) so they are able to modify them with a more user-friendly form.
```
absoludity: The form itself isn't defining the most common properties - the schema is, right? So perhaps, "The goal of this feature is to present the user with the most common parameters which are typically modified before deploying a chart (like username and password) in a more user-friendly form."

docs/developer/basic-form-support.md
```diff
@@ -0,0 +1,289 @@
...
+
+## Create a values.schema.json
+
+This file, introduced with Helm v3, is a [JSON Schema](https://json-schema.org/) that defines the structure of the `values.yaml` file of the chart, including as many validations as needed. If a chart includes its schema, the values used are validated before submitting the new release.
+
+This file can define some or every possible value of the chart. Once it's written it should be included in the Helm package. The proposal to include it in Helm can be found [here](https://github.com/helm/helm/issues/5812).
+
+## Additional annotations used to identify basic parameters
+
+In order to identify, among all the values specified in the schema, the values that should be presented in the form, it's necessary to include some special tags.
```
absoludity: Simpler: In order to identify which values should be presented in the form, ...

docs/developer/basic-form-support.md
```diff
@@ -0,0 +1,289 @@
...
+
+First of all, it's necessary to specify the tag `form`. This field should be unique and it should identify the parameter. All the properties marked with this key in the schema will be represented in the form. For example:
```
absoludity: s/This field/The value of the `form` tag/ (and standardise on either field or tag or key)

docs/developer/basic-form-support.md
```diff
@@ -0,0 +1,289 @@
...
+
+```
+    "wordpressUsername": {
+      "type": "string",
+      "form": "username"
+    },
+```
+
+With the definition above, we are marking the value `wordpressUsername` as a value to be represented in the form (with the ID `username`). Note that we are also specifying the `type` that will be used to render the proper HTML components to represent the input in the form:
```
absoludity: Isn't `type` part of the jsonschema, and it's also being used here to render? In which case "Note that the `type` is also used to render..." (the current sentence implies that we're only specifying it for the rendering, rather than because it's part of the schema).

docs/developer/basic-form-support.md
```diff
@@ -0,0 +1,289 @@
...
+
+![username-input](../img/username-input.png)
+
+In addition to the `type`, there are other tags that can be used to customize the way the parameter is represented:
+
+ - `title` is used to render the title of the parameter. If it's not specified, Kubeapps will use the ID of the component.
```
absoludity: "Kubeapps will use the ID of the component (ie. the value of the `form` field)." (it wasn't 100% clear, but I see now that's what you're referring to).

docs/developer/basic-form-support.md
```diff
@@ -0,0 +1,289 @@
...
+ - `description` is used to include additional information of the parameter.
+ - `default` is used to set a default value. If it's not set, the default value from the `values.yaml` file will be used.
```
absoludity: Oh ... not sure that's a good idea? Why have two places to specify a default, so that you'll get different defaults depending on what UI you use? What is the reason for having a second default here rather than just using the one from the values.yaml always?

docs/developer/basic-form-support.md
andresmgot: okay, yes, it's used for both

docs/developer/basic-form-support.md
andresmgot: sorry, I expressed it the other way. If the value is set in the `values.yaml` it will always be used as the default. In case that is not set and the `default` value is set in the schema, that one will be used.

absoludity: Yeah, but is there any reason to enable setting a default in the schema for a field at all, given the right place for that is already in the values.yaml? If not, I'd personally update this sentence removing any mention of default (or just saying the default is set in the values) and later remove support for a separate default in the schema. +1 either way, but I'm keen to know if there's a need at all for a second way to set the default for a field.

PR #1231, opened 2019-10-17
Simplify usingOIDCToken in Auth module

SimonAlling:
Another nice (indirect) consequence of this PR is that I now have the test suite up and running. :smiley:

### DIFF
--- a/dashboard/src/shared/Auth.ts
+++ b/dashboard/src/shared/Auth.ts
@@ -21,8 +21,7 @@ export class Auth {
   }
 
   public static usingOIDCToken() {
-    const oidc = localStorage.getItem(AuthTokenOIDCKey);
-    return oidc === "true" ? true : false;
+    return localStorage.getItem(AuthTokenOIDCKey) === "true";
   }
 
   public static wsProtocols() {


SimonAlling:
> Whoah - that was a little convoluted. Are you actually using the OIDC support? If so, just note that I'm currently looking to switch to use oauth2-proxy in #1199

I think we are, but I'm just now starting to discover kubeapps, hehe. I don't know whether your PR is relevant to us, but @lindhe might know.

absoludity:
It shouldn't be relevant for any development work you're doing for helm 3 ( given that the authn method shouldn't impact that) - I just wasn't sure whether you were using it in your actual deployment (and just wanted to give a heads-up about the change :) )

lindhe:
We are using OIDC in our current deployment of Kubeapps. But we are not looking to change that with the Helm 3 development. The reason for looking into that part now is just for getting a feel for how the code is structured, tested etc.

However, I do expect that we will need to tinker around a bit with authentication. Particularly, I suspect that we will need to implement some sort of user profile system. But we'll see about that. And it shouldn't have much to do with OIDC anyhow.

PR #1232, opened 2019-10-17
Update chartsvc and chart-repo versions

andresmgot:
Needed for #1182


andresmgot:
_waiting for a fix in a bug in the released image of `chart-repo`_

Issue #676, opened 2018-09-24
Check and document OpenShift support 


migmartri:
We should check if Kubeapps is compatible with OpenShift as well as documenting any setup differences.

migmartri:
Something that I noticed is that `helm status` reports that the mongodb deployment has a different API version

```
==> v1beta1/Deployment
NAME              DESIRED  CURRENT  UP-TO-DATE  AVAILABLE  AGE
kubeapps-mongodb  1        0        0           0          51m

==> v1beta2/Deployment
kubeapps-internal-apprepository-controller  1  1  1  1  51m
kubeapps-internal-chartsvc                  2  2  2  0  51m
kubeapps-internal-dashboard                 2  2  2  2  51m
kubeapps                                    2  2  2  2  51m
kubeapps-internal-tiller-proxy              2  2  2  0  51m
```

and actually this deployment can not schedule any pod.

This is the last entry in the deployment status timeline.

```
  Normal  ScalingReplicaSet  53m   deployment-controller  Scaled up replica set kubeapps-mongodb-5898f98f66 to 1
```

and this is the error in the replica set

```
  Warning  FailedCreate  5s (x14 over 46s)  replicaset-controller  Error creating: pods "kubeapps-mongodb-5898f98f66-" is forbidden: unable to validate against any security context constraint: [fsGroup: Invalid value: []int64{1001}: 1001 is not an allowed group spec.containers[0].securityContext.securityContext.runAsUser: Invalid value: 1001: must be in the ranges: [1000140000, 1000149999]]
```

migmartri:
Another issue I've found is that this command does not work

```
kubectl get secret $(kubectl get serviceaccount kubeapps-operator -o jsonpath='{.secrets[].name}') -o jsonpath='{.data.token}' | base64 --decode
```

 because there is actually two secrets listed inside the service account secrets array and we pick the first one
```
apiVersion: v1
imagePullSecrets:
- name: kubeapps-operator-dockercfg-qflks
kind: ServiceAccount
metadata:
  creationTimestamp: 2018-10-19T18:03:10Z
  name: kubeapps-operator
  namespace: kubeapps
  resourceVersion: "34638"
  selfLink: /api/v1/namespaces/kubeapps/serviceaccounts/kubeapps-operator
  uid: 3d7c5b51-d3c9-11e8-8ef5-080027c965a6
secrets:
- name: kubeapps-operator-dockercfg-qflks
- name: kubeapps-operator-token-tt564
[...]

PR #1233, opened 2019-10-18
Don't depend on sourcing minishift oc-env.

absoludity:
I'd previously assumed that `script/openshift-cluster.mk` would run without the user necessarily having sourced the output of `minishift oc-env`, but this is unnecessary (ie. it's a valid assumption that the user will have done this), and complicates the targets (as it required running only a single command per target after having sourced).

This also fixes `helm dep update` to `helm dep build` (I didn't realise update was pulling charts from requirements.yaml rather than requirements.lock... build is the one we want, as it just ensures that what is in the lock file is present), and updates that target to use the specific mongodb version (otherwise it pulls it every time).

Also fixes the `-reset` target to ensure all items run even if some fail (as if you have an incomplete deployment, it would error if it couldn't delete something that didn't exist etc.)

Ref #676

### DIFF
--- a/chart/kubeapps/templates/apprepository-rbac.yaml
+++ b/chart/kubeapps/templates/apprepository-rbac.yaml
@@ -25,6 +25,7 @@ rules:
   - list
   - update
   - watch
+  - delete
 - apiGroups:
   - batch
   resources:
--- a/script/openshift-cluster.mk
+++ b/script/openshift-cluster.mk
@@ -1,66 +1,67 @@
-# This Makefile assumes that you have:
+# This Makefile aims to help local development with an OpenShift/Minishift cluster. It is not for production
+# use, but does document requirements for OpenShift. The targets assume that you have:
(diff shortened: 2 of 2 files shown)


absoludity:
> _doesn't like Makefiles_

Interest to chat about what you don't like - I won't try to convert you :P For simple projects (where builds/tests are very quick, I agree a Makefile doesn't provide a lot of benefit over similar bash scripts (well, other than automatic parallelization of tasks). The main benefit for any non-trivial project (and for me here) is that if the dependencies are correct, minimal builds (just the things required) are automatic.

A good example here is that it won't bother updating repos and downloading the mongo chart if the file exists. Of course you can encode that in your bash script for every function/target, but here it's automatic. On the other side, it's a bit hacky here because the actual artifacts aren't necessarily local (ie. they might be docs applied to the k8s server). So it's not perfect - happy to update to a better alternative.

Issue #843, opened 2018-11-19
OpenShift: Initial import jobs not created


migmartri:
After installing Kubeapps, it seems that the expected k8s jobs used for chart import are not enqueued.

In the repository controllers you can find:

```bash
E1102 23:42:29.698459       1 controller.go:239] error syncing 'miguel/bitnami': cronjobs.batch "apprepo-sync-bitnami" is forbidden: cannot set an ownerRef on a resource you can't delete: User "system:serviceaccount:miguel:kubeapps-internal-apprepository-controller" cannot delete cronjobs.batch in project "miguel", <nil>
E1102 23:42:29.698811       1 controller.go:239] error syncing 'miguel/incubator': cronjobs.batch "apprepo-sync-incubator" is forbidden: cannot set an ownerRef on a resource you can't delete: User "system:serviceaccount:miguel:kubeapps-internal-apprepository-controller" cannot delete cronjobs.batch in project "miguel", <nil>
E1102 23:42:29.699992       1 controller.go:239] error syncing 'miguel/stable': cronjobs.batch "apprepo-sync-stable" is forbidden: cannot set an ownerRef on a resource you can't delete: User "system:serviceaccount:miguel:kubeapps-internal-apprepository-controller" cannot delete cronjobs.batch in project "miguel", <nil>
E1102 23:42:29.701056       1 controller.go:239] error syncing 'miguel/svc-cat': cronjobs.batch "apprepo-sync-svc-cat" is forbidden: cannot set an ownerRef on a resource you can't delete: User "system:serviceaccount:miguel:kubeapps-internal-apprepository-controller" cannot delete cronjobs.batch in project "miguel", <nil>
E1102 23:49:33.098667       1 controller.go:239] error syncing 'miguel/bitnami': cronjobs.batch "apprepo-sync-bitnami" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: User "system:serviceaccount:miguel:kubeapps-internal-apprepository-controller" cannot update apprepositories/finalizers.kubeapps.com in project "miguel", <nil>
```

It seems that the app-repository-controller role does not have enough permissions specifically missing:

* delete verb to cronjob and job
* update verb to apprepository/finalizers

This can be fixed by updating the  kubeapps-internal-apprepository-controller RBAC role being a working role something like:

```yaml
kubectl get role kubeapps-internal-apprepository-controller  -o yaml
apiVersion: authorization.openshift.io/v1
[...]

PR #1234, opened 2019-10-18
Add cronjob delete perm for app repo controller to avoid startup error

absoludity:
When repeatedly testing installation on openshift (minishift), I'd find that the app repository controller failed to create the cron-jobs, with the error:

```
E1018 01:25:15.595721       1 controller.go:239] error syncing 'kubeapps/incubator': cronjobs.batch "apprepo-sync-incubator" is forbidden: cannot set blockOwnerDeletion in this case because cannot find RESTMapping for APIVersion kubeapps.com/v1alpha1 Kind AppRepository: no matches for kind "AppRepository" in version "kubeapps.com/v1alpha1" 
```

yet, if I force container restarts (just updating the deployment pull policy, for eg), they would restart without issue. So it appears to be a timing issue, that if the app repository controllers are up and running before the RESTMapping for the kubeapps.com/v1alpha is available, it will fail.

I found a reference to this error message in a [commit](https://gitlab.cncf.ci/kubernetes/kubernetes/commit/3b9eb1a8755d744d51a0be47cd85c0350307cc59) which indicates that this check is only done if the user (in our case, app-repo-controller) does not have delete permission on the object. Indeed, if I add delete, the check is skipped during startup and there are no errors on Openshift.

So it's not clear to me if this would be reproducible on vanilla k8s, depending on the timing (we don't appear to have reports of it), but given that we're creating the CronJob resources with an owner reference using metav1.NewControllerRef which sets `BlockOwnerDeletion: true`, I don't see any reason not to include delete here? That is, the cronjob ends up with an owner reference like:

```
kind: CronJob
metadata:
  creationTimestamp: 2019-10-18T02:01:01Z
  name: apprepo-sync-bitnami
  namespace: kubeapps
  ownerReferences:
  - apiVersion: kubeapps.com/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: AppRepository
    name: bitnami
    uid: 1738310b-f14b-11e9-b0c1-52540079aa76
  resourceVersion: "172956"
```
which says that deleting the `bitnami` app repository will cause the GC to first ensure the cron job is deleted.

Too long an explanation for a one-liner :P

### DIFF
--- a/chart/kubeapps/templates/apprepository-rbac.yaml
+++ b/chart/kubeapps/templates/apprepository-rbac.yaml
@@ -25,6 +25,7 @@ rules:
   - list
   - update
   - watch
+  - delete
 - apiGroups:
   - batch
   resources:


prydonius:
Thanks for the digging into this a bit more, I think this is the same issue @migmartri when he was looking into OpenShift support (https://github.com/kubeapps/kubeapps/issues/843). He mentions also adding the update verb for apprepository/finalizers, is that something you found?

absoludity:
> Thanks for the digging into this a bit more, I think this is the same issue @migmartri when he was looking into OpenShift support (#843). He mentions also adding the update verb for apprepository/finalizers, is that something you found?

I did earlier and landed that with https://github.com/kubeapps/kubeapps/pull/1230/files , though I'd just added `apprepositories/finalizers` to the existing `Rule` used by `apprepositories` (ie. it has not just `update`, but `get`, `list` and `watch`).

Also in this PR I've only added `delete` to cronjob, not job (as mentioned in #843) - so I'll check later if it's also needed (it didn't appear to be to get things up and running.

Thanks for pointing out that PR. Fixes #843.

PR #1238, opened 2019-10-18
Fix semver 2.0 tags (+w.x.y-z) handling 

kandeshvari:


kandeshvari:
Seems strange for me. Test fails on commit where wasn't any corresponding changes.

andresmgot:
hello @kandeshvari, thank you for the PR!

The test (actions/repo.test.tsx) is failing because the changes in `actions/repo.ts`. The reason is because in that test we are mocking `axios.get` that is no longer being used:

https://github.com/kubeapps/kubeapps/blob/master/dashboard/src/actions/repos.test.tsx#L16

You need to mock `Chart.fetchChartVersions` instead.

Issue #1220, opened 2019-10-14
Kubeapps not handling semver 2.0 tags (+w.x.y-z)


kandeshvari:
I use semver tags in chart versions like 0.1.14+2.1.0-5

It seems like kubeapps can't process semver tag (https://semver.org/#spec-item-10) properly and it leads to two bugs:

1. I can not upgrade existing release. Dashboard shows page with combo "choose helm repo" and redirects to login page when I choose any option. So I can't upgrade release at all.
2. There is no green label "updates available" on the release card when new version exists.

When I use versions for that chart without tag - all is ok.

absoludity:
Thanks @kandeshvari . It appears to be a definite bug. Triaging and renaming to get it fixed ASAP. Let me know if it's something you'd like to work on, otherwise it'll get picked up when we have bandwidth. Thanks again for creating the issue!

kandeshvari:
@absoludity sure, why not :) But I need some advice where to start from?

andresmgot:
It appears that you have hit two different bugs. First, you are seeing the "Choose helm repo" form because the request to automatically retrieve the repo for your application will look something like:

```
http://kubeapps.local/api/chartsvc/v1/charts?name=mariadb&version=0.1.14+2.1.0-5&appversion=10.3.18
```

As you can see, the version is in the query of the request. The problem is that the symbol `+` means "space" when it's present in a query parameter so it's resolved to `version=0.1.14 2.1.0-5` and the system is not able to find a chart with that version. We would need to escape that character using the method `encodeURIComponent` for the version (and probably the `appversion` too).

That code is here:

https://github.com/kubeapps/kubeapps/blob/master/dashboard/src/shared/Chart.ts#L51

But we would need to probably fix the encoding in these URLs too:

https://github.com/kubeapps/kubeapps/blob/master/dashboard/src/shared/url.ts#L22

Then, the second bug is that, when checking if a chart belongs to a repository (that's what happens when you select a registry), we are directly accessing the URL of the `chartsvc` without auth:

https://github.com/kubeapps/kubeapps/blob/master/dashboard/src/actions/repos.ts#L217

We changed that route to ensure authentication, that's why the server returns a `Not Authorized` response and you are logged out. To fix that, you can change the call to `axios` to the `Chart` method that attaches the proper auth info:

```patch
     try {
-      await axios.get(url.api.charts.listVersions(`${repo}/${chartName}`));
[...]

PR #1239, opened 2019-10-19
Fix tags handling #1220

kandeshvari:


dashboard/src/shared/url.ts
```diff
@@ -23,13 +23,13 @@ export const api = {
...
+      `${api.charts.base}/assets/${id}/versions/${encodeURIComponent(version)}/README.md`,
     getValues: (id: string, version: string) =>
-      `${api.charts.base}/assets/${id}/versions/${version}/values.yaml`,
+      `${api.charts.base}/assets/${id}/versions/${encodeURIComponent(version)}/values.yaml`,
     getSchema: (id: string, version: string) =>
-      `${api.charts.base}/assets/${id}/versions/${version}/values.schema.json`,
+      `${api.charts.base}/assets/${id}/versions/${encodeURIComponent(version)}/values.schema.json`,
     getVersion: (id: string, version: string) =>
-      `${api.charts.base}/charts/${id}/versions/${version}`,
+      `${api.charts.base}/charts/${id}/versions/${encodeURIComponent(version)}`,
```
absoludity: Looking great. It would be great, but not required, if you could add a tests to show your change fixes the issue. For example, for `getChartVersion` you could add a test here:

https://github.com/kubeapps/kubeapps/blob/master/dashboard/src/actions/charts.test.tsx#L87

with a version with a tag (just copy the existing test and add the tag). You should find it fails without your above change, but passes with. Again, no problem if you don't have time to do so.

kandeshvari: No problem :) I've added test as you suggested. Seems good.

PR #1241, opened 2019-10-21
Frontend injects wsprotocols only when not using oidc authn

absoludity:
Follows on from #1199. When QAing #1199, we noted that the websocket requests were 400ing. This was because with 1199, I've switched to no longer white-list requests to the k8s api, so they are sent via the proxy. The proxy handles adding any required auth headers for all requests (websocket requests included).

With this change, they upgrade without issue:
oauth2-proxy-websockets

While there, I also removed the specific (key-cloak-based) cookie name which I'd added because I thought it would be referenced in the frontend and need updating there. Turns out its not at all, so we can just let `oauth2-proxy` use its default cookie name and avoid any confusion.


andresmgot:
Great, thank you! Let's merge the two PRs

PR #1242, opened 2019-10-21
Remove basic form feature flag

andresmgot:
Ref #1182


andresmgot:
_merging this to trigger a new release_

PR #1244, opened 2019-10-22
Use const assertions in Routes.tsx

SimonAlling:
With [`const` assertions](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions), TypeScript infers as specific types as it can (and `readonly` recursively for properties). I think they're great in many situations, so I figured I'd try them out real quick. What do you guys think?

### DIFF
--- a/dashboard/src/containers/RoutesContainer/Routes.tsx
+++ b/dashboard/src/containers/RoutesContainer/Routes.tsx
@@ -19,9 +19,7 @@ import ServiceInstanceViewContainer from "../../containers/ServiceInstanceViewCo
 
 type IRouteComponentPropsAndRouteProps = RouteProps & RouteComponentProps<any>;
 
-const privateRoutes: {
-  [route: string]: React.ComponentType<RouteComponentProps<any>> | React.ComponentType<any>;
-} = {
+const privateRoutes = {
   "/apps/ns/:namespace": AppListContainer,
   "/apps/ns/:namespace/:releaseName": AppViewContainer,
   "/apps/ns/:namespace/new/:repo/:id/versions/:version": AppNewContainer,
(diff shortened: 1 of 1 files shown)


SimonAlling:
> Hi! Thanks for the suggestion. I am not sure though if this has any effect? Even if I set `privateRoutes` to `const`, I am able to modify the value:
> 
> ```
> privateRoutes["/apps/ns/:namespace"] = {} as React.ComponentType<RouteComponentProps<any>>;
> ```

Hm, that's strange. That exact line is a compile error for me (with my `const` assertion changes):

    Cannot assign to '"/apps/ns/:namespace"' because it is a read-only property.ts(2540)

> Maybe that only has effect if you create the variable with `let`? Or am I missing something?

No, it should work with `const` – that's even the primary intended use case, I believe.

SimonAlling:
Aha! Did you perhaps try adding `as const` without removing the explicit `{ [route: string]: ... }` typings? Because the latter override the `const` assertions and bring back mutability.

andresmgot:
> Aha! Did you perhaps try adding as const without removing the explicit { [route: string]: ... } typings? Because the latter override the const assertions and bring back mutability.

Right, that was my issue, good catch!

The downside I see in this approach is that now, the value of the map is not typed so you can include a wrong type by mistake:

```js
const privateRoutes = {
  "/services/instances/ns/:namespace": ServiceInstanceListContainer,
  "foo": "bar",
} as const;
```

In any case, it's true that it has other benefits so I am fine either way.

SimonAlling:
> The downside I see in this approach is that now, the value of the map is not typed so you can include a wrong type by mistake:
> 
> ```js
> const privateRoutes = {
>   "/services/instances/ns/:namespace": ServiceInstanceListContainer,
>   "foo": "bar",
> } as const;
> ```
> 
> In any case, it's true that it has other benefits so I am fine either way.

Yes, but you can actually get a type error at the use site:

```tsx
                                                             ^^^^^^^^^

// Type '"bar"' is not assignable to type 'ComponentClass<any, any> | ... | undefined'.
```

However, this seems to require that `Object.keys` be replaced with `Object.entries`: As far as TypeScript knows, the former only returns plain old strings, whereas the latter makes each property value accessible with as narrow a type as possible (e.g. `ConnectedComponentClass<typeof LoginForm, Pick<ILoginFormProps, "location">>`). Luckily, #1247 does exactly this.

PR #1248, opened 2019-10-23
Enable browser tests

### LINKED ISSUE #1209: Add web browser tests
Our current end-to-end tests only cover the chart installation and tests. In order to avoid major regressions we should include a basic coverage of web browser tests to ensure we don't break the current UX.

andresmgot:
Ref #1209 

This PR adds support for running browser tests in our CI. See inline comment and the docs added for more information.

Test: https://circleci.com/workflow-run/cc32fbe1-f72c-45d8-ba4b-b09944baf6d3


.circleci/config.yml
```diff
@@ -154,7 +154,13 @@ run_e2e_tests: &run_e2e_tests
       DEV_TAG=$(latestReleaseTag)
       IMG_MODIFIER=""
     fi
-    ./script/e2e-test.sh $DEV_TAG $IMG_MODIFIER
+    if ./script/e2e-test.sh $DEV_TAG $IMG_MODIFIER; then
```
andresmgot: I had to run the `e2e` tests as a conditional to be able to gather the screenshots afterwards (and not exiting directly)

script/e2e-test.sh
```diff
@@ -118,6 +118,28 @@ if [[ "$code" != 0 ]]; then
...
+
+# Browser tests
+cd $ROOT_DIR/integration
+kubectl apply -f manifests/executor.yaml
+k8s_wait_for_deployment default integration
+pod=$(kubectl get po -l run=integration -o jsonpath="{.items[0].metadata.name}")
+## Copy latest tests
+kubectl cp ./use-cases ${pod}:/app/
+## Make default user admin
+kubectl create clusterrolebinding default-admin --clusterrole=admin --serviceaccount default:default
```
absoludity: Is this temporary? Or why does the default service account need admin?

Is the plan that in the future different tests can use different roles based on different tokens?

absoludity: Actually, this sounds really dangerous? If I happen to run the browser tests on whatever cluster I have configured, it should *not* be making this change? Let's update to create a specific service account to be used in he ci tests? (and later we can add multiple to test different scenarios of roles?)

andresmgot: I don't expect people executing this script in a non testing environment. Note that in the line 54-55 we are already giving admin privileges to the default service account in the `kube-system` namespace, due to an old bug in Minikube. That can be probably be deleted TBH.

In any case, I used the default service account because it was convenient to get the token but I can use something else.

absoludity: Yeah, I wouldn't expect it either, but if they did and didn't realise, it'd be a bad story. I didn't realise that it was already in use in `kube-system`, but yeah, if it's simple to remove and instead create a service account for use, +100, but I'm ok either way. Thanks for clarifying.

andresmgot:
> lgtm - though I'd love to see us no longer giving admin privs to the default account, even though it is intended for a test cluster.

No, I agree. I am going to update this PR to create a specific admin user for the tests.

Issue #1236, opened 2019-10-18
[basic form] Support for parameters that affect different sections


andresmgot:
Follow up from #1182 

From: https://github.com/helm/charts/pull/18078#issuecomment-543212398

In the WordPress chart (but this pattern is present in other charts as well), there is a parameter `mariadb.enabled` that is being used for: 

  1. Enabling MariaDB (values under `mariadb.*`)
  2. Disabling/Ignoring the external database details (values under `externalDatabase.*`)

Ideally we want to show information about both sections in the basic form. If MariaDB is enabled, we want to allow people to select the size of the volume used by MariaDB. When disabled, it should show the external database details to fill those parameters.

The current solution has the limitation that a section enabler can be associated with just one section (so we either choose to show details of MariaDB or the external database).

Two possibilities to fix this situation are:
 - Support, somehow, to have a relation 1:2 between a section enabler and the sections it affects. This way, if a schema parameter defines the properties `enables: *` and `disables: *`, both sections would be merged in one and displayed depending on the value of the enabler.
 - Change the chart values and replace the boolean parameter by a choice one. That way, we could have a parameter `databaseSelector: mariadb|external`. And generically add support for `choice` parameters that enables a certain section.

I would rather implement the second, but that'd mean that we don't support the first, requiring changes in the existing chart values.

cc/ @juan131

absoludity:
> I would rather implement the second, but that'd mean that we don't support the first, requiring changes in the existing chart values.

It's not clear to me why you prefer this option if it means we'd need to ask chart authors to change their values.yaml to support this? Or do you mean something else?

Also, keen to know what you don't like about your first option? (if it only requires schema changes, and is generic enough to support all situations?). Also, with the first option, does it need to be binary? Could it instead be:
```
enables:
  - section1
  - section2
disables:
  - section3
```
so that it's more generic? And related to that, why would we need to merge the sections to display them - why not keep the sections separate?

andresmgot:
> It's not clear to me why you prefer this option if it means we'd need to ask chart authors to change their values.yaml to support this? Or do you mean something else?

I mean that I'd rather implement the second option from the implementation point of view in Kubeapps because it's a matter of supporting "choice" options in a generic way. It's true that requiring changes in the chart values is a downside important enough to discard that possibility. In any case, yes, let's go with the option that doesn't require changes in the chart.

> Also, keen to know what you don't like about your first option? (if it only requires schema changes, and is generic enough to support all situations?)

[...]

PR #1257, opened 2019-10-28
[basic form] Replace enables/disables property with hidden

andresmgot:
Closes: #1236 

This is a proposal that would "fix" #1236. Instead than supporting a parameter that `enables` or `disables` a section, add a parameter `hidden` that hides the parameter if the given condition is true. For example:

```
        "host": {
          "type": "string",
          "form": true,
          "title": "Database Host",
          "hidden": "mariadb.enabled"
        },
```

The parameter `host` will be hidden if `mariadb.enabled` is `true`. It's also possible to specify a value to match:

```
        "hostname": {
          "type": "string",
          "form": true,
          "title": "Hostname",
          "hidden": {
            "condition": false,
            "value": "ingress.enabled"
          }
        }
```

In that example, the `hostname` will be hidden if `ingress.enabled === false`.

The good thing about this solution is that we don't need to reorder the hierarchy of parameters anymore so I could delete a bunch of code. The downside is that now the `externalDatabase` section is outside the `mariadb` section but with that can be mitigated with the description:

Screenshot from 2019-10-28 18-09-57

when disabling `mariadb`:

Screenshot from 2019-10-28 18-10-06

This requires changes in the current schemas but the changes are minimal.

An collateral improvement is that now the key `form` in the schema doesn't need to be a unique `id`, since the key `form` was only used for enabling or disabling a section but I will change that in a follow up PR if this goes ahead.


andresmgot:
> Right, but a chart author can choose to fix that if they want, right? They're just not being forced to do so?

Right, yes. I meant with the current status of the chart.

> Do we need to be careful with the charts? ie. for the ones you've done, is it worth adding the hidden fields and leaving the existing functionality (so the charts will continue to work with 1.6) until later, or just changing them?

Here the trade-off is confusing developers that want to add this feature to new charts vs people that will use the latest chart but not the latest Kubeapps. In this case, I think it's fine to remove the old fields because the only result, for people using 1.6.0, will be that all the fields will be visible but the form is still usable.

PR #1261, opened 2019-10-29
Additional end-to-end tests

### LINKED ISSUE #1209: Add web browser tests

andresmgot:
Closes #1209 

Add two new tests:

 - Use a non-privileged user to create a release (that fails)
 - Create a new repository

See inline comments.

### DIFF
--- a/chart/kubeapps/templates/apprepository-rbac.yaml
+++ b/chart/kubeapps/templates/apprepository-rbac.yaml
@@ -86,10 +86,7 @@ rules:
   resources:
   - apprepositories
   verbs:
-  - get
-  - create
-  - update
-  - delete
+  - "*"
 - apiGroups:
   - ""
   resources:
--- a/integration/use-cases/create-registry.js
+++ b/integration/use-cases/create-registry.js
@@ -1,3 +1,5 @@
+jest.setTimeout(120000);
+
 test("Deploys an application with the values by default", async () => {
   page.setDefaultTimeout(4000);
   await page.goto(getUrl("/#/login"));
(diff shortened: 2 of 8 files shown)


chart/kubeapps/templates/apprepository-rbac.yaml
```diff
@@ -86,10 +86,7 @@ rules:
   resources:
   - apprepositories
   verbs:
-  - get
-  - create
-  - update
-  - delete
+  - "*"
```
andresmgot: the `write` role was missing the `list` permissions

integration/use-cases/create-registry.js
```diff
@@ -0,0 +1,41 @@
...
+
+  await expect(page).toFillForm("form", {
+    token: process.env.ADMIN_TOKEN
+  });
+
+  await expect(page).toClick("button", { text: "Login" });
+
+  // Double click to show configuration
+  await expect(page).toClick("a", { text: "Configuration" });
+  await expect(page).toClick("a", { text: "Configuration" });
```
andresmgot: I tried to `hover` over the link but it doesn't have any valid selector

chart/kubeapps/templates/apprepository-rbac.yaml
absoludity: So using "*" also adds `patch` and `watch`. Not sure if that's intended.

andresmgot: yes, I mean, I would expect this role to be "admin" for apprepositories

PR #1265, opened 2019-11-04
Remove requirement for cluster-admin for tiller in openshift example.

### LINKED ISSUE #676: Check and document OpenShift support

absoludity:
Fixes #676 .

This PR fixes one remaining issue I had with the development environment on OpenShift, removing the need to use `cluster-admin` for tiller. Together with the work from the past iteration, Kubeapps now installs without issue on OpenShift.

I'm not keen to make a stronger statement about OpenShift support for a few reasons, so have updated the relevant part of the troubleshooting in the README to mention our tentative support and link to the RBAC policies as examples for the related issue.

As per TODOs in the doc, I don't think the RBAC requirements are specific to OpenShift, but rather we should be promoting these as the default (for helm2 / tiller - we don't seem to mention working with non-cluster-admin tiller anywhere that I can see. Checked our [access-control](https://github.com/kubeapps/kubeapps/blob/master/docs/user/access-control.md) doc but that is Kubeapps specific)

### DIFF
--- a/chart/kubeapps/README.md
+++ b/chart/kubeapps/README.md
@@ -179,13 +179,15 @@ Or:
 Error: namespaces "kubeapps" is forbidden: User "system:serviceaccount:kube-system:default" cannot get namespaces in the namespace "kubeapps"
 ```
 
-This usually is an indication that Tiller was not installed with enough permissions to create the resources by Kubeapps. In order to install Kubeapps, you will need to install Tiller with elevated permissions (e.g. as a cluster-admin). For example:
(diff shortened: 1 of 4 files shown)


absoludity:
> Thanks for the update!
> 
> To know your thoughts, what do you think about running our e2e tests against minishift (or similar)? We probably don't want to do so if it's not stable or if it's too slow but if we don't it, it's also likely that the code in the Makefile will become obsolete.

Yes - I'd love to add that along with e2e tests against an oauth2 setup etc. I'm assuming we don't use kind for the normal e2e tests atm, but will take a look soon.

andresmgot:
> I'm assuming we don't use kind for the normal e2e tests atm, but will take a look soon.

Yes, we do. Let me know if you need more info about that.

PR #1266, opened 2019-11-04
Move timeout setting per test

andresmgot:
Ref #1209 

We are still seeing timeout issues. The global configuration appears not to be working.

### DIFF
--- a/integration/use-cases/create-registry.js
+++ b/integration/use-cases/create-registry.js
@@ -1,3 +1,6 @@
+const { setDefaultOptions } = require("expect-puppeteer");
+
+setDefaultOptions({ timeout: 8000 });
 jest.setTimeout(120000);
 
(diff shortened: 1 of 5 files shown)


integration/use-cases/create-registry.js
```diff
@@ -1,20 +1,18 @@
...
+  page.setDefaultTimeout(8000);
+  await page.goto(getUrl("/#/config/repos"));
 
   await expect(page).toFillForm("form", {
     token: process.env.ADMIN_TOKEN
   });
 
   await expect(page).toClick("button", { text: "Login" });
 
-  // Double click to show configuration
```
andresmgot: I had to remove this because the double click in the "configuration" was flaky

andresmgot:
_merging this to fix issues in open PRs_

Issue #1269, opened 2019-11-04
Investigate and demo LDAP for Kubeapps


absoludity:
We have had requests from multiple parties whether LDAP can be used for Kubeapps authn.

Our assumption is that it can but with specific conditions:
 * dex connected via the [LDAP connector](https://github.com/dexidp/dex/blob/master/Documentation/connectors/ldap.md) to the LDAP server, so that username/password authn can be used (with TLS) to authenticate with the server,
 * kubernetes api server using dex as the OpenIDConnect authorization server

We need to investigate and demonstrate that this does indeed work as expected.

PR #1270, opened 2019-11-05
Add cluster-kind.mk with setup for k8s api server oidc configured with test CA

absoludity:
This diff:
 * Adds a cluster-kind.mk which enables a `cluster-kind` target that starts a local k8s environment with OIDC support, configured to trust a test CA,
 * Extracts the cert generation to a script and updates to create certs which can can be used for multiple hostnames (internally k8s will request dex.dex using tls, but we'll be port-forwarding to localhost, and the previous certs used CN=localhost).

I assume there's a way I can check locally that this doesn't impact e2e tests (it shouldn't, but I'd like to check tomorrow).

Q: Is there a reason to store these certs in the repo at all? I could have them as make targets which just get generated if not present etc.?

Ref: #1269


script/cluster-kind.mk
```diff
@@ -0,0 +1,24 @@
...
+	jq ".nodes[0].extraMounts[0].hostPath = \"${PWD}/script/test-certs/ca.cert.pem\"" > \
+	$@
+
+${CLUSTER_CONFIG}: devel/local-dev-apiserver-config.json
+	kind create cluster \
+		--name ${CLUSTER_NAME} \
+		--config=./devel/local-dev-apiserver-config.json \
+		--retain
+
+cluster-kind: ${CLUSTER_CONFIG}
```
andresmgot: you also need a target to load the `kubeconfig-path`, no? `export KUBECONFIG=$(kind get kubeconfig-path --name=kind)`

absoludity: Makefiles don't let you change the external environment, each command is run in a subshell, which is a good thing... so what I want to do is separate the cluster creation helpers (like this one) from the kubeapps deployment configuration helper (which I'll add next). This is because deploying a configured oidc kubeapps example will be the same whether it's on a local kind or minishift or non-local env. So it'll be:

1) Create your configured cluster (ie. with the oidc endpoint)
2) Deploy the example kubeapps (which will configured for oidc)

andresmgot: true, I forgot that

_I am not a enthusiast of Makefiles_

absoludity: > true, I forgot that
> 
> _I am not a enthusiast of Makefiles_

That's right - you mentioned and I wanted to hear more... next standup topic? :) (FWIW, I'm happy to use something else - I just don't see anything that fits the bill. Bash scripts need a lot of extra work to get the same benefits, imo, but I don't care that much).

PR #1275, opened 2019-11-05
Fix test configuration

andresmgot:
Ref #1209 

Restore the global configuration instead of doing it per test file. The problem was that I was not copying the updated config files to the pod that is running the tests in the CI.


script/e2e-test.sh
```diff
@@ -147,7 +147,8 @@ cd $ROOT_DIR/integration
 kubectl apply -f manifests/executor.yaml
 k8s_wait_for_deployment default integration
 pod=$(kubectl get po -l run=integration -o jsonpath="{.items[0].metadata.name}")
-## Copy latest tests
+## Copy config and latest tests
+for f in `ls *.js`; do kubectl cp ./${f} ${pod}:/app/; done
```
absoludity: This looks pretty broad, it'd be nicer if it was just a directory of config, like the `./use-cases` below.

andresmgot: yes, the problem is that some of the configuration files require to be at the same level than the package.json. In any case, it's safe to even do copy all the files to the pod (because that's what the dockefile does) but I don't do that to speed up the process a bit.

Issue #1235, opened 2019-10-18
Improve the UpgradeForm to migrate values from one version to other


andresmgot:
Follow up from #1182 

As discussed in #1144, it will be useful to improve the current behavior when upgrading charts. By default, when a user selects a new version, the current values get overwritten with the default values of the new version. From the point in which the user modifies any value, changing version takes no effect.

We should investigate a better way of migrating changes between versions and if possible show the difference that will be applied before submitting the upgrade.

andresmgot:
**[PROPOSAL]**

This is my proposal of the possible user experience of a user upgrading the app MyApp from version 1.0.0 to 1.0.1:

 1. When clicking in the "Upgrade" button (and after selecting the source repo if necessary), the user will be presented with the current status of the release: MyApp v1.0.0 with the `values` used when the app was first deployed.
 2. The user clicks on the new version: v1.0.1.
   (option a.) The values are kept the same so changes made in the original release can be reused. 
   (option b.) Instead of just leaving the original values, we parse the original values as a yaml `Document`. We can iterate over the elements of the doc and if the new version contains the same section, overwrite it with the previous values. If not, add it to the end.
 3. We can render a new Tab representing the `diff` between the current values and the default values of the new version (ignore the text in the tabs):

![Screenshot from 2019-10-30 16-57-59](https://user-images.githubusercontent.com/4025665/67875507-b7162b80-fb36-11e9-9f37-803afebf5406.png)
![Screenshot from 2019-10-30 16-46-14](https://user-images.githubusercontent.com/4025665/67875509-b9788580-fb36-11e9-965c-9559d77fc968.png)

 4. When clicking "submit", we can show the diff between the default values of the version to upgrade and the values that will be submitted. The user can do a final review at that moment if that's the desired result.
 5. Submit when confirmed.

Regarding option a. vs. option b. in the step 2, I would go for the option 1 because:

 - It's less prone to errors. With automated changes, we are likely going to add/modify something that we shouldn't. 
 - Upgrades will require human intervention anyway. We don't know if we would be helping with the automated merge.
 - It's easier to implement. If after going with option a. we still want to implement option b. we can do so.

Let me know what you think. cc/ @absoludity 

BTW, for the `Diff` view, there is a useful react package: https://www.npmjs.com/package/react-diff-viewer

absoludity:
> **[PROPOSAL]**
> 
[...]

PR #1277, opened 2019-11-06
Add Restore Defaults button

andresmgot:
Ref: #1235 

Add a button to restore the chart version defaults. I am adding a confirmation modal to avoid accidental clicks on the reset button (which causes the loss of any data introduced):

Screencast2019-11-06163227

I am reusing the `ConfirmDialog` already existing for the confirmation modal, that's why there are changes there.


dashboard/src/components/DeploymentFormBody/DeploymentFormBody.tsx
```diff
@@ -131,6 +142,9 @@ class DeploymentFormBody extends React.Component<
           <button className="button button-primary" type="submit">
             Submit
           </button>
+          <button className="button" type="button" onClick={this.openRestoreDefaultValuesModal}>
+            Restore Defaults Values
```
absoludity: Not sure whether "Restore Chart Defaults" would be clearer? Fine either way.

andresmgot: Right, that's true

PR #1283, opened 2019-11-11
Refactor modifications in upgrade and deployment forms

andresmgot:
Follow up of #1278 

This refactor:
- Adds a new property to the `IChartState`: `deployed`. Instead of using `selected` and taking advantage that the first version selected is the one deployed, now we explicitly store that chart version in a different variable. This requires changes in the Redux store but allows us to pre-select the latest version in the upgrade form (not included in this PR)
- Move the logic to manage `modifications` outside the `BasicDeploymentFormBody`. This procedure is specific to the upgrade flow so it's now in the `UpgradeForm` component.
- Since the process of changing the `appValues` when changing the version is different between the deployment and the upgrade form, that's been extracted from the `BasicDeploymentFormBody` as well. In the case of the deployment, we simply apply the default values of the new version while in the upgrade scenario we also apply modifications if they exist.


dashboard/src/components/AppUpgrade/AppUpgrade.test.tsx
```diff
@@ -218,32 +219,52 @@ it("skips the repo selection form if the app contains upgrade info", () => {
...
+
+describe("when receiving new props", () => {
+  it("should set the source repository in the state if present as a property", () => {
+    const repo = { metadata: { name: "stable" } };
+    const wrapper = shallow(<AppUpgrade {...defaultProps} />);
+    wrapper.setProps({ repo });
+    expect(wrapper.state("repo")).toEqual(repo);
+  });
+
+  it("should request the deployed chart when teh app and repo are populated", () => {
```
absoludity: Tiny typo - s/teh/the/

dashboard/src/actions/charts.ts
```diff
@@ -122,10 +131,25 @@ export function getChartVersion(
...
     } catch (e) {
       dispatchError(dispatch, e);
     }
+    return;
+  };
+}
+
+export function getDeployedChartVersion(
+  id: string,
+  version: string,
```
absoludity: Is there some confusion here about a chart version vs a chart revision, or are we defining a function called `getDeployedChartVersion` which takes the chart version as input? *confused*

dashboard/src/components/AppUpgrade/AppUpgrade.tsx
```diff
@@ -39,26 +40,71 @@ interface IAppUpgradeProps {
...
 class AppUpgrade extends React.Component<IAppUpgradeProps, IAppUpgradeState> {
+  public state: IAppUpgradeState = {};
+
   public componentDidMount() {
     const { releaseName, getAppWithUpdateInfo, namespace, fetchRepositories } = this.props;
     getAppWithUpdateInfo(releaseName, namespace);
     fetchRepositories();
   }
 
+  public componentWillReceiveProps(nextProps: IAppUpgradeProps) {
```
absoludity: We seem to keep adding more `componentWillReceiveProps` when we should be removing them as we go, not adding?

See https://reactjs.org/docs/react-component.html#updating

The issue here is that we're adding more code that is dependent on detecting a change in the prop values, rather than working towards the functional `getDerivedStateFromProps`. Even if we can't use that here (it's added in 16.3), let's not add code which we're going to have to remove/update in the future? We should be considering the reasons why react is removing `componentWillReceiveProps` (ie. so that people won't be deriving state based on existing state or old prop values).

dashboard/src/components/DeploymentForm/DeploymentForm.tsx
```diff
@@ -43,12 +43,18 @@ export interface IDeploymentFormState {
...
     isDeploying: false,
     latestSubmittedReleaseName: "",
     valuesModified: false,
   };
 
+  public componentWillReceiveProps(nextProps: IDeploymentFormProps) {
+    if (nextProps.selected.version !== this.props.selected.version && !this.state.valuesModified) {
+      this.setState({ appValues: nextProps.selected.values || "" });
+    }
+  }
```
absoludity: Again, this looks like we're adding more code which calculates state by detecting changes in props, rather than simply a function of the state... I worry we're going to pay for this later when `componentWillReceiveProps` is no longer supported (it'll be removed in 1.17).

Can we first update to React 16.3 (at least, but possibly latest - 16.11), before we add more code like this? At least then we'll be able to use `getDerivedStateFromProps` and will be forced to add `UNSAFE_` to these unsafe methods we're adding?

dashboard/src/components/UpgradeForm/UpgradeForm.tsx
```diff
@@ -41,6 +46,28 @@ class UpgradeForm extends React.Component<IUpgradeFormProps, IUpgradeFormState>
     valuesModified: false,
   };
 
+  public componentWillReceiveProps(nextProps: IUpgradeFormProps) {
```
absoludity: Ditto for `getDerivedStateFromProps`

dashboard/src/components/AppUpgrade/AppUpgrade.tsx
```diff
@@ -39,26 +40,71 @@ interface IAppUpgradeProps {
...
+        // If there is update info about the app we can automatically chose the repository
+        // with the latest version
+        if (app && app.updateInfo) {
+          const repoWithLatest = repos.find(
+            r => r.metadata.name === (app.updateInfo && app.updateInfo.repository.name),
+          );
+          if (repoWithLatest) {
+            this.setState({ repo: repoWithLatest });
+            repo = repoWithLatest;
+            this.setState({ repo });
```
absoludity: Looks like 2 of these lines are redundant?

dashboard/src/components/DeploymentForm/DeploymentForm.tsx
absoludity: Sounds good.

dashboard/src/components/AppUpgrade/AppUpgrade.tsx
```diff
@@ -39,26 +40,71 @@ interface IAppUpgradeProps {
...
 
+  public componentDidUpdate(prevProps: IAppUpgradeProps) {
+    const { repos, app } = this.props;
+    let repo = this.state.repo;
+    // Retrieve the current repo
+    if (!repo) {
+      repo = this.props.repo;
+      if (repo && repo.metadata) {
+        // If the repository comes from the properties, use it
+        this.setState({ repo });
```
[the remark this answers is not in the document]
andresmgot: Exactly, the `repo` will be only available in the `props` when using the `SelectRepoForm` if not, it's obtained from the app info. For simplicity, I stored it in the state.

dashboard/src/components/AppUpgrade/AppUpgrade.tsx
andresmgot: oops, yes

PR #1285, opened 2019-11-13
Local oauth dev with dex

absoludity:
This PR enables a `deploy-dev` target which, combined with the `cluster-kind` target, enables an environment with:
 * api server configured with the oidc plugin using the issuer url of https://dex.dex:3200
 * dex running in the cluster configured with two static users - kubeapps-operator@example.com and kubeapps-user@example.com
 * kubeapps configured to use oidc with the in-cluster dex.dex provider

As the browser also has to be able to access the oidc provider at dex.dex, it additionally requires the user to add `127.0.0.1 dex.dex` to their /etc/hosts. But otherwise, devs can then login as either user to test the experience as that user.

Note: the user `kubeapps-user@example.com` has access to the `kubeapps-user-namespace` namespace only, not `default` (which highlights a real user issue, in that we don't display the users allowed namespaces in the dropdown... though when using a service token we do display and default to the token's namespace).

Ref: #1269


docs/user/manifests/kubeapps-local-dev-users-rbac.yaml
```diff
@@ -0,0 +1,34 @@
...
+metadata:
+  name: kubeapps-operator
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-admin
+subjects:
+- apiGroup: rbac.authorization.k8s.io
+  kind: User
+  name: oidc:kubeapps-operator@example.com
```
andresmgot: I guess you need to set the prefix `oidc:` here because you have configured the API with that prefix. Is that intentional to be more explicit? I have it without prefix in my dev cluster, just in case you prefer to do that without the prefix.

script/deploy-dev.mk
```diff
@@ -0,0 +1,52 @@
...
+
+# The api server does not have service dns entries (in kind or vanilla k8s), so
+# dex is normally required to be available on an external host. Short-circuit
+# that requirement by ensuring dex.dex resolves to the internal IP address on
+# the apiserver so that it can initialise the oidc plugin.
+update-apiserver-etc-hosts:
+	while ! kubectl -n kube-system get po kube-apiserver-kubeapps-control-plane; do \
+		echo "Waiting for api server" && sleep 1; \
+	done
+	kubectl -n kube-system exec kube-apiserver-kubeapps-control-plane -- \
```
andresmgot: I am curious, it doesn't work if you add an ingress to dex, right? The K8s API can't resolve ingress URLs?

docs/user/manifests/kubeapps-local-dev-users-rbac.yaml
absoludity: Yeah, it's intentional (and was recommended) so it's clear in k8s where the authn is coming from (when assigning RBAC etc.)

script/deploy-dev.mk
absoludity: The resolv.conf used on the api-server pod doesn't have any internal nameserver - see https://github.com/kubernetes-sigs/kind/issues/1077 where I dumped /etc/resolv.conf etc. So if I used an actual k8s cluster, I could create an ingress with an externally resolvable hostname and it would work, but I can't see how that can be used for a local setup?

What I was able to do is set the issuer-url for the apiserver to `https://localhost/api/v1/ser..../proxy` which worked fine for the api server, but the issuer-url needs to match the issuer url specified by dex, which cannot be reconciled because the api server proxy url proxies through to the root of the dex service... if dex is configured with the issuerurl as https://localhost/ the apiserver fails because the issuer urls don't match, and if you make the dex one match, then the proxied request 404s (as it's hitting the root /.well-known/..., while dex is serving from /api/v1/.../proxy).

andresmgot: I see, just wondering if I made this work with Minikube changing the `/etc/hosts` or with an ingress (probably the former but I don't remember). Thanks for the info!

PR #1286, opened 2019-11-13
Add diff view

andresmgot:
Ref #1235

Add a new tab in which we show the difference between the default values (or the values of the previous version) and the current values.

Example when deploying:
Screenshot from 2019-11-13 16-13-57

Example when upgrading:
Screenshot from 2019-11-13 16-12-05

If there are no changes:
Screenshot from 2019-11-13 16-18-31

I have changed the titles of the tabs to be more descriptive:
 - Basic --> Form
 - Advanced --> Values (YAML) (I have removed the label in that tab to avoid duplicated text)
 - Changes

At the end, I am using the package `diff2html` for the diff view because I hit an issue with the stable version of `react-diff-viewer`: https://github.com/praneshr/react-diff-viewer/issues/47


dashboard/src/components/DeploymentFormBody/DeploymentFormBody.tsx
```diff
@@ -288,6 +288,33 @@ class DeploymentFormBody extends React.Component<
...
+  private renderDiff = () => {
+    let oldValues = "";
+    let title = "";
+    let emptyDiffText = "";
+    if (this.props.deployedValues) {
+      // If there are already some deployed values (upgrade scenario)
+      // We compare the values from the old release and the new one
+      oldValues = this.props.deployedValues;
+      title = "New values to submit";
+      emptyDiffText = "Values for the new release are identical";
```
absoludity: "The values for the new release are identical to the deployed version." ?

dashboard/src/components/DeploymentFormBody/DeploymentFormBody.tsx
```diff
@@ -288,6 +288,33 @@ class DeploymentFormBody extends React.Component<
...
+
+  private renderDiff = () => {
+    let oldValues = "";
+    let title = "";
+    let emptyDiffText = "";
+    if (this.props.deployedValues) {
+      // If there are already some deployed values (upgrade scenario)
+      // We compare the values from the old release and the new one
+      oldValues = this.props.deployedValues;
+      title = "New values to submit";
```
absoludity: "Difference from deployed version"? (matching how you've labelled the new deployment text below)

PR #1287, opened 2019-11-14
Initial LDAP support in local dev

### LINKED ISSUE #1269: Investigate and demo LDAP for Kubeapps

absoludity:
This PR adds an OpenLDAP deployment so that in addition to the login-by-email added previously, you can also login via LDAP.

Note: This is a WIP as something in this diff results in the api server being unable to verify signatures, so logins fail. I see the following in the apiserver logs:
```
4T05:38:30.142068621Z stderr F E1114 05:38:30.141968       1 authentication.go:65] Unable to authenticate the request due to an error: [invalid bearer token, oidc: verify token: failed to verify signature: failed to verify id token signature]
```
but haven't yet found the cause. Nonetheless, I think it's ready for review otherwise.

Fixes #1269


absoludity:
> Note: This is a WIP as something in this diff results in the api server being unable to verify signatures, so logins fail. I see the following in the apiserver logs:
> 
> ```
> 4T05:38:30.142068621Z stderr F E1114 05:38:30.141968       1 authentication.go:65] Unable to authenticate the request due to an error: [invalid bearer token, oidc: verify token: failed to verify signature: failed to verify id token signature]
> ```
> 
> but haven't yet found the cause. Nonetheless, I think it's ready for review otherwise.

I came to debug this this morning and it is working perfectly - I can log in either via LDAP or just the dex static user and get assigned the correct RBAC. I'll check later when re-creating whether it's related to timing of the keys (ie. they obviously hadn't expired, but perhaps clock-skew put it in the future).

> I am surprised dex works for both, with a connector and static password. Let me know if I can help with something

Yeah, it's pretty neat.

Issue #1200, opened 2019-10-02
Update React version and migrate deprecated hooks


andresmgot:
From: https://github.com/kubeapps/kubeapps/pull/1196/files#r330337078

Hooks like `componentWillReceiveProps` are now deprecated and will be deleted soon. We will need to change those methods and also update the current React version in the way.

PR #1289, opened 2019-11-14
Update React version

andresmgot:
Ref #1200

Update React and some other dependencies.

Note that with the latest React (16.11) the component `componentWillReceiveProps` is still usable (it will be removed in the next major version.

Apart from that, there are few linter issues I needed to fix and automatic changes made by prettier (second commit).

[EDIT]

I found some warnings in the console when running the tests due to old versions of `react-ace` and `react-router` so the last two commits update those. It's better to review commit per commit, ignoring the second one since those are automatic changes.


absoludity:
> Note that with the latest React (16.11) the component `componentWillReceiveProps` is still usable (it will be removed in the next major version.

But importantly, we're forced to realise this when using it by prefixing with `UNSAFE_` (same for `componentWillMount`). Great.

Issue #1249, opened 2019-10-24
[oauth2_proxy] Modify the default login page


andresmgot:
The current login page is bit simple (and maybe confusing?):

![Screenshot from 2019-10-24 12-00-23](https://user-images.githubusercontent.com/4025665/67475129-e66bfa80-f655-11e9-800a-8f9df09bb800.png)

It doesn't give any info that you are login into Kubeapps. It's possible to modify this default view setting the environment variable: `OAUTH2_PROXY_CUSTOM_TEMPLATES_DIR` pointing to a folder with two templates `sign_in.html` and `error.html`.

Taking the source from https://github.com/bitly/oauth2_proxy/blob/master/templates.go, I made a quick test that it works as expected and I included the Kubeappps logo:

![Screenshot from 2019-10-24 12-03-08](https://user-images.githubusercontent.com/4025665/67475341-48c4fb00-f656-11e9-8907-0a95c6a9cb6b.png)

The issue is how we can provide these two files to the deployment without modifying the source docker image (maybe a public S3 bucket mounted as a volume?)

andresmgot:
I guess that's something doable, we just need to forge the proper template

andresmgot:
@absoludity what do you think is better here: build an image `FROM` oauth2_proxy and add the custom page or use an `initContainer` to download the custom page and place it conveniently?

The first option is offline-friendly but it force us to maintain the "forked" image of oauth2_proxy.

absoludity:
I hadn't looked, but didn't think either of those would be necessary. I assumed that we'd be whitelisting the login page and just updating our template there so that it includes the button when configured. Did you check what harbor are doing for the same effect? I assume it'll be something like that.

andresmgot:
I tried modifying the ingress rule for the login page but no white-listing the URL. I will try that and let you know.

PR #1291, opened 2019-11-18
Enable logout when using OIDC for auth

absoludity:
Updates the Header so that it no longer hides the logout link when authenticated via oidc. Instead, an actual anchor link is used to ensure that the SPA doesn't handle the route.

There's an issue which I'll address in a follow-up, in that once the SPA has loaded, it automatically handles `/ -> /#/`, so the second logout works and redirects back to `/` but the app then handles this as usual, no longer hitting the auth-proxy.

We could whitelist the request for `/config.json` but this may not be enough... I'll investigate further as part of #1249.

### DIFF
--- a/dashboard/src/components/Header/Header.test.tsx
+++ b/dashboard/src/components/Header/Header.test.tsx
@@ -1,7 +1,6 @@
 import { shallow } from "enzyme";
 import * as React from "react";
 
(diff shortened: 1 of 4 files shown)


absoludity:
I'm abandoning this one in favour of another cleanup of the cookie-based auth.

PR #1294, opened 2019-11-18
Don't use or store id_token when using cookie-based auth

absoludity:
This is a prequel to re-enabling the logout option for oidc. When I first re-enabled logout, I found that the cookie would be removed as would the id_token (from localStorage), but then the client would just re-get the id_token from the server (`fetchOIDCToken`) and therefore assume it was logged in again.

One of the benefits of using OAuth/OIDC is that the client doesn't need to ever see (or store) the actual access token... the http-only cookie is all that is required (and the auth-proxy adds the id-token to the request), which is not only much safer, but means we can logout also :)

### DIFF
--- a/dashboard/src/shared/Auth.test.ts
+++ b/dashboard/src/shared/Auth.test.ts
@@ -62,7 +62,7 @@ describe("Auth", () => {
   describe("isAuthenticatedWithCookie", () => {
     it("returns true if request to API root succeeds", async () => {
       Axios.get = jest.fn(path => {
-        return { headers: { status: 200 } };
+        return Promise.resolve({ headers: { status: 200 } });
       });
       const isAuthed = await Auth.isAuthenticatedWithCookie();
 
--- a/dashboard/src/shared/Auth.test.ts
+++ b/dashboard/src/shared/Auth.test.ts
@@ -61,23 +61,35 @@ describe("Auth", () => {
 
   describe("isAuthenticatedWithCookie", () => {
     it("returns true if request to API root succeeds", async () => {
-      Axios.head = jest.fn(path => {
+      Axios.get = jest.fn(path => {
         return { headers: { status: 200 } };
       });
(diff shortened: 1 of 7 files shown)


dashboard/src/shared/Auth.ts
```diff
@@ -65,21 +68,23 @@ export class Auth {
...
+  public static async isAuthenticatedWithCookie(): Promise<boolean> {
     try {
-      const { headers } = await Axios.head("");
-      if (headers && headers.authorization) {
-        const tokenMatch = (headers.authorization as string).match(/Bearer\s(.*)/);
-        if (tokenMatch) {
-          return tokenMatch[1];
-        }
-      }
+      await Axios.get(APIBase + "/");
```
andresmgot: this should be `Axios.head`, no? (or maybe the comment is wrong)

dashboard/src/shared/Auth.test.ts
```diff
@@ -58,29 +59,42 @@ describe("Auth", () => {
...
 
-  describe("fetchOIDCToken", () => {
-    it("should fetch a token", async () => {
-      Axios.head = jest.fn(path => {
-        expect(path).toEqual("");
-        return { headers: { authorization: "Bearer foo" } };
+  describe("isAuthenticatedWithCookie", () => {
+    it("returns true if request to API root succeeds", async () => {
+      Axios.get = jest.fn(path => {
+        return { headers: { status: 200 } };
```
andresmgot: nit: this should probably return a promise too (to match the other tests)

absoludity: Updated.

dashboard/src/shared/Auth.ts
absoludity: The comment is wrong (need to check the body of the error response), but already fixed in the following branch.

PR #1295, opened 2019-11-19
Enable logout when logged in via cookie/oauth

absoludity:
This PR un-hides the logout action (it was previously explicitly hidden when logged in via oauth).

The `Auth.unsetAuthCookie` does the work (just redirecting to the configured logout point which removes the cookie - it's impossible to remove the cookie from the client-side JS as it is http-only).

A lot of fun was had (sarcasm) updating the `isAuthenticatedWithCookie` function to handle all cases. It's basically tricky because the the client can't detect an http-only cookie at all.

The result looks like this (sorry, low res):
oidc-login-logout

This is a second prequel to #1249 (after #1294).

### DIFF
--- a/dashboard/public/config.json
+++ b/dashboard/public/config.json
@@ -1,4 +1,5 @@
 {
   "namespace": "default",
-  "appVersion": "DEVEL"
+  "appVersion": "DEVEL",
+  "logoutURI": "/oauth2/sign_out"
 }
--- a/dashboard/src/shared/Auth.ts
+++ b/dashboard/src/shared/Auth.ts
@@ -2,6 +2,7 @@ import Axios, { AxiosResponse } from "axios";
 import * as jwt from "jsonwebtoken";
 const AuthTokenKey = "kubeapps_auth_token";
 const AuthTokenOIDCKey = "kubeapps_auth_token_oidc";
+import { IConfig } from "./Config";
 import { APIBase } from "./Kube";
 
 export const DEFAULT_NAMESPACE = "default";
@@ -20,7 +21,14 @@ export class Auth {
 
   public static unsetAuthToken() {
     localStorage.removeItem(AuthTokenKey);
+  }
+
+  public static unsetAuthCookie(config: IConfig) {
+    // http cookies cannot be deleted (or modified or read) from client-side
+    // JS, so force browser to load the sign-out URI (which expires the
+    // session cookie).
     localStorage.removeItem(AuthTokenOIDCKey);
+    document.location.assign(config.logoutURI);
   }
 
   public static usingOIDCToken() {
--- a/dashboard/src/shared/Auth.ts
+++ b/dashboard/src/shared/Auth.ts
@@ -93,17 +93,14 @@ export class Auth {
       // A non json 403 error response means we did not get through to the API server but
       // instead were rejected by the auth proxy (ie. no http-only cookie).
       // TODO(mnelson): Check why doesn't the auth proxy return a 401 for a request without auth?
-      if (response.headers && response.headers["content-type"] !== "application/json") {
+      if (!response.data || !response.data.message) {
         return false;
       }
       // Finally, the k8s api server nowadays defaults to allowing anonymous
       // requests, so that rather than returning a 401, a 403 is returned if
       // RBAC does not allow the anonymous user access. An http-only cookie
       // will not result in an anonymous request, so...
-      const isAnon =
-        response.data &&
-        response.data.message &&
-        response.data.message.includes("system:anonymous");
+      const isAnon = response.data.message.includes("system:anonymous");
       return !isAnon;
     }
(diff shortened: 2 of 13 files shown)


dashboard/public/config.json
```diff
@@ -1,4 +1,5 @@
 {
   "namespace": "default",
-  "appVersion": "DEVEL"
+  "appVersion": "DEVEL",
+  "logoutURI": "/oauth2/sign_out"
```
andresmgot: I am not sure why the logoutURI is configurable? Isn't that related to oauth2_proxy?

dashboard/src/shared/Auth.ts
```diff
@@ -68,21 +76,35 @@ export class Auth {
...
+      // but RBAC doesn't authorize us).
+      if (response.status !== 403) {
+        return false;
+      }
+      // A non json 403 error response means we did not get through to the API server but
+      // instead were rejected by the auth proxy (ie. no http-only cookie).
+      // TODO(mnelson): Check why doesn't the auth proxy return a 401 for a request without auth?
+      if (response.headers && response.headers["content-type"] !== "application/json") {
+        return false;
+      }
```
andresmgot: is this condition needed? If the response is not an `application/json`, below `response.data.message` will be undefined, right?

absoludity:
> In the gif I am seeing the login page twice, first while we validate the token and then after clicking logout. This may be confusing.
> 
> For the first we can add a loading view but I think that's the current status (it's not related to these changes so we can do that in a different PR/issue).

Yep, the first one is exactly that - `/` -> `/login` but also dispatches a request to check the auth (previously while fetching the oidc token, now while checking auth), so the login form is displayed briefly. Thanks - planning to fix that in another PR.

> 
> The second seems to be permanent, once logout, I would expect the user to be redirected to oauth2_proxy again, isn't that the case?

If I was stopping here, it would be odd, but in the next PR I'm going to add `^/$` to the `skip_auth_regex` so that the app can load and display the login screen initially (rather than the blank screen with one button), and add the button there to initiate the oauth login. The nice thing is, I think we'll be able to support both (ie. you can still login with a token, or click the oauth login). So for that reason I didn't bother forcing a reload (ie. we could check if the referrer is `/oauth2/sign_out` when loading the routes and if so force a reload).

dashboard/public/config.json
absoludity: Yes, but using oauth2_proxy on the kubeapps frontend is just the option we support in the chart - you could also deploy your own oauth2 proxy or whatever on a different host and need a full URI here. That was my thinking anyway, happy to remove it if you think so.

dashboard/src/shared/Auth.ts
absoludity: > If the response is not an application/json, below response.data.message will be undefined, right?

Yes, which would mean that isAnon would be false and the current code assumes it is authed (ie. a 403 that isn't anonymous). I could replace the above condition with one based on `response.data` as you say  - perhaps that'd be simpler? Something like:
```
if (response.data === undefined) {
  return false;
}
const isAnon = response.data.message && ...
```

dashboard/public/config.json
andresmgot: I see, that's fine (I was unsure if that was related to the IdP). We don't support anything else apart than oauth2_proxy but I am fine leaving that.

dashboard/src/shared/Auth.ts
andresmgot: ah, okay, for a moment I thought that with `isAnon = false` we were returning `false`.

Then yes, I'd rather check if the response has data  (and a messsage):

```
if (!response.data || !response.data.message) {
  return false;
}
``` 

Also this allows you to simple return `return response.data.message.includes("system:anonymous");` below since we are already checking that `data` and `message` are defined.

andresmgot:
> If I was stopping here, it would be odd, but in the next PR I'm going to add ^/$ to the skip_auth_regex so that the app can load and display the login screen initially (rather than the blank screen with one button), and add the button there to initiate the oauth login

okay, sounds good to me.

> The nice thing is, I think we'll be able to support both (ie. you can still login with a token, or click the oauth login).

I wouldn't do so. As a user, if an administrator has configured Kubeapps to work with OIDC, I would be confused if I get asked to introduce a token (even if it's optional). I would show either a login page for OIDC or for the token but not both.

dashboard/src/shared/Auth.ts
absoludity: Yep, updated - that's neater.



> Also this allows you to simple return `return response.data.message.includes("system:anonymous");` below since we are already checking that `data` and `message` are defined.

fwiw, I've still left the const, so that the return reads as "return not anonymous":
```
const isAnon = response.data.message.includes("system:anonymous");
return !isAnon;
```

Thanks

PR #1301, opened 2019-11-20
Move dex-related oauth2 options out of default auth options.

absoludity:
**Description of the change**
I realised with https://github.com/kubeapps/kubeapps/pull/1295/files that I had added changes to the default oauth2-proxy chart install, whereas I intended to just test them there, and add them as specific changes for dex.

This change ensures those dex-related options aren't part of the default install when using oauth2 proxy.

### DIFF
--- a/chart/kubeapps/templates/kubeapps-frontend-deployment.yaml
+++ b/chart/kubeapps/templates/kubeapps-frontend-deployment.yaml
@@ -50,9 +50,6 @@ spec:
         - -pass-basic-auth=false
         - -pass-access-token=true
         - -pass-authorization-header=true
-        - -skip-auth-regex=^\/config\.json$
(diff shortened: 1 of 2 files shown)


andresmgot:
It seems there is a conflict here that need to be fixed

absoludity:
> One question, why is this related to just dex?

You're right - they're not... I'm just doing all my testing with dex as the provider. I was thinking that I wanted to finish the work (ie. the next PR) before changing the chart, but these are really needed in the current state. It's not ideal (I'd wanted to get the next PR done to finish it off, but due to other work didn't get to it today). Worst case with this isn't so bad though... oauth users can logout (where they couldn't before), but will end up on the token login page - requiring hitting a refresh to log back in.

PR #1304, opened 2019-11-20
Fix downgrade when the schema is missing

andresmgot:
**Description of the change**

If found another issue, while downgrading from a version that contains a schema to one that doesn't, the form didn't got removed. This is because in the reducer, we were checking that the schema exists to be set in the state (maintaining the previous state otherwise).

### DIFF
--- a/dashboard/src/reducers/charts.ts
+++ b/dashboard/src/reducers/charts.ts
@@ -24,8 +24,8 @@ const chartsSelectedReducer = (
         error: undefined,
         readmeError: undefined,
         version: action.payload.chartVersion,
-        values: action.payload.values ? action.payload.values : state.values,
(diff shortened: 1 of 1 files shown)


dashboard/src/reducers/charts.ts
```diff
@@ -24,8 +24,8 @@ const chartsSelectedReducer = (
         error: undefined,
         readmeError: undefined,
         version: action.payload.chartVersion,
-        values: action.payload.values ? action.payload.values : state.values,
```
absoludity: Is this the same as:
```
values: action.payload.values || state.values,
```
? But fine either way.

andresmgot: it *was* the same, I changed it to just take into account the payload value

PR #1309, opened 2019-11-21
Add Helm 3 support

SimonAlling:
### Description of the change

This PR aims to introduce support for Helm 3, which does not use Tiller. Currently, it's nowhere near complete, but merely intended as a first step on a not yet charted path.

### Benefits

  * Instead of Tiller, we will use the `actions` package from Helm 3. This will improve security by allowing different users to have different permissions, which isn't possible with Tiller. Because Tiller won't be used anymore, Tiller Proxy will be replaced by a component we call Helmer.
  * We've made, and will make, all additions in such a way that current functionality (including Tiller Proxy and Helm 2 support) shouldn't be affected. This is accomplished by the `useHelm2` flag in `values.yaml`: When it's set to the default value of `true`, Kubeapps works exactly like before; when it's set to `false`, Helmer is used instead of Tiller Proxy.
  * The Dashboard shouldn't need to be modified. It sends its REST calls to the same URLs as before; the responses just happen to come from Helmer instead of Tiller Proxy.

### Possible drawbacks

  * Currently, the only thing Helmer can do is list releases (in a specific namespace or in all namespaces). This was the easiest feature to implement, and it serves fairly well as a PoC on how Helm 3 support could be accomplished.
  * `rest.InClusterConfig` and `kubernetes.NewForConfig` are currently called on every REST request handled by Helmer (in `pkg/agent/agent.go`). We weren't sure how to avoid this without compromising security or stability.
  * Helm 3 can't see releases created by Helm 2, and Helmer and Tiller Proxy can't run simultaneously, at least not with our current architecture.

### Applicable issues

  - #1056
  - #1250 (PR)

#### `WithAgentContext`

It may not be obvious why `WithAgentContext` (in `cmd/helmer/internal/handler/handler.go`) exists or why it's implemented the way it is. We tried to explain it as best we could in the comments, but we'd also like to mention that the current implementation is the result of quite a lot of conscious planning and experimenting, where the goal was to utilize the type checker to eliminate the risk of ever using an invalid context.

We think we succeeded in fulfilling that goal, by realizing that although the types of

```golang
func (x X) f_yz(y Y, z Z) R {}
```

and 

```golang
func f_xyz(x X, y Y, z Z) R {}
```

are essentially isomorphic, they differ in that members of the former are much easier to partially apply:

```golang
partiallyAppliedF := myX.f_yz
```

versus

```golang
partiallyAppliedF := func(y Y, z Z) R { return f_xyz(myX, y, z) }
```

Because we want every HTTP handler to create a new `action.Config` for itself (which cannot be done without an `http.Request` containing a token), we've introduced a type `dependentHandler`, whose members (`ListReleases`, `CreateRelease` etc) take the context as a regular argument (like `f_xyz` above) instead of as a "`this` argument" (`f_yz`).

We can then "enforce" the use of our `WithAgentContext` function (by virtue of simplicity), and minimize the risk of creating and using an invalid `agent.Context` and/or `action.Config`, e.g. the "zero context" `agent.Context{}`. An example of this can be seen in `cmd/helmer/main.go`, which also shows why `WithAgentContext` is written in a curried fashion:

```golang
withAgentContext := handler.WithAgentContext(options)

apiv1 := r.PathPrefix("/v1").Subrouter()
apiv1.Methods("GET").Path("/releases").Handler(negroni.New(
    negroni.Wrap(withAgentContext(handler.ListAllReleases)),
))
apiv1.Methods("GET").Path("/namespaces/{namespace}/releases").Handler(negroni.New(
    negroni.Wrap(withAgentContext(handler.ListReleases)),
))
```

#### `useHelm2`

We're not sure about this flag; it feels somewhat counterintuitive to set it to `false` to enable a _new_ feature. We're definitely open to inverting its semantics (together with its name) if you guys prefer that.

#### Trying Helmer

This is how we deploy Helmer in our development environment:

```bash
# Prepare Kubernetes:
minikube start --memory 4000
kubectl create serviceaccount kubeapps-operator
kubectl create clusterrolebinding kubeapps-operator --clusterrole=cluster-admin --serviceaccount=default:kubeapps-operator

# Extract token for later use:
kubectl get secret $(kubectl get serviceaccount kubeapps-operator -o jsonpath='{.secrets[].name}') -o jsonpath='{.data.token}' | base64 --decode && echo

# Prepare Kubeapps:
cd $KUBEAPPS_DIR/chart
sed -i 's/useHelm2: true/useHelm2: false/' kubeapps/values.yaml # or edit the file manually
helm version # Helm 3
helm package kubeapps
cd ..
eval $(minikube docker-env)
IMAGE_TAG=dev make kubeapps/helmer

# Install Kubeapps:
kubectl create namespace kubeapps
helm install kubeapps chart/kubeapps-3.0.3.tgz --namespace kubeapps
kubectl patch deployment kubeapps-internal-helmer -n kubeapps --type=json -p='[{"op": "replace", "path": "/spec/template/spec/containers/0/imagePullPolicy", "value": "IfNotPresent"}]'
kubectl set image -n kubeapps deployment kubeapps-internal-helmer agent=kubeapps/helmer:dev

# Install example charts:
helm repo add bitnami https://charts.bitnami.com/bitnami
helm install silent-parrot bitnami/nginx
helm install vicious-rabbit bitnami/nginx

# Make Dashboard accessible in the browser:
export POD_NAME=$(kubectl get pods --namespace kubeapps -l "app=kubeapps" -o jsonpath="{.items[0].metadata.name}")
kubectl port-forward --namespace kubeapps $POD_NAME 8080:8080
```

The Dashboard with the example charts should now be accessible at `http://localhost:8080`.

After making changes in Helmer:

```bash
eval $(minikube docker-env) # only necessary once per terminal
IMAGE_TAG=dev make kubeapps/helmer
kubectl delete pod -n kubeapps -l app=kubeapps-internal-helmer
# (Helmer should restart in a few moments)
```

#### Multiple commits

Currently, our `helm3` branch consists of four commits. The first two might be relevant even without Helm 3, and it should be possible to cherry-pick them.

#### PR status

Note that, as previously stated, the PR in its current form is **not intended to be merged**; there is still quite a lot of work to do. Feel free to share any thoughts and questions you might have!


SimonAlling:
FYI: @latiif, @lindhe, @NiclasWarefelt and I will be traveling next week, so we might only be able to reply sporadically from Monday on.

pkg/agent/agent.go
```diff
@@ -0,0 +1,94 @@
...
+		panic(err.Error())
+	}
+	config.BearerToken = token
+	config.BearerTokenFile = ""
+	clientset, err := kubernetes.NewForConfig(config)
+	store := createStorage(os.Getenv(driverEnvVar), namespace, clientset)
+	actionConfig.RESTClientGetter = nil
+	actionConfig.KubeClient = kube.New(nil)
+	actionConfig.Releases = store
+	actionConfig.Log = klog.Infof
```
absoludity: Out of interest, why do you want to store the logger in the action config?

SimonAlling: I don't know. [It's part of the `action.Configuration` type](https://github.com/helm/helm/blob/865c46c014cdb7622f97ae287ee92fb8a280f3a9/pkg/action/action.go#L84) from Helm 3.

cmd/helmer/internal/handler/handler.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+	response.NewDataResponse(apps).Write(w)
+}
+
+func ListAllReleases(ctx agent.Context, w http.ResponseWriter, req *http.Request, _ handlerutil.Params) {
+	ListReleases(ctx, w, req, map[string]string{namespaceParam: ""})
+}
+
+// A best effort at extracting the actual token from the Authorization header.
+// We assume that the token is either preceded by tokenPrefix or not preceded by anything at all.
```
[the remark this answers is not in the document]
SimonAlling: I thought maybe sometimes there might just be a raw token. I think [this line](https://github.com/kubeapps/kubeapps/blob/cc9c2d6835a7e4efef5bfe8b8c582cb43d84c55c/chart/kubeapps/templates/kubeapps-frontend-config.yaml#L78) may have played a part in my reaching that conclusion. I'll not object to assuming a `"Bearer "` prefix, though.

cmd/helmer/main.go
```diff
@@ -0,0 +1,118 @@
...
+	userAgentComment string
+	listLimit        int
+	timeout          int64
+)
+
+func init() {
+	settings.AddFlags(pflag.CommandLine) // necessary???
+	pflag.StringVar(&chartsvcURL, "chartsvc-url", "https://kubeapps-internal-chartsvc:8080", "URL to the internal chartsvc")
+	pflag.IntVar(&listLimit, "list-max", 256, "maximum number of releases to fetch")
+	pflag.StringVar(&userAgentComment, "user-agent-comment", "", "UserAgent comment used during outbound requests") // necessary???
```
andresmgot: context on why the `user-agent` is important: https://github.com/kubeapps/kubeapps/issues/767

cmd/helmer/main.go
```diff
@@ -0,0 +1,118 @@
...
+	chartsvcPrefix := "/chartsvc"
+	chartsvcRouter := r.PathPrefix(chartsvcPrefix).Subrouter()
+	// Logos don't require authentication so bypass that step
+	chartsvcRouter.Methods("GET").Path("/v1/assets/{repo}/{id}/logo").Handler(negroni.New(
+		negroni.Wrap(http.StripPrefix(chartsvcPrefix, chartsvcProxy)),
+	))
+	chartsvcRouter.Methods("GET").Handler(negroni.New(
+		authGate,
+		negroni.Wrap(http.StripPrefix(chartsvcPrefix, chartsvcProxy)),
+	))
```
andresmgot: this is why the service won't only interact with "helm"

SimonAlling:
Thanks for your replies, @absoludity and @andresmgot – much appreciated! We'll try to act on your suggestions as soon as we have time. (I've already fixed some of them, as you can see.)

SimonAlling:
I'd also just like to mention that I don't understand why the CI's still failing, now that I've added the missing vendored packages (`k8s.io/klog` and `helm.sh/helm`). Well, I guess it has something to do with `strings.ReplaceAll` not being introduced until Go 1.12, but I tried creating this `Dockerfile`:

```Dockerfile
FROM circleci/golang:1.11 as builder
COPY . /go/src/github.com/kubeapps/kubeapps
WORKDIR /go/src/github.com/kubeapps/kubeapps
```

Then I did

```
docker run -it --rm funny-rabbit make test
```

and it worked just fine.

andresmgot:
> I'd also just like to mention that I don't understand why the CI's still failing, now that I've added the missing vendored packages (`k8s.io/klog` and `helm.sh/helm`). Well, I guess it has something to do with `strings.ReplaceAll` not being introduced until Go 1.12, but I tried creating this `Dockerfile`:
> 
> ```dockerfile
> FROM circleci/golang:1.11 as builder
> COPY . /go/src/github.com/kubeapps/kubeapps
> WORKDIR /go/src/github.com/kubeapps/kubeapps
> ```
> 
> Then I did
> 
> ```
> docker run -it --rm funny-rabbit make test
> ```
> 
> and it worked just fine.

ah, I see, yes, you need to update the version in the CI as well:

https://github.com/kubeapps/kubeapps/blob/master/.circleci/config.yml#L241

There are several occurences of `circleci/golang:1.11`

cmd/helmer/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM golang:1.13 as builder
+COPY . /go/src/github.com/kubeapps/kubeapps
+WORKDIR /go/src/github.com/kubeapps/kubeapps
+
+ARG VERSION
+RUN CGO_ENABLED=0 go build -a -installsuffix cgo -ldflags "-X main.version=$VERSION" ./cmd/helmer
+
+FROM scratch
+COPY --from=builder /go/src/github.com/kubeapps/kubeapps/helmer /agent
```
lindhe: I've read that using the `USER` directive in every Dockerfile is recommended. I have not had time to insert `USER nobody` in this file and test it, just writing this as a reminder.

latiif:
Hello again @andresmgot and @absoludity!

We are working on implementing the *install chart* (`CreateRelease`) functionality for Helm 3 support.

Looking into `tiller-proxy`, we found the `Resolver` interface defined in `/pkg/chart/chart.go` as seen here:

```Go
type Resolver interface {
	ParseDetails(data []byte) (*Details, error)
	GetChart(details *Details, netClient HTTPClient) (*chart.Chart, error)
	InitNetClient(details *Details) (HTTPClient, error)
}
```
The function `GetChart`  returns a pointer to a `Chart`, as defined in `k8s.io/helm/pkg/proto/hapi/chart`; this type differs from the representation used in the Helm 3 library (in `helm.sh/helm/v3/pkg/chart`).
While the two types seem similar on the surface, they are not interchangeable by the type system.

To avoid reinventing the wheel, we wonder **if a function that converts H2 Charts to H3 Charts can be written**. This would allow us to wrap the already existing code for H2 Charts (including repo management and syncing) and reuse it with H3.
```Go
import  h3  "helm.sh/helm/v3/pkg/chart"
import  h2  "k8s.io/helm/pkg/proto/hapi/chart"

//FromH2ChartToH3Chart Converts H2 Charts into H3 Chart represenation
func FromH2ChartToH3Chart(h2.Chart* h2Chart) (h3.Chart,error) {
	panic("TODO")
}
```

Do you think such a function is possible to write, and, if so, is it a good approach?

absoludity:
Hi there @latiif o/

> Do you think such a function is possible to write, and, if so, is it a good approach?

So I'd be very wary of implementing anything which translates between helm2 and helm3 chart structs, as opposed to relying on the helm library itself to do that for us. Of course, helm doesn't provide a way to convert between structs, but given that we only need read access to the chart in both v2 and v3 struct formats, I think we could instead just use the respective helm loaders for each?

To unblock you and hopefully help keep things moving, I've gone ahead and done this in #1362 , which (after review) is landable in master. Take a look and see what you think.

Thanks again for the helm3 work!

SimonAlling:
I think this should be considered done by virtue of #1349, #1359, #1406, #1420, #1421, #1441, #1443, #1462 _et al_.

PR #1311, opened 2019-11-22
Move HTTP/handler utils to a separate package

SimonAlling:
### Description of the change

This PR moves a few types and (mostly pure) functions to a separate package.

### Benefits

The code in question becomes more reusable, which will help with implementing Helm 3 support down the road (see #1309).

### Possible drawbacks

Currently, this will expose some entities that weren't exposed before (`Params`, `WithParams`, `WithoutParams`, `ErrorCode` and `ErrorCodeWithDefault`). If this is undesired, I'd be happy to create the new package at e.g. `internal/handlerutil` or `cmd/internal/handlerutil` instead of `pkg/handlerutil`. (AFAIK, `pkg/internal/handlerutil` won't work because the moved entities need to be imported in `cmd`.)

### DIFF
--- a/cmd/tiller-proxy/internal/handler/handler_test.go
+++ b/cmd/tiller-proxy/internal/handler/handler_test.go
@@ -19,7 +19,6 @@ package handler
 import (
 	"context"
 	"fmt"
-	"net/http"
 	"net/http/httptest"
 	"reflect"
 	"strings"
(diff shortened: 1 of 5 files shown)


cmd/tiller-proxy/internal/handler/handler_test.go
```diff
@@ -49,7 +50,7 @@ func TestErrorCodeWithDefault(t *testing.T) {
 		{fmt.Errorf("This is an unexpected error"), http.StatusUnprocessableEntity, http.StatusUnprocessableEntity},
 	}
 	for _, s := range tests {
-		code := errorCodeWithDefault(s.err, s.defaultCode)
+		code := handlerutil.ErrorCodeWithDefault(s.err, s.defaultCode)
```
andresmgot: since this test is just testing `ErrorCodeWithDefault`, it's better to move this test to kg/handlerutil/handlerutil_test.go

SimonAlling: Yes, I agree.

SimonAlling:
> My only question is whether we need to replicate the `WithParams` etc. in new code. I mentioned in [this comment](https://github.com/kubeapps/kubeapps/pull/1250#issuecomment-548229456) that I don't see why we have them in tiller proxy (that is, over using standard handler func signature `(w http.ResponseWriter, r *httpRequest)` and a simple function to extract args), but i don't care either way. So if you think it's useful to keep using them, great.

You might be right, @absoludity. I'll see what I can do as soon as I have a usable internet connection again. Can't `make` otherwise ...

andresmgot:
Merging this as it is, if you want to add something else, you can add it in a new PR :)

SimonAlling:
OK, @andresmgot! I tried removing the `WithParams` abstraction yesterday, as suggested by @absoludity. No problem according to the typechecker, but I couldn't get a single test to pass without it.

SimonAlling:
By the way, @absoludity, I think I can confirm (after a lengthy and painful debugging session) that the `WithParams` abstraction _is_ necessary to be able to extract the data we want from the `http.Request`. It just doesn't contain any useful data to extract anymore as soon as we're inside the actual handler (e.g. `CreateRelease`).

absoludity:
> It just doesn't contain any useful data to extract anymore as soon as we're inside the actual handler (e.g. `CreateRelease`).

Thanks for digging there. I'm very interested to know why the request would not contain the same data when passed to `CreateRelease` but can look at that separately.

SimonAlling:
> Thanks for digging there. I'm very interested to know why the request would not contain the same data when passed to `CreateRelease` but can look at that separately.

I may have been wrong in stating that it "doesn't contain any useful data to extract _anymore_" above. It seems at least as likely that it never did to begin with. You can see what I tried on [the `remove-withparams` branch in my fork](https://github.com/SimonAlling/kubeapps/commit/41a8cd89b57fe9362dfabc67f157602ce7907a47).

PR #1312, opened 2019-11-22
Upgrade to Go 1.13

SimonAlling:
I think this will make it easier to use external modules going forward. Notably, this change is among the ones we made in order to be able to import `helm.sh/helm/v3` in #1309.

The `gopkg.in/DATA-DOG/go-sqlmock.v1` changes were made by `go mod tidy`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -163,7 +163,7 @@ run_e2e_tests: &run_e2e_tests
     fi
 gke_test: &gke_test
   docker:
-    - image: circleci/golang:1.11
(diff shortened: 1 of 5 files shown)


andresmgot:
as I said in the other PR, do you mind updating the version in the CI too? to avoid inconsistency between the versions

SimonAlling:
Oh, @andresmgot, I almost forgot: I haven't been able to make `make` cache the dependencies; it downloads everything every time. `go build`, however, doesn't.

Issue #651, opened 2018-09-18
Consider MongoDB replacement


andresmgot:
Right now we are using MongoDB to store the information about the available chart of the existing chart repositories. MongoDB is a bit slow to start which causes failures at startup time (#434) apart from other issues (already resolved: #594, #479, #472). We can consider using a more light-weight solution like Redis or etcd since we don't need a lot of functionality.

Having said that, we may want to keep mongodb around if we plan to use it for any future feature. WDYT @migmartri @prydonius @arapulido ?

prydonius:
Thanks @andresmgot. Given the way we currently use MongoDB, it might make sense to switch to something more lightweight like Redis. etcd would actually be overkill IMO, as we don't need the consistency and high availability for this cache store, and writes are slower than MongoDB or Redis.

Switching to a key-value store might mean we can't really query the way we currently do, e.g. something like: https://github.com/kubeapps/kubeapps/blob/master/cmd/chart-repo/utils.go#L244-L248, but I haven't thought to deeply about how we would do that with Redis.

> we may want to keep mongodb around if we plan to use it for any future feature

Indeed, if we ever need to start storing more complex data and need to do more interesting queries, MongoDB would be useful.

migmartri:
I guess that the question is. Are we planning on using MongoDB for more than just caching charts? If that's the case, Redis might be not a good fit, although I do not have experience with Redis apart of simple caching and content allocated storage nor I am familiar with the go clients.

We should also take into account that Monocular will rely on whatever chartsvc backend has, an in the case of kubeapps hub we will need to maybe maintain both (since we store users?)

andresmgot:
Thanks for the input! I will close this issue for the moment since we are not going to take actions unless we discover new blockers.

migmartri:
Thanks @malejpavouk, so it stays in context I am adding to this discussion the link you provided in Slack https://dba.stackexchange.com/questions/191513/posgtresql-gin-btree-ordering as well as https://www.theguardian.com/info/2018/nov/30/bye-bye-mongo-hello-postgres

@andresmgot @prydonius I am reopening this issue but tagging it with needs-more-evidence so we can discuss it further?

fabianschwarzfritz:
Hi @andresmgot ,

thanks for this overall great tooling and keeping the threats together in this repository.

[...]

PR #1313, opened 2019-11-22
MongoDB replacement proposal

andresmgot:
**Description of the change**

This is a design proposal for https://github.com/kubeapps/kubeapps/issues/651

### DIFF
--- a/docs/architecture/design-proposals/replacing-mongodb.md
+++ b/docs/architecture/design-proposals/replacing-mongodb.md
@@ -1,0 +1,99 @@
+## Description of the problem
+
+Due the problems described here: https://github.com/kubeapps/kubeapps/issues/651 and licensing issues, MongoDB is no longer the best solution for our use case. Apart from that, even though is not the goal of this document, we are evaluating the possible effort needed to support other type of assets in Kubeapps (apart than charts) like operators.
+
+The two items above, require a re-design of two micro-services: `chartsvc` and `chart-repo`. While the changes should be kept minimal, at least for the moment, we should plan in advance to support new asset types.
+
+## Requisites:
+
+ - High Availability. Avoid single point of failure.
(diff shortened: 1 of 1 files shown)


lindhe:
I see a lot of talk about Redis or Etcd, both of which makes are pitched as _key-value stores_. But when eyeing through the "Current usage" section, my guts don't say key-value store when seeing that content. Much of that looks more complex, e.g. Collections/charts. I understand that it's possible, but are Redis/Etcd the right tools?

I've only read about them, and heard a security talk about Redis, so I'm not really familiar working with them. Maybe my hunch about their intended use is way off. :)

andresmgot:
> I see a lot of talk about Redis or Etcd, both of which makes are pitched as _key-value stores_. But when eyeing through the "Current usage" section, my guts don't say key-value store when seeing that content. Much of that looks more complex, e.g. Collections/charts. I understand that it's possible, but are Redis/Etcd the right tools?
> 
> I've only read about them, and heard a security talk about Redis, so I'm not really familiar working with them. Maybe my hunch about their intended use is way off. :)
> 
> Relevant? https://dzone.com/articles/great-redis-misapprehension

Right, in this case, I believe that Redis is a suitable solution because it's closer to a cache rather than a database, which is similar to our use case. It's true that the data is not that simple so it may not be the best approach.

> With the outlined options, I'd put my vote to postgres :)

I agree: Its support for JSON objects and it's similarity to MongoDB makes it a good fit. We still need to take care of the persistence or HA but it's the less disruptive.

andresmgot:
Updated the document with a final proposal

absoludity:
Thanks for the update. There's nothing stopping this proposal landing right?

andresmgot:
> Thanks for the update. There's nothing stopping this proposal landing right?

I was just waiting for a final _lgtm_ or in case anyone has any other feedback. Merging this then.

PR #1315, opened 2019-11-25
Oauth logout login 2

### LINKED ISSUE #1249: [oauth2_proxy] Modify the default login page

absoludity:
**Description of the change**
Enables oauth login to use our normal Kubeapps login page (rather than the rather blank oauth2-proxy default).

**Benefits**

A unified login view for Kubeapps, regardless of whether token login or oauth/oidc is being used.

Also ensures we don't see a flash of the login screen after auth'ing.

**Applicable issues**

  - fixes #1249 

**Additional information**
Enables the root route to pass through the proxy, so that our own login form can trigger the authentication as it currently does for token auth also.

oauth2-login-screen

### DIFF
--- a/chart/kubeapps/templates/kubeapps-frontend-deployment.yaml
+++ b/chart/kubeapps/templates/kubeapps-frontend-deployment.yaml
@@ -70,6 +70,7 @@ spec:
             - -skip-auth-regex=^\/config\.json$
             - -skip-auth-regex=^\/favicon.*\.png$
             - -skip-auth-regex=^\/static\/
+            - -skip-auth-regex=^\/$
             {{- range .Values.authProxy.additionalFlags }}
             - {{ . }}
             {{- end }}
--- a/dashboard/public/config.json
+++ b/dashboard/public/config.json
@@ -1,5 +1,7 @@
 {
   "namespace": "default",
   "appVersion": "DEVEL",
+  "authProxyEnabled": true,
+  "loginURI": "/oauth2/start",
   "logoutURI": "/oauth2/sign_out"
 }
--- a/dashboard/public/config.json
+++ b/dashboard/public/config.json
@@ -1,7 +1,7 @@
 {
   "namespace": "default",
   "appVersion": "DEVEL",
-  "authProxyEnabled": true,
+  "authProxyEnabled": false,
   "loginURI": "/oauth2/start",
   "logoutURI": "/oauth2/sign_out"
 }
--- a/dashboard/src/components/LoginForm/LoginForm.tsx
+++ b/dashboard/src/components/LoginForm/LoginForm.tsx
@@ -10,6 +10,7 @@ interface ILoginFormProps {
   authenticated: boolean;
   authenticating: boolean;
   authenticationError: string | undefined;
+  loginURI: string;
   authenticate: (token: string) => any;
   checkCookieAuthentication: () => void;
   location: Location;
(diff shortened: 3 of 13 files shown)


dashboard/public/config.json
```diff
@@ -1,5 +1,7 @@
 {
   "namespace": "default",
   "appVersion": "DEVEL",
+  "authProxyEnabled": false,
+  "loginURI": "/oauth2/start",
   "logoutURI": "/oauth2/sign_out"
```
andresmgot: now that I think about this, maybe it's a bit confusing to call this `loginURI` (or `logoutURI`) since kubeapps has alredy URIs for the login (`/login`), maybe it's better to call it `oauthLoginURI`? Feel free to do that in a different PR or ignore it.

chart/kubeapps/templates/kubeapps-frontend-deployment.yaml
```diff
@@ -70,6 +70,7 @@ spec:
             - -skip-auth-regex=^\/config\.json$
             - -skip-auth-regex=^\/favicon.*\.png$
             - -skip-auth-regex=^\/static\/
+            - -skip-auth-regex=^\/$
             {{- range .Values.authProxy.additionalFlags }}
```
andresmgot: do we need to skip the `/login` route as well?

dashboard/src/components/LoginForm/LoginForm.tsx
```diff
@@ -97,6 +71,64 @@ class LoginForm extends React.Component<ILoginFormProps, ILoginFormState> {
...
+          Your cluster operator has enabled login via an authentication provider.{" "}
+          <a
+            href="https://github.com/kubeapps/kubeapps/blob/master/docs/user/using-an-OIDC-provider.md"
+            target="_blank"
+          >
+            Click here
+          </a>{" "}
+          for more info about using authentication providers with Kubeapps.
+        </p>
+        <a href={this.props.loginURI} className="button button-accent">
```
andresmgot: how is this a link and not a button? (it's okay if it's the same, I am just curious)

chart/kubeapps/templates/kubeapps-frontend-deployment.yaml
absoludity: No - at least, not currently, because we're using a hash-based url scheme. So the `/login` route is actually `/#/login`, which as far as oauth2proxy is concerned, is the `/` path (like all the app routes), even on a refresh.

dashboard/public/config.json
absoludity: Yep, will do.

dashboard/src/components/LoginForm/LoginForm.tsx
absoludity: Do you mean why, rather than how? It's a plain link both in function (ie. just linking to a URI not handled by the app, without the `/#/...` hash uri). But yes, the class ensures it's the same in appearance.

andresmgot: yes, well, I was just surprised that a link with the button classes looks exactly the same than the button

PR #1316, opened 2019-11-25
Include usage information for the oauth2-proxy field 'cookieSecret'

absoludity:
**Description of the change**

Include information for `cookieSecret` option of the auth proxy (including how to generate it).

**Additional information**
As per https://kubernetes.slack.com/archives/C9D3TSUG4/p1574425334139300?thread_ts=1574335985.136900&cid=C9D3TSUG4

cc @lindhe

### DIFF
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -491,7 +491,13 @@ authProxy:
   provider:
   clientID:
   clientSecret:
+  ## cookieSecret is used by oauth2-proxy to encrypt any credentials so that it requires
+  ## no storage. Note that it must be a particular number of bytes. Recommend using the
+  ## following to generate a cookieSecret as per the oauth2 configuration documentation
+  ## at https://pusher.github.io/oauth2_proxy/configuration :
+  ## python -c 'import os,base64; print base64.urlsafe_b64encode(os.urandom(16))'
   cookieSecret:
+  ## Use "example.com" to restrict logins to emails from example.com
   emailDomain: "*"
   ## Additional flags for oauth2-proxy
   ##


lindhe:
While we're on it: is it a good idea to leave this as a mandatory field in `values.yaml`? Feels like a bad idea to code a random cookie into the chart, instead of providing it upon deploy time (e.g., `helm install kubeapps [...] --set authProxy.cookieSecret=$(python -c 'import os,base64; print base64.urlsafe_b64encode(os.urandom(16))')`).

This could be argued for `clientSecret` too, but at least that one is not ephemeral like the random `cookieSecret` is.

andresmgot:
> While we're on it: is it a good idea to leave this as a mandatory field in `values.yaml`? Feels like a bad idea to code a random cookie into the chart, instead of providing it upon deploy time (e.g., `helm install kubeapps [...] --set authProxy.cookieSecret=$(python -c 'import os,base64; print base64.urlsafe_b64encode(os.urandom(16))')`).
> 

yes @lindhe, the cookieSecret is not being coded in the chart (it's just a comment) so you need to set it as `--set ...`. Regarding if it should be mandatory or not, I think it's required but I will leave @absoludity  to confirm that.

lindhe:
I think you misunderstood me slightly. I meant that having this field in `values.yaml` encourages people to write their `cookieSecret` and package the Chart like that. Not having that option in the `values.yaml` but require it on startup might make sense.

However, I think your misunderstanding also has a great point. Why not extend the program so that it automatically sets the `cookieSecret` to something suitable? Or does it ever need to be retrieved/reused?

absoludity:
> While we're on it: is it a good idea to leave this as a mandatory field in `values.yaml`? Feels like a bad idea to code a random cookie into the chart, instead of providing it upon deploy time (e.g., `helm install kubeapps [...] --set authProxy.cookieSecret=$(python -c 'import os,base64; print base64.urlsafe_b64encode(os.urandom(16))')`).

We can't choose whether a user sets a value in the values.yaml or as a `--set` arg though? The values.yaml is really just documenting the available fields (whether you set them in a custom values.yaml of your own, or use the `--set` arg).


> However, I think your misunderstanding also has a great point. Why not extend the program so that it automatically sets the `cookieSecret` to something suitable? 

I did want to make this a little nicer by having it autogenerated by the chart if it's not been set (ie. so it's not mandatory to set it explicitly, like we do for other passwords), but haven't gotten to that yet (I assume I can't depend on python in a chart helper, but will find a platform independent way).

> Or does it ever need to be retrieved/reused?

Yes, if the number of replicas for the kubeapps frontend is increased, the new ones need the cookie secret (not sure if that's what you meant).

PR #1320, opened 2019-11-26
Delete failed installations

### LINKED ISSUE #1163: "has no deployed releases" limbo-state
If you do an initial install, which fails, you'll get into an invalid state where a redeploy ends in "Unable to update the release: rpc error: code = Unknown desc = “aot-ccm-fs-mon-ws-p1” has no deployed releases"
and the only way to fix this is to do a `helm del --purge aot-ccm-fs-mon-ws-p1` and then go through the whole workflow of deploying again.

kubeapps should notice this and allow deployment to continue.

andresmgot:
**Description of the change**

When creating a release, if it fails for some reason (like band inputs) clean up the release afterwards so the same name can be reused.

BTW, there is no an "atomic" flag that we can use in the CreateRelease action for helm, this is the code:

https://github.com/helm/helm/blob/456eb7f4118a635427bd43daa3d7aabf29304f13/pkg/action/install.go#L313

**Possible drawbacks**

If the release failed to be installed but it's not stored as a failed release (for example for an invalid name) we will be returning also an error for the missing release when trying to delete it, which might be confusing.

**Applicable issues**

  - fixes #1163

cc/ @SimonAlling since these changes affect to tiller-proxy

### DIFF
--- a/pkg/proxy/proxy_test.go
+++ b/pkg/proxy/proxy_test.go
@@ -364,7 +364,31 @@ func TestCreateConflictingHelmRelease(t *testing.T) {
 	if err == nil {
 		t.Error("Release should fail, an existing release in a different namespace already exists")
 	}
-	if !strings.Contains(err.Error(), "name that is still in use") {
+	if !strings.Contains(err.Error(), "already exists") {
+		t.Errorf("Unexpected error %v", err)
+	}
+}
+
+func TestCreateExistingHelmRelease(t *testing.T) {
+	ns := "myns"
+	rs := "foo"
+	chartName := "bar"
+	version := "v1.0.0"
+	ch := &chart.Chart{
+		Metadata: &chart.Metadata{Name: chartName, Version: version},
+	}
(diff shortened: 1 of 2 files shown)


pkg/proxy/proxy_test.go
```diff
@@ -364,7 +364,31 @@ func TestCreateConflictingHelmRelease(t *testing.T) {
...
+		Icon:    "icon.png",
+		Name:    "wordpress",
+	}}
+	proxy := newFakeProxy([]AppOverview{app})
+
+	_, err := proxy.CreateRelease(rs, ns, "", ch)
+	if err == nil {
+		t.Error("Release should fail, an existing release in a different namespace already exists")
+	}
+	if !strings.Contains(err.Error(), "already exists") {
```
absoludity: Did you try adding a test for the actual deletion and it was not worth it for now? (given that tiller-proxy is dying).

andresmgot: yes, there is no easy way for the current fake helm implementation to fail doing an installation

PR #1321, opened 2019-11-26
Remove WithoutParams

SimonAlling:
It is a special case of `WithParams` only used for `ListAllReleases`.

(Removing the `WithParams` abstraction altogether seems to be a completely different story – much harder than implied by the types of `CreateRelease` et al., [`mux.Vars`](https://github.com/kubeapps/kubeapps/blob/d904fdfe22579ac753c2d9c7ac17629ea7e25e63/pkg/handlerutil/handlerutil.go#L18) and [`http.HandlerFunc`](https://golang.org/pkg/net/http/#HandlerFunc).)

### DIFF
--- a/cmd/tiller-proxy/internal/handler/handler.go
+++ b/cmd/tiller-proxy/internal/handler/handler.go
@@ -236,7 +236,7 @@ func (h *TillerProxy) UpgradeRelease(w http.ResponseWriter, req *http.Request, p
 }
 
 // ListAllReleases list all releases that Tiller stores
-func (h *TillerProxy) ListAllReleases(w http.ResponseWriter, req *http.Request) {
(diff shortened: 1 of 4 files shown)


SimonAlling:
[This comment](https://github.com/kubeapps/kubeapps/pull/1311#pullrequestreview-321995498) on #1311 is somewhat relevant.

SimonAlling:
I think my inner typechecker is trying to convince me that the empty map is also a map, but I definitely see your point. I tried redefining `WithoutParams` as a wrapper around `WithParams` to dry the code up a bit, but Go's type system is beyond me sometimes.

SimonAlling:
Since the PR doesn't even pass the CI and it doesn't add a lot of value anyway, I'm closing it.

andresmgot:
FWIW the error in the CI is unrelated to your changes, it's just a flaky test. Thanks anyway!

PR #1322, opened 2019-11-27
Allow to configure the command for the sync image

andresmgot:
**Description of the change**

Remove the hardcoded reference to the command executed in the chart-repo image. It's needed to replace `chart-repo` for something else.

**Benefits**

The image and the command can now change.

**Applicable issues**

  - ref #651

### DIFF
--- a/cmd/apprepository-controller/main.go
+++ b/cmd/apprepository-controller/main.go
@@ -33,6 +33,7 @@ var (
 	masterURL        string
 	kubeconfig       string
 	repoSyncImage    string
+	repoSyncCommand  string
 	namespace        string
 	mongoURL         string
 	mongoSecretName  string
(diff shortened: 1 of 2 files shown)


andresmgot:
_merging this as trivial_ cc/ @absoludity

cmd/apprepository-controller/main.go
```diff
@@ -78,6 +79,7 @@ func init() {
 	flag.StringVar(&kubeconfig, "kubeconfig", "", "Path to a kubeconfig. Only required if out-of-cluster.")
 	flag.StringVar(&masterURL, "master", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
 	flag.StringVar(&repoSyncImage, "repo-sync-image", "quay.io/helmpack/chart-repo:latest", "container repo/image to use in CronJobs")
+	flag.StringVar(&repoSyncCommand, "repo-sync-cmd", "/chart-repo", "command used to sync/delete repos for repo-sync-image")
```
absoludity: Looks fine. Only confusion here is why it's called a `Command` rather than an `Endpoint`, but small detail. Thanks!

andresmgot: it's called command because it's the actual path to the binary to execute (it's not an API endpoint)

absoludity: Oh - I misread. Great. thanks.

PR #1328, opened 2019-11-28
Add Delete method for the postgres driver in asset-syncer

### LINKED ISSUE #651: Consider MongoDB replacement

andresmgot:
**Description of the change**

This PR adds the method `Delete` (which is simpler) for the postgres driver in the asset-syncer. To achieve that I have split the `utils.go` (and `utils_test.go`) in two different files, depending if the method required a database connection so we have now `mongodb_utils.go` and `postgresql_utils.go`. Each one of those with their tests.

It's only necessary to review the files `delete.go`, `main.go`, `postgresql_utlils.go` and `postgresql_utils_test.go`. The rest is just moving methods around.

**Applicable issues**

  - ref #651

### DIFF
--- a/cmd/asset-syncer/delete.go
+++ b/cmd/asset-syncer/delete.go
@@ -17,7 +17,10 @@ limitations under the License.
 package main
 
 import (
+	"fmt"
 	"os"
+
+	"database/sql"
 
 	"github.com/kubeapps/common/datastore"
 	"github.com/sirupsen/logrus"
@@ -33,19 +36,6 @@ var deleteCmd = &cobra.Command{
 			cmd.Help()
 			return
 		}
-		mongoURL, err := cmd.Flags().GetString("mongo-url")
-		if err != nil {
-			logrus.Fatal(err)
-		}
-		mongoDB, err := cmd.Flags().GetString("mongo-database")
-		if err != nil {
-			logrus.Fatal(err)
-		}
-		mongoUser, err := cmd.Flags().GetString("mongo-user")
-		if err != nil {
-			logrus.Fatal(err)
-		}
-		mongoPW := os.Getenv("MONGO_PASSWORD")
 		debug, err := cmd.Flags().GetBool("debug")
 		if err != nil {
 			logrus.Fatal(err)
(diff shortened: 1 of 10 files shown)


andresmgot:
merging this PR to send the next one

cmd/asset-syncer/delete.go
```diff
@@ -33,34 +36,69 @@ var deleteCmd = &cobra.Command{
...
 		if debug {
 			logrus.SetLevel(logrus.DebugLevel)
 		}
-		mongoConfig := datastore.Config{URL: mongoURL, Database: mongoDB, Username: mongoUser, Password: mongoPW}
-		dbSession, err := datastore.NewSession(mongoConfig)
-		if err != nil {
-			logrus.Fatalf("Can't connect to mongoDB: %v", err)
+		database, err := cmd.Flags().GetString("database-type")
+		var manager assetManager
+		if database == "mongodb" {
```
absoludity: I'm not sure why we're needing an if-else here, rather than receiving an implementation of the interface you already defined, and just calling the object's delete implementation? I'll probably find out below :)

cmd/asset-syncer/delete.go
```diff
@@ -33,34 +36,69 @@ var deleteCmd = &cobra.Command{
...
+			mongoURL, err := cmd.Flags().GetString("mongo-url")
+			if err != nil {
+				logrus.Fatal(err)
+			}
+			mongoDB, err := cmd.Flags().GetString("mongo-database")
```
absoludity: either way, we really should not be parsing flags in this file... they should be parsed (and validated) in the main.go to create an instance of your interface, avoiding sprinkling `.Fatal` errors throughout the code.

cmd/asset-syncer/main.go
```diff
@@ -42,9 +42,14 @@ func init() {
...
 	for _, cmd := range cmds {
 		rootCmd.AddCommand(cmd)
+		cmd.Flags().String("database-type", "mongodb", "Database to use. Choice: mongodb, postgresql")
 		cmd.Flags().String("mongo-url", "localhost", "MongoDB URL (see https://godoc.org/github.com/globalsign/mgo#Dial for format)")
 		cmd.Flags().String("mongo-database", "charts", "MongoDB database")
 		cmd.Flags().String("mongo-user", "", "MongoDB user")
+		cmd.Flags().String("pg-host", "localhost", "PostgreSQL Hostname")
+		cmd.Flags().String("pg-port", "5432", "PostgreSQL Port")
+		cmd.Flags().String("pg-database", "assets", "PostgreSQL database")
+		cmd.Flags().String("pg-user", "", "PostgreSQL user")
```
absoludity: As above, can we pls parse these in the main.go and create the implementation of your interface here, or fatal here?

andresmgot:
> As per the inline comments, I really think we should:
> 
> 1. have all the parsing for the cmd args in the main.go
> 2. Create an instance of the specific interface implementation for the asset manager there (ie. provide a `NewPostgresAssetManager` and `NewMongoAssetManager` in their respective utils), which stores the required options, then
> 3. the implementations of `Delete` can encapsulate all the specific code, avoiding the if-else in `delete.go` (actually, not even sure it would be required, with the functionality encapsulated in each asset manager implementation).

Let me try. In theory, `delete.go` and `sync.go` are defining the command line for those two methods, that's why the parsing of flags is there (but since those are common, we can probably move them to `main.go`.

PR #1333, opened 2019-12-02
Add sync method for postgresql

andresmgot:
**Description of the change**

`Sync` method for the PostgreSQL driver of the `asset-syncer`. Still missing the tests and the icons fetch logic because I plan to refactor this method first (for the mongodb driver as well).

**Applicable issues**

  - ref #651

### DIFF
--- a/cmd/asset-syncer/sync.go
+++ b/cmd/asset-syncer/sync.go
@@ -17,7 +17,10 @@ limitations under the License.
 package main
 
 import (
+	"fmt"
 	"os"
+
+	"database/sql"
 
 	"github.com/kubeapps/common/datastore"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


cmd/asset-syncer/sync.go
```diff
@@ -34,36 +37,71 @@ var syncCmd = &cobra.Command{
...
 		if debug {
 			logrus.SetLevel(logrus.DebugLevel)
 		}
-		mongoConfig := datastore.Config{URL: mongoURL, Database: mongoDB, Username: mongoUser, Password: mongoPW}
-		dbSession, err := datastore.NewSession(mongoConfig)
-		if err != nil {
-			logrus.Fatalf("Can't connect to mongoDB: %v", err)
+		database, err := cmd.Flags().GetString("database-type")
+		var manager assetManager
+		if database == "mongodb" {
```
absoludity: Same comment as previous PR - let's avoid sprinkling the implementation details throughout, and instead encapsulate the specifics in each interface implementation (and instantiate those in the main.go)

absoludity:
Regarding testing (which you mentioned is still TODO), are you planning to just add a small test double of your `postgresDB` interface? Might be simplest, but keen to hear your thoughts.

andresmgot:
> Regarding testing (which you mentioned is still TODO), are you planning to just add a small test double of your `postgresDB` interface? Might be simplest, but keen to hear your thoughts.

I think you mean a `fake` implementation of the postgresDB interface? For the moment I didn't see it necessary. I will send a specific PR for those tests but it will be simple, just mocking the `Query` methods of Postgres to check some basic functionality.

absoludity:
A fake is a specific type of test-double, like an in-memory database... that'd be a bit much. I wasn't really specifying what type of test-double, but yep, I'd just do a simple stub/spy which also what you are suggesting. A good definition imo is this oldie-but-goodie https://martinfowler.com/bliki/TestDouble.html

absoludity:
Ah no, I see in #1139 you are mocking instead (ie. not returning actual canned responses at all, but just verifying calls). OK.

PR #1335, opened 2019-12-02
Refactor sync method

andresmgot:
**Description of the change**

The goal of this PR is to move anything not related to the database out of the `Sync` method of the database so it can be reused for the different drivers.

Apart from that, I have moved the mongodb methods to the `mongoManager` struct.

**Applicable issues**

  - ref #651


andresmgot:
I am going to land this change, even with the tests failing because I had already prepared a small PR for the tests for postgres. I will fix the mongodb manager in another PR.

PR #1337, opened 2019-12-02
Kubeapps - Add JSON Schema & standardize ingress configuration

juan131:
**Description of the change**

This PR includes a basic JSON schema for the Kubeapps chart. It doesn't cover all the possible values of the chart but just the ones that, in my opinion, are more likely to be modified.

This is valid for Helm 3 to validate some values and, at the same time, for Kubeapps to render a simple form so the chart containing it is easier to configure and deploy. See https://github.com/kubeapps/kubeapps/blob/master/docs/developer/basic-form-support.md for more information.

This is an example of the file rendered by Kubeapps:

Screenshot 2019-12-02 at 15 25 09

This PR also standardizes the way we configure Ingress.

### DIFF
--- a/chart/kubeapps/templates/ingress.yaml
+++ b/chart/kubeapps/templates/ingress.yaml
@@ -22,6 +22,16 @@ spec:
             serviceName: {{ template "kubeapps.fullname" . }}
             servicePort: http
     {{- end }}
+    {{- range .Values.ingress.extraHosts }}
+    - host: {{ .name }}
+      http:
+        paths:
+        - path: {{ default "/" .path }}
+          backend:
+            serviceName: {{ template "kubeapps.fullname" $ }}
+            servicePort: http
+    {{- end }}
+    ## The block below is deprecated and must removed on 3.0.0
     {{- range .Values.ingress.hosts }}
     - host: {{ .name }}
       http:
@@ -31,7 +41,25 @@ spec:
             serviceName: {{ template "kubeapps.fullname" $ }}
             servicePort: http
     {{- end }}
-  {{- if .Values.ingress.tls }}
-  tls: {{- toYaml .Values.ingress.tls | nindent 4 }}
+    ## end of block
+  {{- if or .Values.ingress.tls .Values.ingress.extraTls .Values.ingress.hosts }}
+  tls:
+    ## The block below is deprecated and must removed on 3.0.0
+    {{- range .Values.ingress.hosts }}
+    {{- if .tls }}
+    - hosts:
+        - {{ .name }}
+      secretName: {{ .tlsSecret }}
+    {{- end }}
+    {{- end }}
+    ## end of block
+    {{- if .Values.ingress.tls }}
+    - hosts:
+        - {{ .Values.ingress.hostname }}
+      secretName: {{ printf "%s-tls" .Values.ingress.hostname }}
+    {{- end }}
+    {{- if .Values.ingress.extraTls }}
+    {{- toYaml .Values.ingress.extraTls | nindent 4 }}
+    {{- end }}
   {{- end }}
 {{- end }}
--- a/chart/kubeapps/templates/ingress.yaml
+++ b/chart/kubeapps/templates/ingress.yaml
@@ -13,6 +13,15 @@ metadata:
     {{- end }}
 spec:
   rules:
+    {{- if .Values.ingress.hostname }}
+    - host: {{ .Values.ingress.hostname }}
+      http:
+        paths:
+        - path: /
+          backend:
+            serviceName: {{ template "kubeapps.fullname" . }}
+            servicePort: http
+    {{- end }}
     {{- range .Values.ingress.hosts }}
     - host: {{ .name }}
       http:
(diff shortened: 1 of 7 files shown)


juan131:
@andresmgot I don't like how "arrays" are rendered, I guess we'll need to do sth to improve the UX

chart/kubeapps/values.schema.json
```diff
@@ -0,0 +1,84 @@
+{
+  "$schema": "http://json-schema.org/schema#",
+  "type": "object",
+  "properties": {
+    "frontend": {
+      "type": "object",
+      "title": "Frontend configuration",
+      "form": true,
+      "properties": {
+        "replicaCount": {
```
andresmgot: I would expose the replica count of the other services too: `tiller-proxy`, `chartsvc`, `dashboard` (not the apprepository-controller which should be 1).

chart/kubeapps/values.schema.json
```diff
@@ -0,0 +1,84 @@
...
+      "type": "object",
+      "title": "Chart repositories configuration",
+      "form": true,
+      "properties": {
+        "initialRepos": {
+          "type": "array",
+          "title": "Initial chart repositories",
+          "description": "Initial chart repositories to configure on Kubeapps",
+          "form": true
+        }
```
andresmgot: I would remove this parameter for two reasons:
 - As you said we don't have support for arrays yet so the representation is confusing.
 - This form would be shown only when upgrading (unless you are deploying a second instance of Kubeapps). The `initialRepos` are only meant to be set the first time you install the chart (that's why it's called `initial*`). Changing this parameter when upgrading will have no effect since after installation, repositories are managed using the Kubeapps UI.

chart/kubeapps/values.schema.json
```diff
@@ -0,0 +1,84 @@
...
+              "hidden": {
+                "condition": false,
+                "value": "mongodb.persistence.enabled"
+              }
+            }
+          }
+        }
+      }
+    }
+  }
```
andresmgot: Can you also add the `authProxy` section? It's useful to configure the  SSO :)

chart/kubeapps/templates/ingress.yaml
```diff
@@ -22,12 +31,7 @@ spec:
...
     {{- end }}
-  tls:
-  {{- range .Values.ingress.hosts }}
-    {{- if .tls }}
-    - hosts:
-        - {{ .name }}
-      secretName: {{ .tlsSecret }}
-    {{- end }}
+  {{- if .Values.ingress.tls }}
+  tls: {{- toYaml .Values.ingress.tls | nindent 4 }}
```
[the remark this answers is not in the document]
juan131: I don't think we're including a backwards incompatible change, you only need to adapt the parameters you're setting (but the chart can be upgraded):
```
$ helm install --name kubeapps bitnami/kubeapps \
    --set ingress.enabled=true \
    --set ingress.hosts[0].name=kubeapps.domain \
    --set ingress.hosts[0].tlsSecret=kubeapps.domain-tls
$ helm upgrade kubeapps bitnami/kubeapps \
    --set ingress.enabled=true \
    --set ingress.hostname=kubeapps.domain \
    --set ingress.tls[0].hosts[0]=kubeapps.domain \
    --set ingress.tls[0].secretName=kubeapps.domain-tls
```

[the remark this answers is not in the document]
juan131: If you consider that adapting the parameters should bump the major version, then i don't know if it's worth

chart/kubeapps/templates/ingress.yaml
```diff
@@ -13,6 +13,15 @@ metadata:
...
   rules:
+    {{- if .Values.ingress.hostname }}
+    - host: {{ .Values.ingress.hostname }}
+      http:
+        paths:
+        - path: /
+          backend:
+            serviceName: {{ template "kubeapps.fullname" . }}
+            servicePort: http
+    {{- end }}
```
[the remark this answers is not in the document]
juan131: Sure, I can do sth like that

chart/kubeapps/templates/ingress.yaml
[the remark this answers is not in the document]
juan131: By the way, I adapted the PR so now it supports both ways to configure Ingress (deprecating the old one)

chart/kubeapps/values.schema.json
```diff
@@ -0,0 +1,158 @@
...
+        },
+        "tls": {
+          "type": "boolean",
+          "form": true,
+          "title": "Enable TLS configuration",
+          "hidden": {
+            "condition": false,
+            "value": "ingress.enabled"
+          }
+        }
```
andresmgot: this is great :) I hope this becomes a bitnami standard so we can add this parameters for other charts :)

Also, can you add a parameter to enable `certManager`? That way we can enable TLS with cert-manager directly from the form.

chart/kubeapps/values.schema.json
```diff
@@ -0,0 +1,158 @@
...
+        "clientSecret": {
+          "type": "string",
+          "form": true,
+          "title": "Client Secret",
+          "description": "Secret used to validate the Client ID",
+          "hidden": {
+            "condition": false,
+            "value": "authProxy.enabled"
+          }
+        }
```
andresmgot: the cookie secret is also a mandatory parameter (which is not populated by default). Do you mind adding it?

juan131:
All suggestions addressed @andresmgot

PR #1339, opened 2019-12-03
Unit tests for postgres sync

andresmgot:
**Description of the change**

Unit tests for the `Sync` method of the postgres manager. We cannot test a lot because this depends highly from the posgresql connection and response but we are adding some coverage. End-to-end tests will cover better this functionality.

**Applicable issues**

  - ref #651

### DIFF
--- a/cmd/asset-syncer/postgresql_utils.go
+++ b/cmd/asset-syncer/postgresql_utils.go
@@ -60,6 +60,9 @@ func (m *postgresAssetManager) Sync(charts []chart) error {
 		return err
 	}
 
+	// Remove charts no longer existing in index
+	m.removeMissingCharts(charts)
+
 	// TODO(andresmgot): Fetch and store chart icons
 
 	return nil
@@ -68,8 +71,11 @@ func (m *postgresAssetManager) Sync(charts []chart) error {
 func (m *postgresAssetManager) RepoAlreadyProcessed(repoName, repoChecksum string) bool {
 	var lastChecksum string
 	row := m.db.QueryRow(fmt.Sprintf("SELECT checksum FROM %s WHERE name = '%s'", repositoryTable, repoName))
-	err := row.Scan(&lastChecksum)
-	return err == nil && lastChecksum == repoChecksum
+	if row != nil {
+		err := row.Scan(&lastChecksum)
+		return err == nil && lastChecksum == repoChecksum
+	}
+	return false
 }
 
 func (m *postgresAssetManager) UpdateLastCheck(repoName, checksum string, now time.Time) error {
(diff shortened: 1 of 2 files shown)


cmd/asset-syncer/postgresql_utils.go
```diff
@@ -116,14 +120,16 @@ func (m *postgresAssetManager) importCharts(charts []chart) error {
...
 
-	// Remove charts no longer existing in index
+func (m *postgresAssetManager) removeMissingCharts(charts []chart) error {
+	var chartIDs []string
+	for _, chart := range charts {
+		chartIDs = append(chartIDs, fmt.Sprintf("'%s'", chart.ID))
+	}
 	chartIDsString := strings.Join(chartIDs, ", ")
-	_, err = m.db.Query(fmt.Sprintf("DELETE FROM %s WHERE info ->> 'ID' NOT IN (%s)", chartTable, chartIDsString))
+	_, err := m.db.Query(fmt.Sprintf("DELETE FROM %s WHERE info ->> 'ID' NOT IN (%s)", chartTable, chartIDsString))
```
absoludity: Throughout this file (not just this PR), we seem to be interpolating the query string ourselves, using `fmt.Sprintf`, which is dangerous? As per the signature of the methods in the interface, we should be passing the query string template *and* args to the `Query` command. For example, in this case,

```
_, err := m.db.Query(fmt.Sprintf("DELETE FROM %s WHERE info ->> 'ID' NOT IN (%s)", chartTable, chartIDsString))
```
should be
```
_, err := m.db.Query("DELETE FROM ? WHERE info ->> 'ID' NOT IN (?)", chartTable, chartIDs)
```
or similar?

andresmgot:
> I'm +1ing, but can you please update so that we're not manually interpolating the SQL statements, but rather are letting the db module do the interpolation.
> 

I did try that. What I don't know is why the arguments were not being interpolated when they were used in the first time of the request. An example:

```
"DELETE FROM $1 WHERE info ->> 'ID' NOT IN $2", chartTable, chartIDs
```

The above won't work (I am not sure why, $1 was not being replaced). But this works:

```
"DELETE FROM charts WHERE info ->> 'ID' NOT IN $1", chartIDs
```

I am not sure why, but I can investigate that in a different PR.

> As a follow-up PR, we probably want to use `QueryContext` and `QueryRowContext` instead, so that the request context can be passed through to the db commands. This allows the client to specify the context (ie. any timeout or whatever) which will be passed through and used by the db library also.

Right now we don't have a use case for that context. Do you still want to use those methods even if the context is empty/not used?

absoludity:
> > As a follow-up PR, we probably want to use `QueryContext` and `QueryRowContext` instead, so that the request context can be passed through to the db commands. This allows the client to specify the context (ie. any timeout or whatever) which will be passed through and used by the db library also.
> 
> Right now we don't have a use case for that context. Do you still want to use those methods even if the context is empty/not used?

No no - I keep thinking this is a service triggered by http requests, but it's a cmd-line util.

absoludity:
Actually, `Query` just calls `QueryContext` with a `context.Background` anyway: https://golang.org/src/database/sql/sql.go?s=73205:73261#L1558

So these db queries will be able to block forever. We'll probably want to use `QueryContext` etc., instead, passing a `context.Timeout` with a configured value at some point.

PR #1341, opened 2019-12-03
Refactor flags and manager initialization

andresmgot:
**Description of the change**

Follow up of the conversation here:
https://github.com/kubeapps/kubeapps/pull/1328#pullrequestreview-325900983

The goal is to refactor the current flag initialization to reduce duplicated code and move code to the `main.go` flags.

Unfortunately, it's not possible to execute code in the `main.go` file *after* the flags are populated so still the manager initialization happens in the `sync.go` and `delete.go` files.

**Applicable issues**

  - ref #651

### DIFF
--- a/cmd/asset-syncer/delete.go
+++ b/cmd/asset-syncer/delete.go
@@ -17,12 +17,6 @@ limitations under the License.
 package main
 
 import (
-	"fmt"
-	"os"
-
-	"database/sql"
-
-	"github.com/kubeapps/common/datastore"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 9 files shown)


cmd/asset-syncer/main.go
```diff
@@ -38,21 +47,19 @@ func main() {
 }
 
 func init() {
-	cmds := []*cobra.Command{syncCmd, deleteCmd}
+	rootCmd.PersistentFlags().StringVar(&databaseType, "database-type", "mongodb", "Database to use. Choice: mongodb, postgresql")
```
andresmgot: persisten flags are used to allow these flags to be set either in the root or in the command

cmd/asset-syncer/mongodb_utils.go
```diff
@@ -40,7 +40,21 @@ const (
...
+	dbSession, err := datastore.NewSession(m.mongoConfig)
+	if err != nil {
+		return fmt.Errorf("Can't connect to mongoDB: %v", err)
+	}
+	m.dbSession = dbSession
+	return nil
+}
+
+func (m *mongodbAssetManager) Close() error {
+	return nil
```
andresmgot: `dbSession` don't expose a Close method so we do nothing here

cmd/asset-syncer/sync.go
```diff
@@ -38,67 +34,16 @@ var syncCmd = &cobra.Command{
...
-			defer db.Close()
-			manager = &postgresAssetManager{db}
-		} else {
-			logrus.Fatalf("Unsupported database type %s", database)
+
+		manager, err := newManager(databaseType, databaseURL, databaseName, databaseUser, databasePassword)
+		if err != nil {
+			logrus.Fatal(err)
 		}
+		manager.Init()
```
absoludity: Same here - `err` not being handled, and anyways, if `manager.Init()` was returned from `newManager` it'd dry up both uses.

cmd/asset-syncer/utils.go
```diff
@@ -74,6 +75,27 @@ type assetManager interface {
...
 	UpdateLastCheck(repoName, checksum string, now time.Time) error
+	Init() error
+	Close() error
+}
+
+func newManager(databaseType, databaseURL, databaseName, databaseUser, databasePassword string) (assetManager, error) {
+	if databaseType == "mongodb" {
+		mongoConfig := datastore.Config{URL: databaseURL, Database: databaseName, Username: databaseUser, Password: databasePassword}
+		return &mongodbAssetManager{mongoConfig, nil}, nil
+	} else if databaseType == "postgresql" {
```
absoludity: This could be a bit neater if each util (mongodb, postgresql) defined a `New` function, ie:
```
func NewMongoAssetManager(config *datastore.Config) (assetManager, error)
```
and similar for `NewPostgresAssetManager`. This would mean you don't need to instantiate objects with `nil` fields as you're doing here (more importantly, you don't need to know the implementation details, such as that they have a db field which needs to be nil on instantiation). It would also mean that it could initialize itself (returning the error).

So this whole section would be simply:
```
switch databaseType {
  case MONGODB:
    return NewMongoAssetManager(config)
  case POSTGROSQL:
    return NewPostgresAssetManager(config)
}
return nil, fmt.Errorf("Unsup....")
```

cmd/asset-syncer/main.go
```diff
@@ -22,6 +22,15 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var (
+	databaseType     string
+	databaseURL      string
+	databaseName     string
+	databaseUser     string
+	databasePassword string
```
absoludity: From what I can tell below, we could just have:
```
dbConfig database.Config
```
here, and pass that down, rather than handling all 4 separately?

cmd/asset-syncer/main.go
```diff
@@ -38,21 +47,19 @@ func main() {
 }
 
 func init() {
-	cmds := []*cobra.Command{syncCmd, deleteCmd}
+	rootCmd.PersistentFlags().StringVar(&databaseType, "database-type", "mongodb", "Database to use. Choice: mongodb, postgresql")
+	rootCmd.PersistentFlags().StringVar(&databaseURL, "database-url", "localhost", "MongoDB URL (see https://godoc.org/github.com/globalsign/mgo#Dial for format)")
```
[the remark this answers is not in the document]
andresmgot: oops, missed these

cmd/asset-syncer/main.go
```diff
@@ -38,21 +47,19 @@ func main() {
...
+	rootCmd.PersistentFlags().StringVar(&databaseName, "database-name", "charts", "MongoDB database")
+	rootCmd.PersistentFlags().StringVar(&databaseUser, "database-user", "", "MongoDB user")
+	// see version.go
```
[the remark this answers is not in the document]
andresmgot: just copied what we had but yes, it's related to the user-agent-comment flag

cmd/asset-syncer/delete.go
```diff
@@ -36,67 +30,17 @@ var deleteCmd = &cobra.Command{
...
-			}
-			defer db.Close()
-			manager = &postgresAssetManager{db}
-		} else {
-			logrus.Fatalf("Unsupported database type %s", database)
+		manager, err := newManager(databaseType, databaseURL, databaseName, databaseUser, databasePassword)
+		if err != nil {
+			logrus.Fatal(err)
 		}
+		manager.Init()
```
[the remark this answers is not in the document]
andresmgot: oops, yes, I forgot to handle the error.

I moved the actual database connection to the `Init` method for two reasons:
 - The `newManager` can be tested easily
 - Since `Init` leaves the connection open and it needs to be closed (in the postgresql case), it's more readable if it's done in the same method. I think it's more a bit more readable that way:

```
manager.Init()
defer manager.Close()
```

PR #1342, opened 2019-12-04
Improve tiller-proxy build time with BuildKit

SimonAlling:
### Description of the change

  * [BuildKit](https://docs.docker.com/develop/develop-images/build_enhancements) is used for `docker build`.
  * `COPY . .` is replaced with more explicit directives.
  * `--mount` flag preserves Go build cache between builds.
  * Removing the `-a` flag makes `go build` use Go build cache.

### Benefits

This brought the build time (after modifying the source code) down from 20 seconds to about 5 seconds on my laptop.

### Possible drawbacks

* BuildKit requires Docker 18.09 or later. I don't know if building still works on older versions of Docker.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -120,7 +120,8 @@ exports: &exports
     echo "export HELM_VERSION=v2.14.0" >> $BASH_ENV
 build_images: &build_images
   steps:
-    - setup_remote_docker
+    - setup_remote_docker:
+        version: 18.09.3
     - checkout
     - <<: *exports
     - run: |
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ all: kubeapps/dashboard kubeapps/apprepository-controller kubeapps/tiller-proxy
 
 # TODO(miguel) Create Makefiles per component
 kubeapps/%:
-	docker build -t kubeapps/$*$(IMG_MODIFIER):$(IMAGE_TAG) --build-arg "VERSION=${VERSION}" -f cmd/$*/Dockerfile .
+	DOCKER_BUILDKIT=1 docker build -t kubeapps/$*$(IMG_MODIFIER):$(IMAGE_TAG) --build-arg "VERSION=${VERSION}" -f cmd/$*/Dockerfile .
 
 kubeapps/dashboard:
 	docker build -t kubeapps/dashboard$(IMG_MODIFIER):$(IMAGE_TAG) -f dashboard/Dockerfile dashboard/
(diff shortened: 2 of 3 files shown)


andresmgot:
Thanks for the improvements @SimonAlling! Unfortunately, it seems that CircleCI doesn't have a version of docker new enough to support  buildkit. Maybe you can move that to a variable? so in your local environment, you execute `DOCKER_BUILDKIT=1 make kubeapps/tiller-proxy` and in the CI we run it without the buildkit?

SimonAlling:
> Thanks for the improvements @SimonAlling! Unfortunately, it seems that CircleCI doesn't have a version of docker new enough to support buildkit. Maybe you can move that to a variable? so in your local environment, you execute `DOCKER_BUILDKIT=1 make kubeapps/tiller-proxy` and in the CI we run it without the buildkit?

I could definitely do that, but it doesn't seem trivial. In particular, if I remove `DOCKER_BUILDKIT=1` in `Makefile`, it is not possible to `make`  regularly anymore:

```
$ make kubeapps/tiller-proxy
docker build -t kubeapps/tiller-proxy:dev-2019-12-04T11-54-10-CET --build-arg "VERSION=$(git rev-parse HEAD)" -f cmd/tiller-proxy/Dockerfile .
Sending build context to Docker daemon  64.37MB
Error response from daemon: Dockerfile parse error line 12: Unknown flag: mount
```

However, judging from the CircleCI output, you're using Docker 18.09, [which does support BuildKit](https://docs.docker.com/develop/develop-images/build_enhancements/):

```
Build-agent version 1.0.21372-bd73ea34 (2019-12-03T19:06:06+0000)
Docker Engine Version: 18.09.6
```

So I'm a bit surprised that it doesn't work.

andresmgot:
Maybe it's something that needs to be enabled in the daemon? This is what I see in the CircleCI logs:
```
DOCKER_BUILDKIT=1 docker build -t kubeapps/apprepository-controller-ci:build-90c487be9d916722d09c7d7d2a3dec67104beebb --build-arg "VERSION=$(git rev-parse HEAD)" -f cmd/apprepository-controller/Dockerfile .
buildkit not supported by daemon
make: *** [Makefile:22: kubeapps/apprepository-controller] Error 1
```

SimonAlling:
Yeah, I saw that as well. I'm not very familiar with CircleCI (and I don't have write access anyway); do you have time to try to find a solution? I think almost everyone working with Kubeapps would benefit from this PR.

andresmgot:
It seems it's possible to enable it from the comment here: https://ideas.circleci.com/ideas/CCI-I-1003

Can you replace in `.circleci/config.yaml` the line `- setup_remote_docker` with:

```
      - setup_remote_docker:
          version: 18.09.3
```

That should work :crossed_fingers:

PR #1345, opened 2019-12-04
Use Postgres arguments

andresmgot:
### Description of the change

Follow up of https://github.com/kubeapps/kubeapps/pull/1339#issuecomment-561613522

This also fixes the golang tests for postgres.

### DIFF
--- a/cmd/asset-syncer/postgresql_utils.go
+++ b/cmd/asset-syncer/postgresql_utils.go
@@ -98,7 +98,7 @@ func (m *postgresAssetManager) Sync(charts []chart) error {
 
 func (m *postgresAssetManager) RepoAlreadyProcessed(repoName, repoChecksum string) bool {
 	var lastChecksum string
-	row := m.db.QueryRow(fmt.Sprintf("SELECT checksum FROM %s WHERE name = '%s'", repositoryTable, repoName))
+	row := m.db.QueryRow(fmt.Sprintf("SELECT checksum FROM %s WHERE name = $1", repositoryTable), repoName)
 	if row != nil {
 		err := row.Scan(&lastChecksum)
 		return err == nil && lastChecksum == repoChecksum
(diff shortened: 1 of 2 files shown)


cmd/asset-syncer/postgresql_utils.go
```diff
@@ -98,7 +98,7 @@ func (m *postgresAssetManager) Sync(charts []chart) error {
 
 func (m *postgresAssetManager) RepoAlreadyProcessed(repoName, repoChecksum string) bool {
 	var lastChecksum string
-	row := m.db.QueryRow(fmt.Sprintf("SELECT checksum FROM %s WHERE name = '%s'", repositoryTable, repoName))
+	row := m.db.QueryRow(fmt.Sprintf("SELECT checksum FROM %s WHERE name = $1", repositoryTable), repoName)
```
absoludity: We shouldn't be using `Sprintf` at all for queries? Can you try:
```
row := m.db.QueryRow(`SELECT checksum FROM ? WHERE name = ?`, repositoryTable, repoName)
```
or similar? (ie. just using `?` for arg placement).

andresmgot: I am not sure if I am missing something, are you getting that `?` notation from somewhere?

What you suggest doesn't work, with a syntax error:

```
pq: syntax error at or near "?" 
```

The only documentation I have found is related to arguments. The problem is that `arguments` can only be used with values, not identifiers (from [here](https://www.postgresql.org/docs/9.1/xfunc-sql.html)):

> Arguments to the SQL function are referenced in the function body using the syntax $n: $1 refers to the first argument, $2 to the second, and so on. If an argument is of a composite type, then the dot notation, e.g., $1.name, can be used to access attributes of the argument. The arguments can only be used as data values, not as identifiers. Thus for example this is reasonable:
>
> `INSERT INTO mytable VALUES ($1);`
>
> but this will not work:
>
> `INSERT INTO $1 VALUES (42);`

absoludity: The `?` notation is just what's used in the docs for the `database/sql` package, but it's driver dependent (`database/sql` doesn't force one or the other). And sorry, my mistake, I misread the above: the change you've done is exactly what I'd asked for originally - so that we're passing the args into the query function, rather than interpolating ourselves. The only string interpolation now is just the table name. (I had at first thought you were still using `Sprintf` because of some issue with the query interpolation.)

absoludity: Before I realised, I went to test this out by running the tests, but see tests currently only run against a mockDB. It may be worth setting up something simple, so that if certain env vars are present, it uses a real connection, otherwise defaults to the mockdb. `pq` itself just uses env vars and a docker container. Not urgent though, but useful later when anyone else works on this code.

PR #1347, opened 2019-12-05
Initial plan for kubeapps backend API which can be used by 3rd parties.

absoludity:
### Description of the change

Initial plan with stub handler for implementing a backend API which can be used for 3rd party integrations.

### Benefits
See https://github.com/kubeapps/kubeapps/blob/master/docs/architecture/design-proposals/third-party-add-repository.md

### Possible drawbacks

Until the helm3 work has settled, we don't have an obvious place for the backend API, so I'm including it in tiller-proxy for the moment, but using the frontend config to keep it independent (ie. we can easily move it later).

### Applicable issues
Relates to: #1173 

### Additional information

There are a couple of questions I have regarding the direction, which I'll mention in-line.


chart/kubeapps/templates/kubeapps-frontend-config.yaml
```diff
@@ -79,6 +79,18 @@ data:
         {{- end }}
       }
 
+      location ~* /api/backend {
```
absoludity: So the current config already defines things like `/api/tillerdeploy` and `/api/chartsvc`, so I need `/api/<something>` to maintain the same namespacing, but I'd actually prefer to just use `/api` rather than `/api/backend`. 

Since this will be used by 3rd parties, prefixing with `/backend` doesn't really make sense or help. For example, this endpoint could/should just be `/api/apprepositories`. Thoughts?

cmd/tiller-proxy/main.go
```diff
@@ -181,6 +181,16 @@ func main() {
 		negroni.Wrap(handlerutil.WithParams(h.DeleteRelease)),
 	))
 
+	// Backend routes unrelated to tiller-proxy functionality.
+	// TODO(mnelson): Once the helm3 support is complete and tiller-proxy is being removed,
+	// reconsider where these endpoints live.
+	appreposHandler := handler.AppRepositories{}
+	backendAPIv1 := r.PathPrefix("/backend/v1").Subrouter()
+	backendAPIv1.Methods("POST").Path("/apprepositories").Handler(negroni.New(
+		authGate,
```
absoludity: I've used `authGate` here simply because it appears for all the other handlers, but I don't see why it's necessary here, as I'll be hitting the k8s API with the users credentials (with the `id_token` or `access_token` sent with the request). I assume it's safe to remove here?

chart/kubeapps/templates/kubeapps-frontend-config.yaml
andresmgot: I agree that the `/backend` part is not adding anything to the endpoint definition so I am okay using `/api/apprepositories` instead.

cmd/tiller-proxy/main.go
andresmgot: yes, it's probably not needed. I'd only see it useful if we do something (e.g. fetching credentials) before doing the actual request to the k8s API.

PR #1350, opened 2019-12-05
Update getting-started.md

ruguodangshi:
Because in many cases, kubeapps cannot be accessed through 127.0.0.1,I think it is necessary to provide a way to modify the svc of this kubeapps

### DIFF
--- a/docs/user/getting-started.md
+++ b/docs/user/getting-started.md
@@ -79,6 +79,12 @@ Paste the token generated in the previous step to authenticate and access the Ku
 
 ![Dashboard main page](../img/dashboard-home.png)
 
+***Note:*** If you can't access the dashboard via http://127.0.0.1:8080/ (for example, you don't have a browser in your localhost environment),you can use the following command to change the type of the kubeapps svc to nodePort or other types: 
+
+```bash
+kubectl  edit svc kubeapps -n kubeapps
+``` 
+
 ## Step 4: Deploy WordPress
 
 Once you have the Kubeapps Dashboard up and running, you can start deploying applications into your cluster.
--- a/docs/user/getting-started.md
+++ b/docs/user/getting-started.md
@@ -77,13 +77,7 @@ This will start an HTTP proxy for secure access to the Kubeapps Dashboard. Visit
 
 Paste the token generated in the previous step to authenticate and access the Kubeapps dashboard for Kubernetes.
 
-![Dashboard main page](../img/dashboard-home.png)
-
-***Note:*** If you can't access the dashboard via http://127.0.0.1:8080/ (for example, you don't have a browser in your localhost environment),you can use the following command to change the type of the kubeapps svc to nodePort or other types: 
-
-```bash
-kubectl  edit svc kubeapps -n kubeapps
-``` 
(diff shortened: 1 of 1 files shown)


docs/user/getting-started.md
```diff
@@ -79,6 +79,12 @@ Paste the token generated in the previous step to authenticate and access the Ku
 
 ![Dashboard main page](../img/dashboard-home.png)
 
+***Note:*** If you can't access the dashboard via http://127.0.0.1:8080/ (for example, you don't have a browser in your localhost environment),you can use the following command to change the type of the kubeapps svc to nodePort or other types: 
+
+```bash
+kubectl  edit svc kubeapps -n kubeapps
+``` 
+
```
absoludity: Thanks for working to improve the docs! I'm interested to know your use case for running `kubectl` in an environment without a browser? A couple of thoughts:

So I do all my development on a headless instance and so am similarly using `kubectl` on a machine without a browser, but as I ssh into that machine, I can just use `LocalForward` in my ssh config so that I can still access 127.0.0.1:8080 on my client machine.

There are quite a few different solutions to more advanced setups, I'm just not sure it should be part of the Getting Started guide. Though perhaps we should update the prerequisites from:

```
and [`kubectl`](https://kubernetes.io/docs/tasks/tools/install-kubectl/) installed and configured to talk to your Kubernetes cluster.
```
to
```
and [`kubectl`](https://kubernetes.io/docs/tasks/tools/install-kubectl/) installed on your local machine and configured to talk to your Kubernetes cluster.
```
so that the Getting Started isn't dealing with remote setups (ie. where you are running `kubectl` via ssh - if that's what you're doing like me?)

ruguodangshi: Thank you!I understand what you mean.
Our company is a traditional company. In our company, our project team is responsible for k8s and paas, and we provide k8s and other middleware for other project teams responsible for writing business code. They don't know about k8s, they don't care about installing kubectl, etc. They just want to install the software they want by opening a browser and clicking the deploy button. So I have to change the type of kubeapps svc to nodePort, LoadBalancer, or create an ingress object for it

absoludity: Excellent, thanks for the details. So yes, I was assuming you were setting up as a developer for your own access, but you're actually setting up for your users... and what you've found is that the getting-started guide is not currently helping you there. With that in mind, how about:

```
***Note:*** If you are setting up Kubeapps for other people to access, you will want to use a different service type or setup Ingress rather than using the above `kubectl port-forward`. For detailed information on installing, configuring and upgrading Kubeapps, checkout the [chart README](../../chart/kubeapps/README.md).
```

This repeats the link to the chart readme (which is already mentioned in Step 1: Install Kubeapps, but based on your experience, I think it's worth repeating here.

What do you think? Would that have helped you?

(For what it's worth, I don't think we should be recommending editing the deployment but should always prefer to show how to install with the desired options/ingress/service type in the first place, which is what the chart readme tries to do).

Thanks!

ruguodangshi: Thank you! you are right

absoludity:
@ruguodangshi Did you mean to close it before I merged into Master? If you re-open I can merge it (the above message was just because I first merged master into this branch to update it, so our CI could run again before I merged it into master).

absoludity:
Ah - nm. I see you opened #1353 with the same change. I'll merge that :) Thanks

PR #1355, opened 2019-12-06
Improve apprepository-controller build time

SimonAlling:
See #1342.

I haven't worked with apprepository-controller, but this PR is helpful for `make all` at least.


SimonAlling:
> Out of interest, when are you using `make all`? I generally just build the image for the service that I'm working on, but interested to know what your workflow is.

"Workflow" might be a bit of a stretch, but I tend to `make all` sometimes when I want to be sure that everything is rebuilt. I usually do `make all && make test` before pushing at least.

PR #1363, opened 2019-12-11
Don't panic in ParseDriverType

SimonAlling:
### Description of the change

It rewrites `agent.ParseDriverType` to return an error instead of panicking if it doesn't recognize its input.

### Benefits

* `ParseDriverType` is made more faithful to its declared type.
* It's made more obvious in `main.go` that the application may panic.

### Possible drawbacks

* `ParseDriverType` becomes more verbose.

### Applicable issues

Excerpt from a [related discussion](https://github.com/kubeapps/kubeapps/pull/1359#discussion_r355911774) with @andresmgot in #1359:

> I'm much more inclined to make `agent.ParseDriverType` return an error instead of panicking, because, unlike `createStorage`, it promises that it can handle any `string`. Its current deceitfulness stems from my interpretation of @absoludity's [suggestion](https://github.com/kubeapps/kubeapps/pull/1309#discussion_r349377228) that Kubeops should refuse to start if an invalid driver type has been specified. I'll be more than happy to return an error instead.

### DIFF
--- a/cmd/kubeops/main.go
+++ b/cmd/kubeops/main.go
@@ -50,10 +50,13 @@ func main() {
 		Timeout:   timeout,
 	}
 
-	// Will panic below if an invalid driver type is provided.
 	driverType := defaultHelmDriver
 	if helmDriverArg != "" {
-		driverType = agent.ParseDriverType(helmDriverArg)
+		d, err := agent.ParseDriverType(helmDriverArg)
+		if err != nil {
+			panic(err.Error())
+		}
+		driverType = d // Necessary detour to please typechecker.
 	}
 	withAgentConfig := handler.WithAgentConfig(driverType, options)
 	r := mux.NewRouter()
--- a/cmd/kubeops/main.go
+++ b/cmd/kubeops/main.go
@@ -54,7 +54,7 @@ func main() {
 	if helmDriverArg != "" {
 		d, err := agent.ParseDriverType(helmDriverArg)
 		if err != nil {
-			panic(err.Error())
+			panic(err)
 		}
 		driverType = d // Necessary detour to please typechecker.
 	}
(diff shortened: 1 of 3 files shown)


cmd/kubeops/main.go
```diff
@@ -50,10 +50,13 @@ func main() {
 		Timeout:   timeout,
 	}
 
-	// Will panic below if an invalid driver type is provided.
 	driverType := defaultHelmDriver
 	if helmDriverArg != "" {
-		driverType = agent.ParseDriverType(helmDriverArg)
+		d, err := agent.ParseDriverType(helmDriverArg)
+		if err != nil {
+			panic(err.Error())
```
andresmgot: you can just do `panic(err)`

pkg/agent/agent.go
```diff
@@ -90,16 +91,16 @@ func createStorage(driverType DriverType, namespace string, clientset *kubernete
 	return store
 }
 
-func ParseDriverType(raw string) DriverType {
+func ParseDriverType(raw string) (DriverType, error) {
```
andresmgot: this function is simple but it'd be good to have a simple test

SimonAlling:
> LGTM, I have a couple of small comments, let me know what you think

I have fixed those issues now! :smiley:

pkg/agent/agent_test.go
```diff
@@ -146,3 +146,50 @@ func TestListReleases(t *testing.T) {
...
+			driverType, err := ParseDriverType(tc.input)
+			if err != nil {
+				t.Errorf("%v", err)
+			} else if driverType != tc.output {
+				t.Errorf("expected: %s, actual: %s", tc.output, driverType)
+			}
+		})
+	}
+
+	invalidTestCase := "andresmgot"
```
andresmgot: :joy:

PR #1367, opened 2019-12-11
Asset Syncer: Fetch files for postgres

andresmgot:
### Description of the change

The goal of this PR is to add support for fetching chart files for the PosgreSQL driver as we do for MongoDB. For doing that:

 - I have moved common logic (logic that actually downloads files and icons from the source) to the common `utils`. For doing so I have added the struct `fileImporter` that defines the common methods.
 - The struct `fileImporter` above has, as a property, an `assetManager` which has been extended to define the methods related to the database: `updateIcon`, `filesExist` and `insertFiles`.
 - Each driver (`mongodb_utils` and `postgresql_utils`) define the methods above.

Note: I found that I was not closing the returned `rows` given by the `pq` library so I fixed that here.

### Applicable issues

  - ref #651

### DIFF
--- a/cmd/asset-syncer/utils.go
+++ b/cmd/asset-syncer/utils.go
@@ -19,6 +19,7 @@ package main
 import (
 	"archive/tar"
 	"bytes"
+	"compress/gzip"
 	"crypto/sha256"
 	"crypto/tls"
 	"crypto/x509"
@@ -31,8 +32,10 @@ import (
 	"os"
 	"path"
 	"strings"
+	"sync"
 	"time"
 
+	"github.com/disintegration/imaging"
 	"github.com/ghodss/yaml"
 	"github.com/jinzhu/copier"
 	"github.com/kubeapps/common/datastore"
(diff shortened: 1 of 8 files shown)


cmd/asset-syncer/utils.go
```diff
@@ -257,3 +263,192 @@ func initNetClient(additionalCA string) (*http.Client, error) {
 		},
 	}, nil
 }
+
+type fileImporter struct {
+	manager assetManager
+}
+
+func (f *fileImporter) fetchFiles(charts []chart) {
+	// Process 10 charts at a time
```
andresmgot: no need to review this, the code is the same than before (in mongodb_utils)

cmd/asset-syncer/utils_test.go
```diff
@@ -489,3 +494,135 @@ func Test_newManager(t *testing.T) {
 	}
 
 }
+
+func Test_fetchAndImportIcon(t *testing.T) {
+	t.Run("no icon", func(t *testing.T) {
```
andresmgot: same, this is just the same test replaced

andresmgot:
> LGTM. I'm a little wary of our tests not actually testing with a postgres db, or at least having an option to do so (to verify they do what's expected rather than having to test things manually).

Yes, right now we are not actually testing but once this works end-to-end I will enable the browser tests which will cover most of the code (the same thing we do now with mongodb).

absoludity:
> > LGTM. I'm a little wary of our tests not actually testing with a postgres db, or at least having an option to do so (to verify they do what's expected rather than having to test things manually).
> 
> Yes, right now we are not actually testing but once this works end-to-end I will enable the browser tests which will cover most of the code (the same thing we do now with mongodb).

Great, but I'd still be keen for unit tests that make it trivial for others to make changes here with confidence in the future. Making a change and adding mock expected won't help there, and browser tests are not for iterating TDD.

PR #1368, opened 2019-12-12
Remove panics from pkg/agent

absoludity:
### Description of the change

Small fly-by for some panics in a non-main module. Updated to handle the errors (similar to @SimonAlling 's PR).


pkg/agent/agent.go
```diff
@@ -85,10 +89,9 @@ func createStorage(driverType DriverType, namespace string, clientset *kubernete
 		d := driver.NewMemory()
 		store = storage.Init(d)
 	default:
-		// No (real) enums/ADTs in Go, so no static guarantee against this case.
-		panic("Invalid Helm driver type: " + driverType)
+		return nil, fmt.Errorf("Invalid Helm drive type: %q", driverType)
```
SimonAlling: I personally don't fancy this change. The point of the `DriverType` type is to validate the driver type once and then trust it to be validated. [_Parse, don't validate_](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/), as suggested by Alexis King. I put the `panic` there because it is *possible* to sidestep the type with a string literal, but that *should never happen*™ in practice.

Surely there must be some way to express a plain old sum type in Go? We're not JavaScript, are we?

pkg/agent/agent.go
```diff
@@ -53,24 +54,27 @@ func ListReleases(actionConfig *action.Configuration, namespace string, listLimi
 	return appOverviews, nil
 }
 
-func NewActionConfig(driver DriverType, token, namespace string) *action.Configuration {
+func NewActionConfig(driver DriverType, token, namespace string) (*action.Configuration, error) {
 	actionConfig := new(action.Configuration)
 	config, err := rest.InClusterConfig()
 	if err != nil {
-		panic(err.Error())
+		return nil, err
```
SimonAlling: I like this change!

pkg/agent/agent.go
andresmgot: I'd still see okay to return an error here (which is idiomatic for Go). The function didn't success to create the storage driver but the caller may still wrap the error to give more information or even ignore it. Even if this is not reachable with the current code (which I believe is the case), it's easier to read this way.

SimonAlling: On second thought, even though the `panic` shouldn't™ be reachable today, it will be as soon as another `DriverType` member is added.

Is it really not possible to write a "true" total function from a sum type in Go?

absoludity: I'm not sure if you're saying you're OK with this or not in this second comment. I'm going to land but have a go at something that I think is what you're after in a follow-up. Hope that's OK.

SimonAlling: I am OK with it since I realized that the compiler won't yell at us should we add another `DriverType` member. If you find out a way to write total functions in Go, please enlighten me!

PR #1369, opened 2019-12-12
Improve TestListReleases for /pkg/agent

latiif:
- Check the content of returned listed releases (not just size comparison).
- Complete information about returned apps in test cases.
- Add test case for release with same name, but different versions and namespaces.

### Description of the change
Build and expand on the existing test of `TestListReleases` in `/pkg/agent`, by ensuring the equivalence between what is expected and the returned values of `ListReleases`.

The current test checks only if `|G| = |W|` where `W` is the expected result and `G` is the result of `ListReleases`.
### Benefits

- By populating the expected results of the test cases with `namespace` and `version`, better distinction between releases is achieved.

- This PR adds an additional constraint, namely `∀ g ∊ G : g ∊ W`, which together with the previous one i.e. `|G| = |W|`, ensures that `G = W`.

- The new test case tests the scenario where `ListReleases` returns two apps with same name but different namespaces and versions.

### Applicable issues

  - builds on  #1359

### DIFF
--- a/pkg/agent/agent_test.go
+++ b/pkg/agent/agent_test.go
@@ -2,6 +2,7 @@ package agent
 
 import (
 	"io/ioutil"
+	"strconv"
 	"testing"
 
 	"helm.sh/helm/v3/pkg/action"
@@ -66,6 +67,15 @@ func makeReleases(t *testing.T, config *Config, rels []releaseStub) {
 	}
 }
 
+func strtoi(str string) int {
+	i, err := strconv.Atoi(str)
+	if err != nil {
+		return 0
+	}
+
+	return i
+}
+
 func TestListReleases(t *testing.T) {
 	testCases := []struct {
 		name         string
@@ -86,12 +96,18 @@ func TestListReleases(t *testing.T) {
 			expectedApps: []proxy.AppOverview{
 				proxy.AppOverview{
 					ReleaseName: "airwatch",
+					Namespace:   "default",
+					Version:     "1",
 				},
 				proxy.AppOverview{
 					ReleaseName: "wordpress",
+					Namespace:   "default",
+					Version:     "1",
 				},
 				proxy.AppOverview{
 					ReleaseName: "not-in-default-namespace",
+					Namespace:   "other",
+					Version:     "1",
 				},
 			},
 		},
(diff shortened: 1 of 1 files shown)


pkg/agent/agent_test.go
```diff
@@ -140,9 +183,25 @@ func TestListReleases(t *testing.T) {
...
+
+			//Check if expected and returned contain the same releases, given they have same size
+			m := make(map[releaseStub]bool)
+
+			for _, eapp := range tc.expectedApps {
+				m[releaseStub{eapp.ReleaseName, eapp.Namespace, strtoi(eapp.Version)}] = true
+			}
+
+			for _, app := range apps {
+				rs := releaseStub{app.ReleaseName, app.Namespace, strtoi(app.Version)}
```
absoludity: So I think this is comparing a manually constructed `releaseStub` made from the output `AppOverview`, with another manually constructed `releaseStub` created from the expected `AppOverview`. Why not just compare the output `AppOverview` with the expected `AppOverview` directly? (bonus that they then have the same types).

Also, not sure if it could have been used here, but the `cmp` from the `go-cmp` module is quite nice for comparisons... you might also be able to do something like:
```
import "github.com/google/go-cmp/cmp"
...
    if !cmp.Equal(tc.expectedApps, apps) {
        t.Errorf(cmp.Diff(tc.expectedApps, apps))
    }
```

latiif: Absolutely, comparing `AppOverView`'s makes more sense.
I'll take a look at `go-cmp`, but I have a gut feeling that additional fields need to be filled in the expected `AppOverview` array.

I'll create a new PR since this one is merged & closed.

andresmgot: I tried that while reviewing this PR and it doesn't work (that's why I assumed it was doing the comparison manually). There are still some fields that are not in the expected (like the `status` of the release) and also the order in the slices changes sometimes so the comparison fails.

latiif: I was able to fix it by using the library @absoludity mentioned in PR #1379.

It was done by mapping a unique id (presumably no two releases can have the same name in the same namespaces in the expected results) to a pointer to the expected `AppOverview`.

Then, by generating a unique id for the attained result, looking up it in the mapping, and then and only then, perform deep equality check on both objects.

absoludity: Looks like it got even simpler than that too. Great, thanks @latiif

PR #1373, opened 2019-12-13
Update GetDashToken.cmd script, token.txt should not be deleted

bserratrice:
### Description of the change
Documentation update

### Benefits
Windows script to generate token is currently failing.

### Possible drawbacks
N/A

### Applicable issues
N/A

### Additional information
N/A

### DIFF
--- a/docs/user/getting-started.md
+++ b/docs/user/getting-started.md
@@ -55,8 +55,8 @@ REM Get the Base64 encoded token
 kubectl get secret %ks% -o jsonpath={.data.token} > b64.txt
 
 REM Decode The Token
-DEL token.txt
 certutil -decode b64.txt token.txt
+DEL b64.txt
 ```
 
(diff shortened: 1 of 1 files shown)


docs/user/getting-started.md
```diff
@@ -55,8 +55,8 @@ REM Get the Base64 encoded token
 kubectl get secret %ks% -o jsonpath={.data.token} > b64.txt
 
 REM Decode The Token
-DEL token.txt
```
absoludity: I don't have a windows box to test on (though I'm sure I could online somewhere, but), this line looks as though it was intended to delete a previously generated `token.txt`, whereas in your change it looks like you understood it to be deleting the intermediate `b64.txt` file (which may also be valid).

Does `certutil -decode` fail if the output file already exists? (ie. if you run this script a second time).

I assume from your change that the failure you were seeing is that it can't delete a file that doesn't exist? (the first time this is run).

PR #1374, opened 2019-12-13
Remove the need to validate DriverType

absoludity:
### Description of the change

In https://github.com/kubeapps/kubeapps/pull/1368#discussion_r356994225 @SimonAlling (fairly) didn't like the fact that we were needing to validate the `DriverType` after we'd already parsed it correctly.

This change removes the string `DriverType` constants, replacing them with a `StorageForDriver` function type.

### Benefits
This way, when the driver is parsed we return the corresponding `StorageForDriver` function which can be used to obtain the driver-specific storage, so we no longer need to validate the driver type.

Not sure if this fits what you want @SimonAlling ? Let me know :)

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -94,7 +94,7 @@ func ParseDriverType(raw string) (StorageForDriver, error) {
 	case "memory":
 		return StorageForMemory, nil
 	default:
-		return StorageForMemory, errors.New("Invalid Helm driver type: " + raw)
+		return nil, errors.New("Invalid Helm driver type: " + raw)
 	}
 }
 
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -2,7 +2,6 @@ package agent
 
 import (
 	"errors"
-	"fmt"
 	"strconv"
 
 	"github.com/kubeapps/kubeapps/pkg/proxy"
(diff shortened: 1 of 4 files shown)


pkg/agent/agent.go
```diff
@@ -63,47 +77,24 @@ func NewActionConfig(driver DriverType, token, namespace string) (*action.Config
...
+		return StorageForSecrets, nil
 	case "configmap", "configmaps":
-		return ConfigMap, nil
+		return StorageForConfigMaps, nil
 	case "memory":
-		return Memory, nil
+		return StorageForMemory, nil
 	default:
-		return Memory, errors.New("Invalid Helm driver type: " + raw)
+		return StorageForMemory, errors.New("Invalid Helm driver type: " + raw)
```
SimonAlling: I feel like it might make more sense to return `nil` instead of `StorageForMemory` here, now that `ParseDriverType` returns a function instead of a quasi-enum/string. Can't say whether that should be part of this PR or a new one though.

andresmgot: Agree, I'd return nil here (instead of a valid driver type)

absoludity: I had thought so to, but incorrectly also thought it was not possible. But you're right, the spec says as much too if I'd checked https://golang.org/ref/spec#Function_types . Updated the test to include checking for nil too. Thanks.

Issue #1375, opened 2019-12-13
 container "kubeops" in pod "kubeapps-internal-kubeops-68bf77974-92zhj" is waiting to start: trying and failing to pull image


OPSTime:
**cat kubeapps.custom.values**

> useHelm3: true

**install kubeapps**

> $ helm install  kubeapps bitnami/kubeapps  -n kubeapps -f kubeapps.custom.values 

 **List all pods** 

> NAME                                                          READY   STATUS             RESTARTS   AGE
> apprepo-sync-bitnami-1576206000-hm5xb                         0/1     Completed          0          27m
> apprepo-sync-bitnami-1576206600-m9449                         0/1     Completed          0          17m
> apprepo-sync-bitnami-1576207200-zwc4d                         0/1     Completed          0          7m35s
> apprepo-sync-bitnami-78sxr-v2w2h                              0/1     Completed          0          86m
> apprepo-sync-incubator-1576206000-dk77s                       0/1     Completed          0          27m
> apprepo-sync-incubator-1576206600-925jr                       0/1     Completed          0          17m
> apprepo-sync-incubator-1576207200-25zds                       0/1     Completed          0          7m35s
> apprepo-sync-incubator-wsqhm-29l7x                            0/1     Completed          2          86m
> apprepo-sync-stable-1576205400-wztqs                          0/1     Completed          2          37m
> apprepo-sync-stable-1576206600-qcr94                          0/1     Completed          4          17m
> apprepo-sync-stable-1576207200-rhmjp                          0/1     Completed          1          7m35s
> apprepo-sync-svc-cat-1576206000-hp2zp                         0/1     Completed          0          27m
> apprepo-sync-svc-cat-1576206600-qf4fs                         0/1     Completed          0          17m
> apprepo-sync-svc-cat-1576207200-ctk2s                         0/1     Completed          0          7m35s
> apprepo-sync-svc-cat-nndvz-cd5zg                              0/1     Completed          0          86m
> kubeapps-664c5849d4-cp64n                                     0/1     CrashLoopBackOff   21         86m
> kubeapps-664c5849d4-ncnh7                                     0/1     CrashLoopBackOff   21         86m
> kubeapps-internal-apprepository-controller-67696766b6-khv2g   1/1     Running            0          86m
> kubeapps-internal-chartsvc-677b745bb9-prhlz                   1/1     Running            0          86m
> kubeapps-internal-chartsvc-677b745bb9-xsbmf                   1/1     Running            0          86m
[...]

PR #1376, opened 2019-12-13
Move useHelm3 switch to end until we want it so visible.

absoludity:
### Description of the change
Just noticed this when upgrading Kubeapps itself:

useHelm3-first-option-in-values

Not a big deal, but we probably don't want it to be the first thing a user sees until it actually works.

### DIFF
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -581,6 +581,7 @@ authProxy:
       cpu: 25m
       memory: 32Mi
 
-# true for Helm 3; false for Helm 2:
-useHelm3: false # DOES NOT WORK YET; will act as a feature flag later on when we have Helm 3 support.
+## The useHelm3 feature flag is currently only for developing the in-progress (and incomplete) Helm3 support.
+## If you set it to true, Kubeapps will not work at present.
+useHelm3: false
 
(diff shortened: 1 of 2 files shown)


chart/kubeapps/values.yaml
```diff
@@ -583,3 +580,7 @@ authProxy:
     requests:
       cpu: 25m
       memory: 32Mi
+
+# true for Helm 3; false for Helm 2:
+useHelm3: false # DOES NOT WORK YET; will act as a feature flag later on when we have Helm 3 support.
```
andresmgot: I would move the `DOES NOT WORK YET` ( or something similar to the title instead of being an inline comment)

SimonAlling: Do you mean like this?

```diff
# true for Helm 3; false for Helm 2:
-useHelm3: false # DOES NOT WORK YET; will act as a feature flag later on when we have Helm 3 support.
+# DOES NOT WORK YET; will act as a feature flag later on when we have Helm 3 support.
+useHelm3: false
```

SimonAlling: Hm ... that feels pretty inline as well when I think about it.

absoludity: > Hm ... that feels pretty inline as well when I think about it.

I think he meant simply not having the comment on the same line after the text.

Anyway, updated and expanded the comment a bit).

absoludity:
Side point: the timing of this was interesting: right after I noticed this while upgrading Kubeapps and created this issue, I saw that a user had created #1375 after they tried the flag too.

PR #1379, opened 2019-12-13
Deep Equality Check in /pkg/agent test

latiif:
### Description of the change
A follow up PR for #1369, where the comparisons between expected and attained apps are performed on the actual representation rather than a test stub.

Completed fields 'version' and 'icon' for expected results.
### Benefits

Making sure the equality checks are performed on the entire structure of the attained results.

### Applicable issues
 Related PR #1369 based on the discussion that followed with @absoludity.
### Additional information

No performance drawbacks noticed.


go.mod
```diff
@@ -8,6 +8,7 @@ require (
 	github.com/Masterminds/semver v1.5.0 // indirect
 	github.com/Masterminds/sprig v2.22.0+incompatible // indirect
 	github.com/arschles/assert v1.0.0
+	github.com/davecgh/go-spew v1.1.1
```
latiif: This one slipped through the cracks :slightly_frowning_face:. Needed it for debugging.

pkg/agent/agent_test.go
```diff
@@ -188,24 +206,33 @@ func TestListReleases(t *testing.T) {
...
-				rs := releaseStub{app.ReleaseName, app.Namespace, strtoi(app.Version)}
-				if _, ok := m[rs]; !ok {
-					t.Errorf("got: %v, want: %v", rs, "None")
+				appIdentity := getAppIdentity(app)
+				if expectedApp, ok := m[appIdentity]; !ok {
+					t.Errorf("got: %v, want: %v", &app, "None")
+				} else if !cmp.Equal(*expectedApp, app) {
+					t.Errorf(cmp.Diff(*expectedApp, app))
 				}
 			}
```
andresmgot: you can just do:

```
			if !cmp.Equal(tc.expectedApps, apps) {
				t.Errorf(cmp.Diff(tc.expectedApps, apps))
			}
```

If you reorder the expected apps

go.mod
andresmgot: you should be able to remove it now

pkg/agent/agent_test.go
latiif: Sure thing. It can be done that way with the pair `app.ReleaseName` and `app.Namespace` as keys to sort on.

Did I understand correctly?

andresmgot: nope, I meant that if you reorder the apps in `tc.expectedApps` to match the `apps` you don't need the map nor the iteration over the apps, you can just compare the arrays

pkg/agent/agent_test.go
```diff
@@ -188,24 +206,22 @@ func TestListReleases(t *testing.T) {
...
 
 		})
 	}
 }
 
+// getAppIdentity concatenates the release name and namespace of an app
+func getAppIdentity(app proxy.AppOverview) string {
+	return fmt.Sprintf("%v %v", app.ReleaseName, app.Namespace)
+}
```
absoludity: This looks like it was left over from some testing or earlier sorting? I don't see it being used.

andresmgot:
Hi @latiif, it seems that you have rebased master or something like that and now there is a bunch of commits here that are not related to your change. Can you remove those?

Note: When getting the latest changes from master is better to `merge` that branch instead to avoid this issue.

pkg/agent/agent_test.go
```diff
@@ -188,17 +205,10 @@ func TestListReleases(t *testing.T) {
...
-
-			for _, app := range apps {
-				rs := releaseStub{app.ReleaseName, app.Namespace, strtoi(app.Version)}
-				if _, ok := m[rs]; !ok {
-					t.Errorf("got: %v, want: %v", rs, "None")
+			//Deep equality check of expected aginst attained result
+			for i := 0; i < len(apps); i++ {
+				if !cmp.Equal(apps[i], tc.expectedApps[i]) {
+					t.Errorf(cmp.Diff(apps[i], tc.expectedApps[i]))
 				}
```
andresmgot: you still don't need the `for` loop :)

```
if !cmp.Equal(apps, tc.expectedApps) {
  t.Errorf(cmp.Diff(apps, tc.expectedApps))
}
```

should work for you now

absoludity:
So it looks like the memory driver does not return results with consistent ordering? At least, we're seeing an ordering failure in tests occasionally it seems https://circleci.com/gh/kubeapps/kubeapps/14014

PR #1381, opened 2019-12-13
Refactor database methods for assetsvc

andresmgot:
### Description of the change

Define the `assetManager` interface and move the database related methods to that interface. This enable us to start working on the same methods for postgresql.

### Applicable issues

  - ref #651

### DIFF
--- a/cmd/assetsvc/postgresql_utils.go
+++ b/cmd/assetsvc/postgresql_utils.go
@@ -17,49 +17,21 @@ limitations under the License.
 package main
 
 import (
-	"database/sql"
-	"fmt"
-	"strings"
-
 	"github.com/kubeapps/common/datastore"
 	"github.com/kubeapps/kubeapps/cmd/assetsvc/models"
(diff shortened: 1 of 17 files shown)


cmd/assetsvc/postgresql_utils.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+func (m *postgresAssetManager) getChartsWithFiltes(name, version, appVersion string) ([]*models.Chart, error) {
+	// TODO: Implement this!
+	return nil, nil
+}
+
+func (m *postgresAssetManager) searchCharts(query, repo string) ([]*models.Chart, error) {
+	// TODO: Implement this!
+	return nil, nil
```
absoludity: Fine either way, but when adding stubs like the above, I'd `return nil, fmt.Errorf("Unimplemented")` so you can't ever mistake an empty result for an implemented function.

andresmgot: good point

PR #1385, opened 2019-12-13
DRY up proxy_pass configuration

SimonAlling:
### Description of the change

It makes the `proxy_pass` configuration for Tiller Proxy or Kubeops (depending on the value of `useHelm3`) more DRY.

### Benefits

Suddenly, I couldn't get Kubeapps to start with `useHelm3: true`, making me spend several hours trying various more or less random fixes. (I thought I had messed something up with Docker or Minikube, so I didn't think of bisecting or inspecting the YAML files.) The reason turned out to be [this unconditional `proxy_pass` directive](https://github.com/kubeapps/kubeapps/pull/1347/files#diff-b0b5b871c8bf5a72f95d417c97266ae0R100) introduced in #1347 and fixed by me in #1382.

This PR is intended to prevent something similar from happening again; it's not at all obvious that the `proxy_pass` directives must be conditional.

### Additional information

Might be interesting to @lindhe and @absoludity.

### DIFF
--- a/chart/kubeapps/templates/_helpers.tpl
+++ b/chart/kubeapps/templates/_helpers.tpl
@@ -144,6 +144,17 @@ Create name for the frontend config based on the fullname
 {{- end -}}
 
 {{/*
+Create proxy_pass for the frontend config based on the useHelm3 flag
+*/}}
+{{- define "kubeapps.frontend-config.proxy_pass" -}}
+{{- if .Values.useHelm3 -}}
+http://{{ template "kubeapps.kubeops.fullname" . }}:{{ .Values.kubeops.service.port }}
+{{- else -}}
+http://{{ template "kubeapps.tiller-proxy.fullname" . }}:{{ .Values.tillerProxy.service.port }}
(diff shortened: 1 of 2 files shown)


absoludity:
> The reason turned out to be [this unconditional `proxy_pass` directive](https://github.com/kubeapps/kubeapps/pull/1347/files#diff-b0b5b871c8bf5a72f95d417c97266ae0R100) introduced in #1347 and fixed by me in #1382.

Sorry - I'd intentionally started that work in `tiller-proxy` so as to not get in your way, but failed to see it'd cause nginx to fail when `useHelm3` was set. Perhaps we should add a stub CI test with `useHelm3=true` which at least would have caught this before I landed my branch?

SimonAlling:
You're welcome! Let me also emphasize that I don't think you should be sorry. It was far from obvious that the `proxy_pass` directives must be conditional.

> Perhaps we should add a stub CI test with `useHelm3=true` which at least would have caught this before I landed my branch?

Yeah, I think that would be useful!

PR #1386, opened 2019-12-16
Refactor model definition

andresmgot:
### Description of the change

The definition of a `chart`, `repo`, etc (everything that's stored in the database) was duplicated in both the `assetsvc` and the `asset-syncer` causing some issues because of the JSON serialization. This PR moves that model definition to a common package.

### Applicable issues

  - ref #651


cmd/asset-syncer/utils_test.go
```diff
@@ -581,20 +582,7 @@ func Test_fetchAndImportFiles(t *testing.T) {
...
-		assert.NoErr(t, err)
-		m.AssertExpectations(t)
-	})
-
-	t.Run("valid tarball", func(t *testing.T) {
-		netClient = &goodTarballClient{c: charts[0]}
-		m := mock.Mock{}
-		m.On("One", mock.Anything).Return(errors.New("return an error when checking if files already exists to force fetching"))
-		chartFilesID := fmt.Sprintf("%s/%s-%s", charts[0].Repo.Name, charts[0].Name, cv.Version)
-		m.On("UpsertId", chartFilesID, chartFiles{chartFilesID, testChartReadme, testChartValues, testChartSchema, charts[0].Repo, cv.Digest})
```
absoludity: Why this removal of a test?

andresmgot: completely overlooked this one, this test doesn't have to be removed :/

PR #1389, opened 2019-12-16
Copy CA certs to Kubeops container

SimonAlling:
_"That line can't be very important!"_ — me, 2019


SimonAlling:
No – it was entirely my fault. Thanks to _you_ for realizing what the problem was. :)

PR #1392, opened 2019-12-17
Redefine Repo struct

andresmgot:
### Description of the change

After #1386 (`models` refactor), the `Repo` struct contained auth info that was stored in the `charts` database and was being returned for each chart request. This `auth` info should not be send to the client. To fix this, the `asset-syncer` will work with the struct `RepoInternal` (which has the auth info but it's not stored in the database) and the assetsvc will continue to work with the `Repo` struct.

Apart from that this PR includes a fix that was preventing the postgres driver to properly skip files import.

### DIFF
--- a/cmd/asset-syncer/postgresql_utils.go
+++ b/cmd/asset-syncer/postgresql_utils.go
@@ -172,14 +172,14 @@ func (m *postgresAssetManager) updateIcon(data []byte, contentType, ID string) e
 
 func (m *postgresAssetManager) filesExist(chartFilesID, digest string) bool {
 	rows, err := m.DB.Query(
-		fmt.Sprintf("SELECT * FROM %s WHERE info ->> 'ID' = $1 AND info ->> 'digest' = $2", chartFilesTable),
+		fmt.Sprintf("SELECT * FROM %s WHERE chart_files_id = $1 AND info ->> 'Digest' = $2", chartFilesTable),
 		chartFilesID,
 		digest,
 	)
 	if rows != nil {
 		defer rows.Close()
 	}
-	return err == nil
+	return err == nil && rows.Next()
 }
 
 func (m *postgresAssetManager) insertFiles(chartFilesID string, files models.ChartFiles) error {
--- a/cmd/asset-syncer/postgresql_utils.go
+++ b/cmd/asset-syncer/postgresql_utils.go
@@ -176,10 +176,12 @@ func (m *postgresAssetManager) filesExist(chartFilesID, digest string) bool {
 		chartFilesID,
 		digest,
 	)
+	hasEntries := false
 	if rows != nil {
 		defer rows.Close()
+		hasEntries = rows.Next()
 	}
-	return err == nil && rows.Next()
+	return err == nil && hasEntries
 }
 
 func (m *postgresAssetManager) insertFiles(chartFilesID string, files models.ChartFiles) error {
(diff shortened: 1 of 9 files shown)


cmd/asset-syncer/postgresql_utils.go
```diff
@@ -172,14 +172,14 @@ func (m *postgresAssetManager) updateIcon(data []byte, contentType, ID string) e
...
-		fmt.Sprintf("SELECT * FROM %s WHERE info ->> 'ID' = $1 AND info ->> 'digest' = $2", chartFilesTable),
+		fmt.Sprintf("SELECT * FROM %s WHERE chart_files_id = $1 AND info ->> 'Digest' = $2", chartFilesTable),
 		chartFilesID,
 		digest,
 	)
 	if rows != nil {
 		defer rows.Close()
 	}
-	return err == nil
+	return err == nil && rows.Next()
```
andresmgot: this was the code that was failing to check if the files already exist

cmd/asset-syncer/postgresql_utils.go
```diff
@@ -172,14 +172,16 @@ func (m *postgresAssetManager) updateIcon(data []byte, contentType, ID string) e
...
 		chartFilesID,
 		digest,
 	)
+	hasEntries := false
 	if rows != nil {
 		defer rows.Close()
+		hasEntries = rows.Next()
 	}
-	return err == nil
+	return err == nil && hasEntries
```
andresmgot: this was the code that was failing to check if the files already exist

cmd/asset-syncer/postgresql_utils_test.go
```diff
@@ -124,11 +124,11 @@ func Test_PGfilesExist(t *testing.T) {
 	pgManager := &postgresAssetManager{man}
 	m.On(
 		"Query",
-		`SELECT * FROM files WHERE info ->> 'ID' = $1 AND info ->> 'digest' = $2`,
+		`SELECT * FROM files WHERE chart_files_id = $1 AND info ->> 'Digest' = $2`,
 		[]interface{}{id, digest},
 	)
 	exists := pgManager.filesExist(id, digest)
-	if exists != true {
+	if exists != false {
```
absoludity: The name of this test implies it's testing that it can (successfully) verify that a file exists, but the switch in value here implies it will now call `t.Errorf` if the file exists? It's pretty confusing (imo). The current logic as it is with this change (even though I can't make semantic sense out of it) could be just:
```
if pgManager.filesExist(id, digest) {
    t.Errorf("Failed to check if file exists")
}
```

andresmgot: The goal of the name of the test is to say that it's testing the method `fileExists` (which may appear to say that the file has to exist). The change in the result is caused by the change that we only mark it as valid if the returned rows has at least lenght 1. Before we were just checking if the query returned an error to assume that in that case the file exist.

That's why the result of this test changed. The error keeps being `nil` but the mock query is not returning any row.

It's true that the test would be more valuable if we are actually be able to verify if a file exists. We can do that with the library I am using in #1394 which at least lets you return rows. I will do so once I merge that PR in this branch.

andresmgot:
> So with this we'll go back to not storing any repo auth in the storage backend (pg db) right?

Correct.

cmd/asset-syncer/postgresql_utils_test.go
absoludity: Thanks, that last commit makes the test much clearer.

PR #1394, opened 2019-12-17
Query methods for PostgreSQL

andresmgot:
### Description of the change

Methods to query one or several charts using PostgreSQL. Include tests for `postgresql_utils.go`


andresmgot:
> Not a fan of testing the db queries by a mock db using expectations. Yes, it ensures the current code is tested according to the actual implementation (I'm assuming you've ensured this), but it does not help someone else later updating this code. They will have to update the code, verify it works IRL, grab the resulting queries from the IRL verification, update the mock expectations. If we were using a real db, they can instead just update the code and verify the unit tests still pass (or update the expected _result_).

I agree that be nicer but I am not sure it's worth it. We would need to setup an infra to spin up a container with postgres, configure it, create the database (and tables) and start creating demo data to query it. That's not something trivial to implement. Also, the queries are not something that will change a lot in the future and in any case, the person changing them would need to verify if that works IRL anyway.

PR #1397, opened 2019-12-18
Create handler for app repositories backend API.

absoludity:
<!--
 Before you open the request please review the following guidelines and tips to help it be more easily integrated:

 - Describe the scope of your change - i.e. what the change does.
 - Describe any known limitations with your change.
 - Please run any tests or examples that can exercise your modified code.

 Thank you for contributing!
 -->

### Description of the change

Fills in the stub implementation of the `Create` handler for the app repository backend API
<!-- Describe the scope of your change - i.e. what the change does. -->

Note: this handler is similar in nature to the helm3 work in the respect that it picks up the in-cluster k8s config and uses that with the user token to access the API. But note that it's structured so that:

- the k8s config, with the blanked `BearerToken` is read from the system and created once only when the handler is created,
- an actual working config with the user token is created for each request. This is a copy of the blanked k8s config with the token set.
- it is tested using the (generated) fake clientset for the resource by injecting a dummy `clientsetForConfig` function when testing which just returns the fake. This ensures that external use of this handler will always have the real `clientsetForConfig` function, and only (internal) tests can set a dummy function. In this sense, it may be useful to @SimonAlling as a way to structure the handler code to be testable (warning: client-go testing with fake clientsets is pretty verbose, but once the parts are in place, pretty easy to work with - the setup is the painful part :/)

### Applicable issues

<!-- Enter any applicable Issues here (You can reference an issue using #) -->
  - Ref #1173 

### Additional information

I've added a `--kubeapps-namespace` flag which is effectively a feature flag. Without this being set, a response comes back as:
```
< HTTP/1.1 401 Unauthorized                                                                                                                                                                                                       
< Content-Length: 74                                                                                                                                                                                                              
< Content-Type: text/plain; charset=utf-8                                                                                                                                                                                         < Date: Wed, 18 Dec 2019 04:30:27 GMT                                                                                                                                                                                             < Server: nginx/1.16.1                                                                                                                                                                                                            < X-Content-Type-Options: nosniff                                                                                                                                                                                                 <                                                                                                                                                                                                                                 kubeappsNamespace must be configured to enable app repository handler                                                                                                                                                             EOF                                                                                                                                                                                                                               
```

FWIW, I need to investigate more on the use of the in-cluster config, as when I did an IRL test with the feature switched on, authing with a cookie for my `kubeapps-operator` user, I see the following unexpected result:

```
$ curl -XPOST 'http://localhost:3000/api/v1/apprepositories' -H 'User-Agent: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:70.0) Gecko/20100101 Firefox/70.0' -H 'Accept: application/json, text/p
lain, */*' -H 'Accept-Language: en-US,en;q=0.5' --compressed -H 'DNT: 1' -H 'Connection: keep-alive' -H 'Referer: http://localhost:3000/' -H 'Cookie: _oauth2_proxy=<redacted>' -vvv --data '{"appRepository": {"name": "test-repo", "url": "http://example.com/test-repo"}}'
...
> Content-Length: 79                                                                                                                                                                                                              
> Content-Type: application/x-www-form-urlencoded                                                                                                                                                                                 
>                                                                                                                                                                                                                                 
* upload completely sent off: 79 out of 79 bytes                                                                                                                                                                                  
< HTTP/1.1 403 Forbidden                                                                                                                                                                                                          
< Content-Encoding: gzip                                                                                                                                                                                                          
< Content-Type: text/plain; charset=utf-8                                                                                                                                                                                         
< Date: Wed, 18 Dec 2019 04:32:37 GMT                                                                                                                                                                                             
< Server: nginx/1.16.1                                                                                                                                                                                                            
< X-Content-Type-Options: nosniff                                                                                                                                                                                                 
< Content-Length: 156                                                                                                                                                                                                             
<                                                                                                                                                                                                                                 
apprepositories.kubeapps.com is forbidden: User "system:serviceaccount:kubeapps:kubeapps-internal-tiller-proxy" cannot create resource "apprepositories" in API group "kubeapps.com" in the namespace "kubeapps"                  
```
I don't think this should stop the PR from landing (since it's behind the extra flag which is not set in the chart yet), but tomorrow I plan to find out (a) why it assumes the user accessing is the internal-tiller-proxy service account (ie. does the in-cluster config have not only the token data, but the service account info set, which needs to be blanked out), and (b) why the token (or otherwise) is resulting in a 403.

Maybe it'll be obvious during review, but I need to run :)

### DIFF
--- a/cmd/tiller-proxy/internal/handler/apprepos_handler.go
+++ b/cmd/tiller-proxy/internal/handler/apprepos_handler.go
@@ -82,8 +82,9 @@ func NewAppRepositoriesHandler(kubeappsNamespace string) (*appRepositoriesHandle
 		return nil, err
 	}
 	return &appRepositoriesHandler{
-		config:             *config,
-		kubeappsNamespace:  kubeappsNamespace,
+		config:            *config,
+		kubeappsNamespace: kubeappsNamespace,
+		// See comment in the struct defn above.
 		clientsetForConfig: clientsetForConfig,
 	}, nil
 }
--- a/cmd/tiller-proxy/internal/handler/apprepos_handler.go
+++ b/cmd/tiller-proxy/internal/handler/apprepos_handler.go
@@ -17,18 +17,148 @@ limitations under the License.
 package handler
 
 import (
+	"encoding/json"
+	"io"
 	"net/http"
 
 	log "github.com/sirupsen/logrus"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+
+	"github.com/kubeapps/kubeapps/cmd/apprepository-controller/pkg/apis/apprepository/v1alpha1"
+	clientset "github.com/kubeapps/kubeapps/cmd/apprepository-controller/pkg/client/clientset/versioned"
+	"github.com/kubeapps/kubeapps/pkg/auth"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// AppRepositories handles http requests for operating on app repositories
+// appRepositories handles http requests for operating on app repositories
 // in Kubeapps, without exposing implementation details to 3rd party integrations.
-type AppRepositories struct{}
+type appRepositoriesHandler struct {
+	// The config set internally here cannot be used on its own as a valid
+	// token is required. Call-sites use ConfigForToken to obtain a valid
+	// config with a specific token.
+	config rest.Config
 
-func (a *AppRepositories) Create(w http.ResponseWriter, req *http.Request) {
-	log.Printf("Creating AppRepository")
-	// TODO: Create AppRepository using the k8s client.
+	// The namespace in which (currently) app repositories are created.
+	kubeappsNamespace string
+
+	// clientsetForConfig is a field on the struct only so it can be switched
+	// for a fake version when testing. NewAppRepositoryHandler sets it to the
+	// proper function below so that production code always has the real
+	// version (and since this is a private struct, external code cannot change
+	// the function).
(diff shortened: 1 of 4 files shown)


cmd/tiller-proxy/internal/handler/apprepos_handler.go
```diff
@@ -17,18 +17,148 @@ limitations under the License.
...
+		},
+	)
+	config, err := clientConfig.ClientConfig()
+	if err != nil {
+		return nil, err
+	}
+	return &appRepositoriesHandler{
+		config:             *config,
+		kubeappsNamespace:  kubeappsNamespace,
+		clientsetForConfig: clientsetForConfig,
```
andresmgot: I am not sure why this function is within the `appRepositoriesHandler` struct? As far as I can see it can be called as a normal method (it doesn't depend on the struct properties).

[edit] Got it, you need this because of the tests, to be able to override it. (A comment here explaining it would be nice?)

cmd/tiller-proxy/internal/handler/apprepos_handler.go
```diff
@@ -17,18 +17,148 @@ limitations under the License.
...
+	}
+
+	appRepo, err := appRepositoryForRequestData(req.Body)
+	if err != nil {
+		log.Infof("unable to decode: %v", err)
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+	// TODO(mnelson): validate both required data and request for index
+	// https://github.com/kubeapps/kubeapps/issues/1330
```
andresmgot: it's good that we have a place now to validate this data :+1:

cmd/tiller-proxy/internal/handler/apprepos_handler.go
absoludity: I added a `// See comment in the struct defn above` as there's a 5-liner explanation there :)

PR #1398, opened 2019-12-18
Create tables before syncing content

andresmgot:
### Description of the change

While the goal of this PR was to create programatically the tables used (if not exist), I found several small issues while testing this IRL so I added them to this PR:

 - Use the table names defined in `dbutils`
 - Fix a bug that was deleting the chart from the repositories not being evaluated
 - Remove a debug `fmt.Print`

Note: I am not testing the new code because a test mocking the result doesn't help much and also because this is a basic functionality that will be tested in the e2e tests.

### DIFF
--- a/cmd/asset-syncer/postgresql_utils.go
+++ b/cmd/asset-syncer/postgresql_utils.go
@@ -151,7 +151,7 @@ func (m *postgresAssetManager) removeMissingCharts(charts []models.Chart) error 
 		chartIDs = append(chartIDs, fmt.Sprintf("'%s'", chart.ID))
 	}
 	chartIDsString := strings.Join(chartIDs, ", ")
-	rows, err := m.DB.Query(fmt.Sprintf("DELETE FROM %s WHERE info ->> 'ID' NOT IN (%s)", dbutils.ChartFilesTable, chartIDsString))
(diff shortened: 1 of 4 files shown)


cmd/asset-syncer/postgresql_utils.go
```diff
@@ -134,15 +152,15 @@ func (m *postgresAssetManager) removeMissingCharts(charts []models.Chart) error
 		chartIDs = append(chartIDs, fmt.Sprintf("'%s'", chart.ID))
 	}
 	chartIDsString := strings.Join(chartIDs, ", ")
-	rows, err := m.DB.Query(fmt.Sprintf("DELETE FROM %s WHERE info ->> 'ID' NOT IN (%s)", chartTable, chartIDsString))
+	rows, err := m.DB.Query(fmt.Sprintf("DELETE FROM %s WHERE info ->> 'ID' NOT IN (%s) AND info -> 'repo' ->> 'name' = $1", dbutils.ChartTable, chartIDsString), charts[0].Repo.Name)
```
andresmgot: Had to add the `AND ...` clause because it was deleting charts from other repositories :sweat_smile:

absoludity: Heh - good to catch that one early :P

PR #1399, opened 2019-12-18
Add Helm 3 CreateRelease

latiif:
### Description of the change

This PR implements `CreateRelease` using the Helm 3 API.
### Benefits

- Charts can be installed from the added Helm repos in Kubeapps.
- Validates `values.yaml` against schema files if present.

### Possible drawbacks
- Creating a release works, and it can be listed, but an error message is displayed because `GetRelease` does not exist yet.
- There are no tests yet (Help appreciated).

### Applicable issues

 Based on discussions in #1309 (@lindhe)
 Related PR for `ListReleases` in #1359  (@SimonAlling)

### Additional information
Don't forget to set `useHelm3` to `true` in `values.yaml` to see the changes.

### DIFF
--- a/cmd/kubeops/internal/handler/handler.go
+++ b/cmd/kubeops/internal/handler/handler.go
@@ -4,9 +4,13 @@ import (
 	"net/http"
 
 	"github.com/kubeapps/common/response"
+	appRepo "github.com/kubeapps/kubeapps/cmd/apprepository-controller/pkg/client/clientset/versioned"
 	"github.com/kubeapps/kubeapps/pkg/agent"
 	"github.com/kubeapps/kubeapps/pkg/auth"
+	chartUtils "github.com/kubeapps/kubeapps/pkg/chart"
 	"github.com/kubeapps/kubeapps/pkg/handlerutil"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


cmd/kubeops/internal/handler/handler.go
```diff
@@ -55,3 +88,22 @@ func ListReleases(cfg agent.Config, w http.ResponseWriter, req *http.Request, pa
...
 }
+
+func CreateRelease(cfg agent.Config, w http.ResponseWriter, req *http.Request, params handlerutil.Params) {
+	requireV1Support := false
+	chartDetails, chartMulti, err := handlerutil.ParseAndGetChart(req, cfg.ChartClient, requireV1Support)
+	if err != nil {
+		response.NewErrorResponse(handlerutil.ErrorCode(err), err.Error()).Write(w)
+		return
+	}
+	ch := chartMulti.Helm3Chart
```
lindhe: I get that `requireV1Support` is set to false, but I guess that does not imply that this _is_ in-fact a Helm 3 chart. But here, it seems like we assume that it is a Helm3Chart. It's probably just me not understanding how this works, but I wanted to highlight it in case you've missed a corner-case here.

absoludity: Yeah, it's not an assumption of a helm 3 chart, so much as that we're using the helm3 struct representation of the chart here (which supports both helm3 and helm2 charts), rather than the helm2 struct representation (which supports helm 2 charts only). Perhaps I should have named it `Helm3LoadedChart`... naming is hard :/

pkg/agent/agent.go
```diff
@@ -68,23 +74,65 @@ func ListReleases(actionConfig *action.Configuration, namespace string, listLimi
...
 	actionConfig.Log = klog.Infof
 	return actionConfig, nil
 }
 
+// NewConfigFlagsFromCluster returns ConfigFlags with default values set from within cluster
+func NewConfigFlagsFromCluster(namespace string, clusterConfig *rest.Config) *genericclioptions.ConfigFlags {
+	impersonateGroup := []string{}
+	insecure := false
+
+	// CertFile and KeyFile must be nil for the BearerToken to be used for authentication and authorization instead of the pod's service account.
```
absoludity: Thanks, my first task today was going to be looking for the cause of the service account still being used in a similar PR I had. I'll try the equivalent change but using the `clientcmd` module, which is I think an easier/safer way forward.

cmd/kubeops/internal/handler/handler.go
SimonAlling: I think the name is good. The semantics of the parameter in question is such that when it's `true`, `ParseAndGetChart` should fail if it cannot load a v1 (Helm 2) chart, whereas when it's `false`, it may return only a v2 (Helm 3) chart and `nil` for the v1 chart.

pkg/agent/agent.go
latiif: Great! Let's know how it goes.

PR #1400, opened 2019-12-18
Allow to deploy Kubeapps using a PostgreSQL database

andresmgot:
### Description of the change

Allow to deploy Kubeapps using a PostgreSQL database. It includes the secret creation and the cleanup when deleting the chart. Note that changing from MongoDB to PostgreSQL (or viceversa) while upgrading is not supported (because the secret with the credentials is generated only when installing).

Tested manually (it requires some additional changes in the `apprepository-controller`).

Note: This is not deploying PostgreSQL in HA, that change will come later.

### Applicable issues

  - ref #651


chart/kubeapps/values.schema.json
```diff
@@ -142,37 +142,6 @@
           }
         }
       }
-    },
-    "mongodb": {
```
andresmgot: I am removing this from the schema since the persistence is disabled by default (and it's not really recommended to turn it on) so it didn't make sense to add the clutter of mongo vs postgres.

chart/kubeapps/templates/_helpers.tpl
```diff
@@ -102,10 +112,10 @@ Create name for the apprepository cleanup job
 {{- end -}}
 
 {{/*
-Create name for the mongodb secret bootstrap job
+Create name for the db-secret secret bootstrap job
 */}}
-{{- define "kubeapps.mongodb-jobs-cleanup.fullname" -}}
-{{ template "kubeapps.fullname" . }}-internal-mongodb-jobs-cleanup
+{{- define "kubeapps.db-secret-jobs-cleanup.fullname" -}}
+{{ template "kubeapps.fullname" . }}-internal-db-secret-jobs-cleanup
```
absoludity: Will this break upgrades of existing Kubeapps? (if the k8s secret name changes, I mean).

andresmgot: In theory it should not (it will only break if someone tries to upgrade from mongo to postgres) but I will double check that before merging/releasing this feature.

PR #1403, opened 2019-12-20
Update apprepo create API to include appropriate secret

absoludity:
### Description of the change
Follows on from #1397

1. Fixes issue where clientset was still using service-account auth
2. Updates to also create and reference relevant secret for an app repo with credentials, as well as including other app repo details.

Next PR will update the UI to use this endpoint.

### Benefits

Also demonstrates testing of handlers using k8s clientsets (@SimonAlling - in case it's useful/relevant).

### Applicable issues

  - Ref #1173

### DIFF
--- a/cmd/tiller-proxy/internal/handler/apprepos_handler.go
+++ b/cmd/tiller-proxy/internal/handler/apprepos_handler.go
@@ -18,20 +18,37 @@ package handler
 
 import (
 	"encoding/json"
-	"io"
+	"fmt"
 	"net/http"
 
 	log "github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/kubernetes"
+	corev1typed "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 
 	"github.com/kubeapps/kubeapps/cmd/apprepository-controller/pkg/apis/apprepository/v1alpha1"
-	clientset "github.com/kubeapps/kubeapps/cmd/apprepository-controller/pkg/client/clientset/versioned"
+	apprepoclientset "github.com/kubeapps/kubeapps/cmd/apprepository-controller/pkg/client/clientset/versioned"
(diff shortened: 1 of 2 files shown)


cmd/tiller-proxy/internal/handler/apprepos_handler.go
```diff
@@ -143,23 +173,98 @@ func (a *appRepositoriesHandler) Create(w http.ResponseWriter, req *http.Request
 		return
 	}
 
+	secret := secretForRequest(appRepoRequest, *appRepo)
```
andresmgot: nit: why using the content of `appRepo` instead of the pointer?

cmd/tiller-proxy/internal/handler/apprepos_handler.go
```diff
@@ -143,23 +173,98 @@ func (a *appRepositoriesHandler) Create(w http.ResponseWriter, req *http.Request
...
+		secrets["ca.crt"] = appRepoDetails.CustomCA
+	}
+
+	if len(secrets) == 0 {
+		return nil
+	}
+	blockOwnerDeletion := true
+	return &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: fmt.Sprintf("apprepo-%s-secrets", appRepoDetails.Name),
```
andresmgot: this secret name is generated twice, better to do it in a function so we avoid miss configurations in the future?

cmd/tiller-proxy/internal/handler/apprepos_handler.go
absoludity: No reason, updated.

cmd/tiller-proxy/internal/handler/apprepos_handler.go
absoludity: Also updated. Thanks.

absoludity:
> LGTM (just a couple of minor comments).
> 
> Apart from that, we need to update the tiller-proxy RBAC setting in the chart to be able to create secrets and apprepositories

I don't think so? That is, this handler explicitly uses the users token (from the request). The user needs to have the required RBAC to create app repositories and secrets... it'd be dangerous if we were proxying that. Let me know if I've missed something?

andresmgot:
> I don't think so? That is, this handler explicitly uses the users token (from the request). The user needs to have the required RBAC to create app repositories and secrets... it'd be dangerous if we were proxying that. Let me know if I've missed something?

Right, I forgot that for a moment :upside_down_face:

PR #1404, opened 2019-12-20
Code of conduct

ervin210:
code

### Applicable issues

  - fixes #


andresmgot:
Agreed. It has not been discussed before but I think it's a good idea, if we change the email to `containers@bitnami.com` (which is the maintainer's email that we set in several places) we can merge this PR.

PR #1405, opened 2019-12-20
Add readiness and liveness checks for Kubeops

SimonAlling:
### Description of the change

It does basically the same thing as #1348 but for Kubeops.

### Benefits

* Requests to Kubeops won't fail with `502 Bad Gateway`.

### Applicable issues

This PR is intended to be merged before #1406.

PR #1406, opened 2019-12-20
Add Helm 3 GetRelease

SimonAlling:
### Description of the change

This PR implements `GetRelease` using the Helm 3 API.

The Dashboard doesn't understand the release format used by Helm 3, so we implemented a compatibility layer that translates Helm 3 releases to a Helm 2-similar format suitable for the Dashboard in its current form.

### Benefits

* By and large, it works.
* The Dashboard need not be modified.

### Possible drawbacks

* The original `values.yaml` is _not_ included in the response to the Dashboard, just a generated one without comments and formatting. The same goes for the actual config (`values.yaml` with release-specific modifications). We didn't know where to find the original files.
* There are no tests yet (but there soon will be, thanks to @latiif).

### Applicable issues

* Based on discussions in #1309.
* The logical next step after #1399.
* Kubeops doesn't work at all until #1405 has been merged.

### Additional information

Don't forget to set `useHelm3` to `true` in `values.yaml` to see the changes.

Created in collaboration with @lindhe and @latiif.


cmd/kubeops/internal/handler/dashboardcompat.go
```diff
@@ -0,0 +1,100 @@
...
+		Namespace: h3r.Namespace,
+	}
+}
+
+func compatibleChart(h3c h3chart.Chart) dashboardCompatibleChart {
+	return dashboardCompatibleChart{
+		Files:     compatibleFiles(h3c.Files),
+		Metadata:  *h3c.Metadata,
+		Templates: compatibleTemplates(h3c.Templates),
+		Values:    compatibleConfig(h3c),
```
lindhe: According to comments in the Helm code, "Config is the set of extra Values added to the chart. These values override the default values inside of the chart.". I guess that means that we should not use `compatibleConfig` to set the values. `Config` be the subset of Values that we saw somewhere (you know that yaml that was stripped of comments...), but we should also have a `compatibleValues()` that returns the _actual_ full Values contents.

https://github.com/helm/helm/blob/master/pkg/release/release.go#L29-L31

lindhe:
Could this be the function to use for getting the Values from a Release?

https://github.com/helm/helm/blob/master/pkg/chartutil/values.go#L137

cmd/kubeops/internal/handler/dashboardcompat.go
SimonAlling: You're probably right.

SimonAlling: It seems like `config` is the final, modified `values`. I deployed `bitnami/apache` with the `replicaCount` and `metric.resources.limits` properties manually modified in the GUI, and then compared `values.raw` with `config.raw` (both from the JSON response):

```diff
$ diff json-values.yaml json-config.yaml 
27d26
< 
51d49
< 
55d52
< 
58c55
< replicaCount: 1
---
> replicaCount: 5
86d82
< 
90d85
< 
94d88
< 
98d91
< 
164d156
< 
170d161
< 
179c170
<   secrets:
---
>   secrets: 
225,227c216,218
<     limits: {}
<     #   cpu: 100m
<     #   memory: 128Mi
---
>     limits:
>         cpu: 100m
>         memory: 128Mi
```

Note that both `values.raw` and `config.raw` preserved all YAML comments.

SimonAlling:
We've now implemented `chart.values` and `config` in the JSON response as best we can. We don't know where to find the original `values.yaml` (with comments etc), so we generate it from the internal `map[string]interface{}` representation instead.

go.mod
```diff
@@ -23,6 +23,7 @@ require (
 	github.com/urfave/negroni v1.0.0
 	google.golang.org/grpc v1.25.1
 	gopkg.in/DATA-DOG/go-sqlmock.v1 v1.3.0 // indirect
+	gopkg.in/yaml.v2 v2.2.4
```
latiif: I am curious as why you didn't choose to use `sigs.k8s.io/yaml` instead?

SimonAlling: Well … I just typed away and the import was added automatically when I saved the file. ^^

latiif: I'm not one to judge :)

SimonAlling:
I've noticed that, when installing a chart, its `Raw` property contains the expected files, including `values.yaml`; however, when _getting_ the same chart, its `Raw` property is the empty list – at least as far as I can tell.

cmd/kubeops/internal/handler/dashboardcompat.go
```diff
@@ -0,0 +1,121 @@
...
+	h3chart "helm.sh/helm/v3/pkg/chart"
+	h3 "helm.sh/helm/v3/pkg/release"
+	h2chart "k8s.io/helm/pkg/proto/hapi/chart"
+	h2 "k8s.io/helm/pkg/proto/hapi/release"
+)
+
+// generatedYamlHeader is prepended to YAML generated from the internal map[string]interface{} representation.
+const generatedYamlHeader = "# Not original YAML! Generated from parsed representation."
+
+type dashboardCompatibleRelease struct {
```
absoludity: Is there an issue using the `h2.Release` instead of creating a new struct here? It appears to have the same fields (extra ones serialize with `omitempty`):

https://github.com/kubeapps/kubeapps/blob/master/pkg/agent/agent_test.go#L185

cmd/kubeops/internal/handler/dashboardcompat.go
```diff
@@ -0,0 +1,121 @@
...
+	Name      string                    `json:"name,omitempty"`
+	Info      h2.Info                   `json:"info,omitempty"`
+	Chart     dashboardCompatibleChart  `json:"chart,omitempty"`
+	Config    dashboardCompatibleConfig `json:"config,omitempty"`
+	Manifest  string                    `json:"manifest,omitempty"`
+	Version   int                       `json:"version,omitempty"`
+	Namespace string                    `json:"namespace,omitempty"`
+}
+
+type dashboardCompatibleChart struct {
```
absoludity: Same here with https://github.com/helm/helm/blob/release-2.16/pkg/proto/hapi/chart/chart.pb.go#L24 (although I see you're using `h3chart.Metadata` below - are they the same?)

cmd/kubeops/internal/handler/dashboardcompat.go
```diff
@@ -0,0 +1,121 @@
...
+	Files     []*any.Any                `json:"files,omitempty"`
+	Metadata  h3chart.Metadata          `json:"metadata,omitempty"`
+	Templates []*h2chart.Template       `json:"templates,omitempty"`
+	Values    dashboardCompatibleValues `json:"values,omitempty"`
+}
+
+type dashboardCompatibleValues struct {
```
absoludity: ...and here https://github.com/helm/helm/blob/release-2.16/pkg/proto/hapi/chart/config.pb.go#L22

(in the helm2 code, the `Values` and `Config` appear to be the same thing, which looks reflected in you having the same type).

cmd/kubeops/internal/handler/dashboardcompat.go
```diff
@@ -0,0 +1,121 @@
...
+func compatibleConfig(h3r h3.Release) dashboardCompatibleConfig {
+	return dashboardCompatibleConfig{
+		Raw: valuesToYaml(h3r.Config),
+	}
+}
+
+// valuesToYaml serializes to YAML and prepends an informative header.
+// It assumes that the serialization succeeds.
+func valuesToYaml(values map[string]interface{}) string {
+	marshaled, _ := yaml.Marshal(values)
```
absoludity: Why do you want to ignore the error rather than handling it here? This could result in hard-to-discover bugs later (ie. getting an empty values without any known reason why, I think?)

Actually, this looks like another place where we could just use the helm2 pkg? See

https://github.com/helm/helm/blob/bbdfe5e7803a12bbdf97e94cd847859890cf4050/pkg/tiller/release_server.go#L157

where `Raw` is set using https://github.com/helm/helm/blob/release-2.16/pkg/chartutil/values.go#L45

cmd/kubeops/internal/handler/dashboardcompat.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/golang/protobuf/ptypes/any"
+	"gopkg.in/yaml.v2"
+	h3chart "helm.sh/helm/v3/pkg/chart"
+	h3 "helm.sh/helm/v3/pkg/release"
+	h2chart "k8s.io/helm/pkg/proto/hapi/chart"
+	h2 "k8s.io/helm/pkg/proto/hapi/release"
+)
+
+// generatedYamlHeader is prepended to YAML generated from the internal map[string]interface{} representation.
+const generatedYamlHeader = "# Not original YAML! Generated from parsed representation."
```
andresmgot: I would skip this header, then in the dashboard we can try to recover the comments

cmd/kubeops/internal/handler/dashboardcompat.go
SimonAlling: We originally wanted to write a function from `h3.Release` to `h2.Release`, but arrived at the conclusion that it's impossible in general. Can't remember exactly why, but I think there were several reasons. One might be that the respective `Metadata` types are not isomorphic. The `Values` property in the respective `Chart` types also seems problematic.

I've, however, made a best-effort attempt at using Helm 2 types instead of custom ones. It seems to work; see d67e136.

cmd/kubeops/internal/handler/dashboardcompat.go
SimonAlling: No, `h2chart.Metadata` and `h3chart.Metadata` are different (and not isomorphic).

cmd/kubeops/internal/handler/dashboardcompat.go
SimonAlling: AFAIK, `Values` and `Config` are not the same thing: `Values` is the original `values.yaml` for the chart, whereas `Config` is the _effective_ set of values used for the _release_ – usually `Values` overridden with any custom values set by the user at install time.

cmd/kubeops/internal/handler/dashboardcompat.go
SimonAlling: I thought there would never occur any error there, and abandoning that assumption would contaminate the entire file with `error` and `nil` everywhere without adding any real value (or so I thought).

There _should_ exist a total function from `h3.Release` to some Dashboard-compatible type, shouldn't there? (`newDashboardCompatibleRelease` is my attempt at constructing such a function.) Maybe @andresmgot's suggestion of storing the YAML separately is the best solution; then we wouldn't need to marshal anything at all.

Finally, I don't see how [`YAML`](https://github.com/helm/helm/blob/release-2.16/pkg/chartutil/values.go#L45) could be used here. It works only on members of `Values` from Helm 2, but we have a `map[string]interface{}`.

cmd/kubeops/internal/handler/dashboardcompat.go
SimonAlling: I think the header should be there for clarity until we can do something better, such as storing the YAML separately, as you suggested.

andresmgot: Yes, the problem is that it's giving to the user internal information about a problem they may not be aware of. It seems like an error. It's easier to understand that the parsed YAML is what it's stored in the cluster (with no comments).

SimonAlling: I see!

cmd/kubeops/internal/handler/dashboardcompat.go
andresmgot: In the unlikely case that the `map[string]interface{}` couldn't be marshaled, returning an error here would be helpful to identify the issue. I would either log the error (if you want this function to succeed in any case, even if the values cannot be marshalled) or return it and handle it in an upper level.

cmd/kubeops/internal/handler/dashboardcompat.go
absoludity: Right, here we're just looking at getting data back to the dashboard for presentation, rather than interacting with helm, so I don't *think* isomorphism is essential as it would be if we were generating this for use with helm itself. Thanks for the update!

cmd/kubeops/internal/handler/dashboardcompat.go
absoludity: Sorry - yes, I meant that they appear to use the same type in helm2 rather than requiring two different but identical types - so `Chart.Values` (at https://github.com/helm/helm/blob/release-2.16/pkg/proto/hapi/chart/chart.pb.go#L32) is actually of type `*Config` (whereas here we had two identical types with `dashboardCompatibleValues` and `dashboardCompatibleConfig` I think?)

cmd/kubeops/internal/handler/dashboardcompat.go
SimonAlling: I'd suggest one of these options for now, to keep the code simple:
* Panic if serialization failed.
* Ignore any serialization error (current solution).

The rationale would be the same no matter which one we choose: This won't hit production anyway (because we need to figure out how to store/access `values.yaml`), and it's good enough™ for us to be able to keep working on Helm 3 support.

andresmgot: why don't you log the error instead? A `panic` can be disruptive in the whole process and you don't need to return the error and obscure the code if you just log the error.

SimonAlling: Sounds reasonable. I'll log it instead.

PR #1408, opened 2020-01-02
[kubeops] Implement framework for testing /kubeops/internal/handler

latiif:
### Description of the change
Adds test framework for `/kubeops/internal/handler` to facilitate adding test scenarios in the future for all kinds of actions, such as creating a release, listing all releases, deleting etc.
Changes proposed are analogous to the file [`/tiller-proxy/internal/handler/handler_test.go`](https://github.com/kubeapps/kubeapps/blob/master/cmd/tiller-proxy/internal/handler/handler_test.go).

This PR adds only the skeleton without any test scenarios implemented as they are function-specific and are up to the author of future tests.

### Benefits
- Lays ground for future PR's for testing new Helm 3 functions.
- Uses Helm 3 library for representing releases, charts etc.
- Adds Helm 3 support in `GetChart` in `/pkg/chart/fake`

### Applicable issues

Planned to be used to write tests for #1399 and #1406 and #1359 

### Additional information
This is a minimal PR and future tests for `kubeops/internal/handler` will be based on it.

Signed-off-by: Latiif <abdullatif.alshriaf@combitech.se>

### DIFF
--- a/cmd/kubeops/internal/handler/handler_test.go
+++ b/cmd/kubeops/internal/handler/handler_test.go
@@ -1,0 +1,135 @@
+package handler
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"net/http/httptest"
+	"reflect"
+	"strings"
+	"testing"
+
+	"github.com/kubeapps/kubeapps/pkg/agent"
+	"github.com/kubeapps/kubeapps/pkg/auth"
+	authFake "github.com/kubeapps/kubeapps/pkg/auth/fake"
+	chartFake "github.com/kubeapps/kubeapps/pkg/chart/fake"
+	"helm.sh/helm/v3/pkg/action"
+	"helm.sh/helm/v3/pkg/chart"
+	"helm.sh/helm/v3/pkg/chartutil"
+	kubefake "helm.sh/helm/v3/pkg/kube/fake"
+	"helm.sh/helm/v3/pkg/storage"
+	"helm.sh/helm/v3/pkg/storage/driver"
+
+	"helm.sh/helm/v3/pkg/release"
+)
+
+const defaultListLimit = 256
+
+// newConfigFixture returns an agent.Config with fake clients
+// and memory storage.
+func newConfigFixture(t *testing.T) *agent.Config {
+	t.Helper()
+
+	return &agent.Config{
+		ActionConfig: &action.Configuration{
+			Releases:     storage.Init(driver.NewMemory()),
+			KubeClient:   &kubefake.FailingKubeClient{PrintingKubeClient: kubefake.PrintingKubeClient{Out: ioutil.Discard}},
+			Capabilities: chartutil.DefaultCapabilities,
+			Log: func(format string, v ...interface{}) {
+				t.Helper()
+				t.Logf(format, v...)
+			},
+		},
+		ChartClient: &chartFake.FakeChart{},
+		AgentOptions: agent.Options{
+			ListLimit: defaultListLimit,
+		},
+	}
+}
+
+func TestActions(t *testing.T) {
+	type testScenario struct {
+		// Scenario params
+		Description      string
+		ExistingReleases []release.Release
+		DisableAuth      bool
+		// Request params
+		RequestBody  string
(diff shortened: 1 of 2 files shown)


cmd/kubeops/internal/handler/handler_test.go
```diff
@@ -0,0 +1,135 @@
...
+				t.Errorf("Unexpected body response. Expecting %s, found %s", test.ResponseBody, response.Body)
+			}
+		}
+	}
+}
+
+func extractReleases(storage *storage.Storage) []release.Release {
+	rls, _ := storage.ListReleases()
+	releases := make([]release.Release, len(rls))
+	//cleanup unused properties
```
absoludity: I'm pretty wary of this function claiming to do one thing (extract releases from the storage), while actually doing something else (extract releases from the storage then blank/change certain values).

The `google/go-cmp` has helpers to be able to ignore certain paths within a struct if the comparison test should not compare them, which would be much more explicit

cmd/kubeops/internal/handler/handler_test.go
```diff
@@ -0,0 +1,135 @@
...
+		switch test.Action {
+		default:
+			t.Errorf("Unexpected action %s", test.Action)
+		}
+		// Check result
+		if response.Code != test.StatusCode {
+			t.Errorf("Expecting a StatusCode %d, received %d", test.StatusCode, response.Code)
+		}
+		releases := extractReleases(cfg.ActionConfig.Releases)
+		if !reflect.DeepEqual(releases, test.RemainingReleases) {
```
absoludity: If you instead use the `google/go-cmp` module instead of `DeepEqual`, you'll not only get more useful test output (ie. `cmp.Diff()` pretty-prints the diff between two structs etc.), but won't need to blank certain fields in `extractReleases` (see below).

pkg/chart/fake/chart.go
```diff
@@ -33,18 +35,40 @@ func (f *FakeChart) ParseDetails(data []byte) (*chartUtils.Details, error) {
...
+			Metadata: &chart3.Metadata{
+				Name: details.ChartName,
+			},
+			Values: vals,
+		},
 	}, nil
 }
 
+// Values is a type alias for values.yaml
+type Values map[string]interface{}
```
absoludity: Not sure that there's any benefit hiding the `map[string]interface{}` in a type alias?

cmd/kubeops/internal/handler/handler_test.go
```diff
@@ -0,0 +1,135 @@
...
+		response := httptest.NewRecorder()
+		cfg := newConfigFixture(t)
+		for i := range test.ExistingReleases {
+			err := cfg.ActionConfig.Releases.Create(&test.ExistingReleases[i])
+			if err != nil {
+				t.Errorf("Failed to initiate test: %v", err)
+			}
+		}
+		// Perform request
+		t.Log(test.Description)
```
absoludity: Hmm, I'm not sure that the old tests in the `tiller-proxy/internal/handler/handler_test.go` are a good model to follow... as I see it does the same thing. Rather than `Log`ing the description of the test, the test should be run as a sub-test:
```
t.Run(test.Description, func(t *testing.T){
...
})
```

Take a look at the more recent tests that were added to the agent pkg at: https://github.com/kubeapps/kubeapps/blob/master/pkg/agent/agent_test.go#L185

absoludity:
Forgot to mention, I've also added tests for a different handler in this PR in case it's useful:
https://github.com/kubeapps/kubeapps/pull/1403/files

cmd/kubeops/internal/handler/handler_test.go
latiif: I'll look into `go-cmp` helpers.

latiif:
I opted for using a custom comparator instead of ignoring unwanted fields, which is more explicit and extensible.

pkg/chart/fake/chart.go
latiif: I thought it would make it easier to read function signatures.
_removed in latest commit_

PR #1413, opened 2020-01-07
Push PROD tag only in master or in a tagged commit

andresmgot:
### Description of the change

Avoid issues pushing the image tag `:latest`  when building a branch.

Context: https://kubernetes.slack.com/archives/C9D3TSUG4/p1576769697008300


andresmgot:
This is actually not needed, it's just necessary to change the build filter. Added in #1323

PR #1416, opened 2020-01-08
Switch dashboard to use backend API for creating app repos.

### LINKED ISSUE #1173: Enable third party app to trigger chart deployment UI

absoludity:
### Description of the change

Switches the dashboard from using the k8s API directly to create an app repository and associated secrets, to instead use the backend API to create the app repo (which creates the associated secrets).

### Benefits

This ensures that we dogfood the new API ourselves :)

### Possible drawbacks

I've tested this locally with an auth header and ca.crt to verify the correct secret is created, as well as with a sync pod template, but there could be something I've missed (I've not, for example, verified access using those with an actual private repo).

### Applicable issues

  - fixes #1173

### DIFF
--- a/cmd/tiller-proxy/internal/handler/apprepos_handler.go
+++ b/cmd/tiller-proxy/internal/handler/apprepos_handler.go
@@ -76,7 +76,7 @@ type appRepositoryRequest struct {
 
 type appRepositoryRequestDetails struct {
 	Name               string                 `json:"name"`
-	RepoURL            string                 `json:"repoUrl"`
+	RepoURL            string                 `json:"repoURL"`
 	AuthHeader         string                 `json:"authHeader"`
 	CustomCA           string                 `json:"customCA"`
 	SyncJobPodTemplate corev1.PodTemplateSpec `json:"syncJobPodTemplate"`
--- a/dashboard/src/actions/repos.test.tsx
+++ b/dashboard/src/actions/repos.test.tsx
@@ -195,24 +195,20 @@ describe("installRepo", () => {
       "",
     );
 
-    const authStruct = {
-      header: { secretKeyRef: { key: "authorizationHeader", name: "apprepo-my-repo-secrets" } },
-    };
-
     it("calls AppRepository create including a auth struct", async () => {
       await store.dispatch(installRepoCMDAuth);
       expect(AppRepository.create).toHaveBeenCalledWith(
         "my-repo",
-        "my-namespace",
         "http://foo.bar",
-        authStruct,
-        {},
+        "Bearer: abc",
+        "",
+        "",
       );
     });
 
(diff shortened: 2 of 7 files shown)


cmd/tiller-proxy/internal/handler/apprepos_handler.go
```diff
@@ -76,7 +76,7 @@ type appRepositoryRequest struct {
 
 type appRepositoryRequestDetails struct {
 	Name               string                 `json:"name"`
-	RepoURL            string                 `json:"repoUrl"`
+	RepoURL            string                 `json:"repoURL"`
```
absoludity: This issue was found during real-life testing: used wrong capitalization on the json tag which meant the `repoURL` sent from the dashboard was being ignored. Now it's consistent with `customCA` for example.

dashboard/src/actions/repos.test.tsx
```diff
@@ -271,41 +263,19 @@ spec:
...
-          {},
-          { spec: { containers: [{ env: [{ name: "FOO", value: "BAR" }] }] } },
-        );
-      });
-
-      // Example from https://nealpoole.com/blog/2013/06/code-execution-via-yaml-in-js-yaml-nodejs-module/
-      const unsafeYAMLTemplate =
-        '"toString": !<tag:yaml.org,2002:js/function> "function (){very_evil_thing();}"';
-
-      it("does not call AppRepository create with an unsafe pod template", async () => {
```
absoludity: The frontend no longer parses the yaml. It's now text sent straight to the backend API.

dashboard/src/actions/repos.ts
```diff
@@ -132,68 +130,16 @@ export const installRepo = (
...
-              name: secretName,
-            },
-          };
-          secrets["ca.crt"] = btoa(customCA);
-        }
-      }
-      let syncJobPodTemplateObj = {};
-      if (syncJobPodTemplate.length) {
-        syncJobPodTemplateObj = yaml.safeLoad(syncJobPodTemplate);
-      }
```
absoludity: Frontend no longer needs complication of creating associated secrets etc., since it's done as part of the backend API call implementation.

absoludity:
Sorry, turns out I was testing the wrong images after recreating my dev cluster to work with the new assetsvc etc. I'll re-propose tomorrow with some more changes.

PR #1418, opened 2020-01-08
Add database choice to chart README

andresmgot:
### Description of the change

Add docs of how to use PostgreSQL or MongoDB in the chart.

### DIFF
--- a/chart/kubeapps/README.md
+++ b/chart/kubeapps/README.md
@@ -75,6 +75,12 @@ $ helm install --name kubeapps --namespace kubeapps -f custom-values.yaml bitnam
 
 By default, Kubeapps will track the [community Helm charts](https://github.com/helm/charts) and the [Kubernetes Service Catalog charts](https://github.com/kubernetes-incubator/service-catalog). To change these defaults, override with your desired parameters the `apprepository.initialRepos` object present in the [values.yaml](values.yaml) file.
 
+### Configuring the database to use
+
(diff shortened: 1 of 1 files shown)


andresmgot:
> If the issue is that this secret is only created when (before) the chart is first installed, so that we don't overwrite the randomly-generated, the only solution I can see in a job that in installed with `helm.sh/hook: pre-upgrade` which manually creates the `kubeapps-postgresql` secret iff it doesn't exist. But this would be run on every upgrade until we no longer support upgrading from mongodb and remove it. Thoughts?

Yes, that's what I mentioned. The problem with that is:

 - We would need to maintain that hook. More files to the chart but we have done that before.
 - When upgrading, if the apprepository-controller is not changed, the new database will be created but the sync jobs won't be spin up. Until the next cycle (<10 min) there would be no charts in the app.

Those are no big issues, so we could do that but it's also true that I don't think people will want to modify the database they are using on the fly. They would either want to stick to MongoDB if they are already using that or create a new instance with maybe PostgreSQL.

A different thing would be if we remove MongoDB, then yes, we would need to add that hook to make the upgrade as smooth as possible but I don't think we want to do that (yet).

My take is to not implement that (for the moment) until we have a specific request or until we remove MongoDB but I am open to suggestions. WDYT?

absoludity:
> My take is to not implement that (for the moment) until we have a specific request or until we remove MongoDB but I am open to suggestions. WDYT?

Yeah, I'm OK with that. It'd just be nice if we didn't need a re-install to switch to postgres. I don't think we'll get a request after we've already forced people to re-install, but maybe that's not a big issue. Right, as many users won't need to change anyway.

PR #1419, opened 2020-01-08
Enable replication for PostgreSQL

### LINKED ISSUE #1142: MongoDB is a single point of failure in default deploy
The current default deployment for the Kubeapps chart [explicitly avoids using persistence for MongoDB](https://github.com/kubeapps/kubeapps/blob/master/chart/kubeapps/values.yaml#L257) (see #336 for reasoning) and uses the default MongoDB chart setting for [`replicaSet: { enabled: false}`](https://github.com/helm/charts/blob/master/stable/mongodb/values.yaml#L146).

This leaves MongoDB as a single point of failure which requires time to rebuild if the pod is rescheduled (as found in #1141).

The assumption to this point seems to be that this service is a cache which does not require persistence (see #336), which may be true - but it *is* required to be available. We should investigate whether enabling the default Kubeapps deploy to use a replicaSet for this cache (whether or not we stay with MongoDB or switch to something lighter - see #651), so that it is no longer a single pod and single point of failure that quite seriously affects the user experience when rescheduled (which can happen for a variety of reasons, so needs to be handled gracefully for the user experience). We can also re-evaluate #336 to see whether persistence would improve the situation (but imo using replicasets should make persistence less useful for this cache, unless all pods went down).

andresmgot:
### Description of the change

Deploy PostgreSQL (by default) with one replica. Thanks to that, if any of the two PostgreSQL pods goes down, the service will still work.

The service `kubeapps-postgresql` will reach the `master` instance only while the `kubeapps-postgresql-headless` service hits both the `master` and `slave` replicas.

### Possible drawbacks

The slave replica is read-only. This means that if the master goes down, for the time that that pod is down, the data is immutable.

### Applicable issues

  - fixes #1142

### DIFF
--- a/chart/kubeapps/templates/apprepository-deployment.yaml
+++ b/chart/kubeapps/templates/apprepository-deployment.yaml
@@ -57,7 +57,7 @@ spec:
             - --database-secret-name={{ .Values.postgresql.existingSecret }}
             - --database-secret-key=postgresql-password
             - --database-type=postgresql
-            - --database-url={{ template "kubeapps.postgresql.fullname" . }}:5432
+            - --database-url={{ template "kubeapps.postgresql.fullname" . }}-headless:5432
             - --database-user=postgres
             - --database-name=assets
             {{- end }}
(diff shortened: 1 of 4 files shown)


andresmgot:
> Does this not also require setting num replicas?

The number of replicas is 1 by default when enabling replication.

PR #1421, opened 2020-01-08
Add Helm 3 UpgradeRelease

SimonAlling:
### Description of the change

This PR implements `UpgradeRelease` using the Helm 3 API.

### Benefits

* It will bring us one step closer to Helm 3 support.

### Possible drawbacks

* There are no tests yet.
* `OperateRelease` only supports upgrading at the moment (but it will support testing and rolling back later).

### Applicable issues

* Based on discussions in #1309.
* The logical next step after #1420.

### Additional information

The `switch` statement in `handler.OperateRelease` is currently redundant, but it will make sense when support for testing and rolling back is added (see [its Tiller Proxy counterpart](https://github.com/kubeapps/kubeapps/blob/4b4b14bff89e552506aea7170ec51a4aa32719e5/cmd/tiller-proxy/internal/handler/handler.go#L97:L110)).

### DIFF
--- a/cmd/kubeops/internal/handler/handler.go
+++ b/cmd/kubeops/internal/handler/handler.go
@@ -19,6 +19,8 @@ const (
 	nameParam      = "releaseName"
 )
 
+const requireV1Support = false
+
 // This type represents the fact that a regular handler cannot actually be created until we have access to the request,
 // because a valid action config (and hence agent config) cannot be created until then.
 // If the agent config were a "this" argument instead of an explicit argument, it would be easy to create a handler with a "zero" config.
@@ -91,7 +93,6 @@ func ListAllReleases(cfg agent.Config, w http.ResponseWriter, req *http.Request,
 }
 
 func CreateRelease(cfg agent.Config, w http.ResponseWriter, req *http.Request, params handlerutil.Params) {
-	requireV1Support := false
 	chartDetails, chartMulti, err := handlerutil.ParseAndGetChart(req, cfg.ChartClient, requireV1Support)
 	if err != nil {
 		response.NewErrorResponse(handlerutil.ErrorCode(err), err.Error()).Write(w)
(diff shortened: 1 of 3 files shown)


cmd/kubeops/internal/handler/handler.go
```diff
@@ -109,6 +110,34 @@ func CreateRelease(cfg agent.Config, w http.ResponseWriter, req *http.Request, p
...
+		upgradeRelease(cfg, w, req, params)
+	default:
+		// By default, for maintaining compatibility, we call upgrade.
+		upgradeRelease(cfg, w, req, params)
+	}
+}
+
+func upgradeRelease(cfg agent.Config, w http.ResponseWriter, req *http.Request, params handlerutil.Params) {
+	releaseName := params[nameParam]
+	chartDetails, chartMulti, err := handlerutil.ParseAndGetChart(req, cfg.ChartClient, requireV1Support)
```
andresmgot: here `requireV1Support` is a bit of a readability issue. It seems that it's requiring v1 support. It would more readable if the var is called `requireV2Support` but it may be confusing if its value is `false`. Maybe `notRequireV1Support`?

SimonAlling: I was bugged by the exact same thought! But what do? Wouldn't it be just as confusing to write `const dontRequireV1Support = false`? Or would a comment suffice to mitigate that confusion?

andresmgot: yes, a comment when setting the variable would be enough in my opinion

SimonAlling: I made an

SimonAlling: I felt naming the variable `dontRequireV1Support` or similar would make the code WET, since it would tie the constant name very closely to its value. I've renamed it to `isV1SupportRequired` instead; WDYT?

PR #1422, opened 2020-01-09
Switch dashboard to use backend API for creating app repos

### LINKED ISSUE #1173: Enable third party app to trigger chart deployment UI

absoludity:
### Description of the change

Switches the dashboard from using the k8s API directly to create an app repository and associated secrets, to instead use the backend API to create the app repo (which creates the associated secrets).

### Benefits

This ensures that we dogfood the new API ourselves :) It also simplifies the frontend code, given that the backend API handles the related secret creation etc.

### Possible drawbacks

I've tested this locally with an auth header and ca.crt to verify the correct secret is created, as well as with a sync pod template, but there could be something I've missed (I've not, for example, verified access using those with an actual private repo).

It's not ideal that we're exposing the k8s AppRepository object in the response - I'd prefer not to expose it - but the current dashboard expects this. We can remove it when the other CRUD endpoints are added to the API.

### Applicable issues

  - fixes #1173

### DIFF
--- a/chart/kubeapps/Chart.yaml
+++ b/chart/kubeapps/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kubeapps
-version: 3.2.3
+version: 3.2.4
 appVersion: DEVEL
 description: Kubeapps is a dashboard for your Kubernetes cluster that makes it easy to deploy and manage applications in your cluster using Helm
 icon: https://raw.githubusercontent.com/kubeapps/kubeapps/master/docs/img/logo.png
--- a/chart/kubeapps/Chart.yaml
+++ b/chart/kubeapps/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kubeapps
-version: 3.2.4
+version: 3.2.3
 appVersion: DEVEL
 description: Kubeapps is a dashboard for your Kubernetes cluster that makes it easy to deploy and manage applications in your cluster using Helm
 icon: https://raw.githubusercontent.com/kubeapps/kubeapps/master/docs/img/logo.png
--- a/chart/kubeapps/templates/tiller-proxy-deployment.yaml
+++ b/chart/kubeapps/templates/tiller-proxy-deployment.yaml
@@ -47,7 +47,6 @@ spec:
           args:
             - --host={{ .Values.tillerProxy.host }}
             - --user-agent-comment=kubeapps/{{ .Chart.AppVersion }}
-            - --kubeapps-namespace={{ .Release.Namespace }}
             - --assetsvc-url=http://{{ template "kubeapps.assetsvc.fullname" . }}:{{ .Values.assetsvc.service.port }}
             {{- if .Values.tillerProxy.tls }}
             - --tls
--- a/chart/kubeapps/templates/tiller-proxy-deployment.yaml
+++ b/chart/kubeapps/templates/tiller-proxy-deployment.yaml
@@ -47,6 +47,7 @@ spec:
           args:
             - --host={{ .Values.tillerProxy.host }}
             - --user-agent-comment=kubeapps/{{ .Chart.AppVersion }}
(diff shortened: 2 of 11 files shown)


cmd/tiller-proxy/internal/handler/apprepos_handler.go
```diff
@@ -76,13 +76,18 @@ type appRepositoryRequest struct {
 
 type appRepositoryRequestDetails struct {
 	Name               string                 `json:"name"`
-	RepoURL            string                 `json:"repoUrl"`
+	RepoURL            string                 `json:"repoURL"`
```
absoludity: This issue was found during real-life testing: used wrong capitalization on the json tag which meant the `repoURL` sent from the dashboard was being ignored. Now it's consistent with `customCA` for example.

cmd/tiller-proxy/internal/handler/apprepos_handler.go
```diff
@@ -256,8 +270,8 @@ func secretForRequest(appRepoRequest appRepositoryRequest, appRepo *v1alpha1.App
 			Name: secretNameForRepo(appRepo.Name),
 			OwnerReferences: []metav1.OwnerReference{
 				metav1.OwnerReference{
-					APIVersion:         appRepo.TypeMeta.APIVersion,
-					Kind:               appRepo.TypeMeta.Kind,
+					APIVersion:         "kubeapps.com/v1alpha1",
+					Kind:               "AppRepository",
```
absoludity: IRL testing showed that `appRepo.TypeMeta` is empty. So given that it's constant, I've updated to the actual values.

chart/kubeapps/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kubeapps
-version: 3.2.3
```
andresmgot: we cannot bump the version yet (this would release the `assetsvc` and `asset-syncer`)

chart/kubeapps/templates/tiller-proxy-deployment.yaml
```diff
@@ -47,6 +47,7 @@ spec:
           args:
             - --host={{ .Values.tillerProxy.host }}
             - --user-agent-comment=kubeapps/{{ .Chart.AppVersion }}
+            - --kubeapps-namespace={{ .Release.Namespace }}
```
andresmgot: this is not necessary, the namespace is available in the env var `POD_NAMESPACE` (defined a bit below)

cmd/tiller-proxy/internal/handler/apprepos_handler.go
```diff
@@ -137,6 +142,7 @@ func (a *appRepositoriesHandler) Create(w http.ResponseWriter, req *http.Request
 	if a.kubeappsNamespace == "" {
```
andresmgot: just realized this too, you should be able to use the env var instead of storing the namespace in the struct.

dashboard/src/actions/repos.ts
```diff
@@ -128,72 +127,24 @@ export const installRepo = (
   repoURL: string,
   authHeader: string,
   customCA: string,
-  syncJobPodTemplate: string,
+  syncJobPodTemplate: any,
```
andresmgot: mm, why isn't this a string now? it seems that we are still parsing it as a yaml below

absoludity: Whoops, yes - I changed the type on the incorrect function, I'd meant to verify that `shared/apprepository:Create` accepted an object. Thanks.

PR #1429, opened 2020-01-10
Set unique chart ID column

andresmgot:
### Description of the change

Not setting a unique varchar column with PostgreSQL was causing the application to show duplicates instead of updating the existing chart. Now the chart ID (e.g. `bitnami/apache`) is used as that unique column.

### Possible drawbacks

It requires deleting the existing table (or restarting postgres pods) but the feature has not been released yet.

### DIFF
--- a/cmd/asset-syncer/postgresql_utils.go
+++ b/cmd/asset-syncer/postgresql_utils.go
@@ -26,9 +26,7 @@ import (
 	"github.com/kubeapps/common/datastore"
 	"github.com/kubeapps/kubeapps/pkg/chart/models"
 	"github.com/kubeapps/kubeapps/pkg/dbutils"
-	"github.com/lib/pq"
 	_ "github.com/lib/pq"
-	log "github.com/sirupsen/logrus"
 )
 
 type postgresAssetManager struct {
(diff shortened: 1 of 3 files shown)


cmd/asset-syncer/postgresql_utils.go
```diff
@@ -112,38 +110,22 @@ func (m *postgresAssetManager) UpdateLastCheck(repoName, checksum string, now ti
 }
 
 func (m *postgresAssetManager) importCharts(charts []models.Chart) error {
-	txn, err := m.DB.Begin()
```
absoludity: Why are you also removing the transaction for importing the charts?

andresmgot: I forgot to explain that. The only supported method when doing a transaction is `COPY FROM` (https://godoc.org/github.com/lib/pq#hdr-Bulk_imports). That was the reason for a chart to be added several times to the database. Once I set the unique column, that method started to fail because "there was already an entry with that unique ID".

I tried to use a custom query (instead of COPYing) but arguments were not supported. The only way I found to safely do this was to execute one query per chart (and it didn't have a noticeable impact in the performance).

PR #1430, opened 2020-01-10
Refactor Kubeops routes code

SimonAlling:
### Description of the change

It introduces an abstraction for defining routes, making the definitions clearer, shorter and DRYer.

### Benefits

  * Easier to define routes in `main`.
  * Reduced code repetition.
  * Less clutter, highlighting actual content.
  * Less room for mistakes.

### Possible drawbacks

  * Routes are forced to be very similar in behavior to each other.

### DIFF
--- a/cmd/kubeops/internal/handler/handler.go
+++ b/cmd/kubeops/internal/handler/handler.go
@@ -3,12 +3,14 @@ package handler
 import (
 	"net/http"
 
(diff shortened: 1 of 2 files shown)


SimonAlling:
> `AddRouteWith` is a bit more difficult to read

Yeah. Go's absence of (general) type inference sure doesn't help either.

PR #1433, opened 2020-01-10
[PG] Return latest chart version

andresmgot:
### Description of the change

It's not explicit but the `chartVersion` returned by `getChartsWithFilters` should be the latest version in order to return it to the client. That's is later used by the dashboard to determine if there are updates available. Since the charts are ordered, it should return the first one of the array.

### DIFF
--- a/cmd/assetsvc/postgresql_utils.go
+++ b/cmd/assetsvc/postgresql_utils.go
@@ -142,8 +142,6 @@ func (m *postgresAssetManager) getChartsWithFilters(name, version, appVersion st
 	result := []*models.Chart{}
 	for _, c := range charts {
 		if _, found := containsVersionAndAppVersion(c.ChartVersions, version, appVersion); found {
-			// Return only the latest ChartVersion
-			c.ChartVersions = []models.ChartVersion{c.ChartVersions[0]}
 			result = append(result, c)
 		}
 	}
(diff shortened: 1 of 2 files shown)


cmd/assetsvc/postgresql_utils.go
```diff
@@ -141,9 +141,9 @@ func (m *postgresAssetManager) getChartsWithFilters(name, version, appVersion st
 	}
 	result := []*models.Chart{}
 	for _, c := range charts {
-		if cv, found := containsVersionAndAppVersion(c.ChartVersions, version, appVersion); found {
-			// Return only the interesting ChartVersion
-			c.ChartVersions = []models.ChartVersion{cv}
+		if _, found := containsVersionAndAppVersion(c.ChartVersions, version, appVersion); found {
+			// Return only the latest ChartVersion
+			c.ChartVersions = []models.ChartVersion{c.ChartVersions[0]}
```
absoludity: This looks like the wrong place to change the data? The function is `getChartsWithFilters()`, which should return all charts which match the filter. If I were to see this function and re-use it for some other purpose, it'd be very surprising that it actually *modifies* the result (truncating the chart versions for the chart).

So my question is: why does this function need to truncate the chart versions for the returned charts at all?

The specific call-site in this case (the handler) can choose to overwrite `c.ChartVersions = []models.ChartVersion[c.ChartVersions[0]` if the handler should return just the one chart version, but I don't think it should be part of this util function?

andresmgot: I actually agree, my bad reason to do this is because in the mongodb-utils it's done this way. There, I believe the reason we are doing this is because of the performance. It's encoded in the MongoDB query so it's only returning one version which I believe is faster.

In any case, it's not the case here so I am okay changing the behavior.

PR #1434, opened 2020-01-10
[kubeops] Add tests for GetRelease

latiif:
### Description of the change
Tests new functionality introduced by #1406 . 

### Benefits

- Adds test for GetRelease in `/pkg/agent`
- Adds `action="get"` tests for `/kubeops/handler`
- Adds tests for `dashboardcompat.go` in `/kubeops/internal/handler`

### Possible drawbacks
Skip tests blocked by memory driver bug

### Applicable issues
Tests #1406 

### Additional information
Contains a `TODO` in `/cmd/kubeops/internal/handler/handler_test.go` with instructions on what to do when the [memory driver bug](https://github.com/helm/helm/pull/7372) is fixed.

Signed-off-by: Latiif <abdullatif.alshriaf@combitech.se>

### DIFF
--- a/cmd/kubeops/internal/handler/dashboardcompat_test.go
+++ b/cmd/kubeops/internal/handler/dashboardcompat_test.go
@@ -21,7 +21,6 @@ func TestNewDashboardCompatibleRelease(t *testing.T) {
 		Helm3Release        h3.Release
 		Helm2Release        h2.Release
 		MarshallingFunction func(h2.Release) string
-		ShouldFail          bool
 	}
 	tests := []testScenario{
 		{
@@ -110,68 +109,14 @@ func TestNewDashboardCompatibleRelease(t *testing.T) {
 				},
 			},
 		},
-		{
-			Description:         "Two equivalent releases with different versions",
-			ShouldFail:          true,
-			MarshallingFunction: asResponse,
-			Helm3Release: h3.Release{
-				Name:      "Foo",
-				Namespace: "default",
-				Chart: &h3chart.Chart{
-					Metadata: &h3chart.Metadata{},
-					Values: map[string]interface{}{
-						"port": 8080,
-					},
-				},
-				Info: &h3.Info{
-					Status: h3.StatusDeployed,
-				},
-				Version: 1,
-				Config: map[string]interface{}{
-					"port": 3000,
-					"user": map[string]interface{}{
-						"name":     "user1",
-						"password": "123456",
-					},
-				},
-			},
-			Helm2Release: h2.Release{
-				Name:      "Foo",
-				Namespace: "default",
-				Info: &h2.Info{
-					Status: &h2.Status{
-						Code: h2.Status_DEPLOYED,
-					},
-				},
-				Chart: &h2chart.Chart{
-					Metadata: &h2chart.Metadata{},
-					Values: &h2chart.Config{
-						Raw: "port: 8080\n",
-					},
-				},
-				Version: 5,
-				Config: &h2chart.Config{
-					Raw: "port: 3000\nuser:\n  name: user1\n  password: \"123456\"\n",
-				},
-			},
-		},
-		{
-			Description:         "Incomplete Helm 3 Release",
-			ShouldFail:          true,
-			MarshallingFunction: asResponse,
-			Helm3Release:        h3.Release{Name: "Incomplete"},
-			Helm2Release:        h2.Release{Name: "Incomplete"},
-		},
 	}
 
 	for _, test := range tests {
 		t.Run(test.Description, func(t *testing.T) {
-			// A panic is a failure for the test
+			// Capture the panic and report it in an orderly fashion
 			defer func() {
 				if r := recover(); r != nil {
-					if !test.ShouldFail {
-						t.Errorf("Not expected to fail, yet got a panic: %v. \nStacktrace: \n%s", r, string(debug.Stack()))
-					}
+					t.Errorf("Got a panic: %v. \nStacktrace: \n%s", r, string(debug.Stack()))
 				}
 			}()
 			// Perform conversion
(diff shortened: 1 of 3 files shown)


cmd/kubeops/internal/handler/dashboardcompat_test.go
```diff
@@ -0,0 +1,200 @@
...
+				Version: 1,
+				Config: map[string]interface{}{
+					"port": 3000,
+					"user": map[string]interface{}{
+						"name":     "user1",
+						"password": "123456",
+					},
+				},
+			},
+			Helm2Release: h2.Release{
```
andresmgot: shouldn't this be empty if it fails?

cmd/kubeops/internal/handler/dashboardcompat_test.go
```diff
@@ -0,0 +1,200 @@
...
+				t.Errorf("Expected to fail, but are equal %v", cmp.Diff(h3Marshalled, h2Marshalled))
+			}
+			if !test.ShouldFail && !areEqual {
+				t.Errorf("Not equal: %s", cmp.Diff(h3Marshalled, h2Marshalled))
+			}
+		})
+	}
+}
+
+func asResponse(data h2.Release) string {
```
andresmgot: can you explain why you are using `asResponse`? Why not using directly a marshall function?

cmd/kubeops/internal/handler/dashboardcompat_test.go
```diff
@@ -0,0 +1,200 @@
...
+	"testing"
+)
+
+func TestNewDashboardCompatibleRelease(t *testing.T) {
+	type testScenario struct {
+		// Scenario params
+		Description         string
+		Helm3Release        h3.Release
+		Helm2Release        h2.Release
+		MarshallingFunction func(h2.Release) string
```
andresmgot: this is always `asResponse`, in which case you plan to change this function?

cmd/kubeops/internal/handler/handler_test.go
```diff
@@ -139,13 +175,19 @@ func TestActions(t *testing.T) {
...
 				CreateRelease(*cfg, response, req, test.Params)
 			default:
 				t.Errorf("Unexpected action %s", test.Action)
 			}
 			// Check result
 			if response.Code != test.StatusCode {
+				// TODO Remove this `if` statement after the memory driver bug is fixed
+				if test.StatusCode == 404 && response.Code == 500 {
+					t.SkipNow()
+				}
```
andresmgot: this is a bit difficult to notice, I would rather add a field "skip: bool" to the test struct and explicitly mark some tests to be skipped (and add a link to the Github issue that should be fixed)

cmd/kubeops/internal/handler/dashboardcompat_test.go
latiif: Making the `Helm2Release` would make it fail, sure.
Here we test that all other information are identical, except for the version.

cmd/kubeops/internal/handler/dashboardcompat_test.go
latiif: If we add a test scenario where we test `newDashboardCompatibleRelease` against for example `spew.Sdump`, or a hand-written one where for example not all fields are relevant/populated etc.

cmd/kubeops/internal/handler/dashboardcompat_test.go
andresmgot: I see. The thing I was missing is that the function never fails (I doesn't return an error) so the `ShouldFail` param is a bit misleading. Actually this test is not testing any functional code. As far as I can see, you are giving a wrong expected result and verifying that the test is "failing" (so this is more a test for the test).

If I am correct with the above, I suggest to move this test and remove the `ShouldFail` param for this to be more readable. Also, I would make it so all the Helm2Release are equals to whatever we set in the test. That way, if both releases are not equal, we know exactly how it should behave in that case.

cmd/kubeops/internal/handler/dashboardcompat_test.go
andresmgot: still not sure about this ^

cmd/kubeops/internal/handler/dashboardcompat_test.go
latiif: You are correct. I will remove it. Thanks for pointing it out.

cmd/kubeops/internal/handler/dashboardcompat_test.go
latiif: In the current setting, the result from the function `newDashboardCompatibleRelease` is used in `kubeops/internal/handler` [here](https://github.com/kubeapps/kubeapps/blob/master/cmd/kubeops/internal/handler/handler.go#L170) & [here](https://github.com/kubeapps/kubeapps/blob/master/cmd/kubeops/internal/handler/handler.go#L183), which both marshal the data using `github.com/unrolled/render`. To avoid digging in the aforementioned library _and_ to enable further tests where `newDashboardCompatibleRelease` are marshalled/viewed in different ways, I created `asResponse` as a helper function to mimic the behaviour in `/cmd/kubeops/internal/handler/handler.go`.

If my reasoning seems too stretched I can just bake it into the test.

cmd/kubeops/internal/handler/dashboardcompat_test.go
latiif: After a closer inspection, I can say that:
While it's true that the function `newDashboardCompatibleRelease` doesn't return an error, it still can panic, which it does for this scenario.
```go
{
	Description:         "Incomplete Helm 3 Release",
	ShouldFail:          true,
	MarshallingFunction: asResponse,
	Helm3Release:        h3.Release{Name: "Incomplete"},
	Helm2Release:        h2.Release{Name: "Incomplete"},
},
```

Maybe _`shouldPanic`_ instead of `shouldFail`?

cmd/kubeops/internal/handler/dashboardcompat_test.go
andresmgot: That's fine, a comment explaining that would be helpful for the future

latiif:
Comment added for:
- The field `MarshallingFunction`
- The function `asResponse`

For further improvements or tests on the `dashboardcompat.go`

cmd/kubeops/internal/handler/dashboardcompat_test.go
```diff
@@ -0,0 +1,147 @@
...
+				Config: &h2chart.Config{
+					Raw: "port: 3000\nuser:\n  name: user1\n  password: \"123456\"\n",
+				},
+			},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.Description, func(t *testing.T) {
+			// Capture the panic and report it in an orderly fashion
```
absoludity: I'm confused as to why we're handling a panic that we've discovered in tests, rather than removing the panic from the code? What causes the panic... if I run these tests without the panic handling, it just succeeds.

absoludity: I came back to this later today, found the cause of the panic, added some failing tests and fixed. See what you think: https://github.com/kubeapps/kubeapps/pull/1445

PR #1436, opened 2020-01-10
[helm3] Return deleted time if exists

andresmgot:
### Description of the change

The deletion time was missing in the response given by `GetRelease`. This is used to show a "Purge" button when viewing a deleted release:

Screenshot from 2020-01-10 16-51-08

The tests for `GetRelease` has not been merged yet so I will add it once that's ready.

cc/ @latiif @SimonAlling

### DIFF
--- a/cmd/kubeops/internal/handler/dashboardcompat.go
+++ b/cmd/kubeops/internal/handler/dashboardcompat.go
@@ -7,7 +7,9 @@ package handler
 import (
 	"strings"
 
+	"github.com/golang/protobuf/ptypes"
 	"github.com/golang/protobuf/ptypes/any"
+	"github.com/golang/protobuf/ptypes/timestamp"
 	log "github.com/sirupsen/logrus"
 	"gopkg.in/yaml.v2"
 	h3chart "helm.sh/helm/v3/pkg/chart"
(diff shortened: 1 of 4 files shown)


cmd/kubeops/internal/handler/dashboardcompat.go
```diff
@@ -17,9 +19,17 @@ import (
 )
 
 func newDashboardCompatibleRelease(h3r h3.Release) h2.Release {
+	var deleted *timestamp.Timestamp
+	if !h3r.Info.Deleted.IsZero() {
+		var err error
+		deleted, err = ptypes.TimestampProto(h3r.Info.Deleted.Time)
+		if err != nil {
+			log.Errorf("Failed to parse deletion time %v", err)
```
absoludity: This should, imo, either be a failure (preferred) or we set a deleted time (some epoch) so that the UI at least knows it has been deleted. I don't think ignoring errors (because it's easier than pushing errors up the stack) is going to save us time longer-term.

On another line of thought: why is the UI depending on the timestamp, given that if we have a record of a deleted release, then (I think?) by definition, we have a record which can be purged? (as once it is purged, we have no record of the deleted release).

+1, but see what you think.

andresmgot: > This should, imo, either be a failure (preferred) or we set a deleted time (some epoch) so that the UI at least knows it has been deleted. I don't think ignoring errors (because it's easier than pushing errors up the stack) is going to save us time longer-term.

I didn't want to break the structure of the code (and it's unlikely that this function would fail) but okay, we can return the error.

> On another line of thought: why is the UI depending on the timestamp, given that if we have a record of a deleted release, then (I think?) by definition, we have a record which can be purged? (as once it is purged, we have no record of the deleted release).

The UI depends on the `status.deleted` field to be populated (with any info) to identify a release as deleted. That's the only identifier that we have (or at least had in helm 2, I am not sure if something new is available) to know if a release is deleted.

absoludity: Ah - I see. I'd misread and thought the code already knew this was deleted, but see it's based on the timestamp itself.

PR #1438, opened 2020-01-12
Add tests for DeleteRelease

latiif:
### Description of the change

Tests new functionality introduced by #1420 

### Benefits

- Adds test for DeleteRelease in /pkg/agent
- Adds action="delete" tests for /kubeops/handler

### Possible drawbacks
Skip tests blocked by memory driver bug

### Applicable issues

- Tests #1420 
- Related to #1434 

### Additional information

Contains a `TODO` in `/cmd/kubeops/internal/handler/handler_test.go` with instructions on what to do when the [memory driver bug](https://github.com/helm/helm/pull/7372) is fixed.

### DIFF
--- a/cmd/kubeops/internal/handler/handler_test.go
+++ b/cmd/kubeops/internal/handler/handler_test.go
@@ -192,6 +192,7 @@ func TestActions(t *testing.T) {
 	for _, test := range tests {
 		t.Run(test.Description, func(t *testing.T) {
 			// TODO Remove this `if` statement after the memory driver bug is fixed
+			// Memory Driver Bug: https://github.com/helm/helm/pull/7372
 			if test.Skip {
 				t.SkipNow()
 			}
--- a/cmd/kubeops/internal/handler/handler_test.go
+++ b/cmd/kubeops/internal/handler/handler_test.go
@@ -52,6 +52,7 @@ func TestActions(t *testing.T) {
 		Description      string
 		ExistingReleases []release.Release
 		DisableAuth      bool
+		Skip             bool //TODO: Remove this when the memory bug is fixed
 		// Request params
 		RequestBody  string
 		RequestQuery string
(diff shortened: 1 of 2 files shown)


cmd/kubeops/internal/handler/handler_test.go
```diff
@@ -117,10 +118,83 @@ func TestActions(t *testing.T) {
...
+			// Expected result
+			StatusCode:        404,
+			RemainingReleases: []release.Release{},
+			ResponseBody:      "",
+		},
 	}
 
 	for _, test := range tests {
 		t.Run(test.Description, func(t *testing.T) {
+			// TODO Remove this `if` statement after the memory driver bug is fixed
```
absoludity: Thanks @latiif . Can you reference the actual Issue/PR (PR in this case) to make it easier to audit later? Thanks. (both in this PR and the previous test one)

pkg/agent/agent_test.go
```diff
@@ -324,6 +324,52 @@ func TestListReleases(t *testing.T) {
...
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.description, func(t *testing.T) {
+			cfg := newActionConfigFixture(t)
+			makeReleases(t, cfg, tc.releases)
+			err := DeleteRelease(cfg, tc.releaseToDelete, true)
+			t.Logf("error: %v", err)
+			if isOk := err != nil == tc.shouldFail; !isOk {
```
absoludity: This might be easier to read as:
```
if didFail := err != nil; didFail != tc.shouldFail {

}
```

cmd/kubeops/internal/handler/handler_test.go
```diff
@@ -140,6 +214,8 @@ func TestActions(t *testing.T) {
 			switch test.Action {
 			case "create":
 				CreateRelease(*cfg, response, req, test.Params)
+			case "delete":
+				DeleteRelease(*cfg, response, req, test.Params)
```
absoludity: See what you think, but in https://github.com/kubeapps/kubeapps/pull/1439 I've used a separate test case for Rollback, rather than adding to this one. As mentioned there, I know it's probably just following the precedent from tiller-proxy, but it means tests for the individual handlers can't be focused. For example, here we've had to add a `t.SkipNow()` which is only relevant for a certain Delete action. Anyway, happy for this to land as is, just pointing out in case you choose to separate them. I can re-base my diff once yours lands.

pkg/agent/agent_test.go
latiif: This reads easier 👍

cmd/kubeops/internal/handler/handler_test.go
latiif: Smaller functions are always better.
To avoid overhead complications with multiple test PR's coming in right now, and the eventuality of merge conflicts. I plan to postpone splitting up the `TestActions` function to smaller ones, once the test PR's have landed. If that's okay with you.

absoludity: Yep, that's fine. Let's land this one (and your other one), then tomorrow I'll rebase my `Rollback` one with the separate test and we can go from there.

PR #1439, opened 2020-01-13
Add kubeops Rollback support and tests

absoludity:
### Description of the change

Adds Rollback release support to the helm3 kubeops agent and handler.

### Benefits

With this change, verified locally that rollback works from the dashboard UI.

### Applicable issues

Ref: #1056 

### Additional information

@latiif I've tried to simplify the handler tests a little so that we don't need the extra deref function, and also that each handler can have its own test case while still keeping things dry-ish. I realise that the existing test case for the tiller proxy was one large test case, but adds quite a bit to the cognitive load when maintaining them, IMO, when compared with smaller focused test cases. See what you think.

cc @SimonAlling


cmd/kubeops/internal/handler/handler_test.go
```diff
@@ -46,11 +47,22 @@ func newConfigFixture(t *testing.T) *Config {
...
 
+var releaseComparer = cmp.Comparer(func(x release.Release, y release.Release) bool {
+	return x.Name == y.Name &&
+		x.Version == y.Version &&
+		x.Namespace == y.Namespace &&
+		x.Info.Status == y.Info.Status &&
+		x.Chart.Name() == y.Chart.Name() &&
+		x.Manifest == y.Manifest &&
+		cmp.Equal(x.Config, y.Config) &&
+		cmp.Equal(x.Hooks, y.Hooks)
```
SimonAlling: Maybe a bit weird indentation? I'd probably do something like this by the way:

```go
return and(bool[]{
	x.Name == y.Name,
	x.Version == y.Version,
	// ...
	cmp.Equal(x.Hooks, y.Hooks),
})
```

But it seems like I'd have to implement `and` myself:

```go
func and(xs []bool) bool {
	for _, x := range xs {
		if !x {
			return false
		}
	}
	return true
}
```

cmd/kubeops/internal/handler/handler_test.go
```diff
@@ -147,19 +155,12 @@ func TestActions(t *testing.T) {
...
-					cmp.Equal(x.Hooks, y.Hooks)
-			})
-			if !cmp.Equal(releases, test.RemainingReleases, rlsComparer) {
-				t.Errorf("Unexpected remaining releases. Diff %s", cmp.Diff(releases, test.RemainingReleases, rlsComparer))
+			releases, err := cfg.ActionConfig.Releases.ListReleases()
+			if err != nil {
+				t.Fatalf("%+v", err)
+			}
+			if !cmp.Equal(releases, test.RemainingReleases, releaseComparer) {
+				t.Errorf("Unexpected remaining releases. Diff %s", cmp.Diff(releases, test.RemainingReleases, releaseComparer))
```
SimonAlling: I haven't run the code, but wouldn't something like this be nice?
```suggestion
				t.Errorf("Unexpected remaining releases. Diff:\n%s", cmp.Diff(releases, test.RemainingReleases, releaseComparer))
```

pkg/agent/agent.go
```diff
@@ -100,6 +100,20 @@ func UpgradeRelease(actionConfig *action.Configuration, name, valuesYaml string,
...
+func RollbackRelease(actionConfig *action.Configuration, releaseName string, revision int) (*release.Release, error) {
+	log.Printf("Rolling back %s to revision %d.", releaseName, revision)
+	rollback := action.NewRollback(actionConfig)
+	rollback.Version = revision
+	err := rollback.Run(releaseName)
+	if err != nil {
+		return nil, err
+	}
+
+	// The Helm3 rollback action does not return the new release, unlike the helm2 equivalent,
```
SimonAlling: Maybe like this (if we care about detail)?
```suggestion
	// The Helm 3 rollback action does not return the new release, unlike the Helm 2 equivalent,
```

pkg/agent/agent.go
```diff
@@ -100,6 +100,20 @@ func UpgradeRelease(actionConfig *action.Configuration, name, valuesYaml string,
...
+	// so we grab it explicitly as its required by Kubeapps.
```
SimonAlling: Detail here too:
```suggestion
	// so we grab it explicitly as it's required by Kubeapps.
```

cmd/kubeops/internal/handler/handler_test.go
latiif: Maybe as a Variadic Function?
```go
func and(exps ...bool) bool {
	for _, exp := range exps {
		if !exp {
			return false
		}
	}
	return true
}
```

andresmgot: we would need to maintain extra code. IMO it's fine as it is (I don't see the indentation weird)

SimonAlling: That's neat, @latiif!

SimonAlling: > we would need to maintain extra code. IMO it's fine as it is (I don't see the indentation weird)

Why would `x.Name == y.Name &&` not be indented and the rest of the expression indented one level? What does indentation represent then? If one wants to avoid a separate `and` function, surely the below formatting is more logical?

```go
return (
	x.Name == y.Name &&
	x.Version == y.Version &&
	// ...
	cmp.Equal(x.Hooks, y.Hooks)
)
```

(I'm personally not very fond of it either, because I'd prefer that adding another line 1) be as easy as duplicating and modifying the last one, and 2) not create a three-line diff. But that's orthogonal to my indentation concern.)

absoludity: I actually just moved this code from the existing tests of @latiif 's . Happy to improve it, but don't see a problem as is either.

Anyway, I'll hold off updating and landing this PR until @latiif's test PRs have landed.

SimonAlling: I decided to summarize the three alternatives. Please let me know what you think – not least because I think it's an interesting matter in general (and don't forget my earlier question: "What does indentation represent then?"). :smiley: 

### `and` function
#### Pros
* Very easy to add or remove a condition at the beginning or end.
* Clear diff when adding or removing a condition at the beginning or end.
* DRY: Doesn't repeat `&&` several times. (Debatable, but I for one believe the intended conveyed meaning is "_a_, _b_ and _c_ ANDed together", not "_a_ ANDed with _b_ ANDed with _c_ …".)
* Indentation makes perfect sense.
#### Cons
* `and` must be maintained.

### `return (` with conditions on separate lines
#### Pros
* Very easy to add or remove a condition at the beginning.
* Clear diff when adding or removing a condition at the beginning.
* Indentation makes perfect sense.
#### Cons
* Not as easy to add or remove a condition at the end.
* Not very clear diff when adding or removing a condition at the end.
* WET: Repeats `&&` several times.

### `return x.Name == y.Name &&` (current PR content)
#### Pros
* Saves two lines compared to the other alternatives.
* ?
#### Cons
* Not as easy to add or remove a condition at the beginning or end.
* Not very clear diff when adding or removing a condition at the beginning or end.
* WET: Repeats `&&` several times.
* Indentation doesn't make sense. (Feel free to convince me of the contrary.)

latiif: My issue with splitting up tests for different actions is that it would lead to duplication of code and maintaining a each test function on its own, meaning that we would end up with something similar to this:

```go
func TestGetAction(t *testing.T){
    //Defining the struct for the test scenarios
    type TestScenario struct{
        ⋮
        ⋮
    }
    //Filling of the test scenarios array
    tests := []TestScenario {
        ⋮
        ⋮
        ⋮
    }
    // Performing the actual test and checking of results
   for _, test := range tests {
	    	     ⋮
	    	     ⋮
		}
	}
}

func TestDeleteAction(t *testing.T){
    //Defining the struct for the test scenarios
    type TestScenario struct{
        ⋮
        ⋮
    }
    //Filling of the test scenarios array
    tests := []TestScenario {
        ⋮
        ⋮
        ⋮
    }
    // Performing the actual test and checking of results
   for _, test := range tests {
	    	        ⋮
	    	        ⋮
		}
	}
}
// Do the same for each new action
⋮
⋮
```
This IMHO leads to a bigger code base to maintain, my suggestion is to: 
1. Define the generic structure for test scenarios on a file-level (as opposed to having it inside the `TestXxxAction`.
2. Define a gerenic function that runs the tests extractin common code from the code snippet above (akin to `TestActions` but with an empty `tests=[]TestScenario{}`.
3. With each new action, define a function of the form `func getXxxActionTests() []TestScenario`, that returns tests for the `Xxx` action.
4. Populate the `tests` inside `TestActions` by calling a function that returns desired test scenarios. See below: 
```go
tests := []TestScenario {
    getGetActionTests(),
    getDeleteActionTests(),
    ⋮  
}
```
This would faciliate adding new test scenarios, grouping them and only requires us to maintain one function i.e. `TestActions`.

latiif:
> @latiif I've tried to simplify the handler tests a little so that we don't need the extra deref function, and also that each handler can have its own test case while still keeping things dry-ish. I realise that the existing test case for the tiller proxy was one large test case, but adds quite a bit to the cognitive load when maintaining them, IMO, when compared with smaller focused test cases. See what you think.

My issue with splitting up tests for different actions is that it would lead to duplication of code and maintaining a each test function on its own, meaning that we would end up with something similar to this:

```go
func TestGetAction(t *testing.T){
    //Defining the struct for the test scenarios
    type TestScenario struct{
        ⋮
        ⋮
    }
    //Filling of the test scenarios array
    tests := []TestScenario {
        ⋮
        ⋮
        ⋮
    }
    // Performing the actual test and checking of results
   for _, test := range tests {
	    	     ⋮
	    	     ⋮
		}
	}
}

func TestDeleteAction(t *testing.T){
    //Defining the struct for the test scenarios
    type TestScenario struct{
        ⋮
        ⋮
    }
    //Filling of the test scenarios array
    tests := []TestScenario {
        ⋮
        ⋮
        ⋮
    }
    // Performing the actual test and checking of results
   for _, test := range tests {
	    	        ⋮
	    	        ⋮
		}
	}
}
// Do the same for each new action
⋮
⋮
```
This IMHO leads to a bigger code base to maintain, my suggestion is to: 
1. Define the generic structure for test scenarios on a file-level (as opposed to having it inside the `TestXxxAction`.
2. Define a gerenic function that runs the tests extractin common code from the code snippet above (akin to `TestActions` but with an empty `tests=[]TestScenario{}`.
3. With each new action, define a function of the form `func getXxxActionTests() []TestScenario`, that returns tests for the `Xxx` action.
4. Populate the `tests` inside `TestActions` by calling a function that returns desired test scenarios. See below: 
```go
tests := []TestScenario {
    getGetActionTests(),
    getDeleteActionTests(),
    ⋮  
}
```
This would faciliate adding new test scenarios, grouping them and only requires us to maintain one function i.e. `TestActions`.

cmd/kubeops/internal/handler/handler_test.go
absoludity: > I decided to summarize the three alternatives. Please let me know what you think – not least because I think it's an interesting matter in general (and don't forget my earlier question: "What does indentation represent then?"). smiley

I didn't write the code, so when moving this function so that I could also use it in another test, I explicitly tried not to change Latiif's code (as little as possible, at least). You'll see this function was just copied and pasted in this PR, with its existing indentation.

That said, since it seems to be quite important to you, let me answer your questions with my own thoughts...

> What does indentation represent then?

I don't particularly like the indentation either (visually or semantically), but it's what the gofmt wants, so I don't usually bother thinking more about it. It is formatting after all and gofmt was created for this reason (ie. both to avoid spending too much time discussing formatting in PRs, as well as creating code which is formatted consistently across code bases etc.). So if you look at the actual source for go, you'll see similar formatting of long booleans throughout, just grep for "\&\&$" in the go source.
 
> ### `and` function
> #### Pros
> 
>     * Very easy to add or remove a condition at the beginning or end.
> 
>     * Clear diff when adding or removing a condition at the beginning or end.
> 
>     * DRY: Doesn't repeat `&&` several times. (Debatable, but I for one believe the intended conveyed meaning is "_a_, _b_ and _c_ ANDed together", not "_a_ ANDed with _b_ ANDed with _c_ …".)
> 
>     * Indentation makes perfect sense.
> 
> 
> #### Cons
> 
>     * `and` must be maintained.

I totally agree that using `and` results in the nicest formatting (visually, semantically and future diffs), though I don't agree that it's what we should do with our golang code, for the reasons above (which are also outlined in the original blog post at https://blog.golang.org/go-fmt-your-code too).

That said, even though I don't agree, I'm happy to update and use `and` here, in this specific case, for 3 reasons:
 1. You think it's important so I want to treat it as important,
 2. I do like the formatting more than what gofmt does (even though I don't think this is a good reason to change the code, I do feel the same niggle),
 3. I'm not keen to spend more time discussing the PR rather than landing the PR (and I don't like landing a PR, even with an approval, if a disagreement hasn't been resolved).

So, change done :)

absoludity:
> > @latiif I've tried to simplify the handler tests a little so that we don't need the extra deref function, and also that each handler can have its own test case while still keeping things dry-ish. I realise that the existing test case for the tiller proxy was one large test case, but adds quite a bit to the cognitive load when maintaining them, IMO, when compared with smaller focused test cases. See what you think.
> 
> My issue with splitting up tests for different actions is that it would lead to duplication of code and maintaining each test function on its own,

I think we can avoid duplication of code, but to be clear, I'm in the camp of unit tests not being perfectly DRY, but rather being DAMP (*sigh*, someone must have thought hard to create that acronym :P, but for me the important part for unit tests is to lean further towards readability in the balance of readability vs DRY than I would for actual non-test code.)

So isolating tests for one specific function (in this case, rolling back a release), is not a bad thing, it means I can minimise what future readers need to load into their cognitive stack to read the test and avoids mixing the needs of different functionality when updating test code. Of course I'll do my best for it to remain DRY but I'll try not to do so at the expense of readability and maintainability, so there will be some infrastructure repeated (eg. the actual `t.Run(...)`). Let me explain with your examples below.

> meaning that we would end up with something similar to this:
> 
> ```go
> func TestGetAction(t *testing.T){
>     //Defining the struct for the test scenarios
>     type TestScenario struct{
>         ⋮
>         ⋮
>     }

You'll see that my test doesn't define the struct separately as you have done above. Instead I'm just using an anonymous struct with exactly what I need to test this function. This isn't my idea, but rather, just what the golang authors tend to do in the golang src. For example:

https://golang.org/src/net/http/transport_test.go#L5414 (though if you grep the golang src for `"\[\]struct"` you'll see loads of similar examples where test cases are created in anon structs)

or an example where the anonymous struct is just part of the iteration of the tests:

https://golang.org/src/strings/builder_test.go#L115

>     //Filling of the test scenarios array
>     tests := []TestScenario {
>         ⋮
>         ⋮
>         ⋮
>     }

Yes, the test cases for a specific piece of functionality (in this case, Rollback) needs to be populated with the data for those specific tests only - this is a good thing, IMO.

>     // Performing the actual test and checking of results
>    for _, test := range tests {
> 		t.Run(test.Description, func(t *testing.T) {
> 	    	     ⋮
> 	    	     ⋮
> 		}
> 	}
> }

Yes, the loop to run each set of tests is repeated. This is something which you'll note in the source code of go itself, that is not DRY'd up, but generally left explicit so that each `Test...(t *testing.T)` test is as readable as possible without (a) needing to look up what some helper does, and (b) not needing to accommodate other test functions when updating a helper.

So in cases like this, I try to follow patterns that I see in the golang source code (it is written with the assumption that it will be used in this way). And I have been bitten over the years by test code where DRY has been prioritised over readablitity... you won't notice at first, a small abstraction here, then it gets used in other tests, add a few extra fields to accommodate a specific test, gradually it continues to grow into something hard to read and maintain (as a future reader).

[snip]

> 
> This IMHO leads to a bigger code base to maintain, my suggestion is to:
> 
>     1. Define the generic structure for test scenarios on a file-level (as opposed to having it inside the `TestXxxAction`).

As mentioned, IMO, an anonymous struct is much better suited to the needs of an individual test, rather than adding fields required for testing one piece of functionality which are irrelevant for others. For example, you'll see in my `TestRollbackAction`, I've not (yet) needed the fields: `DisableAuth`, `Skip`, `RequestBody`, `Action`. It may be that in the future I need to add some of them, but that decision can be based on the needs of that one piece of functionality being tested only, and I don't need to think about how it affects other tests, nor look for the definition somewhere else - it's right there. So yes, marginally less DRY, but IMO much more readable, but my opinion isn't so important, so I tend to follow the golang source.

There are cases where a small helper function improves the readability, such as your `createRelease` helper which I've used here (with a small change to return a pointer so that we don't need the `derefReleases` etc.) to make populating the test cases explicit, clear and minimal.

> 
>     2. Define a generic function that runs the tests extracting common code from the code snippet above (akin to `TestActions` but with an empty `tests=[]TestScenario{}`.
> 
>     3. With each new action, define a function of the form `func getXxxActionTests() []TestScenario`, that returns tests for the `Xxx` action.
> 
>     4. Populate the `tests` inside `TestActions` by calling a function that returns desired test scenarios. See below:
> 
> 
> ```go
> tests := []TestScenario {
>     getGetActionTests(),
>     getDeleteActionTests(),
>     ⋮  
> }
> ```
> 
> This would faciliate adding new test scenarios, grouping them and only requires us to maintain one function i.e. `TestActions`.
> 
> (edit: fixed typos)

So yes, these are all excellent suggestions if having DRY code in unit tests is more important than the cognitive load for future readers and maintainers. I do understand that you don't see that it would be more cognitive load, but consider both the extra abstractions (which may not seem like much at first) and the way they tend to grow over time. The aim is that an engineer with no previous context of the specific test can come in, read the tests and see what they would see everywhere else (ie. similar to the golang source, and without needing to look up abstractions), add a failing test, perhaps modify the assertions or test output, and continue.

There's plenty of views on this particular point (DAMP vs DRY in unit tests), and although I do lean the balancing act more towards explicit readability in unit tests over DRY, it's the golang authors experience that I tend to rely on rather than my own.

(Sorry for the long post, I didn't want to land this without replying to your points. Hope it was helpful in some way)

cmd/kubeops/internal/handler/handler_test.go
SimonAlling: Thank you for a well-written reply with good points and important context, @absoludity!

PR #1441, opened 2020-01-13
[Helm3] Return chart version and metadata when listing

andresmgot:
### Description of the change

When listing apps, the expected result (`AppOverview`) is expected to have the chart version and the chart metadata included in the response. This is used to build hte AppList view, showing updates if necessary.

See the tiller-proxy implementation:

https://github.com/kubeapps/kubeapps/blob/8bb060e15ad315c1561d0133cbd30827d292dcd2/pkg/proxy/proxy.go#L216

### Benefits

Before this change:
Screenshot from 2020-01-13 13-42-07

After the change:

Screenshot from 2020-01-13 13-36-11

### Additional information

cc/ @latiif @SimonAlling


pkg/agent/agent.go
```diff
@@ -178,9 +178,22 @@ func ParseDriverType(raw string) (StorageForDriver, error) {
...
+			Home:        r.Chart.Metadata.Home,
+			Sources:     r.Chart.Metadata.Sources,
+			Version:     r.Chart.Metadata.Version,
+			Description: r.Chart.Metadata.Description,
+			Keywords:    r.Chart.Metadata.Keywords,
+			Icon:        r.Chart.Metadata.Icon,
+			ApiVersion:  r.Chart.Metadata.APIVersion,
+			Tags:        r.Chart.Metadata.Tags,
+			AppVersion:  r.Chart.Metadata.AppVersion,
+		},
```
SimonAlling: I think we could and should re-use [`compatibleMetadata`](https://github.com/kubeapps/kubeapps/blob/8bb060e15ad315c1561d0133cbd30827d292dcd2/cmd/kubeops/internal/handler/dashboardcompat.go#L40). I managed to make it typecheck with little effort at least.

This would of course mean said function would have to be moved to somewhere in `pkg`. At closer thought, maybe it makes more sense to put the entire `dashboardcompat.go` in `pkg`. It really is more of a package than a command.

pkg/agent/agent_test.go
```diff
@@ -230,23 +257,31 @@ func TestListReleases(t *testing.T) {
...
+					ChartMetadata: chartv1.Metadata{
+						Version: "1.0.0",
+						Icon:    "https://example.com/icon.png",
+					},
 				},
 				proxy.AppOverview{
 					ReleaseName: "wordpress",
 					Namespace:   "dev",
-					Version:     "2",
+					Version:     "1.0.0",
```
SimonAlling: Is this change trivially correct? If not, maybe add a note in the PR?

pkg/agent/agent_test.go
```diff
@@ -274,23 +313,31 @@ func TestListReleases(t *testing.T) {
...
```
SimonAlling: Is this change trivially correct? If not, maybe add a note in the PR?

pkg/agent/agent.go
andresmgot: That's a good point. At the end we have a package to convert releases from one format to the other so we can reuse it.

pkg/agent/agent_test.go
andresmgot: The version is irrelevant here but I can change the number so the title still applies :+1:

pkg/agent/agent_test.go
andresmgot: In this case it's not needed (the version is not relevant either)

pkg/agent/agent.go
```diff
@@ -183,12 +187,18 @@ func ParseDriverType(raw string) (StorageForDriver, error) {
...
 
-func appOverviewFromRelease(r *release.Release) proxy.AppOverview {
-	return proxy.AppOverview{
-		ReleaseName: r.Name,
-		Version:     strconv.Itoa(r.Version),
-		Icon:        r.Chart.Metadata.Icon,
-		Namespace:   r.Namespace,
-		Status:      r.Info.Status.String(),
+func appOverviewFromRelease(r *release.Release) (proxy.AppOverview, error) {
+	releaseHelm2, err := helm3to2.Convert(*r)
```
absoludity: Converting the release to a helm 2 release seems (1) unnecessary and (2) the only reason for requiring an error in the signature of this function? That is, if `helm3to2.ConvertMaintainers(r.Chart.Metadata.Maintainers)` could be used below instead (ie. just making `compatableMaintainers` public), then we don't need to handle the error at all (nor convert the whole release).

Fine either way, just a thought.

andresmgot: Well, we need the whole metadata (not just the maintainers) so the function to expose would be `compatibleMetadata` but yes, we could do that. I wanted to to keep the methods exposed by the compatibility package as minimum as possible but there is not a good reason for not exposing the metadata conversion.

SimonAlling: I for one see no problem with exporting `helm3to2` functions at will. They're all supposed to be pure and reusable anyway.

SimonAlling: Although one must keep in mind that some of the type signatures in `helm3to2` are essentially lies, since it's impossible in general to write a function from Helm 2 releases to Helm 3 releases.

PR #1444, opened 2020-01-14
Update to test with gke 1.13 and 1.14 rather than 1.12 and 1.13, since 1.12 master no longer available.

absoludity:
### Description of the change

Updates to test against GKE 1.13 and 1.14 rather than 1.12 and 1.13.

Update kubectl version from 1.12 to 1.13.

### Benefits

CircleCI began failing today when testing GKE_1_12_MASTER with the error:
```
ERROR: (gcloud.container.clusters.create) ResponseError: code=400, message=No valid versions with the prefix "1.12" found.

```
It's not clear to my why this errored while GKE_1_12_LATEST_RELEASE did not (though they both use the same `GKE_BRANCH="1.12"` env var, but nonetheless I saw in the [release notes from a week ago](https://cloud.google.com/kubernetes-engine/docs/release-notes) that certain 1.12 cluster versions would no longer be available.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,14 +63,14 @@ workflows:
             - test_dashboard
             - build_go_images
             - build_dashboard
-      - GKE_1_12_MASTER:
+      - GKE_1_14_MASTER:
           <<: *build_on_master
           requires:
             - test_go
             - test_dashboard
             - build_go_images
             - build_dashboard
-      - GKE_1_12_LATEST_RELEASE:
+      - GKE_1_14_LATEST_RELEASE:
           <<: *build_on_master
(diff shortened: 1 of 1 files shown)


andresmgot:
> It's not clear to my why this errored while GKE_1_12_LATEST_RELEASE did not (though they both use the same `GKE_BRANCH="1.12"` env var, but nonetheless I saw in the [release notes from a week ago](https://cloud.google.com/kubernetes-engine/docs/release-notes) that certain 1.12 cluster versions would no longer be available.

You are seeing the job "succeed" because it's being skipped. That job is only performed when the chart version gets updated.

PR #1445, opened 2020-01-14
Test and remove panic possibilities in newDashboardCompatibleRelease

absoludity:
### Description of the change

Related to a question I had at https://github.com/kubeapps/kubeapps/pull/1434#discussion_r366069755 , it was unclear why we were handling panics in tests rather than eliminating them.

I added a number of test cases which caused a panic as mentioned in the above comment, then fixed with a guard clause. While there, I also added some tests for the parsing of the Deleted timestamp.

### Benefits
A number of conditions which could cause a panic have been eliminated.

### Possible drawbacks
The error for the failed parsing of the Deletion timestamp is less explicit (so I could test it without doing string comparisons - *guilty* I'll update it if you think it's important)

### Applicable issues

  - fixes #

### DIFF
--- a/cmd/kubeops/internal/handler/dashboardcompat.go
+++ b/cmd/kubeops/internal/handler/dashboardcompat.go
@@ -19,13 +19,23 @@ import (
 	h2 "k8s.io/helm/pkg/proto/hapi/release"
 )
 
+var (
+	// ErrUnableToConvertWithoutInfo indicates that the input release had nil Info or Chart.
+	ErrUnableToConvertWithoutInfo = fmt.Errorf("unable to convert release without info")
+	// ErrUnableToParseDeletionTime indicates that the deletion time of the h3 chart could not be parsed.
+	ErrFailedToParseDeletionTime = fmt.Errorf("failed to parse deletion time")
+)
+
 func newDashboardCompatibleRelease(h3r h3.Release) (h2.Release, error) {
+	if h3r.Info == nil || h3r.Chart == nil || h3r.Chart.Metadata == nil {
+		return h2.Release{}, ErrUnableToConvertWithoutInfo
+	}
 	var deleted *timestamp.Timestamp
 	if !h3r.Info.Deleted.IsZero() {
 		var err error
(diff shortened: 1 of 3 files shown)


cmd/kubeops/internal/handler/dashboardcompat.go
```diff
@@ -19,13 +19,23 @@ import (
...
+	if h3r.Info == nil || h3r.Chart == nil || h3r.Chart.Metadata == nil {
+		return h2.Release{}, ErrUnableToConvertWithoutInfo
+	}
 	var deleted *timestamp.Timestamp
 	if !h3r.Info.Deleted.IsZero() {
 		var err error
 		deleted, err = ptypes.TimestampProto(h3r.Info.Deleted.Time)
 		if err != nil {
-			return h2.Release{}, fmt.Errorf("Failed to parse deletion time %v", err)
+			return h2.Release{}, ErrFailedToParseDeletionTime
```
andresmgot: I think it's important to have the error text somewhere. That way, if a user reports that it's hitting this error we'd get more info (for example if `.IsZero()` is not working as expected or if the `Time` is no longer valid).

If the only reason to set a fixed message is the test, you can still do `string.Contains(got, want)` instead of `got != want` which is a bit more flexible and unlikely to report a false-positive.

andresmgot:
> But are we sure/How sure are we, the two types of errors are the only ones that could occur?

As far as I can see, the fields that use pointers (which are the ones that may throw a `panic` if they are not defined) are covered (`*h3r.Info`, `*h3r.Chart` and `*h3r.Chart.Metadata`). Now there is not an expected panic (but that doesn't mean that we are missing something) but we can always extend the tests/code if we find new cases in which the code panics.

andresmgot:
Merging this because it will conflict with my PR in #1441 I can open a follow-up PR with my comment later on.

cmd/kubeops/internal/handler/dashboardcompat.go
absoludity: See #1451  Thanks!

PR #1447, opened 2020-01-14
Return error when failing to parse deletion time

andresmgot:
### Description of the change

Context: https://github.com/kubeapps/kubeapps/pull/1445#discussion_r366235802

### DIFF
--- a/pkg/chart/helm3to2/dashboardcompat_test.go
+++ b/pkg/chart/helm3to2/dashboardcompat_test.go
@@ -2,6 +2,7 @@ package helm3to2
 
 import (
 	"net/http/httptest"
+	"strings"
 
 	"github.com/golang/protobuf/ptypes/timestamp"
 	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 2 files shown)


pkg/chart/helm3to2/dashboardcompat_test.go
```diff
@@ -206,8 +207,10 @@ func TestConvert(t *testing.T) {
 		t.Run(test.Description, func(t *testing.T) {
 			// Perform conversion
 			compatibleH3rls, err := Convert(test.Helm3Release)
-			if got, want := err, test.ExpectedError; got != want {
-				t.Errorf("got: %v, want: %v", got, want)
+			if test.ExpectedError != nil || err != nil {
```
absoludity: This line is problematic: I assume you added it because you saw a panic on either `got.Error()` or `want.Error()` (or both) when either of them is nil. So in the case where a test would fail (say a test where `test.ExpectedError == nil` but an error was returned), then you'd still see a panic. To avoid that, you'd need to replace the `||` with `&&`, but then you're effectively skipping this assertion when we don't want to.

We could fix that here easily enough by first testing to ensure we have an error when we didn't expect one (or vice-verse):
```
if gotError, wantError := err != nil, test.ExpectedError != nil; gotError != wantError {
    ...
}
if test.ExpectedError != nil && err != nil {
...(something that uses .Error() on both)
}
```
But, I think this alternative, using the new error wrapping, is simpler and perhaps a good precedent: #1451

andresmgot:
Thanks for the comment and investigating this! I indeed prefer your solution in #1451, it's cleaner and more readable :+1:

PR #1448, opened 2020-01-14
Enable apprepositoriess endpoint for kubeops

andresmgot:
### Description of the change

In order to be able to create apprepositories when using kubeops (Helm 3), it's necessary to use the same code and endpoint than with tiller-proxy.

### DIFF
--- a/cmd/kubeops/main.go
+++ b/cmd/kubeops/main.go
@@ -14,6 +14,7 @@ import (
 	"github.com/heptiolabs/healthcheck"
 	"github.com/kubeapps/kubeapps/cmd/kubeops/internal/handler"
 	"github.com/kubeapps/kubeapps/pkg/agent"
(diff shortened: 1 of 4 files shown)


cmd/kubeops/main.go
```diff
@@ -76,6 +77,18 @@ func main() {
 	addRoute("PUT", "/namespaces/{namespace}/releases/{releaseName}", handler.OperateRelease)
 	addRoute("DELETE", "/namespaces/{namespace}/releases/{releaseName}", handler.DeleteRelease)
 
+	// Backend routes unrelated to kubeops functionality.
+	// TODO(mnelson): Once the helm3 support is complete and tiller-proxy is being removed,
```
absoludity: This TODO can perhaps be removed?

andresmgot: I agree, I just didn't know if you were thinking on something else

PR #1449, opened 2020-01-14
Use kubeops serviceaccount to retrieve charts

andresmgot:
### Description of the change

Kubeops is in charge of downloading the chart info and deploy it. For doing so, it needs to retrieve apprepositories and secrets in the `kubeapps` namespace. We are not using the user token for this task so the user doesn't need to have permissions for the above (context at: #1110).

cc/ @SimonAlling @latiif

### DIFF
--- a/cmd/kubeops/internal/handler/handler.go
+++ b/cmd/kubeops/internal/handler/handler.go
@@ -66,6 +66,9 @@ func WithHandlerConfig(storageForDriver agent.StorageForDriver, options Options)
 		return func(w http.ResponseWriter, req *http.Request, params handlerutil.Params) {
 			namespace := params[namespaceParam]
 			token := auth.ExtractToken(req.Header.Get(authHeader))
+
+			// User configuration and clients, using user token
+			// Used to perform Helm operations
 			restConfig, err := NewInClusterConfig(token)
 			if err != nil {
 				// TODO log details rather than return potentially sensitive details in error.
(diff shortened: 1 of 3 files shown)


cmd/kubeops/internal/handler/handler.go
```diff
@@ -72,19 +72,31 @@ func WithHandlerConfig(storageForDriver agent.StorageForDriver, options Options)
...
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				return
+			}
+			appRepoClient, err := appRepo.NewForConfig(svcRestConfig)
+			if err != nil {
+				// TODO log details rather than return potentially sensitive details in error.
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				return
+			}
+			actionConfig, err := agent.NewActionConfig(storageForDriver, restConfig, userKubeClient, namespace)
```
absoludity: This line belongs on line 81 I think? That is, it is only dependent on `userKubeClient`, and doesn't use `appRepoClient` or `svcKubeClient`.

cmd/kubeops/internal/handler/handler.go
```diff
@@ -72,19 +72,31 @@ func WithHandlerConfig(storageForDriver agent.StorageForDriver, options Options)
...
 			}
-			appRepoClient, err := appRepo.NewForConfig(restConfig)
+			svcRestConfig, err := rest.InClusterConfig()
 			if err != nil {
 				// TODO log details rather than return potentially sensitive details in error.
 				http.Error(w, err.Error(), http.StatusInternalServerError)
 				return
 			}
-			actionConfig, err := agent.NewActionConfig(storageForDriver, restConfig, kubeClient, namespace)
+			svcKubeClient, err := kubernetes.NewForConfig(svcRestConfig)
```
absoludity: Let's document right above this line why we're creating a separate client using service credentials for the chart client.

PR #1450, opened 2020-01-14
Show forbidden errors for Helm 3

andresmgot:
### Description of the change

With Helm 3, we are not pre-processing the chart to return the forbidden actions, we are leaving that to Helm. Then we are just returning the error given by Helm if it exists. We need to adapt the dashboard code and the response a bit in order to give an appropriated response.

### Benefits

Now forbidden errors are easier to debug with Helm 3:

Screenshot from 2020-01-14 17-52-13

### Possible drawbacks

If we compare this solution to the one in tiller-proxy (Helm 2) we have two drawbacks due to just forwarding the error given by Helm:

 - The error message is more confusing (e.g. `create: failed to create: secrets is forbidden ...`
 - Only one forbidden action is returned (the first one to fail I guess) instead of all the forbidden action.

This is the same error for Helm 2 for the sake of comparison:

Screenshot from 2020-01-14 18-00-39

To mitigate this a bit, we could parse the error returned by Helm but that would be prone to errors if they change the format in the future.

### DIFF
--- a/dashboard/src/components/ErrorAlert/PermissionsErrorAlert.tsx
+++ b/dashboard/src/components/ErrorAlert/PermissionsErrorAlert.tsx
@@ -9,12 +9,13 @@ import PermissionsListItem from "./PermissionsListItem";
 interface IPermissionsErrorPage {
   action: string;
   roles: IRBACRole[];
+  rawMessage: string;
   namespace: string;
 }
 
 class PermissionsErrorPage extends React.Component<IPermissionsErrorPage> {
   public render() {
-    const { action, namespace, roles } = this.props;
+    const { action, namespace, roles, rawMessage } = this.props;
     return (
       <UnexpectedErrorAlert
         title={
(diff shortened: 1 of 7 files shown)


dashboard/src/components/ErrorAlert/PermissionsErrorAlert.tsx
```diff
@@ -26,12 +27,20 @@ class PermissionsErrorPage extends React.Component<IPermissionsErrorPage> {
         showGenericMessage={false}
       >
         <div>
-          <p>Ask your administrator for the following RBAC roles:</p>
-          <ul className="error__permisions-list">
-            {roles.map((r, i) => (
-              <PermissionsListItem key={i} namespace={namespace} role={r} />
-            ))}
-          </ul>
+          <p>Ask your administrator for the following permissions:</p>
```
absoludity: I'd put this `<p>` inside the following condition, so that if roles are defined, we say the current msg, but otherwise, if displaying the raw content, we say something more appropriate, such as "The following error was returned by helm:"

absoludity: I've outlined this in https://github.com/kubeapps/kubeapps/pull/1453

andresmgot: :+1: I have added `The following error was returned:` (without `helm`) because this component can be rendered with something else (for example, the service catalog).

PR #1456, opened 2020-01-15
Remove duplicated ingress in chart

andresmgot:
### Description of the change

When enabling the ingress setup, two hosts with identical information are generated. One because the `hostname` is set and another one because the section `hosts` is also configured (the legacy mechanism to set ingresses).

cc/ @juan131

### DIFF
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -77,15 +77,15 @@ ingress:
   ## The list of hostnames to be covered with this ingress record.
   ## Most likely this will be just one host, but in the event more hosts are needed, this is an array
   ##
-  hosts:
-    - name: kubeapps.local
-      path: /
-      ## Set this to true in order to enable TLS on the ingress record
-      ##
-      tls: false
-      ## If TLS is set to true, you must declare what secret will store the key/certificate for TLS
-      ##
-      tlsSecret: kubeapps.local-tls
+  # hosts:
+  #   - name: kubeapps.local
+  #     path: /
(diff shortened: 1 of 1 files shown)


chart/kubeapps/values.yaml
```diff
@@ -77,15 +77,15 @@ ingress:
...
-      tlsSecret: kubeapps.local-tls
+  # hosts:
+  #   - name: kubeapps.local
+  #     path: /
+  #     ## Set this to true in order to enable TLS on the ingress record
+  #     ##
+  #     tls: false
+  #     ## If TLS is set to true, you must declare what secret will store the key/certificate for TLS
+  #     ##
+  #     tlsSecret: kubeapps.local-tls
```
absoludity: Since `.Values.ingress.enabled` is false by default, won't removing the defaults for this deprecated option be a backwards incompatible change? (ie. as per the comment, something that should wait for 3.0.0)

Could we instead just add a condition around [the host block](https://github.com/kubeapps/kubeapps/blob/master/chart/kubeapps/templates/ingress.yaml#L36), like:

```
## The block below is deprecated and must removed on 3.0.0 
{{- if not (eq Values.ingress.hostname .name) }}
- host: {{ .name }
  ...
{{- end }}
```

andresmgot: It won't be a breaking change because:
 - If the user has `enabled: true` and the `hosts` section has some custom parameters, reusing those values would work in the same way.
 - If the user has `enabled: true` and the `hosts` were left by default, upgrading will introduce the new `hostname: kubeapps.local` which will result in the same result.

PR #1457, opened 2020-01-15
Enable atomic installations for Helm 3

andresmgot:
### Description of the change

When installing a chart, if the installation fails, it's useful to cleanup the release so the same name can be reused. This was already implemented for Helm 2 (#1163) but was missing for Helm 3.

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -72,6 +72,7 @@ func CreateRelease(actionConfig *action.Configuration, name, namespace, valueStr
 	cmd := action.NewInstall(actionConfig)
 	cmd.ReleaseName = name
 	cmd.Namespace = namespace
+	cmd.Atomic = true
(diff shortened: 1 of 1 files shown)


lindhe:
The `rollback` command has a `--cleanup-on-fail` flag, perhaps that's relevant too?

andresmgot:
For the upgrade/rollback I think the use case is a bit different. When installing something, if you make a mistake, you want to notice the error and retry (cleaning up the previous release). That's what this PR does so the user don't need to do it manually (or find an error when trying to reuse the name).

In the case of upgrading/rolling back, if any of those steps fails, before doing anything else, you may want to check the status. Let's say that the upgrade runs some data migration, running that twice without the user confirmation can be risky. Also we don't have the issue that the release name being reused.

Let me merge this PR as it is, and if you are thinking on some use case I am missing, we can tackle it in a different issue.

PR #1460, opened 2020-01-16
Remove RawIcon from chart versions

absoludity:
### Description of the change
Removes the `RawIcon` field from the json/bson responses.

I noticed while testing today that listing the (default) catalog is a 6M payload:
catalog-6MB

and looking at a specific wordpress chart is as 14M payload:
wordpress-versions-14MB

A large part of this is because we are now including the `RawIcon` data as part of the responses (for each chart, or for each version of a chart).

### Benefits

The request for the list of charts (default repos) goes down from 6M to 3M. The request for a chart's versions goes down from 14M to 4M.

### Other info
I think this issue (of the huge payloads) appears to have begun with the switch away from the monocular repo, as I notice there that the chart model is explicitly defined to *not* serialize certain fields to (and from) json, including the RawIcon and ChartVersions:

https://github.com/helm/monocular/blob/master/cmd/chartsvc/models/chart.go#L42

It appears that the postgres version is dependent on these being serialized - interested to know the background here.

I think a better solution than what I've tested here would be, as per the TODO, to store the raw icon not in the json column, but a separate column, so it's easy to pull out for serving, but is *not* included in the normal payload.

Thoughts?

### DIFF
--- a/cmd/assetsvc/handler.go
+++ b/cmd/assetsvc/handler.go
@@ -303,7 +303,7 @@ func newChartResponse(c *models.Chart) *apiResponse {
 	return &apiResponse{
 		Type:       "chart",
 		ID:         c.ID,
-		Attributes: chartAttributes(*c),
+		Attributes: blankRawIcon(chartAttributes(*c)),
 		Links:      selfLink{pathPrefix + "/charts/" + c.ID},
 		Relationships: relMap{
 			"latestChartVersion": rel{
@@ -312,6 +312,14 @@ func newChartResponse(c *models.Chart) *apiResponse {
 			},
 		},
 	}
+}
+
+// blankRawIcon returns the same chart data but with a blank raw icon field.
+// TODO(mnelson): The raw icon data should be stored in a separate postgresql column
+// rather than the json field so that this isn't necessary.
+func blankRawIcon(c models.Chart) models.Chart {
+	c.RawIcon = nil
+	return c
 }
 
 func newChartListResponse(charts []*models.Chart) apiListResponse {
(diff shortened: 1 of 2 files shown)


absoludity:
Note that this workaround improves the payloads slightly, but they are still megabytes with lots of redundant data. I'm keen to understand why the json annotations for the chart model went from `-` in monocular (ie. don't serialize at all, for raw_icon and chart_versions etc) to being included in the current Kubeapps model annotations?

andresmgot:
> Note that this workaround improves the payloads slightly, but they are still megabytes with lots of redundant data. I'm keen to understand why the json annotations for the chart model went from `-` in monocular (ie. don't serialize at all, for raw_icon and chart_versions etc) to being included in the current Kubeapps model annotations?

When using mongodb, it's using the `bson` annotation. This is useful when using the `mgo` library to get data. That's not possible with PostgreSQL so we are using the JSON annotation. We would need to manually remove it as you have done in this PR when serializing it as JSON.

I think that answer your question but I may be missing something. Let me know otherwise!

absoludity:
Thanks - it answers for `RawIcon` and `IconContentType`, but not `ChartVersions`, which doesn't have a bson annotation either in monocular:

https://github.com/helm/monocular/blob/master/cmd/chartsvc/models/chart.go#L42

andresmgot:
> Thanks - it answers for `RawIcon` and `IconContentType`, but not `ChartVersions`, which doesn't have a bson annotation either in monocular:
> 
> https://github.com/helm/monocular/blob/master/cmd/chartsvc/models/chart.go#L42

The reason is the same, since the annotation is not `bson: "-"` it's parsing it by default (not ignoring it).

Investigating this, I have found the cause of the issue. Let me open a PR.

PR #1462, opened 2020-01-16
Enable CI builds for Helm 3

andresmgot:
### Description of the change

Enable two more builds to run end-to-end tests using Helm 3 (both using MongoDB and PostgreSQL as database back-ends).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -297,7 +297,7 @@ jobs:
     working_directory: /go/src/github.com/kubeapps/kubeapps
     environment:
       GOPATH: /home/circleci/.go_workspace
-      IMAGES: "apprepository-controller tiller-proxy asset-syncer assetsvc kubeops"
+      IMAGES: "kubeops apprepository-controller tiller-proxy asset-syncer assetsvc"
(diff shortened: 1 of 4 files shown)


andresmgot:
Seem tests are a bit flaky. I'll investigate the cause further.

.circleci/config.yml
```diff
@@ -292,32 +311,50 @@ jobs:
     steps:
       - checkout
       - run: REPO_DOMAIN=kubeapps REPO_NAME=kubeapps ./script/create_release.sh ${CIRCLE_TAG}
-  local_e2e_tests_mongodb:
+  local_e2e_tests_mongodb_helm2:
+    machine: true
+    environment:
+      KUBEAPPS_DB: "mongodb"
+      HELM_VERSION: "v2.14.0"
```
absoludity: Just out of interest, why not 2.16?

andresmgot: We try to use the oldest version possible. Helm always maintains backward compatibility so if our chart works with an older version, it will work with newer ones (but not the other way around).

Issue #1461, opened 2020-01-16
[PostgreSQL] Improve database performance


andresmgot:
### Description:

Follow up of https://github.com/kubeapps/kubeapps/pull/1460

To improve performance when requesting charts we can refactor the `Chart` model so we store the `raw_icon` separately in a different column or table and as a byte blob. That way we would return it only when necessary.

We would need to also verify when, if in any case, we need to return the list of ChartVersions since that's a lot of data.

PR #1467, opened 2020-01-17
Remove chartVersions when listing versions

andresmgot:
### Description of the change

Follow up of https://github.com/kubeapps/kubeapps/pull/1460

Remove the ChartVersion list from the requests: `/charts/{repo}/{chartName}/versions` and `/charts/{repo}/{chartName}/versions/{version}` as the `chartsvc` does. This reduces the sice of the requests from several MBs to less than 500KB.

### DIFF
--- a/cmd/assetsvc/handler_test.go
+++ b/cmd/assetsvc/handler_test.go
@@ -134,7 +134,6 @@ func Test_newChartResponse(t *testing.T) {
 			assert.Equal(t, cResponse.ID, tt.chart.ID, "chart ID should be the same")
 			assert.Equal(t, cResponse.Relationships["latestChartVersion"].Data.(models.ChartVersion).Version, tt.chart.ChartVersions[0].Version, "latestChartVersion should match version at index 0")
 			assert.Equal(t, cResponse.Links.(selfLink).Self, pathPrefix+"/charts/"+tt.chart.ID, "self link should be the same")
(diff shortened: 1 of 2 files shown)


cmd/assetsvc/handler_test.go
```diff
@@ -134,7 +134,6 @@ func Test_newChartResponse(t *testing.T) {
 			assert.Equal(t, cResponse.ID, tt.chart.ID, "chart ID should be the same")
 			assert.Equal(t, cResponse.Relationships["latestChartVersion"].Data.(models.ChartVersion).Version, tt.chart.ChartVersions[0].Version, "latestChartVersion should match version at index 0")
 			assert.Equal(t, cResponse.Links.(selfLink).Self, pathPrefix+"/charts/"+tt.chart.ID, "self link should be the same")
-			assert.Equal(t, len(cResponse.Attributes.(models.Chart).ChartVersions), len(tt.chart.ChartVersions), "number of chart versions in the response should be the same")
 			// We don't send the raw icon down the wire.
 			assert.Nil(t, cResponse.Attributes.(models.Chart).RawIcon)
```
absoludity: Do you think we should update this to include verifying the chart versions is empty?

andresmgot: it's still not clear to me if the chart versions really need to be empty or if it's just our use case in which it's convenient that it's not included. Let's rethink about this in #1461

PR #1469, opened 2020-01-17
Check repo state instead of props when upgrading

andresmgot:
### Description of the change

After https://github.com/kubeapps/kubeapps/pull/1455/ I introduced a regression. When upgrading, if the repository can be auto-resolved, the properties of `AppUpgrade` are not being updated and the condition I introduced was always `true`, causing an infinity loop.

### DIFF
--- a/dashboard/src/components/AppUpgrade/AppUpgrade.tsx
+++ b/dashboard/src/components/AppUpgrade/AppUpgrade.tsx
@@ -93,7 +93,7 @@ class AppUpgrade extends React.Component<IAppUpgradeProps, IAppUpgradeState> {
         repo &&
         repo.metadata &&
         repo.metadata.name &&
-        (prevProps.app !== app || prevProps.repo !== repo)
+        (prevProps.app !== app || this.state.repo !== repo)
       ) {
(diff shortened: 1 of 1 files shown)


dashboard/src/components/AppUpgrade/AppUpgrade.tsx
```diff
@@ -93,7 +94,7 @@ class AppUpgrade extends React.Component<IAppUpgradeProps, IAppUpgradeState> {
         repo &&
         repo.metadata &&
         repo.metadata.name &&
-        (prevProps.app !== app || prevProps.repo !== repo)
+        (prevProps.app !== app || prevRepo !== repo)
```
absoludity: I had expected the test you added with 7b966848d to fail with this change, as it was added to verify that if the props.repo changes it fetches the data? Why is it not failing?

andresmgot: In that test, `state.repo` is empty at the beginning. Then component gets updated with `props.repo`.  `componentDidUpdate` gets executed and the state gets updated. Since `prevRepo !== repo` the function gets called.

PR #1473, opened 2020-01-21
Upgrade end-to-end test

andresmgot:
### Description of the change

This test:
 - Installs a chart using an old version.
 - Checks that the application notifies that there is a new version available.
 - Upgrades to that version.

It uses a new service account "edit" which has permissions to write resources only in its namespace (and don't have permissions to read/write apprepositories).

### Additional information

Requires the changes of https://github.com/kubeapps/kubeapps/pull/1472

### DIFF
--- a/integration/use-cases/upgrade.js
+++ b/integration/use-cases/upgrade.js
@@ -1,0 +1,40 @@
+test("Upgrades an application", async () => {
+  await page.goto(getUrl("/#/login"));
+
+  await expect(page).toFillForm("form", {
+    token: process.env.EDIT_TOKEN
+  });
+
+  await expect(page).toClick("button", { text: "Login" });
+
+  await expect(page).toClick("a", { text: "Catalog" });
+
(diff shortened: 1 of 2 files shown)


integration/use-cases/upgrade.js
```diff
@@ -0,0 +1,40 @@
...
+    "textContent"
+  );
+  const chartVersionValue = await chartVersionElementContent.jsonValue();
+  const latestChartVersion = chartVersionValue.split(" ")[0];
+
+  await expect(page).toSelect("#chartVersion", "7.3.2", { delay: 1000 });
+
+  await expect(page).toClick("button", { text: "Submit", delay: 1000 });
+
+  await expect(page).toMatch("Update Available", { timeout: 60000 });
```
absoludity: Is the 60s because we need to wait for the install to complete? Wonder if it's worth having a repo with a test chart that just creates a secret or something very quick. Not now of course, just a thought.

Issue #1366, opened 2019-12-11
[Question/Feature] Pluggable SCSS styling?


fabianschwarzfritz:
### Description:

Maybe more like a feature request/question than an actual issue. 

For our own installation, we have updated the styling of kubeapps by editing [the dashboard's index.scss](https://github.com/kubeapps/kubeapps/blob/master/dashboard/src/index.scss). We have manually edited the styling by overwriting the style. Oftentimes in the scss we had to use the `!important;` statement to enable correct style overwriting. 

Overwriting the styles worked for us, but the manual search for styles (mostly colors) we wanted to change was quite cumbersome).

It would be really nice if kubeapps would be "themeable" for everyone. Is there a central place where the all used colors and basic styling options are bundled and could be easily overwritten? 

Best Regards,
Fabian

andresmgot:
That's a good point. We have thought about this in the past but never started. Probably it's something we should have in mind while migrating the design system to Clarity (https://github.com/kubeapps/kubeapps/issues/1354)

fabianschwarzfritz:
Hi @andresmgot ,
super that there was some progress on the migration to the new design systems. Is it possible with Clarity to plug in custom styling/theming?
Thanks for any help/update you're able provide 😄 . I am curious about the status and the new capabilities!
Bests, Fabian

andresmgot:
Hi @fabianschwarzfritz, yes there has been some progress but it's still not possible to easily modifying the CSS style. We marked this feature for Kubeapps 2.1 so it will be part of the next feature release (once we release a final 2.0, which is almost out :) ).

fabianschwarzfritz:
Hi @andresmgot,
thanks you very much for looking into this request and for adding the functionality 😄 👍 . 
Best Regards,
Fabian

Issue #1354, opened 2019-12-06
Update Kubeapps styles/classes to use clarity.design


absoludity:
### Description:

At some point in the near future, we need to update the css classes (and possibly components, if React components using clarity are available) to use https://clarity.design/

Identified pre-requisites to complete:
 - [x] Allow to conditionally load Clarity or HEx CSS
 - [x] Allow to conditionally load a component based on a feature flag or alike

Components to be migrated:
 - [x]  ApplicationStatus
 - [x]  AppList
 - [x]  AppUpgrade
 - [x]  AppView
 ~~- [ ]  BindingList~~
 ~~- [ ]  Card~~
 - [x]  Catalog
 - [x]  ChartIcon
 - [x]  ChartView
 - [x]  Config
 - [x]  ConfigLoader
 - [x]  ConfirmDialog
 - [x]  DeploymentForm
 - [x]  DeploymentFormBody
 ~~- [ ]  ErrorAlert~~
 - [ ]  ErrorBoundary
 - [x]  Footer
 - [x]  Header
 ~~- [ ]  Hint~~
 - [x]  InfoCard
 - [x]  Layout
 - [x]  LoadingSpinner
 - [x]  LoadingWrapper
 - [x]  LoginForm
 ~~- [ ]  NotFound~~
 - [x]  OperatorInstance
 - [x]  OperatorInstanceForm
 - [x]  OperatorInstanceFormBody
 - [x]  OperatorInstanceUpdateForm
 - [x]  OperatorList
 - [x]  OperatorNew
 - [x]  OperatorView
 - [x]  PageHeader
 - [X]  PrivateRoute
 ~~- [ ]  SchemaForm~~
 - [x]  SearchFilter
 - [x]  SelectRepoForm
 ~~- [ ]  ServiceClassList~~
 ~~- [ ]  ServiceClassView~~
 ~~- [ ]  ServiceInstanceList~~
 ~~- [ ]  ServiceInstanceView~~
 - [x]  Slider
 ~~- [ ]  TerminalModal~~
 - [x]  UpgradeForm

PR #1474, opened 2020-01-21
Add scroll bars to prevent content jumping

SimonAlling:
### Description of the change

It forces scrollbars to always appear, even if they're not necessary.

### Benefits

Without this addition, page content often jumps around when the user navigates between "pages" or when stuff loads dynamically. Forcing scrollbars at all times removes said annoying behavior.

### Additional information

I'm currently unable to run Kubeapps locally, but I think this should work as intended. Just let me know otherwise, and/or suggest a change.

### DIFF
--- a/dashboard/src/index.scss
+++ b/dashboard/src/index.scss
@@ -4,6 +4,8 @@
 body {
   margin: 0;
   padding: 0;
+  overflow-y: scroll;
+  overflow-x: scroll;
 }
 
 .container-fluid > .row > div {


andresmgot:
I don't think this behaves as you expect. For example, the AppList view:

Screenshot from 2020-01-21 16-51-00

It's weird to see the scroll bars when it's not necessary. In any case, some content still "jumps". This is because the size of the different containers is not fixed. If an application contains 6 services, the list of services will take more than an application with 2. IMO is also better to see the content without having to scroll in different sub-boxes.

SimonAlling:
> I don't think this behaves as you expect. For example, the AppList view:
> 
> Screenshot from 2020-01-21 16-51-00
> 
> It's weird to see the scroll bars when it's not necessary. In any case, some content still "jumps". This is because the size of the different containers is not fixed. If an application contains 6 services, the list of services will take more than an application with 2. IMO is also better to see the content without having to scroll in different sub-boxes.

Seems to behave as I expected from your screenshot. Not sure what you mean by the last sentence, but I can agree that the horizontal scrollbar looks weird. I suggested forcing it to be there because I ran into a situation where it would appear and disappear a large number of times while some content was being loaded, causing an "Are you sure you want to delete this?" confirmation dialog to jump up and down annoyingly.

Anyway, we should definitely have the _vertical_ scrollbar. We simply can't get away from the fact that some pages fit within the screen height (not requiring a scrollbar) while others don't (requiring one). When navigating between "long" and "short" pages, the user experience is interrupted by the header etc jumping horizontally. This doesn't feel right and should be avoided at all costs (and a permanent vertical scrollbar isn't a very large sacrifice anyway).

andresmgot:
>  causing an "Are you sure you want to delete this?" confirmation dialog to jump up and down annoyingly.

I agree that shouldn't happen but to avoid that, confirmation modals are rendered in a fixed place so AFAIK it's not happening, let me know if it's happening for you in some view and we can fix it.

> Anyway, we should definitely have the _vertical_ scrollbar. We simply can't get away from the fact that some pages fit within the screen height (not requiring a scrollbar) while others don't (requiring one).

I don't see the problem with that? I mean, it's a normal behavior. For example, in Github the Notifications page doesn't require a scrollbar because it fits in my window while many other pages don't and we have the scrollbar.

> When navigating between "long" and "short" pages, the user experience is interrupted by the header etc jumping horizontally. This doesn't feel right and should be avoided at all costs (and a permanent vertical scrollbar isn't a very large sacrifice anyway).

That's probably a different problem, since we load the different parts of the application asynchronously, some content parts are loaded sooner than others causing the loaded content to be replaced and "jump". We should minimize that either reducing the loading time and maybe rendering resources in groups (if they are not useful standalone).

SimonAlling:
> I don't see the problem with that? I mean, it's a normal behavior. For example, in Github the Notifications page doesn't require a scrollbar because it fits in my window while many other pages don't and we have the scrollbar.

If "normal" is taken to mean "common", then yes it is, but that doesn't mean it's good. GitHub has the same problem, but not as obvious because their header is left-aligned, so it's not pushed to the left by the vertical scrollbar (except for the notification bell etc on the right-hand side).

> That's probably a different problem, since we load the different parts of the application asynchronously, some content parts are loaded sooner than others causing the loaded content to be replaced and "jump". We should minimize that either reducing the loading time and maybe rendering resources in groups (if they are not useful standalone).

I don't have access to Kubeapps, so I can't try anything at the moment, but I don't understand this point. Content jumping, as I call it, can occur both when actually navigating between pages (less annoying) and when content loads dynamically into an already loaded page (more annoying). The former issue is impossible to avoid, and the latter practically so, without `overflow-y: scroll`. Furthermore, they are both immediately eliminated by that simple CSS rule, whose only downside – a permanently visible vertical scrollbar – is minor.

andresmgot:
IMO we shouldn’t force the scroll bar to appear in all the views. A scroll bar is a pattern that indicates there’s more content in the page and it should be used only for that.

Moreover, scroll bars depend on the browsers and the OS. For example, MacOS hides the scroll bar until you start scrolling. Then, it’s displayed as an hover element, so it doesn’t affect to the page. However, it has the counterpart of my initial statement. You can guess there’s more content but you don’t know it until you start scrolling.

> Content jumping, as I call it, can occur both when actually navigating between pages (less annoying) and when content loads dynamically into an already loaded page (more annoying). 

We could reduce the effect of the second preloading some content (or assuming that some pages will  have enough content to have a scroll bar) but we would only be changing the time in which the header is moving those few pixels.

>  GitHub has the same problem, but not as obvious because their header is left-aligned, so it's not pushed to the left by the vertical scrollbar (except for the notification bell etc on the right-hand side).

That's also a considerable approach, we could move the header to the left and right side so we don't face the issue.

SimonAlling:
> IMO we shouldn’t force the scroll bar to appear in all the views. A scroll bar is a pattern that indicates there’s more content in the page and it should be used only for that.

A valid argument, but scrollbars visually indicate when they're not needed, so unnecessary scrollbars are not that confusing after all. I think content not staying in place (across page loads or otherwise) is much more annoying.
 
> Moreover, scroll bars depend on the browsers and the OS. For example, MacOS hides the scroll bar until you start scrolling. Then, it’s displayed as an hover element, so it doesn’t affect to the page. However, it has the counterpart of my initial statement. You can guess there’s more content but you don’t know it until you start scrolling.

So it seems like Apple agree with me. :slightly_smiling_face: 

> We could reduce the effect of the second preloading some content (or assuming that some pages will have enough content to have a scroll bar) but we would only be changing the time in which the header is moving those few pixels.
> 
> That's also a considerable approach, we could move the header to the left and right side so we don't face the issue.

We still will. The header isn't the only part affected, merely the most obvious one.

Also, I think the improvements you're suggesting are _far_ more difficult to implement than `overflow-y: scroll;`, and still wouldn't really cut it.

andresmgot:
> A valid argument, but scrollbars visually indicate when they're not needed, so unnecessary scrollbars are not that confusing after all. I think content not staying in place (across page loads or otherwise) is much more annoying.

Sorry, I disagree with that. TBH I never noticed the issue until you opened the issue. Having a scroll bar (with minimum scrolling or no scrolling at all) doesn't seem the right solution. Do you have any example of someone doing that?

absoludity:
My 2c: I think we'll find many arguments out there going both ways (even a [19 year old Mozilla issue discussing this back-and-forth](https://bugzilla.mozilla.org/show_bug.cgi?id=76197) ).

Ultimately, I think #1366 will enable Kubeapps installations to use whatever style they want, and regarding the default,  let's get wider design input when #1354 is done? I don't know that we'll resolve this here.

SimonAlling:
> Sorry, I disagree with that. TBH I never noticed the issue until you opened the issue.

You may just be used to Kubeapps having this glitch. I'm sure you would've noticed immediately had we gone from permanent scrollbar (and no jumping) to automatic scrollbar (and jumping).

> Having a scroll bar (with minimum scrolling or no scrolling at all) doesn't seem the right solution. Do you have any example of someone doing that?

Yes: [Facebook](https://facebook.com), [YouTube](https://youtube.com) and [Twitter](https://twitter.com) have `body { overflow-y: scroll; }`.

SimonAlling:
Maybe it's time to revisit this one, not least because #1354 is now closed.

As previously mentioned, `body { overflow-y: scroll; }` is at least fairly common among popular sites – unsurprisingly, since it does improve UX. It's also, AFAIK, the only practically feasible way to prevent content from annoyingly jumping around (due to asynchronous page loading as well as different pages having different amounts of contents).

andresmgot:
Hi @SimonAlling, indeed now is a good time to revisit this. Have you had a change to try the new UI? With the new UI, the scrollbar don't cover the header so you won't see content jumping like you did with the previous UI (at least in the header):

Screenshot from 2020-10-13 15-19-38

Also, note that we are now following the Clarity UI design system, based on similar apps like Harbor that also uses this design system, the scrollbar is not rendered unless is necessary:

Screenshot from 2020-10-13 15-22-28

Here you can see the application layout that Clarity suggests (with no mention to scrollbars unfortunately):

https://clarity.design/documentation/app-layout

Let me close this PR since it doesn't apply to the current UI (and we need to align Kubeapps with other applications in the ecosystem).

PR #1475, opened 2020-01-22
Update documentation for Kubeapps 1.8

andresmgot:
### Description of the change

Update several files to reflect the latest status of Kubeapps (and Helm). This PR is better merged after the new release (1.8.0) is available for users.

PR #1477, opened 2020-01-23
Move Helm 3 feature flag and update comment

SimonAlling:
### Description of the change

It moves the `useHelm3` flag to the top and updates its comment to reflect recent progress with respect to Helm 3 support.

### Benefits

* It's made clear that `useHelm3` can now be enabled.

### Possible drawbacks

* I don't know if we want this yet. Do we?
* Should the comment describe Helm 3 support as "experimental" or similar?

### DIFF
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -8,6 +8,9 @@
 #     - myRegistryKeySecretName
 #   storageClass: myStorageClass
 
+## Set this flag to true to use the Helm 3 backend; false to use Helm 2.
+useHelm3: false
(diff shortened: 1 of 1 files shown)


andresmgot:
Thanks for the update @SimonAlling! Yes, we want to do so, yesterday I opened a PR adapting a bunch of documentation including this. Sorry for not pinging you before, can you take a look? It's here: https://github.com/kubeapps/kubeapps/pull/1475

SimonAlling:
Oh, I missed that one. Obviously makes this PR void.

PR #1480, opened 2020-01-24
Fix upgrade modifications

### LINKED ISSUE #1479: Upgrading chart overwrites custom values
### Description:

Upgrading a chart does not merge in existing values, instead overwriting chart with the values from the upgrade's `values.yaml` only.

Reported by @davidkarlsen in the [#kubeapps channel](https://kubernetes.slack.com/archives/C9D3TSUG4/p1579819532001800).

1. Deploy Wordpress 8.1.0 explicitly setting the `wordpressUsername` value to `custom`:
wordpress-8 1 0-custom
2. Once deployed, click on the immediately available upgrade 

Note that the value for `wordpressUsername` is initially displayed as `custom` but as soon as you switch tabs and come back, or make any other change, it is set back to `user` which overwrites the value when deployed.
wordpress-upgrade-user-back-to-default

`username` should still be `custom` and the diff tab should show the differences of my customisations with the default values.yaml.

David reported this with helm 2.x (and no mongodb), but due to my dev setup, I reproduced it with helm 3 and postgresql.
- Output of `helm version`:

```
version.BuildInfo{Version:"v3.0.1", GitCommit:"7c22ef9ce89e0ebeb7125ba2ebf7d421f3e82ffa", GitTreeState:"clean", GoVersion:"go1.13.4"}
```

```
$ helm list --all-namespaces
dex             dex             1               2020-01-23 22:49:46.981171203 +0000 UTC deployed        dex-2.4.0       2.19.0
kubeapps        kubeapps        1               2020-01-23 22:49:51.46490202 +0000 UTC  deployed        kubeapps-3.3.0  DEVEL
ldap            ldap            1               2020-01-23 22:49:48.576518215 +0000 UTC deployed        openldap-1.2.3  2.4.48
wordpress       default         1               2020-01-23 22:53:15.356890779 +0000 UTC deployed        wordpress-8.1.0 5.3.2

```

```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.3", GitCommit:"5e53fd6bc17c0dec8434817e69b04a25d8ae0ff0", GitTreeState:"clean", BuildDate:"2019-06-06T01:44:30Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.0", GitCommit:"70132b0f130acc0bed193d9ba59dd186f0e634cf", GitTreeState:"clean", BuildDate:"2020-01-14T00:09:19Z", GoVersion:"go1.13.4", Compiler:"gc", Platform:"linux/amd64"}

```

andresmgot:
### Description of the change

Fixes #1479 

Regression added in #1472

We were not taking into account the auto-resolved repository in order to populate the property `repoName`.

Since a `Container` is difficult to test with unit-tests I have covered this fix with the end-to-end test for upgrading.

### DIFF
--- a/dashboard/src/components/AppUpgrade/AppUpgrade.tsx
+++ b/dashboard/src/components/AppUpgrade/AppUpgrade.tsx
@@ -14,7 +14,6 @@ interface IAppUpgradeProps {
   appsError: Error | undefined;
   namespace: string;
   releaseName: string;
-  version: string;
   repoName: string;
   selected: IChartState["selected"];
   deployed: IChartState["deployed"];
--- a/dashboard/src/containers/AppUpgradeContainer/AppUpgradeContainer.test.tsx
+++ b/dashboard/src/containers/AppUpgradeContainer/AppUpgradeContainer.test.tsx
@@ -1,0 +1,50 @@
+import { shallow } from "enzyme";
+import * as React from "react";
+import configureMockStore from "redux-mock-store";
+import thunk from "redux-thunk";
+import AppUpgrade from "../../components/AppUpgrade";
+import Upgrade from "./AppUpgradeContainer";
+
+const mockStore = configureMockStore([thunk]);
+
+const emptyLocation = {
+  hash: "",
+  pathname: "",
+  search: "",
+};
+
+const makeStore = (apps: any, repos: any) => {
+  return mockStore({
+    apps,
+    repos,
+    router: { location: emptyLocation },
+    charts: { selected: {} },
+    config: {},
+  });
+};
+
+describe("LoginFormContainer props", () => {
+  it("repoName is empty if no apps nor repos are available", () => {
+    const store = makeStore({}, { errors: {}, repo: {} });
+    const match = { params: { namespace: "default", releaseName: "foo" } };
+    const wrapper = shallow(<Upgrade store={store} match={match} />);
+    expect(wrapper.find(AppUpgrade).prop("repoName")).toBe(undefined);
+  });
+
+  it("repoName is set using the selected repo", () => {
+    const store = makeStore({}, { errors: {}, repo: { metadata: { name: "stable" } } });
+    const match = { params: { namespace: "default", releaseName: "foo" } };
+    const wrapper = shallow(<Upgrade store={store} match={match} />);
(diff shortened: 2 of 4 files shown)


absoludity:
> Since a `Container` is difficult to test with unit-tests I have covered this fix with the end-to-end test for upgrading.

I don't see an issue here since what you're testing doesn't involve any network, but we need to be careful. Not sure why a `Container` is difficult to unit-test? In particular, here we should be able to do something very similar to https://github.com/kubeapps/kubeapps/blob/master/dashboard/src/containers/HeaderContainer/HeaderContainer.test.tsx#L27 (ie. create a state store with the initial state, verify that the props passed to the component include the correct repoName depending on that state).

andresmgot:
> I don't see an issue here since what you're testing doesn't involve any network, but we need to be careful using e2e tests for things which should be unit tested (end up with more complicated e2e tests, slower e2e tests, etc., if the pattern is followed).

I agree. In this case I added that to the e2e test because it involved a user-facing functionality. I wanted to test that the values changes are preserved when upgrading which is very valuable. It's also not adding more time to the test (in this case).

> Not sure why a `Container` is difficult to unit-test? In particular, here we should be able to do something very similar to https://github.com/kubeapps/kubeapps/blob/master/dashboard/src/containers/HeaderContainer/HeaderContainer.test.tsx#L27 (ie. create a state store with the initial state, verify that the props passed to the component include the correct repoName depending on that state).

Well, yes, it can be tested but we are setting up an artificial environment and we are setting the expected values to get the expected result (which is only to forward a boolean). In this case I think it's more useful to have an e2e test but I am fine adding the unit test as well.

absoludity:
> Well, yes, it can be tested but we are setting up an artificial environment and we are setting the expected values to get the expected result (which is only to forward a boolean).

Yep, a unit test ensuring that the container converts the redux state into the correct args... it is testing exactly the mapStateToProps. It can seem a bit pointless when it's such a small change, but it also sets up the next person to be able to add to the test, rather than adding more to the e2e test.

> In this case I think it's more useful to have an e2e test but I am fine adding the unit test as well.

Thanks :)

PR #1482, opened 2020-01-24
Request resources asynchronously when the socket connection fails

### LINKED ISSUE #1352: Web Socket connections don't upgrade on GKE
### Description:
When running locally with dex/oauth2-proxy, the web socket connections which are used to track changes during an install work perfectly fine, upgrading as expected so the status can be updated:
websocket-upgraded-locally

But on a more production-like setup on GKE, the websocket connection request is not being upgraded, and as a result, the status of an install is not updated:

websocket-screenshot

1. Install Kubeapps on a GKE cluster, using the auth-proxy with an accounts.google.com provider
2. After logging in, install an application and wait for the app to be Ready

App never updates to ready without refreshing the page manually.

App updates to show Ready (or more specifically, the websocket connection requests are upgraded as expected).

- Output of `helm version`:

```
$ helm version
Client: &version.Version{SemVer:"v2.16.1", GitCommit:"bbdfe5e7803a12bbdf97e94cd847859890cf4050", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.16.1", GitCommit:"bbdfe5e7803a12bbdf97e94cd847859890cf4050", GitTreeState:"clean"}

```

```
$ helm --tiller-namespace tiller list kubeapps
NAME            REVISION        UPDATED                         STATUS          CHART           APP VERSION     NAMESPACE
kubeapps        2               Fri Nov 22 15:17:24 2019        DEPLOYED        kubeapps-3.1.0  v1.7.1          kubeapps 

```

```
$ k version
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.3", GitCommit:"5e53fd6bc17c0dec8434817e69b04a25d8ae0ff0", GitTreeState:"clean", BuildDate:"2019-06-06T01:44:30Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"14+", GitVersion:"v1.14.8-gke.17", GitCommit:"188432a69210ca32cafded81b4dd1c063720cac0", GitTreeState:"clean", BuildDate:"2019-11-13T20:47:11Z", GoVersion:"go1.12.11b4", Compiler:"gc", Platform:"linux/amd64"}

```

andresmgot:
### Description of the change

This PR creates an `onError` handler that can be called if the socket connection throws an error. This triggers an interval to re-request resources every 5s. This interval can be closed the same way than the socket calling `closeTimer`.

I have tested this removing the Nginx configuration that allows the socket connection, in the GIF below, you can see how the application gets updated even without the socket. When leaving the view, the timer gets cleared so we don't re-request more those resources.

final_5e2b0e93faec2c0016803a64_5

### Applicable issues

  - fixes #1352

### DIFF
--- a/dashboard/src/actions/kube.tsx
+++ b/dashboard/src/actions/kube.tsx
@@ -41,7 +41,7 @@ export type KubeAction = ActionType<typeof allActions[number]>;
 
 export function getResource(
   ref: ResourceRef,
-  shouldDispatchRequest?: boolean,
+  polling?: boolean,
 ): ThunkAction<Promise<void>, IStoreState, null, KubeAction> {
   return async (dispatch, getState) => {
     const key = ref.getResourceURL();
@@ -58,7 +58,7 @@ export function getResource(
 
     // If it's not the first request, we can skip the request REDUX event
     // to avoid the loading animation
-    if (shouldDispatchRequest !== false) {
+    if (!polling) {
       dispatch(requestResource(key));
     }
     try {
@@ -90,7 +90,7 @@ export function getAndWatchResource(
             // If the Socket fails, create an interval to re-request the resource
             // every 5 seconds. This interval needs to be closed calling closeTimer
             timer = setInterval(async () => {
-              dispatch(getResource(ref, false));
+              dispatch(getResource(ref, true));
             }, 5000);
           },
           closeTimer: () => {
--- a/dashboard/src/actions/kube.tsx
+++ b/dashboard/src/actions/kube.tsx
@@ -18,7 +18,11 @@ export const receiveResourceError = createAction("RECEIVE_RESOURCE_ERROR", resol
 // Takes a ResourceRef to open a WebSocket for and a handler to process messages
 // from the socket.
 export const openWatchResource = createAction("OPEN_WATCH_RESOURCE", resolve => {
-  return (ref: ResourceRef, handler: (e: MessageEvent) => void) => resolve({ ref, handler });
+  return (
+    ref: ResourceRef,
+    handler: (e: MessageEvent) => void,
+    onError: { closeTimer: () => void; onErrorHandler: () => void },
+  ) => resolve({ ref, handler, onError });
 });
 
 export const closeWatchResource = createAction("CLOSE_WATCH_RESOURCE", resolve => {
@@ -37,6 +41,7 @@ export type KubeAction = ActionType<typeof allActions[number]>;
 
 export function getResource(
   ref: ResourceRef,
+  shouldDispatchRequest?: boolean,
 ): ThunkAction<Promise<void>, IStoreState, null, KubeAction> {
   return async (dispatch, getState) => {
     const key = ref.getResourceURL();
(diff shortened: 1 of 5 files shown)


dashboard/src/actions/kube.tsx
```diff
@@ -51,7 +56,11 @@ export function getResource(
       return;
     }
 
-    dispatch(requestResource(key));
+    // If it's not the first request, we can skip the request REDUX event
+    // to avoid the loading animation
+    if (shouldDispatchRequest !== false) {
```
absoludity: Why not `if (shouldDispatchRequest) {` ? Ah, perhaps existing call-sites haven't been updated so get a default of `undefined` which isn't true? Not sure, but wonder if `polling` would be a better arg, which is fine defaulting to false/undefined, so here can be `if(!polling) {...}`. Anyway, fine either way, just found `shouldDispatchRequest` a little confusing at first because it implies it's talking about whether the actual request (to the server) will be dispatched, but it's not, it's just the redux state.

andresmgot: > Why not `if (shouldDispatchRequest) {` ? Ah, perhaps existing call-sites haven't been updated so get a default of `undefined` which isn't true?

yes, it's because of that

> Not sure, but wonder if polling would be a better arg, which is fine defaulting to false/undefined, so here can be if(!polling) {...}. Anyway, fine either way, just found shouldDispatchRequest a little confusing at first because it implies it's talking about whether the actual request (to the server) will be dispatched, but it's not, it's just the redux state.

I think I don't understand what `polling` would mean in this context? I am okay changing the var name though.

dashboard/src/actions/kube.tsx
```diff
@@ -66,13 +75,29 @@ export function getAndWatchResource(
...
+      openWatchResource(
+        ref,
+        (e: MessageEvent) => {
+          const msg = JSON.parse(e.data);
+          const resource: IResource = msg.object;
+          const key = ref.getResourceURL();
+          dispatch(receiveResource({ key, resource }));
+        },
+        {
+          onErrorHandler: () => {
```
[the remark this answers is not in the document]
andresmgot: There are several reasons why the socket may fail: a network error, a forbidden error if the user doesn't have permissions to "watch" a resource, a timeout after sometime... In all of those cases it's fine to jump to the fallback IMO, that's why I am not filtering the error. Are you thinking on a specific case?

dashboard/src/actions/kube.tsx
absoludity: It wasn't so much about changing the name as opposed to choosing something such that the default value could be undefined/false, so that you could just do `if (!polling) {...}` (and perhaps be a little clearer). But totally fine as is, imo.

dashboard/src/actions/kube.tsx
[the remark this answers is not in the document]
absoludity: Nope, I was just wondering if it was intentional that it was a catch all (ie. always drop to the fallback).

PR #1483, opened 2020-01-27
Creates a namespace when typing a new one in the namespace selector

### LINKED ISSUE #1364: Newly creates namespace gone after namespace switch
### Description:

When installing a new chart into a new namespace, the namespace disappears after a namespace switch. 

1. Open the kubeapps dashbaord and select the "default" namespace
1. Type in a new non-existing namespace "new-1" via the dropbownbox's text field. 
1. Go to the catalog, select an aribrary chart and deploy it. The new chart gets deployed into the new namespace "new-1". The installation is successful and I can see it on the cluster via `kubectl` and in the Applications tab in kubeapps. The app is deployed into the new namespace "new-1"
1. Select the "default" namespace in the dropdown in the UI. 
1. Try to select the new namespace "new-1" in the dropdown with the mouse (without typing something in). The namespace does not appear there and seems to be gone. 
1.  Workaround 1: It is only possible to get back to the namespace via the ui, by typing in "new-1" in the dropdown box.  Then the namespace can be selected again.
1. Workaround 2: force reloading the page in the browser. Then the namespace appears in the list and can be selected with the mouse. 

Successfully created namespace via kubeapps UI is not selectable.

To see the namespace in the dropbown list box.

Issue happens independent of which chart is installed. 
Workaround is force-reloading the webbrowser.

 - kubeapps 1.7.1
 - 
- Output of `helm version`:

```
/usr/local/Cellar/helm/2.14.3/bin/helm version
Client: &version.Version{SemVer:"v2.14.3", GitCommit:"0e7f3b6637f7af8fcfddb3d2941fcc7cbebb0085", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.14.3", GitCommit:"0e7f3b6637f7af8fcfddb3d2941fcc7cbebb0085", GitTreeState:"clean"}
```

*Remark: this helm client 2.14.3 was used for the installation*

```
/usr/local/Cellar/helm/2.14.3/bin/helm list
NAME       	REVISION	UPDATED                 	STATUS  	CHART        	APP VERSION	NAMESPACE
naughty-fly	1       	Wed Dec 11 10:13:43 2019	DEPLOYED	mongodb-7.6.0	4.0.13     	new-1
```

```
k version
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T23:42:50Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.3", GitCommit:"b3cbbae08ec52a7fc73d334838e18d17e8512749", GitTreeState:"clean", BuildDate:"2019-11-13T11:13:49Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"linux/amd64"}
```

andresmgot:
### Description of the change

Since Helm 3, the namespace is not being created when deploying an application in a new namespace. This PR adapts kubeapps so it creates the namespace when submitting it from the namespace selector so the installation doesn't fail.

It also re-request the list of namespaces after creating one so this fixes #1364

If a user doesn't have permissions to create a namespace, instead that silently fail, it renders a warning icon with a tooltip text with the content of the error:

Screenshot from 2020-01-27 13-02-47

### Applicable issues

  - fixes #1364

### DIFF
--- a/dashboard/src/actions/namespace.test.tsx
+++ b/dashboard/src/actions/namespace.test.tsx
@@ -2,7 +2,14 @@ import configureMockStore from "redux-mock-store";
 import thunk from "redux-thunk";
 import { getType } from "typesafe-actions";
 import Namespace from "../shared/Namespace";
-import { errorNamespaces, fetchNamespaces, receiveNamespaces, setNamespace } from "./namespace";
+import {
+  createNamespace,
+  errorNamespaces,
+  fetchNamespaces,
+  postNamespace,
+  receiveNamespaces,
+  setNamespace,
+} from "./namespace";
 
 const mockStore = configureMockStore([thunk]);
 
@@ -72,6 +79,45 @@ describe("fetchNamespaces", () => {
     ];
 
     await store.dispatch(fetchNamespaces());
+
+    expect(store.getActions()).toEqual(expectedActions);
+  });
+});
+
+describe("createNamespace", () => {
+  it("dispatches the new namespace and re-fetch namespaces", async () => {
+    Namespace.create = jest.fn();
+    Namespace.list = jest.fn().mockImplementationOnce(() => {
+      return {
+        items: [{ metadata: { name: "overlook-hotel" } }, { metadata: { name: "room-217" } }],
+      };
+    });
+    const expectedActions = [
+      {
+        type: getType(postNamespace),
+        payload: "overlook-hotel",
+      },
+      {
+        type: getType(receiveNamespaces),
+        payload: ["overlook-hotel", "room-217"],
+      },
+    ];
+
+    await store.dispatch(createNamespace("overlook-hotel"));
+    expect(store.getActions()).toEqual(expectedActions);
+  });
+
+  it("dispatches errorNamespace if error creating a namespace", async () => {
+    const err = new Error("Bang!");
+    Namespace.create = jest.fn().mockImplementationOnce(() => Promise.reject(err));
+    const expectedActions = [
+      {
+        type: getType(errorNamespaces),
+        payload: { err, op: "create" },
+      },
+    ];
+
+    await store.dispatch(createNamespace("foo"));
 
     expect(store.getActions()).toEqual(expectedActions);
   });
--- a/dashboard/src/actions/namespace.ts
+++ b/dashboard/src/actions/namespace.ts
@@ -9,17 +9,27 @@ export const setNamespace = createAction("SET_NAMESPACE", resolve => {
   return (namespace: string) => resolve(namespace);
 });
 
+export const postNamespace = createAction("CREATE_NAMESPACE", resolve => {
+  return (namespace: string) => resolve(namespace);
+});
+
 export const receiveNamespaces = createAction("RECEIVE_NAMESPACES", resolve => {
   return (namespaces: string[]) => resolve(namespaces);
 });
 
 export const errorNamespaces = createAction("ERROR_NAMESPACES", resolve => {
-  return (err: Error, op: "list") => resolve({ err, op });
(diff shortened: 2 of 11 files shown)


absoludity:
I'm not convinced that we should interpret selecting a namespace by typing in the name into the select and hitting enter, as the user wanting to create that namespace. How can I check if that namespace exists (or has apps deployed) without creating it?

AIUI, the previous functionality was that helm2 created the namespace when it installed, not when the user filtered by a namespace. Since helm 3 does not do that (and this is an improvement - you can't inadvertently create a typo'd namespace, but rather have to explicitly create one via kubectl), I'm not sure we should necessarily change that?

Possibilities I can think of:
 * Not allowing users to create namespaces via Kubeapps (since Helm no longer does this, for good reason)
 * Explicitly having a "Create new" option in the namespace selector so that it is explicit,
 * Updating our install action so that the previous behavior is maintained (ie. create the namespace on install, though this would be less explicit again).

The second option seems explicit and maintains the same feature, to me. Let me know what you think.

fabianschwarzfritz:
Hi @absoludity and @andresmgot ,

we have decided to keep this behavior and implement the third option which @absoludity  listed. 
We have added the namespace creation in on the server side in the helm installation action. With this the same behavior like with Helm2 is implemented: The namespace is created only when it's not existing yet. and when the user actually installs some software wihtin. _The namespace is not created when there is a typo in the search field and the user hits enter too early!_

Might be that I am a bit biased with this opinion, as it was easier skill wise for us in the team to change the server side go code than the react frontend code. 

I personally think it's a bit confusing for the user to that a search field (used for filtering) can be also used to create namespaces (without installing any charts in it). At least to me, the changes in this Pull Request would be a bit confusing in the UI flow.

From that regard, maybe option 2 (to have an explicit button and action) would be really cool? 

Cheers,
Fabian

lindhe:
> Explicitly having a "Create new" option in the namespace selector so that it is explicit

+1 on that

andresmgot:
Replying to @fabianschwarzfritz:

> we have decided to keep this behavior and implement the third option which @absoludity listed.

I would not go against the Helm 3 behavior, they have their reasons to not create namespaces implicitly. See this issue for more context: https://github.com/helm/helm/issues/5753#issuecomment-502163585

@absoludity @lindhe:

> > Explicitly having a "Create new" option in the namespace selector so that it is explicit
> 
> +1 on that

That's what I did but instead of having an option, I changed the text in the selector when you start to type:

Screenshot from 2020-01-28 13-12-59

If that's not explicit enough I can open a modal or alike to introduce the new namespace and make it more explicit. 

TBH I think we need to refactor completely the namespace selector. Having it as a box in the header doesn't lead to the best user experience and we find issues from time to time but that's a different discussion.

lindhe:
Is there anything to indicate to the user that you _can_ start typing in the field? I think you've made it very clear that you can create a new namespace once you've started typing, but how would the user know that? Do you have a small `+` symbol or something in the drop-down?

andresmgot:
> Is there anything to indicate to the user that you _can_ start typing in the field? I think you've made it very clear that you can create a new namespace once you've started typing, but how would the user know that? Do you have a small `+` symbol or something in the drop-down?

Nothing apart than the blinking cursor when you click. I agree it would be a better UX to at least have that button. I will rework this.

andresmgot:
I am closing this PR and opening a new one since the approach now is a lot different

PR #1485, opened 2020-01-28
Standardize code blocks on docs

juan131:

### Description of the change

This PR updates standardize the markdown syntax for code blocks:

- Use `'bash'` syntax for scripts and commands.
- Use `'console'` syntax for terminal outputs.

### Benefits

- Consistency

### Possible drawbacks

None

### DIFF
--- a/docs/user/migrating-to-v1.0.0-alpha.5.md
+++ b/docs/user/migrating-to-v1.0.0-alpha.5.md
@@ -13,7 +13,7 @@ These are the steps you need to follow to upgrade Kubeapps to this version.
 
 Please follow the steps in [this guide](./securing-kubeapps.md) to install Tiller securely. Don't install the Kubeapps chart yet since it will fail because it will find resources that already exist. Once the new Tiller instance is ready you can migrate the existing releases using the utility command included in `kubeapps` 1.0.0-alpha.5:
 
-```
+```console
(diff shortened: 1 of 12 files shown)


docs/user/migrating-to-v1.0.0-alpha.5.md
```diff
@@ -39,7 +39,7 @@ foo 	1       	Mon Aug  6 12:10:07 2018	DEPLOYED	aerospike-0.1.7	default
 
 Now that we have backed up the releases we should delete existing Kubeapps resources. To do so execute:
 
-```
+```bash
 kubeapps down
```
absoludity: I'm assuming this is no longer relevant? (I've not seen a kubeapps binary since I've been on the project, but heard there was one originally). When do we delete obsolete docs... it may be that this is referred to in some release notes (or perhaps we referred to the tagged version which would be better). Do you think we can just delete this doc?

andresmgot: Yes, this doc is old enough to be deleted

juan131: As I see there are two options:
- A) We delete obsolete documentation (such as this)
- B) We create a folder and put the obsolete docs there

PR #1487, opened 2020-01-29
Add sleep before retrieving tokens

andresmgot:
### Description of the change

There is a race condition in which we try to retrieve the serviceaccount tokens before they are ready. For example:
https://circleci.com/gh/kubeapps/kubeapps/16102

This PR reorders the token retrieval and adds a small sleep to avoid that.


andresmgot:
_merging this to avoid issues in master_

PR #1488, opened 2020-01-29
Add support to 'textarea' render type on basic forms

### LINKED ISSUE #1331: Add support to "textarea" render type on basic forms
**Description**

According to [the documentation](https://github.com/kubeapps/kubeapps/blob/master/docs/developer/basic-form-support.md), `Slider` is the unique "render" type supported so far.

It'd be very useful to include a `textarea` "render" type so we can use it for string parameters used to add configuration blocks.

**Version of Helm, Kubeapps and Kubernetes**:

```console
$ helm list kubeapps
NAME    	REVISION	UPDATED                 	STATUS  	CHART         	APP VERSION	NAMESPACE
kubeapps	1       	Thu Nov 28 12:23:42 2019	DEPLOYED	kubeapps-3.1.0	v1.7.1     	kubeapps
```

juan131:

### Description of the change

This PR adds support for a new render type on basic forms: "textArea"

This new render type can be very useful to add configuration blocks (multiline strings) with a more user-friendly interface. Check the example below using changes at https://github.com/helm/charts/pull/20426:

Screenshot 2020-01-29 at 15 12 05

### Benefits

More possibilities for chart developers to create their **values.schemas.json** 

### Possible drawbacks

None

### Applicable issues

  - fixes #1331


dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/TextAreaParam.tsx
```diff
@@ -0,0 +1,38 @@
...
+export interface IStringParamProps {
+  id: string;
+  label: string;
+  param: IBasicFormParam;
+  handleBasicFormParamChange: (
+    p: IBasicFormParam,
+  ) => (e: React.FormEvent<HTMLTextAreaElement>) => void;
+}
+
+class TextAreaParam extends React.Component<IStringParamProps> {
```
andresmgot: Instead of creating a new Component I would reuse the existing `TextParam`. These two components are mostly the same, the only different is that `TextParam` uses `<input` and this one uses `<textarea`. We can save a lot of code if we reuse the `TextParam`, you just need to filter, for example using the `inputType` parameter there. Something like:

```
let input = <input ...
if (inputType === "textarea") {
 input = <textarea ...
}
    return (
...
            <div className="col-9 margin-t-small">
              {input}
...
```

PR #1489, opened 2020-01-29
Add an option to create a namespace

andresmgot:
### Description of the change

Follow up of #1483 

Add an item to the namespace selector to create a namespace. This opens a modal to introduce the new namespace:

Screencast-2020-01-29-173829

In the case of error, it's shown to the user:

Screenshot from 2020-01-29 17-43-34

### Applicable issues

This is not fixing #1364 yet because it only allows you to create a namespace but you are still able to select a non-existing namespace and try to create an application there (which returns a wrong error).


dashboard/src/components/Header/Header.tsx
```diff
@@ -176,9 +184,8 @@ class Header extends React.Component<IHeaderProps, IHeaderState> {
   private handleNamespaceChange = (ns: string) => {
     const { pathname, push, setNamespace } = this.props;
     const to = pathname.replace(/\/ns\/[^/]*/, `/ns/${ns}`);
-    if (to === pathname) {
-      setNamespace(ns);
-    } else {
+    setNamespace(ns);
```
absoludity: This change just means that `setNamespace` is also called when the path changes, rather than only when it didn't change? Why was it only calling `setNamespace` when it *hadn't* changed in the first place?

dashboard/src/components/Header/NamespaceSelector.test.tsx
```diff
@@ -63,12 +66,35 @@ it("renders the default namespace option if no namespaces provided", () => {
...
-      options: [
-        { label: "All Namespaces", value: "_all" },
-        { label: defaultProps.defaultNamespace, value: defaultProps.defaultNamespace },
-      ],
-    }),
+  expect(select.props()).toMatchObject({
+    options: [
+      { label: "All Namespaces", value: "_all" },
+      { label: defaultProps.defaultNamespace, value: defaultProps.defaultNamespace },
+      { label: "Create New", value: "new" },
```
absoludity: I'm wondering if we need to show the option to "Create New" namespace if we know the user cannot do this. One simple, but not perfect, way is that we could only show the option to create a new namespace when there are more than 2 options in the selector already. It's not perfect as it only indicates the user can read namespaces, but is trivial to implement (EDIT: Actually, forget that, as we can assign RBAC to write namespaces without read, I assume). Not sure if it's worth considering returning this info when getting namespaces, but either way, it'd be another PR. See what you think.

dashboard/src/components/Header/NamespaceSelector.tsx
```diff
@@ -44,11 +58,23 @@ class NamespaceSelector extends React.Component<INamespaceSelectorProps> {
...
+        />
       </div>
     );
   }
 
   private handleNamespaceChange = (value: any) => {
+    if (value.value === "new") {
+      this.setState({ modalIsOpen: true });
+      return;
+    }
```
absoludity: Out of interest, what happens if I create a new namespace named "new", and then try to select it? :P (depending on the answer, you might need a token which is not a valid namespace).

dashboard/src/components/Header/Header.tsx
andresmgot: Yes, the logic was: If we are in a route that doesn't contain the namespace (e.g. /catalog) just call `setNamespace`. In other case, it was relying that a change in the route would trigger the `setNamespace` action by someone else (I assume). But the second was never happening and `setNamespace` was not being called.

dashboard/src/components/Header/NamespaceSelector.test.tsx
andresmgot: IMO hiding things may lead to a worse UX. If I am an unprivileged user and I want to create a namespace, I get a clear error that I don't have enough permissions to do so. If we hide the option it's not clear if I cannot do it or if it's the tool the one that doesn't support it.

dashboard/src/components/Header/NamespaceSelector.tsx
andresmgot: _good catch!_
I always wondered why the All Namespaces choice used `_all` (it does because `_all` is an invalid namespace  name).

fabianschwarzfritz:
Very cool feature and very beneficial when KubeAps is use with Helm3. Thanks @andresmgot !

dashboard/src/components/Header/NamespaceSelector.test.tsx
absoludity: Yep, good point.

PR #1492, opened 2020-01-30
Fix flaky test

andresmgot:
### Description of the change

The end to end test to create a registry is quite flaky:

https://circleci.com/gh/kubeapps/kubeapps/16306
https://circleci.com/gh/kubeapps/kubeapps/16279

Checking the screenshot, the modal that renders the form is not being shown. My guess is that the click in the button is not behaving as expected in the browser so I have added a second click to ensure that the modal is opened. If this doesn't fix the issue I will investigate further.

### DIFF
--- a/integration/use-cases/create-registry.js
+++ b/integration/use-cases/create-registry.js
@@ -8,6 +8,10 @@ test("Creates a registry", async () => {
   await expect(page).toClick("button", { text: "Login" });
 
   await expect(page).toClick("button", { text: "Add App Repository" });
+  // Sometimes, the test fails because the modal doesn't get rendered
+  // Clicking again, just in case, to avoid issues
+  // https://circleci.com/gh/kubeapps/kubeapps/16306
+  await expect(page).toClick("button", { text: "Add App Repository" });
 
   await page.type("#kubeapps-repo-name", "my-repo");
 
--- a/integration/use-cases/create-registry.js
+++ b/integration/use-cases/create-registry.js
@@ -8,10 +8,8 @@ test("Creates a registry", async () => {
   await expect(page).toClick("button", { text: "Login" });
 
   await expect(page).toClick("button", { text: "Add App Repository" });
-  // Sometimes, the test fails because the modal doesn't get rendered
-  // Clicking again, just in case, to avoid issues
-  // https://circleci.com/gh/kubeapps/kubeapps/16306
-  await expect(page).toClick("button", { text: "Add App Repository" });
+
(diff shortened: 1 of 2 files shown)


integration/use-cases/create-registry.js
```diff
@@ -7,6 +7,10 @@ test("Creates a registry", async () => {
 
   await expect(page).toClick("button", { text: "Login" });
 
+  await expect(page).toClick("button", { text: "Add App Repository" });
+  // Sometimes, the test fails because the modal doesn't get rendered
+  // Clicking again, just in case, to avoid issues
+  // https://circleci.com/gh/kubeapps/kubeapps/16306
```
absoludity: If this is the case, would it not mean that when it *does* display after the first click, the second click will get rid of the modal? (at least, that's what happens if I do that manually).

I'm really not keen on fixes like this, as they tend to complicate things down the line. But looking at this scenario, isn't the more likely cause of the error simply that the DOM isn't yet updated to include the modal before the `page.type` below.  [According to the docs](https://github.com/puppeteer/puppeteer/blob/master/docs/api.md#pagetypeselector-text-options), doesn't appear to wait at all (at least it has no option to specify it, like the various `expect` functions do). I assume it  tries to add the first character straight away, which sometimes works, sometimes doesn't, depending on whether the browser has updated the DOM.

So, I think it should be as simple as first expecting the `#kubeapps-repo-name` element to be present first (which will wait by default for up to 30s), before calling `page.type`?

andresmgot: That's a good guess. I assumed that it was waiting but it's true that it's not using `expect`.

> If this is the case, would it not mean that when it _does_ display after the first click, the second click will get rid of the modal? (at least, that's what happens if I do that manually).

It's more of a double click. Based on the above, my guess is that doing a second click was fast enough to be done before the modal is render but slows the test enough so the `page.type` is ready to be executed.

script/e2e-test.sh
```diff
@@ -185,7 +185,7 @@ done
 
 # Run helm tests
 # Retry once if tests fail to avoid temporary issue
-if ! retry_while testHelm "1" "1"; then
+if ! retry_while testHelm "2" "1"; then
```
andresmgot: 1 retry is just one try :)

andresmgot:
Aaaand it didn't work:

https://circleci.com/gh/kubeapps/kubeapps/16370
```
  ● Creates a registry

    TimeoutError: Text not found "Install Repo"

      waiting for function failed: timeout 4000ms exceeded
```

PR #1494, opened 2020-01-31
Fix flaky test

andresmgot:
### Description of the change

Same as https://github.com/kubeapps/kubeapps/pull/1492 apparently that solution didn't work:

https://github.com/kubeapps/kubeapps/pull/1492#issuecomment-580754178

Again, trying with the double-click approach (if fails).

### DIFF
--- a/integration/use-cases/create-registry.js
+++ b/integration/use-cases/create-registry.js
@@ -9,7 +9,14 @@ test("Creates a registry", async () => {
 
   await expect(page).toClick("button", { text: "Add App Repository" });
 
-  await expect(page).toMatch("Install Repo");
+  try {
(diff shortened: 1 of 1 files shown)


andresmgot:
> Is it worth dumping the `console` to see if there are any errors?

The error is what I have in the comment, it's not able to find the button with that text. If you mean the errors in the browser console, that's a different thing (I would need to investigate how to retrieve that).

> Let me know if you're able to reproduce this situation locally, I'd like to dig :)

I am not able to reproduce it locally, the only thing we could do is to retry a job with SSH like https://circleci.com/gh/kubeapps/kubeapps/16370 until it fails. It happens more or less 1 out of 5 times but never got to reproduce it.

Issue #1495, opened 2020-02-03
Prepare AppRepository for being a per-namespace resource


absoludity:
Initial work to enable creating per-namespace AppRepositories, which is a pre-requisite for private registry support. See [design doc](https://docs.google.com/document/d/1YEeKC6nPLoq4oaxs9v8_UsmxrRfWxB6KCyqrh2-Q8x0/edit?ts=5e2adf87#) for more info and discussion.

Parent task: #1256 

Tasks:
- [x] Add ClusterRole for reading AppRepositories only
- [x] API: Update the Kubeapps API endpoint to save the AppRepository in the specific namespace (initially always kubeapps' namespace)
- [x] UI: Update URI for AppRepositories to include namespace
- [x] Chart: Add feature-flag / charm switch for perNamespaceAppRepositories, defaulting to false, so work can continue without affecting users.
- [x] UI: When feature-flag enabled, Display AppRepositories for current namespace or all namespaces

andresmgot:
Asking some questions here to avoid polluting the design doc:

>  API: Update the Kubeapps API endpoint to save the AppRepository in the specific namespace (initially always kubeapps' namespace)

The API currently creates the secrets related to the AppRepository. What do you plan to do with those? If the backend does that we'd need to extend the ClusterRole to read and write secrets in all namespaces (which can be too much). The other option is to move that functionality to the dashboard.

> UI: When feature-flag enabled, Display AppRepositories for current namespace (but to avoid initial confusion, include "Looking for site-wide app repositories? Select 'All Namespaces' in the namespace switcher above" or similar (and vice-versa)

Weren't we going to treat AppRepositories in the `kubeapps` namespace as global? If those keep being global, even if the flag is enabled, people will see the usual catalog.

absoludity:
> Asking some questions here to avoid polluting the design doc:

Comments on the design doc are probably easier to manage (as we can resolve them and keep things clear), but whatever you prefer.

> 
> > API: Update the Kubeapps API endpoint to save the AppRepository in the specific namespace (initially always kubeapps' namespace)
> 
> You are planning to use the user serviceaccount to create the AppRepositories?

[...]

PR #1498, opened 2020-02-04
1495 2 app repository cluster role

absoludity:
Follows #1497 . Updates the app repository controller to be granted a clusterrole and clusterrolebinding to read AppRepositories.

I've used an easily identifiable and descriptive name for the clusterrole (rather than the templated `kubeapps.apprepository.fullname`), while ensuring it is unique per Kubeapps installation (appending the namespace).

Ref #1495

### DIFF
--- a/chart/kubeapps/templates/apprepository-rbac.yaml
+++ b/chart/kubeapps/templates/apprepository-rbac.yaml
@@ -1,5 +1,5 @@
 {{- if .Values.rbac.create -}}
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
   name: {{ template "kubeapps.apprepository.fullname" . }}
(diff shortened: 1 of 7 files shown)


chart/kubeapps/templates/apprepository-rbac.yaml
```diff
@@ -32,6 +32,18 @@ rules:
       - jobs
     verbs:
       - create
+---
+# Kubeapps can read and watch its own AppRepository resources cluster-wide.
```
andresmgot: shouldn't we enable this behind a feature flag for the moment? that way we avoid "breaking" changes in the chart while the feature is being developed

absoludity: I can put it behind a flag, but didn't see much point given that this doesn't affect existing behaviour, just enables the features which will themselves be behind a feature flag to work properly. Let me update.

andresmgot: > given that this doesn't affect existing behaviour

well it affects in the sense that we are now installing a cluster wide resource (apart than the CRD)

absoludity: PTAL :)

PR #1502, opened 2020-02-05
AppRepositories list uses namespaced URI

absoludity:
Ref: #1495 
This diff adds the `featureFlags` hash to the config so the frontend can also use it, which it does to:

- Update the route for the apprepo list when the feature is on
- Update the menu link for the apprepo list when the feature is on.

### DIFF
--- a/dashboard/src/components/Header/Header.tsx
+++ b/dashboard/src/components/Header/Header.tsx
@@ -24,6 +24,7 @@ interface IHeaderProps {
   setNamespace: (ns: string) => void;
   createNamespace: (ns: string) => Promise<boolean>;
   getNamespace: (ns: string) => void;
+  featureFlags: { reposPerNamespace: boolean };
 }
 
 interface IHeaderState {
@@ -32,6 +33,10 @@ interface IHeaderState {
 }
 
 class Header extends React.Component<IHeaderProps, IHeaderState> {
+  public static defaultProps = {
+    featureFlags: { reposPerNamespace: false },
+  };
+
   // Defines the route
   private readonly links: IHeaderLinkProps[] = [
     {
@@ -83,6 +88,9 @@ class Header extends React.Component<IHeaderProps, IHeaderState> {
       this.state.configOpen ? "header__nav__submenu-open" : ""
     }`;
 
+    const reposPath = this.props.featureFlags.reposPerNamespace
+      ? `/config/ns/${namespace.current}/repos`
+      : "/config/repos";
     return (
       <section className="gradient-135-brand type-color-reverse type-color-reverse-anchor-reset">
         <div className="container">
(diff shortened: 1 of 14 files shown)


dashboard/src/containers/RoutesContainer/Routes.test.tsx
```diff
@@ -77,3 +78,51 @@ it("should render a redirect to the login page (when not authenticated)", () =>
...
+          {...emptyRouteComponentProps}
+          namespace={namespace}
+          authenticated={true}
+          featureFlags={{ reposPerNamespace: false }}
+        />
+      </StaticRouter>,
+    )
+      .dive()
+      .dive()
+      .dive();
```
andresmgot: what's the `dive()` for? I guess it's to be able to find the component below but didn't `mount(<StaticRouter ...`  worked for you?

dashboard/src/components/Header/Header.tsx
```diff
@@ -83,6 +88,9 @@ class Header extends React.Component<IHeaderProps, IHeaderState> {
       this.state.configOpen ? "header__nav__submenu-open" : ""
     }`;
 
+    const reposPath = this.props.featureFlags.reposPerNamespace
+      ? `/config/ns/${namespace.current}/repos`
+      : "/config/repos";
```
andresmgot: isn't it a bit easier if we just change the route to `/config/ns/{namespace}/repos` and ignore the namespace in the route in the RepoList component if the flag is not enabled?

absoludity: Yes, this is exactly what I did originally, but based on your review of 1498 (in particular, [this thread](https://github.com/kubeapps/kubeapps/pull/1498#pullrequestreview-352830996)) from which I assumed you were keen to ensure that we don't change current behaviour even if it's not visible to the use, I updated to ensure this change was behind the feature flag also. Maybe I read to much into the comment?

dashboard/src/containers/RoutesContainer/Routes.test.tsx
absoludity: Because `mount` renders the component and all children, it was also rendering the `PrivateRouteContainer` which requires a connected store. `shallow`, where possible, seems to be preferred for both this reason and the fact that you can focus your test on the component level unit-test, rather than forcing to test all the children too (ie. this test won't fail if `RepoListContainer` has an issue, only `RepoListContainer`'s unit tests will).

It does look a bit ugly with the three `dive()`'s though, to get to the `RepoListContainer` only.

dashboard/src/components/Header/Header.tsx
andresmgot: Yes, my comment there was because that was an usability issue (some without permissions to create ClusterRoles wouldn't be able to install the chart). In this case there is no usability issue, just a change in the route. In any case, I am fine with this change.

absoludity: Ah - I wouldn't have thought it likely that someone had perms to create CRDs but not a ClusterRole, but possible, sure. Anyway, worked out well, it was just a bit bigger than expected. Thanks!

Issue #1496, opened 2020-02-03
Sync per-namespace AppRepositories


absoludity:
Support for sync'ing per-namespace AppRepositories, which is a pre-requisite for private registry support. See [design doc](https://docs.google.com/document/d/1YEeKC6nPLoq4oaxs9v8_UsmxrRfWxB6KCyqrh2-Q8x0/edit?ts=5e2adf87#) for more info and discussion.

Parent task: #1256 . Follows (but can be done in parallel with) #1495 

- [x] API: Ensure the app repository credentials (not docker registry secrets) are saved in the specific namespace (using the users' auth)
- [x] API: Ensure the app repository credentials (not docker registry secrets) are copied to Kubeapps' own namespace (with the service account auth) using a canonical name (namespace.repo for eg.)
- [x] API: Ensure both are deleted when an AppRepository is deleted
- [x] chart-sync: With FF enabled, chart-sync should sync AppRepositories across all namespaces

andresmgot:
>  API: Ensure the app repository credentials (not docker registry secrets) are copied to Kubeapps' own namespace (with the service account auth) using a canonical name (namespace.repo for eg.)

This could be a RBAC issue. A user with permissions to access the `kubeapps` namespace will have access to the credentials of all the registries. Maybe it's better to give the apprepository-controller RBAC permissions to create jobs and cronjobs cluster-wide?

absoludity:
> This could be a RBAC issue. A user with permissions to access the `kubeapps` namespace will have access to the credentials of all the registries.

User's don't require permission to access the `kubeapps` namespace for this point - that's what I meant by `copied to Kubeapps' own namespace (with the service account auth)`. More details in "[AppRepository and Secrets per namespace](https://docs.google.com/document/d/1YEeKC6nPLoq4oaxs9v8_UsmxrRfWxB6KCyqrh2-Q8x0/edit?ts=5e2adf87#heading=h.396p9pyw1xwi)" , but with this, a user only requires the normal perms in their own namespace, and Kubeapps only requires the ability to read its own API resource (AppRepositories) in other namespaces.

> Maybe it's better to give the apprepository-controller RBAC permissions to create jobs and cronjobs cluster-wide?

[...]

PR #1509, opened 2020-02-11
Store a copy of the repo secret in Kubeapps' namespace

absoludity:
Ref: #1496 
Updates the apprepo handler so that a copy of an app repository credential is recorded in Kubeapps' own namespace.

### DIFF
--- a/pkg/apprepo/apprepos_handler.go
+++ b/pkg/apprepo/apprepos_handler.go
@@ -39,11 +39,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// coreClientsetInterface provides just the core client
-type coreClientsetInterface interface {
-	CoreV1() corev1typed.CoreV1Interface
-}
-
 // combinedClientsetInterface provides both the app repository clientset and the corev1 clientset.
 type combinedClientsetInterface interface {
 	KubeappsV1alpha1() v1alpha1typed.KubeappsV1alpha1Interface
@@ -70,7 +65,7 @@ type appRepositoriesHandler struct {
 	kubeappsNamespace string
 
 	// The Kubernetes client using the pod serviceaccount
-	svcKubeClient coreClientsetInterface
+	svcKubeClient kubernetes.Interface
 
 	// clientsetForConfig is a field on the struct only so it can be switched
 	// for a fake version when testing. NewAppRepositoryHandler sets it to the
--- a/pkg/apprepo/apprepos_handler.go
+++ b/pkg/apprepo/apprepos_handler.go
@@ -39,6 +39,11 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// coreClientsetInterface provides just the core client
+type coreClientsetInterface interface {
+	CoreV1() corev1typed.CoreV1Interface
+}
+
 // combinedClientsetInterface provides both the app repository clientset and the corev1 clientset.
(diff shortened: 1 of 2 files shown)


pkg/apprepo/apprepos_handler.go
```diff
@@ -65,7 +70,7 @@ type appRepositoriesHandler struct {
 	kubeappsNamespace string
 
 	// The Kubernetes client using the pod serviceaccount
-	svcKubeClient *kubernetes.Clientset
+	svcKubeClient coreClientsetInterface
```
andresmgot: rather than defining our own interface, isn't it better to use `kubernetes.Interface` instead?

pkg/apprepo/apprepos_handler.go
```diff
@@ -310,6 +341,12 @@ func secretNameForRepo(repoName string) string {
 	return fmt.Sprintf("apprepo-%s-secrets", repoName)
 }
 
+// kubeappsSecretNameForRepo returns a name suitable for recording a copy of
+// a per-namespace repository secret in the kubeapps namespace.
+func kubeappsSecretNameForRepo(repoName, namespace string) string {
```
andresmgot: isn't it better to reuse the method `secretNameForRepo`? To avoid confusion between similar methods.

pkg/apprepo/apprepos_handler_test.go
```diff
@@ -188,19 +196,45 @@ func TestAppRepositoryCreate(t *testing.T) {
...
+						_, err = handler.svcKubeClient.CoreV1().Secrets(tc.kubeappsNamespace).Get(kubeappsSecretName, metav1.GetOptions{})
+						if statusErr, ok := err.(*errors.StatusError); ok {
+							status := statusErr.ErrStatus
+							if got, want := status.Code, int32(404); got != want {
+								t.Errorf("got: %d, want: %d", got, want)
+							}
+						} else {
+							t.Errorf("Unable to convert err to StatusError: %+v", err)
+						}
+					}
```
andresmgot: this test body is becoming huge, I wonder if it would be better to start splitting the different checks in different tests

pkg/apprepo/apprepos_handler.go
absoludity: Yep, you're right. I'm used to defining smaller interfaces with just the functionality we need to be able to test, but in this case the test client already exists, so no need.

pkg/apprepo/apprepos_handler.go
absoludity: It is re-using `secretNameForRepo` in its implementation. I could have done this inline, which is perhaps what you mean? The only reason I've pulled it out to a function is because I'll need to generate the same secret name when deleting in a following PR, so rather than have the same inline code to create the kubeapps secret name, I've put it in a fn. Let me know if I missed something?

pkg/apprepo/apprepos_handler_test.go
absoludity: Yeah, I thought the same thing. I'll look at this in the following PRs.

pkg/apprepo/apprepos_handler.go
andresmgot: It's just a bit confusing to know (if I were developing here) which method to use. For me, it'd be better to extend `secretNameForRepo(name, namespace)` allowing to set the namespace empty and adding a comment saying that the namespace is only needed if the secret is meant for the `kubeapps` namespace.

PR #1511, opened 2020-02-11
Show meaningful error when creating a release

### LINKED ISSUE #1440: Clearly differentiate between internal and external application errors
### Description:

When installing/upgrading/deleting an application, the process may fail due to an error in the chart. When this happens, it's not very clear if the error is caused by kubeappss (an internal error) of it's a chart error:

Screenshot from 2020-01-13 11-31-47

We should be able to clearly identify chart errors and maybe point the user to the chart maintainers in that case. 

1.Deploy bitcoind in a k8s cluster +1.16

Related issue: https://github.com/kubeapps/kubeapps/issues/1431

cc/ @fabianschwarzfritz @AlpherJang

### LINKED ISSUE #1464: [Helm3] Incorrect error message when installing to namespace without access
### Description:

If I attempt to install into a namespace to which I don't have access to install (particularly, I can't even read secrets), kubeapps displays the error: "<release-name> already exists, try a different name", which is incorrect (and a bad experience if you keep trying :P).

The actual issue is shown below:
409-incorrect-error

1. Login as a normal user and choose a namespace to which you don't have access (ie. easiest to use a user who doesn't have access to read secrets in default)
2. Deploy a chart

ER: An error along the lines of "You do not have access to install in the namespace 'foo'".
Actual result: see above image.

Putting this as low priority because, realistically, we should not be presenting people with namespaces to which they don't have access, the underlying issue there is that we add `default` without knowing if the user has access. Once that is fixed this issue won't be so obvious, though still worth fixing.

### LINKED ISSUE #1508: Wrong error returned when the installation fails
### Description:

When the application fails to be created, it fails to delete that release (because there is no release). When this happens the error contains "not found" which is considered a 404 causing a wrong response:

Screenshot from 2020-02-10 17-27-37

This is the real message:

```
{"code":404,"message":"Release \"mellow-level\" failed: unable to build kubernetes objects from release manifest: error validating \"\": error validating data: ValidationError(Deployment.metadata): unknown field \"selector\" in io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta. Unable to delete failed release: uninstall: Release not loaded: mellow-level: release: not found"}
```

A similar issue happens when the user doesn't have permissions to get a resource (e.g. ClusterRole). The request fails but the error returned is not relevant:

Screenshot from 2020-02-10 17-32-21

The real error:

```
{"code":409,"message":"Release \"dynamic-weight\" failed: rendered manifests contain a resource that already exists. Unable to continue with install: could not get information about the resource: clusterroles.rbac.authorization.k8s.io \"dynamic-weight-prometheus-alertmanager\" is forbidden: User \"system:serviceaccount:kubeapps:kubeapps-edit\" cannot get resource \"clusterroles\" in API group \"rbac.authorization.k8s.io\" at the cluster scope. Unable to delete failed release: uninstall: Release not loaded: dynamic-weight: release: not found"}
```

1. Deploy the couchdb-operator chart (404)

or

1. Deploy the prometheus chart (409)

andresmgot:
### Description of the change

The goal of this PR is to show an error that really represents the installation status when creating a new release. It gather fixes for #1464, #1508 and #1440.

Now, when there is an error when creating a release, instead of assuming the error, it shows it to the user.

Apart from showing the error, we display a message that changes depending if the error is Kubeapps related (a 500 error) or something else to help people to look for support in the right place.

These are some examples:

Wrong manifest:
Screenshot from 2020-02-11 11-52-58

Masked forbidden error:
Screenshot from 2020-02-11 11-52-19

Invalid manifest:
Screenshot from 2020-02-11 11-47-15

Missing repository:
Screenshot from 2020-02-11 11-54-44

Only when it's a forbidden error, we keep doing some additional parsing:

Screenshot from 2020-02-11 15-54-44

### Applicable issues

  - fixes #1464 
  - fixes #1508
  - fixes #1440

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -3,6 +3,7 @@ package agent
 import (
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/kubeapps/kubeapps/pkg/chart/helm3to2"
 	"github.com/kubeapps/kubeapps/pkg/proxy"
@@ -85,7 +86,7 @@ func CreateRelease(actionConfig *action.Configuration, name, namespace, valueStr
 	if err != nil {
 		// Simulate the Atomic flag and delete the release if failed
 		errDelete := DeleteRelease(actionConfig, name, false)
-		if errDelete != nil {
+		if errDelete != nil && !strings.Contains(errDelete.Error(), "release: not found") {
 			return nil, fmt.Errorf("Release %q failed: %v. Unable to delete failed release: %v", name, err, errDelete)
 		}
 		return nil, fmt.Errorf("Release %q failed and has been uninstalled: %v", name, err)
--- a/dashboard/src/components/DeploymentForm/DeploymentForm.test.tsx
+++ b/dashboard/src/components/DeploymentForm/DeploymentForm.test.tsx
@@ -2,9 +2,14 @@ import { mount, shallow } from "enzyme";
 import * as Moniker from "moniker-native";
 import * as React from "react";
 
-import { IChartState, IChartVersion, UnprocessableEntity } from "../../shared/types";
+import {
+  ForbiddenError,
+  IChartState,
+  IChartVersion,
+  UnprocessableEntity,
+} from "../../shared/types";
 import DeploymentFormBody from "../DeploymentFormBody/DeploymentFormBody";
-import { UnexpectedErrorAlert } from "../ErrorAlert";
+import { PermissionsErrorAlert, UnexpectedErrorAlert } from "../ErrorAlert";
 import DeploymentForm from "./DeploymentForm";
 
 const releaseName = "my-release";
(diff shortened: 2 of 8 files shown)


pkg/agent/agent.go
```diff
@@ -85,7 +86,7 @@ func CreateRelease(actionConfig *action.Configuration, name, namespace, valueStr
 	if err != nil {
 		// Simulate the Atomic flag and delete the release if failed
 		errDelete := DeleteRelease(actionConfig, name, false)
-		if errDelete != nil {
+		if errDelete != nil && !strings.Contains(errDelete.Error(), "release: not found") {
```
absoludity: Could we not do
```
import "helm.sh/helm/v3/pkg/storage/driver"
...
if errDelete != nil && err != driver.ErrReleaseNotFound {
...
```
rather than using string comparisons?

andresmgot: We can't. The errro `ErrReleaseNotFound` is just `release: not found` but the error that `DeleteRelease` is wrapped with more details: `uninstall: Release not loaded: neat-shoe: release: not found`. If we compare both errors, they are not equal.

PR #1512, opened 2020-02-12
1496 2 delete repo creds

absoludity:
Ref: #1496 . Small prequal for adding the apprepository delete URI in the backend. URIs for AppRepositories in the backend should be unique to be resh-ish (consider a `DELETE`). This just updates the route for the Create handler to include the namespace in the URI, which means using the `mux.Vars()`. Had some trouble testing this until I found https://github.com/gorilla/mux/pull/322/files .

While there, I extracted the setup of routes so we can just re-use that in tiller-proxy and kubeops, and while doing that I couldn't see what benefit we were getting from using negroni, and didn't want to import it in the handler module, so removed it.

### DIFF
--- a/pkg/apprepo/apprepos_handler_test.go
+++ b/pkg/apprepo/apprepos_handler_test.go
@@ -24,6 +24,7 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/gorilla/mux"
 	authorizationv1 "k8s.io/api/authorization/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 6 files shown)


pkg/apprepo/apprepos_handler_test.go
```diff
@@ -435,9 +423,10 @@ func TestSecretForRequest(t *testing.T) {
 	}
 
 	testCases := []struct {
-		name    string
-		request appRepositoryRequestDetails
-		secret  *corev1.Secret
+		name             string
+		requestNamespace string
```
andresmgot: is this `requestNamespace` never used?

absoludity: True - it was before this change (as the namespace was set in `secretForRequest` but is no longer. Thanks, it'll be updated in the next PR.

PR #1513, opened 2020-02-12
Move HTTP functionality of apprepo to another package

andresmgot:
### Description of the change

Move HTTP related code out of `pkg/apprepo` so this package can be re-used in other packages (pkg/chart) later on. 

This is part of the refactoring I mention in https://github.com/kubeapps/kubeapps/pull/1507

### Additional information

I still need to add tests for `pkg/http-handler` (to check that it returns the expected response/code but I wanted to send something first to avoid conflicts.


pkg/apprepo/apprepos_handler.go
```diff
@@ -55,9 +54,9 @@ type combinedClientset struct {
 	*kubernetes.Clientset
 }
 
-// appRepositories handles http requests for operating on app repositories
+// AppRepositoriesHandler handles http requests for operating on app repositories
 // in Kubeapps, without exposing implementation details to 3rd party integrations.
-type appRepositoriesHandler struct {
+type AppRepositoriesHandler struct {
```
absoludity: Not sure if this, the conversion from private to public, was done before you created the interface below... I don't see it being accessed outside this module?

absoludity:
Landing to continue work in the handle. Thanks for pushing this so we can move without stacking branches :)

pkg/apprepo/apprepos_handler.go
andresmgot: Apart from that, I was getting a lint warning because the public method NewAppRepositoriesHandler was returning this private struct.

PR #1514, opened 2020-02-13
Add DeleteAppRepository handler and use that from frontend.

absoludity:
Ref #1496 . Adds a delete handler in the backend and switches the dashboard to use that. Following PR will ensure a repo secret is deleted from the Kubeapps namespace when appropriate.

While there, I've moved the use of `mux` back out to the http-handler. We'll still want to pull out the use of `http.Request` I assume.


andresmgot:
_Merging this to iterate in the refactor_

PR #1515, opened 2020-02-13
Delete kubeapps copy of secret if app repo has creds.

absoludity:
Explicitly delete kubeapps copy of secret when app repo with creds is deleted. Ensure normal secret deleted by propagation (tried to test, but as per comment, k8s fake client doesn't do finalizers/propagation, so tested IRL).

Fixed a couple of other things that I found with IRL testing: missing RBAC for service to create/delete secrets in its own namespace, as well as the name of created secrets: removed useless suffix of `-secret`.

Ref: #1496. Follows #1514

### DIFF
--- a/pkg/apprepo/apprepos_handler.go
+++ b/pkg/apprepo/apprepos_handler.go
@@ -212,6 +212,7 @@ func (a *AppRepositoriesHandler) CreateAppRepository(req *http.Request, requestN
 		// https://docs.google.com/document/d/1YEeKC6nPLoq4oaxs9v8_UsmxrRfWxB6KCyqrh2-Q8x0/edit?ts=5e2adf87#heading=h.kilvd2vii0w
 		if requestNamespace != a.kubeappsNamespace {
 			repoSecret.ObjectMeta.Name = kubeappsSecretNameForRepo(appRepo.ObjectMeta.Name, appRepo.ObjectMeta.Namespace)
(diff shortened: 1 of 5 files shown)


pkg/apprepo/apprepos_handler.go
```diff
@@ -227,8 +228,26 @@ func (a *AppRepositoriesHandler) DeleteAppRepository(req *http.Request, repoName
...
+	}
 
-	return clientset.KubeappsV1alpha1().AppRepositories(repoNamespace).Delete(repoName, &metav1.DeleteOptions{})
+	// If the app repo was in a namespace other than the kubeapps one, we also delete the copy of
+	// the repository credentials kept in the kubeapps namespace (the repo credentials in the actual
+	// namespace should be deleted when the owning app repo is deleted).
+	if hasCredentials && repoNamespace != a.kubeappsNamespace {
+		err = clientset.CoreV1().Secrets(a.kubeappsNamespace).Delete(kubeappsSecretNameForRepo(repoName, repoNamespace), &metav1.DeleteOptions{})
+	}
+	return err
```
andresmgot: what I have done other times is just try to delete the secret and ignore the error if it's a NotFound. That way you skip one request and in case we create the secret because any other reason (apart than the Auth.Header or the CustomCA) we are sure we delete it.

andresmgot:
_Merging this to iterate in the refactor_

PR #1516, opened 2020-02-13
Tests for pkg/http-handler

andresmgot:
### Description of the change

Follow up of #1513

Convering `pkg/http-handler` with tests.


pkg/http-handler/http-handler.go
```diff
@@ -39,20 +39,24 @@ type appRepositoryResponse struct {
 	AppRepository v1alpha1.AppRepository `json:"appRepository"`
 }
 
+func returnK8sError(err error, w http.ResponseWriter) {
+	if statusErr, ok := err.(*k8sErrors.StatusError); ok {
+		status := statusErr.ErrStatus
+		log.Infof("unable to create app repo: %v", status.Reason)
```
absoludity: We need to generalise the log message here no?

PR #1518, opened 2020-02-13
Move all k8s resources operations to pkg/apprepo

andresmgot:
### Description of the change

Requires #1517

The goal of this PR is to complete the migration to the existing operations over k8s resources to the package `pkg/apprepo` (the name of the package is to be changed). The advantage of this package is that it allow callers to set a `token`. If set, it uses that token to do requests, if not it uses the backend serviceaccount.

Now the package `pkg/chart` uses this package thanks to the implementation of the methods `GetAppRepository` and `GetSecret` that use the same approach for the `token`.

The fake Handler has been moved to its own package so it can be reused.

### DIFF
--- a/cmd/kubeops/internal/handler/handler.go
+++ b/cmd/kubeops/internal/handler/handler.go
@@ -3,12 +3,13 @@ package handler
 import (
 	"encoding/json"
 	"net/http"
+	"os"
 	"strconv"
 
 	"github.com/gorilla/mux"
 	"github.com/kubeapps/common/response"
-	appRepo "github.com/kubeapps/kubeapps/cmd/apprepository-controller/pkg/client/clientset/versioned"
 	"github.com/kubeapps/kubeapps/pkg/agent"
+	"github.com/kubeapps/kubeapps/pkg/apprepo"
 	"github.com/kubeapps/kubeapps/pkg/auth"
 	chartUtils "github.com/kubeapps/kubeapps/pkg/chart"
 	"github.com/kubeapps/kubeapps/pkg/chart/helm3to2"
@@ -16,15 +17,17 @@ import (
 	log "github.com/sirupsen/logrus"
 	"github.com/urfave/negroni"
 	"helm.sh/helm/v3/pkg/action"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 )
 
 const (
-	authHeader     = "Authorization"
-	namespaceParam = "namespace"
-	nameParam      = "releaseName"
-	authUserError  = "Unexpected error while configuring authentication"
+	authHeader       = "Authorization"
+	namespaceParam   = "namespace"
+	nameParam        = "releaseName"
+	authUserError    = "Unexpected error while configuring authentication"
+	defaultNamespace = metav1.NamespaceSystem
 )
 
 const isV1SupportRequired = false
@@ -91,23 +94,13 @@ func WithHandlerConfig(storageForDriver agent.StorageForDriver, options Options)
 				return
 			}
 
-			// System configuration and clients, using the service serviceaccount
-			// Used to retrieve apprepositories and secrets related to them
-			svcRestConfig, err := rest.InClusterConfig()
-			if err != nil {
-				log.Errorf("Failed to create in-cluster config with service account: %v", err)
-				response.NewErrorResponse(http.StatusInternalServerError, authUserError).Write(w)
-				return
+			kubeappsNamespace := os.Getenv("POD_NAMESPACE")
+			if namespace == "" {
+				namespace = defaultNamespace
 			}
-			svcKubeClient, err := kubernetes.NewForConfig(svcRestConfig)
+			appRepoHandler, err := apprepo.NewAppRepositoriesHandler(kubeappsNamespace)
 			if err != nil {
-				log.Errorf("Failed to create kube client with service account: %v", err)
-				response.NewErrorResponse(http.StatusInternalServerError, authUserError).Write(w)
-				return
-			}
-			appRepoClient, err := appRepo.NewForConfig(svcRestConfig)
-			if err != nil {
-				log.Errorf("Failed to create app repo kube client with service account: %v", err)
+				log.Errorf("Failed to create handler: %v", err)
 				response.NewErrorResponse(http.StatusInternalServerError, authUserError).Write(w)
 				return
 			}
@@ -115,7 +108,7 @@ func WithHandlerConfig(storageForDriver agent.StorageForDriver, options Options)
 			cfg := Config{
 				Options:      options,
 				ActionConfig: actionConfig,
-				ChartClient:  chartUtils.NewChartClient(svcKubeClient, appRepoClient, options.UserAgent),
+				ChartClient:  chartUtils.NewChartClient(appRepoHandler, kubeappsNamespace, options.UserAgent),
 			}
 			f(cfg, w, req, params)
 		}
--- a/cmd/kubeops/internal/handler/handler.go
+++ b/cmd/kubeops/internal/handler/handler.go
@@ -3,7 +3,6 @@ package handler
 import (
 	"encoding/json"
 	"net/http"
-	"os"
 	"strconv"
 
 	"github.com/gorilla/mux"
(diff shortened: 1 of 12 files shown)


absoludity:
> The advantage of this package is that it allow callers to set a `token`. If set, it uses that token to do requests, if not it uses the backend serviceaccount.

This is only for the GetNamespaces method, right? The way you've stated it here sounds like it's general functionality for the package (ie. including other methods like CreateAppRepository etc.)

cmd/kubeops/internal/handler/handler.go
```diff
@@ -91,31 +94,21 @@ func WithHandlerConfig(storageForDriver agent.StorageForDriver, options Options)
...
-				return
-			}
-			svcKubeClient, err := kubernetes.NewForConfig(svcRestConfig)
-			if err != nil {
-				log.Errorf("Failed to create kube client with service account: %v", err)
-				response.NewErrorResponse(http.StatusInternalServerError, authUserError).Write(w)
-				return
+			kubeappsNamespace := os.Getenv("POD_NAMESPACE")
+			if namespace == "" {
+				namespace = defaultNamespace
```
absoludity: Why are we defaulting/guessing that the Kubeapps namespace is the `kube-system` here (`metav1.NamespaceSystem`)?

We should now be able to fail to start, given that `POD_NAMESPACE` is set for both tiller-proxy and kubeops? EDIT: Ah, because this is within a function which has no option of returning an error :/ . If `WithHandlerConfig` requires knowing what the kubeapps namespace is, let's make it a required arg?
EDIT2: And because you've not changed this, it was already like this in `pkg/chart` from where it was moved. It seems like a bad guess to make, potentially a security issue.

My vote would be to pass it as a required arg (and in the main.go, fail if it's not defined). Let me know if you've got other ideas.

pkg/apprepo/apprepos_handler.go
```diff
@@ -151,7 +153,9 @@ func clientsetForConfig(config *rest.Config) (combinedClientsetInterface, error)
 // ConfigForToken returns a new config for a given auth token.
 func (a *AppRepositoriesHandler) ConfigForToken(token string) *rest.Config {
 	configCopy := a.config
-	configCopy.BearerToken = token
+	if token != "" {
+		configCopy.BearerToken = token
+	}
```
absoludity: This looks dangerous - as though it implicitly returns a service account config if you pass in an empty token? `ConfigForToken` should only ever return a config for the specified token, whether or not the token is passed with bad information.

pkg/apprepo/apprepos_handler.go
```diff
@@ -248,6 +252,16 @@ func (a *AppRepositoriesHandler) DeleteAppRepository(repoName, repoNamespace, to
 	return err
 }
 
+// GetAppRepository returns an AppRepository resource from a namespace.
+// Optionally set a token to get the AppRepository using a custom serviceaccount
+func (a *AppRepositoriesHandler) GetAppRepository(repoName, repoNamespace, token string) (*v1alpha1.AppRepository, error) {
+	clientset, err := a.clientsetForRequest(token)
```
absoludity: Right, this looks like we're depending on `clientsetForRequest("")` returning the service account clientset. I'm -1 on that, it's too easy to mistakenly end up with a service account client set. I think we should remove that and do something like:
```
if token != "" {
    clientset, err := a.clientsetForToken(token)
} else {
    clientset = svcKubeClient
}
```

In fact, even then the optional token on this public method is very implicit at call-sites. We could additionally make this fn private (`getAppRepository`) and instead expose two very small (single-line) wrappers:
```
func (a *AppRepositoriesHandler) GetAppRepositoryForUser(repoName, repoNamespace, token string) ...
```
and 
```
func (a *AppRepositoriesHandler) GetAppRepositoryForServiceAccount(repoName, repoNamespace string) ...
```
calling the private method.

pkg/apprepo/apprepos_handler.go
```diff
@@ -385,3 +399,13 @@ func (a *AppRepositoriesHandler) GetNamespaces(token string) ([]corev1.Namespace
 
 	return allowedNamespaces, nil
 }
+
+// GetSecret return the a secret from a namespace using a token if given
+func (a *AppRepositoriesHandler) GetSecret(name, namespace, token string) (*corev1.Secret, error) {
+	userClientset, err := a.clientsetForRequest(token)
```
absoludity: Same here.

pkg/chart/chart.go
```diff
@@ -308,22 +306,17 @@ func (c *ChartClient) InitNetClient(details *Details) (HTTPClient, error) {
...
 
-	namespace := os.Getenv("POD_NAMESPACE")
-	if namespace == "" {
-		namespace = defaultNamespace
-	}
-
 	// We grab the specified app repository (for later access to the repo URL, as well as any specified
 	// auth).
-	appRepo, err := c.appRepoClient.KubeappsV1alpha1().AppRepositories(namespace).Get(details.AppRepositoryResourceName, metav1.GetOptions{})
+	appRepo, err := c.appRepoHandler.GetAppRepository(details.AppRepositoryResourceName, c.kubeappsNamespace, "")
```
absoludity: So this (and the calls below) would use `.GetAppRepositoryForServiceAccount(details.AppRepositoryResourceName, c.kubeappsNamespace)`

andresmgot:
Thanks for the review @absoludity. Check 7a839b5, I think we can both flexibility and explicit token usage as I explain in a inline comment. Let me know what you think about it.

cmd/tiller-proxy/main.go
```diff
@@ -135,7 +130,17 @@ func main() {
 	}
 
 	proxy = tillerProxy.NewProxy(kubeClient, helmClient, timeout)
-	chartClient := chartUtils.NewChartClient(kubeClient, appRepoClient, userAgent())
+	kubeappsNamespace := os.Getenv("POD_NAMESPACE")
+	if kubeappsNamespace == "" {
+		log.Fatalf("POD_NAMESPACE should be defined")
+	}
```
absoludity: Much better, thanks for moving it out to main.go even though it wasn't a change you'd added.

pkg/apprepo/apprepos_handler.go
```diff
@@ -52,33 +52,56 @@ type combinedClientset struct {
...
+	DeleteAppRepository(name, namespace string) error
+	GetNamespaces() ([]corev1.Namespace, error)
+	GetSecret(name, namespace string) (*corev1.Secret, error)
+	GetAppRepository(repoName, repoNamespace string) (*v1alpha1.AppRepository, error)
+}
+
+// AuthHandler exposes handler functionality as a user or the current serviceaccount
+type AuthHandler interface {
+	AsUser(token string) handler
+	AsSVC() handler
```
absoludity: Excellent, much more explicit while still being easy to use for call-sites. Thanks!

The only down-side I can think of here is that it now means that all the methods on the interface can be used with the service account, rather than just `GetNamespaces` and `GetAppRepository`. We could fix that by composing the interface and having AsSVC() return the more limited interface? Not sure. See what you think.

pkg/apprepo/apprepos_handler.go
```diff
@@ -52,33 +52,56 @@ type combinedClientset struct {
...
+	if err != nil {
+		log.Errorf("unable to create clientset: %v", err)
+	}
+	a.clientset = clientset
+	return a
+}
+
+func (a *appRepositoriesHandler) AsSVC() handler {
+	a.clientset = a.svcClientset
+	return a
```
absoludity: These don't look thread safe? Would it be better to return a copy configured with the correct clientset, rather than updating the internal `a.clientset` for the call-site?

Or, stepping back, it may even make more sense if the current `appRepositoryHandler` was something like a handler factory, that only created a handler when you call `AsUser` or `AsSVC`. Not sure about needing a factory, but the main thing is that it seems in its current form here to be mixing responsibilities... more below.

pkg/apprepo/apprepos_handler.go
```diff
@@ -125,13 +148,17 @@ func NewAppRepositoriesHandler(kubeappsNamespace string) (*AppRepositoriesHandle
...
+	}
 
-	return &AppRepositoriesHandler{
+	return &appRepositoriesHandler{
 		config:            *config,
 		kubeappsNamespace: kubeappsNamespace,
 		// See comment in the struct defn above.
 		clientsetForConfig: clientsetForConfig,
-		svcKubeClient:      svcKubeClient,
+		svcClientset:       &combinedClientset{svcAppRepoClient, svcKubeClient},
```
absoludity: So `a.clientset` isn't yet defined (as no one has called `AsUser()` or `AsSVC()` yet), so if a call-site now called `a.CreateAppRepository(..)` they'd get a run-time panic, I think? This is what I meant above about mixing responsibilities. Perhaps what is returned here should *not* implement the interface, and only the handler returned by `AsUser()` or `AsSVC()` should implement the interface. So the object returned here wouldn't be a handler, but enables you to create a handler.

pkg/apprepo/apprepos_handler.go
andresmgot: That's what I wanted, if in the future we need to use this Handler with any the svc account or the user account we don't need to modify the library. Since it's quite explicit now, I don't think there is risk for a developer to misuse the method (nor a reviewer to miss it).

pkg/apprepo/apprepos_handler.go
andresmgot: that's a good point, let me check the possibility of modifying the interfaces to fix this and you next comment.

pkg/apprepo/apprepos_handler.go
andresmgot: agree

andresmgot:
Merging! let me know if you want to discuss something further

absoludity:
> Merging! let me know if you want to discuss something further

LGTM, thanks!

Issue #1521, opened 2020-02-17
Record and display catalog as charts available per namespace


absoludity:
The previous task (#1496) enabled the asset-sync service to sync charts across all namespaces. This task continues that work to enable the dashboard to display the charts available per namespace. See [design doc](https://docs.google.com/document/d/1YEeKC6nPLoq4oaxs9v8_UsmxrRfWxB6KCyqrh2-Q8x0/edit?ts=5e2adf87#) for more info and discussion.

Follows on from #1496, parent task #1256.

- [x] Ensure that per-namespace app repositories are sync'd using their defined repo credentials (sync job needs these passed from kubeapps copy)
- [x] asset-sync: Record the namespace in the assetsvc (db - mongo / postgresql) when sync'ing chart data (so a chart can exist in multiple namespaces, or _all)
- [x] assetsvc: update requests to the chart-svc for charts to expect both the users credentials as well as the namespace being requested, returning charts for the namespace if the user has access.
- [x] UI: Update dashboard to request and display charts per namespace when viewing the catalog (behind the feature flag)

PR #1525, opened 2020-02-18
Use correct secret in kubeapps namespace when syncing an app repo

absoludity:
Updates the cronjob spec so that, when an app repository is from a namespace other than kubeapps, the correct secret ref is used for the repo.

Ref: #1521 .

### DIFF
--- a/cmd/apprepository-controller/controller.go
+++ b/cmd/apprepository-controller/controller.go
@@ -25,6 +25,7 @@ import (
 	appreposcheme "github.com/kubeapps/kubeapps/cmd/apprepository-controller/pkg/client/clientset/versioned/scheme"
 	informers "github.com/kubeapps/kubeapps/cmd/apprepository-controller/pkg/client/informers/externalversions"
 	listers "github.com/kubeapps/kubeapps/cmd/apprepository-controller/pkg/client/listers/apprepository/v1alpha1"
+	"github.com/kubeapps/kubeapps/pkg/kube"
 	log "github.com/sirupsen/logrus"
 	batchv1 "k8s.io/api/batch/v1"
 	batchv1beta1 "k8s.io/api/batch/v1beta1"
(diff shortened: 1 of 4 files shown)


cmd/apprepository-controller/controller.go
```diff
@@ -440,12 +441,12 @@ func newSyncJob(apprepo *apprepov1alpha1.AppRepository, kubeappsNamespace string
...
 			OwnerReferences: ownerReferencesForAppRepo(apprepo, kubeappsNamespace),
 		},
-		Spec: syncJobSpec(apprepo),
+		Spec: syncJobSpec(apprepo, kubeappsNamespace),
 	}
 }
 
 // jobSpec returns a batchv1.JobSpec for running the chart-repo sync job
-func syncJobSpec(apprepo *apprepov1alpha1.AppRepository) batchv1.JobSpec {
+func syncJobSpec(apprepo *apprepov1alpha1.AppRepository, kubeappsNamespace string) batchv1.JobSpec {
```
andresmgot: here (line 457) you also need to change the secretKeyRef?

cmd/apprepository-controller/controller.go
```diff
@@ -582,10 +583,14 @@ func apprepoSyncJobEnvVars(apprepo *apprepov1alpha1.AppRepository) []corev1.EnvV
 		},
 	})
 	if apprepo.Spec.Auth.Header != nil {
+		secretKeyRef := apprepo.Spec.Auth.Header.SecretKeyRef
+		if apprepo.ObjectMeta.Namespace != kubeappsNamespace {
+			secretKeyRef.LocalObjectReference.Name = kube.KubeappsSecretNameForRepo(apprepo.ObjectMeta.Name, apprepo.ObjectMeta.Namespace)
+		}
```
andresmgot: I wonder if it would be simpler to just create a copy of the apprepo in the kubeapps namespace. That would help with the garbage collection and wouldn't require to change these secret references on the fly

cmd/apprepository-controller/controller.go
absoludity: Yes, thanks.

cmd/apprepository-controller/controller.go
absoludity: It would, if we weren't transitioning towards an end goal of syncing across all namespaces and running the jobs in those namespaces. This way we can begin syncing AppRepos across namespaces, we are just still running the sync jobs in the kubeapps namespace until we can update the assetsvc to receive data from the syncer.

Thanks!

PR #1526, opened 2020-02-18
Record the namespace in the assetsvc when syncing chart data.

absoludity:
This PR:
- Adds a new required arg [EDIT: now a kwarg] to the sync command (namespace) and uses this from the app repo controller,
- updates the `initTables` function of the pg manager to include a required column for namespace on the repos table (and updates the unique to use both name and namespace relations),
- adds a foreign key repository_id to the charts table, though I've left this nullable and want to chat about this) [EDIT: it's no longer nullable]
- modifies `UpdateLastCheck` to record the repo namespace. [EDIT: we could actually just update this to a straight update rather than upsert now].
- EDIT: Now includes db unit tests for the new/changed functionality.

E.g. with a running postgres:
```
$ go test ./... -count=1 -v -run TestGetRepoId
=== RUN   TestGetRepoId
=== RUN   TestGetRepoId/it_returns_a_new_ID_if_it_does_not_yet_exist
=== RUN   TestGetRepoId/it_returns_the_existing_ID_if_the_repo_exists_in_the_db
--- PASS: TestGetRepoId (0.06s)
    --- PASS: TestGetRepoId/it_returns_a_new_ID_if_it_does_not_yet_exist (0.03s)
    --- PASS: TestGetRepoId/it_returns_the_existing_ID_if_the_repo_exists_in_the_db (0.03s)
PASS
ok      github.com/kubeapps/kubeapps/cmd/asset-syncer   0.094s
```

E.g. without a running postgres:
```
$ go test ./... -count=1 -v -run TestGetRepoId
=== RUN   TestGetRepoId
--- SKIP: TestGetRepoId (0.00s)
    postgresql_db_test.go:57: skipping postgres tests: dial tcp 127.0.0.1:5432: connect: connection refused
PASS
ok      github.com/kubeapps/kubeapps/cmd/asset-syncer   0.037s

```

A couple of questions:
1. How do we handle schema changes? I don't see anything for handling these. As a result, if I deploy first without these changes, and then update the images for syncer and app-repo controller to the ones from this PR, sync jobs then fail as you would expect with:
```
$ k -n kubeapps logs apprepo-dex-sync-testme-cn2pr-tp89l
time="2020-02-18T04:05:12Z" level=fatal msg="pq: column \"namespace\" of relation \"repos\" does not exist"
```
The only way I can avoid this is by running:
```
k -n kubeapps exec kubeapps-postgresql-0 -- sh -c 'PGPASSWORD=$POSTGRES_PASSWORD psql -d assets -U $POSTGRES_USER -c "DROP TABLE repos;"
```
so the table is re-created. Given that this is only a cache, we could just drop the table and force app repo syncs, but it really looks like we're missing migrations. Given that the postgres support is only new, it may not be an issue, but I'll definitely create an issue so we can do this in the future. [EDIT: resolving this in the next PR]

2. Currently, when a repo is first synced, we do not create a repo record until *after* the charts have been added (via `UpdateLastCheck`). But if we want a foreign key on the chart table, we need to know the repo id before adding the charts. So we could either update to do so, or, the easier option is to just add the repo name and namespace to the chart table also (with the same unique constraint). Thoughts? [EDIT: this has been resolved in this PR]

Follows #1525. Ref #1521


cmd/asset-syncer/postgresql_utils.go
```diff
@@ -62,17 +62,27 @@ func (m *postgresAssetManager) Sync(charts []models.Chart) error {
...
+	_, err = m.DB.Exec(fmt.Sprintf(`
+CREATE TABLE IF NOT EXISTS %s (
+	ID serial NOT NULL PRIMARY KEY,
+	repository_id integer REFERENCES %s (ID),
+	chart_id varchar unique,
+	info jsonb NOT NULL
+)`, dbutils.ChartTable, dbutils.RepositoryTable))
 	if err != nil {
 		return err
 	}
```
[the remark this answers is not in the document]
andresmgot: chart_files_ID contains the repository, e.g. `stable/wordpress-8.1.3`

cmd/asset-syncer/postgresql_utils.go
```diff
@@ -96,30 +117,66 @@ func (m *postgresAssetManager) RepoAlreadyProcessed(repoName, repoChecksum strin
...
+	rows.Next()
+	var id int
+	if err := rows.Scan(&id); err != nil {
+		return 0, err
+	}
+
+	// There should only ever be one ID, but just to be sure
+	if rows.Next() {
+		return 0, fmt.Errorf("expected exactly 1 id returned, got more than one")
+	}
```
andresmgot: maybe you prefer to use QueryRow? I don't think this extra code is necessary?

cmd/asset-syncer/postgresql_utils.go
```diff
@@ -187,7 +244,7 @@ func (m *postgresAssetManager) filesExist(chartFilesID, digest string) bool {
 func (m *postgresAssetManager) insertFiles(chartFilesID string, files models.ChartFiles) error {
 	query := fmt.Sprintf(`INSERT INTO %s (chart_files_ID, info)
```
andresmgot: so yes, we need to either add the namespace to `chart_files_ID` or add the namespace as a different column.

cmd/asset-syncer/sync.go
```diff
@@ -52,14 +52,14 @@ var syncCmd = &cobra.Command{
...
-		r, repoContent, err := getRepo(args[0], args[1], authorizationHeader)
+		repo, repoContent, err := getRepo(namespace, args[0], args[1], authorizationHeader)
 		if err != nil {
 			logrus.Fatal(err)
 		}
 
 		// Check if the repo has been already processed
-		if manager.RepoAlreadyProcessed(r.Name, r.Checksum) {
-			logrus.WithFields(logrus.Fields{"url": r.URL}).Info("Skipping repository since there are no updates")
+		if manager.RepoAlreadyProcessed(repo.Name, repo.Checksum) {
```
andresmgot: this call should also include the namespace?

cmd/asset-syncer/postgresql_utils.go
absoludity: Right, I went with the latter (namespaces as a different column), as I think we use the `chart_files_ID` in the UI which would look a bit odd as `kubeapps/bitnami/dex-1.8.0`.

cmd/asset-syncer/postgresql_utils.go
```diff
@@ -185,12 +233,15 @@ func (m *postgresAssetManager) filesExist(chartFilesID, digest string) bool {
 }
 
 func (m *postgresAssetManager) insertFiles(chartFilesID string, files models.ChartFiles) error {
```
absoludity: Do you know whether the `chartFilesID` arg here is just left over? I checked what gets imported, and the `files.ID` appears identical and so could be used to fill the field too.

andresmgot: I probably missed that so no, there is not a reason

cmd/asset-syncer/postgresql_db_test.go
```diff
@@ -0,0 +1,168 @@
...
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			pam, cleanup := getInitializedManager(t)
+			defer cleanup()
+
+			for _, repo := range tc.existingRepos {
+				_, err := pam.getRepoId(repo.Namespace, repo.Name)
```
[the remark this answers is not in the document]
andresmgot: ah, so this function is _upserting_ (I don't know if that's a real word)

cmd/asset-syncer/postgresql_utils.go
```diff
@@ -62,24 +69,40 @@ func (m *postgresAssetManager) Sync(charts []models.Chart) error {
...
-	))
+
+	_, err = m.DB.Exec(fmt.Sprintf(`
+CREATE TABLE IF NOT EXISTS %s (
+	ID serial NOT NULL PRIMARY KEY,
+	namespace varchar NOT NULL,
+	chart_files_ID varchar NOT NULL,
+	info jsonb NOT NULL,
+	UNIQUE(namespace, chart_files_ID)
+)`, dbutils.ChartFilesTable))
```
andresmgot: check the function `RepoAlreadyProcessed` below, I think that require to use the namespace as well

absoludity: Yes, as does Delete. I plan to update those in a later PR (it doesn't affect things when reposPerNamespace is false).

PR #1527, opened 2020-02-18
Replace kubeops image when releasing a chart

andresmgot:
### Description of the change

The image for `kubeops` was missing from the list that is changed when generating the released chart (so it was pointing to the dev image).


andresmgot:
Merging since this is not covered in the normal CI flow

Issue #1524, opened 2020-02-17
AppView Revamp


andresmgot:
### Description:

The goal of this issue is to redesign the current AppView and improve the UX in different areas:

 - Improve error detection. Make an easy to understand view that can point the user to the piece(s) of the Chart that are failing.
 - Improve usability. Make it easier to modify/upgrade/rollback/test/delete the application.
 - Improve the debugging experience. When something fails while operating a release, it should be possible to detect the cause of the issue and fix it without the need of a terminal. This means being able to report kubernetes events/errors to the user plus being able to read logs.

### Additional information you deem important (e.g. issue happens only occasionally):

Other related issues:
 - https://github.com/kubeapps/kubeapps/issues/962
 - https://github.com/kubeapps/kubeapps/issues/848
 - https://github.com/kubeapps/kubeapps/issues/529

PR #1529, opened 2020-02-19
Design proposal for AppView improvements

andresmgot:
### Description of the change

Design doc for improvements introduced in #1524


docs/architecture/design-proposals/dashboard/application-view-revamp.md
```diff
@@ -0,0 +1,112 @@
...
+
+<img src="./img/credentials-box.png" width="400px">
+
+Note that we will still be showing the different application Secrets along with all the other resources in its own table.
+
+### Compress Application Actions
+
+The current approach is to show one button per action (Upgrade, Delete, Rollback). This list is growing over time (for example, the backend endpoint for running tests is ready to be used). It can also be confusing to show a different list of buttons depending on the release state. One example is the Rollback button that is only rendered if the application has been upgraded. In order to avoid these issues, we can show a clickable menu with the different options, graying out the options not available (potentially showing a tooltip):
```
absoludity: Also not a huge priority imo, though agree it would be nicer. One point related to this that was pointed out a month ago was that people are needing to use "Upgrade" when the really want to reconfigure the existing release (ie. a new release, but with changed config options, no upgrade). I wonder if this is worth considering in relation to actions (ie. providing a way in the UI to reconfigure without upgrading and having to use the current chart). This would also enable us to revert the reversion and default to the latest chart version when *actually* upgrading.

docs/architecture/design-proposals/dashboard/application-view-revamp.md
```diff
@@ -0,0 +1,112 @@
...
+
+<img src="./img/configuration-options.png" width="300px">
+
+### Include Additional Information for Access URLs
+
+The current list of URLs can be improved with two small changes:
+
+- If the browser can access the URL, we could show an icon so the user knows if the URL is working.
```
absoludity: As in, we can fetch a HEAD request or similar, and as long as we get a 200 or a method not allowed, we know it is up?

docs/architecture/design-proposals/dashboard/application-view-revamp.md
```diff
@@ -0,0 +1,112 @@
...
+- If the URL is not working and the cause is well known, we could show additional information for the user to debug it (e.g. Pending IPs when using LoadBalancers in Minikube [link](https://github.com/kubeapps/kubeapps/issues/953)).
```
absoludity: This also sounds very useful. Worth getting input on the UI when Angel or others have time.

docs/architecture/design-proposals/dashboard/application-view-revamp.md
```diff
@@ -0,0 +1,112 @@
...
+
+ <img src="./img/url-list.png">
+
+### Render an Extended Resources Table
+
+Finally, we can show a single resource table with all the resources so users can inspect in detail the different resources of the application without adding too much noise to the view. Apart from the list of resources we are currently showing, we can add the list of pods related to the application (the ones that are related to a Deployment/Statefulset/Daemonset).
```
absoludity: This is looking like too many knobs to me, and moving away from the idea that Kubeapps is enabling you to interact with a more general 'Application', rather than a set of Kubernetes resources. As an example, I don't see a reason for us exposing the individual pods, given that:

1. We can show logs for the more general "service" object with `kubectl logs service/wordpress` (which shows logs for all pods in the deployments matching the service selector, afaict), and
2. If a deployment for a service has a pod in an error state, we can raise that information, and only that (relevant) information, to the focus for that service (eg. scheduling error, not enough cpu or whatever), without needing to drill down so far.

I reckon that we could do something much simpler here by just bringing the services of an app to the forefront and have detailed knobs to see all k8s resources in the background if at all, rather than in the foreground. Another way of thinking about it: I think the design applied here should do to the application view what your work on the deployment form did for the deployment view: focusing on just what the user needs, rather than providing all the knobs and exposing everything. Let's think very carefully about this one.

andresmgot:
@absoludity anything else you think we should cover here?

docs/architecture/design-proposals/dashboard/application-view-revamp.md
andresmgot: > I actually think we could have a tool (if there's not one already) that can be pointed at a deployment and determine the issue, returning a "report" with both human readable and machine readable info. This could be a separate tool (hmm, could be like the `SelfSubjectAccessReview` - with a CRD that returns a resource with the details filled in - could be very useful as a separate tool... anyway). I think we could raise your example error to something like: "The init-container 'foo-container' of the pod 'foo-abc123' has failed to start. The reported error is: 'whatever - insufficient storage'.", with further details in the report, but not attempt to provide the ability to dig into any and every k8s resource in the Kubeapps UI (at least, not yet).

I don't think such a tool exists, for the moment the easiest way to implement something close is allow people to discover errors on their own.

> IMO, I don't think we should be aiming to provide a "complete experience", (again, at least not yet). It may be down the road that once we have the smooth overview experience for the App, that we allow people to do things like drill down and view/edit individual resources, but the people who understand that UX and those resources are going to be the people who are comfortable with the CLI, so I personally think that initially the focus should be on improving things for the user who wants to install stuff and wants to know what went wrong without navigating all the resources (again, big +1 to logs per service, but let's think about going deeper).

I disagree but even if this only helps people that already would be able to fix the issue with a terminal, I think is valuable. If you don't need to leave kubeapps in order to deploy, detect an error and redeploy you are saving people time.

Also, for non experienced users, they would be able to click around until they find some message that says "Error: Whatever..." giving at least the chance to fix it even if they don't know how to use `kubectl` or how retrieve k8s events related to a deployment.

IMO this would be a great feature for Kubeapps (and not something super complicated). It's something that as a user I would really like. For example, Andreas said: "I really like the table showing Application Resources." so I am not alone there :)

absoludity: > 
> I don't think such a tool exists,

Yeah, it could be quite interesting, not just to Kubeapps, but other tools like BKPR. I'll look around out of interest.

> for the moment the easiest way to implement something close is allow people to discover errors on their own.

Hmm, I think we could give it a little more thought than that, to see if we could do something which was a better experience than just exposing the CLI resources, but you could be right.

> 
> > IMO, I don't think we should be aiming to provide a "complete experience", (again, at least not yet). It may be down the road that once we have the smooth overview experience for the App, that we allow people to do things like drill down and view/edit individual resources, but the people who understand that UX and those resources are going to be the people who are comfortable with the CLI, so I personally think that initially the focus should be on improving things for the user who wants to install stuff and wants to know what went wrong without navigating all the resources (again, big +1 to logs per service, but let's think about going deeper).
> 
> I disagree but even if this only helps people that already would be able to fix the issue with a terminal, I think is valuable. If you don't need to leave kubeapps in order to deploy, detect an error and redeploy you are saving people time.

I think we agree on that point though: helping people to be able to fix the issue within Kubeapps is helpful and something we want to do. We already agree that allowing people to dig to logs will be useful (and perhaps raise these to the users focus when there is an error). The only point where we're currently exploring differences is *how* to best expose errors or details: Should we provide direct access from the app view to all relevant k8s resources (similar to Octant?) that they can browse to find the error, or should we try to raise the error up to the app view's main flow (while still allowing people to dig down to some detail where required).

And all I'm trying to say at that point is: I'm not sure. I want to make sure we think carefully about product focus and direction, rather than just doing something which fills the gap.

> 
> Also, for non experienced users, they would be able to click around until they find some message that says "Error: Whatever..." 

Yes, but is "click around until they find some message" the experience we want to enable? If it is possible for us to instead present something clearly such as "Your app 'apache'  did not successfully deploy because there were not enough system resources available to create the foo resources", or "Your app 'foobar' is not successful yet because the deployment 'foobar-database' is failing to start. More details available in the [link]logs[/link]". 

>giving at least the chance to fix it even if they don't know how to use `kubectl` or how retrieve k8s events related to a deployment.

Yes - we definitely want to ensure this is possible which ever way we go.

> 
> IMO this would be a great feature for Kubeapps (and not something super complicated). It's something that as a user I would really like. For example, Andreas said: "I really like the table showing Application Resources." so I am not alone there :)

And I'm agreeing too - that being able to browse all resources would be a great feature - I'm just unsure whether we should show it prominently here on the app view regardless of errors, or whether there is a better option to help users find and fix errors on the app view while still being able to drill down to inspect the resources.

Either way, as above, I'm not wanting you to block on this - it'll be a great step forward either way - I just thought it worth getting more input. Hope that clears things up!

docs/architecture/design-proposals/dashboard/application-view-revamp.md
```diff
@@ -0,0 +1,126 @@
...
+
+Apart from the basic information of the resource, we could add a summary (human-friendly) to know the status of the resource if possible. 
+
+We can also show different buttons in order to show the resource YAML, description and logs if available (available for pods, services and workloads).
+
+<img src="./img/resources-table.png">
+
```
absoludity: "If we later develop a clearer way to capture and present errors to users, clearly identifying the cause without requiring clicking through resources, the resources table will still be useful as a way to dig deeper and learn more without leaving the UI."

PR #1532, opened 2020-02-21
1521 delete tables on upgrade

absoludity:
Ref: #1521 

Following on from #1526, this PR adds a pre-upgrade hook which invalidates the database cache of charts. When postgres is being used, this drops and recreates the tables. For mongodb, it's currently a no-op. I've not added functionality to check for a schema change at the moment, so this will rebuild the cache on every upgrade.

IRL test:
Initial deploy dev environment and verify schema is the original (ie. without namespace field in repos table):
```
$ k -n kubeapps exec kubeapps-postgresql-0 -- sh -c 'PGPASSWORD=$POSTGRES_PASSWORD psql -d assets -U $POSTGRES_USER -c "\d repos;"'
                                    Table "public.repos"
   Column    |       Type        | Collation | Nullable |              Default
-------------+-------------------+-----------+----------+-----------------------------------
 id          | integer           |           | not null | nextval('repos_id_seq'::regclass)
 name        | character varying |           |          |
 checksum    | character varying |           |          |
 last_update | character varying |           |          |
Indexes:
    "repos_pkey" PRIMARY KEY, btree (id)
    "repos_name_key" UNIQUE CONSTRAINT, btree (name)
```

Upgrade with the new sync image (which has the new schema): 
```
helm -n kubeapps upgrade --reuse-values kubeapps ./chart/kubeapps/ --set apprepository.syncImage.tag=2 --set apprepository.image.tag=1
```

See preupgrade job deployed successfully (before I added the hook-delete-policy):
```
$ k -n kubeapps logs kubeapps-internal-apprepository-jobs-preupgrade-jt9zz
time="2020-02-21T00:58:32Z" level=info msg="Successfully invalidated cache"
```

Re-check the table and now see it's updated.
```
  $ k -n kubeapps exec kubeapps-postgresql-0 -- sh -c 'PGPASSWORD=$POSTGRES_PASSWORD psql -d assets -U $POSTGRES_USER -c "\d repos;"'
                                    Table "public.repos"
   Column    |       Type        | Collation | Nullable |              Default
-------------+-------------------+-----------+----------+-----------------------------------
 id          | integer           |           | not null | nextval('repos_id_seq'::regclass)
 namespace   | character varying |           | not null |
 name        | character varying |           | not null |
 checksum    | character varying |           |          |
 last_update | character varying |           |          |
Indexes:
    "repos_pkey" PRIMARY KEY, btree (id)
    "repos_namespace_name_key" UNIQUE CONSTRAINT, btree (namespace, name)
Referenced by:
    TABLE "charts" CONSTRAINT "charts_repository_id_fkey" FOREIGN KEY (repository_id) REFERENCES repos(id)
```

Verified cache is being repopulated as app repo starts.

### DIFF
--- a/chart/kubeapps/templates/_helpers.tpl
+++ b/chart/kubeapps/templates/_helpers.tpl
@@ -98,10 +98,10 @@ Create name for the apprepository-controller based on the fullname
 {{- end -}}
 
 {{/*
-Create name for the apprepository bootstrap job
+Create name for the apprepository pre-upgrade job
 */}}
-{{- define "kubeapps.apprepository-jobs-bootstrap.fullname" -}}
-{{ template "kubeapps.fullname" . }}-internal-apprepository-jobs-bootstrap
+{{- define "kubeapps.apprepository-jobs-preupgrade.fullname" -}}
+{{ template "kubeapps.fullname" . }}-internal-apprepository-jobs-preupgrade
 {{- end -}}
 
 {{/*
--- a/cmd/asset-syncer/postgresql_utils.go
+++ b/cmd/asset-syncer/postgresql_utils.go
@@ -247,3 +247,14 @@ func (m *postgresAssetManager) insertFiles(chartFilesID string, files models.Cha
 	}
 	return err
 }
+
+// InvalidateCache for postgresql deletes and re-writes the schema
+func (m *postgresAssetManager) InvalidateCache() error {
+	tables := strings.Join([]string{dbutils.RepositoryTable, dbutils.ChartTable, dbutils.ChartFilesTable}, ",")
+	_, err := m.DB.Exec(fmt.Sprintf("DROP TABLE IF EXISTS %s CASCADE", tables))
+	if err != nil {
+		return err
+	}
+
+	return m.initTables()
+}
--- a/cmd/asset-syncer/postgresql_utils.go
+++ b/cmd/asset-syncer/postgresql_utils.go
@@ -251,7 +251,7 @@ func (m *postgresAssetManager) insertFiles(chartFilesID string, files models.Cha
 // InvalidateCache for postgresql deletes and re-writes the schema
 func (m *postgresAssetManager) InvalidateCache() error {
 	tables := strings.Join([]string{dbutils.RepositoryTable, dbutils.ChartTable, dbutils.ChartFilesTable}, ",")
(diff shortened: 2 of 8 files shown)


chart/kubeapps/templates/_helpers.tpl
```diff
@@ -98,10 +98,10 @@ Create name for the apprepository-controller based on the fullname
 {{- end -}}
 
 {{/*
-Create name for the apprepository bootstrap job
```
absoludity: I'm assuming this is an old job which no longer exists, as I couldn't see the name being used anywhere.

cmd/asset-syncer/postgresql_utils.go
```diff
@@ -247,3 +247,14 @@ func (m *postgresAssetManager) insertFiles(chartFilesID string, files models.Cha
 	}
 	return err
 }
+
+// InvalidateCache for postgresql deletes and re-writes the schema
+func (m *postgresAssetManager) InvalidateCache() error {
+	tables := strings.Join([]string{dbutils.RepositoryTable, dbutils.ChartTable, dbutils.ChartFilesTable}, ",")
+	_, err := m.DB.Exec(fmt.Sprintf("DROP TABLE IF EXISTS %s CASCADE", tables))
```
andresmgot: is the `CASCADE` necessary?

cmd/asset-syncer/postgresql_utils.go
```diff
@@ -247,3 +247,14 @@ func (m *postgresAssetManager) insertFiles(chartFilesID string, files models.Cha
...
+	if err != nil {
+		return err
+	}
+
+	return m.initTables()
```
andresmgot: this is not really necessary since it will be executed with syncing but I am okay if you want to leave it to leave the tables ready

cmd/asset-syncer/postgresql_utils.go
absoludity: Not here it's not, thanks. I'd originally only deleted the one table and relied on the cascades, but I had to change that as the files table doesn't have a FK in this PR. I've removed the CASCADE and will re-add it when the FK's are in the next PR (I think)>

cmd/asset-syncer/postgresql_utils.go
absoludity: Yeah, the reason for having it here is so that it is sync'd with the correct code (the new, rather than from the still running older pods). We will probably need a transaction here to avoid race conditions with the existing pods too.

Issue #1330, opened 2019-11-28
Validate app repositories before adding them


juan131:
**Description**

When you add an app repository to Kubeapps, no validations are executed.

**Steps to reproduce the issue:**

1. Follow [this guide](https://github.com/kubeapps/kubeapps/blob/master/docs/user/private-app-repository.md) to add you private repository to Kubeapps.
2. Introduce some wrong configuration (e.g. an invalid URL).

**Describe the results you received:**

Kubeapps accepts the wrong configuration without performing any validation:

<img width="1102" alt="Screenshot 2019-11-28 at 13 38 23" src="https://user-images.githubusercontent.com/6740773/69807699-38b8b200-11e6-11ea-91c4-128c2db6d29b.png">

**Describe the results you expected:**

I would expect Kubeapps to validate the configuration and ensure it can connect and list the chars in the app repository.

**Version of Helm, Kubeapps and Kubernetes**:

- Output of `helm list <kubeapps-release-name>`:

```console
$ helm list kubeapps
NAME    	REVISION	UPDATED                 	STATUS  	CHART         	APP VERSION	NAMESPACE
kubeapps	1       	Thu Nov 28 12:23:42 2019	DEPLOYED	kubeapps-3.1.0	v1.7.1     	kubeapps
```

andresmgot:
Thanks @juan131, apart than validating the URL do you think something else needs to be validated?

juan131:
I guess the basic authentication (or any other provided method) should be validated too

andresmgot:
well, the problem is that the browser may not have access to the URL. For example, if I type `http://chartmuseum.default:8080` the application won't be able to validate the access that URL (apart that it's a valid URL) since that's only valid within the cluster.

In any case, it should be easy to validate that at least the URL is valid.

absoludity:
> In any case, it should be easy to validate in the client side that at least the URL is valid.

I don't see much point in just verifying a valid URL, but could the chart-svc (as I see we post directly to the k8s API when creating/refreshing/deleting app repos) have an endpoint to validate chart repo URL+credentials/certs? This is another point where, if we had a backend receiving the request to create the app repository, we could do other things, like validate the request before adding it as a k8s resource.

andresmgot:
Yes, the problem is that right now we don't have the proper service to do so. We would need to use either the chartsvc or tiller-proxy but it's doable.

absoludity:
[...]

PR #1534, opened 2020-02-21
Allow to validate apprepositories

andresmgot:
### Description of the change

Add an endpoint in the backed to validate a given AppRepository. This has been implemented as a different endpoint than the "Create" one to make it optional so users can choose to ignore errors if returned.

### Possible drawbacks

If the repository requires to use a proxy setting some environment variables, that's still not supported.

### Applicable issues

  - Ref: #1330

### Additional information

I couldn't find a good way of testing the HTTP request but tested IRL.

It was necessary to move some code from the `chart` package to `kube`.

### DIFF
--- a/pkg/http-handler/http-handler.go
+++ b/pkg/http-handler/http-handler.go
@@ -18,6 +18,7 @@ package httphandler
 
 import (
 	"encoding/json"
+	"io/ioutil"
 	"net/http"
 	"os"
 
@@ -74,6 +75,25 @@ func CreateAppRepository(handler kube.AuthHandler) func(w http.ResponseWriter, r
 	}
 }
 
+// ValidateAppRepository returns a 200 if the connection to the AppRepo can be established
+func ValidateAppRepository(handler kube.AuthHandler) func(w http.ResponseWriter, req *http.Request) {
+	return func(w http.ResponseWriter, req *http.Request) {
+		token := auth.ExtractToken(req.Header.Get("Authorization"))
+		res, err := handler.AsUser(token).ValidateAppRepository(req.Body)
+		if err != nil {
+			returnK8sError(err, w)
+			return
+		}
+		body, err := ioutil.ReadAll(res.Body)
+		if err != nil {
+			returnK8sError(err, w)
+			return
+		}
+		w.WriteHeader(res.StatusCode)
+		w.Write(body)
+	}
+}
+
 // DeleteAppRepository deletes an App Repository
 func DeleteAppRepository(kubeHandler kube.AuthHandler) func(w http.ResponseWriter, req *http.Request) {
 	return func(w http.ResponseWriter, req *http.Request) {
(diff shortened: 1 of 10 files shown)


pkg/kube/http_utils.go
```diff
@@ -0,0 +1,102 @@
+/*
```
andresmgot: this code and the tests have been moved from the `chart` pkg

pkg/http-handler/http-handler.go
```diff
@@ -74,6 +75,25 @@ func CreateAppRepository(handler kube.AuthHandler) func(w http.ResponseWriter, r
 	}
 }
 
+// ValidateAppRepository returns a 200 if the connection to the AppRepo can be established
+func ValidateAppRepository(handler kube.AuthHandler) func(w http.ResponseWriter, req *http.Request) {
```
absoludity: I think you should be able to add unit tests for this handler by creating an `server := httptest.NewServer(...)` that just returns the index.yaml. Then your test just uses `server.URL` of the test server in the request body for the repo URL?

The test here only needs to verify the 200 response (and response body) when the test server also responds with a 200, and the error response when it does not.

pkg/kube/kube_handler.go
```diff
@@ -278,6 +287,24 @@ func (a *userHandler) DeleteAppRepository(repoName, repoNamespace string) error
 	return err
 }
 
+func (a *userHandler) ValidateAppRepository(appRepoBody io.ReadCloser) (*http.Response, error) {
```
absoludity: A similar setup here would allow testing the URL path trimming - not sure it's necessary though if the handler above is tested.

andresmgot:
For the record, I found an interesting bug. The new API endpoint to create AppRepositories create secrets with data stored in the key `StringData`:

https://github.com/kubeapps/kubeapps/blob/master/pkg/kube/kube_handler.go#L358

While we were just reading `Data` (on its byte form):

https://github.com/kubeapps/kubeapps/blob/master/pkg/chart/chart.go#L340

This inconsistency means that the setup is currently not working. I am not sure if we where storing the data as bytes or strings with the dashboard so I added support for both.

absoludity:
> For the record, I found an interesting bug. The new API endpoint to create AppRepositories create secrets with data stored in the key `StringData`:
> 
> https://github.com/kubeapps/kubeapps/blob/master/pkg/kube/kube_handler.go#L358
> 
> While we were just reading `Data` (on its byte form):
> 
> https://github.com/kubeapps/kubeapps/blob/master/pkg/chart/chart.go#L340
> 
> This inconsistency means that the setup is currently not working (unless I am missing something). I am not sure if we were storing the data as bytes or strings with the dashboard so I added support for both.

`StringData` is just a convenience field... the data is still stored base64 encoded in `data`:
```
$ cat <<EOF | kubectl apply -f -
> apiVersion: v1
> kind: Secret
> metadata:
>   name: my-test-secret
> type: Opaque
> stringData:
>   authorizationHeader: should-be-encoded
>   ca.crt: should-also-be-encoded
> EOF
secret/my-test-secret created

$ kubectl get secret my-test-secret -o yaml
apiVersion: v1
data:
  authorizationHeader: c2hvdWxkLWJlLWVuY29kZWQ=
  ca.crt: c2hvdWxkLWFsc28tYmUtZW5jb2RlZA==
kind: Secret
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"v1","kind":"Secret","metadata":{"annotations":{},"name":"my-test-secret","namespace":"default"},"stringData":{"authorizationHeader":"should-be-encoded","ca.crt":"should-also-be-encoded"},"type":"Opaque"}
  creationTimestamp: "2020-02-24T23:11:54Z"
  name: my-test-secret
  namespace: default
  resourceVersion: "887"
  selfLink: /api/v1/namespaces/default/secrets/my-test-secret
  uid: 31d7c322-d0bb-47f1-89f0-27f7b23a0102
type: Opaque
```
More info: https://kubernetes.io/docs/concepts/configuration/secret/#creating-a-secret-manually

This can be confusing in tests, because the client-go test fakes don't do the translation. Let me know if that was the confusion, or if you still think there is a bug?

andresmgot:
> This can be confusing in tests, because the client-go test fakes don't do the translation. Let me know if that was the confusion, or if you still think there is a bug?

that makes sense. Yes, that's what confused me. Thanks for clarifying

PR #1535, opened 2020-02-24
Add pg tests for assetsvc

absoludity:
Some background work so that the asset service can also be tested against pg.

Some comments inline.

### DIFF
--- a/cmd/assetsvc/postgres_db_test.go
+++ b/cmd/assetsvc/postgres_db_test.go
@@ -121,21 +121,6 @@ func TestGetChart(t *testing.T) {
 			expectedErr:   nil,
 			expectedChart: "my-chart",
 		},
-		// {
-		// 	name: "it returns the chart matching the chartid in the specific namespace",
-		// 	existingCharts: map[string][]models.Chart{
-		// 		"namespace-1": []models.Chart{
-		// 			models.Chart{ID: "chart-1", Name: "incorrect-chart"},
-		// 		},
-		// 		"namespace-2": []models.Chart{
-		// 			models.Chart{ID: "chart-1", Name: "correct-chart"},
-		// 		},
-		// 	},
-		// 	chartId:       "chart-1",
-		// 	namespace:     "namespace-2",
-		// 	expectedErr:   nil,
-		// 	expectedChart: "correct-chart",
-		// },
 	}
 
 	for _, tc := range testCases {
@@ -206,25 +191,6 @@ func TestGetVersion(t *testing.T) {
 			requestedVersion: "1.2.3",
 			expectedVersion:  "1.2.3",
 		},
-		// {
-		// 	name: "it returns the chart version matching the chartid and version",
-		// 	existingCharts: map[string][]models.Chart{
-		// 		"other-namespace": []models.Chart{
-		// 			models.Chart{ID: "chart-1", ChartVersions: []models.ChartVersion{
-		// 				models.ChartVersion{Version: "4.5.6"},
-		// 			}},
-		// 		},
-		// 		"namespace-1": []models.Chart{
-		// 			models.Chart{ID: "chart-1", ChartVersions: []models.ChartVersion{
-		// 				models.ChartVersion{Version: "1.2.3"},
-		// 			}},
-		// 		},
-		// 	},
-		// 	chartId:          "chart-1",
-		// 	namespace:        "namespace-1",
-		// 	requestedVersion: "1.2.3",
-		// 	expectedVersion:  "1.2.3",
-		// },
 	}
 
 	for _, tc := range testCases {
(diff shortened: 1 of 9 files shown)


cmd/assetsvc/postgres_db_test.go
```diff
@@ -0,0 +1,221 @@
...
+		expectedErr      error
+	}{
+		{
+			name:        "it returns an error if the chart does not exist",
+			chartId:     "doesnt-exist-1",
+			namespace:   "doesnt-exist",
+			expectedErr: sql.ErrNoRows,
+		},
+		{
+			name: "it does not ?! return an error if the chart version does not exist",
```
andresmgot: This seems like a bug, it should indeed return a failure if the version doesn't exists

cmd/assetsvc/postgres_db_test.go
```diff
@@ -0,0 +1,221 @@
...
+
+			chart, err := pam.getChartVersion(tc.chartId, tc.requestedVersion)
+
+			if got, want := err, tc.expectedErr; got != want {
+				t.Fatalf("got: %+v, want: %+v", got, want)
+			}
+			if tc.expectedErr != nil {
+				return
+			}
+			// The function just returns the chart with only the one version?
```
andresmgot: answering the comment yes, this function is meant to return a single chartVersion

pkg/chart/models/chart.go
```diff
@@ -71,9 +71,11 @@ type ChartVersion struct {
 	Created    time.Time `json:"created"`
 	Digest     string    `json:"digest"`
 	URLs       []string  `json:"urls"`
-	Readme     string    `json:"readme" bson:"-"`
-	Values     string    `json:"values" bson:"-"`
-	Schema     string    `json:"schema" bson:"-"`
+	// TODO(mnelson): Why are the following three fields repeated here, when they are
+	// already on the ChartFiles struct?
```
andresmgot: I _think_ these values are not really being used, whenever the Readme or the values are needed, the frontend does a specific request and it uses the ChartFiles instead. This was present in the code of the chartsvc and I couldn't verify that.

cmd/assetsvc/postgresql_utils_test.go
```diff
@@ -52,6 +52,22 @@ func (f *fakePGManager) QueryAllCharts(query string, args ...interface{}) ([]*mo
...
+
+func (f *fakePGManager) InitTables() error {
+	return nil
+}
+
+func (f *fakePGManager) EnsureRepoExists(namespace, name string) (int, error) {
+	return 0, nil
+}
+
+func (f *fakePGManager) GetDB() dbutils.PostgresDB {
```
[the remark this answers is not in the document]
andresmgot: mm, I don't fully understand this, rather than modifying the interface, isn't it better to just add the `DB` property to the test struct? (I am not sure if that's the problem).

It seems weird to me, in the functional code, to have `.DB` and `.GetDB` but it's true that I have seen that approach before.

cmd/assetsvc/postgres_db_test.go
absoludity: Updated to expect an error here.

cmd/assetsvc/postgres_db_test.go
absoludity: The bit that is confusing me is chartVersion == chart with only one version (as opposed to some other object). (removed the '?')

cmd/assetsvc/postgresql_utils_test.go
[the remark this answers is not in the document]
absoludity: I can't see how I can add the `DB` property to the test struct. The tests use the non-test `postgresAssetManager` which is defined in the assetsvc as:
```
type postgresAssetManager struct {
    dbutils.PostgresAssetManagerIface
}
```
so the actual `DB` is not exposed at all unless it's available as a method on that interface? (This is slightly different from how it's done in the asset-sync service, unfortunately, which wraps the actual struct rather than the interface, which exposes DB, yet I can't update this one to be the same because the existing assetsvc tests rely on wrapping a fake implementing the same interface).

pkg/chart/models/chart.go
absoludity: I tried deleting them to see what would happen: it turns out they are storing the URL paths to these items for when the handler returns a chart version (see handler.go's `chartVersionAttributes`). I've updated the comment so that's obvious to the next person.

cmd/assetsvc/postgres_db_test.go
andresmgot: yeah, not the most trivial thing. This comes from the mongodb implementation in which when you request a ChartVersion the query filtered all but the matching version (still not perfect but easier to read)

pkg/chart/models/chart.go
andresmgot: ah, true. In any case, I don't think those URLs are necessary (the path is always the same)

PR #1536, opened 2020-02-24
Update nginx image

andresmgot:
### Description of the change

Update Nginx image since debian-9 is no longer maintained,


absoludity:
So I should not have merged this, as it caused a chart release which already includes the new pre-upgrade job which causes a failure (as the code for it isn't included in 1.8.2).

andresmgot:
I would have merged it anyway, the bug was waiting to bite us :)

PR #1537, opened 2020-02-24
Allow to validate apprepositories from the dashboard

### LINKED ISSUE #1330: Validate app repositories before adding them

andresmgot:
### Description of the change

Follow up of #1534 

Validate apprepositories from the dashboard before submitting them. As explained in the other PR, this validation is optional, allowing users to skip it if they think so:

Screenshot from 2020-02-27 11-14-29

### Possible drawbacks

Changing the spec of the sync job is not supported for the validation endpoint so the button is disabled if the template gets modified.

### Applicable issues

  - fixes #1330

### DIFF
--- a/dashboard/src/actions/repos.ts
+++ b/dashboard/src/actions/repos.ts
@@ -25,6 +25,11 @@ export const receiveRepo = createAction("RECEIVE_REPO", resolve => {
   return (repo: IAppRepository) => resolve(repo);
 });
 
+export const checkRepo = createAction("CHECK_REPO");
+export const validatedRepo = createAction("VALIDATED_REPO", resolve => {
+  return (data: any) => resolve(data);
+});
+
 // Clear repo is basically receiving an empty repo
 export const clearRepo = createAction("RECEIVE_REPO", resolve => {
   return () => resolve({} as IAppRepository);
@@ -41,12 +46,15 @@ export const redirect = createAction("REDIRECT", resolve => {
 
 export const redirected = createAction("REDIRECTED");
 export const errorRepos = createAction("ERROR_REPOS", resolve => {
-  return (err: Error, op: "create" | "update" | "fetch" | "delete") => resolve({ err, op });
+  return (err: Error, op: "create" | "update" | "fetch" | "delete" | "validate") =>
+    resolve({ err, op });
 });
 
 const allActions = [
   addRepo,
   addedRepo,
+  checkRepo,
+  validatedRepo,
   clearRepo,
   errorRepos,
   requestRepos,
@@ -170,6 +178,25 @@ export const installRepo = (
       return true;
     } catch (e) {
       dispatch(errorRepos(e, "create"));
+      return false;
+    }
+  };
+};
+
+export const validateRepo = (
+  repoURL: string,
+  authHeader: string,
+  customCA: string,
+): ThunkAction<Promise<boolean>, IStoreState, null, AppReposAction> => {
+  return async dispatch => {
+    try {
+      dispatch(checkRepo());
+      const data = await AppRepository.validate(repoURL, authHeader, customCA);
+      dispatch(validatedRepo(data));
+
+      return true;
+    } catch (e) {
+      dispatch(errorRepos(e, "validate"));
       return false;
     }
   };
--- a/dashboard/src/actions/repos.ts
+++ b/dashboard/src/actions/repos.ts
@@ -25,8 +25,8 @@ export const receiveRepo = createAction("RECEIVE_REPO", resolve => {
   return (repo: IAppRepository) => resolve(repo);
 });
 
-export const checkRepo = createAction("CHECK_REPO");
-export const validatedRepo = createAction("VALIDATED_REPO", resolve => {
+export const repoValidating = createAction("REPO_VALIDATING");
+export const repoValidated = createAction("REPO_VALIDATED", resolve => {
   return (data: any) => resolve(data);
 });
 
@@ -53,8 +53,8 @@ export const errorRepos = createAction("ERROR_REPOS", resolve => {
 const allActions = [
   addRepo,
   addedRepo,
-  checkRepo,
-  validatedRepo,
+  repoValidating,
+  repoValidated,
   clearRepo,
   errorRepos,
   requestRepos,
@@ -190,11 +190,21 @@ export const validateRepo = (
 ): ThunkAction<Promise<boolean>, IStoreState, null, AppReposAction> => {
   return async dispatch => {
     try {
-      dispatch(checkRepo());
+      dispatch(repoValidating());
       const data = await AppRepository.validate(repoURL, authHeader, customCA);
-      dispatch(validatedRepo(data));
-
-      return true;
+      if (data === "OK") {
+        dispatch(repoValidated(data));
+        return true;
+      } else {
+        // Unexpected error
+        dispatch(
+          errorRepos(
+            new Error(`Unable to parse validation response, got: ${JSON.stringify(data)}`),
+            "validate",
+          ),
+        );
+        return false;
+      }
     } catch (e) {
       dispatch(errorRepos(e, "validate"));
       return false;
--- a/dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
+++ b/dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
@@ -1,8 +1,10 @@
 import * as React from "react";
 import { Redirect } from "react-router";
+import { UnexpectedErrorAlert } from "../../../components/ErrorAlert";
 import Hint from "../../../components/Hint";
 
 interface IAppRepoFormProps {
+  validateError?: Error;
   message?: string;
   redirectTo?: string;
   install: (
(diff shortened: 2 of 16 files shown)


absoludity:
> ### Possible drawbacks
> 
> Changing the spec of the sync job is not supported for the validation endpoint so the button is disabled if the template gets modified.

Before looking at the details here, I'm keen to be clear exactly what we can and cannot do, because the most obvious UX would be that we always validate the details.

You mentioned that the sync template can make it difficult/impossible to validate the repository, but it's not obvious why, nor whether this would be the main use-case.

Looking at our docs for the [sync pod template](https://github.com/kubeapps/kubeapps/blob/master/docs/user/private-app-repository.md#modifying-the-synchronization-job), I assume the issue is that in some situations the job may need to be run with certain affinity for nodes with access to the repo, but how would people then install from this repo, given that it's our backend which gets the chart (ie. so would need the same affinity). But I might be missing something?

Secondly, I suspect this situation is a small percentage of the normal use-case, and so think we'd be better off ensuring the expected UX by default, so that we always run the validation when submitting an app repository, if it fails and no sync-template is in use, we block, if it fails and a sync  template is in use, we allow submitting anyway. Does that make sense?

andresmgot:
> Looking at our docs for the [sync pod template](https://github.com/kubeapps/kubeapps/blob/master/docs/user/private-app-repository.md#modifying-the-synchronization-job), I assume the issue is that in some situations the job may need to be run with certain affinity for nodes with access to the repo, but how would people then install from this repo, given that it's our backend which gets the chart (ie. so would need the same affinity). But I might be missing something?

There are several situations in which the sync would need extra configuration:
 - If the repo is behind a proxy, some env vars may be needed (`HTTP_PROXY`, `NO_PROXY`...)
 - If the cluster is using NetworkPolicies it may be needed to run the sync job with certain labels
 - Maybe only some of the nodes has access to the repo so it's necessary to configure the affinity.

Then, once the repo is stored in the database, the backend wouldn't need access to those env vars/labels unless the chart tarballs are also under these conditions.

> Secondly, I suspect this situation is a small percentage of the normal use-case, and so think we'd be better off ensuring the expected UX by default, so that we always run the validation when submitting an app repository, if it fails and no sync-template is in use, we block, if it fails and a sync template is in use, we allow submitting anyway. Does that make sense?

I think blocking is a bit authoritative? There could be a number of reasons of why someone would like to create an AppRepository even if the validation is failing, for example if the repo is temporarily down or it has not been created yet.

Also, if we don't block if the sync template is not empty we could lead to an inconsistent behavior. Maybe the template content has nothing to do with the repo not been reachable but we are not able to know.

absoludity:
> I think blocking is a bit authoritative? There could be a number of reasons of why someone would like to create an AppRepository even if the validation is failing, for example if the repo is temporarily down or it has not been created yet.

Yeah - true, maybe not block - but we could still always do it automatically on submit and then if it fails highlight and still give the option to submit anyway (ie. an extra query/post param).

> 
> Also, if we don't block if the sync template is not empty we could lead to an inconsistent behavior. Maybe the template content has nothing to do with the repo not been reachable but we are not able to know.

Yep, ok.

> 
> What we could do to make this feature easier to use is to automatically do the check without the button 

+1 (and either just save if it passes, or highlight the validation problems with the option of saving anyway)

> and validate the app repository when the info in the form changes.

Not sure what you mean here? I wouldn't think that we'd need/want to validate it any time other than when it's submitted? Ah, or maybe you mean if the form is submitted a second time without any changes we save, otherwise validate? (not sure if it'd be easier to just allow a force save when the validation fails, like an extra arg - but whatever you think, I'll take a look in the morning :) ). Thanks!

andresmgot:
> Not sure what you mean here? I wouldn't think that we'd need/want to validate it any time other than when it's submitted? Ah, or maybe you mean if the form is submitted a second time without any changes we save, otherwise validate? (not sure if it'd be easier to just allow a force save when the validation fails, like an extra arg - but whatever you think, I'll take a look in the morning :) ). Thanks!

I thought about validating without any button. For example, when there are no keystrokes in 1-2 seconds. I am just trying to trying to figure out the friendliest approach. In any case, that would lead to unnecessary checks. 

Another option would be to disable the "save" button if it has not been validated but that could be a bit annoying if I know it will work (for example, when using auto-completion). :thinking: 

Anyway, we can do as you say and make it transparent, validating before saving the repository and allowing to force the submission if the validation fails.

absoludity:
Let me know if this is ready to review again (and if so, can you update the gif? If it's a hassle I'll just build the image and try it out).

andresmgot:
Sorry, yes, I forgot to update the screenshots. It's ready to be re-reviewed. This is the gif version:

Peek 2020-02-27 11-10

dashboard/src/shared/AppRepository.ts
```diff
@@ -47,6 +47,14 @@ export class AppRepository {
     return data;
   }
 
+  public static async validate(repoURL: string, authHeader: string, customCA: string) {
+    const { data } = await axiosWithAuth.post<ICreateAppRepositoryResponse>(
+      url.backend.apprepositories.validate(),
+      { appRepository: { name, repoURL, authHeader, customCA } },
+    );
+    return data;
```
absoludity: So I assume from code above that if this doesn't validate, it errors rather than returning a 200 with the validation failures? Right, checking the handler I see that is the case. I missed that when reviewing the backend. It means our API doesn't differentiate between its own internal errors and an actual request to the repo, nor to be able to return individual field errors later  (such as malformed URL vs missing credentials). Or as another eg, if the repo validation request to the actual repository was returned 500, we'd return that same status code as well as whatever body they send in the response, I think, rather than returning that the request was processed successfully by us, but the validation failed. Anyway, that's related to the backend implementation, not this, but see what you think.

dashboard/src/actions/repos.ts
```diff
@@ -175,6 +183,25 @@ export const installRepo = (
...
+  customCA: string,
+): ThunkAction<Promise<boolean>, IStoreState, null, AppReposAction> => {
+  return async dispatch => {
+    try {
+      dispatch(checkRepo());
+      const data = await AppRepository.validate(repoURL, authHeader, customCA);
+      dispatch(validatedRepo(data));
+
+      return true;
+    } catch (e) {
```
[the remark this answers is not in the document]
andresmgot: The thing is that `AppRepository.validate` throws an error if the validation fails. I can add a `return data === "OK"` for the sake of readability but if the call doesn't fail, it's because the validation succeeded.

dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
```diff
@@ -227,9 +232,22 @@ export class AppRepoForm extends React.Component<IAppRepoFormProps, IAppRepoForm
...
-              <button className="button button-primary" type="submit">
-                Install Repo
+              <button
+                className="button button-primary"
+                type="submit"
+                disabled={this.props.isFetching}
+              >
+                {this.props.isFetching
+                  ? "Validating..."
+                  : `Install Repo ${this.state.validated === false ? "(force)" : ""}`}
```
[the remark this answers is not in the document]
andresmgot: yes, I was using:

```js
validated = undefined // not yet validated
validated = false // validation failed, people can still force
validated = true // validation succeeded, proceed
```

It seems it's not the most readable thing though, I will rewrite it.

dashboard/src/shared/AppRepository.ts
andresmgot: I am not sure I follow, there are two cases here:
 - We are not able to reach the target with the given input. Eg the URL certificate is invalid. In this case we return a 500 because we don't know if the error is related to the service not having the default CA or it's missing a custom CA in the request.
 - We reach the target. In this case, whatever the server returns, we proxy it to the user.

absoludity: > Apart from maybe changing the error code from 500 to something else in the first case, I am not sure what could be an improvement?

Right, it's the 500 we return when we didn't really have an internal server error that makes it obvious in my pov, but more generally, our API should return meaningful status codes for it's own service. So in the latter case that you mention where we reach the target and the validation fails (say incorrect creds), then the request to perform the validation succeeded (200) and should have the results of the validation. If we return a 403/401 it implies that the request to our service failed for that reason.

I'd just expected the return status from our API to be separate from the return status of the 3rd party so that the caller can differentiate between the request to our own API failing vs the request from our service to the external URI failing. Not a big deal, it was just a surprise that I should have noticed when reviewing the backend change.

PR #1538, opened 2020-02-25
Update removeMissingCharts to remove from a namespaced repo only.

absoludity:
Ref: #1521 

Updates the `removeMissingCharts` postgres functionality to remove from the repo (which is namespaced) only.

### DIFF
--- a/cmd/asset-syncer/postgresql_utils.go
+++ b/cmd/asset-syncer/postgresql_utils.go
@@ -50,7 +50,7 @@ func newPGManager(config datastore.Config) (assetManager, error) {
 // These steps are processed in this way to ensure relevant chart data is
 // imported into the database as fast as possible. E.g. we want all icons for
 // charts before fetching readmes for each chart and version pair.
-func (m *postgresAssetManager) Sync(repo models.RepoInternal, charts []models.Chart) error {
+func (m *postgresAssetManager) Sync(repo models.Repo, charts []models.Chart) error {
 	m.InitTables()
 
 	// Ensure the repo exists so FK constraints will be met.
@@ -59,13 +59,13 @@ func (m *postgresAssetManager) Sync(repo models.RepoInternal, charts []models.Ch
 		return err
 	}
 
-	err = m.importCharts(charts, models.Repo{Namespace: repo.Namespace, Name: repo.Name})
+	err = m.importCharts(charts, repo)
 	if err != nil {
 		return err
 	}
 
 	// Remove charts no longer existing in index
-	return m.removeMissingCharts(charts)
+	return m.removeMissingCharts(repo, charts)
 }
 
 func (m *postgresAssetManager) RepoAlreadyProcessed(repoName, repoChecksum string) bool {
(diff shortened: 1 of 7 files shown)


cmd/asset-syncer/postgresql_utils.go
```diff
@@ -110,13 +110,13 @@ func (m *postgresAssetManager) importCharts(charts []models.Chart, repo models.R
...
 }
 
-func (m *postgresAssetManager) removeMissingCharts(charts []models.Chart) error {
+func (m *postgresAssetManager) removeMissingCharts(repo models.Repo, charts []models.Chart) error {
 	var chartIDs []string
 	for _, chart := range charts {
 		chartIDs = append(chartIDs, fmt.Sprintf("'%s'", chart.ID))
 	}
 	chartIDsString := strings.Join(chartIDs, ", ")
-	rows, err := m.DB.Query(fmt.Sprintf("DELETE FROM %s WHERE info ->> 'ID' NOT IN (%s) AND info -> 'repo' ->> 'name' = $1", dbutils.ChartTable, chartIDsString), charts[0].Repo.Name)
```
absoludity: When I wrote the test before this change, i found we weren't deleting the files, ever (as there was no cascade delete relationship)... TDD FTW :)

absoludity:
I'll run it locally to investigate the CI failure tomorrow.

cmd/asset-syncer/postgresql_utils.go
```diff
@@ -156,16 +156,16 @@ func (m *postgresAssetManager) filesExist(chartFilesID, digest string) bool {
 	return err == nil && hasEntries
 }
 
-func (m *postgresAssetManager) insertFiles(chartFilesID string, files models.ChartFiles) error {
+func (m *postgresAssetManager) insertFiles(chartId string, files models.ChartFiles) error {
```
andresmgot: why are you adding the chartID if it's already included in the files ID? (not sure if I am missing something)

absoludity: Do you mean that I could parse the files ID? Eg., in a current mongodb I see the following:
```
> db.charts.find({}, { _id: 1})
{ "_id" : "incubator/kafka" }
{ "_id" : "incubator/buzzfeed-sso" }
{ "_id" : "incubator/kubernetes-vault" }
...
> db.files.find({}, { _id: 1})
{ "_id" : "incubator/kubernetes-vault-0.2.1" }
{ "_id" : "incubator/buzzfeed-sso-0.1.2" }
{ "_id" : "incubator/neo4j-0.3.0" }
```
So yes, we could parse the chart name from there, but I don't want to depend on parsing or assumptions we have about what these will be (eg. `incubator/kubernetes-0-test-0.1.2`), so it seems sensible to just use the actual chartId verbatim?

andresmgot: no, I meant that I don't know why adding the chartID is necessary now when we already have an ID (filesID) but since there are a bunch of changes I guess I am not seeing something.

absoludity: Ah - in the schema itself. The reason is that when I wrote the unit tests for Delete, I saw that currently we were only deleting charts and leaving the files dangling in the file table. To fix this, rather than manually deleting the files, I added an extra foreign key on the files table using the chartID and namespace with an `ON DELETE CASCADE` so that the DB takes care of this via the schema (and the test now passes - see ) https://github.com/kubeapps/kubeapps/blob/master/cmd/asset-syncer/postgresql_db_test.go#L192

PR #1540, opened 2020-02-25
Additional info for application status

andresmgot:
### Description of the change

From the design document at https://github.com/kubeapps/kubeapps/pull/1529

Add a pie chart for the application status showing the number of pods running for the different workloads of an application:

Peek 2020-02-25 13-12

### DIFF
--- a/dashboard/src/components/ApplicationStatus/ApplicationStatus.tsx
+++ b/dashboard/src/components/ApplicationStatus/ApplicationStatus.tsx
@@ -4,8 +4,6 @@ import * as ReactTooltip from "react-tooltip";
 
 import { AlertTriangle } from "react-feather";
 import isSomeResourceLoading from "../../components/AppView/helpers";
-// import Check from "../../icons/Check";
-// import Compass from "../../icons/Compass";
 import { hapi } from "../../shared/hapi/release";
 import {
   IDaemonsetStatus,
@@ -23,7 +21,6 @@ interface IApplicationStatusProps {
   info?: hapi.release.IInfo;
   watchWorkloads: () => void;
   closeWatches: () => void;
-  skipPieChart?: boolean;
 }
 
 interface IWorkload {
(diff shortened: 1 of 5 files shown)


dashboard/src/components/ApplicationStatus/ApplicationStatus.tsx
```diff
@@ -32,7 +53,68 @@ class ApplicationStatus extends React.Component<IApplicationStatusProps> {
...
+          if (status.currentNumberScheduled) {
+            totalPods += status.currentNumberScheduled;
+          }
+          workloads = workloads.concat({
+            name: d.item.metadata.name,
+            replicas: status.currentNumberScheduled || 0,
+            readyReplicas: status.numberReady || 0,
+          });
+        }
+      });
```
absoludity: Looks like the above 3 loops could be a single function call `readyPods, totalPods, workloads = collectResources(daemonsets, readyKey='numberReady', totalKey='currentNumberScheduled')` or similar, and just adding them externally. But fine as is too.

andresmgot:
> Excellent Andres. I don't think you've changed the colours, so in your gif, why does it go from what looks like a successful colour when it's up to 3 pods, but then an error colour after 3 [1]?
> 
> I was going to say, based on your code, that perhaps we should also factor in the health checks, but perhaps it already is and hence the colour change (if the health checks aren't green). If that is the case, it'd seem odd to say Ready when it's red.
> 
> [1] All that with the caveat that I'm slightly colour blind, so am not the best judge :)

I didn't know you were a bit color blind :sweat_smile:. It goes from yellow when not all the pods are ready to green when all the pods are ready. Note that for a pod to be _ready_ the liveness and readiness probes should be passing.

I can change the green for a blue if that helps :) BTW, I am using the colors defined here:

https://design.bitnami.com/category/Foundations/Variables/index.html#Colors

andresmgot:
Changed the color to "Secondary" (blue) when ready and "Action" (orange) when not yet ready, check if it's any better:

Peek 2020-02-26 16-31

absoludity:
Heh - thanks Andres. Those colours are much more obvious to my eyes, but I meant that my eyes are not a good reference. Please land with whichever is more obvious to your eyes :)

PR #1542, opened 2020-02-25
Hide pre-upgrade job unless feature flag set.

absoludity:
I had landed the pre-upgrade template *after* the most recent kubeapps release (1.8.2) thinking that it was safe, as the next chart release would include support for the cache invalidation, but this was not a safe assumption, since we had a chart release on Monday to update the nginx image in #1536 . Since this release included the pre-upgrade template which calls a new command to invalidate cache, the upgrade fails as we have not yet released a kubeapps version including the cache invalidation (see https://kubernetes.slack.com/archives/C9D3TSUG4/p1582667703003300)

I'll land to get the release out.

I'm also keen to update our CI so that on a chart bump, we actually test the upgrade from the previous chart version and don't release.

### DIFF
--- a/chart/kubeapps/Chart.yaml
+++ b/chart/kubeapps/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kubeapps
-version: 3.4.2
+version: 3.4.3
 appVersion: DEVEL
 description: Kubeapps is a dashboard for your Kubernetes cluster that makes it easy to deploy and manage applications in your cluster using Helm
 icon: https://raw.githubusercontent.com/kubeapps/kubeapps/master/docs/img/logo.png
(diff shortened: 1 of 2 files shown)


andresmgot:
Thanks for quickly fixing the issue!

> I'm also keen to update our CI so that on a chart bump, we actually test the upgrade from the previous chart version and don't release.

I agree, we could open an issue for this. We could at least check that for minor and patch versions the upgrade works as expected.

davidkarlsen:
Great - just rolled it out - works as designed.
Thanks!

absoludity:
> Thanks for quickly fixing the issue!

Np, just note that we need to remove the condition as when we're releasing the next kubeapps version, because this is required for the postgresql support.

> 
> > I'm also keen to update our CI so that on a chart bump, we actually test the upgrade from the previous chart version and don't release.
> 
> I agree, we could open an issue for this. We could at least check that for minor and patch versions the upgrade works as expected.

https://github.com/kubeapps/kubeapps/issues/1544

PR #1545, opened 2020-02-27
Setup test infra for mongo tests.

absoludity:
Ref: #1521 

This PR just refactors a little so that the pg and mongo test helpers can live separately (and so be namespaced with the same function names), and adds an initial test using mongodb.

Following PR starts updating the mongo backend safely with namespaces.


absoludity:
> I wouldn't spend too much time in MongoDB testing, I guess we will be deprecating it soon.

Yeah, I was going to ask whether you thought it'd be deprecated or left. I wasn't sure if I'd be needing to support both mongo and postgres when later doing the changes for asset sync jobs in namespaces, if so it'd be worth investing, but if not then I'm keen to avoid mongo as much as possible :)

PR #1551, opened 2020-03-03
Ensure charts imported per namespace when using MongoDB.

absoludity:
Ref #1521 . Requires https://github.com/kubeapps/common/pull/14 to land before landing.

There are three main changes here:

1. Implement `InvalidateCache` for mongodb so that we can ensure on upgrade that the new data and index is used (index added to ensure data consistency for chart id + repo). As a bonus this makes it easier to test (by invalidating the cache before a test etc.)
2. Updates the `importCharts` to import charts per namespace and test this properly.
3. Start updating other functions in the mongo asset-sync (I won't write tests for these) and the assetsvc.

Tested IRL using mongodb (and helm3 which shouldn't affect the test).


cmd/asset-syncer/mongodb_utils_test.go
```diff
@@ -109,15 +117,15 @@ func Test_repoAlreadyProcessed(t *testing.T) {
 	}
 }
 
-func Test_updateLastCheck(t *testing.T) {
+func _Test_updateLastCheck(t *testing.T) {
```
andresmgot: why the change in the name?

pkg/chart/models/chart.go
```diff
@@ -43,7 +43,7 @@ type RepoInternal struct {
 
 // Chart is a higher-level representation of a chart package
 type Chart struct {
-	ID              string             `json:"ID" bson:"_id"`
+	ID              string             `json:"ID" bson:"ID"`
```
andresmgot: `bson:"id"`? (even though I think this is not case sensitive) but as I said before, using just `id` may be tricky, I would change it more

andresmgot:
> I had issues updating the mock tests... let me know if there's something obvious (the failure just says that the expectation wasn't met, but not what the difference between the expectation and reality was, nor was I able to see something by logging the values), so for now I've commented them out.

I don't see anything obvious, probably it's due to the change on the method (Upsert vs UpsertId) and the change in the parameters.

cmd/asset-syncer/mongodb_utils.go
```diff
@@ -50,69 +47,76 @@ func newMongoDBManager(config datastore.Config) assetManager {
...
 }
 
-func (m *mongodbAssetManager) importCharts(charts []models.Chart) error {
+func (m *mongodbAssetManager) importCharts(charts []models.Chart, repo models.Repo) error {
 	var pairs []interface{}
 	var chartIDs []string
 	for _, c := range charts {
+		if c.Repo == nil || c.Repo.Namespace != repo.Namespace || c.Repo.Name != repo.Name {
+			return fmt.Errorf("%w: chart repo: %+v, import repo: %+v", ErrRepoMismatch, c.Repo, repo)
+		}
```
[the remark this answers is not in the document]
andresmgot: > Do you think it's worth following up with that?

If you already have that in mind it's okay. It's just a bit weird to read but it's not hurting that much.

cmd/asset-syncer/mongodb_utils_test.go
```diff
@@ -53,21 +58,26 @@ func Test_importCharts(t *testing.T) {
...
 	})
 	m.On("RemoveAll", bson.M{
-		"_id": "test",
+		"name":      repo.Name,
+		"namespace": repo.Namespace,
 	})
 	manager := getMockManager(m)
-	err := manager.Delete(models.Repo{Name: "test"})
+	err := manager.Delete(repo)
+	// err := manager.UpdateLastCheck(repoNamespace, repoName, checksum, now)
```
andresmgot: _debug_

go.mod
```diff
@@ -38,3 +38,5 @@ require (
 	k8s.io/helm v2.16.0+incompatible
 	sigs.k8s.io/yaml v1.1.0
 )
+
+replace github.com/kubeapps/common => github.com/kubeapps/common v0.0.0-20200304054735-9a4709593163
```
andresmgot: is this line needed? Can't you just set this version?

andresmgot:
Checking the CI, it seems that you've hit an ugly error. With these changes, the chart sync is much more slower (after 15min, the sync is still going on). The problem is worse because at some point, a second sync is triggered (because the 10min interval) so it's even slower.

My assumption is that now, every time we need to insert a chart version, we are upserting based in the chart_id, repo and namespace so mongoDB needs to compare those three fields instead than just checking the ID but it could be something else.

absoludity:
@andresmgot Also please note that I've re-enabled the schema upgrade job which will only work with new images, so we can't bump the chart rev without first bumping the app.

Also, in cases like this, should we still have a flag so that if people do upgrade the chart but leave the app at an old version, they can switch off the schema update? (as it won't work unless using the latest yet-to-be-released kubeapps)

andresmgot:
Nice catch!

> One side-effect of this is that the helm CLI doesn't return until after the post-install job runs, but I don't think that's an issue?

Well, the application is not really usable until the populate* jobs finish so it should be fine.

> @andresmgot Also please note that I've re-enabled the schema upgrade job which will only work with new images, so we can't bump the chart rev without first bumping the app.

Noted

> Also, in cases like this, should we still have a flag so that if people do upgrade the chart but leave the app at an old version, they can switch off the schema update? (as it won't work unless using the latest yet-to-be-released kubeapps)

I still think it would be better to just not do the schema update (that deletes all the stored items) if the schema hasn't changed. It makes the upgrade unnecessary slow in most of the updates.

absoludity:
> Yes, changes in the chart should be backward compatible until there is a new version released to avoid blocking releases of the chart.

Yes, but it's more than that - it a change in the chart which is required for a change in the app. Ah - I've an idea - we can ensure the pre-upgrade job runs, but also allow it to fail for now (or specifically, only attempt running if the cmd-line arg is there. Let me try that.

> In any case, I still think it would be better to just not do the schema update (that deletes all the stored items) if the schema hasn't changed.

Yes, that's exactly what we can do if we had migration support here. Running it on every upgrade is specifically because we don't yet have any migration support (ie. rolling forward, rolling back). I've created #1560 to track that.

> It makes the upgrade unnecessary slow in most of the updates.

Are you talking about the apparent extra time the CLI waits to return because it's waiting for the pre-upgrade job to run? As that will still be the case with any migration.

Actually running the sync's themselves isn't so user-perceivable - and this is where migrations would be much better.

> 
> BTW, investigating what ensureIndex does I found out it's deprecated:
> https://docs.mongodb.com/manual/reference/method/db.collection.ensureIndex/

Yep, I saw it was an alias for createIndex. I'll update to that instead.

Thanks!

andresmgot:
> Are you talking about the apparent extra time the CLI waits to return because it's waiting for the pre-upgrade job to run? As that will still be the case with any migration.
> 
> Actually running the sync's themselves isn't so user-perceivable - and this is where migrations would be much better.

No, I mean that, as far as I understood, the schema "migration" actually deletes the entries in the database and triggers the re-population, isn't that correct? I meant that the time that the sync job is running, users won't be able to use the app.

Thanks for the thoughts!

absoludity:
> No, I mean that, as far as I understood, the schema "migration" actually deletes the entries in the database and triggers the re-population, isn't that correct?

Correct yes. Since there was no test infra nor migration support in our postgres implementation, I decided to invest in the test infra first and postpone adding migration support, instead just invalidating the cache (as it is effectively a cache) so that we have the same code for both postgres and mongodb (and yes, it is necessary in mongodb when we're adding unique indexes - even if you delete all the duplicates you'll then effectively be invalidating the cache).

> I meant that the time that the sync job is running, users won't be able to use the app.

Ah - no, this doesn't seem to be the case. When the sync jobs are triggered, the charts import within a second or two. All the files do take longer, so depending whether the files for the chart you look at right after an upgrade happen to be sync'd yet, the chart will not display. This is the same behaviour when adding a new repo and viewing the charts straight away.

Or am I missing something?

I'm keen for us to add migration support, just not sure on the priority right now? It may make sense to do this as soon as we remove mongodb as an option for the assets? Let me know if I've missed something.

andresmgot:
> I'm keen for us to add migration support, just not sure on the priority right now? It may make sense to do this as soon as we remove mongodb as an option for the assets? Let me know if I've missed something.

No, I don't think we need to spend extra time with that since the schema is something that we change rarely. I think it's just fine if we don't invalidate the cache if there are no changes to avoid that minor "downtime" with every release.

Issue #1552, opened 2020-03-03
[PoC] Evaluate Operators support


andresmgot:
### Description:

The goal of this issue is to kick of the support for Operators in Kubeapps doing a Proof of Concept and evaluate the different requirements.

andresmgot:
Closing this as done! Operators are available in the version v1.9.0

PR #1553, opened 2020-03-03
Operators PoC document

andresmgot:
### Description of the change

Design doc for #1552 

This document describe the target UX and some implementation details to start working in the support for Operators in Kubeapps.


docs/architecture/design-proposals/operators-support-poc.md
```diff
@@ -0,0 +1,111 @@
...
+
+There are two types of Operators: the ones compatible just with Openshift and the ones compatible with any Kubernetes distribution. Both lists are published in GitHub. This is the list of Operators compatible with vanilla Kubernetes distributions: https://github.com/operator-framework/community-operators/tree/master/upstream-community-operators
+
+## PoC User Experience
+
+In this section, we will define the UX for a user that explores and deploys an Operator. In order to simplify the solution, a working environment with `kubectl` will be needed.
+
+*NOTE*: There are several steps in the management of Operators that require cluster-admin privileges. For this first implementation, we will assume that the persona using Kubeapps is a cluster administrator with admin permissions.
+
+1. The user goes to the Operators section. Since they don't have the OLM installed, they get an error:
```
absoludity: The "Operators section" is confusing from a UX POV for me, in that it is initially used here to help admins *install* operators to be available (which then appear in the unified catalog - great), as well as later display the running operator instances for the user (why is this not unified in the applications?)

If the admin option to install operators was available as "Configuration -> Install Operators", which displayed a page showing the same error if OLM is not installed, handled points 3-7 for the admin role of installing the operators etc., and points 8 and 9 remain exactly as you have them, so once an admin had installed the operator, the "operator instance types" would be displayed in the normal catalogue.

For point 10, it's not clear to me why the running operator instances can't be displayed in the app view like other apps, so that we wouldn't need the extra "Operators" link/view at all (in the same way that we don't have a link/view for running charts).

Keen to hear more during the standup, perhaps.

andresmgot: So this approach comes from the current implementation for the service catalog in which the service instances have their own view. It's true that this may not be necessary as far as we clearly differentiate between releases (charts) and operator instances in the app view. We could even merge the service instance(s) view in this.

Let me update the doc with this.

PR #1555, opened 2020-03-03
Add skeleton to include operators in the dashboard

andresmgot:
### Description of the change

Feature flag and dummy components for Operator views explained in #1553.


andresmgot:
> Looks fine, but I'd personally just start with the OperatorsListContainer (if that's the one which will be doing the admin functionality of displaying the operators available for an admin to install on the cluster). But you might have other ideas :)

Well I wanted to show the different views that I explained in #1553 and I believed will be necessary for a minimum end-to-end experience. In any case yes, I agree that the OperatorsListContainer will be the first to be implemented.

PR #1559, opened 2020-03-05
Add operator list

andresmgot:
### Description of the change

Operator List view obtained checking the CRD `packagemanifests.packages.operators.coreos.com`:

Screenshot from 2020-03-06 12-02-29

The tags are the operator provider and the category.

Still needs to write more tests but sending this already to get an early review :)

### Applicable issues

  - ref #1552

### DIFF
--- a/dashboard/src/components/OperatorList/OperatorList.tsx
+++ b/dashboard/src/components/OperatorList/OperatorList.tsx
@@ -1,5 +1,9 @@
 import * as React from "react";
 
+import { IPackageManifest } from "shared/types";
+import { CardGrid } from "../Card";
+import { ErrorSelector } from "../ErrorAlert";
+import InfoCard from "../InfoCard";
 import LoadingWrapper from "../LoadingWrapper";
 import PageHeader from "../PageHeader";
 import OLMNotFound from "./OLMNotFound";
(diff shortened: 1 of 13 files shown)


absoludity:
Looking great! RE:

> The tags are the operator provider and the category.

How can we manage the different sizes depending on the number of tags? Maybe we'll need to watch the overflow to avoid the issue in your image where chart tile is bigger?

dashboard/src/components/OperatorList/OperatorList.tsx
```diff
@@ -8,28 +12,63 @@ export interface IOperatorListProps {
...
+    return (
+      <CardGrid>
+        {this.props.operators.map(operator => {
+          return (
+            <InfoCard
+              key={operator.metadata.name}
+              link={`/operators/ns/${this.props.namespace}/${operator.metadata.name}`}
+              title={operator.metadata.name}
+              // TODO(andresmgot): Icons are protected by RBAC so we need to use the backend to retrieve those
+              // icon={TBD}
```
absoludity: Won't the RBAC be sufficient when logged in as an admin (ie. for demos) until we also store these in the backend?

andresmgot:
> How can we manage the different sizes depending on the number of tags? Maybe we'll need to watch the overflow to avoid the issue in your image where chart tile is bigger?

It seems that the current limit (150px) is not enough. We also need to be careful because setting the limit too low it would unnecessary shorten tags that fits in one line (because the sum between a short and a long tag fits in a row). In any case, I have reduced it to 120px

dashboard/src/components/OperatorList/OperatorList.tsx
andresmgot: yes, the problem is that when using `` the img request doesn't have the authorization header. What we currently do for chart icons is to skip the authorization for logos, we need to do the same for these.

absoludity: The request that gets to the k8s API does have the authorization header if you're using the auth proxy, as cookies get sent with any request. But yes, if you're using the token authentication then it won't. Bummer - the extra complexity to support the token auth is growing.

PR #1561, opened 2020-03-06
Ensure invalidate-cache doesn't run by default.

absoludity:
Since we are going to release a new chart without a new kubeapps version, we need to ensure the cache invalidation is not run (since the subcommand is not available for current kubeapps).

At the same time, we want to ensure that it can be run optionally for dev/CI while testing current master, as it is required for the mongodb namespace work.

I had originally wanted to update the command run so it is run conditionally on whether the binary supports it, but since the container image is build from scratch, I don't have any shell to work with.

### DIFF
--- a/chart/kubeapps/templates/apprepository-jobs-preupgrade.yaml
+++ b/chart/kubeapps/templates/apprepository-jobs-preupgrade.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.featureFlags.invalidateCache }}
 # Ensure db indexes are set and invalidate the chart cache during both install and upgrade.
 apiVersion: batch/v1
 kind: Job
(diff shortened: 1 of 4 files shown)


absoludity:
I did also look at using the templating function `semverCompare` which worked for defined inputs (ie. to only include the job if the `syncImage.tag` was greater than `1.8.3`, but it seems you have to be extremely careful because "If the parser fails, it will cause template execution to halt with an error." ([this is mentioned on the docs of the library used, sprig](http://masterminds.github.io/sprig/semver.html), for a more generic function, but in practise it seems to also be the case for `semverCompare` (I get no output at all from `helm template` if I do `semverCompare "1.8.3" "devel"). We could be careful and only compare if the tag matches a regex, but not sure it's a safe way forward. What do you think?

PR #1566, opened 2020-03-06
Use backend to retrieve operator icons

andresmgot:
### Description of the change

Follow up of #1559 to retrieve operator icons. The backend is the one contacting the K8s API in order to get these resources using its serviceaccount. Since it's necessary to get icons from operators in any namespace this needs to be a ClusterRole/ClusterRoleBinding.

Screenshot from 2020-03-06 18-16-13

### Applicable issues

  - ref #1552

### DIFF
--- a/chart/kubeapps/templates/kubeops-rbac.yaml
+++ b/chart/kubeapps/templates/kubeops-rbac.yaml
@@ -79,5 +79,42 @@ subjects:
     name: {{ template "kubeapps.kubeops.fullname" . }}
     namespace: {{ .Release.Namespace }}
 {{- end -}}
+{{- if .Values.featureFlags.operators }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: "kubeapps:controller:kubeops-operators-{{ .Release.Namespace }}"
+  labels:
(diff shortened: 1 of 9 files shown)


chart/kubeapps/templates/kubeops-rbac.yaml
```diff
@@ -79,5 +79,42 @@ subjects:
     name: {{ template "kubeapps.kubeops.fullname" . }}
     namespace: {{ .Release.Namespace }}
 {{- end -}}
+{{- if .Values.featureFlags.operators }}
```
absoludity: Do we need the same RBAC for tiller-proxy? Guess we can just limit this to helm3 for now anyway?

Also, is this RBAC is only required now because we're directly querying the operators? (as opposed to querying the assetsvc at a later date).

andresmgot:
> It's a bit annoying that we only need this because we are supporting token authz (I think?), otherwise the request for the img would just use the cookie which auth-proxy translates into the required authz header.

That's probably correct (I have my env working with tokens right now).

chart/kubeapps/templates/kubeops-rbac.yaml
andresmgot: true, I forgot to add this for tiller-proxy

Issue #1010, opened 2019-04-19
Pagination for Catalog page


prydonius:
The chartsvc supports pagination as of https://github.com/helm/monocular/releases/tag/v1.5.0, so we should implement this on the UI side of the Catalog page

absoludity:
Let's consider auto-loading on scroll too.

absoludity:
The key thing that relates this task and #2174 is, IMO, that we should be moving towards an interaction-based catalog where some initial (small) subset is presented but items are found by (backend) search and/or filtering. There's already great progress towards this in the UI, but yes, it's about time we did this in the backend :) Moving in this direction will give us a consistent way to display content from hubs in the catalog (which don't have a concept of a list of all the available charts due to the volume).

Thanks for picking it up @antgamdia .

antgamdia:
Here is an update on my initial work on that:

![Peek 2020-12-09 08-44](https://user-images.githubusercontent.com/11535726/101600868-304fa580-39fc-11eb-9400-04f9b20e232f.gif)

My goal was just to get used to the kubeapps frontend development and finish the server-side pagination logic. We should decide which UX is the best for our use case.

> Endless scrolling loads content continuously as the user scrolls down the page, eliminating the need for pagination. 
It is recommended for info with a flat structure, where each unit belongs at the same level of hierarchy and has similar chances of being interesting to users.

> This approach is not recommended for finding tasks, in which people just want to locate specific content. Also, locating a previously found item on an extremely long page is inefficient. It's much easier for people to remember that the item is on page 3 than it is to guess where the item is positioned.


I personally do think that having the endless scrolling is good for Kubeapps, since it's a good option for users just interested in discovering the catalog. On the other hand, we should still support other use cases; wherefore the server-side filtering should be improved. I think the combination of good filters (category, catalog, etc. and free text) could improve the overall UX of the Kubeapps catalog.

andresmgot:
> I personally do think that having the endless scrolling is good for Kubeapps, since it's a good option for users just interested in discovering the catalog. On the other hand, we should still support other use cases; wherefore the server-side filtering should be improved. I think the combination of good filters (category, catalog, etc. and free text) could improve the overall UX of the Kubeapps catalog.

I do agree :+1: So, you are thinking on doing server-side filtering for the left-side filters (provider, category...) and the name and also have auto-loading in the front-end right?

[...]

PR #1567, opened 2020-03-09
1521 send namespace assetsvc 2

absoludity:
Ref #1521, Follows  #1562 

Updates all other handlers *except* search in the assetsvc to use namespace. Able to TDD with postgres for some handlers, but left mongo untested (so extra eyes would be good :) ).

EDIT: I went ahead and removed the search handler and related code. There was a comment there from Andres saying it wasn't used, which I double-checked in the dashboard code. I see only search filters on the list views, rather than any reference to a separate search URI. Good thing too, because the current code looks slightly broken (in that the AND clause for repo is only AND'd with the first of the OR clauses for both postgres and mongo, I think?) and I was going to have to write unit tests to confirm then fix then also support namespace :)


cmd/assetsvc/handler.go
```diff
@@ -273,33 +273,6 @@ func listChartsWithFilters(w http.ResponseWriter, req *http.Request, params Para
...
 }
 
-// searchCharts returns the list of charts that matches the query param in any of these fields:
-//  - name
-//  - description
-//  - repository name
-//  - any keyword
-//  - any source
-//  - any maintainer name
-func searchCharts(w http.ResponseWriter, req *http.Request, params Params) {
```
andresmgot: this will be needed if we implement the pagination (#1010) but we can recover the code in that case.

PR #1569, opened 2020-03-09
Update dependencies

andresmgot:
### Description of the change

Updates for:

 - Mongodb
 - PostgreSQL
 - Nginx
 - kubectl
 - oauth2_proxy

The only major upgrade is for oauth2_proxy (the branch 4.X is no longer maintained). The only breaking change in this image is:

```
#321 Add reverse proxy boolean flag to control whether headers like X-Real-Ip are accepted.
This defaults to false. Usage behind a reverse proxy will require this flag to be set to avoid logging the reverse proxy IP address.
```

I believe that doesn't affect our use case but can you try this latest image @absoludity to be sure? (I don't have an OIDC setup right now). Thanks!

### DIFF
--- a/chart/kubeapps/Chart.yaml
+++ b/chart/kubeapps/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kubeapps
-version: 3.4.4
+version: 3.4.5
 appVersion: DEVEL
 description: Kubeapps is a dashboard for your Kubernetes cluster that makes it easy to deploy and manage applications in your cluster using Helm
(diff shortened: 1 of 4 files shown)


absoludity:
> I believe that doesn't affect our use case but can you try this latest image @absoludity to be sure? (I don't have an OIDC setup right now). Thanks!

I don't keep one setup, but it's easy enough to do so. Confirming that deploying with either mongodb or postgresql with this chart works fine (tested login, viewing apps and catalog only) :)

andresmgot:
Great, thanks!

PR #1573, opened 2020-03-10
Update the dashboard's assetsvc URIs to be namespaced.

absoludity:
Follows #1567 , Ref #1521 

Updates the frontend URIs to match the new assetsvc's namespaced URIs. Currently it just uses the config kubeapps namespace so that this can land. I'll continue by making this dependent on the selected namespace, when the feature flag is set.

### DIFF
--- a/dashboard/src/shared/Chart.ts
+++ b/dashboard/src/shared/Chart.ts
@@ -38,15 +38,6 @@ export default class Chart {
     return data;
   }
 
-  public static async exists(id: string, version: string, repo: string) {
-    try {
-      await axiosWithAuth.get(URL.api.charts.getVersion(`${repo}/${id}`, version));
-    } catch (e) {
-      return false;
-    }
-    return true;
-  }
-
   public static async listWithFilters(name: string, version: string, appVersion: string) {
     const url = `${Chart.APIEndpoint}/charts?name=${name}&version=${encodeURIComponent(
       version,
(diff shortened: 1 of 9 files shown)


dashboard/src/shared/Chart.ts
```diff
@@ -4,49 +4,40 @@ import { IChart, IChartVersion } from "./types";
...
-  public static async exists(id: string, version: string, repo: string) {
-    try {
-      await axiosWithAuth.get(URL.api.charts.getVersion(`${repo}/${id}`, version));
-    } catch (e) {
-      return false;
-    }
-    return true;
-  }
-
   public static async listWithFilters(name: string, version: string, appVersion: string) {
```
andresmgot: this function also requires to get the namespace, no?

dashboard/src/shared/Chart.ts
```diff
@@ -4,49 +4,40 @@ import { IChart, IChartVersion } from "./types";
...
   }
 
-  public static async getSchema(id: string, version: string) {
-    const { data } = await axiosWithAuth.get<JSONSchema4>(URL.api.charts.getSchema(id, version));
+  public static async getSchema(namespace: string, id: string, version: string) {
+    const { data } = await axiosWithAuth.get<JSONSchema4>(URL.api.charts.getSchema(namespace, id, version));
     return data;
   }
 
-  public static async exists(id: string, version: string, repo: string) {
```
andresmgot: why did you delete this function?
[edit] I see, apparently it's not being used

dashboard/src/shared/Chart.ts
absoludity: Indeed, updated, thanks.

PR #1574, opened 2020-03-10
Return namespaced self-links in responses.

absoludity:
While testing #1573 I found the icons weren't rendering because the frontend uses the icon url in the response [1], which hadn't been updated to be namespaced. Turns out there were a few others that I've also updated.

[1] Oddly, we don't have the dashboard using the similar links for the readme or schema.

Ref: #1521


andresmgot:
> [1] Oddly, we don't have the dashboard using the similar links for the readme or schema.

TBH I don't think it's needed for the logo either since the path is always the same

absoludity:
> > [1] Oddly, we don't have the dashboard using the similar links for the readme or schema.
> 
> TBH I don't think it's needed for the logo either since the path is always the same

Not sure I agree. They're resources on an API, which should (and currently do) provide URIs for sub-resources so clients don't need to know or care about the path. But it matters not, we're not changing anything here :)

PR #1576, opened 2020-03-10
Operator View

### LINKED ISSUE #1552: [PoC] Evaluate Operators support

andresmgot:
### Description of the change

Operator View. All the information can be extracted from the current definition of a `PackageManifest`:

Screenshot from 2020-03-10 11-04-21

### Applicable issues

  - ref #1552

### DIFF
--- a/dashboard/src/components/ChartView/ChartReadme.scss
+++ b/dashboard/src/components/ChartView/ChartReadme.scss
@@ -32,3 +32,7 @@
 .ChartReadme tr {
   border-bottom: 1px solid #eee;
 }
+
+.ChartReadme img {
+  max-width: 100%;
+}
--- a/dashboard/src/components/OperatorView/OperatorCapabilityLevel.test.tsx
+++ b/dashboard/src/components/OperatorView/OperatorCapabilityLevel.test.tsx
@@ -1,0 +1,29 @@
+import { shallow } from "enzyme";
+import * as React from "react";
+import CapabiliyLevel from "./OperatorCapabilityLevel";
+
+[
+  {
+    name: "basic install level",
+    expectedLevels: 1,
+    level: "Basic Install",
+  },
+  {
+    name: "auto pilot level",
+    expectedLevels: 5,
+    level: "Auto Pilot",
+  },
+  {
+    name: "unknown level",
+    expectedLevels: 0,
+    level: "Foo",
+  },
+].forEach(t => {
+  it(t.name, () => {
+    const wrapper = shallow(<CapabiliyLevel level={t.level} />);
+    const satisfiedLevels = wrapper
+      .find(".capabilityLevelIcon")
+      .filterWhere(s => s.prop("stroke") === "#1598CB");
+    expect(t.expectedLevels).toEqual(satisfiedLevels.length);
+  });
+});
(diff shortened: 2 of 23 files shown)


dashboard/src/components/ChartView/ChartReadme.scss
```diff
@@ -32,3 +32,7 @@
 .ChartReadme tr {
   border-bottom: 1px solid #eee;
 }
+
+.ChartReadme img {
+  max-width: 100%;
```
absoludity: Guessing you've verified this doesn't affect existing chart views?

dashboard/src/components/OperatorView/OperatorCapabilityLevel.test.tsx
```diff
@@ -0,0 +1,29 @@
+import { shallow } from "enzyme";
+import * as React from "react";
+import CapabiliyLevel from "./OperatorCapabilityLevel";
+
+[
+  {
+    name: "basic install level",
+    expectedLevels: 1,
+    level: "Basic Install",
```
absoludity: I was surprised that these capability level's are not identifiers (but checked upstream and they're not). But perhaps they should at least be constants in the module?

dashboard/src/components/OperatorView/OperatorCapabilityLevel.tsx
```diff
@@ -0,0 +1,37 @@
...
+  level: string;
+}
+
+const levels = {
+  "Basic Install": 1,
+  "Seamless Upgrades": 2,
+  "Full Lifecycle": 3,
+  "Deep Insights": 4,
+  "Auto Pilot": 5,
+};
```
absoludity: Perhaps something like:
```
const BASIC_INSTALL = "Basic Install";
...
const levels = {
  BASIC_INSTALL: 1,
  ...
};
```
so the keys are defined once here? (and can be referenced in tests as per my other comment).
```

File: dashboard/src/components/OperatorView/OperatorDescription.test.tsx [hunk 4]
```diff
@@ -0,0 +1,11 @@
+import { shallow } from "enzyme";
+import * as React from "react";
+import * as ReactMarkdown from "react-markdown";
+import OperatorDescription from "./OperatorDescription";
+
+it("renders a description", () => {
+  const wrapper = shallow(<OperatorDescription description="# Title!\n" />);
+  const markdown = wrapper.find(ReactMarkdown);
+  expect(markdown).toExist();
+  expect(markdown.html()).toContain("Title!");
```
absoludity: Without context, isn't it `<h1>Title!</h1>` or something that we should be checking here?

dashboard/src/components/ChartView/ChartReadme.scss
andresmgot: it does affect charts as well, the thing is that since it's not common to have images in those chart readmes I have never seen an image escaping the component boundaries. With this change it should be fine.

dashboard/src/components/OperatorView/OperatorDescription.test.tsx
andresmgot: well, yes, the goal of this component is not to test the `Markdown` component but it's more complete that way.

PR #1578, opened 2020-03-10
Add installation instructions for operators

andresmgot:
### Description of the change

Add installation instructions for Operators. This is a provisional approach while we develop a proper form to create these Operators through Kubeapps. In this case, we point to the publicly available YAML that the operator hub provides:

Screenshot from 2020-03-10 16-27-05

There are two types of operators: namespaced or non-namespaced (global). This information is available in the channels that the PackageManifest specifies. The manifest published in the OperatorHub refers to the default channel so we need to check that one. 

Namespaced operators are installed in a namespace called `my-<operator>` while global are installed in the `operators` namespace.

### Applicable issues

  - ref #1552

### DIFF
--- a/dashboard/src/components/OperatorView/OperatorInstallation.tsx
+++ b/dashboard/src/components/OperatorView/OperatorInstallation.tsx
@@ -1,0 +1,68 @@
+import * as React from "react";
+import { X } from "react-feather";
+import * as Modal from "react-modal";
+
+// TODO(andresmgot) This is a temporary component
+// A form should be rendered instead asking for the operator
+// details and creating a Subscription
+
+interface IOperatorInstallationProps {
+  name: string;
+  namespaced: boolean;
+  closeModal: () => void;
+  modalIsOpen: boolean;
+}
+
+class OperatorInstallation extends React.Component<IOperatorInstallationProps> {
+  public render() {
+    return (
+      <Modal
+        className="centered-modal"
+        isOpen={this.props.modalIsOpen}
+        onRequestClose={this.props.closeModal}
+        contentLabel="Modal"
+      >
+        <div className="container">
+          <div className="row">
+            <div className="col-10">
+              <h5>Install {this.props.name}</h5>
+            </div>
+            <div className="col-2 text-r">
+              <a onClick={this.props.closeModal} style={{ color: "inherit" }}>
+                <X />
+              </a>
+            </div>
+          </div>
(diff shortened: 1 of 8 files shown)


dashboard/src/components/OperatorView/OperatorInstallation.tsx
```diff
@@ -0,0 +1,68 @@
...
+            </div>
+            <div className="Terminal__Tab">
+              <pre className="Terminal__Code">
+                <code>kubectl create -f https://operatorhub.io/install/{this.props.name}.yaml</code>
+              </pre>
+            </div>
+          </section>
+          {this.props.namespaced ? (
+            <span>
+              This Operator will be installed in the <code>my-{this.props.name}</code> namespace and
```
absoludity: Strange that this isn't configurable? (I tried using qparams on the URL too).

dashboard/src/components/OperatorView/OperatorView.test.tsx
```diff
@@ -40,32 +71,20 @@ it("shows an error if the operator doesn't have any channel defined", () => {
...
-      ],
+      ...defaultOperator.status,
+      channels: [{ name: "alpha" }, defaultOperator.status.channels[0]],
     },
   };
   const wrapper = shallow(<OperatorView {...defaultProps} operator={operator as any} />);
-  expect(wrapper).toMatchSnapshot();
+  expect(wrapper.find(OperatorDescription).prop("description")).toEqual(
+    "this is a testing operator",
+  );
```
absoludity: Shouldn't something here be checking the channel? (ie. that the default channel is still beta? Or is the test name inaccurate?)

dashboard/src/components/OperatorView/OperatorInstallation.tsx
andresmgot: you mean the target namespace? Yes, it's kind of a convention to make the public yaml easy to create/use (you can check for example https://operatorhub.io/install/etcd.yaml

Once we change the approach to have a proper form, that namespace will be configurable

dashboard/src/components/OperatorView/OperatorView.test.tsx
andresmgot: it's indirectly checking the channel, here the default channel is `beta` even though it's not the first one. Since it's getting the right description (the one for the beta channel), it's choosing the right channel.

PR #1575, opened 2020-03-10
Avoid invalidate cache flag while testing the latest release

andresmgot:
### Description of the change

Fixes master builds disabling the invalidateCache flag while testing the latest release:  https://circleci.com/gh/kubeapps/kubeapps/18448

PR #1580, opened 2020-03-11
Update assetsvc chart test for namespaced URIs

absoludity:
Last change for namespaced assetsvc to pass on CI, but I'm not landing this as it also would mean we can't release the chart for a previous version.

Other thoughts I had:
 * Temporarily update the assetsvc to still respond for non-namespaced requests: this would turn in to a security issue later when we check to ensure the requestor has perms, also quite a bit of code just to work around a CI issue.
 * Update the test URI to something like the live probe: fails to be a useful test at that point
 * I don't think we can choose the URL conditionally because (a) it's going to depend on which app version is installed, and (b) helm template tags for testing semvers is dangerous (exits on error).

It seems like we really should be keeping a branch for the currently released chart, and if we want to update the currently released chart we land to that branch and release, while master of the chart is always for master of the app code. Has this been talked about in the past (other than my mentioning it once before)?

### DIFF
--- a/chart/kubeapps/templates/tests/test-assetsvc.yaml
+++ b/chart/kubeapps/templates/tests/test-assetsvc.yaml
@@ -16,5 +16,5 @@ spec:
       command:
         - sh
         - -c
-        - curl -o /tmp/output $ASSETSVC_HOST:$ASSETSVC_PORT/v1/charts && cat /tmp/output && cat /tmp/output | grep wordpress
+        - curl -o /tmp/output $ASSETSVC_HOST:$ASSETSVC_PORT/v1/ns/{{ .Release.Namespace }}/charts && cat /tmp/output && cat /tmp/output | grep wordpress
   restartPolicy: Never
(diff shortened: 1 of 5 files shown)


andresmgot:
TBH these tests are not that useful now that we have proper browser tests. Maybe we can do something similar to what I did in https://github.com/kubeapps/kubeapps/pull/1575 and just not execute the chart tests if we are testing the last release. That would "fix" the CI.

Regarding a possible release, I think we are close enough to have the registries per namespace feature ready to be released so we can either merge it in master or wait until we want to release and leave it in the branch (we are not actively working in a different feature that affects the same code so you won't see conflicts).

absoludity:
> TBH these tests are not that useful now that we have proper browser tests. Maybe we can do something similar to what I did in #1575 and just not execute the chart tests if we are testing the last release. That would "fix" the CI. Then we can re-enable them after the next release.

Yeah - I'm happy with that. Added a similar condition so it's only run if testing the latest release.

> 
> Regarding a possible release, I think we are close enough to have the registries per namespace feature ready to be released so we can either merge it in master or wait until we want to release and leave it in the branch (we are not actively working in a different feature that affects the same code so you won't see conflicts).

+1 for the first option (landing it in master, so we can find and fix issues quickly, rather than waiting).

Thanks.

PR #1582, opened 2020-03-12
Show operator instance types in the catalog

andresmgot:
### Description of the change

This PR adds support for showing Operator instance types in the Catalog view along with available charts. To improve the discover-ability of those instances, there is a column filter that you can use to list only charts or operators (which is only rendered if there are operators available).

Screenshot from 2020-03-12 13-41-21

Fun fact: The icons are now encrypted in base64 in the CSV definition

### Applicable issues

  - ref #1552


dashboard/src/components/Catalog/Catalog.tsx
```diff
@@ -77,16 +97,96 @@ class Catalog extends React.Component<ICatalogProps, ICatalogState> {
...
+                  </label>
+                </div>
+                <div>
+                  <label className="checkbox" key="listoperators">
+                    <input
+                      type="checkbox"
+                      checked={listOperators}
+                      onChange={this.toggleListOperators}
+                    />
+                    <span>Operator Instances</span>
```
absoludity: We're using the technical name "operator instances" here, but I wonder if it'd be better from a user pov to just filter by "operators" (vs charts)?

dashboard/src/components/Catalog/Catalog.tsx
```diff
@@ -77,16 +97,96 @@ class Catalog extends React.Component<ICatalogProps, ICatalogState> {
...
+          icon: `data:${csv.spec.icon[0].mediatype};base64,${csv.spec.icon[0].base64data}`,
+          version: crd.version,
+          description: crd.description,
+          type: "operator",
+          csv: csv.metadata.name,
+          namespace: this.props.namespace,
+        });
+      });
+    });
+    return result;
```
absoludity: Looks like this is returning all charts followed by all operators? Should the catalog result be sorted in some way?

dashboard/src/components/Catalog/Catalog.tsx
andresmgot: Right, I used Operators first but I thougth it could be confusing with the operators view. In any case you are probably right.

dashboard/src/components/Catalog/Catalog.tsx
andresmgot: It is. Probably is more intuitive to have them shuffled and ordered alphabetically

PR #1586, opened 2020-03-16
Fix icon URL with namespace

andresmgot:
### Description of the change

Notice that the URL in tiller-proxy/kubeops had the namespace missing.


andresmgot:
> Thanks. I was wondering why this worked without issue for me, perhaps because of what I mentioned the other day related to the operator work: because I'm using the auth-proxy, the auth token is being sent with the query for logos. Not so without?

Probably yes. I was getting a 401.

PR #1589, opened 2020-03-17
Review from #1584

andresmgot:
### Description of the change

Clicked "merge" too soon in #1584 by mistake :/

### DIFF
--- a/dashboard/src/actions/operators.test.tsx
+++ b/dashboard/src/actions/operators.test.tsx
@@ -1,5 +1,6 @@
 import configureMockStore from "redux-mock-store";
 import thunk from "redux-thunk";
+import { IResource } from "shared/types";
 import { getType } from "typesafe-actions";
 import actions from ".";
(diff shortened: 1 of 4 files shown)


dashboard/src/actions/operators.test.tsx
```diff
@@ -193,14 +194,15 @@ describe("getCSV", () => {
...
+    const resource = {} as IResource;
+    Operators.createResource = jest.fn(() => resource);
     const expectedActions = [
       {
         type: getType(operatorActions.creatingResource),
       },
       {
         type: getType(operatorActions.resourceCreated),
-        payload: true,
+        payload: resource,
```
absoludity: The fact that this can change means the test isn't actually verifying the resource, I think? But fine either way.

andresmgot: yep, since the method is mocked the no validation is made

PR #1592, opened 2020-03-18
Enable authz check for specific namespace.

absoludity:
Ref #1521 This PR updates the AuthGate used by the assetsvc and tiller-proxy to verify whether the user has access to a particular namespace (or cluster-wide access to all namespaces), rather than whether they've valid credentials to authenticate.

It can't land as is (and will fail CI), as I need to update the frontend so that the namespace is always included (even if we're not running with the feature flag) so that users who do not have access to the kubeapps namespace can still view the catalog.

### DIFF
--- a/pkg/auth/auth_test.go
+++ b/pkg/auth/auth_test.go
@@ -33,6 +33,8 @@ import (
 
 type fakeK8sAuth struct {
 	DiscoveryCli discovery.DiscoveryInterface
+	canIResult   bool
+	canIError    error
 }
 
 func (u fakeK8sAuth) Validate() error {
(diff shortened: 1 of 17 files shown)


pkg/auth/auth_test.go
```diff
@@ -33,6 +33,8 @@ import (
 
 type fakeK8sAuth struct {
 	DiscoveryCli discovery.DiscoveryInterface
+	canIResult   bool
+	canIError    error
```
absoludity: This test change was while I was first created unit tests for the new ValidateForNamespace before realising it's a one-line function proxying to the existing `CanI` function, but I left the changes as it's more extensible, allowing individual tests to decide what the result of the `CanI` call should be.

absoludity:
> It's true that this auth check makes more sense in the assetsvc rather than in the backend service because all the calls but the ones for the assetsvc hits the K8s API. If the user doesn't have permission to do something, the error from the K8s API would me more descriptive than ours.

It wasn't for that reason that I was keen to move it. Actually, are you sure that's even correct? In kubeops we don't use the AuthGate except for the assetsvc. And in tiller-proxy, we still need the AuthGate because the k8s API won't complain at all when tiller-proxy makes requests? Let's chat about this in the standup?

andresmgot:
> > It's true that this auth check makes more sense in the assetsvc rather than in the backend service because all the calls but the ones for the assetsvc hits the K8s API. If the user doesn't have permission to do something, the error from the K8s API would me more descriptive than ours.
> 
> It wasn't for that reason that I was keen to move it. Actually, are you sure that's even correct? In kubeops we don't use the AuthGate except for the assetsvc. And in tiller-proxy, we still need the AuthGate because the k8s API won't complain at all when tiller-proxy makes requests? Let's chat about this in the standup?

No, actually it's the same use case. After passing the AuthGate, we check if the user has permissions to do whatever is included in the chart so that AuthGate step is also optional for tiller-proxy. That's why we would only need it in the assetsvc.

PR #1594, opened 2020-03-19
OperatorInstance view

andresmgot:
### Description of the change

Draft for the OperatorInstance view:
Screenshot from 2020-03-19 13-32-37

For the moment we are showing just some basic info. The current status (as a YAML) and the installation values along with a Card with basic information about the CRD and the ClusterServiceVersion. The plan is to keep adding things but in other PRs to facilitate the review. The next things I plan to add are:

 - Make the Delete button works (currently is a dummy button)
 - Retrieve resources associated to the operator instance and show the same things than for other apps:
   - Application status (worklads running)
   - Resource tables.

We should also be able to modify the instance but I will leave that for later.

### Applicable issues

  - ref #1552

### DIFF
--- a/dashboard/src/components/AppList/CustomResourceListItem.tsx
+++ b/dashboard/src/components/AppList/CustomResourceListItem.tsx
@@ -1,8 +1,8 @@
 import * as React from "react";
 
-import UnexpectedErrorPage from "components/ErrorAlert/UnexpectedErrorAlert";
 import placeholder from "../../placeholder.png";
 import { IClusterServiceVersion, IResource } from "../../shared/types";
+import UnexpectedErrorPage from "../ErrorAlert/UnexpectedErrorAlert";
 import InfoCard from "../InfoCard";
 
 interface ICustomResourceListItemProps {
--- a/dashboard/src/components/AppList/CustomResourceListItem.tsx
+++ b/dashboard/src/components/AppList/CustomResourceListItem.tsx
@@ -6,19 +6,20 @@ import InfoCard from "../InfoCard";
 
 interface ICustomResourceListItemProps {
   resource: IResource;
-  csv?: IClusterServiceVersion;
+  csv: IClusterServiceVersion;
 }
 
 class CustomResourceListItem extends React.Component<ICustomResourceListItemProps> {
   public render() {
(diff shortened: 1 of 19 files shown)


dashboard/src/components/AppList/CustomResourceListItem.tsx
```diff
@@ -6,19 +6,20 @@ import InfoCard from "../InfoCard";
...
-    const icon = csv?.spec.icon
+    const icon = csv.spec.icon
       ? `data:${csv.spec.icon[0].mediatype};base64,${csv.spec.icon[0].base64data}`
       : placeholder;
+    const crd = csv.spec.customresourcedefinitions.owned.find(c => c.kind === resource.kind);
     return (
       <InfoCard
         key={resource.metadata.name}
-        link={`/operators-instances/ns/${resource.metadata.namespace}/${resource.metadata.name}`}
+        link={`/operators-instances/ns/${resource.metadata.namespace}/${csv.metadata.name}/${crd?.name}/${resource.metadata.name}`}
```
absoludity: Won't the `crd?.name` leave either the string 'undefined' or at best, '' so there will be two slashes?

dashboard/src/components/OperatorInstanceForm/OperatorInstanceForm.tsx
```diff
@@ -227,7 +227,9 @@ class DeploymentFormBody extends React.Component<
     this.setState({ submittedResourceName: resourceName });
     const created = await createResource(namespace, resource.apiVersion, resourceType, resource);
     if (created) {
-      push(`/operators-instances/ns/${namespace}/${resourceName}`);
+      push(
+        `/operators-instances/ns/${namespace}/${csv?.metadata.name}/${crd?.name}/${resourceName}`,
```
absoludity: Ditto here.

dashboard/src/components/AppList/CustomResourceListItem.tsx
andresmgot: yes, the thing is that `crd` will never be empty if `resource` is populated since the crd is required to obtain the resource. Typescript is not helping here a lot but I will try to make it more readable.

PR #1595, opened 2020-03-19
Operator instance delete button

andresmgot:
### Description of the change

Follow up of #1594. Reuses the ConfirmDialog component to delete an Operator instance:

Peek 2020-03-19 16-55

### Applicable issues

  - ref #1552

### DIFF
--- a/dashboard/src/components/OperatorInstance/OperatorInstance.tsx
+++ b/dashboard/src/components/OperatorInstance/OperatorInstance.tsx
@@ -46,7 +46,12 @@ class OperatorInstance extends React.Component<IOperatorInstanceProps, IOperator
   }
 
   public componentDidUpdate(prevProps: IOperatorInstanceProps) {
-    const { resource, csv } = this.props;
+    const { csvName, crdName, instanceName, namespace, getResource, resource, csv } = this.props;
+    if (prevProps.namespace !== namespace) {
+      getResource(namespace, csvName, crdName, instanceName);
+      return;
+    }
+
     if (csv !== prevProps.csv || resource !== prevProps.resource) {
       if (csv && resource) {
         this.setState({
--- a/dashboard/src/components/OperatorInstance/OperatorInstance.tsx
+++ b/dashboard/src/components/OperatorInstance/OperatorInstance.tsx
@@ -1,11 +1,13 @@
+import { RouterAction } from "connected-react-router";
 import * as yaml from "js-yaml";
 import * as React from "react";
 
 import placeholder from "../../placeholder.png";
-import { IClusterServiceVersion, IResource } from "../../shared/types";
+import { IClusterServiceVersion, IClusterServiceVersionCRD, IResource } from "../../shared/types";
 import AppNotes from "../AppView/AppNotes";
 import AppValues from "../AppView/AppValues";
 import Card, { CardContent, CardFooter, CardGrid, CardIcon } from "../Card";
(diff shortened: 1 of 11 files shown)


dashboard/src/components/OperatorInstance/OperatorInstance.tsx
```diff
@@ -21,19 +23,41 @@ export interface IOperatorInstanceProps {
...
+  };
+
   public componentDidMount() {
     const { csvName, crdName, instanceName, namespace, getResource } = this.props;
     getResource(namespace, csvName, crdName, instanceName);
   }
 
+  public componentDidUpdate(prevProps: IOperatorInstanceProps) {
+    const { resource, csv } = this.props;
+    if (csv !== prevProps.csv || resource !== prevProps.resource) {
```
absoludity: What happens if people change the namespace? I assume the controller tries to get the resource and we'll 404 appropriately so it's not relevant here?

dashboard/src/components/OperatorInstance/OperatorInstance.tsx
```diff
@@ -93,6 +125,28 @@ class OperatorInstance extends React.Component<IOperatorInstanceProps> {
...
+    });
+  };
+
+  private handleDeleteClick = async () => {
+    const { namespace, resource } = this.props;
+    const { crd } = this.state;
+    const deleted = await this.props.deleteResource(namespace, crd!.name.split(".")[0], resource!);
+    this.closeModal();
+    if (deleted) {
+      this.props.push(`/apps/ns/${namespace}`);
```
absoludity: Seems odd to be pushing down `push` as a function to the props (it'd be like sending `dispatch` as a prop to the component, rather than using `dispatch` at the controller. I'd expect instead to see a function `deleted()` which is provided as a prop (or `onDelete` or whatever) which handles the `push` at the controller. But not a big deal.

dashboard/src/components/OperatorInstance/OperatorInstance.tsx
andresmgot: good point, I didn't test the namespace change here (TBH I am just testing the happy path for the moment until I have something working end-to-end)

dashboard/src/components/OperatorInstance/OperatorInstance.tsx
andresmgot: we are using the same approach in different places, that's why I used it:

```
src/containers/HeaderContainer/HeaderContainer.ts
37:    push: (path: string) => dispatch(push(path)),

src/containers/AppNewContainer/AppNewContainer.tsx
47:    push: (location: string) => dispatch(push(location)),

src/containers/AppViewContainer/AppViewContainer.tsx
39:    push: (location: string) => dispatch(push(location)),

src/containers/AppUpgradeContainer/AppUpgradeContainer.tsx
56:    push: (location: string) => dispatch(push(location)),

src/containers/ServiceClassViewContainer/ServiceClassViewContainer.ts
50:    push: (location: string) => dispatch(push(location)),
```

In any case, what do you mean by `controller`? Why are you comparing `push` to `dispatch`?

absoludity: Generally I thought we try to keep components being dumb - they just react to UI events by either setting props or calling a function like, in this case, `deleted()`, without knowing what it does, so that something else which controls the component (ie. the thing with the brains :) ) can take the appropriate action. But no big deal, just unexpected.

PR #1596, opened 2020-03-20
Use namespace from the chart when rendering responses.

absoludity:
Ref #1521. A lot of churn, but only three real changes:

1. Remove the AuthGate from tiller proxy (as discussed yesterday, tiller-proxy checks the user auth explicitly in the handlers), which allows us to restrict it to the assetsvc use-case. (EDIT: CI showed that this also required removing the implicit passing of values via the request context which the AuthGate set, now removed, as it's not particularly type-safe).
2. Update AuthGate to allow access to the global repositories via the assetsvc (ie. in kubeapps' own namespace)
3. Ensure the assetsvc uses the namespace of the chart when rendering results, rather than the requested namespace for the query, since we're including the global charts in a request for specific namespaces and we need icons and other links to match the asset.

Unrelated: When do you think we could switch the default for `allowNamespaceDiscovery` to true?

### DIFF
--- a/cmd/tiller-proxy/internal/handler/handler.go
+++ b/cmd/tiller-proxy/internal/handler/handler.go
@@ -43,7 +43,10 @@ func returnForbiddenActions(forbiddenActions []auth.Action, w http.ResponseWrite
 
 // TillerProxy client and configuration
 type TillerProxy struct {
+	// TODO(mnelson): was DisableUserAuthCheck just an initial compatibility flag when the auth check
+	// was added? (ie. can it be removed).
 	DisableUserAuthCheck bool
+	CheckerForRequest    auth.CheckerForRequest
 	ListLimit            int
 	ChartClient          chartUtils.Resolver
 	ProxyClient          proxy.TillerClient
@@ -73,7 +76,11 @@ func (h *TillerProxy) CreateRelease(w http.ResponseWriter, req *http.Request, pa
 			response.NewErrorResponse(handlerutil.ErrorCode(err), err.Error()).Write(w)
 			return
 		}
-		userAuth := req.Context().Value(auth.UserKey).(auth.Checker)
+		userAuth, err := h.CheckerForRequest(req)
+		if err != nil {
+			response.NewErrorResponse(handlerutil.ErrorCode(err), err.Error()).Write(w)
+			return
+		}
 		forbiddenActions, err := userAuth.GetForbiddenActions(params["namespace"], "create", manifest)
 		if err != nil {
 			response.NewErrorResponse(handlerutil.ErrorCode(err), err.Error()).Write(w)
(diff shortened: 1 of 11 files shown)


cmd/tiller-proxy/internal/handler/handler.go
```diff
@@ -43,7 +43,10 @@ func returnForbiddenActions(forbiddenActions []auth.Action, w http.ResponseWrite
 
 // TillerProxy client and configuration
 type TillerProxy struct {
+	// TODO(mnelson): was DisableUserAuthCheck just an initial compatibility flag when the auth check
+	// was added? (ie. can it be removed).
```
andresmgot: yes, it was also a way for people to disable the auth in case they want to skip authentication. I don't think it's being used anyway so we can remove it if we want.

script/deploy-dev.mk
```diff
@@ -34,7 +34,10 @@ deploy-dev: deploy-dex deploy-openldap update-apiserver-etc-hosts
 		--values ./docs/user/manifests/kubeapps-local-dev-values.yaml \
 		--values ./docs/user/manifests/kubeapps-local-dev-auth-proxy-values.yaml \
 		--set useHelm3=true \
+		--set allowNamespaceDiscovery=true \
 		--set postgresql.enabled=true \
+		--set postgresql.volumePermissions.enabled=false \
+		--set postgresql.volumePermissions.image.pullPolicy=IfNotPresent \
```
andresmgot: curious, why this?

pkg/auth/auth.go
```diff
@@ -96,6 +96,7 @@ type Action struct {
 // Checker for the exported funcs
 type Checker interface {
 	Validate() error
```
andresmgot: is Validate being used anymore now?

andresmgot:
> Unrelated: When do you think we could switch the default for allowNamespaceDiscovery to true?

We can do that and bumping the major chart version. Maybe when we release this feature and the operators PoC.

cmd/tiller-proxy/internal/handler/handler.go
absoludity: Great, I'll do that separately.

pkg/auth/auth.go
absoludity: No, it's no longer used. I'll remove that too.

script/deploy-dev.mk
absoludity: Woops. It was because I was having issues with postgres taking too long to start up when working on a slower connection (it's been trickier running locally rather than on dockerdev). So I was ensuring I could use an image I'd already loaded on the cluster. Removed.

absoludity:
> > Unrelated: When do you think we could switch the default for allowNamespaceDiscovery to true?
> 
> We can do that and bumping the major chart version. Maybe when we release this feature and the operators PoC.

Sounds good.

PR #1597, opened 2020-03-20
Update yaml dep

andresmgot:
### Description of the change

Fixes an issue I found when upgrading the `yaml` dependency: https://github.com/eemeli/yaml/issues/142

### DIFF
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -65,7 +65,7 @@
     "semver": "^5.6.0",
     "ts-jest": "^22.0.1",
     "typesafe-actions": "^2.0.4",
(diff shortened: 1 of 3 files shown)


eemeli:
I'd recommend holding off on this for a couple of days, as the `yaml/browser/...` paths aren't guaranteed to always stay constant between minor releases. This really should "just work", and will be fixed in the next patch release of `yaml` very shortly.

andresmgot:
Yes, in any case the tests are not happy with the change anyway so I am closing the PR. Thanks for the heads up @eemeli!

PR #1599, opened 2020-03-23
Add ApplicationStatus and resource tables to the OpeartorInstance

andresmgot:
### Description of the change

Show the ApplicationStatus along with the AccessURL table and all the other resource tables:

Screenshot from 2020-03-23 19-09-28

The CRD doesn't have a specific reference to the resources it will create, instead it just list the different Kinds that it will create (e.g. Service, Deployment). Because of that, we need to list all those resources in the given namespace and filter out the ones that doesn't have the Custom Resource as owner.

There are several minor things that I found while working on this that require special attention:

 - When getting and watching a list of resources, first you get the resource list (e.g. DeploymentList) but then, you receive individual responses for each of the resoucers (e.g a Deployment) not the full list again.
 - When getting a list (e.g. DeploymentList) the resources within the list doesn't include its `kind` (I guess it's assumed from the list) so we cannot rely in that property. I am using the `selfLink` in this PR instead.

I may be missing more things to explain, please let me know if there is something that it's difficult to understand.

### Applicable issues

  - ref #1552

### DIFF
--- a/dashboard/src/shared/ResourceAPIVersion.ts
+++ b/dashboard/src/shared/ResourceAPIVersion.ts
@@ -1,0 +1,15 @@
+// We explicitly define the apiVersions here, just in case a generic pluralizer
+// isn't sufficient. Note that apiVersions may change over time.
+export const ResourceKindsWithAPIVersions = {
+  ClusterRole: "rbac.authorization.k8s.io/v1",
+  ConfigMap: "v1",
+  DaemonSet: "apps/v1",
+  Deployment: "apps/v1",
+  Ingress: "extensions/v1beta1",
+  Secret: "v1",
+  Service: "v1",
+  StatefulSet: "apps/v1",
+  Pods: "v1",
+} as const;
+
+export type ResourceAPIVersion = keyof typeof ResourceKindsWithAPIVersions;
--- a/dashboard/src/actions/kube.test.tsx
+++ b/dashboard/src/actions/kube.test.tsx
@@ -4,8 +4,8 @@ import { getType } from "typesafe-actions";
 
 import actions from ".";
 import { Kube } from "../shared/Kube";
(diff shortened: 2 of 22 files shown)


dashboard/src/shared/ResourceAPIVersion.ts
```diff
@@ -0,0 +1,15 @@
+// We explicitly define the apiVersions here, just in case a generic pluralizer
+// isn't sufficient. Note that apiVersions may change over time.
```
absoludity: Couple of questions: why the new file (ie. why not define this const in `Kube.ts`).

Second, I was wondering why you were defining these when, as you say, they'll change depending on the k8s server, but then saw it'd take a bunch of queries to populate that info (eg. `/apis/rbac.authorization.k8s.io/` to get the versions, then `/apis/rbac.authorization.k8s.io/v1/` to find the available kinds and plural names, then for each api). Looks easy to hard-code :)

Checking what kubectl does, when I run `kubectl get apprepositories -v10` it is able to check cached versions of the server resources (i assume populated from the above queries).

andresmgot: > Couple of questions: why the new file (ie. why not define this const in Kube.ts).

I followed the same approach than with the resource plurals that are defined as well in a standalone file (ResourceKinds.ts).

> Second, I was wondering why you were defining these when, as you say, they'll change depending on the k8s server, but then saw it'd take a bunch of queries to populate that info (eg. /apis/rbac.authorization.k8s.io/ to get the versions, then /apis/rbac.authorization.k8s.io/v1/ to find the available kinds and plural names, then for each api). Looks easy to hard-code :)

Yes, apart from the complexity of doing several requests, it would be async calls and also I am not sure how RBAC behaves in that case so it seemed better to just hardcode it for the moment.

PR #1600, opened 2020-03-24
Avoid re-auth on All Namespaces when selecting catalog

absoludity:
The simple part of this PR was just switching the `authGate` to return a Forbidden rather than Unauthorized when a user requests the charts for a namespace to which they don't have access. This stopped the frontend from logging out, but a little more work was needed to display an error with an action the user can take.

This issue does not arise normally, because we already check whether a user has access to a namespace when selecting the namespace (ie. before we even attempt to get charts for the namespace), but because we currently use `All Namespaces` as the default when a user logs in, even though the user may not have access to resources across all namespaces, the issue appears when logging in as such a user and then selecting the catalog.

The correct solution, IMO, is to select the first namespace to which the user has access as the default, but we can't do this until `allowNamespaceDiscovery` is always on (?). Without that ability yet, I've instead displayed an error with an action for the user to take (selecting another namespace to which they have access). See what you think.

### DIFF
--- a/dashboard/src/components/Catalog/Catalog.tsx
+++ b/dashboard/src/components/Catalog/Catalog.tsx
@@ -65,7 +65,7 @@ class Catalog extends React.Component<ICatalogProps, ICatalogState> {
     } = this.props;
     const { listCharts, listOperators } = this.state;
     if (error) {
-      const isForbidden = error.constructor == ForbiddenError;
+      const isForbidden = error.constructor === ForbiddenError;
       return (
         <MessageAlert
           level={"error"}
--- a/dashboard/src/components/Catalog/Catalog.tsx
+++ b/dashboard/src/components/Catalog/Catalog.tsx
@@ -2,7 +2,7 @@ import { RouterAction } from "connected-react-router";
 import * as React from "react";
 import { Link } from "react-router-dom";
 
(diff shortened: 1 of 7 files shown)


dashboard/src/components/Catalog/Catalog.tsx
```diff
@@ -59,11 +59,24 @@ class Catalog extends React.Component<ICatalogProps, ICatalogState> {
...
     } = this.props;
     const { listCharts, listOperators } = this.state;
+    if (error) {
+      return (
+        <MessageAlert
+          level={"error"}
+          children={
+            <div>
+              <h5>Unable to fetch catalog</h5>
+              There was an error fetching the catalog. Please choose a namespace above to which you have access.
```
andresmgot: this could be misleading if the error is because any other reason, I would add a condition:

```
There was an error fetching the catalog. {namespace === definedNamespaces.all && "Please choose a namespace above to which you have access."}
```
Or check if the error is a Forbidden error

andresmgot:
> The correct solution, IMO, is to select the first namespace to which the user has access as the default, but we can't do this until allowNamespaceDiscovery is always on (?)

I am okay if you want to set the default to true already

PR #1602, opened 2020-03-24
Several fixes for operators

andresmgot:
### Description of the change

I have tried with a different operator and found several issues:

 - [unrelated to operators] Some ingresses doesn't define rules, it's possible to specify just a `backend` [ref](https://kubernetes.io/docs/concepts/services-networking/ingress/#single-service-ingress). I have adapted the AccessURLItem to cover that case.
 - The `kind` property is not available when the resource is extracted from a list. Use `selfLink` instead.
 - When a list is empty, show a message rather than an empty list.
 - Avoid to show a message "No charts found" for the catalog view in a namespace if there are operators available.
 - Transform names into lower case before ordering items in the catalog view (if not, names starting with an upper case went first, instead that ordering things alphabetically).
 - If a ClusterServiceVersion doesn't define the resources that its CRD generates, pull all the different resources instead of rendering an empty view.
 - The `status` and the `spec` of an operator instance is missing sometimes. Don't show the related information in that case.

This an example of those fixes for the jaeger operator:

Screenshot from 2020-03-24 16-40-50

### Applicable issues

  - ref #1552


dashboard/src/shared/types.ts
```diff
@@ -269,7 +273,7 @@ export interface IPackageManifest extends IResource {
 export interface IClusterServiceVersionCRDResource {
   kind: string;
   name: string;
-  version: string;
+  version?: string;
```
absoludity: Why isn't this required for the resource?

dashboard/src/reducers/charts.test.ts
```diff
@@ -49,6 +49,7 @@ describe("chartReducer", () => {
             ...initialState.selected,
             error,
           },
+        },
```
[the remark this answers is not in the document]
andresmgot: I think so, I got the error trying to commit this, the pre-check that is done before commiting was failing so I guess it's an issue with `react-scripts-ts`.

dashboard/src/shared/types.ts
andresmgot: actually it's required but it was not being used, re-added

PR #1604, opened 2020-03-25
Use per-namespace app repositories by default

absoludity:
Enables `allowNamespaceDiscovery` and `reposPerNamespace` to be on by default (as well as the invalidateCache which is required whenever we release a new Kubeapps version).

Updates the documentation for private app repos to match.

If we don't have any separate chart releases in the next few days, it might be good to land this and test it locally before releasing. Happy either way.

### DIFF
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -12,9 +12,9 @@
 ## If you set it to true, Kubeapps will not work with releases installed with Helm 2.
 useHelm3: false
 
-## Enable this feature flag to allow users to discover available namespaces (only the ones they have access).
-## If you set it to true, Kubeapps creates a ClusterRole to be able to list namespaces.
-allowNamespaceDiscovery: false
+## Disable this feature flag to disallow users to discover available namespaces (only the ones they have access).
+## When set to true, Kubeapps creates a ClusterRole to be able to list namespaces.
+allowNamespaceDiscovery: true
 
 ## The frontend service is the main reverse proxy used to access the Kubeapps UI
 ## To expose Kubeapps externally either configure the ingress object below or
@@ -638,9 +638,8 @@ authProxy:
       cpu: 25m
       memory: 32Mi
 ## Feature flags
-## These are used to switch on in development features or new features not yet released.
+## These are used to switch on in development features or new features which are ready to be released.
 featureFlags:
-  reposPerNamespace: false
-  # The invalidateCache flag must be set to default to true when we next release the Kubeapps app.
-  invalidateCache: false
+  reposPerNamespace: true
+  invalidateCache: true
   operators: false
--- a/dashboard/src/components/AppUpgrade/AppUpgrade.tsx
+++ b/dashboard/src/components/AppUpgrade/AppUpgrade.tsx
@@ -121,6 +121,7 @@ class AppUpgrade extends React.Component<IAppUpgradeProps> {
         isFetching={this.props.reposIsFetching}
         error={this.props.chartsError}
         kubeappsNamespace={this.props.kubeappsNamespace}
+        namespace={this.props.namespace}
         repoError={this.props.repoError}
         repo={this.props.repo}
         repos={this.props.repos}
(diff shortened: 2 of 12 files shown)


absoludity:
@andresmgot I'll fix the conflict, but I'm not sure why this is failing on CI with the *admin* user not having access to fetch AppRepos in the kubeapps namespace? (see https://19293-107555654-gh.circle-artifacts.com/0/integration/reports/screenshots/creates-a-registry.png , which is from `integration/use-cases/create-registry.js` which logs in with `ADMIN_TOKEN`. I'll look more closely in the morning.

andresmgot:
> @andresmgot I'll fix the conflict, but I'm not sure why this is failing on CI with the _admin_ user not having access to fetch AppRepos in the kubeapps namespace? (see https://19293-107555654-gh.circle-artifacts.com/0/integration/reports/screenshots/creates-a-registry.png , which is from `integration/use-cases/create-registry.js` which logs in with `ADMIN_TOKEN`. I'll look more closely in the morning.

The error could be misleading, if the component is receiving a forbidden error and the error message is not parse-able as a Forbidden error it shows the default required RBAC roles error that is the one you are seeing so maybe what's failing is not exactly fetching app repositories. Also note that the admin user has RBAC permissions to manage apprepositories only in the `kubeapps` namespace ([ref](https://github.com/kubeapps/kubeapps/blob/master/script/e2e-test.sh#L233), since that namespace is now like a "global" namespace that can be also the error if we are requesting all apprepositories.

absoludity:
> > @andresmgot I'll fix the conflict, but I'm not sure why this is failing on CI with the _admin_ user not having access to fetch AppRepos in the kubeapps namespace? (see https://19293-107555654-gh.circle-artifacts.com/0/integration/reports/screenshots/creates-a-registry.png , which is from `integration/use-cases/create-registry.js` which logs in with `ADMIN_TOKEN`. I'll look more closely in the morning.
> 
> The error could be misleading, if the component is receiving a forbidden error and the error message is not parse-able as a Forbidden error it shows the default required RBAC roles error that is the one you are seeing. Another possibility is that the admin user has RBAC permissions to manage apprepositories only in the `kubeapps` namespace ([ref](https://github.com/kubeapps/kubeapps/blob/master/script/e2e-test.sh#L233), since that namespace is now like a "global" namespace that can be also the error if we are requesting apprepositories in all namespaces.

Thanks for the background. That helped me see that the issue I had initially was because we were only giving write access to kubeapps-operator, where I explicitly updated this in the PR so that the UI requires both read and write to use the add-app repo UI. I've added read access and it gets past there, but fails elsewhere. Landing your PR #1606 which I'll merge and retry...

absoludity:
> > > @andresmgot I'll fix the conflict, but I'm not sure why this is failing on CI with the _admin_ user not having access to fetch AppRepos in the kubeapps namespace? (see https://19293-107555654-gh.circle-artifacts.com/0/integration/reports/screenshots/creates-a-registry.png , which is from `integration/use-cases/create-registry.js` which logs in with `ADMIN_TOKEN`. I'll look more closely in the morning.
> > 
> > 
> > The error could be misleading, if the component is receiving a forbidden error and the error message is not parse-able as a Forbidden error it shows the default required RBAC roles error that is the one you are seeing. Another possibility is that the admin user has RBAC permissions to manage apprepositories only in the `kubeapps` namespace ([ref](https://github.com/kubeapps/kubeapps/blob/master/script/e2e-test.sh#L233), since that namespace is now like a "global" namespace that can be also the error if we are requesting apprepositories in all namespaces.
> 
> Thanks for the background. That helped me see that the issue I had initially was because we were only giving write access to kubeapps-operator, where I explicitly updated this in the PR so that the UI requires both read and write to use the add-app repo UI. I've added read access and it gets past there, but fails elsewhere. Landing your PR #1606 which I'll merge and retry...

So, frustratingly, this appears to be similar to something we've experienced in the past requiring a second click. In this case, the first click appears to only give the button focus, works fine with the second :/ See https://github.com/kubeapps/kubeapps/pull/1604/commits/d7b74104d8f9c85d1c520eeb69ab7fef856bd16c for the fix.

andresmgot:
> So, frustratingly, this appears to be similar to something we've experienced in the past requiring a second click. In this case, the first click appears to only give the button focus, works fine with the second :/ See [d7b7410](https://github.com/kubeapps/kubeapps/commit/d7b74104d8f9c85d1c520eeb69ab7fef856bd16c) for the fix.

it's indeed weird to see this behavior from time to time.. I wonder if it's an issue in our side or the test framework behaving weird with modals.

PR #1607, opened 2020-03-25
Show operators and instances in all namespaces

andresmgot:
### Description of the change

Fix the operator list and the application list for operators when selecting "All Namespaces" in the namespace selector.

### Applicable issues

  - ref #1552

### DIFF
--- a/dashboard/src/shared/url.ts
+++ b/dashboard/src/shared/url.ts
@@ -9,6 +9,10 @@ export const app = {
   },
 };
 
+function withNS(namespace: string) {
+  return namespace === "_all" ? "" : `namespaces/${namespace}`;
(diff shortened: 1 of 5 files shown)


dashboard/src/shared/url.ts
```diff
@@ -9,6 +9,10 @@ export const app = {
   },
 };
 
+function withNS(namespace: string) {
+  return namespace === "_all" ? "" : `namespaces/${namespace}`;
```
absoludity: If you made this `namespaces/${namespace}/` (ie. so that the non-all case has a trailing slash), then you could avoid the double-slash below when namespace is `_all`?

PR #1608, opened 2020-03-25
Add icon to AppList items

andresmgot:
### Description of the change

In order to differentiate between charts and operator instances, I have added an icon with the logo of both solutions (instead of using tags). If we are fine with this I can apply the same to the Catalog.

Screenshot from 2020-03-26 10-58-33

### Applicable issues

  - ref #1552


andresmgot:
> Actually, does that mean we can remove the "operator" tag now? (given that we don't have a "chart" tag)?

Yes, I prefer to use two tags because it's similar to the chart card but we can use something useful like the operator name.

PR #1609, opened 2020-03-26
Fix ci test.

absoludity:


absoludity:
@andresmgot jfyi: all four e2e tests fail, you can even see that the button is focused (the border): https://19575-107555654-gh.circle-artifacts.com/0/integration/reports/screenshots/creates-a-registry.png

If it is something with the test-framework or modal dialogs, there must be something in my previous change #1604 that is triggering it.

andresmgot:
I don't think this is a button issue, the other issue it was not reproducible (only failed 1 out of 5 times more or less). If this is reproducible probably the error is somewhere else. I think that the problem can be that the application is failing to create the apprepository but the UI is not rendering the error. I can try to reproduce it.

andresmgot:
mm, it's weird, it works for me locally in master:

```
▶ INTEGRATION_ENTRYPOINT=http://kubeapps.local:30080 ADMIN_TOKEN=${admin_token} VIEW_TOKEN=${view_token} EDIT_TOKEN=${edit_token} yarn start:window -- create-registry
yarn run v1.21.1
warning From Yarn 1.0 onwards, scripts don't require "--" for options to be forwarded. In a future version, any explicit "--" will be forwarded as-is to the scripts.
$ INTEGRATION_HEADLESS=false jest --runInBand create-registry
Waiting http://kubeapps.local:30080 to be ready before running the tests (60s)
http://kubeapps.local:30080 is ready!
 PASS  use-cases/create-registry.js (17.897s)
```

Let me try in a PR.

andresmgot:
Ah, so I found the error, the problem is that the error is shown before the title so it's not in the screenshot:

Screenshot from 2020-03-26 12-56-46

It's indeed a forbidden error (my operator user was overpowered, that's why I didn't reproduce it at first)

PR #1610, opened 2020-03-26
Restore test, fix apprepository-rbac

andresmgot:
### Description of the change

Follow up of #1609

As explained there, the current test is giving a false positive in master. The reason is quite tricky. With the changes added in the RBAC templates, we were removing the blank line at the end of apprepository-rbac.yaml. This causes that the `Role` affected (`repositories-write`) was not being created and it fails quietly (note the `create---`):

```yaml
---
# Source: kubeapps/templates/apprepository-rbac.yaml
# Define role, but no binding, so users can be bound to this role
kind: Role
metadata:
  name: kubeapps2-repositories-write
rules:
  - apiGroups:
      - kubeapps.com
    resources:
      - apprepositories
    verbs:
      - "*"
  - apiGroups:
      - ""
    resources:
      - secrets
    verbs:
      - create---
# Kubeapps can read and watch its own AppRepository resources cluster-wide.
kind: ClusterRole
```

So giving write access to the operator serviceaccount was not doing anything. Fun bug...

### DIFF
--- a/chart/kubeapps/templates/apprepository-rbac.yaml
+++ b/chart/kubeapps/templates/apprepository-rbac.yaml
@@ -93,7 +93,7 @@ rules:
       - secrets
     verbs:
       - create
-{{- end -}}
+{{ end }}
 {{- if .Values.featureFlags.reposPerNamespace -}}
 ---
 # Kubeapps can read and watch its own AppRepository resources cluster-wide.
--- a/integration/use-cases/create-registry.js
+++ b/integration/use-cases/create-registry.js
@@ -31,7 +31,7 @@ test("Creates a registry", async () => {
   let retries = 3;
   while (retries > 0) {
     try {
-      await expect(page).toMatch("gitlab", { timeout: 2000 });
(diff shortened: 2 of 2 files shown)


andresmgot:
_merging this to fix master_

absoludity:
> So giving write access to the operator serviceaccount was not doing anything. Fun bug...

Wow, fun indeed :P. Thanks Andres. So I assume k8s itself makes this obvious (with an error along the lines of the referenced role in the role-binding is not found? Wondering if we need to ensure that's exposed in the future.

I'm surprised it actually validated that yaml?

absoludity:
Great debugging, btw!

andresmgot:
> So I assume k8s itself makes this obvious (with an error along the lines of the referenced role in the role-binding is not found?

Nope, it just says forbidden and that the user doesn't have permissions. It took a bit for me to notice that the role was missing.

> I'm surprised it actually validated that yaml?

Me too :man_shrugging:

PR #1611, opened 2020-03-26
Allow to update operator instances

andresmgot:
### Description of the change

Adds a form to update (modify) operator instances. I've refactored the OperatorInstanceForm so it can be used for creating and updating resources.

Screenshot from 2020-03-26 18-14-23

Note: I have also included a unrelated change (sorry). The AppList component was showing operators even when the feature flag was not enabled.

### Applicable issues

  - ref #1552


dashboard/src/actions/operators.ts
```diff
@@ -214,6 +214,26 @@ export function deleteResource(
...
+  apiVersion: string,
+  resource: string,
+  name: string,
+  body: object,
+): ThunkAction<Promise<boolean>, IStoreState, null, OperatorAction> {
+  return async dispatch => {
+    dispatch(creatingResource());
+    try {
+      const r = await Operators.updateResource(namespace, apiVersion, resource, name, body);
+      dispatch(resourceCreated(r));
```
absoludity: Should this be a `resourceUpdated(r)` action?

dashboard/src/components/OperatorInstance/OperatorInstance.tsx
```diff
@@ -128,8 +128,21 @@ class OperatorInstance extends React.Component<IOperatorInstanceProps, IOperator
...
+      csv,
+      instanceName,
+      csvName,
+      crdName,
+      namespace,
+      push,
+    } = this.props;
     const { resources } = this.state;
+    const onUpdateClick = () =>
+      push(`/operators-instances/ns/${namespace}/${csvName}/${crdName}/update/${instanceName}`);
```
absoludity: Starting to think it'd be worthwhile to define the URL template literals for UI routes in one place, and refer to them by name? (not necessarily now, but just to think about).

dashboard/src/actions/operators.ts
andresmgot: I didn't want to add more boilerplate code but yes, it could be confusing

dashboard/src/components/OperatorInstance/OperatorInstance.tsx
andresmgot: you mean like the URLs that are already in `dashboard/src/shared/url.ts`? You are right, better not to add URL literals out of there.

PR #1613, opened 2020-03-27
Show the list of operators installed

andresmgot:
### Description of the change

In the Operators list, show different sections to browse the catalog of operators or to list the ones installed:

Screenshot from 2020-03-30 15-57-29

### Applicable issues

  - ref #1552


andresmgot:
> I could be missing something, but don't the cards appearing on the "Installed" tab also appear in the "Browse" tab (unless we explicitly remove them). Asking because what I'd expect is just a simple list (rather than 2 tabs) which shows all operators, but clearly prioritizes and identifies the installed operators (perhaps by having them as the first section of the list).
> 
> This is what I see most UI lists of apps or whatever doing, which makes more sense to me. +1, but see what you think.

I was probably biased from other views, it's a valid point (and would play better with namespace changes) so I'm fine using sections in the same view.

dashboard/src/components/OperatorList/OperatorList.test.tsx
```diff
@@ -75,11 +112,41 @@ it("skips the error if the OLM is not installed", () => {
...
+  expect(
+    wrapper
+      .find(CardGrid)
+      .last()
+      .children(),
+  ).not.toExist();
+  expect(wrapper).toMatchSnapshot();
+});
+
+it("render the operator list with installed operators", () => {
```
absoludity: You've re-used the same test name here.

dashboard/src/components/OperatorList/OperatorList.test.tsx
```diff
@@ -75,11 +112,41 @@ it("skips the error if the OLM is not installed", () => {
...
+  const wrapper = shallow(
+    <OperatorList {...defaultProps} isOLMInstalled={true} operators={[sampleOperator]} csvs={[]} />,
   );
   expect(wrapper.find(OLMNotFound)).not.toExist();
   expect(wrapper.find(InfoCard)).toExist();
```
absoludity: It's not clear to me if this should be asserting that the "Installed" section is not present when nothing is installed. That is, should this just show the available operators without even an 'Installed" section when there are non installed?

EDIT: I see from the code and snapshot below that this is indeed the way it renders, so I guess I expected to see here something like an assertion that the h3 heading "Installed" does not exist (but does above). Then again. the snapshot does that for us - *shrug*.

andresmgot: > EDIT: I see from the code and snapshot below that this is indeed the way it renders, so I guess I expected to see here something like an assertion that the h3 heading "Installed" does not exist (but does above). Then again. the snapshot does that for us - _shrug_.

That's exactly what I am doing in the lines 144 and 127

absoludity: Not sure how I missed 144 :)

PR #1614, opened 2020-03-30
Update getChartVersions* helpers to include the chart namespace

absoludity:
Although I'd already updated the getChart* helpers to ensure we requested the chart its specific namespace (not necessarily the current namespace being browsed by the user), the getChartVersion* functions had not been updated, so were always requesting with the Kubeapps namespace.

This won't yet pass CI: I still need to test locally, and there is one more function (getChartReadme) which needs updating, which I'll do before landing.


dashboard/src/actions/charts.test.tsx
```diff
@@ -82,10 +80,10 @@ describe("fetchChartVersions", () => {
       { type: getType(actions.charts.requestCharts) },
       { type: getType(actions.charts.receiveChartVersions), payload: response },
     ];
-    await store.dispatch(actions.charts.fetchChartVersions("foo"));
+    await store.dispatch(actions.charts.fetchChartVersions("chart-namespace", "foo"));
     expect(store.getActions()).toEqual(expectedActions);
     expect(axiosGetMock.mock.calls[0][0]).toBe(
-      `api/assetsvc/v1/ns/${namespace}/charts/foo/versions`,
+      `api/assetsvc/v1/ns/chart-namespace/charts/foo/versions`,
```
andresmgot: the linter would change this to "api/assetsvc..." (with double quotes) but why did you remove the var?

dashboard/src/containers/AppUpgradeContainer/AppUpgradeContainer.tsx
```diff
@@ -47,12 +50,12 @@ function mapDispatchToProps(dispatch: ThunkDispatch<IStoreState, null, Action>)
     checkChart: (repo: string, chartName: string) =>
       dispatch(actions.repos.checkChart(repo, chartName)),
     clearRepo: () => dispatch(actions.repos.clearRepo()),
-    fetchChartVersions: (id: string) => dispatch(actions.charts.fetchChartVersions(id)),
+    fetchChartVersions: (namespace: string, id: string) => dispatch(actions.charts.fetchChartVersions(namespace, id)),
     fetchRepositories: () => dispatch(actions.repos.fetchRepos()),
     getAppWithUpdateInfo: (releaseName: string, ns: string) =>
```
andresmgot: here you can see that there are other functions that require the namespace as well: checkChart, fetchRepositories, getAppWithUpdateInfo (that last one is used in many places).

dashboard/src/actions/charts.test.tsx
absoludity: That was a mistake (must have been blindly fixing tests). Switch this (and similar ones) back to use `${namespace}` (and use `chart-namespace` for that var).

dashboard/src/containers/AppUpgradeContainer/AppUpgradeContainer.tsx
absoludity: Updated `checkChart` and `fetchRepositories`, though `getAppWithUpdateInfo` already included the namespace (just hidden with a 2-letter var ns, I updated it and put the namespace arg first for consistency).

Need to test the built images IRL, I suspect I'll need to change the inputs for the SelectRepo component.

absoludity:
> +1 LGTM let me know if you find trouble testing this

Thanks. I got to spend some time looking at the remaining issues, and they revolve around the assumption in our routes that we only care about the namespace in which the user is deploying (no longer true, as we also care about the namespace in which the chart is defined). That is, if you are browsing the catalog in namespace-x, which shows the global charts (ie. from kubeapps namespace), and click on a chart, the route needs to include both namespaces. I'm updating this in a separate PR based on this one (but it's not ready), and updating the privateRoutes as:
```
--- dashboard/src/containers/RoutesContainer/Routes.tsx
+++ dashboard/src/containers/RoutesContainer/Routes.tsx
@@ -26,12 +26,12 @@ type IRouteComponentPropsAndRouteProps = RouteProps & RouteComponentProps<any>;
 const privateRoutes = {
   "/apps/ns/:namespace": AppListContainer,
   "/apps/ns/:namespace/:releaseName": AppViewContainer,
-  "/apps/ns/:namespace/new/:repo/:id/versions/:version": AppNewContainer,
+  "/apps/ns/:namespace/new/:reponamespace/:repo/:id/versions/:version": AppNewContainer,
   "/apps/ns/:namespace/upgrade/:releaseName": AppUpgradeContainer,
   "/catalog/ns/:namespace": CatalogContainer,
   "/catalog/ns/:namespace/:repo": CatalogContainer,
-  "/charts/ns/:namespace/:repo/:id": ChartViewContainer,
-  "/charts/ns/:namespace/:repo/:id/versions/:version": ChartViewContainer,
+  "/ns/:namespace/charts/:reponamespace/:repo/:id": ChartViewContainer,
+  "/ns/:namespace/charts/:reponamespace/:repo/:id/versions/:version": ChartViewContainer,
   "/config/brokers": ServiceBrokerListContainer,
   "/services/brokers/:brokerName/classes/:className": ServiceClassViewContainer,
   "/services/brokers/:brokerName/instances/ns/:namespace/:instanceName": ServiceInstanceViewContainer,
``` 
But it looks like it's going to be worth making them all consistent, ie. so all the `/apps/ns` ones are `/ns/:namespace/apps/...` as well. What do you think?

andresmgot:
I don't think we need two different namespaces in the route, that would be confusing (like if we were supporting deploying charts in `ns1` only defined in `ns2`.

It's the work of the backend to return global charts, for example: `/charts/ns/default/bitnami/wordpress` should still return the chart from the global repository `bitnami`.

absoludity:
> I don't think we need two different namespaces in the route, that would be confusing (like if we were supporting deploying charts in `ns1` only defined in `ns2`).

I'm not sure I see an alternative... we need to be able to uniquely reference a chart in a namespaced repository (as the repo/chart-name is *not* unique)? Let's say you're browsing the catalog for "All namespaces" at:
```
/catalog/ns/_all
```
which shows both the `bitnami/wordpress` chart from the `kubeapps` namespace as well as the `my-repo/fluentd-elasticsearch` from the `kubeapps-user-namespace` (and possibly even another `my-repo/fluentd-elasticsearch` from another namespace). When you click on the elasticsearch one, the route will be:
```
/catalog/ns/_all/my-repo/fluentd-elasticsearch
```
which does not identify the chart. We could update the assetsvc to return the *first* chart matching `my-repo/fluentd-elasticsearch` across all namespaces, but that is a work-around for not having a unique way to identify the chart. Of course, using "All namespaces" highlights the issue, but it's the same if you consider browsing `/catalog/ns/kubeapps-user-namespace` (in that you need a unique way to refer to a chart given that there can be multiple repos available in that namespace from the `kubeapps` namespace - or other namespaces later).

Currently if you ask the assetsvc for all the charts available in a given namespace, it returns both those for repos in that namespace as well as the global namespace, *but* the returned data includes the actual repository (including namespace) to which the chart belongs. Whereas it sounds like you're saying it should pretend the all charts are actually from that namespace? This leads to duplication of a lot of data (and things like us busting the browser cache for logos etc.).

This is why I thought the route for deploying a chart should identify both the current namespace to which we're deploying, as well as uniquely identifying the versioned chart:
`/ns/:namespace/apps/new/:reponamespace/:repo/:id/versions/:version` (and similar for the chartview, as it is the path to deploying an app).

But keen to hear if you have an alternative.

> 
> It's the work of the backend to return global charts, for example: `/charts/ns/default/bitnami/wordpress` should still return the chart from the global repository `bitnami` (stored in the `kubeapps` namespace).

As above, this means that:
`/chart/ns/<anything>/my-repo/wordpress` returns the chart from the global repository even if it's not the chart I intended (eg. I was specifically browsing the `my-repo` repo in another namespace when I selected `wordpress`.

I really don't (yet) see why we would not want to be able to uniquely address a chart in the assetsvc, which has implications for the dashboard. Let me know your thoughts.

andresmgot:
> which shows both the `bitnami/wordpress` chart from the `kubeapps` namespace as well as the `my-repo/fluentd-elasticsearch` from the `kubeapps-user-namespace` (and possibly even another `my-repo/fluentd-elasticsearch` from another namespace). When you click on the elasticsearch one, the route will be:
> 
> ```
> /catalog/ns/_all/my-repo/fluentd-elasticsearch
> ```
> 
> which does not identify the chart. We could update the assetsvc to return the _first_ chart matching `my-repo/fluentd-elasticsearch` across all namespaces, but that is a work-around for not having a unique way to identify the chart (and leaves to chance whether it's actually correct).

In that case, I think we need to redirect to the specific namespace. It's the same that we do if you list applications in "All namespaces", once you select an application it would redirect to the application namespace. If it's a global chart I would redirect it to `/catalog/ns/kubeapps/bitnami/wordpress`.

This doesn't affect if the user is in their namespace so going to `/catalog/ns/my-ns/bitnami/wordpress` would have the same solution.

This is not great from a "All namespaces" UX but it's the best I can think of since the apprepository is actually in the kubeapps namespace.

What we should treat carefully is namespaced repositories that use the same ID than global ones. If someone tries to define a repository `bitnami` in their namespace, that should return a conflict error.

> Whereas it sounds like you're saying it should pretend the all charts are actually from that namespace? This leads to duplication of a lot of data (and things like us busting the browser cache for logos etc.).

Indeed that's what I meant but I am not sure why that requires to duplicate data? I'd assume that the backend would return always data from global namespaces.

> As above, this means that:
> `/chart/ns/<anything>/my-repo/wordpress` returns the chart from the global repository even if it's not the chart I intended (eg. I was specifically browsing the `my-repo` repo in another namespace when I selected `wordpress`.

That should return a 404 if the repository `my-repo` doesn't have a `wordpress` instance. If it's `bitnami/wordpress` then yes, it should return the global chart.

> I really don't (yet) see why we would not want to be able to uniquely address a chart in the assetsvc, which has implications for the dashboard. Let me know your thoughts.

I am reluctant to do so because that would add another vector of complexity, we would need to handle for every request two different namespaces which I believe is more .

absoludity:
> In that case, I think we need to redirect to the specific namespace. It's the same that we do if you list applications in "All namespaces", once you select an application it would redirect to the application namespace. If it's a global chart I would redirect it to `/catalog/ns/kubeapps/bitnami/wordpress`.

This doesn't work generally, as I think you mention below, as the `/ns/kubeapps` in the above URL is the value in the namespace switcher. So if we consider another user viewing the available charts for install in the `kubeapps-user-namespace`, which includes the globally available charts, when they click on `bitnami/wordpress` they cannot be redirected to the `kubeapps` namespace, as they don't have access to view that namespace. The user needs to be able to view a chart from another namespace in their target namespace, and the only unambiguous way I can see is `/ns/target-namespace/catalog/kubeapps/bitnami/wordpress`. To avoid this, you're suggesting to have different behaviour in other namespaces:

> This doesn't affect if the user is in their namespace so going to `/catalog/ns/my-ns/bitnami/wordpress` would have the same solution.

But this is ambiguous, and is trying to present chart repositories as though they are not namespaced, with a set of rules to ensure the name is unique without the namespace. 

> 
> This is not great from a "All namespaces" UX but it's the best I can think of since the apprepository is actually in the kubeapps namespace.
> 
> What we should treat carefully is namespaced repositories that use the same ID than global ones. If someone tries to define a repository `bitnami` in their namespace, that should return a conflict error.

So repository names aren't namespaced, but rather the repo name is required to be unique across namespaces. Note this doesn't just affect repos in the global namespace, as later when we allow users to share a repo across multiple namespaces we'd have the same issue - as long as we are trying to reference namespaced repositories ambiguously (ie. without the namespace).

> Indeed that's what I meant but I am not sure why that requires to duplicate data? I'd assume that the backend would return always data from global namespaces.

It does currently (we may later add a switch to enable just repos from the namespace), but the difference is that it returns the actual chart data, so the client sees that charts from the global namespace are from the global namespace (they have `Repo` properties which include namespace, and unique links which unambiguously identify them), where as I think you mean they should appear as if they were defined in that namespace, so `bitnami/wordpress` would appear as being from a repo `{name: 'bitnami', namespace: 'one-namespace'}` when viewed from one namespace, and `{name: 'bitnami', namespace: 'other-namespace'}` when viewed from `other-namespace`. 


> I am reluctant to do so because that would add another vector of complexity, we would need to handle for every request two different namespaces which I believe is more difficult to handle.

It's not required for every request, it is only when a URL is intended to identify a chart (from a specific namespace) to be installed in a specific namespace (ie. the new app URI, or viewing a potential chart (from a specific namespace) from the context of a target namespace).

> I prefer the complexity to be in the backend when possible.

Same, but I don't think we should try to avoid referencing namespaced repositories by their namespace as I think that will end up more complicated in the long run. Trying to narrow down the problem definition that we're discussing, I think we agree that we should be able to view the list of charts available for install in a namespace with:

`/ns/target-namespace/charts`

and that we should be able to view a specific chart within the target namespace with:

`/ns/target-namespace/charts/<chart-identifier>`

I think we differ only in the value of `<chart-identifier>`? In that you want it to be without the namespace of the chart, a typical chart-id that we're used to such as `bitnami/wordpress`, where as I think that to uniquely identify a chart within a namespaced repository in kubeapps, we must include that namespace: `kubeapps/bitnami/wordpress` (or `other-namespace/foo-repo/bar-chart`). Is that right? Sorry for the long reply - I'll aim to come back after you've had a chance to read so we can talk about it in higher bandwidth.

andresmgot:
> where as I think you mean they should appear as if they were defined in that namespace, so `bitnami/wordpress` would appear as being from a repo `{name: 'bitnami', namespace: 'one-namespace'}` when viewed from one namespace, and `{name: 'bitnami', namespace: 'other-namespace'}` when viewed from `other-namespace`.

I don't know if that's relevant but no, in that case, the repo information should be righ so it should return `{name: 'bitnami', namespace: 'kubeapps'}` regardless of the namespace.

> Note this doesn't just affect repos in the global namespace, as later when we allow users to share a repo across multiple namespaces we'd have the same issue - as long as we are trying to reference namespaced repositories ambiguously (ie. without the namespace).
> 

Here is where I think we are proposing different things. For the sake of simplicity, if a repo is not global, it should be copied, in the database, for each of the target namespaces. I am not sure if you were thinking on something else?

I am thinking here as if the charts were also instances of a CRD. You are either requesting a global instance (in this case, hosted in the Kubeapps namespace, but that can be hidden) or a resource in a namespace so you cannot request a namespaced resource if you are not in that namespace.

PR #1615, opened 2020-03-30
Show proper message when there are no workloads or URLs

andresmgot:
### Description of the change

Two small bug fixes:
 - When there are no workloads, say that instead of "Ready" since we cannot ensure that the application is actually ready.
 - When there are no URL items, skip the URL table

Before these changes:
Screenshot from 2020-03-30 17-11-13

After:
Screenshot from 2020-03-30 17-14-17

### Applicable issues

  - ref #1552

### DIFF
--- a/dashboard/src/components/AppView/AccessURLTable/AccessURLTable.test.tsx
+++ b/dashboard/src/components/AppView/AccessURLTable/AccessURLTable.test.tsx
@@ -5,7 +5,6 @@ import * as React from "react";
 import itBehavesLike from "../../../shared/specs";
 
 import ResourceRef from "shared/ResourceRef";
-import LoadingWrapper from "../../../components/LoadingWrapper";
 import { IIngressSpec, IResource, IServiceSpec, IServiceStatus } from "../../../shared/types";
 import AccessURLItem from "./AccessURLItem";
 import AccessURLTable from "./AccessURLTable";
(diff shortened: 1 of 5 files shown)


andresmgot:
> +1 to say something other than Ready, but would have thought it'd be better to do that only, ie. still show the circle, but rather than "Ready", just have "No workload" or similar? (so it still shows zero pods, but that's expected, just means it won't jump or change significantly when updating).
> 

The thing is that showing 0 Pods would be wrong. For example, the spark-cluster from the screenshot does have Pods (defined by a replication-controller that doesn't define an ownerReference). So if we say that there are no pods or simply "no workloads" it would be wrong: there are, but we cannot detect them.

Also, the circle is meant so show progress. If there is no progress to show I would find it a bit weird.

absoludity:
> The thing is that showing 0 Pods would be wrong. For example, the spark-cluster from the screenshot does have Pods (defined by a replication-controller that doesn't define an ownerReference). So if we say that there are no pods or simply "no workloads" it would be wrong: there are, but we cannot detect them.
> 
> Also, the circle is meant so show progress. If there is no progress to show I would find it a bit weird.

Oh, I thought it was a temporary state - ie. No pods/workloads currently (but there will be soon, ie. the UI would update). If that's not the case, great.

PR #1622, opened 2020-04-02
Consistent namespaced routes

absoludity:
This is some background cleanup which I need for pending work but is worthwhile in and of itself: use a consistent route for namespaced resources. The only change that needs reading is the Routes.tsx changes, the rest are just matching those new routes.

Most of the routes are just moving the namespaced-portion of the route to the front so they're consistent and not splitting nouns, the exception being the app upgrade route where I've also moved the 'upgrade' verb for consistency.. It also prepares the way for referencing namespaced chart ids, but even if that doesn't go forward (pending discussion), it's still worthwhile imo.


dashboard/src/containers/RoutesContainer/Routes.tsx
```diff
@@ -25,19 +25,19 @@ import ServiceInstanceViewContainer from "../../containers/ServiceInstanceViewCo
...
-  "/apps/ns/:namespace/new/:repo/:id/versions/:version": AppNewContainer,
-  "/apps/ns/:namespace/upgrade/:releaseName": AppUpgradeContainer,
-  "/catalog/ns/:namespace": CatalogContainer,
-  "/catalog/ns/:namespace/:repo": CatalogContainer,
+  "/ns/:namespace/apps": AppListContainer,
+  "/ns/:namespace/apps/:releaseName": AppViewContainer,
+  "/ns/:namespace/apps/:releaseName/upgrade": AppUpgradeContainer,
+  "/ns/:namespace/apps/new/:repo/:id/versions/:version": AppNewContainer,
+  "/ns/:namespace/catalog": CatalogContainer,
+  "/ns/:namespace/catalog/:repo": CatalogContainer,
```
andresmgot, line 33: I guess we would need to change also the routes for operators (line 70). I can take care of that

PR #1625, opened 2020-04-02
Catch forbidden errors when checking if OLM is installed

andresmgot:
### Description of the change

When the user doesn't have permissions to list operators, we were showing the message for the OLM being missing instead of the forbidden message. This PR fixes that so the proper message is displayed:

Screenshot from 2020-04-02 13-08-20

### Applicable issues

  - Ref #1552


dashboard/src/actions/operators.test.tsx
```diff
@@ -44,6 +44,26 @@ describe("checkOLMInstalled", () => {
...
+      {
+        type: getType(operatorActions.checkingOLM),
+      },
+      {
+        type: getType(operatorActions.errorOperators),
+        payload: new Error("nope"),
+      },
+      {
+        type: getType(operatorActions.OLMNotInstalled),
+      },
```
absoludity, line 62: It's not obvious to me why we'd still dispatch an OLMNotInstalled after dispatching an error: we know there was an error, we don't know that the OLM is not installed?

dashboard/src/components/OperatorList/OperatorList.tsx
```diff
@@ -52,16 +52,27 @@ class OperatorList extends React.Component<IOperatorListProps> {
...
+          error={error}
+          action="list"
+          resource="Operators"
+          namespace={this.props.namespace}
+        />
+      );
+    }
+    if (!isOLMInstalled) {
+      return <OLMNotFound />;
+    }
```
absoludity, line 75: Ah - is it because here we're only handling the one specific type of error, and ignoring any other, hence needing the OLMNotInstalled? Why not just handle the error?

dashboard/src/actions/operators.test.tsx
andresmgot: Well, yes, at the beginning this was me trying to be extra-verbose. `OLMNotInstalled` action is indeed not needed (even less now that this may be triggered if we are not able to verify if it's installed).

andresmgot:
_Unrelated error in the CI, merging this_

PR #1628, opened 2020-04-03
Fix namespaced chart click pt 2

absoludity:
Follows on from #1614, ref #1521 .

Based on discussions with @andresmgot , this PR is (I think) a good compromise between two viewpoints: the dashboard route does not require a second namespace for the chart (which could be confusing) to be able to find the chart, but the url does indicate whether it's a global chart or local chart.

This in turn means that the dashboard has enough info to make *namespaced* requests to the assetsvc for all chart resources - either the current namespace or the global one, so the assetsvc is not required to do crazy things like returning `bitnami/wordpress` from the `kubeapps` namespace when requesting it from `other-namespace`, which in turn also means we don't need to worry about duplicate repo names in global/local namespaces.

See what you think. It can land, but I still need to follow up with a fix for deploying a chart from a non-global namespace (it's still looking for the repo in the kubeapps namespace).


dashboard/src/components/Catalog/CatalogItem.tsx
```diff
@@ -37,18 +38,19 @@ function trimDescription(desc: string): string {
 
 const CatalogItem: React.SFC<ICatalogItemProps> = props => {
   const { item } = props;
-  const { icon, name, repoName, version, description, type, namespace, id, csv } = item;
+  const { icon, name, repo, version, description, type, namespace, id, csv } = item;
   const iconSrc = icon || placeholder;
   let link;
   let tag1;
   let subIcon;
```
andresmgot, line 45: if the repo is mandatory and it's needed to render this card, you can add a check `if (!repo) return null` or smth like that to avoid all the `repo?.` that may indicate that the repo is not mandatory

dashboard/src/components/Catalog/CatalogItem.tsx
```diff
@@ -37,18 +38,19 @@ function trimDescription(desc: string): string {
...
     tag1 = (
-      <Link className="ListItem__content__info_tag_link" to={`/catalog/${repoName}`}>
-        {repoName}
+      <Link className="ListItem__content__info_tag_link" to={`/catalog/${repo?.name}`}>
+        {repo?.name}
       </Link>
     );
-    link = `/charts/${repoName}/${name}`;
+    const chartsSegment = namespace !== repo?.namespace ? "global-charts" : "charts";
+    link = `/ns/${namespace}/${chartsSegment}/${repo?.name}/${name}`;
```
andresmgot, line 53: feel free to ignore this suggestion but as a matter of style I would prefer the global route to be something like `ns/foo/charts/bitnami/wordpress?global=true`. Also, with that you don't need to define two different routes for the same container.

dashboard/src/components/ChartView/ChartDeployButton.tsx
```diff
@@ -31,7 +33,7 @@ class ChartDeployButton extends React.Component<IChartDeployButtonProps, IChartD
         {this.state.clicked && (
           <Redirect
             push={true}
-            to={`/ns/${namespace}/apps/new/${repoName}/${chartName}/versions/${versionStr}`}
+            to={`/ns/${namespace}/apps/${newSegment}/${repoName}/${chartName}/versions/${versionStr}`}
```
andresmgot, line 36: same here about the query param `global`

dashboard/src/components/Catalog/CatalogItem.tsx
absoludity: > if the repo is mandatory and it's needed to render this card, you can add a check `if (!repo) return null` or smth like that

That would be swallowing an error in a different way, imo? We need to be rendering something - a chart or an operator. If it's a chart, then the repo is required. Currently we would swallow the error by rendering a chart with invalid links etc. If this was the way to fix it, I'd instead suggest throwing an error if it is a chart type and repo is null, but...

> to avoid all the `repo?.` that may indicate that the repo is not mandatory

So the `repo` is necessary if the catalog item is a chart. I think the issue might be the data modelling here: we should be specifying requirements for a chart catalog item separately from an operator catalog item (though they may share many properties), so that this type requirement (that a repo is required for a chart) can be explicit. Thoughts?

dashboard/src/components/Catalog/CatalogItem.tsx
absoludity: That would be using the query params for hierarchical data, rather than for non-hierarchical data (sorting / filtering), which isn't great, imo. That is, the Uniform Resource Identifier should identify the hierarchical resource using the path only (see https://tools.ietf.org/html/rfc3986#section-3.4, or better explained specifically for restful apis at https://stackoverflow.com/questions/30967822/when-do-i-use-path-params-vs-query-params-in-a-restful-api ).

The easy test here is, does the path without a query params identify a resource? This is why either `global` or the explicit chart namespace is required in the URI IMO, to identify the resource unambiguously.

dashboard/src/components/ChartView/ChartDeployButton.tsx
absoludity: Yeah, for similar reasons, using a query param would indicate an optional, non-hierarchical, piece of data - this URI path unambiguously identifies the hierarchy of the chart (ie. a chart in the global repository called `${repoName}` with the specific `${chartName}` and `${versionStr}`.

PR #1633, opened 2020-04-04
Update imgs in getting started doc. Re-word/order Step4

tiffanyfay:
### Description of the change
Updated all outdated images in getting started. I also cropped several since they take up unnecessary space. 

Step 4 made it sound like you the credentials were to access the deployment, but they're just for logging in to the WordPress account which isn't needed to view it or show how Kubeapps itself works; I edited it as a result. Added "note" to the start of the note to match consistency with a previous one (however they still don't match the GKE note).

### Benefits
The docs now match what things look like when you start. Screenshots take up less space so the doc is smaller/there's less scrolling involved. Step 4 is more clear.

### Additional information
https://github.com/tiffanyfay/kubeapps/blob/docs-getting-started/docs/user/getting-started.md to see how it compares to https://github.com/kubeapps/kubeapps/blob/master/docs/user/getting-started.md


docs/user/getting-started.md
```diff
@@ -94,27 +93,31 @@ Paste the token generated in the previous step to authenticate and access the Ku
...
 
   ![WordPress installation](../img/wordpress-installation.png)
 
-- Click the "Submit" button. The application will be deployed. You will be able to track the new Kubernetes deployment directly from the browser.
+- Click the "Submit" button. The application will be deployed. You will be able to track the new Helm deployment directly from the browser. The status will be shown at the top and you can also look at the individual resources lower in the page.
 
   ![WordPress deployment](../img/wordpress-deployment.png)
 
-To obtain the WordPress username and password, refer to the "Notes" section of the deployment page, which contains the commands you will need to run to obtain the credentials for the deployment.
+To access your new WordPress site, you can run commands in the "Notes" section to get the URLs or simply click a URL (HTTP and HTTPS) shown.
```
absoludity, line 112: "you can run **the** commands in the..." ?

PR #1643, opened 2020-04-07
Fix broken link in Deploy App buton

andresmgot:
### Description of the change

The button that takes to the catalog in the Application List is broken due to the change in the routes.

The same happens for the link to the repo in the cards of the catalog.


absoludity:
@andresmgot This would be a blocker for the release right? (I thought I had the same fix in my stacked PR, but turns out I've not rebased to current master with the new routes).

andresmgot:
The images have been already released but we can just release v1.9.1

PR #1645, opened 2020-04-08
Update docs code blocks chart/kubeapps/readme

tiffanyfay:
### Description of the change
* Removes $ from code commands because it makes copy-pasting easier. Otherwise you have to make sure you skip the $, or you get an error (at least on my mac with zsh).
* Swaps `console` type to `bash` type for consistency and also because console is only really useful if you're showing commands to be run and the output to distinguish between them.

### Benefits
See above

### DIFF
--- a/chart/kubeapps/README.md
+++ b/chart/kubeapps/README.md
@@ -47,8 +47,8 @@ To install the chart with the release name `kubeapps`:
 For Helm 2:
 
 ```console
-$ helm repo add bitnami https://charts.bitnami.com/bitnami
-$ helm install --name kubeapps --namespace kubeapps bitnami/kubeapps
+helm repo add bitnami https://charts.bitnami.com/bitnami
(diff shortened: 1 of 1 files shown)


tiffanyfay:
Should the two blocks in https://github.com/kubeapps/kubeapps/pull/1645/files#diff-fd1223ba99ae8afb3850192caf05b551L198 be `bash` also? I didn't put anything since they're errors.

absoludity:
> Should the two blocks in https://github.com/kubeapps/kubeapps/pull/1645/files#diff-fd1223ba99ae8afb3850192caf05b551L198 be `bash` also? I didn't put anything since they're errors.

The link doesn't drop me at a specific line, but I assume you mean the ones at:

https://github.com/kubeapps/kubeapps/blob/master/chart/kubeapps/README.md#troubleshooting

then I'm not sure it matters (trying it here in the editor with or without doesn't change the presentation, and not sure it matters semantically for the errors). I'll merge as is :)

PR #1650, opened 2020-04-08
Fix card shrink in Catalog

andresmgot:
### Description of the change

When there are less than 4 cards in the catalog, these cards shrink 

Screenshot from 2020-04-08 17-32-51

We need to set the CSS properties to fill the whole row.

Screenshot from 2020-04-08 17-40-11

### DIFF
--- a/dashboard/src/components/Catalog/Catalog.tsx
+++ b/dashboard/src/components/Catalog/Catalog.tsx
@@ -143,7 +143,7 @@ class Catalog extends React.Component<ICatalogProps, ICatalogState> {
                 </div>
               </div>
             )}
-            <div className={this.shouldRenderOperators() ? "col-10" : ""}>
+            <div className={this.shouldRenderOperators() ? "col-10" : "col-12"}>
               <CardGrid>{items}</CardGrid>
             </div>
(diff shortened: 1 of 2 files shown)


dashboard/src/components/Catalog/Catalog.tsx
```diff
@@ -143,7 +143,7 @@ class Catalog extends React.Component<ICatalogProps, ICatalogState> {
                 </div>
               </div>
             )}
-            <div className={this.shouldRenderOperators() ? "col-10" : ""}>
+            <div className={this.shouldRenderOperators() ? "col-10" : "col-12"}>
```
absoludity, line 146: Why is this dependent on whether we're rendering operators? That is, why not use the same width ("col-10") either way? If it's because the filter is not present, then we should use a function that makes that obvious, perhaps?

andresmgot: yes, it's because if there are no operators we are skipping the column that allows you to filter by chart/operator. This function is used several times (line 103 and 122), I guess it's more obvious in the full context of the function.

PR #1652, opened 2020-04-09
Fix namespaced chart deploy backend

absoludity:
Updates ChartDetails to include both the app repo namespace as well as the user token from the request, so that requests can be made on behalf of the user for the app repo in the specified namespace.

Ref #1521.

### DIFF
--- a/pkg/handlerutil/handlerutil.go
+++ b/pkg/handlerutil/handlerutil.go
@@ -79,9 +79,8 @@ func ParseAndGetChart(req *http.Request, cu chartUtils.Resolver, requireV1Suppor
 	if err != nil {
 		return nil, nil, err
 	}
-	chartDetails.UserToken = auth.ExtractToken(req.Header.Get("Authorization"))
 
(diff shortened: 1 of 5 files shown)


pkg/handlerutil/handlerutil.go
```diff
@@ -78,6 +79,8 @@ func ParseAndGetChart(req *http.Request, cu chartUtils.Resolver, requireV1Suppor
 	if err != nil {
 		return nil, nil, err
 	}
+	chartDetails.UserToken = auth.ExtractToken(req.Header.Get("Authorization"))
```
andresmgot, line 82: I would rather use a different param for the UserToken for two reasons: that way we are not returning a partially empty struct from `ParseDetails` and also because we avoid having non-related chart info y the chart details.

absoludity: Yep, makes sense. I'll separate `UserToken` out of `ChartDetails` tomorrow. Thanks.

PR #1660, opened 2020-04-13
Update chart requirements

andresmgot:
### Description of the change

It's no longer needed to use the version 7.8.10 for MongoDB


absoludity:
> 
> It's no longer needed to use the version 7.8.10 for MongoDB

s/needed/possible/ ?

andresmgot:
the chart is missing in some Azure registry so it's not possible to use that chart there

PR #1661, opened 2020-04-13
Omit chart sources if they don't exist

andresmgot:
### Description of the change

I was trying to deploy chartmuseum with Kubeapps and it no longer contains `sources` in the chart information which caused a crash in the application.


andresmgot:
_merging this to tag a new release_

Issue #1522, opened 2020-02-17
Support docker registry secrets for AppRepositories


absoludity:
As part of implementing support for app repositories using private docker registries (#1256), we need to support linking docker registry secrets to AppRepositories. See [design doc](https://docs.google.com/document/d/1YEeKC6nPLoq4oaxs9v8_UsmxrRfWxB6KCyqrh2-Q8x0/edit?ts=5e2adf87#) for more info and discussion.

- [x] Update the AppRepository resource to support the additional field dockerRegistrySecrets (a list of strings containing the names of any secrets that should be made available by the repo)
- [x] [UI] Enable selecting from existing dockerconfigjson secrets in the given namespace, as required for an AppRepository.
- [x] [UI] Enable adding new dockerconfigjson secrets while creating/editing an AppRepository
- [x] Ensuring that repos available in all namespaces cannot have docker registry secrets added (as the secrets are available only in the namespace of the app repository).

PR #1662, opened 2020-04-14
Docker registry secrets

absoludity:
Ref: #1522 

Enables AppRepositories to be created with a list of docker registry secrets which should be included for matching registries.

Also ensures that a global AppRepository is not able to be created (or validated) with docker registry secrets.

### DIFF
--- a/cmd/apprepository-controller/pkg/apis/apprepository/v1alpha1/types.go
+++ b/cmd/apprepository-controller/pkg/apis/apprepository/v1alpha1/types.go
@@ -36,15 +36,15 @@ type AppRepository struct {
 
 // AppRepositorySpec is the spec for an AppRepository resource
 type AppRepositorySpec struct {
-	Type string            `json:"type"`
-	URL  string            `json:"url"`
-	Auth AppRepositoryAuth `json:"auth,omitempty"`
+	Type               string                 `json:"type"`
+	URL                string                 `json:"url"`
+	Auth               AppRepositoryAuth      `json:"auth,omitempty"`
+	ResyncRequests     uint                   `json:"resyncRequests"`
+	SyncJobPodTemplate corev1.PodTemplateSpec `json:"syncJobPodTemplate"`
 	// DockerRegistrySecrets is a list of dockerconfigjson secrets which exist
 	// in the same namespace as the AppRepository and should be included
 	// automatically for matching images.
-	DockerRegistrySecrets []string               `json:"dockerRegistrySecrets,omitempty"`
-	ResyncRequests        uint                   `json:"resyncRequests"`
-	SyncJobPodTemplate    corev1.PodTemplateSpec `json:"syncJobPodTemplate"`
+	DockerRegistrySecrets []string `json:"dockerRegistrySecrets,omitempty"`
 }
 
 // AppRepositoryAuth is the auth for an AppRepository resource
--- a/cmd/apprepository-controller/pkg/apis/apprepository/v1alpha1/types.go
+++ b/cmd/apprepository-controller/pkg/apis/apprepository/v1alpha1/types.go
@@ -36,11 +36,15 @@ type AppRepository struct {
 
 // AppRepositorySpec is the spec for an AppRepository resource
 type AppRepositorySpec struct {
-	Type               string                 `json:"type"`
-	URL                string                 `json:"url"`
-	Auth               AppRepositoryAuth      `json:"auth,omitempty"`
-	ResyncRequests     uint                   `json:"resyncRequests"`
-	SyncJobPodTemplate corev1.PodTemplateSpec `json:"syncJobPodTemplate"`
+	Type string            `json:"type"`
+	URL  string            `json:"url"`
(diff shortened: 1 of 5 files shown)


pkg/kube/kube_handler.go
```diff
@@ -139,10 +139,13 @@ type appRepositoryRequestDetails struct {
 	RepoURL            string                 `json:"repoURL"`
 	AuthHeader         string                 `json:"authHeader"`
 	CustomCA           string                 `json:"customCA"`
+	RegistrySecrets    []string               `json:"registrySecrets"`
 	SyncJobPodTemplate corev1.PodTemplateSpec `json:"syncJobPodTemplate"`
 	ResyncRequests     uint                   `json:"resyncRequests"`
 }
 
+var ErrGlobalRepositoryWithSecrets = fmt.Errorf("docker registry secrets cannot be set for app repositories available in all namespaces")
```
andresmgot, line 147: you need to add a comment for exported var or names (the linter complains for me)

pkg/http-handler/http-handler.go
```diff
@@ -80,7 +80,7 @@ func CreateAppRepository(handler kube.AuthHandler) func(w http.ResponseWriter, r
 func ValidateAppRepository(handler kube.AuthHandler) func(w http.ResponseWriter, req *http.Request) {
 	return func(w http.ResponseWriter, req *http.Request) {
 		token := auth.ExtractToken(req.Header.Get("Authorization"))
-		res, err := handler.AsUser(token).ValidateAppRepository(req.Body)
+		res, err := handler.AsUser(token).ValidateAppRepository(req.Body, mux.Vars(req)["namespace"])
```
andresmgot, line 83: this is unrelated to the docker registries, right? (it's to run the validation in the proper namespace)
[EDIT] A well, I see that it's used to check if the namespace is the global one and have secrets

pkg/kube/kube_handler.go
```diff
@@ -139,10 +139,13 @@ type appRepositoryRequestDetails struct {
 	RepoURL            string                 `json:"repoURL"`
 	AuthHeader         string                 `json:"authHeader"`
 	CustomCA           string                 `json:"customCA"`
+	RegistrySecrets    []string               `json:"registrySecrets"`
```
andresmgot, line 142: So this is a different approach than what we do with repositories credentials. There we ask for credentials rather than secrets and we create those secrets on behalf of the user. That has the advantage that we can use garbage collection to clean up stuff and it's easier for people to use it. Is you plan to move to that approach once we support setting multiple namespaces for an AppRepository maybe?

cmd/apprepository-controller/pkg/apis/apprepository/v1alpha1/types.go
```diff
@@ -36,11 +36,15 @@ type AppRepository struct {
...
-	Auth               AppRepositoryAuth      `json:"auth,omitempty"`
-	ResyncRequests     uint                   `json:"resyncRequests"`
-	SyncJobPodTemplate corev1.PodTemplateSpec `json:"syncJobPodTemplate"`
+	Type string            `json:"type"`
+	URL  string            `json:"url"`
+	Auth AppRepositoryAuth `json:"auth,omitempty"`
+	// DockerRegistrySecrets is a list of dockerconfigjson secrets which exist
+	// in the same namespace as the AppRepository and should be included
+	// automatically for matching images.
+	DockerRegistrySecrets []string               `json:"dockerRegistrySecrets,omitempty"`
```
andresmgot, line 45: nit: I would move this param to the last (at first sight I thought you were adding 3 new params to the spec)

absoludity: Done.

pkg/kube/kube_handler.go
absoludity: > So this is a different approach than what we do with repositories credentials. There we ask for credentials rather than secrets and we create those secrets on behalf of the user.

The change here doesn't necessarily mean a different approach, rather it's just the way we link any secrets to the app repository (via the name, with the necessary assumption of the same namespace). The UI can still require credentials from the user which we use to create the secrets, or we could allow selecting from the available dockerconfigjson secrets already present.

> That has the advantage that we can use garbage collection to clean up stuff and it's easier for people to use it.

We chatted about this, but just ftr, I don't think these dockerconfigjson secrets should be garbage collected when the app repository is deleted. It should be no problem to delete the app repository and still leave an existing deployment which was deployed from that app repository. Furthermore, if a user increases the replica sets and has a pod spawned on a different node, that node should be able to pull the image from the private docker registry. So the docker registry secrets are a little different from the app repository secrets (ie. they're not for accessing things to install, but rather are run-time requirements)

> Is your plan to move to that approach once we support setting multiple namespaces for an AppRepository maybe?

Not necessarily. The important thing there is that it should be easy and intuitive to link the secrets to an app repo, and we'd take care of the rest (ensuring those secrets are available in the other namespaces using the users creds).

PR #1665, opened 2020-04-14
Automatic code-style change

andresmgot:
### Description of the change

cc/ @absoludity I got this after doing some commit. I guess this is because you are skipping the pre-commit hook?


absoludity:
> 
> cc/ @absoludity I got this after doing some commit. I guess this is because you are skipping the pre-commit hook?

Yep, and CI is not failing if it's a no-op.

PR #1669, opened 2020-04-14
Fix OLM check for older k8s versions

### LINKED ISSUE #1668: OLM installation check not working with Kubernetes <1.17
### Description:

While developing this feature, I was using a local cluster using Kubernetes 1.17. In this version, CRD version is `v1` while in previous versions it was `v1beta1`. Since we are using the presence of the CRD `clusterserviceversion` as the check to know if OLM is installed, this always return false for Kubernetes clusters prior to 1.17 (GKE latest version is 1.15).

andresmgot:
### Description of the change

Fixes #1668

Rather than checking if the CRD exists, we are listing operators. If the OLM is installed it should return the list of Operators (even if the list is an empty array).

### Applicable issues

  - fixes #1668


dashboard/src/shared/Operators.ts
```diff
@@ -1,10 +1,11 @@
 import * as urls from "../shared/url";
 import { axiosWithAuth } from "./AxiosInstance";
+import { definedNamespaces } from "./Namespace";
 import { IClusterServiceVersion, IK8sList, IPackageManifest, IResource } from "./types";
 
 export class Operators {
   public static async isOLMInstalled() {
-    const { status } = await axiosWithAuth.get(urls.api.operators.crd);
+    const { status } = await axiosWithAuth.get(urls.api.operators.operators(definedNamespaces.all));
```
absoludity, line 8: Does this work if you don't have RBAC for all namespaces?

andresmgot: well, no, typically the admin should have cluster-wide privileges but it's true that it's not necessary for this

Issue #1617, opened 2020-03-30
Deploying with private docker registries


absoludity:
Following on from #1522, as part of implementing support for app repositories with private docker registries (#1256), we need to ensure that when deploying or upgrading any resources, we can identify docker registry domains for which we have secrets available, and ensure `imagePullSecrets` are included for the relevant pod specifying the required secrets (if not already present). See [design doc](https://docs.google.com/document/d/1YEeKC6nPLoq4oaxs9v8_UsmxrRfWxB6KCyqrh2-Q8x0/edit?ts=5e2adf87#) for more info and discussion.

- [x] Add non-exposed functionality to backend to build map of registry domains to k8s secret names by querying secrets of type kubernetes.io/dockerconfigjson in the target namespace using the users credentials.
- [x] Use this map to parse rendered yaml templates, identifying image registry domains and appending required imagePullSecrets to the rendered yaml templates.
- [x] Update the (potentially helm3-only) install/upgrade functions to render the full template first and apply via k8s (ie. verify this works without any actual changes)
- [x] Behind a feature-flag, switch the (potentially helm3-only) install/upgrade functions to use the modified rendered yaml templates including the extra imagePullSecrets where present.

PR #1670, opened 2020-04-15
Add initial helm post renderer.

absoludity:
Ref #1617 . Adds an initial helm post-renderer which will be updated in subsequent PRs to add the image pulls secret when relevant.


pkg/agent/docker_secrets_postrenderer.go
```diff
@@ -0,0 +1,57 @@
...
+		}
+		resourceList = append(resourceList, resource)
+	}
+
+	// TODO: Update relevant container images with image pull secrets
+
+	modifiedManifests = bytes.NewBuffer([]byte{})
+	encoder := yaml.NewEncoder(modifiedManifests)
+	defer encoder.Close()
+	encoder.SetIndent(2)
```
andresmgot, line 47: isn't this the default?

pkg/agent/docker_secrets_postrenderer_test.go
```diff
@@ -0,0 +1,91 @@
...
+  - command:
+    - sh
+    - -c
+    - echo 'foo'
+    env:
+    - name: SOME_ENV
+      value: env_value
+    image: example.com/bitnami/nginx:1.16.1-debian-10-r42
+    name: container-name
+  restartPolicy: Never
```
andresmgot, line 69: mmh, I see it's in alphabetical order, what does it? Is it reliable?

pkg/agent/docker_secrets_postrenderer.go
```diff
@@ -0,0 +1,57 @@
+package agent
+
+import (
+	"bytes"
+	"io"
+
+	"gopkg.in/yaml.v3"
```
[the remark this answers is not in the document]
absoludity: It was just to use the latest. The only difference that's relevant is that v3 has a default tab intent of 4 spaces, as well as a `SetIndent()` function. I'll switch to v2 since I'd not realised v3 isn't yet stable/released.

pkg/agent/docker_secrets_postrenderer.go
absoludity: It is the default for v2 but not v3.

pkg/agent/docker_secrets_postrenderer_test.go
absoludity: Yes, the output is reliably ordered. I assume the reason is that when the import happens, ordered yaml dicts are imported into explicitly unordered go maps.

PR #1671, opened 2020-04-16
Implement adding imagePullSecrets to Pod resources.

absoludity:
Ref #1617. Implements adding imagePullSecrets to Pod resources (only - I'll update to handle other resources which also include pod templates).

### DIFF
--- a/pkg/agent/docker_secrets_postrenderer.go
+++ b/pkg/agent/docker_secrets_postrenderer.go
@@ -2,7 +2,9 @@ package agent
 
 import (
 	"bytes"
+	"fmt"
 	"io"
+	"strings"
 
 	"gopkg.in/yaml.v2"
 )
@@ -39,7 +41,27 @@ func (r *DockerSecretsPostRenderer) Run(renderedManifests *bytes.Buffer) (modifi
 		resourceList = append(resourceList, resource)
 	}
 
-	// TODO: Update relevant container images with image pull secrets
+	// TODO(mnelson): If re-rendering the entire manifest creates issues, we
+	// could instead find the correct byte position and insert the image pull
+	// secret into the byte stream at the relevant points, but this will be
+	// more complex.
+	for _, resourceItem := range resourceList {
+		resource, ok := resourceItem.(map[interface{}]interface{})
+		if !ok {
+			continue
+		}
+		podSpec, err := getResourcePodSpec(resource)
+		if err != nil {
+			return nil, err
+		}
+		if podSpec == nil {
+			continue
+		}
+		err = r.updatePodSpecWithPullSecrets(podSpec)
+		if err != nil {
+			return nil, err
+		}
+	}
 
 	modifiedManifests = bytes.NewBuffer([]byte{})
 	encoder := yaml.NewEncoder(modifiedManifests)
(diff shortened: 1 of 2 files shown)


pkg/agent/docker_secrets_postrenderer.go
```diff
@@ -54,3 +76,105 @@ func (r *DockerSecretsPostRenderer) Run(renderedManifests *bytes.Buffer) (modifi
...
+		}
+		image, ok := container["image"].(string)
+		if !ok {
+			// NOTE: in https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#container-v1-core
+			// the image is optional to allow higher level config management to default or override (such as
+			// deployments or statefulsets), but both only define pod templates which in turn define containers?
+			return fmt.Errorf("pod container did not contain an image reference: %+v", container)
+		}
+
+		// Ignore (dockerhub) image refs without a domain.
```
andresmgot, line 126: This comment makes more sense within the `if` block below. Also, what happens if the user wants to add an imagePullSecret for a private image in the dockerhub?

pkg/agent/docker_secrets_postrenderer.go
```diff
@@ -54,3 +76,105 @@ func (r *DockerSecretsPostRenderer) Run(renderedManifests *bytes.Buffer) (modifi
...
+		domainSplit := strings.Split(image, "/")
+		if len(domainSplit) == 1 {
+			continue
+		}
+		imageDomain := domainSplit[0]
+
+		secretName, ok := r.secrets[imageDomain]
```
andresmgot, line 133: I am not sure if we are just interested in a domain? For example, for a gcr container (`gcr.io/my_project/image:tag`), should we check `gcr.io` or `gcr.io/my-project`?

pkg/agent/docker_secrets_postrenderer_test.go
```diff
@@ -89,3 +150,243 @@ other: doc
...
+			if got, want := err != nil, tc.expectErr; got != want {
+				t.Fatalf("got: %t, want: %t. err: %+v", got, want, err)
+			}
+
+			if got, want := result, tc.result; !cmp.Equal(got, want) {
+				t.Errorf("mismatch (-want +got):\n%s", cmp.Diff(want, got))
+			}
+		})
+	}
+}
```
andresmgot, line 392: nice test coverage :+1:

pkg/agent/docker_secrets_postrenderer.go
```diff
@@ -54,3 +76,105 @@ func (r *DockerSecretsPostRenderer) Run(renderedManifests *bytes.Buffer) (modifi
...
+// we limit our assumptions of the untyped handling to the following, with
+// clear errors when the assumption fails:
+// - The pod spec includes a 'containers' key with a slice value (with a clear error if this is not the case)
+// - Each container value is a map with an 'image' key and string value.
+func (r *DockerSecretsPostRenderer) updatePodSpecWithPullSecrets(podSpec map[interface{}]interface{}) error {
+	containersObject, ok := podSpec["containers"]
+	if !ok {
+		return fmt.Errorf("pod spec did not include a containers key: %+v", podSpec)
+	}
+	containers, ok := containersObject.([]interface{})
```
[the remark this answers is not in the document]
absoludity: > there is a package in `client-go` that can help with this kind of conversions and parameters retrieval: https://godoc.org/github.com/contiv/client-go/pkg/apis/meta/v1/unstructured

You meant to link to the one owned by kubernetes in apimachinery (rather than `contiv`) right? https://godoc.org/github.com/kubernetes/apimachinery/pkg/apis/meta/v1/unstructured

If Helm's API passed a JSON stream, rather than a YAML stream, I could use that nicely. But as it is, I'd either need to convert the YAML to JSON to parse into the `unstructured.Unstructured`, update, then marshal it out to JSON (using `unstructured` and then back to YAML, or explicitly convert the `map[interface{}]interface{}` to `map[string]interface{}` to be able to use `unstructured.Unstructured.SetUnstructuredContent()` to create the object in the first place. Let me know if I've missed something - happy to try.

> 
> Also, I would try to use the actual struct types of the resources we are handling instead keep using type conversions from `interface{}` and literal keys. It's easier to just access `podSpec.Containers[0].Image` rather than doing several conversions.

Yes, I had gone this way initially for the same reason, but realized that we'd then be making assumptions about the templated YAML that are not ours to make (ie. which client-go version we're using affects the resources that we import). IMO we should be heading the other way - ideally we would not change the rest of the YAML at all (no changes in ordering, no re-rendering except the addition of the image pull secret), but to do that is a little more complex.

> It's even more useful when we add more complex types like a `CronJob` in which I don't necessarily know the full path till the image name.

They're actually pretty trivial - they just have a `spec.template.spec` as do all the others (I left a comment linking to https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#podtemplatespec-v1-core ).

pkg/agent/docker_secrets_postrenderer.go
absoludity: So it turns out that for dockerhub the secret's registry domain will be the FQDN: https://index.docker.io/v1 . I'll deal with this in a followup.

pkg/agent/docker_secrets_postrenderer.go
absoludity: Docker credentials are per server (though it turns out this can be the domain name or the fqdn of the server).

PR #1673, opened 2020-04-17
Handle dockerhub secrets special case. Allow optional protocol.

absoludity:
Ref: #1617 . Updates the imagePullSecret injection in the post renderer for two cases:

- Dockerhub images use "docker.io" as the registry domain in the image ref (docker's `reference` library even adds it as such), while the credential in docker configs must be keyed with "https://index.docker.io/v1/".
- general registry domains in the docker credentials can be listed with or without an explicit protocol (https), so we update to append the imagePullSecret as long as the domain itself matches.


absoludity:
Closing as I need to target this branch in the kubeapps repo.

Issue #1666, opened 2020-04-14
Allow to edit AppRepositories


andresmgot:
### Description:

The AppRepository form is evolving in complexity, it would be good for users to be able to edit existing AppRepositories without having to delete and re-create them.

PR #1676, opened 2020-04-17
Allow to modify the existing App Repositories

andresmgot:
### Description of the change

This PR adds the possibility of modifying existing AppRepositories from the dashboard:

Peek 2020-04-17 16-55

It's still needed to implement this functionality in the backend.

Now, apart from fetching apprepositories, we are fetching secrets owned by an apprepository. With that information we recover the information shown in the form so the user can re-edit them.

Note: There were an "update" method that was being used to just increase the resyncRequests. This functionality has been moved to its own method since the workflow is different.

### Applicable issues

  - ref #1666 

### Additional information

I thought this one was an easy one, it turned out not to be :)


dashboard/src/components/Config/AppRepoList/AppRepoList.tsx
```diff
@@ -117,17 +130,27 @@ class AppRepoList extends React.Component<IAppRepoListProps> {
                       repo={repo}
                       renderNamespace={renderNamespace}
                       namespace={namespace}
+                      errors={errors}
+                      validating={validating}
+                      validate={validate}
+                      secret={repoSecrets.find(secret =>
```
absoludity, line 136: So it's just the first secret which has the app repo as owner (which should be safe for now, since we won't want ownership for the registry secrets).

dashboard/src/actions/repos.ts
```diff
@@ -130,12 +144,36 @@ export const fetchRepos = (
       dispatch(requestRepos(namespace));
       const repos = await AppRepository.list(namespace);
       dispatch(receiveRepos(repos.items));
+      const secrets = await Secret.list(namespace);
+      const repoSecrets = secrets.items?.filter(s =>
```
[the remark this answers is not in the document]
absoludity: On a related thought, would it not be neater to actually use kubeapps' own API to retrieve the app repositories in the format expected? (ie. with the auth header etc., similar to the format that we submit them in to the backend).

dashboard/src/actions/repos.ts
```diff
@@ -175,6 +205,36 @@ export const installRepo = (
...
+      const syncJobPodTemplateObj = parsePodTemplate(syncJobPodTemplate);
+      const ns = getTargetNS(getState, namespace);
+      dispatch(requestRepoUpdate());
+      const data = await AppRepository.update(
+        name,
+        ns,
+        repoURL,
+        authHeader,
+        customCA,
+        syncJobPodTemplateObj,
```
absoludity, line 227: Like this - when updating the rep, we just pass the relevant data to the backend. Similarly, we could have a getRepo which returns the relevant data for the frontend (or other consumers) without them needing to separately load secrets etc.

absoludity:
Also, we might need to replace that matrix of buttons with more subtle delete, edit, refresh icons on hover?

andresmgot:
> I'm OK with this landing, but think we'd have a simpler dashboard implementation if we first added a backend `apprepositories.get` (both singular and plural) which returned the json including the app repo credentials, similar to the existing `create` and the `update` which you added.

yep, I agree. I started this PR assuming there would be no changes in the backend which turned out to be something not worth doing. Since we will need changes in the backend anyway we can include `get` and others if we have time. I will use this branch as base and start merging PRs here.

One comment worth mentioning is that even if we have a `get` method we would need some processing because the format of the API is a bit different than the format of the form (again my fault). We should still translate the AuthHeader to a username:password for example. In any case it's true that we wouldn't need to fetch secrets here.

andresmgot:
> Also, we might need to replace that matrix of buttons with more subtle delete, edit, refresh icons on hover?

Agree

andresmgot:
I am hiding the Modify/Edit button to be able to merge this PR and unblock the registry secrets listing

PR #1677, opened 2020-04-20
Update helm module and link post renderer.

absoludity:
Uses the post renderer currently without any secrets (which results in a no-op for the postrenderer).

I had to update to helm 3.1.2 to use the postrenderer arg. Unrelated, but 3.1.2 had [a change which updated the helm Memory driver to be namespace aware](https://github.com/helm/helm/commit/e6d2d10bad8a872478783b0b1483cf467c05741b), which required a few changes to existing test suits.

Ref: #1617


cmd/kubeops/internal/handler/handler_test.go
```diff
@@ -182,7 +182,6 @@ func TestActions(t *testing.T) {
 			Description: "Get a simple release",
 			ExistingReleases: []*release.Release{
 				createRelease("foo", "foobar", "default", 1, release.StatusDeployed),
-				createRelease("oof", "oofbar", "dev", 1, release.StatusDeployed),
```
andresmgot, line 185: do you need to delete these or it was just for debugging purposes?

absoludity: It was useless now that the memory driver is namespace aware, since it is never seen by the test (unless you explicitly changed the namespace to "dev").

PR #1678, opened 2020-04-20
Update postrenderer to support pod spec templates.

absoludity:
Just updates the post renderer to support getting the pod spec from all other resources which includes pod specs as a template.

### DIFF
--- a/pkg/agent/docker_secrets_postrenderer.go
+++ b/pkg/agent/docker_secrets_postrenderer.go
@@ -196,13 +196,30 @@ func getResourcePodSpec(resource map[interface{}]interface{}) map[interface{}]in
 		return nil
 	}
 
-	// TODO: Update to support other kinds with pod specs as part of pod templates:
-	// https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#podtemplatespec-v1-core
 	switch kind {
 	case "Pod":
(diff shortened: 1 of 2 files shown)


pkg/agent/docker_secrets_postrenderer.go
```diff
@@ -206,6 +204,25 @@ func getResourcePodSpec(resource map[interface{}]interface{}) map[interface{}]in
 			return nil
 		}
 		return podSpec
+	case "DaemonSet", "Deployment", "Job", "PodTemplate", "ReplicaSet", "ReplicationController", "StatefulSet":
+		// These resources all include a spec.template.spec PodSpec.
```
andresmgot, line 208: what about a CronJob? 

Also, we should also cover the case in which the resource is a List of any of those components.

absoludity: > what about a CronJob?

Yes, `JobTemplateSpec` is in a CronJob also, I need to include this as a separate case.

Similarly a `PodTemplate` is in a `PodTemplateList` too, but...

> 
> Also, we should also cover the case in which the resource is a List of any of those resources.

...yes, this needs to be handled not here, but in the parent case iterating the resources. I'll do this in a follow-up PR.

PR #1679, opened 2020-04-20
Allow to select docker registry credentials

### LINKED ISSUE #1522: Support docker registry secrets for AppRepositories

andresmgot:
### Description of the change

Allow to add docker registry credentials in the spec of an AppRepository. This is still not ready to be reviewed but I'm sending a draft to validate the UX first:

Peek 2020-04-20 18-33

### Applicable issues

  - fixes #1522

### Additional information

_still needs to be properly tested_

### DIFF
--- a/dashboard/src/actions/repos.test.tsx
+++ b/dashboard/src/actions/repos.test.tsx
@@ -313,6 +313,7 @@ describe("installRepo", () => {
     "",
     "",
     "",
+    [],
   );
 
   context("when authHeader provided", () => {
@@ -323,6 +324,7 @@ describe("installRepo", () => {
       "Bearer: abc",
       "",
       "",
+      [],
     );
 
     it("calls AppRepository create including a auth struct", async () => {
@@ -334,6 +336,7 @@ describe("installRepo", () => {
         "Bearer: abc",
         "",
         {},
+        [],
       );
     });
 
@@ -356,6 +359,7 @@ describe("installRepo", () => {
       "",
       "This is a cert!",
       "",
+      [],
     );
 
     it("calls AppRepository create including a auth struct", async () => {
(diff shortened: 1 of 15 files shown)


absoludity:
Thanks Andres! Yes, the UX looks exactly as I'd expect (ie. can select existing `kubernetes.io/dockerconfigjson` secrets as well as create new ones).

I'd move the "Server" field up to be the first piece of info after the secret name in the form.

Separate to this PR, we may want to use some validation on the server too (a few ways that could go).

absoludity:
> Separate to this PR, we may want to use some validation on the server too (a few ways that could go).

We will need to ensure that this is disabled (but probably good to be visible) when not using helm3 (already in the dashboard config). Also need to ensure it's disabled or explicitly not available for global app repos (in fact, if it's not visible for global repos, you won't need a feature flag, as it'll only display on per-app repos).

andresmgot:
:+1: thanks for the comments! I will try to split these changes so they are easier to review

PR #1680, opened 2020-04-21
Postrenderer support resourcelists

absoludity:
Based on [feedback from previous PR](https://github.com/kubeapps/kubeapps/pull/1678#discussion_r411754535), this updates to handle both CronJobs and all resource list types.

Ref: #1617

### DIFF
--- a/pkg/agent/docker_secrets_postrenderer.go
+++ b/pkg/agent/docker_secrets_postrenderer.go
@@ -198,32 +198,27 @@ func getResourcePodSpec(resource map[interface{}]interface{}) map[interface{}]in
 
 	switch kind {
 	case "Pod":
-		podSpec, ok := resource["spec"].(map[interface{}]interface{})
-		if !ok {
-			log.Errorf("invalid resource: non-map pod spec. %+v", resource)
-			return nil
-		}
-		return podSpec
+		return getMapForKeys([]string{"spec"}, resource)
 	case "DaemonSet", "Deployment", "Job", "PodTemplate", "ReplicaSet", "ReplicationController", "StatefulSet":
 		// These resources all include a spec.template.spec PodSpec.
 		// https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#podtemplatespec-v1-core
-		spec, ok := resource["spec"].(map[interface{}]interface{})
-		if !ok {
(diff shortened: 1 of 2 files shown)


pkg/agent/docker_secrets_postrenderer.go
```diff
@@ -52,6 +52,57 @@ func NewDockerSecretsPostRenderer(secrets map[string]string) (*DockerSecretsPost
...
+// search for pod templates.
+var resourceListTypes map[string]struct{} = map[string]struct{}{
+	"CronJobList":               struct{}{},
+	"DaemonSetList":             struct{}{},
+	"DeploymentList":            struct{}{},
+	"JobList":                   struct{}{},
+	"PodTemplateList":           struct{}{},
+	"ReplicaSetList":            struct{}{},
+	"ReplicationControllerList": struct{}{},
+	"StatefulSetList":           struct{}{},
```
andresmgot, line 65: AFAIK, just `"List"` is also a valid resource type

pkg/agent/docker_secrets_postrenderer.go
```diff
@@ -52,6 +52,57 @@ func NewDockerSecretsPostRenderer(secrets map[string]string) (*DockerSecretsPost
...
+			continue
+		}
+
+		kind, ok := kindValue.(string)
+		if !ok {
+			log.Errorf("invalid resource: non-string resource kind. %+v", resource)
+			continue
+		}
+		if _, ok := resourceListTypes[kind]; ok {
+			if items, ok := resource["items"]; ok {
```
andresmgot, line 86: can't we just check if the resource has the key `"items"` to know if it's a list? That may be a slightly slower (since we would be processing list for other than workloads) but safer and we would need to maintain less hard-coded types.

pkg/agent/docker_secrets_postrenderer.go
absoludity: As far as I know, this is only used for the top level which the renderer never sees, but I've added it anyway.

pkg/agent/docker_secrets_postrenderer.go
absoludity: We could, and it would work now. I thought it better to be explicit, since we don't control the API, but perhaps it would be simpler. Updating.

PR #1681, opened 2020-04-21
Generate and use secretsPerDomain.

absoludity:
Adds a function to create the map of docker registry servers to secret names and passes that in during Install and Upgrade functions.

I had to expose `kube.Handler`, but it was actually odd that it wasn't already exposed (it's an interface with public methods, and we use the public methods already). I haven't yet checked other uses of this, but will do. (I'll fix the errors).

With this, I hope to test deploying with a private repo tomorrow, which together with your UI work, might mean we can have a video demo or similar for the end of the week!

Ref: #1617

### DIFF
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -91,7 +91,7 @@ type Resolver interface {
 	ParseDetails(data []byte) (*Details, error)
 	GetChart(details *Details, netClient kube.HTTPClient, requireV1Support bool) (*ChartMultiVersion, error)
 	InitNetClient(details *Details, userAuthToken string) (kube.HTTPClient, error)
-	RegistrySecretsPerDomain() (map[string]string, error)
+	RegistrySecretsPerDomain() map[string]string
 }
 
 // ChartClient struct contains the clients required to retrieve charts info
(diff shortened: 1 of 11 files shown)


pkg/chart/chart.go
```diff
@@ -338,3 +352,42 @@ func (c *ChartClient) GetChart(details *Details, netClient kube.HTTPClient, requ
...
+// RegistrySecretsPerDomain checks the app repo and available secrets
+// to return the secret names per registry domain.
+//
+// These are actually calculated during InitNetClient when we already have a
+// k8s client with the user token.
+func (c *ChartClient) RegistrySecretsPerDomain() map[string]string {
+	return c.registrySecretsPerDomain
+}
+
+func getRegistrySecretsPerDomain(appRepoSecrets []string, namespace string, client kube.Handler) (map[string]string, error) {
```
andresmgot, line 365: unless I am missing something, you would need to return here `client kube.AuthHandler`. The `handler` is intentionally not exported so we don't mistakenly use it without setting it up with a token first (because if you don't do so, it would use the backend serviceaccount token if I recall correctly).

absoludity: `client kube.AuthHandler` is an interface defining the `AsUser(token)` method, so yes, I could pass that as well as the token to this method so we're explicit. It'd be nicer to explicitly have an interface `kube.UserHandler` to force the client here to be the result of `AsUser(token)`, but I'll leave that for another day, and just update to use `AuthHandler` and pass in the token. Thanks.

PR #1684, opened 2020-04-21
Select docker registry creds

andresmgot:
### Description of the change

Fork of #1679 to include just the registry credentials selection (without the form to add new ones).

Screenshot from 2020-04-21 16-31-32

### Applicable issues

  - ref #1522

### DIFF
--- a/dashboard/src/actions/repos.ts
+++ b/dashboard/src/actions/repos.ts
@@ -158,6 +158,9 @@ export const fetchRepos = (
       dispatch(requestRepos(namespace));
       const repos = await AppRepository.list(namespace);
       dispatch(receiveRepos(repos.items));
+      // TODO(andresmgot): Create an endpoint for returning credentials related to an AppRepository
+      // to avoid listing secrets
+      // https://github.com/kubeapps/kubeapps/issues/1686
       const secrets = await Secret.list(namespace);
       const repoSecrets = secrets.items?.filter(s =>
         s.metadata.ownerReferences?.some(ownerRef => ownerRef.kind === "AppRepository"),
@@ -312,6 +315,9 @@ export function fetchImagePullSecrets(
   return async dispatch => {
     try {
       dispatch(requestImagePullSecrets(namespace));
+      // TODO(andresmgot): Create an endpoint for returning just the list of secret names
+      // to avoid listing all the secrets with protected information
+      // https://github.com/kubeapps/kubeapps/issues/1686
       const secrets = await Secret.list(namespace);
       const imgPullSecrets = secrets.items?.filter(
         s => s.type === "kubernetes.io/dockerconfigjson",
--- a/dashboard/src/actions/repos.ts
+++ b/dashboard/src/actions/repos.ts
@@ -72,10 +72,6 @@ export const receiveImagePullSecrets = createAction("RECEIVE_IMAGE_PULL_SECRETS"
   return (secrets: ISecret[]) => resolve(secrets);
 });
 
-export const createImagePullSecret = createAction("CREATE_IMAGE_PULL_SECRET", resolve => {
-  return (secret: ISecret) => resolve(secret);
-});
-
 const allActions = [
   addRepo,
   addedRepo,
@@ -99,7 +95,6 @@ const allActions = [
   redirected,
   requestImagePullSecrets,
   receiveImagePullSecrets,
-  createImagePullSecret,
 ];
 export type AppReposAction = ActionType<typeof allActions[number]>;
 
@@ -319,26 +314,6 @@ export function fetchImagePullSecrets(
       dispatch(receiveImagePullSecrets(imgPullSecrets));
     } catch (e) {
       dispatch(errorRepos(e, "fetch"));
-    }
-  };
-}
-
-export function createDockerRegistrySecret(
-  name: string,
-  user: string,
-  password: string,
-  email: string,
-  server: string,
-  namespace: string,
-): ThunkAction<Promise<boolean>, IStoreState, null, AppReposAction> {
-  return async dispatch => {
-    try {
-      const secret = await Secret.createPullSecret(name, user, password, email, server, namespace);
-      dispatch(createImagePullSecret(secret));
-      return true;
-    } catch (e) {
-      dispatch(errorRepos(e, "fetch"));
-      return false;
     }
   };
 }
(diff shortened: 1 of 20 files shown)


dashboard/src/actions/repos.ts
```diff
@@ -287,3 +300,20 @@ export function checkChart(
...
   };
 }
+
+export function fetchImagePullSecrets(
+  namespace: string,
+): ThunkAction<Promise<void>, IStoreState, null, AppReposAction> {
+  return async dispatch => {
+    try {
+      dispatch(requestImagePullSecrets(namespace));
+      const secrets = await Secret.list(namespace);
```
absoludity, line 310: IMO, ideally we'd not be pulling the full secrets across the net to the client (where MITM is a threat), but rather just getting the list of secret names (from a backend API or similar). It's a separate topic, but I'm not sure if Kubeapps could generally never see actual secrets client side (other than when they are created)? If it would be possible (ie. we only ever request the actual secret resource when explicitly asked to by the user to display in the UI), then we should create an issue and reference it here, IMO.

dashboard/src/components/Config/AppRepoList/AppRepoForm.test.tsx
```diff
@@ -123,5 +178,30 @@ describe("when the repository info is already populated", () => {
...
+        metadata: {
+          name: "bar",
+        },
+      } as ISecret;
+      wrapper.setProps({ imagePullSecrets: [secret, secret2] });
+
+      expect(wrapper.state()).toMatchObject({
+        selectedImagePullSecrets: { foo: true, bar: false },
+      });
+    });
```
absoludity, line 205: We'll need to think about what to do if the existing secret referenced by the app repo has been deleted (I'm erroring during the deploy in this case).

dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
```diff
@@ -387,4 +442,35 @@ export class AppRepoForm extends React.Component<IAppRepoFormProps, IAppRepoForm
...
+    secrets.forEach(secret => {
+      let selected = false;
+      // If it has been already selected
+      if (selectedImagePullSecrets[secret.metadata.name]) {
+        selected = true;
+      }
+      // Or if it's already selected in the existing repo
+      if (repo?.spec?.dockerRegistrySecrets?.some(s => s === secret.metadata.name)) {
+        selected = true;
+      }
```
absoludity, line 471: Here I guess: if the secret on the app repo *isn't* available, we have an issue (or they should be explicitly removed by the user from the app repo).

dashboard/src/components/Config/AppRepoList/AppRepoForm.test.tsx
```diff
@@ -60,6 +67,54 @@ it("should not call the install method when the validation fails unless forced",
...
+it("should call the install method with the selected docker credentials", async () => {
+  const validate = jest.fn(() => true);
+  const install = jest.fn(() => true);
+  const wrapper = shallow(<AppRepoForm {...defaultProps} validate={validate} onSubmit={install} />);
+  wrapper.setState({ selectedImagePullSecrets: { "repo-1": true } });
+
+  const button = wrapper.find("form");
+  button.simulate("submit", { preventDefault: jest.fn() });
+  // wait for async functions
+  await wait(1);
```
absoludity, line 91: ugh - that's going to suck to maintain. Is this because it doesn't help to `await button.simulate` etc?

dashboard/src/actions/repos.ts
andresmgot: well, this is not the only place in which we are fetching secrets. To be able to edit an AppRepository and modify the credentials used we are also fetching those. In any case, it's true that we can limit the scope of what the application receives.

dashboard/src/components/Config/AppRepoList/AppRepoForm.test.tsx
andresmgot: We are only listing in the form existing secrets. If an AppRepository references a secret that no longer exists, when updating it with the dashboard, it will be automatically removed.

dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
andresmgot: yes, missing secrets won't be shown.

dashboard/src/components/Config/AppRepoList/AppRepoForm.test.tsx
andresmgot: well, it's not that difficult to maintain (the previous approach I had was worse, see for example https://github.com/kubeapps/kubeapps/pull/1684/files#diff-19c2034062c1971cbf14e6f364bcef47L184). It's just one line that allows to wait for mocked async functions to finish before evaluating tests results. Note that this is only needed when the Component code calls to something `async` and we wait for the result to update the state.

> Is this because it doesn't help to await button.simulate etc?

Yes, `button.simulate` doesn't return a promise so waiting that does nothing.

dashboard/src/components/Config/AppRepoList/AppRepoForm.test.tsx
absoludity: Right, that's what I think could be an issue: Scenario: User 1 adds an app repository with a specific secret. User two not using kubeapps at some point later kubectl delete's the secret . User 1 later edits the app repo and *silently* the secret is removed as a requirement of the app repo, without them noticing. People deploy charts and they fail.

I just wonder if it'd be better to make the user aware: "This app repository referred to the secret 'my-private-repo-creds' which is no longer present. Please ensure secrets are present for all private registries used" or something?

dashboard/src/components/Config/AppRepoList/AppRepoForm.test.tsx
absoludity: This seems to handle the issue via `setImmediate`, which apparently though supported in node (so should work fine here) isn't going to be a standard? https://github.com/enzymejs/enzyme/issues/823#issuecomment-492984956

andresmgot: yep, that doesn't seem that good either, at least now it's clear what it's doing.

dashboard/src/components/Config/AppRepoList/AppRepoForm.test.tsx
andresmgot: I agree but that would be a weird corner case, it's not common for people to edit an apprepository that's working for a while. I'm sure that people would notice before because the images referenced by the charts won't be pulled anymore.

PR #1688, opened 2020-04-22
Add scripts to deploy Kubeapps + Harbor (integrated) for testing purposes

juan131:

### Description of the change

This PR adds a couple of scripts that allow to create/delete a setup in your cluster where you have both Kubeapps and Harbor deployed on different namespaces, and integrated.

The scripts will also provide basic information about how to access both Kubeapps and Harbor to start playing with private repositories. 

### Benefits

This can be very useful for testing purposes.
 
### Possible drawbacks

None


script/setup-kubeapps-harbor.sh
```diff
@@ -0,0 +1,108 @@
...
+harbor_deployments=(
+  "harbor-chartmuseum"
+  "harbor-clair"
+  "harbor-core"
+  "harbor-jobservice"
+  "harbor-nginx"
+  "harbor-notary-server"
+  "harbor-notary-signer"
+  "harbor-portal"
+  "harbor-registry"
```
absoludity, line 46: When demoing kubeapps-harbor integration, I usually use:
```
--set persistence.enabled=false \
--set postgresql.persistence.enabled=false \
--set redis.master.persistence.enabled=false \
--set jobservice.replicas=0 \
--set registry.replicas=0 \
--set clair.replicas=0 \
--set notary.enabled=false
```
so that less resources are required locally (I often use a kind cluster). I'm not saying you should do the same here, but it might be worth considering, or at least making them optional?

juan131:
@absoludity I split the logic on several scripts and, I added some logic to add flags/helpers that make them easier to use

script/delete-harbor.sh
```diff
@@ -0,0 +1,36 @@
...
+        *)
+            echo "Invalid command line flag $1" >&2
+            return 1
+            ;;
+    esac
+    shift
+done
+
+# Uninstall Harbor
+silence kubectl delete ns "$namespace"
```
absoludity, line 32: I'm guessing you know this works, but wouldn't this be a race condition, or leave helm dangling, or make the next line redundant? That is, shouldn't we uninstall harbor first. Ah, it looks like you've repeated the namespace delete below afterwards, so I'm guessing you meant to remove this one?

script/setup-harbor.sh
```diff
@@ -0,0 +1,123 @@
...
+. "${ROOT_DIR}/script/liblog.sh"
+
+# Axiliar functions
+print_menu() {
+    local script
+    script=$(basename "${BASH_SOURCE[0]}")
+    log "${RED}NAME${RESET}"
+    log "    $(basename -s .sh "${BASH_SOURCE[0]}")"
+    log ""
+    log "${RED}SINOPSIS${RESET}"
```
absoludity, line 27: small typo - synopsis

script/setup-harbor.sh
```diff
@@ -0,0 +1,123 @@
...
+deployments=(
+    "harbor-chartmuseum"
+    "harbor-clair"
+    "harbor-core"
+    "harbor-jobservice"
+    "harbor-nginx"
+    "harbor-notary-server"
+    "harbor-notary-signer"
+    "harbor-portal"
+    "harbor-registry"
```
absoludity, line 108: Not necessary for this PR, but I'd love to be able to limit the services I want with harbor (for local demos I just run core, chartmuseam, registry and nginx (I think). We can build on that later.

script/setup-kubeapps-harbor.sh
```diff
@@ -0,0 +1,31 @@
...
+info "-- Harbor installation --"
+info "-------------------------"
+echo
+"$ROOT_DIR"/script/setup-harbor.sh --namespace "harbor"
+# Install Kubeapps
+info "---------------------------"
+info "-- Kubeapps installation --"
+info "---------------------------"
+echo
+"$ROOT_DIR"/script/setup-kubeapps.sh --namespace "kubeapps" --initial-repos "harbor-library" "http://harbor.harbor.svc.cluster.local/chartrepo/library"
```
absoludity, line 31: Excellent, this is exactly what I meant by splitting up the files. Great work.

script/delete-harbor.sh
juan131: I introduced this line twice by mistake. I'll fix it.

juan131:
Re-requesting review from you both @absoludity @andresmgot since I added the section in the developer docs as requested.

script/setup-kubeapps-harbor.sh
juan131: I enabled the possibility to disable Clair, Notary and Chartmuseum.
We can improve this in the future to disable other features (e.g. persistence).

docs/developer/testing-environment.md
```diff
@@ -0,0 +1,52 @@
...
+- [setup-harbor](../../script/setup-harbor.sh).
+
+These scripts will create the necessary namespaces, install the charts, wait for them to be available, and perform any extra action that might be needed. Find detailed information about how to use these scripts running the commands below:
+
+```bash
+./setup-kubeapps.sh --help
+./setup-harbor.sh --help
+```
+
+You can also use the [setup-kubeapps-harbor](../../script/setup-kubeapps-harbor.sh) script which is a wrapper that uses both the scripts mentioned above with some default values:
```
absoludity, line 42: I'd reverse the order here, and mention setup-kubeapps-harbor.sh first, as the wrapper, then the more detailed ones if you want them. But absolutely fine either way.

PR #1690, opened 2020-04-23
Fix YAML parsing for keys with dots and slashes

### LINKED ISSUE #1683: Don't need Yaml to automatically convert to json
image

image

kubeapps version: 1.8.2  and  1.9.2

I don't need Yaml to automatically convert to json.

andresmgot:
### Description of the change

So there are two issues when parsing a YAML that have `.` and `/`:
 - `.` were interpreted as objects so if the key was `a.b`, it was interpreted as `a: {b:}`
 - `/` is escaped as `~1` to avoid interpreting them as objects and we were not replacing that back.

I have modified our code to simply use `/` as the notification for objects (e.g. `a.b.c` -> `a/b/c`) which is the default for JSON patch. I have also added a step to unescape the character `~1`.

### Applicable issues

  - fixes #1683

### DIFF
--- a/dashboard/src/shared/schema.ts
+++ b/dashboard/src/shared/schema.ts
@@ -2,6 +2,7 @@
 // In particular, it doesn't contain definitions for `get` and `set`
 // that are used in this package
 import * as AJV from "ajv";
+import * as jsonpatch from "fast-json-patch";
 import * as jsonSchema from "json-schema";
 import { isEmpty, set } from "lodash";
 import * as YAML from "yaml";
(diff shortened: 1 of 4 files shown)


dashboard/src/shared/schema.ts
```diff
@@ -73,12 +74,31 @@ function getDefinedPath(allElementsButTheLast: string[], doc: YAML.ast.Document)
...
+    path
+      // ignore the first slash, if exists
+      .replace(/^\//, "")
+      // split by slashes
+      .split("/")
+  );
+}
+
+function unescapePath(path: string[]): string[] {
+  // jsonpath scapes slashes to not mistake then with objects so we need to revert that
```
absoludity, line 88: typo "scapes"

PR #1691, opened 2020-04-23
Minor fixes in the docker-credentials form 

andresmgot:
### Description of the change

I learned the hard way that `https://index.docker.io/v1` is not valid (it requires the trailing `/`), should we enforce that?

### DIFF
--- a/dashboard/src/components/Config/AppRepoList/AppRepoAddDockerCreds.tsx
+++ b/dashboard/src/components/Config/AppRepoList/AppRepoAddDockerCreds.tsx
@@ -89,7 +89,7 @@ export class AppRepoAddDockerCreds extends React.Component<IAppRepoFormProps, IA
                   id="kubeapps-docker-cred-server"
                   value={this.state.server}
                   onChange={this.handleServerChange}
-                  placeholder="https://index.docker.io/v1"
+                  placeholder="https://index.docker.io/v1/"
                   required={true}
                 />
               </div>
--- a/dashboard/src/components/Config/AppRepoList/AppRepoAddDockerCreds.tsx
+++ b/dashboard/src/components/Config/AppRepoList/AppRepoAddDockerCreds.tsx
@@ -62,7 +62,7 @@ export class AppRepoAddDockerCreds extends React.Component<IAppRepoFormProps, IA
             );
           })
         ) : (
-          <span className="margin-b-small">No existing credentials found.</span>
+          <div className="margin-b-small">No existing credentials found.</div>
         )}
         {this.state.showSecretSubForm && (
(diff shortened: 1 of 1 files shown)


absoludity:
> I learned the hard way that `https://index.docker.io/v1` is not valid (it requires the trailing `/`), should we enforce that?

As in k8s won't use the secret for dockerhub if the trailing slash isn't included? If that's the case, yes, we may even want a radio allowing selecting between dockerhub and a custom one. Or do you mean kubeapps treated it differently (it shouldn't have).

absoludity:
> > I learned the hard way that `https://index.docker.io/v1` is not valid (it requires the trailing `/`), should we enforce that?
> 
> As in k8s won't use the secret for dockerhub if the trailing slash isn't included? If that's the case, yes, we may even want a radio allowing selecting between dockerhub and a custom one. Or do you mean kubeapps treated it differently (it shouldn't have).

I'm not 100% but https://github.com/kubernetes/kubernetes/blob/master/pkg/credentialprovider/keyring.go#L106 woulds seem to strip off any `/v1/` when adding the credential to the basic provider (haven't checked if that's used). But the way I read that code, it should also be fine to just use the domain "index.docker.io" but the docs say otherwise, so unsure.

I also found while playing today that I could only use the domain of my private registry for the image pull secret, it wouldn't work with the full URL with path etc. Still need to dig some more.

andresmgot:
I have tested it a bit more but yes, as you point out, the paths `/v1/` and `/v2/` get specifically removed but it doesn't do that with `/v1` or `/v2`. Then the docker auth cli is just interested in the domain AFAIK. This is what I have tested:

```
domains ==> work

docker.io yes
index.docker.io yes
index.docker.io/v1 no
index.docker.io/v1/ yes
index.docker.io/v1/ yes
https://index.docker.io yes
https://index.docker.io/v1 no
https://index.docker.io/v1/ yes
https://index.docker.io/v2/ yes
```

The protocol is irrelevant both domains docker.io and index.docker.io works so for the sake of simplicify I will leave the placeholder as `docker.io`.

absoludity:
> The protocol is irrelevant and both domains docker.io and index.docker.io work. For the sake of simplicity I will leave the placeholder as `docker.io`.

Careful, the fact that it works regardless of the protocol doesn't mean it's irrelevant. There may be some security reason, such as old docker clients using http (and sending the credentials over http) if the protocol is not specified. Or it could depend on the credential store used.

Since the k8s docks specifically say to use "https://index.docker.io/v1/", including the protocol (see https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#create-a-secret-by-providing-credentials-on-the-command-line ) I think we should use it.

PR #1692, opened 2020-04-23
Do not logout when getting a 401

### LINKED ISSUE #1675: User gets logged out after failing to add app repository
### Description:

User gets logged out when trying to add an app repository but doesn't have permissions to do so.

1. Have an RBAC enabled cluster, and make sure you have the role that allows listing app repos but not to create them
2. Install Kubeapps
3. Log into Kubeapps
4. Go to app repos page
5. Click create app repo
6. Fill in some info and click "Install Repo"
A relevant error message is displayed since we're not allowed to create app repos, but it's only displayed for a split second before we are logged out.

A relevant error message is displayed since we're not allowed to create app repos, but it's only displayed for a split second before we are logged out.

I expect to get an error message telling me that I am not able to create an app repository and what rights I need to do so as well as staying logged in.

I don't believe this is exclusive to creating an app repository as I believe we also ran into this when trying to delete an app repository while not having the rights to do so.

@absoludity said on slack that they had also run into this problem while testing locally.

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.2.0-rc.1", GitCommit:"7bffac813db894e06d17bac91d14ea819b5c2310", GitTreeState:"clean", GoVersion:"go1.13.10"}
```

```
NAME        NAMESPACE    REVISION    UPDATED                                     STATUS      CHART             APP VERSION
kubeapps    kubeapps     1           2020-04-15 12:43:32.762055057 +0200 CEST    deployed    kubeapps-3.5.2    v1.9.2
```

```
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.11", GitCommit:"d94a81c724ea8e1ccc9002d89b7fe81d58f89ede", GitTreeState:"clean", BuildDate:"2020-03-16T20:10:55Z", GoVersion:"go1.12.17", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.7", GitCommit:"c83d931fb9bece427bc63a02349755e0f8696d3e", GitTreeState:"clean", BuildDate:"2020-01-31T20:09:49Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"linux/amd64"}
```

andresmgot:
### Description of the change

I believe I hit the same issue than #1675 but differently, I was testing a private repository and I introduced a typo in the password, instead of showing an error I got logged out.

AFAIK we should only logout when the error comes from the auth proxy (either if it's a 403 or a 401) so this PR fixes that. Let me know what you think @absoludity.

### Applicable issues

  - fixes #1675


absoludity:
> 
> I believe I hit the same issue than #1675 but differently, I was testing a private repository and I introduced a typo in the password, instead of showing an error I got logged out.

Do you mean that the repo validation check caused a logout? If so, the fix for that should be in the validation code. That is, doing a validation check on an external site should *never* cause a logout even if the response is crafted to mimic exactly what we'd expect from an auth proxy 401. We should not be using a code path which potentially logs out depending on the response of a third-party site?

We should be able to make the validation http request and handle (or ignore) all errors there in the validation code (perhaps whatever helper we're using should have an option to ensure it doesn't try to handle the errors, or we just use a separate Axios instance without the error handling).

> 
> AFAIK we should only logout when the error comes from the auth proxy

I haven't looked closely right now but afair this is the difficulty: detecting if a 401 or 403 originates from the auth proxy itself rather than the k8s API was not trivial. Checking for a 401 or 403 was too broad, but I need to run it locally and look again. Either way, that is very much unrelated to requests to external sites which should not even be evaluating those error handling conditions for our internal requests.


> (either if it's a 403 or a 401) so this PR fixes that.


> Let me know what you think @absoludity.
> ### Applicable issues
> 
>     * fixes #1675

andresmgot:
> Do you mean that the repo validation check caused a logout? If so, the fix for that should be in the validation code. That is, doing a validation check on an external site should _never_ cause a logout even if the response is crafted to mimic exactly what we'd expect from an auth proxy 401. We should not be using a code path which potentially logs out depending on the response of a third-party site?

I think you misunderstood me. That was the behavior I was seeing _before_ this PR and it's what I am fixing, now when I introduce wrong credentials I just get the error as it should be:

Screenshot from 2020-04-24 12-03-21

But it's true that if the response is crafted to look exactly like a response of the auth proxy (not this specific case) we will still be logging out the user. In any case, that issue is already present in master so this PR is not making that worse. 

The question would be, can we just ignore 401 errors and treat them as any other error? The response would be yes if the auth proxy only returns 403 for an expired token but we don't know that for sure.

> We should be able to make the validation http request and handle (or ignore) all errors there in the validation code (perhaps whatever helper we're using should have an option to ensure it doesn't try to handle the errors, or we just use a separate Axios instance without the error handling).

The validation code is quite simple, we are just proxying whatever the repository returns, I don't think we need to change that.

absoludity:
> I think you misunderstood me. That was the behavior I was seeing _before_ this PR 

That's what I understood :)

> and it's what I am fixing, now when I introduce wrong credentials I just get the error as it should be:
> 
> Screenshot from 2020-04-24 12-03-21
> 
> But it's true that if the response is crafted to look exactly like a response of the auth proxy (not this specific case) we will still be logging out the user.

Well, I think this issue here with the validation is a direct implication of the point I was trying to make at https://github.com/kubeapps/kubeapps/pull/1537#discussion_r385475425 . Our validation API (in the backend) should not (IMO) be returning the 4xx error from the external site. The request to our backend API was not a 4xx, it was a successfully authenticated request to the backend and it verified that the user app repo cannot be validate with the user input. But because we currently return the same status code as the external request, the dashboard cannot tell the difference between a 4xx with the backend and a 4xx with the external site.

Our backend API should be returning a 200 response (as the request to the backend was fine) and the data payload should indicate that the users info for the app repo did not validate. The dashboard gets the data from the successful backend call to determine the outcome, rather than assuming an error means it did not validate.

> In any case, that issue is already present in master so this PR is not making that worse.
> 
> The question would be, can we just ignore 401 errors and treat them as any other error? The response would be yes if the auth proxy only returns 403 for an expired token but we don't know that for sure.

I'm keen to look at this separately (and plan to do so next). Adding code in our generic error handling for requests which caters for responses from external sites is not, I don't think, what we should be doing here.

> 
> > We should be able to make the validation http request and handle (or ignore) all errors there in the validation code (perhaps whatever helper we're using should have an option to ensure it doesn't try to handle the errors, or we just use a separate Axios instance without the error handling).
> 
> The validation code is quite simple, we are just proxying whatever the repository returns,

Right, it's *too* simple IMO, we can no longer differentiate between a 4xx from the backend and a 4xx from the external site.

andresmgot:
> Our backend API should be returning a 200 response (as the request to the backend was fine) and the data payload should indicate that the users info for the app repo did not validate. The dashboard gets the data from the successful backend call to determine the outcome, rather than assuming an error means it did not validate.

Okay, I buy you that. In any case, I don't know why we should be automatically logging out users because of a 401. This may happen in other places, for example, if a chart points to an icon with an URL that is protected with basic auth, we would may a 401 and log out the user.

absoludity:
> Okay, I buy you that. In any case, I don't know why we should be automatically logging out users because of a 401. This may happen in other places, for example, if a chart points to an icon with an URL that is protected with basic auth, we would may a 401 and log out the user.

The only way we would logout the user is if we requested the icon via axios using the axios instance that we have setup for talking to the k8s api, which yes, we should not be doing?  That's what I meant above (when I'd incorrectly assumed the validation was sending the request from the dashboard) with: "(perhaps whatever helper we're using should have an option to ensure it doesn't try to handle the errors, or we just use a separate Axios instance without the error handling)". If we are fetching an icon (or anything) from an external site, we should just be creating a new axios instance (which won't have our error handling which is specific to kubeapps authn).

andresmgot:
> The only way we would logout the user is if we requested the icon via axios using the axios instance that we have setup for talking to the k8s api, which yes, we should not be doing? That's what I meant above (when I'd incorrectly assumed the validation was sending the request from the dashboard) with: "(perhaps whatever helper we're using should have an option to ensure it doesn't try to handle the errors, or we just use a separate Axios instance without the error handling)". If we are fetching an icon (or anything) from an external site, we should just be creating a new axios instance (which won't have our error handling which is specific to kubeapps authn).

Right, we are not using the axios instance with auth for that. Anyway, I will open a different PR for the validation stuff.

PR #1694, opened 2020-04-24
Make the nodePort customizable

andresmgot:
### Description of the change

Being able to customize the nodePort chosen with the chart. It can be used like this:

```
--set frontend.service.type=NodePort --set frontend.service.nodePort=31080
```


chart/kubeapps/values.yaml
```diff
@@ -119,6 +119,9 @@ frontend:
     ## ref: http://kubernetes.io/docs/user-guide/services/#type-loadbalancer
     ##
     # loadBalancerIP:
+    ## Set a specific NodePort (only when frontend.service.type="LoadBalancer")
```
absoludity, line 122: ^^ copy-pasta? (LoadBalancer)

andresmgot:
(I removed the version increase to avoid releasing #1693 already)

PR #1700, opened 2020-04-28
Update docs for private app repositories.

absoludity:
Update private app repo docs to include info about associating image pull secrets with an app repo.

Rendered preview in the branch at:
https://github.com/absoludity/kubeapps/blob/app-repo-docs/docs/user/private-app-repository.md

cc @tiffanyfay , given your improvements to the docs, in case you have suggestions before we release with this (though we can improve the docs after release obviously, so no pressure :) )


tiffanyfay:
Thanks! I can look at this in the morning, if that's not too late. I am going to sleep now-- I am on US Pacific Time.

tiffanyfay:
And like you said, I can make comments after so I don't become the blocker for this.

PR #1704, opened 2020-04-29
Default to postgres rather than mongodb.

absoludity:
While thinking about roadmap goals, I realised this is IMO long overdue. We need to be able to start improving the asset svc as we add multi-cluster support and/or improve operator integration etc., and it's going to be painful if we have to continue doing so for both mongodb and postgres.

### DIFF
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -513,7 +513,7 @@ dashboard:
 ##
 mongodb:
   ## Whether to deploy a mongodb server to satisfy the applications database requirements.
-  enabled: true
+  enabled: false
   ## Kubeapps uses MongoDB as a cache and persistence is not required
   ##
   persistence:
@@ -546,7 +546,7 @@ mongodb:
 ##
 postgresql:
   ## Whether to deploy a postgresql server to satisfy the applications database requirements.
-  enabled: false
+  enabled: true
   ## Enable replication for high availability
   replication:
     enabled: true


absoludity:
> I believe that there is no problem with this but we need to double check that there are no issues upgrading from one db to the other

Yep, will do before landing.

absoludity:
So I just tested upgrading from a mongodb to postgres and it fails because the pre-upgrade pod can't run as there is no db secret:
```
  Warning  Failed     2s (x4 over 27s)  kubelet, kubeapps-control-plane  Error: secret "kubeapps-db" not found
```
This is because it assumes that the postgres secret already exists, but our `db-secret-bootstrap.yaml` is run `pre-install`. I don't see a way we can run it `pre-upgrade-only-if-not-exists`.

On pre-upgrade we need to:
* ensure a db secret exists if it doesn't already
* ensure postgres exists if it doesn't already
* run the pre-upgrade job

though I'm not sure this is possible. Thoughts? Otherwise it's going to be a breaking change for a major version :/

absoludity:
> I don't see a way we can run it `pre-upgrade-only-if-not-exists`.

Actually, this is now possible with helm 3.1, via the lookup function, mentioned here: https://github.com/helm/charts/issues/5167#issuecomment-619137759

With that, we could always ensure the postgres secret is available before the upgrade. But to use that we also need to be assuming helm 3.1, so I don't see how we can use this before then.

So possibly:
 * Move to helm3 by default (make 3.1)
 * Remove helm 2 support
 * Then support upgrading to postgres

Or: release a new major version which defaults to helm 3 and postgres by default, and requires re-install of Kubeapps. I'll convert this into an issue and close this PR. https://github.com/kubeapps/kubeapps/issues/1713

andresmgot:
_replied in the issue linked_

PR #1706, opened 2020-04-29
Enable modify button for app repositories

### LINKED ISSUE #1666: Allow to edit AppRepositories

andresmgot:
### Description of the change

Requires #1705 

I have enabled the modify button which was mostly ready unless two minor things:

 - I needed to re-fetch secrets in case the modification modify any of the related app repo secrets
 - I have added the updatedRepo event in the reducer to update the updated app repo in the state.

Since the reducer for app repos didn't exist, I have covered part of that file with unit tests.

### Applicable issues

  - fixes #1666

### DIFF
--- a/dashboard/src/actions/repos.ts
+++ b/dashboard/src/actions/repos.ts
@@ -35,6 +35,10 @@ export const receiveReposSecrets = createAction("RECEIVE_REPOS_SECRETS", resolve
   return (secrets: ISecret[]) => resolve(secrets);
 });
 
+export const receiveReposSecret = createAction("RECEIVE_REPOS_SECRET", resolve => {
+  return (secret: ISecret) => resolve(secret);
+});
+
 export const requestRepo = createAction("REQUEST_REPO");
 export const receiveRepo = createAction("RECEIVE_REPO", resolve => {
   return (repo: IAppRepository) => resolve(repo);
@@ -89,6 +93,7 @@ const allActions = [
   receiveRepo,
   receiveRepos,
   receiveReposSecrets,
+  receiveReposSecret,
   resetForm,
   errorChart,
   requestRepo,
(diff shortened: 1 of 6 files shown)


dashboard/src/actions/repos.ts
```diff
@@ -248,6 +256,8 @@ export const updateRepo = (
         registrySecrets,
       );
       dispatch(repoUpdated(data.appRepository));
+      // Re-fetch secrets in case there are updates
+      dispatch(fetchRepoSecrets(namespace));
```
absoludity, line 260: So given the different secret associated with an AppRepository - helm repository secrets and docker registry image pull secrets, this could be confusing? Perhaps a more explicit comment:

```
// Re-fetch the helm repo secret that could have been modified with the updated headers
// so that if the user chooses to edit the app repo again, they will see the current value.
```
if I have understood correctly? Also, don't we expect only ever *one* secret for the helm repository access? If so, why aren't we encoding that anywhere? Finally, would the code be simpler if the AppRepository actually stored in a field the name of the helm repository secret?

andresmgot: we can request just one secret, yes

dashboard/src/actions/repos.ts
```diff
@@ -248,6 +271,12 @@ export const updateRepo = (
         registrySecrets,
       );
       dispatch(repoUpdated(data.appRepository));
+      // Re-fetch the helm repo secret that could have been modified with the updated headers
+      // so that if the user chooses to edit the app repo again, they will see the current value.
+      if (data.appRepository.spec?.auth) {
+        const secretName = data.appRepository.spec.auth.header.secretKeyRef.name;
```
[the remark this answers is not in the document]
andresmgot: Good catch, yes, that's correct.

PR #1707, opened 2020-04-30
Update docs to clearly state that the imagePullSecret feature is Helm3 only

absoludity:
Wondering if the UI should clearly state as much (and not allow imagePullSecrets to be configured)?


andresmgot:
> Wondering if the UI should clearly state as much (and not allow imagePullSecrets to be configured)?

We could add a note. AFAIK we don't know which version of helm we are running in the Dashboard but we can say that the feature is only supported for Helm 2. You can add that in this PR

absoludity:
> > Wondering if the UI should clearly state as much (and not allow imagePullSecrets to be configured)?
> 
> We could add a note. AFAIK we don't know which version of helm we are running in the Dashboard but we can say that the feature is only supported for Helm 2. You can add that in this PR

absoludity:
> > Wondering if the UI should clearly state as much (and not allow imagePullSecrets to be configured)?
> 
> We could add a note. AFAIK we don't know which version of helm we are running in the Dashboard

We could easily add it to the config.json feature flags.

> but we can say that the feature is only supported for Helm 2. You can add that in this PR

Good enough for now :) Done.

PR #1714, opened 2020-05-04
Elaborate message for global app repositories

andresmgot:
### Description of the change

I believe that the message that explains global namespaces is not clear. People may think that selecting "All Namespaces" in the namespace selector would create a cluster resource rather than storing it in the Kubeapps namespace:

https://kubernetes.slack.com/archives/C9D3TSUG4/p1588170819044300

Suggestions?

### DIFF
--- a/dashboard/src/components/Config/AppRepoList/AppRepoList.tsx
+++ b/dashboard/src/components/Config/AppRepoList/AppRepoList.tsx
@@ -186,9 +186,8 @@ class AppRepoList extends React.Component<IAppRepoListProps> {
           <MessageAlert header="Looking for other app repositories?">
             <div>
               <p className="margin-v-normal">
-                You can view App Repositories across all namespaces by selecting "All Namespaces" or
-                "{kubeappsNamespace}" above, if you have permission to view App Repositories
-                cluster-wide. These will be stored in the {kubeappsNamespace} namespace.
+                You can view App Repositories across all namespaces by selecting "All Namespaces"
+                above, if you have permission to view App Repositories cluster-wide.
               </p>
               <p className="margin-v-normal">
                 Kubeapps now enables you to create App Repositories in your own namespace that will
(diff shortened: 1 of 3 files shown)


dashboard/src/components/Config/AppRepoList/AppRepoList.tsx
```diff
@@ -186,8 +186,9 @@ class AppRepoList extends React.Component<IAppRepoListProps> {
           <MessageAlert header="Looking for other app repositories?">
             <div>
               <p className="margin-v-normal">
-                You can view App Repositories across all namespaces by selecting "All Namespaces"
-                above, if you have permission to view App Repositories cluster-wide.
+                You can view App Repositories across all namespaces by selecting "All Namespaces" or
+                "{kubeappsNamespace}" above, if you have permission to view App Repositories
+                cluster-wide. These will be stored in the {kubeappsNamespace} namespace.
```
absoludity, line 191: I worry this change might create more confusion, see if you think my suggestion any better.

Note that, first, when *viewing* app repositories, selecting "All Namespaces" does request the app repositories cluster-wide (showing both those in kubeapps namespace as well as other namespaces). Selecting the `kubeapps` namespace will show only the app repositories present in the kubeapps namespace. They are in fact two different things when viewing, the confusion is that *creating* an app repo when either is selected is the same as we don't create cluster-wide app repositories resources. So I think it's clearer to limit the text "You can view App Repositories across all namespaces" to "All namespaces".

Next, the message you are adjusting is displayed to users viewing app repos when the selector is *not* "All namespaces" (ie. I'm not sure Andreas would have seen it here, it is not visible when All namespaces is selected).

So, it would be clearer, imo, to leave this message (specific to viewing app repos) and instead show a separate message in the form when *creating* or editing an app repository for "All Namespaces" or `kubeapps`: "This App Repository will be created in the {kubeappsNamespace} namespace and charts will be available in all namespaces for installation."

Make sense?

andresmgot: yep, makes sense, that also works for me

andresmgot: Screenshot from 2020-05-05 13-01-22

Issue #1708, opened 2020-04-30
Cover Operators and Private Registries with end-to-end tests


andresmgot:

absoludity:
Still seeing 
  ● Deploys an Operator

    TimeoutError: Navigation Timeout Exceeded: 20000ms exceeded

       8 |       } catch (e) {
       9 |         // Refresh since the chart will get a bit of time to populate
    > 10 |         await page.reload({
         |                    ^

I'd bee keen to close this ticket only when it's not requiring more work? (Sorry, I perhaps shouldn't have moved it to Done yesterday prematurely).

andresmgot:
No problem, that issue you point is fixed here: https://github.com/kubeapps/kubeapps/pull/1729

PR #1716, opened 2020-05-04
Browser test for adding a private Helm and Docker registry

andresmgot:
### Description of the change

For this test we install chartmuseum and we upload a chart to it (just a mirror from the bitnami/apache). We setup basic authentication.

Then, in the browser test, we create a dummy docker-registry secret and we use it. To check if the docker-registry is attached, we check the deployment manifest, looking for the imagePullSecret.

### Applicable issues

  - ref #1708


integration/use-cases/create-private-registry.js
```diff
@@ -0,0 +1,96 @@
...
+    "#kubeapps-docker-cred-server",
+    "https://index.docker.io/v1/"
+  );
+  await page.type("#kubeapps-docker-cred-username", "user");
+  await page.type("#kubeapps-docker-cred-password", "password");
+  await page.type("#kubeapps-docker-cred-email", "user@example.com");
+  await expect(page).toClick("button", { text: "Submit" });
+
+  // Select the new secret
+  await expect(page).toClick("label", { text: secret });
```
absoludity, line 51: I meant to ask about this: in the UX we should automatically select the secret which was just created, don't you think? (It'd be easy to create it and not select it).

script/e2e-test.sh
```diff
@@ -271,12 +330,15 @@ pod=$(kubectl get po -l run=integration -o jsonpath="{.items[0].metadata.name}")
 for f in *.js; do
   kubectl cp "./${f}" "${pod}:/app/"
 done
+## Support for Docker registry secrets are not supported for Helm2, skipping that test
+if [[ "${HELM_VERSION:-}" =~ "v2" ]]; then
+  rm use-cases/create-private-registry.js
```
absoludity, line 335: We can't use https://jestjs.io/docs/en/configuration#testpathignorepatterns-arraystring ? No problem if it's a hassle to change.

integration/use-cases/create-private-registry.js
andresmgot: it's not exactly trivial to do this since the component just re-renders with a new set of secrets in the props but we can take that separately.

script/e2e-test.sh
andresmgot: that will be a bit more elegant, yes

PR #1717, opened 2020-05-05
Fix operators test for GKE

andresmgot:
### Description of the change

Fix the GKE tests in master. There are two problems for the Operator test:

 - The OLM cannot be installed in GKE 1.14, the manifest fails to be installed so we need to skip the test in this build.
 - The newer image that I am using to avoid the issue for limits too low doesn't work for GKE 1.15 so I need to use the default manifest and let the pod to be killed several times because of OOM until it populates the operator catalog (which is slow).

[EDIT]

At the end I had to disable the test also for GKE 1.15, it was too flaky because the operatorhub populator pod gets continuously restarted.

### Applicable issues

  - ref #1708

### DIFF
--- a/script/e2e-test.sh
+++ b/script/e2e-test.sh
@@ -92,7 +92,7 @@ installOLM() {
 
     kubectl apply -f ${url}/crds.yaml
 
-    if [[ -z "$GKE_ADMIN" ]]; then
+    if [[ -z "$GKE_BRANCH" ]]; then
       # The Pod that populates the catalog gets OOM Killed due to very low limits
       # This has been fixed here: https://github.com/operator-framework/operator-lifecycle-manager/pull/1389
       # But the fix has not been published yet. To workaround the issue we are using a newer image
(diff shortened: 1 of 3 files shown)


script/e2e-test.sh
```diff
@@ -91,12 +103,17 @@ installOLM() {
 
     kubectl apply -f ${url}/crds.yaml
 
-    # The Pod that populates the catalog gets OOM Killed due to very low limits
-    # This has been fixed here: https://github.com/operator-framework/operator-lifecycle-manager/pull/1389
-    # But the fix has not been published yet. To workaround the issue we are using a newer image
-    # This will be fixed in a version > 0.14.2
-    kubectl apply -f "${ROOT_DIR}/script/manifests/olm.yaml"
-
+    if [[ "${GKE_BRANCH-}" != "1.15" ]]; then
```
absoludity, line 106: Should this code still be here if we're not installing OLM for 1.15 either?

andresmgot: right, I forgot to delete it

Issue #1713, opened 2020-05-04
Switch to use postgresql by default


absoludity:
Currently, while we support Helm < 3.1, we cannot (AFAICT) support upgrading to a chart where Postgresql is used by default for the backend, without requiring manual intervention by the user to ensure (at least) that the database secret exists (normally done pre-install).

With Helm 3.1, we can use the lookup template tag to create the secret if it does not already exist during an upgrade.

So while we could gradually:
 * Move to helm3 by default (make 3.1)
 * Remove helm 2 support
 * Then support upgrading to postgres via the above lookout,
it is going to drawn out over quite a long time.

We need to be able to drop mongodb support so we can begin other larger tasks without being hindered by supporting two backends (with minimal functional testing for mongodb in the go code base).

Another option would be for us to decide when we can drop helm 2 support. According to the [helm2 support contract](https://github.com/helm/helm/blob/master/CONTRIBUTING.md#support-contract-for-helm-2), helm 2 will stop accepting bug fixes this month (May 2020), while November 2020 is the end (no more security patches accepted etc.). Depending on our strategy for that, we may be able to switch to postgres-only at the same time (ie. remove support for mongodb).

andresmgot:
something that would work today is adding a second container to the `apprepository-jobs-preupgrade` job that creates the secret (using `kubectl`) if it doesn't exist. WDYT? Any downside of doing that?

absoludity:
> something that would work today is adding a second container to the `apprepository-jobs-preupgrade` job that creates the secret (using `kubectl`) if it doesn't exist. WDYT? Any downside of doing that?

That's a neat idea - I'd thought of updating the pre-upgrade job (which is the asset-sync binary) to do that, but decided it'd be pretty ugly, but using an init-container or similar would be much neater, using kubectl from a bash script. We just need to pass in the `postgresql.exstingSecret` value for the secret name. Thanks!

PR #1718, opened 2020-05-07
Create postgres secret if it doesn't exist during upgrade.

absoludity:
Based on the suggestion at https://github.com/kubeapps/kubeapps/issues/1713#issuecomment-623341571 , this adds an init container to the database post-upgrade job so that, if upgrading from mongo to postgres, we automatically create a secret.

Note that I had to use a more privileged service account - I've just used the apprepository-controller's sa (and added get/write secrets), but will create a separate service account for the post-upgrade job, with RBAC (or do you think we should just use the default service account, with the extra RBAC, wouldn't think so).

Note also that I've switched from pre to post upgrade so that postgresql itself is always available. I want to test more to ensure there isn't a race between the sync jobs and the post upgrade job.


absoludity:
Hmm, looks like the e2e tests are all failing with:

```Waiting for deployment "olm-operator" rollout to finish: 0 of 1 updated replicas are available...
deployment "olm-operator" successfully rolled out
Waiting for deployment "catalog-operator" rollout to finish: 0 of 1 updated replicas are available...
error: deployment "catalog-operator" exceeded its progress deadline```

We may need to disable olm installation on the e2e tests too? :/

**Comment by andresmgot** [comment]
> but will create a separate service account for the post-upgrade job, with RBAC (or do you think we should just use the default service account, with the extra RBAC, wouldn't think so).

Yes please, create its own serviceaccount, it's more clear that way.

> Hmm, looks like the e2e tests are all failing with:

That's weird, maybe it's a transient error, let me check.

**Comment by andresmgot** [comment]
> > Hmm, looks like the e2e tests are all failing with:
> 
> That's weird, maybe it's a transient error, let me check.

So the problem is that the image that I was using to avoid the OOM issue is no longer available (`quay.io/operator-framework/olm@sha256:fa0163f45f762be20a943ab866b291acf71b8407233f018c0a0cb154feca38ef`) so I will open a separate issue to use latest while they release a new version (which apparently will happen soon).

**Comment by absoludity** [comment]
Looks like operator tests are still failing intermittently (at least, this only failed on one of the 4 e2e tests at https://circleci.com/gh/kubeapps/kubeapps/22526?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link ).
```
Deploys an Operator

    TimeoutError: Navigation Timeout Exceeded: 20000ms exceeded

       8 |       } catch (e) {
       9 |         // Refresh since the chart will get a bit of time to populate
    > 10 |         await page.reload({
         |                    ^
      11 |           waitUntil: ["networkidle0", "domcontentloaded"],
      12 |           timeout: 20000,
      13 |         });
```

andresmgot:
> Looks like operator tests are still failing intermittently (at least, this only failed on one of the 4 e2e tests at https://circleci.com/gh/kubeapps/kubeapps/22526?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link ).
> 
> ```
> Deploys an Operator
> 
>     TimeoutError: Navigation Timeout Exceeded: 20000ms exceeded
> 
>        8 |       } catch (e) {
>        9 |         // Refresh since the chart will get a bit of time to populate
>     > 10 |         await page.reload({
>          |                    ^
>       11 |           waitUntil: ["networkidle0", "domcontentloaded"],
>       12 |           timeout: 20000,
>       13 |         });
> ```

mmh, that's the code that reload a page, we have been used that since the beginning but the operators require more refreshes, I suspect that's the issue. I can try to increase the delay between refreshes to avoid that issue.

PR #1719, opened 2020-05-07
Use latest image for OLM

andresmgot:
### Description of the change

The previous image is no longer available. Use the `latest` image until they release a new version.

https://github.com/kubeapps/kubeapps/pull/1718#issuecomment-625123423


andresmgot:
_merging to fix master_

Issue #1336, opened 2019-12-02
Add support to "complex conditions" for hidden items on basic forms


juan131:
**Description**

According to [the documentation](https://github.com/kubeapps/kubeapps/blob/master/docs/developer/basic-form-support.md#subsections):

> The tag hidden can be a string pointing to the parameter that needs to be true to hide the element or an object to also set the value that the pointed value needs to match.

This doesn't give us much space to create "hidden rules" when two or more parameters need to be true, or to use other conditions such as "matchString".

E.g. check: https://github.com/helm/charts/pull/19300

**Version of Helm, Kubeapps and Kubernetes**:

- Output of `helm list <kubeapps-release-name>`:

```console
$ helm list kubeapps
NAME    	REVISION	UPDATED                 	STATUS  	CHART         	APP VERSION	NAMESPACE
kubeapps	1       	Thu Nov 28 12:23:42 2019	DEPLOYED	kubeapps-3.1.0	v1.7.1     	kubeapps
```

andresmgot:
Thanks @juan131 for filling the issue. From the description I see two different requests:

 1. Allow to set an array of conditions instead of just one. Then the parameter can be hidden if all or any of the conditions match.
 2. Allow other type of comparisons. Right now it's only possible to match exact values.

I believe that for adding the schema for redmine you only need the first, right?

juan131:
Yes @andresmgot 
For the Remine case, I only need the 1st one.

PR #1720, opened 2020-05-07
Add support to multiple conditions to hide an element on basic forms

juan131:

### Description of the change

This PR adds support for multiple conditions to hide an element on basic forms.

As it's explained in the developer docs, developers will be able to use an array of conditions, and the conditional operator to consider an element to be hidden.

### Benefits

More possibilities for chart developers to create their values.schemas.json

### Possible drawbacks

None that I'm aware of

### Applicable issues

  - fixes https://github.com/kubeapps/kubeapps/issues/1336

### DIFF
--- a/dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/BasicDeploymentForm.test.tsx
+++ b/dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/BasicDeploymentForm.test.tsx
@@ -131,8 +131,17 @@ it("should hide an element if it depends on a param (object)", () => {
       path: "foo",
       type: "string",
       hidden: {
-        value: "bar",
-        condition: "enabled",
+        conditions: [
+          {
+            value: "bar",
+            reference: "enabled",
+          },
+          {
+            value: "baz",
+            reference: "disabled",
+          },
+        ],
+        operator: "and",
       },
     },
     {
@@ -140,7 +149,7 @@ it("should hide an element if it depends on a param (object)", () => {
       type: "string",
     },
   ] as IBasicFormParam[];
-  const appValues = "foo: 1\nbar: enabled";
+  const appValues = "foo: 1\nbar: enabled\nbaz: disabled";
   const wrapper = shallow(
     <BasicDeploymentForm {...defaultProps} params={params} appValues={appValues} />,
   );
--- a/dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/BasicDeploymentForm.test.tsx
+++ b/dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/BasicDeploymentForm.test.tsx
@@ -125,7 +125,31 @@ it("should hide an element if it depends on a param (string)", () => {
   expect(hiddenParam).toExist();
 });
 
-it("should hide an element if it depends on a param (object)", () => {
+it("should hide an element if it depends on a single param (object)", () => {
+  const params = [
+    {
+      path: "foo",
+      type: "string",
+      hidden: {
+        value: "bar",
+        path: "enabled",
+      },
+    },
+    {
+      path: "bar",
+      type: "string",
+    },
+  ] as IBasicFormParam[];
+  const appValues = "foo: 1\nbar: enabled";
+  const wrapper = shallow(
+    <BasicDeploymentForm {...defaultProps} params={params} appValues={appValues} />,
+  );
+
+  const hiddenParam = wrapper.find("div").filterWhere(p => p.prop("hidden") === true);
+  expect(hiddenParam).toExist();
+});
+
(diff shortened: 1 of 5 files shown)


dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/BasicDeploymentForm.test.tsx
```diff
@@ -131,16 +131,25 @@ it("should hide an element if it depends on a param (object)", () => {
       path: "foo",
       type: "string",
       hidden: {
-        value: "bar",
-        condition: "enabled",
+        conditions: [
+          {
+            value: "bar",
+            reference: "enabled",
```
andresmgot, line 137: nit: I would rename `reference` with `path` to make it more clear that is a jsonpath

dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/BasicDeploymentForm.tsx
```diff
@@ -42,8 +42,33 @@ class BasicDeploymentForm extends React.Component<IBasicDeploymentFormProps> {
...
+          case "and":
+            // Every value matches its reference in
+            // all the conditions
+            isHiddenParam = true;
+            hidden.conditions.forEach(c => {
+              if (getValue(this.props.appValues, c.value) !== c.reference) {
+                isHiddenParam = false;
+              }
+            });
+            return isHiddenParam;
```
andresmgot, line 58: a neater solution:

```
return hidden.conditions.every(condition => getValue(this.props.appValues, condition.value) === condition.reference)
```

dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/BasicDeploymentForm.tsx
```diff
@@ -42,8 +42,33 @@ class BasicDeploymentForm extends React.Component<IBasicDeploymentFormProps> {
...
+          case "or":
+            // It is enough if the value matches its reference in
+            // any of the conditions
+            isHiddenParam = false;
+            hidden.conditions.forEach(c => {
+              if (getValue(this.props.appValues, c.value) === c.reference) {
+                isHiddenParam = true;
+              }
+            });
+            return isHiddenParam;
```
andresmgot, line 68: same than above but with `hidden.conditions.some`

dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/BasicDeploymentForm.test.tsx
```diff
@@ -148,3 +148,36 @@ it("should hide an element if it depends on a param (object)", () => {
...
+it("should hide an element if it depends on multiple params (object)", () => {
+  const params = [
+    {
+      path: "foo",
+      type: "string",
+      hidden: {
+        conditions: [
+          {
+            value: "bar",
+            path: "enabled",
```
andresmgot, line 161: there has been a change in the meaning. Before `value` meant the path to the element and `condition` the value it should have (yes, a bit weird). So for the object `value: mariadb.enabled, condition: true` it meant that `mariadb.enabled === true`.

You have added `path` but this `path` is actually the value, and the `value` is the path :). In your example, `path` should be `bar` (or if it were an inner object, it would be `bar/foo`) and the `value` should be `enabled`.

Hope I explained myself :)

dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/BasicDeploymentForm.tsx
```diff
@@ -42,8 +42,31 @@ class BasicDeploymentForm extends React.Component<IBasicDeploymentFormProps> {
...
+                c => getValue(this.props.appValues, c.value) === c.path,
+              );
+            case "or":
+              // It is enough if the value matches the referenced
+              // value (via jsonpath) in any of the conditions
+              return hidden.conditions.some(
+                c => getValue(this.props.appValues, c.value) === c.path,
+              );
+            default:
+              return false;
```
andresmgot, line 65: I would choose a default (probably `and` is better as a default) instead of returning always false if the operator is not defined.

dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/BasicDeploymentForm.test.tsx
juan131: Yes, fixing it

dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/BasicDeploymentForm.tsx
juan131: okey!

Issue #1703, opened 2020-04-28
Allow to customize Operators installation


andresmgot:
### Description:

Currently, we are pointing to the public manifest available in the OperatorHub to install Operators. This has two inconvenients:

 - Access to the OperatorHub is required.
 - The installation cannot be customized (e.g. the namespace is hardcoded).

We should offer a form to be able to customize the installation parameters.

PR #1723, opened 2020-05-07
Add form to deploy an Operator

andresmgot:
### Description of the change

Adds a form to deploy operators (rather than just showing instructions):

Screenshot from 2020-05-07 17-58-38

An Operator is formed by two CRDs a `Subscription` that's created per Operator and a `OperatorGroup` which is created per namespace.

Note: A namespaced Operator cannot be installed globally nor in the `operators` namespace.

TODO:
 - Adapt e2e test
 - Update docs (will do in a different PR)
 - Disable deployment form for Operators already deployed (will do in a different PR)

### Applicable issues

  - ref #1703


dashboard/src/components/OperatorNew/OperatorNew.test.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+it("calls getOperator when changing the namespace the component", () => {
+  const getOperator = jest.fn();
+  const wrapper = shallow(<OperatorNew {...defaultProps} getOperator={getOperator} />);
+  expect(getOperator).toHaveBeenCalledWith(defaultProps.namespace, defaultProps.operatorName);
+  wrapper.setProps({ namespace: "foo" });
+  expect(getOperator).toHaveBeenCalledWith("foo", defaultProps.operatorName);
+});
+
+it("parses the default channel when receiving the operator", () => {
```
absoludity, line 62: Is it implicit here that the 0th (first) channel is the default one? (if so, clearer as "parses the first channel when receiving the operator")

dashboard/src/components/OperatorNew/OperatorNew.test.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const wrapper = shallow(<OperatorNew {...defaultProps} />);
+  wrapper.setProps({ operator: defaultOperator });
+  expect(wrapper.state()).toMatchObject({
+    updateChannel: defaultOperator.status.channels[0],
+    updateChannelGlobal: false,
+    installationModeGlobal: false,
+  });
+});
+
+it("shows an error if it exists", () => {
```
absoludity, line 72: if it doesn't exist?

dashboard/src/components/OperatorNew/OperatorNew.test.tsx
```diff
@@ -0,0 +1,124 @@
...
+    <OperatorNew
+      {...defaultProps}
+      namespace="operators"
+      createOperator={createOperator}
+      push={push}
+    />,
+  );
+  wrapper.setProps({ operator: defaultOperator });
+  wrapper.find("form").simulate("submit");
+  await wait(1);
```
absoludity, line 120: hmph... did you try some of the things I mentioned last time? Are you certain we're not adding potential flaky tests which are going to bite us in the future? I'd much prefer that we not continue sprinkling the tests with wait, and instead (in order of my preference, but I don't care):
1) test the onSubmit handler separately - we don't need to test that simulating the form's submit event triggers the handler, necessarily,
2) using mount instead of shallow, and simulating on the button instead (see https://github.com/enzymejs/enzyme/issues/308 for other ideas, though I've not tested them).

dashboard/src/containers/OperatorNewContainer/OperatorNewContainer.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+function mapStateToProps(
+  { operators, namespace }: IStoreState,
+  { match: { params } }: IRouteProps,
+) {
+  return {
+    namespace: namespace.current,
+    isFetching: operators.isFetching,
+    operator: operators.operator,
+    error: operators.errors.fetch || operators.errors.create,
```
absoludity, line 26: Why not let the component handle `operator.errors` instead? (if there can be multiple concurrently - not sure).

dashboard/src/shared/Operators.ts
```diff
@@ -95,4 +95,65 @@ export class Operators {
...
+    name: string,
+    channel: string,
+    installPlanApproval: string,
+    csv: string,
+  ) {
+    // First create the OperatorGroup if needed
+    await this.createOperatorGroupIfNotExists(namespace);
+    // Now create the subscription
+    const { data: result } = await axiosWithAuth.post<IResource>(
+      urls.api.operators.subscription(namespace, name),
```
absoludity, line 110: Something to think about longer term, rather than for this PR: is the fact that we're interacting with the k8s API directly from the dashboard for all these calls going to be something which helps or hinders us when we start looking at multi-cluster support? (or neither even). If we are able to run kubeops in the one cluster where Kubeapps is installed only, and have it operate with the users credentials on multiple clusters, then the dashboard would only ever interact with kubeops. If on the other hand, we expect the dashboard to be interacting directly with the k8s API, then some functional awareness of the dashboard to determine which cluster is necessary. It may be easy enough do a mixture of both if needed.

Perhaps interacting directly with the k8s API would make it easier to extract functionality into plugins. When using an authProxy, I've no problem at all security-wise with doing this, but when using tokens stored in the browser client it's a bit of a worry. Thoughts?

dashboard/src/containers/OperatorNewContainer/OperatorNewContainer.tsx
andresmgot: both errors cannot happen at the same time. It either fails to gather the operator info (fetch) or once the info is available it fails to create an operator (create). I did this because it's slightly easier to manage in the component logic.

dashboard/src/shared/Operators.ts
andresmgot: IMO I would rather not expose the whole Kubernetes API and use kubeops to return only the necessary information, although I know that cause a higher bar to implement new functionality (and it's why I am using it for the Operators stuff). If we have the time to refactor that, it'd be great.

absoludity: I don't think that we'll ever be given the time to do things like this - especially if we keep adding to it. If we're considering it as tech-debt, let's not add it in the first place (and consider updating it with any changes to the operator support work, or consider it part of moving the operator work out of alpha/beta).

Issue #1726, opened 2020-05-09
Upgrade to Helm 3.2.0+


batiati:
Cannot install a chart that uses the function "fromJsonArray" released on [Helm 3.2.0](https://github.com/helm/helm/releases/tag/v3.2.0)

```
Release "XYZ" failed and has been uninstalled: parse error at (/templates/_helpers.tpl:74): function "fromJsonArray" not defined
```

It works from helm install command

batiati:
Workarround: refactory the json file, adding a root element and use the "fromJson" function instead of "fromJsonArray".

This:
```
[ ... ]
```

Became this:
```
{
    Items: [ ... ]
}
```

absoludity:
Thanks. Yes, I only just updated Kubeapps services to 3.1, but we can update again easily enough :)

PR #1731, opened 2020-05-12
Test helm update to 3.2

absoludity:
Testing update to helm 3.2 to address #1726

This is ready for review now. The relevant changes are:

1. cmd/apprepository-controller/hack/update-codegen.sh: I've included instructions to simplify the process next time
2. The signature for the k8s client changed to include a context (I've used `context.TODO()` throughout as we don't currently pass the context through),
3. The test behaviour of the helm client changed, as it now creates a release record even if there are no resources to create (ie. k8s client's Create is never called), so updated the test so the client throws the error earlier (on client Bulid).


andresmgot:
_sounds good, doesn't work_

absoludity:
> _sounds good, doesn't work_

Heh, yes. I've got changes locally to actually update the generated code, I'll get back to it later or tomorrow.

absoludity:
So after merging master (small conflict), CI is failing with an odd docker-related message:
```
Successfully tagged kubeapps/dashboard-ci:build-5ec9dbd9f6450d3f15b9a612e1a9e5479eacd2d6
Error response from daemon: client version 1.40 is too new. Maximum supported API version is 1.39

Exited with code exit status 1
```
So it builds and tags the images fine, but then fails to push because the docker client is too new. Have you seen something similar @andresmgot ?

andresmgot:
> So it builds and tags the images fine, but then fails to push because the docker client is too new. Have you seen something similar @andresmgot ?

Mmh, no, maybe they (CircleCI) have updated the build image and the docker cli version but not the docker daemon (which is in a VM) :/

absoludity:
> Mmh, no, maybe they (CircleCI) have updated the build image and the docker cli version but not the docker daemon (which is in a VM) :/

Retested today and it's working, so they've fixed the discrepancy. Thanks!

PR #1733, opened 2020-05-12
Fix selfLink assumption

### LINKED ISSUE #1695: Bitnami Redis chart causes JS Error
### Description:

I've deployed the redis chart from the Bitnami charts repository at https://charts.bitnami.com/bitnami and when I try to view the deployed instance in Kubeapps it causes a type error in the Chrome console and Kubeapps pops up a page that says `Sorry! Something went wrong.`

1. Deploy redis chart from bitnami charts repo
2. Attempt to view the deployed redis chart
3. Kubeapps UI fails to render the output

Failed to view deployed chart

View the deployed chart

I'm running a devel version of kubeapps since I'm creating a PR for modifying some OAuth URI variables in the dashboard config.
- Console Error:
```
TypeError: Cannot read property 'type' of undefined
    at t.b (ServiceItem.tsx:41)
    at Yo (react-dom.production.min.js:153)
    at ys (react-dom.production.min.js:261)
    at yu (react-dom.production.min.js:246)
    at vu (react-dom.production.min.js:246)
    at uu (react-dom.production.min.js:239)
    at react-dom.production.min.js:123
    at t.unstable_runWithPriority (scheduler.production.min.js:19)
    at Wi (react-dom.production.min.js:122)
    at Ki (react-dom.production.min.js:123)
```

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.2.0", GitCommit:"e11b7ce3b12db2941e90399e874513fbd24bcb71", GitTreeState:"clean", GoVersion:"go1.13.10"}
```

```
kubeapps        services        91              2020-04-24 14:20:31.165057811 -0700 PDT deployed        kubeapps-3.5.3  DEVEL
```

```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.8", GitCommit:"ec6eb119b81be488b030e849b9e64fda4caaf33c", GitTreeState:"clean", BuildDate:"2020-03-12T21:00:06Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.8", GitCommit:"ec6eb119b81be488b030e849b9e64fda4caaf33c", GitTreeState:"clean", BuildDate:"2020-03-12T20:52:22Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
```

andresmgot:
### Description of the change

When the `kind` of a resource is not available (for example, when receiving an element of a list), we rely in the `selfLink` of the component to discover its type. We were looking in the whole `selfLink` which caused that if a resource is deployed in the namespace "services" or is called "services" it would be treated as a service. To fix that we just look for the place in which the type is located.

### Applicable issues

  - fixes #1695

### DIFF
--- a/dashboard/src/components/AppView/ResourceTable/ResourceItem/ResourceTableItem.test.tsx
+++ b/dashboard/src/components/AppView/ResourceTable/ResourceItem/ResourceTableItem.test.tsx
@@ -7,6 +7,7 @@ import { IKubeItem, IResource } from "../../../../shared/types";
 import DeploymentItemRow from "./DeploymentItem/DeploymentItem";
 import OtherResourceItem from "./OtherResourceItem";
 import ResourceTableItem from "./ResourceTableItem";
+import SecretItem from "./SecretItem";
 
 const kubeItem: IKubeItem<IResource> = {
   isFetching: false,
@@ -158,5 +159,35 @@ context("when there is a valid resouce", () => {
       <ResourceTableItem {...defaultProps} resource={kubeItem} name={cm.metadata.name} />,
     );
     expect(wrapper.find(OtherResourceItem)).toExist();
+  });
+
+  it("ignores the namespace if it matches a known type", () => {
+    const deployment = {
+      metadata: {
+        name: "foo",
+        selfLink: "/api/v1/namespaces/deployments/secrets/kubernetes",
+      },
+      status: { replicas: 1, updatedReplicas: 1, availableReplicas: 1 },
+    } as IResource;
+    kubeItem.item = deployment;
+    const wrapper = shallow(
+      <ResourceTableItem {...defaultProps} resource={kubeItem} name={deployment.metadata.name} />,
+    );
+    expect(wrapper.find(SecretItem)).toExist();
+  });
+
+  it("ignores the resource name if it matches a known type", () => {
+    const deployment = {
+      metadata: {
+        name: "foo",
+        selfLink: "/api/v1/namespaces/default/secrets/services",
+      },
+      status: { replicas: 1, updatedReplicas: 1, availableReplicas: 1 },
+    } as IResource;
+    kubeItem.item = deployment;
+    const wrapper = shallow(
+      <ResourceTableItem {...defaultProps} resource={kubeItem} name={deployment.metadata.name} />,
+    );
+    expect(wrapper.find(SecretItem)).toExist();
   });
 });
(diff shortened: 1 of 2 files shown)


dashboard/src/components/AppView/ResourceTable/ResourceItem/ResourceTableItem.test.tsx
```diff
@@ -159,4 +160,34 @@ context("when there is a valid resouce", () => {
...
+      status: { replicas: 1, updatedReplicas: 1, availableReplicas: 1 },
+    } as IResource;
+    kubeItem.item = deployment;
+    const wrapper = shallow(
+      <ResourceTableItem {...defaultProps} resource={kubeItem} name={deployment.metadata.name} />,
+    );
+    expect(wrapper.find(SecretItem)).toExist();
+  });
+
+  it("ignores the resource name if it matches a known type", () => {
```
absoludity, line 179: I think what you mean in both cases is something like: "it parses the type from the position in the self-link". I was confused at first as it's not that anything is being ignored, but rather, prior to this change, the position was not being used and so any matching type string would match?

dashboard/src/components/AppView/ResourceTable/ResourceItem/ResourceTableItem.tsx
```diff
@@ -86,21 +86,34 @@ class WorkloadItem extends React.Component<IResourceItemProps> {
...
+
+  private typeFromSelfLink(link: string) {
+    const parsedLink = link.split("/");
+    if (parsedLink.length < 2) {
+      // Unknown selflink
+      return "";
+    }
+    // For a single resource, the type is the second-to-last item
+    // e.g. /api/v1/namespaces/default/pods/foo
+    return parsedLink[parsedLink.length - 2];
```
absoludity, line 115: Minor point, but it seems like `typeFromSelfLink` is also where the switch should belong: it's the function which encapsulates returning the resource type (ie. not the string, but the object). Up to you.

dashboard/src/components/AppView/ResourceTable/ResourceItem/ResourceTableItem.test.tsx
andresmgot: Yes, maybe I didn't explained myself. I wanted to say that if a secret is deployed in the namespace `services` or is called `services` it still gets detected as a secret

dashboard/src/components/AppView/ResourceTable/ResourceItem/ResourceTableItem.tsx
andresmgot: the goal of this function was to be a bit more self-readable, if it's not I will just use `renderResource` above

PR #1739, opened 2020-05-14
Avoid HTTP error codes when running validation

### LINKED ISSUE #1699: Introducing wrong credentials when creating a App Repository cause a logout
### Description:

Details at https://github.com/kubeapps/kubeapps/pull/1692

We should return a 200 and send a JSON response when validating a repo.

andresmgot:
### Description of the change

Avoid just forwarding the response when running the App Repository validation.

Details at #1692

### Applicable issues

  - fixes #1699

### DIFF
--- a/dashboard/src/actions/repos.test.tsx
+++ b/dashboard/src/actions/repos.test.tsx
@@ -699,7 +699,7 @@ describe("checkChart", () => {
 describe("validateRepo", () => {
   it("dispatches repoValidating and repoValidated if no error", async () => {
     AppRepository.validate = jest.fn(() => {
-      return { Code: 200, Message: "OK" };
+      return { code: 200, message: "OK" };
     });
     const expectedActions = [
       {
@@ -707,7 +707,7 @@ describe("validateRepo", () => {
       },
       {
         type: getType(repoActions.repoValidated),
-        payload: { Code: 200, Message: "OK" },
+        payload: { code: 200, message: "OK" },
       },
     ];
 
--- a/dashboard/src/actions/repos.test.tsx
+++ b/dashboard/src/actions/repos.test.tsx
@@ -735,7 +735,7 @@ describe("validateRepo", () => {
 
   it("dispatches checkRepo and errorRepos when the validation cannot be parsed", async () => {
     AppRepository.validate = jest.fn(() => {
-      return { statusCode: 409 };
+      return { error: "forbidden" };
     });
     const expectedActions = [
       {
@@ -744,7 +744,7 @@ describe("validateRepo", () => {
       {
         type: getType(repoActions.errorRepos),
         payload: {
-          err: new Error('Unable to parse validation response, got: {"statusCode":409}'),
+          err: new Error('Validation failed, got: {"error":"forbidden"}'),
           op: "validate",
         },
       },
--- a/dashboard/src/actions/repos.test.tsx
+++ b/dashboard/src/actions/repos.test.tsx
@@ -698,14 +698,16 @@ describe("checkChart", () => {
 
 describe("validateRepo", () => {
   it("dispatches repoValidating and repoValidated if no error", async () => {
-    AppRepository.validate = jest.fn(() => "OK");
+    AppRepository.validate = jest.fn(() => {
+      return { Code: 200, Message: "OK" };
+    });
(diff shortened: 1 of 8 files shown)


pkg/http-handler/http-handler.go
```diff
@@ -111,22 +110,12 @@ func UpdateAppRepository(handler kube.AuthHandler) func(w http.ResponseWriter, r
...
-		res, err := handler.AsUser(token).ValidateAppRepository(req.Body, mux.Vars(req)["namespace"])
+		err := handler.AsUser(token).ValidateAppRepository(req.Body, mux.Vars(req)["namespace"])
 		if err != nil {
-			returnK8sError(err, w)
-			return
-		}
-		body, err := ioutil.ReadAll(res.Body)
-		if err != nil {
-			returnK8sError(err, w)
+			w.Write([]byte(err.Error()))
```
absoludity, line 115: Wouldn't this mean that we're now parsing the response body and string matching to determine an error?

pkg/http-handler/http-handler_test.go
```diff
@@ -253,15 +254,18 @@ func TestValidateAppRepository(t *testing.T) {
...
 			name:         "it should return OK if no error is detected",
 			expectedCode: 200,
+			expectedBody: "OK",
 		},
 		{
 			name:         "it should return the error code if given",
 			err:          fmt.Errorf("Boom"),
-			expectedCode: 500,
+			expectedCode: 200,
+			expectedBody: "Boom",
```
absoludity, line 268: Right, this no longer returns the error code. Sorry if it wasn't clear on the previous MP, but what I'd expect here is that *our* API call returns a 200 (as you've done), because it has done its job successfully, but the response data should contain the result of the validation. So here I'd expect that the (json) response include both the status code of the external request, as well as the error message (which you've included as the body).

So a successful validation returns a 200 with a body:
```
{
  valid: true
}
```

and for a validation that didn't succeed:
```
{
  valid: false,
  error: {
    code: 500,
    message: "Boom"
  }
}
```

or something similar. Yes, it works if we depend on the body text being the exact string "OK", but it limits what info we can give the user (ie. with the current change, a blank response from the external URL would give no info, whereas including both the externally received status code with the message will be clearer). But even without that, I'd be keen to avoid an API which requires the client to parse a string response to choose how to handle it.

andresmgot: it's true that we are missing information without the status code

dashboard/src/actions/repos.test.tsx
```diff
@@ -735,7 +737,7 @@ describe("validateRepo", () => {
 
   it("dispatches checkRepo and errorRepos when the validation cannot be parsed", async () => {
     AppRepository.validate = jest.fn(() => {
-      return { statusCode: 409 };
+      return { error: "forbidden" };
```
absoludity, line 740: Shouldn't this be returning a `{ code: 409, message: "forbidden" }`? EDIT: Ah, the point is that the error is the same one in that is stringified in the payload below, was just confused by the unexpected data.

pkg/kube/kube_handler.go
```diff
@@ -380,14 +387,26 @@ func getValidationCliAndReq(appRepoBody io.ReadCloser, requestNamespace, kubeapp
...
-func (a *userHandler) ValidateAppRepository(appRepoBody io.ReadCloser, requestNamespace string) (*http.Response, error) {
+func doValidationRequest(cli HTTPClient, req *http.Request) (ValidationResponse, error) {
+	res, err := cli.Do(req)
+	if err != nil {
+		// If the request fail, it's not an internal error
+		return ValidationResponse{Code: 400, Message: err.Error()}, nil
+	}
+	body, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		return ValidationResponse{}, fmt.Errorf("Unable to parse validation response. Got: %v", err)
```
absoludity, line 398: If you use a pointer type for the returned `*ValidationResponse` this can just be nil? (and below)

dashboard/src/actions/repos.test.tsx
andresmgot: right, I keep setting dummy data in the tests but it's true that it's more readable with a real result.

PR #1748, opened 2020-05-20
Fix rollback bug

andresmgot:
### Description of the change

After doing the rollback, it's re-requesting the App but the parameters were switched and a 404 error is being showed now after running a rollback.

### DIFF
--- a/dashboard/src/actions/apps.test.tsx
+++ b/dashboard/src/actions/apps.test.tsx
@@ -183,22 +183,22 @@ describe("delete applications", () => {
     App.delete = deleteAppOrig;
   });
   it("delete an application", async () => {
-    await store.dispatch(actions.apps.deleteApp("foo", "default", false));
+    await store.dispatch(actions.apps.deleteApp("default", "foo", false));
     const expectedActions = [
       { type: getType(actions.apps.requestDeleteApp) },
       { type: getType(actions.apps.receiveDeleteApp) },
     ];
     expect(store.getActions()).toEqual(expectedActions);
-    expect(deleteAppMock.mock.calls[0]).toEqual(["foo", "default", false]);
+    expect(deleteAppMock.mock.calls[0]).toEqual(["default", "foo", false]);
   });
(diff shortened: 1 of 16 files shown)


dashboard/src/actions/apps.test.tsx
```diff
@@ -363,16 +363,20 @@ describe("rollbackApp", () => {
...
 
     const expectedActions = [
       { type: getType(actions.apps.requestRollbackApp) },
       { type: getType(actions.apps.receiveRollbackApp) },
       { type: getType(actions.apps.requestApps) },
+      { type: getType(actions.apps.selectApp), payload: true },
     ];
     expect(store.getActions()).toEqual(expectedActions);
     expect(App.rollback).toHaveBeenCalledWith("my-release", "default", 1);
+    // It should re-get the app after the rollback, note that the parameters are switched
```
absoludity, line 378: Can we instead update `App.getRelease` (and call-sites) to be consistent with all the other signatures in `shared/App.ts`?

andresmgot: actually the "problem" is `getAppWithUpdateInfo` and `getRelease`, all the other functions has the `releaseName` before the `namespace`. I didn't want to change the functions though because it was riskier... Anyway, will try

andresmgot: well, actually no, `getAppWithUpdateInfo` is used in many places along with other functions that use the `namespace, name` structure which I believe is better because is closer to an API call in Kubernetes, we would need to change all App.* functions to use that approach instead... still quite risky

absoludity: But it has to be done, no? I'll +1 this so you can land it if you want, but I don't understand why we wouldn't fix it to be consistent now and avoid confusion / risk in the future.

andresmgot: yes, I just didn't have the time yesterday to complete this, updating this

PR #1749, opened 2020-05-21
Switch to nodeport for dev env on kind cluster.

absoludity:
While trialing an idea for an easy-to-demo kubeapps VM, I found I could simplify access to the Kubeapps+OIDC setup. This updates the local dev environment (currently used solely by me, I believe, as it's undocumented and dev only) so that:

1. Both dex and kubeapps use a service type of NodePort. Since the kind-cluster is a single node (container), this allows the k8s node (the docker container) to be addressable both on the host and inside the cluster (including from the API server)
2. Since the kind k8s API server already has a valid CA cert which includes v3 extensions for additional alternative name, including the IP address of the node, we can configure the api server to use this for the oidc ca cert and use the same on dex, simplifying the setup.

With this change, no port-forwards are required to be run for dex and kubeapps to be reachable on the host, so it just works :)

I'll later follow up with a couple of changes, but not high priorities right now:
1. Update kubeapps to https using the same cert
2. Ensure we don't depend on the assumption of the kind docker network being the first bridge network (see hard-coded IPs)
3. Replace the makefiles with additions to @juan131 's scripts.


docs/user/manifests/kubeapps-local-dev-dex-values.yaml
```diff
@@ -3,15 +3,18 @@ certs:
   web:
     create: false
 config:
-  issuer: https://dex.dex:32000
+  # TODO: Don't assume that kind network will be the first non-default bridged network (ie. 172.18.0.2)
+  # This means we'll need to generate the certs as well, rather than pre-configuring them.
+  # Or check if we can use cert generated for tls with k8s node?
+  issuer: https://172.18.0.2:32000
```
andresmgot, line 9: I am not sure how you got this, I believe that in my case is 172.17.0.2 (I may be wrong and this IP appears when using a NodePort)

absoludity: Do you mean that the `kind` bridge network is `172.17.0.0/16` for you? That's unexpected. Locally, I see the default bridge network `bridge` on that range, while kind's custom network on the next available `172.18.0.0/16`, as shown here:

```
docker network inspect bridge kind --format "{{.Name}} {{.IPAM.Config}}"

bridge [{172.17.0.0/16  172.17.0.1 map[]}]
kind [{172.18.0.0/16  172.18.0.1 map[]} {fc00:f853:ccd:e793::/64  fc00:f853:ccd:e793::1 map[]}]
```

This also matches what's seen in the docker network tutorial for the default `bridge` and the first custom one? https://docs.docker.com/network/network-tutorial-standalone/

absoludity:
> interesting, you still need to have a gcloud project configured for this, right?

Nope (nor did you previously). It's using dex in an example mode with config user data as it did before. No gcloud project required.

docs/user/manifests/kubeapps-local-dev-dex-values.yaml
andresmgot: I don't have a "kind" network (only `bridge`):

```
▶ docker network inspect bridge kind --format "{{.Name}} {{.IPAM.Config}}"
bridge [{172.17.0.0/16  172.17.0.1 map[]}]
Error: No such network: kind
```

Probably something is misconfigured on my side though because between restarts I just re-start the `kind-control-plane` container which is probably not enough. BTW, how do you handle it when you restart your laptop?

andresmgot:
> > interesting, you still need to have a gcloud project configured for this, right?
> 
> Nope (nor did you previously). It's using dex in an example mode with config user data as it did before. No gcloud project required.

Right, I forgot that there were an option to just use static data there.

docs/user/manifests/kubeapps-local-dev-dex-values.yaml
absoludity: On both points, I'm guessing the issue is that you're not using the 0.8.x release of kind, where both (breaking) changes were added: https://github.com/kubernetes-sigs/kind/releases :) Sorry, I should have mentioned that was required.

Issue #1752, opened 2020-05-26
[multi-cluster] Configuration support for additional clusters


absoludity:
For multi-cluster support, Kubeapps will require install/upgrade time configuration specifying any additional clusters which should be available to users.

Parent task: #1753 

- [x] Update frontend config so that the current default cluster is just one of potentially many clusters
- [x] Update dashboard URLs to include a ‘default’ cluster explicitly
- [x] Enable configuring additional clusters in the chart configuration and updating the frontend nginx config accordingly

absoludity:
Complete.

PR #1757, opened 2020-06-01
Add a shared k8s-api-proxy.conf and default cluster location.

absoludity:
Prep work so that we can soon route /api/clusters/cluster-name/ from
configuration to the correct k8s cluster. Adds a default cluster which
uses the local api server (to the pod), while leaving the existing
location for temporary backwards compat.

Ref: #1752

### DIFF
--- a/chart/kubeapps/templates/kubeapps-frontend-config.yaml
+++ b/chart/kubeapps/templates/kubeapps-frontend-config.yaml
@@ -8,6 +8,24 @@ metadata:
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
 data:
+  k8s-api-proxy.conf: |-
+    # Disable buffering for log streaming
+    proxy_buffering off;
+    # Hide Www-Authenticate to prevent it triggering a basic auth prompt in
+    # the browser with some clusters
+    proxy_hide_header Www-Authenticate;
+
+    # Keep the connection open with the API server even if idle (the default is 60 seconds)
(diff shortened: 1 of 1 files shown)


chart/kubeapps/templates/kubeapps-frontend-config.yaml
```diff
@@ -29,28 +47,23 @@ data:
         return 200 "healthy\n";
       }
 
+      # The default cluster running on the same cluster as Kubeapps.
+      location ~* /api/clusters/default {
```
andresmgot, line 51: so the plan here is to add some templating here to generate blocks like this one and point to different clusters, right?

chart/kubeapps/templates/kubeapps-frontend-config.yaml
```diff
@@ -29,28 +47,23 @@ data:
...
+        rewrite /api/clusters/default/(.*) /$1 break;
+        rewrite /api/clusters/default / break;
+        proxy_pass https://kubernetes.default;
+        include "./server_blocks/k8s-api-proxy.conf";
```
andresmgot, line 55: if `vhost.conf` and `k8s-api-proxy.conf` are at the same level shouldn't this be ` include "./k8s-api-proxy.conf";`?

absoludity: No - I thought so too initially, but they're not relative to the file, they're relative to the CWD, which is set on the CLI or in the bitnami nginx case, configured to `/opt/bitnami/nginx/conf`.

PR #1760, opened 2020-06-02
Run postupgrade job only when upgrading

andresmgot:
### Description of the change

This job creates the db secret (if it hasn't been created at installation time) and cleans up the database. As far as I can tell none of these processes are necessary when installing the app. Related to #1754

### DIFF
--- a/chart/kubeapps/templates/apprepository-jobs-postupgrade.yaml
+++ b/chart/kubeapps/templates/apprepository-jobs-postupgrade.yaml
@@ -1,11 +1,11 @@
 {{- if .Values.featureFlags.invalidateCache }}
-# Ensure db indexes are set and invalidate the chart cache during both install and upgrade.
+# Ensure db indexes are set and invalidate the chart when upgrading.
 apiVersion: batch/v1
 kind: Job
 metadata:
   name: {{ template "kubeapps.apprepository-job-postupgrade.fullname" . }}
   annotations:
-    helm.sh/hook: post-upgrade,post-install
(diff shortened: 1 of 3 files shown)


andresmgot:
_it's failing for MongoDB so I am probably forgetting something_

absoludity:
Postgres handles this because the `Sync()` command calls `InitTables` as postgres supports create table if it doesn't exist, whereas currently we'd need to update dbutils.mongodbutils with something that ensures the indicies exist. That said, I think we should move in the opposite direction, so that when we are pg-only, we always run the migration post-install/post-upgrade ?

How is this change helping #1754, given both post-install or post-upgrade are run after all resources are created?

andresmgot:
> That said, I think we should move in the opposite direction, so that when we are pg-only, we always run the migration post-install/post-upgrade ?

Why would be the migration needed in the first install? (maybe I am missing something).

> How is this change helping #1754, given both post-install or post-upgrade are run after all resources are created?

Is `post-upgrade` run after installing a chart? That would be weird? My point here was that if the job running the `postupgrade` action fails, the installation fails while this seemed to me like a step we can skip.

andresmgot:
_as talked in person, we agreed on initializing tables and collections in the first sync (if they don't exist) so we don't need to run the upgrade job when installing the chart_

Merging this PR to cut a release, let me know if you find any issue and I will open a follow up PR.

absoludity:
Sorry, yes that's fine. I didn't manage to come back and +1 after our chat.

Issue #1761, opened 2020-06-03
[multi-cluster] Update Kubeops backend to support multicluster


absoludity:
The kubeops backend which currently interfaces between user requests to install/upgrade/delete releases and the helm api interacting with the cluster needs to be updated to support interacting with different clusters.

Parent task #1753

- [x] Additional clusters to be included in kubeops configuration
- [x] Kubeops routes updated to support clusters without breaking existing endpoints
- [x] Shared backend routes updated to support clusters without breaking existing endpoints
- [x] ~~Update AuthGate helper (checks namespace access before routing through to assetsvc) to support additional clusters.~~ (not required as app repos only on defualt cluster currently).
- [x] Switch from InClusterConfig (with token) to explicit config with TLSClientConfig set so that the helm action config is correct.

PR #1765, opened 2020-06-03
Update kubeops service to include arg for additonal cluster config.

absoludity:
First part of #1761 . Follow up will enable adding the mounted file to the pod when additional clusters defined.

### DIFF
--- a/cmd/kubeops/main_test.go
+++ b/cmd/kubeops/main_test.go
@@ -1,0 +1,86 @@
+package main
+
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/kubeapps/kubeapps/cmd/kubeops/internal/handler"
+)
+
+func TestParseAdditionalClusterConfig(t *testing.T) {
+	testCases := []struct {
+		name           string
+		configJSON     string
+		expectedErr    bool
+		expectedConfig map[string]handler.AdditionalClusterConfig
+	}{
+		{
+			name:       "parses a single additional cluster",
+			configJSON: `[{"name": "cluster-2", "apiServiceURL": "https://example.com", "certificateAuthorityData": "abcd"}]`,
+			expectedConfig: map[string]handler.AdditionalClusterConfig{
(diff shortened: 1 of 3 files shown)


cmd/kubeops/main_test.go
```diff
@@ -0,0 +1,86 @@
...
+			}
+
+			if got, want := config, tc.expectedConfig; !cmp.Equal(want, got) {
+				t.Errorf("mismatch (-want +got):\n%s", cmp.Diff(want, got))
+			}
+		})
+	}
+}
+
+func createConfigFile(t *testing.T, content string) string {
```
andresmgot, line 73: To avoid having to create temporary files, I would move the `ioutil.ReadFile` action outside the `parse` function. It's not a big deal but this would prevent the tests from running in a read-only environment like a container.

cmd/kubeops/main_test.go
```diff
@@ -0,0 +1,86 @@
...
+					CertificateAuthorityData: "abcd",
+				},
+			},
+		},
+		{
+			name: "parses multiple additional clusters",
+			configJSON: `[
+	{"name": "cluster-2", "apiServiceURL": "https://example.com/cluster-2", "certificateAuthorityData": "abcd"},
+	{"name": "cluster-3", "apiServiceURL": "https://example.com/cluster-3", "certificateAuthorityData": "efgh"}
+]`,
```
andresmgot, line 35: wondering if it would make more sense to have this as a YAML to be more similar to a `.kube/config` but it's not the same so I guess it's okay as JSON too

cmd/kubeops/main_test.go
absoludity: Yeah - even when running on a read-only container we can mount a `tmpfs` aiui. The reason I include as much as possible in the function rather than in the main is exactly so we're testing as much as we can. CLI passes a path, this function verifies that it can return the config when given that path.

cmd/kubeops/main_test.go
absoludity: Yeah, I was unsure about this but went with json just because of the stdlib support. I don't think it'll be hard to configure this in the chart, but I'll find out next.

PR #1769, opened 2020-06-04
Enable kubeops to talk to other configured clusters.

absoludity:
Not yet tested IRL, I still need to check that the `kubehandler` here
works as expected or requires a separate in cluster config etc., but
this should maintain the current behaviour for non-multi-cluster.

### DIFF
--- a/cmd/kubeops/internal/handler/handler.go
+++ b/cmd/kubeops/internal/handler/handler.go
@@ -92,8 +92,10 @@ func NewClusterConfig(inClusterConfig *rest.Config, token string, cluster string
 func WithHandlerConfig(storageForDriver agent.StorageForDriver, options Options) func(f dependentHandler) handlerutil.WithParams {
 	return func(f dependentHandler) handlerutil.WithParams {
 		return func(w http.ResponseWriter, req *http.Request, params handlerutil.Params) {
-			cluster := params[clusterParam]
-			if cluster == "" {
+			// Don't assume the cluster name was in the url for backwards compatability
+			// for now.
+			cluster, ok := params[clusterParam]
+			if !ok {
 				cluster = defaultClusterName
 			}
 			namespace := params[namespaceParam]
--- a/cmd/kubeops/internal/handler/handler.go
+++ b/cmd/kubeops/internal/handler/handler.go
@@ -2,6 +2,7 @@ package handler
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"strconv"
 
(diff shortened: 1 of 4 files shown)


cmd/kubeops/internal/handler/handler.go
```diff
@@ -59,14 +62,27 @@ type Config struct {
...
+func NewClusterConfig(inClusterConfig *rest.Config, token string, cluster string, additionalClusters map[string]AdditionalClusterConfig) (*rest.Config, error) {
+	config := rest.CopyConfig(inClusterConfig)
 	config.BearerToken = token
 	config.BearerTokenFile = ""
+
+	if cluster == defaultClusterName {
+		return config, nil
+	}
+
+	clusterConfig, ok := additionalClusters[cluster]
```
andresmgot, line 75: I would rename `clusterConfig` (I thought it was a `rest.Config` but it seems to be short information we store). Maybe `clusterParams` or `clusterProperties`

cmd/kubeops/internal/handler/handler_test.go
```diff
@@ -20,11 +20,16 @@ import (
...
 	helmTime "helm.sh/helm/v3/pkg/time"
+	"k8s.io/client-go/rest"
 
 	"helm.sh/helm/v3/pkg/release"
 )
 
-const defaultListLimit = 256
+const (
+	defaultListLimit        = 256
+	KUBERNETES_SERVICE_HOST = "KUBERNETES_SERVICE_HOST"
```
andresmgot, line 30: this would give me a lint warning: `don't use ALL_CAPS in Go names; use CamelCase`
also, for exported consts you need to give it a comment

cmd/kubeops/main.go
```diff
@@ -90,12 +90,19 @@ func main() {
 	// Routes
 	// Auth not necessary here with Helm 3 because it's done by Kubernetes.
 	addRoute := handler.AddRouteWith(r.PathPrefix("/v1").Subrouter(), withHandlerConfig)
+	// Deprecate non-cluster-aware URIs.
 	addRoute("GET", "/releases", handler.ListAllReleases)
 	addRoute("GET", "/namespaces/{namespace}/releases", handler.ListReleases)
 	addRoute("POST", "/namespaces/{namespace}/releases", handler.CreateRelease)
 	addRoute("GET", "/namespaces/{namespace}/releases/{releaseName}", handler.GetRelease)
 	addRoute("PUT", "/namespaces/{namespace}/releases/{releaseName}", handler.OperateRelease)
 	addRoute("DELETE", "/namespaces/{namespace}/releases/{releaseName}", handler.DeleteRelease)
```
andresmgot, line 99: I guess it's better to do add a redirect in the nginx conf rather than maintaining here both set of routes. Can we do that?

cmd/kubeops/internal/handler/handler.go
absoludity: Renamed to `additionalCluster` given that it's coming from the `additionalClusters` map.

cmd/kubeops/internal/handler/handler_test.go
absoludity: Woops, these were left over from a different approach. Also installed golint in my new env :)

cmd/kubeops/main.go
absoludity: We could add redirect(s) in the nginx conf, but I don't know that it's better in this case. Here I'm ensuring they still work while the frontend has not been updated, but will remove them once the frontend always sends the cluster name. I won't be keeping them beyond that.

Issue #1544, opened 2020-02-27
CI should test chart upgrades


absoludity:
Currently our CI verifies that an installation of kubeapps succeeds and does end-2-end testing on that.

When relevant (ie. a chart version change), we should additionally check the install from the previous version and upgrade to the new version. This would have prevented #1542 from being a problem.

PR #1771, opened 2020-06-04
Bump chart version. Fix upgrade issues

andresmgot:
### Description of the change

Bump chart version to release v1.10.1.

Fixes a couple of issues:

 - AppRepositories were moved out of the `post-install` hook but this is a breaking change. When upgrading, the AppRepositories get detected as a new resource but conflicting with the one existing. I have restored the hook annotation for the moment.
 - There was a typo in the secret creation for PostgreSQL when upgrading from MongoDB.

### DIFF
--- a/chart/kubeapps/Chart.yaml
+++ b/chart/kubeapps/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kubeapps
-version: 3.6.0
+version: 3.7.0
 appVersion: DEVEL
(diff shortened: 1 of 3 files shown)


andresmgot:
_merging this to trigger the release_

absoludity:
I'm moving the "CI should test chart upgrades" (#1544) into the discussion for the next iteration. It's not fair that you have to deal with untested upgrades with manual QA during the release process. Thanks for the fixes.

PR #1772, opened 2020-06-04
Clarify token usage in getting started doc.

absoludity:
Ensure people are aware that creating service-tokens for user login is only for easy demoing of Kubeapps, but SSO should be used for any real setup.

### DIFF
--- a/docs/user/getting-started.md
+++ b/docs/user/getting-started.md
@@ -33,7 +33,7 @@ The above commands will deploy Kubeapps into the `kubeapps` namespace in your cl
 
 ## Step 2: Create a Kubernetes API token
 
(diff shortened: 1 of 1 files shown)


absoludity:
Andres, I'm landing this now as I need to link to the Getting started for a discussion, but noticed that we still don't clarify the use of service tokens. Happy to update retrospectively.

andresmgot:
> but noticed that we still don't clarify the use of service tokens. Happy to update retrospectively.

What do you mean?

The changes LGTM btw.

absoludity:
I just meant that our getting started still gave the impression that using service account tokens and handing those out to your users was a valid way of doing user authn. We had never clarified that it's just for demo setup and perhaps also when you don't have access to user-auth on your cluster, but not recommended.

PR #1791, opened 2020-06-15
Update kube package handlers to be multi-cluster aware.

absoludity:
Moves NewClusterConfig (and tests) into the shared kube package so that
it can be used by other packages. Updates the AsUser helper (which
returns clientsets) to use NewClusterConfig, requiring a call-signature
change.

Also updates NewClusterConfig so as to not re-use default clusters TLS config if none configured for the selected cluster.

I'll followup with a related small fix to ensure we don't swallow errors
in AsUser.

Ref: #1761

### DIFF
--- a/pkg/kube/kube_handler.go
+++ b/pkg/kube/kube_handler.go
@@ -41,6 +41,44 @@ import (
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 )
 
+const (
+	// DefaultClusterName is the string used to identify the internal cluster.
+	DefaultClusterName = "default"
+)
+
+// AdditionalClusterConfig contains required info to talk to additional clusters.
+type AdditionalClusterConfig struct {
+	Name                     string `json:"name"`
+	APIServiceURL            string `json:"apiServiceURL"`
+	CertificateAuthorityData string `json:"certificateAuthorityData,omitempty"`
+}
+
+// AdditionalClustersConfig is an alias for a map of additional cluster configs.
+type AdditionalClustersConfig map[string]AdditionalClusterConfig
+
+// NewClusterConfig returns a copy of an in-cluster config with a custom token and/or custom cluster host
+func NewClusterConfig(inClusterConfig *rest.Config, token string, cluster string, additionalClusters AdditionalClustersConfig) (*rest.Config, error) {
+	config := rest.CopyConfig(inClusterConfig)
+	config.BearerToken = token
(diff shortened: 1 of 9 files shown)


pkg/kube/kube_handler.go
```diff
@@ -41,6 +41,44 @@ import (
...
+	additionalCluster, ok := additionalClusters[cluster]
+	if !ok {
+		return nil, fmt.Errorf("cluster %q has no configuration", cluster)
+	}
+
+	config.Host = additionalCluster.APIServiceURL
+	config.TLSClientConfig = rest.TLSClientConfig{}
+	if additionalCluster.CertificateAuthorityData != "" {
+		config.TLSClientConfig.CAData = []byte(additionalCluster.CertificateAuthorityData)
+	}
```
andresmgot, line 78: apart than the CA data, is there something we should worry about in the `TLSClientConfig`?

absoludity: Nope, not that I can see - the other fields are either not relevant (client certificate authentication) or have sensible defaults (`Insecure`, `ServerName`):

```
go doc k8s.io/client-go/rest TLSClientConfig       
package rest // import "k8s.io/client-go/rest"

type TLSClientConfig struct {
        // Server should be accessed without verifying the TLS certificate. For testing only.
        Insecure bool
        // ServerName is passed to the server for SNI and is used in the client to check server
        // ceritificates against. If ServerName is empty, the hostname used to contact the
        // server is used.
        ServerName string

        // Server requires TLS client certificate authentication
        CertFile string
        // Server requires TLS client certificate authentication
        KeyFile string
        // Trusted root certificates for server
        CAFile string

        // CertData holds PEM-encoded bytes (typically read from a client certificate file).
        // CertData takes precedence over CertFile
        CertData []byte
        // KeyData holds PEM-encoded bytes (typically read from a client certificate key file).
        // KeyData takes precedence over KeyFile
        KeyData []byte
        // CAData holds PEM-encoded bytes (typically read from a root certificates bundle).
        // CAData takes precedence over CAFile
        CAData []byte

        // NextProtos is a list of supported application level protocols, in order of preference.
        // Used to populate tls.Config.NextProtos.
        // To indicate to the server http/1.1 is preferred over http/2, set to ["http/1.1", "h2"] (though the server is free to ignore that preference).
        // To use only http/1.1, set to ["http/1.1"].
        NextProtos []string
}

```

Issue #1762, opened 2020-06-03
[multi-cluster] UX integration


absoludity:
Parent task: #1753 

- [x] Update dashboard config to include cluster names.
- [x] Update routes to include the cluster (possibly allowing current routes for the default cluster)
- [x] Update all requests to backend/kubeops to include current cluster (obviously no helm 2 support, but will need to update tiller-proxy routes to match)
- [x] Add cluster selector when additional clusters are configured
- [x] Update config for app repos to include cluster in route (and disallow route for non-default clusters for now).

PR #1793, opened 2020-06-15
Move AppListView route to be cluster-aware.

absoludity:
First of several PRs which will transition existing relevant routes in the UI to
be cluster-aware, defaulting to the "default" cluster so as to not break
the existing experience.

Ref: #1762

### DIFF
--- a/dashboard/src/components/Header/Header.tsx
+++ b/dashboard/src/components/Header/Header.tsx
@@ -42,17 +42,17 @@ class Header extends React.Component<IHeaderProps, IHeaderState> {
     {
       children: "Applications",
       exact: true,
-      namespaced: true,
+      clustered: true,
       to: "apps",
     },
     {
       children: "Catalog",
-      namespaced: true,
+      clustered: false,
       to: "catalog",
     },
     {
       children: "Service Instances (alpha)",
-      namespaced: true,
+      clustered: false,
       to: "services/instances",
     },
   ];
@@ -117,7 +117,7 @@ class Header extends React.Component<IHeaderProps, IHeaderState> {
                 <ul className="header__nav__menu" role="menubar">
                   {this.links.map(link => (
                     <li key={link.to}>
-                      <HeaderLink {...link} currentNamespace={namespace.current} />
+                      <HeaderLink {...link} currentCluster={namespace.cluster} currentNamespace={namespace.current} />
                     </li>
                   ))}
                 </ul>
--- a/dashboard/src/components/Header/Header.tsx
+++ b/dashboard/src/components/Header/Header.tsx
@@ -1,17 +1,14 @@
 import * as React from "react";
+import { LogOut, Settings } from "react-feather";
 import { NavLink } from "react-router-dom";
+import "react-select/dist/react-select.css";
 import logo from "../../logo.svg";
-
 import { INamespaceState } from "../../reducers/namespace";
 import { definedNamespaces } from "../../shared/Namespace";
-import HeaderLink, { IHeaderLinkProps } from "./HeaderLink";
-import NamespaceSelector from "./NamespaceSelector";
-
-// Icons
-import { LogOut, Settings } from "react-feather";
-
-import "react-select/dist/react-select.css";
+import { app } from "../../shared/url";
 import "./Header.css";
+import HeaderLink from "./HeaderLink";
+import NamespaceSelector from "./NamespaceSelector";
 
 interface IHeaderProps {
   authenticated: boolean;
@@ -36,26 +33,6 @@ class Header extends React.Component<IHeaderProps, IHeaderState> {
   public static defaultProps = {
     featureFlags: { reposPerNamespace: false, operators: false },
   };
-
-  // Defines the route
-  private readonly links: IHeaderLinkProps[] = [
-    {
-      children: "Applications",
-      exact: true,
-      clustered: true,
-      to: "apps",
-    },
-    {
-      children: "Catalog",
-      clustered: false,
-      to: "catalog",
-    },
-    {
-      children: "Service Instances (alpha)",
-      clustered: false,
-      to: "services/instances",
-    },
-  ];
 
   constructor(props: any) {
     super(props);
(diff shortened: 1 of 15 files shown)


dashboard/src/components/Header/Header.tsx
```diff
@@ -42,17 +42,17 @@ class Header extends React.Component<IHeaderProps, IHeaderState> {
...
       children: "Applications",
       exact: true,
-      namespaced: true,
+      clustered: true,
       to: "apps",
     },
     {
       children: "Catalog",
-      namespaced: true,
+      clustered: false,
```
andresmgot, line 50: why is the catalog not "clustered"?

dashboard/src/components/Header/Header.tsx
```diff
@@ -117,7 +117,7 @@ class Header extends React.Component<IHeaderProps, IHeaderState> {
                 <ul className="header__nav__menu" role="menubar">
                   {this.links.map(link => (
                     <li key={link.to}>
-                      <HeaderLink {...link} currentNamespace={namespace.current} />
+                      <HeaderLink {...link} currentCluster={namespace.cluster} currentNamespace={namespace.current} />
```
andresmgot, line 120: we probably want to store the `cluster` in a different object. `namespace.cluster` is a bit confusing.

dashboard/src/components/Header/HeaderLink.tsx
```diff
@@ -18,8 +20,10 @@ class HeaderLink extends React.Component<IHeaderLinkProps> {
   };
 
   public renderInternalLink() {
-    const { currentNamespace, namespaced, to } = this.props;
-    const link = currentNamespace && namespaced ? `/ns/${currentNamespace}/${to}` : to;
+    const { currentCluster, currentNamespace, clustered, to } = this.props;
+
+    let link = currentNamespace ? `/ns/${currentNamespace}/${to}` : to;
+    link = clustered ? `/c/${currentCluster}${link}` : link;
```
andresmgot, line 26: let's move all the links to `shared/url` so we have a single place with all the links. We can have a section `header` there with these links.

dashboard/src/containers/RoutesContainer/Routes.tsx
```diff
@@ -101,7 +102,7 @@ class Routes extends React.Component<IRoutesProps> {
   }
   private rootNamespacedRedirect = () => {
     if (this.props.namespace && this.props.authenticated) {
-      return <Redirect to={`/ns/${this.props.namespace}/apps`} />;
+      return <Redirect to={`/c/${this.props.cluster}/ns/${this.props.namespace}/apps`} />;
```
andresmgot, line 105: this link should be obtained from `shared/url` as well

absoludity:
> Thanks @absoludity I have some minor suggestions.

Updated :)
 
> BTW, I use vscode with the typescript plugin, that helps a lot for example it automatically order imports and fix some lint issues when saving files.

Huh - so I've been using vscode for the past month or so (since getting a beefier local computer), and the Javascript/Typescript plugin (from Microsoft) is actually built-in (you can install a plugin for the nightly build, but I assume you're not doing that). It does highlight any lint errors (but they don't stand out to my slightly colour-blind eyes), but it was not automatically ordering imports. So I needed the following config (and installing the tslint extension from MS):
```
    "[typescript]": {
        "editor.codeActionsOnSave": {
            "source.organizeImports": true,
            "source.fixAll.tslint": true
        }
    },
    "[typescriptreact]": {
        "editor.codeActionsOnSave": {
            "source.organizeImports": true,
            "source.fixAll.tslint": true
        }
    }

```
though I don't see it fixing any lint on save yet (I'll look at this later). Thanks :)

andresmgot:
LGTM but there are a couple of comments/questions that you forgot to reply.

> So I needed the following config (and installing the tslint extension from MS):

Maybe those two options are conflicting. I have just the tslint option enabled globally (not per language) and it works for me:

```
  "editor.codeActionsOnSave": {
    "source.fixAll.tslint": false
  },
```

absoludity:
> Sorry - I meant to add a review with all the comments I had pending, but turns out I just replied to your main comment and left the comments pending :/ You should see them now.

@andresmgot Thanks for the +1, I'll land but am still keen for your input on the state, in particular, before I refactor that (though we can do that on the separate PR, I guess - I'll move ahead as suggested above).

dashboard/src/components/Header/Header.tsx
andresmgot: If it's something temporary it's better to leave a comment saying why.

IMO, the catalog URI should be eventually clustered:
 - "global" App repos may differ between clusters
 - namespaced repos should be eventually supported

Also, in my opinion, the multi-cluster support shouldn't be claimed as "complete" if secondary clusters have less functionality than the main one. WDYT?

dashboard/src/components/Header/Header.tsx
andresmgot: Cool, thanks for the explanation. I'd go with the second too so we can populate the selectors and show everything. I am not worried about refreshing this data since it's just a matter of refreshing the page.

dashboard/src/components/Header/HeaderLink.tsx
andresmgot: that's fine, thanks

dashboard/src/components/Header/Header.tsx
absoludity: > If it's something temporary it's better to leave a comment saying why.

As above, the code no longer exists since updating to use `shared/url`.
> 
> IMO, the catalog URI should be eventually clustered:

In a following PR, you'll see that I did add the `/c/:cluster/` prefix to the catalog URI, because importantly, the URI defines the catalog available for install in that location (cluster+namespace), ie. the target.

> 
>     * "global" App repos may differ between clusters

Global app repos are global for the Kubeapps installation in the current design. If I switched to browse a different Kubeapps installed on a different cluster (which happened to also support a union of clusters as the first installation), it may have a different set of global app repos. But I don't plan for a single Kubeapps instalation to have different global app repos depending on the cluster - there is one Kubeapps installation (with a set of global app repos) which supports multiple clusters. Let's chat about this tonight?

> 
>     * namespaced repos should be eventually supported

Yep, they should, but this is tied together with deciding on the future of AppRepos with OCI in the mix, which we should do separately.

> 
> Also, in my opinion, the multi-cluster support shouldn't be claimed as "complete" if secondary clusters have less functionality than the main one. WDYT?

Do you mean in that initially it won't be possible to add a namespaced app repository to an additional cluster? If so, as above, if not - let me know :)

dashboard/src/components/Header/Header.tsx
absoludity: Great, I started that yesterday, should have it done today.

dashboard/src/components/Header/Header.tsx
andresmgot: > Global app repos are global for the Kubeapps installation in the current design. If I switched to browse a different Kubeapps installed on a different cluster (which happened to also support a union of clusters as the first installation), it may have a different set of global app repos. But I don't plan for a single Kubeapps instalation to have different global app repos depending on the cluster - there is one Kubeapps installation (with a set of global app repos) which supports multiple clusters. Let's chat about this tonight?

I think that's a valid use case that we should take into account (even if we don't implement it for the moment). As administrator, I may want to enable certain repositories for everyone with access to one cluster and other repos for a second cluster (in a dev vs prod environment for example).

absoludity: Definitely - I just mean that I would do that as part of the *one* Kubeapps installation - enabling the configuration of different repos on different clusters as part of the one instance of Kubeapps' configuration.

PR #1801, opened 2020-06-17
Update remaining routes for apps, catalogs and charts to be cluster-aware

absoludity:
Also found three existing issues along the way, two of which are fixed
and one documented for follow-up fixes:

1. optional-but-not-optional ICatalogItemProps.repo for chart catalog
items (documented in #1803 )
2. Current namespace lost when clicking on links for specific chart
versions. Fixed.
3. ChartHeader was using a non-namespaced link for the repo (which was
404ing). Fixed.

Ref: #1762


dashboard/src/components/Catalog/CatalogItem.tsx
```diff
@@ -45,15 +46,18 @@ const CatalogItem: React.SFC<ICatalogItemProps> = props => {
   let subIcon;
   if (type === "chart") {
     tag1 = (
+      // TODO (absoludity): It appears that repo is an optional prop only because the item
+      // may be an operator. We could instead use polymorphism with two separate
+      // types ICatalogItemChart and ICatalogItemOperator which both implement
+      // a required interface for use in the component?
```
andresmgot, line 52: interfaces/inheritance are discouraged in react (if I understood your comment).

What we could do is move the link/icon resolution to the caller in where we are sure that for example, `repo` is populated for charts and remove this `if/else` from here.

dashboard/src/components/ChartView/ChartVersionsList.tsx
```diff
@@ -21,13 +21,17 @@ class ChartVersionsList extends React.Component<IChartVersionsListProps, IChartV
...
     const versions = this.state.showAll ? this.props.versions : this.props.versions.slice(0, 5);
     const items = versions.map(v => {
+      const chartData = v.relationships.chart.data;
       const selectedClass =
         this.props.selected.attributes.version === v.attributes.version
           ? "type-bold type-color-action"
           : "";
+      // TODO(absoludity): the namespace in the url call below should be the current namespace, not the repo namespace.
+      // For example, if a global chart is being viewed from `my-namespace`, we don't want to switch to the `kubeapps`
+      // namespace when selecting a specific chart version (as the app currently does).
```
andresmgot, line 31: this is a bug, is there anything preventing you from fixing it now? (I'm worried we forget about this)

dashboard/src/components/Catalog/CatalogItem.tsx
absoludity: > interfaces/inheritance are discouraged in react (if I understood your comment).

I don't think interfaces are discouraged? What you normally want to discourage is indeed inheritance (ie. prefer composition over inheritance) but I didn't mention inheritance here - it'd be composition.
 
> What we could do is move the link/icon resolution to the caller in where we are sure that for example, `repo` is populated for charts and remove this `if/else` from here.

We could - but not sure it's neater, it just avoids dealing with the fact that we're representing two different things (charts and operators) with a single implementation and if/else's.

Anyway, let me move the comment to an issue so we can discuss there (created #1803). I may take a look at it tomorrow, as it's bitten me a couple of times.

dashboard/src/components/ChartView/ChartVersionsList.tsx
absoludity: Just that I wasn't sure how many hoops I'd need to jump through to pass the namespace down at the time, so was going to do it in a followup PR. Just checked though and namespace is available in the parent `ChartView` component, so it's trivial - done.

dashboard/src/components/AppView/AppControls/UpgradeButton.test.tsx
```diff
@@ -17,7 +18,7 @@ it("renders a redirect when clicking upgrade", () => {
 
   button.simulate("click");
   expect(push.mock.calls.length).toBe(1);
-  expect(push.mock.calls[0]).toEqual(["/ns/default/apps/foo/upgrade"]);
+  expect(push.mock.calls[0]).toEqual([url.app.apps.upgrade("foo", "default")]);
 });
 
 context("when a new version is available", () => {
```
andresmgot, line 24: BTW, just noticed that the "Delete" button redirects to an invalid link (AppControls.tsx:86) generating a 404

absoludity: Fixed in #1809 (also audited for other hard-coded links there). Thanks.

PR #1806, opened 2020-06-18
Use two separate types for IChartCatalogItem and IOperatorCatalogItem.

### LINKED ISSUE #1803: ICatalogItemProps encapsulates two different types with conditionals
Currently the `ICatalogItemProps` interface encapsulates the fields for both a Chart and an Operator, so that the intention of required fields becomes mired in conditionals. For example, the `repo` field is required if the implementation represents a chart, but not if it represents an operator, so it is marked as a non-required prop (`repo?`). This means that to use the repo in code which is specific to a chart - where the repo is actually required, I have to dance around this with:

```
url.app.charts.get(name, repo || {} as IRepo, namespace);
```

so that TypeScript is happy (without the `|| {}` it rightly complains that `repo` could be undefined yet it is a required arg for the call).

We should instead solve this properly so that required fields for a type are actually required.

absoludity:
It's not perfect - there's more that could be DRY'd up and encapsulated,
but with this change you can actually use TypeScript's type system to
ensure the correct fields are present.

I've not done an IRL test, this is just to demo the thought.

Fixes #1803 (maybe). Ref #1762

### DIFF
--- a/dashboard/src/components/Catalog/Catalog.tsx
+++ b/dashboard/src/components/Catalog/Catalog.tsx
@@ -9,7 +9,7 @@ import { MessageAlert } from "../ErrorAlert";
 import LoadingWrapper from "../LoadingWrapper";
 import PageHeader from "../PageHeader";
 import SearchFilter from "../SearchFilter";
-import CatalogItem, { ICatalogItem } from "./CatalogItem";
+import CatalogItem, { ICatalogItemProps, IChartCatalogItem, IOperatorCatalogItem } from "./CatalogItem";
 
 interface ICatalogProps {
   charts: IChartState;
@@ -102,9 +102,10 @@ class Catalog extends React.Component<ICatalogProps, ICatalogState> {
     const filteredCharts = this.filteredCharts(allItems);
     const filteredCSVs = this.shouldRenderOperators() ? this.filteredCSVs(csvs) : [];
     const catalogItems = this.getCatalogItems(filteredCharts, filteredCSVs);
-    const items = catalogItems.map(c => (
-      <CatalogItem key={`${c.type}/${c.repo?.name || c.csv}/${c.name}`} item={c} />
-    ));
+    const items = catalogItems.map(c => {
+      const keyComponent = c.type === "operator" ? (c.item as IOperatorCatalogItem).csv : (c.item as IChartCatalogItem).repo.name;
+      return <CatalogItem type={c.type} key={`${c.type}/${keyComponent}/${c.item.name}`} item={c.item} />
+    });
     return (
       <section className="Catalog">
         <PageHeader>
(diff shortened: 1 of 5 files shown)


dashboard/src/components/Catalog/Catalog.tsx
```diff
@@ -9,7 +9,7 @@ import { MessageAlert } from "../ErrorAlert";
 import LoadingWrapper from "../LoadingWrapper";
 import PageHeader from "../PageHeader";
 import SearchFilter from "../SearchFilter";
-import CatalogItem, { ICatalogItem } from "./CatalogItem";
+import CatalogItem, { ICatalogItemProps, IChartCatalogItem, IOperatorCatalogItem } from "./CatalogItem";
```
absoludity, line 12: The specific interfaces are only required for generating the segment of the key that differs depending on the type... not sure if that segment is required for uniqueness? (see below, line 106, where it includes either the csv or the repo name).

dashboard/src/components/Catalog/Catalog.tsx
```diff
@@ -102,9 +102,10 @@ class Catalog extends React.Component<ICatalogProps, ICatalogState> {
     const filteredCharts = this.filteredCharts(allItems);
     const filteredCSVs = this.shouldRenderOperators() ? this.filteredCSVs(csvs) : [];
     const catalogItems = this.getCatalogItems(filteredCharts, filteredCSVs);
-    const items = catalogItems.map(c => (
-      <CatalogItem key={`${c.type}/${c.repo?.name || c.csv}/${c.name}`} item={c} />
-    ));
+    const items = catalogItems.map(c => {
+      const keyComponent = c.type === "operator" ? (c.item as IOperatorCatalogItem).csv : (c.item as IChartCatalogItem).repo.name;
+      return <CatalogItem type={c.type} key={`${c.type}/${keyComponent}/${c.item.name}`} item={c.item} />
```
andresmgot, line 107: this seems unnecessary complex for just the key but I don't have a better proposal :thinking:

dashboard/src/components/Catalog/CatalogItem.tsx
```diff
@@ -38,32 +42,55 @@ function trimDescription(desc: string): string {
...
   }
+};
+
+const OperatorCatalogItem: React.SFC<IOperatorCatalogItem> = props => {
+  const { icon, name, csv, version, description, namespace, id } = props;
+  const iconSrc = icon || placeholder;
+  // Cosmetic change, remove the version from the csv name
+  const csvName = props.csv.split(".v")[0];
+  const tag1 = <span>{csvName}</span>;
+  const link = `/ns/${namespace}/operators-instances/new/${csv}/${id}`;
```
andresmgot, line 60: this link should be obtained from a function :)

dashboard/src/components/Catalog/Catalog.tsx
absoludity: It's just for the required unique key, I think? so we could instead just use the index... I wasn't sure as I was only refactoring this code, so left it as it was - let me know if there's a reason for the more structured key?

dashboard/src/components/Catalog/CatalogItem.tsx
absoludity: Yeah - grepping for `"/ns/.*operator"` shows quite a few operator URLs in non-test code that aren't using the shared urls helper (or even have a url helper defined as in this case for the operator-instance new).

PR #1807, opened 2020-06-18
Test chart upgrade

### LINKED ISSUE #1544: CI should test chart upgrades

andresmgot:
### Description of the change

The goal of this PR is to add a basic test to ensure we don't break the upgrade process (at least with the flags we use in our CI system). To avoid having another build or run the upgrade test for every build I am adding the env var `TEST_UPGRADE` to the PostgreSQL + Helm 3 scenario (the "recommended" setup).

The test is just installing the `kubeapps` chart from the `bitnami` repository and upgrading it afterwards with the status of the current `chart` folder.

### Benefits

Breaking changes in the chart will be noticed.

### Applicable issues

  - fixes #1544

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -334,6 +334,7 @@ jobs:
     environment:
       HELM_VERSION: "v3.0.2"
       KUBEAPPS_DB: "postgresql"
+      TEST_UPGRADE: "1"
     <<: *local_e2e_steps
   GKE_1_15_MASTER:
     <<: *gke_test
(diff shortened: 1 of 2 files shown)


andresmgot:
> Nice solution Andres. +1 for only testing with the recommended setup for now - we can't test all combinations (and the sooner we get the combinations down, the better), but one thing this won't allow is testing a mongodb install upgrading to pg, for example?

yes, we are not covering that use case (which is a corner case anyway). When doing "weird" changes to the default values we need to hand them carefully anyway.

PR #1808, opened 2020-06-21
Basic Forms: Hide parameters based on the deployment event (install or upgrade)

### LINKED ISSUE #1798: Basic Forms: Install-only values
### Feature

Allow to render some values as read-only for upgrades, allowing the user's input only during the chart installation.

### Benefits

Depending on the chart's implementation, changing some values after the installation can leave to unexpected results or inconsistent deployment.

For example, changing the `database` from `mariadb` to `postgresql` after the installation, may leave to data loss if the chart cannot handle such data migration.

### Drawbacks

This validation may be more suitable to be implemented on Helm itself.

batiati:
### Proposal

Fixes #1798

This pull request contains the implementation of a proposal to enhance the `hidden` condition to evaluate the deployment events `install` or `upgrde`.

It is presented as a new property `hidden.event` and can be used alone or combined with the existing rules (single condition or an array of conditions)

### Example

Parameters can also be hidden based on the deployment event `install` or `upgrade`, wherever they are no longer relevant.

```json
    "mariadb": {
      "type": "object",
      "properties": {
        "enabled": {
          "type": "boolean",
          "title": "Use a new MariaDB database hosted in the cluster",
          "form": "useSelfHostedDatabase",
          "hidden": {
            "event": "upgrade"
          }
        }
      }
    },
```

The `event` condition can be combined into a array of conditions in the same way:

```json
    "mariadb": {
      "type": "object",
      "properties": {
        "enabled": {
          "type": "boolean",
          "title": "Use a new MariaDB database hosted in the cluster",
          "form": "useSelfHostedDatabase",
          "hidden": {
            "conditions": [
            {
              "event": "upgrade"
            },
            {
              "path": "mariadb/enabled",
              "value": "true"
            }],
            "operator": "or"
          }
        }
      }
    },
```


dashboard/src/components/DeploymentFormBody/DeploymentFormBody.tsx
```diff
@@ -19,6 +19,7 @@ import Differential from "./Differential";
 import "./Tabs.css";
 
 export interface IDeploymentFormBodyProps {
+  deploymentEvent: DeploymentEvent;
```
andresmgot, line 22: since the deployment event is now explicit, can you modify here the condition in line 149? There we are guessing if we are deploying or upgrading based on the fact that the `releaseVersion` is set but I prefer the explicit approach.

andresmgot:
Also, it seems that there are some errors in the CI that could be unrelated to your changes, can you merge your branch with the latest status of master?

batiati:
I was thinking about the way both conditions `event` and `path` are evaluated if presented in the same clause, like this:

```
          "hidden": {
              "event": "upgrade"
              "path": "mariadb/enabled",
              "value": "true"
          }
```

The current implementation uses the `and` operator, which is quite useless for most scenarios, for instance, during the `install` event this field will be visible all the time, despite the `path` condition.

It would make more sense if it was evaluated using `or` operation, resulting in the same effect of:

```
          "hidden": {
            "conditions": [
            {
              "event": "upgrade"
            },
            {
              "path": "mariadb/enabled",
              "value": "true"
            }],
            "operator": "or"
          }
```

andresmgot:
> I was thinking about the way both conditions `event` and `path` are evaluated if presented in the same clause, like this:
> 
> ```
>           "hidden": {
>               "event": "upgrade"
>               "path": "mariadb/enabled",
>               "value": "true"
>           }
> ```
> 
> The current implementation uses the `and` operator, which is quite useless for most scenarios, for instance, during the `install` event this field will be visible all the time, despite the `path` condition.

IMO I would just not support that and force to either use the `event` or the `path/value` for clarity.

batiati:
> IMO I would just not support that and force to either use the `event` or the `path/value` for clarity.

Agreed!

PR #1813, opened 2020-06-23
Initialize the additional clusters state when receiving config.

absoludity:
### Description of the change

Initializes the additional clusters redux state when receiving the application configuration.

### Benefits

Part of #1762

### DIFF
--- a/dashboard/src/reducers/cluster.ts
+++ b/dashboard/src/reducers/cluster.ts
@@ -131,7 +131,7 @@ const clusterReducer = (
         default: state.clusters.default,
       };
       const config = action.payload as IConfig;
-      config.featureFlags.additionalClusters.forEach(cluster => {
+      config.featureFlags.additionalClusters?.forEach(cluster => {
         clusters[cluster.name] = {
           currentNamespace: "default",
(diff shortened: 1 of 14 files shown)


dashboard/src/reducers/cluster.ts
```diff
@@ -123,6 +125,22 @@ const clusterReducer = (
           },
         };
       }
+    case getType(actions.config.receiveConfig):
+      // Initialize the additional clusters when receiving the config.
+      const clusters = {
+        default: state.clusters.default,
+      };
+      const config = action.payload as IConfig;
+      config.featureFlags.additionalClusters.forEach(cluster => {
```
andresmgot, line 134: small fail-safe in case additionalClusters is not defined (like in a local config file): `config.featureFlags.additionalClusters &&  config.featureFlags.additionalClusters.forEach...`

absoludity: We can use `config.featureFlags.additionalClusters?.forEach` . Added a failing test and fixed with that (had to delete the property from a valid `IConfig` to stop ts from complaining :)

PR #1819, opened 2020-06-24
Enforce the dashboard code to check js prettification during build.

absoludity:
# Description of the change

Check the pretty-fication (?) of dashboard source files during build. This means that building an image locally will fail as well as CI. 

### Benefits

Can't get approval to land code that hasn't had prettier run.
```
IMAGE_TAG=1 make kubeapps/dashboard
...
src/shared/Secret.test.ts
src/shared/url.ts
Code style issues found in the above file(s). Forgot to run Prettier?
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
The command '/bin/sh -c yarn run prettier-check' returned a non-zero code: 1
make: *** [Makefile:25: kubeapps/dashboard] Error 1
```
### Possible drawbacks

Haven't checked if it creates tmp files on the image - if so, will need to delete them, or run this as a multi-staged docker file? EDIT: it's already multi stage, so no issue here.

Also, there seem to be similar targets already, but they don't work (on my container):
```
yarn tslint-check
yarn run v1.21.1
$ tslint-config-prettier-check ./tslint.ci.json
Conflict rule(s) detected in ./tslint.ci.json:
  quotemark
error Command failed with exit code 1.
```

Haven't yet checked if these are unused / stale. EDIT: appear to be default files from create-react-app.


dashboard/Dockerfile
```diff
@@ -5,9 +5,9 @@ COPY package.json yarn.lock /app/
 RUN yarn install --frozen-lockfile
 
 RUN mkdir /app/src
-RUN curl -L https://unpkg.com/@bitnami/hex@3.2.0/dist/hex.min.css > /app/src/hex.min.css
-
 COPY . /app
+RUN yarn run prettier-check
+RUN curl -L https://unpkg.com/@bitnami/hex@3.2.0/dist/hex.min.css > /app/src/hex.min.css
```
andresmgot, line 10: why did you move the `curl`? (seems like I had a reason to do it before the copy but tbh I don't remember why)

absoludity: Because the minified css in `hex.min.css` then fails the prettier-check :)

andresmgot: ah, I see, `css` files are ignored by our .gitignore, we should only checck `scss` files

absoludity: Right - updated, thanks.

PR #1821, opened 2020-06-24
Layout v2 component

andresmgot:
### Description of the change

The `Layout` is the first self-written component that we render. I am following the same approach than in #1814 to conditionally load one file or another depending on the UI in the configuration. Note that the requisites to do this are:

 - The component should have a container. This is needed to pass the featureFlags as a property. Most of our components already have this but it was not the case for the `Layout` component.
 - The index file of the component will decide which of the two versions it should load.
 - A second file (I've chosen to use the notation `.v2`) should be loaded for the Clarity UI.

Result:

With the `hex` UI:

Screenshot from 2020-06-24 12-27-20

With the `clarity` UI:

Screenshot from 2020-06-24 12-26-49

Note that the Layout component is just a wrapper, we still need to work in all the components of the different views.

### Applicable issues

  - ref #1354 

### Additional information

The change of base branch messed a bit the git history but I've also moved the `UISelector` to the layout component, I believe there is a better place rather than in the `Header` component.


dashboard/src/containers/LayoutContainer/LayoutContainer.ts
```diff
@@ -0,0 +1,17 @@
...
+  router: RouteComponentProps<{}>;
+}
+
+function mapStateToProps({ config: { featureFlags } }: IState) {
+  return {
+    UI: featureFlags.ui,
+  };
+}
+
+export default connect(mapStateToProps)(Layout);
```
absoludity, line 17: Note: you might to wait or my branch updating our redux/react-router dependencies etc. as you'll no longer need a container to access state (as it can be accessed via hooks). Similarly, updating connected-react-router forces adding the router state to the `IStoreState` as you've done manually here (ie. it'll already be done in my branch).

andresmgot: I prefer to merge this before so I don't have conflicts since it's quite simple, in any case, the route thing is not needed.

PR #1822, opened 2020-06-25
WIP: trying to update react/redux etc. I'll detail inline.

absoludity:
Not correct or finished, but in-progress with some explanations in line, in case it helps.

Note that this is the state *before* I tried ejecting. If we instead try the blog post https://joshblog.net/2018/upgrading-a-react-and-typescript-app-from-react-scripts-ts-to-create-react-app-v2-1/ , we'll probably want to start with master, do that process, then once it works, update the libs here (and related changes).

Let me know how you go, I'm happy to continue tomorrow if needed.


dashboard/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:8.9 AS build
+FROM bitnami/node:14.4.0-prod AS build
```
absoludity, line 1: I needed to update beyond 8.9 because one of the packages required a later node. I wasn't sure why we weren't using bitnami/node, so just switched to that.

dashboard/package.json
```diff
@@ -23,21 +23,24 @@
     "@types/vfile": "^4.0.0",
     "@types/vfile-message": "^2.0.0",
     "@types/ws": "^6.0.0",
-    "@types/yaml": "^1.0.2",
+    "@types/yaml": "^1.9.7",
     "ajv": "^6.10.2",
     "axios": "^0.19.0",
-    "connected-react-router": "^4.5.0",
+    "connected-react-router": "^6.8.0",
```
absoludity, line 29: This change requires some migration, as per https://github.com/supasate/connected-react-router/blob/master/FAQ.md#how-to-migrate-from-v4-to-v5v6

dashboard/package.json
```diff
@@ -50,9 +53,9 @@
     "react-markdown": "^4.2.2",
     "react-minimal-pie-chart": "^6.0.1",
     "react-modal": "^3.1.11",
-    "react-redux": "^5.0.6",
+    "react-redux": "^7.2.0",
```
absoludity, line 56: Also a big change. Not sure if there are specific migration steps.

dashboard/package.json
```diff
@@ -62,10 +65,10 @@
     "redux": "^4.0.0",
     "redux-devtools-extension": "^2.13.5",
     "redux-thunk": "^2.2.0",
+    "seamless-immutable": "^7.1.4",
```
absoludity, line 68: I also explicitly install 4 or 5 packages which were generating peer dependency warnings when building (afaict, this means you need to explicitly install a version yourself - which I did, but may not be required).

dashboard/package.json
```diff
@@ -62,10 +65,10 @@
...
     "semver": "^5.6.0",
-    "ts-jest": "^22.0.1",
```
absoludity, line 66: I wasn't sure why this was in our dependencies (since it should only be for devDependencies below right?).

dashboard/src/actions/apps.test.tsx
```diff
@@ -8,6 +8,8 @@ import Chart from "../shared/Chart";
 import { definedNamespaces } from "../shared/Namespace";
 import { IAppState, UnprocessableEntity } from "../shared/types";
 
+// TODO: do we need to mock the actual modules - https://stackoverflow.com/a/54781786 Check.
+
```
absoludity, line 12: I was at this point when I stopped: I wanted to try testing this file, but was no longer able to test since `react-scripts-ts` no longer worked. I needed to update the mocks below to satisfy the type checking, but we may also need to add
```
jest.mock("../shared/App");
jest.mock("../shared/Chart");
```
here as well.

dashboard/src/components/AppList/AppListItem.test.tsx
```diff
@@ -23,13 +22,13 @@ it("renders an app item", () => {
   const card = wrapper.find(InfoCard).shallow();
   expect(
     card
-      .find(Link)
+      .find("Link")
```
absoludity, line 25: Required to satisfy typescript (forget the error).

dashboard/src/containers/RoutesContainer/Routes.test.tsx
```diff
@@ -83,30 +83,3 @@ it("should render a redirect to the login page (when not authenticated)", () =>
   expect(wrapper.find(NotFound)).not.toExist();
   expect(wrapper.find(Redirect).prop("to")).toEqual("/login");
 });
-
-describe("Routes depending on feature flags", () => {
```
absoludity, line 87: Test is no longer relevant since the ff was removed (and the .dive().dive().dive() no longer got to the same element.)

dashboard/src/reducers/index.ts
```diff
@@ -11,7 +13,8 @@ import kubeReducer from "./kube";
 import operatorReducer from "./operators";
 import reposReducer from "./repos";
 
-const rootReducer = combineReducers<IStoreState>({
+export default (history: History<LocationState>) => combineReducers<IStoreState>({
```
absoludity, line 16: This change is part of the migration documented for connected-react-router.

dashboard/src/shared/types.ts
```diff
@@ -354,6 +356,7 @@ export interface IAppState {
 }
 
 export interface IStoreState {
+  router: RouterState<LocationState>;
```
absoludity, line 359: Ditto, as per connected-react-router migration.

dashboard/src/store/index.ts
```diff
@@ -1,16 +1,16 @@
...
 
-import rootReducer from "../reducers";
+import createRootReducer from "../reducers";
 
 // Use Hash based routing to support deploying Kubeapps in arbitrary URL subpaths
 export const history = createHashHistory();
 
 export default createStore(
-  connectRouter(history)(rootReducer), // add router state to reducer
+  createRootReducer(history), // add router state to reducer
```
absoludity, line 13: Ditto.

dashboard/Dockerfile
andresmgot: btw, the latest LTS version is v12.18.1

PR #1823, opened 2020-06-25
Replace react-scripts-ts with official react-scripts

### LINKED ISSUE #1637: Replace package react-scripts-ts
### Description:

`react-scripts-ts` is used to build the Dashboard project. Unfortunately this package is not being maintained (last update was in March 2019) and we keep finding issues because of it. We should replace it for a native way of building the application (probably using webpack directly).

andresmgot:
### Description of the change

The official `react-scripts` package supports TS so we don't need to keep using the outdated `react-script-ts`. I based the migration in the steps described here: https://joshblog.net/2018/upgrading-a-react-and-typescript-app-from-react-scripts-ts-to-create-react-app-v2-1/. For that I had to:

 - Upgrade some dependencies.
 - Update the `package.json` and `tsconfig.json` with some new properties/config options.
 - Fix a ton of linter/syntax errors.
 - Fix associated tests.

### Applicable issues

  - fixes #1637


dashboard/package.json
```diff
@@ -140,9 +140,24 @@
...
+    "production": [
+      ">0.2%",
+      "not dead",
+      "not op_mini all"
+    ],
+    "development": [
+      "last 1 chrome version",
+      "last 1 firefox version",
+      "last 1 safari version"
+    ]
```
absoludity, line 161: This looks interesting!? https://create-react-app.dev/docs/supported-browsers-features/#configuring-supported-browsers

dashboard/src/components/AppUpgrade/AppUpgrade.tsx
```diff
@@ -9,14 +9,14 @@ import SelectRepoForm from "../SelectRepoForm";
 import UpgradeForm from "../UpgradeForm";
 
 export interface IAppUpgradeProps {
-  app: IRelease;
+  app?: IRelease;
```
absoludity, line 12: Eek, why are these changing to optional props? (and then the bangs below saying they're always defined)

dashboard/src/components/AppUpgrade/AppUpgrade.tsx
```diff
@@ -27,7 +27,7 @@ export interface IAppUpgradeProps {
     values?: string,
     schema?: JSONSchema4,
   ) => Promise<boolean>;
-  fetchChartVersions: (namespace: string, id: string) => Promise<IChartVersion[]>;
+  fetchChartVersions: (namespace: string, id: string) => Promise<any>;
```
absoludity, line 30: Interested to test why this is needed too (though I trust that it is).

dashboard/src/components/AppView/AppView.test.tsx
```diff
@@ -33,7 +33,7 @@ describe("AppViewComponent", () => {
     namespace: "weee",
   });
 
-  const validProps: IAppViewProps = {
+  const validProps = {
```
absoludity, line 36: Ditto (marking now so if I land this I can check these points myself).

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -19,7 +19,7 @@ import ResourceTable from "./ResourceTable";
 export interface IAppViewProps {
   namespace: string;
   releaseName: string;
-  app: IRelease;
+  app?: IRelease;
```
absoludity, line 22: Ditto

dashboard/src/components/ConfigLoader/ConfigLoader.tsx
```diff
@@ -1,7 +1,8 @@
 import * as React from "react";
 
 import UnexpectedErrorPage from "../../components/ErrorAlert/UnexpectedErrorAlert";
-import LoadingWrapper, { ILoadingWrapperProps } from "../../components/LoadingWrapper";
+import LoadingWrapper from "../../components/LoadingWrapper";
+import { ILoadingWrapperProps } from "../LoadingWrapper/LoadingWrapper";
```
absoludity, line 5: These look like relative paths to the same component?

dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/BasicDeploymentForm.tsx
```diff
@@ -141,7 +141,7 @@ class BasicDeploymentForm extends React.Component<IBasicDeploymentFormProps> {
         if (param.render === "textArea") {
           paramComponent = (
             <TextParam
-              label={param.title || name}
+              label={param.title || param.path}
```
absoludity, line 144: Sheesh - where was `name` even defined here?

dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/Subsection.tsx
```diff
@@ -44,7 +44,7 @@ class Subsection extends React.Component<ISubsectionProps> {
   }
 
   private handleChildrenParamChange = (param: IBasicFormParam) => {
-    return (e: React.FormEvent<HTMLInputElement>) => {
+    return (e: React.FormEvent<any>) => {
```
absoludity, line 47: I'll check this too.

dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/TextParam.tsx
```diff
@@ -6,9 +6,7 @@ export interface IStringParamProps {
   label: string;
   inputType?: string;
   param: IBasicFormParam;
-  handleBasicFormParamChange: (
-    p: IBasicFormParam,
-  ) => (e: React.FormEvent<HTMLInputElement | HTMLTextAreaElement | HTMLSelectElement>) => void;
+  handleBasicFormParamChange: (param: IBasicFormParam) => (e: React.FormEvent<any>) => void;
```
absoludity, line 9: Check.

dashboard/src/components/LoadingWrapper/index.ts
```diff
@@ -1,3 +1,3 @@
-import LoadingWrapper, { ILoadingWrapperProps, LoaderType } from "./LoadingWrapper";
-export { ILoadingWrapperProps, LoaderType };
+import LoadingWrapper, { LoaderType } from "./LoadingWrapper";
+export { LoaderType };
```
absoludity, line 2: Ah - related to one of my questions above.

dashboard/src/components/OperatorView/OperatorView.tsx
```diff
@@ -19,7 +19,7 @@ interface IOperatorViewProps {
   namespace: string;
   error?: Error;
   push: (location: string) => RouterAction;
-  getCSV: (namespace: string, name: string) => Promise<IClusterServiceVersion>;
+  getCSV: (namespace: string, name: string) => Promise<any>;
```
absoludity, line 22: Check.

dashboard/src/components/UpgradeForm/UpgradeForm.tsx
```diff
@@ -33,7 +33,7 @@ export interface IUpgradeFormProps {
   ) => Promise<boolean>;
   push: (location: string) => RouterAction;
   goBack: () => RouterAction;
-  fetchChartVersions: (namespace: string, id: string) => Promise<IChartVersion[]>;
+  fetchChartVersions: (namespace: string, id: string) => Promise<any>;
```
absoludity, line 36: Check

dashboard/src/shared/Config.test.ts
```diff
@@ -31,7 +31,6 @@ describe("Config", () => {
   });
 
   it("does not returns the overriden namespace if NODE_ENV=production", async () => {
-    process.env.NODE_ENV = "production";
```
absoludity, line 34: Guessing there was some other reason for removing this, but it appears to make this contradict the previous one above, yet they both pass?

dashboard/src/components/AppUpgrade/AppUpgrade.tsx
andresmgot: they are undefined while the call to `getAppWithUpdateInfo` is being executed, we can avoid the bangs though

dashboard/src/components/AppUpgrade/AppUpgrade.tsx
andresmgot: well, actually is not any, but it could be undefined

dashboard/src/components/AppView/AppView.test.tsx
andresmgot: here is because if I let the `IAppViewProps`, then `validProps.app` may be undefined (according to ts) but if I don't set the type, `validProps.app` is always defined and I don't need to change any test.

dashboard/src/components/AppView/AppView.tsx
andresmgot: same reason, app will be undefined while fetching

dashboard/src/components/ConfigLoader/ConfigLoader.tsx
andresmgot: nope, (notice the inner LoadingWrapper) but it could be

dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/BasicDeploymentForm.tsx
andresmgot: this is tricky, `name` is a global value (that resolves to `nodejs`, weird). It's difficult to notice because the linter didn't complain, now it seem it does :)

PR #1824, opened 2020-06-25
Update jest-tests

andresmgot:
### Description of the change

Update the jest-test package and adapt the existing tests.

### DIFF
--- a/dashboard/src/components/ConfigLoader/ConfigLoader.tsx
+++ b/dashboard/src/components/ConfigLoader/ConfigLoader.tsx
@@ -8,6 +8,7 @@ import LoadingWrapper, {
 interface IConfigLoaderProps extends ILoadingWrapperProps {
   getConfig: () => void;
   error?: Error;
+  children?: any;
 }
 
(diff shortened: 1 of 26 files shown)


dashboard/src/components/ConfigLoader/ConfigLoader.tsx
```diff
@@ -7,6 +7,7 @@ import { ILoadingWrapperProps } from "../LoadingWrapper/LoadingWrapper";
 interface IConfigLoaderProps extends ILoadingWrapperProps {
   getConfig: () => void;
   error?: Error;
+  children?: any;
```
absoludity, line 10: `any`? Guessing this can be specific too?

andresmgot: well, in this case there are too many types:
```
'string | (string & {}) | (string & ReactElement<any, string | ((props: any) => ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)> | null) | (new (props: any) => Component<...>)>) | (string & ReactNodeArray) | (string & ReactPortal) | undefined'
```

PR #1825, opened 2020-06-26
Small adjustments to #1823

absoludity:
Just a couple of small changes I wanted to check from #1823.

### DIFF
--- a/dashboard/src/shared/Config.test.ts
+++ b/dashboard/src/shared/Config.test.ts
@@ -13,7 +13,7 @@ describe("Config", () => {
 
     defaultJSON = require("../../public/config.json");
 
-    moxios.stubRequest("config.json", { status: 200, response: defaultJSON });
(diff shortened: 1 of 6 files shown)


dashboard/src/shared/Config.test.ts
```diff
@@ -13,7 +13,7 @@ describe("Config", () => {
 
     defaultJSON = require("../../public/config.json");
 
-    moxios.stubRequest("config.json", { status: 200, response: defaultJSON });
+    moxios.stubRequest("config.json", { status: 200, response: { ...defaultJSON } });
```
absoludity, line 16: This was the reason it was passing either way - the actual defaultJSON data was being *updated* by the code, rather than a copy of the data. Perhaps the require only happens once - I'm not sure, but the test now fails if you comment out line 36 below (where as it didn't when you removed it due to TS complaining about a read-only value).

PR #1826, opened 2020-06-26
1762 Pass the current cluster through to calls of url.app.apps.new.

absoludity:
Starts the job of including the cluster data in all URLs, but explicitly does one component by using redux hooks (in a function component) - this is the first commit changing `ChartDeployButton`, while the other is done the normal way, passing the cluster value down through the hierarchy from the container (in the second commit, `DeploymentFormBody`).

I know we can't (and don't want to) go re-writing existing components generally, but I was interested to see how it works so updated a simple component.

Ref: #1762

It's not ready, but please take a look and see if you agree with the direction for testing function-based components with redux state. I think we can make it easy to test such functions with state.

Also done:
 * Update param url ordering
 * fix build failure:
```
TypeScript error in /app/src/containers/Root.tsx(15,10):
Type '{ children: Element; }' has no properties in common with type 'IntrinsicAttributes & Pick<ClassAttributes<ConfigLoader> & IConfigLoaderProps, "type" | "ref" | "size" | "key">'.  TS2559

    13 |     return (
    14 |       <Provider store={store}>
  > 15 |         <ConfigLoaderContainer>
       |          ^
    16 |           <ConnectedRouter history={history}>
    17 |             <React.Suspense fallback={null}>
    18 |               <Layout headerComponent={HeaderContainer}>
```
Turns out that with newer `react-redux`, the `connect` function no longer assumes that `children` is a prop, so connected components need to define it explicitly.


dashboard/package.json
```diff
@@ -50,7 +50,7 @@
     "react-markdown": "^4.2.2",
     "react-minimal-pie-chart": "^6.0.1",
     "react-modal": "^3.1.11",
-    "react-redux": "^5.0.6",
+    "react-redux": "^7.2.0",
```
absoludity, line 53: Needed for the redux hooks as per https://react-redux.js.org/api/hooks (well, update to 7.1)

dashboard/src/components/ChartView/ChartDeployButton.tsx
```diff
@@ -1,42 +1,38 @@
...
-        )}
-      </div>
+  if (clicked) {
+    const newAppURL = app.apps.new(
+      props.version,
+      currentCluster,
+      props.namespace,
+      props.version.attributes.version,
     );
+    dispatch(push(newAppURL));
```
absoludity, line 27: Switched to a dispatch as otherwise we also need to wrap the component being tested in a `<Router>` (to use Redirect - tests failed because the Redirect should not be used outside a `Router`, though not sure why the change to a function component caused that. It seems cleaner too (dispatching / calling a function, though haven't yet tested IRL)

andresmgot: that's what we do in the containers so it makes sense:
```
src/containers/AppNewContainer/AppNewContainer.tsx:56:    push: (location: string) => dispatch(push(location)),
```

dashboard/src/components/ConfigLoader/ConfigLoader.tsx
```diff
@@ -6,9 +6,9 @@ import LoadingWrapper, {
 } from "../../components/LoadingWrapper/LoadingWrapper";
 
 interface IConfigLoaderProps extends ILoadingWrapperProps {
+  children: JSX.Element;
```
andresmgot, line 9: hmm, this didn't work for me when I tried :shrug:

dashboard/src/components/DeploymentFormBody/DeploymentFormBody.tsx
```diff
@@ -148,13 +149,15 @@ class DeploymentFormBody extends React.Component<
...
     // changes the route but in the case of the upgrade it only changes the state
-    const { namespace, selected, deploymentEvent } = this.props;
+    const { cluster, namespace, selected, deploymentEvent } = this.props;
 
     if (deploymentEvent === "upgrade") {
       const { chartID, chartNamespace, getChartVersion } = this.props;
       getChartVersion(chartNamespace, chartID, e.currentTarget.value);
     } else {
-      this.props.push(url.app.apps.new(selected.version!, namespace, e.currentTarget.value));
+      this.props.push(
```
andresmgot, line 158: `getChartVersion` above will need the `cluster` param as well, isn't it?

absoludity: I'm doing the changes systematically in consumable PRs. First going through and making sure all the UI urls correctly use the cluster (rather than just using the "default"), then I'll do the backend calls.

PR #1828, opened 2020-06-26
Fix warnings

andresmgot:
### Description of the change

After changing to `react-scripts`, when running the application, there were a lot of warning. This PR fixes those. There were a bunch of accessibility issues like no having anchors without a `href`. I tried to add those warnings to the CI to fail if they exists but couldn't find a way to do so (I think because we are still using `tslint` when `typescript-eslint` is now recommended.

### DIFF
--- a/dashboard/src/components/AppView/AccessURLTable/AccessURLItem/AccessURLItem.tsx
+++ b/dashboard/src/components/AppView/AccessURLTable/AccessURLItem/AccessURLItem.tsx
@@ -22,7 +22,12 @@ const AccessURLItem: React.SFC<IAccessURLItem> = props => {
             } type-small margin-r-small padding-tiny padding-h-normal`}
           >
             {URLItem.isLink ? (
-              <a className="padding-tiny padding-h-normal" href={l} target="_blank">
+              <a
+                className="padding-tiny padding-h-normal"
+                href={l}
+                target="_blank"
+                rel="noopener noreferrer"
+              >
                 {l}
               </a>
             ) : (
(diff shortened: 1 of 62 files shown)


dashboard/src/components/AppView/AccessURLTable/AccessURLItem/AccessURLItem.tsx
```diff
@@ -22,7 +22,12 @@ const AccessURLItem: React.SFC<IAccessURLItem> = props => {
             } type-small margin-r-small padding-tiny padding-h-normal`}
           >
             {URLItem.isLink ? (
-              <a className="padding-tiny padding-h-normal" href={l} target="_blank">
+              <a
+                className="padding-tiny padding-h-normal"
+                href={l}
```
absoludity, line 27: Worst 1-letter variable eva! :P (already there, I know).

dashboard/src/shared/utils.ts
```diff
@@ -1,5 +1,5 @@
 export function escapeRegExp(str: string) {
-  return str.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, "\\$&");
+  return str.replace(/[-[\]/{}()*+?.\\^$|]/g, "\\$&");
```
absoludity, line 2: Hmm, this change isn't a mechanical one? What was the motivation here - looks like the regex for replacement was escaping the actual regex chars when it didn't need to? I can't produce any difference for different inputs (though this would be much easier to see if we already had a test and showed it didn't change).

absoludity: nm, I see it is just a lint warning for `no-useless-escape`.

andresmgot:
> Looks like it's mostly mechanical, with an exception on the `escapeRegExp` function?

I also had to change a couple of links with buttons because there is one rule that doesn't allow you to use anchors without `href`

PR #1831, opened 2020-06-29
First version for the Header component using Clarity

andresmgot:
### Description of the change

First functional component using Clarity UI. This PR partially implements the Header componenet:

Screenshot from 2020-06-29 18-40-35

For the moment nor the context selector nor the menu button works (it requires, among other things, the changes in #1826 to be able to reuse other components.

I am introducing Clarity Web Components and that requires a react wrapper that I took from https://stackblitz.com/edit/react-ts-wrapper-5w8nhf. This is used only for icons at the moment.

### Applicable issues

  - ref #1354


dashboard/src/components/Clarity/converter/reactWrapper.ts
```diff
@@ -0,0 +1,75 @@
+// Code from: https://stackblitz.com/edit/react-ts-wrapper-5w8nhf
+import React from "react";
+
+export function createReactComponent<BaseComponent extends HTMLElement>(elementName: string) {
+  return class ReactWrapperComponent extends React.Component<
```
absoludity, line 5: Does this (or the future version of `createReactComponent`) force us to use class components? Is there a way to do function components? Perhaps it's possible and would avoid needing to replicate the old class methods?

dashboard/src/components/Header/Header.v2.tsx
```diff
@@ -0,0 +1,99 @@
...
+  logout: () => void;
+  cluster: IClusterState;
+  defaultNamespace: string;
+  push: (path: string) => void;
+  setNamespace: (ns: string) => void;
+  createNamespace: (ns: string) => Promise<boolean>;
+  getNamespace: (ns: string) => void;
+}
+
+class Header extends React.Component<IHeaderProps> {
```
absoludity, line 31: Since this is just a render function, why not a function component here?

dashboard/src/components/Clarity/converter/reactWrapper.ts
andresmgot: it shouldn't, I saw examples of function components with the other wrapper available. I will know more when I migrate the Header component (or other) to a function component.

dashboard/src/components/Header/Header.v2.tsx
andresmgot: I was just waiting for you to merge your PR, I can do that now in a separate PR.

PR #1833, opened 2020-06-30
Update url.app.apps.{get|upgrade} with required cluster arg.

absoludity:
Continues url updates for cluster requirement. Ref #1762 .

Starting to get some duplication of test setup, as the function components which use state need providers etc. I looked at options for creating the shared code, but `react-scripts` doesn't support the jest `setupFiles` option, so it might just be a separate `src/helpers.test.tsx` exporting some functions like `getWrapper(state, component, props)`? What do you think? (not for this PR, but I will cleanup the dups once decided).

The other thing that I had to wrangle here is that once wrapped in a provider, it seems that even our class-based components are no longer classes after wrapping - at least, I had to remove calls to `wrapper.state()` (also tried `wrapper.find(ComponentName).state()`) as jest kept failing saying that `state()` is only available on class-based components.

### DIFF
--- a/dashboard/src/components/AppView/AppControls/AppControls.test.tsx
+++ b/dashboard/src/components/AppView/AppControls/AppControls.test.tsx
@@ -2,16 +2,45 @@ import { mount, shallow } from "enzyme";
 import context from "jest-plugin-context";
 import * as React from "react";
 import Modal from "react-modal";
-import { Redirect } from "react-router";
+import { Provider } from "react-redux";
+import configureMockStore, { MockStore } from "redux-mock-store";
+import thunk from "redux-thunk";
+
+import { IRelease } from "shared/types";
+import { IStoreState } from "shared/types";
 import RollbackButtonContainer from "../../../containers/RollbackButtonContainer";
 import { hapi } from "../../../shared/hapi/release";
 import itBehavesLike from "../../../shared/specs";
 import * as url from "../../../shared/url";
 import ConfirmDialog from "../../ConfirmDialog";
-
-import { IRelease } from "shared/types";
 import AppControls, { IAppControlsProps } from "./AppControls";
 import UpgradeButton from "./UpgradeButton";
+
+const mockStore = configureMockStore([thunk]);
+
+// TODO(absoludity): As we move to function components with (redux) hooks we'll need to
+// be including state in tests, so we may want to put things like initialState
+// and a generalized getWrapper in a test helpers or similar package?
+const initialState = {
+  apps: {},
+  auth: {},
+  catalog: {},
+  charts: {},
+  config: {},
+  kube: {},
+  clusters: {
+    currentCluster: "default-cluster",
+  },
+  repos: {},
+  operators: {},
+} as IStoreState;
+
+const getWrapper = (store: MockStore, props: IAppControlsProps) =>
+  mount(
+    <Provider store={store}>
+      <AppControls {...props} />
+    </Provider>,
+  );
 
 const namespace = "bar";
 const defaultProps = {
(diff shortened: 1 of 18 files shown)


andresmgot:
> Continues url updates for cluster requirement. Ref #1762 .
> 
> Starting to get some duplication of test setup, as the function components which use state need providers etc. I looked at options for creating the shared code, but `react-scripts` doesn't support the jest `setupFiles` option, so it might just be a separate `src/helpers.test.tsx` exporting some functions like `getWrapper(state, component, props)`? What do you think? (not for this PR, but I will cleanup the dups once decided).
>

That's fine. I would use the `shared` folder instead though (to avoid polluting the root folder). Something like `shared/test-helpers.ts`.
 
> The other thing that I had to wrangle here is that once wrapped in a provider, it seems that even our class-based components are no longer classes after wrapping - at least, I had to remove calls to `wrapper.state()` (also tried `wrapper.find(ComponentName).state()`) as jest kept failing saying that `state()` is only available on class-based components.

I guess there should be an analogue way of getting the state?

dashboard/src/components/AppView/AppControls/AppControls.test.tsx
```diff
@@ -2,17 +2,46 @@ import { mount, shallow } from "enzyme";
...
 import { hapi } from "../../../shared/hapi/release";
 import itBehavesLike from "../../../shared/specs";
 import * as url from "../../../shared/url";
 import ConfirmDialog from "../../ConfirmDialog";
-
-import { IRelease } from "shared/types";
 import AppControls, { IAppControlsProps } from "./AppControls";
 import UpgradeButton from "./UpgradeButton";
 
+const mockStore = configureMockStore([thunk]);
```
andresmgot, line 19: I am a bit confused, the `AppControls` component is still a class, why all these changes are needed? Is it because of the change in `react-redux`?

dashboard/src/components/AppView/AppControls/AppControls.test.tsx
```diff
@@ -21,54 +50,55 @@ const defaultProps = {
...
 
-  const confirm = wrapper
-    .find(".AppControls")
-    .children()
-    .find(ConfirmDialog);
+  const confirm = appControls.children().find(ConfirmDialog);
   expect(confirm.exists()).toBe(true);
   confirm.props().onConfirm(); // Simulate confirmation
 
-  expect(wrapper.state("deleting")).toBe(true);
```
andresmgot, line 42: we should block this until we find a way to get the state? We are using the state in several tests, we would lose coverage. What's your plan here?

dashboard/src/components/AppView/AppControls/AppControls.test.tsx
andresmgot: ah, you commented it in the PR:

> it seems that even our class-based components are no longer classes after wrapping - at least, I had to remove calls to wrapper.state()

that's weird...

dashboard/src/components/AppView/AppControls/AppControls.test.tsx
absoludity: Sorry, I forgot to include the info I found: Most people (myself included) recommend *not testing state* (ie. not testing implementation details) but instead validating what's rendered (the effect of the state. See this answer: https://stackoverflow.com/a/58432424 which I'd mostly agree with).

The quick and dirty solution on the same answer can actually be used by us (small tweak because we can't use `shallow` here, so don't have `dive()`. I had tried `appControls.state()` above which didn't work (same error), but what I didn't try, which *does* work, is explicitly `wrapper.find(AppControls).state()`. So I'm updating to that so we have a record of how we can do a quick-n-dirty change, but ultimately we should not be testing implementation details IMO.

andresmgot: Generally I agree but sometimes is much faster to just test the state, is a trade-off.

PR #1835, opened 2020-06-30
Updates url.app.{catalog|repo} to require the cluster arg.

absoludity:
Follows #1833 . Ref #1762 .

Only unexpected here was the lack of ability to use `shallow` when rendering a connected component, and hence unable to do a small targeted snapshot.

A few fly-by fixes for non-typed data.

### DIFF
--- a/dashboard/src/components/AppList/AppList.test.tsx
+++ b/dashboard/src/components/AppList/AppList.test.tsx
@@ -205,6 +205,11 @@ it("filters apps", () => {
 });
 
 it("clicking 'List All' checkbox should trigger toggleListAll", () => {
+  const mockFetchAppsWithUpdateInfo = jest.fn();
+  const updatedProps: IAppListProps = {
+    ...defaultProps,
+    fetchAppsWithUpdateInfo: mockFetchAppsWithUpdateInfo,
+  };
   const apps = {
     isFetching: false,
     items: [],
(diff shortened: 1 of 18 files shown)


dashboard/src/components/AppList/AppList.test.tsx
```diff
@@ -219,20 +220,12 @@ it("clicking 'List All' checkbox should trigger toggleListAll", () => {
...
-      })}
-    />,
-  );
+  const wrapper = shallow(<AppList {...defaultProps} apps={apps} />);
   const checkbox = wrapper.find('input[type="checkbox"]');
   expect(apps.listingAll).toBe(false);
   checkbox.simulate("change");
   // The last call to fetchApps should list all the apps
-  const fetchCalls = defaultProps.fetchAppsWithUpdateInfo.mock.calls;
+  const fetchCalls = (defaultProps.fetchAppsWithUpdateInfo as jest.Mock).mock.calls;
```
andresmgot, line 228: this kind of explicit changes is why I removed the type of the `defaultProps` object. I am not sure if it's worth to do so since actually the inner elements will differ from a `IAppListProps` for example.

absoludity: It's pretty easy to work around while still keeping things typed. I just updated with this change to just have a reference to the item first (and it's cleaner anyway):

```
 it("clicking 'List All' checkbox should trigger toggleListAll", () => {
+  const mockFetchAppsWithUpdateInfo = jest.fn();
+  const updatedProps: IAppListProps = {
+    ...defaultProps,
+    fetchAppsWithUpdateInfo: mockFetchAppsWithUpdateInfo,
+  }
   const apps = {
     isFetching: false,
     items: [],
@@ -220,12 +225,12 @@ it("clicking 'List All' checkbox should trigger toggleListAll", () => {
     ],
     listingAll: false,
   } as IAppState;
-  const wrapper = shallow(<AppList {...defaultProps} apps={apps} />);
+  const wrapper = shallow(<AppList {...updatedProps} apps={apps} />);
   const checkbox = wrapper.find('input[type="checkbox"]');
   expect(apps.listingAll).toBe(false);
   checkbox.simulate("change");
   // The last call to fetchApps should list all the apps
-  const fetchCalls = (defaultProps.fetchAppsWithUpdateInfo as jest.Mock).mock.calls;
+  const fetchCalls = mockFetchAppsWithUpdateInfo.mock.calls;
   expect(fetchCalls[fetchCalls.length - 1]).toEqual(["default", true]);
 });
```

andresmgot: yes but you are still fooling typescript because `defaultProps.fetchAppWithUpdateInfo` will be reported as a `func()...` while it's a `Mock`.

PR #1838, opened 2020-07-01
Update calls to the kubeops service to require cluster.

absoludity:
Follows #1837 Ref #1762 

Updates all calls related to the kubeops backend to require and use the cluster argument.

Last but not least will be the update to the shared backend API.

### DIFF
--- a/cmd/tiller-proxy/main.go
+++ b/cmd/tiller-proxy/main.go
@@ -156,6 +156,7 @@ func main() {
 	}
 
 	// Routes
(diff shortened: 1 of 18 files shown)


andresmgot:
You have some tests failing because I assume the parameter is not available for tiller-proxy.

absoludity:
> You have some tests failing because I assume the parameter is not available for tiller-proxy.

True, updated tiller-proxy with the same aliases. Thanks.

PR #1840, opened 2020-07-01
Clarity components: CardGrid, InfoCard, PageHeader, SearchFilter

andresmgot:
### Description of the change

Started working on the AppList view. This contains several subcomponents:
 - PageHeader
Screenshot from 2020-07-01 18-14-50

 - SearchFilter
Screenshot from 2020-07-01 18-15-13

 - CardGrid
 - InfoCard
Screenshot from 2020-07-06 16-49-11

These are just some of the components included but they are ready to be merged. Will keep adding more of those.

This PR also includes some fixes for the changes related to the `cluster` changes in the URLs.

### Applicable issues

  - ref #1354

### DIFF
--- a/dashboard/src/components/Card/CardGrid.v2.tsx
+++ b/dashboard/src/components/Card/CardGrid.v2.tsx
@@ -1,11 +1,12 @@
 import * as React from "react";
+import Row from "../js/Row";
 
 import "./CardGrid.v2.css";
 
 function CardGrid(props: { children: JSX.Element }) {
   return (
     <div className="card-grid">
-      <div className="clr-row">{props.children}</div>
+      <Row>{props.children}</Row>
     </div>
   );
 }
--- a/dashboard/src/components/Card/CardGrid.v2.tsx
+++ b/dashboard/src/components/Card/CardGrid.v2.tsx
@@ -1,0 +1,13 @@
+import * as React from "react";
+
+import "./CardGrid.v2.css";
+
+function CardGrid(props: { children: JSX.Element }) {
+  return (
+    <div className="card-grid">
+      <div className="clr-row">{props.children}</div>
+    </div>
+  );
+}
+
+export default CardGrid;
--- a/dashboard/src/components/InfoCard/InfoCard.v2.tsx
+++ b/dashboard/src/components/InfoCard/InfoCard.v2.tsx
@@ -1,0 +1,84 @@
+import { ClarityIcons, infoCircleIcon } from "@clr/core/icon-shapes";
+import * as React from "react";
+import { Link } from "react-router-dom";
+import { CdsIcon } from "../Clarity/clarity";
+
+import placeholder from "../../placeholder.png";
+import "./InfoCard.v2.css";
+
+ClarityIcons.addIcons(infoCircleIcon);
+
+export interface IInfoCardProps {
+  title: string;
+  info: string | JSX.Element;
+  link?: string;
+  icon?: string;
+  banner?: string;
+  subIcon?: string;
+  description?: string | JSX.Element;
+  tag1Class?: string;
+  tag1Content?: string | JSX.Element;
+  tag2Class?: string;
+  tag2Content?: string | JSX.Element;
+}
+
+function InfoCard(props: IInfoCardProps) {
+  const {
+    title,
+    link,
+    info,
+    description,
+    tag1Content,
+    tag1Class,
+    tag2Content,
+    tag2Class,
+    banner,
+    subIcon,
+  } = props;
+  const icon = props.icon ? props.icon : placeholder;
+  return (
+    <div className="clr-col-lg-3 clr-col-12">
+      <Link to={link || "#"} className="card clickable">
+        <div className="card-header">{title}</div>
+        <div className="card-block">
+          <div className="card-media-block">
+            <img src={icon} className="card-media-image" alt="icon" />
+            <div className="card-media-description">
+              <span className="card-media-text">{description}</span>
+            </div>
+          </div>
+          {banner && (
+            <div className="alert alert-info alert-sm" role="alert">
+              <div className="alert-items">
+                <div className="alert-item static">
+                  <div className="alert-icon-wrapper">
+                    <CdsIcon className="alert-icon" shape="info-circle" />
+                  </div>
+                  <span className="alert-text">{banner}</span>
+                </div>
+              </div>
+            </div>
+          )}
+        </div>
+        <div className="card-footer">
+          <div className="clr-row">
+            <div className="clr-col-6" style={{ padding: 0 }}>
+              {info}
+            </div>
+            <div className="clr-col-6 label-section">
+              {tag1Content && (
+                <span className={`label ${tag1Class || "label-info"}`}>{tag1Content}</span>
+              )}
+              {tag2Content && (
+                <span className={`label ${tag2Class || "label-info"}`}>{tag2Content}</span>
+              )}
+              {subIcon && <img src={subIcon} alt="icon" />}
+            </div>
+          </div>
+        </div>
+      </Link>
+    </div>
+  );
+}
+
+export default InfoCard;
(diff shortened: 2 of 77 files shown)


dashboard/src/components/Card/CardGrid.v2.tsx
```diff
@@ -0,0 +1,13 @@
+import * as React from "react";
+
+import "./CardGrid.v2.css";
+
+function CardGrid(props: { children: JSX.Element }) {
+  return (
+    <div className="card-grid">
+      <div className="clr-row">{props.children}</div>
```
Angelmmiguel, line 8: You can reuse the `Row` internal component here

dashboard/src/components/InfoCard/InfoCard.v2.tsx
```diff
@@ -0,0 +1,84 @@
...
+  info: string | JSX.Element;
+  link?: string;
+  icon?: string;
+  banner?: string;
+  subIcon?: string;
+  description?: string | JSX.Element;
+  tag1Class?: string;
+  tag1Content?: string | JSX.Element;
+  tag2Class?: string;
+  tag2Content?: string | JSX.Element;
```
Angelmmiguel, line 22: Can we make these props more explicit? `tag1` and `tag2` may be confusing about the content they contain:

```
categoryClass:
category:
statusClass:
status:
```

Also, you can define a `StatusTag` and `CategoryTag` component that sets the colors automatically. In this way, the component that calls `InfoCard` doesn't need to know the CSS class for the tags.

dashboard/src/components/InfoCard/InfoCard.v2.tsx
```diff
@@ -0,0 +1,84 @@
...
+    tag1Content,
+    tag1Class,
+    tag2Content,
+    tag2Class,
+    banner,
+    subIcon,
+  } = props;
+  const icon = props.icon ? props.icon : placeholder;
+  return (
+    <div className="clr-col-lg-3 clr-col-12">
```
Angelmmiguel, line 40: You can reuse the internal `Column` component

dashboard/src/components/InfoCard/InfoCard.v2.tsx
```diff
@@ -0,0 +1,84 @@
...
+      <Link to={link || "#"} className="card clickable">
+        <div className="card-header">{title}</div>
+        <div className="card-block">
```
Angelmmiguel, line 43: You can use the internal `Card` component. It contains several subcomponents such as `CardHeader` and `CardBlock`.

dashboard/src/components/InfoCard/InfoCard.v2.tsx
```diff
@@ -0,0 +1,84 @@
...
+              <div className="alert-items">
+                <div className="alert-item static">
+                  <div className="alert-icon-wrapper">
+                    <CdsIcon className="alert-icon" shape="info-circle" />
+                  </div>
+                  <span className="alert-text">{banner}</span>
+                </div>
+              </div>
+            </div>
+          )}
```
Angelmmiguel, line 61: You can reuse the internal `Alert` component

dashboard/src/components/InfoCard/InfoCard.v2.tsx
```diff
@@ -0,0 +1,84 @@
...
+            <div className="clr-col-6 label-section">
+              {tag1Content && (
+                <span className={`label ${tag1Class || "label-info"}`}>{tag1Content}</span>
+              )}
+              {tag2Content && (
+                <span className={`label ${tag2Class || "label-info"}`}>{tag2Content}</span>
+              )}
+              {subIcon && <img src={subIcon} alt="icon" />}
+            </div>
+          </div>
```
Angelmmiguel, line 77: Same about `Row` and `Column`

dashboard/src/components/Layout/UISelector/Clarity.scss
```diff
@@ -1,6 +1,5 @@
-@import "node_modules/@clr/core/styles/global.min.css";
-@import "node_modules/@clr/city/css/bundles/default.min.css";
-
-body {
-  font-family: "Clarity City", "Avenir Next", "Helvetica Neue", Arial, sans-serif;
+.clr-row {
+  // By default, a row has negative margin. This causes
+  // that a row for the whole body surpasses its limits
+  margin: 0;
```
Angelmmiguel, line 4: This negative margin is required by the column class. You should always use `Column` inside a `Row`. Removing this margin can cause issues in the Clarity grid system.

dashboard/src/components/PageHeader/PageHeader.v2.tsx
```diff
@@ -0,0 +1,13 @@
+import * as React from "react";
+
+import "./PageHeader.v2.css";
+
+interface IPageHeaderProps {
+  children: Array<boolean | JSX.Element> | JSX.Element;
+}
+
+function PageHeader(props: IPageHeaderProps) {
+  return <header className="kubeapps-header clr-row">{props.children}</header>;
```
Angelmmiguel, line 10: For this, I'd keep `header` as a separate HTML tags:

```suggestion
  return <header className="kubeapps-header">
    <Row>
      {props.children}
    </Row>
  </header>;
```

Rows should be used as independent items to avoid collisions between the styles.

dashboard/src/components/SearchFilter/SearchFilter.v2.tsx
```diff
@@ -0,0 +1,43 @@
...
+        <CdsIcon size="sm" shape="search" />
+        <input
+          id="search"
+          type="text"
+          className="padding-l-bigger"
+          placeholder={props.placeholder}
+          autoComplete="off"
+          onChange={handleChange}
+          value={props.value}
+        />
```
Angelmmiguel, line 37: You can use the internal `Input` component.

dashboard/src/components/SearchFilter/SearchFilter.v2.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    props.onChange(e.currentTarget.value);
+  };
+  const handleSubmit = (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+    props.onSubmit(props.value);
+  };
+  return (
+    <div className="search-box">
+      <form onSubmit={handleSubmit}>
```
Angelmmiguel, line 27: Include the `aria-label` and `role` for accessibility:

```suggestion
      <form onSubmit={handleSubmit} role="search" aria-label="Search on the site">
```

With these changes, screen readers will notify the users this is a search box.

dashboard/src/components/js/Alert/Alert.js
```diff
@@ -0,0 +1,109 @@
...
+import cs from "classnames";
+import {
+  ClarityIcons,
+  infoCircleIcon,
+  exclamationCircleIcon,
+  checkCircleIcon,
+  exclamationTriangleIcon,
+  timesIcon,
+} from "@clr/core/icon-shapes";
+import { CdsIcon } from "../../Clarity/clarity";
```
andresmgot, line 14: cc/ @Angelmmiguel note the new way to import icons. The react wrapper has been extracted (`../../Clarity/clarity`) from https://stackblitz.com/edit/react-ts-wrapper-5w8nhf. I'd assume that there is a similar wrapper for plain JS?

dashboard/src/components/InfoCard/InfoCard.v2.tsx
Angelmmiguel: mmm still think it's not trivial to understand the purpose, but it's not a blocker for this PR :+1:

dashboard/src/components/js/Alert/Alert.js
Angelmmiguel: Yeah, the approach is pretty similar, although they provide you some helpers to install a certain group of icons directly.

I'm curious about how importing the same icon in two different components affects to the library. Not sure about this use case since we load all the icons in a single file internally.

PR #1841, opened 2020-07-01
Fix Header v2 compilation

andresmgot:
### Description of the change

Fix build in master


andresmgot:
Thanks for merging it.

> This must have landed before you approved and landed the build changes?

I am not sure what you mean? Yes, this should've been done before merging the change in the `Header.v2` component the problem is that we created the "conflicting" PRs at the same time (and I didn't notice).

PR #1842, opened 2020-07-02
Clarity Login form

andresmgot:
### Description of the change

Login page:

Screenshot from 2020-07-02 13-53-00

In the case OIDC is enabled:

Screenshot from 2020-07-02 13-55-50

### Applicable issues

  - ref #1354 

### Additional information

The test and the component logic has been just adapted from the existing LoginForm component.

### DIFF
--- a/dashboard/src/components/LoginForm/LoginForm.v2.tsx
+++ b/dashboard/src/components/LoginForm/LoginForm.v2.tsx
@@ -12,13 +12,14 @@ import TokenLogin from "./TokenLogin";
 
 ClarityIcons.addIcons(infoCircleIcon);
 
-interface ILoginFormProps {
+export interface ILoginFormProps {
   authenticated: boolean;
   authenticating: boolean;
   authenticationError: string | undefined;
   oauthLoginURI: string;
   authenticate: (token: string) => any;
   checkCookieAuthentication: () => void;
+  appVersion: string;
   location: Location;
 }
 
@@ -65,7 +66,7 @@ function LoginForm(props: ILoginFormProps) {
         )}
         <div className="login-moreinfo">
           <a
-            href="https://github.com/kubeapps/kubeapps/blob/master/docs/user/access-control.md"
+            href={`https://github.com/kubeapps/kubeapps/blob/${props.appVersion}/docs/user/access-control.md`}
             target="_blank"
             rel="noopener noreferrer"
           >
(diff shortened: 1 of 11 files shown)


dashboard/src/components/LoginForm/LoginForm.v2.tsx
```diff
@@ -0,0 +1,145 @@
...
+  private oauthLogin = () => {
+    return (
+      <>
+        <section className="title">
+          <h3 className="welcome">Welcome to</h3>
+          Kubeapps
+          <h5 className="hint">
+            Your cluster operator has enabled login via an authentication provider.
+          </h5>
+        </section>
```
Angelmmiguel, line 83: `section` blocks should be used to identify a block of a website that acts as a single unit. In this case, the "title" doesn't represent a block, but the full login block does.

Please, update the `section` to wrap all this block instead only the title. 

In addition to that, ensure you create a title that represents the block itself. For a screen reader, the current heading title (`Welcome to`) doesn't provide enough information. `Kubeapps` string should be part of this title too. You can use a `span` tag to change the style of the `Kubeapps` string if it's required. 

Regarding the hint, a heading is not an appropriate HTML tag for it. You can use a paragraph for this.

Once the previous comments are done, you should link the section with both title and hint sections. For this, give an ID to them and link them using the aria properties:

```
<section aria-labelledby="login-title" aria-describedby="login-desc">
  <h3 id="login-title">
    Welcome to
    <span>Kubeapps</span>
  </h3>
  <p id="login-desc">
    ...
  </p>
  <div>
    ... Login form ...
  </div>
</section>
```

dashboard/src/components/LoginForm/LoginForm.v2.tsx
```diff
@@ -0,0 +1,145 @@
...
+  private tokenLogin = () => {
+    return (
+      <>
+        <section className="title">
+          <h3 className="welcome">Welcome to</h3>
+          Kubeapps
+          <h5 className="hint">
+            Your cluster operator should provide you with a Kubernetes API token.
+          </h5>
+        </section>
```
Angelmmiguel, line 110: Same here :)

dashboard/src/components/LoginForm/LoginForm.v2.tsx
Angelmmiguel: There are still some work about accessibility in the Clarity docs. The approach I suggested is more appropriate for this case.

dashboard/src/components/LoginForm/LoginForm.v2.tsx
```diff
@@ -0,0 +1,82 @@
...
+            authenticationError={props.authenticationError}
+            oauthLoginURI={props.oauthLoginURI}
+          />
+        ) : (
+          <TokenLogin
+            authenticationError={props.authenticationError}
+            token={token}
+            handleTokenChange={handleTokenChange}
+          />
+        )}
```
Angelmmiguel, line 66: Much better! :+1:

PR #1843, opened 2020-07-02
Update backend url for namespaces to require cluster.

absoludity:
Updates `url.backend.namespaces.list()` to require the cluster arg, and all affected call-sites/dependencies.

Ref #1762

### DIFF
--- a/dashboard/src/shared/Namespace.ts
+++ b/dashboard/src/shared/Namespace.ts
@@ -1,31 +1,38 @@
 import { axiosWithAuth } from "./AxiosInstance";
-import { APIBase } from "./Kube";
 import * as url from "./url";
 
 import { ForbiddenError, IResource, NotFoundError } from "./types";
 
 export default class Namespace {
-  public static async list() {
+  public static async list(cluster: string) {
+    // This call is hitting an actual backend endpoint (see pkg/http-handler.go)
+    // while the other calls (create, get) are hitting the k8s API via the
+    // frountend nginx.
     const { data } = await axiosWithAuth.get<{ namespaces: IResource[] }>(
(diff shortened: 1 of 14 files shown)


dashboard/src/shared/Namespace.ts
```diff
@@ -1,31 +1,38 @@
...
 
-  public static async create(name: string) {
-    const { data } = await axiosWithAuth.post<IResource>(`${APIBase}/api/v1/namespaces/`, {
-      apiVersion: "v1",
-      kind: "Namespace",
-      metadata: {
-        name,
+  public static async create(cluster: string, name: string) {
+    const { data } = await axiosWithAuth.post<IResource>(
+      `api/clusters/${cluster}/api/v1/namespaces/`,
```
andresmgot, line 19: this URL should be in the `url` lib

dashboard/src/shared/Namespace.ts
```diff
@@ -1,31 +1,38 @@
...
+    );
     return data;
   }
 
-  public static async get(name: string) {
+  public static async get(cluster: string, name: string) {
     try {
-      const { data } = await axiosWithAuth.get<IResource>(`${APIBase}/api/v1/namespaces/${name}`);
+      const { data } = await axiosWithAuth.get<IResource>(
+        `api/clusters/${cluster}/api/v1/namespaces/${name}`,
```
andresmgot, line 34: same

dashboard/src/shared/Namespace.ts
absoludity: I've done this as part of a larger change to improve the direct calls to the k8s api from the dashboard, so they are easily identifiable, as part of #1877 .

PR #1844, opened 2020-07-02
1762 add select cluster component

absoludity:
The intent was to add the select cluster component and then test the multi-cluster integration.

As expected, there were quite a few adjustments needed to demo end-to-end. This PR isn't to review/land (lots of TODOs, insecure comms etc. which I did to demo rather than debugging), but just as a record for me to work from when I return in 2 weeks :)

Note to self: Instead of updating the IResource with a `cluster` attribute, add it to the `ResourceRef` (like we do with the release namespace).

jfyi @andresmgot (I'll double check that there's no bug-fixes here which need to land before a release).

Ref #1762

### DIFF
--- a/cmd/kubeops/internal/handler/handler.go
+++ b/cmd/kubeops/internal/handler/handler.go
@@ -22,7 +22,7 @@ import (
 
 const (
 	authHeader     = "Authorization"
-	clusterParam   = "clusters"
+	clusterParam   = "cluster"
 	namespaceParam = "namespace"
 	nameParam      = "releaseName"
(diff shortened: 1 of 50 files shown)


cmd/kubeops/internal/handler/handler.go
```diff
@@ -22,7 +22,7 @@ import (
 
 const (
 	authHeader     = "Authorization"
-	clusterParam   = "clusters"
+	clusterParam   = "cluster"
```
absoludity, line 25: This bug doesn't break the non-multicluster situation as below we default to the default cluster if the param is empty for backwards compat. at the moment.

Issue #1839, opened 2020-07-01
Authentication header bypassing-authentication


batiati:
### Description:

I'm trying to bypass the login form injecting the authentication header on the NGINX ingress, following the instructions as stated on this doc:

https://github.com/kubeapps/kubeapps/blob/master/docs/user/login-alternatives.md#bypassing-authentication

![image](https://user-images.githubusercontent.com/5728002/86246304-28ecbc00-bb81-11ea-8605-da4600e0f3fe.png)

But it still showing the Login form requesting the token.
I'm not using any OAuth2/OIDC provider, I just wanted to publish an internal demo without the need of authenticate.

From the ingress log, it seems that the `add_header` directive needs an additional argument:

```
Error: exit status 1
2020/07/01 11:50:04 [emerg] 56664#56664: invalid number of arguments in "add_header" directive in /tmp/nginx-cfg745122493:760
nginx: [emerg] invalid number of arguments in "add_header" directive in /tmp/nginx-cfg745122493:760
nginx: configuration file /tmp/nginx-cfg745122493 test failed
```
I tried to fix it putting the [always parameter ](http://nginx.org/en/docs/http/ngx_http_headers_module.html#add_header), which solved the problem on the NGINX ingress log, but had no effect on kubeapps login form.

### Version of Helm, Kubeapps and Kubernetes:

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.2.0", GitCommit:"e11b7ce3b12db2941e90399e874513fbd24bcb71", GitTreeState:"clean", GoVersion:"go1.13.10"}
```

- Output of `helm list <kubeapps-release-name>`:

```
NAME            NAMESPACE       REVISION        UPDATED                                 STATUS          CHART           APP VERSION
kubeapps        kubeapps        2               2020-06-29 20:57:44.7007997 -0300 -03   deployed        kubeapps-3.7.2  v1.10.2
```

- Output of `kubectl version`:

```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.8", GitCommit:"ec6eb119b81be488b030e849b9e64fda4caaf33c", GitTreeState:"clean", BuildDate:"2020-03-12T21:00:06Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #1846, opened 2020-07-03
Update login-alternatives.md

andresmgot:
### Description of the change

The "bypassing authentication" method no longer works as explained. Rather than updating those instructions I believe it's better to not support that approach for security reasons. Also I updated the docs  to make it clear  that the preferred way of authenticating is using OIDC.

### Applicable issues

  - ref https://github.com/kubeapps/kubeapps/issues/1839

### DIFF
--- a/docs/user/login-alternatives.md
+++ b/docs/user/login-alternatives.md
@@ -1,33 +1,11 @@
 # Kubeapps Login
 
-By default, when first visiting Kubeapps, a login form is shown for the user to introduce a Kubernetes API token:
+By default, Kubeapps allow you to login using a Kubernetes service account token. While this is enabled by default to make it easier to get started, the **recommended way of authenticate in Kubeapps is using an OAuth2/OIDC provider**. Depending on your cluster provider, the instructions may differ. Check this [guide](./using-an-OIDC-provider.md) for more information.
+
+When using the service account authentication, the login form is shown for the user to introduce a Kubernetes API token:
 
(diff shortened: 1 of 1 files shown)


docs/user/login-alternatives.md
```diff
@@ -1,33 +1,11 @@
...
-By default, when first visiting Kubeapps, a login form is shown for the user to introduce a Kubernetes API token:
+By default, Kubeapps allow you to login using a Kubernetes service account token. While this is enabled by default to make it easier to get started, the **recommended way of authenticate in Kubeapps is using an OAuth2/OIDC provider**. Depending on your cluster provider, the instructions may differ. Check this [guide](./using-an-OIDC-provider.md) for more information.
+
+When using the service account authentication, the login form is shown for the user to introduce a Kubernetes API token:
 
 ![Dashboard Login](../img/dashboard-login.png)
 
 The goal of the login form is to identify the user and associate it with a Kubernetes service account. This identity information will be used by Kubeapps to authenticate the user against the Kubernetes API. You can find more information about access control in Kubeapps in this [document](./access-control.md).
 
 However, it's possible to disable the form or delegate the authentication to an OAuth2/OIDC provider so users don't need to introduce a token in the login form.
```
batiati, line 11: I think it's better to remove the part `... it's possible to disable the form ... `  too.

andresmgot: Good catch, thanks @batiati I missed that.

PR #1849, opened 2020-07-07
LoadingWrapper Clarity Component

andresmgot:
### Description of the change

Update for the LoadingWrapper to use the Spinner based on Clarity UI.

### Applicable issues

  - ref #1354

### DIFF
--- a/dashboard/src/components/LoadingWrapper/LoadingWrapper.v2.tsx
+++ b/dashboard/src/components/LoadingWrapper/LoadingWrapper.v2.tsx
@@ -4,16 +4,13 @@ import Spinner from "../js/Spinner";
 
 export interface ILoadingWrapperProps {
   loaded?: boolean;
-  size?: string;
+  small?: boolean;
+  medium?: boolean;
   children?: any;
 }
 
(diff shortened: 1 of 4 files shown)


dashboard/src/components/LoadingWrapper/LoadingWrapper.v2.tsx
```diff
@@ -0,0 +1,19 @@
...
+  loaded?: boolean;
+  size?: string;
+  children?: any;
+}
+
+function LoadingWrapper(props: ILoadingWrapperProps) {
+  return props.loaded ? (
+    props.children
+  ) : (
+    <Spinner medium={props.size === "medium"} small={props.size === "small"} />
```
Angelmmiguel, line 15: You can keep `medium` and `small` props. It's more readable when using them:

```
<LoadingWrapper medium />
```

In this way, you don't need to go to the implementation to see the available sizes in the JSX.

andresmgot: was thinking on maintaining the current API but yes, it's easier to set the size that way

Angelmmiguel:
Still LGTM!

PR #1851, opened 2020-07-07
AppList v2

andresmgot:
### Description of the change

Rest of components of the AppList view: 

 - AppList
 - AppListGrid
 - AppListItem
 - CustomResourceListItem

Screenshot from 2020-07-07 17-59-55

(The header button in the top-right corner is a bit misplaced, will fix in a different PR).

### Applicable issues

  - ref #1354

### DIFF
--- a/dashboard/src/components/AppList/AppList.v2.test.tsx
+++ b/dashboard/src/components/AppList/AppList.v2.test.tsx
@@ -65,11 +65,6 @@ context("while fetching apps", () => {
     props: { ...defaultProps, isFetchingResources: true },
   });
 
-  it("matches the snapshot", () => {
-    const wrapper = shallow(<AppList {...props} />);
-    expect(wrapper).toMatchSnapshot();
-  });
-
   it("renders a Application header", () => {
     const wrapper = shallow(<AppList {...props} />);
     expect(wrapper.find("h1").text()).toContain("Applications");
--- a/dashboard/src/components/AppList/AppList.v2.test.tsx
+++ b/dashboard/src/components/AppList/AppList.v2.test.tsx
@@ -1,0 +1,260 @@
+import { mount, shallow } from "enzyme";
+import context from "jest-plugin-context";
+import * as React from "react";
+
+import SearchFilter from "components/SearchFilter/SearchFilter.v2";
+import { BrowserRouter as Router } from "react-router-dom";
+import itBehavesLike from "../../shared/specs";
+import { IAppOverview, IAppState } from "../../shared/types";
+import Alert from "../js/Alert";
+import AppList, { IAppListProps } from "./AppList.v2";
+import AppListItem from "./AppListItem.v2";
+import CustomResourceListItem from "./CustomResourceListItem.v2";
+
+let props = {} as any;
+
+const defaultProps: IAppListProps = {
+  apps: {} as IAppState,
+  filter: "",
+  namespace: "default",
+  cluster: "defaultc",
+  pushSearchFilter: jest.fn(),
+  fetchAppsWithUpdateInfo: jest.fn(),
+  getCustomResources: jest.fn(),
+  isFetchingResources: false,
+  customResources: [],
+  csvs: [],
+  featureFlags: { operators: true, additionalClusters: [], ui: "hex" },
+  appVersion: "v1.0.0",
+};
+
+context("when changing props", () => {
+  it("should fetch apps in the new namespace", async () => {
+    const fetchAppsWithUpdateInfo = jest.fn();
+    const getCustomResources = jest.fn();
+    mount(
+      <Router>
+        <AppList
+          {...defaultProps}
+          fetchAppsWithUpdateInfo={fetchAppsWithUpdateInfo}
+          getCustomResources={getCustomResources}
+          namespace="foo"
+        />
+      </Router>,
+    );
+    expect(fetchAppsWithUpdateInfo).toHaveBeenCalledWith("foo", true);
+    expect(getCustomResources).toHaveBeenCalledWith("foo");
+  });
+
+  it("should update the filter", () => {
+    const wrapper = mount(
+      <Router>
+        <AppList {...defaultProps} filter="foo" />
+      </Router>,
+    );
+    expect(wrapper.find(SearchFilter).prop("value")).toEqual("foo");
+  });
+});
+
+context("while fetching apps", () => {
+  props = { ...defaultProps, apps: { isFetching: true } };
+
+  itBehavesLike("aLoadingComponent", { component: AppList, props });
+  itBehavesLike("aLoadingComponent", {
+    component: AppList,
+    props: { ...defaultProps, isFetchingResources: true },
+  });
+
+  it("matches the snapshot", () => {
+    const wrapper = shallow(<AppList {...props} />);
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it("renders a Application header", () => {
+    const wrapper = shallow(<AppList {...props} />);
+    expect(wrapper.find("h1").text()).toContain("Applications");
+  });
+
+  it("shows the search filter and deploy button", () => {
+    const wrapper = shallow(<AppList {...props} />);
+    expect(wrapper.find("SearchFilter")).toExist();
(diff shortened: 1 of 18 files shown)


dashboard/src/components/AppList/AppList.v2.test.tsx
```diff
@@ -0,0 +1,260 @@
...
+
+  itBehavesLike("aLoadingComponent", { component: AppList, props });
+  itBehavesLike("aLoadingComponent", {
+    component: AppList,
+    props: { ...defaultProps, isFetchingResources: true },
+  });
+
+  it("matches the snapshot", () => {
+    const wrapper = shallow(<AppList {...props} />);
+    expect(wrapper).toMatchSnapshot();
```
Angelmmiguel, line 70: I usually use `toMatchSnapshot` only with small portions of the components to avoid failing tests due to minimal changes. Just a comment on this :)

dashboard/src/components/AppList/AppList.v2.tsx
```diff
@@ -0,0 +1,102 @@
...
+  getCustomResources: (ns: string) => void;
+  customResources: IResource[];
+  isFetchingResources: boolean;
+  csvs: IClusterServiceVersion[];
+  featureFlags: IFeatureFlags;
+  appVersion: string;
+}
+
+function AppList(props: IAppListProps) {
+  const [filter, setFilter] = useState("");
```
Angelmmiguel, line 35: Glad to see we're moving the codebase to hooks!

dashboard/src/components/AppList/AppList.v2.tsx
```diff
@@ -0,0 +1,102 @@
...
+    pushSearchFilter,
+    customResources,
+    appVersion,
+    csvs,
+  } = props;
+
+  useEffect(() => {
+    fetchAppsWithUpdateInfo(namespace, true);
+    getCustomResources(namespace);
+    setFilter(filterProps);
```
Angelmmiguel, line 53: Are these props something that may change outside of this component? If this is not going to change, you can initialize the state directly. I see `filter` is the one used on the code.

```
  const [filter, setFilter] = useState(props.filter);
```

In addition to this, React recommends you to use an `useEffect` block per behavior. For example, I see here two different blocks:

1. Fetching resources
2. Setting filters

You can split it into two blocks:

```
useEffect(() => {
  fetchAppsWithUpdateInfo(namespace, true);
  getCustomResources(namespace);
}, [namespace, fetchAppsWithUpdateInfo, getCustomResources]);


useEffect(() => {
  setFilter(filterProps);
}, [filterProps]);
```

The purpose of this division is to avoid complex "side effects" that may affect the performance

dashboard/src/components/AppList/AppList.v2.tsx
```diff
@@ -0,0 +1,102 @@
...
+            </>
+          </Row>
+        </Column>
+        <Column span={2}>
+          <div className="header-button">
+            <Link to={url.app.catalog(cluster, namespace)}>
+              <CdsButton status="primary">Deploy</CdsButton>
+            </Link>
+          </div>
+        </Column>
```
Angelmmiguel, line 79: From this component POV, this structure is not correct. The `Row` component must be placed as a parent of `Column`. If `PageHeader` already defines a `Row`, I'd remove it from there and left the components to define it so it's more explicit. 

Also, the fragment (`<></>`) is not required at this point.

dashboard/src/components/AppList/AppList.v2.tsx
```diff
@@ -0,0 +1,102 @@
...
+      </PageHeader>
+      <main>
```
Angelmmiguel, line 81: Is `main` exposed in part of the app? `main` is a landmark to notify screen readers and other software about the main content of the page. We must keep only one `main` HTML tag per page

dashboard/src/components/AppList/AppListGrid.tsx
```diff
@@ -0,0 +1,70 @@
...
+}
+
+function AppListGrid(props: IAppListProps) {
+  const { appList, customResources, cluster, namespace, appVersion, filter } = props;
+  const filteredReleases = (appList || []).filter(a =>
+    new RegExp(escapeRegExp(filter), "i").test(a.releaseName),
+  );
+  const filteredCRs = customResources.filter(cr =>
+    new RegExp(escapeRegExp(filter), "i").test(cr.metadata.name),
+  );
```
Angelmmiguel, line 31: Just for keeping components simpler, I use to move this kind of functions into a `component/utils/filters.tsx` folder.

dashboard/src/components/AppList/AppListItem.v2.tsx
```diff
@@ -0,0 +1,59 @@
...
+        <div className="clr-row">
+          <div className={`clr-col-2 info-icon ${updateAvailable ? "is-success" : ""}`}>
+            <CdsIcon shape={updateAvailable ? "circle-arrow" : "bundle"} />
+          </div>
+          <div className="clr-col-10">
+            <span>App: {app.chartMetadata.appVersion}</span>
+            <br />
+            <span>Chart: {app.chartMetadata.version}</span>
+          </div>
+        </div>
```
Angelmmiguel, line 48: Here you can use `Row` and `Column`. In addition, I'd add an `aria-label` to mark this content as a group and notify screen readers about what they are going to read here:

```suggestion
        <div className="clr-row" aria-label="Chart version information">
          <div className={`clr-col-2 info-icon ${updateAvailable ? "is-success" : ""}`}>
            <CdsIcon shape={updateAvailable ? "circle-arrow" : "bundle"} />
          </div>
          <div className="clr-col-10">
            <span>App: {app.chartMetadata.appVersion}</span>
            <br />
            <span>Chart: {app.chartMetadata.version}</span>
          </div>
        </div>
```

dashboard/src/components/AppList/AppList.v2.test.tsx
andresmgot: yep, this test was copied from the one existing but it's pretty useless.

dashboard/src/components/AppList/AppList.v2.tsx
andresmgot: > Are these props something that may change outside of this component?

Yep, it can be changed changing URL.

> You can split it into two blocks:

Nice, didn't know about that :+1:

dashboard/src/components/AppList/AppList.v2.tsx
andresmgot: yep, this second `main` shouldn't be here

dashboard/src/components/AppList/AppList.v2.tsx
```diff
@@ -50,51 +50,54 @@ function AppList(props: IAppListProps) {
...
+              <Row>
+                <h1>Applications</h1>
+                <SearchFilter
+                  key="searchFilter"
+                  placeholder="search apps..."
+                  onChange={setFilter}
+                  value={filter}
+                  onSubmit={pushSearchFilter}
+                />
+              </Row>
```
Angelmmiguel, line 74: I think this `Row` is not required

dashboard/src/components/AppList/AppListGrid.tsx
andresmgot: That's just hiding things under the carpet? I prefer not having the extra step if the filter is not going to be shared among different components.

dashboard/src/components/AppList/AppList.v2.tsx
andresmgot: It is to avoid having the `h1` and the filter in different lines.

dashboard/src/components/AppList/AppListGrid.tsx
Angelmmiguel: I always try to have components as simple as I can (in terms of code). The idea is to separate concerns and responsibilities. For components, they should:

1. Initialize based on properties
1. Run all the hooks to manage the state
1. Prepare stuff to be represented
1. JSX code

For this reason, I like to avoid having extra methods inside the components, but this is no a blocker :)

Issue #1832, opened 2020-06-29
Deleting a namespaced app repo does not delete cronjob nor assets


absoludity:
### Description:

Deleting a namespaced app repo leaves the sync cronjob in place and so assets are still returned (but unable to be deployed) after deletion.

### Steps to reproduce the issue:

1. Install `my-repo` in the default namespace using `https://charts.bitnami.com/bitnami`
2. Verify the cronjob is installed, sync performed, and results appear in the catalog
3. Delete the `my-repo` repo in the default namespace

### Describe the results you received:

Results are still shown for the my-repo repository in the catalog, but cannot be deployed.
The cronjob is still present in the kubeapps namespace for syncing the data.

### Describe the results you expected:

Results should no longer appear, nor should the sync cronjob still be present.

### Version of Helm, Kubeapps and Kubernetes:
- Output of `helm version`:

```
helm version
version.BuildInfo{Version:"v3.2.0", GitCommit:"e11b7ce3b12db2941e90399e874513fbd24bcb71", GitTreeState:"clean", GoVersion:"go1.13.10"}```
```
- Output of `helm list <kubeapps-release-name>`:

```
helm list -n kubeapps
NAME            NAMESPACE       REVISION        UPDATED                                         STATUS          CHART                 APP VERSION
chartmuseum     kubeapps        1               2020-06-24 16:18:45.924973082 +1000 AEST        deployed        chartmuseum-2.13.0    0.12.0     
kubeapps        kubeapps        1               2020-06-24 15:57:01.541583472 +1000 AEST        deployed        kubeapps-3.7.2        v1.10.2    
```

BUT note running a development dashboard image.

- Output of `kubectl version`:

```
k version
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"clean", BuildDate:"2020-04-16T11:56:40Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"clean", BuildDate:"2020-04-30T20:19:45Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}

```

andresmgot:
[...]

PR #1852, opened 2020-07-07
Added namespace field to intialRepos which changes the namespace of t…

rarick:
…he created AppRepository

### Description of the change

Added optional namespace field to `apprepository.intialRepos` to allow users to specify the namespace of the apprepository.

### Benefits

Currently RBAC prevents apprepositories from being read across namespaces. However, there is currently no way to create intial apprepositories via the values.yaml in a namespace other than default/kubeapps. This will allow for users to create these apprepositories on startup in a specific namespace.

### Possible drawbacks

Things that need to be addressed before this PR goes in (that I'm not entirely sure how to do):
- Update the version number
- It appears that when an apprepository is created in another namespace, it's not cleaned up by `helm uninstall ...`. I'm not entirely familiar with helm, so I could use help on this
- I'm not acquainted with the codebase, so I'd like a maintainer to verify that this namespace change isn't breaking anything I'm unaware of

### Applicable issues

  - fixes #

### Additional information

Please feel free to reach out to me, I'll be as responsive as possible. I'm also in the #kubeapps channel of the kubernetes slack.

### DIFF
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -240,7 +240,7 @@ apprepository:
   #   # It should start with -----BEGIN CERTIFICATE----- or
   #   # -----BEGIN RSA PRIVATE KEY-----
   #   caCert:
-  #   # Create this apprepository in some other namespace
+  #   # Create this apprepository in a custom namespace
   #   namespace:
   # https://github.com/kubeapps/kubeapps/issues/478#issuecomment-422979262
   ## AppRepository Controller containers' resource requests and limits
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -240,6 +240,8 @@ apprepository:
   #   # It should start with -----BEGIN CERTIFICATE----- or
   #   # -----BEGIN RSA PRIVATE KEY-----
   #   caCert:
+  #   # Create this apprepository in some other namespace
+  #   namespace:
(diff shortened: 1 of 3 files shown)


andresmgot:
Hi @rarick, thanks for the PR!

> * Update the version number

Yes, bumping the patch version should be enough: 3.7.3. The CI will take care of the rest.

> * It appears that when an apprepository is created in another namespace, it's not cleaned up by `helm uninstall ...`. I'm not entirely familiar with helm, so I could use help on this

Yes, we are aware of the issue. It's still open here: https://github.com/kubeapps/kubeapps/issues/1832

> * I'm not acquainted with the codebase, so I'd like a maintainer to verify that this namespace change isn't breaking anything I'm unaware of

Thanks for checking. Yes, this change is fine :+1:

chart/kubeapps/values.yaml
```diff
@@ -240,6 +240,8 @@ apprepository:
   #   # It should start with -----BEGIN CERTIFICATE----- or
   #   # -----BEGIN RSA PRIVATE KEY-----
   #   caCert:
+  #   # Create this apprepository in some other namespace
```
andresmgot, line 243: I would say `Create this apprepository in a custom namespace`

rarick:
Awesome! Thanks for the feedback. I'll get those patched ASAP.

PR #1865, opened 2020-07-14
Catalog v2

andresmgot:
### Description of the change

Catalog view for Clarity:

Peek 2020-07-15 18-28

### Applicable issues

  - ref #1354


dashboard/src/components/Catalog/Catalog.v2.tsx
```diff
@@ -0,0 +1,152 @@
...
+    fetchCharts,
+    namespace,
+    pushSearchFilter,
+    getCSVs,
+    csvs,
+    repo,
+    filter: propsFilter,
+  } = props;
+  const [searchFilter, setSearchFilter] = useState(propsFilter);
+  const [typeFilter, setTypeFilter] = useState([] as string[]);
```
Angelmmiguel, line 51: For the type filter, I'd use a `radio` button with the options:

- All
- Charts
- Operators

It's not common to display a checkbox already marked, but the default behavior should be the "All" one. Also, conceptually it's easier to identify you click on an option to include only those items than excluding the group.

dashboard/src/components/Catalog/Catalog.v2.tsx
```diff
@@ -0,0 +1,152 @@
...
+              <h5>Filters</h5>
+              {csvs.length > 0 && (
+                <>
+                  <div className="filter-label">Application Type:</div>
+                  <FilterGroup
+                    name="apptype"
+                    options={["Operators", "Charts"]}
+                    onChange={setTypeFilter}
+                  />
+                </>
```
Angelmmiguel, line 121: Comment about using a radio button in this case

dashboard/src/components/Catalog/Catalog.v2.tsx
```diff
@@ -0,0 +1,152 @@
...
+            page.
+          </Alert>
+        )}
+        <Row>
+          <Column span={2}>
+            <div className="filters-menu">
+              <h5>Filters</h5>
+              {csvs.length > 0 && (
+                <>
+                  <div className="filter-label">Application Type:</div>
```
Angelmmiguel, line 115: Use a real `label` HTML tag here.

dashboard/src/components/Catalog/Catalog.v2.tsx
```diff
@@ -0,0 +1,152 @@
...
+                    options={["Operators", "Charts"]}
+                    onChange={setTypeFilter}
+                  />
+                </>
+              )}
+              {allRepos.length > 0 && (
+                <>
+                  <div className="filter-label">Application Repository:</div>
+                  <FilterGroup name="apprepo" options={allRepos} onChange={setRepoFilter} />
+                </>
```
Angelmmiguel, line 127: I'd add some vertical spacing between filters. You can use a `margin-top`

dashboard/src/components/Catalog/CatalogItems.tsx
```diff
@@ -0,0 +1,66 @@
...
+        id: c.id,
+        name: c.attributes.name,
+        icon: c.attributes.icon ? `api/assetsvc/${c.attributes.icon}` : undefined,
+        version: c.relationships.latestChartVersion.data.app_version,
+        description: c.attributes.description,
+        repo: c.attributes.repo,
+        namespace,
+      },
+    };
+  });
```
Angelmmiguel, line 27: Use `useMemo` here. You don't want to compute this on every render of this component, but only when `charts` changes.

dashboard/src/components/Catalog/CatalogItems.tsx
```diff
@@ -0,0 +1,66 @@
...
+              csv: csv.metadata.name,
+              namespace,
+            },
+          };
+        });
+      } else {
+        return [];
+      }
+    })
+    .flat();
```
Angelmmiguel, line 50: Same about this.

dashboard/src/components/Catalog/Catalog.v2.tsx
```diff
@@ -0,0 +1,152 @@
...
+      <LoadingWrapper loaded={!isFetching}>
+        {error && (
+          <Alert theme="danger">Found en error fetching the catalog: {error.message}</Alert>
+        )}
+        {charts.length === 0 && csvs.length === 0 && (
+          <Alert theme="warning">
+            Charts not found. Manage your Helm chart repositories in Kubeapps by visiting the{" "}
+            <Link to={app.config.apprepositories(namespace)}>App repositories configuration</Link>{" "}
+            page.
+          </Alert>
```
Angelmmiguel, line 107: Why is this a warning? We can transform this into a "Welcome" message to start working with Kubeapps.

A welcome message that suggest you the next step you need to accomplish to configure your application

dashboard/src/components/Catalog/CatalogItem.v2.tsx
```diff
@@ -0,0 +1,40 @@
...
+  item: IChartCatalogItem | IOperatorCatalogItem;
+}
+
+function CatalogItem(props: ICatalogItemProps) {
+  if (props.type === "operator") {
+    const item = props.item as IOperatorCatalogItem;
+    return OperatorCatalogItem(item);
+  } else {
+    const item = props.item as IChartCatalogItem;
+    return ChartCatalogItem(item);
```
Angelmmiguel, line 36: Just for consistency, use `JSX` here too:

```suggestion
    return <OperatorCatalogItem { ...item} />;
  } else {
    const item = props.item as IChartCatalogItem;
    return <ChartCatalogItem {...item} />;
```

dashboard/src/components/Catalog/Catalog.v2.tsx
andresmgot: mmh, for consistency with the rest of filters, I would use checkboxes. My idea is to have the typical filter checkbox list in which marking none of them is the same result than marking all (not sure if my screenshot confused you). An example from amazon:

Screenshot from 2020-07-15 17-42-05

Marking none of the items show all of them, click in one filter only those elements.

dashboard/src/components/Catalog/Catalog.v2.tsx
andresmgot: This is because if the catalog is empty, it's likely that something is going wrong. By default, kubeapps includes several repos so this should always have something but it's true that it can be the case in which a user deletes all repos.

andresmgot: Screenshot from 2020-07-15 18-21-53

dashboard/src/components/Catalog/Catalog.v2.tsx
Angelmmiguel: Okey! We can left it based on checkboxes. Since they're not active by default, I think it's fine :)

PR #1873, opened 2020-07-20
Resource Tabs

andresmgot:
### Description of the change

Following up on #1872, use tabs to show the different resource types:

Screenshot from 2020-07-20 17-14-32

This PR also includes a generic "Tabs" type.

### Applicable issues

  - ref #1354

### DIFF
--- a/dashboard/src/components/Tabs/Tabs.tsx
+++ b/dashboard/src/components/Tabs/Tabs.tsx
@@ -3,25 +3,26 @@ import React, { useState } from "react";
 import "./Tabs.scss";
 
 interface ITabsProps {
+  id: string;
   columns: string[];
   data: Array<string | JSX.Element | JSX.Element[]>;
 }
 
-export default function Tabs({ columns, data }: ITabsProps) {
+export default function Tabs({ id, columns, data }: ITabsProps) {
   const [selected, setSelected] = useState(0);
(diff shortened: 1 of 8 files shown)


andresmgot:
> Is the centered text the default alignment for tables? I wonder if there's a special class for the main column (my eyes expected the left-most column in your screenshot to be left-aligned, but that's anecdotal).

Will address this feedback in the other PR.

dashboard/src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,49 @@
...
+  return (
+    <div className="tabs">
+      <ul className="nav" role="tablist">
+        {columns.map((column, index) => {
+          return (
+            <li role="presentation" className="nav-item" key={`${column}-${index}`}>
+              <button
+                id={`tab${index}`}
+                className={`btn btn-link nav-link tab-button ${selected === index ? "active" : ""}`}
+                aria-controls={`panel${index}`}
```
Angelmmiguel, line 24: To avoid having issues with duplicated `id` in the same view, I'd ask parent components to define a `prefix`/`id` for a `Tabs` component. In that way, you can create unique ids per `Tabs` component:

```
`${id}-content-${index}
```

PR #1875, opened 2020-07-20
Chart info in app view

andresmgot:
### Description of the change

Adaptation of the chart info:

Screenshot from 2020-07-20 18-18-43
Screenshot from 2020-07-20 18-18-29

### Applicable issues

  - ref #1354


dashboard/src/components/AppView/ChartInfo.v2.tsx
```diff
@@ -0,0 +1,40 @@
...
+  }
+  return (
+    <>
+      <section className="chartinfo-subsection" aria-labelledby="chartinfo-description">
+        <h5 className="chartinfo-subsection-title" id="chartinfo-description">
+          Description
+        </h5>
+        <span>{metadata?.description}</span>
+      </section>
+      {notes}
```
absoludity, line 34: Rather than adding empty div's, couldn't we just `let notes` above be undefined initially, so that here we can just use notes if it's defined?

dashboard/src/components/AppView/ChartUpdateInfo.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+interface IChartInfoProps {
+  app: IRelease;
+}
+
+export default function ChartUpdateInfo(props: IChartInfoProps) {
+  const { app } = props;
+  // If update is not set yet we cannot know if there is
+  // an update available or not
+  if (app.updateInfo) {
```
absoludity, line 22: Could be a little easier to read calling out to a separate function: `const updateInfo = getUpdateInfoForApp(app)` to be returned below? Fine either way, but either way, there is quite a lot of logic here that should be tested.

andresmgot: _now it should be more readable, and covered by tests_

dashboard/src/components/AppView/ChartInfo/ChartInfo.v2.tsx
```diff
@@ -0,0 +1,44 @@
...
+      </section>
+    );
+  }
+  return (
+    <>
+      {descriptionInfo}
+      {versionsInfo}
+      <ChartUpdateInfo app={app} />
+    </>
+  );
```
Angelmmiguel, line 41: I'd keep all the JSX code in the same place. In that way, you only need to check the returned value to know the final code you're returning. Just a suggestion.

```suggestion
  return (
    <>
      {metadata && (
        <section className="chartinfo-subsection" aria-labelledby="chartinfo-description">
          <h5 className="chartinfo-subsection-title" id="chartinfo-description">
            Description
          </h5>
          <span>{metadata.description}</span>
        </section>
      )}
      {metadata && (
        <section className="chartinfo-subsection" aria-labelledby="chartinfo-versions">
          <h5 className="chartinfo-subsection-title" id="chartinfo-versions">
            Versions
          </h5>
          <div>
            {metadata.appVersion && <div>App Version: {metadata.appVersion}</div>}
            <span>Chart Version: {metadata.version}</span>
          </div>
        </section>
      )}
      <ChartUpdateInfo app={app} />
    </>
  );
```

dashboard/src/components/AppView/ChartInfo/ChartUpdateInfo.tsx
```diff
@@ -0,0 +1,71 @@
...
+        <CdsIcon shape="check-circle" size="md" solid={true} /> Up to date
+      </div>
+    );
+  }
+  if (chartMetadata.appVersion !== updateInfo.appLatestVersion) {
+    // There is a new application version
+    return (
+      <div className="color-icon-info">
+        <CdsIcon shape="circle-arrow" size="md" solid={true} /> A new version for{" "}
+        {chartMetadata.name} is available: {updateInfo.appLatestVersion}.
```
Angelmmiguel, line 42: Can we make `updateInfo.appLatestVersion` a link to the README file of that version. I think it's an interesting information.

dashboard/src/components/AppView/ChartInfo/ChartInfo.v2.tsx
```diff
@@ -0,0 +1,44 @@
...
+  let descriptionInfo;
+  if (metadata) {
+    versionsInfo = (
+      <section className="chartinfo-subsection" aria-labelledby="chartinfo-versions">
+        <h5 className="chartinfo-subsection-title" id="chartinfo-versions">
+          Versions
+        </h5>
+        <div>
+          {metadata.appVersion && <div>App Version: {metadata.appVersion}</div>}
+          <span>Chart Version: {metadata.version}</span>
```
Angelmmiguel, line 22: I'd emphasize the final data here. Since both label and value are using the same style, it's not clear to me what's important in this section. Maybe showing the versions using a bold text is enough:

```suggestion
          {metadata.appVersion && <div>App Version: <strong>{metadata.appVersion}</strong></div>}
          <span>Chart Version: <strong>{metadata.version}</strong></span>
```

dashboard/src/components/AppView/ChartInfo/ChartUpdateInfo.tsx
andresmgot: I wouldn't do so for two reasons:
 - READMEs are mostly identical between versions, the new README won't likely have new information.
 - (more important) The README is shown in the catalog view for that chart. There the action item is "Deploy", that may confuse the user that would deploy a new chart rather than upgrading the existing one. What we had before was a link to the upgrade form but since there is a button in the header I thought it was not that helpful.

Issue #1686, opened 2020-04-22
Avoid requesting secrets in a namespace


andresmgot:
### Description:

In order to be able to update the existing AppRepositories and to be able to select Docker credentials, we are listing all the secrets in a namespace in the Dashboard. We should be able to limit the info requested in order to avoid potential security issues.

absoludity:
In particular, there is no need to fetch secrets while listing app repositories (looks like it's currently done just as a short-cut). We should only be fetching secrets which are required for editing the app repository currently.

PR #1877, opened 2020-07-21
Update api server urls to use url helper.

absoludity:
Follow on from #1843, where [Andres' review mentioned that the namespace
url should be using our url helper](https://github.com/kubeapps/kubeapps/pull/1843#discussion_r449450126). While there, I've grouped all the
existing helpers which hit the api server directly under `api.k8s` as
well as a few more, so that we can clearly identify (and
potentially remove - see #1686) where the dashboard assumes direct access to the k8s
api.

It also updates the shared Secrets helper to be cluster aware (though the call-sites for app repository functionality are not, and will-not be for now).

Ref #1762

### DIFF
--- a/dashboard/src/shared/Secret.ts
+++ b/dashboard/src/shared/Secret.ts
@@ -1,6 +1,6 @@
 import { axiosWithAuth } from "./AxiosInstance";
-import { APIBase } from "./Kube";
 import { IK8sList, IOwnerReference, ISecret } from "./types";
+import * as url from "./url";
 
 export default class Secret {
   public static async create(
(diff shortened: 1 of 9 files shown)


dashboard/src/shared/Secret.ts
```diff
@@ -1,54 +1,35 @@
 import { axiosWithAuth } from "./AxiosInstance";
-import { APIBase } from "./Kube";
-import { IK8sList, IOwnerReference, ISecret } from "./types";
+import { IK8sList, ISecret } from "./types";
+import * as url from "./url";
 
 export default class Secret {
-  public static async create(
```
andresmgot, line 6: has this function gone somewhere else?

andresmgot: ah, I see, this is not needed because the secrets are now being created in the backend.

PR #1878, opened 2020-07-21
Pass additionalClusters to NewHandler and SetupDefaultRoutes.

absoludity:
When testing end-to-end, one of the issues I found in the backend was that we weren't using the `additionalClusters` configuration when setting up handlers and the default routes.

This change fixes that and removes some lint.

Ref: #1762

### DIFF
--- a/cmd/kubeops/internal/handler/handler.go
+++ b/cmd/kubeops/internal/handler/handler.go
@@ -22,7 +22,7 @@ import (
 
 const (
 	authHeader     = "Authorization"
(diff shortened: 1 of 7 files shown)


absoludity:
> I don't see where the new parameter (additionalClustersConfig) is being used?

`additionalClustersConfig` was already a private field on `kube.kubeHandler` and was already being used (and tested) to setup the required config during `AsUser`, the problem was that it was never being set by a call-site, as `NewHandler` didn't take it as an arg and hence never set the field.

PR #1879, opened 2020-07-21
Send cluster through to login form and actions.

absoludity:
More code pulled out of the demo branch at #1844 . In particular, this PR updates the Auth helper, actions and usage in login forms to all include the cluster from the state.

Follows #1877, Ref: #1762

### DIFF
--- a/dashboard/src/actions/auth.ts
+++ b/dashboard/src/actions/auth.ts
@@ -75,12 +75,9 @@ export function expireSession(): ThunkAction<Promise<void>, IStoreState, null, A
   };
 }
 
-export function checkCookieAuthentication(cluster: string): ThunkAction<
-  Promise<void>,
-  IStoreState,
-  null,
-  AuthAction
-> {
+export function checkCookieAuthentication(
+  cluster: string,
+): ThunkAction<Promise<void>, IStoreState, null, AuthAction> {
   return async dispatch => {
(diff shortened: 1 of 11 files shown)


dashboard/src/actions/auth.ts
```diff
@@ -25,14 +25,15 @@ const allActions = [setAuthenticated, authenticating, authenticationError, setSe
...
+  cluster: string,
   token: string,
   oidc: boolean,
 ): ThunkAction<Promise<void>, IStoreState, null, AuthAction> {
   return async dispatch => {
     dispatch(authenticating());
     try {
       if (!oidc) {
-        await Auth.validateToken(token);
+        await Auth.validateToken(cluster, token);
```
andresmgot, line 36: what's the expected UX if someone tries to use several clusters without OIDC?

absoludity: My preference would be that we detect the issue when parsing the config (config reducer) and state that it's not supported. Right now you'd simply not be able to login. What do you think?

andresmgot: we could do that at installation time? We can fail if a second cluster is given and OIDC is not enabled (https://austindewey.com/2018/12/28/helm-tricks-input-validation-with-required-and-fail/)

absoludity: Even better. Added to 1764.

PR #1881, opened 2020-07-22
Add a cluster selector to header when more than one cluster.

absoludity:
# Description of the change

Adds a cluster selector to the header iff more than one cluster is configured.

### Applicable issues

#1762 

### Additional information
I've also updated previous code I'd landed to use the new `mountWrapper` helper (and updated the helper lib to be able to create a store with some merged attributes).

### DIFF
--- a/dashboard/src/components/Header/ClusterSelector.test.tsx
+++ b/dashboard/src/components/Header/ClusterSelector.test.tsx
@@ -1,4 +1,5 @@
 import * as React from "react";
+import * as ReactRedux from "react-redux";
 
 import { getStore, mountWrapper } from "shared/specs/mountWrapper";
 import ClusterSelector, { IClusterSelectorProps } from "./ClusterSelector";
@@ -39,6 +40,8 @@ it("dispatches fetchNamespaces and calls onChange prop when changed", () => {
     ...defaultProps,
     onChange,
   };
+  const mockDispatch = jest.fn();
+  const spyOnUseDispatch = jest.spyOn(ReactRedux, "useDispatch").mockReturnValue(mockDispatch);
 
   const wrapper = mountWrapper(store, <ClusterSelector {...props} />);
   const select = wrapper.find("Select");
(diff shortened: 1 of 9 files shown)


dashboard/src/components/Header/ClusterSelector.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+it("dispatches fetchNamespaces and calls onChange prop when changed", () => {
+  const store = getStore({});
+  const onChange = jest.fn();
+  const props = {
+    ...defaultProps,
+    onChange,
+  };
+  const mockDispatch = jest.fn();
+  const spyOnUseDispatch = jest.spyOn(ReactRedux, "useDispatch").mockReturnValue(mockDispatch);
```
andresmgot, line 44: :+1: `jest.spyOn` was what I was looking for. BTW, you may want to move this to a `beforeEach` method to avoid copying it for every test.

dashboard/src/components/Header/ClusterSelector.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const select = wrapper.find("Select");
+  expect(select).toExist();
+
+  const selectOnChange = select.prop("onChange");
+  expect(selectOnChange).toBeDefined();
+  selectOnChange!({ value: "other" } as any);
+
+  // fetchNamespaces returns an async thunk action - hand to test more than dispatch
+  // was called once.
+  expect(mockDispatch).toBeCalledTimes(1);
```
andresmgot, line 56: what I've done is mocking the action:

```
  actions.kube = {
    ...actions.kube,
    getAndWatchResource: jest.fn(),
    closeWatchResource: jest.fn(),
  };
```

dashboard/src/components/Header/Header.test.tsx
```diff
@@ -163,3 +163,39 @@ it("doesn't call getNamespace when selecting all namespaces", () => {
   expect(setNamespace).toHaveBeenCalledWith("_all");
   expect(getNamespace).not.toHaveBeenCalled();
 });
+
+describe("ClusterSelector", () => {
+  it("does not render the cluster switcher when there is only one cluster", () => {
```
andresmgot, line 168: I am not sure if we should hide the selector, it would help promoting the feature. In any case, I am not hiding it in the new UI.

dashboard/src/components/Header/ClusterSelector.test.tsx
absoludity: Yep, if there was ore than one test using it.

dashboard/src/components/Header/Header.test.tsx
absoludity: Yep, once the feature is released, I agree, but right now it'd be confusing to see a cluster selector and not be able to use it? (Fine for the new UI, which is also not released).

andresmgot: yes, I mean once we release :+1:

PR #1882, opened 2020-07-22
AppSecrets section and first row completion

andresmgot:
### Description of the change

The goal of this PR is to include a new component that shows the secrets of the application in a more prominent place, along with some style changes and internal to finish the "first" row of the application view. The goal of this first row is to try to show everything needed to access the repo: readiness, access URL and credentials:

Screenshot from 2020-07-22 12-10-41

### Applicable issues

  - ref #1354

### DIFF
--- a/dashboard/src/shared/specs/mountWrapper.tsx
+++ b/dashboard/src/shared/specs/mountWrapper.tsx
@@ -8,9 +8,9 @@ import thunk from "redux-thunk";
 
 import { IStoreState } from "../../shared/types";
 
-export const mockStore = configureMockStore([thunk]);
+const mockStore = configureMockStore([thunk]);
 
-export const initialState = {
+const initialState = {
   apps: {},
   auth: {},
   catalog: {},
--- a/dashboard/src/shared/specs/mountWrapper.tsx
+++ b/dashboard/src/shared/specs/mountWrapper.tsx
@@ -15,7 +15,7 @@ const initialState = {
   auth: {},
   catalog: {},
   charts: {},
-  config: {},
+  config: { featureFlags: {} },
   kube: {
     items: {},
   },
--- a/dashboard/src/shared/specs/mountWrapper.tsx
+++ b/dashboard/src/shared/specs/mountWrapper.tsx
@@ -7,15 +7,17 @@ import thunk from "redux-thunk";
 
 import { IStoreState } from "../../shared/types";
 
-const mockStore = configureMockStore([thunk]);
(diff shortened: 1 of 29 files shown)


dashboard/src/shared/specs/mountWrapper.tsx
```diff
@@ -7,15 +7,17 @@ import thunk from "redux-thunk";
 
 import { IStoreState } from "../../shared/types";
 
-const mockStore = configureMockStore([thunk]);
+export const mockStore = configureMockStore([thunk]);
 
-const initialState = {
+export const initialState = {
```
absoludity, line 12: I don't think we need these extra exports, if we use `getStore(objectToMergeWithInitialState)` that I landed earlier.  See what you think.

dashboard/src/shared/utils.ts
```diff
@@ -33,3 +36,28 @@ export function trimDescription(desc: string): string {
...
+  resources: { [s: string]: IKubeItem<IResource | IK8sList<IResource, {}>> },
+) {
+  const result: Array<IKubeItem<IResource | ISecret>> = [];
+  refs.forEach(ref => {
+    const kubeItem = resources[ref.getResourceURL()];
+    if (kubeItem) {
+      const itemList = kubeItem.item as IK8sList<IResource | ISecret, {}>;
+      if (itemList && itemList.items) {
+        itemList.items.forEach(i => {
+          result.push({
```
absoludity, line 51: So this won't handle a list of lists I don't think? But maybe that's never going to be a thing (don't know of a situation in k8s of such a list anyway).

absoludity:
> The goal of this first row is to try to show everything needed to access the repo: readiness, access URL and credentials:

Not all the credentials are actually needed, I don't think? For eg., in the example screenshot, it's only the WP secret which is relevant and needed. I wonder if it'd be worth differentiating.

andresmgot:
> > The goal of this first row is to try to show everything needed to access the repo: readiness, access URL and credentials:
> 
> Not all the credentials are actually needed, I don't think? For eg., in the example screenshot, it's only the WP secret which is relevant and needed. I wonder if it'd be worth differentiating.

Yes, they are not. In other cases it will be even less relevant for secrets like encoded certs but we have no way of knowing which of them are "important" to build that list.

dashboard/src/shared/utils.ts
andresmgot: I don't know if a list of lists is allowed by the K8s API, certainly I've never seen one.

Issue #1855, opened 2020-07-08
Kind value is being dropped from binding creation


ralphsmith80:
### Description:

Unable to create a Service Binding via the Kubeapps dashboard. This maybe related to #990 .

I have a video link you can download [here](https://1drv.ms/u/s!Au-fZXWZl4eHkHNbaKzvCUCgQ8JR).

### Steps to reproduce the issue:

1. Follow steps on the [service catalog walkthrough](https://svc-cat.io/docs/walkthrough/) up to [step 4](https://svc-cat.io/docs/walkthrough/#step-4---creating-a-new-serviceinstance)
2. Create a service instance via kubeapps
3. Create a service binding via kubeapps 

### Describe the results you received:

Dashboard is returning an error when creating a Service Binding.

### Describe the results you expected:

Service Binding should be created.

### Additional information you deem important (e.g. issue happens only occasionally):

Error response:
```
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {
    
  },
  "status": "Failure",
  "message": "ServiceBinding in version \"v1beta1\" cannot be handled as a ServiceBinding: unmarshalerDecoder: Object 'Kind' is missing in '{\"metadata\":{\"name\":\"mini-binding\"},\"spec\":{\"instanceRef\":{\"name\":\"mini-instance\"},\"parameters\":{}}}', error found in #10 byte of ...|ters\":{}}}|..., bigger context ...|nceRef\":{\"name\":\"mini-instance\"},\"parameters\":{}}}|...",
  "reason": "BadRequest",
  "code": 400
}
```

The error message seems to imply that the `kind` property is missing. I attempted to issue the same request via Postman with the added `"kind": "ServiceBinding"` metadata key:value but that failed with the same error response. 

### Version of Helm, Kubeapps and Kubernetes:

- Output of `helm version`: 

```
version.BuildInfo{Version:"v3.2.4", GitCommit:"0ad800ef43d3b826f31a5ad8dfbb4fe05d143688", GitTreeState:"dirty", GoVersion:"go1.14.3"}
```

- Output of `helm list <kubeapps-release-name>`:

Installed via [Quick Start in README](https://github.com/kubeapps/kubeapps#quick-start). Version in UI is noted as [v1.10.2](https://github.com/kubeapps/kubeapps/releases/tag/v1.10.2)

```
Error: "helm list" accepts no arguments
```

- Output of `kubectl version`:

```
[...]

PR #1884, opened 2020-07-22
Fix/#1855

ralphsmith80:
This fix adds missing post body parameter to `ServiceBinding:create`. These parameters are required.

I didn't use a close commit for this because while this does solve the problem of creating a binding I'm still seeing a 404 error fetching the secret after the binding is created. I will look at that further but I would like to keep the commits focused and small.

The second commit in this PR is an update to the developer documentation that provide an alternative to Telepresence for running the dashboard locally. I needed to do that myself since I could not get past the Mac SIP issue with Telepresence without disabling it entirely which I did not want to do.

As noted above I don't want to but `fixes` because it solves some of the issues with Binding but I'm not sure yet if it solves all the problems.

### DIFF
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -172,5 +172,6 @@
       "last 1 firefox version",
       "last 1 safari version"
     ]
-  }
+  },
+  "proxy": "http://127.0.0.1:8080"
 }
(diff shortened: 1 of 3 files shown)


ralphsmith80:
Looks like I can't manually link the issue either so I'll say relates to #1855 .

ralphsmith80:
I was on the fence if the fetch for the binding secret was on my broker so I ran through this with the minibroker and an instance of the mariadb. While the initial fetch for the secret failed after refreshing the page it worked as expected.  So I think we can close #1855 . 

The 404 issue is an issue not properly waiting or handling the async behavior of k8s creating the secret after the binding is created. 

## Initial binding creation
Binding-Created

binding-secret-fail-response

binding-secret-fails

## After page refresh
binding-secret-success-response

binding-secret

PR #1886, opened 2020-07-23
Update namespace actions/reducer to be cluster aware.

absoludity:
### Description of the change

Updates the namespace actions and the reducer to be cluster-aware, updating state per-cluster.

### Benefits

Required for multicluster.

### Possible drawbacks

None

### Applicable issues

Ref: #1762

### DIFF
--- a/dashboard/src/components/AppList/AppList.tsx
+++ b/dashboard/src/components/AppList/AppList.tsx
@@ -59,7 +59,11 @@ class AppList extends React.Component<IAppListProps, IAppListState> {
       namespace,
     } = this.props;
     // refetch if new namespace or error removed due to location change
-    if (prevProps.namespace !== namespace || (!error && prevProps.apps.error)) {
+    if (
+      prevProps.namespace !== namespace ||
+      prevProps.cluster !== cluster ||
+      (!error && prevProps.apps.error)
+    ) {
       fetchAppsWithUpdateInfo(cluster, namespace, listingAll);
       if (this.props.featureFlags.operators) {
         getCustomResources(namespace);
--- a/dashboard/src/actions/auth.ts
+++ b/dashboard/src/actions/auth.ts
@@ -3,7 +3,7 @@ import { ActionType, createAction } from "typesafe-actions";
 
 import { Auth } from "../shared/Auth";
 import { IStoreState } from "../shared/types";
-import { clearNamespaces, NamespaceAction } from "./namespace";
+import { clearClusters, NamespaceAction } from "./namespace";
 
 export const setAuthenticated = createAction("SET_AUTHENTICATED", resolve => {
   return (authenticated: boolean, oidc: boolean, defaultNamespace: string) =>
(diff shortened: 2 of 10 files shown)


dashboard/src/components/AppList/AppList.tsx
```diff
@@ -59,7 +59,11 @@ class AppList extends React.Component<IAppListProps, IAppListState> {
       namespace,
     } = this.props;
     // refetch if new namespace or error removed due to location change
-    if (prevProps.namespace !== namespace || (!error && prevProps.apps.error)) {
+    if (
+      prevProps.namespace !== namespace ||
+      prevProps.cluster !== cluster ||
+      (!error && prevProps.apps.error)
+    ) {
```
andresmgot, line 66: there are a lot of comparisons like this one:

```
dashboard/src/components/OperatorNew/OperatorNew.tsx
63:    if (prevProps.namespace !== this.props.namespace) {

dashboard/src/components/OperatorList/OperatorList.tsx
88:    if (prevProps.namespace !== this.props.namespace) {

dashboard/src/components/OperatorView/OperatorView.tsx
40:    if (prevProps.namespace !== this.props.namespace) {

dashboard/src/components/AppView/AppView.tsx
93:    if (prevProps.namespace !== namespace) {

dashboard/src/components/AppList/AppList.tsx
62:    if (prevProps.namespace !== namespace || (!error && prevProps.apps.error)) {

dashboard/src/components/ServiceInstanceList/ServiceInstanceList.tsx
76:      (prevProps.namespace !== namespace || (error && !prevProps.error))

dashboard/src/components/Config/AppRepoList/AppRepoList.tsx
100:    if (prevProps.namespace !== namespace || (prevProps.errors.fetch && !fetch)) {

dashboard/src/components/OperatorInstance/OperatorInstance.tsx
62:    if (prevProps.namespace !== namespace) {

dashboard/src/components/ServiceInstanceView/ServiceInstanceView.tsx
94:    if (prevProps.namespace !== namespace) {

```

It is a pain to maintain those... using function components help so if you miss something I should be able to notice it in the new UI.

absoludity: I updated the two relevant ones in 12628de (appview and apprepolist), as we're not yet supporting operators across clusters.

absoludity:
Rubber-ducking here: So I've fixed the e2e issue with the operator test in https://github.com/kubeapps/kubeapps/pull/1886/commits/bfe0bab23f541cfcaa76e907297c4c0dbd6b8439 (the operator views don't include the cluster in the route, so after deleting the operator, the operator instance was switching to an `undefined` cluster. But now I'm consistently seeing two e2e tests fail - mongodb-helm3 and postgresql-helm3.

Both fail during `create-private-registry` with `Element button (text: "Add new credentials") not found` and sure enough, the artifact shows that the whole section is not displayed:

creates-a-private-registry

which, from the code, should be displayed if `namespace !== kubeappsNamespace`. Yet the test is starting at the URL `/#/config/ns/default/repos`, stays at that namespace when I run the test locally, and is passing for the helm2 tests nor can I trigger it locally manually. There must be something I'm missing - do you mind being a second pair of eyes for me @andresmgot ?

absoludity:
nm, I can now reproduce locally (after switching to a service account token associated with the kubeapps namespace, like ci does, rather than the default namespace). Ends up with a route using the default ns, but displaying the kubeapps namespace data and selector:

namespace-issue-1

PR #1887, opened 2020-07-23
Improved secret style

andresmgot:
### Description of the change

Follow up of https://github.com/kubeapps/kubeapps/pull/1882#pullrequestreview-453869187

Peek 2020-07-23 13-00

### Applicable issues

  - ref #1354


dashboard/src/components/AppView/ResourceTable/ResourceItem/SecretItem/SecretItemDatum.v2.tsx
```diff
@@ -13,11 +16,29 @@ interface ISecretItemDatumProps {
...
+  };
   const decodedValue = atob(value);
 
   return (
     <Row>
+      <Column span={5}>
+        <div className="secret-datum-text">{name}</div>
+      </Column>
+      <Column span={5}>
+        <input
```
andresmgot, line 35: Note that I couldn't use the `Input` component, the style was not properly applied :/ 

Screenshot from 2020-07-23 13-05-37

dashboard/src/components/AppView/ResourceTable/ResourceItem/SecretItem/SecretItemDatum.v2.scss
```diff
@@ -21,3 +21,7 @@
     outline: none;
   }
 }
+
+.secret-datum-content {
+  border-bottom: 1px solid var(--clr-forms-border-color, #b3b3b3) !important;
```
andresmgot, line 26: I needed this because by default, `readOnly` input fields don't have the border-bottom but I wanted it to show the length of the input field.

Angelmmiguel: Interesting!

dashboard/src/components/AppView/ResourceTable/ResourceItem/SecretItem/SecretItemDatum.v2.tsx
Angelmmiguel: Np. I think this is a special case of Input usage

dashboard/src/components/AppView/ResourceTable/ResourceItem/SecretItem/SecretItemDatum.v2.tsx
```diff
@@ -13,11 +16,29 @@ interface ISecretItemDatumProps {
...
 function SecretItemDatum({ name, value }: ISecretItemDatumProps) {
   const [hidden, setHidden] = React.useState(true);
+  const [copied, setCopied] = React.useState(false);
   const toggleDisplay = () => setHidden(!hidden);
+  const setCopiedTrue = () => {
+    setCopied(true);
+    setTimeout(() => {
+      setCopied(false);
+    }, 1000);
+  };
```
Angelmmiguel, line 26: This code may cause an error if the user navigates to a different page when the tooltip is showed. This component will be unmounted and the `setCopied` method will disappear from the context. To avoid this, you can use `useRef` to keep the timeout reference and `useEffect` to control it:

```suggestion
  const [copied, setCopied] = React.useState(false);
  const toggleDisplay = () => setHidden(!hidden);
  const copyTimeout = useRef(null);

  // You can now remove this intermediate method.
  const setCopiedTrue = () => {
    setCopied(true);
  };

  useEffect(() => {
    if (copied) {
      copyTimeout.current = setTimeout(() => {
        setCopied(false);
      }, 1000);
    }

    return () => {
      if (copyTimeout.current != null) {
        clearTimeout(copyTimeout.current);
      }
    }
  }, [copied])
```

dashboard/src/components/AppView/ResourceTable/ResourceItem/SecretItem/SecretItemDatum.v2.tsx
```diff
@@ -13,11 +16,29 @@ interface ISecretItemDatumProps {
...
   const decodedValue = atob(value);
 
   return (
     <Row>
+      <Column span={5}>
+        <div className="secret-datum-text">{name}</div>
```
Angelmmiguel, line 32: I'd change this to a label and connect it to the `input`

dashboard/src/components/AppView/ResourceTable/ResourceItem/SecretItem/SecretItemDatum.v2.tsx
```diff
@@ -33,11 +54,18 @@ function SecretItemDatum({ name, value }: ISecretItemDatumProps) {
...
-      <Column span={11}>
-        <div className="secret-datum-text" id={`secret-item-datum-${name}-ref`}>
-          <span>
-            {name}: {hidden ? `${decodedValue.length} bytes` : <strong>{decodedValue}</strong>}
-          </span>
+      <Column span={1}>
+        <button className="secret-datum-icon" aria-expanded={!hidden} onClick={setCopiedTrue}>
+          <div data-tip={true} data-for="app-status">
+            <CopyToClipboard text={decodedValue}>
+              <CdsIcon shape="copy-to-clipboard" size="md" solid={true} />
```
Angelmmiguel, line 61: Add a `aria-label` to this icon so it notifies screen readers its purpose:

```suggestion
              <CdsIcon shape="copy-to-clipboard" size="md" solid={true} aria-label={`Copy ${name} secret value to the clipboard`}/>
```

dashboard/src/components/AppView/ResourceTable/ResourceItem/SecretItem/SecretItemDatum.v2.tsx
andresmgot: good catch!

andresmgot:
> Perhaps we could have a way - if a schema has been included for a chart (for easy deployment), it would tend to include secrets that are relevant to users, I'd think? So we could do something like: if the chart has a schema and the schema includes some secret fields, highlight those credentials only (and otherwise highlight all or none - they're still in the resources below). Not suggesting for this PR.

The schema uses the `values.yaml`, there is no way to map `.values.wordpressPassword` to the secret containing those credentials.

Ideally, what app developers could do is tagging secrets with an annotation. Something like `kubeapps.com/credentials/username: foo, kubeapps.com/credentials/password: wordpress-password` (being `wordpress-password` the key of the secret data). That way we could ensure those are the app credentials and we can highlight only that.

PR #1888, opened 2020-07-23
Adding eslint

andresmgot:
### Description of the change

I noticed some eslint warnings when running the app locally. We were not taking into account eslint warnings in the build process, some fixes are useful.


andresmgot:
> Do you mind adding ` && yarn run eslint` (assuming it fails if there is lint found) as part of this so that it's not possible to pass CI if it hasn't been run?

I was going to do so but it's not necessary since the `test_dashboard` job already executes `yarn run lint` (and this script executes `lint-js lint-css eslint`).

PR #1889, opened 2020-07-24
Update all call-sites to ensure ResourceRefs use cluster.

absoludity:

### Description of the change

Follows on from #1886, updating ResourceRefs to include the cluster (in the constructor as a required arg) then updating all the call-sites to use it (mainly in tests, other than AppView and OperatorInstance (which uses fromCRD)).

It was a pain to audit because the typechecker can't find everything, as the 3rd arg is an optional string. Anyway, I think I've audited all call-sites now.

This is slightly different to how I did the change in the demo branch, and has the benefit that the resource refs just calculate multicluster URLs once created, so users of the objects don't need to care.

### Benefits

Required for multicluster support.

### Possible drawbacks

Danger of missing a call-site.

### Applicable issues

Ref: #1762

### DIFF
--- a/dashboard/src/shared/ResourceRef.ts
+++ b/dashboard/src/shared/ResourceRef.ts
@@ -5,6 +5,7 @@ import { IClusterServiceVersionCRDResource, IK8sList, IResource } from "./types"
 
 export function fromCRD(
   r: IClusterServiceVersionCRDResource,
+  cluster: string,
   namespace: string,
   ownerReference: any,
 ) {
@@ -17,7 +18,7 @@ export function fromCRD(
   // TODO(andresmgot): This won't work for new resource types, we would need to dinamically
   // resolve those
   const resourceNamespace = r.kind.startsWith("Cluster") ? "" : namespace;
-  return new ResourceRef(resource, resourceNamespace, {
+  return new ResourceRef(resource, cluster, resourceNamespace, {
     metadata: { ownerReferences: [ownerReference] },
   });
 }
@@ -25,6 +26,7 @@ export function fromCRD(
 // ResourceRef defines a reference to a namespaced Kubernetes API Object and
 // provides helpers to retrieve the resource URL
 class ResourceRef {
+  public cluster: string;
   public apiVersion: string;
   public kind: ResourceKind;
   public name: string;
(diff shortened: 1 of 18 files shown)


dashboard/src/shared/ResourceRef.ts
```diff
@@ -33,9 +35,8 @@ class ResourceRef {
 
   // Creates a new ResourceRef instance from an existing IResource. Provide
   // defaultNamespace to set if the IResource doesn't specify a namespace.
-  // TODO: add support for cluster-scoped resources, or add a ClusterResourceRef
-  // class.
-  constructor(r: IResource, defaultNamespace?: string, defaultFilter?: any) {
+  constructor(r: IResource, cluster: string, defaultNamespace?: string, defaultFilter?: any) {
```
andresmgot, line 38: for the maintainability of this I would:

 - Make the `defaultNamespace` mandatory. It's easier to understand calls like `new ResourceRef({}, "cluster-a", namespace)`. Also you would notice if you are missing the cluster in some call.
 - Remove the `defaultFilter` param. AFAIK, it's only used in the function `fromCRD` above. You can move it to a setter.

absoludity: > for the maintainability of this I would:

These are both not things I changed, but always keen to improve things...

> 
>     * Make the `defaultNamespace` mandatory. It's easier to understand calls like `new ResourceRef({}, "cluster-a", namespace)`. Also you would notice if you are missing the cluster in some call.

It's not clear to me why we even have this arg... the docstring says in case an IResource doesn't have a namespace, but it's a required field on IResource. I think we are only ever showing namespaced resources? Smells either like a hack to work-around bad input, or old cruft that hasn't been removed? Let me know, but either way I'm keen to change that in a separate branch and move forward here.


> 
>     * Remove the `defaultFilter` param. AFAIK, it's only used in the function `fromCRD` above. You can move it to a setter.

+1. It's a public field, so just set it in `fromCRD`.

andresmgot: > It's not clear to me why we even have this arg... the docstring says in case an IResource doesn't have a namespace, but it's a required field on IResource. I think we are only ever showing namespaced resources? Smells either like a hack to work-around bad input, or old cruft that hasn't been removed? Let me know, but either way I'm keen to change that in a separate branch and move forward here.

This parameter is needed because we parse the manifests installed by helm to retrieve the different resources. With helm, it's not necessary to set the namespace. If a resource doesn't have a namespace, it's installed in the namespace of the release.

absoludity: Thanks for the explanation. That being the case, it seems odd to require it as a default even if it's not necessary (rather than just leaving it as an optional default). But I'll take a closer look.

PR #1891, opened 2020-07-24
Disable host check for dev setup

andresmgot:
### Description of the change

After https://github.com/kubeapps/kubeapps/pull/1884, the expected host is `127.0.0.1` when running the server locally, this breaks my local dev environment because I am using `kubeapps.local`. Disabling the host check since we don't need it for the dev setup.

### DIFF
--- a/docs/developer/dashboard.md
+++ b/docs/developer/dashboard.md
@@ -68,6 +68,16 @@ yarn run start
 
 As an alternative to using [Telepresence](https://www.telepresence.io/) you can use the default [Create React App API proxy](https://create-react-app.dev/docs/proxying-api-requests-in-development/) functionality.
 
+First add the desired host:port to the package.json:
+
+```patch
+-  }
++  },
++  "proxy": "http://127.0.0.1:8080"
+```
+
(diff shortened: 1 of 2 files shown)


andresmgot:
> Have you tried what's recommended in the [docs with specifying your public host in the the `.env.development` file](https://create-react-app.dev/docs/proxying-api-requests-in-development/#invalid-host-header-errors-after-configuring-proxy)?

It didn't work, I assume because we are accessing the service through telepresence. I had to remove the setting from the package.json.

docs/developer/dashboard.md
```diff
@@ -68,6 +68,16 @@ yarn run start
...
 
+First add the desired host:port to the package.json:
+
+```patch
+-  }
++  },
++  "proxy": "http://127.0.0.1:8080"
+```
+
+> **NOTE**: The [proxy](../../dashboard/package.json#L176) `key:value` has already added to the `package.json` for convenience but you can change the `host:port` values to meet your needs.
```
absoludity, line 79: This note is no longer correct, I think?

andresmgot: right, forgot to update the comment

PR #1892, opened 2020-07-24
Chart View v2 and other minor changes

andresmgot:
### Description of the change

Adapted version of the ChartView:

Peek 2020-07-24 16-01

Other minor changes I fixed along the way:
 - I gathered all the icons import to the `clarity` wrapper so we don't need to add them in every component.
 - Added an icon to the AppList "Deploy" button, to match the one in the ChartView.
 - Refactored the "left-menu" style so it can be reused in different componetns.

### Applicable issues

  - #1354


andresmgot:
> Not sure, but did you perhaps push the changes to the deploymentform here by mistake? Looks like a different PR. +1'd though.

Yep, things that happen the last minute of a Friday :)

PR #1897, opened 2020-07-28
Do not allow app repo listing for other clusterss.

absoludity:
### Description of the change

Ensures that if a user navigates to an additional cluster while viewing the config for app repos, or visits the config with the additional cluster already selected, that an informative message is displayed regarding their options:

apprepos-default-only

### Benefits

We don't leave people not knowing why app repos aren't working for additional clusters.

### Applicable issues

Follows #1893 Ref: #1762

### Additional information

The next PR will ensure that an `--insecure` flag is passed before accepting additional clusters without CAData, as well as fixing the issue when CAData is provided and calls fail.

### DIFF
--- a/dashboard/src/components/Config/AppRepoList/AppRepoList.test.tsx
+++ b/dashboard/src/components/Config/AppRepoList/AppRepoList.test.tsx
@@ -1,6 +1,7 @@
 import { shallow } from "enzyme";
 import * as React from "react";
 
+import { MessageAlert } from "components/ErrorAlert";
 import AppRepoList from "./AppRepoList";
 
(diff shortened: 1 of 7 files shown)


andresmgot:
I remember doing a comment in this PR regarding the message but I cannot find it. Did I forgot to send it or it was lost in Github?

absoludity:
I never saw one, nor does my email client record any message other than the approval. Happy to update it though?

andresmgot:
oh, I guess I misclicked. My comment was to try to make the message shorter (the second paragraph is too detailed?) and include a button to redirect to the global namespace in the default cluster (which is the one that affects other clusters).

PR #1898, opened 2020-07-28
Deployment Form v2

andresmgot:
### Description of the change

New deployment form:

Peek 2020-07-28 16-19

Note: At the end of the gif, the deployment fails because the user I was logged in didn't have enough permissions.

This PR includes some changes in the ChartView component since I extracted the sub-header (the one with the version selector) to a standalone component. This is called "ChartHeader".

### Applicable issues

  - ref #1354 

### Additional information

When working on this, I noticed a bug in the BasicDeploymentForm due to #1828 in which the switch-case was not properly handled so it's fixed here as well.


dashboard/src/components/ChartView/ChartHeader.v2.tsx
```diff
@@ -0,0 +1,92 @@
+import Column from "components/js/Column";
```
absoludity, line 1: I think this is only looks untested because you actually just extracted it from the chartview below, so still covered by the tests there (I assume).

andresmgot: yes, it's both tested with ChartView and the DeploymentForm

PR #1900, opened 2020-07-28
Update pinned bitnami charts in values.yaml

absoludity:
### Description of the change

Updates the bitnami charts in the values.yaml to their latest versions.

### Benefits

Some providers delete older tags (odd?) breaking chart deployments without manual configuration.

### Possible drawbacks

See if CI passes without issue.

### Additional information

Longer term, we should not pin these in our Kubeapps main branch, but pin them to the current latest when releasing the Kubeapps chart.

### DIFF
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -105,7 +105,7 @@ frontend:
   image:
     registry: docker.io
     repository: bitnami/nginx
-    tag: 1.18.0-debian-10-r38
+    tag: 1.19.1-debian-10-r16
     ## Specify a imagePullPolicy
     ## ref: http://kubernetes.io/docs/user-guide/images/#pre-pulling-images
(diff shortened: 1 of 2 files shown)


chart/kubeapps/values.yaml
```diff
@@ -646,7 +646,7 @@ authProxy:
   image:
     registry: docker.io
     repository: bitnami/oauth2-proxy
-    tag: 5.1.0-debian-10-r24
+    tag: 5.1.1-debian-10-r58
```
andresmgot, line 649: BTW, can you give a try to 6.x?

absoludity: 6.x requires option changes in our chart. I'll try to do that today separately.

absoludity: See #1904

PR #1901, opened 2020-07-29
App Upgrade v2

andresmgot:
### Description of the change

New upgrade flow:

Peek 2020-07-29 18-05

Select repo form:

Screenshot from 2020-07-29 18-10-26

### Possible drawbacks

I am not that happy with the version selector field. It's quite hidden and in this flow is critical, will revisit later to make it more prominent.

### Applicable issues

  - ref #1354 

### Additional information

This PR includes some unrelated fixes (adding the `cluster` param in some places that it was missing).

### DIFF
--- a/dashboard/src/components/DeploymentFormBody/DeploymentFormBody.v2.tsx
+++ b/dashboard/src/components/DeploymentFormBody/DeploymentFormBody.v2.tsx
@@ -145,7 +145,7 @@ function DeploymentFormBody({
       </div>
       <div className="deployment-form-control-buttons">
         <CdsButton status="primary" type="submit">
-          <CdsIcon shape="deploy" inverse={true} /> Deploy v{selected.version?.attributes.version}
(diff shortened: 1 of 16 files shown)


dashboard/src/components/DeploymentFormBody/DeploymentFormBody.v2.tsx
```diff
@@ -145,7 +145,7 @@ function DeploymentFormBody({
       </div>
       <div className="deployment-form-control-buttons">
         <CdsButton status="primary" type="submit">
-          <CdsIcon shape="deploy" inverse={true} /> Deploy v{chartVersion}
+          <CdsIcon shape="deploy" inverse={true} /> Deploy v{selected.version?.attributes.version}
```
absoludity, line 148: In what cases is `selected.version` undefined?

andresmgot: here it's never undefined because it's checked before but typescript did't detect it, rewriting to make it more clear

PR #1903, opened 2020-07-30
Update oauth2-proxy from 5.x to 6.x

absoludity:
### Description of the change

The only breaking change is that flags now require -- .
For details of breaking changes in 6.x:
https://github.com/oauth2-proxy/oauth2-proxy/blob/master/CHANGELOG.md#breaking-changes-1

### Benefits

Using more secure latest oauth2-proxy (project migrated earlier this year).

### Possible drawbacks

Users who have set `.Values.authProxy.additionalFlags` will need to update those (auth-proxy container will refuse to start).

### DIFF
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -668,10 +668,10 @@ authProxy:
   ## Additional flags for oauth2-proxy
   ##
   additionalFlags: []
-  # - -ssl-insecure-skip-verify
-  # - -cookie-secure=false
-  # - -scope=https://www.googleapis.com/auth/userinfo.email https://www.googleapis.com/auth/cloud-platform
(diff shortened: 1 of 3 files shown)


chart/kubeapps/values.yaml
```diff
@@ -646,7 +646,7 @@ authProxy:
   image:
     registry: docker.io
     repository: bitnami/oauth2-proxy
-    tag: 5.1.1-debian-10-r58
+    tag: 6.0.0-debian-10-r23
     ## Specify a imagePullPolicy
     ## ref: http://kubernetes.io/docs/user-guide/images/#pre-pulling-images
     ##
```
andresmgot, line 652: the comments in lines 671-674 need to be updated as well

PR #1906, opened 2020-07-30
Rollback and delete buttons for AppView v2

andresmgot:
### Description of the change

Delete modal:

Screenshot from 2020-07-30 15-46-25

Rollback modal:
Screenshot from 2020-07-30 15-46-12

### Applicable issues

  - ref #1354

### DIFF
--- a/dashboard/src/components/AppView/AppView.v2.tsx
+++ b/dashboard/src/components/AppView/AppView.v2.tsx
@@ -18,6 +18,8 @@ import { IK8sList, IRelease, IResource } from "../../shared/types";
 import * as url from "../../shared/url";
 import LoadingWrapper from "../LoadingWrapper/LoadingWrapper.v2";
 import AccessURLTable from "./AccessURLTable/AccessURLTable.v2";
+import DeleteButton from "./AppControls/DeleteButton/DeleteButton.v2";
+import RollbackButton from "./AppControls/RollbackButton/RollbackButton.v2";
 import AppNotes from "./AppNotes.v2";
 import AppSecrets from "./AppSecrets";
 import AppValues from "./AppValues/AppValues.v2";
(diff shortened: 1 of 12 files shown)


dashboard/src/components/AppView/AppView.v2.tsx
```diff
@@ -207,9 +210,7 @@ export default function AppView({
         </div>
       </PageHeader>
 
-      {error && (
-        <Alert theme="danger">Unable to load the application. Received: {error.message}</Alert>
-      )}
+      {error && <Alert theme="danger">Found an error: {error.message}</Alert>}
```
absoludity, line 213: "Found an error" is a bit vague perhaps? I assume you changed this because "Unable to load the application" is not generic enough (or just incorrect)? Perhaps `The previous operation falied with: {error.message}`

dashboard/src/components/ConfirmDialog/ConfirmDialog.v2.tsx
```diff
@@ -22,17 +24,19 @@ function ConfirmDialog({
...
+      {error && <Alert theme="danger">Found error: {error.message}</Alert>}
       {loading === true ? (
         <div className="confirmation-modal">
-          <div>Loading, please wait</div>
+          <span>Loading, please wait</span>
           <LoadingWrapper loaded={false} />
         </div>
       ) : (
         <div className="confirmation-modal">
-          <div>{confirmationText || "Are you sure you want to delete this?"}</div>
```
absoludity, line 35: I wouldn't use the word "this" in these (not your change, I know). Ah - I see, it's just the generic text if `confirmationText` isn't provided. Perhaps we should just make `confirmationText` required (not necessary here).

dashboard/src/components/AppView/AppView.v2.tsx
andresmgot: yes, the problem is that the error can appear either when loading the app or when upgrading/deleting/rollingback so it's not always after an "operation"

PR #1915, opened 2020-07-31
AppRepoList v2

andresmgot:
### Description of the change

First PR for adding support for the App Repo list management view:

Screenshot from 2020-08-18 17-16-57

Since this view is quite complex, this PR only contains the code related to the repository table and  the different buttons. I will send other PRs for:

 - The form for adding/editing a repo
 - The split between global and namespace repositories as detailed here: https://www.figma.com/proto/q8ZXJAU2JD1riZg2c47OeE/Wireframes?node-id=108%3A3164&scaling=min-zoom

Note that the buttons in the table are "outlined" rather than "flat" because I found a weird behavior: it's not possible to set the color to red ("danger") for flat buttons, the parameter gets ignored.

### DIFF
--- a/dashboard/src/components/Config/AppRepoList/AppRepoButton.v2.tsx
+++ b/dashboard/src/components/Config/AppRepoList/AppRepoButton.v2.tsx
@@ -1,0 +1,82 @@
+import React, { useState } from "react";
+
+import actions from "actions";
+import { CdsButton, CdsIcon } from "components/Clarity/clarity";
+import Alert from "components/js/Alert";
+import Modal from "components/js/Modal/Modal";
+import { useDispatch, useSelector } from "react-redux";
+import { Action } from "redux";
+import { ThunkDispatch } from "redux-thunk";
+import { IAppRepository, ISecret, IStoreState } from "../../../shared/types";
+import "./AppRepo.css";
+import { AppRepoForm } from "./AppRepoForm.v2";
+
+interface IAppRepoAddButtonProps {
+  namespace: string;
+  kubeappsNamespace: string;
+  text?: string;
+  primary?: boolean;
+  repo?: IAppRepository;
+  secret?: ISecret;
+}
+
+export function AppRepoAddButton({
+  text,
+  primary,
+  namespace,
+  kubeappsNamespace,
+  repo,
+  secret,
+}: IAppRepoAddButtonProps) {
+  const dispatch: ThunkDispatch<IStoreState, null, Action> = useDispatch();
+  const [modalIsOpen, setModalOpen] = useState(false);
+  const openModal = () => setModalOpen(true);
+  const closeModal = () => setModalOpen(false);
+  const { errors } = useSelector((state: IStoreState) => state.repos);
+  // Primary by default
+  const isPrimary = primary !== false;
+
+  const onSubmit = (
+    name: string,
+    url: string,
+    authHeader: string,
+    customCA: string,
+    syncJobPodTemplate: string,
+    registrySecrets: string[],
+  ) =>
+    dispatch(
+      actions.repos.updateRepo(
+        name,
+        namespace,
+        url,
+        authHeader,
+        customCA,
+        syncJobPodTemplate,
+        registrySecrets,
+      ),
+    );
+
+  return (
+    <>
+      <CdsButton onClick={openModal} action={isPrimary ? "solid" : "outline"}>
+        {isPrimary ? <CdsIcon shape="plus-circle" inverse={true} /> : <></>}{" "}
+        {text || "Add App Repository"}
+      </CdsButton>
+      <Modal showModal={modalIsOpen} onModalClose={closeModal}>
+        {errors.create && (
+          <Alert theme="danger">
+            Found an error creating the repository: {errors.create.message}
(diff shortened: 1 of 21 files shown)


dashboard/src/components/Config/AppRepoList/AppRepoButton.v2.tsx
```diff
@@ -0,0 +1,82 @@
...
+  repo,
+  secret,
+}: IAppRepoAddButtonProps) {
+  const dispatch: ThunkDispatch<IStoreState, null, Action> = useDispatch();
+  const [modalIsOpen, setModalOpen] = useState(false);
+  const openModal = () => setModalOpen(true);
+  const closeModal = () => setModalOpen(false);
+  const { errors } = useSelector((state: IStoreState) => state.repos);
+  // Primary by default
+  const isPrimary = primary !== false;
```
absoludity, line 37: This looks like it's handling the possibility of an `undefined` for primary as it's optional. In which case, why not just define the default above in the function args as `primary = true, ...` (see https://stackoverflow.com/a/32596200 )

dashboard/src/components/Config/AppRepoList/AppRepoButton.v2.tsx
```diff
@@ -0,0 +1,82 @@
...
+  return (
+    <>
+      <CdsButton onClick={openModal} action={isPrimary ? "solid" : "outline"}>
+        {isPrimary ? <CdsIcon shape="plus-circle" inverse={true} /> : <></>}{" "}
+        {text || "Add App Repository"}
+      </CdsButton>
+      <Modal showModal={modalIsOpen} onModalClose={closeModal}>
+        {errors.create && (
+          <Alert theme="danger">
+            Found an error creating the repository: {errors.create.message}
```
absoludity, line 68: "An error occurred while creating the repository: {...}"?

dashboard/src/components/Config/AppRepoList/AppRepoList.v2.tsx
```diff
@@ -0,0 +1,125 @@
...
+            currently create a private AppRepository for a particular namespace of an additional
+            cluster. We may in the future support AppRepositories on additional clusters but for now
+            you will need to switch back to your default cluster.
+          </p>
+        </Alert>
+      ) : (
+        <>
+          {errors.fetch && (
+            <Alert theme="danger">
+              Found an error fetching repositories: {errors.fetch.message}
```
absoludity, line 69: I'm guessing you're just re-using the existing messages, so fine to leave as is, but normally an application doesn't find an error, an error occurs... hence suggesting "An error occurred while fetching repositories: {...}" etc.

dashboard/src/components/Config/AppRepoList/AppRepoRefreshAllButton.v2.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { useDispatch, useSelector } from "react-redux";
+import { IStoreState } from "shared/types";
+
+export function AppRepoRefreshAllButton() {
+  const [refreshing, setRefreshing] = useState(false);
+  const { repos } = useSelector((state: IStoreState) => state.repos);
+  const dispatch = useDispatch();
+
+  const handleResyncAllClick = async () => {
+    // Fake timeout to show progress
```
absoludity, line 14: Why are we faking showing progress when the actual repo is being refreshed? IMO either we shouldn't pretend and instead just use a user info "Repository refresh requested. It may take x minutes to complete.", or it should show actual progress? Again, fine without, just wondering what the longer-term plan is for this "progress".

dashboard/src/shared/specs/mountWrapper.tsx
```diff
@@ -31,7 +36,7 @@ export const defaultStore = mockStore(initialState);
 // getStore returns a store initialised with a merge of
 // the initial state with any passed extra state.
 export const getStore = (extraState: object) => {
-  const state = { ...initialState };
+  const state = cloneDeep(initialState);
```
absoludity, line 39: Ouch - did you find that by tests writing state then seen in other tests?

absoludity (approved):
Great, just a few questions, no blockers.

dashboard/src/shared/specs/mountWrapper.tsx
andresmgot: Yep, `{...object}` only performs a shallow copy, keeping references to inner objects:

```
> a = {a: 1, b: {c: 2}}
{ a: 1, b: { c: 2 } }
> b = {...a}
{ a: 1, b: { c: 2 } }
> b.b.c = 3
3
> a
{ a: 1, b: { c: 3 } }
```

absoludity: Yep yep, I realise, but also realised that I wrote that code the other week and was saying "ouch" because you may have had to find that bug the hard way (ie. tests failing because of contaminated test data)... hopefully you just noticed it while passing by and didn't suffer as a result :)

andresmgot: sorry, I found this with weird test results :P but it was easy to spot

PR #1926, opened 2020-08-06
Ensure chart will not render if additional clusters configured without oidc

### LINKED ISSUE #1579: Cannot use external auth proxy service without patching config
As mentioned by @plnordquist and summarised in [this comment](https://github.com/kubeapps/kubeapps/issues/1557#issuecomment-596844913), Kubeapps works fine with the default auth proxy service built into the chart, but if you setup your own auth proxy (say, for eg, if you don't want to use oauth2-proxy) and so set the chart value `authProxyEnabled` to false, then Kubeapps also currently [blanks out the `oauthLoginURI`](https://github.com/kubeapps/kubeapps/blob/master/dashboard/src/containers/LoginFormContainer/LoginFormContainer.ts#L17) in the Login form, so that login fails.

Currently, to work around this, you need to deploy with `authProxyEnabled` to false so the chart doesnt' deploy its own auth proxy, but then manually update the dashboard config map to set it to true, so the dashboard allows login.

Note: It may or may not be as simple as removing the condition which blanks the `oauthLoginURI`. I'm guessing there's a reason we did that but failed to leave a useful comment.

### LINKED ISSUE #1917: [multi-cluster] Fail installation if non-oidc clusters configured
### Description:

We don't support multi-cluster with non-oidc clusters (ie. using service token authn) so need to ensure we fail with an appropriate message when people try (whether that's during chart installation, or re-configuration).

See [thread](https://github.com/kubeapps/kubeapps/pull/1879#discussion_r457879441).

Part of #1753

absoludity:
### Description of the change

Ensures that the chart won't render if additional clusters are configured but the oidc auth proxy is not configured.

If people use an externally configured auth proxy, we also want to enable configuring additional clusters.

Note that since we switched to oauth2-proxy (well, in particular, started supporting http-only cookies for auth), people have experienced issues with using an external auth proxy, as the frontend assumed cookie authn iff `.Values.authProxy.enabled` (see #1579). The new `.Values.authProxy.externallyEnabled` means we can trivially fix that in the frontend also, as done below.

So, configuring additional clusters but not using oidc fails:
```
$ helm install kubeapps ./chart/kubeapps --namespace kubeapps --values ./docs/user/manifests/kubeapps-local-dev-additional-kind-cluster.yaml --set useHelm3=true

Error: template: kubeapps/templates/kubeapps-frontend-deployment.yaml:88:15: executing "kubeapps/templates/kubeapps-frontend-deployment.yaml" at <fail "additionalClusters can be configured only when using an authenticate proxy for cluster oidc authentication.">: error calling fail: additionalClusters can be configured only when using an authenticate proxy for cluster oidc authentication.
```

Install works fine with the `make deploy-dev` target that configures the internal oauth2-proxy running on the frontend. Additionally, the new `authProxy.externallyEnabled` value will also enable additional clusters:

```
$ helm install kubeapps ./chart/kubeapps --namespace kubeapps --values ./docs/user/manifests/kubeapps-local-dev-additional-kind-cluster.yaml --set useHelm3=true --set authProxy.externallyEnabled=true
...
(install works fine and login page assumes the oauth redirect urls as expe8cted)

```

### Benefits

We can now ensure people don't install Kubeapps with multi-cluster configured without an auth proxy.

Also, people can now use an external auth proxy without manual updating of config maps.

### Applicable issues

  - fixes #1917 
  - fixes #1579

### DIFF
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -644,6 +644,9 @@ testImage:
 authProxy:
   # Set to true to enable the OIDC proxy
   enabled: false
+  # Set to true if an external auth proxy is setup to provide cookie authentication
+  # at the oauthLoginURI and oauthLogoutURI values below.
+  externallyEnabled: false
   ## Bitnami OAuth2 Proxy image
   ## ref: https://hub.docker.com/r/bitnami/oauth2-proxy/tags/
   ##
--- a/chart/kubeapps/templates/dashboard-config.yaml
+++ b/chart/kubeapps/templates/dashboard-config.yaml
@@ -41,7 +41,7 @@ data:
     {
       "namespace": "{{ .Release.Namespace }}",
       "appVersion": "{{ .Chart.AppVersion }}",
-      "authProxyEnabled": {{ .Values.authProxy.enabled }},
+      "authProxyEnabled": {{ or .Values.authProxy.enabled .Values.authProxy.externallyEnabled }},
       "oauthLoginURI": {{ .Values.authProxy.oauthLoginURI | quote }},
       "oauthLogoutURI": {{ .Values.authProxy.oauthLogoutURI | quote }},
       "featureFlags": {{ .Values.featureFlags | toJson }}
--- a/chart/kubeapps/templates/kubeapps-frontend-deployment.yaml
+++ b/chart/kubeapps/templates/kubeapps-frontend-deployment.yaml
@@ -83,6 +83,8 @@ spec:
           {{- if .Values.authProxy.resources }}
           resources: {{- toYaml .Values.authProxy.resources | nindent 12 }}
           {{- end }}
+        {{- else }}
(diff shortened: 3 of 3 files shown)


chart/kubeapps/values.yaml
```diff
@@ -644,6 +644,9 @@ testImage:
 authProxy:
   # Set to true to enable the OIDC proxy
   enabled: false
+  # Set to true if an external auth proxy is setup to provide cookie authentication
+  # at the oauthLoginURI and oauthLogoutURI values below.
```
andresmgot, line 648: the comment's are not that clear, people may enable both `authProxy.enabled` and `authProxy.externallyEnabled` (mostly because we don't say anything about not setting `enabled` to true). I think the UX would be better if the params are:

```
enabled: false
deployOauth2Proxy: true
```

so if people want to use their own auth proxy they disable that. WDYT?

absoludity: I think it's only the history that makes this tricky. Let me open a new PR with better comments and discuss this there.

Issue #1894, opened 2020-07-28
[multi-cluster] Update remaining non-multi-cluster routes and provide info


absoludity:
There are some routes remaining which have not been updated to include the cluster in the route (such as service brokers, operators).

Similar to #1893 and #1897 which update the app repo config route, we'll need to update the remaining ones and define the behaviour for clusters other than the default (ie. whether we support the functionality, or provide info why it's only available on the default kubeapps cluster).

- [x] Address some [feedback on the info msg](https://github.com/kubeapps/kubeapps/pull/1897#issuecomment-666983924).
- [ ] Update service broker routes
- [ ] Update operator routes

PR #1929, opened 2020-08-07
Ensure service brokers are default cluster only

absoludity:
### Description of the change

Updates the service broker configuration to be clear its only supported on the default cluster:

1894-service-brokers

While there, also included a commit addressing the [feedback from Andres](https://github.com/kubeapps/kubeapps/pull/1897#issuecomment-666983924) from the similar app repository info

1894-app-repositories

### Benefits

Users who switch clusters when viewing the service broker config won't mistakenly think it's supported there.

### Applicable issues

  - Ref #1894

### DIFF
--- a/dashboard/src/components/Config/ServiceBrokerList/ServiceBrokerList.test.tsx
+++ b/dashboard/src/components/Config/ServiceBrokerList/ServiceBrokerList.test.tsx
@@ -5,6 +5,7 @@ import * as React from "react";
 import ServiceBrokerList from ".";
 import {
   ErrorSelector,
+  MessageAlert,
   ServiceBrokersNotFoundAlert,
   ServiceCatalogNotInstalledAlert,
 } from "../../../components/ErrorAlert";
@@ -19,6 +20,7 @@ let defaultProps = {
   errors: {},
   checkCatalogInstalled: jest.fn(),
   isInstalled: true,
+  cluster: "default",
 };
 
 beforeEach(() => {
@@ -30,6 +32,7 @@ beforeEach(() => {
     errors: {},
     checkCatalogInstalled: jest.fn(),
     isInstalled: true,
+    cluster: "default",
   };
 });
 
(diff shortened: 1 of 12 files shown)


dashboard/src/components/Config/ServiceBrokerList/ServiceBrokerList.test.tsx
```diff
@@ -52,6 +55,18 @@ context("if the service broker is not installed", () => {
   });
 });
 
+context("if the service brokers are accessed on an additional cluster", () => {
+  it("shows an alert with info", () => {
```
fjagugar, line 59: I would find useful to also test the opposite: If it is accessed on the default cluster, the MessageAlert is not present.

I see that later on we have a test that is comparing the rendered page with  snapshot. Is it expected that test to cover the mentioned "opposite test"?

dashboard/src/components/Header/Header.tsx
```diff
@@ -139,7 +139,9 @@ class Header extends React.Component<IHeaderProps, IHeaderState> {
                         <NavLink to={reposPath}>App Repositories</NavLink>
                       </li>
                       <li role="none">
-                        <NavLink to="/config/brokers">Service Brokers</NavLink>
+                        <NavLink to={app.config.brokers(clusters.currentCluster)}>
+                          Service Brokers
+                        </NavLink>
```
fjagugar, line 144: Should this link be displayed only if the currentCluster is the default one or is it actually expected that users reach this endpoint in order to get the error message and understand why is not available?

dashboard/src/components/Config/ServiceBrokerList/ServiceBrokerList.test.tsx
absoludity: > I would find useful to also test the opposite: If it is accessed on the default cluster, the MessageAlert is not present.

+1, added.


> 
> I see that later on we have a test that is comparing the rendered page with snapshot. Is it expected that test to cover the mentioned "opposite test"?

Well it would, but I don't like relying on snapshots for covering unit tests either - so I agree with you and just added a unit test instead.

dashboard/src/components/Header/Header.tsx
absoludity: Yes, it is expected that the menu item is there (when service brokers are configured), rather than not displaying options for reasons which are not clear to the user (seems to be Kubeapps UX strategy - in form the user of what RBAC is missing).

That said, I did land an earlier PR which does not display this menu option unless service brokers are configured.

PR #1930, opened 2020-08-07
Only show the Service Instances menu if the broker is configured.

### LINKED ISSUE #605: Feature/addon disablement option
Currently, Kubeapps focused in two main feature sets:

- Deploy and manage applications using Helm.
- Integration with k8s service catalog.

These two features are exposed via menu entries on the top navbar.
selection_684

I'd like to open the discussion of if we should allow users to "enable/disable" addons, being the first one the service catalog.

My take on this is that it makes sense to offer this functionality because:

* While deploying applications is the core of the dashboard, service catalog is an additional, optional capability. 
* It is not supported in all platforms, for example EKS does not support it yet.
* In some cases you might not be interested in any of the existing brokers or on this feature at all.
* Although we are working on improving it, this feature still needs some polishing.

Enabling/Disabling features could be done initially as part of a setting (enabled by default so we do not change the current behavior) in the values.yaml file so it can be tweaked at deployment time. In the future, we might want to evaluate a different way in which an admin enable/disables addons.

What do you think?

absoludity:
### Description of the change

Updates the main menu to display the Service Instances (alpha?!) menu item only if a broker has been configured.

I think this option won't even be visible in the new UI. I'm only doing this here so that I don't need to make all the routes for service instances cluster aware, plus we had a recent request along these lines on the related issue (and it'll be a few weeks yet before the new UI is complete).

### Benefits

Users don't see a useless menu entry when service brokers have not even been configured.

### Applicable issues

  - fixes #605

### DIFF
--- a/dashboard/src/components/Header/Header.test.tsx
+++ b/dashboard/src/components/Header/Header.test.tsx
@@ -25,9 +25,29 @@ const defaultProps = {
   getNamespace: jest.fn(),
   featureFlags: { operators: false, additionalClusters: [], ui: "hex" },
   appVersion: "",
+  isServiceCatalogInstalled: false,
 };
 it("renders the header links and titles", () => {
   const wrapper = shallow(<Header {...defaultProps} />);
+  const menubar = wrapper.find(".header__nav__menu").first();
+  const items = menubar.children().map(p => p.props().children.props);
+  const expectedItems = [
+    { children: "Applications", to: app.apps.list("default", "default") },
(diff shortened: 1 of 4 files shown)


andresmgot:
I guess this is okay, we should update a bit the docs under `user/service-catalog.md` to just point to install the catalog using Helm. Also, we can delete this component: dashboard/src/components/ErrorAlert/ServiceCatalogNotInstalledAlert.tsx

PR #1932, opened 2020-08-07
Update operator instance routes

### LINKED ISSUE #1894: [multi-cluster] Update remaining non-multi-cluster routes and provide info

absoludity:
### Description of the change

Updates operator instance routes and adds relevant info when cluster switched from default.

### Applicable issues

  - fixes #1894 
Follows #1931

### DIFF
--- a/dashboard/src/components/AppList/__snapshots__/AppList.test.tsx.snap
+++ b/dashboard/src/components/AppList/__snapshots__/AppList.test.tsx.snap
@@ -136,6 +136,7 @@ exports[`when custom resources available matches the snapshot 1`] = `
       <div>
         <CardGrid>
           <CustomResourceListItem
+            cluster="defaultc"
             csv={
               Object {
                 "spec": Object {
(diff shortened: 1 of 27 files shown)


dashboard/src/components/AppList/__snapshots__/AppList.test.tsx.snap
```diff
@@ -136,6 +136,7 @@ exports[`when custom resources available matches the snapshot 1`] = `
       <div>
         <CardGrid>
           <CustomResourceListItem
+            cluster="defaultc"
```
aruiz14, line 139: Is this a typo?

absoludity: No - all the tests were already using `default` for the default namespace value, so I wanted to use something different for the default cluster value for comparisons in tests. Perhaps I'll switch to something more explicit `default-cluster` in future tests (and update this one).

Issue #1933, opened 2020-08-10
[multi-cluster] Support per-cluster oidc clientids


absoludity:
### Description:

Some multi-cluster environments, such as Tanzu Kubernetes Grid, recommend configuring Dex with a different client-id per cluster (see, for example, the static clients section of the [dex config specified when enabling gangway](https://docs.vmware.com/en/VMware-Tanzu-Kubernetes-Grid/1.0/vmware-tanzu-kubernetes-grid-10/GUID-manage-instance-gangway-aws.html)).

Most oidc proxies (oauth2-proxy, vouch, keycloak-gatekeeper) are designed to proxy for a single oauth client only. Kubeapps default oidc setup uses oauth2 and so has this same limitation.

Currenty it's not possible to configure Kubeapps with separate client ids per cluster in the multi-cluster configuration.

The main result of this is that we'll need to ensure authn can be per cluster (as each JWT is signed using the client secret, which will be different for each cluster). As a result, we can't assume a single auth proxy (configured with a single client).

Plan for investigation
- [x] Try [istio's identity and access adapter](https://github.com/istio-ecosystem/authservice) which allows different client IDs for different routes.
- [ ] Update chart config so that frontend is always the UI endpoint, rather than the authProxy when enabled (as we can't assume a single one)
- [x] Update login URI to be cluster specific (though default can still work for all if a separate clientID not used)
- [ ] When switching clusters in the UI, update to be able login to selected cluster if not already logged in (ie. when browser doesn't send the required cookie) rather than logging out.

Workaround: Scott mentioned that he can update the provider so that it just uses the one client id for all clusters to work around this issue, but we'd prefer to find a solution that works ootb with TKG's default extensions.

absoludity:
> Most oidc proxies (oauth2-proxy, vouch, keycloak-gatekeeper) are designed to proxy for a single oauth client only. Kubeapps default oidc setup uses oauth2 and so has this same limitation.

The simplest setup here would be to have multiple instances of the oauth2-proxy running on the frontend, though longer-term it would be nicer to have a single process (as it won't scale well for many clusters).

absoludity:
istio's authservice is dependent on istio infrastructure (there is no default way to redirect traffic through the proxy to the actual service, other than creating an [auth policy](https://github.com/istio-ecosystem/authservice/blob/master/bookinfo-example/config/bookinfo-authn-policy-template.yaml), afaict.

absoludity:
[...]

PR #1934, opened 2020-08-11
Update the login URL to be cluster aware.

absoludity:
### Description of the change

Updates the login URL to be cluster aware.

### Benefits

Some multicluster setups configure different `--oidc-client-id`'s per cluster, so the URL for login needs to include the cluster to know which configuration to use when authenticating. 

This should not affect the generic multi-cluster setup where a single client id is used for the cluster api servers.

### Applicable issues

  - Ref #1933


absoludity:
Turns out I don't need this at all.

PR #1936, opened 2020-08-12
Switch the default for IPv6 support to false.

absoludity:
In  #1910 this flag was added but defaulting to true. This has caught out a number of people as [dual stack IPv4/IPv6 is still alpha in kubernetes](https://kubernetes.io/docs/concepts/services-networking/dual-stack/), and not supported on some standard cluster setups (eg. Tanzu Kubernetes Grid).


absoludity:
Thanks @alemorcuq

PR #1938, opened 2020-08-13
Support local dev setup of clientid-per-cluster.

absoludity:
### Description of the change

Updates devel environment to support a setup where the additional cluster uses its own oauth2 clientid.

### Benefits

Allows local testing of a required cluster configuration.

### Applicable issues

  - ref #1933 

### Additional information

I've not made this the default setup for multicluster until we know more about longer-term authn for setups like TKG.

This solution (an nginx+oauth2-proxy in front of the kubeapps frontend) was the second attempt, after trying first to use istio's authproxy.


absoludity:
Closing in favour of #1939

Issue #1571, opened 2020-03-09
Kubeapps Helm Chart does not allow configuring NGINX for large headers


plnordquist:
### Description:

With the current configuration of the kubeapps Helm chart, I am unable to configure the nginx serving the dashboard and the frontend nginx to support large headers with oauth2-proxy.

### Steps to reproduce the issue:

1. Configure an OIDC provider for Kubernetes, in my case Keycloak
1. Deploy the kubeapps helm chart with oauth2-proxy configured to talk to Keycloak
2. Attempt to load kubeapps in the browser

### Describe the results you received:

NGINX returns a 400 Bad Request Request Header Or Cookie Too Large

### Describe the results you expected:

The application loads properly.

### Additional information you deem important (e.g. issue happens only occasionally):

My Keycloak is configured to add group membership to the token and that's how the tokens are so large.  The workaround I currently use is to add `large_client_header_buffers 4 32k;` to the vhost.conf file inside the configmaps that control nginx.  For the `kubeapps-frontend-config` configmap I've placed the directive above the server block in the vhost.conf file.  For the `kubeapps-internal-dashboard-config` configmap I've placed the directive above inside the server block above the location block in the vhost.conf file.  I'm also running the oauth2-proxy outside of the Helm chart and I've customized the deployment as referenced here https://github.com/kubeapps/kubeapps/issues/1557#issuecomment-596741738.  I can try deploying the chart with oauth2-proxy inside the chart if it's necessary to test this issue fully.

### Version of Helm, Kubeapps and Kubernetes:

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.1.1", GitCommit:"afe70585407b420d0097d07b21c47dc511525ac8", GitTreeState:"clean", GoVersion:"go1.13.8"}
```

- Output of `helm list <kubeapps-release-name>`:

```
NAME                    NAMESPACE       REVISION        UPDATED                                 STATUS          CHART                           APP VERSION
kubeapps                services        14              2020-03-06 11:48:36.334056678 -0800 PST deployed        kubeapps-3.4.3                  v1.8.2
```

- Output of `kubectl version`:

```
[...]

PR #1940, opened 2020-08-17
Ensure local ingress-nginx has larger proxy-buffer sizes.

absoludity:
### Description of the change

It turns out the issue with LDAP login I'd found on the previous #1939 wasn't specific to my own env, as @vrabbi also gets a 500 when logging in via LDAP on a TKG cluster.

I found the issue in my local ingres-nginx logs:

```
2020/08/17 00:56:00 [error] 50#50: *76751 upstream sent too big header while reading response header from upstream, client: 172.18.0.1, server: localhost, request: "GET /oauth2/callback?code=of4d542tl3zhphasig2cuvisx&state=0c27e351936a66b7540a438a745687ba%3A%2F HTTP/2.0", upstream: "http://10.244.0.15:3000/oauth2/callback?code=of4d542tl3zhphasig2cuvisx&state=0c27e351936a66b7540a438a745687ba%3A%2F", host: "localhost", referrer: "https://172.18.0.2:32000/auth/ldap?req=ahe3hmcoa2u367pprrdv2v6qh"
172.18.0.1 - - [17/Aug/2020:00:56:00 +0000] "GET /oauth2/callback?code=of4d542tl3zhphasig2cuvisx&state=0c27e351936a66b7540a438a745687ba%3A%2F HTTP/2.0" 502 157 "https://172.18.0.2:32000/auth/ldap?req=ahe3hmcoa2u367pprrdv2v6qh" "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:79.0) Gecko/20100101 Firefox/79.0" 170 0.011 [kubeapps-kubeapps-http] [] 10.244.0.15:3000 0 0.012 502 c75dfa925e4f6d837cff3defbf8d81cd
```

So it appears the cookie created by auth-proxy, when using LDAP, is too large. After some frustrating trials using the normal settings mentioned in other issues (#1571 and others mention `large-client-header-buffers`), I found they didn't help at all, while the proxy-buffer settings did. Setting to 8k works fine.

Surprisingly, I didn't need to modify similar / related settings on the kubeapps frontend or dashboard nginx configs.

### Benefits

Can now login with LDAP over TLS using oauth2-proxy.

### Possible drawbacks

I'm not sure whether the similar setting can be tweaked on TKG, @vrabbi?

### Applicable issues

Follows #1939, ref #1933 

### Additional information

I tried to switch to use the bitnami ingress nginx chart, but it's not obvious what settings are needed specifically for kind. Since we can pull a specific deployment config for kind and update with the required settings, I've just done that for now.


absoludity:
> isn't it better to install the ingress controller with helm?

Yep, see the Additional information above. I need to find out the specific Kind changes to be able to use the helm chart.

BTW, I accidentally closed this PR by deleting the branch it was based on (after landing that), so have recreated in #1944 which I'll land instead (same PR, just rebased on master).

PR #1945, opened 2020-08-18
Improve wording and semantics for external oauth proxy support.

absoludity:
### Description of the change

On #1926 Andres mentioned:

> the comment's are not that clear, people may enable both `authProxy.enabled` and `authProxy.externallyEnabled` (mostly because we don't say anything about not setting `enabled` to true). I think the UX would be better if the params are:
> 
> ```
> enabled: false
> deployOauth2Proxy: true
> ```
> 
> so if people want to use their own auth proxy they disable that. WDYT?

I think in both my original PR and your suggestion above, the confusion is that it becomes ambiguous exactly what the `authProxy` section of the values is controlling (an init container resource? enabling an externally defined auth proxy? configuring oauth login/logout URIs for the dashboard?). And it's hard to find a way which doesn't have some ambiguity without breaking the existing options. If we were starting from scratch, I'd decouple the dashboard configuration from any auth proxy deployment config, but we don't have that option, so what I've suggested here is to at least decouple the internal authProxy configuration from the option for an external auth proxy, and updated the comments to be clearer. 

See what you think. As above, if we were breaking existing options, I'd also move the `authProxy,oauth{Login|Logout}URI`s into the `dashboard: ` config. Hopefully this should be clearer that you either use the internal authProxy or indicate you're using an external one. I can add a condition so it fails if you set both to true, but I'm not sure that's necessary or a good idea (imagining a scenario where Kubeapps is exposed via two networks, one internal, one external with ingress - using different clients of the same auth provider or some bizarre setup).

### DIFF
--- a/chart/kubeapps/templates/dashboard-config.yaml
+++ b/chart/kubeapps/templates/dashboard-config.yaml
@@ -41,7 +41,7 @@ data:
     {
       "namespace": "{{ .Release.Namespace }}",
       "appVersion": "{{ .Chart.AppVersion }}",
-      "authProxyEnabled": {{ or .Values.authProxy.enabled .Values.externalAuthProxy }},
+      "authProxyEnabled": {{ .Values.authProxy.enabled }},
       "oauthLoginURI": {{ .Values.authProxy.oauthLoginURI | quote }},
       "oauthLogoutURI": {{ .Values.authProxy.oauthLogoutURI | quote }},
       "featureFlags": {{ .Values.featureFlags | toJson }},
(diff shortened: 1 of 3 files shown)


absoludity:
> Got your point but it's true that it's confusing to have the parameter "external" out of the auth proxy configuration (mostly because of the URIs). What if we let it as:
> 
> ```
> authProxy:
>    # Comment
>   enabled: false
>   # Comment
>   external: false
> ```
> 
> and then, we only deploy the authProxy if `enabled == true && external == false`? That would be more readable and safe for the user.

OK, updated to that - it's a good compromise :) Please just re-check the comments for me as I had to reword them all.

andresmgot (commented):
New comments look good to me :+1: Thanks!

Issue #1764, opened 2020-06-03
[multi-cluster] Document, demo and publicise multi-cluster availability


absoludity:
To remove the feature flag we'd want to have at least the first item below complete.

Parent task: #1753 

- [x] Document the multi-cluster support with example configuration etc.
- [x] Move multi-cluster config out of featureFlags
- [x] Record demo running on TKG clusters
- [x] Support docs team to get blog post published including demo and docs.
- [x] Improve scripts for testing multi-cluster locally

Issue #1942, opened 2020-08-17
Support running Kubeapps on a TKG Management Cluster


vrabbi:
### Description:

After discussing at length the multi cluster support and adapting it to TKG with @absoludity i believe there is still one key functionality that should be added in the multi cluster area to fully support TKG.

In TKG, the management Cluster is not OIDC enabled, however the Workload Clusters are.
Within the TKG Management Cluster it is suggested to run shared services that provide a unified service toolkit for the workload clusters. this currently is scoped OOTB to include Dex and Harbor (if using TKG+ licensing).

Adding support for running in a non OIDC enabled cluster would be a huge advantage as that would allow kubeapps to follow the TKG methodology and architecture.

### Additional information you deem important (e.g. issue happens only occasionally):

possible solutions i have thought of are:
#### option 1 ####
add 2 additional variables to the helm chart:
1. disableLocalCluster - a boolean that defaults to false and can be set in conjunction with configuring additional Clusters and adding the defaultCluster variable.
2. defaultCluster - the name of the cluster from the additional clusters that should be used as the default cluster when logging in.

#### option 2 ####
add 1 additonal variable to the helm chart:
1. disableLocalCluster - a boolean disabling the local cluster.
in this option there would be no defaultCluster rather there would be a homepage indicating the user must select a cluster. from the context list.

#### Pros and Cons ####
option 1 is probably easier to implement but has 2 main drawbacks:
1. As workload clusters are dynamic and come up and down in certain environments, having a default cluster may prove challenging for some customers.
2. overall in a multi cluster solution you run the risk of customers provisioning charts to the wrong cluster

option 2 is probaly harder to implement but has a few key benefits:
1. makes the user select the cluster which provides an extra security guard around users provisioning charts in the wrong cluster
2. allows kubeapps to be deployed in the management cluster before any additional clusters are deployed and allows them to be more dynamic without making many changes to the kubeapps deployment. (only need to add the scopes and the cluster details in the additional clusters array)

absoludity:
So a couple of thoughts on this:

I've left a bit of room to update the configuration semantics, in particular, I don't know that we need a `disableLocalCluster` option at all if we instead update the `clusters` option to be definitive, that is, if we assume that it defines all the clusters available as targets for installations - including the cluster on which Kubeapps is installed.

Including the cluster on which Kubeapps is installed is just a cluster without an `apiServiceURL`, so for example:
```
clusters:
  - name: default
  - name: additional-1
[...]

PR #1947, opened 2020-08-18
Move multi-cluster config from featureFlags.

absoludity:
### Description of the change

Moves the multi-cluster configuration out of the `featureFlags` to the root `.Values.clusters` and properly documents.

Note that the change from `additionalClusters` to `clusters` is intentional as per the comment and related issue.

### Benefits

I can now document the setup without a featureFlag :)

### Possible drawbacks

This is mostly a mechanical change in the frontend, so the whole config is sent. I'm going to follow-up with another PR to only send the name and apiurl, (ie. not send the CAdata unnecessarily, or any serviceToken which is just sending sensitive data unnecessarily to the frontend).

### Applicable issues

Ref: #1764


chart/kubeapps/templates/kubeops-config.yaml
```diff
@@ -10,5 +10,5 @@ metadata:
     heritage: {{ .Release.Service }}
 data:
   additional-clusters.conf: |-
```
andresmgot, line 12: should this be `clusters.conf`?

chart/kubeapps/values.yaml
```diff
@@ -19,6 +19,27 @@ allowNamespaceDiscovery: true
...
+## cluster on which Kubeapps is installed. In the future a populated list here will be definitive, so that
+## it is possible to configure Kubeapps to target other clusters only (ie. not target the cluster on which
+## Kubeapps is installed), as per https://github.com/kubeapps/kubeapps/issues/1942 .
+## The base64-encoded certificateAuthorityData can be obtained from the additional cluster's kube config
+## file, for example:
+## kubectl --kubeconfig ~/.kube/kind-config-kubeapps-additional config view --raw -o jsonpath='{.clusters[0].cluster.certificate-authority-data}'
+#
+# clusters:
+# - name: second-cluster
+#   apiServiceURL: https://second-cluster:6443
```
andresmgot, line 35: after renaming this to `clusters`, it's a bit confusing the sample data (which implies additional clusters)

chart/kubeapps/templates/kubeops-config.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.featureFlags.additionalClusters -}}
+{{- if .Values.clusters -}}
```
andresmgot, line 1: you may want to check the length of the array here rather than a truth-y value

chart/kubeapps/templates/kubeops-config.yaml
absoludity: It will be updated (and the related deployment template, kubeops args, etc.) , but I've left that for when I get to #1942 . It was important in this PR to do the chart change mainly - I included the frontend changes only because the multicluster config was initially passed through directly as part of the feature flag, otherwise it wasn't really special.

chart/kubeapps/values.yaml
absoludity: Yeah, that's what I tried to make explicit in the preceding comment? I'm renaming it now so that we don't release and then change the value name for #1942.

absoludity (commented):
Thanks

PR #1948, opened 2020-08-18
Sanitize the multicluster config when sending to frontend.

absoludity:
### Description of the change

Just sanitizes the templated output so it looks like:

```
  config.json: |-
    {
      "namespace": "kubeapps",
      "appVersion": "DEVEL",
      "authProxyEnabled": true,
      "oauthLoginURI": "/oauth2/start",
      "oauthLogoutURI": "/oauth2/sign_out",
      "featureFlags": {"invalidateCache":true,"operators":false,"ui":"hex"},
      "clusters": [{"apiServiceURL":"https://172.18.0.3:6443","name":"second-cluster"},{"apiServiceURL":"https://example.com:6443","name":"third-cluster"}]
    }
```

Not sure if there's a better way to filter a list of dicts?

### DIFF
--- a/chart/kubeapps/templates/dashboard-config.yaml
+++ b/chart/kubeapps/templates/dashboard-config.yaml
@@ -50,5 +50,4 @@ data:
         {{- $sanitizedClusters = append $sanitizedClusters (pick . "name" "apiServiceURL") }}
       {{- end }}
       "clusters": {{ $sanitizedClusters | toJson }}
-      ]
     }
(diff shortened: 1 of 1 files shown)


absoludity:
Converting to draft while fixing a dashboard issue.

absoludity:
> I am not sure what was the problem before? You wanted to avoid typos or is it because the output was not readable?

Mainly avoid sending sensitive info (the potential `serviceToken`) over the wire to the frontend to be stored in state there etc., when it's not needed at all on the frontend.

andresmgot:
> > I am not sure what was the problem before? You wanted to avoid typos or is it because the output was not readable?
> 
> Mainly avoid sending sensitive info (the potential `serviceToken`) over the wire to the frontend to be stored in state there etc., when it's not needed at all on the frontend.

Got it

PR #1949, opened 2020-08-19
Fix the version matcher for react-scripts so that serialize-javascript is updated.

absoludity:
### Description of the change

When updating from `react-scripts-ts` to `react-scripts` we missed that we'd [pinned the version matcher for react-scripts to `3.4.1` rather than `^3.4.1`](https://github.com/kubeapps/kubeapps/pull/1823/files#diff-8a93884670d48666f7da0af337381d92R126).

A side-effect of this is that we were still depending (in our dev env) on out-of-date packages.

This updates to use the correct matcher to get updated packages.

### DIFF
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -137,7 +137,7 @@
     "node-sass-chokidar": "^1.3.3",
     "npm-run-all": "^4.1.3",
     "prettier": "^1.10.2",
-    "react-scripts": "3.4.1",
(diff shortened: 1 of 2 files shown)


absoludity:
I'm going to land this straight upgrade so we can tag images etc., but let me know if you see an issue.

andresmgot (commented):
:+1:

PR #1957, opened 2020-08-21
Allow to list repositories using the backend

andresmgot:
### Description of the change

This is needed to show the list of both the app repositories in the global namespace and other namespaces. Note that in the case that the request is for the `kubeapps` namespace, the request is made using the backend serviceaccount. This is to allow unprivileged users to list global apprepositories.

### DIFF
--- a/pkg/kube/kube_handler_test.go
+++ b/pkg/kube/kube_handler_test.go
@@ -331,6 +331,58 @@ func TestAppRepositoryCreate(t *testing.T) {
 	}
 }
 
+func TestAppRepositoryList(t *testing.T) {
+	testCases := []struct {
+		name             string
+		requestNamespace string
+		existingRepos    map[string][]repoStub
+	}{
+		{
+			name:             "it gets repos from the global namespace",
+			requestNamespace: kubeappsNamespace,
+			existingRepos: map[string][]repoStub{
+				"kubeapps": {repoStub{name: "test-repo"}},
+			},
+		},
+		{
+			name:             "it gets repos from a namespace",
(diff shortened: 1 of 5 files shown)


pkg/kube/kube_handler_test.go
```diff
@@ -331,6 +331,58 @@ func TestAppRepositoryCreate(t *testing.T) {
...
+			handler := userHandler{
+				kubeappsNamespace: kubeappsNamespace,
+				svcClientset:      cs,
+			}
+			if tc.requestNamespace != kubeappsNamespace {
+				handler = userHandler{
+					kubeappsNamespace: kubeappsNamespace,
+					clientset:         cs,
+				}
+			}
```
absoludity, line 372: This seems backwards? The test should not be setting either the `svcClientset` or the `clientset`, but rather, it should set both (perhaps with different fake data?) and ensuring that the code uses the correct handler.

Hmm, actually, I think I see your intention: the code would error here if it chose the incorrect clientset, so you're setting either one or the other to ensure an error is raised if the incorrect code path went? It's a bit fragile, but ok - not backwards as I initially thought. Perhaps a comment so the intention is clear, if that is the intention?

absoludity (approved):
Great to see one more direct query from the UI to api server removed :)

pkg/kube/kube_handler_test.go
andresmgot: Yes, you got the intention, the goal here is to fail if the test is trying to use the wrong clientset, will add a comment about that.

PR #1959, opened 2020-08-24
Split between global and namespaced repositories

andresmgot:
### Description of the change

Finish the AppRepoList view adding both global and namespaced repositories:

Screenshot from 2020-08-24 11-41-51

### Applicable issues

  - ref #1354

### DIFF
--- a/dashboard/src/components/Config/AppRepoList/AppRepoControl.v2.tsx
+++ b/dashboard/src/components/Config/AppRepoList/AppRepoControl.v2.tsx
@@ -33,7 +33,10 @@ export function AppRepoControl({
     return async () => {
       await dispatch(actions.repos.deleteRepo(repoName, repoNamespace));
       if (repoNamespace !== kubeappsNamespace) {
-        // Re-fetch repos in both namespaces
+        // Re-fetch repos in both namespaces because otherwise, the state
+        // will be updated only with the repos of repoNamespace and removing
+        // the global ones.
+        // TODO(andresmgot): This can be refactored once hex UI is dropped
         dispatch(actions.repos.fetchRepos(repoNamespace, kubeappsNamespace));
       } else {
         dispatch(actions.repos.fetchRepos(repoNamespace));
--- a/dashboard/src/components/Config/AppRepoList/AppRepoControl.v2.tsx
+++ b/dashboard/src/components/Config/AppRepoList/AppRepoControl.v2.tsx
@@ -2,11 +2,12 @@ import React, { useState } from "react";
 
 import { CdsButton } from "components/Clarity/clarity";
 import { useDispatch } from "react-redux";
-import { IAppRepository, ISecret } from "shared/types";
+import { Action } from "redux";
+import { ThunkDispatch } from "redux-thunk";
+import { IAppRepository, ISecret, IStoreState } from "shared/types";
 import actions from "../../../actions";
 import ConfirmDialog from "../../ConfirmDialog/ConfirmDialog.v2";
 import { AppRepoAddButton } from "./AppRepoButton.v2";
-
 import "./AppRepoControl.css";
 
 interface IAppRepoListItemProps {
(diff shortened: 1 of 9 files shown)


dashboard/src/components/Config/AppRepoList/AppRepoList.v2.tsx
```diff
@@ -84,41 +130,53 @@ function AppRepoList({ cluster, namespace, kubeappsNamespace }: IAppRepoListProp
...
+                    valign="center"
+                    columns={tableColumns}
+                    data={getTableData(globalRepos, namespace !== kubeappsNamespace)}
+                  />
+                ) : (
+                  <p>No global repositories found.</p>
+                )}
+                {namespace !== kubeappsNamespace && (
+                  <>
+                    <h3>Namespace Repositories</h3>
```
absoludity, line 157: I'm not sure about the terminology here. If you use the above, it should be "Namespaced Repositories", but I wonder if we can do better with `<h3>{namespace} Repositories</h3>`, which, next to "Global Repositories" would be clear, I think?

dashboard/src/components/Config/AppRepoList/AppRepoList.v2.tsx
```diff
@@ -84,41 +130,53 @@ function AppRepoList({ cluster, namespace, kubeappsNamespace }: IAppRepoListProp
...
+                    <p>
+                      Namespaced Repositories are only available in the current namespace. To change
```
absoludity, line 159: Similar here - this could read "The following Repositories are available in the { namespace } namespace only. ..."

dashboard/src/components/Config/AppRepoList/AppRepoList.v2.tsx
```diff
@@ -84,41 +130,53 @@ function AppRepoList({ cluster, namespace, kubeappsNamespace }: IAppRepoListProp
...
+                    </p>
+                    {namespaceRepos.length ? (
+                      <Table
+                        valign="center"
+                        columns={tableColumns}
+                        data={getTableData(namespaceRepos, false)}
+                      />
+                    ) : (
+                      <p>
+                        The current namespace doesn't have any repository. Click on the button "Add
```
absoludity, line 170: "repository" should be plural here, but could also use "The { namespace } namespace doesn't have any repositories. ..." ?

absoludity (approved):

dashboard/src/components/Config/AppRepoList/AppRepoList.v2.tsx
andresmgot: The thing that bugs me is having a lowercase at the beginning: `<h3>default Repositories</h3>`, I'd rather do something like Namespace Repositories: default

dashboard/src/components/Config/AppRepoList/AppRepoControl.v2.tsx
```diff
@@ -26,11 +27,17 @@ export function AppRepoControl({
...
-  const dispatch = useDispatch();
+  const dispatch: ThunkDispatch<IStoreState, null, Action> = useDispatch();
 
   const handleDeleteClick = (repoName: string, repoNamespace: string) => {
-    return () => {
-      dispatch(actions.repos.deleteRepo(repoName, repoNamespace));
+    return async () => {
+      await dispatch(actions.repos.deleteRepo(repoName, repoNamespace));
+      if (repoNamespace !== kubeappsNamespace) {
+        // Re-fetch repos in both namespaces
```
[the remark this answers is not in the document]
andresmgot: Yes, I am a bit limited because I don't want to change the state definition or the calls done from the current UI. The thing is that calling `fetchRepos` cleans the current state. So if in the state, there are repos from `ns1` and `ns2` and we re-fetch only `ns2`, the repos from `ns1` will dissapear.

Added a comment to clarify it.

dashboard/src/components/Config/AppRepoList/AppRepoList.v2.tsx
absoludity: Yeah, it would look a bit odd, unless it had formatting for code or similar (ie. `<h3><code>{ namespace}</code> Repositories<h3>`). Your one is great too. Note that in this case it should be singular: `Namespace Repository: default` (unless the intention is that there's more than one there, as it was with the global repositories), but not a problem.

andresmgot: mmh, in this case `default` is not the "namespace repository" what `Namespace Repositories: default` means is "namespaced repositories available in the default namespace" that's why I set it plural. Not sure if my English is right here?

absoludity: Oh right - yes, that's fine. I'd missed that we want to list multiple "namespaced" app repos for a single namespace, potentially. All good :)

PR #1961, opened 2020-08-24
Categories filter

### LINKED ISSUE #1960: Allow to filter by Chart Categories
Some charts (from Bitnami) now include an annotation for specify its category: https://github.com/bitnami/charts/blob/master/bitnami/apache/Chart.yaml#L22

We can use that info to build a filter and be able to filter by category.

andresmgot:
### Description of the change

Extend the `Chart` model in the backend to include the category of the chart and apply required changes in the dashboard to use those:

Screenshot from 2020-08-24 16-01-27

Note that chart and operator categories are used.

### Applicable issues

  - fixes #1960


absoludity (approved):
Great.

andresmgot:
> Great. Are there any migration issues we need to worry about here?

Nope, if a chart doesn't have annotations or the "category" key is empty, the `assetsvc` will return `category: ""` anyway, which is resolved as "Unkown" in the catalog.

PR #1965, opened 2020-08-25
Operator List v2

andresmgot:
### Description of the change

Operator List view for Clarity. These are different renders:

OLM not found:

Screenshot from 2020-08-25 11-58-49

Empty catalog:

Screenshot from 2020-08-25 11-59-05

Normal catalog:

Screenshot from 2020-08-25 12-14-10

### Possible drawbacks

For the moment, I haven't implemented the logic that shows if an operator is installed or not. The current logic checks if there are CSVs matching a PackageManifest, but when you install an operator, what's first installed is a Subscription. A controller watches subscriptions and eventually creates the CSV but is not immediate so someone may try to install an operator several times before it's available. This require new logic and API calls but is less important than having the operator flow working end to end, so I will get back to this later.

### Applicable issues

  - ref #1354


dashboard/src/components/OperatorList/OperatorList.v2.test.tsx
```diff
@@ -0,0 +1,262 @@
...
+        },
+      ],
+    },
+  },
+} as any;
+
+// itBehavesLike("aLoadingComponent", {
+//   component: OperatorList,
+//   props: { ...defaultProps, isFetching: true },
+// });
```
absoludity, line 89: Left over or meant to be uncommented?

dashboard/src/components/OperatorList/OperatorList.v2.test.tsx
```diff
@@ -0,0 +1,262 @@
...
+        operators: { isOLMInstalled: true, operators: [sampleOperator, sampleOperator2] },
+      }),
+      <OperatorList {...defaultProps} />,
+    );
+    expect(wrapper.find(InfoCard).length).toBe(2);
+
+    // Filter category "security"
+    const input = wrapper.find("input").findWhere(i => i.prop("value") === "security");
+    input.simulate("change", { target: { value: "security" } });
+    expect(wrapper.find(InfoCard).length).toBe(1);
```
absoludity, line 245: Should really verify that it's `sampleOperator` right? (I mean, verify that the one item is the correct one).

absoludity (approved):
lgtm

dashboard/src/components/OperatorList/OperatorList.v2.test.tsx
andresmgot: oops, yes, forgot to rewrite

dashboard/src/components/OperatorList/OperatorList.v2.test.tsx
andresmgot: right

PR #1967, opened 2020-08-26
Default to the container env proxy vars if none set from apprepo.

absoludity:
Just a follow-up commit that didn't make it into #1966 before it was merged.

### DIFF
--- a/pkg/kube/http_utils.go
+++ b/pkg/kube/http_utils.go
@@ -122,17 +122,26 @@ func InitNetClient(appRepo *v1alpha1.AppRepository, caCertSecret, authSecret *co
 func getProxyConfig(appRepo *v1alpha1.AppRepository) *httpproxy.Config {
 	template := appRepo.Spec.SyncJobPodTemplate
 	proxyConfig := httpproxy.Config{}
+	defaultToEnv := true
 	if len(template.Spec.Containers) > 0 {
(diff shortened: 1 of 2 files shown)


andresmgot (commented):
And now the tests are failing :)

absoludity:
> And now the tests are failing :)

Hah, that's interesting. It's failing because the cicircle env actually sets these vars and I hadn't cleared them for the test (only saved and overwrote the ones I was testing).

andresmgot (approved):

absoludity:
So I'm still not sure I want to land this PR. I'd originally done it because of your [original comment](https://github.com/kubeapps/kubeapps/pull/1966#discussion_r477105304), but after I clarified there, I'm not sure whether you still think it should land. I don't see why we would want this default to env behaviour now (yes, backwards compat, but the change in #1966 already ensures the original intention of #1097 without requiring the manual env ar setting). Let me know what you think.

andresmgot:
It doesn't hut, does it? Is there any drawback of merging this?

absoludity:
> It doesn't hut, does it? Is there any drawback of merging this?

Just extra code which in the future may cause confusion given it's not needed. But OK, I think you think it's worthwhile keeping the backwards compat in case someone is manually doing this. Merging.

PR #1971, opened 2020-08-27
Update dashboard components to use config.kubeappsCluster pt1

absoludity:
### Description of the change

Part 1 of updating the dashboard components to use the `kubeappsCluster` config and remove "default" assumptions.

### Benefits

Work towards #1942

### DIFF
--- a/dashboard/src/actions/repos.ts
+++ b/dashboard/src/actions/repos.ts
@@ -118,8 +118,9 @@ export const deleteRepo = (
   return async (dispatch, getState) => {
     const {
       clusters: {
+        currentCluster,
         clusters: {
-          default: { currentNamespace },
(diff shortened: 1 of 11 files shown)


dashboard/src/actions/repos.ts
```diff
@@ -118,8 +118,9 @@ export const deleteRepo = (
   return async (dispatch, getState) => {
     const {
       clusters: {
+        currentCluster,
         clusters: {
-          default: { currentNamespace },
+          [currentCluster]: { currentNamespace },
```
absoludity, line 123: I was surprised I can do this, makes it quite neat (using a reference from the same object destructuring).

andresmgot (approved):

PR #1972, opened 2020-08-27
Update operator components to use config.kubeappsCluster

absoludity:
Follows #1971 , this time all the operator components.

Ref: #1942

PR #1974, opened 2020-08-28
Remove assumption of "default" cluster for operator URLs pt1

absoludity:
Follows #1972 , related to #1942,  updating the following to no longer assume that the cluster on which Kubeapps is installed is the "default":

- Catalog and Service binding URLs
- all operator URLs.

For the service catalog I've just pulled the cluster out of the config (as I don't know that we're going to be furthering the service catalog work and we'd need to route them using cluster as well), whereas for the operator functionality I've done it properly, passing through from the components through the dispatch functions to actions, to the shared helpers and urls.

### DIFF
--- a/dashboard/src/actions/catalog.test.tsx
+++ b/dashboard/src/actions/catalog.test.tsx
@@ -18,6 +18,7 @@ const clusterClass = { metadata: { name: "cluster-class" } } as any;
 let store: any;
 const testArgs = {
   releaseName: "my-release",
+  kubeappsCluster: "kubeappsCluster",
   namespace: "my-namespace",
   className: "my-class",
   planName: "myPlan",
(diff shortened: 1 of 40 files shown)


andresmgot (approved):
What would be part 2?

absoludity:
> What would be part 2?

The chart change to ensure that the default cluster can be configured (though as per the branch name, I'd stuck a few 0.1 and 0.2 parts in front :) )

absoludity:
> > What would be part 2?
> 
> The chart change to ensure that the default cluster can be configured (though as per the branch name, I'd stuck a few 0.1 and 0.2 parts in front :) )

Oh right, I see. I think I'd initially planned on using two PRs for the operator URLs, but ended up doing it all in one. That's all :)

PR #1978, opened 2020-08-28
Operator Instance form v2

andresmgot:
### Description of the change

You know the drill. Form to create a Operator Instance. Still need to adapt the component to update those.

Screenshot from 2020-08-28 11-46-24

### Applicable issues

  - ref #1354


dashboard/src/components/DeploymentFormBody/AdvancedDeploymentForm.v2.tsx
```diff
@@ -7,7 +7,7 @@ import "ace-builds/src-noconflict/theme-xcode";
 export interface IAdvancedDeploymentForm {
   appValues?: string;
   handleValuesChange: (value: string) => void;
-  children: JSX.Element;
+  children?: JSX.Element;
```
absoludity, line 10: Why would we ever have a form without any children? Seems like we're throwing away some benefit of the type checker, but maybe I'm missing something?

dashboard/src/components/OperatorInstanceFormBody/__snapshots__/OperatorInstanceFormBody.v2.test.tsx.snap
```diff
@@ -0,0 +1,8 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`loading spinner matches the snapshot 1`] = `
+<LoadingWrapper
+  loaded={false}
+  type={0}
+/>
```
absoludity, line 7: Talking of snapshots, this one looks like the other extreme of snapshots - is there much use? (as in, a unit test can confirm this trivially without losing any clarity, imo). But fine either way.

absoludity (approved):
FWIW, I've done most of the work now on operator routes and actions etc., so that we could make it available on other clusters when we want (probably just needs some wording changes when installing the OLM on the specific cluster etc.)

dashboard/src/components/OperatorInstanceFormBody/__snapshots__/OperatorInstanceFormBody.v2.test.tsx.snap
andresmgot: yep, this is just that a function that is called `itBehavesLike` generates it. This method is just syntax sugar to check if the LoadingWrapper is rendered but it also creates a snapshot.

andresmgot:
> FWIW, I've done most of the work now on operator routes and actions etc., so that we could make it available on other clusters when we want (probably just needs some wording changes when installing the OLM on the specific cluster etc.)

:+1: I have an item in my TODO list to do some QA setting multicluster. We can add a follow up task to enable Operators in other clusters if it's trivial.

PR #1981, opened 2020-08-28
Fix upgrades for apps installed from a global repository

### LINKED ISSUE #1980: Apps installed from a global repository cannot be upgraded
### Description:

When trying to upgrade an app installed from a global repository, this error comes up:

image

1. Install kubeapps, by default
2. Install Apache in the `default` namespace
3. Try to upgrade

There are two issues here:

 - The dashboard is not able to auto-discover the update info. This is because we are only asking for charts in the app namespace.
 - The form that allows to select the source of the application (SelectRepoForm) is no longer working (since we are only requesting app repos within the app namespace).

andresmgot:
### Description of the change

Partially fixes #1980 from the dashboard. WDYT?

The SelectRepoForm still need to be fixed though (but that's a different issue).

### DIFF
--- a/dashboard/src/actions/apps.test.tsx
+++ b/dashboard/src/actions/apps.test.tsx
@@ -9,6 +9,7 @@ import { definedNamespaces } from "../shared/Namespace";
 import { IAppState, UnprocessableEntity } from "../shared/types";
 
 const mockStore = configureMockStore([thunk]);
+const kubeappsNamespace = "kubeapps";
 
 let store: any;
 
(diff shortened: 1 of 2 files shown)


absoludity:
> 
> Partially fixes #1980 from the dashboard. WDYT?

Looking at the issue, I don't think this is the right approach. The backend was meant to be returning all charts that target the given namespace (ie. both global and namespace-specific). I'll push an alternative PR in a few for you to look at and add more context to the bug.

andresmgot:
Dropping this in favor of #1983

PR #1983, opened 2020-08-31
Include global charts when filtering for charts available in a namespace

### LINKED ISSUE #1980: Apps installed from a global repository cannot be upgraded

absoludity:
### Description of the change

Fixes #1980 first by ensuring the integration test fails as it should have and then updating the assetsvc function in question to do what was intended - return charts from both the specific namespace as well as the global charts.

Still TODO: update the mongodb driver in the same way.

### Benefits

Can upgrade charts deployed from the global namespace.

### Applicable issues

  - fixes #1980 

### Additional information

More of the problem investigation and description on the issue.

### DIFF
--- a/cmd/assetsvc/postgresql_utils.go
+++ b/cmd/assetsvc/postgresql_utils.go
@@ -160,7 +160,16 @@ func containsVersionAndAppVersion(chartVersions []models.ChartVersion, version, 
 }
 
 func (m *postgresAssetManager) getChartsWithFilters(namespace, name, version, appVersion string) ([]*models.Chart, error) {
-	charts, err := m.QueryAllCharts(fmt.Sprintf("SELECT info FROM %s WHERE repo_namespace = $1 AND info ->> 'name' = $2", dbutils.ChartTable), namespace, name)
+	clauses := []string{"info ->> 'name' = $1"}
+	queryParams := []interface{}{name, namespace}
+	if namespace != dbutils.AllNamespaces {
+		queryParams = append(queryParams, m.GetKubeappsNamespace())
+		clauses = append(clauses, "(repo_namespace = $2 OR repo_namespace = $3)")
+	} else {
+		clauses = append(clauses, "repo_namespace = $2")
+	}
+	dbQuery := fmt.Sprintf("SELECT info FROM %s WHERE %s ORDER BY info ->> 'ID' ASC", dbutils.ChartTable, strings.Join(clauses, " AND "))
+	charts, err := m.QueryAllCharts(dbQuery, queryParams...)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


absoludity:
FWIW, you can see the passing postgresql but failing mongodb integration tests after I'd landed the postgres fix only, at https://app.circleci.com/pipelines/github/kubeapps/kubeapps/969/workflows/d41894e2-b3ff-485b-b5dc-7e5d604bf50d

cmd/assetsvc/postgresql_utils.go
```diff
@@ -160,7 +160,16 @@ func containsVersionAndAppVersion(chartVersions []models.ChartVersion, version,
...
-	charts, err := m.QueryAllCharts(fmt.Sprintf("SELECT info FROM %s WHERE repo_namespace = $1 AND info ->> 'name' = $2", dbutils.ChartTable), namespace, name)
+	clauses := []string{"info ->> 'name' = $1"}
+	queryParams := []interface{}{name, namespace}
+	if namespace != dbutils.AllNamespaces {
+		queryParams = append(queryParams, m.GetKubeappsNamespace())
+		clauses = append(clauses, "(repo_namespace = $2 OR repo_namespace = $3)")
+	} else {
+		clauses = append(clauses, "repo_namespace = $2")
+	}
+	dbQuery := fmt.Sprintf("SELECT info FROM %s WHERE %s ORDER BY info ->> 'ID' ASC", dbutils.ChartTable, strings.Join(clauses, " AND "))
```
andresmgot, line 171: why the ordering?

integration/use-cases/upgrade.js
```diff
@@ -1,14 +1,12 @@
 test("Upgrades an application", async () => {
-  await page.goto(getUrl("/#/login"));
+  await page.goto(getUrl("/#/c/default/ns/default/catalog"));
```
andresmgot, line 2: I thought about changing this as well but the problem is that now we are not covering the case of an upgrade in the kubeapps namespace :man_shrugging: In any case, since both cases are covered with unit tests is fine.

andresmgot (approved):
:+1: I like this solution better.

cmd/assetsvc/postgresql_utils.go
absoludity: Just so it's clearly deterministic (in tests as well as functionality).

integration/use-cases/upgrade.js
absoludity: Yeah - we could easily add another, but I think testing the deploy and upgrade of a global chart in a namespace other than the `kubeapps` one is by far more relevant for the integration tests.

PR #1989, opened 2020-09-01
Fix operator code landed between #1974 proposed and landed.

absoludity:
I didn't land #1974 immediately due to another issue and in the following day other call-sites for the changed URLs landed so that master was broken when I landed #1974


absoludity:
Landing to unbreak master.

PR #1991, opened 2020-09-01
Fix opacity and select issue

andresmgot:
### Description of the change

The script `create-react-app` doesn't play well with percentages in SCSS (apparently): https://stackoverflow.com/questions/58853844/the-opacity-value-was-changed-to-1-after-building-the-reacjs-project which caused the `opacity` of the context selector to be `1%`.

Apart from the fix for the above, this PR includes another small fix for the namespace and cluster selector that was not setting a default `value`.

### Applicable issues

  - ref #1354

### DIFF
--- a/dashboard/src/components/Header/ContextSelector.scss
+++ b/dashboard/src/components/Header/ContextSelector.scss
@@ -4,14 +4,14 @@
 }
 
 .kubeapps-dropdown {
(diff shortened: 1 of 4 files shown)


andresmgot:
_merging to tag a release_

absoludity:
Great, +1

PR #1993, opened 2020-09-01
Fix version selector in update form

andresmgot:
### Description of the change

Small fix for the update form. There, the version in the selector was forced to `currentVersion` so even after changing it, the currentVersion was always displayed in the selector.


absoludity:
+1, good catch. Also, I'm deleting these branches to clean up the remote repo a bit. It'd be good to get it back to just branches that are used.

PR #1995, opened 2020-09-02
Remove dependence on const kube.DefaultClusterName, moving to config.

absoludity:
### Description of the change

Another prep branch for the actual change of 1942: this PR removes the const `kube.DefaultClusterName` to instead be included in the clustersConfig for `kubeops`, still using the value `default`, though this will change in the next PR

I've one concern about the change of the cli arg, which I'll note below.

Ref #1942

### DIFF
--- a/cmd/kubeops/main.go
+++ b/cmd/kubeops/main.go
@@ -27,16 +27,16 @@ import (
 	"k8s.io/helm/pkg/helm/environment"
 )
 
-const additionalClustersCAFilesPrefix = "/etc/additional-clusters-cafiles"
+const clustersCAFilesPrefix = "/etc/clusters-cafiles"
 
 var (
-	additionalClustersConfigPath string
-	assetsvcURL                  string
-	helmDriverArg                string
-	listLimit                    int
-	settings                     environment.EnvSettings
-	timeout                      int64
-	userAgentComment             string
+	clustersConfigPath string
+	assetsvcURL        string
+	helmDriverArg      string
+	listLimit          int
+	settings           environment.EnvSettings
+	timeout            int64
+	userAgentComment   string
 )
 
 func init() {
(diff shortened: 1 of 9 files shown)


cmd/kubeops/main.go
```diff
@@ -47,7 +47,7 @@ func init() {
 	pflag.StringVar(&userAgentComment, "user-agent-comment", "", "UserAgent comment used during outbound requests")
 	// Default timeout from https://github.com/helm/helm/blob/b0b0accdfc84e154b3d48ec334cd5b4f9b345667/cmd/helm/install.go#L216
 	pflag.Int64Var(&timeout, "timeout", 300, "Timeout to perform release operations (install, upgrade, rollback, delete)")
-	pflag.StringVar(&additionalClustersConfigPath, "additional-clusters-config-path", "", "Configuration for additional clusters")
```
absoludity, line 50: I was checking `pflag` to see if there's any alias option so that I could keep the old flag name as well, but don't see any. It would probably be worth keeping the flag so that the binary doesn't barf if passed the old flag name for now? Actually, I'm going to do that change right now.

absoludity:
#1997 (when passing and landed in master) should resolve the CI issues in this PR (once rebased on master).

cmd/kubeops/main.go
andresmgot: People don't interact with these flags, just with the chart values so I don't see the need of maintaining the old flag. Also, it would be the same if we finally remove the branch in two versions, isn't it? (And I'm pretty sure we won't remember to delete it).

andresmgot (approved):

cmd/kubeops/main.go
absoludity: The problem isn't people interacting with the flags directly, but rather indirectly if and when they change the image that they use. Given that the chart can be updated independently of the images used, it's really a pain to maintain forwards and backwards compatibility. All I've done here is ensure that the new kubeops image generated with this code, can happily run with both the currently released chart as well as the new one (based on changes in the next PR). It does nothing about the existing image on the new chart.

But perhaps you mean that people don't actually run a different image than the default provided with the chart, which I imagine is generally the case. It'd be nice to make that an explicit assumption, then yes, we wouldn't need to bother doing this (ie. rolling back is only ever rolling back the chart+image in lock-step).

andresmgot: > But perhaps you mean that people don't actually run a different image than the default provided with the chart, which I imagine is generally the case. It'd be nice to make that an explicit assumption, then yes, we wouldn't need to bother doing this (ie. rolling back is only ever rolling back the chart+image in lock-step).

Yes, if someone changes the default image, they are walking into untested realms and we cannot ensure that it will work. We have changed flags before (for example for the PosgreSQL change) and we didn't have backward compatibility changes. In any case, it's true that now we could release a new chart version without an app version.

absoludity: We should then be explicit about that, given that helm (and our chart) allows you to update those independently. When a chart upgrade happens, it does the chart changes (ie. container arguments) at the same time as the new image by default, but people might want to pin the image so it'd be good to be explicit.

PR #1997, opened 2020-09-02
1942 app repos use cluster

absoludity:
### Description of the change

Updates shared helpers for AppRepositories, ServiceBrokers and ServiceInstances to no longer rely on the non-cluster-aware backend URIs.

### Benefits

Required for #1995 to be able to pass CI.

### Applicable issues

Ref: #1942

### DIFF
--- a/dashboard/src/actions/repos.ts
+++ b/dashboard/src/actions/repos.ts
@@ -125,7 +125,7 @@ export const deleteRepo = (
       },
     } = getState();
     try {
-      await AppRepository.delete(name, namespace);
+      await AppRepository.delete(currentCluster, name, namespace);
       dispatch(fetchRepos(currentNamespace));
       return true;
     } catch (e) {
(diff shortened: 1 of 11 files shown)


dashboard/src/actions/repos.ts
```diff
@@ -139,9 +139,12 @@ export const resyncRepo = (
   name: string,
   namespace: string,
 ): ThunkAction<Promise<void>, IStoreState, null, AppReposAction> => {
-  return async dispatch => {
+  return async (dispatch, getState) => {
+    const {
+      clusters: { currentCluster },
+    } = getState();
```
andresmgot, line 145: why don't we do do this rather than passing around `cluster` all the time in other actions? :sweat_smile:

andresmgot (approved):

dashboard/src/actions/repos.ts
absoludity: Not sure if it was asked in jest or seriousness :) . In the long run, functional code is much easier to work with, test, maintain etc., but I took a short-cut here that I didn't take with the normal components because it's not clear to me whether we'll be keeping the service catalog / service instances in the future.

andresmgot: well, it was a serious question. I don't think this is a shortcut since at the end is the same than what we are doing (getting the current cluster from the state). unless I'm missing something. IMO, sacrifice a bit of complexity for the tests vs adding the parameter "cluster" or "namespace" all over the place is something reasonable. Anyway, this is just me thinking out loud.

Issue #1907, opened 2020-07-30
[clarity UI] Several improvements


andresmgot:
### Description:

Issue to gather feedback and possible improvements while we work in the new UI:
(this is mostly a list for me to not forget about these items)

**General** 
 - [x] Refactor subheader component
 - [x] Revisit error messages
 - [x] Upgrade Clarity

**App List**
 - [x] Remove the `Status: ` prefix of the status label
 - [x] Standardize footer: App <app>, Version: <version>
 - [x] Move update info to a more prominent place (title?)
 - [x] Vertically center the description of cards
 ~~- [ ] Add status for operator instances~~
 - [x] Fix icon issues in bigger screens

**Catalog**
 - [x] Add summary of filters applied (make it possible to clear them?)
 - [x] Remove colon (:) from filters
 - [x] Operator source as a filter

**App View**
 - [x] Chart section: merge versions and update info

**Deployment Form**
 - [x] Highlight if there are changes to review
 - [x] Review differential tab style, fix scrolling
 - [x] Fix slow typing

**Upgrade Form**
 - [x] Make the version more prominent

**Context Selector**
 - [x] Allow to select "All Namespaces"
 - [x] Allow to create new namespaces
 - [x] Test Multicluster setup
 - [x] Fix misbehaving selector (selects first by default)

**Chart View**
 - [x] Issue with README rendered for lists

**App Repo List**
 - [x] Link to app repository catalog

**Operator List**
 - [x] Installed vs Available operators
 - [x] Do not pin to a specific OLM version
 - [x] Remove check for default cluster only

PR #1998, opened 2020-09-02
Refactor PageHeader v2

andresmgot:
### Description of the change

When adding components, each one of them used PageHeader differently. This PR refactor those so they use a common and easy to use PageHeader.

### Applicable issues

  - ref #1907


absoludity (approved):
Looks neater :)

PR #2000, opened 2020-09-03
Update chart's clusters list to be definitive.

### LINKED ISSUE #1942: Support running Kubeapps on a TKG Management Cluster

### LINKED ISSUE #1943: Change Default Cluster Name in multi cluster setup
### Description:

Currently in a multi cluster setup, the local cluster which is the default cluster is named default in the context dropdown for selecting clusters. 
As this is a multi cluster solution where i believe all clusters are first class and not just partially supported that it would be better to change this to be the cluster name and not "default".
having it with the name default will confuse users as they wont know what cluster this is.

absoludity:
### Description of the change

Updates the `.Values.clusters` list of clusters to be definitive, enabling the cluster on which kubeapps is installed to have a name other than `default` or to not be a target cluster at all for users.

### Benefits

See #1942 and #1943 

TL;DR is that with this change, the cluster on which Kubeapps is running can be called whatever the user likes, but still defaults to `default`, and this is included in the definitive list of `.Values.clusters`.

### Applicable issues

  - fixes #1942 
  - fixes #1943

### DIFF
--- a/dashboard/src/reducers/cluster.ts
+++ b/dashboard/src/reducers/cluster.ts
@@ -155,15 +155,10 @@ const clusterReducer = (
           namespaces: [],
         };
       });
-      if (Object.keys(clusters).length === 0) {
-        clusters.default = {
-          currentNamespace: "default",
-          namespaces: [],
-        };
-      }
+
       return {
         ...state,
-        currentCluster: action.payload.kubeappsCluster,
+        currentCluster: config.clusters[0],
         clusters,
       };
     default:
--- a/dashboard/src/reducers/cluster.ts
+++ b/dashboard/src/reducers/cluster.ts
@@ -146,17 +146,21 @@ const clusterReducer = (
       }
       break;
     case getType(actions.config.receiveConfig):
-      // Initialize the additional clusters when receiving the config.
-      const clusters = {
-        ...state.clusters,
-      };
+      // Initialize the clusters when receiving the config.
       const config = action.payload as IConfig;
+      const clusters: IClustersMap = {};
       config.clusters?.forEach(cluster => {
         clusters[cluster] = {
           currentNamespace: "default",
           namespaces: [],
         };
       });
+      if (Object.keys(clusters).length === 0) {
+        clusters.default = {
+          currentNamespace: "default",
+          namespaces: [],
+        };
+      }
       return {
         ...state,
         currentCluster: action.payload.kubeappsCluster,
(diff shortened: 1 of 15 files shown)


absoludity:
Marking this ready for review, it still has one CI test failing which I need to look into (fail when missing permissions), but I'll do that first thing.

dashboard/src/reducers/cluster.ts
```diff
@@ -146,20 +146,19 @@ const clusterReducer = (
...
       break;
     case getType(actions.config.receiveConfig):
-      // Initialize the additional clusters when receiving the config.
-      const clusters = {
-        ...state.clusters,
-      };
+      // Initialize the clusters when receiving the config.
       const config = action.payload as IConfig;
+      const clusters: IClustersMap = {};
       config.clusters?.forEach(cluster => {
```
andresmgot, line 152: so `clusters` should be defined, right? `config.clusters.forEach`

andresmgot (approved):

dashboard/src/reducers/cluster.ts
absoludity: Yep, indeed.

PR #2001, opened 2020-09-03
fix frontend service port when using external proxy

### LINKED ISSUE #1996: Cannot use external auth proxy without patching frontend service
### Description:

When using external auth proxy, the Kubeapps service is not accessible. In the frontend service the value of `targetPort` should be `http` instead of `proxy`.

https://github.com/kubeapps/kubeapps/blob/ea44804980d7238e94096d9c3ba499438a64d169/chart/kubeapps/templates/kubeapps-frontend-service.yaml#L16

The above line should be changed to
```
{{- if and .Values.authProxy.enabled (not .Values.authProxy.external) }}
```

zahid0:
### Description of the change
Fixes frontend service port when using external proxy

### Applicable issues

  - fixes #1996

### DIFF
--- a/chart/kubeapps/templates/kubeapps-frontend-service.yaml
+++ b/chart/kubeapps/templates/kubeapps-frontend-service.yaml
@@ -13,7 +13,7 @@ spec:
   {{- end }}
   ports:
     - port: {{ .Values.frontend.service.port }}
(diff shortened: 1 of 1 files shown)


absoludity (approved):
Great, thanks!

PR #2004, opened 2020-09-03
Refactor cards

andresmgot:
### Description of the change

 - [X] Remove the `Status: ` prefix of the status label.
 - [X] Standardize footer: App <app>, Version: <version>. Since operators don't include the app version but it's important for charts, the format is now `App: name vVersion / Chart: version` for charts and `App: name / Operator: version` for operators. 
 - [X] Move update info to a more prominent place (title?). Since the title already had an icon (helm or OLM icon) I moved the type icon as a subtle background image in the description so I can add a tooltip with the update info in the title.
 - [X] Vertically center the description of cards
 ~~- [ ] Add status for operator instances~~ This cannot be done since there is no a generic way to obtain a short status from arbitrary instances
 - [X] Fix icon issues in bigger screens

Screenshots:

Screenshot from 2020-09-03 11-57-33

Screenshot from 2020-09-03 11-44-43

### Applicable issues

  - ref #1907

### DIFF
--- a/dashboard/src/components/AppList/AppListItem.v2.test.tsx
+++ b/dashboard/src/components/AppList/AppListItem.v2.test.tsx
@@ -1,5 +1,7 @@
+import Tooltip from "components/js/Tooltip";
 import { shallow } from "enzyme";
 import * as React from "react";
+import { defaultStore, mountWrapper } from "shared/specs/mountWrapper";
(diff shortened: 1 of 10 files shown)


absoludity (approved):
I like the more subtle background icon for the type.

PR #2005, opened 2020-09-03
Upgrade Clarity to v4

andresmgot:
### Description of the change

Use the latest version of Clarity. This has the advantage that we can now remove the temporary wrapper we had and use new core components (for the moment we are just using `CdsButton` and `CdsIcon`).

Not much to review here, only a couple of manual changes that I've commented.

### Applicable issues

  - ref #1907


dashboard/src/components/Layout/UISelector/Clarity.tsx
```diff
@@ -1,8 +1,69 @@
...
+  infoCircleIcon,
+  plusCircleIcon,
+  refreshIcon,
+  rewindIcon,
+  searchIcon,
+  successStandardIcon,
+  timesCircleIcon,
+  timesIcon,
+  trashIcon,
+  uploadCloudIcon,
```
andresmgot, line 29: we still need to register the icons used in the app

absoludity (approved):

PR #2012, opened 2020-09-07
Update multi-cluster instructions with definitive list of clusters.

absoludity:
### Description of the change

Updates the multi-cluster documentation to match the changes included as part of the solution to #1942 and #1943 .

While there, I've included some other debugging and common problems when configuring OIDC for k8s clusters based on early feedback.

View the updated doc at https://github.com/absoludity/kubeapps/blob/1942-docs-update/docs/user/deploying-to-multiple-clusters.md


andresmgot (approved):
:+1:

PR #2015, opened 2020-09-08
Add option for all namespaces in the namespace selector

andresmgot:
### Description of the change

Added the item for "All Namespaces" in the context selector as we have it today:

Screenshot from 2020-09-08 13-32-08

For the moment I am just matching current functionality, it's not the goal to rethink about the desired UX.

I have also fixed a small bug I noticed. The `select` input for the namespace were showing the wrong namespace when the namespace was changed outside the selector (for example when clicking on a link).

### Applicable issues

  - ref #1907

### DIFF
--- a/dashboard/src/components/Header/ContextSelector.scss
+++ b/dashboard/src/components/Header/ContextSelector.scss
@@ -56,3 +56,12 @@
     }
   }
 }
+
+.kubeapps-create-new-ns {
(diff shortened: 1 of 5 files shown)


absoludity (approved):
Personally I'd prefer we at least only display All Namespaces if the user has access to all namespaces, rather than providing an option they can't use.

PR #2016, opened 2020-09-08
Add form to create a new namespace

andresmgot:
### Description of the change

Allow people to create and use new namespaces:

Peek 2020-09-08 17-10

### Applicable issues

  - ref #1907 

### Additional information

I refactored the `ContextSelector` component to avoid using props so we can eventually delete the container.

### DIFF
--- a/dashboard/src/components/Header/ContextSelector.scss
+++ b/dashboard/src/components/Header/ContextSelector.scss
@@ -56,3 +56,12 @@
     }
   }
 }
+
+.kubeapps-create-new-ns {
+  display: flex;
+  justify-content: flex-end;
+  cds-button {
+    --color: var(--clr-header-font-color, #fafafa);
+    outline: 0px;
+  }
+}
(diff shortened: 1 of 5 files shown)


absoludity (approved):
Are you handling users who don't have RBAC to create namespaces (ie. all non-admin users, effectively) in the same way? (ie. showing the required RBAC). Similar to the previous PR, I think for these admin-type functions (viewing all namespaces, creating namespaces) we'd be better not not provide an option that can't be used (which is different to some missing RBAC for something a normal user could be able to use).

andresmgot:
> Are you handling users who don't have RBAC to create namespaces (ie. all non-admin users, effectively) in the same way? (ie. showing the required RBAC). Similar to the previous PR, I think for these admin-type functions (viewing all namespaces, creating namespaces) we'd be better not not provide an option that can't be used (which is different to some missing RBAC for something a normal user could be able to use).

For the moment we don't have an API to check that so let me open a new issue to properly design that. Note that for the creation of namespaces it's trivial to check RBAC permissions but for listing all namespaces we can either check if the user has permissions to list namespaces (which is not the same thing) or if they have permissions to get secrets in everywhere (which maybe is too much?).

PR #2019, opened 2020-09-09
Refactor update info

andresmgot:
### Description of the change

The version and update info are currently separated so it's a bit confusing. Also, it's more difficult to notice if there are updates available. This is the current status:

Screenshot from 2020-09-09 11-46-49

In this PR I have moved all the version info to the same section and use an alert instead:

Screenshot from 2020-09-09 11-45-28

In the case there are no updates, is more subtle:

Screenshot from 2020-09-09 11-45-39

### Applicable issues

  - ref #1907

### DIFF
--- a/dashboard/src/components/AppView/AppView.v2.tsx
+++ b/dashboard/src/components/AppView/AppView.v2.tsx
@@ -200,7 +200,7 @@ export default function AppView({
       ) : (
         <Row>
           <Column span={3}>
-            <ChartInfo app={app} />
+            <ChartInfo app={app} cluster={cluster} />
           </Column>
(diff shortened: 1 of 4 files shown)


absoludity (approved):
I think it's great being more noticeable and in the expected place, just wonder if we should also provide the ability to act on it (ie. a small upgrade link/button right there?)

andresmgot:
> I think it's great being more noticeable and in the expected place, just wonder if we should also provide the ability to act on it (ie. a small upgrade link/button right there?)

Good point. Added:
Screenshot from 2020-09-10 15-52-01

PR #2022, opened 2020-09-09
Split between installed and available operators

andresmgot:
### Description of the change

Use `subscriptions.operators.coreos.com` to differ if an operator has been installed or not (this is the entity created when installing an operator).

Screenshot from 2020-09-09 15-57-09

### Applicable issues

  - ref #1907


absoludity (approved):
Great.

PR #2024, opened 2020-09-10
Fix bugs for multicluster selector

andresmgot:
### Description of the change

I tried the multi-cluster support using the local dev environment to check how it behaves with the new UI. Found several bugs that has been fixed here:

 - When selecting a new cluster, the namespace selector was not being populated with the available namespaces of that selector.
 - The action `setNamespace` was not cluster-aware. This caused a race condition in which that is called before changing the `currentCluster` in the state, causing requesting resources from the wrong cluster.
 - When selection a cluster, a default namespace should be `selected`.
 - The `cluster` state reducer had a hard-coded reference to the `default` cluster.

### Applicable issues

  - ref #1907

### DIFF
--- a/dashboard/src/actions/namespace.test.tsx
+++ b/dashboard/src/actions/namespace.test.tsx
@@ -32,7 +32,12 @@ interface ITestCase {
 }
 
 const actionTestCases: ITestCase[] = [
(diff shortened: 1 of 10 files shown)


absoludity (approved):
Thanks for the fixes.

PR #2027, opened 2020-09-11
Added .Values.frontend.largeClientHeaderBuffers to accommodate large …

rarick:
…OIDC and LDAP cookies

### Description of the change

Added a new value in the `frontend` section to allow for the configuration of the nginx `large_client_header_buffers` value. This was an issue internally at my company, and is also brought up in issue #1571.

### Benefits

Allows for configuration of the `large_client_header_buffers` field.

### Possible drawbacks

An additional setting for configuration and documenting.

### Applicable issues

  -  #1571 

### Additional information

It could be preferable to just set this option permanently rather than having it as a configuration option, but I've made it configurable in this pull request.

### DIFF
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -204,9 +204,8 @@ frontend:
   proxypassAccessTokenAsBearer: false
   ## Set large_client_header_buffers in nginx config
   ## Can be required when using OIDC or LDAP due to large cookies
-  ## Example value: "4 32k"
   #
-  largeClientHeaderBuffers: ""
+  largeClientHeaderBuffers: "4 32k"
 
 ## AppRepository Controller is the controller used to manage the repositories to
 ## sync. Set apprepository.initialRepos to configure the initial set of
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -202,6 +202,11 @@ frontend:
   ## Some K8s distributions such as GKE requires it
   ##
   proxypassAccessTokenAsBearer: false
+  ## Set large_client_header_buffers in nginx config
+  ## Can be required when using OIDC or LDAP due to large cookies
+  ## Example value: "4 32k"
+  #
+  largeClientHeaderBuffers: ""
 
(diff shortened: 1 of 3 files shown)


chart/kubeapps/values.yaml
```diff
@@ -202,6 +202,11 @@ frontend:
   ## Some K8s distributions such as GKE requires it
   ##
   proxypassAccessTokenAsBearer: false
+  ## Set large_client_header_buffers in nginx config
+  ## Can be required when using OIDC or LDAP due to large cookies
+  ## Example value: "4 32k"
+  #
+  largeClientHeaderBuffers: ""
```
absoludity, line 209: Thanks @rarick . I'd be +1 to make the default "4 32k", given that I'm keen to have the OIDC setup as the default one, though it's still good for it to be configurable (people can switch it off or make it bigger if, for eg., there setup has many groups being encoded into the OIDC id_token)

What do you think?

rarick: That sounds good to me. In the scenario that broke our setup, group membership in the token pushed the cookie size to just over 8k, so 32k should be fine for the vast majority of users. I'll get that in right away 👍

absoludity: Yeah, I had a similar thing while testing OIDC + LDAP (#1944, but in that case it was on the ingress nginx for a dev setup). Excellent, thanks for the change!

absoludity (commented):
Note that it's failing CI because of the error:
```
2020/09/14 03:35:17 [emerg] 1#0: invalid number of arguments in "large_client_header_buffers" directive in /opt/bitnami/nginx/conf/server_blocks/vhost.conf:13
nginx: [emerg] invalid number of arguments in "large_client_header_buffers" directive in /opt/bitnami/nginx/conf/server_blocks/vhost.conf:13
```
because you quoted the value, where as it's not meant to be quoted: http://nginx.org/en/docs/http/ngx_http_core_module.html#large_client_header_buffers

Works fine without the `| quote` :) I'll +1 and land if you can update removing that. Thanks!

absoludity (approved):
Great, thanks!

Issue #2026, opened 2020-09-11
[Multi-cluster] Auth to asset service fails


absoludity:
### Description:

While testing the multi-cluster support on Tanzu Kubernetes Grid, I found in that environment that although the general authn was working fine, the specific requests to the asset svc were failing, so that the catalog could not be fetched:

![tkg-multicluster](https://user-images.githubusercontent.com/497518/92862909-a6b12b80-f43e-11ea-9f73-8e8625487cc2.png)

### Steps to reproduce the issue:

1. Setup a TKG management cluster with an oidc-enabled workload cluster
2. Configure the multi-cluster support with Kubeapps running on the management cluster
3. Log in to kubeapps and view the catalog

### Describe the results you received:

As per the screenshot, the kubeops svc (in particular, the authgate) checks if the user can get secrets in the namespace fails, even though the user is able to do so.

### Describe the results you expected:

Catalog is displayed without issue (as it is in the local multi-cluster setup)

### Additional information you deem important (e.g. issue happens only occasionally):

The request comes back as unauthorized.

### Version of Helm, Kubeapps and Kubernetes:

Using devel Kubeapps.

absoludity:
Initial investigations show this to be an unexpected (on my part) networking configuration, in that I can curl the API server on its external dns using a token without issue, but when doing the same curl (with the same token) from a pod within the cluster using the internal IP, I get
```
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {
    
  },
  "status": "Failure",
  "message": "Unauthorized",
  "reason": "Unauthorized",
  "code": 401
}
```

Is this the AWS specific locked-down env you were talking about @vrabbi ?

vrabbi:
I don't believe so. I think this is a kubeadm thing I saw before in the cluster API slack channel overall. As of the last versions I believe this is the expected behavior. What version of k8s did you install with tkg?

On Sep 14, 2020 07:43, Michael Nelson <notifications@github.com> wrote:

Initial investigations show this to be an unexpected (on my part) networking configuration, in that I can curl the API server on its external dns using a token without issue, but when doing the same curl (with the same token) from a pod within the cluster using the internal IP, I get

{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {

  },
  "status": "Failure",
  "message": "Unauthorized",
  "reason": "Unauthorized",
  "code": 401
}


[...]

PR #2031, opened 2020-09-15
Update assetsvc URIs to be cluster aware.

absoludity:
Required to be able to verify a users access to charts in a given namespace in the authgate check. 

Ref #2026


andresmgot (approved):
LGTM so far

PR #2034, opened 2020-09-15
Add a version selector in the upgrade form

andresmgot:
### Description of the change

The version selector is a bit difficult to notice, since this is important in the upgrade flow, I have also added a version selector in the main body of the form (as there were before):

Screenshot from 2020-09-15 15-32-31

### Applicable issues

  - ref #1907

### DIFF
--- a/dashboard/src/components/UpgradeForm/UpgradeForm.v2.tsx
+++ b/dashboard/src/components/UpgradeForm/UpgradeForm.v2.tsx
@@ -6,6 +6,7 @@ import * as YAML from "yaml";
 
 import ChartSummary from "components/Catalog/ChartSummary";
 import ChartHeader from "components/ChartView/ChartHeader.v2";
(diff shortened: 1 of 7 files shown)


dashboard/src/components/UpgradeForm/UpgradeForm.v2.tsx
```diff
@@ -181,6 +184,18 @@ function UpgradeForm({
           </Column>
           <Column span={9}>
             <form onSubmit={handleDeploy}>
+              <div className="upgrade-form-version-selector">
+                <label className="centered deployment-form-label deployment-form-label-text-param">
+                  Version To Upgrade
```
absoludity, line 189: "Upgrade to Version" ?

"Version to Upgrade" implies that the displayed version is being upgrade, rather than that it's the target of the upgrade.

absoludity (approved):

PR #2035, opened 2020-09-15
Fix some state issues

andresmgot:
### Description of the change

Noticed a couple of issues related to the existing state when upgrading a second application:

 - The first application `schema` was being rendered.
 - The "Changes" tab was using the first application deployed values as old values.

### Applicable issues

  - ref #1907

### DIFF
--- a/dashboard/src/components/UpgradeForm/UpgradeForm.v2.tsx
+++ b/dashboard/src/components/UpgradeForm/UpgradeForm.v2.tsx
@@ -100,9 +100,16 @@ function UpgradeForm({
       const values = applyModifications(newModifications, deployed.values);
       setModifications(newModifications);
       setAppValues(values);
-      setDeployedValues(values);
     }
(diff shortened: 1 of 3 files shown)


dashboard/src/components/UpgradeForm/UpgradeForm.v2.tsx
```diff
@@ -100,10 +100,17 @@ function UpgradeForm({
       const values = applyModifications(newModifications, deployed.values);
       setModifications(newModifications);
       setAppValues(values);
-      setDeployedValues(values);
     }
   }, [deployed.values, appCurrentValues, modifications]);
 
+  useEffect(() => {
+    if (deployed.values) {
+      // Calculate modifications from the default values
```
absoludity, line 108: Cut'n'paste error? I think this should be "from the deployed values"?

absoludity (approved):

Issue #2039, opened 2020-09-16
Remove MongoDB backend suppport for 2.0 pre-release


absoludity:
Moving forward we're removing support for using mongodb as the backend cache storage for Kubeapps as part of the major release.

PR #2040, opened 2020-09-16
Remove MongoDB option from chart.

absoludity:
### Description of the change

Removes MongoDB as an option for the backend in the chart.

Subsequent PR will remove from the scripts and code.

### Applicable issues

Ref: #2039


andresmgot (approved):
Thanks!

PR #2046, opened 2020-09-18
add apiVersion check for ingress apiVersion

Siegfriedk:
Ingress apiVersion extensions/v1beta1 got deprecated and is now removed
in 1.16.

https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/

Its the same approach as the grafana (and prometheus-operator) chart are taking:
https://github.com/grafana/helm-charts/blob/main/charts/grafana/templates/ingress.yaml#L6

### Description of the change

It adds a helm based capability check to switch between both apiVersions;

This is possible as the ingress apiVersions are compatible (which is not always the case, see DaemonSet)

This change enables the current Ingress spec to be used in Kubernetes Version >= 1.16 while keeping compatibility to Kubernetes Versions < 1.16.

### DIFF
--- a/chart/kubeapps/templates/ingress.yaml
+++ b/chart/kubeapps/templates/ingress.yaml
@@ -1,5 +1,9 @@
 {{- if .Values.ingress.enabled -}}
+{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1" }}
+apiVersion: networking.k8s.io/v1beta1
+{{ else }}
 apiVersion: extensions/v1beta1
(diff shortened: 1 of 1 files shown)


absoludity (commented):
Thanks @Siegfriedk . The change itself looks good, can you just revert the chart version change for now, as we don't want to release a new chart version right now (we're in the process of finalising changes for a chart and app release in a few days). Thanks.

absoludity (approved):
Great, thanks.

PR #2048, opened 2020-09-21
Remove mongodb support from asset svcs and re-write postgres tests.

### LINKED ISSUE #2039: Remove MongoDB backend suppport for 2.0 pre-release

absoludity:
### Description of the change

As part of #2039, this PR removes the support for MongoDB in the assetsvc and asset-sync services.

### Benefits

Less code to maintain, higher velocity moving forward with new features.

### Applicable issues

  - fixes #2039

### Additional information

This PR was particularly painful for two main reasons:

1. The existing tests for the db-agnostic sections of the code weren't db agnostic, but dependent on the mongo support (that's just historical fact  since we initially only had mongo support, I assume)
2. The existing fake manager that we had for the postgresql tests couldn't be used for the higher-level tests as returning results from the fake pg manager was a very bad experience. I've instead added and used `sqlmock` which makes it much easier to handle db tests
3. The assetsvc uses a global var (?) to store the manager (not great for a bunch of reasons, but in this case I updated the tests to be clearer that this was the case). Less relevant right now but we'd not be able to run tests in parallel.


cmd/asset-syncer/utils.go
```diff
@@ -184,6 +179,7 @@ func chartsFromIndex(index *helmrepo.IndexFile, r *models.Repo) []models.Chart {
 		}
 		charts = append(charts, newChart(entry, r))
 	}
+	sort.Slice(charts, func(i, j int) bool { return charts[i].ID < charts[j].Name })
```
andresmgot, line 182: I don't understand this sorting, shouldn't this be (if needed) `charts[i].Name < charts[j].Name`?

cmd/assetsvc/postgresql_utils.go
```diff
@@ -71,7 +71,7 @@ func (m *postgresAssetManager) getPaginatedChartList(namespace, repo string, pag
 	dbQuery := fmt.Sprintf("SELECT info FROM %s %s ORDER BY info ->> 'name' ASC", dbutils.ChartTable, repoQuery)
 	charts, err := m.QueryAllCharts(dbQuery, queryParams...)
 	if err != nil {
-		return nil, 0, nil
+		return nil, 0, err
```
andresmgot, line 74: :sweat_smile:

andresmgot (approved):
Thanks and sorry for the pain of adapting the tests.

cmd/asset-syncer/utils.go
absoludity: `charts[i].ID < charts[j].ID` is what I'd meant, thanks. The tests were failing intermittently because the order was not deterministic. Using Name still won't be deterministic as you can have `bitnami/apache` and `aardvark/apache`, the order of which would still be indeterminate if it's on the name only, I think?

cmd/assetsvc/postgresql_utils.go
absoludity: Yeah, updated tests actually found this :)

PR #2056, opened 2020-09-23
Update frontend deps before release.

absoludity:
### Applicable issues

  - fixes #

### DIFF
--- a/dashboard/yarn.lock
+++ b/dashboard/yarn.lock
@@ -2,24 +2,24 @@
 # yarn lockfile v1
 
 
-"@babel/code-frame@7.8.3", "@babel/code-frame@^7.0.0", "@babel/code-frame@^7.0.0-beta.35", "@babel/code-frame@^7.8.3":
+"@babel/code-frame@7.8.3":
   version "7.8.3"
   resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.8.3.tgz#33e25903d7481181534e12ec0a25f16b6fcf419e"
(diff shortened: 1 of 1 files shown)


absoludity:
I've gone ahead with the release without this for now. We appear to have a lot of frontend packages at versions that are no longer supported and should be updated. Let's do this as part of the next release?

andresmgot:
I guess so

antgamdia:
I'm also concerned about the hardcoded import of Clarity ([Component](https://github.com/kubeapps/kubeapps/blob/master/dashboard/src/components/Layout/Clarity.tsx#L35) and [Dockerfile](https://github.com/kubeapps/kubeapps/blob/master/dashboard/Dockerfile#L9)).
Dockerfile manually adds Clarity ui@4.0.1 whereas the package.json is also declaring [^2.3.4](https://github.com/kubeapps/kubeapps/blob/fee5e8130a5aaf9c088d0e884e4046ab4713550b/dashboard/package.json#L10), which is definitely incompatible with versions > 2.x.x.

For a newcomer dev it is a bit misleading to have to manually add a dependency that is apparenly included in the package.json.
Perhaps we can point the import to `import "../../../node_modules/@clr/ui/clr-ui.min.css";` instead of `import "../../clr-ui.min.css";`. IMHO, this will be compatible with both online / offline approaches.

andresmgot:
Good catch @antgamdia! feel free to send a PR to remove that ambiguity (we can also remove the references to `hex.min.css`)

absoludity:
Indeed. Good catch . I'll close this PR since it's way out of date now anyway - I had checked for a quick update before a release but that wasn't possible.

Issue #2028, opened 2020-09-14
Pre-release of Kubeapps 2.0


andresmgot:
### Description:

The goal of this issue is to gather requirements we need to fulfill before cutting a release candidate for Kubeapps 2.0. We are targeting September 23th to make the new version available. We can split the issues to finish if they need changes in the code base or not:

**Codebase related:**
 - [x] Finish improvements/bug fixes planned in #1907 related to the new UI.
 - [x] [Multi-cluster] Auth to asset service fails #2026
 - [x] [Upgrade Flow] Allow to select global or namespaced repositories #1987
 - [x] Remove flag and support for Helm v2
 - [x] Remove flag and support for MongoDB
 - [x] Remove feature flag for Operators
 - [x] Upgrade PostgreSQL dependency
 - [x] Migrate e2e tests
 - [x] Upgrade Helm dependency (v3.3.2 - Security patch)

**Documentation and Promotion**
 - [x] Documentation for Multicluster setup #1764
 - [x] Demo video for Multicluster setup #1764
 - [x] Update current documentation to reflect new UI
 - [x] Demo video presenting the new UI
 - [x] Update kubeapps.com
 - [x] Upgrade from 1.11 guide

andresmgot:
Everything done here :) Kubeapps 2.0.0-beta.2 already available.

PR #2060, opened 2020-09-23
Update documentation to Kubeapps 2.0

andresmgot:
### Description of the change

Adapt documentation to remove references to Helm 2 and the old UI.

### Applicable issues

  - ref #2028


README.md
```diff
@@ -15,15 +15,12 @@ Kubeapps is a web-based UI for deploying and managing applications in Kubernetes
 
 Use the Helm chart to install the latest version of Kubeapps: 
 
-***Note:*** Kubeapps supports both Helm 2 and Helm 3. For detailed instructions on how to install and use Kubeapps follow the [Getting Started Guide](docs/user/getting-started.md).
+***Note:*** Kubeapps 2.0 and onwards supports Helm 3 only. For detailed instructions on how to install and use Kubeapps follow the [Getting Started Guide](docs/user/getting-started.md).
```
absoludity, line 18: I'd be a bit more specific here? "supports the Helm 3 API only, which in most cases still allows installing charts made for Helm 2", or something similar?

chart/kubeapps/README.md
```diff
@@ -286,3 +251,21 @@ kubectl apply -f <repo name>.yaml
 ```
 
 After that you should be able to access the new version of Kubeapps. If the above doesn't work for you or you run into any other issues please open an [issue](https://github.com/kubeapps/kubeapps/issues/new).
+
+### Upgrading to 2.0
+
+Kubeapps 2.0 (Chart version 4.0.0) introduces some breaking changes:
+
+ - Helm 2 is no longer supported. If you are still using some Helm 2 charts, [migrate them with the available tools](https://helm.sh/docs/topics/v2_v3_migration/). If you are facing any issue managing this migration and Kubeapps, please open a new issue!
```
absoludity, line 259: As per that link, "Helm 2 charts (apiVersion=v1) are still installable" and that page is more about migrating your environment rather than charts, though it does mention how to update a chart to the new v2 api. Again, I think it's good to point people here, but also be clear that many helm2 charts (v1) just continue to work when deployed with helm 3.

docs/user/deploying-to-multiple-clusters.md
```diff
@@ -8,9 +8,10 @@ Once configured, you can select the cluster to which you are deploying in the sa
 
 When you have selected the target cluster and namespace, you can browse the catalog as normal and deploy apps to the chosen target cluster and namespace as you would normally.
 
-You can watch a brief demonstration of deploying to an additional cluster (we will update this demo in a few weeks with the new UI):
+<!-- TODO: Re-publish the Video  -->
+<!-- You can watch a brief demonstration of deploying to an additional cluster (we will update this demo in a few weeks with the new UI):
 
-[![Demo of Kubeapps with multiple clusters](https://img.youtube.com/vi/KIoW4zZDtdY/0.jpg)](https://www.youtube.com/watch?v=KIoW4zZDtdY)
+[![Demo of Kubeapps with multiple clusters](https://img.youtube.com/vi/KIoW4zZDtdY/0.jpg)](https://www.youtube.com/watch?v=KIoW4zZDtdY) -->
```
absoludity, line 14: Will do :)

absoludity (approved):
LGTM

docs/user/deploying-to-multiple-clusters.md
absoludity: Can you pls uncomment and update to use https://www.youtube.com/watch?v=pzVMZGTK0vU

README.md
andresmgot: okay, TBH, most users won't know the difference between the Helm 3 API and the chart specification versioning so will try to simplify it.

absoludity: Yeah, I agree it doesn't need to mention the API, I just mean that helm 2 charts are still supported outside of corner cases. There's nothing that needs to be done to still install those charts.

PR #2063, opened 2020-09-24
Add support for `kubectl alpha debug` for debugging scratch images.

absoludity:
Trying out as it'd be useful for debugging. But when running the debug command, it's just hanging for me at:

```
k -n kubeapps alpha debug -it kubeapps-internal-kubeops-74cd96c749-7jvs5 --image=busybox --target kubeops -- bash
Defaulting debug container name to debugger-ll99d.
```

Need to dig to find out why. The ephemeral containers are being created, I can see them on the pod, it's just not attaching apparently.


andresmgot:
Not sure what could be the reason but `busybox` doesn't have `bash` available.

absoludity:
> Not sure what could be the reason but `busybox` doesn't have `bash` available.

Yeah, it's the same without any command or explicitly with `sh`. At least for me. I'll dig a big more some other time, as it'd be nice to use.

PR #2071, opened 2020-09-28
Update the access-control doc to prefer OIDC.

absoludity:
### Description of the change

While writing some unrelated Kubeapps content, I went to link to the
access-control user doc and saw that it is still referring to service tokens as
the normal way to authenticate (and even referring to the generic token auth on
k8s, which should not be used for similar reasons).

### DIFF
--- a/docs/user/using-an-OIDC-provider.md
+++ b/docs/user/using-an-OIDC-provider.md
@@ -5,7 +5,7 @@ OpenID Connect (OIDC) is a simple identity layer on top of the OAuth 2.0 protoco
 It is possible to configure your Kubernetes cluster to use an OIDC provider in order to manage accounts, groups and roles with a single application. Additionally, some managed Kubernetes environments enable authenticating via plain OAuth2 (GKE).
 This guide will explain how you can use an existing OAuth2 provider, including OIDC, to authenticate users within Kubeapps.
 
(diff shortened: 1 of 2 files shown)


docs/user/using-an-OIDC-provider.md
```diff
@@ -5,6 +5,8 @@ OpenID Connect (OIDC) is a simple identity layer on top of the OAuth 2.0 protoco
 It is possible to configure your Kubernetes cluster to use an OIDC provider in order to manage accounts, groups and roles with a single application. Additionally, some managed Kubernetes environments enable authenticating via plain OAuth2 (GKE).
 This guide will explain how you can use an existing OAuth2 provider, including OIDC, to authenticate users within Kubeapps.
 
+For a complete worked example of this process on a specific Kubernetes environment, one of the Kubeapps developers has written a series detailing the installation of [Kubeapps on a set of VMware TKG clusters with OpenID Connect].
```
andresmgot, line 8: you forgot to add the link here

andresmgot (approved):

docs/user/using-an-OIDC-provider.md
absoludity: Woops, thanks.

PR #2072, opened 2020-09-28
Fix style for several Access URLs

andresmgot:
### Description of the change

Fix the style for the Access URL table when there is more than one URL per svc/ingress (before it was shown everything in the same line without spaces):

Screenshot from 2020-09-28 12-49-09


andresmgot:
_merging this a trivial to tag a new patch release_

PR #2081, opened 2020-09-29
Fix wrong namespace when requesting chart info

### LINKED ISSUE #2075: Chart new version indicator when select all namespaces
Hello

Would be nice to see the indicator icon when new chart version will appear when clicking on "All namespaces"
Currently the indicator is been shown if you select specific namespace

Selection_008

andresmgot:
### Description of the change

Fix an issue causing the request for the chart to use the current namespace (`_all` if all namespaces is selected) rather than the app namespace.

Apart from that, I found a small bug that two releases in different namespaces were using the same `key` in the list.

### Applicable issues

  - fixes #2075

### DIFF
--- a/dashboard/src/actions/apps.ts
+++ b/dashboard/src/actions/apps.ts
@@ -22,9 +22,7 @@ export const receiveApps = createAction("RECEIVE_APPS", resolve => {
   return (apps: hapi.release.Release[]) => resolve(apps);
 });
 
-export const listApps = createAction("REQUEST_APP_LIST", resolve => {
-  return (listingAll: boolean) => resolve(listingAll);
-});
+export const listApps = createAction("REQUEST_APP_LIST");
 
 export const receiveAppList = createAction("RECEIVE_APP_LIST", resolve => {
   return (apps: IAppOverview[]) => resolve(apps);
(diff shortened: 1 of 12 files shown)


dashboard/src/actions/apps.ts
```diff
@@ -244,7 +244,7 @@ export function fetchAppsWithUpdateInfo(
       dispatch(
         getAppUpdateInfo(
           cluster,
-          namespace,
+          app.namespace,
```
absoludity, line 247: A couple of things confuse the code here, imo:


* the `all` function argument is actually `allStatuses` not all apps across namespaces - can we rename it to be explicit (it's always passed as true from the only call-site afaict, so we could even get rid of it).
* the `fetchApps` and subsequent `App.listApps` have functionality to support fetching apps across all namespaces which is the cause of this change, but it'd be a lot easier to reason about this code if it was obvious why `namespace` can't be used here (because it could have the value `_all`. Perhaps a comment, or renaming it `namespace_or_all` or something.

absoludity (approved):

dashboard/src/actions/apps.ts
andresmgot: > * the `all` function argument is actually `allStatuses` not all apps across namespaces - can we rename it to be explicit (it's always passed as true from the only call-site afaict, so we could even get rid of it).

Yes, since we have deleted the code for the Hex UI we can delete it.

> * the `fetchApps` and subsequent `App.listApps` have functionality to support fetching apps across all namespaces which is the cause of this change, but it'd be a lot easier to reason about this code if it was obvious why `namespace` can't be used here (because it could have the value `_all`. Perhaps a comment, or renaming it `namespace_or_all` or something.

:+1:

PR #2084, opened 2020-09-30
Enable manually switching repos-per-namespace to false. Fixes #1923

### LINKED ISSUE #1923: Multiple kubeapps on single cluster cannot distinguish each others global and local app repos
### Description:

Having two kubeapps installations interfere with each other AppRepositories.
See:  https://kubernetes.slack.com/archives/C9D3TSUG4/p1596549628152100

1. create a ns1, install kubeapps into it, defining an initial repo: https://github.com/kubeapps/kubeapps/blob/master/chart/kubeapps/values.yaml#L226
2. create a ns2, install kubeapps into it, defining an initial repo as in 1, set helm3=true
3. Observe that the catalog of charts is there, but when you click an app to upgrade it it will fail because it cannot find the chart.
4. k -n ns1 get cronjobs will show TWO cronjobs, despite there is only one repo added for this installation, i.e. it discovers the apprepo defined in both ns1 and ns2 - same for the other install.

# Fix
If I manually delete the extraneous jobs it works again.

I noticed the problem when upgrading from 3.7.1 to 3.7.4 (chart version)

absoludity:
### Description of the change

See the recent discussion on #1923.

Switching this flag on, for a specific case required by @davidkarlsen , would require updating the apprepository controller deployment to set the flag as `--repos-per-namespace=false`.

That said, I'm guessing you'd also need this flag on the deployment running tiller-proxy?

### Benefits

Enables manually configuring the app-repository controller deployment to only watch the kubeapps namespace.

### Possible drawbacks

Requires manual configuration, but this is OK IMO as it's a fix for a temporary
issue (where David is running Kubeapps with tiller-proxy in one namespace and
without in another.

### Applicable issues

  - fixes #1923

### DIFF
--- a/cmd/apprepository-controller/main.go
+++ b/cmd/apprepository-controller/main.go
@@ -69,7 +69,13 @@ func main() {
 
 	// We're interested in being informed about cronjobs in kubeapps namespace only, currently.
 	kubeInformerFactory := kubeinformers.NewSharedInformerFactoryWithOptions(kubeClient, 0, kubeinformers.WithNamespace(namespace))
-	apprepoInformerFactory := informers.NewSharedInformerFactory(apprepoClient, 0)
+	// Enable app repo scanning to be manually set to scan the kubeapps repo only. See #1923.
+	var apprepoInformerFactory informers.SharedInformerFactory
+	if reposPerNamespace {
+		apprepoInformerFactory = informers.NewSharedInformerFactory(apprepoClient, 0)
+	} else {
+		apprepoInformerFactory = informers.NewFilteredSharedInformerFactory(apprepoClient, 0, namespace, nil)
+	}
 
 	controller := NewController(kubeClient, apprepoClient, kubeInformerFactory, apprepoInformerFactory, namespace)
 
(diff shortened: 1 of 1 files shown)


andresmgot (approved):

davidkarlsen:
This will do exactly what we hoped - thanks a lot. Will the toggle be available in kubeapps helm-chart too?

absoludity:
> This will do exactly what we hoped - thanks a lot. Will the toggle be available in kubeapps helm-chart too?

No, for now I'm just making it a manual option you'll need to edit on the deployment. Until we sort out what we want to enable longer-term in this regard, I just want to unblock you without making a decision about the chart UI.

absoludity:
Note that this will only be available in the 2.x releases, not your tiller-proxy as mentioned above.

andresmgot:
> No, for now I'm just making it a manual option you'll need to edit on the deployment. Until we sort out what we want to enable longer-term in this regard, I just want to unblock you without making a decision about the chart UI.

I would be fine (while we maintain the flag) to be able to set it from the chart using the `apprepository` configuration (rather than making it a feature flag). Something like `apprepository.reposPerNamespace`. WDYT?

absoludity:
Less keen, but not opposed. The reason I didn't make it a chart option is because currently we only know of this use-case. That said, in separate conv. you mentioned the scenario of a cluster setup so that different users can to test kubeapps in their own namespace, which we saw demo'd - agree it would be helpful there. OK.

absoludity:
See #2094

PR #2085, opened 2020-09-30
Update docs related to access control

### LINKED ISSUE #1927: Documentation for users access management
this is just a suggestion to improve documentation about configuring user's access management.

[The section for read/write app repositories](https://github.com/kubeapps/kubeapps/blob/master/docs/user/access-control.md#app-repositories), talks about binding the role "apprepositories-read". But you dont provide the yaml for that. It is the same than role "kubeapps-apps-read", but it would be confusing for people just following the tutorial and not enough K8s knowledge (like me). By the way, the role name is "apprepositories-read" or" kubeapps-apprepositories-read"?.

It happens the same in the section for write app repositories.

The way of executing the kubectl command changes from what it has been done in previous steps (wold be confusing). An environment variable instead of using --namespace. I suggest using the same way. 

In any case the command seems also confusing about the rolebinding:

```
kubectl create -n $KUBEAPPS_NAMESPACE rolebinding example-kubeapps-repositories-read \
  --clusterrole=kubeapps:$KUBEAPPS_NAMESPACE:apprepositories-read \
  --serviceaccount default:example
```

--clusterole==kubeapps:$KUBEAPPS_NAMESPACE:apprepositories-read

shouldnt be just..?

--clusterole==apprepositories-read

Again, the role name is apprepositories-read, but the one for reading apps is kubeaps-apprepositories-read.

For the apprepositories write section. Shouldnt we use an existing kubernetes role write, like we did with role edit in the apps write?

I am just trying to clarify that, because it is taking me a lot of time to reproduce the whole process (maybe because I am pretty new to Kubernetes)

many thanks

EDIT: As per a [comment from Curtis below](https://github.com/kubeapps/kubeapps/issues/1927#issuecomment-674467601), the confusion may be that the docs indicate `kubeapps:$KUBEAPPS_NAMESPACE:apprepositories-read` where as it may now be a const: `kubeapps:kubeapps:apprepositories-read` (we should double-check why it was changed - I think i did it as part of the per-namespace app repos work, but need to check the details).

andresmgot:
### Description of the change

Checking #1927, it's true that this doc is a bit confusing. In summary I have:

 - Fix some commands that were using the role `kubeapps-applications-read` rather than the ones created by Kubeapps.
 - The `kubeapps-applications-read` clusterrole is actually quite confusing since rather than being a role to just allow read access for app repos, it was more of a cluster-role admin. I have deleted that file and role to avoid confusion.
 - I have also modified the `-write` clusterrole to be able to edit Secrets, this is necessary to modify and remove app repo credentials.
 - I have also removed the `-read` commands from the docs to avoid noise, usually people are interested in either managing repos or not interact with them.

### Applicable issues

  - fixes #1927


docs/user/access-control.md
```diff
@@ -51,82 +50,66 @@ Kubeapps.
...
-Kubeapps `apprepositories-write` ClusterRole in the namespace Kubeapps is installed in
-for users to create and refresh App Repositories in Kubeapps
-
-```bash
-export KUBEAPPS_NAMESPACE=kubeapps
-kubectl create -n $KUBEAPPS_NAMESPACE rolebinding example-kubeapps-repositories-write \
-  --clusterrole=kubeapps:$KUBEAPPS_NAMESPACE:apprepositories-write \
-  --serviceaccount default:example
-```
+The above command command allows people to create app repositories in the Kubeapps namespace, these are called "Global Repositories" since they will be available in any namespace Kubeapps is available. In the other hand, it's also possible to create "Namespaced Repositories" that will be available just in a single namespace. For doing so, users need to have permissions to create app repositories in those namespaces. Read the next section to know how to create those roles.
```
absoludity, line 88: "On the other hand" (not in :) )

absoludity (approved):
Thanks for the extra effort for the better wording throughout too.

Just note, if you grep for kubeapps-applications-read you'll see it in the bash helper scripts too, so we need to switch it there also before landing this.

PR #2090, opened 2020-10-01
Configuration for probot-stale. Fixes #576

### LINKED ISSUE #576: Look into adding a stale issue/PR closer bot
This will help us clean up the issue queue

absoludity:
### Description of the change

I've enabled probot-stale on the kubeapps/kubeapps repo only. This is just the
default configuration as per https://github.com/probot/stale

### Benefits

Is it not obvious? :)

### Applicable issues

  - fixes #576

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
+# Label to use when marking an issue as stale
+staleLabel: wontfix
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
```
andresmgot, line 2: I would increase this to 90 days. In case we want to push back something to the next quarter.

.github/stale.yml
```diff
@@ -0,0 +1,17 @@
...
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
```
andresmgot, line 8: `kind/bug`?

andresmgot (approved):

.github/stale.yml
absoludity: Yeah, makes sense - if we've triaged as a bug.

PR #2091, opened 2020-10-01
Use cluster ServiceToken to retrieve operator logos

### LINKED ISSUE #2076: GetOperatorLogo is not cluster aware
### Description:

After installing the OLM on a k8s cluster which is not the cluster on which Kubeapps is installed, and then browsing to the list of operators to make available, none of the logos are rendered. All requests return a 500 as shown:

operators-logo-500

It appears that the request is going through to the cluster on which Kubeapps is installed (which in this case, does not have the OLM installed, hence the api server is unable to provide the resource).

Note that if you deploy an operator, then the instance does show the icon so must be routed to the correct cluster. It's just on the available operators page.

1. Install Kubeapps in a multi-cluster setup
2. Install the OLM on the cluster on which Kubeapps is not installed
3. Browse the available operators

No logos appear

Logos appear

andresmgot:
### Description of the change

Similar to the approach when retrieving namespaces in other clusters, where a internal service token is used, use that token to retrieve logos. Note that it's not possible to use the user token because the logos are requested as `img` `src` so no authentication info is given for that request.

### Applicable issues

  - fixes #2076

### DIFF
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -288,7 +288,11 @@ func (c *Client) parseDetailsForHTTPClient(details *Details, userAuthToken strin
 	}
 	if details.AppRepositoryResourceNamespace == c.kubeappsNamespace {
 		// If we're parsing a global repository (from the kubeappsNamespace), use a service client.
-		client = c.appRepoHandler.AsSVC()
+		// AppRepositories are only allowed in the default cluster for the moment
+		client, err = c.appRepoHandler.AsSVC(c.kubeappsCluster)
+		if err != nil {
+			return nil, nil, nil, fmt.Errorf("unable to get app repository %q: %v", details.AppRepositoryResourceName, err)
+		}
 	}
 	appRepo, err := client.GetAppRepository(details.AppRepositoryResourceName, details.AppRepositoryResourceNamespace)
 	if err != nil {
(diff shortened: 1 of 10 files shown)


pkg/chart/chart.go
```diff
@@ -288,7 +288,11 @@ func (c *Client) parseDetailsForHTTPClient(details *Details, userAuthToken strin
 	}
 	if details.AppRepositoryResourceNamespace == c.kubeappsNamespace {
 		// If we're parsing a global repository (from the kubeappsNamespace), use a service client.
-		client = c.appRepoHandler.AsSVC()
+		// AppRepositories are only allowed in the default cluster for the moment
+		client, err = c.appRepoHandler.AsSVC(c.kubeappsCluster)
+		if err != nil {
+			return nil, nil, nil, fmt.Errorf("unable to get app repository %q: %v", details.AppRepositoryResourceName, err)
```
absoludity, line 294: "unable to create client"? (or is this a user-facing error msg?)

pkg/kube/kube_handler_test.go
```diff
@@ -530,9 +530,23 @@ func TestDeleteAppRepository(t *testing.T) {
...
+							CAFile:                   "",
+							ServiceToken:             "fake",
+							Insecure:                 true,
+						},
+					},
+				},
 			}
 
-			err := handler.AsSVC().DeleteAppRepository(tc.repoName, tc.requestNamespace)
+			cli, _ := handler.AsSVC("cluster")
```
absoludity, line 548: Just pls handle the error for completeness. I know it seems unnecessary but I did recently get a nil pointer error the other week only to find it was because of an unhandled error.

absoludity (approved):
Great, thanks.

Issue #865, opened 2018-12-04
Evaluate to use allIDs/byID pattern for Redux


andresmgot:
Context: https://github.com/kubeapps/kubeapps/pull/847#discussion_r238124382

Docs: https://redux.js.org/recipes/structuringreducers/normalizingstateshape#designing-a-normalized-state

PR #2093, opened 2020-10-01
Force re-request of the chart version when upgrading

### LINKED ISSUE #2092: Wrong "Changes" when upgrading
### Description:

When checking a second application, after clicking on "Upgrade" the first application is used as the "previous" state so the "Changes" tab shows the difference between one application and the other:

Screenshot from 2020-10-01 16-29-58

andresmgot:
### Description of the change

Fixes #2092

### DIFF
--- a/dashboard/src/components/UpgradeForm/UpgradeForm.tsx
+++ b/dashboard/src/components/UpgradeForm/UpgradeForm.tsx
@@ -121,10 +121,10 @@ function UpgradeForm({
   }, [deployed.values, modifications]);
 
   useEffect(() => {
-    if (deployed.chartVersion?.attributes.version && !version) {
+    if (deployed.chartVersion?.attributes.version) {
       getChartVersion(cluster, repoNamespace, chartID, deployed.chartVersion.attributes.version);
     }
-  }, [getChartVersion, cluster, repoNamespace, chartID, deployed.chartVersion, version]);
+  }, [getChartVersion, cluster, repoNamespace, chartID, deployed.chartVersion]);
 
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


dashboard/src/components/UpgradeForm/UpgradeForm.tsx
```diff
@@ -121,10 +121,10 @@ function UpgradeForm({
   }, [deployed.values, modifications]);
 
   useEffect(() => {
-    if (deployed.chartVersion?.attributes.version && !version) {
+    if (deployed.chartVersion?.attributes.version) {
```
absoludity, line 124: It seems to me we should either be clearing `version` in the state when changing routes, or ensuring that it is specific to the ID of the chart, rather than just re-requesting to ensure it's the right one. Sounds like a use-case for #865 ?

absoludity (approved):
Unblocking, but I think we should instead be storing the chart by id rather than just space for one chart in the state?

dashboard/src/components/UpgradeForm/UpgradeForm.tsx
andresmgot: Agree, the state can benefit from the refactor (but better fix this bug now)

absoludity: FWIW, it's not about whether the state can benefit from a refactor IMO, but whether we/contributors will save time in the long-term by being able to reason `is this chart already loaded in the state`, rather than creating a function, later finding a bug, updating with a cludge which re-loads the chart every time, etc. A little bit of house-keeping with each PR goes a long way.

PR #2094, opened 2020-10-02
Enable watching only the namespace in which Kubeapps is running.

absoludity:
### Description of the change

Follows on from #2084 and the discussion there.

I initially didn't want to expose this via a chart option, but @andresmgot pointed out that there's another case where this would be useful: a cluster setup for training and learning where different users each have Kubeapps installed in their own namespace (in which case, we wouldn't want each installation picking up the app repositories of the other installations).


andresmgot (approved):

andresmgot:
cc/ @davidkarlsen

PR #1299, opened 2019-11-19
Kubeapps chart - Lint chart

juan131:
Signed-off-by: juan131 <juan@bitnami.com>

**Description of the change**

- Change chart YAML(s) to make it follow Bitnami standards (and [Helm Best Practices](https://helm.sh/docs/topics/chart_best_practices))
- Some utilities such as **yamllint**, **kubeval** or **helm lint** have been used to ensure there are no syntax errors on the chart's manifests (see https://github.com/bitnami/charts/pull/1452)

PR #2086, opened 2020-09-30
Update apprepositories-secret.yaml

obeyler:
### Description of the change
only remove a '-' it causes Json/yaml  encode/decode failure

### Benefits

make the repository with seflf certificate possible

PR #2096, opened 2020-10-05
Fix initialRepos setup when there are multiple secrets

andresmgot:
### Description of the change

Reported in slack:

https://kubernetes.slack.com/archives/C9D3TSUG4/p1601819799005800

If several initialRepos have an associated secret, only the last one gets created. The three dashes line was (accidentally?) removed in #1299.

### DIFF
--- a/chart/kubeapps/Chart.yaml
+++ b/chart/kubeapps/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kubeapps
-version: 4.0.2
+version: 4.0.3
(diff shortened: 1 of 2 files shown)


absoludity (approved):
That's the second PR in a week from accidental fall-out in #1299 (the other being #2086). Smaller PRs for review? Testing of templated output (would have caught the issue in #2086, perhaps not here).

andresmgot:
> Smaller PRs for review?

That'd have helped if this was not the result of an automatic change.

PR #2102, opened 2020-10-07
List apprepos in all namespaces

andresmgot:
### Description of the change

Fix the case in which "All Namespaces" is selected and the user list AppRepositories.

### DIFF
--- a/dashboard/src/components/Config/AppRepoList/AppRepoButton.test.tsx
+++ b/dashboard/src/components/Config/AppRepoList/AppRepoButton.test.tsx
@@ -107,3 +107,19 @@ it("calls updateRepo when submitting and there is a repo available", () => {
   (wrapper.find(AppRepoForm).prop("onSubmit") as any)();
   expect(updateRepo).toHaveBeenCalled();
 });
+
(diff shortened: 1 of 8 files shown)


absoludity (approved):
I had to run this to verify how the issue was fixed - great. The only odd UX for me here is that I can "Add app repository" while browsing all namespaces, but can't edit/refresh/delete them. Not a big issue, not sure what would be best either (probably being able to edit/refresh/delete if I have perms without having to change to the kubeapps namespace).

andresmgot:
> The only odd UX for me here is that I can "Add app repository" while browsing all namespaces, but can't edit/refresh/delete them. Not a big issue, not sure what would be best either (probably being able to edit/refresh/delete if I have perms without having to change to the kubeapps namespace).

I agree, I found it confusing too. I think it's better to disable the form in that case so you need to select a namespace (the same you need to do so for creating apps or operators).

andresmgot:
Apart from disabling the button:

Screenshot from 2020-10-08 13-22-43

I have fixed the link, and edit and delete buttons for namespaced repos that were broken if "All namespaces" was selected.

PR #2103, opened 2020-10-07
Fix setting number values in form

andresmgot:
### Description of the change

Found that the default value for number values was not being set in the form (e.g. `replicas: 1`). Turned out that `isEmpty(1)` returns `true`.


dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/TextParam.test.tsx
```diff
@@ -201,5 +197,22 @@ it("a change in the param property should update the current value", () => {
       value: "foo",
     },
   });
-  expect(input.prop("value")).toBe("");
```
andresmgot, line 204: this test was not properly testing what it should

absoludity (approved):
Yeah, `isEmpty` is testing for empty *objects*: "Objects are considered empty if they have no own enumerable string keyed properties."

Issue #1756, opened 2020-05-30
Certain config elements become json when going to the application deployment page


davidkarlsen:
### Description:

Certain config values become json in the upgrade page of an app.

### Steps to reproduce the issue:

1. Go to applications and select an application, scroll down the the "installation values" section and observe that all values are yaml
2. Click upgrade, observe that the section of install values now is a mix of yaml and json as outlined below

### Describe the results you received:

```
env: [ { name: TZ, value: Europe/Oslo } ]

finods:
  {
    application: aml-cmt-api,
    applicationEnvironment: p1,
    group: eos,
    systemShortname: aml
  }
```

### Describe the results you expected:
```
env:
- name: TZ
  value: Europe/Oslo
finods:
  application: aml-cmt-api
  applicationEnvironment: p1
  group: eos
  systemShortname: aml
```

### Version of Helm, Kubeapps and Kubernetes:

- Output of `helm version`: 

```
helm: N/A - kubeapps runs helm
```

- Output of `helm list <kubeapps-release-name>`:

```
kubeapps-helm3  kubeapps-helm3  2               2020-05-29 11:42:07.153037457 +0000 UTC deployed        kubeapps-3.6.0  v1.10.0    
```

- Output of `kubectl version`:

```
kubectl version
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.3", GitCommit:"2e7996e3e2712684bc73f0dec0200d64eec7fe40", GitTreeState:"clean", BuildDate:"2020-05-21T14:51:23Z", GoVersion:"go1.14.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.5+icp-ee", GitCommit:"eb4df6c6fb47f5b4fd1ed8bfbfe2d0ed5ea636e1", GitTreeState:"clean", BuildDate:"2019-05-08T02:18:32Z", GoVersion:"go1.11.5", Compiler:"gc", Platform:"linux/amd64"}
```

davidkarlsen:
Releated: https://github.com/kubeapps/kubeapps/issues/1683

absoludity:
[...]

PR #2104, opened 2020-10-08
Avoid irrelevant changes in diff

### LINKED ISSUE #1293: Changes tab shows a lot of whitespace differences
This is a follow-up to #1235. While testing a local deploy, I noticed the changes tab is showing a lot of white-space. I think this is known (and due to the yaml parsing library's handling of whitespace for comments - though I don't see an [upstream issue](https://github.com/eemeli/yaml/issues)?). Just noting so we can at some point fix it (probably upstream). See:

kubeapps-whitespace-changes

andresmgot:
### Description of the change

Avoid showing non important changes (like deleted blank lines), when showing the difference between the default values and the ones to submit.

Screenshot from 2020-10-08 15-34-36

Note that these blank lines (and other changes) are performed when the values are parsed as YAML objects (in order to react to a change in the form).

### Applicable issues

  - fixes #1293

### DIFF
--- a/dashboard/src/components/DeploymentFormBody/DeploymentFormBody.test.tsx
+++ b/dashboard/src/components/DeploymentFormBody/DeploymentFormBody.test.tsx
@@ -1,0 +1,64 @@
+import * as React from "react";
+import { act } from "react-dom/test-utils";
+import { defaultStore, mountWrapper } from "shared/specs/mountWrapper";
+import { IChartState, IChartVersion } from "shared/types";
+import BasicDeploymentForm from "./BasicDeploymentForm";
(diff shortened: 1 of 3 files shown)


absoludity (approved):
Great - more than a few times I've had to manually avoid this in a video demo. But will this do more than whitespace formatting and also reformat default some yaml to json values as in #1756, or is that just when inserting data?

andresmgot:
> But will this do more than whitespace formatting and also reformat default some yaml to json values as in #1756, or is that just when inserting data?

Yes, that's only when data has been modified so it's unrelated to this PR. That issue is related to how JSON patches are applied (when calculating modifications when upgrading).

PR #2122, opened 2020-10-22
Fix resource table when working with lists

### LINKED ISSUE #2120: An error occurred: Cannot read property 'name' of undefined
### Description:

`An error occurred: Cannot read property 'name' of undefined` during operator object install

1. Install kubeapps
2. Choose an operator, like etcd
3. change the name (or possibly use the default name but I changed the name)
4. deploy it
5. See `An error occurred: Cannot read property 'name' of undefined`

`An error occurred: Cannot read property 'name' of undefined`
Show information of deployment described in the documentation 

```
helm version
version.BuildInfo{Version:"v3.2.4", GitCommit:"0ad800ef43d3b826f31a5ad8dfbb4fe05d143688", GitTreeState:"clean", GoVersion:"go1.13.12"}

helm list -n kubeapps
kubeapps        kubeapps        1               2020-10-21 12:30:32.2962579 -0400 EDT   deployed        kubeapps-4.0.4  v2.0.0

kubectl version
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.2", GitCommit:"f5743093fd1c663cb0cbc89748f730662345d44d", GitTreeState:"clean", BuildDate:"2020-09-16T13:41:02Z", GoVersion:"go1.15", Compiler:"gc", Platform:"windows/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.2", GitCommit:"f5743093fd1c663cb0cbc89748f730662345d44d", GitTreeState:"clean", BuildDate:"2020-09-16T13:32:58Z", GoVersion:"go1.15", Compiler:"gc", Platform:"linux/amd64"}
```

andresmgot:
### Description of the change

At the moment, the code was assuming that there were one resource reference per resource in the state. Sometimes the reference points to a list so there will be more resources than references.

This PR also handles the case in which there are no references at all (but this is not happening IRL).

### Applicable issues

  - fixes #2120

### DIFF
--- a/dashboard/src/components/AppView/ResourceTable/ResourceTable.tsx
+++ b/dashboard/src/components/AppView/ResourceTable/ResourceTable.tsx
@@ -91,7 +91,7 @@ function ResourceTable({ id, title, resourceRefs }: IResourceTableProps) {
     () =>
       resources.map((resource, index) => {
         // When the resourceRef is a list, the list of references will be just one
-        const ref = resourceRefs[index] || resourceRefs[0];
+        const ref = resourceRefs.length === 1 ? resourceRefs[0] : resourceRefs[index];
         if (ref) {
           return getData(
             ref.name,
--- a/dashboard/src/components/AppView/ResourceTable/ResourceTable.tsx
+++ b/dashboard/src/components/AppView/ResourceTable/ResourceTable.tsx
@@ -83,17 +83,25 @@ function ResourceTable({ id, title, resourceRefs }: IResourceTableProps) {
     flattenResources(resourceRefs, state.kube.items),
   );
 
-  const columns = useMemo(() => getColumns(resourceRefs[0]), [resourceRefs]);
+  const columns = useMemo(
+    () => (resourceRefs.length ? getColumns(resourceRefs[0]) : OtherResourceColumns),
+    [resourceRefs],
+  );
   const data = useMemo(
     () =>
-      resources.map((resource, index) =>
-        getData(
-          resourceRefs[index].name,
-          columns.map(c => c.accessor),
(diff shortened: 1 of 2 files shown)


dashboard/src/components/AppView/ResourceTable/ResourceTable.tsx
```diff
@@ -83,17 +83,25 @@ function ResourceTable({ id, title, resourceRefs }: IResourceTableProps) {
...
-        getData(
-          resourceRefs[index].name,
-          columns.map(c => c.accessor),
-          columns.map(c => c.getter),
-          resource,
-        ),
-      ),
+      resources.map((resource, index) => {
+        // When the resourceRef is a list, the list of references will be just one
+        const ref = resourceRefs[index] || resourceRefs[0];
```
absoludity, line 94: This looks a little ambiguous to me, in that if there were 3 resource refs and 5 resources it would get the refs for the first 3 iterations and then the 0th for the fourth and fifth (not that that's possible for the data you expect). If you expect that the length of resourceRefs will be 1 specifically in this case, why not make that explicit:

```
const ref = (resourceRefs.length === 1) ? resourceRefs[0] : resourceRefs[index];  
```
so that if our expectations aren't met, we'll still get an error as we should, rather than masking it with odd behaviour?

absoludity (approved):

dashboard/src/components/AppView/ResourceTable/ResourceTable.tsx
andresmgot: Got it, yes, it's also more readable that way.

PR #2123, opened 2020-10-22
Avoid double v prefix in the AppList

andresmgot:
### Description of the change

The appVersion field in the `Chart.yaml` may contain already the `v` prefix, and we are prefixing another `v` again:

Screenshot from 2020-10-22 13-04-18

In any case, the `appVersion` doesn't need that prefix so I am also removing it from our chart release process.

### Applicable issues

I'd swear I fixed this already but I couldn't find it, maybe I forgot to create the PR :thinking:

### DIFF
--- a/script/chart_sync_utils.sh
+++ b/script/chart_sync_utils.sh
@@ -71,7 +71,7 @@ replaceImage() {
 updateRepo() {
     local targetRepo=${1:?}
     local targetTag=${2:?}
-    local targetTagWithoutV=${targetTag/v/}
+    local targetTagWithoutV=${targetTag#v}
     local targetChartPath="${targetRepo}/${CHART_REPO_PATH}"
     local chartYaml="${targetChartPath}/Chart.yaml"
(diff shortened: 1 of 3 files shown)


script/chart_sync_utils.sh
```diff
@@ -71,6 +71,7 @@ replaceImage() {
 updateRepo() {
     local targetRepo=${1:?}
     local targetTag=${2:?}
+    local targetTagWithoutV=${targetTag/v/}
```
absoludity, line 74: A bit pedantic, but you probably want to strip from the front of the var, rather than replacing anywhere in the var?
```
$ VERSIONTEST=v1.2.3 && echo ${VERSIONTEST#v}
1.2.3

$ VERSIONTEST=1.2.3v && echo ${VERSIONTEST#v}
1.2.3v

$ VERSIONTEST=1.2.3v && echo ${VERSIONTEST/v/}
1.2.3
```

absoludity (approved):

PR #2126, opened 2020-10-26
Change `-` to `--` for auth-proxy additionalFlags

lindhe:
I believe this changed with a recent version of auth-proxy, and that we
just missed updating this documentation.

Just close this PR if I'm wrong.


andresmgot (approved):
That's correct. Thanks!

PR #2130, opened 2020-10-28
Show upgrade form even after failure

### LINKED ISSUE #2121: A failure during the upgrade process causes changes to be lost
### Description:

The content of the form is hidden when an upgrade error happens, forcing the user to refresh and lose the changes.

andresmgot:
### Description of the change

Avoid to lose changes when an upgrade returns an error:

Screenshot from 2020-10-28 18-29-55

Different things in this PR:
 - I have split errors in the state depending on the action that causes it. This allows to differentiate between an error fetching the app and the error upgrading the app.
 - Avoid setting the state as "isFetching" when upgrading the app since that causes the UpgradeForm to re-render. The ideal solution is to split this as well to isFetching/isCreating/isUpgrading... but I didn't go that far.The upgrading loading wrapper is handled as in-component state.
 - Ensure that the upgrade form is still shown even if there is an error.

### Applicable issues

  - fixes #2121


dashboard/src/components/AppUpgrade/AppUpgrade.tsx
```diff
@@ -12,7 +12,8 @@ export interface IAppUpgradeProps {
   app?: IRelease;
   appsIsFetching: boolean;
   chartsIsFetching: boolean;
-  appsError: Error | undefined;
+  getError?: Error;
+  upgradeError?: Error;
```
absoludity, line 16: It's not clear to me why you're adding different error props (get, upgrade) with a plain Error type, rather than just leaving the one error prop and [custom error types](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#Custom_Error_Types)? Wouldn't that be neater (and not lead to adding an error prop per error type across components)?

dashboard/src/reducers/apps.ts
```diff
@@ -19,8 +19,14 @@ const appsReducer = (
       return { ...state, isFetching: true };
     case getType(actions.apps.receiveApps):
       return { ...state, isFetching: false, items: action.payload };
-    case getType(actions.apps.errorApps):
-      return { ...state, isFetching: false, error: action.payload };
+    case getType(actions.apps.errorGetApp):
+      return { ...state, isFetching: false, getError: action.payload };
```
absoludity, line 23: The same in the state - not sure why we can't just have one error which can be different custom types.

absoludity (approved):
+1 but I think it'd be much neater if we avoid adding props/fields per error type and instead use custom error types.

dashboard/src/components/AppUpgrade/AppUpgrade.tsx
andresmgot: I am not sure if that would be more neat? For two reasons:
 - It's technically possible (although it won't happen) that there were two errors colliding (error fetching an deleting an app for example). If there were only one prop for the error, we would lose information about one of them.
 - We would need to check the error type anyway. For example, in this PR, we would need to check `if (err.Constructor === FetchError)` or alike in the `AppUpgrade` component, and in case it's not, forward the error to the `UpgradeForm` which is not much more neat.

Are you thinking on something I'm missing?

absoludity: > I am not sure if that would be more neat?

Neater that the alternative of adding a new error prop for each possible error? :P 

> For two reasons:
> 
>     * It's technically possible (although it won't happen) that there were two errors colliding (error fetching an deleting an app for example). If there were only one prop for the error, we would lose information about one of them.

I'm not sure what the issue is here - we normally don't continue after the first error - and even if this was possible in practise, it's not hard to create a [custom error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#ES6_Custom_Error_Class) handling whatever situation you want.

> 
>     * We would need to check the error type anyway. For example, in this PR, we would need to check `if (err.Constructor === FetchError)` or alike in the `AppUpgrade` component, and in case it's not, forward the error to the `UpgradeForm` which is not much more neat.

The neatness isn't in the handling of the error, but in not having to add an extra prop for each type of error we can encounter here.

That said, I think the handling can be just as neat. Here are some different scenarios. Let's say we were handling different types of errors in the one spot:

```
switch (err.Constructor) {
  case FetchError:
    ...
    break;
  case ...
}
```

Or just wanting to handle one type of error as you've mentioned:
```
if (err.Constructor === FetchError) {...}
```

although, I'm not saying you should, but remember you can put whatever syntactical sugar you like in our error handling, so it could look like:

```
if errors.IsA(err, FetchError) {}
```

I think your last point (correct me - not 100%) is that in the `AppUpgrade` component you'd need to check the error type and so conditionally forward it to the `UpgradeForm`? I'd just set the error prop unconditionally, you already handle a fetch error here and don't continue:
```
  if (errors.IsA(error, FetchError)) {
    return <Alert theme="danger">Unable to retrieve the current app: {error.message}</Alert>;
  }
  ...
    <UpgradeForm
    ...
    error={error}
    ...
    />
```

So I'd expect the `UpgradeForm` to do the same:
```
if (errors.IsA(error, UpgradeError)) { 
    return <Alert theme="danger">Some specific upgrade error handling</Alert>;
}
```

(Unrelated to this discussion, you might even want to throw the error when a leaf component doesn't handle an error so it's caught and handled generically by the error boundary if our specific handling doesn't handle it).

> 
> 
> Are you thinking on something I'm missing?

Only that the issue of neatness is, in my mind, centered around whether we need to add an extra prop for each different type of error, as opposed to handling a single error according to its type.

I've +1'd earlier, I just worry that this creeps in as a pattern in other places (or already has) so would be great to think about and handle the right way now. Up to you.

andresmgot: > > I am not sure if that would be more neat?
> 
> Neater that the alternative of adding a new error prop for each possible error? :P
> 

You would need to add a new error class for each possible error anyway (that's why I say it's more or less the same).

> > For two reasons:
> > ```
> > * It's technically possible (although it won't happen) that there were two errors colliding (error fetching an deleting an app for example). If there were only one prop for the error, we would lose information about one of them.
> > ```
> 
> I'm not sure what the issue is here - we normally don't continue after the first error - and even if this was possible in practise, it's not hard to create a [custom error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#ES6_Custom_Error_Class) handling whatever situation you want.
> 

No, the thing is that if we only store one error, we could overwrite the original error. I am thinking about this case:

 - A request is send to delete app A
 - The deletion is done but some error occurred so we set a deletion error
 - After the deletion is done, a refresh is trigger to retrieve information again
 - An error occurs because A no longer exists so we set a fetch error

The fetch error above would remove the information about the deletion error (which is the important one). That scenario doesn't happen IRL but that was what I meant.

> I think your last point (correct me - not 100%) is that in the `AppUpgrade` component you'd need to check the error type and so conditionally forward it to the `UpgradeForm`? I'd just set the error prop unconditionally, you already handle a fetch error here and don't continue:
> 
> ```
>   if (errors.IsA(error, FetchError)) {
>     return <Alert theme="danger">Unable to retrieve the current app: {error.message}</Alert>;
>   }
>   ...
>     <UpgradeForm
>     ...
>     error={error}
>     ...
>     />
> ```
> 
> So I'd expect the `UpgradeForm` to do the same:
> 
> ```
> if (errors.IsA(error, UpgradeError)) { 
>     return <Alert theme="danger">Some specific upgrade error handling</Alert>;
> }
> ```

That was the behavior prior to this PR, the goal is to not return when there is an upgrade error to allow people to modify the UpgradeForm with new data so the error can be fixed, without losing already introduced input.

> (Unrelated to this discussion, you might even want to throw the error when a leaf component doesn't handle an error so it's caught and handled generically by the error boundary if our specific handling doesn't handle it).

We do that with the error boundary, but that is only for unexpected errors thrown (not errors set in the Redux state). We could change the approach and have all the errors handled by the error boundary but that would require a fairly big refactor.

> > Are you thinking on something I'm missing?
> 
> Only that the issue of neatness is, in my mind, centered around whether we need to add an extra prop for each different type of error, as opposed to handling a single error according to its type.
> 
> I've +1'd earlier, I just worry that this creeps in as a pattern in other places (or already has) so would be great to think about and handle the right way now. Up to you.

You still need to add error classes for each different type of error but it's true that it can be reused in other scenarios (so `FetchError` can be used both for apps and operators for example).

dashboard/src/actions/apps.ts
```diff
@@ -8,12 +8,17 @@ import { hapi } from "../shared/hapi/release";
...
+  DeleteError,
+  FetchError,
   IAppOverview,
   IChartUpdateInfo,
   IChartVersion,
   IRelease,
   IStoreState,
+  RollbackError,
   UnprocessableEntity,
+  UpgradeError,
```
absoludity, line 21: Just a thought as I see this repeated, we could instead have a `shared/errors` then here just do `import * as errors from "shared/errors";` , then within the code just use `errors.CreateError` for example. Fine either way.

dashboard/src/actions/apps.ts
```diff
@@ -50,12 +55,9 @@ export const requestRollbackApp = createAction("REQUEST_ROLLBACK_APP");
...
 export const receiveRollbackApp = createAction("RECEIVE_ROLLBACK_APP_CONFIRMATION");
 
-export const errorApps = createAction("ERROR_APPS", resolve => {
-  return (err: Error) => resolve(err);
-});
-
-export const errorDeleteApp = createAction("ERROR_DELETE_APP", resolve => {
-  return (err: Error) => resolve(err);
+export const errorApp = createAction("ERROR_APP", resolve => {
+  return (err: FetchError | CreateError | UpgradeError | RollbackError | DeleteError) =>
```
absoludity, line 59: Similarly, we could just do `(err: CustomError)` here, or if we want to not include the existing servererror, define all these app errors as extending an `AppError` and use that? (At least, I assume the typescript annotation here takes that into account, haven't tried it).

dashboard/src/components/AppUpgrade/AppUpgrade.tsx
```diff
@@ -95,13 +95,10 @@ function AppUpgrade({
     }
   }, [getDeployedChartVersion, app, chart, repoName, repoNamespace, cluster]);
 
-  if (appsError) {
-    return (
-      <Alert theme="danger">
-        An error occurred while processing the application: {appsError.message}
-      </Alert>
-    );
+  if (error && error.constructor === FetchError) {
```
absoludity, line 98: Just as an example, this could be:
```
if (errors.IsA(e, errors.FetchError) {
...
```
if we added that to the errors module for use. Again, not necessary, just mentioning.

absoludity (approved):
Still approved :)

dashboard/src/components/AppUpgrade/AppUpgrade.tsx
andresmgot: I prefer to avoid that syntactic sugar because it confuses the type checking. For example, I am defining `error: FetchError | UpgradeError`, if for some mistake I write `isA(e, DeleteError)` the compiler won't complain (but it will never be true) while if I do `e.constructor === DeleteError` I get the compilation error saying that the condition will never be true.

PR #2131, opened 2020-10-29
Fix wrong link generation when the ingress URL contains a regex  (#1120)

### LINKED ISSUE #1120: Regex in path gets included
When using regex in the path, that too gets included.

Screenshot from 2019-08-09 11-11-29

Migrated from #673

antgamdia:
### Description of the change

In #1120 , it was noticed that a link (i.e., <a ...href=....> tag) is being created for each host/path in the ingress URL tables.
The ingress configuration allows users to enter wildcards and regular expressions, however, the hyperlinks cannot target to these URL since they are not reacheble.
The change proposed in this PR is simply to detect whether the given ingress URL is, actually, a valid URL (see discussion of what mean by a valid URL in #1120). If so, it renders the <a> html tag; otherwise, it just renders an < span>.

### Benefits

For users, there will not be any misleading hyperlink targets pointing nowhere.

### Possible drawbacks

Due to the current regex, it is possible for some false positives/negatives to be included. Nevertheless, to the besy of my knowledge, there is no upstream accepted solution for doing so.

### Applicable issues

  - fixes #1120

### Additional information

N/A


dashboard/src/components/AppView/AccessURLTable/AccessURLTable.tsx
```diff
@@ -166,7 +182,13 @@ export default function AccessURLTable({ ingressRefs, serviceRefs }: IAccessURLT
       .concat(
         allIngresses.map(ingress => {
           return {
-            url: ingress.item ? getAnchors(GetURLItemFromIngress(ingress.item).URLs) : "Unknown",
+            url: ingress.item
+              ? GetURLItemFromIngress(ingress.item).URLs.map(
+                  // check whether each URL is, indeed, a valid URL.
+                  // If so, render the <a>, othersiwe, render a simple <span>
+                  url => (IsURL(url) ? getAnchor(url) : getSpan(url)),
```
andresmgot, line 189: great, can you create a test in AccessURLTable.test for covering the case in which the `path` is a regexp? You can verify that there is no anchor in that case.

Tip: You can copy the test setup from "should show the table with available ingresses"

andresmgot (commented):
Good job figuring this out :) I have some small comments.

dashboard/src/components/AppView/AccessURLTable/AccessURLTable.test.tsx
```diff
@@ -178,6 +178,41 @@ context("when the app contains ingresses", () => {
...
+    const ingressRefs = [{ name: "svc", getResourceURL: jest.fn(() => url) } as any];
+    const state = { kube: { items: { [url]: ingressItem } } };
+    const store = getStore(state);
+    const wrapper = mountWrapper(
+      store,
+      <AccessURLTable {...defaultProps} ingressRefs={ingressRefs} />,
+    );
+    expect(wrapper.find("Table")).toExist();
+    expect(
+      wrapper.find("a").findWhere(a => a.prop("href") === "http://foo.bar/ready(/|$)(.*)"),
```
andresmgot, line 211: I would do `expect(wrapper.find("a")).not.toExistst()` so we are sure we are not parsing the URL in a weird way and generate a link.

dashboard/src/components/AppView/AccessURLTable/AccessURLTable.test.tsx
```diff
@@ -178,6 +178,41 @@ context("when the app contains ingresses", () => {
...
+    ).not.toExist();
+    expect(wrapper.find("span").first()).toHaveText("http://foo.bar/ready(/|$)(.*)");
+    expect(wrapper).toMatchSnapshot();
```
andresmgot, line 214: you can omit the snapshot

dashboard/src/components/AppView/AccessURLTable/AccessURLTable.test.tsx
```diff
@@ -178,6 +178,41 @@ context("when the app contains ingresses", () => {
...
+    const store = getStore(state);
+    const wrapper = mountWrapper(
+      store,
+      <AccessURLTable {...defaultProps} ingressRefs={ingressRefs} />,
+    );
+    expect(wrapper.find("Table")).toExist();
+    expect(
+      wrapper.find("a").findWhere(a => a.prop("href") === "http://foo.bar/ready(/|$)(.*)"),
+    ).not.toExist();
+    expect(wrapper.find("span").first()).toHaveText("http://foo.bar/ready(/|$)(.*)");
```
andresmgot, line 213: This is a bit risky, it would fail if we add any other `span` before. Better to do `wrapper.find("span").findWhere(s => s.text() === "http://...` so we find it wherever it is.

andresmgot (approved):
Thanks for the changes! I have some small comments but you can merge once you apply those.

absoludity (approved):
Nice work Antonio!

dashboard/src/components/AppView/AccessURLTable/AccessURLTable.test.tsx
antgamdia: The span element seems not to have a .text(), so I first have to render the cheerio component and then access the actual text: `wrapper.find("span").findWhere(s => s.render().text() ...`, if you have a better solution, please let me know.

dashboard/src/components/AppView/AccessURLTable/AccessURLTable.tsx
```diff
@@ -77,18 +77,34 @@ function flattenIngresses(ingresses: Array<IKubeItem<IResource | IK8sList<IResou
...
 
 function getAnchors(URLs: string[]) {
-  return URLs.map(URL => (
+  return URLs.map(URL => getAnchor(URL));
+}
+
+function getAnchor(URL: string) {
+  return (
     <div className="margin-b-sm">
       <a href={URL} target="_blank" rel="noopener noreferrer" key={URL}>
```
andresmgot, line 86: running the tests for this, I noticed several warnings like `Warning: Each child in a list should have a unique "key" prop.`. For this function, the `key={URL}` needs to be in the `div`. Note that any component returned in an array should have the `key` property so you need to update `getSpan` and `getUnknown`.

dashboard/src/components/AppView/AccessURLTable/AccessURLTable.tsx
```diff
@@ -77,18 +77,34 @@ function flattenIngresses(ingresses: Array<IKubeItem<IResource | IK8sList<IResou
...
+}
+
+function getUnknown() {
+  return <span>Unknown</span>;
 }
 
 function getNotes(resource?: IResource) {
   if (!resource) {
-    return <span>Unknown</span>;
+    return getUnknown;
```
andresmgot, line 107: just noticed that this should be `getUnknown()`

andresmgot (changes requested):

dashboard/src/components/AppView/AccessURLTable/AccessURLTable.test.tsx
andresmgot: I double checked and the method `.text()` should be present (in fact, it should exist for any element). Tried this and worked:

```
    const matchingSpans = wrapper.find("span").findWhere(s => s.text().includes("foo.bar/ready"));
```

antgamdia: Mmm, don't know either... but now it's working as you mention. Perhaps a misconfiguration in my linter led me to think there was an error with the .text() method. Thank you for your comments!

andresmgot (approved):
Thanks for the changes

PR #2132, opened 2020-10-29
Consistent semver usage of @clr/ui dependency

antgamdia:
### Description of the change

For newcomers developers, it is a bit weird to have to manually include the Clarity dependency (as dockerfile did). Despite the docs mentions the only command to run the dashboard is `yarn install` and `yarn run start`, it was necessary to manually put the clarity `clr-ui.min.css` file.
What's more, the clarity dependency was already included in the package.json, but with major version = 2 (instead of 4).

This PR removes every manual importation of the  `clr-ui.min.css` file (and the old one from the HEX UI) and changes the importation to the `node_modules` folder (`../../../node_modules/@clr/ui/clr-ui.min.css` instead of `../../clr-ui.min.css`).

### Benefits

Since developers already have to exec `yarn install` in order to install the required deps, including Clarity in the list of deps improves traceability and enhances the dev experience for newcomers.

### Possible drawbacks

Dockerfile relied on a very specific version of Clarity instead of leveraging from the semver. This PR sets the clarity dep in `^4.0.1`, that is, anything 4.X.X will be accepted.

### Applicable issues

N/A

### Additional information

If this PR does not become accepted, a change in the Dashboard developer documentarios will be required to explain that is required to manually add a dependency.

### DIFF
--- a/dashboard/src/components/Layout/Clarity.tsx
+++ b/dashboard/src/components/Layout/Clarity.tsx
@@ -32,7 +32,7 @@ import {
 import "@clr/core/icon/register.js";
 import * as React from "react";
 
-import "../../clr-ui.min.css";
+import "../../../node_modules/@clr/ui/clr-ui.min.css";
 
 Icons.addIcons(
   angleIcon,
--- a/dashboard/src/components/Layout/Clarity.tsx
+++ b/dashboard/src/components/Layout/Clarity.tsx
@@ -32,7 +32,7 @@ import {
 import "@clr/core/icon/register.js";
 import * as React from "react";
 
-import "../../../node_modules/@clr/ui/clr-ui.min.css";
+import "@clr/ui/clr-ui.min.css";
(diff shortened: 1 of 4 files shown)


dashboard/src/components/Layout/Clarity.tsx
```diff
@@ -32,7 +32,7 @@ import {
 import "@clr/core/icon/register.js";
 import * as React from "react";
 
-import "../../clr-ui.min.css";
+import "../../../node_modules/@clr/ui/clr-ui.min.css";
```
andresmgot, line 35: the `node_modules` is in the path so you can do:

```
import "@clr/ui/clr-ui.min.css";
```

andresmgot (approved):
Thanks @antgamdia! Just a minor comment.

andresmgot (approved):
LGTM, thanks!

PR #2138, opened 2020-11-03
Add a /refresh endpoint for forcing an update in an AppRepository

### LINKED ISSUE #2062: [FEATURE REQUEST] refresh repository by webhook
Hello

Per the conversion here: #2057  and the comment https://github.com/kubeapps/kubeapps/issues/1738#issuecomment-629087081
Would be nice to see webhook refresh mechanism of the repositories

When triggering the webhook, the related repository get refreshed

antgamdia:
### Description of the change

As discussed in #2062 , an endpoint to trigger a refresh in a given repository was requested. Internally, a refresh is being triggered when resyncRequests property changes. This PR creates an endpoint for doing so.
It also has implications on the frontend side, since it is no longer need to use de old logic (GET the appRepo, increase the resyncRequests and PUT it back).

### Benefits

This PR enables third-party integrations (i.e., Harbor webhooks) to manually trigger an update in a given AppRepository whenever needed.
Furthermore, on the frontend side, we no longer need to perform two requests, one is enough.

### Possible drawbacks

I can't see any negative implications in the short term. Perhaps we should re-think webhook and subscription mechanisms in the future and revisit this /refresh approach

### Applicable issues

  - fixes #2062

### Additional information

For example, a `POST /api/v1/clusters/default/namespaces/kubeapps/apprepositories/bitnami/refresh` will trigger a refresh in the `bitnami` appRepository in the namespace `kubeapps`.

### DIFF
--- a/dashboard/src/shared/AppRepository.ts
+++ b/dashboard/src/shared/AppRepository.ts
@@ -18,7 +18,7 @@ export class AppRepository {
 
   public static async resync(cluster: string, name: string, namespace: string) {
     const { data } = await axiosWithAuth.post(
-      url.backend.apprepositories.refresh(cluster, name, namespace),
+      url.backend.apprepositories.refresh(cluster, namespace, name),
       null,
     );
     return data;
--- a/dashboard/src/shared/AppRepository.ts
+++ b/dashboard/src/shared/AppRepository.ts
@@ -17,12 +17,9 @@ export class AppRepository {
   }
 
   public static async resync(cluster: string, name: string, namespace: string) {
-    const repo = await AppRepository.get(cluster, name, namespace);
-    repo.spec.resyncRequests = repo.spec.resyncRequests || 0;
-    repo.spec.resyncRequests++;
-    const { data } = await axiosWithAuth.put(
-      AppRepository.getSelfLink(cluster, namespace, name),
-      repo,
+    const { data } = await axiosWithAuth.post(
+      url.backend.apprepositories.refresh(cluster, name, namespace),
+      null,
     );
     return data;
   }
--- a/dashboard/src/shared/url.ts
+++ b/dashboard/src/shared/url.ts
@@ -80,7 +80,7 @@ export const backend = {
       backend.apprepositories.base(cluster, namespace),
     list: (cluster: string, namespace: string) => backend.apprepositories.base(cluster, namespace),
     validate: (cluster: string) => `${backend.apprepositories.base(cluster, "kubeapps")}/validate`,
-    delete: (cluster: string, namespace: string, name: string) =>
+    delete: (cluster: string, name: string, namespace: string) =>
       `${backend.apprepositories.base(cluster, namespace)}/${name}`,
     refresh: (cluster: string, namespace: string, name: string) =>
       `${backend.apprepositories.base(cluster, namespace)}/${name}/refresh`,
--- a/dashboard/src/shared/url.ts
+++ b/dashboard/src/shared/url.ts
@@ -80,9 +80,9 @@ export const backend = {
       backend.apprepositories.base(cluster, namespace),
     list: (cluster: string, namespace: string) => backend.apprepositories.base(cluster, namespace),
     validate: (cluster: string) => `${backend.apprepositories.base(cluster, "kubeapps")}/validate`,
-    delete: (cluster: string, name: string, namespace: string) =>
+    delete: (cluster: string, namespace: string, name: string) =>
       `${backend.apprepositories.base(cluster, namespace)}/${name}`,
-    refresh: (cluster: string, name: string, namespace: string) =>
+    refresh: (cluster: string, namespace: string, name: string) =>
       `${backend.apprepositories.base(cluster, namespace)}/${name}/refresh`,
     update: (cluster: string, namespace: string, name: string) =>
       `${backend.apprepositories.base(cluster, namespace)}/${name}`,
--- a/dashboard/src/shared/url.ts
+++ b/dashboard/src/shared/url.ts
@@ -82,6 +82,8 @@ export const backend = {
     validate: (cluster: string) => `${backend.apprepositories.base(cluster, "kubeapps")}/validate`,
     delete: (cluster: string, name: string, namespace: string) =>
       `${backend.apprepositories.base(cluster, namespace)}/${name}`,
+    refresh: (cluster: string, name: string, namespace: string) =>
+      `${backend.apprepositories.base(cluster, namespace)}/${name}/refresh`,
     update: (cluster: string, namespace: string, name: string) =>
       `${backend.apprepositories.base(cluster, namespace)}/${name}`,
   },
--- a/pkg/http-handler/http-handler.go
+++ b/pkg/http-handler/http-handler.go
@@ -160,6 +160,37 @@ func UpdateAppRepository(handler kube.AuthHandler) func(w http.ResponseWriter, r
 	}
 }
 
+// RefreshAppRepository forces a refresh in a given apprepository (by updating resyncRequests property)
+func RefreshAppRepository(handler kube.AuthHandler) func(w http.ResponseWriter, req *http.Request) {
+	return func(w http.ResponseWriter, req *http.Request) {
+		requestNamespace, requestCluster := getNamespaceAndCluster(req)
+		repoName := mux.Vars(req)["name"]
+		token := auth.ExtractToken(req.Header.Get("Authorization"))
+
+		clientset, err := handler.AsUser(token, requestCluster)
+		if err != nil {
+			returnK8sError(err, w)
+			return
+		}
+
+		appRepo, err := clientset.RefreshAppRepository(repoName, requestNamespace)
+		if err != nil {
+			returnK8sError(err, w)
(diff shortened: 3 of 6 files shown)


absoludity (commented):
> This PR enables third-party integrations (i.e., Harbor webhooks) to manually trigger an update in a given AppRepository whenever needed.

Although I like the generals changes here (simplifying refresh), I just want to check: how does this work as a third-party webhook integration, when it depends on the user's authentication? Is the harbor webhook only ever triggered via the users browser session as opposed to a background process?

That is, the new endpoint assumes the user is logged in to kubeapps (so that their cookie is being used by oauth2-proxy to include their id_token as the bearer, etc.). I assume if you hit this URL via curl or a private browser window, it'd *not* work (at least I hope it wouldn't currently :) ). So it may work if the webhook is triggered from the users' browser in harbor *and* they're already auth'd with Kubeapps, but I'm guessing otherwise it'd fail (ie. logout of kubeapps and then try from harbor).

We could enable the refresh to happen without auth creds with specific conditions (since it's idempotent and non-destructive) - interested to hear your thoughts, or where I've misunderstood :)

pkg/http-handler/http-handler.go
```diff
@@ -274,12 +305,14 @@ func SetupDefaultRoutes(r *mux.Router, clustersConfig kube.ClustersConfig) error
...
 	r.Methods("GET").Path("/namespaces/{namespace}/operator/{name}/logo").Handler(http.HandlerFunc(GetOperatorLogo(backendHandler)))
+
 	r.Methods("GET").Path("/clusters/{cluster}/namespaces").Handler(http.HandlerFunc(GetNamespaces(backendHandler)))
 	r.Methods("GET").Path("/clusters/{cluster}/apprepositories").Handler(http.HandlerFunc(ListAppRepositories(backendHandler)))
 	r.Methods("GET").Path("/clusters/{cluster}/namespaces/{namespace}/apprepositories").Handler(http.HandlerFunc(ListAppRepositories(backendHandler)))
 	r.Methods("POST").Path("/clusters/{cluster}/namespaces/{namespace}/apprepositories").Handler(http.HandlerFunc(CreateAppRepository(backendHandler)))
 	r.Methods("POST").Path("/clusters/{cluster}/namespaces/{namespace}/apprepositories/validate").Handler(http.HandlerFunc(ValidateAppRepository(backendHandler)))
 	r.Methods("PUT").Path("/clusters/{cluster}/namespaces/{namespace}/apprepositories/{name}").Handler(http.HandlerFunc(UpdateAppRepository(backendHandler)))
+	r.Methods("POST").Path("/clusters/{cluster}/namespaces/{namespace}/apprepositories/{name}/refresh").Handler(http.HandlerFunc(RefreshAppRepository(backendHandler)))
```
andresmgot, line 315: I would change this for a `GET` request, the body is not relevant.

andresmgot (commented):
Thanks @antgamdia! I have a few comments.

andresmgot:
> Although I like the generals changes here (simplifying refresh), I just want to check: how does this work as a third-party webhook integration, when it depends on the user's authentication? Is the harbor webhook only ever triggered via the users browser session as opposed to a background process?
> 
> That is, the new endpoint assumes the user is logged in to kubeapps (so that their cookie is being used by oauth2-proxy to include their id_token as the bearer, etc.). I assume if you hit this URL via curl or a private browser window, it'd _not_ work (at least I hope it wouldn't currently :) ). So it may work if the webhook is triggered from the users' browser in harbor _and_ they're already auth'd with Kubeapps, but I'm guessing otherwise it'd fail (ie. logout of kubeapps and then try from harbor).
> 
> We could enable the refresh to happen without auth creds with specific conditions (since it's idempotent and non-destructive) - interested to hear your thoughts, or where I've misunderstood :)

Yes, with these changes, only authenticated users (with permissions to modify AppRepos) would be able to trigger the refresh. My suggestion is to write some documentation about how to create a serviceaccount with those permissions and use its token as a Bearer token when configuring, for example, the Harbor webhook (as explained in here: https://github.com/kubeapps/kubeapps/issues/2062#issuecomment-721229852).

dashboard/src/shared/url.ts
```diff
@@ -82,6 +82,8 @@ export const backend = {
     validate: (cluster: string) => `${backend.apprepositories.base(cluster, "kubeapps")}/validate`,
     delete: (cluster: string, name: string, namespace: string) =>
       `${backend.apprepositories.base(cluster, namespace)}/${name}`,
+    refresh: (cluster: string, name: string, namespace: string) =>
```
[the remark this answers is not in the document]
antgamdia: +1 the ordering is pretty odd. I considered changing to _cluster, namespace, name,_ but I opted for preserving the existing order for the sake of uniformity. 
I'll send another PR to solve it when this one becomes merged :)

dashboard/src/shared/AppRepository.ts
```diff
@@ -17,12 +17,9 @@ export class AppRepository {
   }
 
   public static async resync(cluster: string, name: string, namespace: string) {
-    const repo = await AppRepository.get(cluster, name, namespace);
-    repo.spec.resyncRequests = repo.spec.resyncRequests || 0;
-    repo.spec.resyncRequests++;
-    const { data } = await axiosWithAuth.put(
-      AppRepository.getSelfLink(cluster, namespace, name),
-      repo,
+    const { data } = await axiosWithAuth.post(
```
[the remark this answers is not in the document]
antgamdia: Yes, technically it is perfectly feasible. However, webhooks (almost) always use `POST` method for sending requests mainly for: I) semantically, "refresh" is an action and it is something you want to get executed, not simply retrieve a value; II) it will trigger a subsequent operation that may not be idempotent.

At the moment, this POST method will ignore any payload sent in the body, that's the reason for sending null at the frontend.

pkg/http-handler/http-handler.go
antgamdia: See my comments above

antgamdia:
> My suggestion is to write some documentation about how to create a serviceaccount with those permissions and use its token as a Bearer token when configuring,

+1 to this idea. Harbor indeed only supports authentication in webhooks by manually setting the HTTP Authorization header when configuring it. Therefore, I'll go for the idea of the serviceaccount for hooks.
I'll investigate and document it in this PR.

dashboard/src/shared/AppRepository.ts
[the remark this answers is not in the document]
andresmgot: :thinking: okay, makes sense

dashboard/src/shared/url.ts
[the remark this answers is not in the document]
andresmgot: both options work for me

andresmgot:
> I'll investigate and document it in this PR.

You can do a separate PR for it if you want. Typically split PRs are better.

absoludity (approved):
> > My suggestion is to write some documentation about how to create a serviceaccount with those permissions and use its token as a Bearer token when configuring,
> 
> +1 to this idea. Harbor indeed only supports authentication in webhooks by manually setting the HTTP Authorization header when configuring it. Therefore, I'll go for the idea of the serviceaccount for hooks.
> I'll investigate and document it in this PR.

Great - sounds good. Small concern that the permission required by an external webhook would be to edit `AppRepository` resources (not just the count field, so for example, the cred could be mis-used to change the source of the app repo), but not a big issue right now, and something that will be solved by open policy agent in the not-too-distant future (well, already, it's just not common to use it extensively yet).

pkg/kube/kube_handler.go
```diff
@@ -446,18 +446,13 @@ func (a *userHandler) UpdateAppRepository(appRepoBody io.ReadCloser, requestName
...
-
-	// retrive the repo object with name=repoName
+	// Retrieve the repo object with name=repoName
 	appRepo, err := a.clientset.KubeappsV1alpha1().AppRepositories(requestNamespace).Get(context.TODO(), repoName, metav1.GetOptions{})
 	if err != nil {
 		return nil, err
 	}
 
-	// an updated is forced if the ResyncRequests property changes,
+	// An updated is forced if the ResyncRequests property changes,
```
andresmgot, line 455: `An update`

dashboard/src/shared/url.ts
```diff
@@ -80,8 +80,10 @@ export const backend = {
       backend.apprepositories.base(cluster, namespace),
     list: (cluster: string, namespace: string) => backend.apprepositories.base(cluster, namespace),
     validate: (cluster: string) => `${backend.apprepositories.base(cluster, "kubeapps")}/validate`,
-    delete: (cluster: string, name: string, namespace: string) =>
+    delete: (cluster: string, namespace: string, name: string) =>
```
andresmgot, line 83: you have changed the order of the parameters here in this `delete` but not in the places it's used, maybe you should leave this change out of this PR.

antgamdia: Totally agree, it was wrongly staged and committed while working on #2147. My fault!

antgamdia:
Comments from this PR addressed. Also edited a test step that is no longer required. Merged the latest master commits to integrate changes on #2147.

PR #2140, opened 2020-11-04
Add search (crtl+F) in ACE editors

antgamdia:
### Description of the change

ACE text editors have lots of features out-of-the-box ready to be used. For instance,  the keybind Crtl+F should trigger a search box for searching within the text. However, this feature was not enabled. This PR is to support that.

### Benefits
Crtl+F will now trigger a searchbox to appear.

image

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

See the full list of ACE modules here: https://github.com/ajaxorg/ace-builds/blob/master/ace-modules.d.ts


andresmgot (approved):
Thanks for the improvement :)

PR #2141, opened 2020-11-04
Update app manifest to match with actual favicons

antgamdia:
# Description of the change

Wrong favicon declaration was causing a mismatch and, therefore, a console warning was being thrown: `Error while trying to use the following icon from the Manifest: http://localhost:8080/favicon.ico (Resource size is not correct - typo in the Manifest?)`.

This PR simply declares the existing favicons in the manifest.json.

### Benefits

No error is thrown in the console.

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

N/A

### DIFF
--- a/dashboard/public/manifest.json
+++ b/dashboard/public/manifest.json
@@ -4,7 +4,27 @@
   "icons": [
     {
       "src": "favicon.ico",
-      "sizes": "192x192",
+      "sizes": "32x32",
+      "type": "image/x-icon"
+    },
+    {
+      "src": "favicon-16x16.png",
+      "sizes": "16x16",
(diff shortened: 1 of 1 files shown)


andresmgot (approved):
I need to understand better what's the purpose of this manifest, it sometimes fail to be downloaded but the app is not affected.

antgamdia:
> I need to understand better what's the purpose of this manifest,

AFAIK it only contains information to install apps on phones. These icons are supposed to be used when a web app is installed (or pinned) on the home screen of a mobile phone. More info: https://developer.mozilla.org/en-US/docs/Web/Manifest

PR #2145, opened 2020-11-05
Add design proposal for sso without oidc api server.

absoludity:
### Description of the change

Adds the overview and link to the design proposal for supporting SSO without requiring k8s API server configuration.


andresmgot (approved):
Regarding the two options listed in the document (using certs or id_tokens), I believe you are working in a POC for the former, right?

absoludity:
> Regarding the two options listed in the document (using certs or id_tokens), I believe you are working in a POC for the former, right?

Yep, as that's what I can already test against with pinniped (and it also doesn't require oauth setup on workload clusters either).

PR #2147, opened 2020-11-05
Change ordering in some arguments to "cluster, namespace, other args"

antgamdia:
### Description of the change

`cluster, others, namespace` and `cluster, namespace, others` orderings were mixed up across the code. This PR review each file using these methods and changes the order properly. Tests changed accordingly.

### Benefits

Now the two first arguments will be `cluster, namespace`. It was quite odd to be constantly mixing these two orderings and may lead to accidental errors due to a parameter misteaching (note normally they are both _string_ so no type alert will be thrown).

### Possible drawbacks

Despite the best effort applied in looking up each method usage, it is slightly possible to have missed some reference.

### Applicable issues

- This PR has been generated after the discussion in PR #2138.

### Additional information

N/A

### DIFF
--- a/dashboard/src/actions/catalog.test.tsx
+++ b/dashboard/src/actions/catalog.test.tsx
@@ -323,8 +323,8 @@ describe("removeBinding", () => {
     expect(store.getActions().length).toBe(0);
     expect(ServiceBinding.delete).toHaveBeenCalledWith(
       testArgs.kubeappsCluster,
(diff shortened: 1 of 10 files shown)


andresmgot (approved):
Oof, risky PR this one :joy: It's difficult to notice errors because both parameters (name, namespace) are strings but I didn't notice any (and the tests are passing) so my +1. Thanks!

PR #2149, opened 2020-11-05
Allow to bypass the oauth loading page

### LINKED ISSUE #2109: Skip the user login page when OIDC is used
**Question/Feature request**
Is there any way to skip the user login page when OIDC login is enabled? We're integrating kubeapp start link in another dashboard and the user is already logged via OIDC in the parent dashboard. Something similar to Harbor https://github.com/goharbor/harbor/pull/9311

andresmgot:
### Description of the change

As per #2109, allow users to configure Kubeapps so it skips the login/welcome page and redirect directly to the OAuth login URI.

To use it, a value is exposed in the chart: `--set authProxy.skipKubeappsLoginPage=true`.

### Applicable issues

  - fixes #2109

### DIFF
--- a/chart/kubeapps/Chart.yaml
+++ b/chart/kubeapps/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kubeapps
-version: 4.0.4
+version: 4.0.5
 appVersion: DEVEL
 description: Kubeapps is a dashboard for your Kubernetes cluster that makes it easy to deploy and manage applications in your cluster using Helm
 icon: https://raw.githubusercontent.com/kubeapps/kubeapps/master/docs/img/logo.png
--- a/dashboard/src/actions/auth.ts
+++ b/dashboard/src/actions/auth.ts
@@ -99,10 +99,9 @@ export function checkCookieAuthentication(
     const isAuthed = await Auth.isAuthenticatedWithCookie(cluster);
     if (isAuthed) {
       await dispatch(authenticate(cluster, "", true));
-      return true;
     } else {
       dispatch(setAuthenticated(false, false, ""));
-      return false;
     }
+    return isAuthed;
   };
 }
--- a/dashboard/src/actions/auth.ts
+++ b/dashboard/src/actions/auth.ts
@@ -90,7 +90,7 @@ export function expireSession(): ThunkAction<Promise<void>, IStoreState, null, A
 
 export function checkCookieAuthentication(
   cluster: string,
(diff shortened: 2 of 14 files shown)


chart/kubeapps/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kubeapps
-version: 4.0.4
+version: 4.0.5
```
absoludity, line 3: Maybe land this after @antgamdia 's PR with the ClusterRole, so they're both included in the chart update (once you've reviewed his).

dashboard/src/actions/auth.ts
```diff
@@ -99,8 +99,10 @@ export function checkCookieAuthentication(
     const isAuthed = await Auth.isAuthenticatedWithCookie(cluster);
     if (isAuthed) {
       await dispatch(authenticate(cluster, "", true));
+      return true;
     } else {
       dispatch(setAuthenticated(false, false, ""));
+      return false;
     }
```
absoludity, line 106: Perhaps just 
```
return isAuthed;
```
instead?

dashboard/src/components/LoginForm/LoginForm.test.tsx
```diff
@@ -25,31 +28,16 @@ const defaultProps = {
...
-    mount(<LoginForm {...props} checkCookieAuthentication={checkCookieAuthentication} />);
-    expect(checkCookieAuthentication).toHaveBeenCalled();
-  });
-
-  it("does not call checkCookieAuthentication when oauthLoginURI not provided", () => {
-    const checkCookieAuthentication = jest.fn();
-    mount(<LoginForm {...defaultProps} checkCookieAuthentication={checkCookieAuthentication} />);
-    expect(checkCookieAuthentication).not.toHaveBeenCalled();
-  });
-});
```
absoludity, line 51: Why are these removed?

absoludity (approved):
Nice, thanks.

chart/kubeapps/Chart.yaml
andresmgot: well, @antgamdia changes cannot be used until the new endpoint (`/refresh`) is available so the change in the chart is not going to be usable yet.

dashboard/src/components/LoginForm/LoginForm.test.tsx
andresmgot: Because this was already been covered in the tests under "oauth login form" (sorry I forgot to mention that).

PR #2150, opened 2020-11-06
Webhook documentation

antgamdia:
### Description of the change

Documented the process of configuring a webhook (exemplified using Harbor) and the subsequent creation/granting of the service account.

### Benefits

Users will now be able to configure their platforms to use the /refresh endpoint implemented in #2138.

### Possible drawbacks

Despite trying to be platform-agnostic, the only real example documented is Harbor. Perhaps other platforms should be included, don't know.

### Applicable issues

  - suggested by #2062.
  - implemented in #2138.

### Additional information

This PR is only intended for documenting the process implemented in #2138.


docs/user/syncing-apprepository-webhook.md
```diff
@@ -0,0 +1,159 @@
...
+## Creating and granting a ServiceAccount for authenticating requests
+
+Since the webhook endpoint will be used outside the user browser session (i.e., no OICD login), an ad-hoc ServiceAccount ought to be created; next, it has to be granted with sufficient permissions to `get` and `update` the `AppRepository` objects. Finally, the ServiceAccount token must be retrieved.
+
+### Creating the Kubernetes ServiceAccount, ClusterRole and RoleBinding objects
+
+First, the `ServiceAccount` and the `ClusterRole` objects are created. In this example, the ServiceAccount *apprepositories-webhook* is being created.
+
+```bash
+cat <<EOF | kubectl apply -f -
```
absoludity, line 82: Is there a reason that you're not explicitly putting the service account in a specific namespace of the AppRepository? (I mean, I know it can go in any namespace, but I thought it'd be clearer to keep them together in the one).

docs/user/syncing-apprepository-webhook.md
```diff
@@ -0,0 +1,159 @@
...
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: apprepositories-webhook
+---
+kind: ClusterRole
```
absoludity, line 89: A couple of related thoughts: I assume you're using a ClusterRole so it can be re-used in other namespaces, but the person creating the webhook might not have cluster-admin. I wonder if the Kubeapps chart should be updated with this PR to include the creation of this cluster role when installing Kubeapps, so that any Kubeapps user can setup a webhook for an app repository in their namespace? See what you think.

docs/user/syncing-apprepository-webhook.md
```diff
@@ -0,0 +1,159 @@
...
+      - kubeapps.com
+    resources:
+      - apprepositories
+    verbs:
+      - get
+      - update
+EOF
+```
+
+Next, the `ClusterRole` *apprepositories-webhook* (with `get` and `update` permissions over `apprepositories`) has to be associated to a certain (or multiple) namespace(s) by means of a `RoleBinding`:
```
absoludity, line 104: Since you give an example of how to bind it in multiple namespaces below, this sentence might be simpler as "...to be bound to the relevant namespace(s) by means of a `RoleBinding`:"

docs/user/syncing-apprepository-webhook.md
```diff
@@ -0,0 +1,159 @@
...
+
+> To bind it in multiple namespaces, create as RoleBinding (changing `namespace: my-namespace` to match your needs) as you need.
```
absoludity, line 106: s/as you need/for each namespace/

docs/user/syncing-apprepository-webhook.md
```diff
@@ -0,0 +1,159 @@
...
+The `<TOEKN>`  for the ServiceAccount *apprepositories-webhook* is retrieved:
+
+```bash
+kubectl get secret $(kubectl get serviceaccount apprepositories-webhook -o jsonpath='{range .secrets[*]}{.name}{"\n"}{end}' | grep apprepositories-webhook) -o jsonpath='{.data.token}' -o go-template='{{.data.token | base64decode}}' && echo
+```
+This value will be the Bearer token to be passed in the `Authentication` HTTP header.
+
+## Configuring a webhook in Harbor
+
+A high-level description of the main steps is presented herein; please refer to the [official Harbor documentation](https://goharbor.io/docs/1.10/working-with-projects/project-configuration/configure-webhooks/) for further information.
```
absoludity, line 138: s/herein/below/ (simple :) )

docs/user/syncing-apprepository-webhook.md
```diff
@@ -0,0 +1,159 @@
...
+
+1. Log in to the Harbor interface with an account that has at least project administrator privileges.
+2. Go to  *Projects*, select a project, and select  *Webhooks*.
+3. Enter this URL: `https://<KUBEAPPS_URL>/api/v1/clusters/<CLUSTER_NAME>/namespaces/<NAMESPACE_NAME>/apprepositories/<APPREPO_NAME>/refresh`. Modify `<KUBEAPPS_URL>` , `<CLUSTER_NAME>`, `<NAMESPACE_NAME>`, `<APPREPO_NAME>` to match your needs. 
+	> For instance: `https://<KUBEAPPS_URL>/api/v1/clusters/default/namespaces/kubeapps/apprepositories/bitnami/refresh` will update the App Repository `bitnami` in the namespace `kubeapps` of the cluster `default`.
+	
+4.  As authentication header enter `Bearer <TOKEN>`
+	> Note `<TOKEN>`is the value retrieved [before](#retrieving-the-serviceaccount-token).
+	
+5. Select as many events as you need to trigger notifications.
```
absoludity, line 148: Perhaps you could append to step 5: " The image below shows the common events that you would want for triggering a resync." and then updating the image to just select the relevant ones? (I could be missing something, but I think it'd just be "Chart Uploaded" and "Chart Deleted"? (possibly also Artifact pushed / deleted for future OCI stuff?)

absoludity (approved):
Excellent @antgamdia . I'm +1'ing with a bunch of text suggestions inline that are all optional - see what you think. If you do decide to add the ClusterRole to the chart itself, perhaps don't land, but otherwise fire away :)

docs/user/manifests/kubeapps-local-dev-webhook-rbac.yaml
```diff
@@ -0,0 +1,35 @@
...
+  name: apprepositories-webhook
+  namespace: kubeapps
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: apprepositories-webhook
+subjects:
+- kind: ServiceAccount
+  name: apprepositories-webhook
+  namespace: default
```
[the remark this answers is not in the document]
antgamdia: Yep, I just supposed the whole files should be there regardless of any existing mention in the docs. I'll delete this file since it is actually unnecessary. Thanks for letting me know!

docs/user/manifests/kubeapps-local-dev-webhook-rbac.yaml
```diff
@@ -0,0 +1,35 @@
...
+      - apprepositories
+    verbs:
+      - get
+      - update
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: apprepositories-webhook
+  namespace: kubeapps
```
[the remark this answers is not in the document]
antgamdia: I actually forgot to mention it in the docs. Thank you for bringing it to my attention!

But, yes, I was thinking about what's might be the best approach for the example. Whereas using the kubeapps namespace will work seamlessly for users who just deploy the default kubeapps chart with the Bitnami repo, it has, as you mention, security concerns. Given that the general docs also include the cluster-admin for the sake of simplicity, I opted to exemplify using the non-namespaced repo.

Nevertheless, since the target user using webhooks may be the one who is installing a new custom repo (i.e., Harbor), I believe we should choose the namespaced repo approach in the docs. I'll modify the docs accordingly.

docs/user/syncing-apprepository-webhook.md
```diff
@@ -0,0 +1,159 @@
...
+```bash
+kubectl get secret $(kubectl get serviceaccount apprepositories-webhook -o jsonpath='{range .secrets[*]}{.name}{"\n"}{end}' | grep apprepositories-webhook) -o jsonpath='{.data.token}' -o go-template='{{.data.token | base64decode}}' && echo
+```
+
+- In your platform:
+	- Webhook URL:    `https://<KUBEAPPS_URL>/api/v1/clusters/<CLUSTER_NAME>/namespaces/<NAMESPACE_NAME>/apprepositories/<APPREPO_NAME>/refresh`
+	- Method: `POST`
+	- Authentication header: `Bearer <TOKEN>`
+
+> For instance: `https://<KUBEAPPS_URL>/api/v1/clusters/default/namespaces/kubeapps/apprepositories/bitnami/refresh` will update the App Repository `bitnami` in the namespace `kubeapps` of the cluster `default`.
```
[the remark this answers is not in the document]
antgamdia: Totally agree! I just wanted to summarize the information, but... yep, it is finally quite long and definitively not worthwhile!

docs/user/syncing-apprepository-webhook.md
```diff
@@ -0,0 +1,159 @@
...
+	- Method: `POST`
+	- Authentication header: `Bearer <TOKEN>`
+
+> For instance: `https://<KUBEAPPS_URL>/api/v1/clusters/default/namespaces/kubeapps/apprepositories/bitnami/refresh` will update the App Repository `bitnami` in the namespace `kubeapps` of the cluster `default`. 
+
+
+
+## Creating and granting a ServiceAccount for authenticating requests
+
+Since the webhook endpoint will be used outside the user browser session (i.e., no OICD login), an ad-hoc ServiceAccount ought to be created; next, it has to be granted with sufficient permissions to `get` and `update` the `AppRepository` objects. Finally, the ServiceAccount token must be retrieved.
```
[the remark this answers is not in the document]
antgamdia: +1 it's an occupational hazard due to my past years in the Academia :-P
Thanks for pointing this out!

docs/user/syncing-apprepository-webhook.md
antgamdia: I just intended to express that the actual ServiceAccount is independent of whichever chosen namespace for the RoleBinding and one is able to grant this sa to any namespace.
I'm not sure here whether it'd be clearer to use the same namespace; don't know :S

docs/user/syncing-apprepository-webhook.md
antgamdia: Mmmm, it really makes sense! I'm going to test this idea out and then I'll do another PR for the chart thing so we can continue the discussion there.
Thank you for the valuable comments!

antgamdia: Follow up PR here: #2151

absoludity (approved):
Great, thanks for the updates!

docs/user/syncing-apprepository-webhook.md
```diff
@@ -0,0 +1,115 @@
...
+  name: apprepositories-webhook
+  namespace: default
+EOF
+```
+> Note: remember that the namespace `kubeapps` is special as every App Repository added here is considered as a **global** one and it is shared across every namespace. Please take it into consideration before granting the ServiceAccount with permissions on the `kubeapps` namespace.
+
+
+### Retrieving the ServiceAccount token
+
+The `<TOEKN>`  for the ServiceAccount *apprepositories-webhook* is retrieved:
```
andresmgot, line 85: typo: `<TOKEN>` :)

PR #2151, opened 2020-11-06
Add apprepositories-webhook ClusterRole in kupeapps chart

antgamdia:
### Description of the change

After the [discussion](https://github.com/kubeapps/kubeapps/pull/2150#discussion_r518708591) in PR #2150. A ClusterRole can be added by default during the Kubeapps installation via Helm.

### Benefits

We henceforth will have an `apprepositories-webhook` ClusterRole and users just have to create a ServiceAccount and a RoleBinding to start using the webhook endpoint.

### Possible drawbacks

I'm not aware right now of the consequences of changing the chart here. Perhaps there are some requirements. For now, I'm sending this PR as a draft.

### Applicable issues

  - Related PR #2150

### Additional information

N/A


absoludity (approved):

chart/kubeapps/templates/apprepository-rbac.yaml
```diff
@@ -142,4 +142,19 @@ rules:
       - secrets
     verbs:
       - '*'
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: "kubeapps:{{ .Release.Namespace }}:apprepositories-webhook"
```
andresmgot, line 149: I'd call this `kubeapps:{{ .Release.Namespace }}:apprepositories-refresh` to follow the same naming than the others.

andresmgot (approved):
Just one naming suggestion.

chart/kubeapps/templates/apprepository-rbac.yaml
antgamdia: Good idea. Since it involves changing the docs, I will also send a PR for that purpose.

antgamdia:
I've just noticed that this PR was closed but not merged (I initially configured it as a draft, so it couldn't be merged). 
Merging now.

chart/kubeapps/templates/apprepository-rbac.yaml
```diff
@@ -142,4 +142,19 @@ rules:
       - secrets
     verbs:
       - '*'
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: "kubeapps:{{ .Release.Namespace }}:apprepositories-refresh"
+  labels:{{ include "kubeapps.extraAppLabels" . | nindent 4 }}
+  app: {{ template "kubeapps.apprepository.fullname" . }}
```
absoludity, line 151: We missed the indentation here - `app:` is a label, not a field on metadata - which caused CI to fail, but we currently have false positives so it makes a genuine failure easier to ignore. We should fix that :) Anyway, I've fixed it as part of #2162 .

antgamdia: My bad! I didn't notice it! Thanks for pointing out and fix it!!!

PR #2155, opened 2020-11-10
Convert SVG images to PNG

andresmgot:
### Description of the change

Transform SVG images to PNG as we do with all the logos.

### Possible drawbacks

It's possible that some images don't reproduce the right colors (even though I checked the ones from the `stable` repository and they seem fine). Also, the resolution may be lower. The file size doesn't get increased since after resizing them to (max) 160x160px the size is more or less the same.

### Additional information

Example response (from a previously svg):

Screenshot from 2020-11-10 15-44-18

Screenshot from 2020-11-10 15-32-18


absoludity (approved):
Great, thanks Andres.

PR #2156, opened 2020-11-10
2144 pull secrets in jobs fix

### LINKED ISSUE #2144: Offline Install: AppRepo sync jobs do not receive required imagePullSecrets
## Description:

In #1751 support was added to allow offline installation of Kubeapps. If that offline installation is from an internal mirror, the `global.imageRegistry=REPO_URL` option can be set [as per the docs](https://github.com/kubeapps/kubeapps/blob/master/docs/user/offline-installation.md#4-install-kubeapps), which works. If the internal mirror also requires pull secrets, the additional `global.imagePullSecrets=SECRET_NAME` option needs to be set, which also works for the initial installation, but this value is *not* passed through to the app repository controller and so is not used for the cronjobs/syncjobs that try to pull the related kubeapps image.

So the kubeapps installation works, but sync jobs go into image pull backoff:

kubeapps

This was originally reported on [#kubeapps in kubernetes slack](https://kubernetes.slack.com/archives/C9D3TSUG4/p1604511793033400).

1. Setup an internal docker registry with kubeapps images mirrored, [as per the docs](https://github.com/kubeapps/kubeapps/blob/master/docs/user/offline-installation.md#4-install-kubeapps).
2. Ensure the internal docker registry is private, requiring docker credentials
3. Create a secret in your `kubeapps` namespace with the docker credentials
4. Install kubeapps setting both the `global.imageRegistry` and `global.imagePullSecrets` options

Kubeapps installs fine but sync jobs fail with ImagePullBackoff

Kubeapps installs fine and sync jobs run without issue.

antgamdia:
### Description of the change

The problem described in #2144 and replicated following the steps in https://github.com/kubeapps/kubeapps/issues/2144#issuecomment-724805837 is pretty straightforward to be solved. `ImagePullSecrets` value is not being passed through the created jobs.

In order to make apprepository-controller accept this value, it should be passed as new arg param in the go binary.
For this purpose, the param `repo-sync-image-pullsecrets` has been created. It accepts a list of comma separated values with then names of the secrets used as ImagePullSecrets.

For this param to work, it is necessary that the Helm chart properly handles the values.global.imagePullSecrets. This PR also introduces a change in the chart for generating a comma-separated list based on the list of ImagePullSecrets.

### Benefits

We henceforth will be able to perform an installation of Kubeapps in an air-gapped environment since every component is able to be retrieved locally if necessary.

An offline installation (supposing the registry is already populated) will be as simple as executing:
`helm install kubeapps ./<MY_LOCAL_PREDOWNLOADED_KUBEAPPS_FOLDER> --set global.imageRegistry=MY_REGISTRY_URL> --set global.imagePullSecrets[0]=<MY_IMAGEPULLSECRET>`

### Possible drawbacks

I'm not aware if setting the` global.imagePullSecrets` value in Helm is the only valid option. Maybe this PR will not work for users setting manually ImagePullSecrets directly in the YAML (since the helper function for generating the comma-separated argument is precisely using global.imagePullSecrets).

### Applicable issues

  - fixes #2144

### Additional information

N/A

### DIFF
--- a/chart/kubeapps/templates/_helpers.tpl
+++ b/chart/kubeapps/templates/_helpers.tpl
@@ -258,11 +258,3 @@ Returns a JSON list of cluster names only (without sensitive tokens etc.)
     {{- end }}
     {{- $sanitizedClusters | toJson }}
 {{- end -}}
-
-{{/*
-Returns a comma separated string from a list
-*/}}
-{{- define "kubeapps.joinListWithComma" -}}
-{{- $local := dict "first" true -}}
-{{- range $k, $v := . -}}{{- if not $local.first -}},{{- end -}}{{- $v -}}{{- $_ := set $local "first" false -}}{{- end -}}
-{{- end -}}
--- a/chart/kubeapps/templates/_helpers.tpl
+++ b/chart/kubeapps/templates/_helpers.tpl
@@ -258,3 +258,11 @@ Returns a JSON list of cluster names only (without sensitive tokens etc.)
     {{- end }}
     {{- $sanitizedClusters | toJson }}
 {{- end -}}
+
+{{/*
+Returns a comma separated string from a list
+*/}}
+{{- define "kubeapps.joinListWithComma" -}}
+{{- $local := dict "first" true -}}
+{{- range $k, $v := . -}}{{- if not $local.first -}},{{- end -}}{{- $v -}}{{- $_ := set $local "first" false -}}{{- end -}}
+{{- end -}}
--- a/cmd/apprepository-controller/controller.go
+++ b/cmd/apprepository-controller/controller.go
@@ -273,7 +273,7 @@ func (c *Controller) syncHandler(key string) error {
 		if errors.IsNotFound(err) {
 			log.Infof("AppRepository '%s' no longer exists so performing cleanup of charts from the DB", key)
 			// Trigger a Job to perfrom the cleanup of the charts in the DB corresponding to deleted AppRepository
-			_, err = c.kubeclientset.BatchV1().Jobs(namespace).Create(context.TODO(), newCleanupJob(name, namespace, c.conf), metav1.CreateOptions{})
+			_, err = c.kubeclientset.BatchV1().Jobs(namespace).Create(context.TODO(), newCleanupJob(name, c.conf), metav1.CreateOptions{})
 			return nil
 		}
 		return fmt.Errorf("Error fetching object with key %s from store: %v", key, err)
@@ -281,27 +281,27 @@ func (c *Controller) syncHandler(key string) error {
 
 	// Get the cronjob with the same name as AppRepository
 	cronjobName := cronJobName(apprepo)
-	cronjob, err := c.cronjobsLister.CronJobs(c.conf.Namespace).Get(cronjobName)
+	cronjob, err := c.cronjobsLister.CronJobs(c.conf.KubeappsNamespace).Get(cronjobName)
 	// If the resource doesn't exist, we'll create it
 	if errors.IsNotFound(err) {
 		log.Infof("Creating CronJob %q for AppRepository %q", cronjobName, apprepo.GetName())
-		cronjob, err = c.kubeclientset.BatchV1beta1().CronJobs(c.conf.Namespace).Create(context.TODO(), newCronJob(apprepo, c.conf.Namespace, c.conf), metav1.CreateOptions{})
+		cronjob, err = c.kubeclientset.BatchV1beta1().CronJobs(c.conf.KubeappsNamespace).Create(context.TODO(), newCronJob(apprepo, c.conf), metav1.CreateOptions{})
 		if err != nil {
 			return err
 		}
 
 		// Trigger a manual Job for the initial sync
-		_, err = c.kubeclientset.BatchV1().Jobs(c.conf.Namespace).Create(context.TODO(), newSyncJob(apprepo, c.conf.Namespace, c.conf), metav1.CreateOptions{})
+		_, err = c.kubeclientset.BatchV1().Jobs(c.conf.KubeappsNamespace).Create(context.TODO(), newSyncJob(apprepo, c.conf), metav1.CreateOptions{})
 	} else if err == nil {
 		// If the resource already exists, we'll update it
-		log.Infof("Updating CronJob %q in namespace %q for AppRepository %q in namespace %q", cronjobName, c.conf.Namespace, apprepo.GetName(), apprepo.GetNamespace())
-		cronjob, err = c.kubeclientset.BatchV1beta1().CronJobs(c.conf.Namespace).Update(context.TODO(), newCronJob(apprepo, c.conf.Namespace, c.conf), metav1.UpdateOptions{})
+		log.Infof("Updating CronJob %q in namespace %q for AppRepository %q in namespace %q", cronjobName, c.conf.KubeappsNamespace, apprepo.GetName(), apprepo.GetNamespace())
+		cronjob, err = c.kubeclientset.BatchV1beta1().CronJobs(c.conf.KubeappsNamespace).Update(context.TODO(), newCronJob(apprepo, c.conf), metav1.UpdateOptions{})
 		if err != nil {
 			return err
 		}
 
 		// The AppRepository has changed, launch a manual Job
-		_, err = c.kubeclientset.BatchV1().Jobs(c.conf.Namespace).Create(context.TODO(), newSyncJob(apprepo, c.conf.Namespace, c.conf), metav1.CreateOptions{})
+		_, err = c.kubeclientset.BatchV1().Jobs(c.conf.KubeappsNamespace).Create(context.TODO(), newSyncJob(apprepo, c.conf), metav1.CreateOptions{})
 	}
 
 	// If an error occurs during Get/Create, we'll requeue the item so we can
@@ -320,7 +320,7 @@ func (c *Controller) syncHandler(key string) error {
 		return fmt.Errorf(msg)
 	}
 
-	if apprepo.GetNamespace() == c.conf.Namespace {
+	if apprepo.GetNamespace() == c.conf.KubeappsNamespace {
 		c.recorder.Event(apprepo, corev1.EventTypeNormal, SuccessSynced, MessageResourceSynced)
 	}
 	return nil
@@ -409,11 +409,11 @@ func ownerReferencesForAppRepo(apprepo *apprepov1alpha1.AppRepository, childName
 // newCronJob creates a new CronJob for a AppRepository resource. It also sets
 // the appropriate OwnerReferences on the resource so handleObject can discover
 // the AppRepository resource that 'owns' it.
-func newCronJob(apprepo *apprepov1alpha1.AppRepository, repoNamespace string, config Config) *batchv1beta1.CronJob {
+func newCronJob(apprepo *apprepov1alpha1.AppRepository, config Config) *batchv1beta1.CronJob {
 	return &batchv1beta1.CronJob{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:            cronJobName(apprepo),
-			OwnerReferences: ownerReferencesForAppRepo(apprepo, repoNamespace),
+			OwnerReferences: ownerReferencesForAppRepo(apprepo, config.KubeappsNamespace),
 			Labels:          jobLabels(apprepo),
 		},
 		Spec: batchv1beta1.CronJobSpec{
@@ -423,7 +423,7 @@ func newCronJob(apprepo *apprepov1alpha1.AppRepository, repoNamespace string, co
 			// https://github.com/kubernetes/kubernetes/issues/54870
 			ConcurrencyPolicy: "Replace",
 			JobTemplate: batchv1beta1.JobTemplateSpec{
-				Spec: syncJobSpec(apprepo, repoNamespace, config),
+				Spec: syncJobSpec(apprepo, config),
 			},
 		},
 	}
@@ -431,18 +431,18 @@ func newCronJob(apprepo *apprepov1alpha1.AppRepository, repoNamespace string, co
 
 // newSyncJob triggers a job for the AppRepository resource. It also sets the
 // appropriate OwnerReferences on the resource
-func newSyncJob(apprepo *apprepov1alpha1.AppRepository, repoNamespace string, config Config) *batchv1.Job {
+func newSyncJob(apprepo *apprepov1alpha1.AppRepository, config Config) *batchv1.Job {
 	return &batchv1.Job{
 		ObjectMeta: metav1.ObjectMeta{
 			GenerateName:    cronJobName(apprepo) + "-",
-			OwnerReferences: ownerReferencesForAppRepo(apprepo, repoNamespace),
+			OwnerReferences: ownerReferencesForAppRepo(apprepo, config.KubeappsNamespace),
 		},
-		Spec: syncJobSpec(apprepo, repoNamespace, config),
+		Spec: syncJobSpec(apprepo, config),
 	}
 }
 
 // jobSpec returns a batchv1.JobSpec for running the chart-repo sync job
-func syncJobSpec(apprepo *apprepov1alpha1.AppRepository, repoNamespace string, config Config) batchv1.JobSpec {
+func syncJobSpec(apprepo *apprepov1alpha1.AppRepository, config Config) batchv1.JobSpec {
 	volumes := []corev1.Volume{}
 	volumeMounts := []corev1.VolumeMount{}
 	if apprepo.Spec.Auth.CustomCA != nil {
@@ -450,7 +450,7 @@ func syncJobSpec(apprepo *apprepov1alpha1.AppRepository, repoNamespace string, c
 			Name: apprepo.Spec.Auth.CustomCA.SecretKeyRef.Name,
 			VolumeSource: corev1.VolumeSource{
 				Secret: &corev1.SecretVolumeSource{
-					SecretName: secretKeyRefForRepo(apprepo.Spec.Auth.CustomCA.SecretKeyRef, apprepo, repoNamespace).Name,
+					SecretName: secretKeyRefForRepo(apprepo.Spec.Auth.CustomCA.SecretKeyRef, apprepo, config).Name,
 					Items: []corev1.KeyToPath{
 						{Key: apprepo.Spec.Auth.CustomCA.SecretKeyRef.Key, Path: "ca.crt"},
 					},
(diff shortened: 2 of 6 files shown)


antgamdia:
Ups, chart test is failing; gonna work on that asap!

cmd/apprepository-controller/controller.go
```diff
@@ -619,3 +629,22 @@ func dbFlags() []string {
 		"--database-name=" + dbName,
 	}
 }
+
+// getImagePullSecretsRefs gets the []LocalObjectReference of Secrets from the
+// comma separated list passed in the repoSyncImagePullSecrets arg
```
absoludity, line 634: I think generally it's a good idea to parse the command-line input in the main.go (or call it from there) and then *pass* the result down, rather than referencing a global var from the command. So in this case, I'd be keen that we update the `NewController` function above to receive the `imagePullSecrets []corev1.LocalObjectReference`.

It's not essential, but when working on larger projects you'll find it gets very messy if different files are all just referencing global vars - generally it should be avoided.

absoludity: Also, whether you move this helper or not, can you pls add some unit tests. In particular, showing that spaces in the string don't result in `LocalObjectReference`'s with spaces.

absoludity (commented):
Great work Antonio. Couple of thoughts below, see what you think.

andresmgot (commented):
Some more comments :) Thanks!

chart/kubeapps/templates/_helpers.tpl
```diff
@@ -258,3 +258,11 @@ Returns a JSON list of cluster names only (without sensitive tokens etc.)
...
     {{- $sanitizedClusters | toJson }}
 {{- end -}}
+
+{{/*
+Returns a comma separated string from a list
+*/}}
+{{- define "kubeapps.joinListWithComma" -}}
+{{- $local := dict "first" true -}}
+{{- range $k, $v := . -}}{{- if not $local.first -}},{{- end -}}{{- $v -}}{{- $_ := set $local "first" false -}}{{- end -}}
+{{- end -}}
```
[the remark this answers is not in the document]
antgamdia: Thanks for the tip! Added to my vscode config :)

cmd/apprepository-controller/controller.go
```diff
@@ -480,6 +481,13 @@ func syncJobSpec(apprepo *apprepov1alpha1.AppRepository, kubeappsNamespace strin
 	if len(podTemplateSpec.Spec.Containers) == 0 {
 		podTemplateSpec.Spec.Containers = []corev1.Container{{}}
 	}
+
+	// Populate ImagePullSecrets spec
+	if len(podTemplateSpec.Spec.ImagePullSecrets) == 0 {
+		log.Info("Populating ImagePullSecrets")
```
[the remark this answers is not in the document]
antgamdia: Totally agree with you. Extracted to a single debug log in the main.go for getting more information about the rest of the args.

cmd/apprepository-controller/main.go
```diff
@@ -91,6 +92,7 @@ func init() {
 	flag.StringVar(&kubeconfig, "kubeconfig", "", "Path to a kubeconfig. Only required if out-of-cluster.")
 	flag.StringVar(&masterURL, "master", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
 	flag.StringVar(&repoSyncImage, "repo-sync-image", "quay.io/helmpack/chart-repo:latest", "container repo/image to use in CronJobs")
+	flag.StringVar(&repoSyncImagePullSecrets, "repo-sync-image-pullsecrets", "", "optional reference to secrets in the same namespace to use for pulling the image used by this pod")
```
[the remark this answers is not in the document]
antgamdia: I did try this approach and I was able to have a []string with the values. However, when generating automatically these args from the Chart template, I considered that a comma-separated string would be easier to understand.  I personally see having multiple values for the same flag it's a bit confusing. 

Nevertheless, as you mention, with the array flag approach we could have both options and will let users decide. So, ok, let's implement it with an array flag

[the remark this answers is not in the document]
andresmgot: I believe the `,` approach won't work out of the box with the `flag` package. You will need to code it yourself, following the example in the URL, it would be something like:

```
func (i *arrayFlags) Set(value string) error {
    *i = append(*i, strings.Split(value, ",")...)
    return nil
}
```

But you would need also being careful not to include spaces and that kind of things. For simplicity I would just go to use as many `--repo-syn-image-pullsecrets` as needed but up to you (also note that setting these flags is something internal and automatic, not something the user needs to do on their own).

antgamdia:
Adding changes from this PR review. I've noticed that a number of global variables from the flags was directly used at the controller. While addressing the change proposed in https://github.com/kubeapps/kubeapps/pull/2156#discussion_r521155364, I decided to extract all these globar vars to a different structure and, then, pass it entirely to the controller.

In order to address the change in https://github.com/kubeapps/kubeapps/pull/2156#discussion_r521157891, the parsing logic of the flags has been extracted so that we are able to invoke it from the tests. I've created a new test file for the unit tests of the parsing process of the flags.

Finally, as a result of the discussion in https://github.com/kubeapps/kubeapps/pull/2156#discussion_r521208562, the chart template has been updated accordingly.

Thanks for your valuable comments!!

cmd/apprepository-controller/main.go
```diff
@@ -30,29 +35,94 @@ import (
...
+// besides, it contains the []LocalObjectReference for the PullSecrets
+// so they can be shared across the jobs
+type Config struct {
+	MasterURL                string
+	Kubeconfig               string
+	RepoSyncImage            string
+	RepoSyncImagePullSecrets arrayFlags
+	ImagePullSecretsRefs     []corev1.LocalObjectReference
+	RepoSyncCommand          string
+	Namespace                string
```
absoludity, line 50: I'd strongly prefer this to be `KubeappsNamespace` explicitly, as it gets confusing looking through code that refers to the namespaces of different resources etc.

cmd/apprepository-controller/main.go
```diff
@@ -87,18 +159,38 @@ func main() {
...
+// if the flag arg it's a comma-separated list, split it by "," and append
+// otherwise append it directly (dups allowed)
+func (i *arrayFlags) Set(value string) error {
+	if strings.Contains(value, ",") {
+		*i = append(*i, strings.Split(strings.TrimSpace(value), ",")...)
+
+	} else {
+		*i = append(*i, strings.TrimSpace(value))
+	}
+	return nil
```
absoludity, line 195: So this is a bit of custom code we'll need to test and maintain to support both multiple args as well as comma-separated args. When you find yourself doing this, it's often a good idea to step back and ask whether others have already solved this. In this case, we've used spf13/pflag as a drop-in replacement for go's standard flag lib. It actually provides this functionality out of the box. See https://pkg.go.dev/github.com/spf13/pflag#StringSlice .

cmd/apprepository-controller/main.go
```diff
@@ -30,29 +35,94 @@ import (
...
+	Crontab                  string
+	ReposPerNamespace        bool
+
+	// Args are the positional (non-flag) command-line arguments.
+	Args []string
+}
+
+// parseFlags parses the command-line arguments provided to the program.
+// Typically os.Args[0] is provided as 'programName' and os.args[1:] as 'args'.
+func parseFlags(progname string, args []string) (config *Config, output string, err error) {
```
absoludity, line 66: I'm not sure, but it looks like you've created this function to avoid having the global vars even exist? If so, that's admirable, but above and beyond what I meant with my earlier comment - and may not be worth the extra complexity in setup. I think it's clearer to just use the `flag.Parse()` and is what most go-lang apps do, I'd just (generally) avoid accessing those vars anywhere but in the main.go so that it doesn't create problems later on refactoring work (you'll often see people create their sub packages used for this reason, to ensure the boundary is clear and there's no access to the vars in the main package - but no need for that here yet).

absoludity (commented):
Hi Antonio. A few things hav been added here, which are all worthwhile discussing, but means the PR has drifted from what you'd originally intended and the review process then starts to extend. I don't mind if you want to push ahead and deal with the extra changes, but my recommendation (for practising having laser-focussed PRs in the future) would be to:

1. Realise this isn't urgent - no need to push when you've done enough for the day - get a good rest and look again in the morning with a fresh perspective :)
2. revert the change which switches to use the app repo namespace instead of kubeapps namespace (see inline comments)
3. Update to use spf13/pflag which already supports the exact parsing functionality you're after (ie. `--myflag "a,b" --myflag "c"` becomes `["a", "b", "c"]`), simplifying and removing your related custom code accordingly. It's a drop-in replacement, so shouldn't affect any other flags etc.

I think it's a great idea to encapsulate the config into a struct passed to the controller's `New()` function :) And don't block on me to land this - if you address my comments and Andres gives a +1, you're good to land.

cmd/apprepository-controller/main.go
absoludity: Ah, I re-read your comment and see that you extracted it so that this function is testable. I don't mind if you want to leave it, but you might find that when switching to `pflag` (which supports the parsing you want) that there's actually nothing to test (you'll just be creating the `Config` struct with the parsed values). See what you think.

cmd/apprepository-controller/controller.go
```diff
@@ -101,7 +101,7 @@ func NewController(
 	apprepoclientset clientset.Interface,
 	kubeInformerFactory kubeinformers.SharedInformerFactory,
 	apprepoInformerFactory informers.SharedInformerFactory,
-	kubeappsNamespace string) *Controller {
+	conf *Config) *Controller {
```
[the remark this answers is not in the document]
andresmgot: > Is there a reason to pass by ref here, rather than by value? (other than optimization which I don't think is significant here).

I am fine using refs (pointers) as default when using structs

cmd/apprepository-controller/controller.go
```diff
@@ -273,35 +273,35 @@ func (c *Controller) syncHandler(key string) error {
...
+			_, err = c.kubeclientset.BatchV1().Jobs(namespace).Create(context.TODO(), newCleanupJob(name, namespace, c.conf), metav1.CreateOptions{})
 			return nil
 		}
 		return fmt.Errorf("Error fetching object with key %s from store: %v", key, err)
 	}
 
 	// Get the cronjob with the same name as AppRepository
 	cronjobName := cronJobName(apprepo)
-	cronjob, err := c.cronjobsLister.CronJobs(c.kubeappsNamespace).Get(cronjobName)
+	cronjob, err := c.cronjobsLister.CronJobs(c.conf.Namespace).Get(cronjobName)
```
andresmgot, line 284: it's already in other comments, but please keep the "kubeapps namespace"

cmd/apprepository-controller/controller.go
```diff
@@ -273,35 +273,35 @@ func (c *Controller) syncHandler(key string) error {
...
 	// If the resource doesn't exist, we'll create it
 	if errors.IsNotFound(err) {
 		log.Infof("Creating CronJob %q for AppRepository %q", cronjobName, apprepo.GetName())
-		cronjob, err = c.kubeclientset.BatchV1beta1().CronJobs(c.kubeappsNamespace).Create(context.TODO(), newCronJob(apprepo, c.kubeappsNamespace), metav1.CreateOptions{})
+		cronjob, err = c.kubeclientset.BatchV1beta1().CronJobs(c.conf.Namespace).Create(context.TODO(), newCronJob(apprepo, c.conf.Namespace, c.conf), metav1.CreateOptions{})
```
andresmgot, line 288: `newCronJob(apprepo, c.conf.Namespace, c.conf)` --> If you are already passing the conf, why pass the namespace as a different parameter?

cmd/apprepository-controller/controller_test.go
```diff
@@ -13,20 +13,44 @@ import (
 )
 
 func Test_newCronJob(t *testing.T) {
-	dbURL = "postgresql.kubeapps"
-	dbName = "assets"
-	dbUser = "admin"
-	dbSecretName = "postgresql"
-	const kubeappsNamespace = "kubeapps"
+	var kubeconfig = ""
```
andresmgot, line 16: just for styling, it's more common to use `kubeconfig := ""` rather than `var kubeconfig = ""`. The `var` notation is typically only used when you don't know the initial value (or if you don't want to set it) e.g. `var kubeConfig *kube.Config`

cmd/apprepository-controller/main.go
```diff
@@ -30,29 +35,94 @@ import (
...
+	conf, output, err := parseFlags(os.Args[0], os.Args[1:])
+
+	if err != nil {
+		log.Fatal("got error:", err)
+		log.Fatal("output:\n", output)
+		os.Exit(1)
+	}
+
+	log.WithFields(log.Fields{
+		"kubeconfig":                  conf.Kubeconfig,
```
andresmgot, line 106: make sure this doesn't logs secrets (like the serviceaccount token in the kubeconfig). I am not sure if this is just the path to the kubeconfig file or the actual content.

cmd/apprepository-controller/main.go
```diff
@@ -87,18 +159,38 @@ func main() {
...
+	if len(imagePullSecretsRefsParam) == 0 {
+		return nil
+	}
+
+	// getting and appending a []LocalObjectReference for each ImagePullSecret passed
+	for _, imagePullSecretName := range imagePullSecretsRefsParam {
+		imagePullSecretsRefs = append(imagePullSecretsRefs, []corev1.LocalObjectReference{
+			{Name: imagePullSecretName},
+		}...)
+	}
```
andresmgot, line 177: you can simplify this:

```
imagePullSecretsRefs = append(imagePullSecretsRefs, corev1.LocalObjectReference{Name: imagePullSecretName})
```

cmd/apprepository-controller/main.go
andresmgot: yep, since you are changing this much how the flags are set, we can use a more complete library like the one Michael suggests.

cmd/apprepository-controller/main_test.go
```diff
@@ -0,0 +1,153 @@
+package main
```
andresmgot, line 1: As far as I can tell, this is just testing that the flags behave as expected (which is not needed if we use a different package).

andresmgot (commented):
Thanks @antgamdia. As @absoludity suggests, there are two major things I'd change: fix the replacement of namespace -> kubeappsNamespace and use `spflag` rather than just `flag` so you don't need the extra testing.

cmd/apprepository-controller/controller.go
antgamdia: I completely get your points in having kubeappsNamespace instead of repoNamespace. In the beginning, I saw that both names were being used, so I opted to preserve the most generic one.
I was not aware of https://github.com/kubeapps/kubeapps/pull/2156#discussion_r521683721, so that's my bad, gonna change it asap :)
And, as you mention, having the c.conf makes it possible to avoid passing this parameter again and again.

cmd/apprepository-controller/main.go
antgamdia: No, AFAIK it is just the path to the kubeconfig file (`/home/agamez/.kube/config ` in my case). In fact, it is calling BuildConfigFromFlags method:
` BuildConfigFromFlags(masterUrl, kubeconfigPath string) (*restclient.Config, error)` which only needs the path, not the file itself.

The log trace I'm getting is:

`DEBUG[0000] apprepository-controller configured with these args:  crontab="*/10 * * * *" database-name=charts database-secret-key=postgresql-root-password database-secret-name=kubeapps-db database-url=localhost database-user=root kubeconfig=/home/agamez/.kube/config master= namespace=default repo-sync-cmd=/chart-repo repo-sync-image="demo.goharbor.io/docker/bitnami/kubeapps-asset-syncer:2.0.0-scratch-r2" repo-sync-image-pullsecrets="[harbor]" repos-per-namespace=true user-agent-comment=`

It does trace the secret names and the DB info, but just because this is the required input for the command to work.
However, it is just a debug log, I don't see it's a big deal, but we can directly remove it or change it to a _trace_ level.

antgamdia:
Comments addressed. 
I've migrated to the pflags package with major no changes involved except for the ImagePullSecret fla (removed the custom parsing logic).
I've left the test for the command line arguments, since it can be useful for detecting errors if we change the flag lib in the future.
Changed back reponamespace to kubeappsNamespace.

Thanks for your comments again!

cmd/apprepository-controller/controller.go
```diff
@@ -553,38 +558,38 @@ func cronJobName(apprepo *apprepov1alpha1.AppRepository) string {
 }
 
 // deleteJobName returns a unique name for the Job to cleanup AppRepository
-func deleteJobName(reponame, reponamespace string) string {
-	return fmt.Sprintf("apprepo-%s-cleanup-%s", reponamespace, reponame)
+func deleteJobName(reponame string, config Config) string {
+	return fmt.Sprintf("apprepo-%s-cleanup-%s", config.KubeappsNamespace, reponame)
 }
```
andresmgot, line 563: be careful with this, `reponamespace` is the namespace in which the app repository is created, here (and in other places), it's expected to use the repository namespace, not the namespace in which kubeapps is installed.

cmd/apprepository-controller/controller.go
```diff
@@ -595,27 +600,27 @@ func apprepoSyncJobEnvVars(apprepo *apprepov1alpha1.AppRepository, kubeappsNames
...
 
 // apprepoCleanupJobArgs returns a list of args for the repo cleanup container
-func apprepoCleanupJobArgs(repoName, repoNamespace string) []string {
+func apprepoCleanupJobArgs(repoName string, config Config) []string {
 	return append([]string{
 		"delete",
 		repoName,
-		"--namespace=" + repoNamespace,
-	}, dbFlags()...)
+		"--namespace=" + config.KubeappsNamespace,
```
andresmgot, line 616: same here `repoNamespace !== kubeappsNamespace`

cmd/apprepository-controller/main.go
andresmgot: :+1: thanks for double checking!

cmd/apprepository-controller/main.go
```diff
@@ -87,18 +155,19 @@ func main() {
...
-	flag.StringVar(&crontab, "crontab", "*/10 * * * *", "CronTab to specify schedule")
+// getImagePullSecretsRefs gets the []string of Secrets names from the
+// StringSliceVar flag list passed in the repoSyncImagePullSecrets arg
+func getImagePullSecretsRefs(imagePullSecretsRefsArr []string) []corev1.LocalObjectReference {
+	var imagePullSecretsRefs []corev1.LocalObjectReference
+
+	// if no repoSyncImagePullSecrets arg passed, return nil, as usual
+	if len(imagePullSecretsRefsArr) == 0 {
+		return nil
+	}
```
andresmgot, line 166: This `if` is not actually needed, if `imagePullSecretsRefsArr` is empty, the `range` will be skipped and it will return `nil` anyway.

andresmgot (commented):
Thanks @antgamdia, there are still a few places in which the kubeapps namespace and the repository namespace got confused, please take a look. Apart from that looking fine :+1:

cmd/apprepository-controller/controller.go
antgamdia: Thanks for pointing this out! I've changed it a little bit for avoiding further confusion: now the `apprepo` is directly being passed through instead of manually passing the repoName and repoNamespace strings

cmd/apprepository-controller/main.go
```diff
@@ -17,42 +17,108 @@ limitations under the License.
...
+	return &conf, buf.String(), nil
+}
 
 func main() {
-	flag.Parse()
+	conf, output, err := parseFlags(os.Args[0], os.Args[1:])
+
+	if err != nil {
+		log.Fatal("got error:", err)
+		log.Fatal("output:\n", output)
```
absoludity, line 97: These look like debugging? Perhaps make it more specific to the context, something like:
```
log.Fatal("error parsing command-line arguments: ", err)
log.Fatal(output)
```

cmd/apprepository-controller/main.go
```diff
@@ -17,42 +17,108 @@ limitations under the License.
...
+		os.Exit(1)
```
absoludity, line 98: You can also use `log.Exit(1)` which will also ensure logrus has a chance to cleanup before the binary exits.

absoludity (approved):
Great, thanks @antgamdia . Few small comments but land when you're ready :)

antgamdia:
Landing finally! Thank you for your comments and continuous improvements across this PR :)

PR #2157, opened 2020-11-10
[bitnami/kubeapps] Major version. Adapt Chart to apiVersion: v2

carrodher:
**Description of the change**

- Bump `apiVersion` from `v1` to `v2` in _Chart.yaml_
- Fields in _Chart.yaml_ file has been ordered alphabetically (standardization for all Bitnami Helm Charts)
- Move dependency information from the _requirements.yaml_ to the _Chart.yaml_
- After running `helm dependency update`, a new _Chart.lock_ file is generated containing the same structure used in the previous _requirements.lock_
- Update _README.md_
- There is a new PostgreSQL subchart major version containing the same changes described here about the apiVersion and also a terminology renaming from master/slave to primary/read

**Possible drawbacks**

Helm v2 is no longer supported

### DIFF
--- a/chart/kubeapps/Chart.yaml
+++ b/chart/kubeapps/Chart.yaml
@@ -1,19 +1,23 @@
-apiVersion: v1
-name: kubeapps
-version: 4.0.5
+annotations:
+  category: Infrastructure
+apiVersion: v2
 appVersion: DEVEL
+dependencies:
+  - name: postgresql
+    repository: https://charts.bitnami.com/bitnami
+    version: '10.X.X'
 description: Kubeapps is a dashboard for your Kubernetes cluster that makes it easy to deploy and manage applications in your cluster using Helm
+home: https://kubeapps.com
 icon: https://raw.githubusercontent.com/kubeapps/kubeapps/master/docs/img/logo.png
 keywords:
   - helm
   - dashboard
   - service catalog
   - deployment
-home: https://kubeapps.com
+maintainers:
+  - email: containers@bitnami.com
+    name: Bitnami
+name: kubeapps
 sources:
   - https://github.com/kubeapps/kubeapps
-maintainers:
-  - name: Bitnami
-    email: containers@bitnami.com
-annotations:
-  category: Infrastructure
+version: 5.0.0
--- a/chart/kubeapps/README.md
+++ b/chart/kubeapps/README.md
@@ -252,12 +252,21 @@ After that you should be able to access the new version of Kubeapps. If the abov
 - Move dependency information from the *requirements.yaml* to the *Chart.yaml*
 - After running `helm dependency update`, a *Chart.lock* file is generated containing the same structure used in the previous *requirements.lock*
 - The different fields present in the *Chart.yaml* file has been ordered alphabetically in a homogeneous way for all the Bitnami Helm Charts
+- In the case of PostgreSQL subchart, apart from the same changes that are described in this section, there are also other major changes due to the master/slave nomenclature was replaced by primary/readReplica. [Here](https://github.com/bitnami/charts/pull/4385) you can find more information about the changes introduced.
 
 **Considerations when upgrading to this version**
 
 - If you want to upgrade to this version from a previous one installed with Helm 3, you shouldn't face any issues
 - If you want to upgrade to this version using Helm 2, this scenario is not supported as this version doesn't support Helm 2 anymore
 - If you installed the previous version with Helm 2 and wants to upgrade to this version with Helm 3, please refer to the [official Helm documentation](https://helm.sh/docs/topics/v2_v3_migration/#migration-use-cases) about migrating from Helm 2 to 3
+
+Apart from the previous considerations, due to the PostgreSQL major version bump, it's necessary to remove the existing statefulsets:
+
+> Note: The command below assumes that Kubeapps has been deployed in the kubeapps namespace using "kubeapps" as release name, if that's not the case, adapt the command accordingly.
+
+```console
+$ kubectl delete statefulset -n kubeapps kubeapps-postgresql-master kubeapps-postgresql-slave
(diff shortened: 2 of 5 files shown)


chart/kubeapps/README.md
```diff
@@ -242,6 +242,29 @@ kubectl apply -f <repo name>.yaml
 
 After that you should be able to access the new version of Kubeapps. If the above doesn't work for you or you run into any other issues please open an [issue](https://github.com/kubeapps/kubeapps/issues/new).
 
+### Upgrading to 5.0.0
+
+[On November 13, 2020, Helm v2 support was formally finished](https://github.com/helm/charts#status-of-the-project), this major version is the result of the required changes applied to the Helm Chart to be able to incorporate the different features added in Helm v3 and to be consistent with the Helm project itself regarding the Helm v2 EOL.
```
absoludity, line 247: Hmm, I'm guessing this comes from a standard text you're adding to lots of charts? I'd personally reword the first statement to "On November 13, 2020, Helm 2 support ended" . The important part is just "Helm 2" rather than "Helm v2", since the "v2" really refers to the chart API version which is actually the most recent one.

chart/kubeapps/README.md
```diff
@@ -242,6 +242,29 @@ kubectl apply -f <repo name>.yaml
...
+
+**What changes were introduced in this major version?**
+
+- Previous versions of this Helm Chart use `apiVersion: v1` (installable by both Helm 2 and 3), this Helm Chart was updated to `apiVersion: v2` (installable by Helm 3 only). [Here](https://helm.sh/docs/topics/charts/#the-apiversion-field) you can find more information about the `apiVersion` field.
+- Move dependency information from the *requirements.yaml* to the *Chart.yaml*
+- After running `helm dependency update`, a *Chart.lock* file is generated containing the same structure used in the previous *requirements.lock*
```
andresmgot, line 253: should the `Chart.lock` be part of the repo?

chart/kubeapps/Chart.yaml
```diff
@@ -1,19 +1,23 @@
...
-name: kubeapps
-version: 4.0.5
+annotations:
+  category: Infrastructure
+apiVersion: v2
 appVersion: DEVEL
+dependencies:
+  - name: postgresql
+    repository: https://charts.bitnami.com/bitnami
+    version: '10.X.X'
```
andresmgot, line 8: would the major upgrade of PG work out of the box or is there any additional step to run? (I guess it's only the chart v1 deprecation but just to double check).

andresmgot (approved):

chart/kubeapps/README.md
carrodher: I just replaced all the Helm vX with Helm X, it was copied from the official doc, they use both nomenclatures to refer to the Helm tool itself, for example, https://helm.sh/blog/covid-19-extending-helm-v2-bug-fixes/
> When Helm v3 was released in November 2019...

In this case, it can't be referred to apiVersion since v3 doesn't exist. 

But yes, to avoid confusion I just removed the `v`

chart/kubeapps/README.md
carrodher: Yes, it is going to be part of the repo once the new postgresql major version appears. Now, running `helm dep update` fails because the new major is not in the repo (that is why the CI is failing).

chart/kubeapps/Chart.yaml
carrodher: For now yes, we are also evaluating if we should include as part of this PostgreSQL major a rewording of the master/slave nomenclature to something like primary/readReplica; but it is something we're still evaluating. If finally we also apply those changes, I will update the upgrade notes according to that

andresmgot:
@carrodher can you merge master in your branch? To fix the issue with the CI

chart/kubeapps/README.md
```diff
@@ -242,6 +242,38 @@ kubectl apply -f <repo name>.yaml
 
 After that you should be able to access the new version of Kubeapps. If the above doesn't work for you or you run into any other issues please open an [issue](https://github.com/kubeapps/kubeapps/issues/new).
 
+### Upgrading to 5.0.0
```
andresmgot, line 245: Can you rename this title? (In the other title it says "Upgrading to 2.0" (talking about the app version, not the chart cersion). We can use this PR to also release the new version so this can be: `Upgrading to 2.0.1 (Chart 5.0.0)`.

chart/kubeapps/README.md
```diff
@@ -242,6 +242,38 @@ kubectl apply -f <repo name>.yaml
...
+
+- Previous versions of this Helm Chart use `apiVersion: v1` (installable by both Helm 2 and 3), this Helm Chart was updated to `apiVersion: v2` (installable by Helm 3 only). [Here](https://helm.sh/docs/topics/charts/#the-apiversion-field) you can find more information about the `apiVersion` field.
+- Move dependency information from the *requirements.yaml* to the *Chart.yaml*
+- After running `helm dependency update`, a *Chart.lock* file is generated containing the same structure used in the previous *requirements.lock*
+- The different fields present in the *Chart.yaml* file has been ordered alphabetically in a homogeneous way for all the Bitnami Helm Charts
+- In the case of PostgreSQL subchart, apart from the same changes that are described in this section, there are also other major changes due to the master/slave nomenclature was replaced by primary/readReplica. [Here](https://github.com/bitnami/charts/pull/4385) you can find more information about the changes introduced.
+
+**Considerations when upgrading to this version**
+
+- If you want to upgrade to this version from a previous one installed with Helm 3, you shouldn't face any issues
```
andresmgot, line 259: well, this is not true if it's necessary to delete postgresql statefulsets, no?

andresmgot:
Merging this to trigger the chart release for 2.0.1

Issue #2018, opened 2020-09-09
Refactor namespace management UX


andresmgot:
### Description:

Follow up of https://github.com/kubeapps/kubeapps/pull/2016#issuecomment-689438445

For the moment, we are everyone to list app in all namespaces and to create namespaces. If they don't have RBAC permissions, they will eventually get an error which may not be the best UX.

We need to implement the backend logic to enable/disable these features. Potentially hiding or disabling the related buttons in the UI.

andresmgot:
## Current status

After successfully logging in the system, the user is presented with the context selector in which they will be able to select the namespace and cluster to work on. By default, the namespace associated to the service account is selected. If the account is not associated to a namespace, "All Namespaces" is selected.

If the user has permissions to list namespaces, the list of all namespaces available is shown. In other case, the `kubeops` serviceaccount is used to filter out the namespaces in which the user doesn't have permissions.

When changing the context, the user is redirected to the AppList view since it's the safest thing to do.

## Proposed changes

 1. Even if a user has permissions to list namespaces, remove from the list the namespaces in which the user doesn't have permissions. This prevents from errors selecting a wrong namespace.
 2. If the user doesn't have an associated namespace, use the first one available (in alphabetical order). This will autoselect the available namespace in case a user is only allowed to work in a single namespace.
 3. Apart from filtering out the namespaces in which the user doesn't have permissions, remove the ones in "terminating" state (#543).
 4. Remove the "All Namespaces"  option from the context selector. This option only makes sense in certain views: ApplicationList and OperatorList. Using "All Namespaces" in other views like the Catalog or the Deployment Form is troublesome. Rather than showing this option in the Context Selector, we can add a checkbox in the views interested: "[ ] Show applications from all namespaces".
 5. Store the last namespace used in the browser console. That way, when reloading or logging in again, the work from the previous session can be resumed. Remove this parameter when intentionally logging out from the system (so other users will use their preferred namespace). If the stored namespace is not in the list of allowed namespace, fallback to the default behavior.
 6. Disable the "Create Namespace" button when the user doesn't have permissions to create a namespace. When hovering the disabled button, information text about the missing permissions should be shown.
[...]

PR #2158, opened 2020-11-11
Filter out forbidden and terminating namespaces

andresmgot:
### Description of the change

This PR address two minor issues when listing namespaces:

 - Filter out namespaces in which the user don't have permissions (regardless if the user has permission to list namespace). This is to avoid showing unusable namespaces for the user.
 - Filter out namespaces that are in terminating status (which are not usable either).

### Applicable issues

  - ref #2018 (points 1 and 3)

### DIFF
--- a/pkg/kube/kube_handler.go
+++ b/pkg/kube/kube_handler.go
@@ -629,9 +629,9 @@ func KubeappsSecretNameForRepo(repoName, namespace string) string {
 	return fmt.Sprintf("%s-%s", namespace, secretNameForRepo(repoName))
 }
 
-func filterAllowedNamespaces(userClientset combinedClientsetInterface, namespaces *corev1.NamespaceList) ([]corev1.Namespace, error) {
+func filterAllowedNamespaces(userClientset combinedClientsetInterface, namespaces []corev1.Namespace) ([]corev1.Namespace, error) {
 	allowedNamespaces := []corev1.Namespace{}
-	for _, namespace := range namespaces.Items {
+	for _, namespace := range namespaces {
 		res, err := userClientset.AuthorizationV1().SelfSubjectAccessReviews().Create(context.TODO(), &authorizationapi.SelfSubjectAccessReview{
 			Spec: authorizationapi.SelfSubjectAccessReviewSpec{
 				ResourceAttributes: &authorizationapi.ResourceAttributes{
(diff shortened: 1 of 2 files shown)


pkg/kube/kube_handler.go
```diff
@@ -649,36 +649,49 @@ func filterAllowedNamespaces(userClientset combinedClientsetInterface, namespace
...
+}
+
+func filterActiveNamespaces(namespaces []corev1.Namespace) []corev1.Namespace {
+	readyNamespaces := []corev1.Namespace{}
+	for _, namespace := range namespaces {
+		if namespace.Status.Phase == corev1.NamespaceActive {
+			readyNamespaces = append(readyNamespaces, namespace)
+		}
+	}
+	return readyNamespaces
```
absoludity, line 662: Functional code for the win :)

pkg/kube/kube_handler.go
```diff
@@ -649,36 +649,49 @@ func filterAllowedNamespaces(userClientset combinedClientsetInterface, namespace
 			allowedNamespaces = append(allowedNamespaces, namespace)
 		}
 	}
-	namespaces.Items = allowedNamespaces
-	return namespaces, nil
+	return allowedNamespaces, nil
```
absoludity, line 652: With this change, it might make sense to update the signature to accept `namespaces []corev1.Namespace` and just update callsites to call it with `.Items` on the list? Only reason is that then this can match the functional filter that you've created below (which is nicer, imo :) ). Whatever you think.

pkg/kube/kube_handler.go
```diff
@@ -649,36 +649,49 @@ func filterAllowedNamespaces(userClientset combinedClientsetInterface, namespace
...
 			return nil, err
 		}
 	}
 
-	return namespaces.Items, nil
+	// Filter namespaces in which the user has permissions to write (secrets) only
+	namespaceList, err := filterAllowedNamespaces(a.clientset, namespaces)
+	if err != nil {
+		return nil, err
+	}
```
absoludity, line 686: Not sure why this filter has been moved out to here: isn't it a superfluous (and relatively expensive) check if the request was done with the user's own credentials? (if you decide to move it back, just remember to update the wording of the second test which changed accordingly, I think)

absoludity (approved):

pkg/kube/kube_handler.go
andresmgot: It's not superfluous since it's covering the use case in which a user has permissions to list namespaces (which is harmless) but don't have permissions to access resources in all the namespaces (not an admin). This prevents people from using namespaces that they don't have access.

I'm thinking on the case, for example, I (andres@corp.com) have permissions to list namespaces but only have write access to my namespace `foo`. If `bar` exists, Kubeapps will autoselect the first namespace in alphabetical order (which is `bar`) so by default I will get a forbidden error if I try to do anything without changing the namespace.

PR #2161, opened 2020-11-12
Refactor default namespace resolution

andresmgot:
### Description of the change

This is the first PR, preparing a following-up to be able to store the last used namespace in the browser storage. In this PR:

 - I have decoupled the authentication flow from the namespace selection. No `auth` action will set/modify the `cluster` (or namespace) state.
 - Centralize the logic to decide a default namespace and do it only once, after fetching the namespaces information. This will be stored in the `cluster.currentNamespace` so components don't need to look at other places for this info.

### Benefits

Getting rid of some spaghetti code :) 

### Possible drawbacks

The application will now wait for the API to return the existing namespaces before selecting one (which includes a minimal delay when loading the app).

### Applicable issues

  - ref #2018


dashboard/src/shared/Namespace.ts
```diff
@@ -50,3 +51,18 @@ export default class Namespace {
 export const definedNamespaces = {
   all: "_all",
 };
+
+export function getCurrentNamespace(currenNS: string, availableNS: string[]) {
```
antgamdia, line 55: Minor typo, perhaps you mean curren**t**NS here?

andresmgot:
_I am debugging an issue in the e2e test since it seems that it's clicking in the Catalog link before the default namespace is resolved, causing a 404 :)_

dashboard/src/components/Header/Header.test.tsx
```diff
@@ -41,7 +51,33 @@ it("renders the header links and titles", () => {
...
+      ...defaultState,
+      auth: { authenticated: false },
+    }),
+    <Header />,
+  );
+  const items = wrapper.find(".nav-link");
+  expect(items).not.toExist();
+});
+
+it("should skip the links the namespace info is not available", () => {
```
absoludity, line 65: links *if the* namespace info...

dashboard/src/containers/RoutesContainer/Routes.test.tsx
```diff
@@ -58,9 +59,9 @@ it("should render a redirect to the login page", () => {
     <StaticRouter location="/" context={{}}>
       <Routes
         {...emptyRouteComponentProps}
-        cluster={"default"}
-        namespace={""}
-        authenticated={true}
+        cluster={""}
+        currentNamespace={""}
+        authenticated={false}
```
absoludity, line 64: I'm confused what this test is now testing compared to the following? It looks like they are both testing that the redirect is rendered when unauthed (and that the value of the state should be irrelevant - so only one test required?)

dashboard/src/containers/RoutesContainer/Routes.tsx
```diff
@@ -63,20 +65,17 @@ class Routes extends React.Component<IRoutesProps> {
         {Object.entries(privateRoutes).map(([route, component]) => (
           <PrivateRouteContainer key={route} exact={true} path={route} component={component} />
         ))}
-        <PrivateRouteContainer
-          key={reposPath}
-          exact={true}
-          path={reposPath}
-          component={RepoListContainer}
-        />
```
absoludity, line 71: Was this a temporary route we'd added during the switch to clustered/namespaced routes? Right, then we've updated `reposPath` but never removed the now redundant route?

dashboard/src/shared/Namespace.ts
```diff
@@ -50,3 +51,18 @@ export default class Namespace {
 export const definedNamespaces = {
   all: "_all",
 };
+
+export function getCurrentNamespace(currentNS: string, availableNS: string[]) {
+  if (currentNS) {
```
absoludity, line 56: Don't understand why we even have `currentNS` as an arg - why call this function if we already have the `currentNS`?

dashboard/src/shared/Namespace.ts
```diff
@@ -50,3 +51,18 @@ export default class Namespace {
...
+    // If a namespace has been already selected, use it
+    return currentNS;
+  }
+  // Try to get a namespace from the auth token
+  const tokenNS = Auth.defaultNamespaceFromToken(Auth.getAuthToken() || "");
+  if (tokenNS) {
+    // Return the default namespace in the token
+    return tokenNS;
```
absoludity, line 64: We should validate that this ns is in `availableNS` - we had a recent report because in Edukates the token is from a different namespace to which the user has no access.

absoludity (approved):
Great to see all the `defaultNamespace` removal!

dashboard/src/containers/RoutesContainer/Routes.test.tsx
andresmgot: yes, the only difference is that the `cluster` and `currentNamespace` info is populated, which should be ignored. Updating the test description for being more clear.

dashboard/src/containers/RoutesContainer/Routes.tsx
andresmgot: seems so, there was no reason to keep it as a separated case.

dashboard/src/shared/Namespace.ts
andresmgot: yes,  it was to avoid doing the check every time I had to call this function.

dashboard/src/shared/Namespace.ts
andresmgot: good point.

PR #2162, opened 2020-11-16
DRY up config in tests.

absoludity:
### Description of the change

Just a follow-up of https://github.com/kubeapps/kubeapps/pull/2156#discussion_r522538006  (sorry, wasn't very clear with what I meant there), removing the extra code which is repeated in many tests.

Also removes the defers which were used to reset test config, instead creating the test config (functionally) with no state shared between tests.


cmd/apprepository-controller/controller_test.go
```diff
@@ -979,12 +842,10 @@ func Test_newSyncJob(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			if tt.config.UserAgentComment != "" {
-				defaultConfig.UserAgentComment = tt.config.UserAgentComment
-				defer func() { defaultConfig.UserAgentComment = "" }()
```
absoludity, line 984: It wasn't clear to me why you were modifying `defaultConfig` at all here, given that you pass `tt.config` into your function below. But either way, although it's an interesting use of `defer` to reset the shared state for your test functions, it's a much better idea (IMO) to not have any shared state that requires resetting if you can avoid it (eg. what I did here by having each test construct the `Config` for each test).

cmd/apprepository-controller/controller_test.go
```diff
@@ -336,15 +295,11 @@ func Test_newCronJob(t *testing.T) {
...
-			}
-			if tt.config.Crontab != "" {
-				defaultConfig.Crontab = tt.config.Crontab
-				defer func() { defaultConfig.Crontab = "" }()
-			}
-			result := newCronJob(tt.apprepo, tt.config)
```
absoludity, line 347: See comment below about the defer funcs here.

cmd/apprepository-controller/controller_test.go
```diff
@@ -13,32 +13,17 @@ import (
...
-				KubeappsNamespace:        defaultConfig.KubeappsNamespace,
-				ReposPerNamespace:        defaultConfig.ReposPerNamespace,
-				DBURL:                    defaultConfig.DBURL,
-				DBUser:                   defaultConfig.DBUser,
-				DBName:                   defaultConfig.DBName,
-				DBSecretName:             defaultConfig.DBSecretName,
-				DBSecretKey:              defaultConfig.DBSecretKey,
-				UserAgentComment:         defaultConfig.UserAgentComment,
-				Crontab:                  defaultConfig.Crontab,
-			},
```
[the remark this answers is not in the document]
antgamdia: Ooook, now I see your point; totally agree with you. Since we already have the default config, there's no need to repeat it in each test. Thanks!

cmd/apprepository-controller/controller_test.go
antgamdia: I didn't know either what was the reason for having this `defer` here. I just supposed it was the proper way to reset the state and I didn't want to change more things in the previous PR.
However, with this change, it is much clearer now :)

antgamdia (approved):

andresmgot (approved):

Issue #2143, opened 2020-11-04
Kubeapps UI should gracefully handle token expiration


doddatpivotal:
![Screen Shot 2020-11-04 at 11 15 03 AM](https://user-images.githubusercontent.com/37634387/98136506-018b6000-1e8f-11eb-9699-338fd921dc5b.png)

I choose not to use a service account to access kubeapps, but rather the token generated by the vsphere-login kubectl plugin.  This sets the context of my kubeconfig with a token that does have an expiration date.

I've used that token to login to kubeapps and am able to interact based upon my personal user account permissions.

However, if I come back to kubeapps after some time and my token has expired, I see this message.  If I click logout, the message remains and I am not brought to the login page.  If I refresh my browser, finally I am brought to the login page.

It is my expectation that kubeapps should handle the expired token condition, and redirect me directly to the login page (perhaps with an info message indicating that the token is expired).

I can not use OIDC authentication because at this point, vSphere 7u1 does not offer an OIDC auth provider.

antgamdia:
I've just looked into the issue and the solution doesn't seem complicated. Perhaps if a redirect is forced after every logout the UX improves.

doddatpivotal:
That would incrementally be better.  Would love to see a better error message than "minified react error".  And potentially an auto logout, redirect to login.

andresmgot:
What we had in the past, was a modal saying "Your session has expired, please reload the page" and a button to reload (or simply redirect if the person refreshes the page. See the code at dashboard/src/components/PrivateRoute/PrivateRoute.tsx:28. What I don't know is if after the UI revamp that code is still been reached so we may need to refactor that a bit.

antgamdia:
Mmm, looking into this issue a little bit more,  the error message is not just a 401/403 problem, but, indeed, there is something causing an endless redirection loop (see the error message: https://reactjs.org/docs/error-decoder.html/?invariant=185).

antgamdia:
@doddatpivotal, would you mind to share more detailed information about how to reproduce this error? I'd need some docs or references so that I can set up the environment to investigate what's happening. 
(feel free to contact me in slack at [@]agamez)

antgamdia:
This issue is also appearing when performing these steps:

1. Login with token
2. Edit the URL deleting namespace: https://clusterURL/#/c/default/ns/namespaceName/catalog to -> https://clusterURL/#/c/default/ns//catalog
[...]

PR #2163, opened 2020-11-16
Handling 403 status codes with "anonymous"

antgamdia:
k8s api server nowadays defaults to allowing anonymous requests, so that rather than returning a 401, a 403 is returned.
Currently, we capture some 403 errors for displaying useful information (e.g., "user x doesn't have access to namespace y"). 
However, some edge cases remained unhandled. For instance, a user was able to 'bypass' UI authentication by simply entering " " (whitespace). Internally, it resulted in an anonymous authentication and the API returned 403 instead of 401.

### Description of the change

This PR is to add a new handled 403 case so that the user gets redirected to the login page instead of viewing a meaningless error message.

### Benefits

Users using tokens with expiration (i.e., #2143) will leverage from a better UX.

### Possible drawbacks

I'm not fully aware of what possibly be the behavior when using OIDC. That's the reason why the anonymous 403 is only being handled when is not OIDC login.

### Applicable issues

  - related #2143 

### Additional information

N/A


dashboard/src/shared/AxiosInstance.ts
```diff
@@ -66,6 +66,15 @@ export function addErrorHandling(axiosInstance: AxiosInstance, store: Store<ISto
...
             dispatchErrorAndLogout(message);
+            return Promise.reject(new UnauthorizedError(message));
+          }
+          // If not using auth proxy, an anonymous response is sent when
+          // a serviceaccount token expires (eg., delete secret xxx)
+          // or the token is managed externally (eg., vsphere kubectl login)
+          // In this case, force reauthentication
+          if (!Auth.usingOIDCToken() && Auth.isAnonymous(response)) {
+            dispatchErrorAndLogout(message);
+            return Promise.reject(new UnauthorizedError(message));
```
andresmgot, line 77: these two ifs can be handled as one `if is403FromAuthProxy || isAnonymous` (the other comparison is irrelevant)

Also, why are you returning early? In theory `message` can be a JSON message (not sure for this use case) and it's handled later.

andresmgot (commented):
Thanks! I have a minor comment. Also, you are missing some tests to cover the changes.

dashboard/src/shared/AxiosInstance.ts
antgamdia: Yes, I simply added another in case we should take other actions rather than logout. But, yes, if we gonna do the same in both alternatives we can merge them.

Regarding the rejection, I just thought that the logout action would be the last action and, consequently, there's no need to try to parse the JSON message (particularly, when it is an anonymous 403 it will always fail). But ok, I'll remove the early rejections just in case.

dashboard/src/shared/Auth.ts
```diff
@@ -94,6 +94,21 @@ export class Auth {
...
+  // requests, so that rather than returning a 401, a 403 is returned if
+  // RBAC does not allow the anonymous user access.
+  public static isAnonymous(response: AxiosResponse): boolean {
+    return (
+      (response?.data?.message && response.data.message.includes("system:anonymous")) ||
+      (response?.data &&
+        response.data &&
+        typeof response.data === "string" &&
+        response.data.includes("system:anonymous"))
+    );
```
andresmgot, line 109: this condition is getting a bit messy :) Here is an alternative:

```
const msg = get(response, "data.message") || get(response, "data");
return typeof msg === "string" && msg.includes("system:anonymous");
```

(being `get` lodash.get)

dashboard/src/shared/AxiosInstance.test.ts
```diff
@@ -181,6 +181,43 @@ describe("createAxiosInterceptorWithAuth", () => {
...
+          '{"metadata":{},"status":"Failure","message":"selfsubjectaccessreviews.authorization.k8s.io is forbidden: User "system:anonymous" cannot create resource "selfsubjectaccessreviews" in API group "authorization.k8s.io" at the cluster scope","reason":"Forbidden","details":{"group":"authorization.k8s.io","kind":"selfsubjectaccessreviews"},"code":403} {"namespaces":null}',
+      },
+      status: 403,
+    });
+
+    try {
+      await axios.get(testPath);
+    } catch (error) {
+      expect(store.getActions()).toEqual(expectedActions);
+    }
```
andresmgot, line 217: nitpick, if the axios method doesn't throw the error you may get a false negative:

```
    try {
      await axios.get(testPath);
    } catch (error) {
      // compare here the error, if needed
    }
```

andresmgot (approved):
Thanks! I have just a couple of minor comments.

dashboard/src/shared/Auth.ts
antgamdia: Not aware of lodash.get. Much simpler now, thanks for the tip!

PR #2164, opened 2020-11-16
Bump CRD from v1beta1 to v1 (k8s  1.16+)

antgamdia:
### Description of the change

In k8s 1.16+ a deprecation warning of `apiextensions.k8s.io/v1beta1` in favor of `apiextensions.k8s.io/v1` appeared. In 1.22+ it will be fully deprecated.

This PR bumps `apiextensions.k8s.io/v1beta1` to `apiextensions.k8s.io/v1`.

### Benefits

Users won't get a deprecation warning. Besides, k8s 1.22+ users still will be able to install Kubeapps.

### Possible drawbacks

Prior to <1.16 it won't be compatible (but these versions already reached EOL).

### Applicable issues

N/A

### Additional information

Perhaps it should be landed with other chart-related PRs to also bump up the chart version.

### DIFF
--- a/chart/kubeapps/crds/apprepository-crd.yaml
+++ b/chart/kubeapps/crds/apprepository-crd.yaml
@@ -1,4 +1,4 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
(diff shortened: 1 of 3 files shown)


andresmgot (approved):
Thanks. Also note that the CRD is created on the first installation so only people installing a new version of Kubeapps (not upgrading) will get this change.

PR #2165, opened 2020-11-16
Store the last namespace used in the browser storage

andresmgot:
### Description of the change

Another item for #2018, store the latest namespace used in the browser so it's easier to resume previous work.

When logging out of the system (or when the session expires), this information is cleaned up.

### Applicable issues

  - ref #2018

### DIFF
--- a/dashboard/src/shared/Namespace.ts
+++ b/dashboard/src/shared/Namespace.ts
@@ -1,8 +1,7 @@
-import { axiosWithAuth } from "./AxiosInstance";
-import * as url from "./url";
-
 import { Auth } from "./Auth";
+import { axiosWithAuth } from "./AxiosInstance";
 import { ForbiddenError, IResource, NotFoundError } from "./types";
+import * as url from "./url";
 
 export default class Namespace {
   public static async list(cluster: string) {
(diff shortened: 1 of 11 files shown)


dashboard/src/shared/Namespace.ts
```diff
@@ -52,11 +51,35 @@ export const definedNamespaces = {
...
+const namespaceKey = "kubeapps_namespace";
+
+function getStoredNamespace(cluster: string) {
+  const ns = localStorage.getItem(namespaceKey) || "{}";
+  return JSON.parse(ns)[cluster] || "";
+}
+
+export function setStoredNamespace(cluster: string, namespace: string) {
+  const ns: string = localStorage.getItem(namespaceKey) || "{}";
+  const nsObject = JSON.parse(ns);
```
absoludity, line 64: Shouldn't we be handling exceptions (`SyntaxError`) both here and 5 lines up? (just defaulting to `{}` if there is a syntax error, which is perhaps what you intended?)

absoludity (approved):
Given that sessions can be quite short, I wonder if it's worthwhile leaving the data there for a session expiry and only clearing it on an intentional logout?

andresmgot:
> Given that sessions can be quite short, I wonder if it's worthwhile leaving the data there for a session expiry and only clearing it on an intentional logout?

I thought about this but since a session may have expired a minute ago or a month ago, there is no way of knowing that. Since there is nothing stored in the browser (and you need to re-login), and you may choose a different account, I think it's "safer" to just clean up everything (even though trying to use a namespace by default is harmless, it's just to follow the same approach for all the info).

dashboard/src/shared/Namespace.ts
andresmgot: In theory, this info is not something the user will modify on their own so a `SyntaxError` won't happen but it's true that if someone tamper this data, it may cause a DoS.

PR #2166, opened 2020-11-17
Throw an error if there is no namespace to use

andresmgot:
### Description of the change

When a user don't have an namespace with write access, show an error (the previous behavior was to hang loading):

Screenshot from 2020-11-17 11-23-44

### Applicable issues

  - ref #2018


dashboard/src/actions/namespace.ts
```diff
@@ -48,6 +48,16 @@ export function fetchNamespaces(
     try {
       const namespaceList = await Namespace.list(cluster);
       const namespaceStrings = namespaceList.namespaces.map((n: IResource) => n.metadata.name);
+      if (namespaceStrings.length === 0) {
+        dispatch(
+          errorNamespaces(
+            cluster,
+            new Error("The current account does not have access to any namespace"),
```
absoludity, line 55: just needs a plural for "namespaces" (odd in English, [any is used with plural or uncountable nouns only](https://dictionary.cambridge.org/grammar/british-grammar/any)). Or to use the singular, "does not have access to a namepsace.", but I think "any" is better.

dashboard/src/components/ErrorBoundary/ErrorBoundary.tsx
```diff
@@ -20,11 +22,14 @@ class ErrorBoundary extends React.Component<IErrorBoundaryProps, IErrorBoundaryS
...
-      return <Alert theme="danger">An error occurred: {propsError.message}</Alert>;
-    }
-    if (stateError) {
-      return <Alert theme="danger">An error occurred: {stateError.message}</Alert>;
+    const err = propsError || stateError;
+    if (err) {
+      return (
+        <Alert theme="danger">
+          An error occurred: {err.message}.{" "}
+          <CdsInlineButton onClick={this.props.logout}>Logout</CdsInlineButton>
```
absoludity, line 30: It looks like this change would mean that all errors caught by the boundary will now have a logout button? Why would that be wanted?

dashboard/src/reducers/cluster.ts
```diff
@@ -95,10 +95,6 @@ const clusterReducer = (
         },
       };
     case getType(actions.namespace.errorNamespaces):
-      // Ignore error listing namespaces since those are expected
```
absoludity, line 98: I love seeing code like this removed :P (ignoring errors because they're expected, as it inevitably leads to debugging issues later).

absoludity (approved):
Only thing I'm unsure about is why we'd want to have the logout button for all errors displayed by the error boundary - or maybe that's not the case?

andresmgot:
> Only thing I'm unsure about is why we'd want to have the logout button for all errors displayed by the error boundary - or maybe that's not the case?

The error boundary shows to kind of errors:
 - Namespace related errors (like the screenshot), when it's not possible to fetch or create namespaces.
 - Uncaught errors. If in the application, something `throws` an error it will be shown here. These errors are unexpected and should not happen (may be caused by bugs).

For the first type, it may makes sense to logout and use a more privileged account. For the second it makes less sense, I can differentiate between the two (one is set through props, the other through the state).

absoludity (approved):
> > Only thing I'm unsure about is why we'd want to have the logout button for all errors displayed by the error boundary - or maybe that's not the case?
> 
> The error boundary shows to kind of errors:
> 
>     * Namespace related errors (like the screenshot), when it's not possible to fetch or create namespaces.
> 
>     * Uncaught errors. If in the application, something `throws` an error it will be shown here. These errors are unexpected and should not happen (may be caused by bugs).
> 
> 
> For the first type, it may makes sense to logout and use a more privileged account. For the second it makes less sense, I can differentiate between the two (one is set through props, the other through the state).
> 
> Let me know what you think.

Yeah, I think that's OK, though using the distinction of "props" vs "state" might be easy now, but we may need to revisit as soon as we have a state-based error which doesn't require logout (or vice versa)?

andresmgot:
> Yeah, I think that's OK, though using the distinction of "props" vs "state" might be easy now, but we may need to revisit as soon as we have a state-based error which doesn't require logout (or vice versa)?

Yes, that makes sense.

PR #2168, opened 2020-11-17
Revert "Bump CRD from v1beta1 to v1 (k8s  1.16+)"

antgamdia:
Reverts kubeapps/kubeapps#2164

I will open an issue for following up on this change.


andresmgot (approved):
:+1: Thanks!

FTR: We still support GKE 1.15 in our CI

PR #2171, opened 2020-11-18
Add can-i endpoint

andresmgot:
### Description of the change

Add a "can-i" endpoint, similar to `kubectl can-i` to check if the user can perform a certain action. This will be used in a follow-up PR to check if the user has permissions to create namespaces.

### Applicable issues

  - ref #2018


absoludity:
I'm a little bit wary of exposing a mirror of the api from the k8s api itself? (as in, same request payload with the SubjectAccessReview, just moved to our own API and simplifying the response to a bool). Not saying we shouldn't, I just want to understand the bigger picture of why you think this API is needed. Typically our Kubeapps API would be providing specific functionality that may combine a bunch of calls to the k8s API.

From what you've mentioned in the description, it sounds like the *UI* will use this API to check if the user can create namespaces, I guess for the existing functionality. So I wonder if an extensible alternative would be to expose a "/user-config" API endpoint for Kubeapps which returns the info Kubeapps needs, something like:

```
{
  perms: {
    create_namespaces: true,
    manage_crds: false,
    list_namespaces: ...,
  },
  ...
}
```
so we could populate state on load (and refresh when requested if needed), but using a model that Kubeapps 

Thoughts? If you want to land this for now anyway, I don't mind if @antgamdia +1's and you land, I just think the discussion might be worth having, given our existing API is very haphazard and it would be great to start a longer-term plan moving towards something better.

pkg/kube/kube_handler.go
```diff
@@ -254,6 +255,10 @@ type appRepositoryRequest struct {
 	AppRepository appRepositoryRequestDetails `json:"appRepository"`
 }
 
+type selfSubjectAccessRequest struct {
+	AppRepository appRepositoryRequestDetails `json:"appRepository"`
```
absoludity, line 259: erm, confused? Guessing this is a copy-n-paste that was left over after you found `authornizationapi.ResourceAttributes`?

andresmgot:
> I'm a little bit wary of exposing a mirror of the api from the k8s api itself? (as in, request payload with the resource attributes from the SubjectAccessReview, just moved to our own API and simplifying the response to a bool). Not saying we shouldn't, I just want to understand the bigger picture of why you think this API is needed. Typically our Kubeapps API would be providing specific functionality that may combine a bunch of calls to the k8s API.
> 
> From what you've mentioned in the description, it sounds like the _UI_ will use this API to check if the user can create namespaces, I guess for the existing functionality. So I wonder if an extensible alternative would be to expose a "/user-config" API endpoint for Kubeapps which returns the info Kubeapps needs, something like:
> 
> ```
> {
>   perms: {
>     create_namespaces: true,
>     manage_crds: false,
>     list_namespaces: ...,
>   },
>   ...
> }
> ```
> 
> so we could populate state on load (and refresh when requested if needed), but using a model that Kubeapps
> 
> Thoughts? If you want to land this for now anyway, I don't mind if @antgamdia +1's and you land, I just think the discussion might be worth having, given our existing API is very haphazard and it would be great to start a longer-term plan moving towards something better.

You can see how I am using this at #2172. So yes, basically I needed an endpoint to check if the user has permissions to do something (in this case, creating namespaces). If tomorrow, we need to check something else we can just reuse this endpoint without the need to modify the backend, that's the rationale behind this implementation (vs creating something just for namespaces).

I don't think there is a problem "mirroring" this endpoint? Can you elaborate why you are wary of this? Is it because you think this specific use case is dangerous or you are way of this becoming a "standard" and expose in the future other k8s endpoints?

antgamdia:
Interesting discussion... 

TL; DR: i) +1 to land this PR; ii)  I'd suggest taking a look on the `SelfSubjectRulesReview` just to verify if it's (or it isn't) useful; iii) before implementing high-level endpoints I'd re-think the current set of Kubeapps APIs and its current usage.

I don't think mirroring the k8s API for this scenario is a problem. It encapsulates the k8s API complexity of simply determining whether or not I have permissions for some action.
In fact, the k8s has two similar (but different) endpoints for doing so:  
  - `SelfSubjectRulesReview`  determines which actions you can perform in a namespace
  - `SelfSubjectAccessReview` checks whether or the current user can perform an action (not filling in a spec.namespace means "in all namespaces")

I'm keen to have a good, well-defined (and documented in OpenAPI!!) Kubeapps API, even if it just proxies the request from UI.
I personally think we should not expose the actual k8s API so that the UI directly talks to it. So, as for me, +1 to this PR.

At the moment, I feel the consumption of APIs is a bit chaotic. We're still serving deprecated endpoints (non-cluster-aware ones), some parts and (at least for me) it is unclear to rapidly determine who talks to who.
However, it is out of the scope of this PR. It is more related to a refactor and documenting effort.

Regarding the long-term solution, I agree with you @absoludity , having this concept of "user-config" though a single endpoint could be worthwhile. Perhaps this `SelfSubjectRulesReview` could be an initial approach for that.
Anyway, the user-config, once loaded in the UI, should have its own lifecycle (when it should be created, updated, evicted, etc...)

But, as a first step. I do think we should have the "can I" endpoint and then think of more useful and higher-level endpoints such as user-config. But before doing so, I'd stop to reflect on the "past. present and future" on the whole set of APIs in Kubeapps.

absoludity:
> You can see how I am using this at #2172. So yes, basically I needed an endpoint to check if the user has permissions to do something (in this case, creating namespaces). If tomorrow, we need to check something else we can just reuse this endpoint without the need to modify the backend,

Right, then we'd have two `can-i` calls from the frontend. When we find we need to know a third thing about the users perms for some other aspect, we'd call the `can-i` again, as opposed to being able to modify the backend to return more info in the one call. That's all I meant.

> that's the rationale behind this implementation (vs creating something just for namespaces).

I'm not suggesting creating something just for namespaces (see my example above) - it's whether Kubeapps would be better encapsulating what we need to know about the user (extensible in the future), rather than mirroring the lower level K8s subject access review to be called multiple times in the future.

> I don't think there is a problem "mirroring" this endpoint? Can you elaborate why you are wary of this? Is it because you think this specific use case is dangerous or you are way of this becoming a "standard" and expose in the future other k8s endpoints?

No no, no danger.  Yes to the latter - I'm wary in terms of maintenance if we continue that pattern into the future, as it's easy to just add a mirror endpoint, rather than thinking carefully about the API which is extensible for the Kubeapps UX itself.

But again, happy to see this landed, I'm just keen for us to start discussing and thinking longer term about our API and where we want it to be.

Thanks!

andresmgot:
Replying inline to some of your comments:

> > * `SelfSubjectAccessReview` checks whether or the current user can perform an action (not filling in a spec.namespace means "in all namespaces")

Yes, that's what I am using in this PR.

> Regarding the long-term solution, I agree with you @absoludity , having this concept of "user-config" though a single endpoint could be worthwhile. Perhaps this `SelfSubjectRulesReview` could be an initial approach for that.

The `SelfSubjectRulesReview` response is a bit complex to parse (and maybe too detailed) so I would rather avoid that and let the API tell us if one action is allowed or not.

> Right, then we'd have two `can-i` calls from the frontend. When we find we need to know a third thing about the users perms for some other aspect, we'd call the `can-i` again, as opposed to being able to modify the backend to return more info in the one call. That's all I meant.

I don't think that's a problem? It's probably better to compose that view at different times. For example, I would like to know if I can create namespaces but I may not care if I can create `PackageManifests` (operators). Also it helps to decouple a little  bit the backend from the frontend.

> No no, no danger. Yes to the latter - I'm wary in terms of maintenance if we continue that pattern into the future, as it's easy to just add a mirror endpoint, rather than thinking carefully about the API which is extensible for the Kubeapps UX itself.

I agree that this should be considered in a case by case but I think the `can-i` endpoint makes sense (it's also true that it's not directly mirroring the API since it's simplifying the `selfsubjectaccessreview` endpoint).

pkg/kube/kube_handler.go
andresmgot: yep :) forgot to delete this

absoludity (approved):
Sorry, I thought @antgamdia already had before I went to bed, otherwise I would have with my last comment. +1'd

> For example, I would like to know if I can create namespaces but I may not care if I can create `PackageManifests` (operators).

You would only have info in that user config of things the UX needs to know - so if it wanted to know one it'd want all (for decisions in the UX). You won't need them all in the one place of the UI code, but they'd all be needed to display different parts of the UI.

> Also it helps to decouple a little bit the backend from the frontend.

I'm not sold on this one - adding 1:1 k8s API calls (even if it's only part of the spec that is required by our call) is, IMO, only marginally better than doing the call directly from the frontend (a similar function can simplify the call there) . Anyway, I agree this is part of a larger discussion about the future of an actual consistent API etc.

Thanks for the discussion, and sorry for the unintended delay!

PR #2172, opened 2020-11-18
Check if the user has permissions to create namespaces

andresmgot:
### Description of the change

Use the new endpoint (#2171) to check if the user has permissions to create a namespace and disable the button if not:

Screenshot from 2020-11-18 15-34-57

### Applicable issues

  - ref #2018

### DIFF
--- a/dashboard/src/components/Header/ContextSelector.test.tsx
+++ b/dashboard/src/components/Header/ContextSelector.test.tsx
@@ -77,10 +77,12 @@ it("shows the current cluster", () => {
       foo: {
         currentNamespace: "default",
         namespaces: ["default"],
+        canCreateNS: true,
       },
       bar: {
         currentNamespace: "default",
         namespaces: ["default"],
+        canCreateNS: true,
       },
     },
   } as IClustersState;
@@ -152,4 +154,19 @@ it("shows an error creating a namespace", () => {
   const wrapper = mountWrapper(getStore({ clusters }), <ContextSelector />);
   // The error will be within the modal
   expect(wrapper.find(CdsModal).find(Alert)).toExist();
+});
+
+it("disables the create button if not allowed", () => {
+  const clusters = {
+    currentCluster: "foo",
+    clusters: {
+      foo: {
+        currentNamespace: "default",
+        namespaces: ["default"],
+        canCreateNS: false,
+      },
+    },
+  } as IClustersState;
+  const wrapper = mountWrapper(getStore({ clusters }), <ContextSelector />);
+  expect(wrapper.find(".flat-btn")).toBeDisabled();
 });
--- a/dashboard/src/components/Header/ContextSelector.test.tsx
+++ b/dashboard/src/components/Header/ContextSelector.test.tsx
@@ -127,7 +127,7 @@ it("submits the form to create a new namespace", () => {
   actions.namespace.createNamespace = createNamespace;
   const wrapper = mountWrapper(defaultStore, <ContextSelector />);
 
-  const modalButton = wrapper.find(CdsButton).filterWhere(b => b.text() === "Create Namespace");
+  const modalButton = wrapper.find(".flat-btn");
   act(() => {
     (modalButton.prop("onClick") as any)();
   });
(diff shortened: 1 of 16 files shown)


dashboard/src/components/Header/ContextSelector.tsx
```diff
@@ -178,7 +179,14 @@ function ContextSelector() {
...
                 </form>
               </CdsModal>
-              <CdsButton status="inverse" size="sm" action="flat" onClick={openNewNSModal}>
+              <CdsButton
+                disabled={!canCreateNS}
+                status="inverse"
+                size="sm"
+                action="flat"
+                className="flat-btn"
+                onClick={openNewNSModal}
```
andresmgot, line 188: needs to conditionally add `title`

andresmgot:
This requires review cc/ @antgamdia @absoludity

dashboard/src/shared/specs/mountWrapper.tsx
```diff
@@ -32,6 +32,7 @@ export const initialState = {
       "default-cluster": {
         currentNamespace: "default",
         namespaces: ["default", "other"],
+        canCreateNS: true,
```
antgamdia, line 35: Just out of curiosity, why this default value is set as `true` whereas the one in cluster.ts is `false` ?

antgamdia:
Just to check if I'm understanding the general behavior: 

once a user authenticates and sets its cluster, the can-i endpoint determines if new ns can be created. This info is stored in the State (session scoped?).
How often is this canCreateNs been reevaluated? Each time Header.tsx is mounted (ie, each refresh) ?

dashboard/src/shared/specs/mountWrapper.tsx
andresmgot: the `cluster.ts` contains initial empty values. The empty value for booleans is `false`. But in this case, this `mountWrapper` is used in tests and it's mocking an already populated state (so it's not empty), and I set the value to `true` to maintain previous behavior (since before, the "Create Namespace" button was always enabled).

andresmgot:
> This info is stored in the State (session scoped?).
> How often is this canCreateNs been reevaluated? Each time Header.tsx is mounted (ie, each refresh) ?

Yes, the state is stored in the `Redux` state. This state is shared between components but gets refreshed everytime the application gets loaded (aka after each refresh). We are not using session cookies and we are using the browser storage for things like the user serviceaccount token or the latest namespace used.

antgamdia (approved):

dashboard/src/components/Header/ContextSelector.test.tsx
```diff
@@ -153,3 +155,18 @@ it("shows an error creating a namespace", () => {
...
+    clusters: {
+      foo: {
+        currentNamespace: "default",
+        namespaces: ["default"],
+        canCreateNS: false,
+      },
+    },
+  } as IClustersState;
+  const wrapper = mountWrapper(getStore({ clusters }), <ContextSelector />);
+  expect(wrapper.find(".flat-btn")).toBeDisabled();
```
absoludity, line 171: A few lines above we're doing `.find(CdsButton).filterWhere(b => b.text() === "Create Namespace");`, it'd be worth updating that to use the new class name (so it's clearly the same node in both cases and sets the right precedent for future changes).

dashboard/src/components/Header/ContextSelector.tsx
```diff
@@ -178,7 +179,15 @@ function ContextSelector() {
                   </CdsModalActions>
                 </form>
               </CdsModal>
-              <CdsButton status="inverse" size="sm" action="flat" onClick={openNewNSModal}>
+              <CdsButton
+                disabled={!canCreateNS}
+                title={canCreateNS ? "" : "missing permissions"}
```
absoludity, line 184: We may as well include a useful title when the user can create a NS? "Create a new namespace in the current cluster" or similar? Also, I think it'd be worth at least a phrase for the missing permissions, perhaps? "You don't have permission to create namespaces on the cluster." Fine either way, just a thought.

absoludity (approved):
LGTM, thanks!

PR #2173, opened 2020-11-18
Support charts with a '/' in their names

### LINKED ISSUE #1811: Charts with slashes (/) in the name cannot be rendered
### Description:

We noticed that if a chart contains a slash in the name (e.g. `library/kubeapps`), the chart fails to be rendered, giving a 404:

Screenshot from 2020-06-22 12-07-33
Screenshot from 2020-06-22 12-07-57

This is because `library/` gets interpreted as a subpath.

1. Create Harbor instance
2. Upload a chart to the default repo (`library`)
3. Import the default repo to Kubeapps.

To properly show the catalog, escaping the slash.

antgamdia:
Currently, charts repositories using a trailing slash as part of the chart name are not working in Kubeapps. It is due to a wrong interpretation of / as a path separator.

### Description of the change

-  asset-syncer should handle the tarball unpacking when the chart name contains a / character.
-  dashboard should properly encode the / character in a chart name as `%2F` so that the router is able to understand it's a path param.
-  assetsvc router should NOT prematurely decode the `%2F` so that it can be passed directly to the corresponding methods.
-  nginx shoud NOT decode `%2F` when proxy_pass. Note that decoding it's the default behavior and some changes have to be made (more info)

### Benefits

This PR intends to fix issue #1811. From now on, harbor instances and chart repos such as [https://repo.chartcenter.io ](url)can be used.

### Possible drawbacks

Since it is a transversal change (dashboard->nginx->assetsvc->asset-syncer->DB), I expect some bugs to appear. That's why this PR is still a draft. A more thorough testing effort is yet required.

### Applicable issues

  - fixes #1811

### Additional information

These changes do not apply to operators, just to charts.

I'm still working on this PR.

EDIT 1: added problems sections to let you know what I'm currently working on.

#### Problems found so far:

[P1] Charts become subsumed if names match: --> solved
- Repo1 contains repo1/chart
- Repo2 contains repo2/repo%2Fchart
- Result: repo1/chart is NOT displayed. repo2/repo%2Fchart IS displayed. 
- Expected: both of them should be displayed

[P2] Chart icons no longer displayed --> solved

[P3] 404 often obtained when getting deployment info --> solved

[P4] CI fails when retrieving new chart versions --> working on that 

In mirrored charts (ie, name=`repo2/repo%2Fchart`), the name once it has been installed in Helm becomes `repo2/chart`. At the moment, not able to retrieve the full name from a Helm release.

(personal notes: how are we using pagination? is it implemented in the UI?) --> ok. not used at all


andresmgot:
> (personal notes: how are we using pagination? is it implemented in the UI?)

We are not using pagination at the moment (even though the `assetsvc` supports it).

Please let us know when you want us to take a look to the code

antgamdia:
> We are not using pagination at the moment (even though the `assetsvc` supports it).

Thank you for the info! I'm using the jfrog repo for testing the charts with slashes. The response I'm getting contains almost 1900 charts and the loading/rendering process becomes pretty slow.

The point now is some of these jfrog repos are something like "jfrog/bitnami/cassandra" and it is, somehow, preventing the Cassandra bitnami chart to appear in the bitnami repo. I will have to do with the assetsvc, but I'm still figuring out what's really happening.
Once I fix these issues I will convert this draft PR to normal one.

andresmgot:
> The point now is some of these jfrog repos are something like "jfrog/bitnami/cassandra" and it is, somehow, preventing the Cassandra bitnami chart to appear in the bitnami repo. I will have to do with the assetsvc, but I'm still figuring out what's really happening.

That's probably because it's a mirror. When the assetsvc finds the same chart in two repos (with the same checksum), it replies with just one for simplicity (maybe is not that intuitive :sweat_smile: ).

antgamdia:
> That's probably because it's a mirror. When the assetsvc finds the same chart in two repos (with the same checksum), it replies with just one for simplicity (maybe is not that intuitive  ).

Mmm...good point, it can be the issue. The thing is the UX is not the best: adding a jfrog repo makes almost every bitnami chart disappear. I'm gonna take a look at that logic to see if some improvement is feasible.

andresmgot:
> Mmm...good point, it can be the issue. The thing is the UX is not the best: adding a jfrog repo makes almost every bitnami chart disappear. I'm gonna take a look at that logic to see if some improvement is feasible.

The behavior can be disabled with a flag and I would say it should be already disabled but since you are not the first person reporting that I guess it's not.

antgamdia:
> The behavior can be disabled with a flag and I would say it should be already disabled but since you are not the first person reporting that I guess it's not.

Yep, it's weird. I'm not still able to determine what's happening. Now I guess the problem could be in the asses-syncer... I'll try to debug the queries executed by the cronjobs

antgamdia:
> Yep, it's weird. I'm not still able to determine what's happening. Now I guess the problem could be in the asses-syncer... I'll try to debug the queries executed by the cronjobs

Nop, it was just the post-filtering after a successful query. All the assets are being properly saved into the DB with their full name (eg: jfrog/bitnami/wordpress). I've added a couple of commits fixing the icons and the filtering.
Digging into these changes, I've opened a new issue #2174 for improving the loading times.

I'm solving now another problem pointed out by our CI

antgamdia:
Well, this PR takes me a little bit more than I expected. I've tried to manually test the functionality and everything seems to be properly working. Nevertheless, there are plenty of changes across each Kubeapps component so I'd like to get a review and +1 from you both @absoludity and @andresmgot before merging.

Next, I will try to highlight the main changes in each part:

## Nginx

  1. **Chart->Nginx configuration**: with `rewrite ^ $request_uri` we are forcing $request_uri to have the same value as the one matched in location. That is, if a `/` is sent encoded as `%2F` it will remain intact. The default behavior for Nginx is to encode these characters before proxy_pass. 
  **Problem found**: `?` was also being escaped and, wherefore, services downstream were not properly getting query params.
  With `?$args` we force these params to appear.
  **Drawback**: in some cases, we get something like: `myURI%3Fparam=value?param=value`. 
  Honestly, I wasn't able to come up with a better solution, though.

## Assetsvc

  2. **assetsvc/handler**: 
    2.1 changed the duplicate detection scope. Now it is bounded to a repository name. Why? Mirrored charts (eg, jfrog mirroring bitnami repo) were replacing the 'original' ones.
    2.2 Sometimes, mux routes are not being matched properly. I added a workaround, in case we receive certain params, the request is forwarded from `listCharts` to `listChartsWithFilters`.
    2.3 Since params now come unescaped (because of 1.), we have to decode them prior to usage (see `decodeParam` method).
    2.4 Due to 1. (drawback), chart name param appears as `params["chartName"]` and `req.FormValue("name")`. Logic to get both params has been added.

  3. **assetsvc/handler**: tests adapted accordingly:
    3.1 added repo name when missing
    3.2  added test cases for encoded params, eg, name: `foo%2Fmy-chart`).
    3.3 changed `name+'/'+repo` id generation to `getEncodedChartIDString(name,id)` 

  4. **assetsvc/postgresql_utils**: 
    4.1 Changed duplicate detection logic as in 2.1. Now two chart are the same if they have the same digest AND they belong to the same repo.
    4.2 Added logic to decode param to store them properly into the DB (ie, `jfrog/bitnami%2Fwordpress` becomes `jfrog/bitnami/wordpress`)
    4.3 Added a fallback query when a char/file is not being retrieved. Why? When upgrading a mirrored chart (eg, jfrog/bitnami/wordpress), we only have (at that point) the repo and the name (ie "jfrog" and "Wordpress"). This changes adds a fallback query `id="jfrog whatever wordpress"` so that it can be retrieved.

  5. **assetsvc/postgresql_utils_test**: added chart names with slashes to ensure everything is working even if a chart contains a slash

## Asset-syncer

  6. **asset-syncer/utils**: created `getNameFromLastSlash` method for retrieving the chart name if it contains multiple slashes (eg, `foo/bar/mychart` will return `mychart`). This chart name is being used for uncompressing the tarball file and getting the right files.

  7. **asset-syncer/utils_test**: add test case for the method `getNameFromLastSlash`

## Dashboard UI

  8. **dashboard/repos**: remove duplicated repos. Two repos are the same if both uid matches. It was causing problems in the upgrade form.

  9. **dashboard/repos.test**: added uid to test case to check duplicates (see 8.)

  10. **dashboard/CatalogItems**: changed icon URL from `api/assetsvc/<param>`  `api/assetsvc<param>` when param already have a `/`

  11. **dashboard/chartHeader**: changed icon URL from `api/assetsvc/<param>`  `api/assetsvc<param>` when param already have a `/`

  12. **dashboard/Chart**:  changed URL from `.../charts?name=` to  `.../charts/?name=` so that it properly arrives at the assetsvc after changes in 1. and 2.

  13. **dashboard/url**:  properly encode chart name when generating URL (ie, `/` becomes `%2F`)


## Very minor things:

  1. **asset-syncer/main**: Changed default asset-syncer URL to be `localhost` to `localhost:5432`

  2. **asset-syncer.md**: removed old and unused `database-type` param from docs

  3. **assetsvc.md**: removed old and unused `database-type` param from docs

cmd/assetsvc/handler.go
```diff
@@ -336,3 +379,25 @@ func newChartVersionListResponse(c *models.Chart) apiListResponse {
 
 	return cvl
 }
+
+func decodeParam(param string, w http.ResponseWriter) string {
+	decodedParam, err := url.PathUnescape(param)
+	if err != nil {
+		log.WithError(err).Errorf("could not decode param %s", decodedParam)
```
absoludity, line 386: you should be using `param` not `decodedParam` here, as if an error was returned then `decodedParam` will be an empty string, I would expect.

cmd/assetsvc/handler.go
```diff
@@ -336,3 +379,25 @@ func newChartVersionListResponse(c *models.Chart) apiListResponse {
...
+		if w != nil {
+			response.NewErrorResponse(http.StatusBadRequest, "could not decode param: "+param).Write(w)
+		}
+	}
+	return decodedParam
+}
+
+func getEncodedChartID(c *models.Chart) string {
```
absoludity, line 394: Not sure why this fn is required - it'd be clearer for call-sites to just call what you've got on the following line?

cmd/assetsvc/handler.go
```diff
@@ -336,3 +379,25 @@ func newChartVersionListResponse(c *models.Chart) apiListResponse {
...
+	return getEncodedChartIDString(c.Repo.Name, c.Name)
+}
+
+func getEncodedChartIDString(chartRepoName, chartName string) string {
+	return getChartID(chartRepoName, url.PathEscape(chartName))
+}
+func getChartID(chartRepoName, chartName string) string {
+	return fmt.Sprintf("%s/%s", chartRepoName, chartName)
```
absoludity, line 402: Out of interest, why is the `/` between the repo and chart also not encoded, assuming this is sent over the wire as well?

cmd/assetsvc/postgresql_utils.go
```diff
@@ -92,11 +101,21 @@ func (m *postgresAssetManager) getPaginatedChartList(namespace, repo string, pag
 	return charts, 1, nil
 }
 
-func (m *postgresAssetManager) getChart(namespace, chartID string) (models.Chart, error) {
+func (m *postgresAssetManager) getChart(namespace, chartIDUnescaped string) (models.Chart, error) {
 	var chart models.ChartIconString
+	chartID, _ := url.PathUnescape(chartIDUnescaped)
```
absoludity, line 106: I'm guessing you meant to change this, but we can't ignore errors like this, it needs to be:
```
chartID, err := url.PathUnescape(chartIDUnescaped)
if err != nil {
    return models.Chart{}, err
}
```

cmd/assetsvc/postgresql_utils.go
```diff
@@ -92,11 +101,21 @@ func (m *postgresAssetManager) getPaginatedChartList(namespace, repo string, pag
...
 	err := m.QueryOne(&chart, fmt.Sprintf("SELECT info FROM %s WHERE repo_namespace = $1 AND chart_id = $2", dbutils.ChartTable), namespace, chartID)
 	if err != nil {
-		return models.Chart{}, err
```
absoludity, line 99: You've put an explanation in the PR description, but that would need to be here also. You said there that: "Added a fallback query when a char/file is not being retrieved. Why? When upgrading a mirrored chart (eg, jfrog/bitnami/wordpress), we only have (at that point) the repo and the name (ie "jfrog" and "Wordpress"). This changes adds a fallback query id="jfrog whatever wordpress" so that it can be retrieved."

What I don't understand is why you say you only have the repo and the name? It appears that you have the repo and the ID (since you're splitting it to get the name). I see it's something to do with a difference when upgrading a mirrored chart - but the `alikeChartID` that you create appears be just transforming "bitnami/wordpress" to "bitnami%wordpress" ? What is it actually matching in this case in the DB?

dashboard/src/actions/repos.ts
```diff
@@ -208,12 +208,25 @@ export const fetchRepos = (
...
         dispatch(receiveRepos(repos.items));
       } else {
-        let totalRepos = repos.items;
+        const totalRepos = repos.items;
         await Promise.all(
           otherNamespaces.map(async otherNamespace => {
             dispatch(requestRepos(otherNamespace));
             const otherRepos = await AppRepository.list(currentCluster, otherNamespace);
-            totalRepos = totalRepos.concat(otherRepos.items);
+            // Avoid addiing duplicated repos: if two repos have the same uid, skip one.
```
absoludity, line 216: If two different namespaces have the same repo added, does that mean you won't see one of them? Why? From the PR description this appears to be related to the upgrade form (presenting a repo to choose), but in that case (which is upgrading in a specific namespace) I'd expect the only possibility is if the namespace has the same repo as a global repo? Are there other call-sites of this function that aren't related to upgrading? (like listing all repos) where we would want to see repos duplicated across different namespaces?

dashboard/src/components/Catalog/CatalogItems.tsx
```diff
@@ -14,13 +14,20 @@ export default function CatalogItems({ charts, csvs, cluster, namespace }: ICata
   const chartItems: ICatalogItemProps[] = useMemo(
     () =>
       charts.map(c => {
+        let icon;
+        if (c.attributes?.icon) {
+          icon = c.attributes.icon.startsWith("/")
+            ? c.attributes.icon
+            : "/".concat(c.attributes.icon);
+          icon = "api/assetsvc".concat(icon);
```
absoludity, line 22: perhaps the above 4 lines could be just:
```
icon = "api/assetsvc/".concat(icon.replace(/^\//, "");
```

dashboard/src/components/ChartView/ChartHeader.tsx
```diff
@@ -26,6 +26,11 @@ export default function ChartHeader({
   deployButton,
   selectedVersion,
 }: IChartHeaderProps) {
+  let icon = placeholder;
+  if (chartAttrs?.icon) {
+    icon = chartAttrs.icon.startsWith("/") ? chartAttrs.icon : "/".concat(chartAttrs.icon);
+    icon = "api/assetsvc".concat(icon);
```
absoludity, line 32: Same here.

absoludity (commented):
Thanks @antgamdia ! That's a mammoth effort throughout the codebase - I'm guessing you learned a huge amount (and hope it wasn't too frustrating!)

> Well, this PR takes me a little bit more than I expected.

Please continue to update your expectations so that it's expected when things take longer than you originally envisage :)

> I've tried to manually test the functionality and everything seems to be properly working. Nevertheless, there are plenty of changes across each Kubeapps component so I'd like to get a review and +1 from you both @absoludity and @andresmgot before merging.

Thanks a lot for all the detailed notes about the changes. I've gone through and left a bunch of comments/questions. See what you think.

absoludity:
BTW, no rush to respond on this given you've got training etc.

cmd/assetsvc/postgresql_utils.go
```diff
@@ -121,11 +140,21 @@ func (m *postgresAssetManager) getChart(namespace, chartID string) (models.Chart
 	}, nil
 }
 
-func (m *postgresAssetManager) getChartVersion(namespace, chartID, version string) (models.Chart, error) {
+func (m *postgresAssetManager) getChartVersion(namespace, chartIDUnescaped, version string) (models.Chart, error) {
+	chartID, _ := url.PathUnescape(chartIDUnescaped)
```
andresmgot, line 144: here you should handle the error as well

cmd/assetsvc/postgresql_utils.go
andresmgot: Maybe this is explained elsewhere but why don't we store the chartID in the database escaped? So we don't need the extra escaping/unescaping.

cmd/assetsvc/postgresql_utils.go
andresmgot: >This changes adds a fallback query id="jfrog whatever wordpress" so that it can be retrieved

If that's correct, we may be retrieving the wrong wordpress (i.e. jfrog/impersonator/wordpress).

cmd/assetsvc/postgresql_utils_test.go
```diff
@@ -108,16 +108,16 @@ func Test_getChartFiles(t *testing.T) {
 	pgManager, mock, cleanup := getMockManager(t)
 	defer cleanup()
 
-	expectedFiles := models.ChartFiles{ID: "foo"}
+	expectedFiles := models.ChartFiles{ID: "fo/o"}
```
andresmgot, line 111: it's generally better to write new test. If this test fails we don't know if it's because the basic functionality is broken or because of an error with the new logic related to `/` in the name.

dashboard/src/actions/repos.ts
andresmgot: This is comparing `uid` so the same repo in two different namespaces won't be identified as equals. Not sure why you need this code though, seems like an error requesting the same resources twice, which should not happen (so we should fix that).

dashboard/src/actions/repos.ts
```diff
@@ -208,12 +208,25 @@ export const fetchRepos = (
...
+                  r?.metadata?.uid ? otherRepo.metadata.uid === r?.metadata?.uid : -1,
+                );
+                if (index === -1) {
+                  totalRepos.push(otherRepo);
+                }
+              } else {
+                // if no uid, add it anyway
+                totalRepos.push(otherRepo);
+              }
+            });
```
andresmgot, line 229: I still think this should not be needed but you can replace all this code simply doing at the end:

```
totalRepos = uniqBy(totalRepos, "metadata.uid")
```

(uniqBy from lodash)

dashboard/src/components/Catalog/CatalogItems.tsx
andresmgot: or more simply: `api/assetsvc/${trimStart(c.attributes.icon, "/")}`

Again `trimStart` from lodash :) whenever you need to do something with strings, arrays or objects, lodash have many useful functions.

andresmgot (commented):
Thanks @antgamdia! I agree with all @absoludity comments. I have one major question though, why don't we store in the database the encoded name? If we do so, we should not need changes in the assetsvc, no? We would just need to encode the name in the asset-syncer and decode it in the dashboard (I may be missing something).

cmd/assetsvc/handler.go
```diff
@@ -100,13 +101,14 @@ func min(a, b int) int {
 
 func uniqChartList(charts []*models.Chart) []*models.Chart {
 	// We will keep track of unique digest:chart to avoid duplicates
-	chartDigests := map[string]bool{}
+	chartDigests := map[string]map[string]bool{}
```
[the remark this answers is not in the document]
antgamdia: Mmm, I see your point @andresmgot. I also considered taking this apart in other PR, but since the problem[1] that appeared as a consequence of this change, I just left it in this PR.

But, ok, I  agree, let's revert this change and open another PR for discussing it.

[1] 
```
1. Add bitnami repo
2. Go to bitnami charts -> there are plenty of charts, eg, "bitnami/wordpress"
3. Add jfrog repo
4. Go to bitnami charts -> only 4 charts, there is no longer a "bitnami/wordpress"
```

cmd/assetsvc/handler.go
```diff
@@ -120,20 +122,31 @@ func getPaginatedChartList(namespace, repo string, pageNumber, pageSize int, sho
...
 	}
-	response.NewDataResponseWithMeta(cl, meta).Write(w)
 }
 
 // getChart returns the chart from the given repo
 func getChart(w http.ResponseWriter, req *http.Request, params Params) {
-	chartID := fmt.Sprintf("%s/%s", params["repo"], params["chartName"])
-	chart, err := manager.getChart(params["namespace"], chartID)
+	repo := decodeParam(params["repo"], w)
+	chartID := getChartID(repo, params["chartName"]) // chartName remains unscaped
```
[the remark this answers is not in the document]
antgamdia: I just postponed the transformation to the postgresql manager. There is no further handler (after removing the aforementioned workaround). It is perfectly possible to decode the param in the handler; gonna change it.
 
edit: however, if we go with the idea of storing names encoded into the DB, these changes should remain.

cmd/assetsvc/handler.go
```diff
@@ -242,11 +277,19 @@ func getChartVersionSchema(w http.ResponseWriter, req *http.Request, params Para
 
 // listChartsWithFilters returns the list of repos that contains the given chart and the latest version found
 func listChartsWithFilters(w http.ResponseWriter, req *http.Request, params Params) {
-	charts, err := manager.getChartsWithFilters(params["namespace"], params["chartName"], req.FormValue("version"), req.FormValue("appversion"))
+	// Workaround for retrieving chart name from both "chartName" or "name" params
```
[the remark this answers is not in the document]
antgamdia: It was needed because of the nginx problem, despite mux didn't capture the param properly with the `params["chartName"]` it still was retrievable from the request itself. Now it is no longer necessary.

cmd/assetsvc/handler.go
antgamdia: Yep, totally right!

cmd/assetsvc/handler.go
antgamdia: In the database, IDs are being stored like that (eg. `bitnami/wordpress`), so we need to pass this `/`to the manager.
If we store now the encoded name, we would move from `jfrog/bitnami/wordpress`to `jfrog/bitnami%2Fwordpress`

cmd/assetsvc/postgresql_utils.go
antgamdia: > > This changes adds a fallback query id="jfrog whatever wordpress" so that it can be retrieved
> 
> If that's correct, we may be retrieving the wrong wordpress (i.e. jfrog/impersonator/wordpress).

Yes, you're right. Indeed, I wanted to discuss this point with you: 
AFAIK, from Helm we're not able to determine whether the release of the chart `bitnami/wordpress` comes from `bitnami/wordpress`, `jfrog/bitnami/wordpress` or even `jfrog/impersonator/wordpress`.

Given that I just assumed _ok, as a chart repository, I don't want to have repeated charts, so, if I mirror something, I'd ensure there is solely one chart_. It is a too optimistic assumption, I know.
With this solution/workaround: it definitively may retrieve the wrong chart, but, I wasn't able to come up with a better solution. 
Note that, without this, users won't be able to upgrade charts from repos with slashes. That's why I opted for this solution.

If you have an alternative approach to overcome this problem, I really will appreciate it! If not, perhaps we should change the frontend logic and let the user decide which chart to choose if the fallback query returns more than one result.

dashboard/src/actions/repos.ts
antgamdia: If I leave this part unchanged (in the upstream Kubeapps version) and then add a new repo as global, I get the following behavior:
image
And the corresponding ` Encountered two children with the same key, 'kubeapps/jfrog'. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.`

I guess it is due to an unintended side effect when concatenating 
`AppRepository.list(currentCluster, namespace)` and `AppRepository.list(currentCluster, otherNamespace)` because they are global repos and they appear in both listings.

So, perhaps the best option so far is just to move this change to a separate PR and discuss it there.

cmd/assetsvc/postgresql_utils_test.go
antgamdia: Absolutely right! I'm gonna add the new test for adding the slash case instead of squashing the old one. My bad!

dashboard/src/actions/repos.ts
antgamdia: I'll do it in another PR, but... lodash is such a good thing for helping with these basic functions. I should think about it more often!

cmd/assetsvc/handler.go
```diff
@@ -336,3 +379,25 @@ func newChartVersionListResponse(c *models.Chart) apiListResponse {
...
 	return cvl
 }
+
+func decodeParam(param string, w http.ResponseWriter) string {
+	decodedParam, err := url.PathUnescape(param)
+	if err != nil {
+		log.WithError(err).Errorf("could not decode param %s", decodedParam)
+		if w != nil {
+			response.NewErrorResponse(http.StatusBadRequest, "could not decode param: "+param).Write(w)
+		}
```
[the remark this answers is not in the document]
antgamdia: Thanks for the improvement. You're totally right; I forgot adding the `return` so my error handling was being completely useless... 
It's a bit tedious, perhaps adding a middleware/filter/whatever it is called in go for handling this could be interesting.
I will address this change following the more idiomatic approach.

This is also related to the management of the routes: it's really annoying to have to define a apiv1.Methods("GET").Path.(...).Queries(...)` per each combination of path/params instead of having one per path.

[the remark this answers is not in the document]
absoludity: > This is also related to the management of the routes: it's really annoying to have to define a apiv1.Methods("GET").Path.(...).Queries(...)` per each combination of path/params instead of having one per path.

+100 . For a separate branch, but yes, I've wanted to see if we can just have one route regardless of params - not sure whether it's a result of the lib being used or our use of that lib.

antgamdia:
Thank you for your fruitful comments on this PR. I have been working on addressing your suggestions, namely:

- The `rewrite /api/assetsvc([^?]*)` made my day. Now it properly matches the query params so that we can get rid of the workarounds in mux. Thanks a lot @absoludity !
- Remove 'own' implementation of `path.base()` and other lodash-able algorithms
- Error handling in url encode/decode in main code and tests
- Added new tests for the 'with slash' cases instead of replacing them
- Extracted "make duplicated chart repo-scoped" (https://github.com/kubeapps/kubeapps/pull/2173#discussion_r531535059), I'll open a PR soon.
- Extracted "remove duplicated charts when upgrading" (https://github.com/kubeapps/kubeapps/pull/2173#discussion_r531663523), I'll open a PR soon.
- Fallback queries: pros and cons.  **Need your comments** on that --> https://github.com/kubeapps/kubeapps/pull/2173#discussion_r531625248. On the one hand, having them solves the problem, but, at the same time, it can return undesired results. 

I'm pushing these commits to this PR, but I'll continue working on the @andresmgot's suggestion: storing the encoded name directly into the database.

cmd/assetsvc/postgresql_utils.go
andresmgot: _talked about this in slack_

Let's do two different calls, one expecting exact matches (for the deployment use case) and one for a non-exact match (for the upgrade use case, in which we only have the repo id and the chart name).

Also, not sue if it will be included in this PR, it would be ideal to return the list of repos that contain the chart if more than one matches the query.

antgamdia:
> Let's do two different calls, one expecting exact matches (for the deployment use case) and one for a non-exact match (for the upgrade use case, in which we only have the repo id and the chart name).

I was about to implement this, but I'm unsure of the scope of this change and I wanted to discuss it a little bit more. My main point is that I don't want to enlarge this PR yet more...

Since it is a low level change (posgtresql asset manager), creating additional methods would imply changing the interface (https://github.com/kubeapps/kubeapps/blob/master/cmd/assetsvc/utils.go#L24).  Besides, in order to be able to call them separately, an additional handler should be created (https://github.com/kubeapps/kubeapps/blob/master/cmd/assetsvc/main.go#L47).

From the frontend side, the first time we notice the error (I mean, if not enabled fallback queries) is when retrieving the chart versions. It will call https://github.com/kubeapps/kubeapps/blob/master/dashboard/src/shared/Chart.ts#L20.
Should I create an alternative version of fetchChartVersions for the fallback mode and repeat this approach for every upgrade-related call?

I don't know, but, it seems this suggestion would imply many other changes and I want to be sure we all are on the same page before digging into the code :P

antgamdia:
> Let's do two different calls, one expecting exact matches (for the deployment use case) and one for a non-exact match (for the upgrade use case, in which we only have the repo id and the chart name).

As discussed during our conversation, let's leave this change as is, and will try to land an improvement in another separated PR.

Therefore, (finally!!!!), this PR is ready for the second round of revisions.

cmd/assetsvc/postgresql_utils.go
```diff
@@ -92,11 +103,28 @@ func (m *postgresAssetManager) getPaginatedChartList(namespace, repo string, pag
 	return charts, 1, nil
 }
 
-func (m *postgresAssetManager) getChart(namespace, chartID string) (models.Chart, error) {
+func (m *postgresAssetManager) getChart(namespace, chartIDUnescaped string) (models.Chart, error) {
+	return m.getChartWithFallback(namespace, chartIDUnescaped, enableFallbackQueryMode)
+}
+
+func (m *postgresAssetManager) getChartWithFallback(namespace, chartIDUnescaped string, withFallback bool) (models.Chart, error) {
```
andresmgot, line 110: this chartIDUnescaped gets me confused. As far as I know this would be always escaped, no? The thing is that sometimes this chartID would be incomplete. So valid values for this function would be:
 1.  `bitnami/wordpress` for the normal bitnami wordpress
 2. `jfrog/bitnami%2Fwordpress` when installing the chart but from the jfrog repo
 3. `jfrog/wordpress` when upgrading the chart, when we don't have the `bitnami` info

am I wrong?

chart/kubeapps/templates/kubeapps-frontend-config.yaml
```diff
@@ -80,9 +80,16 @@ data:
...
+      # see https://serverfault.com/a/906479
+      # Ex: from "/api/assetsvc/what$2Fever?param=value"
+      #     it matches as $1="/what$2Fever" and $args="param=value"
+      #     downstream services will receive "/assetsvc/what$2Fever?param=value"
       location ~* /api/assetsvc {
-        rewrite /api/assetsvc/(.*) /assetsvc/$1 break;
-        rewrite /api/assetsvc /assetsvc break;
+        rewrite ^ $request_uri; # pass the encoded url downstream as is, 
+        rewrite /api/assetsvc([^?]*) /assetsvc$1?$args break;
+        rewrite /api/assetsvc /assetsvc?$args break;
```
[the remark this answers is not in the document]
antgamdia: Yep, I've just checked it and you're right, we can delete this line. Thx!

cmd/asset-syncer/utils.go
```diff
@@ -323,9 +324,10 @@ func (f *fileImporter) importWorker(wg *sync.WaitGroup, icons <-chan models.Char
 		}
 	}
 	for j := range chartFiles {
-		log.WithFields(log.Fields{"name": j.Name, "version": j.ChartVersion.Version}).Debug("importing readme and values")
-		if err := f.fetchAndImportFiles(j.Name, r, j.ChartVersion); err != nil {
-			log.WithFields(log.Fields{"name": j.Name, "version": j.ChartVersion.Version}).WithError(err).Error("failed to import files")
+		name := j.Name
+		log.WithFields(log.Fields{"name": name, "version": j.ChartVersion.Version}).Debug("importing readme and values")
+		if err := f.fetchAndImportFiles(name, r, j.ChartVersion); err != nil {
+			log.WithFields(log.Fields{"name": name, "version": j.ChartVersion.Version}).WithError(err).Error("failed to import files")
```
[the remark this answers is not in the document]
antgamdia: Oh, didn't notice this change. It's just a reminiscence of a former change there (encoding j.Name before passing it through). That's why I renamed it. 
Gonna revert this change!

cmd/assetsvc/handler.go
```diff
@@ -120,8 +121,20 @@ func getPaginatedChartList(namespace, repo string, pageNumber, pageSize int, sho
...
+	namespace, err := url.PathUnescape(params["namespace"])
+	if err != nil {
+		handleDecodeError(params["namespace"], w, err)
+		return
+	}
+	repo, err := url.PathUnescape(params["repo"])
+	if err != nil {
+		handleDecodeError(params["repo"], w, err)
+		return
+	}
```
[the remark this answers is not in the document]
antgamdia: You're right, there's no real need to unescape parms. 
However, note the former nginx behavior prior to this PR was to unescaped everything before the proxy_pass. So all the functions in handler.go do expect something unescaped.
In order to preserve the same behavior and to prevent unnoticed side effect, I opted for manually decoding each param before using it.

cmd/assetsvc/handler.go
```diff
@@ -267,11 +383,11 @@ func newChartResponse(c *models.Chart) *apiResponse {
 		Type:       "chart",
 		ID:         c.ID,
 		Attributes: blankRawIconAndChartVersions(chartAttributes(namespace, *c)),
-		Links:      selfLink{chartPath + c.ID},
+		Links:      selfLink{chartPath + getChartID(c.Repo.Name, c.Name)},
```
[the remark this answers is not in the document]
antgamdia: Prior to the 'storing the encoded name directly into the db' change, a function would perform the encoding of the name (`getEncodedChartID`). Then I simply renamed the function to point to the non-encoding version (`getChartID`).

Currently, this function is just concatenating strings with a slash. I agree with you in that's no longer necessary.

cmd/assetsvc/handler_test.go
```diff
@@ -179,10 +196,10 @@ func Test_newChartListResponse(t *testing.T) {
 			clResponse := newChartListResponse(tt.input)
 			assert.Equal(t, len(clResponse), len(tt.result), "number of charts in response should be the same")
 			for i := range tt.result {
-				assert.Equal(t, clResponse[i].Type, "chart", "response type is chart")
-				assert.Equal(t, clResponse[i].ID, tt.result[i].ID, "chart ID should be the same")
-				assert.Equal(t, clResponse[i].Relationships["latestChartVersion"].Data.(models.ChartVersion).Version, tt.result[i].ChartVersions[0].Version, "latestChartVersion should match version at index 0")
-				assert.Equal(t, clResponse[i].Links.(selfLink).Self, pathPrefix+"/ns/"+namespace+"/charts/"+tt.result[i].ID, "self link should be the same")
+				assert.Equal(t, "chart", clResponse[i].Type, "response type is chart")
+				assert.Equal(t, getChartID(tt.result[i].Repo.Name, tt.result[i].Name), getChartID(tt.result[i].Repo.Name, tt.result[i].Name), "chart ID should be the same")
```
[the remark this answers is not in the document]
antgamdia: Good catch, thx! It's my bad; I updated many asserts because they were flipped `(actual, expected)` instead of `(expected, actual)` and I just did a wrong copy&pase...

cmd/assetsvc/postgresql_utils.go
antgamdia: Yes, totally right! We may have these three possible values. The `...Unescaped` var name is very confusing now. It made sense when dealing with encoding/decoding, but since it is directly stored as-is, I think we can rename them back.

andresmgot (approved):
Thanks for the changes! LGTM, we can do the getChart(...) in a follow-up PR

antgamdia:
Thanks! Before merging this monster PR, I'd like to have also the +1 from @absoludity

absoludity:
Ok, I'll take a look again tomorrow.

chart/kubeapps/templates/kubeapps-frontend-config.yaml
```diff
@@ -80,9 +80,15 @@ data:
         include "./server_blocks/k8s-api-proxy.conf";
       }
 
+      # Forward '/api/assetsvc' to '/assetsvc'
+      # but preserving the encoding (eg. '%2F' is not converted to '/')
+      # see https://serverfault.com/a/906479
+      # Ex: from "/api/assetsvc/what$2Fever?param=value"
+      #     it matches as $1="/what$2Fever" and $args="param=value"
+      #     downstream services will receive "/assetsvc/what$2Fever?param=value"
```
absoludity, line 88: Thanks for the clear explanation for our future selves!

cmd/asset-syncer/utils.go
```diff
@@ -429,9 +430,21 @@ func (f *fileImporter) fetchAndImportFiles(name string, r *models.RepoInternal,
...
+	// ie., "foo%2Fbar" should return "foo/bar"
+	decodedName, err := url.PathUnescape(name)
+	if err != nil {
+		log.Errorf("Cannot decode %s", name)
+		return err
+	}
+
+	// get last part of the name
+	// ie., "foo/bar" should return "bar"
+	fixedName := path.Base(decodedName) // get
```
absoludity, line 443: Not sure what the inline comment at the end of this line is for? Perhaps left over after moving it to above?

cmd/assetsvc/postgresql_utils.go
```diff
@@ -31,6 +31,9 @@ import (
 // TODO(mnelson): standardise error API for package.
 var ErrChartVersionNotFound = errors.New("chart version not found")
 
+// Temporal flag
```
absoludity, line 34: Seems odd that I'm commenting on a temporary comment for a temporary flag, but I get the impression that you like to learn and improve in everything you do :) I think you mean "temporary" as opposed to  "permanent", rather than "temporal" as opposed to "spatial". See https://english.stackexchange.com/questions/7539/temporal-vs-temporary

absoludity (approved):
Great - excellent investigation and scope-limited solution :)

cmd/assetsvc/postgresql_utils.go
antgamdia: Hahaha, thank you for your comment! :) False friends are often a bit tricky. In Spanish, `temporal` does mean both things (temporal and temporary)

antgamdia:
It seems it isn't passing one ci check, but previously it was though. Perhaps it's due to the ci changes. I don't know yet, but I'm not able to reproduce it locally.
I will check it, don't want to break something when merging

PR #2178, opened 2020-11-23
Fix upgrade test

andresmgot:
### Description of the change

Due to a new input in the apache form, the `#replicaCount-1` selector is not available (now it's `replicaCount-2`). In this PR I'm using a different selector to avoid that issue in the future (but it will still break if there is another number input).

### Applicable issues

  - fixes master builds


antgamdia (approved):
LGTM. As you mentioned, it will fail if new inputs are added, but at least it fixes the errors when building on master right now.

PR #2179, opened 2020-11-23
Remove All Namespaces option. Adapt AppList

andresmgot:
### Description of the change

As per #2018, the global namespace option "All Namespaces" causes some UX like errors when trying to deploy an application. This PR removes the option so only the views that makes sense are able to list resources in all namespaces.

In this PR, I have adapted the view AppList to this new behavior, note that is still possible to obtain an URL to list all applications:

Peek 2020-11-23 17-34

Note that I have also refactored the `AppList` component so the `AppListContainer` is no longer needed.

### Applicable issues

  - ref #2018

### Additional information

I will adapt the rest of views (`AppRepoList`, `AppRepoControl`) in a different PR to avoid accumulating more changes here.


antgamdia:
Mmm, I want to checkout the changes and give them a try. I'm not sure about the overall UX of this change, I mean: is this use case so frequent to be displayed in the header? Is the resulting URL with the query param a little bit ugly?
However, I totally agree with the general idea, but, it's just I want to double check.
Anyway, as for me, +1 to this PR

absoludity:
Regarding the URL: Why not update the URL rather than adding a query param which effectively makes the URL mean "apps in namespace default and all apps". I mean, would there be an issue using going from `/c/default-cluster/ns/mynamespace/` -> `/c/default-cluster/ns/_all/` when the switch is toggled? We'd obviously want this to not update the current namespace in the state, so that toggling back works fine, as does clicking on any other (namespaced) link. Not a big issue, just wondering what the drawbacks are that led to adding a query param with conflicting meaning.

dashboard/src/actions/repos.ts
```diff
@@ -295,12 +282,11 @@ export const updateRepo = (
     } = getState();
     try {
       const syncJobPodTemplateObj = parsePodTemplate(syncJobPodTemplate);
-      const ns = getTargetNS(getState, namespace);
```
absoludity, line 298: Great to see this extra complexity being removed!

dashboard/src/components/Config/AppRepoList/AppRepoControl.test.tsx
```diff
@@ -67,7 +68,8 @@ it("deletes the repo and refreshes list", async () => {
   expect(fetchRepos).toHaveBeenCalledWith(defaultProps.namespace, defaultProps.kubeappsNamespace);
 });
 
-it("deletes the repo and refreshes list (in other namespace)", async () => {
+// TODO: Re-enable when the repo view is refactored
+xit("deletes the repo and refreshes list (in other namespace)", async () => {
```
absoludity, line 72: In retrospect, would it have been an option to update the AppList component without disabling the "All Namespaces" functionality (so other components continued to work) until you'd finished? Otherwise you may need to hold off landing this PR until the others are ready (not a big issue either).

dashboard/src/components/Config/AppRepoList/AppRepoControl.tsx
```diff
@@ -33,19 +32,8 @@ export function AppRepoControl({
...
-        // TODO(andresmgot): This can be refactored once hex UI is dropped
-        dispatch(actions.repos.fetchRepos(currentNamespace, kubeappsNamespace));
-      } else {
-        if (currentNamespace === definedNamespaces.all || currentNamespace === kubeappsNamespace) {
-          dispatch(actions.repos.fetchRepos(currentNamespace));
-        } else {
-          dispatch(actions.repos.fetchRepos(currentNamespace, kubeappsNamespace));
-        }
-      }
+      // repoNamespace?
```
absoludity, line 35: Yes, this is related to my earlier question: I guess the scenario you're questioning here is like the one I described, except that in this case you delete a repo whic could have been from the current namespace or from another. I assume we'd need to refetch repos from the namespace of the deleted repo?

dashboard/src/containers/AppListContainer/AppListContainer.tsx
```diff
@@ -1,37 +0,0 @@
...
-  return {
-    fetchAppsWithUpdateInfo: (cluster: string, ns: string) =>
-      dispatch(actions.apps.fetchAppsWithUpdateInfo(cluster, ns)),
-    pushSearchFilter: (filter: string) => dispatch(actions.shared.pushSearchFilter(filter)),
-    getCustomResources: (cluster: string, namespace: string) =>
-      dispatch(actions.operators.getResources(cluster, namespace)),
-  };
-}
-
-export default connect(mapStateToProps, mapDispatchToProps)(AppList);
```
absoludity, line 37: Woo - thanks for doing the conversion too.

dashboard/src/shared/Namespace.ts
```diff
@@ -48,9 +48,9 @@ export default class Namespace {
 }
 
 // Set of namespaces used accross the applications as default and "all ns" placeholders
-export const definedNamespaces = {
-  all: "_all",
-};
+// export const definedNamespaces = {
+//   all: "_all",
+// };
```
absoludity, line 53: I guess this is ready to delete now?

absoludity (approved):

absoludity:
So it's just not clear to me whether you plan to hold off landing this (since some tests had to be disabled for now) until you've got the remaining changes done and tests re-enabled?

andresmgot:
> I mean: is this use case so frequent to be displayed in the header? 

We have no data about how frequent it is but it's indeed useful to get a glance of everything installed in a cluster quickly. Hiding it in a less prominent would be counter-productive for people looking for this feature.

> Is the resulting URL with the query param a little bit ugly?

It's as short and self-explanatory as I could get. If I only use `all` it could mean different things.

> I mean, would there be an issue using going from `/c/default-cluster/ns/mynamespace/apps` -> `/c/default-cluster/ns/_all/apps` when the switch is toggled? (or even `/c/default-cluster/apps`). We'd obviously want this to not update the current namespace in the state, so that toggling back works fine, as does clicking on any other (namespaced) link. Not a big issue, just wondering what the drawbacks are that led to adding a query param with conflicting meaning.

The major issue is that if the state is not populated, there is no way to switch back to the single namespace scenario. If a user shares a link `/c/default-cluster/ns/_all/apps` to another user, there is no namespace to switch back. For that case, we could fallback to the default namespace (the one encoded in the token or the first one of the list) but the logic would start to get a bit complicated. Not sure what's better? Let me try.

> So it's just not clear to me whether you plan to hold off landing this (since some tests had to be disabled for now) until you've got the remaining changes done and tests re-enabled?

Yes, my plan was using this branch as base and merge it when the follow-up PRs are merged here.

dashboard/src/actions/repos.test.tsx
```diff
@@ -493,32 +493,15 @@ describe("installRepo", () => {
...
-      "my-repo",
-      "kubeapps-namespace",
-      "http://foo.bar",
-      "",
-      "",
-      {},
-      [],
-    );
-  });
-
```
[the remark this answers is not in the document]
andresmgot: Yes, you are assuming right. I think the toggle will be explanatory enough: "**Show** repos in all namespaces". The problem is that the repos are indeed more confusing because we have: single namespace, global namespace and all namespaces. Making that distinction is difficult (I wonder if we should have gone for the `ClusterAppRepository` CRD option vs global repositories)

dashboard/src/components/Config/AppRepoList/AppRepoControl.test.tsx
andresmgot: well, yes, it would've been in a inconsistent state anyway so I prefer to use this temporary branch since there are not that many changes more.

dashboard/src/components/Config/AppRepoList/AppRepoControl.tsx
andresmgot: Yes, I added the comment to not forget about that (and simplify this implementation).

absoludity: Yep, but shouldn't the following line use `repoNamespace` instead of `currentNamespace`, since that's the list of repos which will have changed?

andresmgot: I had to re-check this. No, if we are in the "all namespaces" scenario, we should delete the repo from its namespace but then we need to fech repos from all namespaces (`currentNamespace` is an empty string in that scenario).

absoludity: Ah - so perhaps a comment that "currentNamespace will either be the same as repoNamespace or empty if browsing across all namespaces".

andresmgot:
_this is not ready to be merged_

PR #2180, opened 2020-11-23
Fix version retrieval when pushing commit

andresmgot:
### Description of the change

When syncing the chart with the bitnami repo, the version in the commit message is wrong since it's picking the PosgreSQL dependency:

https://github.com/bitnami/charts/commit/cbe4ff5431459d4f037d905aa9ed2cbc89e6c16a

Note that the `Chart.yaml` now contains both versions:

```yaml
dependencies:
  - name: postgresql
    version: '10.X.X'
name: kubeapps
version: 5.0.0
```

### DIFF
--- a/script/chart_sync_utils.sh
+++ b/script/chart_sync_utils.sh
@@ -106,7 +106,7 @@ commitAndPushChanges() {
         cd -
         return 1
     fi
-    local chartVersion=$(grep -w version: ${chartYaml} | awk '{print $2}')
+    local chartVersion=$(grep -e '^version:' ${chartYaml} | awk '{print $2}')
     git add --all .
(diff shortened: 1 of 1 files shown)


antgamdia (approved):
LGTM

absoludity (approved):
Good enough for now. I was going to suggest https://github.com/mikefarah/yq if there's a lot of other places where we pull specific values out of yaml, but seems not.

andresmgot:
> Good enough for now. I was going to suggest https://github.com/mikefarah/yq if there's a lot of other places where we pull specific values out of yaml, but seems not.

Yep, thought the same but installing it just for that was not worth it.

PR #2182, opened 2020-11-24
Hide all-namespace switch if not allowed

andresmgot:
### Description of the change

Depends on: #2179

Normal users won't have permission to list apps in all namespaces so hide the switch in that case.

### DIFF
--- a/dashboard/src/components/AppList/AppList.tsx
+++ b/dashboard/src/components/AppList/AppList.tsx
@@ -11,6 +11,7 @@ import { useEffect, useState } from "react";
 import { useDispatch, useSelector } from "react-redux";
 import { useLocation } from "react-router";
 import { Link } from "react-router-dom";
+import { Kube } from "shared/Kube";
 import { IStoreState } from "../../shared/types";
 import * as url from "../../shared/url";
 import PageHeader from "../PageHeader/PageHeader";
@@ -35,6 +36,7 @@ function AppList() {
 
   const [searchFilter, setSearchFilter] = useState("");
   const [allNS, setAllNS] = useState(false);
+  const [canSetAllNS, setCanSetAllNS] = useState(false);
   const [namespace, setNamespace] = useState(currentNamespace);
   const toggleListAllNS = () => {
     submitFilters(!allNS);
@@ -74,6 +76,12 @@ function AppList() {
   }, [dispatch, cluster, namespace]);
 
   useEffect(() => {
+    // In order to be able to list applications in all namespaces, it's necessary to be able
+    // to list/get secrets in all of them.
+    Kube.canI(cluster, "", "secrets", "list", "").then(allowed => setCanSetAllNS(allowed));
+  }, [cluster]);
+
+  useEffect(() => {
     setSearchFilter(searchQuery);
   }, [searchQuery]);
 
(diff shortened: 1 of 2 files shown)


dashboard/src/components/AppList/AppList.tsx
```diff
@@ -94,16 +102,18 @@ function AppList() {
...
-              </CdsToggle>
-            </CdsToggleGroup>
+            {canSetAllNS && (
+              <CdsToggleGroup className="flex-v-center">
+                <CdsToggle>
+                  <label>Show apps in all namespaces</label>
+                  <input
+                    type="checkbox"
+                    onChange={toggleListAllNS}
+                    checked={allNSQuery === "yes" || allNS}
```
antgamdia, line 112: Just out of curiosity: why did you choose "yes"/"no" instead of "true"/"false" ?

antgamdia (approved):
LGTM

dashboard/src/components/AppList/AppList.tsx
andresmgot: it's closer to what a human would say? :smile: It was just to make clear that this is not automatically parsed as a boolean

dashboard/src/components/AppList/AppList.tsx
```diff
@@ -73,6 +75,12 @@ function AppList() {
     dispatch(actions.operators.getResources(cluster, namespace));
   }, [dispatch, cluster, namespace]);
 
+  useEffect(() => {
+    // In order to be able to list applications in all namespaces, it's necessary to be able
+    // to list/get secrets in all of them.
+    Kube.canI(cluster, "", "secrets", "list", "").then(allowed => setCanSetAllNS(allowed));
```
absoludity, line 81: I'm still not sure why having a single "userConfig" check when the app loads to get back the user config and store that in state wouldn't be better than using `canI` wherever the UI wants to make a decision. The UI here wants to know if the user can view apps across all namespaces, but the UI has to jump down to lower level implementation of how to determine that - listing secrets across namespaces. Our backend API could encapsulate those details and just provide the user config (what they can and can't do).

absoludity (approved):
LGTM

dashboard/src/components/AppList/AppList.tsx
andresmgot: IMO, this is not "userConfig" since we are just talking about permissions (for me "config" would mean something that the user can change and configure at will). So if we want to move this to a single call in the backend it would be "userPermissions" (since at the moment, the backend has nothing configurable as far as I can remember).

Now, the "problems" with that are two (again, in my opinion):
 - It's slower to implement. I would need to implement things in two places, write tests in two places, modify the API and build the image. 
 - The response wouldn't be structured. If someone takes a look to the backend standalone, it wouldn't make sense to see why we are returning only certain permissions (create namespaces, list secrets, list apprepositories) while it's only necessary in a very specific point of the UI.

On the other hand, it's true that here the frontend needs to know that the permission to list apps maps to the permission to list secrets (and that's a decision made in the backend). Not sure if avoiding that assumption in the UI is worth the effort/downsides of the other approach.

I will merge this as it is in the dev branch but we can keep the discussion and I can open a different PR.

absoludity: Sure, `userConfig` could be `userPerms`, that's just naming. Regarding the other point about structure, yes - it'd be encapsulating all the info that the UI needs about the user to operate - totally agree it'd be slightly more work but for the benefit of encapsulating that info in the one call/place.

> I will merge this as it is in the dev branch but we can keep the discussion and I can open a different PR.

Yep, i don't feel strongly, just worth discussion.

PR #2184, opened 2020-11-25
Allow to list repositories in all namespaces

andresmgot:
### Description of the change

This is the second view in which listing resources in all namespaces is used:

Screenshot from 2020-11-25 13-07-51

Requires #2179

### Applicable issues

  - ref #2018

### Additional information

Will add more comments inline explaining some decissions.


dashboard/src/actions/repos.ts
```diff
@@ -116,16 +116,10 @@ export const deleteRepo = (
...
-        currentCluster,
-        clusters: {
-          [currentCluster]: { currentNamespace },
-        },
-      },
+      clusters: { currentCluster },
     } = getState();
     try {
       await AppRepository.delete(currentCluster, namespace, name);
-      dispatch(fetchRepos(currentNamespace));
```
andresmgot, line 128: it's not necessary to re-fetch repos here since the component is already doing this (also it's out of the scope for the function that deletes a repo)

dashboard/src/components/Config/AppRepoList/AppRepoControl.tsx
```diff
@@ -12,28 +12,27 @@ import "./AppRepoControl.css";
...
   const closeModal = () => setModalOpen(false);
   const dispatch: ThunkDispatch<IStoreState, null, Action> = useDispatch();
 
-  const handleDeleteClick = (repoName: string, repoNamespace: string, currentNamespace: string) => {
+  const handleDeleteClick = (repoName: string, repoNamespace: string) => {
     return async () => {
       await dispatch(actions.repos.deleteRepo(repoName, repoNamespace));
-      // repoNamespace?
-      dispatch(actions.repos.fetchRepos(currentNamespace));
+      refetchRepos();
```
andresmgot, line 35: the logic to re-fetch repos is in the AppList component so this component don't need to be aware of that complexity.

dashboard/src/components/Config/AppRepoList/AppRepoList.test.tsx
```diff
@@ -122,23 +154,27 @@ xdescribe("global and namespaced repositories", () => {
     ).toExist();
   });
 
-  it("shows the global repositories with the buttons disabled if the current namespace is other", () => {
+  it("shows the global repositories with the buttons disabled if the current user is not allowed to modify them", () => {
```
andresmgot, line 157: rather than basing the decision if the buttons should be disabled based on the current namespace, check if the user has permissions to perform the action.

dashboard/src/components/Config/AppRepoList/AppRepoList.tsx
```diff
@@ -16,36 +21,75 @@ import { AppRepoDisabledControl } from "./AppRepoDisabledControl";
...
-  kubeappsNamespace: string;
-}
-
-function AppRepoList({
-  cluster,
-  namespace,
-  kubeappsCluster,
-  kubeappsNamespace,
-}: IAppRepoListProps) {
+function AppRepoList() {
```
andresmgot, line 24: I've refactored this component to avoid the container

dashboard/src/components/Config/AppRepoList/AppRepoList.tsx
```diff
@@ -144,25 +200,13 @@ function AppRepoList({
...
-                  {kubeappsCluster &&
-                    (kubeappsCluster !== cluster || namespace !== kubeappsNamespace) && (
-                      <>
-                        Administrators can go to the{" "}
-                        <Link to={app.config.apprepositories(kubeappsCluster, kubeappsNamespace)}>
-                          {kubeappsNamespace}
-                        </Link>{" "}
-                        namespace to manage them.
-                      </>
-                    )}
```
andresmgot, line 159: The link to the `kubeapps` namespace is no longer needed since it's not a requirement to edit these repos.

dashboard/src/actions/repos.ts
absoludity: It's not clear to me how the component does this, since the callback you've created is dependent on `namespace` and `kubeappsNamespace` but not any current state.

I realise I must be wrong (as you've tested this) but afaict, the `useEffect` you've created will only run your callback if the dependencies to that callback change, I thought?

Edit: OK, I think I see it's because the callback `refetchRepos` is explicitly called when the component does a delete. It's still unclear to me why the callback still executes (rather than returning the memoized result) given that the dependencies it lists haven't changed?

dashboard/src/components/Config/AppRepoList/AppRepoControl.tsx
absoludity: You meant AppRepoList right? Anyway great, yes, this component only needs to know to call `refetchRepos()`.

absoludity (approved):

dashboard/src/actions/repos.ts
andresmgot: > Edit: OK, I think I see it's because the callback refetchRepos is explicitly called when the component does a delete. It's still unclear to me why the callback still executes (rather than returning the memoized result) given that the dependencies it lists haven't changed?

Yes, that's a bit tricky, what's memoized is the "body" of the function, not the function itself. I needed to do this because if not, the reference to `refetchRepos` would be different with each render. With this trick, the `useEffect` hook is only executed once. When, for example, the `namespace` changes, the `refetchRepos` ref changes and it gets executed again.

absoludity: Great, thanks - the comment makes it much clearer.

Issue #2177, opened 2020-11-23
Add FAQ in docs


antgamdia:
...and populate it initially with "Can Kubeapps be served under a sub-path?".

andresmgot:
We have a "Troubleshooting" section: https://github.com/kubeapps/kubeapps/tree/master/chart/kubeapps#troubleshooting

We can have a similar section for the FAQ

PR #2185, opened 2020-11-25
Add documentation for the Makefile scripts

antgamdia:
### Description of the change

Starting setting up the environment for the OIDC auth mode is not a straightforward task for beginners. Despite there exists a bunch of scripts in a Makefile to do so, they are not documented at all.
This PR is to simply add some hints and comments on how to use these scripts.
Furthermore, it also updates the Makefile file itself, otherwise, I wasn't able to run it.

### Benefits

Newcomers developers will set up the OIDC env more easily.

### Possible drawbacks

The changes in the .mk scripts are not tested. It 'just works' in my machine. Any suggestion to improve them is welcomed.

### Applicable issues

- Related to https://github.com/kubeapps/kubeapps/issues/2177, but just because I needed to set up this OIDC env to write the docs.

### Additional information

N/A

### DIFF
--- a/docs/developer/using-makefiles.md
+++ b/docs/developer/using-makefiles.md
@@ -1,0 +1,53 @@
+# How to set up the environment using the provided makefiles
+
+The main file is [Makefile](https://github.com/kubeapps/kubeapps/blob/master/Makefile), which will compile and prepare the production assets for then generating a set of Docker images. It is the starting point when you want to build the Kubeapps different components.
+
+For setting up the environment for running Kubeapps, we also provide (as it) makefiles for:
+* Creating a multicluster environment with Kind ([cluster-kind.mk](https://github.com/kubeapps/kubeapps/blob/master/script/cluster-kind.mk))
+* Deploying and configuring the components for getting Kubeapps running with OIDC login using Dex ([deploy-dev.mk](https://github.com/kubeapps/kubeapps/blob/master/script/deploy-dev.mk)).
+
+## Makefile for generating images
+
+### Commands
+
+Find below a list of the most used commands:
+
+``` bash
+make # will make all the kubeapps images
+make kubeapps/dashboard
+make kubeapps/apprepository-controller
+make kubeapps/kubeops
+make kubeapps/assetsvc
+make kubeapps/asset-syncer
+```
+
+> You can set the image tag manually: `IMAGE_TAG=myTag make`
+
+## Makefile for setting up the environment
+
+### Prerequisites
+
+* Install `mkcert`; you gan get it from the [official repository](https://github.com/FiloSottile/mkcert/releases).
+* Get the Kind network IP and replace it when necessary.
+    * Retrieve the IP by executing  `kubectl --namespace=kube-system get pods -o wide | grep kube-apiserver-kubeapps-control-plane  | awk '{print $6}'`
+    * Replace `172.18.0.2` with the previous IP the following files:
+        * [kubeapps-local-dev-additional-apiserver-config.yaml](../user/manifests/kubeapps-local-dev-additional-apiserver-config.yaml)
+        * [kubeapps-local-dev-additional-kind-cluster.yaml](../user/manifests/kubeapps-local-dev-additional-kind-cluster.yaml)
+        * [kubeapps-local-dev-apiserver-config.yaml](../user/manifests/kubeapps-local-dev-apiserver-config.yaml)
+        * [kubeapps-local-dev-auth-proxy-values.yaml](../user/manifests/kubeapps-local-dev-auth-proxy-values.yaml)
+        * [kubeapps-local-dev-dex-values.yaml](../user/manifests/kubeapps-local-dev-dex-values.yaml)
+### Commands
+
+```bash
+# Create two cluster with RBAC and Nginx Ingress controller
+# and configure the kube apiserver with the oidc flags
+make multi-cluster-kind
+
+# Install dex (identity service using OIDC),
+# install openldap and add default users,
+# generate certs for tls,
+# and deploy kubeapps with the proper configuration 
+make deploy-dev
+```
+
+> Many commands are using `...|| true`, so you will have to manually make sure that every command is, indeed, running OK by checking the logs.
--- a/docs/developer/using-makefiles.md
+++ b/docs/developer/using-makefiles.md
@@ -1,10 +1,12 @@
-# How to set up the environment using the provided makefiles
+# How to set up the environment using the provided makefile targets
 
 The main file is [Makefile](https://github.com/kubeapps/kubeapps/blob/master/Makefile), which will compile and prepare the production assets for then generating a set of Docker images. It is the starting point when you want to build the Kubeapps different components.
 
-For setting up the environment for running Kubeapps, we also provide (as it) makefiles for:
+For setting up the environment for running Kubeapps, we also provide (as is) makefile targets for:
 * Creating a multicluster environment with Kind ([cluster-kind.mk](https://github.com/kubeapps/kubeapps/blob/master/script/cluster-kind.mk))
 * Deploying and configuring the components for getting Kubeapps running with OIDC login using Dex ([deploy-dev.mk](https://github.com/kubeapps/kubeapps/blob/master/script/deploy-dev.mk)).
+
+> Disclaimer: these files are not being actively maintained, as they are solely intended for helping Kubeapp developers to set up the environment. If you are a contributor and you are having troubles, please feel free to [open an issue](https://github.com/kubeapps/kubeapps/issues/new).
 
 ## Makefile for generating images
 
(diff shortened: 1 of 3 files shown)


script/cluster-kind.mk
```diff
@@ -15,36 +15,36 @@ ${CLUSTER_CONFIG}:
...
+		--timeout=120s || true
 
 cluster-kind: ${CLUSTER_CONFIG}
 
 # dex will be running on the same node as the API server in the dev environment, so we can
 # reuse the key and cert from the apiserver, which already includes v3 extensions
 # for the correct alternative name (using the IP address).
 devel/dex.crt:
-	kubectl -n kube-system cp kube-apiserver-kubeapps-control-plane:etc/kubernetes/pki/apiserver.crt ./devel/dex.crt
+	docker cp kubeapps-control-plane:/etc/kubernetes/pki/apiserver.crt ./devel/dex.crt || true
```
antgamdia, line 34: I wasn't able to perform a `kubectl cp`... it started complaining about "executable file not found in $PATH”: unknown", so I just decided to go through Docker to perform this operation

docs/developer/using-makefiles.md
```diff
@@ -0,0 +1,53 @@
...
+
+> You can set the image tag manually: `IMAGE_TAG=myTag make`
+
+## Makefile for setting up the environment
+
+### Prerequisites
+
+* Install `mkcert`; you gan get it from the [official repository](https://github.com/FiloSottile/mkcert/releases).
+* Get the Kind network IP and replace it when necessary.
+    * Retrieve the IP by executing  `kubectl --namespace=kube-system get pods -o wide | grep kube-apiserver-kubeapps-control-plane  | awk '{print $6}'`
```
absoludity, line 32: Unfortunately, I don't think this will work, unless you mean that afterwards you `make delete-cluster-kind && make cluster-kind`, as some of these files (the apiserver-config ones) are config for the cluster api server itself, which has to know where to find dex.

I haven't found a way to provide an IP address for dex which is available to:

1. the api server and oauth2-proxy within the cluster,
2. the browser outside of the cluster

other than using a bridge network IP. I used to instead rely on setting an `/etc/hosts` line for `dex.local` but had to manually update the `/etc/hosts` of the api server (that's the orphaned .PHONY target you found, `update-apiserver-etc-hosts`) which was fiddly as it's not persistent (and may not even work now with distroless). Let me know if you've ideas, as I suspect we'll continue to have this issue with `tilt` or bash scripts.

absoludity (commented):
Thanks Antonio. I'm not sure about documenting the OIDC make targets (see comment below) but I'm not against it, just wary (I don't want us to end up supporting them but that's probably not an issue since it's only in dev).

I don't think we can just add `|| true` to the targets - this just hides errors. The 3 big issues that you found (well, at least 2, I guess you found the 3rd also but may have hidden it with an `|| true`) are:

1. Requirement to discover the nodes IP address on the bridge network so that Dex can be reached both inside and outside the cluster at the same address. We *could* solve this, but I've never bothered, by inspecting the `kind` network (`docker network inspect kind`) and setting the value as the next available IP on that network.

2. Kubernetes 1.19 uses distroless for the apiserver container, so we can't `kubectl cp` on there - thanks for the fix

3. openldap can't be installed from the old stable repo - it fails to download. I had to delete that repo and add `charts.helm.sh/stable` as per https://helm.sh/docs/faq/#i-am-getting-a-warning-about-unable-to-get-an-update-from-the-stable-chart-repository

But with that, everything just worked on a new machine setup, so I'd be really keen for you to remove the `|| true`'s  and find out what's failing. I suspect part of the issue may be trying to re-run targets without first `make reset-dev` to delete kubeapps, dex, openldap etc.

Let me know! Thanks for the docs.

script/cluster-kind.mk
antgamdia: Ahh, now everything makes sense. I wasn't aware of the k8s move to distroless. Thank you for letting me know :)

docs/developer/using-makefiles.md
```diff
@@ -0,0 +1,53 @@
...
+make multi-cluster-kind
+
+# Install dex (identity service using OIDC),
+# install openldap and add default users,
+# generate certs for tls,
+# and deploy kubeapps with the proper configuration 
+make deploy-dev
+```
+
+> Many commands are using `...|| true`, so you will have to manually make sure that every command is, indeed, running OK by checking the logs.
```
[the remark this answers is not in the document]
antgamdia: Ok, let's remove the `.. || true`. I fully understand it is not a good practice at all. 
I did that just to get more agility when launching different targets each time. 

For instance, I create two clusters, then I delete the first one, but I want to keep the additional one to avoid keep waiting for it to be ready. I simply relaunch `multi-cluster-kind` and ignore the errors complaining about the cluster xxxx already exists.
Analogously, I exec `deploy-dependencies`, then delete dex, then exec it again. It complains about openldap, etc, but I just ignore these errors though.

Otherwise, make just stop the execution and gives the error trace.

But, I agree, I can still do these tricks in local, but I should leave the upstream version clean.

docs/developer/using-makefiles.md
```diff
@@ -0,0 +1,53 @@
...
+make kubeapps/dashboard
+make kubeapps/apprepository-controller
+make kubeapps/kubeops
+make kubeapps/assetsvc
+make kubeapps/asset-syncer
+```
+
+> You can set the image tag manually: `IMAGE_TAG=myTag make`
+
+## Makefile for setting up the environment
```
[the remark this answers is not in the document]
antgamdia: I completely get your point, that's the tradeoff...
But, I'd rather opt for having, at least, something. It is not the perfect solution, but it saves a lot of time in setting up the full OIDC env when it is just needed for minor things.
For me, it was really useful to have the makefile when working on the subpath issue, but I do miss a document just giving hints about how to start, I had to figure out by myself.
I don't think we should actively maintain these docs, but, if they have been useful for a newcomer dev like me, perhaps they might be useful for someone else. Half a loaf is better than none, I guess

So, as for me, I'd add a disclaimer (` Disclaimer: these files are not being actively maintained, as they are solely intended for helping Kubeapp developers to set up the environment. If you are a contributor and you are having troubles, please feel free to open an issue.`) and remain the files and doc.

Regarding Tilt, definitively we should give it a try.

docs/developer/using-makefiles.md
antgamdia: Mmm, that's true, you're right. Since I was constantly creating and deleting the kind cluster, it indeed worked for me.
I don't see at the moment a way to sort out this issue rather than yours. 
I'm going to update the docs with this information and let's wait until we come up with a better environment setup with tilt.

docs/developer/using-makefiles.md
```diff
@@ -0,0 +1,58 @@
+# How to set up the environment using the provided makefile targets
+
+The main file is [Makefile](https://github.com/kubeapps/kubeapps/blob/master/Makefile), which will compile and prepare the production assets for then generating a set of Docker images. It is the starting point when you want to build the Kubeapps different components.
+
+For setting up the environment for running Kubeapps, we also provide (as is) makefile targets for:
+* Creating a multicluster environment with Kind ([cluster-kind.mk](https://github.com/kubeapps/kubeapps/blob/master/script/cluster-kind.mk))
+* Deploying and configuring the components for getting Kubeapps running with OIDC login using Dex ([deploy-dev.mk](https://github.com/kubeapps/kubeapps/blob/master/script/deploy-dev.mk)).
+
+> Disclaimer: these files are not being actively maintained, as they are solely intended for helping Kubeapp developers to set up the environment. If you are a contributor and you are having troubles, please feel free to [open an issue](https://github.com/kubeapps/kubeapps/issues/new).
```
absoludity, line 9: Great, thanks.

docs/developer/using-makefiles.md
```diff
@@ -0,0 +1,58 @@
...
+
+> You can set the image tag manually: `IMAGE_TAG=myTag make`
+
+## Makefile for setting up the environment
+
+### Prerequisites
+
+* Install `mkcert`; you can get it from the [official repository](https://github.com/FiloSottile/mkcert/releases).
+* Get the Kind network IP and replace it when necessary.
+    * Retrieve the IP for Dex:
```
absoludity, line 34: Retrieve the node's IP address on the `kind` bridge network so it can be used by Dex

docs/developer/using-makefiles.md
```diff
@@ -0,0 +1,58 @@
...
+    It is a requirement to discover the nodes IP address on the bridge network so that Dex can be reached both inside and outside the cluster at the same address. 
+    You can get this IP by inspecting the kind network (`docker network inspect kind`) and setting the value as **the next available IP on that network**.
```
absoludity, line 36: (if you don't already have any kind clusters launched, this will be the first address after the gateway, ie. something like 172.x.0.2).

docs/developer/using-makefiles.md
```diff
@@ -0,0 +1,58 @@
...
+    
+        * Another way to do so is to execute `kubectl --namespace=kube-system get pods -o wide | grep kube-apiserver-kubeapps-control-plane  | awk '{print $6}'`, but it won't work unless you exec `make delete-cluster-kind && make cluster-kind`, as some of these files (the apiserver-config ones) are config for the cluster apiserver itself, which has to know where to find dex.
```
absoludity, line 38: Just to add a little more clarity: "Another way to do so is to start the environment with `make cluster-kind` and manually verify the IP address by executing `kubectl ....`, but you will need to re-create the cluster after you've updated the config files (below) by executing `make delet....`, as some ...."

script/deploy-dev.mk
```diff
@@ -5,32 +5,36 @@
...
-	helm install ldap stable/openldap --namespace ldap \
+	kubectl --kubeconfig=${CLUSTER_CONFIG} create namespace ldap
+	helm --kubeconfig=${CLUSTER_CONFIG} repo add stable https://charts.helm.sh/stable
+	helm --kubeconfig=${CLUSTER_CONFIG} install ldap stable/openldap --namespace ldap \
 		--values ./docs/user/manifests/kubeapps-local-dev-openldap-values.yaml
 
+# Get mkcert from https://github.com/FiloSottile/mkcert/releases
 devel/localhost-cert.pem:
-	mkcert -key-file ./devel/localhost-key.pem -cert-file ./devel/localhost-cert.pem localhost 172.18.0.2
+	mkcert -key-file ./devel/localhost-key.pem -cert-file ./devel/localhost-cert.pem localhost 172.19.0.2
```
absoludity, line 23: Oh - good point, we need to include `deploy-dev.mk` in the files requiring updating (to get a cert for the correct IP - I was wondering why my browser still warned me yesterday when I tried the setup).

script/deploy-dev.mk
```diff
@@ -40,10 +44,16 @@ deploy-dev: deploy-dependencies
 	@echo "  kubeapps-user-ldap@example.org:password"
 	@echo "to authenticate with the corresponding permissions."
 
+reset-dev-kubeapps:
+	helm --kubeconfig=${CLUSTER_CONFIG} -n kubeapps delete kubeapps
+	kubectl -kubeconfig=${CLUSTER_CONFIG} delete jobs -n kubeapps --all
+	kubectl delete namespace --wait kubeapps
```
absoludity, line 50: Is there a reason the first two lines are needed? I normally just delete the namespace (which delete the helm release as it's a secret in the ns, and deletes all the jobs)?

absoludity (approved):
Excellent, thanks Antonio. I agree that it's better to have some docs with the disclaimer :)

script/deploy-dev.mk
antgamdia: Ups, this file was uploaded with my config (172.19.x.x, instead of 172.18.x.x), but yes, I should include it in that list of files.

script/deploy-dev.mk
antgamdia: Well, there is no strong reason besides my impatience... when deleting the ns, I have to keep waiting without any information about what's happening. However, if I manually delete the helm release, then the jobs and finally, everything else, I receive more information and feel happier to see each step succeeding.

But, as for the makefile target, if you think it's better to just have the ns deletion since it's clearer, go ahead, I don't mind.

script/deploy-dev.mk
antgamdia: Ups, this was an unintended change, I just uploaded my config. However, as you mention, yep, we should add this file into that list

script/deploy-dev.mk
absoludity: I don't care either way - I was just interested to know. Feedback on the process is good, definitely.

Issue #2152, opened 2020-11-06
How to configure kubeapps to be served behind a reverse proxy subpath


bsctl:
## Question:
How to configure kubeapps to be served behind a reverse proxy subpath?
For example, having a main domain https://main.domain, we want to serve it as https://main.domain/kubapps

absoludity:
It should be possible with the nginx-ingress (if you're using that): https://github.com/kubernetes/ingress-nginx/tree/master/docs/examples/rewrite#examples , though we're not testing this setup at the moment. If you're using oauth, you'll need to also set the `authProxy.oauth{Login|Logout}URI`.

bsctl:
@absoludity we used haproxy ingress controller as following:

```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  annotations:
    haproxy.org/path-rewrite: /catalog/(.*) /\1
  labels:
    app: kubeapps
    app.kubernetes.io/instance: catalog
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: kubeapps
    chart: kubeapps-4.0.4
    helm.sh/chart: kubeapps-4.0.4
    heritage: Helm
    release: catalog
  name: catalog-kubeapps
  namespace:
spec:
  rules:
  - host: console.example.com
    http:
      paths:
      - backend:
          serviceName: catalog-kubeapps
          servicePort: http
        path: /catalog
        pathType: ImplementationSpecific
status:
  loadBalancer: {}
```

Intentions are to serve Kubeapps under https://console.example.com/catalog

It seems to work except for the some assets, i.e chart icons that are not loaded by the browser (404), eg.

`https://console.example.com/api/assetsvc/v1/ns/cmp-system/assets/bitnami/rabbitmq/logo`

Is there any way to make Kubeapps proxy-aware, for example as Grafana does? I mean introduce an option like `serve_from_sub_path: 'true'`

antgamdia:
Hi @bsctl, I've been trying to reproduce your issue, but I'm not able to.

I'm using Kubeapps v2.0.0 (chart version 4.0.4) with a simple local cluster with [Kind](https://kind.sigs.k8s.io/) and a [HAProxy ingress controller](https://haproxy-ingress.github.io/v0.11/docs/getting-started/),  I used your ingress configuration as follows and every logo is being loaded well with this URL pattern: 
[...]

PR #2186, opened 2020-11-25
Add FAQ section in the chart readme (+ focus on subpath and OIDC)

antgamdia:
After the question received in https://github.com/kubeapps/kubeapps/issues/2152 we noticed there was no documentation explaining how to set up Kubeapps to be accessed in a subpath (eg., console.example.com/catalog).

### Description of the change

This PR adds a new FAQ section in the chart readme and, specifically, focuses upon explaining the changes in different parts to have a running Kubeapps with OIDC under a subpath.
Besides the `authProxy.oauth{Login|Logout}URI` flags already documented, Dex has to be configured with the `--proxy-prefix=/subpath/oauth2` flag

### Benefits

Users deploying Kubeapps with OIDC under a subpath will have docs to do so.

### Possible drawbacks

Despite I have manually tested the basic functionality was working fine, perhaps there exists some problem we are not aware of yet (such as the chart names containing slashes).

### Applicable issues

- Related https://github.com/kubeapps/kubeapps/issues/2152
- Fixes https://github.com/kubeapps/kubeapps/issues/2177

### Additional information

N/A

### DIFF
--- a/chart/kubeapps/README.md
+++ b/chart/kubeapps/README.md
@@ -155,16 +155,16 @@ kubectl delete namespace kubeapps
 
 ## FAQ
 
-- [How to install Kubeapps for demo purposes?](#how-to-install-kubeapps-for-demo-purposes-)
-- [How to install Kubeapps in production scenarios?](#how-to-install-kubeapps-in-production-scenarios-)
-- [How to use Kubeapps?](#how-to-use-kubeapps-)
+- [How to install Kubeapps for demo purposes?](#how-to-install-kubeapps-for-demo-purposes)
+- [How to install Kubeapps in production scenarios?](#how-to-install-kubeapps-in-production-scenarios)
+- [How to use Kubeapps?](#how-to-use-kubeapps)
 - [How to configure Kubeapps with Ingress](#how-to-configure-kubeapps-with-ingress)
   * [Serving Kubeapps in a subpath](#serving-kubeapps-in-a-subpath)
-- [Can Kubeapps be installed in more than one cluster?](#can-kubeapps-be-installed-in-more-than-one-cluster-)
-- [How to install Kubeapps without Internet connection?](#how-to-install-kubeapps-without-internet-connection-)
-- [Does Kubeapps support private repositories?](#does-kubeapps-support-private-repositories-)
-- [Does Kubeapps support Operators?](#does-kubeapps-support-operators-)
-- [More questions?](#more-questions-)
+- [Can Kubeapps be installed in more than one cluster?](#can-kubeapps-be-installed-in-more-than-one-cluster)
+- [How to install Kubeapps without Internet connection?](#how-to-install-kubeapps-without-internet-connection)
+- [Does Kubeapps support private repositories?](#does-kubeapps-support-private-repositories)
+- [Does Kubeapps support Operators?](#does-kubeapps-support-operators)
+- [More questions?](#more-questions)
 
 ### How to install Kubeapps for demo purposes?
 
--- a/chart/kubeapps/README.md
+++ b/chart/kubeapps/README.md
@@ -153,6 +153,104 @@ If you have dedicated a namespace only for Kubeapps you can completely clean the
 kubectl delete namespace kubeapps
 ```
 
+## FAQ
+
+- [How to install Kubeapps for demo purposes?](#how-to-install-kubeapps-for-demo-purposes-)
+- [How to install Kubeapps in production scenarios?](#how-to-install-kubeapps-in-production-scenarios-)
+- [How to use Kubeapps?](#how-to-use-kubeapps-)
+- [How to configure Kubeapps with Ingress](#how-to-configure-kubeapps-with-ingress)
+  * [Serving Kubeapps in a subpath](#serving-kubeapps-in-a-subpath)
+- [Can Kubeapps be installed in more than one cluster?](#can-kubeapps-be-installed-in-more-than-one-cluster-)
+- [How to install Kubeapps without Internet connection?](#how-to-install-kubeapps-without-internet-connection-)
+- [Does Kubeapps support private repositories?](#does-kubeapps-support-private-repositories-)
+- [Does Kubeapps support Operators?](#does-kubeapps-support-operators-)
+- [More questions?](#more-questions-)
+
+### How to install Kubeapps for demo purposes?
+
+This snippet shows how to install Kubeapps for exclusively **demo purposes** as explained in [getting started](https://github.com/kubeapps/kubeapps/blob/master/docs/user/getting-started.md).
+
+```bash
+
+# Create namespace kubeapps
+kubectl create namespace kubeapps
+
+# Add the Bitnami chart repository and install Kubeapps
+helm repo add bitnami https://charts.bitnami.com/bitnami
+helm install kubeapps --namespace kubeapps bitnami/kubeapps
+
+# Create a serviceaccount with a clusterrolebinding
+kubectl create serviceaccount kubeapps-operator
+kubectl create clusterrolebinding kubeapps-operator --clusterrole=cluster-admin --serviceaccount=default:kubeapps-operator
+
+# Retrieve the token in linux/macOS
+kubectl get secret $(kubectl get serviceaccount kubeapps-operator -o jsonpath='{range .secrets[*]}{.name}{"\n"}{end}' | grep kubeapps-operator-token) -o jsonpath='{.data.token}' -o go-template='{{.data.token | base64decode}}' && echo
+
+# Make Kubeapps accesible from your local address
+kubectl port-forward -n kubeapps svc/kubeapps 8080:80
+
+# Visit http://127.0.0.1:8080 in you browser and enter the token you obtained before.
+```
+
+### How to install Kubeapps in production scenarios?
+
+For any user-facing installation you should [configure an OAuth2/OIDC provider](https://github.com/kubeapps/kubeapps/blob/master/docs/user/using-an-OIDC-provider.md) to enable secure user authentication with Kubeapps and the cluster.
+Please also refer to the [Access Control](https://github.com/kubeapps/kubeapps/blob/master/docs/user/access-control.md) documentation to configure fine-grained access control for users.
+
+### How to use Kubeapps?
+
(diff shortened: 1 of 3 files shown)


chart/kubeapps/README.md
```diff
@@ -153,6 +153,104 @@ If you have dedicated a namespace only for Kubeapps you can completely clean the
...
+  * [Serving Kubeapps in a subpath](#serving-kubeapps-in-a-subpath)
+- [Can Kubeapps be installed in more than one cluster?](#can-kubeapps-be-installed-in-more-than-one-cluster)
+- [How to install Kubeapps without Internet connection?](#how-to-install-kubeapps-without-internet-connection)
+- [Does Kubeapps support private repositories?](#does-kubeapps-support-private-repositories)
+- [Does Kubeapps support Operators?](#does-kubeapps-support-operators)
+- [More questions?](#more-questions)
+
+### How to install Kubeapps for demo purposes?
+
+This snippet shows how to install Kubeapps for exclusively **demo purposes** as explained in [getting started](https://github.com/kubeapps/kubeapps/blob/master/docs/user/getting-started.md).
```
absoludity, line 171: Rather than duplicating (and maintaining duplicate info), why not just refer to the getting started for his answer?

chart/kubeapps/README.md
```diff
@@ -153,6 +153,104 @@ If you have dedicated a namespace only for Kubeapps you can completely clean the
...
+```
+
+### How to install Kubeapps in production scenarios?
+
+For any user-facing installation you should [configure an OAuth2/OIDC provider](https://github.com/kubeapps/kubeapps/blob/master/docs/user/using-an-OIDC-provider.md) to enable secure user authentication with Kubeapps and the cluster.
+Please also refer to the [Access Control](https://github.com/kubeapps/kubeapps/blob/master/docs/user/access-control.md) documentation to configure fine-grained access control for users.
+
+### How to use Kubeapps?
+
+Have a look to [this guide](https://github.com/kubeapps/kubeapps/blob/master/docs/user/dasboard.md) for knowing how to use the Kubeapps dashboard.
```
absoludity, line 202: I think we should generally avoid non-semantic links ("this guide" or "click here" etc.), so this could be reworded as "Have a look at the [Kubeapps dashboard documentation] for information about using Kubeapps." or similar?

chart/kubeapps/README.md
```diff
@@ -153,6 +153,104 @@ If you have dedicated a namespace only for Kubeapps you can completely clean the
...
+
+```bash
+helm install kubeapps --namespace kubeapps \
+  --set ingress.enabled=true \
+  --set ingress.hostname=example.com \
+    bitnami/kubeapps
+```
+#### Serving Kubeapps in a subpath
+
+Eventually, you may want to serve Kubeapps with a subpath, for instance `http://example.com/subpath`. If you are using the ingress configuration provided by the Kubeapps chart, note that currently it does not support customizing that `path` value, so, consequently you have to manually patch the ingress object:
```
absoludity, line 216: I'd not use the word "Eventually" as this implies that if people use Kubeapps long enough, they will end up wanting to do this, which I don't think is the case? Perhaps just drop the "Eventually, " and instead "You may want to server..."

chart/kubeapps/README.md
```diff
@@ -153,6 +153,104 @@ If you have dedicated a namespace only for Kubeapps you can completely clean the
...
+
+Besides, if you are using the OAuth2/OIDC login (more information in [this guide](https://github.com/kubeapps/kubeapps/blob/master/docs/user/using-an-OIDC-provider.md)), you will need, also, to configure the different URLs:
+
+```bash
+helm install kubeapps bitnami/kubeapps \
+  --namespace kubeapps \
+  # ... other OIDC flags 
+  --set authProxy.oauthLoginURI="/subpath/oauth2/login" \
+  --set authProxy.oauthLogoutURI="/subpath/oauth2/logout" \
+  --set authProxy.additionalFlags="{<other flags>,--proxy-prefix=/subpath/oauth2}"
```
absoludity, line 230: Excellent, thanks for looking into the issue there!

chart/kubeapps/README.md
```diff
@@ -153,6 +153,104 @@ If you have dedicated a namespace only for Kubeapps you can completely clean the
...
+```
+
+### Can Kubeapps be installed in more than one cluster?
```
absoludity, line 233: Hmm, the point of the multicluster support is that you *don't* need to install kubeapps in more than one cluster to be able to use Kubeapps to install apps into more than one cluster. Perhaps: "Can Kubeapps install apps into more than one cluster?"

chart/kubeapps/README.md
```diff
@@ -153,6 +153,104 @@ If you have dedicated a namespace only for Kubeapps you can completely clean the
...
+
+Yes! Kubeapps 2.0+ supports multicluster environments. Check out [this guide](https://github.com/kubeapps/kubeapps/blob/master/docs/user/deploying-to-multiple-clusters.md) to know more.
+
+### How to install Kubeapps without Internet connection?
+
+Yes! Follow [this guide](https://github.com/kubeapps/kubeapps/blob/master/docs/user/offline-installation.md) to discover how to perform an installation in an air-gapped scenario.
```
absoludity, line 239: If you want to start with "Yes!", perhaps update the title to a "Can Kubeapps be installed without ..."?

docs/user/using-an-OIDC-provider.md
```diff
@@ -108,6 +116,13 @@ helm install kubeapps bitnami/kubeapps \
   --set authProxy.cookieSecret=$(echo "not-good-secret" | base64) \
   --set authProxy.additionalFlags="{--cookie-secure=false}"
 ```
+> If you are serving Kubeapps under a subpath (eg., "example.com/subpath") you also need to set the `authProxy.oauthLoginURI` and `authProxy.oauthLogoutURI` flags, as well as the additional flag `--proxy-prefix`. For instance:
```
absoludity, line 119: Perhaps just have a separate section for this new info in this same doc, "Serving Kubeapps from a subpath while supporting OIDC" and referring to it in all 3 places?

absoludity (approved):
Excellent! A few suggestions, see what you think.

chart/kubeapps/README.md
antgamdia: Right... I first thought about having a collection of snippets for each question, but, since we already have a bunch of guides, I opted to directly point to them, as you suggest. However, given that the chart doc is the starting point for some users, I think perhaps it is a good idea to have a full example of how to start running Kubeapps.

I'd rather suggest adding this snippet in the TL;DR or even just add a link to the getting started there.
I'm just considering users who just want to give it a try without thoroughly read the docs.

chart/kubeapps/README.md
antgamdia: Totally agree with that, we should avoid such bad names for links. I just wrote it after solving my problems with OIDC and wasn't able to come up with a better naming :P . Gonna reword all the links! Thank you for pointing this out

docs/user/using-an-OIDC-provider.md
antgamdia: Yep, I think it's the best option for avoiding duplication. The point is: currently, changing the path is not supported in the chart values, so you have to manually patch the ingress object.

If we move these instructions to a separated section, perhaps we are giving much more importance to something not supported out-of-the-box by Kubeapps and wherefore I just decided to put it as a side note.

What do you think? Should we: 
i) support changing ingress path in the values.yaml or not?
ii) move these side notes to a separate section regardless of i) ?

antgamdia: Update: after a conversation with @andresmgot , it seems it is possible to change the path by using the [Values.ingress.extraHosts](https://github.com/kubeapps/kubeapps/blob/master/chart/kubeapps/templates/ingress.yaml#L33). I'll check it just to confirm and will update the docs accordingly.

chart/kubeapps/README.md
```diff
@@ -153,6 +156,87 @@ If you have dedicated a namespace only for Kubeapps you can completely clean the
...
+You may want to serve Kubeapps with a subpath, for instance `http://example.com/subpath`, you have to set the proper Ingress configuration. If you are using the ingress configuration provided by the Kubeapps chart, you will have to set the `ingress.extraHosts` parameter:
+
+```bash
+helm install kubeapps --namespace kubeapps \
+    --set ingress.enabled=true
+    --set ingress.extraHosts[0].name="console.example.com"
+    --set ingress.extraHosts[0].path="/catalog"
+    bitnami/kubeapps
+```
+> This will also create an additional ingress rule serving Kubeapps at `kubeapps.local/`. In case you don't want it, you can always patch the ingress object generated before (where we just set the `ingress.hostname`) by executing: `kubectl patch ingress kubeapps -n kubeapps --type=json -p='[{"op": "replace", "path": "/spec/rules/0/http/paths/0/path", "value":"/subpath"}]'`
```
andresmgot, line 206: You can simply use `--set ingress.hostname=""` to avoid the default hostname.

docs/user/using-an-OIDC-provider.md
```diff
@@ -130,7 +139,6 @@ helm install kubeapps bitnami/kubeapps \
   --set authProxy.additionalFlags="{--cookie-secure=false,--scope=https://www.googleapis.com/auth/userinfo.email https://www.googleapis.com/auth/cloud-platform}" \
   --set frontend.proxypassAccessTokenAsBearer=true
 ```
-
```
andresmgot, line 133: I think this empty line is needed.

andresmgot (approved):
Thanks for the docs! Just a couple of minor comments.

chart/kubeapps/README.md
antgamdia: Ah, so the `{{ if .Values.ingress.hostname }}`  will return false if an empty string is passed. Interesting... If so, that's much much better than patching the ingress :facepalm: 
Thanks for the tip!

docs/user/using-an-OIDC-provider.md
antgamdia: Yep, my bad accidentally deleted it

chart/kubeapps/README.md
absoludity: Yep, fair point.

PR #2187, opened 2020-11-25
Change namespace in location rather than redirect to the app list

### LINKED ISSUE #2018: Refactor namespace management UX

andresmgot:
### Description of the change

Ref #2018

In the linked issue, there are two (last) items regarding the "change context" button:

 - Disable it for some views
 - Just change the namespace in the state

This PR address the second, and after some thought, I am not sure about the first. After implementing this, I noticed that most of the views (12 out of 17) work without any further change. This is because after migrating to react hooks, components are forced to re-fetch resources if the namespace changes in the state.

My goal is, after this PR and since the fixes should be easy for the components failing, to allow to change context in any view (as in Kubeapps 1.x) without redirecting to the app list.

Note that I will use this branch as base while I fix those. Let me know what you think.

### Applicable issues

  - fixes #2018


dashboard/src/components/Header/ContextSelector.tsx
```diff
@@ -57,7 +59,14 @@ function ContextSelector() {
     setStateNamespace(event.target.value);
   const changeContext = () => {
     dispatch(actions.namespace.setNamespace(cluster, namespace));
-    dispatch(push(app.apps.list(cluster, namespace)));
+    // Regex matching a namespaced route: e.g. /c/cluster/ns/namespace
+    const nsRegex = /^\/c\/([^/]*)\/ns\/[^/]*\//;
+    if (nsRegex.test(location.pathname)) {
+      // Change the namespace in the route
+      history.push(
+        location.pathname.replace(nsRegex, `/c/$1/ns/${namespace}/`).concat(location.search),
```
absoludity, line 67: Nice. I was wondering how the two substitutions are treated differently, but see the braces are required for template literals but it'll ignore the `$1` without braces.

dashboard/src/components/Header/ContextSelector.tsx
```diff
@@ -70,7 +79,7 @@ function ContextSelector() {
     if (created) {
       closeNewNSModal();
       dispatch(actions.namespace.setNamespace(cluster, newNS));
-      dispatch(push(app.apps.list(cluster, newNS)));
+      history.push(app.apps.list(cluster, newNS));
```
absoludity, line 82: I'm confused why you're not dispatching this one, when creating a new namespace? Wouldn't we want to switch to the new namespace in that case, rather than just pushing the update to the history? Or what am I missing? What if I create a new namespace when I'm on a non-namespaced view? Or perhaps that's the intention (for a namespaced view, the state is updated due to the url causing the components to update, but for a non-namespaced it doesn't affect the current components).

absoludity (approved):
> My goal is, after this PR and since the fixes should be easy for the components failing, to allow to change context in any view (as in Kubeapps 1.x) without redirecting to the app list.
> 
> Note that I will use this branch as base while I fix those. Let me know what you think.

Yep, sounds good. I think this is what we discussed on the issue a while back ([avoiding disabling the context switch](https://github.com/kubeapps/kubeapps/issues/2018#issuecomment-705150062)).

dashboard/src/components/Header/ContextSelector.tsx
andresmgot: No, there is no behavioral change here. The only thing is that I am using now `useHistory` from `react-router` rather than directly `push` from `connected-react-router`. I am doing this to be able to easily mock `push` in the tests. `push` returns a `Thunkdispatch` so that's why you need to dispatch it but it's not necessary for `history.push`.

absoludity: Huh, I would have thought (but haven't checked) that `dispatch(push(...))` would be resulting in a redux `LOCATION_CHANGE` action, which we handle in certain reducers, whereas I wouldn't have thought `history.push(...)` was related to redux at all (so no action being handled.

andresmgot: > whereas I wouldn't have thought `history.push(...)` was related to redux at all (so no action being handled.

`history.push` also triggers a `LOCATION_CHANGE`, in fact, any change in the URL triggers that

absoludity: OIC - so the extra `dispatch` around the `dispatch(push(...))` was not related to dispatching a `LOCATION_CHANGE` itself. Makes sense, thanks.

PR #2189, opened 2020-11-26
Adapt components to change the namespace in the context

andresmgot:
### Description of the change

After #2187, adapt all the components to be able to change the context from it. As explained there, most of the components didn't require any change because after a change in the namespace, the related resources get reloaded but there are some exceptions. I will explain those inline.

### Possible drawbacks

In the future we may introduce changes that break this behavior again.

### Applicable issues

  - ref #2018

### DIFF
--- a/dashboard/src/actions/charts.ts
+++ b/dashboard/src/actions/charts.ts
@@ -1,10 +1,9 @@
 import { JSONSchema4 } from "json-schema";
-import { Dispatch } from "redux";
 import { ThunkAction } from "redux-thunk";
 import { ActionType, createAction } from "typesafe-actions";
 
 import Chart from "../shared/Chart";
-import { ForbiddenError, IChart, IChartVersion, IStoreState, NotFoundError } from "../shared/types";
+import { FetchError, IChart, IChartVersion, IStoreState, NotFoundError } from "../shared/types";
 
 export const requestCharts = createAction("REQUEST_CHARTS");
 
@@ -60,16 +59,6 @@ const allActions = [
 
 export type ChartsAction = ActionType<typeof allActions[number]>;
 
-function dispatchError(dispatch: Dispatch, err: Error) {
-  if (err.message.match("could not find")) {
-    dispatch(errorChart(new NotFoundError(err.message)));
-  } else if (err.message.match("Unable to validate user")) {
-    dispatch(errorChart(new ForbiddenError(err.message)));
-  } else {
-    dispatch(errorChart(err));
-  }
-}
-
 export function fetchCharts(
   cluster: string,
   namespace: string,
@@ -83,7 +72,7 @@ export function fetchCharts(
         dispatch(receiveCharts(charts));
       }
     } catch (e) {
-      dispatchError(dispatch, e);
+      dispatch(errorChart(new FetchError(e.message)));
     }
   };
 }
@@ -102,7 +91,7 @@ export function fetchChartVersions(
       }
       return versions;
     } catch (e) {
-      dispatchError(dispatch, e);
+      dispatch(errorChart(new FetchError(e.message)));
       return [];
     }
   };
@@ -139,7 +128,7 @@ export function getChartVersion(
         dispatch(selectChartVersion(chartVersion, values, schema));
       }
     } catch (e) {
-      dispatchError(dispatch, e);
+      dispatch(errorChart(new FetchError(e.message)));
     }
   };
 }
@@ -158,7 +147,7 @@ export function getDeployedChartVersion(
         dispatch(receiveDeployedChartVersion(chartVersion, values, schema));
       }
     } catch (e) {
-      dispatchError(dispatch, e);
+      dispatch(errorChart(new FetchError(e.message)));
     }
   };
 }
--- a/dashboard/src/actions/charts.ts
+++ b/dashboard/src/actions/charts.ts
@@ -7,6 +7,8 @@ import { FetchError, IChart, IChartVersion, IStoreState, NotFoundError } from ".
 
 export const requestCharts = createAction("REQUEST_CHARTS");
 
+export const requestChart = createAction("REQUEST_CHART");
+
 export const receiveCharts = createAction("RECEIVE_CHARTS", resolve => {
   return (charts: IChart[]) => resolve(charts);
 });
@@ -46,6 +48,7 @@ export const errorReadme = createAction("ERROR_README", resolve => {
 
 const allActions = [
   requestCharts,
+  requestChart,
   errorChart,
   receiveCharts,
   receiveChartVersions,
@@ -122,7 +125,7 @@ export function getChartVersion(
 ): ThunkAction<Promise<void>, IStoreState, null, ChartsAction> {
   return async dispatch => {
     try {
-      dispatch(requestCharts());
+      dispatch(requestChart());
       const { chartVersion, values, schema } = await getChart(cluster, namespace, id, version);
       if (chartVersion) {
         dispatch(selectChartVersion(chartVersion, values, schema));
--- a/dashboard/src/components/AppView/AppView.test.tsx
+++ b/dashboard/src/components/AppView/AppView.test.tsx
@@ -82,7 +82,7 @@ describe("AppViewComponent", () => {
     expect(wrapper.find(LoadingWrapper)).toExist();
   });
 
-  it("renders only a fetch error if exists", () => {
+  it("renders a fetch error only", () => {
     const wrapper = mountWrapper(
       defaultStore,
       <AppViewComponent {...validProps} error={new FetchError("not found")} />,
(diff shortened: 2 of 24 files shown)


dashboard/src/actions/charts.ts
```diff
@@ -60,16 +59,6 @@ const allActions = [
...
-function dispatchError(dispatch: Dispatch, err: Error) {
-  if (err.message.match("could not find")) {
-    dispatch(errorChart(new NotFoundError(err.message)));
-  } else if (err.message.match("Unable to validate user")) {
-    dispatch(errorChart(new ForbiddenError(err.message)));
-  } else {
-    dispatch(errorChart(err));
-  }
-}
-
```
andresmgot, line 72: this was a remnant of the 1.x implementation where the error type was read by the ErrorSelector.

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -146,6 +146,9 @@ export default function AppView({
     setResourceRefs(parseResources(parsedManifest, cluster, app.namespace));
   }, [app, cluster]);
 
+  if (error && error.constructor === FetchError) {
+    return <Alert theme="danger">Application not found. Received: {error.message}</Alert>;
+  }
```
andresmgot, line 151: This was an error I saw in several places, where the components was partially rendered with wrong information from the previous namespace. To avoid that, early return if there is an error fetching resources.

dashboard/src/containers/AppNewContainer/AppNewContainer.tsx
```diff
@@ -30,7 +30,7 @@ function mapStateToProps(
     chartNamespace: params.global === "global" ? config.kubeappsNamespace : params.namespace,
     cluster: params.cluster,
     chartVersion: params.version,
-    error: apps.error,
+    error: apps.error || charts.selected.error,
```
andresmgot, line 33: when rendering the deployment form we also need to render the error when fetching the chart if exists.

dashboard/src/reducers/charts.ts
```diff
@@ -88,8 +88,6 @@ const chartsReducer = (
         isFetching: false,
         selected: chartsSelectedReducer(state.selected, action),
       };
-    case getType(actions.namespace.setNamespaceState):
-      return { ...initialState };
```
andresmgot, line 92: This was deleting the chart state when changing the namespace. I believe this was to re-fetch chart information when checking different applications but it was not necessary. 

In fact, this was causing an issue: Since charts don't depend on the context namespace (they depend on the chart namespace), after cleaning up the state, charts were not being re-fetched, since functions that fetch them don't depend on the namespace.

dashboard/src/shared/url.ts
```diff
@@ -3,9 +3,15 @@ import { IChartVersion, IRepo } from "./types";
 
 export const app = {
   apps: {
-    new: (cluster: string, namespace: string, cv: IChartVersion, version: string) => {
+    new: (
+      cluster: string,
+      namespace: string,
+      cv: IChartVersion,
+      version: string,
+      globalNamespace: string,
```
andresmgot, line 11: this change (adding as a parameter the `globalNamespace`) is necessary, because when using the `kubeapps` namespace, the link was rendered as non global: e.g. `/c/default/ns/kubeapps/apps/new/bitnami/apache`. When changing the namespace in that scenario, just the namespace gets changed in the URL so it would turn to `/c/default/ns/foo/apps/new/bitnami/apache`, rather than `/c/default/ns/foo/apps/new-from-global/bitnami/apache` which is the correct link.

dashboard/src/shared/url.ts
```diff
@@ -19,17 +25,16 @@ export const app = {
...
-    },
-    version: (
+    get: (
       cluster: string,
       namespace: string,
       chartName: string,
-      chartVersion: string,
       repo: IRepo,
-    ) => `${app.charts.get(cluster, namespace, chartName, repo)}/versions/${chartVersion}`,
+      globalNamespace: string,
```
andresmgot, line 33: same for `charts` vs `global-charts`

dashboard/src/reducers/charts.ts
absoludity: It may have been before certain data was stored per namespace in the state. I'd think we'd certainly want to ensure that any data which is not stored per namespace is either cleared or updated to be stored per namespace, but yeah, resetting to the initial state may no longer be necessary if things are stored per namespace.

dashboard/src/shared/url.ts
absoludity: Right - good find.

dashboard/src/actions/charts.ts
andresmgot: why would we? I mean, both are 404 errors, it would be difficult to differ between those. Also, the body of the error is still shown to the user so it's not that we are hiding anything.

dashboard/src/components/AppView/AppView.test.tsx
```diff
@@ -81,6 +82,15 @@ describe("AppViewComponent", () => {
     expect(wrapper.find(LoadingWrapper)).toExist();
   });
 
+  it("renders only a fetch error if exists", () => {
```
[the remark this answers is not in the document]
andresmgot: I mean, if the error exists

dashboard/src/components/DeploymentForm/DeploymentForm.test.tsx
```diff
@@ -70,6 +72,23 @@ describe("renders an error", () => {
     expect(wrapper.find(Alert).exists()).toBe(true);
     expect(wrapper.find(Alert).html()).toContain("wrong format!");
   });
+
+  it("only renders a fetch error if exists", () => {
```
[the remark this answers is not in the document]
andresmgot: also meant the error here

dashboard/src/reducers/charts.test.ts
```diff
@@ -38,22 +38,4 @@ describe("chartReducer", () => {
...
-            ...initialState.selected,
-            error,
-          },
-        },
-        {
-          type: getType(actions.namespace.setNamespaceState) as any,
-        },
-      ),
-    ).toEqual({ ...initialState });
-  });
```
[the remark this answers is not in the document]
andresmgot: (explaining in the comment below)

dashboard/src/reducers/charts.ts
andresmgot: I am not sure if I'm following your reasoning but no data is stored per namespace (unless the namespace info itself). The way it's structured now, the app info or the chart info is stored independently and it may require to be re-fetched when the namespace change.

In this case, the chart state doesn't depend on the user namespace but on the chart namespace. This will be the same for namespaced repositories but no for global repositories That's why it should not be cleaned up when changing namespaces.

dashboard/src/components/AppView/AppView.tsx
andresmgot: Yes, it would be more clean if the state doesn't contain wrong info (in this case I believe is the component state, not the redux state the one containing old info, need to check) but we need this early return anyway. If not, we would render an empty (or loading gif) header and the error in a subcomponent, rather than at the top.

dashboard/src/reducers/charts.ts
absoludity: My reasoning was: if we only store the current chart in the state (rather than charts per namespace), then when the user switches namespaces we'd want to avoid displaying a chart which may not be relevant in the new namespace (the issue), so we'd need to be sure we clear the state (which is what I think this code intended, but as you say, didn't account for the difference between the user-selected namespace and the chart namespace, but you've fixed that by moving it).

If we stored charts per (chart) namespace, we wouldn't need to do this, I don't think? (though there may be other issues doing that). Anyway, I'm fine with the way you've chosen.

absoludity (approved):

dashboard/src/actions/charts.ts
andresmgot: right, it's true that it can be a Forbidden error but in any case, we are not using that distinction now

dashboard/src/reducers/charts.ts
andresmgot: > If we stored charts per (chart) namespace, we wouldn't need to do this, I don't think? (though there may be other issues doing that). Anyway, I'm fine with the way you've chosen.

yes, we could restructure the store to be aware of the chart namespace (and in general, that's something we could do for improving performance and caching stuff) but it would require a slighly big refactor that I am not sure would pay off for the benefit it gives.

dashboard/src/components/OperatorInstanceForm/OperatorInstanceForm.tsx
```diff
@@ -82,6 +82,10 @@ export default function DeploymentFormBody({
 
   useEffect(() => {
     dispatch(actions.operators.getCSV(cluster, namespace, csvName));
+    // Clean up component state
+    setDefaultValues("");
+    setCRD(undefined);
+    setIcon(placeholder);
```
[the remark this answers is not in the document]
andresmgot: right, my reasoning was for the function to be more readable but it's not 100% secure

PR #2190, opened 2020-11-30
Change scope of uniqChartList from all to a single repo

antgamdia:
### Description of the change

Make uniqChartList function repo-scoped instead of being unscoped. 

Current behavior leads to this scenario:

1. Add bitnami repo (https://charts.bitnami.com/bitnami)
2. Go to bitnami charts -> there are plenty of charts, eg, "bitnami/wordpress"
3. Add jfrog repo (https://repo.chartcenter.io)
4. Go to bitnami charts -> only 4 charts, there is no longer a "bitnami/wordpress"

### Benefits

Adding a new repo with mirrored charts won't prevent the old repo to still have its charts.

### Possible drawbacks

With this PR, the whole Kubeapps catalog may have duplicated charts in different repositories.

### Applicable issues

Following-up the PR https://github.com/kubeapps/kubeapps/pull/2173#discussion_r531535059

### Additional information

The main point is that, from a UX perspective, the sudden disappearance of a number of charts in an old repo when adding a new one could be misleading and frustrating for users.

### DIFF
--- a/cmd/assetsvc/handler.go
+++ b/cmd/assetsvc/handler.go
@@ -100,13 +100,15 @@ func min(a, b int) int {
 
 func uniqChartList(charts []*models.Chart) []*models.Chart {
 	// We will keep track of unique digest:chart to avoid duplicates
-	chartDigests := map[string]bool{}
+	// chartDigests records when we process a specific chart digest for a specific repository.
+	chartDigests := map[string]map[string]bool{}
 	res := []*models.Chart{}
 	for _, c := range charts {
 		digest := c.ChartVersions[0].Digest
 		// Filter out the chart if we've seen the same digest before
-		if _, ok := chartDigests[digest]; !ok {
-			chartDigests[digest] = true
+		if _, ok := chartDigests[digest][c.Repo.Name]; !ok {
+			chartDigests[digest] = map[string]bool{}
+			chartDigests[digest][c.Repo.Name] = true
 			res = append(res, c)
 		}
 	}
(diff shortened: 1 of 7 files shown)


andresmgot (commented):
Hey, so following up with my previous comment. The goal of this function is to actually behave as you describe in the body of the PR, it's supposed to filter duplicated charts over different repositories. The problem is that when filtering by repo, it's the UI the one that does the filtering, not the backend (so `bitnami/*` charts may have been previously filtered).

This feature can be disabled (I thought it was disabled by default but it's not), doing queries with the parameter `showDuplicates=1`. A good change could be to replace the parameter to `removeDuplicates` so the default is the opposite.

In any case, I think this feature is no longer useful, so we can just delete it. It's probably better to just delete this function (`uniqChartList`) and `showDuplicates`.

absoludity:
> Hey, so following up with my previous comment. The goal of this function is to actually behave as you describe in the body of the PR, it's supposed to filter duplicated charts over different repositories.

I assume the original intention to filter duplicate charts over different repositories was so that you wouldn't see two lots of the same chart when browsing the available catalog, *not* that you would see zero lots of the chart when browsing a specific catalog because it's a duplicate in another repo that you're currently not viewing (which is what I think Antonio is pointing out). 

> The problem is that when filtering by repo, it's the UI the one that does the filtering, not the backend (so `bitnami/*` charts may have been previously filtered).

Yes, that's the cause, but the problem is that charts disappear unexpectedly.

> 
> In any case, I think this feature is no longer useful, so we can just delete it. It's probably better to just delete this function (`uniqChartList`) and `showDuplicates`.

+1 .

antgamdia:
Yes, I agree with you. If this feature is no longer useful, let's remove it. 
Pushing right now the change.

andresmgot (approved):
:+1: thanks!

PR #2191, opened 2020-11-30
Filter out dups repos by UID

antgamdia:
### Description of the change

Currently, adding the [bitnami](https://charts.bitnami.com/bitnami) and the [jfrog](https://repo.chartcenter.io) repos, and then trying to upgrade an app, results in the following behavior:
image

And the corresponding error in js console: `Encountered two children with the same key, 'kubeapps/jfrog'. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.` 

I guess it is due to an unintended side effect when concatenating
`AppRepository.list(currentCluster, namespace)` and `AppRepository.list(currentCluster, otherNamespace)` because they are global repos and they appear in both listings.

### Benefits

This PR filters out duplicated repos (duplicated= same uid)

### Possible drawbacks

I don't see any problem or drawback at the moment

### Applicable issues

Following-up the PR https://github.com/kubeapps/kubeapps/pull/2173#discussion_r531663523

### Additional information

N/A

### DIFF
--- a/dashboard/src/actions/repos.test.tsx
+++ b/dashboard/src/actions/repos.test.tsx
@@ -267,10 +267,10 @@ describe("fetchRepos", () => {
     AppRepository.list = jest
       .fn()
       .mockImplementationOnce(() => {
-        return { items: [{ name: "repo1" }] };
+        return { items: [{ name: "repo1", metadata: { uid: "123" } }] };
       })
       .mockImplementationOnce(() => {
-        return { items: [{ name: "repo2" }] };
+        return { items: [{ name: "repo2", metadata: { uid: "321" } }] };
       });
 
     const expectedActions = [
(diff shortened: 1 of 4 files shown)


andresmgot (commented):
Thanks for addressing this! I think it's better if we avoid requesting things twice (rather than doing extra filtering). In `SelectRepoForm`, you just need to avoid requesting repos for both namespaces (current and global) if both are the same.

antgamdia:
Good idea! I've added this check prior to dispatch the `fetchRepos` action. However, I've left the UID-based dup check just in case, do you believe we should also remove it?

dashboard/src/components/SelectRepoForm/SelectRepoForm.tsx
```diff
@@ -34,7 +34,13 @@ function SelectRepoForm({ cluster, namespace, chartName }: ISelectRepoFormProps)
...
 
   useEffect(() => {
-    dispatch(actions.repos.fetchRepos(namespace, kubeappsNamespace));
+    if (namespace !== kubeappsNamespace) {
+      // Normal namespace, show local and global repos
+      dispatch(actions.repos.fetchRepos(namespace, kubeappsNamespace));
+    } else {
+      // Global namespace, show global repos only
+      dispatch(actions.repos.fetchRepos(kubeappsNamespace));
+    }
```
andresmgot, line 43: can you add a test for this as well?

andresmgot (approved):
It's okay to leave the extra filtering. The only thing missing is a test for the fix in SelectRepoForm :+1:

dashboard/src/actions/repos.test.tsx
```diff
@@ -339,6 +339,41 @@ describe("fetchRepos", () => {
...
+      .fn()
+      .mockImplementationOnce(() => {
+        return { items: [{ name: "repo1", metadata: { uid: "123" } }] };
+      })
+      .mockImplementationOnce(() => {
+        // will be skipped
+        return {
+          items: [
+            { name: "repo1", metadata: { uid: "321" } },
+            { name: "repo2", metadata: { uid: "123" } }, // same uid
```
andresmgot, line 354: you mean `// different uid`. In any case, I'd avoid inline comments if they are not super necessary, usually it only makes sense to the one writing them :)

PR #2192, opened 2020-11-30
Fix CSS display for img-container

andresmgot:
### Description of the change

I noticed that some `<img` where not properly rendered when the `src`  is encoded in base64:

Screenshot from 2020-11-30 12-28-59

Change the `display` property fixes the issue (the rest of icons keep working as usual)

Screenshot from 2020-11-30 12-29-15

### DIFF
--- a/dashboard/src/components/PageHeader/PageHeader.scss
+++ b/dashboard/src/components/PageHeader/PageHeader.scss
@@ -6,7 +6,7 @@
   background-color: white;
 
   .img-container {
-    display: flex;
+    display: contents;
     align-items: center;
     img {
       max-height: 3rem;


antgamdia (approved):
Good catch! I'm just wondering about the [current support](https://caniuse.com/css-display-contents) to `display: contents`. We are only using it at the `#login-submit-button`.
Right now, I'm not really aware of the difference between `contents` and other more widespread alternatives like `flex`, so, if you consider it is the right way, go ahead!

andresmgot:
> Good catch! I'm just wondering about the [current support](https://caniuse.com/css-display-contents) to `display: contents`. We are only using it at the `#login-submit-button`.
> Right now, I'm not really aware of the difference between `contents` and other more widespread alternatives like `flex`, so, if you consider it is the right way, go ahead!

Yes, it's properly supported in modern browsers. Not a CSS expert but basically, the `contents` allow to "ignore" the hierarchy of current `div` but still allows to set some properties (like centering the image) and setting boundaries. That's enough to fix the issue (which is setting the div as `flex` with no children `width`).

Issue #2181, opened 2020-11-24
Turn pinniped-proxy POC into production ready service.


absoludity:
In #2154 a proof-of-concept for the `pinniped-proxy` service was developed and tested to ensure that the integration works as expected for our use-case.

Moving forward to make this a production-ready proxy that can be used in Kubeapps requires some additional work:
 
- [x] Proper access logs . Currently we're just exposing logs for errors and debugging. To be useful as a proxy that can be debugged and inspected more generally, we'll want formatted access and error logs.
- [x] Talk to pinniped-concierge API rather than using pinniped CLI. The POC just calls out to a bundled pinniped CLI binary for the credential exchange, but the CLI is designed for single-user use and could take decisions that would not be compatible with a multi-user scenario. We'll need to do the exchange using the pinniped API (I assume).
- [x] Multi-cluster support: The POC was done with multi-cluster support in mind, but it's not yet been implemented and tested.
- [x] When targeting the local cluster (https://kubernetes.default) the proxy should not require passing the cert auth data, since it can be obtained from pod itself. This will make the more common single-cluster case simpler but also better support the multi-cluster case.
- [x] Improved test coverage and modularization.
- [x] Ensure auth failures via the proxy are reported as such (currently a 500)

PR #2193, opened 2020-12-01
Add initial pinniped-proxy command and developer docs.

absoludity:
### Description of the change

Adds an initial `pinniped-proxy` command which does nothing other than using the structured options package to create a CLI with one option for setting the port on which the binary should listen.

### Applicable issues

  - Ref: #2181

### Additional information

As per #2181, I've already done the two major changes which I wanted to make this service production ready. One of the remaining tasks is to refactor and test the code sufficiently, so I'll use the opportunity to do so while proposing the code in small pieces that *hopefully* help to familiarize with Rust and the main libraries I've chosen for the proxy, as well as the tooling.

I'll continue to update the dev doc as I later add the support for building images etc.

More details in-line.


cmd/pinniped-proxy/src/cli.rs
```diff
@@ -0,0 +1,19 @@
+use structopt::StructOpt;
```
absoludity, line 1: So this PR introduces the structured options crate (package). You can read more about it at https://docs.rs/structopt/0.3.21/structopt/ , but think of it as similar to the golang `cobra` module.

cmd/pinniped-proxy/src/cli.rs
```diff
@@ -0,0 +1,19 @@
...
+
+#[derive(StructOpt, Debug)]
```
absoludity, line 3: This is an example of invoking a macro - in this case, we're asking the Rust compiler to automatically derive functionality to be able to print debug output of the struct below, as well as derive functionality for the struct below to be used as the structured options for a command-line CLI.

cmd/pinniped-proxy/src/cli.rs
```diff
@@ -0,0 +1,19 @@
...
+/// A proxy server which converts k8s API server requests with bearer tokens to
+/// requests with short-lived X509 certs exchanged by pinniped.
+///
+/// pinniped-proxy proxies incoming requests with an `Authorization: Bearer
+/// token` header, exchanging the token via the pinniped aggregate API for x509
+/// short-lived client certificates, before forwarding the request onwards
+/// to the destination k8s API server.
```
absoludity, line 10: The triple slash comments are documentation comments. You'll see that the above is reformatted and used in the output of the binary with the `-h` help option.

cmd/pinniped-proxy/src/cli.rs
```diff
@@ -0,0 +1,19 @@
...
+pub struct Options {
+    #[structopt(
+        short = "p",
+        long = "port",
+        default_value = "3333",
+        help = "Specify the port on which pinniped-proxy listens."
+    )]
```
absoludity, line 17: The macro allows the code here to just be a simple struct with a single field, with all the required functionality derived (short-name, default value, help etc.)

cmd/pinniped-proxy/src/main.rs
```diff
@@ -0,0 +1,9 @@
+use structopt::StructOpt;
+
+// Ensure the root crate is aware of the child modules.
+mod cli;
+
+fn main() {
+    let opt = cli::Options::from_args();
+    println!("{:#?}", opt);
```
absoludity, line 8: A simple main function to begin with, just does a debug print (`{:#?}`) of the options. More interesting is checking the help for the command (included in the dev docs below, since it won't change).

If you remove the `Debug` from the derive macro in the `cli` module, you'll find it will no longer compile as this line here requires an implementation for displaying the struct in debug.

andresmgot (approved):
okay, so at least I understood this PR :)

antgamdia (approved):
Wow, thank you for such detailed comments; they are really useful to get a better understanding of each part!

PR #2194, opened 2020-12-01
Add stub proxy server.

absoludity:
### Description of the change

Adds a stub proxy server using the recommended libraries (crates) hyper and tokio.

The main crate introduced in this PR is hyper - "hyper is a fast and correct HTTP implementation written in and for Rust.". Hyper is itself is built on a lower-level crate, tokio: "A runtime for writing reliable network applications without compromising speed.", but it only features via a macro (see inline comments).

Follows on from #2193 

You can run the server with `cargo run` and curl the address to see the response as you'd expect.

### Applicable issues

  - Ref #2181 

### Additional information

Before reading this PR, if you're interested, I'd recommend reading briefly about [Rust's error handling with the `Result` enum](https://doc.rust-lang.org/stable/book/ch09-02-recoverable-errors-with-result.html). Where Golang functions tend to return a tuple with (value, error), where error can be nil, idiomatic rust instead returns a `Result` which can be either the Ok type with your value `Ok(value)` or `Err(e)`. This results in a common pattern for error handling when calling functions of:

```
let my_value = match some_function() {
    Ok(value) => value,
    Err(e) => return Err(e), // pass the error up the stack by returning it here
};
```
which is so common that the somewhat familiar `?` operator can be used, so that the following is equivalent:

```
let my_value = some_function()?;
```
Anyway, more details in the link.

### DIFF
--- a/cmd/pinniped-proxy/Cargo.toml
+++ b/cmd/pinniped-proxy/Cargo.toml
@@ -11,4 +11,8 @@ edition = "2018"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+anyhow = "1.0"
+hyper = "0.13"
 structopt = "0.3"
+# Update to 0.3 once apparent issue with the nascent 0.3.5 is fixed.
+tokio = { version = "0.2", features = ["full"] }
--- a/cmd/pinniped-proxy/src/main.rs
+++ b/cmd/pinniped-proxy/src/main.rs
@@ -1,9 +1,40 @@
+use std::convert::Infallible;
+
+use anyhow::Result;
+use hyper::{Server, service::{make_service_fn, service_fn}};
 use structopt::StructOpt;
 
 // Ensure the root crate is aware of the child modules.
 mod cli;
+mod service;
 
-fn main() {
+#[tokio::main]
+async fn main() -> Result<()> {
     let opt = cli::Options::from_args();
-    println!("{:#?}", opt);
+
+    // For every incoming connection, we make a new hyper `Service` to handle
+    // all incoming HTTP requests on that connection. This is done by passing a
+    // closure to the hyper `make_service_fn` which returns our custom `make_svc`
+    // function that can be used for each connection. 
+    let make_svc = make_service_fn(|_conn| {
+        // The closure just returns an async block that runs a service to handle
+        // all requests for the connection.
+        async {
+            // `service_fn` is a helper from the hyper crate which converts a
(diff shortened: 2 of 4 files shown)


cmd/pinniped-proxy/Cargo.toml
```diff
@@ -11,4 +11,8 @@ edition = "2018"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+anyhow = "1.0"
```
absoludity, line 14: Oh - one other dependency that I started using here. https://docs.rs/anyhow/1.0.34/anyhow/ makes it even easier to handle different types of errors with context etc.. I'll explain a bit more in a later PR.

cmd/pinniped-proxy/src/main.rs
```diff
@@ -1,9 +1,40 @@
...
+use anyhow::Result;
+use hyper::{Server, service::{make_service_fn, service_fn}};
 use structopt::StructOpt;
 
 // Ensure the root crate is aware of the child modules.
 mod cli;
+mod service;
 
-fn main() {
+#[tokio::main]
```
absoludity, line 11: The only place we actually see the lower level `tokio` library is here - a macro that converts the main function into an async function which can be run using `tokio`.

cmd/pinniped-proxy/src/main.rs
```diff
@@ -1,9 +1,40 @@
...
+
+    let server = Server::bind(&addr).serve(make_svc);
+
+    println!("Listening on http://{}", addr);
+
+    // Run the server for ever. If it returns with an error, return the
+    // result, otherwise, if it completes, we return Ok.
+    server.await?;
+
+    Ok(())
```
andresmgot, line 39: what is this `Ok(())` doing? I guess that if everything goes fine, it never executes?

andresmgot (approved):
Thanks for the extra mile explaining everything :+1:

cmd/pinniped-proxy/src/main.rs
absoludity: It's just returning an empty Ok result. If you look back on line 12 at the signature of the `main` function, you'll see it returns a `Result<()>` which means that the function will return one of the two Result enums - either Ok (with a matching empty type of `()`) or an error. It is much clearer when the result has a value, such as `Ok(some_number)`, but a main function either succeeds or it errors. Different types of errors may be returned, but if the main function succeeds, there's no value associated with it.

PR #2195, opened 2020-12-01
Add logging helper, demo tests and traits.

absoludity:
### Description of the change

This PR follows #2194 and introduces the logging library and demonstrates both tests and traits.

The main library used here is [pretty_env_logger](https://crates.io/crates/pretty_env_logger) which is a colourized version of a more basic env logger, which just allows setting the log level via the environment (`RUST_LOG=info`, for eg.). This provides macros to emit logs at each level.

Running the stub proxy now and hitting it with a few curls shows, first with `RUST_LOG=info`:
```
$ RUST_LOG=info cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.09s
     Running `target/debug/pinniped-proxy`
 INFO  pinniped_proxy::service > GET /some/url/other 200 OK
 INFO  pinniped_proxy::service > PUT /some/url/other 200 OK
```

while repeating this with `RUST_LOG=debug`:

```
$ RUST_LOG=debug cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.07s
     Running `target/debug/pinniped-proxy`
 DEBUG hyper::proto::h1::io > read 92 bytes 
 DEBUG hyper::proto::h1::io > parsed 3 headers 
 DEBUG hyper::proto::h1::conn > incoming body is empty 
 INFO  pinniped_proxy::service > GET /some/url/other 200 OK
 DEBUG hyper::proto::h1::io    > flushed 96 bytes 
 DEBUG hyper::proto::h1::io    > read 0 bytes 
 DEBUG hyper::proto::h1::conn  > read eof 
 DEBUG hyper::proto::h1::io    > read 92 bytes 
 DEBUG hyper::proto::h1::io    > parsed 3 headers 
 DEBUG hyper::proto::h1::conn  > incoming body is empty 
 INFO  pinniped_proxy::service > PUT /some/url/other 200 OK
 DEBUG hyper::proto::h1::io    > flushed 96 bytes 
 DEBUG hyper::proto::h1::io    > read 0 bytes 
 DEBUG hyper::proto::h1::conn  > read eof 
```

### Applicable issues

  - Ref #2181 

### Additional information

There are a couple of other features of the language which I've introduced here: traits and tests. The tests are pretty self explanatory - I'll leave a few inline comments below. Traits are a little like interfaces in golang: they define functionality that other objects can choose to implement. Most of the common trait functionality can be derived automatically, but I've explicitly implemented the Display trait for the LogData struct, so that it knows how to display a basic log line. More details inline. 

I'm trying to do these PRs so that they are actually a good learning introduction to rust. Let me know if it's useful. I'd recommend reading the section of the rust book on [Understanding Ownership](https://doc.rust-lang.org/stable/book/ch04-00-understanding-ownership.html) with this PR.


cmd/pinniped-proxy/src/logging.rs
```diff
@@ -0,0 +1,108 @@
+use std::cmp::PartialEq;
+use std::fmt;
+
+use hyper::{Body, Request, Response};
+
+#[derive(Clone, Debug, Default, PartialEq)]
```
absoludity, line 6: So here we're allowing four traits to be derived automatically:
* Clone enables an `object.clone()` method which, well, returns a clone of the struct. This becomes important in rust where the compiler helps you to be very explicit about ownership of variables as they get passed around. Cloning a var gives you a way to ensure the function you're passing the var into can own its own copy.
* Debug enables formatting the struct into a string using the `{:?}` format (e.g.: `info!("log data is: {:?}", log_data)`)
* Default enables the creation of a struct with sensible defaults (or you can implement it explicitly for different defaults)
* PartialEq enables equality testing of two instances of the struct (again, you can implement it explicitly for custom comparisons)

cmd/pinniped-proxy/src/logging.rs
```diff
@@ -0,0 +1,108 @@
...
+        };
+        write!(f, "{} {} {}", self.method, self.target_uri, status)
+    }
+}
+
+/// request_log_data returns a new LogData with the relevant data from the request. 
+pub fn request_log_data(req: &Request<Body>) -> LogData {
+    LogData {
+        target_uri: req.uri().clone(),
+        method: req.method().clone(),
```
absoludity, line 36: Here's an example of cloning so that the LogData that we're returning does not refer to memory that is a part of the `req` reference - instead, we've created new copies of the data we need so that LogData is independent. If I don't clone them, the compiler tells us that we've implicitly moved the ownership of the uri from the `req` reference to the LogData struct, which violates concurrency principles and would stop the compiler from being able to determine when it can free the memory:
```
error[E0507]: cannot move out of a shared reference
  --> src/logging.rs:35:21
   |
35 |         target_uri: *req.uri(),
   |                     ^^^^^^^^^^ move occurs because value has type `Uri`, which does not implement the `Copy` trait

error: aborting due to previous error
```

cmd/pinniped-proxy/src/logging.rs
```diff
@@ -0,0 +1,108 @@
...
+
+/// response_log_data returns a new LogData including the relevant data from the response.
+pub fn response_log_data(resp: &Response<Body>, log_data: LogData) -> LogData {
+    let mut updated_log_data = log_data.clone();
+    updated_log_data.status_code = Some(resp.status());
+    updated_log_data
+}
+
+#[cfg(test)]
+mod tests {
```
absoludity, line 49: Tests in rust modules are included as a sub module within the same file. With this change, running `cargo test` results in:

```
$ cargo test
   Compiling pinniped-proxy v0.1.0 (/home/michael/dev/kubeapps/cmd/pinniped-proxy)
    Finished test [unoptimized + debuginfo] target(s) in 2.17s
     Running target/debug/deps/pinniped_proxy-6b227f04e120477c

running 4 tests
test logging::tests::display_including_status ... ok
test logging::tests::display_without_status ... ok
test logging::tests::populate_from_request ... ok
test logging::tests::populate_from_response ... ok

test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

cmd/pinniped-proxy/src/service.rs
```diff
@@ -1,7 +1,17 @@
...
-    Ok(Response::new(Body::from("pinniped-proxy stub\n")))
+use crate::logging;
+
+pub async fn proxy(req: Request<Body>) -> Result<Response<Body>, Infallible> {
+    let log_data = logging::request_log_data(&req);
+
+    // TODO: actual proxying to happen here.
+    let response = Response::new(Body::from("pinniped-proxy stub\n"));
+
+    info!("{}", logging::response_log_data(&response, log_data));
```
absoludity, line 14: Small update to our proxy stub function so that it now records the data for logging before and after the (soon-to-be) actual proxying of the request, and includes the log at the info level.

cmd/pinniped-proxy/src/logging.rs
antgamdia: > Debug enables formatting the struct into a string using the {:?} format (e.g.: info!("log data is: {:?}", log_data))

I don't know if I'm getting it: you are enabling the `Debug` trait, so you are letting the compiler implement the, let's say, toString() of this struct. However, you aren't using it in this file, aren't you?
I guess it is intended for other call-sites when you may need to log the value of a log_data object.

So, I suppose that when implementing a new struct, we always should add certain traits. Isn't there any way to ask the compiler to do so with every struct in the project? Don't know, just asking

cmd/pinniped-proxy/src/logging.rs
antgamdia: Is this approach typical in Rust? I mean, why not having the tests in a separate file?

antgamdia (approved):
Thank you, again, for all the useful and instructive comments! I really appreciate them!

cmd/pinniped-proxy/src/logging.rs
andresmgot: I believe it's used in this file. Clone in line 35, debug in any print (like in line 28) and default and PartialEq in the tests. What I don't know is how you map these `derive` to the functions used afterwards.

andresmgot (approved):

cmd/pinniped-proxy/src/logging.rs
absoludity: > I believe it's used in this file. Clone in line 35, debug in any print (like in line 28) and default and PartialEq in the tests.

Yep, that's correct. The debug output formatter is the `{:?}` or `{:#?}`.

> What I don't know is how you map these `derive` to the functions used afterwards.

The compiler just adds implementations to the struct for them automatically. You can implement them manually (and have to for some, such as the Display trait, which can't be derived automatically since it depends or your own needs). More info at https://doc.rust-lang.org/book/ch10-02-traits.html

> So, I suppose that when implementing a new struct, we always should add certain traits. Isn't there any way to ask the compiler to do so with every struct in the project? Don't know, just asking

Others have asked too (especially about the debug trait, which seems obvious): https://github.com/rust-lang/prev.rust-lang.org/issues/1073

cmd/pinniped-proxy/src/logging.rs
absoludity: One of the benefits of having it as a sub module is that you can `use super::*` to have access to all the private types etc. So it appears that unit tests always go in the same file - https://doc.rust-lang.org/book/ch11-03-test-organization.html#the-tests-module-and-cfgtest . Integration tests go in a `tests` directory: https://doc.rust-lang.org/book/ch11-03-test-organization.html#integration-tests

PR #2198, opened 2020-12-02
Add https module with code to extract target server URL from header.

absoludity:
### Description of the change

Follows #2195

This PR adds an https module with functionality to extract and validate the k8s api server url from the HTTP headers. No new concepts I don't think, but shows some good examples of using `match` and testing.

### Applicable issues

  - Ref #2181 

### Additional information

Might be a good spot, while reviewing, to read about [Enums and Pattern Matching](https://doc.rust-lang.org/book/ch06-00-enums.html)

### DIFF
--- a/cmd/pinniped-proxy/Cargo.toml
+++ b/cmd/pinniped-proxy/Cargo.toml
@@ -18,3 +18,4 @@ pretty_env_logger = "0.4"
 structopt = "0.3"
 # Update to 0.3 once apparent issue with the nascent 0.3.5 is fixed.
 tokio = { version = "0.2", features = ["full"] }
+url = "2.1"
--- a/cmd/pinniped-proxy/src/https.rs
+++ b/cmd/pinniped-proxy/src/https.rs
@@ -1,0 +1,124 @@
+use anyhow::Result;
+use hyper::{HeaderMap, header::HeaderValue};
+use url::Url;
+
+const DEFAULT_K8S_API_SERVER_URL: &str = "https://kubernetes.local";
+const HEADER_K8S_API_SERVER_URL: &str = "PINNIPED_PROXY_API_SERVER_URL";
+const INVALID_SCHEME_ERROR: &'static str = "invalid scheme, https required";
+
+/// validate_url returns a result containing the validated url or an error if it is invalid.
+fn validate_url(u: String) -> Result<String> {
+    let result = Url::parse(&u);
+    match result {
+        Ok(url) => match url.scheme() {
+            "https" => Ok(u),
+            _ => Err(anyhow::anyhow!(INVALID_SCHEME_ERROR)),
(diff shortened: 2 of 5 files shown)


cmd/pinniped-proxy/Cargo.toml
```diff
@@ -17,4 +17,5 @@ log = "0.4"
 pretty_env_logger = "0.4"
 structopt = "0.3"
 # Update to 0.3 once apparent issue with the nascent 0.3.5 is fixed.
-tokio = { version = "0.2", features = ["full"] }
\ No newline at end of file
+tokio = { version = "0.2", features = ["full"] }
```
antgamdia, line 20: So, apparently, you can decide which compilation options you want for a given package. Interesting!

cmd/pinniped-proxy/src/https.rs
```diff
@@ -0,0 +1,124 @@
...
+const HEADER_K8S_API_SERVER_URL: &str = "PINNIPED_PROXY_API_SERVER_URL";
+const INVALID_SCHEME_ERROR: &'static str = "invalid scheme, https required";
+
+/// validate_url returns a result containing the validated url or an error if it is invalid.
+fn validate_url(u: String) -> Result<String> {
+    let result = Url::parse(&u);
+    match result {
+        Ok(url) => match url.scheme() {
+            "https" => Ok(u),
+            _ => Err(anyhow::anyhow!(INVALID_SCHEME_ERROR)),
```
antgamdia, line 15: What's the meaning of this underscore?

antgamdia (approved):

cmd/pinniped-proxy/src/https.rs
antgamdia: You mentioned it in another PR:

> The _ is just to avoid warnings about unused variables (I'll use the in later PRs), similar to golang (though go forces you rather than warns).

Thanks!

cmd/pinniped-proxy/Cargo.toml
```diff
@@ -17,4 +17,5 @@ log = "0.4"
...
+url = "2.1"
```
andresmgot, line 21: you are not adding a blank line at the end for any reason?

andresmgot (approved):

cmd/pinniped-proxy/Cargo.toml
absoludity: Yeah, and can therefore save compile time depending on what you need.

cmd/pinniped-proxy/Cargo.toml
absoludity: Yeah, I was working on a different computer and didn't have my editor there setup to do it automatically. I did notice and fix it in the following one, I think.

PR #2199, opened 2020-12-02
Add functions to extract cert auth data from headers.

absoludity:
### Description of the change

Follows #2198 and adds functions to extract the api server CA cert from the request headers.

I'll follow this up with a change that doesn't require the CA cert in the request headers when targeting the default cluster api server (https://kubernetes.local), as per #2181 .

### Applicable issues

  - fRef: #2181 

### Additional information

Two other changes here:
* adding the `handle_error` function in the service to handle errors with an appropriate HTTP response and use this to handle errors for the 3 `https::` functions so far.
* improve the logging, you can run with `RUST_LOG=debug` and curl the service to see details such as:

```
 DEBUG hyper::proto::h1::io > read 122 bytes 
 DEBUG hyper::proto::h1::io > parsed 4 headers 
 DEBUG hyper::proto::h1::conn > incoming body is empty 
 DEBUG pinniped_proxy::https  > no PINNIPED_PROXY_API_SERVER_URL header present, defaulting to https://kubernetes.local
 DEBUG pinniped_proxy::https  > failed to base64 decode PINNIPED_PROXY_API_SERVER_CERT header data: Invalid byte 32, offset 3.
 INFO  pinniped_proxy::service > GET / 400 Bad Request
 DEBUG hyper::proto::h1::io    > flushed 111 bytes 
 DEBUG hyper::proto::h1::io    > read 0 bytes 
 DEBUG hyper::proto::h1::conn  > read eof 
```

### DIFF
--- a/cmd/pinniped-proxy/src/service.rs
+++ b/cmd/pinniped-proxy/src/service.rs
@@ -11,6 +11,7 @@ pub async fn proxy(req: Request<Body>) -> Result<Response<Body>, Infallible> {
 
     // TODO: actual proxying to happen here.
     let _ = https::get_api_server_url(req.headers());
+    let _ = https::get_api_server_cert_auth_data(req.headers());
     let response = Response::new(Body::from("pinniped-proxy stub\n"));
 
     info!("{}", logging::response_log_data(&response, log_data));
--- a/cmd/pinniped-proxy/src/service.rs
+++ b/cmd/pinniped-proxy/src/service.rs
@@ -32,7 +32,9 @@ pub async fn proxy(req: Request<Body>) -> Result<Response<Body>, Infallible> {
     Ok(response)
 }
 
-/// handle_error converts an error into a 500 message.
+/// handle_error converts an error into a BAD_REQUEST response.
+/// 
+/// We may need to expand this to give different responses for different errors.
 fn handle_error(e: Error, log_data: logging::LogData) -> Result<Response<Body>, Infallible> {
     let response = Response::builder()
         .status(StatusCode::BAD_REQUEST)
--- a/cmd/pinniped-proxy/src/service.rs
+++ b/cmd/pinniped-proxy/src/service.rs
@@ -1,6 +1,7 @@
 use std::convert::Infallible;
 
-use hyper::{Body, Request, Response};
+use anyhow::Error;
+use hyper::{Body, Request, Response, StatusCode};
 use log::info;
 
 use crate::logging;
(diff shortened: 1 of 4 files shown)


cmd/pinniped-proxy/src/service.rs
```diff
@@ -9,11 +10,34 @@ use crate::https;
 pub async fn proxy(req: Request<Body>) -> Result<Response<Body>, Infallible> {
     let log_data = logging::request_log_data(&req);
 
+    let _ = match https::get_api_server_url(req.headers()) {
```
absoludity, line 13: The `_` is just to avoid warnings about unused variables (I'll use the in later PRs), similar to golang (though go forces you rather than warns).

antgamdia (approved):
Great! Thank you for splitting this PR into smaller ones. It is being quite interesting, like a Netflix series, but in Rust: looking forward to seeing the next episode :P

absoludity:
> Great! Thank you for splitting this PR into smaller ones. It is being quite interesting, like a Netflix series, but in Rust: looking forward to seeing the next episode :P

Heh - great. FWIW, I'm not just splitting the larger PR into smaller ones, but actually fixing things and doing the remaining work as I go (not sure it'd be worth the time just to split them up).

cmd/pinniped-proxy/src/service.rs
```diff
@@ -9,11 +10,36 @@ use crate::https;
...
+        Ok(u) => u,
+        Err(e) => return handle_error(e, log_data),
+    };
+    // TODO: don't call this if we're using https://kubernetes.local, instead
+    // grab the data from the file system.
+    let cert_auth_data = match https::get_api_server_cert_auth_data(req.headers()) {
```
andresmgot, line 19: who was setting this auth data? where is the cert coming from?

andresmgot (approved):
Thanks!

cmd/pinniped-proxy/src/service.rs
absoludity: In our specific case, the Kubeapps frontend, which has the multi-cluster config (collection of api server urls together with the cert auth data for each, where required), adds these headers when forwarding the request to pinniped-proxy. This allows pinniped-proxy to receive the request, exchange the credential, then forward the request on to the api server (with the cert auth) together with the client auth certs.

andresmgot: I see, but this is not implemented yet, no? (or I am missing something?)

absoludity: Correct - Currently the service is just logging requests - I'll add the proxying and credential exchange next.

PR #2200, opened 2020-12-02
Increase reliability of e2e tests

### LINKED ISSUE #2176: CI flakiness
There are many spurious errors in the CI that slows us down.

andresmgot:
### Description of the change

The goal of this PR is to avoid, as much as possible, transient errors caused by flaky tests. The biggest issue is that these are not reproducible locally and only happens sometimes in the CI.

Most of this PR is adding retries/bigger timeouts (I didn't have the time to debug each of the issues).

The most troublesome test is the operator one. The OLM, apparently arbitrary misbehaves, so almost each of the actions is susceptible to fail. It will still fail from time to time after this PR (1/10 builds fails when running my tests).

In any case, the reliability should be increased with this PR since prior to this ~6/10 builds failed.

### Applicable issues

  - fixes #2176 

### Additional information

Feel free to merge this PR if you want while I'm OOO.

I'll add some more comments inline.

FTR: I took notes of the errors I keep finding (as you can see many different): https://docs.google.com/spreadsheets/d/1cZpm8aieunGuFKwNWshQTwJJ9JsbaOahRmD1ud-fCRE/edit#gid=0

10/10 builds working: https://app.circleci.com/pipelines/github/kubeapps/kubeapps/1494/workflows/c707011a-20d5-43b6-80ba-7ee0158dc973

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,8 +40,8 @@ workflows:
           matrix:
             parameters:
               # Enable the line below to do bundle builds to debug CI issues
-              # number: ["1"]
-              number: ["1", "2", "3", "4", "5", "6", "7", "8", "9", "10"]
+              number: ["1"]
+              # number: ["1", "2", "3", "4", "5", "6", "7", "8", "9", "10"]
           requires:
             - test_go
             - test_dashboard
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,4 @@
-version: 2
+version: 2.1
 
 ## Build conditions
 # Build in any branch or tag
@@ -37,6 +37,11 @@ workflows:
           <<: *build_always
       - local_e2e_tests:
           <<: *build_always
+          matrix:
+            parameters:
+              # Enable the line below to do bundle builds to debug CI issues
+              # number: ["1"]
+              number: ["1", "2", "3", "4", "5", "6", "7", "8", "9", "10"]
           requires:
             - test_go
             - test_dashboard
@@ -127,7 +132,7 @@ build_images: &build_images
     - <<: *exports
     - run: |
         mkdir -p images/
-        read -ra IMG_ARRAY <<< "$IMAGES"
+        read -ra IMG_ARRAY \<<< "$IMAGES"
         if [[ -n "${CIRCLE_TAG}" ]]; then
           makeArgs="VERSION=${CIRCLE_TAG}"
         fi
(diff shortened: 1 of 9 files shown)


.circleci/config.yml
```diff
@@ -218,7 +223,14 @@ local_e2e_steps: &local_e2e_steps
...
         chmod +x kind-Linux-amd64
         sudo mv kind-Linux-amd64 /usr/local/bin/kind
-    - run: kind create cluster
+    - run: |
+        if kind create cluster; then
+          echo "Cluster created"
+        else
+          echo "Cluster creation failed, retrying"
+          kind delete cluster || true
+          kind create cluster
```
andresmgot, line 232: `kind` fails sometimes to create the cluster.

integration/manifests/executor.yaml
```diff
@@ -24,5 +24,7 @@ spec:
         - /dev/null
         image: kubeapps/integration-tests:latest
         name: integration
-        resources: {}
+        resources:
+          requests:
+            memory: 100Mi
```
andresmgot, line 29: this avoid `core-dumped` errors when running the browser tests in the executor pod.

integration/use-cases/lib/utils.js
```diff
@@ -1,11 +1,25 @@
...
     while (retriesLeft > 0) {
       try {
         await toCheck();
         break;
       } catch (e) {
+        if(testName) {
+          await page.screenshot({
+            path: path.join(__dirname, `../../${screenshotsFolder}/${testName}-${retries - retriesLeft}.png`),
+          });
+        }
```
andresmgot, line 18: this is mostly for debugging a specific step with several retries. I have removed the places in which I used it (to avoid polluting results with other screenshots) but left this code in case we need it in the future.

script/e2e-test.sh
```diff
@@ -141,8 +141,11 @@ installOrUpgradeKubeapps() {
     helm upgrade --install kubeapps-ci --namespace kubeapps "${chartSource}" \
       ${invalidateCacheFlag} \
       "${img_flags[@]}" \
-      --set featureFlags.ui=clarity \
-      --set featureFlags.operators=true
+      --set frontend.replicaCount=1 \
+      --set kubeops.replicaCount=1 \
+      --set assetsvc.replicaCount=1 \
+      --set dashboard.replicaCount=1 \
+      --set postgresql.replication.enabled=false
```
andresmgot, line 148: did this to reduce resource consumption

absoludity (approved):
Thanks Andres. I'm very keen to land this so that we can keep moving, but also very keen that we not consider this finished until we are able to avoid sprinkling sleep statements throughout - that pattern will only multiply and lead to longer and longer CI with less and less knowledge of why things are failing.

Thanks also for documenting all the issues you were seeing. Great that we can trigger a bundle of tests for this.

PR #2201, opened 2020-12-03
Add an FAQ for configuring global private repositories.

absoludity:
### Description of the change

The question of why we only support imagePullSecrets for namespaced
repositories comes up often enough to answer with a FAQ.

Rendered at https://github.com/absoludity/kubeapps/blob/faq-global-private-repositories/chart/kubeapps/README.md#why-cant-i-configure-global-private-repositories


antgamdia (approved):
Thank you for updating the FAQ: It's great to see this section is becoming useful.

Unrelated strictly to this PR, but I was wondering if we should include a link to the FAQ section directly into the main Readme (not the chart readme) so that people looking for kubeapps directly in Github are able to promptly find this section.

absoludity:
> Thank you for updating the FAQ: It's great to see this section is becoming useful.
> 
> Unrelated strictly to this PR, but I was wondering if we should include a link to the FAQ section directly into the main Readme (not the chart readme) so that people looking for kubeapps directly in Github are able to promptly find this section.

Yep, good idea. Done.

PR #2202, opened 2020-12-03
Add actual proxying (without credential exchange yet)

absoludity:
### Description of the change

Adds code to proxy the actual request, even though the credential exchange isn't yet implemented.

Follows #2199

### Applicable issues

  - ref #2181

### Additional information

With this PR you can proxy to your running local k8s api server, all that's missing is the credential exchange.

```
$ curl http://127.0.0.1:3333/api/v1 -H "PINNIPED_PROXY_API_SERVER_URL: https://127.0.0.1:33613" -H "PINNIPED_PROXY_API_SERVER_CERT: $CA_DATA"
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {
    
  },
  "status": "Failure",
  "message": "forbidden: User \"system:anonymous\" cannot get path \"/api/v1\"",
  "reason": "Forbidden",
  "details": {
    
  },
  "code": 403
}
```

with the logs showing:

```
$ RUST_LOG=info cargo run
   Compiling pinniped-proxy v0.1.0 (/home/michael/dev/kubeapps/cmd/pinniped-proxy)
    Finished dev [unoptimized + debuginfo] target(s) in 26.50s
     Running `target/debug/pinniped-proxy`
 INFO  pinniped_proxy::service > GET https://127.0.0.1:33613/api/v1 403 Forbidden
```

### DIFF
--- a/cmd/pinniped-proxy/src/https.rs
+++ b/cmd/pinniped-proxy/src/https.rs
@@ -83,7 +83,10 @@ pub fn cert_for_cert_data(cert_data: Vec<u8>) -> Result<Certificate> {
 /// rewrite_request directs the specified request to the targeted backend api
 /// server.
 pub fn rewrite_request(mut req: Request<hyper::Body>, k8s_api_server_url: String) -> Result<Request<hyper::Body>> {
-    // Update the request URI.
+    // Update the request URI from
+    // http://pinniped-proxy:1234/api/v1/foo?bar
+    // to
+    // https://k8s-api-server-url:5678/api/v1/foo?bar
     let uri_string = format!(
         "{}{}",
         k8s_api_server_url,
--- a/cmd/pinniped-proxy/src/https.rs
+++ b/cmd/pinniped-proxy/src/https.rs
@@ -1,7 +1,13 @@
 use anyhow::Result;
-use hyper::{HeaderMap, header::HeaderValue};
+use hyper::{
+    client::connect::HttpConnector,
+    Client,
+    HeaderMap,
+    header::HeaderValue, Request,
+};
+use hyper_tls::HttpsConnector;
 use log::debug;
-use native_tls::Certificate;
+use native_tls::{Certificate, TlsConnectorBuilder};
 use url::Url;
 
 const DEFAULT_K8S_API_SERVER_URL: &str = "https://kubernetes.local";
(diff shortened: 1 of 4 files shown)


cmd/pinniped-proxy/src/https.rs
```diff
@@ -73,11 +80,56 @@ pub fn cert_for_cert_data(cert_data: Vec<u8>) -> Result<Certificate> {
...
 }
 
+/// rewrite_request directs the specified request to the targeted backend api
+/// server.
+pub fn rewrite_request(mut req: Request<hyper::Body>, k8s_api_server_url: String) -> Result<Request<hyper::Body>> {
+    // Update the request URI.
+    let uri_string = format!(
+        "{}{}",
+        k8s_api_server_url,
+        req.uri().path_and_query().map(|x| x.as_str()).unwrap_or("")
```
antgamdia, line 90: I guess this part is for concatenating `k8s_api_server_url` to the rest of the request uri, but,just for the sake of clarity, would it be a good idea to add an example as a comment so that we can get a better idea of the big picture?

cmd/pinniped-proxy/src/https.rs
```diff
@@ -73,11 +80,56 @@ pub fn cert_for_cert_data(cert_data: Vec<u8>) -> Result<Certificate> {
...
+    );
+    let uri: hyper::Uri = uri_string.parse()?;
+    *req.uri_mut() = uri.clone();
+
+    // Update the host header.
```
antgamdia, line 95: I'm not sure, but, since we are proxing here, should we add the [`X-Forwarded-xxxx` ](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded) headers also?

antgamdia (approved):

cmd/pinniped-proxy/src/https.rs
absoludity: Yep, will do.

cmd/pinniped-proxy/src/https.rs
absoludity: I had done this initially, then removed as in this case, there is no info being lost during the proxying since the request comes with the target destination in the headers (ie. it's not a reverse proxy hiding backend services, but an authenticating proxy for known backend services). I'll have another think about it, let me know if you've more thoughts.

PR #2206, opened 2020-12-07
Add exchange of credentials via pinniped-concierge.

absoludity:
### Description of the change

Adds the pinniped module with the functionality to exchange credentials via the pinniped-concierge API.
This is then called within the proxy service to update the tls connection to use the client identity.

### Benefits

With this we can then curl the proxy with an api request to see the result:

```
curl -H "PINNIPED_PROXY_API_SERVER_URL: https://127.0.0.1:41543" -H "PINNIPED_PROXY_API_SERVER_CERT: ${CACERT_DATA}" http://localhost:3333/api/v1 -H "Authorization: Bearer pinny-the-seal:password123"
{
  "kind": "APIResourceList",
  "groupVersion": "v1",
  "resources": [
  ...
```

### Applicable issues

  - ref #2181

### Additional information

I've not added tests for the pinniped module's main functionality. I need to look into testing network functionality.


cmd/pinniped-proxy/src/pinniped.rs
```diff
@@ -0,0 +1,141 @@
...
+
+    let auth_token = match authorization.to_string().strip_prefix("Bearer ") {
+        Some(a) => a.to_string(),
+        None => authorization.to_string(),
+    };
+    let token_creds: Api<TokenCredentialRequest> = Api::namespaced(client.clone(), &pinniped_namespace);
+    let mut cred_request = TokenCredentialRequest::new("", TokenCredentialRequestSpec {
+        token: Some(auth_token),
+        authenticator: corev1::TypedLocalObjectReference {
+            name: env::var(PINNIPED_AUTHENTICATOR_NAME).with_context(|| format!("error retrieving {}", PINNIPED_AUTHENTICATOR_NAME))?,
```
absoludity, line 125: `with_context` is a helper of the anyhow error library that enables changed errors, a little like golang's newer errors. So for example, if I run with the incorrect authenticator type var, I'll see:

```
 ERROR pinniped_proxy::service > Failed to exchange credentials

Caused by:
    0: error retrieving PINNIPED_AUTHENTICATOR_TYPE
    1: environment variable not found
```

cmd/pinniped-proxy/Cargo.toml
```diff
@@ -15,9 +15,16 @@ anyhow = "1.0"
 base64 = "0.13"
 hyper = "0.13"
 hyper-tls = "0.4"
+kube = { version = "0.43.0", features = ["derive"] }
+kube-runtime = "0.43.0"
+k8s-openapi = { version = "0.9.0", default-features = false, features = ["v1_17"] }
```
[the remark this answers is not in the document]
antgamdia: Which are the implications of enabling `features = ["v1_17"]`?  Does it accept a range of k8s versions?

antgamdia (approved):

cmd/pinniped-proxy/Cargo.toml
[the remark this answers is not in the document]
absoludity: I assume it's just the feature set that is enabled ie., in our case, it'll be assuming the 1.17 API. I don't see that it will affect us much (a little like using the kubectl 1.17 cli) especially since we're mainly using the lib for CRDs.

cmd/pinniped-proxy/src/https.rs
```diff
@@ -33,6 +35,22 @@ fn validate_url(u: String) -> Result<String> {
     }
 }
 
+/// include_client_cert updates a tls connection to be built with a client cert for authentication.
+/// 
+/// The client cert is obtained by exchanging the authorization token for a client identity via
+/// pinniped.
+pub async fn include_client_identity_for_headers<'a>(mut tls_builder: &'a mut TlsConnectorBuilder, request_headers: HeaderMap<HeaderValue>, k8s_api_server_url: &str, k8s_api_ca_cert_data: &[u8]) -> Result<&'a mut TlsConnectorBuilder> {
```
andresmgot, line 42: :astonished: this syntax is crazy

absoludity: Yeah, it's pretty unfamiliar at first, but necessary if you want to want the compiler to be able to determine the lifetimes of variables. It's usually not required to mention it explicitly, but is required here because there are two arguments being passed by reference, so we clearly identify that it's `tls_builder` which is required to have the same lifetime as the return value (because the return value references it). The book has a good overview: https://doc.rust-lang.org/book/ch10-03-lifetime-syntax.html

PR #2209, opened 2020-12-08
Load default cert authority data for kubernetes.default

absoludity:
### Description of the change

Follows #2206 and updates to load the default cert authority data for use when the request doesn't include this *and* the url is for the https://kubernetes.default url.

### Benefits

Less configuration for the single-cluster use-case.

### Applicable issues

  - ref #2181 - one of the items identified there.

### Additional information

See inline comments.

### DIFF
--- a/cmd/pinniped-proxy/src/cli.rs
+++ b/cmd/pinniped-proxy/src/cli.rs
@@ -12,8 +12,16 @@ pub struct Options {
     #[structopt(
         short = "p",
         long = "port",
+        env = "PINNIPED_PROXY_PORT",
         default_value = "3333",
         help = "Specify the port on which pinniped-proxy listens."
     )]
-    pub port: u16, 
+    pub port: u16,
+    #[structopt(
+        long = "default-ca-cert",
+        env = "PINNIPED_PROXY_DEFAULT_CA_CERT",
+        default_value = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt",
+        help = "Specify the file path to the cert authority for the default api server https://kubernetes.default"
+    )]
+    pub default_ca_cert: String,
 }
--- a/cmd/pinniped-proxy/src/https.rs
+++ b/cmd/pinniped-proxy/src/https.rs
@@ -12,9 +12,9 @@ use url::Url;
 
 use crate::pinniped;
 
-const DEFAULT_K8S_API_SERVER_URL: &str = "https://kubernetes.local";
+pub const DEFAULT_K8S_API_SERVER_URL: &str = "https://kubernetes.local";
 const HEADER_K8S_API_SERVER_URL: &str = "PINNIPED_PROXY_API_SERVER_URL";
-const HEADER_K8S_API_SERVER_CA_CERT: &str = "PINNIPED_PROXY_API_SERVER_CERT";
+pub const HEADER_K8S_API_SERVER_CA_CERT: &str = "PINNIPED_PROXY_API_SERVER_CERT";
 const INVALID_SCHEME_ERROR: &'static str = "invalid scheme, https required";
 
 /// validate_url returns a result containing the validated url or an error if it is invalid.
@@ -36,7 +36,7 @@ fn validate_url(u: String) -> Result<String> {
 }
 
 /// include_client_cert updates a tls connection to be built with a client cert for authentication.
-/// 
+///
 /// The client cert is obtained by exchanging the authorization token for a client identity via
 /// pinniped.
 pub async fn include_client_identity_for_headers<'a>(mut tls_builder: &'a mut TlsConnectorBuilder, request_headers: HeaderMap<HeaderValue>, k8s_api_server_url: &str, k8s_api_ca_cert_data: &[u8]) -> Result<&'a mut TlsConnectorBuilder> {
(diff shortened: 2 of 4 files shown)


cmd/pinniped-proxy/src/cli.rs
```diff
@@ -12,8 +12,16 @@ pub struct Options {
     #[structopt(
         short = "p",
         long = "port",
+        env = "PINNIPED_PROXY_PORT",
```
absoludity, line 15: Nice, I hadn't realised that the `structopt` crate also enabled env vars as alternatives for options.

cmd/pinniped-proxy/src/cli.rs
```diff
@@ -12,8 +12,16 @@ pub struct Options {
...
         default_value = "3333",
         help = "Specify the port on which pinniped-proxy listens."
     )]
-    pub port: u16, 
+    pub port: u16,
+    #[structopt(
+        long = "default-ca-cert",
+        env = "PINNIPED_PROXY_DEFAULT_CA_CERT",
+        default_value = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt",
```
absoludity, line 23: Using the normal location by default, but enabling configuration for local use (or custom k8s setups).

cmd/pinniped-proxy/src/https.rs
```diff
@@ -71,18 +71,12 @@ pub fn get_api_server_url(request_headers: &HeaderMap<HeaderValue>) -> Result<St
...
-            Ok(data) => Ok(data),
-            Err(e) => {
-                debug!("failed to base64 decode {} header data: {}", HEADER_K8S_API_SERVER_CA_CERT, e);
-                Err(anyhow::anyhow!(e))
-            }
-        },
-        None => {
-            debug!("header {} required but not present", HEADER_K8S_API_SERVER_CA_CERT);
-            Err(anyhow::anyhow!("header {} required but not present", HEADER_K8S_API_SERVER_CA_CERT))
+pub fn get_api_server_cert_auth_data(cacert_header: &HeaderValue) -> Result<Vec<u8>> {
```
absoludity, line 74: Just updating this function to take the actual header value as an arg instead of the headers, which removes the need to error if it's not present in the headers (moved that out to the service).

cmd/pinniped-proxy/src/main.rs
```diff
@@ -17,17 +18,26 @@ async fn main() -> Result<()> {
     pretty_env_logger::init();
     let opt = cli::Options::from_args();
 
+    // Load the default certificate authority data on startup once.
+    let default_ca_data = fs::read_to_string(opt.default_ca_cert.clone())
+        .with_context(|| format!("error loading default-ca-cert at {}", opt.default_ca_cert))?;
```
absoludity: `.with_context()` is a feature of the `anyhow` error library which allows adding some context to a result before using the `?` operator, so that the error on the cmdline looks like:
```
Error: unable to load default-ca-cert at /var/run/secrets/kubernetes.io/serviceaccount/ca.crt

Caused by:
    No such file or directory (os error 2)
```

rather than just:
```
No such file or directory (os error 2)
```

cmd/pinniped-proxy/src/main.rs
```diff
@@ -17,17 +18,26 @@ async fn main() -> Result<()> {
     pretty_env_logger::init();
     let opt = cli::Options::from_args();
 
+    // Load the default certificate authority data on startup once.
+    let default_ca_data = fs::read_to_string(opt.default_ca_cert.clone())
+        .with_context(|| format!("unable to load default-ca-cert at {}", opt.default_ca_cert))?;
```
absoludity, line 23: `.with_context` is a trait of the `anyhow` package which enables adding context to a result before potentially returning an error. So instead of seeing just the following on the cmd line:

```
No such file or directory (os error 2)
```

you instead get:

```
Error: unable to load default-ca-cert at /var/run/secrets/kubernetes.io/serviceaccount/ca.crt

Caused by:
    No such file or directory (os error 2)
```

cmd/pinniped-proxy/src/main.rs
```diff
@@ -17,17 +18,26 @@ async fn main() -> Result<()> {
...
+
     // For every incoming connection, we make a new hyper `Service` to handle
     // all incoming HTTP requests on that connection. This is done by passing a
     // closure to the hyper `make_service_fn` which returns our custom `make_svc`
-    // function that can be used for each connection. 
+    // function that can be used for each connection.
     let make_svc = make_service_fn(|_conn| {
+        let default_ca_data = default_ca_data.clone();
```
absoludity, line 30: At first this can look a little odd - why are we ghosting the variable which is already available in scope? The reason is that this closure needs to be able to reference the ca data in an async block below, but the async block, being async, could live longer than the referenced var from the parent scope. So we clone it here so the value is owned inside this closure, and similarly below when defining the service fn closure so it owns its own copy of the value.

cmd/pinniped-proxy/src/main.rs
```diff
@@ -17,17 +18,26 @@ async fn main() -> Result<()> {
...
         // The closure just returns an async block that runs a service to handle
         // all requests for the connection.
         async {
             // `service_fn` is a helper from the hyper crate which converts a
-            // function that returns a Response into a `Service`.
-            Ok::<_, Infallible>(service_fn(service::proxy))
+            // function that returns a Response into a `Service`. We pass a
+            // closure to service_fn here so we can pass the default certificate
+            // authority data.
```
absoludity, line 37: Let me know if this comment isn't clear enough. The `service_fn` helper expects a function which takes a request and returns a response, but we want to use a function which additionally accepts the default ca data, so I'm creating a closure which has the expected signature wrapping our proxy which additionally passes the default_ca_data from the closures scope.

cmd/pinniped-proxy/src/main.rs
antgamdia: Interesting! Much more meaningful errors. Having a look at the docs, it seems that for enabling backtraces, you have to set the envars `RUST_BACKTRACE=1` and/or `RUST_LIB_BACKTRACE=0`

cmd/pinniped-proxy/src/main.rs
antgamdia: Thanks for the explanation; it's interesting how in Rust one should be aware of this ownership stuff. But, in the end, it's worthwhile as you get thread-safety out of the box

cmd/pinniped-proxy/src/main.rs
antgamdia: I'm not catching on to this, but I guess I should fetch the current branch and take a closer look at the code.

antgamdia (approved):

cmd/pinniped-proxy/src/main.rs
absoludity: Yep, both the log-level and whether you display full back traces are just env vars.

PR #2210, opened 2020-12-11
Don't shadow method name with field name.

absoludity:
### Description of the change

No functional changes, it just clears up some (but not all) confusion that I found here while preparing some related work.

### Benefits

Hopefully clearer that the field formally known as `svcClientset` on the `kubeHandler` is the clientset using the service account credentials for the cluster on which Kubeapps is running. Also reduces confusion of the same name being used as a field and method on the struct.

Also found an unhandled error.

### Possible drawbacks

None.

### Applicable issues

  - Ref: #2181

### DIFF
--- a/pkg/kube/kube_handler.go
+++ b/pkg/kube/kube_handler.go
@@ -124,8 +124,9 @@ type kubeHandler struct {
 	// The namespace in which (currently) app repositories are created on the default cluster.
 	kubeappsNamespace string
 
-	// clientset using the pod serviceaccount on the default cluster.
-	svcClientset combinedClientsetInterface
+	// kubeappsSvcClientset is the clientset using the pod serviceaccount on the
+	// cluster on which kubeapps is installed.
+	kubeappsSvcClientset combinedClientsetInterface
 
 	// Configuration for additional clusters which may be requested.
 	clustersConfig ClustersConfig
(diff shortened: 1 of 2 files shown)


pkg/kube/kube_handler.go
```diff
@@ -143,7 +144,7 @@ type userHandler struct {
 	// The namespace in which (currently) app repositories are created.
 	kubeappsNamespace string
 
-	// clientset using the pod serviceaccount for the default cluster
+	// clientset using the pod serviceaccount for the specific cluster
 	svcClientset combinedClientsetInterface
```
andresmgot, line 148: if you are renaming this to kubeappsSVCClientset, why don't you change this one as well? (It's the same thing)

pkg/kube/kube_handler.go
```diff
@@ -221,7 +223,7 @@ func (a *kubeHandler) AsUser(token, cluster string) (handler, error) {
 		return nil, err
 	}
 
-	svcClientset, err := a.svcClientSet(cluster, config)
+	svcClientset, err := a.getSvcClientsetForCluster(cluster, config)
```
andresmgot, line 226: +1

andresmgot (approved):

antgamdia (approved):

pkg/kube/kube_handler.go
absoludity: I didn't think it was the same thing. The one I've renamed to `kubeappsSvcClientSet` uses the service account on which kubeapps is installed, always, where as this `svcClientset` is for the current cluster in a multicluster env (think requesting namespaces to populate the selector). Yes, it will be the same thing in a single cluster environment but isn't generally.

PR #2211, opened 2020-12-11
Ensure Kubeops proxies user requests via pinniped proxy when configured.

absoludity:
### Description of the change

When testing pinniped-proxy with Kubeapps, requests from the browser client to the Kubeapps frontend for the k8s API were successfully proxied via pinniped-proxy, hitting the k8s api server with generated certs which were auth'd by the api server successfully.

But what I hadn't counted on was the requests from the browser client to our Kubeapps frontend for the kubeops service, which creates its own `client-go` kubernetes client set for user requests. This PR ensures that these are also proxied via pinniped-proxy when configured for that cluster.

### Benefits

Kubeops can also successfully interact with the k8s api server of each cluster using pinniped.

### Possible drawbacks

~I still need to do the IRL (in real life) test to ensure this works as expected.~ Edit: finally found and fixed a number of small issues so that this works IRL.

### Applicable issues

  - Ref: #2181 

### Additional information

Follows #2210 which just cleaned up some naming.

### DIFF
--- a/cmd/kubeops/main_test.go
+++ b/cmd/kubeops/main_test.go
@@ -19,7 +19,7 @@ func TestParseClusterConfig(t *testing.T) {
 	}{
 		{
 			name:       "parses a single cluster",
-			configJSON: `[{"name": "cluster-2", "apiServiceURL": "https://example.com", "certificateAuthorityData": "Y2EtY2VydC1kYXRhCg==", "serviceToken": "abcd"}]`,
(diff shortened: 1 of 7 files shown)


antgamdia (approved):
Thank you for the changes, especially for the useful comments and the links explaining why some parts are required.
It's interesting how the RoundTripperis important here to set the headers when proxying.

absoludity:
I have some further changes that I'll need, in terms of shuffling the config a little to enable multi-cluster support, but I'll do that in a followup branch.

PR #2213, opened 2020-12-11
Required chart configuration changes to enable Kubeapps configured with pinniped-proxy

absoludity:
### Description of the change

The chart changes in this PR were used for an IRL local demo that Kubeapps can be configured to proxy requests to specific clusters via pinniped so that although the cluster itself may not be configured with OIDC, Kubeapps can itself use OIDC (with the requests being routed so that the oidc credential is exchanged by pinniped for a certificate auth).

The changes in https://github.com/absoludity/kubeapps/pull/2 are used for the local dev env setup (and have been extracted from this current PR).

### Benefits

Demonstrates that we can run Kubeapps with OIDC on a cluster not configured for OIDC (using pinniped).

### Possible drawbacks

I've noticed that logging in and using Kubeapps works fine generally with this, but there is one issue when viewing the release detail page in that XHR requests appear to get backlogged. Even switching to another route does not stop the new XHR requests from being queued. Not sure yet if this is related to changes in master or this PR, but I'll need to investigate further separately.

### Applicable issues

  - Ref: #2181 

### Additional information

I originally had the pinniped-proxy service running as a container on the frontend because I'd thought all requests to the service would be directly from the frontend, forgetting that some requests are from kubeops.

Adapting this so that the kubeops service can also proxy requests required the separate k8s clusterip service in this PR, but I didn't yet pull the container out itself to a separate deployment. It may be that it's related to the above backlog issue (requests going frontend->kubeops->frontend(pinniped)->k8s server potentially locking the frontend), but need to check. Not sure if it's worth moving the pinniped-proxy container out to its own deployment now? What do you think?

#### Testing instructions (for posterity)

```
# Create a single kind cluster without any oidc config (see change in cluster-kind.md)
make cluster-kind

# setup pinniped concierge in the cluster
make deploy-pinniped

# Update the jwt-authenticator with the newly created clusters' CA
less ~/.kube/kind-config-kubeapps
# Copy the `certificate-authority-data` then edit `kubeapps-pinniped-jwt-authenticator.yaml` and set the data for `.spec.tls.certificateAuthorityData`.

make add-pinniped-jwt-authenticator

# Deploy kubeapps with the modified values to use pinniped-proxy, ensuring any
# kubeops/pinniped-proxy images have been loaded into the cluster.
make deploy-dev
```


andresmgot:
I just took a quick look, but shouldn't the request be: `http://kubeapps.kubeapps:3333` rather than `31333` since kubeops is accessing the proxy from within the cluster?

absoludity:
> I just took a quick look, but shouldn't the request be: `http://kubeapps.kubeapps:3333` rather than `31333` since kubeops is accessing the proxy from within the cluster?

Yeah, I had tried that first then updated to 31333 while testing things and forgot to remove.

andresmgot:
> Adapting this so that the kubeops service can also proxy requests required the separate k8s clusterip service in this PR, but I didn't yet pull the container out itself to a separate deployment. It may be that it's related to the above backlog issue (requests going frontend->kubeops->frontend(pinniped)->k8s server potentially locking the frontend), but need to check. Not sure if it's worth moving the pinniped-proxy container out to its own deployment now? What do you think?

For me it makes sense to have the pinniped proxy running as part of the frontend, with the oauth2_proxy, as a single point to handle authentication. What would be the benefit of moving it to its own deployment? Why would it lock the frontend? As far as I can tell we won't need to scale it on its own and we are not going to define its own network policy either. In any case, we can always move it to its own deployment later if we need to.

chart/kubeapps/templates/kubeapps-frontend-config.yaml
```diff
@@ -61,10 +61,29 @@ data:
       location ~* /api/clusters/{{ .name }} {
         rewrite /api/clusters/{{ .name }}/(.*) /$1 break;
         rewrite /api/clusters/{{ .name }} / break;
+
+      # Helm returns a nil pointer error when accessing foo.bar if foo doesn't
+      # exist, even with the `default` function.
+      # See https://github.com/helm/helm/issues/8026#issuecomment-756538254
+      {{- $pinnipedConfig := .pinnipedConfig | default dict }}
+      {{- if and $.Values.pinnipedProxy.enabled $pinnipedConfig.exchangeCredentials }}
```
andresmgot, line 69: Would it make sense to simplify this and only allow all-or-none configuration regarding pinniped? (just configuring all the clusters with the same pinniped instance if given)

I am okay leaving this as it is since it's already working but the additional flexibility may be adding unnecessary complexity to the chart and config.

cmd/pinniped-proxy/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+ADD ./cmd/pinniped-proxy ./
+RUN rm ./target/release/deps/pinniped_proxy*
+RUN cargo build --release
+
+# Put it all together on a scratch image.
+# TODO: Turns out we'll need to include libssl on the scratch image, perhaps compiling libssl or copying dyn libs listed by ldd.
+# FROM scratch
+# COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+# FROM bitnami/minideb:buster-snapshot-20201121T213529Z
+FROM debian:10.6-slim
```
andresmgot, line 20: why have you commented out `minideb`?

andresmgot (approved):

chart/kubeapps/templates/kubeapps-frontend-config.yaml
```diff
@@ -61,10 +61,29 @@ data:
...
+        # If pinniped proxy is enabled *and* the current cluster is configured
+        # to exchange credentials then we route via pinnipedProxy to exchange
+        # credentials for client certificates.
+        {{- if .apiServiceUrl }}
+        proxy_set_header PINNIPED_PROXY_API_SERVER_URL {{ .apiServiceURL }};
+        {{- end }}
+        {{- if .certificateAuthorityData }}
+        proxy_set_header PINNIPED_PROXY_API_SERVER_CERT {{ .certificateAuthorityData }};
+        {{- end }}
+        proxy_pass http://localhost:3333;
```
antgamdia, line 79: Shouldn't we also parametrize this? I believe we're doing so in other `proxy_pass` but I'm unsure if it makes sense here.

antgamdia (approved):

chart/kubeapps/templates/kubeapps-frontend-config.yaml
absoludity: > Would it make sense to simplify this and only allow all-or-none configuration regarding pinniped? (just configuring all the clusters with the same pinniped instance if given)

You (as the person installing Kubeapps) may not have that option if you are given access to setup Kubeapps with existing clusters, some of which use pinniped, some straight oidc, some different pinniped authenticators etc. So yes, it would be simpler if we had the guarantee that "pinniped-concierge will always be setup with the same namespace and authenticator details for all clusters", but given that we don't, I didn't want to enforce that as a restriction.

> 
> I am okay leaving this as it is since it's already working but the additional flexibility may be adding unnecessary complexity to the chart and config.

I've tried to do this so it doesn't have unnecessary complexity for the user: so the defaults assume that pinniped is install in the same namespace and the same authenticator in all clusters, but can be changed.

chart/kubeapps/templates/kubeapps-frontend-config.yaml
absoludity: Yes, you're right. And this would be another reason to instead define `values.pinnipedProxy.URL` or similar as discussed on the previous PR. I'll do that in a follow-up so we can discuss the change there.

cmd/pinniped-proxy/Dockerfile
absoludity: I don't remember. I was originally trying to use scratch, but as above, that wasn't possible (yet). I must have tried minideb but had issues. I'll leave the comment and try in the followup PR.

Issue #2174, opened 2020-11-20
Add server-side chart filtering by repo


antgamdia:
### Description:

When adding repos with a big number of charts, let's say 1890 (eg, https://repo.chartcenter.io), the UI allows filtering by repository. However, this filtering just takes place on the client-side since the same request is being made (/api/assetsvc/v1/clusters/default/namespaces/kubeapps/charts), regardless of the selected filter.

Whereas the assetsvc does have a way to filter out charts by a certain query param, it is not fully implemented.

### Steps to reproduce the issue:

1. Add a "big" repo like https://repo.chartcenter.io a
2  Adds a "small" repo like https://charts.bitnami.com/bitnami
3. Filter by the small one
![image](https://user-images.githubusercontent.com/11535726/99799787-8b942380-2b33-11eb-9f60-ea314d57270a.png)
4. Wait more than 2-3 endless seconds to load

### Describe the results you received:

When a big repo is added, it makes slow the loading process of every repo.

### Describe the results you expected:

I'd expect the same delay as before adding the big repo.

### Additional information you deem important (e.g. issue happens only occasionally):

Most logic is already implemented, it is just sending the proper API calls and handle the params accordingly

### Version of Helm, Kubeapps and Kubernetes:

Kubeapps version fb1adf02d7101eed08f5dbe118c06393a268dcf7

antgamdia:
Rel #1010

antgamdia:
An initial set of server-side stuff to implement:

- [X] Add pagination in the assetsvc for listing charts (PR #2216)
- [ ]  List all the different chart categories
- [ ]  Custom search for charts (e.g., id, name, description, category...) (perhaps autocompletion support?)

antgamdia:
Closing as every PR related to server-side filtering has been merged.
The UI change discussion continues in https://github.com/kubeapps/kubeapps/issues/1010

PR #2216, opened 2020-12-14
Implement pagination in assetsvc

antgamdia:
### Description of the change

Add the missing server-side pagination. There was already some logic for this feature, but it was not fully implemented yet.
This PR is to support both `page` and `size` query parameters when listing: i) all charts; ii) all charts within a repo.

If no `size`, defaults to `size=0` (i.e., no size limit at all)
If no `page` (or `page=0`), defaults to `page=1` (i.e., first page)

```
 .../api/assetsvc/v1/clusters/default/namespaces/default/charts/bitnami
 { "data": [ ...(79 objects)... ], "meta": { "totalPages": 1, "totalCharts": 79 }}

 .../api/assetsvc/v1/clusters/default/namespaces/default/charts/bitnami?size=10
 { "data": [ ...(10 objects)... ], "meta": { "totalPages": 8, "totalCharts": 79 }}

 .../api/assetsvc/v1/clusters/default/namespaces/default/charts/bitnami?page=<whatever>
 { "data": [ ...(79 objects)... ], "meta": { "totalPages": 1, "totalCharts": 79 }}

 .../api/assetsvc/v1/clusters/default/namespaces/default/charts/bitnami?page=2&size=78
 { "data": [ ...(1 objects)... ], "meta": { "totalPages": 2, "totalCharts": 79 }}
```

### Benefits

Frontend calls to assetsvc will be able to be paginated while preserving the current behavior if no parameters are passed.

### Possible drawbacks

The current implementation uses `LIMIT` and `OFFSET` PostgreSQL statements. There are better ways to paginate results if we focus on performance. Nevertheless, it is a much more understandable and readable approach than using keyset, TID scan or custom cursors [1].

### Applicable issues

  - Related #2174 

### Additional information

[1] https://www.citusdata.com/blog/2016/03/30/five-ways-to-paginate/


cmd/assetsvc/handler.go
```diff
@@ -100,14 +103,14 @@ func min(a, b int) int {
 }
 
 func getPaginatedChartList(namespace, repo string, pageNumber, pageSize int) (apiListResponse, interface{}, error) {
-	charts, totalPages, err := manager.getPaginatedChartList(namespace, repo, pageNumber, pageSize)
-	return newChartListResponse(charts), meta{totalPages}, err
+	charts, totalPages, totalCharts, err := manager.getPaginatedChartList(namespace, repo, pageNumber, pageSize)
```
andresmgot, line 106: why is the totalCharts number necessary?

cmd/assetsvc/postgresql_utils.go
```diff
@@ -71,12 +72,30 @@ func (m *postgresAssetManager) getPaginatedChartList(namespace, repo string, pag
...
 	if err != nil {
-		return nil, 0, err
+		return nil, 0, 0, err
+	}
+
+	dbCountQuery := fmt.Sprintf("SELECT count(info) FROM %s %s", dbutils.ChartTable, repoQuery)
+	count, err := m.QueryCount(dbCountQuery, queryParams...)
+	if err != nil {
+		return nil, 0, 0, err
 	}
```
andresmgot, line 92: this is calculating the total number of charts, right? (regardless of the pagination). If that's the case, you just need to do the query in case the pagination is requested (pageSize > 0) so I would do this query within the `if` below (removing the `count > 0` condition).

andresmgot (commented):
Thanks @antgamdia! I have a couple of questions.

cmd/assetsvc/handler.go
antgamdia: It is not strictly necessary at all. In fact, one can roughly estimate this number by size*totalPages. However, if, in the future, we want to display the total amount of results in the UI, we will have this number directly in the meta information.
But, as you pointed out, currently it is not being useful.

cmd/assetsvc/postgresql_utils.go
antgamdia: Mmm, I see your point. We can skip this query as you mention, with no pagination this totalCharts would be just `len(charts)`. I'll address that.

cmd/assetsvc/handler.go
andresmgot: okay, so simplicity matters, I would remove this. The less we need to maintain the better :)

antgamdia: As requested, removed totalChart; simplicity wins.
If, eventually, we want to get it back again, we will just have to revert 
the commit a2e0e4cea35c05838cd0f4db003ba5529d5cc48a

cmd/assetsvc/handler.go
```diff
@@ -71,17 +71,19 @@ type count struct {
...
-	page := req.FormValue("page")
-	size := req.FormValue("size")
-	pageInt, err := strconv.ParseUint(page, 10, 64)
+// getPageAndSizeParams extracts the page number and the page size of a request. Default (page,size) = (1, 0) if not set
+func getPageAndSizeParams(req *http.Request) (int, int) {
+	pageNumber := req.FormValue("page")
+	pageSize := req.FormValue("size")
+
+	pageNumberInt, err := strconv.ParseUint(pageNumber, 10, 64)
 	if err != nil {
```
andresmgot, line 80: I would handle the case in which the page is `0` and set it to `1` in that case.

cmd/assetsvc/postgresql_utils.go
```diff
@@ -71,12 +72,30 @@ func (m *postgresAssetManager) getPaginatedChartList(namespace, repo string, pag
 		repoQuery = strings.Join(clauses, " AND ")
 		repoQuery = "WHERE " + repoQuery
 	}
-	dbQuery := fmt.Sprintf("SELECT info FROM %s %s ORDER BY info ->> 'name' ASC", dbutils.ChartTable, repoQuery)
+
+	paginationClause := ""
+	if pageSize > 0 && pageNumber > 0 {
```
andresmgot, line 77: `pageNumber` confused me a bit, since setting the number to 0 would disable the pagination. Added a comment above, it's better to fallback in that case to `1` so this condition is not necessary.

andresmgot (approved):
Thanks! I have just an extra comment.

cmd/assetsvc/postgresql_utils.go
antgamdia: The double check here is just for also support the same logic (page=0 disable pagination) also in the tests, where this method is directly invoked. 
I agree with your previous comment about explicitly handling the page=0 case.  I'm gonna update it!

andresmgot: > The double check here is just for also support the same logic (page=0 disable pagination) also in the tests, where this method is directly invoked.

It's a bit weird that setting `pageNumber = 0` has different outputs in different places (either set it to 1 or disable pagination). IMO it would be more clear to either set it to 1 as well here or return an error if the input is not valid (with a comment explaining why this has been sanitized before).

andresmgot (approved):
still +1 when you take a look to the comment.

PR #2219, opened 2020-12-15
Add a getChartCategories endpoint

antgamdia:
### Description of the change

Before this PR, the only way to retrieve the list of distinct chart categories was to fetch all charts and, then, manually retrieves these unique categories. This PR is to add an endpoint for doing so but at the server-side and leveraging from the DB.

The proposed response in this PR is the list of distinct categories (`name`) and the count of charts belonging to each category (`count`). An example is depicted below:
```
{"data":[{"name":"Analytics","count":10},{"name":"ApplicationServer","count":2},{"name":"CertificateAuthority","count":1},{"name":"CMS","count":5},{"name":"CRM","count":2},{"name":"Database","count":12},{"name":"DeveloperTools","count":7},{"name":"E-Commerce","count":3},{"name":"E-Learning","count":1},{"name":"Forum","count":2},{"name":"HumanResourceManagement","count":1},{"name":"Infrastructure","count":24},{"name":"LogManagement","count":1},{"name":"MachineLearning","count":3},{"name":"ProjectManagement","count":2},{"name":"Wiki","count":2},{"name":"WorkFlow","count":1}],"meta":{"totalPages":0}}
```

Particulary, there are two new endpoints:
`/clusters/{cluster}/namespaces/{namespace}/charts/categories`
`/clusters/{cluster}/namespaces/{namespace}/charts/{repo}/categories`

### Benefits

Getting the list of categories no longer will involve a whole fetching every chart at client-side.

### Possible drawbacks

N/A

### Applicable issues

- Related #2174

### Additional information

In the current version of UI we are not displaying the number of charts. I added the count just for providing a more complete response, but I'm open to simply return the `select distinct ...` if you may want to.

### DIFF
--- a/cmd/assetsvc/handler.go
+++ b/cmd/assetsvc/handler.go
@@ -111,9 +111,9 @@ func getPaginatedChartList(namespace, repo string, pageNumber, pageSize int) (ap
 	return newChartListResponse(charts), meta{totalPages}, err
 }
 
-func getAllChartCategories(namespace, repo string) (apiChartCategoryListResponse, interface{}, error) {
+func getAllChartCategories(namespace, repo string) (apiChartCategoryListResponse, error) {
 	chartCategories, err := manager.getAllChartCategories(namespace, repo)
-	return newChartCategoryListResponse(chartCategories), meta{}, err
+	return newChartCategoryListResponse(chartCategories), err
 }
 
 // listCharts returns a list of charts based on filter params
@@ -154,13 +154,13 @@ func getChartCategories(w http.ResponseWriter, req *http.Request, params Params)
 		return
 	}
 
-	chartCategories, meta, err := getAllChartCategories(namespace, repo)
+	chartCategories, err := getAllChartCategories(namespace, repo)
 	if err != nil {
 		log.WithError(err).Error("could not fetch categories")
 		response.NewErrorResponse(http.StatusInternalServerError, "could not fetch chart categories").Write(w)
 		return
 	}
-	response.NewDataResponseWithMeta(chartCategories, meta).Write(w)
+	response.NewDataResponse(chartCategories).Write(w)
 }
 
 // getChart returns the chart from the given repo
--- a/cmd/assetsvc/handler.go
+++ b/cmd/assetsvc/handler.go
@@ -52,12 +52,7 @@ type apiResponse struct {
 
 type apiListResponse []*apiResponse
 
-type apiChartCategoryResponse struct {
-	Name  string `json:"name"`
-	Count int    `json:"count"`
-}
-
-type apiChartCategoryListResponse []*apiChartCategoryResponse
+type apiChartCategoryListResponse []*models.ChartCategory
 
 type selfLink struct {
 	Self string `json:"self"`
@@ -410,9 +405,8 @@ func newChartResponse(c *models.Chart) *apiResponse {
 	}
 }
 
-func newChartCategoryResponse(c *models.ChartCategory) *apiChartCategoryResponse {
-	return &apiChartCategoryResponse{
-
+func newChartCategoryResponse(c *models.ChartCategory) *models.ChartCategory {
+	return &models.ChartCategory{
 		Name:  c.Name,
 		Count: c.Count,
 	}
--- a/cmd/assetsvc/handler.go
+++ b/cmd/assetsvc/handler.go
@@ -52,6 +52,13 @@ type apiResponse struct {
 
 type apiListResponse []*apiResponse
 
+type apiChartCategoryResponse struct {
+	Name  string `json:"name"`
+	Count int    `json:"count"`
+}
+
+type apiChartCategoryListResponse []*apiChartCategoryResponse
+
 type selfLink struct {
 	Self string `json:"self"`
 }
@@ -104,6 +111,11 @@ func getPaginatedChartList(namespace, repo string, pageNumber, pageSize int) (ap
 	return newChartListResponse(charts), meta{totalPages}, err
 }
 
+func getAllChartCategories(namespace, repo string) (apiChartCategoryListResponse, interface{}, error) {
+	chartCategories, err := manager.getAllChartCategories(namespace, repo)
+	return newChartCategoryListResponse(chartCategories), meta{}, err
+}
+
 // listCharts returns a list of charts based on filter params
 func listCharts(w http.ResponseWriter, req *http.Request, params Params) {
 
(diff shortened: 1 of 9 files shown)


cmd/assetsvc/handler.go
```diff
@@ -104,6 +111,11 @@ func getPaginatedChartList(namespace, repo string, pageNumber, pageSize int) (ap
 	return newChartListResponse(charts), meta{totalPages}, err
 }
 
+func getAllChartCategories(namespace, repo string) (apiChartCategoryListResponse, interface{}, error) {
+	chartCategories, err := manager.getAllChartCategories(namespace, repo)
+	return newChartCategoryListResponse(chartCategories), meta{}, err
```
andresmgot, line 116: the `meta{}` argument is not necessary

cmd/assetsvc/main_test.go
```diff
@@ -105,6 +105,142 @@ func Test_GetCharts(t *testing.T) {
...
+			"two charts - same category",
+			[]*models.Chart{
+				{Repo: testRepo, ID: "my-repo/my-chart", Category: "cat1", ChartVersions: []models.ChartVersion{{Version: "0.0.1", Digest: "123"}}},
+				{Repo: testRepo, ID: "my-repo/dokuwiki", Category: "cat1", ChartVersions: []models.ChartVersion{{Version: "1.2.3", Digest: "1234"}, {Version: "1.2.2", Digest: "12345"}}},
+			},
+			[]*models.ChartCategory{
+				{Name: "cat1", Count: 1},
+				{Name: "cat2", Count: 2},
+				{Name: "cat3", Count: 3},
+			},
```
andresmgot, line 132: I don't understand this `expected`, if the input is only to `cat1` charts

cmd/assetsvc/main_test.go
```diff
@@ -105,6 +105,142 @@ func Test_GetCharts(t *testing.T) {
...
+
+			res, err := http.Get(ts.URL + pathPrefix + "/clusters/default/namespaces/my-namespace/charts/categories")
+			assert.NoError(t, err)
+			defer res.Body.Close()
+
+			assert.Equal(t, res.StatusCode, http.StatusOK, "http status code should match")
+
+			var b bodyAPIListResponse
+			json.NewDecoder(res.Body).Decode(&b)
+			assert.Len(t, *b.Data, len(tt.expectedChartCategories))
```
andresmgot, line 169: you are not using `tt.charts` anywhere?

cmd/assetsvc/postgresql_utils.go
```diff
@@ -55,6 +55,31 @@ func exists(current []string, str string) bool {
...
+		clauses = append(clauses, "(repo_namespace = $1 OR repo_namespace = $2)")
+	}
+	if repo != "" {
+		queryParams = append(queryParams, repo)
+		clauses = append(clauses, fmt.Sprintf("repo_name = $%d", len(queryParams)))
+	}
+	repoQuery := ""
+	if len(clauses) > 0 {
+		repoQuery = strings.Join(clauses, " AND ")
+		repoQuery = "WHERE " + repoQuery
```
andresmgot, line 72: it's more readable if you do `repoQuery = "WHERE " + strings.Join(...`

cmd/assetsvc/main_test.go
```diff
@@ -105,6 +105,142 @@ func Test_GetCharts(t *testing.T) {
...
+			defer cleanup()
+
+			rows := sqlmock.NewRows([]string{"name", "count"})
+			for _, chartCategories := range tt.expectedChartCategories {
+				rows.AddRow(chartCategories.Name, chartCategories.Count)
+			}
+
+			mock.ExpectQuery("SELECT (info ->> 'category')*").
+				WithArgs("my-namespace", kubeappsNamespace, tt.repo).
+				WillReturnRows(rows)
```
andresmgot, line 229: same here, you are just testing that adding `expectedCategories` runs the expected query. That's fine but you need to update the test names/inputs to really reflect what you were trying.

cmd/assetsvc/postgresql_utils_test.go
```diff
@@ -220,6 +220,62 @@ func Test_getChartsWithFilters_withSlashes(t *testing.T) {
...
+
+			rows := sqlmock.NewRows([]string{"name", "count"})
+			for _, chartCategories := range tt.expectedChartCategories {
+				rows.AddRow(chartCategories.Name, chartCategories.Count)
+			}
+
+			expectedParams := []driver.Value{"other-namespace", "kubeapps"}
+			if tt.repo != "" {
+				expectedParams = append(expectedParams, "bitnami")
+			}
```
andresmgot, line 264: don't understand this `if`, you are adding `bitnami` by default?

cmd/assetsvc/postgresql_utils_test.go
```diff
@@ -220,6 +220,62 @@ func Test_getChartsWithFilters_withSlashes(t *testing.T) {
...
+			mock.ExpectQuery("SELECT (info ->> 'category')*").
+				WithArgs(expectedParams...).
+				WillReturnRows(rows)
+
+			chartCategories, err := pgManager.getAllChartCategories(tt.namespace, tt.repo)
+			if err != nil {
+				t.Fatalf("Found error %v", err)
+			}
```
andresmgot, line 272: same here about tt.availableCharts

cmd/assetsvc/handler.go
```diff
@@ -52,6 +52,13 @@ type apiResponse struct {
 
 type apiListResponse []*apiResponse
 
+type apiChartCategoryResponse struct {
+	Name  string `json:"name"`
+	Count int    `json:"count"`
+}
```
andresmgot, line 58: this is the same than your struct `ChartCategory` below, you can just use that one.

andresmgot (commented):
Thanks!

cmd/assetsvc/main_test.go
antgamdia: My bad! I just copied the list from another test and didn't notice that

cmd/assetsvc/main_test.go
antgamdia: I thought it would useful, but, the mock query only uses the `tt.expectedChartCategories` so we can remove it.

cmd/assetsvc/postgresql_utils_test.go
antgamdia: My bad again... it should be `if tt.repo != "" { expectedParams = append(expectedParams, tt.repo) }`

andresmgot (approved):
Changes LGTM, thanks!

PR #2220, opened 2020-12-17
Add search parameters to assetsvc

antgamdia:
### Description of the change

This PR is to add a more complete range of server-side filtering and to refactor how some WHERE clauses were being generated.

For instance, before this PR, for looking up a certain chart with a given set of `version` and `apversion`, we performed a non-where `select from` and, later, we would iterate over each result to perform the find operation. Now, we leverage from the PostgreSQL jsonb operators and query using `"(info->'chartVersions' @> '[{\"version\":\"%s\",\"app_version\":\"%s\"}]'::jsonb)"` 

Particularly `/charts` now accepts  `name`, `version`, `appversion`, `repos`, `categories`, `q`, `page`, `size`.

Transitorily, the `/charts/{repo}` still exist for backwards compatibility. In the case  `/charts/repo1?repos=repo2,repo3`, the current behavior is just `WHERE ...(repo=repo1 OR repo=repo2 OR repo=repo3)`.

**Note**: due to the upcoming holidays, PTOs and so on, this PR is based on the latest unmerged changes in #2219.
_**Changes exclusively belonging to this PR are in commit 28ea3ff.**_ Wherefore this PR ~~is~~ was marked as a draft, as it requires the previous one to be merged.

### Benefits

All the existing frontend filters can now be performed at the backend. The refactor in the generation of the WHERE clauses should improve the overall performance.
Furthermore, the existing frontend calls should not be affected by this PR.

### Possible drawbacks

N/A

### Applicable issues

- Related #2174

### Additional information

The existing logic for retrieving one result (instead of a paginated list of results) remains unchanged. Perhaps we can refactor it as well in a follow-up PR, but if you may consider this PR is a good place for it, I'm open to implement it too.


andresmgot:
> **Note**: due to the upcoming holidays, PTOs and so on, this PR is based on the latest unmerged changes in #2219.
> _**Changes exclusively belonging to this PR are in commit [28ea3ff](https://github.com/kubeapps/kubeapps/commit/28ea3ff50af35e5b544e51d8f5d75a9e9a66b2de).**_ Wherefore this PR is marked as a draft, as it requires the previous one to be merged.

What I do in this situation is pushing branches in the origin repo (kubeapps/kubeapps). If you do so, you can create a PR against a branch different than `master`, in this case, you would create a branch against `assetsvc-listCategories`. Once the PR has been accepted you can either:
 - Merge the second PR into the first and then into master
 - Merge the first PR into master and change the base branch of the second to `master` (it requires to rebase changes)
 
The second is a bit more complicated so I usually do the first (only if both PRs have been accepted).

antgamdia:
Thank you for comments! That's a good option for PR with dependencies such as this one. I'll definitively take it into account for future PRs :)

cmd/assetsvc/postgresql_utils.go
```diff
@@ -39,6 +39,16 @@ type postgresAssetManager struct {
...
 
+type chartQuery struct {
+	namespace   string
+	chartName   string
+	version     string
+	appVersion  string
+	searchQuery string
+	repos       []string
+	categories  []string
+}
```
antgamdia, line 50: I don't know if it is the best place to put this struct, especially due to this is not an exclusive model for postgresql , perhaps in utils.go?

cmd/assetsvc/main.go
```diff
@@ -44,18 +44,17 @@ func setupRoutes() http.Handler {
 	// Routes
 	apiv1 := r.PathPrefix(pathPrefix).Subrouter()
 	// TODO: mnelson: Seems we could use path per endpoint handling empty params? Check.
-	apiv1.Methods("GET").Path("/clusters/{cluster}/namespaces/{namespace}/charts").Queries("name", "{chartName}", "version", "{version}", "appversion", "{appversion}").Handler(WithParams(listChartsWithFilters))
```
antgamdia, line 47: I decided to get rid of the .Queries, it doesn't make sense for me to have to manually define a new line for each combination of params. I'd rather handle it at the controller so that we can also manage empty values accordingly.

cmd/assetsvc/postgresql_utils.go
```diff
@@ -235,34 +213,77 @@ func (m *postgresAssetManager) getChartFilesWithFallback(namespace, filesID stri
 	return chartFiles, nil
 }
 
-func containsVersionAndAppVersion(chartVersions []models.ChartVersion, version, appVersion string) (models.ChartVersion, bool) {
-	for _, ch := range chartVersions {
-		if ch.Version == version && ch.AppVersion == appVersion {
-			return ch, true
-		}
+func (m *postgresAssetManager) getPaginatedChartListWithFilters(cq chartQuery, pageNumber, pageSize int) ([]*models.Chart, int, error) {
+	whereQuery, whereQueryParams := m._generateWhereClause(cq)
```
antgamdia, line 217: Now the logic for applying filters in the WHERE clause is shared by using `m._generateWhereClause`

cmd/assetsvc/postgresql_utils.go
```diff
@@ -235,34 +213,77 @@ func (m *postgresAssetManager) getChartFilesWithFallback(namespace, filesID stri
...
+			fmt.Sprintf("(info ->> 'sources' ILIKE $%d) OR ", len(whereQueryParams)) +
+			fmt.Sprintf("(info -> 'maintainers' ->> 'name' ILIKE $%d))", len(whereQueryParams))
+		whereClauses = append(whereClauses, searchClause)
+	}
+	if len(whereClauses) > 0 {
+		whereQuery = "WHERE " + strings.Join(whereClauses, " AND ")
+	}
+
+	return whereQuery, whereQueryParams
 }
```
antgamdia, line 289: This is the main part of this PR, the `_generateWhereClause` method. Given a `chartQuery` it generates all the WHERE clauses. 
Currently it joins every clause using AND, but we can customize it by passing the operator as a new param. If it's worthwhile, I'm willing to add it.

cmd/assetsvc/postgresql_utils.go
```diff
@@ -47,54 +57,22 @@ func newPGManager(config datastore.Config, kubeappsNamespace string) (assetManag
...
-	}
-	if repo != "" {
-		queryParams = append(queryParams, repo)
-		clauses = append(clauses, fmt.Sprintf("repo_name = $%d", len(queryParams)))
-	}
-	repoQuery := ""
-	if len(clauses) > 0 {
-		repoQuery = "WHERE " + strings.Join(clauses, " AND ")
-	}
+func (m *postgresAssetManager) _getPaginatedChartList(whereQuery string, whereQueryParams []interface{}, pageNumber, pageSize int) ([]*models.Chart, int, error) {
```
andresmgot, line 75: I am not sure why you changed this name but please follow the current convention for naming. You don't need the underscore for telling that some method is internal since methods starting with an lowercase letter are already not exported.

cmd/assetsvc/postgresql_utils.go
```diff
@@ -235,34 +213,77 @@ func (m *postgresAssetManager) getChartFilesWithFallback(namespace, filesID stri
...
+			"(info->>'name' = $%d)", len(whereQueryParams),
+		))
 	}
-	result := []*models.Chart{}
-	for _, c := range charts {
-		if _, found := containsVersionAndAppVersion(c.ChartVersions, version, appVersion); found {
-			result = append(result, c)
+	if cq.version != "" && cq.appVersion != "" {
+		whereClauses = append(whereClauses, fmt.Sprintf(
+			"(info->'chartVersions' @> '[{\"version\":\"%s\",\"app_version\":\"%s\"}]'::jsonb)", cq.version, cq.appVersion,
```
andresmgot, line 244: this could be a SQL injection problem (we are not escaping version nor appVersion from the request, right?)

cmd/assetsvc/postgresql_utils.go
andresmgot: I don't think it's needed to do that atm

andresmgot: Also, you need to add tests for this function so you can ensure it behaves as expected. You can try with a matrix that given some inputs, produces the expected query and params.

andresmgot (commented):
Thanks @antgamdia! I have some comments.

cmd/assetsvc/postgresql_utils.go
antgamdia: Absolutely agree with you. I tried to add $-parameters, but Postgresql does not allow to parameterize inside of a jsonb literal and I ended by simply doing so; totally my bad! Thanks for pointing this out.

Changing it to `(info->'chartVersions' @> $1::jsonb)`, where `$1` would be `fmt.Sprintf('[{"version":"%s","app_version":"%s"}]', cq.version, cq.appVersion)`
Therefore, the injection cannot break out of scope of $, I guess.

andresmgot (approved):
Changes look good to me, you can merge master to avoid the e2e issue you are seeing and runnning the tests.

andresmgot:
I meant to merge master into this branh to get the final green light from the CI :)

antgamdia:
Oh! I didn't get that, my bad! I can revert the merge and do it properly if you want to.

andresmgot:
No need, the build is master came up green :)

PR #2224, opened 2021-01-03
Update chart deps and images.

absoludity:
- Update chart dependencies and images.
- ~~Update frontend deps.~~ (we need to do these more carefully as results in ts errors).

### Description of the change

Follows the README instructions for manually updating the chart dependencies, default images used and frontend deps.

### DIFF
--- a/chart/kubeapps/Chart.lock
+++ b/chart/kubeapps/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: postgresql
   repository: https://charts.bitnami.com/bitnami
-  version: 10.0.0
-digest: sha256:f8c58847aa69098396a2c24071459d6da8790b499590ebee84cc3cbe25fe2167
-generated: "2020-11-18T15:24:50.839372537Z"
+  version: 10.2.0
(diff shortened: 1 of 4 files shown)


antgamdia:
> * ~Update frontend deps.~ (we need to do these more carefully as results in ts errors).

Thanks for updating all the deps! I also tried updating frontend ones (just to see what happens) and noticed these errors. Perhaps we should just bump up the patch versions x.x.x+1 here and perform the upgrade in a follow-up PR.

antgamdia (approved):

absoludity:
> > * ~Update frontend deps.~ (we need to do these more carefully as results in ts errors).
> 
> Thanks for updating all the deps! I also tried updating frontend ones (just to see what happens) and noticed these errors. Perhaps we should just bump up the patch versions x.x.x+1 here and perform the upgrade in a follow-up PR.

Last time I had a similar issue it was because we'd incorrectly specified some version conditions (effectively stopping more recent versions of a couple of libs) which meant others couldn't update. But yes, +1 for a short-term solution if the longer term requires more time. I've just landed this so the frontend changes can be done independently.

> 
> btw, shouldn't we also update the go deps?

Sorry, I didn't see this edit before I landed the PR. Yes, that would be a good addition to include in the README instructions too.

PR #2225, opened 2021-01-04
Fix link to chartmuseum chart.

absoludity:
### Description of the change

The google stable chart repo is no longer valid. We need to install chartmuseum
during CI from its new location as per
https://github.com/helm/chartmuseum/pull/393

The current URL results in a 403 now.

### Benefits

CI works again?


absoludity:
Landing to unblock CI.

PR #2229, opened 2021-01-07
Multi-cluster pinniped-proxy support

absoludity:
- Update concierge params to be default only (later overridable for multi-cluster)
- Switch PinnipedProxyURL to kubeapps-wide and support concierge options.

### Description of the change

While doing IRL testing, it was apparent that the configuration was a little
confused. The Kubeapps configuration for PinnipedProxyURL should not be per
cluster (and in the cluster config), but rather have one value for the Kubeapps installation only (since there is just the one service URL when enabled).

Similarly, the setup of pinniped-concierge (which runs on each cluster) may
differ between clusters, so it needs to be possible to set the required values for
requests to pinniped-concierge to match a per cluster setup, even if we currently demo one cluster only.

Chart changes required for IRL testing are in #2213 .

### Benefits

The new config supports future multicluster setups with pinniped.

### Applicable issues

  - Ref #2181


pkg/kube/kube_handler.go
```diff
@@ -63,15 +63,29 @@ type ClusterConfig struct {
...
+	// that should be used for any credential exchange.
+	PinnipedConfig PinnipedConciergeConfig `json:"pinnipedConfig,omitempty"`
+}
+
+// PinnipedConciergeConfig enables each cluster configuration to specify the
+// pinniped-concierge installation to use for any credential exchange.
+type PinnipedConciergeConfig struct {
+	// ExchangeCredentials just flags whether this cluster should use
+	// pinniped to exchange credentials.
+	ExchangeCredentials bool `json:"exchangeCredentials"`
```
andresmgot, line 77: I would call this var just `Enable`, that way it's self explanatory without Pinniped context

cmd/kubeops/main_test.go
```diff
@@ -77,9 +78,27 @@ func TestParseClusterConfig(t *testing.T) {
...
+						Name:                     "cluster-2",
+						APIServiceURL:            "https://example.com",
+						CertificateAuthorityData: "ca-cert-data\n",
+						ServiceToken:             "abcd",
+						PinnipedConfig: kube.PinnipedConciergeConfig{
+							ExchangeCredentials: true,
+						},
+					},
+				},
+				PinnipedProxyURL: "http://kubeapps-internal-pinniped-proxy.kubeapps:3333",
```
andresmgot, line 99: would it be better to store the URL with the rest of configuration? That way everything is together. I found it a bit confusing to have a defined proxy URL if later on it's disabled in the `PinnipedConfig`

andresmgot (approved):
Thanks, I just have a couple of comments regarding readability

antgamdia (approved):

cmd/kubeops/main_test.go
absoludity: Yeah, the example here in the test doesn't really show why it's required for a multi-cluster setup. In such a setup, each cluster can have `pinniped-concierge` (not `-proxy`) installed, potentially in a different namespace with a different type of authenticator - or not... one of the clusters may also be configured with the OIDC options and not need pinniped. That's why we need per cluster `PinnipedConfig`. But there will only ever be one `PinnipedProxyURL` for a Kubeapps installation, no matter how many clusters there are, since it is proxying requests to all the clusters which are configured with pinniped (using each cluster's pinniped-concierge to exchange credentials).

So it didn't make sense to have the `PinnipedProxyURL` within the per-cluster `PinnipedConfig`, as that would enable invalid configurations which we cannot handle. I would have preferred for this to be `values.pinnipedProxy.URL` (or `.Endpoint`) or similar, but I also needed to be able to feed the value down through the kubeops CLI to the `pkg/kube` handlers, which only receive the cluster config. So rather than adding yet another arg, I placed it here in the config.

Perhaps the ideal here would be to make it part of `values.pinnipedProxy.URL` and use that to populate the `ClustersConfig.PinnipedProxyURL` (and making the field omitted from the json, so it isn't ever passed in as part of the clusters config). Let me know what you think, happy to update in a further PR.

andresmgot: > So it didn't make sense to have the PinnipedProxyURL within the per-cluster PinnipedConfig, as that would enable invalid configurations which we cannot handle.

I get the point that having different  URLs per cluster is not a valid configuration but I still think that having all the Pinniped configuration in the same place (from this service POV, it would be the `PinnipedConfig`).

> Perhaps the ideal here would be to make it part of values.pinnipedProxy.URL and use that to populate the ClustersConfig.PinnipedProxyURL (and making the field omitted from the json, so it isn't ever passed in as part of the clusters config). Let me know what you think, happy to update in a further PR.

I've lost track of the places in which you configure this but yes, the chart values is the right place to enforce a single URL (and any other info) for later generate a `Config` per cluster if needed.

PR #2234, opened 2021-01-08
Automatically resolve apiVersions and plurals

### LINKED ISSUE #2049: [NEED HELP] Clarity UI - Websocket errors
### Description:

Hi, 

I've installed kubeapps on an AKS cluster and I linked it with Azure AD to let our user to authenticate through Azure. On the **hex** UI everything seems to works but when I tried to enable the new UI _(clarity)_ with the feature flag provided, I have several websocket errors and after couple of minute the application disconnect me and I'm not able to connect again when I click again on `login with OIDC provider` I have a blank page and I need to wait some time to be able to login again.

Here you can see the message that I have on the Google Chrome console:

kubeapps-wss-bug

Just for information, the user that I'm using to login is a cluster admin. I'm using NGINX as ingress controller and I have setup the following annotations:

```
nginx.ingress.kubernetes.io/proxy-read-timeout: "3600"
nginx.ingress.kubernetes.io/proxy-send-timeout: "3600"
```

### Flags used

```
featureFlags:
  invalidateCache: true
  operators: false
  # ui is a WIP feature for the migration to Clarity design system.
  ui: clarity
```

Can someone help me on this issue ?

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.2.3", GitCommit:"8f832046e258e2cb800894579b1b3b50c2d83492", GitTreeState:"clean", GoVersion:"go1.13.12"}
```

```
NAME                    NAMESPACE       REVISION        UPDATED                                 STATUS          CHART                   APP VERSION
kubeapps                blue-dev        1               2020-09-21 10:41:07.1137078 +0200 CEST  deployed        kubeapps-3.9.2          v1.11.3
```

```
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.3", GitCommit:"2e7996e3e2712684bc73f0dec0200d64eec7fe40", GitTreeState:"clean", BuildDate:"2020-05-20T12:52:00Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.11", GitCommit:"b93345fce4b93610948e30f12c767f5dabd3d570", GitTreeState:"clean", BuildDate:"2020-08-24T20:09:56Z", GoVersion:"go1.12.17", Compiler:"gc", Platform:"linux/amd64"}
```

andresmgot:
### Description of the change

(sorry for the lengthy PR :pray:)

Fixes #2049 (the issue related to `undefined` being set in URLs). In summary, rather than relying (only) in a hardcoded list of plurals and apiVersions, resolve those using the Kubernetes API. This is used to resolve `kind`s to a valid URL. For example, an object with `kind: "Role", name: "foo"` is requested with the URL `.../apis/rbac.authorization.k8s.io/v1/namespaces/kubeapps/roles/foo`. Most of the changes are caused by the fact that  now this information is resolved asynchronously in the redux state.

There are two calls made:
 - `/apis` - Returns all the different groups and apiVersions
 - `/apis/<apiVersion>` - Returns, among other info, the plurals of the different resources included in that API group.

The different kinds and groups are stored in the `Kube` state.

Note that for maintaining backward compatibility, (e.g. for users that don't have permissions to access those endpoints), the state is populated with a set of default apiVersions and plurals.

### Benefits

Avoid wrong URLs.

### Possible drawbacks

If a user don't have permissions to retrieve the different API groups and the default info is no longer valid (e.g. a resource moved to a different version), the URLs will still be wrong.

Also, there is no feedback for the user if they cannot retrieve that information. I can address that in a follow up PR and show a warning.

### Applicable issues

  - fixes #2049


dashboard/src/shared/ResourceRef.ts
```diff
@@ -30,19 +30,29 @@ export function fromCRD(
...
   // Creates a new ResourceRef instance from an existing IResource. Provide
   // defaultNamespace to set if the IResource doesn't specify a namespace.
-  constructor(r: IResource, cluster: string, defaultNamespace?: string) {
+  constructor(
+    r: IResource,
+    cluster: string,
+    plural: string,
+    namespaced: boolean,
+    defaultNamespace?: string,
+  ) {
```
absoludity, line 48: I think this is the main change here - adding the extra data (plural, namespaced) to the constructor so that it can be used. What's not clear to me is why we need to use `plural` at all. I guess I'll wait for the non-draft PR with an explanation :) Thanks!

andresmgot: Yes. The `plural` is needed to build the URL from the `kind` (which is the only information that we have for sure). So we can resolve, for example, `kind: Role` -> `/apis/rbac.authorization.k8s.io/v1/namespaces/kubeapps/roles/<name>`

This was hardcoded at https://github.com/kubeapps/kubeapps/pull/2234/files/6f856ca662950a5d97f960825b23af58458eb761#diff-f6be03870e629fd9c812827f468e126171394db8ac54eea771d90cd6d8c0925a

absoludity: https://github.com/kubernetes/kubernetes/issues/18622

In particular:

```
The current impl (changed in November, I think) now pluralizes ListOptions as listoptionses, PodLogOptions as podlogoptionses, etc. I'm as big a LotR fan as anyone, but...
```
:)

absoludity (approved):

dashboard/src/shared/ResourceRef.ts
andresmgot: :sweat_smile: it would have been nice to not have plurals... even though we'd still need to discover the apiVersions and if the resources are namespaced or not.

Issue #1647, opened 2020-04-08
Move app repo sync to run in repo namespace (and cluster)


absoludity:
Currently, when creating an AppRepository, if the app repo namespace isn't kubeapps (ie. this is a per-namespace AppRepository), for a private repo we save a copy of the repository secret in kubeapps own namespace using the service account clientset. This enables the existing assetsync service to be able to sync private AppRepositories in other namespaces.

It is not ideal and is a temporary work-around until the asset-sync is updated to run cronjobs in other namespaces with the assetsvc receiving the data. See the [relevant section of the design doc for details](https://docs.google.com/document/d/1YEeKC6nPLoq4oaxs9v8_UsmxrRfWxB6KCyqrh2-Q8x0/edit?ts=5e2adf87#heading=h.kilvd2vii0w).

This issue is a placeholder to refer to in code, to ensure that we move the sync job to the user namespace (and send data from there to the assetsvc) and remove the code creating the copy.

ppbaena:
To be closed as it is not considered in the current roadmap for Kubeapps, and it's a long time since it was requested.

PR #2235, opened 2021-01-11
Fix namespaced initial repos

### LINKED ISSUE #2233: Namespaced initial repos with authorization not working
### Description:

We added an initial repo for a specific namespace, by having the following in our `values.yaml`:

```YAML
apprepository:
    initialRepos:
    -   name: bitnami
    -   name: my-private-repo
        url: https://my-private-repo.com
        authorizationHeader: "Bearer <REDACTED>"
        namespace: my-namespace
``` 

While the `bitnami` repository was successfully created, the app repo `my-private-repo` wasn't. 
We suspect this was due to missing secrets in both kubeapps' namespace and the target namespace i.e. `my-namespace`. 

Namespaced repos (without authorization) work, and global repos with authorization also work. The problem seems to come up we have a namespaced repositories _with_ authorization.

We suspect the culprit piece of code lies in the Helm chart, as this scenario wasn't updated when the namespaced repositories were introduced.

After installing kubeapps with the previous excerpt from `values.yaml`, we get a pod that is stuck in state `CreateContainerConfigError`. Upon running `kubectl describe <>` on said pod we see this error message:
```
Error: secret "my-namespace-apprepo-my-private-repo" not found
```
When we check for secrets in the kubeapps' namespace, we see no such secret indeed:
```
> kubectl get secrets -n kubeapps
NAME                                                          TYPE                                  DATA   AGE
apprepo-my-private-repo-secrets                               Opaque                                1      5m16s
default-token-99n9w                                           kubernetes.io/service-account-token   3      62m
kubeapps-db                                                   Opaque                                2      5m16s
kubeapps-internal-apprepository-controller-token-6g85j        kubernetes.io/service-account-token   3      5m16s
kubeapps-internal-apprepository-job-postupgrade-token-xfp79   kubernetes.io/service-account-token   3      5m16s
kubeapps-internal-kubeops-token-gtp8n                         kubernetes.io/service-account-token   3      5m16s
sh.helm.release.v1.kubeapps.v1                                helm.sh/release.v1                    1      5m16s
```
Additionally a secret should be present in the target namespace i.e. `my-namespace`, but upon inspection there is no such secret either:
```
> kubectl get secret -n my-namespace
NAME                  TYPE                                  DATA   AGE
default-token-r64sg   kubernetes.io/service-account-token   3      8m14s
```

1. Install chart with an initial repo that's namespaced and requires authorization.
```bash
helm install kubeapps bitnami/kubeapps --version 5.0.0 --namespace kubeapps --values values.yaml 
```
where `values.yaml` is:
```yaml
apprepository:
    initialRepos:
    -   name: bitnami
    -   name: my-private-repo
        url: https://my-private-repo.com
        authorizationHeader: "Bearer <REDACTED>"
        namespace: my-namespace
```

The private repository doesn't work at all.

I expected the namespaced private repository to work.

It works perfectly fine when manually added afterwards via the GUI.

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.4.1", GitCommit:"c4e74854886b2efe3321e185578e6db9be0a6e29", GitTreeState:"dirty", GoVersion:"go1.15.4"}
```

- Output of `helm list -n <kubeapps-namespace> `:

```
NAME    	NAMESPACE	REVISION	UPDATED                                	STATUS  	CHART         	APP VERSION
kubeapps	kubeapps 	1       	2021-01-08 15:10:59.216099396 +0100 CET	deployed	kubeapps-5.0.0	2.0.1      
```

```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.4", GitCommit:"d360454c9bcd1634cf4cc52d1867af5491dc9c5f", GitTreeState:"clean", BuildDate:"2020-11-12T10:37:08Z", GoVersion:"go1.15.4", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"clean", BuildDate:"2020-04-16T11:48:36Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
```

kristinnardal2:
### Description of the change

For namespaced repos we need to create two identical secrets containing
ca.crt and/or authorizationHeader. We need one in the namespace that the
app repository is scoped to and one in the kubeapps release namespace.

The name of the secret in the kubeapps release namespace is expected to
be of the format {{ .namespace }}-apprepo-{{ .name }} by the
apprepository controller.

Fixes #2233

### DIFF
--- a/chart/kubeapps/Chart.yaml
+++ b/chart/kubeapps/Chart.yaml
@@ -20,4 +20,4 @@ maintainers:
 name: kubeapps
 sources:
   - https://github.com/kubeapps/kubeapps
-version: 5.0.0
+version: 5.1.0
--- a/chart/kubeapps/Chart.yaml
+++ b/chart/kubeapps/Chart.yaml
@@ -20,4 +20,4 @@ maintainers:
 name: kubeapps
 sources:
   - https://github.com/kubeapps/kubeapps
-version: 5.1.0
+version: 5.0.1
--- a/chart/kubeapps/templates/apprepositories-secret.yaml
+++ b/chart/kubeapps/templates/apprepositories-secret.yaml
@@ -4,6 +4,9 @@ apiVersion: v1
 kind: Secret
 metadata:
   name: {{ template "kubeapps.apprepository-secret.name" . }}
+  {{- if .namespace }}
+  namespace: {{ .namespace }}
+  {{- end }}
   annotations:
     "helm.sh/hook": pre-install
   labels:{{ include "kubeapps.labels" $ | nindent 4 }}
@@ -17,5 +20,25 @@ data:
     {{ .authorizationHeader | b64enc }}
   {{- end }}
 ---
+{{/* credentials are required in the release namespace for syncer jobs */}}
+{{- if .namespace }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: "{{ .namespace }}-apprepo-{{ .name }}"
+  annotations:
+    "helm.sh/hook": pre-install
+  labels:{{ include "kubeapps.labels" $ | nindent 4 }}
+data:
+  {{- if .caCert }}
+  ca.crt: |-
+    {{ .caCert | b64enc }}
+  {{- end }}
+  {{- if .authorizationHeader }}
+  authorizationHeader: |-
+    {{ .authorizationHeader | b64enc }}
+  {{- end }}
+---
+{{- end }}
 {{- end }}
 {{- end }}


chart/kubeapps/Chart.yaml
```diff
@@ -20,4 +20,4 @@ maintainers:
 name: kubeapps
 sources:
   - https://github.com/kubeapps/kubeapps
-version: 5.0.0
```
andresmgot, line 23: we can consider this a bug fix, we can call this version 5.0.1

andresmgot (approved):
Thanks @kristinnardal2! Looks good to me. There is a minor nitpick issue: the name of the secret would be different in the apprepo namespace if the repo is created with the UI vs the chart (`apprepo-foo` vs `apprepo-foo-secrets`) but I think that's okay since the secret in the repo namespace doesn't need to have a fixed name. Am I right @absoludity?

chart/kubeapps/templates/apprepositories-secret.yaml
```diff
@@ -4,6 +4,28 @@ apiVersion: v1
...
+  {{- if .caCert }}
+  ca.crt: |-
+    {{ .caCert | b64enc }}
+  {{- end }}
+  {{- if .authorizationHeader }}
+  authorizationHeader: |-
+    {{ .authorizationHeader | b64enc }}
+  {{- end }}
+---
+{{/* credentials are required in the release namespace for syncer jobs */}}
```
absoludity, line 23: Until we get time to tackle #1647 :)

absoludity:
> Thanks @kristinnardal2! Looks good to me. There is a minor nitpick issue: the name of the secret would be different in the apprepo namespace if the repo is created with the UI vs the chart (`apprepo-foo` vs `apprepo-foo-secrets`) but I think that's okay since the secret in the repo namespace doesn't need to have a fixed name. Am I right @absoludity?

Yes. Ideally they'd be the same (that was an oversight when adding the per-namespace repos) but I don't see a problem (and it's less change to the chart as is).

kristinnardal2:
> > Thanks @kristinnardal2! Looks good to me. There is a minor nitpick issue: the name of the secret would be different in the apprepo namespace if the repo is created with the UI vs the chart (`apprepo-foo` vs `apprepo-foo-secrets`) but I think that's okay since the secret in the repo namespace doesn't need to have a fixed name. Am I right @absoludity?
> 
> Yes. Ideally they'd be the same (that was an oversight when adding the per-namespace repos) but I don't see a problem (and it's less change to the chart as is).

We (me and @latiif) considered changing the name to match the ones created by the UI but didn't really know if it was intended or not.

chart/kubeapps/templates/apprepositories-secret.yaml
kristinnardal2: That's a nice issue, would simplify this quite a bit!

Would you like this linked in the code or is it fine as it is?

andresmgot:
Let's leave the name as it is but can you change the version to `5.0.1` rather than `5.1.0` @kristinnardal2 ? That way me make clear that we are not adding new functionality to the chart, just fixing what was supposed to be there already.

kristinnardal2:
> Let's leave the name as it is but can you change the version to `5.0.1` rather than `5.1.0` @kristinnardal2 ? That way me make clear that we are not adding new functionality to the chart, just fixing what was supposed to be there already.

Sounds good, I've changed the version to `5.0.1` now.

PR #2236, opened 2021-01-12
Fix old reference to helm in cluster requirement.

absoludity:
### Description of the change

While browsing the getting started for other reasons, I noticed it still
refers to Helm as an in-cluster requirement (from the tiller days).

### Benefits

Don't confuse users.

### Possible drawbacks

None.


antgamdia (approved):
More clear now! Thanks for the update

andresmgot (approved):

PR #2238, opened 2021-01-12
Fix default value in forms when using hidden

### LINKED ISSUE #1913: Bug: Form "hidden" feature does not operate on default value
### Description:

I have the following scenario in my `values.schema.json`:

```
"dropdown": {
  "type": "string",
  "form": true,
  "default": "a",
  "enum": ["a", "b"]
},
"a": {
  "type": "string",
  "form": true,
  "hidden": {
    "path": "dropdown",
    "value": "b"
  }
},
"b": {
  "type": "string",
  "form": true,
  "hidden": {
    "path": "dropdown",
    "value": "a"
  }
}
```

When the form page is loaded, both `a` and `b` will **NOT** be hidden.

I would expect `b` to be hidden, since the dropdown has a default value of `a`.  However, it does not appear that the form is recognizing the default value of this dropdown and marking `b` as hidden.

I have not tested if this is solely an interaction that occurs with `enum`. It could possibly occur with other default values as well.

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.2.4", GitCommit:"0ad800ef43d3b826f31a5ad8dfbb4fe05d143688", GitTreeState:"clean", GoVersion:"go1.13.12"}
```

```
CHART         	APP VERSION
kubeapps-3.7.2	v1.10.2    
```

```
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.0+724e12f93f", GitCommit:"7207690010f08f2ae526ee3300d4a18ea65997fe", GitTreeState:"clean", BuildDate:"2020-01-06T18:32:40Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.5", GitCommit:"e6503f8d8f769ace2f338794c914a96fc335df0f", GitTreeState:"clean", BuildDate:"2020-06-26T03:39:24Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
```

antgamdia:
### Description of the change

This PR simply adds a new check when evaluating whether or not a form field should be rendered. If the current `getValue(appValues, path)` is undefined (that is, there is no actual value in the `values.yaml` file) it tries to obtain the value form the hidden property (via `param.hidden?.value`).

### Benefits

Users won't have to specify a value in `values.yaml` if there already exists a default value in `hidden.value` of the `values.schema.json`

### Possible drawbacks

N/A

### Applicable issues

  - fixes #1913

### Additional information

N/A

### DIFF
--- a/dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/Param.tsx
+++ b/dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/Param.tsx
@@ -72,7 +72,13 @@ export default function Param({
     paramDeploymentEvent?: DeploymentEvent,
   ): boolean => {
     if (paramDeploymentEvent == null) {
-      return getValue(appValues, path) === (expectedValue ?? true);
+      let val = getValue(appValues, path);
+      // no value is provided, but there is a default one in the schema (hidden.value)
+      // https://github.com/kubeapps/kubeapps/issues/1913
+      if (!val && typeof param.hidden === "object") {
+        val = param.hidden?.value;
+      }
+      return val === (expectedValue ?? true);
     } else {
       return paramDeploymentEvent === deploymentEvent;
     }
--- a/dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/Param.tsx
+++ b/dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/Param.tsx
@@ -10,8 +10,8 @@ import TextParam from "./TextParam";
 interface IParamProps {
   appValues: string;
   param: IBasicFormParam;
+  allParams: IBasicFormParam[];
   id: string;
-  otherParams: IBasicFormParam[];
   handleBasicFormParamChange: (
     p: IBasicFormParam,
   ) => (e: React.FormEvent<HTMLInputElement | HTMLTextAreaElement | HTMLSelectElement>) => void;
@@ -22,7 +22,7 @@ interface IParamProps {
 export default function Param({
   appValues,
   param,
-  otherParams,
+  allParams,
   id,
   handleBasicFormParamChange,
   handleValuesChange,
@@ -68,6 +68,19 @@ export default function Param({
     }
   };
 
+  const getParamMatchingPath = (params: IBasicFormParam[], path: string): any => {
+    let targetParam;
+    for (const p of params) {
+      if (p.path === path) {
+        targetParam = p;
+        break;
+      } else if (p.children && p.children?.length > 0) {
+        targetParam = getParamMatchingPath(p.children, path);
+      }
+    }
+    return targetParam;
+  };
+
   const evalCondition = (
     path: string,
     expectedValue?: any,
@@ -78,7 +91,7 @@ export default function Param({
       // retrieve the value that the property pointed by path should have to be hidden.
       // https://github.com/kubeapps/kubeapps/issues/1913
       if (val === undefined) {
-        const target = otherParams.find(p => p.path === path);
+        const target = getParamMatchingPath(allParams, path);
         val = target?.value;
       }
       return val === (expectedValue ?? true);
@@ -105,6 +118,7 @@ export default function Param({
         handleValuesChange={handleValuesChange}
         appValues={appValues}
         param={param}
+        allParams={allParams}
         deploymentEvent={deploymentEvent}
       />
     );
--- a/dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/Param.tsx
+++ b/dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/Param.tsx
@@ -11,6 +11,7 @@ interface IParamProps {
   appValues: string;
   param: IBasicFormParam;
   id: string;
+  otherParams: IBasicFormParam[];
   handleBasicFormParamChange: (
     p: IBasicFormParam,
   ) => (e: React.FormEvent<HTMLInputElement | HTMLTextAreaElement | HTMLSelectElement>) => void;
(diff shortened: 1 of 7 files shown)


dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/Param.tsx
```diff
@@ -72,7 +72,13 @@ export default function Param({
     paramDeploymentEvent?: DeploymentEvent,
   ): boolean => {
     if (paramDeploymentEvent == null) {
-      return getValue(appValues, path) === (expectedValue ?? true);
+      let val = getValue(appValues, path);
+      // no value is provided, but there is a default one in the schema (hidden.value)
+      // https://github.com/kubeapps/kubeapps/issues/1913
+      if (!val && typeof param.hidden === "object") {
+        val = param.hidden?.value;
+      }
```
andresmgot, line 80: `hidden.value` is not the default value, it's the value that the property pointed by `path` should have to be hidden. Here you are saying that the value of the property is whatever makes it hidden so it will always be true :) 

I was wrong in the issue description because I thought that `param` was the parameter pointed by `path` but it's not. You need to retrieve that `param`. To do so, you need to modify the properties of this component and include the rest of the parameters, that way you can look for it. I tried this code and it works:

```js
  const evalCondition = (
    path: string,
    expectedValue?: any,
    paramDeploymentEvent?: DeploymentEvent,
  ): boolean => {
    if (paramDeploymentEvent == null) {
      let val = getValue(appValues, path);
      if (val === undefined) {
        const target = otherParams.find(p => p.path === path);
        val = target?.value;
      }
      return val === (expectedValue ?? true);
    } else {
      return paramDeploymentEvent === deploymentEvent;
    }
  };
```

Also note the condition, I am using `val === undefined` rather than `!val` since for example `val = false` is a valid value.

andresmgot (changes requested):
This is not doing what you think :) I have a suggestion. Also you will need to write some tests, the tests for the `Param` component are in the `BasicDeploymentForm.test.tsx` file (for historical reasons :)).

If you want me to give you a hand with this let me know, I can send the PR myself if you are busy.

dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/Param.tsx
antgamdia: Mmm, ok, now I see... I misunderstood what `hidden.value` was actually doing. Thanks for the clarification!

>  Also you will need to write some tests, the tests for the Param component are in the BasicDeploymentForm.test.tsx file (for historical reasons :)).

I looked for a `Param.test.tsx` and didn't find anything; thanks for letting me know :). Adding a new test case.

> I am using val === undefined rather than !val ...

I thought val=false wasn't valid, but now, having a better understanding of what `hidden.value` does, I know it was.

dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/Param.tsx
```diff
@@ -11,6 +11,7 @@ interface IParamProps {
   appValues: string;
   param: IBasicFormParam;
   id: string;
+  otherParams: IBasicFormParam[];
```
andresmgot, line 14: nitpick: The name `otherParams` implies that the property excludes the current param but it's not excluding it. Probably `allParams` is a better name.

dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/Subsection.tsx
```diff
@@ -46,6 +46,7 @@ function Subsection({
           return (
             <Param
               param={childrenParam}
+              otherParams={param.children ? param.children : []}
```
andresmgot, line 49: you need to modify here the component as well and include `allParams` as a property to pass it down here.

dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/Param.tsx
```diff
@@ -72,7 +74,14 @@ export default function Param({
     paramDeploymentEvent?: DeploymentEvent,
   ): boolean => {
     if (paramDeploymentEvent == null) {
-      return getValue(appValues, path) === (expectedValue ?? true);
+      let val = getValue(appValues, path);
+      // retrieve the value that the property pointed by path should have to be hidden.
+      // https://github.com/kubeapps/kubeapps/issues/1913
+      if (val === undefined) {
+        const target = otherParams.find(p => p.path === path);
```
andresmgot, line 81: mmh, just noticed that the params are not a flat array but it has nested params (for the case of a subsection). You need to add some recursion here, if not this would only detect top-level params.

andresmgot (commented):
Thanks @antgamdia! Just noticed a case we should cover to complete the issue.

dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/Subsection.tsx
antgamdia: Mmm, I think I'm not getting what you mean here. Do you mean creating the `allParams` property in the `ISubsectionProps` ?

My idea was to pass as `allParams` every children param as part of the subsection in order to make non-top-level params to work as well.
I used the example (with two levels) I wrote in https://github.com/kubeapps/kubeapps/issues/1913#issuecomment-750844297 and I worked fine.

dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/Param.tsx
antgamdia: That's true, totally agree.

antgamdia:
Thank you for your comments! Now it should work even if we mix different depth levels, for instance:

hidden

andresmgot (approved):
great, thanks :)

PR #2241, opened 2021-01-13
Add OIDC debugging info to the doc.

### LINKED ISSUE #2208: FAQ entry or otherwise document steps for debugging oidc logins
### Description:

We often get questions about how to debug oidc logins where the user logs in to the SSO successfully but Kubeapps says it's a 401.

These are actually easy once you know how, but we really need to document the know-how :) (probably at https://github.com/kubeapps/kubeapps/blob/master/docs/user/using-an-OIDC-provider.md ).

First, if you temporarily set the `--set-authorization-header=true` option for oauth2 proxy, the actual `id_token` will be included in the Authorization header of the response back to the client, so you can then:

1) base64 decode the token (or paste into https://jwt.io if it's a local throw-away credential for testing) to verify the token has the correct user/groups etc. that you expect, and
2) use the token with `curl -H "Authorization: Bearer $TOKEN" ...` to reproduce the issue without depending on Kubeapps etc.

Once you can reproduce, there are a couple of possibilities for the cause which commonly trip people up:

1. Sometimes people don't have a k8s cluster api server even configured for oidc (not realising that was necessary). Easy to check the api server oidc args with something like:
```
k -n kube-system get po -l component=kube-apiserver -o yaml | grep oidc
      - --oidc-ca-file=/etc/kubernetes/pki/apiserver.crt
      - --oidc-client-id=default
      - --oidc-groups-claim=groups
      - '--oidc-groups-prefix=oidc:'
      - --oidc-issuer-url=https://172.18.0.2:32000
      - --oidc-username-claim=email
      - '--oidc-username-prefix=oidc:'
```

2. The most common problem is that the api server is using an `oidc-username-prefix` (whether explicitly specified or the default - the [docs specify that the default for both the username and group prefix is `oidc:`](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#configuring-the-api-server) but I've always explicitly defined it). So for a token identifying a user of `myname@example.com`, the RBAC would need to be for the user `oidc:myname@example.com` .

absoludity:
### Description of the change

As per #2208, add debugging instructions to help resolve OIDC login issues with Kubeapps.

### Benefits

People can resolve some issues themselves more easily.

### Applicable issues

  - fixes #2208


docs/user/using-an-OIDC-provider.md
```diff
@@ -241,3 +241,59 @@ spec:
...
 EOF
 ```
+
+## Debugging auth failures when using OIDC
+
+If you find after configuring your OIDC/OAuth2 setup following the above instructions, that although you can successfully authenticate with your provider you are nonetheless unable to login to Kubeapps but instead see a 403 or 401 request in the browser's debugger, then you will need to investigate *why* the Kubernetes cluster is not accepting your credential.
+
+### Viewing the JWT id token
+
+The easiest way to check the credential that is being used is to temporarily set the `--set-authorization-header=true` option for oauth2 proxy of the kubeapps deployment. Edit the `kubeapps` deployment in the specific namespace on your cluster and add the option to the command for the oauth2 proxy container. Once the deployment runs a new container with the extra option, Kubeapps will then include the `id_token` that is being used to authenticate you with the Kubernetes api server in the response back to the browser.
```
antgamdia, line 251: Perhaps I'd show an example of how to set this flag, maybe using `kubectl patch deployment kubeapps ...` or directly with helm flags, so that users can go straight to add the flag without thinking of in which deployment or place they should set the flag.

docs/user/using-an-OIDC-provider.md
```diff
@@ -241,3 +241,59 @@ spec:
...
+
+To view the token, in your browser debugger's Requests tab, watch for the request to `/api/clusters/default` or similar which will have a 40X status. Click on this request to view the headers and in the Response headers look for the `Authentication` header. The value here will be the base64-encoded `id_token`. Copy the value.
```
antgamdia, line 253: Do you mean `Authorization` instead of `Authentication` header?  

Perhaps (my former teacher's point of view... :P), it would be useful to have a screenshot for helping users to understand what they should copy and paste. Some people copy `Authorization: Bearer xxxxx`, other `Bearer xxxxx` and other `xxxxx`.

antgamdia (approved):

docs/user/using-an-OIDC-provider.md
andresmgot, line 251: > Edit the `kubeapps` deployment

People often get confused with the deployment names. I would specify to ignore the ones called `kubeapps-internal-*`. Also, if you can include a YAML snippet with an example of an added flag I think it will be much appreciated.

docs/user/using-an-OIDC-provider.md
andresmgot, line 253: > Requests tab,

I chrome is called "Network" tab.

Also, as above, a screenshot with an example of an `id_token` would be ideal.

andresmgot (approved):
Great guide! I have a couple of suggestions to make it even more helpful. Also, to make it more discoverable, you can link this guide from the FAQ in the chart README

docs/user/using-an-OIDC-provider.md
absoludity: Good idea. Done.

absoludity: Thanks, yep, I've removed that now in place of a copy-n-pastable patch.

docs/user/using-an-OIDC-provider.md
absoludity: Yes I did, and good point - done.

absoludity: It's "Network" in firefox too - fixed.

PR #2242, opened 2021-01-13
Remove AppView container, use react hooks

andresmgot:
### Description of the change

Refactor AppView to avoid using `containers` in favor of react hooks.

### Benefits

Remove one TODO :)


antgamdia (approved):
So the main modifications are:

1) Change `mapStateToProps` to something like `... = useSelector((state: IStoreState) => state);`

2) Change `mapDispatchToProps` simply `useDispatch(...)' 

3) Modify `/Routes.tsx` to point to the proper file

I'll try to do the same for the Catalog view as soon as I finish the other PRs.

absoludity (approved):

andresmgot:
> So the main modifications are:
> 
> 1. Change `mapStateToProps` to something like `... = useSelector((state: IStoreState) => state);`
> 2. Change `mapDispatchToProps` simply `useDispatch(...)`
> 3. Modify `/Routes.tsx` to point to the proper file
> 
> I'll try to do the same for the Catalog view as soon as I finish the other PRs.

Yes, also I am using `useParams()` to retrieve the parameters in the URL. Most of the changes are related to the tests, since it's not trivial to mock all these hooks.

PR #2244, opened 2021-01-13
Use ChartQuery in getChartCategories

antgamdia:
### Description of the change

`getAllChartCategories` method was not using the `ChartQuery` object to receive the information.. Also, this PR allow using both `?repo` and  `?repos` for backwards compatibility.

### Benefits

Retrieving all distinct categories will work even if we choose a particular repo (o list of repos).

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

### DIFF
--- a/cmd/assetsvc/handler.go
+++ b/cmd/assetsvc/handler.go
@@ -90,8 +90,8 @@ func getPageAndSizeParams(req *http.Request) (int, int) {
 	return int(pageNumberInt), int(pageSizeInt)
 }
 
-func getAllChartCategories(namespace, repo string) (apiChartCategoryListResponse, error) {
-	chartCategories, err := manager.getAllChartCategories(namespace, repo)
+func getAllChartCategories(cq ChartQuery) (apiChartCategoryListResponse, error) {
+	chartCategories, err := manager.getAllChartCategories(cq)
 	return newChartCategoryListResponse(chartCategories), err
 }
 
@@ -109,7 +109,12 @@ func getChartCategories(w http.ResponseWriter, req *http.Request, params Params)
 		return
 	}
 
-	chartCategories, err := getAllChartCategories(namespace, repo)
+	cq := ChartQuery{
+		namespace: namespace,
+		repos:     append(strings.Split(strings.TrimSpace(req.FormValue("repos")), ","), repo),
+	}
+
+	chartCategories, err := getAllChartCategories(cq)
 	if err != nil {
 		log.WithError(err).Error("could not fetch categories")
 		response.NewErrorResponse(http.StatusInternalServerError, "could not fetch chart categories").Write(w)
--- a/cmd/assetsvc/handler.go
+++ b/cmd/assetsvc/handler.go
@@ -90,6 +90,57 @@ func getPageAndSizeParams(req *http.Request) (int, int) {
 	return int(pageNumberInt), int(pageSizeInt)
 }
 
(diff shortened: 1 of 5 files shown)


cmd/assetsvc/handler.go
```diff
@@ -109,7 +109,12 @@ func getChartCategories(w http.ResponseWriter, req *http.Request, params Params)
 		return
 	}
 
-	chartCategories, err := getAllChartCategories(namespace, repo)
+	cq := ChartQuery{
+		namespace: namespace,
+		repos:     append(strings.Split(strings.TrimSpace(req.FormValue("repos")), ","), repo),
+	}
```
absoludity, line 115: Perhaps if you extract the above 15 lines into something like `extractChartQueryFromRequest()` it'd be easy to test in isolation? As it is, I think we need a test for the repos - it looks like you're creating a list by appending the current repo (from the url) with those from the submitted form data, but we may (?) want to ensure that is unique? Anyway, with that tested, +1

absoludity (approved):

cmd/assetsvc/handler.go
antgamdia: Good idea indeed. I still think we should improve and normalize how we retrieve the params. It's such a mess being mixing up params from `req.FormValue("foo")` and from `params["foo"]`... but, yep, for this PR, I totally agree with you.

Regarding the uniqueness, it won't have any side effect if we pass dups since, in the end, it will become something like `WHERE ...OR repo=X OR repo=X`

andresmgot (commented):
I am okay if you want to add the tests @absoludity mentions and merge this PR but as I said in https://github.com/kubeapps/kubeapps/pull/2246 I think we don't need to filter categories by repository (only by namespace).

antgamdia:
I've extracted this chartQuery generation logic and the decoding one as well. Consequently, I've added two new test cases.
Even if we decide not to filter by repo, this refactor change could be still useful.

cmd/assetsvc/handler.go
```diff
@@ -90,26 +90,72 @@ func getPageAndSizeParams(req *http.Request) (int, int) {
...
-func getAllChartCategories(namespace, repo string) (apiChartCategoryListResponse, error) {
-	chartCategories, err := manager.getAllChartCategories(namespace, repo)
+func extractDecodedNamespaceAndRepoAndVersionParams(params Params) (string, string, string, []string, []error) {
+	paramErr := []string{}
+
+	namespace, namespaceErr := url.PathUnescape(params["namespace"])
+	repo, repoErr := url.PathUnescape(params["repo"])
+	version, versionErr := url.PathUnescape(params["version"])
+
+	if namespaceErr != nil || repoErr != nil || versionErr != nil {
```
andresmgot, line 100: the idiomatic thing to do is to early return if there is an error, returning an empty response and the error:

```
namespace, err := url.PathUnescape(params["namespace"])
if err != nil {
  return "", "", "", params["namespace"], err
}
```

(I know you are just grouping errors, but the extra complexity is not necessary)

andresmgot (approved):
LGTM, I would simplify this a bit but up to you.

cmd/assetsvc/handler.go
antgamdia: Yes, I kinda agree. It is not a big deal not to display every parameter with decoding errors and the complexity is not necessary. I've just updated the PR with this suggestion, thanks!

andresmgot (approved):
Thanks!

PR #2245, opened 2021-01-13
Fix unhandled rejection

antgamdia:
### Description of the change

Very minor fix of an unhandled rejection.

### Benefits

If this error is thrown, the app will show a friendly message.

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

- Change initially introduced in #2221 but removed from there to keep it as simple as possible.

### DIFF
--- a/dashboard/src/actions/repos.ts
+++ b/dashboard/src/actions/repos.ts
@@ -185,8 +185,12 @@ export const fetchRepoSecret = (
     const {
       clusters: { currentCluster },
     } = getState();
-    const secret = await Secret.get(currentCluster, namespace, name);
-    dispatch(receiveReposSecret(secret));
+    try {
+      const secret = await Secret.get(currentCluster, namespace, name);
+      dispatch(receiveReposSecret(secret));
+    } catch (e) {
+      dispatch(errorRepos(e, "fetch"));
+    }
   };
 };
 
--- a/dashboard/src/actions/repos.ts
+++ b/dashboard/src/actions/repos.ts
@@ -162,14 +162,18 @@ export const fetchRepoSecrets = (
     const {
       clusters: { currentCluster },
     } = getState();
-    // TODO(andresmgot): Create an endpoint for returning credentials related to an AppRepository
-    // to avoid listing secrets
-    // https://github.com/kubeapps/kubeapps/issues/1686
-    const secrets = await Secret.list(currentCluster, namespace);
-    const repoSecrets = secrets.items?.filter(s =>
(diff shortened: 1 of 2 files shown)


absoludity (approved):
Not necessary, but if there is already a test case for this, it'd be great to document the change with a test (normally I'd write the test first to see it fail, then fix, for these small errors). Anyway, fine as is.

antgamdia:
Thank you, I've added a couple of tests and, because of that, I've fixed another rejection :)

dashboard/src/actions/repos.ts
```diff
@@ -162,14 +162,18 @@ export const fetchRepoSecrets = (
...
+      // to avoid listing secrets
+      // https://github.com/kubeapps/kubeapps/issues/1686
+      const secrets = await Secret.list(currentCluster, namespace);
+      const repoSecrets = secrets.items?.filter(s =>
+        s.metadata.ownerReferences?.some(ownerRef => ownerRef.kind === "AppRepository"),
+      );
+      dispatch(receiveReposSecrets(repoSecrets));
+    } catch (e) {
+      dispatch(errorRepos(e, "fetch"));
+    }
```
andresmgot, line 176: This has a caveat. There is a situation in which "it's okay" to fail retrieving repo secrets. If you are in a namespace, this function would be called twice, one for the current namespace and another to retrieve repos in the `kubeapps` namespace (to show global repositories, read-only). In that case, non-privileged users won't be able to retrieve secrets from the `kubeapps` repository but we don't want to fail because that's fine (people need to go to the `kubeapps` namespace to manage global repos).

What I would do, to avoid the unhandled rejection and allow that situation, is adapt `fetchRepos` (line 190) for the case in which several namespaces are given and avoid calling `fetchRepoSecrets` for the kubeapps namespace. Ping me if you need more info about this.

andresmgot (changes requested):
Sorry, there is a feature taking advantage of this bug :)

dashboard/src/actions/repos.ts
andresmgot: Nevermind, I was double-checking this when I found out that `fetchRepos` is luckily not calling `fetchRepoSecrets` with the global namespace (because of the order of the params). This is highly weak but nothing to do with this PR, it can be merged as it is.

andresmgot (approved):
Nevermind, I just found out that we are not hitting the bug I described. I will send a PR myself making this more robust but you can merge the PR as it is.

dashboard/src/actions/repos.ts
antgamdia: Great! Thank you for having a look at this issue!! Merging then :)

PR #2246, opened 2021-01-13
Retrieve chart categories using the backend

antgamdia:
### Description of the change

Currently, the set of chart categories is generated by iterating over the whole list of charts, getting every category, selecting the unique ones and sorting them.
This PR is simply to move this logic to the backend by using the already existing endpoint.

### Benefits

The number of operations to render the catalog (especially when a repo is large) will be reduced.

### Possible drawbacks
~~This PR (at the moment) is not filtering by *operator* categories, so, as a consequence filtering operator categories will stop working.~~ Removing the dynamic loading depending on which repo is select should avoid this problem :)

### Applicable issues

N/A

### Additional information

Change initially introduced in #2221 but removed from there to keep it as simple as possible.


dashboard/src/actions/charts.ts
```diff
@@ -80,6 +95,26 @@ export function fetchCharts(
   };
 }
 
+export function fetchChartCategories(
+  cluster: string,
+  namespace: string,
+  repos: string,
+): ThunkAction<Promise<IChartCategory[]>, IStoreState, null, ChartsAction> {
```
andresmgot, line 102: since you are dispatching the categories (and later or retrieving them from the redux state), you don't need this function to return the array: `ThunkAction<Promise<void>`

dashboard/src/components/Catalog/Catalog.tsx
```diff
@@ -124,15 +132,16 @@ function Catalog(props: ICatalogProps) {
...
+    categories
+      .map(c => categoryToReadable(c.name))
       .concat(flatten(csvs.map(c => getOperatorCategories(c)))),
   ).sort();
 
   useEffect(() => {
     fetchCharts(cluster, namespace, repo);
     getCSVs(cluster, namespace);
-  }, [cluster, namespace, repo, fetchCharts, getCSVs]);
+    fetchChartCategories(cluster, namespace, currentRepo);
```
andresmgot, line 143: I don't think we should remove categories if a repository is selected (to only show categories of that repository). It would be a weird UX. Some examples:

https://artifacthub.io/packages/search?page=1&kind=3
https://operatorhub.io/?provider=%5B%22Alibaba+Cloud%22%5D

You can see that all the categories are still shown even if the current filters don't contain all of them.

dashboard/src/actions/charts.ts
```diff
@@ -80,6 +95,26 @@ export function fetchCharts(
...
+  return async dispatch => {
+    dispatch(requestChartsCategories());
+    try {
+      const categories = await Chart.fetchChartCategories(cluster, namespace, repos);
+      if (categories) {
+        dispatch(receiveChartCategories(categories));
+      }
+      return categories;
+    } catch (e) {
+      dispatch(errorChart(new FetchError(e.message)));
```
andresmgot, line 112: It would be better to create a new type of error for this (e.g. `errorChartCagetories`). Note that if you receive an error here, you need to clean up the state for `categories` in the `reducer`.

andresmgot (commented):
Thanks for splitting this @antgamdia ! Much easier to review and give, hopefully, helpful comments :)

dashboard/src/components/Catalog/Catalog.tsx
```diff
@@ -124,15 +132,16 @@ function Catalog(props: ICatalogProps) {
   const allRepos = uniq(charts.map(c => c.attributes.repo.name));
   const allProviders = uniq(csvs.map(c => c.spec.provider.name));
   const allCategories = uniq(
-    charts
-      .map(c => categoryToReadable(c.attributes.category))
+    categories
+      .map(c => categoryToReadable(c.name))
       .concat(flatten(csvs.map(c => getOperatorCategories(c)))),
```
andresmgot, line 137: also, forgot to mention, right now, we have a category `Unknown` for charts that don't include a category. You will need to add it to the `categories` list and remove it from `categoryToReadable`

antgamdia: > You will need to add it to the `categories` list and remove it from `categoryToReadable`:

Mmm, I'm not getting this. Currently, `categoryToReadable` is doing `return category === "" ? "Unknown" : ...`
And the server's response is `{"data":[{"name":"","count":2027}]}`. So, imho, it's seamlessly working.

Why do you think we should change it?

dashboard/src/components/Catalog/Catalog.tsx
antgamdia: Mmm, I get your point but I don't any strong opinion about that. 
On the one hand, leaving the whole list of categories would result in a more consistent view, since users won't see any disappearing elements.
But..., on the other hand, considering users having their repos without categories, when they suddenly add, let's say, bitnami repo, a list of categories will appear and they won't be able to get rid of them.

We are returning the number of charts in each category, perhaps we can display this number to prevent users from clicking on non-existent categories.

I don't know, I'm open to both options as I mentioned. What's your opinion, @absoludity ?

dashboard/src/components/Catalog/Catalog.tsx
andresmgot: ah, okay, I thought that the API was not returning a value for `name: ""`. If that's the case then it's okay.

dashboard/src/components/Catalog/Catalog.tsx
andresmgot: Let me put it the other way. If you first select a category (e.g. "CMS") and then select a repo that doesn't contain that category (e.g. "my-repo"), this would remove the category from list. 

In that scenario, the user cannot deselect the category, so they would be stuck with an empty list of apps. If to fix that you automatically remove the category filter, the whole list of applications will be shown but the user is still expecting CMS apps because they have not interacted with that filter again.

(hope I explained myself)

antgamdia: Thank you for the clarification. However, I think I'm missing something beacause I believe the scenario you mention will still work. Here is a gif with the current state (ignore the catalog empty message):

Even if the CMS categories continue being selected, since it follows an OR logic, as soon as the user clicks on an existing category, charts will appear.

filtersCategories

absoludity: Thanks for the gif.

> Even if the CMS categories continue being selected, since it follows an OR logic, as soon as the user clicks on an existing category, charts will appear.

and importantly, the currently active filters (`Category: CMS` and `Repository: harbor`) are still displayed above and so can still be deselected individually.

> I don't know, I'm open to both options as I mentioned. What's your opinion, @absoludity ?

Same - I don't feel strongly either way and so, as always in that circumstance, I'd defer to people who do. Although I do think it'd be a simpler UX to only see the categories that are relevant (given that, as above, there is no issue deselecting active filters which are no longer relevant) there are others who argue that people will be looking to the place where they turned a filter on to be able to turn it off again and so recommend having *both* visible (see Be redundant when displaying applied filters at in this [Filtering best practises article](https://uxdesign.cc/crafting-a-kickass-filtering-ux-beea1798d64b)). So I'd go with Andres' recommendation (unless you feel less strongly Andres when you see that you can deselect them trivially).

antgamdia: Thank you for your comments and for sharing the article; from this point of view, it kinda makes sense not to hide unuseful categories.
Moreover, I've just discovered another reason for going this way: these blue top level filters (`filter-summary`) are **not** accessible: keyboard navigation is not possible, so people with visual problems won't be able to remove the filters from this section. Hence, this redundancy is not being guaranteed for everybody, wherefore I also agree with the initial Andrés' recommendation.

antgamdia:
This PR should be now ready for review. The look&feel after this PR should be the same since it simply avoids getting the chart categories from the chart array and instead performs an API call to do so.

A change required in subsequent PR is to perform initial loading actions detecting the first time the component is rendered. AFAIK, `useEffect` always is monitoring for changes in the dependencies array.

dashboard/src/actions/charts.test.tsx
```diff
@@ -76,6 +76,66 @@ describe("fetchCharts", () => {
...
+    const expectedActions = [
+      { type: getType(actions.charts.requestChartsCategories) },
+      { type: getType(actions.charts.receiveChartCategories), payload: response },
+    ];
+    await store.dispatch(actions.charts.fetchChartCategories(cluster, namespace, "foo,bar"));
+    expect(store.getActions()).toEqual(expectedActions);
+    expect(axiosGetMock.mock.calls[0][0]).toBe(
+      `api/assetsvc/v1/clusters/${cluster}/namespaces/${namespace}/charts/categories?repos=foo,bar`,
+    );
+  });
```
andresmgot, line 104: since you are not finally using the `repo` as an argument, you can remove this related code

dashboard/src/components/Catalog/Catalog.tsx
```diff
@@ -73,30 +74,37 @@ function Catalog(props: ICatalogProps) {
...
     csvs,
     repo,
     filter: propsFilter,
   } = props;
   const dispatch = useDispatch();
   const [filters, setFilters] = useState(initialFilterState());
+  const initialRepoValue = propsFilter[filterNames.REPO]
+    ? String(propsFilter[filterNames.REPO])
+    : "";
+  const [currentRepo, setCurrentRepo] = useState(initialRepoValue);
```
andresmgot, line 93: same, you don't need this code, no?

dashboard/src/components/Catalog/Catalog.tsx
```diff
@@ -124,15 +132,22 @@ function Catalog(props: ICatalogProps) {
...
-  }, [cluster, namespace, repo, fetchCharts, getCSVs]);
+    if (firstUpdate.current) {
+      // actions when the component is mounted for the first time
+      firstUpdate.current = false;
+      setCurrentRepo("");
+      fetchChartCategories(cluster, namespace);
+      fetchCharts(cluster, namespace, repo);
+      getCSVs(cluster, namespace);
+    }
+  }, [dispatch, getCSVs, fetchCharts, fetchChartCategories, cluster, namespace, currentRepo, repo]);
```
andresmgot, line 150: I don't understand these changes, can't you just add `fetchChartCategories`? Why is the `firstUpdate` check needed?

andresmgot (commented):
Thanks @antgamdia! It's looking good so far, I just have some comments because I think we can simplify this even further.

dashboard/src/components/Catalog/Catalog.tsx
antgamdia: After looking at it a little bit closer,  `firstUpdate` can be removed and replaced by better handling of the state properties.

dashboard/src/components/Catalog/Catalog.tsx
antgamdia: Not in this PR, but the point is that `repo` property is undefined, so just created this prop in the component's state. It will be useful for filtering charts from a specific repo, but to keep this PR as simple as possible, I'm temporarily removing this code.

andresmgot (approved):
Thanks! LGTM

PR #2247, opened 2021-01-14
Ensure containers run as as non-root.

absoludity:
### Description of the change

When running Kubeapps on a TMC cluster with TMC's default settings all our services errored with:

```
  Warning  Failed     1s (x5 over 29s)  kubelet, ip-10-0-1-198.eu-west-1.compute.internal  Error: container has runAsNonRoot and image will run as root
```

### Benefits

If CI doesn't find any side effects, it means Kubeapps can run in security contexts requiring non-root containers without any extra config.

### Possible drawbacks

See what CI says, otherwise none.


antgamdia (approved):
Nice, it seems CI agrees with the change :)

andresmgot (approved):
FTR, the bitnami images (used in the published chart) already contains this change. For example:
https://github.com/bitnami/bitnami-docker-kubeapps-apprepository-controller/blob/master/2/scratch/Dockerfile#L8

My +1 to this change to make these images closer to the ones in the chart.

PR #2248, opened 2021-01-14
Allow to inject CSS code

### LINKED ISSUE #1366: [Question/Feature] Pluggable SCSS styling?

andresmgot:
### Description of the change

Allow to inject CSS code from using the chart. This CSS code is dumped to a file and mounted in the `kubeapps-dashboard` service so it's served on its own.

Apart from that, this PR moves the logo configuration to a CSS file so it can be overwritten too.

Note that even though people can add CSS code, this is not a complete "theme" system (which would require much more work, like parameterizing the current colors, fonts...). 

### Benefits

People can tune the current style.

### Applicable issues

  - fixes #1366

### Additional information

Example of a nice customization:

Screenshot from 2021-01-14 13-04-26


antgamdia (approved):
Interesting! As you mentioned, it is an initial work that will cover some styling requirements, but it is not a complete theming solution. 

I don't know if we should also mount the icons (https://github.com/kubeapps/kubeapps/tree/master/dashboard/public) so that people can modify them.

Finally, I don't know either injecting css by mounting it could potentially have any undesired side (security) effect, like adding some arbitrary code to be executed. Just speaking my mind, anyway.

andresmgot:
> I don't know if we should also mount the icons (https://github.com/kubeapps/kubeapps/tree/master/dashboard/public) so that people can modify them.

Yeah, that could be an additional thing to extract or parameterize but it's difficult to know what to put without requirements. With this as a base, we can start adding things in the future.

> Finally, I don't know either injecting css by mounting it could potentially have any undesired side (security) effect, like adding some arbitrary code to be executed. Just speaking my mind, anyway.

It should not because it requires access to the Kubernetes cluster but let me bring some other reviewer in case there is something I'm not seeing.

dashboard/src/components/Header/Header.tsx
```diff
@@ -47,11 +46,9 @@ function Header() {
...
       <div className="container">
         <header className="header header-7">
-          <div className="branding">
-            <NavLink to="/">
-              <img src={logo} alt="Kubeapps logo" className="kubeapps__logo" />
-            </NavLink>
-          </div>
+          <NavLink to="/">
+            <div className="kubeapps__logo" />
+          </NavLink>
```
Angelmmiguel, line 51: This is causing the link not to be accessible because it doesn't contain any text inside. If you want to proceed with this solution, I would add a hidden text so screen readers can read at least "Homepage".

This is a common implementation of a `.sr-only` CSS class:

```
.sr-only {
	height: 1px;
	margin: -1px;
	overflow: hidden;
	padding: 0;
	position: absolute;
	width: 1px;
	white-space: nowrap;
}
```

Then, you can use it inside:

```suggestion
          <NavLink to="/">
            <div className="kubeapps__logo">
              <span className="sr-only">Homepage</span>
            </div>
          </NavLink>
```

dashboard/src/index.tsx
```diff
@@ -17,6 +17,14 @@ addAuthHeaders(axiosWithAuth);
 
 ReactDOM.render(<Root />, document.getElementById("root") as HTMLElement);
 
+// Allow to load custom styling.
+// The dashboard webserver will return this style file.
+const link = document.createElement("link");
+link.href = "/custom_style.css";
+link.type = "text/css";
+link.rel = "stylesheet";
+document.getElementsByTagName("head")[0].appendChild(link);
```
Angelmmiguel, line 26: If this file will be always present and you don't need to confirm anything with JS code, place the style link directly on the [index.html](https://github.com/kubeapps/kubeapps/blob/master/dashboard/public/index.html) template.

Angelmmiguel (changes requested):
LGTM in general! Added two comments before approving it 👍 

Regarding security, I don't see any security flag by adding an arbitrary CSS file to the browser more than hiding / showing elements in the UI. The only thing that worries a me a little bit is that you can load images for tracking purposes.

andresmgot:
Thanks for the comments @Angelmmiguel!

absoludity (approved):
Nice. Yeah I think in the past the only security issue would have been that people can add a dangerous SVG as a background image, but these days that's less of an issue (if at all?) in that the browser won't run any JS in the SVG unless used within an `svg` tag (I'd not realised this when we recently discussed svg related issues).

Angelmmiguel (approved):
LGTM!! 👏

mecampbellsoup:
Hey guys - is usage of this feature documented anywhere? I see two spots:

* https://github.com/kubeapps/kubeapps/blob/3d6491545b579b89aeb9039248fbec5b55085086/chart/kubeapps/templates/dashboard-config.yaml#L47-L48
* https://github.com/kubeapps/kubeapps/blob/3d6491545b579b89aeb9039248fbec5b55085086/chart/kubeapps/templates/dashboard-config.yaml#L22-L24

It _looks_ to me like the custom CSS rules declared in the former are served by NGINX via the location in the latter?

andresmgot:
Hey, at the moment, the only documentation is in the chat values: https://github.com/kubeapps/kubeapps/blob/master/chart/kubeapps/values.yaml#L492

It's not super advanced but you can inject some CSS code using the Helm chart. And yes, it works as you mention.

PR #2249, opened 2021-01-14
Move text filter to the backend

antgamdia:
### Description of the change

`getAllChartCategories` method was not using the `ChartQuery` object to receive the information. Also, this PR allows using both `?repo` and  `?repos` for backwards compatibility.

### Benefits

The logic to filter **charts** will be delegated to the backend-side.

### Possible drawbacks

It'll stop searching the whole displayed catalog and instead it'll go to the backend to retrieve the matching **charts**. 
~~As a consequence, looking up an operator from the textbox will stop working.~~
Possible solution:  adapt the current logic by merging the regex client-side filtering for operators with the backend filtering for charts ---> implemented

### Applicable issues

This PR depends on https://github.com/kubeapps/kubeapps/pull/2246 to be merged.

### Additional information


andresmgot:
Let us know when you want a review of this. You need to use #2246 as base of this PR to be able to review it (but for that, you need to push the branch `antgamdia:getCategoriesFromBackend` to this repo rather than your fork.

antgamdia:
> Let us know when you want a review of this. You need to use #2246 as base of this PR to be able to review it (but for that, you need to push the branch `antgamdia:getCategoriesFromBackend` to this repo rather than your fork.

Thank you. The point is that there are some failing tests I'm still trying to fix that affects both PRs. For instance, this PR seems to be breaking some tests in `Catalog.test.tsx` related to the search box, probably it will be a matter of updating the component or adding some property in the test, but I'm still figuring this out.
Once I have a kinda stable #2246 I'll update this PR and will do what you said if it doesn't become merged.

antgamdia:
After merging the previous PR, this one is ready for review. Some aspects I'd like to highlight:
  - the existing tests are passing (with minor adaptations), but probably more tests could and should be added.
  - the subsequent PRs https://github.com/kubeapps/kubeapps/pull/2251 and https://github.com/kubeapps/kubeapps/pull/2252 just add minor things using this one as the base. Once each of them become merged, the pagination itself can be implemented
  - despite the fact that, conceptually, substituting the search filter to the backend is not a big deal, there have been many issues while trying to avoid inconsistencies between the search term.
    - The `currentSearchQuery` is in the component's state, so users won't notice a delay while typing.
    -  `filters[filterNames.SEARCH][0]` is the `?Search=xxxx` query parameter at the URL. Users may want to bookmark a search, so queries without having typed anything in the search box should still work.
        - Similarly, when changing namespaces, the search must be invoked again, but using a different namespace.
    - Regardless of how search has been invoked, `search.query`, at the redux state, will always have the actual searched value.
    - All of these variables will be eventually consistent; especially note that `currentSearchQuery` will accept any value at any rate, but it will be passed through with a debounce (300ms), so the actual searched value may differ. This is the intended behavior for preventing users to overload the assetsvc API.
    - Users should be able to clear the search filter in different ways: i) clear single filter (blue top tag), but not accessible to keyboard navigation; ii) using clear all filters (but all the filters will be wiped out); iii) so a third option is needed: removing the search terms should result in deleting the search filter (obvious, but not trivial to implement).

Perhaps there is a simpler approach and I'm adding unnecessary complexity here, but I haven't been able to figure out a better solution that solves the aforementioned problems   :(

Current status:

textSearch

dashboard/src/components/Catalog/Catalog.tsx
```diff
@@ -132,23 +155,66 @@ function Catalog(props: ICatalogProps) {
...
-  }, [dispatch, fetchCharts, cluster, namespace, repo]);
+    if (namespaceUpdate.current !== namespace || !currentSearchQuery.length) {
+      // otherwise, fetching will be managed by setSearchFilter
+      fetchCharts(cluster, namespace, repo, currentSearchQuery);
+    }
+    namespaceUpdate.current = namespace;
+  }, [dispatch, fetchCharts, currentSearchQuery, cluster, namespace, repo]);
+
+  const debouncedfetchChartsSearch = useCallback(
+    debounce((q: string) => {
```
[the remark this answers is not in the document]
andresmgot: Also, I am not sure if `debounce` will ever work as you expect here. Probably you want to move it out of the component. Every time there is a new property or change in the state, this component will be re-rendered (the code will be executed again). This means that `debounce` will be called again, creating a new instance.

In other components, I have written the same functionality but on my own. See `src/components/DeploymentFormBody/AdvancedDeploymentForm.tsx` for an example (for the `onChange` function).

andresmgot (commented):
Thanks for the PR @antgamdia. I have the feeling that this should be muuch more simple than how it is right now. In theory you just need to modify three simple things:

 - The call to `fetchCarts` should include the `query`
 - The current filter that removes charts based on the `searchFilter` should be deleted
 - To avoid unnecessary calls, you should gather changes in the searchFilter before re-doing a request.

The third point should be even optional. I gave that a quick try (without the third point) and had a more-or-less solution working. This is the minimal code I tried:

```patch
diff --git a/dashboard/src/actions/charts.ts b/dashboard/src/actions/charts.ts
index cebe3f75..2590b512 100644
--- a/dashboard/src/actions/charts.ts
+++ b/dashboard/src/actions/charts.ts
@@ -86,11 +86,12 @@ export function fetchCharts(
   cluster: string,
   namespace: string,
   repo: string,
+  query: string,
   return async dispatch => {
     dispatch(requestCharts());
     try {
-      const charts = await Chart.fetchCharts(cluster, namespace, repo);
+      const charts = await Chart.fetchCharts(cluster, namespace, repo, query);
       if (charts) {
         dispatch(receiveCharts(charts));
       }
diff --git a/dashboard/src/components/Catalog/Catalog.tsx b/dashboard/src/components/Catalog/Catalog.tsx
index 35baace0..2e6e78ff 100644
--- a/dashboard/src/components/Catalog/Catalog.tsx
+++ b/dashboard/src/components/Catalog/Catalog.tsx
@@ -33,7 +33,7 @@ interface ICatalogProps {
   charts: IChartState;
   repo: string;
   filter: ParsedQs;
-  fetchCharts: (cluster: string, namespace: string, repo: string) => void;
+  fetchCharts: (cluster: string, namespace: string, repo: string, query: string) => void;
   cluster: string;
   namespace: string;
   kubeappsNamespace: string;
@@ -137,12 +137,12 @@ function Catalog(props: ICatalogProps) {
     getCSVs(cluster, namespace);
   }, [dispatch, getCSVs, fetchChartCategories, cluster, namespace]);
 
-  useEffect(() => {
-    fetchCharts(cluster, namespace, repo);
-  }, [dispatch, fetchCharts, cluster, namespace, repo]);
-
   // Only one search filter can be set
   const searchFilter = filters[filterNames.SEARCH][0] || "";
+  useEffect(() => {
+    fetchCharts(cluster, namespace, repo, searchFilter);
+  }, [dispatch, fetchCharts, cluster, namespace, repo, searchFilter]);
+
   const setSearchFilter = (searchTerm: string) => {
     setFilters({
       ...filters,
@@ -155,7 +155,6 @@ function Catalog(props: ICatalogProps) {
       () => filters[filterNames.TYPE].length === 0 || filters[filterNames.TYPE].includes("Charts"),
     )
     .filter(() => filters[filterNames.OPERATOR_PROVIDER].length === 0)
-    .filter(c => new RegExp(escapeRegExp(searchFilter), "i").test(c.id))
     .filter(
       c =>
         filters[filterNames.REPO].length === 0 ||
diff --git a/dashboard/src/containers/CatalogContainer/CatalogContainer.tsx b/dashboard/src/containers/CatalogContainer/CatalogContainer.tsx
index 849fae2b..99214713 100644
--- a/dashboard/src/containers/CatalogContainer/CatalogContainer.tsx
+++ b/dashboard/src/containers/CatalogContainer/CatalogContainer.tsx
@@ -27,8 +27,8 @@ function mapStateToProps(
 
 function mapDispatchToProps(dispatch: ThunkDispatch<IStoreState, null, Action>) {
   return {
-    fetchCharts: (cluster: string, namespace: string, repo: string) =>
-      dispatch(actions.charts.fetchCharts(cluster, namespace, repo)),
+    fetchCharts: (cluster: string, namespace: string, repo: string, query: string) =>
+      dispatch(actions.charts.fetchCharts(cluster, namespace, repo, query)),
     fetchChartCategories: (cluster: string, namespace: string) =>
       dispatch(actions.charts.fetchChartCategories(cluster, namespace)),
     pushSearchFilter: (filter: string) => dispatch(actions.shared.pushSearchFilter(filter)),
diff --git a/dashboard/src/shared/Chart.ts b/dashboard/src/shared/Chart.ts
index da4139e1..be4deae4 100644
--- a/dashboard/src/shared/Chart.ts
+++ b/dashboard/src/shared/Chart.ts
@@ -4,9 +4,14 @@ import { IChart, IChartCategory, IChartVersion } from "./types";
 import * as URL from "./url";
 
 export default class Chart {
-  public static async fetchCharts(cluster: string, namespace: string, repo: string) {
+  public static async fetchCharts(
+    cluster: string,
+    namespace: string,
+    repos: string,
+    query: string,
+  ) {
     const { data } = await axiosWithAuth.get<{ data: IChart[] }>(
-      URL.api.charts.list(cluster, namespace, repo),
+      URL.api.charts.list(cluster, namespace, repos, query),
     );
     return data.data;
   }
diff --git a/dashboard/src/shared/url.ts b/dashboard/src/shared/url.ts
index b1121795..71a24f17 100644
--- a/dashboard/src/shared/url.ts
+++ b/dashboard/src/shared/url.ts
@@ -117,8 +117,10 @@ export const api = {
       `${api.charts.base(cluster, namespace)}/charts/${id}`,
     getVersion: (cluster: string, namespace: string, id: string, version: string) =>
       `${api.charts.get(cluster, namespace, id)}/versions/${encodeURIComponent(version)}`,
-    list: (cluster: string, namespace: string, repo?: string) =>
-      `${api.charts.base(cluster, namespace)}/charts${repo ? `/${repo}` : ""}`,
+    list: (cluster: string, namespace: string, repos: string, query: string) =>
+      `${api.charts.base(cluster, namespace)}/charts?${query ? "&q=" + query : ""}${
+        repos ? `&repos=${repos}` : ""
+      }`,
     getChartCategories: (cluster: string, namespace: string) =>
       `${api.charts.base(cluster, namespace)}/charts/categories`,
     listVersions: (cluster: string, namespace: string, id: string) =>
```

I believe that is going to be easier for you to start from that point and solve the pending issue (updating the searchbox properly and gather changes before submitting). Let me know if you want to do some pair programming for this so we can find a simple solution together.

dashboard/src/actions/charts.test.tsx
```diff
@@ -71,7 +72,54 @@ describe("fetchCharts", () => {
       throw new Error("something went wrong");
     });
     axiosWithAuth.get = axiosGetMock;
-    await store.dispatch(actions.charts.fetchCharts(cluster, namespace, "foo"));
+    await store.dispatch(actions.charts.fetchCharts(cluster, namespace, "foo", ""));
+    expect(store.getActions()).toEqual(expectedActions);
+  });
+});
+
+describe("fetchChartCategories", () => {
```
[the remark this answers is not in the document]
antgamdia: Not at all. Indeed this change was introduced and merged in https://github.com/kubeapps/kubeapps/pull/2246. I'm removing it since it has been accidentally duplicated during the merge.

dashboard/src/components/Catalog/Catalog.tsx
```diff
@@ -166,13 +232,34 @@ function Catalog(props: ICatalogProps) {
...
+    .filter(
+      c =>
+        filters[filterNames.REPO].length === 0 ||
+        filters[filterNames.REPO].includes(c.attributes.repo.name),
+    )
+    .filter(
+      c =>
+        filters[filterNames.CATEGORY].length === 0 ||
+        filters[filterNames.CATEGORY].includes(categoryToReadable(c.attributes.category)),
+    );
```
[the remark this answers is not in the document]
antgamdia: But this `search.items` array is totally different from the `charts` one. I'd like continue searching without losing all the already fetched charts (e.g., repo with 2000 charts, search for 'wordpress', clear search. As a user, I don't want to wait again for the 2000 charts to be re-fetched)

dashboard/src/components/Catalog/Catalog.tsx
```diff
@@ -166,13 +232,34 @@ function Catalog(props: ICatalogProps) {
...
         filters[filterNames.TYPE].length === 0 || filters[filterNames.TYPE].includes("Operators"),
     )
     .filter(() => filters[filterNames.REPO].length === 0)
-    .filter(c => new RegExp(escapeRegExp(searchFilter), "i").test(c.metadata.name))
+    .filter(c => {
+      const regex = new RegExp(escapeRegExp(currentSearchQuery), "i");
+      return (
+        regex.test(c.metadata.name) ||
+        c?.spec?.customresourcedefinitions?.owned.find(crd => regex.test(crd.displayName))
+      );
```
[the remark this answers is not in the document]
antgamdia: Currently we do not suppport backend filtering for operators. This change is just to search using more fields when filtering operators.

dashboard/src/components/Catalog/Catalog.test.tsx
```diff
@@ -113,28 +114,29 @@ it("behaves like a loading wrapper", () => {
...
-
-  it("filters modifying the search box", () => {
-    const wrapper = mountWrapper(defaultStore, <Catalog {...populatedProps} />);
     act(() => {
       (wrapper.find(SearchFilter).prop("onChange") as any)("bar");
+      (wrapper.find(SearchFilter).prop("submitFilters") as any)("bar");
     });
     wrapper.update();
-    expect(wrapper.find(InfoCard)).toHaveLength(1);
+    expect(fetchCharts).toHaveBeenCalled();
```
andresmgot, line 139: `.toHaveBeenCalledWith(...)` to ensure that it's using the proper query

dashboard/src/reducers/charts.ts
```diff
@@ -54,11 +55,16 @@ const chartsReducer = (
...
+      return { ...state, isFetching: true };
     case getType(actions.charts.requestChartsCategories):
       return { ...state, isFetching: true };
     case getType(actions.charts.receiveCharts):
-      return { ...state, isFetching: false, items: action.payload };
+      return {
+        ...state,
+        isFetching: false,
+        items: !action?.meta.length ? action.payload : state.items,
+        searchItems: action?.meta.length ? action.payload : state.searchItems,
```
andresmgot: why are you checking `action?.meta.length`?

dashboard/src/components/Catalog/Catalog.test.tsx
```diff
@@ -14,6 +14,7 @@ const defaultChartState = {
   selected: {} as IChartState["selected"],
   deployed: {} as IChartState["deployed"],
   items: [],
+  searchItems: [],
```
andresmgot, line 17: I'd add a comment saying why we are splitting `items` and `searchItems`. Also, to make it even more clear, I would change `items` for an object:

```
items: {
 all: [],
 filtered: [],
},
```

antgamdia:
I've just added some missing test cases after having implemented the text filter in a simpler approach (thx a lot @andresmgot). This PR should be now ready for a review if the CI has no objections :)

dashboard/src/components/Catalog/Catalog.tsx
```diff
@@ -135,27 +137,43 @@ function Catalog(props: ICatalogProps) {
...
+      c =>
+        filters[filterNames.REPO].length === 0 ||
+        filters[filterNames.REPO].includes(c.attributes.repo.name),
+    )
+    .filter(
+      c =>
+        filters[filterNames.CATEGORY].length === 0 ||
+        filters[filterNames.CATEGORY].includes(categoryToReadable(c.attributes.category)),
+    );
+  const filteredSearchedCharts = searchedCharts
```
andresmgot, line 172: rather than repeating all these lines, do the check here:

```
const filteredCharts = searchFilter.length ? searchedCharts : charts 
```

dashboard/src/components/Catalog/Catalog.test.tsx
andresmgot: Also, I am not sure this distinction is providing any caching. Note that even if we already have "all" items fetched from before, if the search term is removed, we are fetching all the charts again anyway. That renders the loading wrapper and ignore the current state. Am I wrong?

dashboard/src/reducers/charts.ts
antgamdia: This is the `query` value, just to know if the `receiveCharts` comes from a search query or not.

antgamdia:
As per a discussion, we should get rid of the "cache" in `searchedCharts` and just wipe out the current downloaded charts and just start over again when a new search query is introduced.
The rationale is just to keep this PR as simple as possible and implement it as a follow-up PR.

Note now we are not leveraging from this implemented feature (since a new `fetchCharts` is being called), it was just an implementation needed/desirable for the pagination itself, but out of the scope of this PR.

andresmgot (approved):
Thanks for the changes :) LGTM

PR #2250, opened 2021-01-15
Support for closable errors. Add kind-resolution error as closable

andresmgot:
### Description of the change

Follow up of #2234

At the moment, if a request to resolve API kinds fails, we fallback to the default set with a silent error. With this PR, we show a closable warning to inform about the issue:

Screenshot from 2021-01-15 13-37-21

### Benefits

Now it's possible to use closable errors :)

### DIFF
--- a/dashboard/src/components/Header/Header.tsx
+++ b/dashboard/src/components/Header/Header.tsx
@@ -27,7 +27,6 @@ function Header() {
     if (authenticated && clusters.currentCluster) {
       dispatch(actions.namespace.fetchNamespaces(clusters.currentCluster));
       dispatch(actions.namespace.canCreate(clusters.currentCluster));
-      dispatch(actions.kube.getResourceKinds(clusters.currentCluster));
     }
   }, [dispatch, authenticated, clusters.currentCluster]);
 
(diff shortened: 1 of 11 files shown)


dashboard/src/components/Header/Header.tsx
```diff
@@ -27,7 +27,6 @@ function Header() {
     if (authenticated && clusters.currentCluster) {
       dispatch(actions.namespace.fetchNamespaces(clusters.currentCluster));
       dispatch(actions.namespace.canCreate(clusters.currentCluster));
-      dispatch(actions.kube.getResourceKinds(clusters.currentCluster));
```
andresmgot, line 30: I have moved this request to the `Layout` component since the error is displayed there.

antgamdia (approved):
Great! So, from now on, we'll have a closable `AlertGroup` for other errors and warnings.
Is the default style the one in the issue? If so, I'd add a small top margin.

Should we now revisit all the existing modals in order to determine if they should be dismissable or not?

andresmgot:
> Is the default style the one in the issue? If so, I'd add a small top margin.

Yes, the component itself should be unaware of where it's injected but it's true that in here, it's missing a bit of margin from the top (I am going to add it to the parent component, to keep AlertGroup generic).

> Should we now revisit all the existing modals in order to determine if they should be dismissable or not?

I guess you mean "alerts" (not "modals", which are popup windows) but no, current errors are "hard" errors, meaning that having an error cause a malfunction in the app and should be fixed (cannot be ignored). In any case, if we find out an error which is not super important, we can make it dismissable but no need to review now all of them.

PR #2251, opened 2021-01-16
Use repos from state in catalog

antgamdia:
### Description of the change

This PR uses the `repos` property, which already exists in the global state, and adds the component state variable `currentRepo` for tracking the selected repos (used in the /chart request)  

### Benefits

The logic to getting all the repositories will no longer require iterate over the whole list of charts; instead, repos will be retrieved from the state.

### Possible drawbacks

N/A

### Applicable issues

This PR depends on https://github.com/kubeapps/kubeapps/pull/2249 to be merged.

### Additional information

### DIFF
--- a/dashboard/src/components/Catalog/Catalog.test.tsx
+++ b/dashboard/src/components/Catalog/Catalog.test.tsx
@@ -4,7 +4,7 @@ import FilterGroup from "components/FilterGroup/FilterGroup";
 import InfoCard from "components/InfoCard/InfoCard";
 import Alert from "components/js/Alert";
 import { act } from "react-dom/test-utils";
-import { defaultStore, getStore, mountWrapper } from "shared/specs/mountWrapper";
+import { defaultStore, getStore, initialState, mountWrapper } from "shared/specs/mountWrapper";
 import { IAppRepository, IChart, IChartState, IClusterServiceVersion } from "../../shared/types";
 import SearchFilter from "../SearchFilter/SearchFilter";
 import Catalog, { filterNames } from "./Catalog";
@@ -27,7 +27,7 @@ const defaultProps = {
   fetchChartCategories: jest.fn(),
   fetchRepos: jest.fn(),
   pushSearchFilter: jest.fn(),
-  cluster: "default",
+  cluster: initialState.config.kubeappsCluster,
   namespace: "kubeapps",
   kubeappsNamespace: "kubeapps",
   csvs: [],
@@ -137,7 +137,7 @@ describe("filters by the searched item", () => {
       (wrapper.find(SearchFilter).prop("onChange") as any)("bar");
     });
     wrapper.update();
-    expect(fetchCharts).toHaveBeenCalledWith("default", "kubeapps", "", 1, 0, "bar");
+    expect(fetchCharts).toHaveBeenCalledWith("default-cluster", "kubeapps", "", 1, 0, "bar");
   });
 });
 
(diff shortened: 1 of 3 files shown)


absoludity:
Did you forget to set the base branch on this PR, or is it really 24 commits? :) (Guessing you did, so will hold off).

antgamdia:
> Did you forget to set the base branch on this PR, or is it really 24 commits? :) (Guessing you did, so will hold off).

Ups, sorry! This PR should be marked as a draft, I forgot to do so! (The few changes here are the ones from the commit https://github.com/kubeapps/kubeapps/pull/2251/commits/cd29e9ebb9e0db5c629695714ef88503de16ed96).
All my latest PRs just depends upon https://github.com/kubeapps/kubeapps/pull/2249

antgamdia:
I have a failing test here, but before going deeper I just want to confirm you are ok with the approach, @andresmgot .

dashboard/src/components/Catalog/Catalog.tsx
```diff
@@ -125,14 +135,31 @@ function Catalog(props: ICatalogProps) {
...
   ).sort();
 
+  // We do not currently support app repositories on additional clusters.
+  const supportedCluster = cluster === kubeappsCluster;
+  useEffect(() => {
+    if (!namespace) {
+      // All Namespaces
+      fetchRepos("");
+      return;
+    }
```
andresmgot, line 153: not having a namespace is not a possibility in the Catalog view.

dashboard/src/components/Catalog/Catalog.tsx
```diff
@@ -141,8 +168,8 @@ function Catalog(props: ICatalogProps) {
   // Only one search filter can be set
   const searchFilter = propsFilter[filterNames.SEARCH]?.toString() || "";
   useEffect(() => {
-    fetchCharts(cluster, namespace, repo, searchFilter);
-  }, [fetchCharts, cluster, namespace, repo, searchFilter]);
+    fetchCharts(cluster, namespace, currentRepo, searchFilter);
```
andresmgot, line 171: why do you need to store a separate state variable? won't it work simply using the filters already in the state? Something like:

```js
const reposFilter = filters[filterNames.REPOS]?.join(",") || "";
 useEffect(() => {
    fetchCharts(cluster, namespace, reposFilter, searchFilter);
  }, [fetchCharts, cluster, namespace, reposFilter, searchFilter]);
```

andresmgot (commented):
The approach makes sense to me but I think you don't need the new state var

dashboard/src/components/Catalog/Catalog.tsx
antgamdia: You're right, if we already have the `filters` in the state it doesn't really make sense. I wrongly thought that, perhaps, `reposFilter`  was not always referring to the current value,

dashboard/src/components/Catalog/Catalog.test.tsx
```diff
@@ -212,15 +213,17 @@ describe("filters by application repository", () => {
...
+    const fetchRepos = jest.fn();
+    const wrapper = mountWrapper(store, <Catalog {...populatedProps} fetchRepos={fetchRepos} />);
     // The repo name is "foo"
     const input = wrapper.find("input").findWhere(i => i.prop("value") === "foo");
     input.simulate("change", { target: { value: "foo" } });
     // It should have pushed with the filter
     const historyAction = store
       .getActions()
       .find(action => action.type === "@@router/CALL_HISTORY_METHOD");
+    expect(fetchRepos).toHaveBeenCalledWith("kubeapps");
```
absoludity, line 226: So this tests the case where the namespace is kubeapps, but we could also update a separate test to ensure that
```
expect(fetchRepos).toHaveBeenCalledWith("otherNamespace", true);
```
?

absoludity (approved):

dashboard/src/components/Catalog/Catalog.test.tsx
andresmgot: :heavy_plus_sign:

andresmgot (approved):
LGTM

PR #2254, opened 2021-01-18
Refactor fetchRepos action

andresmgot:
### Description of the change

Follow up of https://github.com/kubeapps/kubeapps/pull/2245#discussion_r557286054

When `fetchRepos` is called for more than one namespace (any namespace + global namespace), the secrets from the global namespace are not needed so they should not be fetched. That's the current behavior but this PR ensures that, making it a boolean the parameter to fetch global repos or not.


antgamdia (approved):
Thank you for addressing this!

PR #2256, opened 2021-01-18
Draft: Use backend to search in catalog

andresmgot:
### Description of the change

Proposal of https://github.com/kubeapps/kubeapps/pull/2249#pullrequestreview-570398961 as proper code

This is enough to get the search-box working with backend filtering:

Peek 2021-01-18 19-02

(It can be polished a bit but it's just the idea, to avoid too many changes).

cc/ @antgamdia

### DIFF
--- a/dashboard/src/components/SearchFilter/SearchFilter.tsx
+++ b/dashboard/src/components/SearchFilter/SearchFilter.tsx
@@ -15,8 +15,19 @@ export interface ISearchFilterProps {
 }
 
 function SearchFilter(props: ISearchFilterProps) {
+  const [value, setValue] = React.useState(props.value);
+  const [timeout, setTimeoutState] = React.useState({} as NodeJS.Timeout);
   const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
-    props.onChange(e.currentTarget.value);
+    // Copy is needed to avoid losing the reference
+    const valueCopy = e.currentTarget.value;
+    setValue(e.currentTarget.value);
+    // Gather changes before submitting
+    clearTimeout(timeout);
+    setTimeoutState(
+      setTimeout(() => {
+        props.onChange(valueCopy);
+      }, 300),
+    );
   };
   const handleSubmit = (e: React.FormEvent<HTMLFormElement>) => {
     e.preventDefault();
@@ -37,7 +48,7 @@ function SearchFilter(props: ISearchFilterProps) {
               placeholder={props.placeholder}
               autoComplete="off"
               onChange={handleChange}
-              value={props.value}
+              value={value}
               {...Input.defaultProps}
             />
           </Column>
--- a/dashboard/src/actions/charts.test.tsx
+++ b/dashboard/src/actions/charts.test.tsx
@@ -39,7 +39,7 @@ describe("fetchCharts", () => {
       { type: getType(actions.charts.requestCharts) },
       { type: getType(actions.charts.receiveCharts), payload: response },
     ];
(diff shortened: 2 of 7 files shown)


dashboard/src/components/SearchFilter/SearchFilter.tsx
```diff
@@ -15,8 +15,19 @@ export interface ISearchFilterProps {
...
+    // Copy is needed to avoid losing the reference
+    const valueCopy = e.currentTarget.value;
+    setValue(e.currentTarget.value);
+    // Gather changes before submitting
+    clearTimeout(timeout);
+    setTimeoutState(
+      setTimeout(() => {
+        props.onChange(valueCopy);
+      }, 300),
+    );
```
andresmgot, line 30: FTR, I tried lodash `debounce` but I couldn't make it work as this is doing.

antgamdia:
Thank you for the initial working code, much simpler than my proposed one :P. My main concerns here are:

- You're overwriting IChartsState.Items (since you are re-using the unchanged code of fetchCharts).
- Categories are eventually lost, but perhaps it's just a problem with the reducer (default state is being set)
- Search term (the actual text on the input field) could be eventually a bit laggy resulting in an unresponsive UI sometimes. 
  - To keep this PR as simple as possible, we can address it in a future PR (if it's a concerning problem)

I'd like to remove all the overhead in my original PR, but, at the same time, I'd like to still have separate arrays for both the whole list of charts and the filtered ones.

andresmgot:
> * You're overwriting IChartsState.Items (since you are re-using the unchanged code of fetchCharts).
> * Search term (the actual text on the input field) could be eventually a bit laggy resulting in an unresponsive UI sometimes.
> 
> I'd like to remove all the overhead in my original PR, but, at the same time, I'd like to still have separate arrays for both the whole list of charts and the filtered ones.

Yes, that would be a good performance improvement (doing a request with the empty search-box is the slowest). It can be implemented as a follow-up PR though.

> * Categories are eventually lost, but perhaps it's just a problem with the reducer (default state is being set)

Yes, that's a problem. I assumed it was because my local copy was missing some of your code. Can that be the reason?

antgamdia:
> es, that's a problem. I assumed it was because my local copy was missing some of your code. Can that be the reason?

Yes, in the `requestCharts` you're setting `initialState`, which has `categories: []`. Replacing `initialState` by `state` should fix it.
```javascript
case getType(actions.charts.requestCharts):
      return { ...initialState, isFetching: true };
```

I don't know what's the purpose of removing the `items` there. Perhaps it can be replace by `startFetching` for both filtered/unfiltered search (addressing therefore https://github.com/kubeapps/kubeapps/pull/2249#discussion_r559497099) just setting the `isFetching` (and later on, the pagination status)

andresmgot:
> I don't know what's the purpose of removing the `items` there.

This is to avoid showing "wrong" charts. Imaging this case: loads charts from `namespace1` -> `items` is populated with those charts. Changes to `namespace2` and it fails to retrieve charts in that namespace. The state at that point will contain the charts from `namespace1` and the error. In any case, that can be easily fixed removing `items` in case of error so I am okay not removing `items`.

> Perhaps it can be replace by `startFetching` for both filtered/unfiltered search (addressing therefore [#2249 (comment)](https://github.com/kubeapps/kubeapps/pull/2249#discussion_r559497099)) just setting the `isFetching` (and later on, the pagination status)

I am not sure what would be fixed with that :confused:

andresmgot:
BTW, closing this, is not meant to be merged.

Issue #2232, opened 2021-01-08
OCI Integration for Chart Registries


andresmgot:
[Design Document] https://github.com/kubeapps/kubeapps/blob/master/docs/architecture/design-proposals/OCI-registries.md

Based in the design document, we'll go ahead an implement the option 3 explained there: Allow OCI Registries, requiring the user to add the list of repositories to be used.

The list of items required to implement the feature are:

 - [x] Adapt the CRD spec (XS)
 - [x] #asset-syncer Fetch repository index (S)
 - [x] #asset-syncer Fetch and import files (M)
   ~~- [ ] (extra) #asset-syncer Improve memory footprint (https://github.com/kubeapps/kubeapps/pull/2283#discussion_r565730063)~~
   - [x] Improve performance
   ~~- [ ] Limit number of charts synced~~
 - [x] #kubeops Parse chart details (XS)
 - [x] #kubeops Initialize Net cli (M)
 - [x] #kubeops Registry secrets per domain (XS)
 - [x] #kubeops Get (pull) charts (M)
   - [x] (extra) #kubeops allow insecure TLS
 - [x] #kubeops Validate registry (S)
 - [x] (optional) #kubeops Validate given repos (S)
 - [x] #dashboard Adapt App Repo form (M-L)
   ~~- [ ] (extra) Allow to select same creds for charts & docker images~~
 - [x] Documentation
 - [x] Add support for initial-repos in chart

andresmgot:
I've moved the remaining items to https://github.com/kubeapps/kubeapps/issues/2425 so we can call this done

PR #2257, opened 2021-01-19
Update CRD spec to include list of repositories

andresmgot:
### Description of the change

As explained in the design document, the option chosen to implement OCI registries requires a small modification of the AppRepository CRD to include the list of repositories.

### Applicable issues

  - ref #2232

### DIFF
--- a/cmd/apprepository-controller/pkg/apis/apprepository/v1alpha1/types.go
+++ b/cmd/apprepository-controller/pkg/apis/apprepository/v1alpha1/types.go
@@ -45,6 +45,9 @@ type AppRepositorySpec struct {
 	// in the same namespace as the AppRepository and should be included
 	// automatically for matching images.
 	DockerRegistrySecrets []string `json:"dockerRegistrySecrets,omitempty"`
+	// In case of an OCI type, the list of repositories is needed
+	// as there is no API for the index
+	Repositories []string `json:"repositories,omitempty"`
 }
 
 // AppRepositoryAuth is the auth for an AppRepository resource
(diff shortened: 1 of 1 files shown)


antgamdia (approved):
Great! So, this is just a list of repository names for CRD to have a kind of reference/index with the repos that it contains, I guess

andresmgot:
> Great! So, this is just a list of repository names for CRD to have a kind of reference/index with the repos that it contains, I guess

Yes, since OCI registries don't have an index endpoint, in this case, it's manual.

cmd/apprepository-controller/pkg/apis/apprepository/v1alpha1/types.go
```diff
@@ -45,6 +45,9 @@ type AppRepositorySpec struct {
 	// in the same namespace as the AppRepository and should be included
 	// automatically for matching images.
 	DockerRegistrySecrets []string `json:"dockerRegistrySecrets,omitempty"`
+	// In case of an OCI type, the list of repositories is needed
+	// as there is no API for the index
+	Repositories []string `json:"repositories,omitempty"`
```
absoludity, line 50: Just a thought, feel free to ignore: This object to which you're adding the `Repositories` is the spec of an `AppRepository`. Generally, an `AppRepository` as a repository of apps (whether that's a helm repository or something else in the future, such as, in this case, an OCI registry), so it initially hits some dissonance to then see a field called `Repositories`. Might be clearer to name it `OCIRepositories` so `appRepository.Spec.Repositories` becomes `appRepository.Spec.OCIRepositories`. See what you think.

andresmgot: makes sense to me

PR #2260, opened 2021-01-20
Refactor: Move asset-syncer methods to an interface

andresmgot:
### Description of the change

This PR doesn't include any functional change, two changes have been included:

 - The AppRepository `type` is now passed as an argument to the asset-syncer (but currently it's being ignored)
 - Methods depending on the `type` has been moved to an interface (`Repo`) so I can implement those for the OCI repository type in the next PR.

### Applicable issues

  - ref #2232

### DIFF
--- a/pkg/chart/models/chart.go
+++ b/pkg/chart/models/chart.go
@@ -30,6 +30,7 @@ type Repo struct {
 	Namespace string `json:"namespace"`
 	Name      string `json:"name"`
 	URL       string `json:"url"`
+	Type      string `json:"type"`
 }
 
 // RepoInternal holds the App repository details including auth and checksum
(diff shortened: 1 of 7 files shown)


pkg/chart/models/chart.go
```diff
@@ -30,15 +30,16 @@ type Repo struct {
...
 }
 
 // RepoInternal holds the App repository details including auth and checksum
 type RepoInternal struct {
 	Namespace           string `json:"namespace"`
 	Name                string `json:"name"`
 	URL                 string `json:"url"`
+	Type                string `json:"type"`
 	AuthorizationHeader string `bson:"-"`
-	Checksum            string
```
absoludity, line 41: Just need to update the comment above, given that the checksum is no longer held here.

absoludity (approved):

PR #2261, opened 2021-01-20
Rename repositories property in AppRepository CRD

andresmgot:
### Description of the change

Follow up of https://github.com/kubeapps/kubeapps/pull/2257#discussion_r560575419


andresmgot:
_merging this as trivial since I need the change for other work_

PR #2263, opened 2021-01-20
Fix linter warnings

andresmgot:
### Description of the change

Fix a bunch of warnings vscode was reporting to me.


antgamdia (approved):
Thank you! When I edited this file in the past I saw the warnings, but I simply suppose there was a reason to keep them. 
Happy to see you got rid of them :)

PR #2252, opened 2021-01-16
Add pagination params 

antgamdia:
### Description of the change

This PR is to introduce the `page` and `size` params (defaulted to `1`, `0`, that is, no pagination whatsoever). The functionality is the very same as the parent branch of this PR.

### Benefits

Not many... just intended for easing the endless scrolling feature.

### Possible drawbacks

N/A

### Applicable issues

This PR depends on https://github.com/kubeapps/kubeapps/pull/2249 to be merged.

### Additional information

- Change initially introduced in https://github.com/kubeapps/kubeapps/pull/2221 but removed from there to keep it as simple as possible.

PR #2264, opened 2021-01-20
Infinite scrolling pagination

### LINKED ISSUE #1010: Pagination for Catalog page

antgamdia:
### Description of the change

This PR adds endless scrolling in the Catalog view as a result of the analysis and discussions of the [design document](https://via.vmw.com/2020-kubeapps-ui-proposal).

We already implemented the server-side filtering for the most expensive calls: 1) filter by text; 2) filter by repository.
Moreover, the available categories are retrieved from the server-side (instead of getting unique names each time on client-side).

In this PR we support endless pagination, so once the user has reached the bottom of the page, new chunks will be requested.

### Benefits

Users won't experience a long delay when rendering the whole set of charts, since the catalog will be retrieved little by little.

### Possible drawbacks

~~Currently, there are issues when triggering many requests at a time. For instance, for replicating this failure scenario, rapidly click on a repository checkbox multiple times and scroll up and down. In the network tab, some requests don't make sense: repeated ones, race conditions. etc~~
Currently, we don't retry failed requests. 

### Applicable issues

- fixes #1010

### Additional information

Change initially introduced in #2221 but removed from there to keep it as simple as possible.

Specifically, this PR is being sent after having implemented the necessary code in other places for the pagination to work, namely:

- [x] - Minor change in assetsvc (https://github.com/kubeapps/kubeapps/pull/2244)
- [x] - Minor unhandled rejection (https://github.com/kubeapps/kubeapps/pull/2245)
- [x] - Add logic to retrieve the list of categories via backend (https://github.com/kubeapps/kubeapps/pull/2246)
- [x] -Add logic to move text search to backend + force fetch when the namespace is changed (required for the implemented logic) (https://github.com/kubeapps/kubeapps/pull/2249)
- [x] -  Get allRepos from the state instead of iterating all charts to get them (https://github.com/kubeapps/kubeapps/pull/2251)
- [x] -  Change the endpoint to the paginated ones, but without actually using it (https://github.com/kubeapps/kubeapps/pull/2252)

Some implementation details:

#### First render:
`page`=`1`
`hasRequestedFirstPage`=`false`
`hasLoadedFirstPage`=`false`
`isFetching`=`false`
useEffect(initial)-> `fetchCharts` with `page`=`1`
`isFetching`=`false`
useEffect(`isFetching` changed to `true`) -> `hasRequestedFirstPage`=`true`
useEffect(`hasRequestedFirstPage` changed to `true`) -> `hasLoadedFirstPage`=`true`

#### Scrolling
user scrolls down and, because `hasLoadedFirstPage` is `true` -> `page`=`page + 1`
useEffect(`page` changed to)-> `fetchCharts` with `page`=`2`


andresmgot:
It's usually better to review finished PRs but I am okay giving some early feedback

dashboard/src/components/Catalog/Catalog.tsx
```diff
@@ -220,6 +237,58 @@ function Catalog(props: ICatalogProps) {
...
+                  (!filters[filterNames.TYPE].length ||
+                    filters[filterNames.TYPE].find((type: string) => type === "Charts")) &&
+                  // Disable scrolling if all the charts have been fetched
+                  // && !isFetching
+                  !hasFinished &&
+                  // Enable scrolling just if every parameter is already updated
+                  clusterUpdate.current === cluster &&
+                  namespaceUpdate.current === namespace &&
+                  reposFilterUpdate.current === reposFilter &&
+                  searchFilterUpdate.current === searchFilter
```
andresmgot, line 275: I don't understand these conditions, I believe is trying to avoid duplicated calls but can you elaborate?

dashboard/src/reducers/charts.ts
```diff
@@ -56,14 +59,20 @@ const chartsReducer = (
...
   switch (action.type) {
     case getType(actions.charts.requestCharts):
-      return { ...state, isFetching: true };
+      state.records.set(action.payload, false);
+      return { ...state, isFetching: true, records: state.records };
     case getType(actions.charts.requestChartsCategories):
       return { ...state, isFetching: true };
     case getType(actions.charts.receiveCharts):
+      state.records.set(state.page, true);
+      state.records.set(state.page + 1, false);
```
andresmgot, line 68: do you need this line? Isn't the default `false`?

dashboard/src/reducers/charts.ts
```diff
@@ -56,14 +59,20 @@ const chartsReducer = (
...
       return {
         ...state,
         isFetching: false,
-        items: action.payload,
+        hasFinished: action?.meta,
+        items: uniqBy([...state.items, ...action.payload], "id"), // TODO(agamez): handle undesired requests to avoid this workaround
+        page: action?.meta ? state.page : state.page + 1, // if action.meta==true, it's the last chunk
```
andresmgot, line 74: I would include the page in the payload

dashboard/src/components/Catalog/Catalog.tsx
```diff
@@ -168,9 +172,22 @@ function Catalog(props: ICatalogProps) {
...
+  // fetchCharts is only called in observeBorder
   useEffect(() => {
-    fetchCharts(cluster, namespace, reposFilter, page, size, searchFilter);
-  }, [fetchCharts, cluster, namespace, reposFilter, page, size, searchFilter]);
+    clusterUpdate.current = cluster;
+    namespaceUpdate.current = namespace;
+    reposFilterUpdate.current = reposFilter;
+    searchFilterUpdate.current = searchFilter;
+    resetRequestCharts();
+  }, [resetRequestCharts, cluster, namespace, reposFilter, searchFilter]);
```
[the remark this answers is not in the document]
andresmgot: If I understand the problem here, if you are in page 3 and then you click in a repo filter, you want to reset the page counter, right? (So you request the first page again). Or you are trying some performance improvement?

What I would do, to try to simplify this a bit (and potentially avoid the `refs`), would be to have the `page` in the component state (rather than in redux), and everytime there is a change in the filters, reset the page. I am not sure if that'd trigger two requests though (one for the page: 0 and old filters and one for page: 0 and the new filters or just the second).

dashboard/src/reducers/charts.ts
```diff
@@ -56,14 +59,20 @@ const chartsReducer = (
...
       return { ...state, isFetching: true };
     case getType(actions.charts.receiveCharts):
+      state.records.set(state.page, true);
+      state.records.set(state.page + 1, false);
       return {
         ...state,
         isFetching: false,
-        items: action.payload,
+        hasFinished: action?.meta,
+        items: uniqBy([...state.items, ...action.payload], "id"), // TODO(agamez): handle undesired requests to avoid this workaround
```
andresmgot, line 73: yep, +1 to the TODO, we should get rid of those duplicated requests

andresmgot (commented):
Some comments here, the major two things I would change is having the `page` in the Catalog state and use the total number of pages that the API should return to know if you are in the latest page or not.

dashboard/src/actions/charts.ts
```diff
@@ -90,15 +93,19 @@ export function fetchCharts(
...
 ): ThunkAction<Promise<void>, IStoreState, null, ChartsAction> {
   return async dispatch => {
-    dispatch(requestCharts(query));
-    try {
-      const charts = await Chart.fetchCharts(cluster, namespace, repos, page, size, query);
-      dispatch(receiveCharts(charts));
-    } catch (e) {
-      dispatch(errorChart(new FetchError(e.message)));
+    const latestElement = Array.from(records.keys()).pop() || 1;
+    if (records.get(page) === false && page <= latestElement) {
```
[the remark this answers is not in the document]
antgamdia: It is just an extra condition for my unfruitful attempts to avoid the problems of duplicating requests.
The rationale behind this is `latestElement` is being set in `receiveCharts`: here we set to `false` the next page that should be requested (otherwise, it should be `undefined`).

So, if is properly processes page 6, it will set `7: false`. During the next request, `(records.get(7) === false && 7 <= 7)`

If something is wrong, the map could possibly include `7:false, 8:false` , the `latestElement` would be 8 whereas `page` is still 7. 

I'm sure there exists a better solution for the root cause.

Regarding the renaming, it's not the `lastPage`, but the `lastPendingPage`, but I'm ok with whichever name.

dashboard/src/actions/charts.ts
```diff
@@ -90,15 +93,19 @@ export function fetchCharts(
...
+      dispatch(requestCharts(page, query));
+      try {
+        const charts = await Chart.fetchCharts(cluster, namespace, repos, page, size, query);
+        dispatch(receiveCharts(charts, charts.length === 0));
```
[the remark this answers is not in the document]
antgamdia: Yes, I agree, I'm going to change this condition and leverage from the `totalPages` value already present in the API response. Totally, right!

dashboard/src/components/Catalog/Catalog.tsx
```diff
@@ -220,6 +237,58 @@ function Catalog(props: ICatalogProps) {
         intersection(filters[filterNames.CATEGORY], getOperatorCategories(c)).length,
     );
 
+  // avoid repeating the same requests with the same params with useCallback
+  // avoid overloading the backend with a max rate of 2 reqs/second (= debounce 0.5s)
+  const debouncedFetchCharts = useCallback(
+    debounce(() => {
+      fetchCharts(cluster, namespace, reposFilter, page, size, records, searchFilter);
+    }, 500),
```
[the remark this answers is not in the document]
antgamdia: As far as I could check, the denounce is working accordingly working with the specified time (1s in the gif):

debounce

Other if we remove the `debounce` and the `useCallback` each request is being duplicated since we wouldn't store the returned values of the function. Here's an example:

nodebounce

dashboard/src/components/Catalog/Catalog.tsx
antgamdia: Yes, I'd want to explain it a little bit more as it is just another attempt to avoid dups:

Whenever a change occurs and it is detected by the `useEffect`, I (try to) prevent these values to be actually updated before triggering the reset. It is being done by means of the `useRef`. The good point here is that mutating the `.current` won't cause a re-render.

If we remove these conditions, duplicated requests would appear.

dashboard/src/reducers/charts.ts
antgamdia: I think so; when there is no `key`, the `map.get(xxx)` will return `undefined` instead of `false` (and this behavior is being used in https://github.com/kubeapps/kubeapps/pull/2264#discussion_r561066033)

dashboard/src/components/Catalog/Catalog.tsx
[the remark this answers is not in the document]
antgamdia: I used the `useRef` beacuse of I explained in https://github.com/kubeapps/kubeapps/pull/2264#discussion_r561084979.

Mmm, but, although we store `page` in the component state, we will still have it as one of the dependencies of the `useEffect`, so, regardless of how it is stored, it would cause a refresh whenever it is updated, wouldn't it?

I don't know, anyway I will give it a try for sure.

antgamdia:
> having the `page` in the Catalog state and use the total number of pages that the API should return to know if you are in the latest page or not.

Thank you, as usual, for your comments :) 

Regarding the `page` in the Catalog state, I'm not completely sure about that. I'm using the state's page here to (try to) avoid duplicates. See what you think.
https://github.com/kubeapps/kubeapps/blob/1f22d91c57268f6e5d2bbc5b770bad57f32ca942/dashboard/src/reducers/charts.ts#L69 

The second point has been addressed, a much better approach to detect the end of the pagination!

dashboard/src/actions/charts.ts
```diff
@@ -90,15 +94,19 @@ export function fetchCharts(
...
 ): ThunkAction<Promise<void>, IStoreState, null, ChartsAction> {
   return async dispatch => {
-    dispatch(requestCharts(query));
-    try {
-      const charts = await Chart.fetchCharts(cluster, namespace, repos, page, size, query);
-      dispatch(receiveCharts(charts));
-    } catch (e) {
-      dispatch(errorChart(new FetchError(e.message)));
+    const lastPendingPage = Array.from(records.keys()).pop() || 1;
+    if (records.get(page) === false && page <= lastPendingPage) {
```
andresmgot, line 102: Following the conversation at https://github.com/kubeapps/kubeapps/pull/2264#discussion_r561018190

As a general comment, code should not be added "hoping" to behave somehow, it has to be proven to behave that way with unit tests. I don't see any unit tests enforcing the behavior of this so I'd assume this is not needed. Code that is not "protected" by unit tests is damned to be broken in the future when we don't remember what's doing and we decide to clean it up.

Regarding this issue in particular, I don't think is the duty of the function `fetchCharts` to decide if it should `fetchCharts` or not based on the state. It's the duty of the caller, in this case the `Catalog` component, to not request twice a page (and it's easier to add safeguards and tests there).

dashboard/src/actions/charts.ts
```diff
@@ -90,15 +94,19 @@ export function fetchCharts(
...
+      dispatch(requestCharts(page, query));
+      try {
+        const response = await Chart.fetchCharts(cluster, namespace, repos, page, size, query);
+        dispatch(receiveCharts({ items: response.data, page }, response?.meta?.totalPages <= page));
```
andresmgot, line 106: I would also add the `totalPages` to the payload unless using the meta parameter as you are doing gives some benefit, that way, if in the future you need to add more info, you don't need to guess if it's in one parameter or the other.

dashboard/src/components/Catalog/Catalog.tsx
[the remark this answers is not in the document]
andresmgot: yes, but that's what we want, no? If we are at `page: 2, repo: ""` and we change it to `repo: "bitnami` and **at the same time** (this is what I don't know if it's possible) `page: 0` we would do a single `fetchChart` with `page: 0, repo: "bitnami"`. Isn't that correct?

dashboard/src/components/Catalog/Catalog.tsx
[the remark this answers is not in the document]
andresmgot: Nice, thanks for double-checking this. You will need to add some tests for this though. Also, I think that the usage of `debounce` is legitimate but `useCallback` seems a hack for fixing the problem of not re-requesting the same page (in theory we should be taking the count of the pages we have already requested, and in fact we are, and use that to decide to trigger or not the function).

In any case, I am fine with this as long as you ensure it with a test.

dashboard/src/components/Catalog/Catalog.tsx
andresmgot: that's fine then but as above, you'd need to ensure this with a test. First, comment this code, and write a test ensure it's failing (there are more than one request for the same params), then enable the functional code and the test should pass.

dashboard/src/reducers/charts.ts
```diff
@@ -56,14 +59,26 @@ const chartsReducer = (
...
       return {
         ...state,
         isFetching: false,
-        items: action.payload,
+        hasFinished: action?.meta,
+        items: areMatchingPages
+          ? uniqBy([...state.items, ...action.payload.items], "id")
+          : state.items, // if pages don't match, ignore payload to avoid duplicates
+        page: action?.meta ? action.payload.page : action.payload.page + 1, // if action.meta==true, it's the last chunk
+        records: state.records,
```
andresmgot, line 81: I think we can simplify all this if we store `items` as a map. If we store:

```
 items[action.payload.page] = action.payload.items
```

You will get unique items per page and, I think, you won't need the `records` since you already have a map.

dashboard/src/reducers/charts.test.ts
```diff
@@ -102,61 +105,81 @@ describe("chartReducer", () => {
...
-      items: [],
+      hasFinished: true,
+      items: [chartItem],
+      page: 2,
+      records: new Map<number, boolean>()
+        .set(1, true)
+        .set(2, true)
+        .set(3, false),
     });
   });
```
andresmgot, line 172: I'd add a couple more of tests, one to check that receiving two pages of charts includes all the charts as `items` and another checking that if you add the same page twice the items are not duplicated.

dashboard/src/shared/types.ts
```diff
@@ -95,6 +104,7 @@ export interface IChartAttributes {
 
 export interface IChartState {
   isFetching: boolean;
+  hasFinished: boolean;
```
andresmgot, line 107: nitpipck: `hasFinished` could mean anything, a better name would be `finishedFetching`

andresmgot (commented):
Thanks @antgamdia this is looking better! As a summary, I have two major comments:

 - There are many tests missing to ensure the behavior you are coding.
 - The decision to fetch or no charts should be made in the Catalog. Thanks to that, you may be able to remove (or not) the pages from the state.
 
Let me know what you think or if you are stuck testing something.

antgamdia:
> * The decision to fetch or no charts should be made in the Catalog. Thanks to that, you may be able to remove (or not) the pages from the state.

Again, thank you for your comments!

Just for the record, as you suggested in 1-1 conversations, we could do something like:

```javascript
const [pagesRequested, setPagesRequested] = useState([] as number[]);
if (
    entry.isIntersecting &&
    !isFilteringOnlyOperators && 
    !hasFinished &&
    !pagesRequested.include(page)
  ) {
      setPagesRequested(pagesRequested.concat(page));
      debouncedFetchCharts();
  }
```

antgamdia:
As I mentioned in the (offline) discussion, I have not been able to move the `page` and its corresponding logic to the component. Besides my unfruitful attempts to do so, multiple requests are being generated. 
Currently, in this PR this problem is addressed with a hacky combination of `useRef`, `useCallback` and `debounce`...

This situation also prevented me to write some test cases, I will ping you to discuss a little bit more on how to address them.
As far as I can remember, the current missing tests are:

- Check that `resetRequestCharts` is invoked after a change is detected in `cluster, namespace, reposFilter, searchFilter`.
- Check the `debounce` is really giving a rate of 2req/second.
- Test the `intersection API` to simulate a scrolling event.

Apart from these three major (at least for me) tests, the PR will be finished.

dashboard/src/actions/charts.test.tsx
```diff
@@ -62,33 +81,6 @@ describe("fetchCharts", () => {
     );
   });
 
-  it("fetches charts with query", async () => {
```
andresmgot, line 53: why have you deleted this test? (even if the query is not sent in the action it's useful to ensure it's being used)

dashboard/src/reducers/charts.ts
```diff
@@ -70,14 +71,15 @@ const chartsReducer = (
         state.records.set(state.page, true); // mark the current page as successfully retrieved
         state.records.set(state.page + 1, false); // mark the next page as pending (not undefined)
       }
+      const isLastPage = action.payload.page >= action.payload.totalPages;
```
andresmgot: shouldn't this be `page === totalPages`? (not sure why the `>=`, if that's possible.)

dashboard/src/components/Catalog/Catalog.tsx
```diff
@@ -234,123 +311,145 @@ function Catalog(props: ICatalogProps) {
...
+                        filters[filterNames.TYPE].find((type: string) => type === "Charts")) && (
+                        <div className="endPageMessage">
+                          <LoadingWrapper loaded={false} />
+                          <span>
+                            {!charts.length && !csvs.length
+                              ? "Loading catalog..."
+                              : "Scroll down to discover more applications"}
+                          </span>
+                        </div>
+                      )}
```
andresmgot, line 439: two minor things here:
 - I would use the small loading spinner, I think there is a property for that
 - Would remove the messages (including "no remaining applications"), it's clear that it's loading when the loading wrapper is shown and it's weird when using filters. For example if I search "wordpress" I see the message "scroll down..." but there is no need to scroll down, is just a matter of letting it load.

andresmgot:
Another usability issue, I see that operators are loaded first, before getting the first page:

Peek 2021-01-22 11-59

We should show only the loading wrapper before we get the first page.

andresmgot:
Sorry, I am adding comments as I see them (I am checking the code locally).

I am also getting repeated requests, this is the load of the second page:

Screenshot from 2021-01-22 12-02-50

Is that expected?

Also, I think we should use smaller page sizes, 50 rather of 100 would be better IMO

dashboard/src/actions/charts.test.tsx
antgamdia: The action no longer has the "query" as part of the payload. Why would we have to still have this test for a non-existent parameter? 

```javascript 
export const requestCharts = createAction("REQUEST_CHARTS", resolve => {
  return (page?: number) => resolve(page);
});
```

So now the tests are testing there `page` parameter, so new test cases have been added for:

```
fetches charts from a repo (first page)
fetches charts from a repo (middle page)
fetches charts from a repo (last page)
fetches charts from a repo (already processed page)
fetches charts from a repo (not-yet-requested page)
fetches charts from a repo (already-requested page)
```

dashboard/src/reducers/charts.ts
antgamdia: Yes, in theory, we reach the end when the `page===totalPages`. It was just to cover the edge case in which two `receiveCharts` arrive at the same time and, "lastPage" and "lastPage+1" are both marked to be fetched. So I directly consider `lastPage+n` as the same case.

However, it should occur and, if it does, we can just ignore subsequents `lastPage+n` request.

dashboard/src/components/Catalog/Catalog.tsx
antgamdia: - I'll test the small spinner out, I remembered to have seen this prop and will add some screenshots for us to compare.
- I'm not sure about that, searching for WordPress perhaps results in a few charts (and "no remaining applications" won't be displayed if searching). However, search results can possibly be paginated, so these messages do make sense for me. Again, I will change it and will post here some gifs for us to decide.

antgamdia:
> We should show only the loading wrapper before we get the first page.

I don't get your point here, this is how it was working so far.  What if only operators, and not charts, are present?
For doing so, we should have a kinda `hasStartedFetching` or similar check to be sure that we only are rendering them (charts/ops) if: 1) we already have charts OR 2) a chart retrieving process has initiated but there are no results.
Perhaps we can address it in another PR (we also have the issue when entering filters with ",", I was wondering about creating an issue with all this _future work_.

antgamdia:
> Sorry, I am adding comments as I see them (I am checking the code locally).

Not a problem at all! And thank you for checking manually the code :) 

> Also, I think we should use smaller page sizes, 50 rather of 100 would be better IMO

Yes, it was something I also wanted to discuss. For this PR, I can hardcoded it to 50, that's great. However, perhaps we should move it to the `Config.ts` so that it becomes more easily to be changed.

antgamdia:
After a bunch of test cases and minor bug fixes after addressing the previous comments and code suggestions, there is a stable version. There are no duplicated requests and the overall UX should be also ok.
Some test cases are still pending; honestly, I can't figure out how to test in-depth how the pagination currently works (notes below) and how to better test the scrolling.

Given that the end of the iteration is coming, I would suggest prioritizing the essential pending aspects you may consider to this PR to be merged. And, next, creating an issue for documenting the further actions for further iterations.
For instance: i) search text box is not always in sync with the ?Search param; ii) I'd like to have a configurable `size` rather than hardcoding it; iii) there is an issue when filter values include a comma.

See what you think :)




#### First render:
`page`=`1`
`hasRequestedFirstPage`=`false`
`hasLoadedFirstPage`=`false`
`isFetching`=`false`
useEffect(initial)-> `fetchCharts` with `page`=`1`
`isFetching`=`false`
useEffect(`isFetching` changed to `true`) -> `hasRequestedFirstPage`=`true`
useEffect(`hasRequestedFirstPage` changed to `true`) -> `hasLoadedFirstPage`=`true`

#### Scrolling
user scrolls down and, because `hasLoadedFirstPage` is `true` -> `page`=`page + 1`
useEffect(`page` changed to)-> `fetchCharts` with `page`=`2`

dashboard/src/reducers/charts.test.ts
```diff
@@ -99,64 +111,189 @@ describe("chartReducer", () => {
...
+        type: getType(actions.charts.receiveCharts) as any,
+        payload: { items: [chartItem], page: 2, totalPages: 3 } as IReceiveChartsActionPayload,
+      },
+    );
+    expect(state).toEqual({
+      ...initialState,
+      isFetching: false,
+      hasFinishedFetching: false,
+      items: [chartItem],
+      // page: 3,
```
andresmgot, line 151: remove comment

dashboard/src/reducers/charts.test.ts
```diff
@@ -99,64 +111,189 @@ describe("chartReducer", () => {
...
```
andresmgot, line 168: same

dashboard/src/components/Catalog/Catalog.test.tsx
```diff
@@ -196,6 +318,128 @@ describe("filters by application type", () => {
...
+        }
+      />,
+    );
+    expect(wrapper.find("CatalogItems").prop("page")).toBe(1);
+    expect(setState).toHaveBeenCalledWith(1);
+    expect(mockUseEffect).toHaveBeenCalledTimes(9);
+    expect(resetRequestCharts).toHaveBeenNthCalledWith(1);
+  });
+});
+
```
andresmgot, line 442: I wanted to check if it's possible to mock the observer to test that at least we are changing the page. I managed to do so:

```js
  it("changes page", () => {
    const setState = jest.fn();
    const setPage = jest.fn();
    const charts = {
      ...defaultChartState,
      hasFinishedFetching: false,
      isFetching: false,
      items: [],
    } as any;
    spyOnUseState = jest
      .spyOn(React, "useState")
      //  @ts-ignore
      .mockImplementation((init: any) => {
        if (init === false) {
          // Mocking the result of hasLoadedFirstPage to simulate that is already loaded
          return [true, setState];
        }
        if (init === 1) {
          // Mocking the result of setPage to ensure it's called
          return [1, setPage];
        }
        return [init, setState];
      });

    // Mock intersection observer
    const observe = jest.fn();
    const unobserve = jest.fn();
    window.IntersectionObserver = jest.fn(callback => {
      (callback as (e: any) => void)([{ isIntersecting: true }]);
      return { observe, unobserve } as any;
    });

    mountWrapper(defaultStore, <Catalog {...populatedProps} charts={charts} />);
    expect(setPage).toHaveBeenCalledWith(2);
  });
```

Ideally we should check that `fetchCharts` is called with the second page, rather than mocking `setPage` but if we don't mock `setPage`, it enters an infinite loop increasing the page.

andresmgot (commented):
Thanks! The functional code looks good to me (caveat below), I just have some minor comments regarding the tests.

> iii) there is an issue when filter values include a comma.

I'm able to reproduce that bug and it's quite important. Can you open another PR to fix it before doing a release? It should be easy to fix (just ignore commas for the search filter).

dashboard/src/components/Catalog/Catalog.test.tsx
andresmgot: BTW, you may need to also mock `window.IntersectionObserverEntry`, I commented that part of the code when trying this

dashboard/src/components/Catalog/Catalog.test.tsx
```diff
@@ -1,23 +1,25 @@
...
+  hasFinishedFetching: false,
   selected: {} as IChartState["selected"],
   deployed: {} as IChartState["deployed"],
   items: [],
   categories: [],
   updatesInfo: {},
   page: 1,
-  size: 0,
+  size: 20,
+  records: new Map<number, boolean>().set(1, false),
```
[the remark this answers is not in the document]
antgamdia: Yes, `records` was unnoticedly left there. But, `hasFinishedFetching` is still part of the `IChartState`, so it should remain, shouldn't it?

dashboard/src/actions/charts.ts
```diff
@@ -93,10 +96,16 @@ export function fetchCharts(
...
     try {
-      const charts = await Chart.fetchCharts(cluster, namespace, repos, page, size, query);
-      dispatch(receiveCharts(charts));
+      const response = await Chart.fetchCharts(cluster, namespace, repos, page, size, query);
+      dispatch(
+        receiveCharts({
+          items: response.data,
+          page,
+          totalPages: response.meta.totalPages,
+        } as IReceiveChartsActionPayload),
```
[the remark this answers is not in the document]
antgamdia: Yes, this is not necessary at all, but I just wanted to type the object so that we can get transpiler errors when changing its props in the future. However, I'm ok removing it.

dashboard/src/components/Catalog/Catalog.test.tsx
antgamdia: FTR:
The condition was: `if ("IntersectionObserver" in window && "IntersectionObserverEntry" in window && "isIntersecting" in window.IntersectionObserverEntry.prototype)`

Since mocking `window.IntersectionObserverEntry.prototype` is causing problems, we decided to get rid of this condition.
In the future, if we decide to implement a fallback mechanism (fetch all pages instead of one-by-one), we can add it back.

dashboard/src/components/Catalog/Catalog.test.tsx
```diff
@@ -359,26 +340,14 @@ describe("pagination", () => {
     );
 
     expect(wrapper.find("CatalogItems").prop("page")).toBe(1);
-    expect(setState).toHaveBeenCalledWith(1);
-    expect(mockUseEffect).toHaveBeenNthCalledWith(3, [false, false]); // [hasRequestedFirstPage, isFetching])
-    expect(fetchCharts).toHaveBeenNthCalledWith(1, "default-cluster", "kubeapps", "", 1, 20, ""); // [hasRequestedFirstPage, isFetching])
-    expect(mockUseEffect).toHaveBeenCalledTimes(7);
+    expect(wrapper.find("CatalogItems>CatalogItem>ChartCatalogItem").length).toBe(0);
```
andresmgot, line 343: you can just do `find(ChartCatalogItem)`, no?

dashboard/src/components/Catalog/Catalog.test.tsx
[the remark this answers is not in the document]
andresmgot: this test is still not there, no?

andresmgot (approved):
Thanks for the changes, this is almost done :) Accepting it already once you add that test I mention

dashboard/src/components/Catalog/Catalog.test.tsx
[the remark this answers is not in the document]
antgamdia: Ups!! I totally forgot and missed this comment, sorry!

absoludity:
Don't wait for a +1 from me, unless there's something specific you want input on. I'll trust Andres' review :)

dashboard/src/components/Catalog/Catalog.test.tsx
```diff
@@ -435,6 +435,46 @@ describe("pagination and chart fetching", () => {
...
+      <Catalog {...populatedProps} resetRequestCharts={resetRequestCharts} charts={charts} />,
+    );
+
+    const input = wrapper.find("input").findWhere(i => i.prop("value") === "foo");
+    input.simulate("change", { target: { value: "foo" } });
+
+    spyOnUseState.mockRestore();
+    expect(setPage).toHaveBeenCalledWith(2); // changes page to 2
+    expect(resetRequestCharts).toHaveBeenCalled(); // but receives a change in a filter
+    expect(wrapper.find("CatalogItems").prop("page")).toBe(1); // and page is again 1
```
andresmgot, line 476: I believe this is not ensuring what you want. You can verify it commenting the `useEffect(...)` that reset the page and the charts. If you do that, this test will still pass I believe.

I would do a much simpler test, without the `setState` mock. I would just check that after the filter change, `resetRequestCharts` is called twice (I belive it's called once the first time the component is rendered). We would not be checking that the page number is reset but since it's in the same piece of logic it's worth it not to add extra complexity in the test.

dashboard/src/components/Catalog/Catalog.test.tsx
```diff
@@ -366,7 +366,7 @@ describe("pagination and chart fetching", () => {
       />,
     );
     expect(wrapper.find("CatalogItems").prop("page")).toBe(1);
-    expect(wrapper.find("CatalogItems>CatalogItem>ChartCatalogItem").length).toBe(0);
+    expect(wrapper.find("ChartCatalogItem").length).toBe(0);
```
andresmgot, line 369: btw, just fyi in case you don't know, the difference between using `wrapper.find(Component)` and `wrapper.find("Component")` is that the second verifies that there is a component named  "Component" while the first ensures that the exact component (the one you are importing in the test file) is present.

You don't need to change this, it's just that I usually do the former since it's more accurate, in case there are two components that are named the same.

andresmgot (approved):
Thanks, still +1 but I would change the test.

dashboard/src/components/Catalog/Catalog.test.tsx
antgamdia: Ahhh, good to know! Thanks for the tip!

dashboard/src/components/Catalog/Catalog.test.tsx
antgamdia: I just commented 
`  useEffect(() => {
    setPage(1);
    resetRequestCharts();}, ...` and the test is not passing. Besides, when moving to the simpler test, it seems `resetRequestCharts` is called just once (but, I agree with you, it should be twice; I will check it)

antgamdia: As per the offline conversation: `setPage` is not really changing the page `expect(wrapper.find("CatalogItems").prop("page")).toBe(2); ---> false`.
Besides, implementing the test for testing this is not that easy. For instance, trying to set a repo search filter with 'checked=true` didn't work.

The agreed decision has been just to add a //TODO here and go ahead with this PR (so that the release can be out asap).

antgamdia:
Merging finally!!!  :crossed_fingers:

PR #2265, opened 2021-01-20
Add small padding bottom to the OperatorList

antgamdia:
### Description of the change

This PR simply introduces a small bottom padding to the Operator catalog, so the UI remains more or less consistent with the general chart catalog.

### Benefits

Operator catalog will no longer be like:
image

... to be like:

image

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

N/A

### DIFF
--- a/dashboard/src/components/OperatorList/OperatorList.scss
+++ b/dashboard/src/components/OperatorList/OperatorList.scss
@@ -1,0 +1,3 @@
+.operatorListContainer {
+  padding-bottom: 5vh;
+}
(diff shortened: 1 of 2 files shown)


absoludity (approved):
Thanks for the before screenshot, an after screenshot next time would be handy to compare :) +1

antgamdia:
>  screenshot next time would be handy to compare :) +1

I already have it indeed, but totally forgot to paste it in the issue :stuck_out_tongue_closed_eyes: 
Description updated

andresmgot (approved):

PR #2266, opened 2021-01-20
Clear search filter when deleting text

antgamdia:
### Description of the change

After the changes in #2249, some minor bugs are popping up. This PR is just to force the clear when the text input has been deleted.

The current (master) status is:

Note that "Clear all" and `?Search=` remains despite the user input has been deleted.

textnox

### Benefits

Users won't be shown a "clear all" even if they already cleared the input, that is:

textok

### Possible drawbacks

N/A (I hope so)

### Applicable issues

N/A

### Additional information

Sometimes I'm noticing a sync problem between the text at the search filter (the `?Search=xxx` value) and the text input content.  Perhaps we can look into this other issue in another PR.

### DIFF
--- a/dashboard/src/components/Catalog/Catalog.tsx
+++ b/dashboard/src/components/Catalog/Catalog.tsx
@@ -175,7 +175,7 @@ function Catalog(props: ICatalogProps) {
   const setSearchFilter = (searchTerm: string) => {
     const newFilters = {
       ...filters,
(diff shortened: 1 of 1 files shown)


absoludity (approved):
Unrelated to this PR, but do you know why the whole content flashes when deleting text in your screenshot? (as though there's some intermediate state where the categories and content are blank)

andresmgot (approved):

PR #2267, opened 2021-01-21
Fix websocket behavior in Firefox

### LINKED ISSUE #2258: All k8s resources for a release polled continuously even after navigating away on Firefox
### Description:

Watching the network requests while viewing a release shows that:
1. a bunch of connection upgrade requests go through to watch the resources,
2. the dashboard additionally begins polling all resources on a timeout
3. all resources continue to be polled after switching to another page.

These continue to accumulate as you visit other app releases so that you end up flooding the backend.

Questions:
1. Why are we both upgrading to a websocket successfully and then also polling? Was this from the work to resolve the issues with websockets? I would have thought we should only be polling if the websocket connections failed.
2. Why are we watching/polling resources on completed releases? Yes, they could go into a different state which the websocket will handle, but if we're polling we should not be polling with the same frequency as when initially running an app?
3. Why are we watching/polling *all* resources (ie. not just those that change state, but also those that don't, such as roles, clusterroles etc etc.)
4. Why are we continuing to watch/poll the resources after navigating away from the release page (guessing this is just a bug)

1. Open your Kubeapps URL and notice in the Network tab
2. Navigate to view the Kubeapps release and watch the network tab for a minute. You'll see ~30 101 upgrades to websocket, followed by ~30-40 requests every 5s
3. Navigate back to the Apps and watch the polling continue
4. navigate back to the Kubeapps release and watch more items being added to the polling.

Repeate 3-4 and you'll be flooding the backend in no time :)

The (unnecessary?) polling accumulates until the backend is flooded with requests

1. Polling wouldn't be required when the websocket upgrade is successful
2. Polling wouldn't poll all resources, just those that change
3. Even for resources that can change after a release, polling would not be as frequent
4. Polling wouldn't continue after navigating away from a release

andresmgot:
### Description of the change

In Firefox, there is an issue that a warning is thrown when opening a websocket: https://github.com/socketio/socket.io/issues/2016

This causes Kubeapps to fallback to the periodic polling mechanism but it's still receiving events from the websocket. It's also losing the reference to the "timer" that needs to be cleaned when the component is unmounted.

This PR includes two fixes:

 - If the websocket receives an event, remove the timer so the fallback method is not executed.
 - The timer is stored in the Kube state, identified uniquely by the resource is polling so it's not possible to create two timers for the same resource.

### Applicable issues

  - fixes #2258

### DIFF
--- a/dashboard/src/actions/kube.test.tsx
+++ b/dashboard/src/actions/kube.test.tsx
@@ -120,7 +120,7 @@ describe("getAndWatchResource", () => {
         payload: {
           ref,
           handler: expect.any(Function),
-          onError: { onErrorHandler: expect.any(Function) },
+          onError: expect.any(Function),
         },
       },
     ];
@@ -169,7 +169,7 @@ describe("getAndWatchResource", () => {
         payload: {
           ref: r,
           handler: expect.any(Function),
-          onError: { onErrorHandler: expect.any(Function) },
+          onError: expect.any(Function),
         },
       },
     ];
@@ -238,7 +238,7 @@ describe("getAndWatchResource", () => {
         payload: {
           ref: r,
           handler: expect.any(Function),
-          onError: { onErrorHandler: expect.any(Function) },
+          onError: expect.any(Function),
         },
       },
     ];
@@ -256,7 +256,7 @@ describe("getAndWatchResource", () => {
     );
 
     const watchFunction = (testActions[1].payload as any).handler as (e: any) => void;
-    const onErrorFunction = (testActions[1].payload as any).onError.onErrorHandler as () => void;
+    const onErrorFunction = (testActions[1].payload as any).onError as () => void;
     onErrorFunction();
     watchFunction({ data: `{"object": ${JSON.stringify(svc)}}` });
     const newActions = [
--- a/dashboard/src/actions/kube.test.tsx
+++ b/dashboard/src/actions/kube.test.tsx
@@ -120,7 +120,7 @@ describe("getAndWatchResource", () => {
         payload: {
           ref,
           handler: expect.any(Function),
-          onError: { onErrorHandler: expect.any(Function), closeTimer: expect.any(Function) },
+          onError: { onErrorHandler: expect.any(Function) },
         },
       },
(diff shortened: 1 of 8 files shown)


absoludity (approved):
Excellent, thanks for investigating and fixing. Did you find anything from mozilla about why the error is created if the ws connection is fine? If it is in some odd state, it could be that we remove the backup timers, but then stop getting updates on the websocket? (not important, just thinking about possibilities when we don't know why the error is being raised).

andresmgot:
> Did you find anything from mozilla about why the error is created if the ws connection is fine? If it is in some odd state, it could be that we remove the backup timers, but then stop getting updates on the websocket? (not important, just thinking about possibilities when we don't know why the error is being raised).

Yes, I forgot to update this. The root of the error is that the component (ApplicationSummary) that do the websockets gets mounted/unmounted a couple of times when the app is loading, causing the websocket to be opened-closed-opened-closed-opened and the error is thrown if the websocket is closed before it finishes to connect. That's a warning in Chrome but an error in Firefox.

I unsuccessfully tried to address that error moving the code to a component that doesn't get unmounted (AppView) but the problem persists, I guess it's something related to how `useEffect` cleanup action works.

absoludity:
> > Did you find anything from mozilla about why the error is created if the ws connection is fine? If it is in some odd state, it could be that we remove the backup timers, but then stop getting updates on the websocket? (not important, just thinking about possibilities when we don't know why the error is being raised).
> 
> Yes, I forgot to update this. The root of the error is that the component (ApplicationSummary) that do the websockets gets mounted/unmounted a couple of times when the app is loading, causing the websocket to be opened-closed-opened-closed-opened and the error is thrown if the websocket is closed before it finishes to connect. That's a warning in Chrome but an error in Firefox.
> 
> I unsuccessfully tried to address that error moving the code to a component that doesn't get unmounted (AppView) but the problem persists, I guess it's something related to how `useEffect` cleanup action works.

Oh nice. That'd be great as a comment with a link to more info in case someone ever  wants to jump in (or as a link to an issue or something), if you get a chance.

Issue #2268, opened 2021-01-22
Enable the internal pinniped-proxy to run with TLS certs 


absoludity:
### Description:

For the future case of Kubeapps running via pinniped-proxy, the internal requests proxied from our backend via the pinniped-proxy service should be on a secure channel. Currently requests from our backend to the pinniped-proxy service are without TLS (as we've not yet added TLS support to pinniped-proxy)

antgamdia:
Have a look at the changelog, now kube-rs and its related projects have updated the HTTP client to make it more configurable. Perhaps useful for this issue: 
https://github.com/clux/kube-rs/blob/master/CHANGELOG.md#tls-enhancements

antgamdia:
I've just come across this TODO in the code that perhaps is relevant for this issue:

https://github.com/vmware-tanzu/kubeapps/blob/85f72b22af7b29332a896a257f907caf88f2ab39/pkg/agent/httpConfigForCluster.go#L28-L31

absoludity:
Yeah, it actually references this issue :P (usually create an issue when adding the todo). Thanks!

antgamdia:
> Yeah, it actually references this issue :P (usually create an issue when adding the todo). Thanks!

@absoludity , did you have a chance to look at this //TODO when working on this issue and you left it unchanged for a reason? Or is it still something to do ? 
I just came across it now that I'm working on #5014 , but there's no rush.

PR #2269, opened 2021-01-22
Helm with pinniped

absoludity:
### Description of the change

While testing and demoing Kubeapps on a non-oidc cluster using pinniped, I've found that although helm releases are fetched correctly, I was not able to create a new release (deploy an app). Instead I'd get the cryptic message: Release "..." failed and has been uninstalled: unable to build kubernetes objects from release manifest: unknown

deploy-error

I assumed the issue was in pinniped-proxy, so began debugging there only to find that certain requests were hitting pinniped-proxy without the `Authorization` header with the bearer token. Confused, I then looked to kubeops where we create the requests via the helm API. After a lot of hunting around, the issue was pinpointed in that *something* in the internals of `genericclioptions.ConfigFlags.ToRestConfig()` was stripping off the bearer token if the API host was not secure, which makes sense for general (and external) interaction with an API server. This was only happening in requests from helm that use the discovery client on the API server. Everything else just works with the existing config. See the unit-tests which fail without this change.

Currently the internal pinniped-proxy service isn't (yet) running with TLS (though I'd like to make that happen separately - see #2268 ), so the only way I found to ensure the correct config (with the bearer token) is used for these specific helm requests to the discovery endpoints, was to implement parts of the `genericclioptions.RESTClientGetter` interface with a version that returns the original config.
 
### Benefits

I can install apps in Kubeapps on a non-oidc cluster.

### Possible drawbacks

It's not specific to this change (as we were already using http for traffic to the internal pinniped-proxy service) but we should update that to tls-only.

### Applicable issues

  - Ref #2181 

### Additional information

I got sick of looking for the actual code which is removing or not including the token, but I'm really keen to find it. If you're keen to get lost in the depths of client-go and join the hunt, I won't stop you :P

### DIFF
--- a/pkg/agent/agent_test.go
+++ b/pkg/agent/agent_test.go
@@ -15,6 +15,7 @@ import (
 	"helm.sh/helm/v3/pkg/storage"
 	"helm.sh/helm/v3/pkg/storage/driver"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 	chartv1 "k8s.io/helm/pkg/proto/hapi/chart"
 
 	"github.com/google/go-cmp/cmp"
@@ -703,6 +704,45 @@ func TestUpgradeRelease(t *testing.T) {
 			//check original version is superseded
 			rel, err = cfg.Releases.Get(tc.release, revisionBeingUpdated)
 			if got, want := rel.Info.Status, release.StatusSuperseded; got != want {
+				t.Errorf("got: %q, want: %q", got, want)
+			}
+		})
+	}
+}
+
+func TestNewConfigFlagsFromCluster(t *testing.T) {
+	testCases := []struct {
+		name   string
+		config rest.Config
+	}{
+		{
+			name: "bearer token remains for an https host",
+			config: rest.Config{
+				Host:        "https://example.com/",
+				APIPath:     "",
+				BearerToken: "foo",
+			},
+		},
+		{
+			name: "bearer token remains for an http host",
+			config: rest.Config{
+				Host:        "http://example.com/",
+				APIPath:     "",
+				BearerToken: "foo",
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			restClientGetter := NewConfigFlagsFromCluster("namespace-a", &tc.config)
+
+			config, err := restClientGetter.ToRESTConfig()
+			if err != nil {
+				t.Fatalf("%+v", err)
+			}
+
+			if got, want := config.BearerToken, tc.config.BearerToken; got != want {
 				t.Errorf("got: %q, want: %q", got, want)
 			}
 		})
--- a/pkg/agent/httpConfigForCluster.go
+++ b/pkg/agent/httpConfigForCluster.go
@@ -1,0 +1,112 @@
+package agent
+
+import (
+	"path/filepath"
+	"regexp"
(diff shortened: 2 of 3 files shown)


pkg/agent/httpConfigForCluster.go
```diff
@@ -0,0 +1,112 @@
...
+type configForCluster struct {
+	config         *rest.Config
+	discoveryBurst int
+	*genericclioptions.ConfigFlags
+}
+
+// ToRESTConfig has the main difference from the original implementation,
+// returning the config as is.
+func (f *configForCluster) ToRESTConfig() (*rest.Config, error) {
+	log.Debug("configForCluster.ToRESTConfig called")
```
andresmgot, line 39: do you want to keep these debug messages?

pkg/agent/httpConfigForCluster.go
```diff
@@ -0,0 +1,112 @@
...
+
+// overlyCautiousIllegalFileCharacters matches characters that *might* not be supported.  Windows is really restrictive, so this is really restrictive
+var overlyCautiousIllegalFileCharacters = regexp.MustCompile(`[^(\w/\.)]`)
+
+// computeDiscoverCacheDir takes the parentDir and the host and comes up with a "usually non-colliding" name.
+func computeDiscoverCacheDir(parentDir, host string) string {
+	// strip the optional scheme from host if its there:
+	schemelessHost := strings.Replace(strings.Replace(host, "https://", "", 1), "http://", "", 1)
+	// now do a simple collapse of non-AZ09 characters.  Collisions are possible but unlikely.  Even if we do collide the problem is short lived
+	safeHost := overlyCautiousIllegalFileCharacters.ReplaceAllString(schemelessHost, "_")
```
andresmgot, line 110: why don't you use the `sha` of the host?

pkg/agent/agent_test.go
```diff
@@ -708,3 +709,42 @@ func TestUpgradeRelease(t *testing.T) {
...
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			restClientGetter := NewConfigFlagsFromCluster("namespace-a", &tc.config)
+
+			config, err := restClientGetter.ToRESTConfig()
+			if err != nil {
+				t.Fatalf("%+v", err)
+			}
+
+			if got, want := config.BearerToken, tc.config.BearerToken; got != want {
```
andresmgot, line 745: at least I understand this test :sweat_smile:

andresmgot (approved):
This should have been difficult to debug. Thanks!

pkg/agent/agent_test.go
absoludity: Heh, yeah - I was so happy to be able to reproduce the actual issue in a unit test before fixing it, once I'd found the issue.

pkg/agent/httpConfigForCluster.go
absoludity: Yes, I did - they were incredibly helpful for debugging, but then, hmm, they are very specific to debugging this issue. Perhaps I'll remove. Thanks.

pkg/agent/httpConfigForCluster.go
absoludity: I didn't want to change this code from the original implementation at all - I just needed to reproduce it here in this implementation, as it's (a private) part of the default implementation for `genericclioptions.ConfigFlags`.

PR #2272, opened 2021-01-22
Implement checksum method for OCI registries

andresmgot:
### Description of the change

Add the method to get the "checksum" of an OCI registry. To achieve this, the method retrieves the list of tags for each repository (app) and calculates the checksum of the aggregated list.

### Possible drawbacks

For large registries, a request will be made per repository which might bloat the registry API.

### Applicable issues

  - ref #2232

### DIFF
--- a/cmd/asset-syncer/utils.go
+++ b/cmd/asset-syncer/utils.go
@@ -249,7 +249,6 @@ func doReq(url, authHeader string) ([]byte, error) {
 
 // Checksum returns the sha256 of the repo
 func (r *OCIRegistry) Checksum() (string, error) {
-	// repositoryAPIurl := fmt.Sprintf("%s://%s/v2%s", url.Scheme, url.Host, url.Path)
 	content := []byte{}
 	for _, appName := range r.repositories {
 		url, err := parseRepoURL(r.RepoInternal.URL)
--- a/cmd/asset-syncer/utils.go
+++ b/cmd/asset-syncer/utils.go
@@ -23,7 +23,6 @@ import (
 	"crypto/sha256"
 	"crypto/tls"
 	"crypto/x509"
-	"errors"
 	"fmt"
 	"image"
 	"io"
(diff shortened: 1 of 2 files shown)


cmd/asset-syncer/utils.go
```diff
@@ -222,10 +221,51 @@ type OCIRegistry struct {
...
-	return "", nil
+	content := []byte{}
+	for _, appName := range r.repositories {
+		url, err := parseRepoURL(r.RepoInternal.URL)
+		if err != nil {
+			return "", err
+		}
+		// Retrieve the list of tags to add it to the list
+		// Caveat: Mutated image tags won't be detected as new
+		url.Path = path.Join("v2", url.Path, appName, "tags/list")
```
absoludity, line 260: `"tags", "list")` since we're using `path.Join` anyway?

cmd/asset-syncer/utils.go
```diff
@@ -222,10 +221,51 @@ type OCIRegistry struct {
...
+	}
+
+	if res.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("request failed: %v", err)
+	}
+
+	return ioutil.ReadAll(res.Body)
+}
+
 // Checksum returns the sha256 of the repo
```
absoludity, line 250:
```
// by concatenating tags for all repositories within the registry and returning the sha256.
```

cmd/asset-syncer/utils.go
```diff
@@ -222,10 +221,51 @@ type OCIRegistry struct {
...
-	// TBD
-	return "", nil
+	content := []byte{}
+	for _, appName := range r.repositories {
+		url, err := parseRepoURL(r.RepoInternal.URL)
+		if err != nil {
+			return "", err
+		}
+		// Retrieve the list of tags to add it to the list
+		// Caveat: Mutated image tags won't be detected as new
```
absoludity, line 259: Not necessary now, but I guess we could have a list of exceptions (`latest`) for which we actually grab the digest?

cmd/asset-syncer/utils_test.go
```diff
@@ -707,3 +707,70 @@ func Test_fetchAndImportFiles(t *testing.T) {
...
+}
+
+type authenticatedChecksumHTTPClient struct{}
+
+func (h *authenticatedChecksumHTTPClient) Do(req *http.Request) (*http.Response, error) {
+	w := httptest.NewRecorder()
+
+	// Ensure we're sending the right Authorization header
+	if !strings.Contains(req.Header.Get("Authorization"), "Bearer ThisSecretAccessTokenAuthenticatesTheClient") {
+		w.WriteHeader(500)
```
absoludity, line 733: Fine as is, but given these functions (`Do`) are test only, if you pass a `t *testing.T` to the function you can `t.Fail()` here rather than having to return an http error and interpret that in the code.

cmd/asset-syncer/utils_test.go
```diff
@@ -707,3 +707,70 @@ func Test_fetchAndImportFiles(t *testing.T) {
...
+	return w.Result(), nil
+}
+
+type authenticatedChecksumHTTPClient struct{}
+
+func (h *authenticatedChecksumHTTPClient) Do(req *http.Request) (*http.Response, error) {
+	w := httptest.NewRecorder()
+
+	// Ensure we're sending the right Authorization header
+	if !strings.Contains(req.Header.Get("Authorization"), "Bearer ThisSecretAccessTokenAuthenticatesTheClient") {
```
absoludity, line 732: This should be equality I think?

absoludity (approved):
Would be good if the oci api actually had an endpoint for the digest of the registry, but maybe that'd be unmaintainable when adding large blobs.

cmd/asset-syncer/utils.go
andresmgot: In theory, `latest` should not be used for charts: https://github.com/helm/community/blob/master/hips/hip-0006.md#will-i-be-able-to-use-custom-tags-such-as-latest

cmd/asset-syncer/utils_test.go
andresmgot: well, this is also testing that getting an error code in the response is translated to a function error.

Issue #2253, opened 2021-01-16
[Question or Feature Request] Make bearer token location in browser storage configurable


mecampbellsoup:
## Background

We are writing a new authentication backend server for our cloud PaaS product/platform.

If we write an OIDC-compliant auth server (i.e. our own IdP), we can configure kubeapps to use this viva `oauth2-proxy` for authenticating users in a more seamless way than having them copy-paste tokens. This is all well-documented e.g. [here](https://github.com/kubeapps/kubeapps/blob/master/docs/user/using-an-OIDC-provider.md).

However, it will be much faster for us to implement a simple [webhook token auth server](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#webhook-token-authentication) backend instead. Moreover we aren't in love with JWTs in general and are leaning towards tried-and-true session management instead.

We intend to host kubeapps on our cloud platform's domain, e.g. `cloud.coreweave.com/apps`, so if we are able to write a cookie for this domain as part of normal user login flow on `cloud.coreweave.com`, we just need a way to tell kubeapps to read some value from a cookie and use it for bearer token auth.

I found [these lines of code in the frontend app]()https://github.com/kubeapps/kubeapps/blob/9a7b52c6911e8d97f470677722043b3ae765037e/dashboard/src/shared/Auth.ts#L10-L12 whereby the `kubeapps_auth_token` is read out of the browser's local storage.

## Current Behavior

The kubeapps frontend _always_ looks in `localStorage` for the token to authenticate requests to the kubeapps backend.

## Proposal

Make the choice of "in-browser token storage backend" configurable, e.g `document.cookie` instead of `localStorage` for the `kubeapps_auth_token`. That would be a simple and elegant solution to our use-case, I believe.

We could write a cookie containing an opaque token for that user after they login as normal. Upon navigating to `/apps` we can render kubeapps, and kubeapps will now be configured to read from cookies (instead of local storage, the current default token location in the browser) in order to obtain the auth token to be sent w/ requests to the backend.

absoludity:
Hi there, just a bit of background / clarification (in case it's not obvious)

> whereby the `kubeapps_auth_token` is read out of the browser's local storage.

This is only for demo's or similar where a K8s service account token is provided directly to Kubeapps and used for authentication. We don't recommend this for anything other than demo (nor storing any credential client side generally).

[...]

PR #2274, opened 2021-01-22
feat(auth): enable implicit cookie-based SSO authentication

mecampbellsoup:
## Description

The purpose of this PR is to enable cookie SSO authentication configuration. (What a mouthful!)

More specifically, the idea is to facilitate a use-case where:

1. k8s API is configured to do custom webhook token auth
2. A user visiting kubeapps has already signed in with this hypothetical "cloud provider" - imagine kubeapps is hosted at `apps.company.com` and the user signed in already at `cloud.company.com`
3. `cloud.company.com` has written an `HttpOnly` cookie to `.company.dom` domain
4. This cookie can then be presented to the k8s API and bearer token auth performed against it

To facilitate such a flow, I see ~4 issues/things that need changed or configured:

1. The dashboard `/login` page currently relies on [local storage](https://github.com/mecampbellsoup/kubeapps/blob/9a7b52c6911e8d97f470677722043b3ae765037e/dashboard/src/shared/Auth.ts#L6-L19) to store tokens. It has been explained to me by @absoludity that this is [development-only behavior](https://github.com/kubeapps/kubeapps/issues/2253#issuecomment-761953949). For my proposed flow we would need the dashboard to simply try to make a request to e.g. `http://localhost:3000/api/clusters/default` to determine if the user is authenticated by the existing cookie stored in the browser. I [tried](https://github.com/mecampbellsoup/kubeapps/blob/375df7bb616812f0fe164ee1b0d91dd6870b57bd/dashboard/src/reducers/auth.ts#L16) to use [`Auth.isAuthenticatedWithCookie`](https://github.com/mecampbellsoup/kubeapps/blob/9a7b52c6911e8d97f470677722043b3ae765037e/dashboard/src/shared/Auth.ts#L110) but didn't want to spend hours messing around with async/await issues in the event this proposal is immediately shot down.
2. The name of the cookie to be read would need to be [configurable in the Helm chart](https://github.com/mecampbellsoup/kubeapps/blob/375df7bb616812f0fe164ee1b0d91dd6870b57bd/chart/kubeapps/values.yaml#L555-L557).
3. The kubeapps backend auth logic I believe should try to auth _its_ requests to k8s using the cookie if configured. I took a stab at that change [here](https://github.com/mecampbellsoup/kubeapps/blob/375df7bb616812f0fe164ee1b0d91dd6870b57bd/pkg/auth/authgate.go#L24-L32). 
3. I have to edit the frontend NGINX conf (code snippet of this below) to add `proxy_set_header Authorization "Bearer $cookie_sessionid";`. I can do this manually but would also be nice if it were configurable I think.

Here is how I amended my kubeapps frontend NGINX config (stored in the ConfigMap). Note the additions of `proxy_set_header`:

```
      # Helm returns a nil pointer error when accessing foo.bar if foo doesn't
      # exist, even with the `default` function.
      # See https://github.com/helm/helm/issues/8026#issuecomment-756538254
        # Otherwise we route directly through to the clusters with existing credentials.
        proxy_set_header Authorization "Bearer $cookie_sessionid";
        proxy_pass https://kubernetes.default;
      }

      # TODO: The following location is left for backwards compat but will no longer
      # be needed once clients are sending the cluster name.
      # Using regexp match instead of prefix one because the application can be
      # deployed under a specific path i.e /kubeapps
      location ~* /api/kube {
        rewrite /api/kube/(.*) /$1 break;
        rewrite /api/kube / break;
        proxy_set_header Authorization "Bearer $cookie_sessionid";
        proxy_pass https://kubernetes.default;
      }
```

## Alternative Solution

As an alternative, it would be simpler if we just put an auth gateway in front of our kubeapps deployment. This gateway could be the step at which the `sessionid` cookie is authenticated, and set as a header `Authorization: Bearer my-cookie-value`. However, issue number 1 in my list above still exists - i.e. the kubeapps frontend simply does not think you are logged in unless you (1) submit an auth token manually (local storage route, supposed to be non-production deployment), or (2) configure an OAuth proxy via Helm chart `.Values.authProxy`.

## Valediction

As I said, the purpose of this PR is to clarify the conversation I've been having with @absoludity in https://github.com/kubeapps/kubeapps/issues/2253.

I hope to generate some discussion and wanted to have pseudo-code to illustrate my intentions.

Thanks for taking the time to take a look!

### DIFF
--- a/dashboard/src/reducers/auth.ts
+++ b/dashboard/src/reducers/auth.ts
@@ -13,10 +13,17 @@ export interface IAuthState {
 }
 
 const getInitialState: () => IAuthState = (): IAuthState => {
-  const token = Auth.getAuthToken() || "";
+  // let authenticated = await Auth.isAuthenticatedWithCookie("default");
+  let authenticated = false;
+  if (!authenticated) {
+    const token = Auth.getAuthToken() || "";
+    if (token !== "") {
+      authenticated = true;
+    }
+  }
   return {
     sessionExpired: false,
-    authenticated: token !== "",
+    authenticated,
     authenticating: false,
     oidcAuthenticated: Auth.usingOIDCToken(),
   };
--- a/pkg/auth/authgate.go
+++ b/pkg/auth/authgate.go
@@ -22,7 +22,13 @@ type CheckerForRequest func(clustersConfig kube.ClustersConfig, req *http.Reques
 func AuthCheckerForRequest(clustersConfig kube.ClustersConfig, req *http.Request) (Checker, error) {
 	token := ExtractToken(req.Header.Get("Authorization"))
 	if token == "" {
-		return nil, fmt.Errorf("Authorization token missing")
+		// TODO: Get `sessionid` cookie name from Helm config `authProxy.cookieName`
+		cookie, err := req.Cookie("sessionid")
+		if err != nil {
+			return nil, fmt.Errorf("Authorization token missing")
+		} else {
+			token = cookie.Value
+		}
 	}
 	clusterName := mux.Vars(req)["cluster"]
 	return NewAuth(token, clusterName, clustersConfig)
--- a/chart/kubeapps/templates/dashboard-config.yaml
+++ b/chart/kubeapps/templates/dashboard-config.yaml
@@ -52,6 +52,7 @@ data:
       "kubeappsNamespace": "{{ .Release.Namespace }}",
       "appVersion": "v{{ .Chart.AppVersion }}",
       "authProxyEnabled": {{ .Values.authProxy.enabled }},
+      "cookieName": {{ .Values.authProxy.cookieName }},
       "oauthLoginURI": {{ .Values.authProxy.oauthLoginURI | quote }},
       "oauthLogoutURI": {{ .Values.authProxy.oauthLogoutURI | quote }},
       "authProxySkipLoginPage": {{ .Values.authProxy.skipKubeappsLoginPage }},
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -552,6 +552,10 @@ testImage:
 # Auth Proxy configuration for OIDC support
 # ref: https://github.com/kubeapps/kubeapps/blob/master/docs/user/using-an-OIDC-provider.md
 authProxy:
+  ## Set to the name of your auth cookie. Enables implicit session-based SSO.
+  # TODO: Default to "" as this should be an opt-in feature.
+  cookieName: "sessionid"
+
   ## Set to true if Kubeapps should configure the OAuth login/logout URIs defined below.
   #
   enabled: false
--- a/dashboard/public/config.json
+++ b/dashboard/public/config.json
@@ -3,6 +3,7 @@
   "kubeappsNamespace": "kubeapps",
   "appVersion": "DEVEL",
   "authProxyEnabled": false,
+  "cookieName": "sessionid",
   "authProxySkipLoginPage": false,
   "oauthLoginURI": "/oauth2/start",
   "oauthLogoutURI": "/oauth2/sign_out",


pkg/auth/authgate.go
```diff
@@ -22,7 +22,13 @@ type CheckerForRequest func(clustersConfig kube.ClustersConfig, req *http.Reques
...
 	token := ExtractToken(req.Header.Get("Authorization"))
 	if token == "" {
-		return nil, fmt.Errorf("Authorization token missing")
+		// TODO: Get `sessionid` cookie name from Helm config `authProxy.cookieName`
+		cookie, err := req.Cookie("sessionid")
+		if err != nil {
+			return nil, fmt.Errorf("Authorization token missing")
+		} else {
+			token = cookie.Value
+		}
```
absoludity, line 31: This backend code should not be requiring any cookie, since the authorization header should already be set by the frontend proxy.

absoludity (commented):
Hey there Matt. So from what I understand, I don't think you'll need to do any changes at all to Kubeapps (dashboard or backend), as we already have settings that tell the dashboard to assume an auth proxy is in use (ie. assume that there's a cookie set and it shouldn't ask about tokens). Though there could be some assumptions of OIDC (such as the login page asking you to click a button labelled "Login via OIDC provider", but we can ensure that label is configurable.

So instead, I think the options you'll need to deploy with are as follows (check the values.yaml for more info, but I'll comment below): 

```
authProxy:
  # enabled just ensures that kubeapps uses the oauth login/logout urls (which you can 
  # also configure as you need) to redirect the user for login or logout - not oauth specific.
  enabled: true 

  # external just ensures that Kubeapps assumes you're setting up your own auth proxy
  # so it won't run an `oauth2-proxy` container on the frontend nginx.
  external: true

  # skipKubeappsLoginPage tells kubeapps to redirect directly to the login url you provide when
  # unauthenticated.
  skipKubeappsLoginPage: true
```

With those settings your Kubeapps deployment won't ask for a token and will always be checking if its authenticated with a request to the api server. It will of course fail until you add your extra snippet to convert the cookie to an Authorization header, but once that's done, you should be able to verify that all requests heading to the api server (or Kubeapps backend) are being sent with the Authorization header (and so if everything is setup correctly, your webhook should be able to validate them).

pkg/auth/authgate.go
mecampbellsoup: Ah, makes sense, got it 👍

dashboard/src/reducers/auth.ts
```diff
@@ -13,10 +13,17 @@ export interface IAuthState {
...
+  if (!authenticated) {
+    const token = Auth.getAuthToken() || "";
+    if (token !== "") {
+      authenticated = true;
+    }
+  }
   return {
     sessionExpired: false,
-    authenticated: token !== "",
+    authenticated,
```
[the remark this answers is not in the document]
mecampbellsoup: OK sure, will revert, the code changes in this were largely just... thinking out loud, you might say 😄

mecampbellsoup:
> It will of course fail until you add your extra nginx config snippet to convert the cookie to an Authorization header

So at this point we have realized that we should just use our existing auth service as a [k8s authenticating proxy](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#authenticating-proxy) - instead of this "kubeapps -> auth proxy -> webhook token auth -> k8s API -> kubeapps" song and dance.

The goal is for the flow to be just "kubeapps -> auth proxy -> k8s API -> kubeapps".

In testing this I've hard-coded in the header I need for that auth strategy as follows:
```
      location ~* /api/clusters/default {
        rewrite /api/clusters/default/(.*) /$1 break;
        rewrite /api/clusters/default / break;

      # Helm returns a nil pointer error when accessing foo.bar if foo doesn't
      # exist, even with the `default` function.
      # See https://github.com/helm/helm/issues/8026#issuecomment-756538254
        # Otherwise we route directly through to the clusters with existing credentials.
        proxy_set_header X-Remote-User matt;
        proxy_pass https://kubernetes.default;
      }
```
In production if we auth our k8s API with an authenticating proxy as I'm proposing here, it's not clear to me how/at which step I would convert the cookie to the required `X-Remote-User` (and other) header(s).

In any case, you've been very helpful, I think I just need to slog through this but I definitely feel we are heading in the right direction. 😄  If you have any further thoughts on this stuff please let me know!

mecampbellsoup:
> > It will of course fail until you add your extra nginx config snippet to convert the cookie to an Authorization header
> 
> So at this point we have realized that we should just use our existing auth service as a [k8s authenticating proxy](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#authenticating-proxy) - instead of this "kubeapps -> auth proxy -> webhook token auth -> k8s API -> kubeapps" song and dance.
> 
> The goal is for the flow to be just "kubeapps -> auth proxy -> k8s API -> kubeapps".
> 
> In testing this I've hard-coded in the header I need for that auth strategy as follows:
> 
> ```
>       location ~* /api/clusters/default {
>         rewrite /api/clusters/default/(.*) /$1 break;
>         rewrite /api/clusters/default / break;
> 
>       # Helm returns a nil pointer error when accessing foo.bar if foo doesn't
>       # exist, even with the `default` function.
>       # See https://github.com/helm/helm/issues/8026#issuecomment-756538254
>         # Otherwise we route directly through to the clusters with existing credentials.
>         proxy_set_header X-Remote-User matt;
>         proxy_pass https://kubernetes.default;
>         include "./server_blocks/k8s-api-proxy.conf";
>       }
> ```
> 
> In production if we auth our k8s API with an authenticating proxy as I'm proposing here, it's not clear to me how/at which step I would convert the cookie to the required `X-Remote-User` (and other) header(s).
> 
> I know the auth proxy is supposed to set those headers, before it proxies the request to the k8s API, but nowhere in my kubeapps config have I described/configured the auth proxy URL. Is it take from `oauthLoginURI` or something?
> 
> In any case, you've been very helpful, I think I just need to slog through this but I definitely feel we are heading in the right direction. 😄

I suppose I could simply update the `nginx.conf` to swap `http://kubernetes.default` for `http://my-authenticating-proxy-service-thing`...? @absoludity

absoludity:
> > > It will of course fail until you add your extra nginx config snippet to convert the cookie to an Authorization header
> > 
> > So at this point we have realized that we should just use our existing auth service as a [k8s authenticating proxy](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#authenticating-proxy) - instead of this "kubeapps -> auth proxy -> webhook token auth -> k8s API -> kubeapps" song and dance.
> > The goal is for the flow to be just "kubeapps -> auth proxy -> k8s API -> kubeapps".

It's not clear to me why you're not doing what you originally proposed (convert cookie value to Authorization header as requests hit the Kubeapps frontend) - it seems simplest given the info you've described so far? There is no song and dance, it'd just be "kubeapps -> k8s API" (no auth proxy, just the k8s api then using whatever webhook you have). But I'm probably missing context?

> > In any case, you've been very helpful, I think I just need to slog through this but I definitely feel we are heading in the right direction. smile

:) I am getting lost in the details and switching plans now though.

> 
> I suppose I could simply update the `nginx.conf` to swap `http://kubernetes.default` for `http://my-authenticating-proxy-service-thing`...? @absoludity

I don't think so - Kubeapps talks to the k8s api from the nginx frontend yes, but also from the kubeops backend. Again, with the limited info I have, it'd be simpler if kubeapps just dealt with an Authorization header set by your snippet in the frontend as that would work for all requests.

> 
> One more thing I want to throw in here... k8s authenticating proxy **requires** a client cert file to be configured:
> 
> > --requestheader-client-ca-file Required. PEM-encoded certificate bundle. A valid client certificate must be presented and validated against the certificate authorities in the specified file before the request headers are checked for user names.
> 
> i.e. clients must present a PEM certificate that the server can validate against its `requestheader-client-ca-file`. Is this something configurable with `ingress` config block, e.g. https://github.com/kubeapps/kubeapps/blob/master/chart/kubeapps/values.yaml#L53? (I don't think I would want to enable ingress though since I want the kubeapps frontend to be the only way of interacting w/ kubeapps backend.)

I've not used authenticating proxies headers like this before, but I don't think this will work for Kubeapps as is (as you'd need to do this for every service which talks to the k8s API - both the nginx frontend as well as kubeops backend).

Sorry I can't be of more help! I'm keen to understand why the cookie -> authz header doesn't work for your webhook auth (or did you mean that you're no longer wanting to do web hook auth). Seems like this is taking more work than just using an oauth2 compliant service (like dex) which works out of the box and can talk to your upstream auth service.

mecampbellsoup:
> (or did you mean that you're no longer wanting to do web hook auth)

Yes, correct. Ha. (More context below, if you're curious.)

> It's not clear to me why you're not doing what you originally proposed (convert cookie value to Authorization header as requests hit the Kubeapps frontend) - it seems simplest given the info you've described so far? There is no song and dance, it'd just be "kubeapps -> k8s API" (no auth proxy, just the k8s api then using whatever webhook you have). But I'm probably missing context?

Ha, yea, sorry, I have really ruined this thread/convo! 😭 

After initial interest in the k8s [webhook token auth](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#webhook-token-authentication) pattern, we realized we could just put _everything_ in our cluster behind an API gateway like [Kong](https://github.com/Kong/kong/) or [HAProxy](https://www.haproxy.com/blog/using-haproxy-as-an-api-gateway-part-1/).

We are really enamored of the pattern where auth is performed upfront, in a single place (the gateway), and every app upstream of that then has a consistent "API" (for lack of a better word) - i.e. Grafana, Prometheus, k8s, kubeapps, everything can refer to a request header - set by the gateway - like `X-Request-User` that will contain the username or email of the authenticated user.

(In e.g. [Grafana auth proxy](https://grafana.com/docs/grafana/latest/auth/auth-proxy/) you would set `header_name=X-Request-User` in the `[auth.proxy]` section of `grafana.ini`. In k8s, you do something similar when configuring auth proxy authentication, via [`--requestheader-username-headers` apiserver configuration.](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#authenticating-proxy))

> Again, with the limited info I have, it'd be simpler if kubeapps just dealt with an Authorization header set by your snippet in the frontend as that would work for all requests.

So, it sounds like, as long as our API gateway sets an `Authorization` header, we should be pretty much good to go? Makes sense. The frontend will detect it, and try it w/ the k8s API, so as long as that succeeds the user should be logged in.

Shame, I was really looking forward to making a PR/working on kubeapps! Though I know we have some other ideas for contributing lined up, e.g. around charts for custom kube operators (i.e. deploying operators via kubeapps) but TBH I am not involved in that (yet...).

Thanks, as always, your time is very much appreciated. 😃

absoludity:
> So, it sounds like, as long as our API gateway sets an `Authorization` header, we should be pretty much good to go? Makes sense. The frontend will detect it, and try it w/ the k8s API, so as long as that succeeds the user should be logged in.

Yes, this is correct :)

> Shame, I was really looking forward to making a PR/working on kubeapps! Though I know we have some other ideas for contributing lined up, e.g. around charts for custom kube operators (i.e. deploying operators via kubeapps) but TBH I am not involved in that (yet...).

Great - let's see :)

> Thanks, as always, your time is very much appreciated. smiley
 :+1:

mecampbellsoup:
> Yes, this is correct :)
> 
> >

So, requests from kubeapps to k8s will present the `Authorization` token for authentication? That worked well for webhook token auth (our webhook endpoint would just lookup the session from the `Authorization "Bearer $cookie_sessionid"`); if we use an authenticating proxy instead I suppose our proxy will have to write in the configured `X-Request-User` header per the k8s auth proxy [docs](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#authenticating-proxy).

> The API server can be configured to identify users from request header values, such as X-Remote-User. It is designed for use in combination with an authenticating proxy, which sets the request header value.

> --requestheader-username-headers Required, case-insensitive. Header names to check, in order, for the user identity. The first header containing a value is used as the username.

In summary, AFAIUI, our API gateway needs to:

1. Add `Authorization` header so _kubeapps_ believes the request is already authenticated; and
2. Add the auth proxy headers for k8s API (assuming we are using an auth proxy and not a webhook endpoint for k8s auth).

mecampbellsoup:
Hi @absoludity I am _just about_ there but don't understand how to bypass the 'paste token here' login page. I wrote up my question with more detail in [Slack](https://kubernetes.slack.com/archives/C9D3TSUG4/p1612386018008900). Cheers!

absoludity:
> Hi @absoludity I am _just about_ there but don't understand how to bypass the 'paste token here' login page. I wrote up my question with more detail in [Slack](https://kubernetes.slack.com/archives/C9D3TSUG4/p1612386018008900). Cheers!

I think that's answered above in https://github.com/kubeapps/kubeapps/pull/2274#pullrequestreview-575020525 ? Let me know if not.

mecampbellsoup:
> > Hi @absoludity I am _just about_ there but don't understand how to bypass the 'paste token here' login page. I wrote up my question with more detail in [Slack](https://kubernetes.slack.com/archives/C9D3TSUG4/p1612386018008900). Cheers!
> 
> I think that's answered above in [#2274 (review)](https://github.com/kubeapps/kubeapps/pull/2274#pullrequestreview-575020525) ? Let me know if not.

I very nearly have this working now - I had to apply the follow changes to the NGINX configuration:

```
      location ~* /api/clusters/default {
        rewrite /api/clusters/default/(.*) /$1 break;
        rewrite /api/clusters/default / break;

      # Helm returns a nil pointer error when accessing foo.bar if foo doesn't
      # exist, even with the `default` function.
      # See https://github.com/helm/helm/issues/8026#issuecomment-756538254
        # Otherwise we route directly through to the clusters with existing credentials.
        proxy_set_header Host apps.coreweave.test;
        proxy_pass https://gatekeeper-kubernetes-ingress.default;
      }

      # TODO: The following location is left for backwards compat but will no longer
      # be needed once clients are sending the cluster name.
      # Using regexp match instead of prefix one because the application can be
      # deployed under a specific path i.e /kubeapps
      location ~* /api/kube {
        rewrite /api/kube/(.*) /$1 break;
        rewrite /api/kube / break;
        proxy_set_header Host apps.coreweave.test;
        proxy_pass https://gatekeeper-kubernetes-ingress.default;
      }
```
1. Change `kubernetes.default` to `gatekeeper-kubernetes-ingress.default` (which is our HAProxy service)
2. Add `proxy_set_header Host apps.coreweave.test` so HAProxy knows to execute the k8s API ingress

It looked like the browser entered some kind of infinite loop but I'm going to look at it with fresh eyes in the morning. It definitely looks promising though!

mecampbellsoup:
> > > Hi @absoludity I am _just about_ there but don't understand how to bypass the 'paste token here' login page. I wrote up my question with more detail in [Slack](https://kubernetes.slack.com/archives/C9D3TSUG4/p1612386018008900). Cheers!
> > 
> > 
> > I think that's answered above in [#2274 (review)](https://github.com/kubeapps/kubeapps/pull/2274#pullrequestreview-575020525) ? Let me know if not.
> 
> I very nearly have this working now - I had to apply the follow changes to the NGINX configuration:
> 
> ```
>       location ~* /api/clusters/default {
>         rewrite /api/clusters/default/(.*) /$1 break;
>         rewrite /api/clusters/default / break;
> 
>       # Helm returns a nil pointer error when accessing foo.bar if foo doesn't
>       # exist, even with the `default` function.
>       # See https://github.com/helm/helm/issues/8026#issuecomment-756538254
>         # Otherwise we route directly through to the clusters with existing credentials.
>         proxy_set_header Host k8s.coreweave.test;
>         proxy_pass https://gatekeeper-kubernetes-ingress.default;
>         include "./server_blocks/k8s-api-proxy.conf";
>       }
> 
>       # TODO: The following location is left for backwards compat but will no longer
>       # be needed once clients are sending the cluster name.
>       # Using regexp match instead of prefix one because the application can be
>       # deployed under a specific path i.e /kubeapps
>       location ~* /api/kube {
>         rewrite /api/kube/(.*) /$1 break;
>         rewrite /api/kube / break;
>         proxy_set_header Host k8s.coreweave.test;
>         proxy_pass https://gatekeeper-kubernetes-ingress.default;
>         include "./server_blocks/k8s-api-proxy.conf";
>       }
> ```
> 
> 1. Change `kubernetes.default` to `gatekeeper-kubernetes-ingress.default` (which is our HAProxy service)
> 2. Add `proxy_set_header Host apps.coreweave.test` so HAProxy knows to execute the k8s API ingress
> 
> It looked like the browser entered some kind of infinite loop but I'm going to look at it with fresh eyes in the morning. It definitely looks promising though!

Alright, now that my brain is functioning again... I wrote up more details in [Slack](https://kubernetes.slack.com/archives/C9D3TSUG4/p1612457562010400?thread_ts=1612386018.008900&cid=C9D3TSUG4) but in short, it I have requests from kubeapps' NGINX proxied to k8s authenticating as desired, but requests from kubeapps to the kubeapps backend are not authenticating (i.e. 403 response).

I believe this is because the kubeapps backend expects a token, correct?

PR #2278, opened 2021-01-25
Add initial API docs portal

### LINKED ISSUE #2222: Do we have the swagger API doc of Dashboard or KubeOPS?
Do we have the swagger API doc of Dashboard or KubeOPS?

When I want to use kubeapps through API, I did not find API documentation

Or, do we plan to create a Swagger API?

antgamdia:
### Description of the change

As the number of APIs used in Kubeapps is increasingly growing, having good documentation is essential to have a clear idea of what's under the hood.
This PR is to add an Interactive documentation portal.

### Benefits

Users will better understand what the Kubeapps components are exposing and, developers (newcomers, especially) will rapidly get an overview of what is inside Kubeapps.

### Possible drawbacks

Currenty, we don't follow any kind of API-first design when developing new components, as a result, all the API information is disseminated across the code. Maintaining API docs is always tedious, so we should check a way to automatically generate documentation from the source code.

### Applicable issues

  - related #2222 

### Additional information

It is still a draft PR since I haven't finished the API model - I did my best to discover the APIs (current and deprecated ones), but I  want to clarify a couple of things before going deeper.

Related to https://github.com/kubeapps/kubeapps/pull/2291

Some screenshots:

image

image


dashboard/src/components/ApiDocs/index.tsx
```diff
@@ -0,0 +1,21 @@
+import React from "react";
+import { Auth } from "shared/Auth";
+import SwaggerUI from "swagger-ui-react";
+import "swagger-ui-react/swagger-ui.css";
+
+// Request interface needed for avoiding type error in the requestInterceptor
+interface Request {
+  // tslint:disable-line
```
andresmgot, line 8: why are you disabling the linter here?

dashboard/src/components/ApiDocs/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+  [k: string]: any;
+}
+
+function authenticate(req: Request): Request {
+  const token = Auth.getAuthToken() || "";
+  req.headers.Authorization = `Bearer ${token}`;
+  req.url = req.url.replace("127.0.0.1:8080", `${window.location.host}`);
+  return req;
+}
```
andresmgot, line 17: can you try to use the current axios interceptor? (`axiosWithAuth` from `src/shared/AxiosInstance.ts`). This won't work when using OIDC for example.

dashboard/src/components/Header/Menu.tsx
```diff
@@ -86,6 +86,10 @@ function Menu({ clusters, appVersion, logout }: IContextSelectorProps) {
                 .
                 <br />
                 {appVersion}
+                <br />
+                <Link to={"/docs"} key="deploy-button">
```
andresmgot, line 90: why the `key`?

dashboard/src/components/ApiDocs/index.tsx
antgamdia: Because ts interfaces generally should start with `I`, but since I need it to override the type used in `requestInterceptor`.
Perhaps it is an issue with the `@types/swagger-ui-react module`.

dashboard/src/components/ApiDocs/index.tsx
antgamdia: Mmm, I don't get your point here :S 
I just want to exact the bearer token so that it can be used more easily. Otherwise, this token can also be manually entered:
image

However, the engine for performing the requests is not configurable AFAIK. Look at the props it exposes:

```javascript
    spec?: object | string;
    url?: string;
    onComplete?: (system: System) => void;
    requestInterceptor?: (req: Request) => Request | Promise<Request>;
    responseInterceptor?: (res: Response) => Response | Promise<Response>;
    docExpansion?: 'list' | 'full' | 'none';
    defaultModelExpandDepth?: number;
    plugins?: Plugin[];
    supportedSubmitMethods?: string[];
    deepLinking?: boolean;
```

dashboard/src/components/Header/Menu.tsx
antgamdia: My bad!  unnoticed during copypasting :S

antgamdia:
BTW, don't you think the color of the links is too dark?
image

dashboard/src/components/ApiDocs/index.tsx
andresmgot: okay, I guess it's okay, the only thing is that I would not add the header if `getAuthToken()` returns nothing (to avoid messing up with the cookies set if oauth2_proxy is enabled).

dashboard/src/components/ApiDocs/index.tsx
andresmgot: ah, that's weird, the swagger package is defining the Request interface but it's not exposing it.. Anyway, can you add a comment with that info?

andresmgot:
> BTW, don't you think the color of the links is too dark?
> image

Yes, probably we should use a thicker white style for the link (but I don't know which class should be used for a dark-themed link).

Also, can you change the icon? That one is used when the image cannot be loaded, Harbor uses "network-globe":

Screenshot from 2021-01-26 16-22-43

antgamdia:
> Also, can you change the icon? That one is used when the image cannot be loaded, Harbor uses "network-globe":

Mmm, the problem is we're using the old clarity icon system (@clr/icons instead of @cds/core) and "network-globe" is not available there.
As a result, I'm getting:

image

So: 1) we can set the `file` icon since it's working right now; 2) we can use the network-globe and wait until we upgrade the versions; 3) we can test each icon one by one and select the best one given the constraints.
I personally would leave the "file" one


The newest versions are
`    "@cds/city": "^1.1.0",
    "@cds/core": "^5.0.0",
    "@cds/react": "^5.0.0",`
But we should address this migration in another future moment :P 

https://clarity.design/foundation/icons/

andresmgot:
you don't need to upgrade, you just need to add the icon to the import list, check `components/Layout/Clarity.tsx`

antgamdia:
> Layout/Clarity.tsx

:man_facepalming:  Ouch! I supposed we already have the whole set of icons in react with @clr/icons  . Thx for letting me know!

dashboard/src/components/ApiDocs/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+// Request interface needed for avoiding type error in the requestInterceptor
+// it is being used, but it is not exported, so we define it here
+// tslint:disable-next-line
+interface Request {
+  [k: string]: any;
+}
+
+function authenticate(req: Request): Request {
+  const token = Auth.getAuthToken() || "";
+  // Otherwise, OIDC login is being used and we cannot automatically set the token here
```
absoludity, line 15: If the user is logged in, the cookie will be set (you can check) and the auth-proxy will be updating the request with the Authorization header anyway (ie. the user can test the API as themselves). Happy to check that once landed.

dashboard/public/openapi.yaml
```diff
@@ -0,0 +1,1901 @@
...
+          type: string
+        type:
+          type: string
+  securitySchemes:
+    bearerAuth:
+      type: http
+      scheme: bearer
+      bearerFormat: JWT
+security:
+  - bearerAuth: []
```
absoludity, line 1901: Woah - what did you use to generate and manually assemble that?

antgamdia: Currently, it has been a full-manual process using the types already defined in go :grimacing: 
There a couple of tools for doing some automation (https://github.com/go-swagger/go-swagger or https://github.com/swaggo/swag) but given how we are defining the APIs (multiple places, params fuzzily defined, proxypassing with nginx, etc) probably it won't work... I mean, it's something I want to check, but don't think it'll work.

antgamdia:
Given that you already review this PR and the only in-progress thing is the OAS doc itself, I've removed the API documentation itself to keep this PR as simple as possible.
I'll send another PR with the single OAS doc.

So, after merging this PR this doc portal will appear:
image

andresmgot (commented):
LGTM, but let's hold this until the new release is created to not include this page without the actual API docs.

dashboard/public/openapi.yaml
absoludity: Wow, that's 2k of manual openapi.yaml - impressive :)

andresmgot (approved):
We can land this already

PR #2279, opened 2021-01-25
Change comma by pipe to concat filters in the catalog

antgamdia:
### Description of the change

Following up https://github.com/kubeapps/kubeapps/pull/2221#issuecomment-760943588 and https://github.com/kubeapps/kubeapps/pull/2264#pullrequestreview-575212694.

This PR is to workaround this bug by simply change the default character for splitting and joining, the comma `,`, to another less frequent but also meaningful, the pipe `|`.

### Benefits

Filters values having a comma as part of their names won't break the UI anymore: 

image

### Possible drawbacks

I don't think so, but I'm not fully aware whether or not this "comma"-split way to generate the URL is being used anywhere else. 

### Applicable issues

N/A

### Additional information

N/A

### DIFF
--- a/dashboard/src/components/Catalog/Catalog.tsx
+++ b/dashboard/src/components/Catalog/Catalog.tsx
@@ -69,7 +69,7 @@ export function filtersToQuery(filters: any) {
   const activeFilters = Object.keys(filters).filter(f => filters[f].length);
   if (activeFilters.length) {
     const filterQueries = activeFilters.map(
-      filter => `${filter}=${filters[filter].map((f: string) => encodeURIComponent(f)).join(",")}`,
+      filter => `${filter}=${filters[filter].map((f: string) => encodeURIComponent(f)).join("|")}`,
     );
     query = "?" + filterQueries.join("&");
   }
@@ -107,7 +107,7 @@ function Catalog(props: ICatalogProps) {
   useEffect(() => {
     const newFilters = {};
     Object.keys(propsFilter).forEach(filter => {
-      newFilters[filter] = propsFilter[filter]?.toString().split(",");
+      newFilters[filter] = propsFilter[filter]?.toString().split("|");
     });
     setFilters({
       ...initialFilterState(),
@@ -167,7 +167,7 @@ function Catalog(props: ICatalogProps) {
 
   // Only one search filter can be set
   const searchFilter = propsFilter[filterNames.SEARCH]?.toString() || "";
-  const reposFilter = filters[filterNames.REPO]?.join(",") || "";
+  const reposFilter = filters[filterNames.REPO]?.join("|") || "";
   useEffect(() => {
     fetchCharts(cluster, namespace, reposFilter, page, size, searchFilter);
   }, [fetchCharts, cluster, namespace, reposFilter, page, size, searchFilter]);
--- a/dashboard/src/components/Catalog/Catalog.tsx
+++ b/dashboard/src/components/Catalog/Catalog.tsx
@@ -64,6 +64,10 @@ export function initialFilterState() {
   return result;
 }
 
+const tmpStrRegex = /__/g;
+const tmpStr = "__";
+const commaRegex = /,/g;
+
 export function filtersToQuery(filters: any) {
   let query = "";
   const activeFilters = Object.keys(filters).filter(f => filters[f].length);
@@ -75,7 +79,7 @@ export function filtersToQuery(filters: any) {
     const filterQueries = activeFilters.map(
       filter =>
         `${filter}=${filters[filter]
-          .map((f: string) => encodeURIComponent(f.replace(/,/g, "__")))
+          .map((f: string) => encodeURIComponent(f.replace(commaRegex, tmpStr)))
           .join(",")}`,
     );
     query = "?" + filterQueries.join("&");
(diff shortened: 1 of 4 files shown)


dashboard/src/components/Catalog/Catalog.tsx
```diff
@@ -107,7 +107,7 @@ function Catalog(props: ICatalogProps) {
   useEffect(() => {
     const newFilters = {};
     Object.keys(propsFilter).forEach(filter => {
-      newFilters[filter] = propsFilter[filter]?.toString().split(",");
+      newFilters[filter] = propsFilter[filter]?.toString().split("|");
```
andresmgot, line 110: I think we can leave the `,` as a separator since it's a common practice, we don't need to change that. The thing is that we don't support (or we don't want to support?) arrays for the text search. I would just ignore the `,` for the text search here:

```js
newFilters[filter] = filterNames.SEARCH === filter ? propsFilter[filter]?.toString() : propsFilter[filter]?.toString().split(",");
```

andresmgot (changes requested):

dashboard/src/components/Catalog/Catalog.tsx
andresmgot: Also, we need a test :)

antgamdia: I don't think it is a good solution. Indeed, I did notice this problem when filtering operators; look at this comment: https://github.com/kubeapps/kubeapps/pull/2221#issuecomment-760943588

What do you think? If implementing your solution, we will still have the same issue in operators (and there are plenty of companies with "," in their names...)

andresmgot: got it, then it's better to properly encode the values rather than just making the bug less probable

antgamdia: (oh no, encoding again!! :roll_eyes: ) 
According to the RFC 3986?, the characters that may be subject to URL %-encoding are:
`!     #     $     %     &     '     (     )     *     +     ,     /     :     ;     =     ?     @     [     ]` 

But, they are considered sub-delimiters and there is no actual need to encode them even if encodeURIComponent does.
However, the `qs` parser doesn't support it yet (https://github.com/ljharb/qs/issues/237).

Indeed, `qs` is decoding all the parameters, so we don't have control over the values we received.
For instance, even if I push the filters properly like `Provider=something1,some%2Cthing2`, the component will always receive `something1,some,thing2`, so the split/join won't work.
So, I don't think encoding it's a straightforward solution, perhaps we should dig deeper into the `qs` module and re-implement the parsing function or directly substitute `qs` with another alternative.

I'd come to an agreement with the delimiter character to use and would leave this PR as is (perhaps adding the ignore you mentioned for the text search as well) since it is "workarounding" an existing bug that prevents us from releasing.

Anyway, I'm open to implementing other alternatives, of course :)

andresmgot: > (oh no, encoding again!!  )

:laughing:

andresmgot: okay, so I would still like to maintain the `,` as a separator because it's the standard but it's true that encoding won't help. Let's replace `,` in the filters with a highly improbable string (I suggest a double underscore `__`) and then undo the change when reading the filter. WDYT?

antgamdia: Yes, I totally agree with this approach. Just a nitpick comment: if we use a non-to-be-encoded character like `_`, the URL will be like: ` ...?Provider=Lightbend__%20Inc` (the underscore is visible).
What if we use another symbol that should be encoded', for instance, the whitespace character in braille: `⠀` (U+2800)  (%E2%A0%80)
Anyway, I'm gonna start working with "__" and will update the PR asap.

andresmgot: I don't mind what characters to use, use what you think is best

dashboard/src/components/Catalog/Catalog.tsx
```diff
@@ -107,7 +114,8 @@ function Catalog(props: ICatalogProps) {
   useEffect(() => {
     const newFilters = {};
     Object.keys(propsFilter).forEach(filter => {
-      newFilters[filter] = propsFilter[filter]?.toString().split("|");
+      const filterValue = propsFilter[filter]?.toString() || "";
+      newFilters[filter] = filterValue.split(",").map(a => a.replace(/__/g, ","));
```
andresmgot, line 118: it's usually good to avoid regexps (they are more difficult to compute). You can do `replaceAll("__", ",")`

andresmgot (approved):
Thanks! I would avoid the regexp but it's a minor thing. LGTM

dashboard/src/components/Catalog/Catalog.tsx
antgamdia: That's was my first option indeed but... tests stop working! The problem is `replaceAll` is not so widely supported, (see https://caniuse.com/mdn-javascript_builtins_string_replaceall). In nodejs, it's supported from >=15, so Jest/Enzyme complains about it's not a function.
So I just replace the replaceAll by a normal replace :P 

Also, I think we should create the regex outside the loop indeed.

andresmgot: :O thought that was there from before, nevermind then

PR #2283, opened 2021-01-27
Pull OCI charts

andresmgot:
### Description of the change

This PR includes the code that pull charts from an OCI registry in order to collect the chart metadata (digest and Chart.yaml) and the rest of files we are interested in (values, readme, schema). 

The goal was to directly use Helm code as a library but the code that performs what we want is under a private package (`helm.sh/helm/v3/internal/experimental/registry`) so I copied part of that package here to use it. More info in the link you can find in the file `helm.go`.

Important to mention a big difference between the traditional flow and the OCI behavior. In the former, the `Charts()` method only returned metadata from the repository `index.yaml` and the rest of the files (readme, etc) was fetched in `FetchFiles()`. For the OCI, we need to pull the artifact to obtain its metadata so we are also extracting the files at that point (so `FetchFiles()` basically does nothing now).

### Possible drawbacks

We will miss fixes from the forked code.

Also, one difference from traditional helm charts is that I'm pulling charts sequentially (not in parallel). That's marked as future point of improvement.

### Applicable issues

  - ref #2232


cmd/asset-syncer/utils_test.go
```diff
@@ -737,6 +738,16 @@ func (h *authenticatedChecksumHTTPClient) Do(req *http.Request) (*http.Response,
...
+type fakeOCIPuller struct {
+	content  *bytes.Buffer
+	checksum string
+	err      error
+}
+
+func (f *fakeOCIPuller) pullOCIChart(ociFullName string) (*bytes.Buffer, string, error) {
+	return f.content, f.checksum, f.err
+}
+
```
absoludity, line 750: Nice'n'simple fake!

cmd/asset-syncer/utils.go
```diff
@@ -358,17 +361,83 @@ func extractFilesFromBuffer(buf *bytes.Buffer) (*artifactFiles, error) {
...
+			// Pull Chart
+			// TODO: Run in parallel
+			ref := path.Join(url.Host, url.Path, fmt.Sprintf("%s:%s", appName, tag))
+			chartBuffer, digest, err := r.puller.pullOCIChart(ref)
+			if err != nil {
+				return nil, err
+			}
+
+			// Extract
+			files, err := extractFilesFromBuffer(chartBuffer)
```
absoludity, line 383: So we're loading the files for every tag into memory here I think. Is that going to be a problem if there's many tags for an app? (or do we implicitly do this already with the current non-oci?)

absoludity (approved):

cmd/asset-syncer/utils.go
andresmgot: we are extracting one tag at a time, it should not be a problem

absoludity: > we are extracting one tag at a time,

Not sure what you mean? The code above iterates all the tags for the app and appears to extract the file contents into memory for each tag (ie. could be 200 copies of the same/similar readme, icons etc. in memory). Probably not that big, but I'm just not sure if we were doing this before or explicitly appending the file data later (one at a time).

Or how do you mean that it currently does them one at a time (as in, writes the file and frees the memory before getting the next tag)? 

> it should not be a problem

Hopefully not :)

andresmgot: sorry, yes, I was thinking just about this `files` var but it's true that we are storing all these files in `ChartVersions` for later. Taking Jenkins as an example, the three files there (readme, values, schema) (note that the icons are not fetched here), are ~60KB. The problem is that this is per version. If we have 10 versions it would be 600KB. And then if we have 200 apps the count goes up to 120MB, it can really be a problem for large repositories.

Just thinking out loud here but maybe it's better to just store the pointer to the `Reader` to those files so we can read and store the files just when needed. Will evaluate that.

PR #2284, opened 2021-01-28
Handle pinniped authentication errors as unauthorized.

### LINKED ISSUE #2181: Turn pinniped-proxy POC into production ready service.

absoludity:
### Description of the change

Up until now, if pinniped-proxy received a response from pinniped identifying that it could not exchange credentials (such as "authentication failed") it would return a 500.

This PR updates to recognise a successful request to pinniped that doesn't contain credentials and returns some context of a specific error so it can be handled differently in the service.

### Benefits

We don't return 500 internal server errors for errors that aren't internal server errors.

The user just sees a 401 request with a limited message.

### Possible drawbacks

An error such as this is a configuration error (ie. the user does not have access yet, or the jwt authenticator has the incorrect audience (this is how I tested it)), so checking logs is required to solve.

We should update the oidc debugging faq with pinniped specific steps once we document this work.

### Applicable issues

  - fixes #2181

### Additional information

Here's an example request returning a 401 from pinniped-proxy (after I manually changed the audience of the jwt authenticator):

unauthorized-by-pinniped

### DIFF
--- a/cmd/pinniped-proxy/Cargo.lock
+++ b/cmd/pinniped-proxy/Cargo.lock
@@ -73,11 +73,32 @@ version = "0.3.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fb1ff21a63d3262af46b9f33a826a8d134e2d0d9b2179c86034948b732ea8b2a"
 dependencies = [
- "bytes",
+ "bytes 0.5.6",
  "flate2",
  "futures-core",
  "memchr",
  "pin-project-lite 0.1.11",
+]
+
+[[package]]
+name = "async-stream"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3670df70cbc01729f901f94c887814b3c68db038aad1329a418bae178bc5295c"
+dependencies = [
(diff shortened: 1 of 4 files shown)


andresmgot:
> so checking logs is required to solve

Why is that? Why not adding the error info, if it exists, to the error message? Unless you mean pinniped itself logs, not the proxy logs.

andresmgot (approved):

andresmgot:
_merging this in case we want to announce the pinniped integration as part of the upcoming release_

absoludity:
> > so checking logs is required to solve
> 
> Why is that? Why not adding the error info, if it exists, to the error message? Unless you mean pinniped itself logs, not the proxy logs.

Correct, the pinniped logs. Pinniped itself (being an authentication mechanism) doesn't actually give the client any useful information other than an "authentication failed" string when the authentication fails - similar to a 401. The CR status either has the credentials or it has no credentials and a string. So we give that same info to the client, but figuring out why the credentials weren't returned requires checking logs of pinniped itself. I've only ever seen a configuration error causing this (such as the one I used to test it, updating the expected audience of the jwt authenticator).

PR #2285, opened 2021-01-28
Update pinniped-proxy runtime to bitnami/minideb. Add to circleci

### LINKED ISSUE #2181: Turn pinniped-proxy POC into production ready service.

absoludity:
### Description of the change

Cleans up the Dockerfile used to build pinniped-proxy, updating to use bitnami/minideb for the runtime base.

Also adds the testing and building of pinniped-proxy to circleci, but does not depend on it for any other images/tests.

### Benefits

Start building the pinniped-proxy image.

### Possible drawbacks

I'll check with @carrodher about landing this and whether we need to wait for equivalent config in the bitnami pipeline.

### Applicable issues

  - fixes #2181

### Additional information

@andresmgot You mentioned recently that the build image might be an issue since it's not part of the bitnami catalog, but I'm not sure how that's different to the golang images which use a dockerhub image for building - ah, unless during the bitnami release it's switched for a specific bitnami go image in that case?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,11 +29,15 @@ workflows:
           <<: *build_always
       - test_dashboard:
           <<: *build_always
+      - test_pinniped_proxy:
+          <<: *build_always
       - test_chart_render:
           <<: *build_always
       - build_go_images:
           <<: *build_always
       - build_dashboard:
+          <<: *build_always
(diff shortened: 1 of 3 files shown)


absoludity:
Closing as I'd pushed to my local master for this PR. New PR at #2286

PR #2286, opened 2021-01-28
Update pinniped-proxy runtime to bitnami/minideb. Add to circleci

### LINKED ISSUE #2181: Turn pinniped-proxy POC into production ready service.

absoludity:
### Description of the change

Cleans up the Dockerfile used to build pinniped-proxy, updating to use bitnami/minideb for the runtime base.

Also adds the testing and building of pinniped-proxy to circleci, but does not depend on it for any other images/tests.

### Benefits

Start building the pinniped-proxy image.

### Possible drawbacks

I'll check with @carrodher about landing this and whether we need to wait for equivalent config in the bitnami pipeline.

### Applicable issues

  - fixes #2181

### Additional information

@andresmgot You mentioned recently that the build image might be an issue since it's not part of the bitnami catalog, but I'm not sure how that's different to the golang images which use a dockerhub image for building - ah, unless during the bitnami release it's switched for a specific bitnami go image in that case?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -306,7 +306,8 @@ jobs:
     <<: *build_images
   build_pinniped_proxy:
     docker:
-      - image: circleci/rust
+      # We're building the image in a docker container anyway so just re-use the golang image already in use.
+      - image: circleci/golang:1.13
     environment:
       IMAGES: "pinniped-proxy"
     <<: *build_images
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -307,7 +307,6 @@ jobs:
   build_pinniped_proxy:
     docker:
       - image: circleci/rust
-    working_directory: /go/src/github.com/kubeapps/kubeapps
     environment:
       IMAGES: "pinniped-proxy"
     <<: *build_images
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,11 +29,15 @@ workflows:
           <<: *build_always
       - test_dashboard:
           <<: *build_always
+      - test_pinniped_proxy:
+          <<: *build_always
       - test_chart_render:
           <<: *build_always
       - build_go_images:
           <<: *build_always
       - build_dashboard:
+          <<: *build_always
+      - build_pinniped_proxy:
           <<: *build_always
       - local_e2e_tests:
           <<: *build_always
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -294,6 +304,12 @@ jobs:
     environment:
       IMAGES: "dashboard"
     <<: *build_images
+  build_pinniped_proxy:
+    docker:
+      - image: circleci/rust
```
andresmgot, line 309: Honestly this image does nothing since we are then building in a docker container, I am not sure what would be faster, if reusing the `golang` image or doing this.

cmd/pinniped-proxy/Dockerfile
```diff
@@ -12,12 +13,7 @@ ADD ./cmd/pinniped-proxy ./
...
 RUN cargo build --release
 
-# Put it all together on a scratch image.
-# TODO: Turns out we'll need to include libssl on the scratch image, perhaps compiling libssl or copying dyn libs listed by ldd.
-# FROM scratch
-# COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
-# FROM bitnami/minideb:buster-snapshot-20201121T213529Z
-FROM debian:10.6-slim
+FROM bitnami/minideb:buster-snapshot-20210126T024633Z
 RUN apt-get update && apt-get install -y ca-certificates libssl1.1 && rm -rf /var/lib/apt/lists/*
```
andresmgot, line 17: Not sure if you replied this before but why not using `scratch` and copy the `ca-certificates`? That's what we do with other services.

andresmgot (approved):
LGTM but I have a question regarding the base image.

andresmgot:
> ah, unless during the bitnami release it's switched for a specific bitnami go image in that case?

Yep, AFAIK they use their own base images for building and compiling all the images.

.circleci/config.yml
absoludity: Ah - ok. I'll switch it to the golang one then.

cmd/pinniped-proxy/Dockerfile
absoludity: It's not just the ca-certificates. The proxy dynamically links to libssl. There are ways to compile with static linking which I tried, but it becomes quite complicated (or uses a rando base image for the build), so I started simple with a minimal base image with libssl available.

andresmgot: got it, it would be ideal to statically link it

PR #2287, opened 2021-01-28
Create pinnipedProxyURL based on existing params.

absoludity:
### Description of the change

Without this change the pinniped-proxy-url used by kubeops was relying on a default value (kubeapps namespace).

This updates the location for pinniped-proxy in both the kubeops and frontend deployments, based on existing values.

### Benefits

Pinniped-proxy will work when using a namespace other than kubeapps for the kubeapps deployment.

### Possible drawbacks

None.

### Applicable issues

  - Ref #2181


andresmgot (approved):

andresmgot:
_merging this in case we want to announce the pinniped integration as part of the upcoming release_

PR #2295, opened 2021-01-29
Several fixes to the pull process

andresmgot:
### Description of the change

After testing the code related to OCI registries IRL, I noticed the three issues fixed here (the first one is the bigger):

 - An OCI endpoint can contain any kind of artifact (docker images or helm charts). We need to filter only the Helm charts, if not it will fail trying to pull a docker image as a chart. For simplicity, I have moved the methods related to the OCI API to an interface (`ociAPICli`).
 - If a chart contained several READMEs, the last one read was the one stored (e.g. `apache/vhosts/README.md`). Fixed to only take into account files in the root directory.
 - For "large" text files (e.g. Jenkins README), `tarReader.Read` lost a bunch of bytes at the end, filling those with "0"s (TBH, not sure why). Using `ioutil.ReadAll` fixes the issue.

### Applicable issues

  - ref #2232

### DIFF
--- a/cmd/asset-syncer/utils.go
+++ b/cmd/asset-syncer/utils.go
@@ -37,6 +37,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/containerd/containerd/remotes/docker"
 	"github.com/disintegration/imaging"
 	"github.com/ghodss/yaml"
 	"github.com/jinzhu/copier"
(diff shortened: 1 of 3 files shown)


cmd/asset-syncer/utils.go
```diff
@@ -325,6 +391,10 @@ func extractFilesFromBuffer(buf *bytes.Buffer) (*artifactFiles, error) {
 		}
 
 		compressedFileName := header.Name
+		if len(strings.Split(compressedFileName, "/")) > 2 {
```
absoludity, line 394: Might be more straight forward to use `filepath.Dir(compressedFileName) != "/"` or similar?

absoludity (approved):

cmd/asset-syncer/utils.go
andresmgot: for doing so, we need the name of the app, it would be something like `filepath.Dir(compressedFileName) != appName` but I don't have that name in this function (and I don't think it's worth it to add it just for this).

absoludity: Right, OK - I misread and thought the path was just `/Readme.md`. You could still use https://golang.org/pkg/path/filepath/#SplitList (which is effectively the same as your code, just using the filepath library which is OS independent for slashes etc., but not so relevant here, I agree). All good :)

PR #2296, opened 2021-01-29
Add authentication information when pulling OCI charts

andresmgot:
### Description of the change

Following up on #2295. This time adding authentication. It's just needed to set the authorization header in the "resolver".

### Applicable issues

  - ref #2232

### DIFF
--- a/cmd/asset-syncer/utils.go
+++ b/cmd/asset-syncer/utils.go
@@ -37,6 +37,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/containerd/containerd/remotes/docker"
 	"github.com/disintegration/imaging"
 	"github.com/ghodss/yaml"
 	"github.com/jinzhu/copier"
@@ -254,6 +255,15 @@ func doReq(url string, headers map[string]string) ([]byte, error) {
 	}
 
 	if res.StatusCode != http.StatusOK {
+		if err == nil {
+			errC, err := ioutil.ReadAll(res.Body)
+			if err != nil {
+				return nil, fmt.Errorf("failed to read error content: %v", err)
+			}
+			if len(errC) > 0 {
+				return nil, fmt.Errorf("request failed: %v", string(errC))
+			}
+		}
 		return nil, fmt.Errorf("request failed: %v", err)
 	}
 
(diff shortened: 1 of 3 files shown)


cmd/asset-syncer/utils.go
```diff
@@ -254,6 +255,15 @@ func doReq(url string, headers map[string]string) ([]byte, error) {
 	}
 
 	if res.StatusCode != http.StatusOK {
+		if err == nil {
```
andresmgot, line 258: This is unrelated but I noticed that when getting an error, the message was in the body of the response so I am extracting it. I will add a unit test for this.

absoludity: I'm not sure what this condition here is doing given that 5 lines above we're already returning if the error is non-nil (ie. I'd expect that it is always nil by line 258 and so this is superfluous - perhaps left over from an earlier edit?)

cmd/asset-syncer/utils_test.go
```diff
@@ -192,6 +193,19 @@ func Test_syncURLInvalidity(t *testing.T) {
 	}
 }
 
+func Test_getOCIRepo(t *testing.T) {
+	t.Run("it should add the auth header to the resolver", func(t *testing.T) {
+		repo, _ := getOCIRepo("namespace", "test", "https://test", "Basic auth", []string{})
+		// The header property is private so we need to use reflect to get its value
+		resolver := repo.(*OCIRegistry).puller.(*ociPuller).resolver
+		resolverValue := reflect.ValueOf(resolver)
+		headerValue := reflect.Indirect(resolverValue).FieldByName("header")
```
absoludity, line 202: yikes, there must be a better way here?

absoludity: It's not clear to my why you can't access `repo.AuthorizationHeader`, since it's a promoted field from the embedded `RepoInternal` struct? For example, this works: https://play.golang.org/p/6jdAMxnRqHZ

andresmgot: I can access that field but it would not test the code added in this PR. The goal is to test that the `resolver` has that header.

absoludity: Yeah, OK. There are other ways around it, but probably no prettier. What I've done in similar situations is to pull out the one line which creates the resolver into its own function, say `createResolver(header string)` or similar, which will be the one piece of untested code, then pass the function as an arg to `getOCIRepo`. In tests you can then pass a fake of the same function and just ensure it's called with the header. But yes, it's also not ideal and leaves the one line untested. +1 with what you think best.

absoludity (approved):

cmd/asset-syncer/utils_test.go
andresmgot: Yes, also, the `resolver` is already an interface and I have another interface for the `puller`. It's already getting difficult to read with those interfaces, better to skip that one for that line.

PR #2297, opened 2021-01-30
enabled TTL lifetime mechanism for finished Jobs

fbalicchia:
Hi, 
With this issue we enabled TTL lifetime for sync jobs using this
[clean-up-finished-jobs-automatically](https://kubernetes.io/docs/concepts/workloads/controllers/job/#clean-up-finished-jobs-automatically) 

We add `TTLSecondsAfterFinished` parameter and if initialized default is 1h

In this manner namespace where KubeApps is installed is not filled of old jobs and sysadmin is no constricted to make a cleanup periodically

In according with documentation to enabled this features is necessary set `true` to `TTLAfterFinished` featureGates, IMHO is necessary to add this information in documentation, Could you suggest where I can insert this informations ? 

In KinD for example to enable featureGates is necessary create cluster with a snipped like this 

```
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
featureGates:
  TTLAfterFinished: true
nodes:
- role: control-plane
  image: kindest/node:v1.18.8
  extraPortMappings:
  - containerPort: 31080
    hostPort: 80
  - containerPort: 31443
    hostPort: 443
```


fbalicchia:
I don't really understand why  e2e fail cause. Making the same test on master fork it seems to works 
[fork_master](https://circleci.com/gh/fbalicchia/kubeapps/tree/master). Thanks for help

cmd/apprepository-controller/controller.go
```diff
@@ -69,6 +70,8 @@ const (
 	// MessageResourceSynced is the message used for an Event fired when an
 	// AppRepsitory is synced successfully
 	MessageResourceSynced = "AppRepository synced successfully"
+	//DefaultLifeTimeTTL max sync lifetime job //https://kubernetes.io/docs/concepts/workloads/controllers/job/#clean-up-finished-jobs-automatically
+	DefaultLifeTimeTTL = 3600
```
absoludity, line 74: I'm not sure why we need this constant here, since it already defaults to "3600" from the command-line arg default that you've set? If you agree, perhaps just delete this const and move the helpful comment to the main.go?

cmd/apprepository-controller/controller.go
```diff
@@ -564,6 +569,24 @@ func jobLabels(apprepo *apprepov1alpha1.AppRepository) map[string]string {
 	}
 }
 
+// ttlLifetimeJobs return time to live set by user if is nill return default
+func ttlLifetimeJobs(config Config) *int32 {
+	var result int32
+
+	if config.TTLSecondsAfterFinished == "" {
```
absoludity, line 576: Given the default command-line arg, a value of "" implies it's been explicitly overridden as "", in which case, I think we should not set the field? (as per the docs, if the field is not set, it won't be handled by the TTL controller).

To do this, you'd need to modify you earlier code only set the `TTLSecondsAfterFinished` field if `config.TTLSecondsAfterFinished != ""`. Thoughts?

absoludity (commented):
Looking good, thanks! A couple of thoughts below.

absoludity:
> In according with documentation to enabled this features is necessary set `true` to `TTLAfterFinished` featureGates, IMHO is necessary to add this information in documentation, Could you suggest where I can insert this informations ?

Hmm, I'd not realised this feature was still alpha (I initially just saw it was added in 1.12 and so thought it must be available). If the field can be added without enabling the feature (and have no effect, but not cause resource validation issues), then that's fine. (Edit: it seems that's the case, given CI now passes).


> I don't really understand why e2e fail cause, Making the same test on master fork it seems to works
> [fork_master](https://circleci.com/gh/fbalicchia/kubeapps/tree/master).

I re-ran it and it passed :) The errors on the original look like there was issues with the instantiated cluster itself.

cmd/apprepository-controller/controller.go
fbalicchia: You are right absolutely

absoludity (approved):
Great, thanks!

Issue #2240, opened 2021-01-12
Remove Helm exclusive features from the chart


andresmgot:
### Description:

Some users want to deploy Kubeapps in their clusters but they don't have access to Helm (e.g. in a gitops workflow). In order to achieve that, they generate the chart YAML (`helm template ...`) and store the resulting manifests to apply them later. For that approach to work out of the box, it's necessary to adapt the pieces of the chart that rely in Helm-specific functionality. This is, the use of helm hooks (`pre-install`, `post-upgrade`, etc) and the use of functions like `.Capabilities`.

This is the list of templates that require some kind of adaptation:
 - [x] 1/ chart/kubeapps/templates/apprepositories.yaml
   - Uses the `post-install` hook.
   - If the hook is removed and it's installed at the same time than the CRD (AppRepository) it returns an error. `Error from server (NotFound): error when creating "kubeapps.yaml": the server could not find the requested resource (post apprepositories.kubeapps.com)`
   - Need to validate if the linked secrets are necessary (for private repositories).
   - There are three alternatives: 1. The CRD is installed before installing the rest of the manifest. 2. The work of populating the default apprepositories is moved to a job. 3. Remove the initialRepositories.
  
 - [x] 2/ chart/kubeapps/templates/apprepository-jobs-cleanup.yaml
   - Uses the `post-delete` hook.
   - Cleanup action to remove apprepositories installed in the `kubeapps` namespace. It is not mandatory, resources will be left behind but that's fine. Actually it's only removing global repositories so currently namespaced repositories are left behind. I would delete this job anyway.

 - [x] 3/ chart/kubeapps/templates/apprepositories-secret.yaml
   - Uses the `pre-install` hook
   - Generate the secrets needed for the app-repositories installed at 1.
   - We need to verify if the installation fails to create the apprepo if the secret doesn't exists. If not it can just be moved out of the hook.

 - [x] 4/ chart/kubeapps/templates/db-secret-jobs-cleanup.yaml
   - hook: post-delete
   - removes the db secret (it  should not be needed if the secret is not installed in a hook)

 - [x] 5/ chart/kubeapps/templates/db-secret-bootstrap.yaml
   - hook: pre-install
   - Creates the secret for the database. The database may fail to be installed if the secret doesn't exist already. Needs to verify if there is an alternative for this.

 - [x] 6/ chart/kubeapps/templates/apprepository-jobs-postupgrade.yaml
   - hook: post-upgrade
[...]

PR #2305, opened 2021-02-03
Remove the helm2 crd resource now in crds directory.

absoludity:
### Description of the change

While testing our chart with kapp I noticed that we still have the helm2 CRD
resource definition with its custom condition. Helm 3 uses the crd in the
chart's crds directory and installs that, by which time this resource is never
rendered anyway as the condition is always false.

### Benefits

No longer see warning about the invalid (helm2) hook.

### Possible drawbacks

None.


andresmgot (approved):
:+1:  This is 7/ in https://github.com/kubeapps/kubeapps/issues/2240

absoludity:
> +1 This is 7/ in #2240

Nice, ticked it off there :)

Issue #2237, opened 2021-01-12
Decide how to perform a periodic review of the project's dependencies


antgamdia:
As already pointed out in https://github.com/kubeapps/kubeapps/issues/2228 and https://github.com/kubeapps/kubeapps/pull/2224, we should perform a periodic analysis of Kubeapps' dependencies in order to be aware of any vulnerability or deprecation.

Currently, the process of how to update deps is unclear (besides some attempts to update it, e.g., https://github.com/kubeapps/kubeapps/pull/2231).
I personally think the process is twofold:

- [x] Add an automatic dependency scanner. There exist solutions, including but not limited to, such as [Snyk](https://snyk.io) or [Trivy](https://github.com/aquasecurity/trivy#circleci) that perform this task. 
- [x] Add an automatic bot to bump up the versions.
- [ ] Update the docs with a consistent process that includes: 
  1. when we should update the deps (each version increase? when a vulnerability with severity>=medium appears?, each time a new version is published? Note that dependabot [can be configured](https://docs.github.com/en/github/administering-a-repository/enabling-and-disabling-version-updates#enabling-github-dependabot-version-updates) for bumping up versions periodically.  (example here: https://github.com/antgamdia/kubeapps/blob/master/.github/dependabot.yml)
  2. which deps should be updated (e.g., in releases, bump up minor/patch, weekly check security alerts and merge PRs)
  3. how to proceed when the version bump implies changes in the code
  4. in which cases we should ignore a security warning (e.g., a vulnerable dep, with no current updates but only used in build-time)

This issue is just for us to add our comments and to reach an agreement on what to do regarding dependencies.

antgamdia:
As suggested in offline discussions, here is a very initial evaluation of both Snyk and Trivy tools:

https://via.vmw.com/2021-kubeapps-deps-tool

If you prefer to move there the discussion on how we plan to update deps (in general, not regarding security updates) , I'm also ok with that.

antgamdia:
Regarding the idea of how to keep our deps up to date, I'd start by setting up Dependabot properly. Now it is only alerting about (some) security updates, but it is much more powerful. It is integrated within Github and it will just send PRs. WDYT?
https://docs.github.com/en/github/administering-a-repository/enabling-and-disabling-version-updates

andresmgot:
Sounds good. Regarding the security report, I am okay using snyk.

antgamdia:
Snyk added! Let's see how it behaves.
[...]

PR #2309, opened 2021-02-03
Add dependabot config file

antgamdia:
### Description of the change

This PR is to simply add the config file for Dependabot. This is just a proposal, feel free to customize any parameter.

Dependabot will be watching 1) dashboard npm deps; 2) integration npm deps; 3) go deps; 4) pinniped proxy cargo deps
For each category, a maximum of 2 concurrent PR is set (so we don't have tons of PR to review) with a daily frequency.

This process can also be set to `weekly` during a certain day/time. I'm ok if you rather go for this periodicity.

### Benefits

Our deps will be finally up to date  :)

### Possible drawbacks

We'll have more work to review the generated PRs, but it's a good practice to keep ours deps up to date as much as possible.

### Applicable issues

  - related  #2237

### Additional information

After merging, perhaps we should disable the snyk feature to do the same (but it's just click a toggle button, let's see how it's work for a couple of days).
Also, I'll try to update these docs to reference we're using this automatic system: https://github.com/kubeapps/kubeapps/blob/master/docs/developer/release-process.md

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,12 +9,6 @@ updates:
       interval: "daily"
     open-pull-requests-limit: 2
 
-  - package-ecosystem: "npm"
-    directory: "/integration"
-    schedule:
-      interval: "daily"
-    open-pull-requests-limit: 2
-
   - package-ecosystem: "gomod"
     directory: "/"
     schedule:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,28 @@
...
+    directory: "/dashboard"
+    schedule:
+      interval: "daily"
+    open-pull-requests-limit: 2
+
+  - package-ecosystem: "npm"
+    directory: "/integration"
+    schedule:
+      interval: "daily"
+    open-pull-requests-limit: 2
```
andresmgot, line 16: I would not add here (nor in snyk) the integration package.json, that's only used for testing, it's not part of the application.

PR #2310, opened 2021-02-03
ChartClient internal refactor

andresmgot:
### Description of the change

The goal of this PR is to prepare the code to be able to implement the methods required to pull a chart from an OCI registry. The two major changes (and the reason why this PR is a bit bulky) are:

 - I have split the interface `ChartClient` between the methods that depends on the repository from the ones that don't (so the implementation for the OCI case doesn't repeat the same methods).
 - At the moment, the Client is instantiated at the very beginning, before the application repository is being fetched. The problem is that we need the application repository info _before_ the client is instantiated (to differ between a regular repository or an OCI one). Because of that, I had to extract some of the logic previously present in the Client to other functions (and some interfaces required for testing purposes).

I will add more comments inline but note that there is no functional change here.

### Applicable issues

  - ref #2232


pkg/agent/agent_test.go
```diff
@@ -153,7 +153,7 @@ func TestCreateReleases(t *testing.T) {
 			namespace: "default",
 			version:   1,
 			existingReleases: []releaseStub{
-				releaseStub{"otherchart", "default", 1, "1.0.0", release.StatusDeployed},
+				{"otherchart", "default", 1, "1.0.0", release.StatusDeployed},
```
andresmgot, line 156: these are just lint warnings

pkg/kube/kube_handler.go
```diff
@@ -208,13 +208,9 @@ type ValidationResponse struct {
...
-// This interface is explicitly private so that it cannot be used in function
-// args, so that call-sites cannot accidentally pass a service handler in place
-// of a user handler.
-// TODO(mnelson): We could instead just create a UserHandler interface which embeds
-// this one and adds one method, to force call-sites to explicitly use a UserHandler
-// or ServiceHandler.
-type handler interface {
+// AuthedHandler exposes Kubernetes resources as either a given user or as a service
+// Note: This should be retrieved as the result of AuthHandler
+type AuthedHandler interface {
```
andresmgot, line 213: I needed to expose this but I am not sure I understood your comment here: https://github.com/kubeapps/kubeapps/pull/1681#discussion_r412652252

absoludity (commented):
Thanks Andres, some thoughts:

> 
> The goal of this PR is to prepare the code to be able to implement the methods required to pull a chart from an OCI registry. The two major changes (and the reason why this PR is a bit bulky) are:
> 
>     * I have split the interface `ChartClient` between the methods that depends on the repository from the ones that don't (so the implementation for the OCI case doesn't repeat the same methods).

It looks like you've removed part of the interface and put the code inline, rather than splitting the interface? I don't think you need to split the interface at all if the goal is just to avoid repeating the same methods in the implementation of the OCI case. For example, I think you could use [struct embedding](https://golang.org/doc/effective_go.html#embedding) so that the OCI implementation struct can embed the non-OCI implementation and just define the functions that differ? Or you could define a struct with the implementations of certain functions required by the interface which are the same, then embed that in the OCI and non-OCI implementations so that both satisfy the interface?

That would encapsulate the functionality without having to inline it, I think? (it may also remove the need to make `handler` public, not sure).

> 
>     * At the moment, the Client is instantiated at the very beginning, before the application repository is being fetched. The problem is that we need the application repository info _before_ the client is instantiated (to differ between a regular repository or an OCI one). Because of that, I had to extract some of the logic previously present in the Client to other functions (and some interfaces required for testing purposes).

I probably don't have enough context yet (sorry, I can look again in the morning with fresh eyes), but it's not clear to me why you can't still instantiate the client at the very beginning by updating `chartutils.NewChartClient()` so that it fetches the repo and returns the correct client? (it'd obviously need a signature update to include the possibility of an error)?

> 
> 
> I will add more comments inline but note that there is no functional change here.

Yeah, I get that there's no functional change, but I worry it leaves the code in a slightly worse state (longer handler functions where things are inlined and therefore more confusing to follow) than it was (and it was already a little convoluted, not excluding my own contributions there). I just want you to be sure that there's not a better way to be able to (1) provide two separate client implementations for the original simpler interface, and (2) avoid the security issue of call-sites not needing to explicitly identify whether they are using a user client or service client.



> ### Applicable issues
> 
>     * ref #2232

pkg/kube/kube_handler.go
andresmgot: Yes, the thing is that I now I am in that position, I want the user of `AuthedHandler` to call it regardless if it's a `UserHandler` or a `ServiceHandler`:

https://github.com/kubeapps/kubeapps/pull/2310/files#diff-7882fe4bc8aaae69a41f78e120b9b973ddec3b087e8eacec396457cb06675f29R274

Note that the decision is made there and the rest is used whatever that resolves. In any case, it's true that it can be risky and misused. Will revert this and try to find a better solution (but it may require to duplicate code).

andresmgot:
> It looks like you've removed part of the interface and put the code inline, rather than splitting the interface?

Not really, let me elaborate on this. The interface that `ChartClient` implemented had four methods: `ParseDetails`, `GetChart`, `InitNetClient`, `RegistrySecretsPerDomain`. `GetChart` and `InitNetClient` (now `InitClient`) are still in the interface so no changes there. `ParseDetails` had nothing to do with the information within a `ChartClient` so that's a standalone function now and `RegistrySecretsPerDomain` was in the interface just because those secrets were retrieved in `InitNetClient` and stored in the struct but that was weird because it was not necessary for `InitNetClient`, it was done there just because the required info to retrieve the secrets (`AppRepo` spec) was request at that step. Now the method to retrieve the `AppRepo` has been moved to its own function so we don't need to store nor retrieve unnecessary info in the wrong "hook".

> I probably don't have enough context yet (sorry, I can look again in the morning with fresh eyes), but it's not clear to me why you can't still instantiate the client at the very beginning by updating `chartutils.NewChartClient()` so that it fetches the repo and returns the correct client? (it'd obviously need a signature update to include the possibility of an error)?

`NewChartClient` is returning an instance of `ChartClient` (that implements `InitClient` and `GetChart`). All that is dependent of a Helm Repository. What I will need to call is `NewOCIClient` that also implements those methods for the OCI case. For knowing which client to instantiate, it's necessary to retrieve the `AppRepo` first (to read `AppRepo.Spec.Type`). And for retrieving the `AppRepo` we need to `ParseDetails` from the request before.

> but I worry it leaves the code in a slightly worse state (longer handler functions where things are inlined and therefore more confusing to follow) than it was (and it was already a little convoluted, not excluding my own contributions there)

I know this is (was?) some spaghetti code, hopefully now is less entangled, at least I see it that way after working on this for a couple of days. The handler functions are now longer because they are explicitly retrieving the needed instances: chart details, kubernetes client, appRepository and the actual chart. Before the same information was "randomly" retrieved but hidden under the carpet in the different hooks and stored in private struct fields. On top of that, it worked only if these hooks were called in the right order. It still like that a little bit (you need to first call `InitClient` before `GetChart` for it to work) but at least it's reduced and required to be able to call `GetChart` with a mocked client.

pkg/kube/kube_handler.go
andresmgot: Adding a bit more of color here. There are two things that will be retrieved with the handler: an apprepo and the secrets related to that repo (e.g. CA file or basic auth). For that we can use either the svc handler (for global apprepos) or the user handler (for namespaced apprepos). We would need to duplicate the logic that selects a client depending on the repo namespace (unless I find a better solution but I cannot think of any).

cmd/kubeops/internal/handler/handler.go
```diff
@@ -164,7 +172,24 @@ func ListAllReleases(cfg Config, w http.ResponseWriter, req *http.Request, _ han
 
 // CreateRelease creates a release.
 func CreateRelease(cfg Config, w http.ResponseWriter, req *http.Request, params handlerutil.Params) {
-	chartDetails, chartMulti, err := handlerutil.ParseAndGetChart(req, cfg.ChartClient, isV1SupportRequired)
+	chartDetails, err := handlerutil.ParseRequest(req)
+	if err != nil {
+		returnErrMessage(err, w)
+		return
+	}
+	appRepo, caCertSecret, authSecret, err := chart.GetAppRepoAndRelatedSecrets(chartDetails.AppRepositoryResourceName, chartDetails.AppRepositoryResourceNamespace, cfg.KubeHandler, cfg.Token, cfg.Cluster, cfg.Options.KubeappsNamespace)
```
absoludity, line 180: Thanks for DRYing up into `GetAppRepoAndRelatedSecrets`.

pkg/chart/chart.go
```diff
@@ -364,11 +346,19 @@ func (c *Client) GetChart(details *Details, netClient kube.HTTPClient, requireV1
 
 // RegistrySecretsPerDomain checks the app repo and available secrets
 // to return the secret names per registry domain.
-//
-// These are actually calculated during InitNetClient when we already have a
-// k8s client with the user token.
-func (c *Client) RegistrySecretsPerDomain() map[string]string {
-	return c.registrySecretsPerDomain
+func RegistrySecretsPerDomain(appRepo *appRepov1.AppRepository, cluster string, userAuthToken string, authHandler kube.AuthHandler) (map[string]string, error) {
+	registrySecretsPerDomain, err := getRegistrySecretsPerDomain(
```
absoludity, line 350: I don't think there's any need for the indirection now, you could just have the implementation of (the private) `getRegistrySecretsPerDomain` here now, I think? (what's here is essentially now `return getRegistrySecretsPerDomain(...)`)

pkg/chart/fake/chart.go
```diff
@@ -17,25 +17,19 @@ limitations under the License.
...
 )
 
-type FakeChart struct{}
-
-func (f *FakeChart) ParseDetails(data []byte) (*chartUtils.Details, error) {
-	details := &chartUtils.Details{}
-	err := json.Unmarshal(data, details)
-	return details, err
-}
+// Chart implements Resolver inteface
```
absoludity, line 28: Note, a while back we noted that `Chart` was confusing for the implementation and had renamed in the non-fake module to `Client` (since what was called `Chart` was really a chart client). We should probably rename it here too so that...

pkg/chart/fake/chart.go
```diff
@@ -17,25 +17,19 @@ limitations under the License.
...
+type Chart struct{}
 
-func (f *FakeChart) GetChart(details *chartUtils.Details, netClient kube.HTTPClient, requireV1Support bool) (*chartUtils.ChartMultiVersion, error) {
+// GetChart fake
+func (f *Chart) GetChart(details *chartUtils.Details, repoURL string, requireV1Support bool) (*chartUtils.ChartMultiVersion, error) {
```
absoludity, line 32: This won't be odd (`Chart.GetChart` but instead `ChartClient.GetChart`)

absoludity (approved):
Thanks for expanding on the reasoning and for finding a way to keep the handler explicit.

cmd/kubeops/internal/handler/handler.go
```diff
@@ -104,7 +107,10 @@ func WithHandlerConfig(storageForDriver agent.StorageForDriver, options Options)
 			cfg := Config{
 				Options:      options,
 				ActionConfig: actionConfig,
-				ChartClient:  chartUtils.NewChartClient(kubeHandler, options.ClustersConfig.KubeappsClusterName, options.KubeappsNamespace, options.UserAgent),
```
absoludity, line 107: We did not catch in this review that the chart client being removed here was for the `options.ClustersConfig.KubeappsClusterName` cluster, whereas below when calling `chart.GetAppRepoAndRelatedSecrets` we used `config.Cluster` instead. Fixed in #2377

PR #2311, opened 2021-02-03
Remove code related to charts v2

andresmgot:
### Description of the change

Follow up of #2310

I noticed that there was a bunch of code related to loading v2 charts but later on, only the v3 chart version was being taken into account so I cleaned it up.

### Applicable issues

  - ref #2232


absoludity (approved):
Woohoo to removing temporary code that was only meant to support us during the transition from 2->3 :)

PR #2334, opened 2021-02-05
Enable validation for OCI appRepos

andresmgot:
### Description of the change

Adding the endpoint to enable validation for OCI registries. Rather than just doing a request to validate the registry, we are doing a request per repository (app) to validate that there are no typos there (we could not validate just the registry anyway, because the root path of the API `https://repo/v2` is protected, even for public repos, at least for Harbor).

### Applicable issues

  - ref #2232

### DIFF
--- a/pkg/kube/kube_handler.go
+++ b/pkg/kube/kube_handler.go
@@ -23,6 +23,8 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"net/url"
+	"path"
 	"strings"
 
 	"github.com/kubeapps/kubeapps/cmd/apprepository-controller/pkg/apis/apprepository/v1alpha1"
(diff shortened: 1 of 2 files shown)


pkg/kube/kube_handler.go
```diff
@@ -586,13 +589,63 @@ func doValidationRequest(cli HTTPClient, req *http.Request) (*ValidationResponse
...
+			parsedURL.RawQuery = q.Encode()
+			req, err := http.NewRequest("GET", parsedURL.String(), nil)
+			if err != nil {
+				return nil, err
+			}
+			result = append(result, req)
+		}
+		break
+	default:
+		req, err := http.NewRequest("GET", repoURL+"/index.yaml", nil)
```
absoludity, line 619: We should probably use `path.Join` everywhere.

pkg/kube/kube_handler.go
```diff
@@ -586,13 +589,63 @@ func doValidationRequest(cli HTTPClient, req *http.Request) (*ValidationResponse
...
+		for _, repoName := range appRepo.Spec.OCIRepositories {
+			parsedURL, err := url.ParseRequestURI(repoURL)
+			if err != nil {
+				return nil, err
+			}
+			parsedURL.Path = path.Join("v2", parsedURL.Path, repoName, "tags", "list")
+			q := parsedURL.Query()
+			q.Add("n", "1")
+			parsedURL.RawQuery = q.Encode()
+			req, err := http.NewRequest("GET", parsedURL.String(), nil)
```
absoludity, line 611: Does `oras` not provide helpers for getting the tag list of a repo? Ah, or is this so both OCI and non-oci repos can be validated by a bunch of plain requests?

absoludity (approved):

pkg/kube/kube_handler.go
andresmgot: nope, `oras` only exposes methods to push and pull artifacts, not to interact with an OCI API (probably there is some package for that, but this is simple enough to just use the API with a request).

PR #2323, opened 2021-02-04
OCI GetChart

andresmgot:
### Description of the change

Follow up of #2311

Now it's time to implement the actual interface for the OCI case. 

### Possible drawbacks

As explained in a `TODO`, using a CA cert won't work here. The only alternative with ORAS is to use an `insecure` flag but I didn't included that (yet?).

### Applicable issues

  - ref #2232

PR #2336, opened 2021-02-05
Allow to create OCI apprepos

andresmgot:
### Description of the change

Enable the usage of OCI registries for AppRepositories: 

Screenshot from 2021-02-08 14-00-11

Note that using a CA cert is not supported as explained here: https://github.com/kubeapps/kubeapps/pull/2323#issuecomment-775029689

### Applicable issues

  - ref #2232

### DIFF
--- a/dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
+++ b/dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
@@ -436,21 +436,25 @@ export function AppRepoForm(props: IAppRepoFormProps) {
               </label>
             </Column>
             <Column span={9}>
-              <div className="column-valing-center clr-control-container">
-                <div hidden={type !== TYPE_OCI}>
-                  <label className="clr-control-label" htmlFor="kubeapps-repo-username">
-                    List of Repositories
-                  </label>
-                  <div className="clr-textarea-wrapper">
-                    <textarea
-                      id="kubeapps-oci-repositories"
-                      rows={4}
-                      className="clr-textarea"
-                      placeholder={"nginx, jenkins"}
-                      value={ociRepositories}
-                      onChange={handleOCIRepositoriesChange}
-                    />
-                  </div>
+              <div
+                className="column-valing-center clr-control-container"
+                hidden={type !== TYPE_OCI}
+              >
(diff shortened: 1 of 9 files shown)


dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
```diff
@@ -372,6 +395,72 @@ export function AppRepoForm(props: IAppRepoFormProps) {
           </Row>
         </div>
       </div>
+      <div className="clr-form-control">
+        <label className="clr-control-label">Repository Type</label>
+        <span className="clr-form-description">
+          Charts can be stored either in a Helm repository (e.g. ChartMuseum) or in a OCI Registry.
+          Select the one that applies.
```
absoludity, line 402: "Select which type of repository you are adding." (though we're confusing two different types of repositories here, another option:
"Select which type of chart catalog you are adding."

absoludity (approved):
LGTM, just two thoughts:

1) I wonder if in the UI we should rename "AppRepository" to "AppCatalog" so that an app catalog can have either a helm repository or an OCI registry? But that's unrelated to this PR.
2) Do you think we should have this UI change behind a feature flag? (or are you happy/ok with the sync times for it to be used and us to start maintaining it as is?)

andresmgot:
> * I wonder if in the UI we should rename "AppRepository" to "AppCatalog" so that an app catalog can have either a helm repository or an OCI registry? But that's unrelated to this PR.

I think that can be more confusing, `AppRepository` is what we have been called all this time and the meaning (from Kubeapps point of view) is the same. Also the CRD, values.yaml, controller.. all those point to app repositories.

> * Do you think we should have this UI change behind a feature flag? (or are you happy/ok with the sync times for it to be used and us to start maintaining it as is?)

I don't think so. Better to get feedback if possible (also, I may able to improve performance before we cut a release).

dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
andresmgot: "Select the chart storage type"?

andresmgot: I am going to merge this but let me know if you don't agree, I can open another PR for this.

PR #2337, opened 2021-02-06
Allow apiServiceURLs to include subpath in NGINX location

mecampbellsoup:
### Description of the change

More context in Slack can be found [here](https://kubernetes.slack.com/archives/C9D3TSUG4/p1612546656047300?thread_ts=1612465715.011500&cid=C9D3TSUG4).

In order to use kubeapps with our API gateway (authenticating proxy), I need to be able to configure an optional k8s API URL _path_ (in addition to the base service URL).

This way I can write an ingress rule that matches that path, and set the base k8s service URL to my ingress service. Voila - kubeapps sits behind our API gateway, like all our other upstream services!

One additional change is needed: In the kubeops NGINX locations (i.e. those that proxy via `proxy_pass {{ include "kubeapps.frontend-config.proxy_pass" . -}};`), I need to be able to set an additional proxy header so as to include an `Authorization: Bearer some-token` that kubeops will [in turn](https://github.com/kubeapps/kubeapps/blob/e49b5ee672018e2293b66af1c1e4be11f32782ee/pkg/auth/authgate.go#L23) use when making its request to the k8s backend.

### Benefits

kubeapps charts will now be compatible with [k8s authenticating proxy](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#authenticating-proxy) auth strategies like ours!

### Possible drawbacks

As these values would be strictly opt-in there should be no drawbacks.

### Applicable issues

Resolves https://github.com/kubeapps/kubeapps/issues/2253.

### DIFF
--- a/chart/kubeapps/templates/_helpers.tpl
+++ b/chart/kubeapps/templates/_helpers.tpl
@@ -6,13 +6,6 @@ Expand the name of the chart.
 {{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
 
-{{- define "kubeapps.parseAPIServiceURL" -}}
-{{- $parsed := urlParse . }}
-{{- $baseURL := (printf "%s://%s" (get $parsed "scheme") (get $parsed "host")) }}
-{{- $path := get $parsed "path" }}
-{{ dict "baseURL" $baseURL "path" $path | toYaml }}
-{{- end -}}
-
 {{/*
 Create a default fully qualified app name.
 We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
--- a/chart/kubeapps/templates/_helpers.tpl
+++ b/chart/kubeapps/templates/_helpers.tpl
@@ -6,6 +6,13 @@ Expand the name of the chart.
 {{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
 
+{{- define "kubeapps.parseAPIServiceURL" -}}
+{{- $parsed := urlParse . }}
+{{- $baseURL := (printf "%s://%s" (get $parsed "scheme") (get $parsed "host")) }}
+{{- $path := get $parsed "path" }}
+{{ dict "baseURL" $baseURL "path" $path | toYaml }}
+{{- end -}}
+
 {{/*
 Create a default fully qualified app name.
 We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
--- a/chart/kubeapps/templates/kubeapps-frontend-config.yaml
+++ b/chart/kubeapps/templates/kubeapps-frontend-config.yaml
@@ -62,9 +62,9 @@ data:
     {{/* We need to split the API service URL(s) into the base url and the path segment so
            those configurations using a path can be appropriately rewritten below while
            ensuring the proxy_pass statement is given the base URL only. */}}
-    {{- $parsed := urlParse (default "" .apiServiceURL) }}
-    {{- $apiServiceBaseURL := ternary "" (printf "%s://%s" (get $parsed "scheme") (get $parsed "host")) (empty .apiServiceURL)}}
-    {{- $apiServiceURLPath := get $parsed "path" }}
+    {{- $parsed := urlParse (default "https://kubernetes.default" .apiServiceURL) }}
+    {{- $apiServiceBaseURL := urlJoin (pick $parsed "scheme" "host") }}
+    {{- $apiServiceURLPath := $parsed.path }}
         rewrite /api/clusters/{{ .name }}/(.*) {{ $apiServiceURLPath }}/$1 break;
         rewrite /api/clusters/{{ .name }} {{ $apiServiceURLPath }}/ break;
 
@@ -85,7 +85,7 @@ data:
         proxy_pass http://kubeapps-internal-pinniped-proxy.{{ $.Release.Namespace }}:{{ $.Values.pinnipedProxy.service.port }};
     {{- else }}
         # Otherwise we route directly through to the clusters with existing credentials.
-        proxy_pass {{ default "https://kubernetes.default" $apiServiceBaseURL }};
+        proxy_pass {{ $apiServiceBaseURL }};
     {{- if .certificateAuthorityData }}
         proxy_ssl_trusted_certificate "./server_blocks/{{ .name }}-ca.pem";
     {{- end }}
--- a/chart/kubeapps/templates/kubeapps-frontend-config.yaml
+++ b/chart/kubeapps/templates/kubeapps-frontend-config.yaml
@@ -103,6 +103,9 @@ data:
         proxy_set_header Authorization "Bearer $http_x_forwarded_access_token";
         {{- end }}
 
+        {{- if .Values.frontend.proxypassExtraSetHeader }}
+        proxy_set_header {{ .Values.frontend.proxypassExtraSetHeader }};
+        {{- end }}
         proxy_pass {{ include "kubeapps.frontend-config.proxy_pass" . -}};
       }
 
@@ -112,6 +115,9 @@ data:
         proxy_read_timeout 10m;
         rewrite /api/kubeops/(.*) /$1 break;
         rewrite /api/kubeops / break;
+        {{- if .Values.frontend.proxypassExtraSetHeader }}
+        proxy_set_header {{ .Values.frontend.proxypassExtraSetHeader }};
+        {{- end }}
         proxy_pass {{ include "kubeapps.frontend-config.proxy_pass" . -}};
 
         {{- if .Values.frontend.proxypassAccessTokenAsBearer }}
@@ -130,6 +136,9 @@ data:
         proxy_set_header Authorization "Bearer $http_x_forwarded_access_token";
         {{- end }}
 
+        {{- if .Values.frontend.proxypassExtraSetHeader }}
+        proxy_set_header {{ .Values.frontend.proxypassExtraSetHeader }};
+        {{- end }}
         proxy_pass {{ include "kubeapps.frontend-config.proxy_pass" . -}};
       }
 
--- a/chart/kubeapps/templates/kubeapps-frontend-config.yaml
+++ b/chart/kubeapps/templates/kubeapps-frontend-config.yaml
@@ -59,8 +59,12 @@ data:
       # the cluster). See clusters option.
       {{- range .Values.clusters }}
       location ~* /api/clusters/{{ .name }} {
-        rewrite /api/clusters/{{ .name }}/(.*) /$1 break;
-        rewrite /api/clusters/{{ .name }} / break;
+        {{- $apiServiceURL := default "https://kubernetes.default" .apiServiceURL }}
+        {{- $parsed := urlParse $apiServiceURL }}
+        {{- $baseURL := (printf "%s://%s" (get $parsed "scheme") (get $parsed "host")) }}
+        {{- $path := get $parsed "path" }}
+        rewrite /api/clusters/{{ .name }}/(.*) {{ $path }}/$1 break;
+        rewrite /api/clusters/{{ .name }} {{ $path }}/ break;
 
       {{/* Helm returns a nil pointer error when accessing foo.bar if foo doesn't
            exist, even with the `default` function.
@@ -79,7 +83,7 @@ data:
         proxy_pass http://kubeapps-internal-pinniped-proxy.{{ $.Release.Namespace }}:{{ $.Values.pinnipedProxy.service.port }};
       {{- else }}
         # Otherwise we route directly through to the clusters with existing credentials.
-        proxy_pass {{ default "https://kubernetes.default" .apiServiceURL }};
+        proxy_pass {{ $baseURL }};
         {{- if .certificateAuthorityData }}
         proxy_ssl_trusted_certificate "./server_blocks/{{ .name }}-ca.pem";
         {{- end }}
--- a/chart/kubeapps/templates/kubeapps-frontend-config.yaml
+++ b/chart/kubeapps/templates/kubeapps-frontend-config.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: {{ template "kubeapps.frontend-config.fullname" . }}
-  labels:{{ include "kubeapps.extraAppLabels" . | nindent 4 }}
+  labels: {{ include "kubeapps.extraAppLabels" . | nindent 4 }}
     app: {{ template "kubeapps.frontend-config.fullname" . }}
 data:
   k8s-api-proxy.conf: |-
@@ -22,10 +22,10 @@ data:
     # Google Kubernetes Engine requires the access_token as the Bearer when talking to the k8s api server.
     proxy_set_header Authorization "Bearer $http_x_forwarded_access_token";
     {{- end }}
-  {{- range .Values.clusters }}
+    {{- range .Values.clusters }}
   {{- if .certificateAuthorityData }}
   {{ .name }}-ca.pem: {{ .certificateAuthorityData }}
-   {{- end }}
+  {{- end }}
   {{- end}}
   vhost.conf: |-
     # Retain the default nginx handling of requests without a "Connection" header
@@ -40,12 +40,12 @@ data:
 
     server {
       listen 8080;
-  {{- if .Values.frontend.largeClientHeaderBuffers }}
+    {{- if .Values.frontend.largeClientHeaderBuffers }}
       large_client_header_buffers {{ .Values.frontend.largeClientHeaderBuffers }};
-  {{- end }}
-  {{- if .Values.enableIPv6 }}
+    {{- end }}
+    {{- if .Values.enableIPv6 }}
       listen [::]:8080;
-  {{- end}}
+    {{- end}}
       server_name _;
 
       location /healthz {
@@ -57,40 +57,40 @@ data:
       # Ensure each cluster can be reached (should only be
       # used with an auth-proxy where k8s credentials never leave
       # the cluster). See clusters option.
-      {{- range .Values.clusters }}
+    {{- range .Values.clusters }}
       location ~* /api/clusters/{{ .name }} {
-        {{- $apiServiceURL := default "https://kubernetes.default" .apiServiceURL }}
-        {{- $parsed := urlParse $apiServiceURL }}
-        {{- $baseURL := (printf "%s://%s" (get $parsed "scheme") (get $parsed "host")) }}
-        {{- $path := get $parsed "path" }}
+    {{- $apiServiceURL := default "https://kubernetes.default" .apiServiceURL }}
+    {{- $parsed := urlParse $apiServiceURL }}
+    {{- $baseURL := (printf "%s://%s" (get $parsed "scheme") (get $parsed "host")) }}
+    {{- $path := get $parsed "path" }}
         rewrite /api/clusters/{{ .name }}/(.*) {{ $path }}/$1 break;
         rewrite /api/clusters/{{ .name }} {{ $path }}/ break;
 
-      {{/* Helm returns a nil pointer error when accessing foo.bar if foo doesn't
+    {{/* Helm returns a nil pointer error when accessing foo.bar if foo doesn't
            exist, even with the `default` function.
            See https://github.com/helm/helm/issues/8026#issuecomment-756538254 */}}
-      {{- $pinnipedConfig := .pinnipedConfig | default dict }}
-      {{- if and $.Values.pinnipedProxy.enabled $pinnipedConfig.enable }}
+    {{- $pinnipedConfig := .pinnipedConfig | default dict }}
+    {{- if and $.Values.pinnipedProxy.enabled $pinnipedConfig.enable }}
         # If pinniped proxy is enabled *and* the current cluster is configured
         # to exchange credentials then we route via pinnipedProxy to exchange
         # credentials for client certificates.
-        {{- if .apiServiceURL }}
+    {{- if .apiServiceURL }}
         proxy_set_header PINNIPED_PROXY_API_SERVER_URL {{ .apiServiceURL }};
-        {{- end }}
-        {{- if .certificateAuthorityData }}
+    {{- end }}
+    {{- if .certificateAuthorityData }}
         proxy_set_header PINNIPED_PROXY_API_SERVER_CERT {{ .certificateAuthorityData }};
-        {{- end }}
+    {{- end }}
         proxy_pass http://kubeapps-internal-pinniped-proxy.{{ $.Release.Namespace }}:{{ $.Values.pinnipedProxy.service.port }};
-      {{- else }}
+    {{- else }}
         # Otherwise we route directly through to the clusters with existing credentials.
         proxy_pass {{ $baseURL }};
-        {{- if .certificateAuthorityData }}
+    {{- if .certificateAuthorityData }}
         proxy_ssl_trusted_certificate "./server_blocks/{{ .name }}-ca.pem";
-        {{- end }}
-      {{- end }}
+    {{- end }}
+    {{- end }}
         include "./server_blocks/k8s-api-proxy.conf";
       }
-      {{- end }}
+    {{- end }}
 
       # Forward '/api/assetsvc' to '/assetsvc'
       # but preserving the encoding (eg. '%2F' is not converted to '/')
@@ -102,14 +102,14 @@ data:
         rewrite ^ $request_uri; # pass the encoded url downstream as is,
         rewrite /api/assetsvc([^?]*) /assetsvc$1?$args break;
 
-        {{- if .Values.frontend.proxypassAccessTokenAsBearer }}
+    {{- if .Values.frontend.proxypassAccessTokenAsBearer }}
         # Google Kubernetes Engine requires the access_token as the Bearer when talking to the k8s api server.
         proxy_set_header Authorization "Bearer $http_x_forwarded_access_token";
-        {{- end }}
+    {{- end }}
 
-        {{- if .Values.frontend.proxypassExtraSetHeader }}
+    {{- if .Values.frontend.proxypassExtraSetHeader }}
         proxy_set_header {{ .Values.frontend.proxypassExtraSetHeader }};
-        {{- end }}
+    {{- end }}
         proxy_pass {{ include "kubeapps.frontend-config.proxy_pass" . -}};
       }
 
@@ -119,15 +119,15 @@ data:
         proxy_read_timeout 10m;
         rewrite /api/kubeops/(.*) /$1 break;
         rewrite /api/kubeops / break;
-        {{- if .Values.frontend.proxypassExtraSetHeader }}
+    {{- if .Values.frontend.proxypassExtraSetHeader }}
         proxy_set_header {{ .Values.frontend.proxypassExtraSetHeader }};
-        {{- end }}
+    {{- end }}
         proxy_pass {{ include "kubeapps.frontend-config.proxy_pass" . -}};
 
-        {{- if .Values.frontend.proxypassAccessTokenAsBearer }}
+    {{- if .Values.frontend.proxypassAccessTokenAsBearer }}
         # Google Kubernetes Engine requires the access_token as the Bearer when talking to the k8s api server.
         proxy_set_header Authorization "Bearer $http_x_forwarded_access_token";
-        {{- end }}
+    {{- end }}
       }
 
       # The route for the Kubeapps backend API is not prefixed.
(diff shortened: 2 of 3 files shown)


chart/kubeapps/templates/kubeapps-frontend-config.yaml
```diff
@@ -58,10 +58,7 @@ data:
       # used with an auth-proxy where k8s credentials never leave
       # the cluster). See clusters option.
       {{- range .Values.clusters }}
-      location ~* /api/clusters/{{ .name }} {
-        rewrite /api/clusters/{{ .name }}/(.*) /$1 break;
-        rewrite /api/clusters/{{ .name }} / break;
```
mecampbellsoup, line 63: @absoludity @antgamdia unfortunately, this change doesn't seem to work as expected.

```
172.17.0.24 - - [05/Feb/2021:22:59:20 +0000] "GET /api/clusters/default/ HTTP/1.1" 502 552 "http://apps.coreweave.test:31035/" "Mozilla/5.0 (Macintosh; Intel Mac OS X 11_2_0) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/88.0.4324.146 Safari/537.36"
2021/02/05 22:59:20 [error] 27#27: *5 no resolver defined to resolve gatekeeper-kubernetes-ingress.default, client: 172.17.0.24, server: _, request: "GET /api/clusters/default/ HTTP/1.1", host: "apps.coreweave.test:31035", referrer: "http://apps.coreweave.test:31035/"
```
Either of you guys see what might be wrong?

Reading the existing config, it just seems to remove `/api/clusters/{{ .name }}` from the URL before proxying the request to `.apiServiceUrl`.

My thinking was to just place the "extra" group (i.e. what comes after `/api/clusters/default`)  into the location itself so I can then `proxy_pass http://some-host/k8s/$1` and thereby avoid [this NGINX error](https://stackoverflow.com/questions/53353572/proxy-pass-cannot-have-uri-part-in-location-given-by-regular-expression).

chart/kubeapps/templates/kubeapps-frontend-config.yaml
```diff
@@ -79,7 +76,7 @@ data:
         proxy_pass http://kubeapps-internal-pinniped-proxy.{{ $.Release.Namespace }}:{{ $.Values.pinnipedProxy.service.port }};
       {{- else }}
         # Otherwise we route directly through to the clusters with existing credentials.
-        proxy_pass {{ default "https://kubernetes.default" .apiServiceURL }};
+        proxy_pass {{ default "https://kubernetes.default" .apiServiceURL }}/$1;
```
mecampbellsoup, line 79: By putting the "extra" match group into the `proxy_pass` directive I expected to preserve existing behavior of this location and avoid [this error](https://stackoverflow.com/questions/53353572/proxy-pass-cannot-have-uri-part-in-location-given-by-regular-expression).

chart/kubeapps/templates/kubeapps-frontend-config.yaml
```diff
@@ -130,6 +133,9 @@ data:
         proxy_set_header Authorization "Bearer $http_x_forwarded_access_token";
         {{- end }}
 
+        {{- if .Values.frontend.proxypassExtraSetHeader }}
+        proxy_set_header {{ .Values.frontend.proxypassExtraSetHeader }};
```
mecampbellsoup, line 137: I need this so that I can add an Authorization header that will be used by my API gateway when authenticating k8s service requests.

chart/kubeapps/templates/kubeapps-frontend-config.yaml
absoludity: > Reading the existing config, it just seems to remove `/api/clusters/{{ .name }}` from the URL before proxying the request to `.apiServiceUrl`.

That's correct - the kubeapps uri defines the cluster name and we use that to proxy the request to the correct api service url with the remaining path.

> 
> My thinking was to just place the "extra" group (i.e. what comes after `/api/clusters/default`) into the location itself so I can then `proxy_pass http://some-host/k8s/$1` and thereby avoid [this NGINX error](https://stackoverflow.com/questions/53353572/proxy-pass-cannot-have-uri-part-in-location-given-by-regular-expression).

I don't understand what needs changing here, given that I think you want the following (I think?):
```
http://kubeapps.some-host/k8s/api/v1/namespaces -> http://gateway.some-host/k8s/api/v1/namespaces
```

which wouldn't require any location/rewrite changes from the above, just the proxy pass. But there must be something I'm missing. Can you let us know exact examples of what you're trying to achieve right here (sorry if you've already said this and I've lost it in the other threads/PRs).

It looks like what you have here is already a workaround for some other issue, I'd like to see the original thing you tried, know why you're trying it and what the issue is, before looking at possible workarounds :)

chart/kubeapps/templates/kubeapps-frontend-config.yaml
absoludity: If you look at the [answer to your link](https://stackoverflow.com/a/53354944), you'll see that what we've used is exactly what the answer suggests as their (2) option - straight location with a rewrite and `break`. The error you mention seems, from the question, to stem from not including a `break` in the rewrite, but that's just from a glance.

chart/kubeapps/templates/kubeapps-frontend-config.yaml
absoludity: Nice, glad that part works without issue :)

chart/kubeapps/values.yaml
```diff
@@ -198,6 +198,7 @@ frontend:
   ## Some K8s distributions such as GKE requires it
   ##
   proxypassAccessTokenAsBearer: false
+  proxypassExtraSetHeader: Authorization "Bearer $cookie_sessionid";
```
absoludity, line 201: Sweet, that'll be a nice addition for the chart. Ooh, and that will make it trivial to do "anonymous" access demos (as in, the user can deploy kubeapps with a pre-configured bearer token of a service-account so no login is required for demos on throwaway clusters (obviously demo only, very dangerous for anything exposed).

chart/kubeapps/templates/kubeapps-frontend-config.yaml
mecampbellsoup: @absoludity sure thing, here is what I am currently trying to do.

### NGINX configuration for k8s API location proxy

I have `.apiServiceURI` set to `http://gatekeeper-kubernetes-ingress.default/k8s/` - this way I can use the `/k8s` portion of the URI in my API gateway so that it proxies ingress to the k8s API backend (and I don't have to set some special `proxy_set_header Host`).

```
      location ~* /api/clusters/default {
        rewrite /api/clusters/default/(.*) /$1 break;
        rewrite /api/clusters/default / break;

      # Helm returns a nil pointer error when accessing foo.bar if foo doesn't
      # exist, even with the `default` function.
      # See https://github.com/helm/helm/issues/8026#issuecomment-756538254
        # Otherwise we route directly through to the clusters with existing credentials.
        proxy_pass http://gatekeeper-kubernetes-ingress.default/k8s/;
      }
```

However, this is not a valid NGINX configuration:

```
(⎈ minikube:default)➜  gatekeeper git:(mc/cloud-charts) ✗ k logs -f gatekeeper-kubeapps-67c86c5574-g24sr
nginx 17:00:25.14
nginx 17:00:25.14 Welcome to the Bitnami nginx container
nginx 17:00:25.15 Subscribe to project updates by watching https://github.com/bitnami/bitnami-docker-nginx
nginx 17:00:25.15 Submit issues and feature requests at https://github.com/bitnami/bitnami-docker-nginx/issues
nginx 17:00:25.15
nginx 17:00:25.15 INFO  ==> ** Starting NGINX setup **
nginx 17:00:25.24 INFO  ==> Validating settings in NGINX_* env vars
nginx 17:00:25.25 INFO  ==> Initializing NGINX

nginx 17:00:25.26 INFO  ==> ** NGINX setup finished! **
nginx 17:00:25.35 INFO  ==> ** Starting NGINX **
2021/02/12 17:00:25 [emerg] 1#1: "proxy_pass" cannot have URI part in location given by regular expression, or inside named location, or inside "if" statement, or inside "limit_except" block in /opt/bitnami/nginx/conf/server_blocks/vhost.conf:33
nginx: [emerg] "proxy_pass" cannot have URI part in location given by regular expression, or inside named location, or inside "if" statement, or inside "limit_except" block in /opt/bitnami/nginx/conf/server_blocks/vhost.conf:33
```

chart/kubeapps/templates/kubeapps-frontend-config.yaml
mecampbellsoup: @absoludity right but I just cannot do `proxy_pass http://my-host.com/k8s` inside this location block (due to the trailing `/k8s` AFAIUI).

chart/kubeapps/templates/kubeapps-frontend-config.yaml
absoludity: That makes sense, but it's what I already understood. The bit that's not clear to me is why what you've stated you need requires a change to the location. Let me use some examples:

Is this an example of a URI you want to proxy here:
```
https://my-kubeapps-domain.com/api/clusters/cluster-name/k8s/api/v1/namespaces
```
?

If so, the existing:

```
      location ~* /api/clusters/{{ .name }} {
```

should match the URL, and the existing:

```
        rewrite /api/clusters/{{ .name }}/(.*) /$1 break;
        rewrite /api/clusters/{{ .name }} / break;
```

should ensure that the *path* of the request is rewritten from `/api/clusters/cluster-name/k8s/api/v1/namespaces` to `/k8s/api/v1/namespaces` . Which is, afaict, exactly what you want? So I'm guessing something didn't quite work when you tried that and hence you're now trying to use an explicit regex is the location etc.? But I'm not sure - that's what I'm trying to understand... why the default config for the location+rewrite didn't do what you want.

On the other hand, if what you're trying to do is have an incoming URL *without* the `/k8s/` such as:

```
https://my-kubeapps-domain.com/api/clusters/cluster-name/api/v1/namespaces
```

and want that routed to your `http://gatekeeper-kubernetes-ingress.default` with the extra `/k8s/`, ie. so the path would be rewritten to `/k8s/api/v/namespaces`, then it should just require a small update to the rewrite rules:

```
        rewrite /api/clusters/{{ .name }}/(.*) /k8s/$1 break;
        rewrite /api/clusters/{{ .name }} /k8s/ break;
``` 

It's just not clear to me which of those (or something else) you're trying to do. If that works, then yes, we'll need to find some way that (1) allows you to include the `k8s` path in the `.apiServerURL` while stripping it off for the `proxy_path` directive (should be easy), and (2) possibly be able to include a subpath in the rewrite.

chart/kubeapps/templates/kubeapps-frontend-config.yaml
absoludity: Yes, the solution I've linked to, similar to our existing config, doesn't try to append paths to the `proxy_path` directive, but rather uses a rewrite to set the path.

chart/kubeapps/templates/kubeapps-frontend-config.yaml
mecampbellsoup: > Is this an example of a URI you want to proxy here:
> 
> ```
> https://my-kubeapps-domain.com/api/clusters/cluster-name/k8s/api/v1/namespaces
> ```

No, I don't think so. AFAIUI, this URL is generated by the kubeapps dashbaord by looking at the [generated k8s API URL](https://github.com/mecampbellsoup/kubeapps/blob/a0164150cd0a0908763618151cdc7e5968d7d1e3/dashboard/src/shared/Kube.ts#L6). So the dashboard's generated URLs will follow the pattern e.g. `/api/clusters/default/can-i` and `/api/clusters/default/namespaces` and not `/k8s/can-i` and `/k8s/namespaces`.

> and want that routed to your `http://gatekeeper-kubernetes-ingress.default` with the extra `/k8s/`, ie. so the path would be rewritten to `/k8s/api/v/namespaces`

Yes, I believe this is what I want!

I think I know what to do now with this NGINX config changeset; I will work on this and ping you once I have pushed something that works. My thinking is that we should add an optional cluster attribute (to the clusters list in `values.yaml`) like `.apiServiceLocation` that, when set, is simply concatenated with the `.apiServiceURL` in the kubeops-related locations, and is added to the rewrite path for what I'm calling the "directly-to-k8s API" location (i.e. `/api/clusters/default`).

pkg/kube/kube_handler.go
```diff
@@ -136,7 +137,13 @@ func NewClusterConfig(inClusterConfig *rest.Config, userToken string, cluster st
 		return config, nil
 	}
 
-	config.Host = clusterConfig.APIServiceURL
+	if clusterConfig.APIServicePath != "" {
+		//config.Host = path.Join(clusterConfig.APIServiceURL, clusterConfig.APIServicePath)
+		baseURL, _ := url.Parse(clusterConfig.APIServiceURL)
+		config.Host, _ = baseURL.Parse(clusterConfig.APIServicePath)
```
mecampbellsoup, line 143: @absoludity I wasn't sure which of the 2 approaches here ☝️ were more idiomatic for Golang (i.e. line 141, or lines 142-143). Any opinion?

pkg/kube/kube_handler.go
```diff
@@ -47,6 +47,7 @@ import (
 type ClusterConfig struct {
 	Name                     string `json:"name"`
 	APIServiceURL            string `json:"apiServiceURL"`
+	APIServicePath           string `json:"apiServicePath"`
```
mecampbellsoup, line 50: @absoludity the idea here is I have created a separate parameter for the k8s API service "path" (e.g. the `/k8s` in my setup).

This way I can simply concatenate `APIServiceURL` with `APIServicePath` here in the kubeops backend code, and conditionally write the NGINX rewrite rules (see above in this diff).

chart/kubeapps/templates/kubeapps-frontend-config.yaml
```diff
@@ -59,8 +59,13 @@ data:
       # the cluster). See clusters option.
       {{- range .Values.clusters }}
       location ~* /api/clusters/{{ .name }} {
+        {{- if .apiServicePath }}
```
mecampbellsoup, line 62: @absoludity do you know how Helm treats undefined values? It errors, right (null pointer)?

If instead if just passed an empty string or something, I could remove this `if/else` and just rely on:

```
rewrite /api/clusters/{{ .name }}/(.*) /{{ .apiServicePath }}/$1 break;
```
... becomes:
```
rewrite /api/clusters/default/(.*) //$1 break;
``` 

NGINX [merges slashes](https://nginx.org/en/docs/http/ngx_http_core_module.html#merge_slashes) by default.

pkg/kube/kube_handler.go
```diff
@@ -136,7 +137,14 @@ func NewClusterConfig(inClusterConfig *rest.Config, userToken string, cluster st
 		return config, nil
 	}
 
-	config.Host = clusterConfig.APIServiceURL
+	if clusterConfig.APIServicePath != "" {
+		//config.Host = path.Join(clusterConfig.APIServiceURL, clusterConfig.APIServicePath)
+		baseURL, _ := url.Parse(clusterConfig.APIServiceURL)
+		fullURL, _ := baseURL.Parse(clusterConfig.APIServicePath)
+		config.Host = fullURL.String()
```
mecampbellsoup, line 144: @absoludity I wasn't sure which of the 2 approaches here ☝️ were more idiomatic for Golang (i.e. line 141, or lines 142-144). Any opinion?

mecampbellsoup:
Hm, for some reason the `local_e2e_steps` step is [failing](https://app.circleci.com/pipelines/github/kubeapps/kubeapps/2016/workflows/6ace6868-1c5c-4e63-8625-bfbbb1380ec9/jobs/36983).

I guess it's related to the [kubeapps rollout failing](https://app.circleci.com/pipelines/github/kubeapps/kubeapps/2016/workflows/6ace6868-1c5c-4e63-8625-bfbbb1380ec9/jobs/36983/parallel-runs/0/steps/0-110)?

chart/kubeapps/templates/kubeapps-frontend-config.yaml
absoludity: Although adding this extra `apiServicePath` to both the chart and the kubeops codebase will work, I think we can achieve the identical result without it and avoid prematurely introducing extra change and complexity into the `kubeops` codebase. We can do this by using `APIServiceURL` as a URL that can also contain a path - the only reason you were unable to use `APIServiceURL` with a path was because you were trying to pass it verbatim (ie. with the path) through to `proxy_pass`. But we can avoid that in the helm template itself.

So, if we allow `apiServiceURL` to be a URL, which can include a path - such as `https://second-cluster:6443/k8s` then we can actually extract the path from that easily enough in the helm template using [urlParse](https://helm.sh/docs/chart_template_guide/function_list/#urlparse) to pull out the apiServicePath from the resulting dict.

Note that the path will either be an empty string, or a string beginning with a `/`, so you won't need your conditional here, and can instead simply do the following always (note the `/` is no longer needed as it will be in the `apiServicePath`:

```
        rewrite /api/clusters/{{ .name }}/(.*) {{ apiServicePath }}/$1 break;
        rewrite /api/clusters/{{ .name }} {{ apiServicePath }}/ break;
``` 

If you need a hand with this, let us know and we can do an example for you separately. Similarly, when you have the `proxy_pass` command, we can ensure that the `scheme://host:port` only is sent (using the values from urlparse).

With that, I don't think we'll need to change the `kubeops` backend at all, and the existing `apiServiceURL` can define a URL with a path as needed? Let me know what you think or if you have any trouble with that.

absoludity (approved):
Excellent work Matt!

Small suggestion below - see what you think but fine without.

chart/kubeapps/templates/kubeapps-frontend-config.yaml
```diff
@@ -57,36 +57,40 @@ data:
...
+    {{- range .Values.clusters }}
       location ~* /api/clusters/{{ .name }} {
-        rewrite /api/clusters/{{ .name }}/(.*) /$1 break;
-        rewrite /api/clusters/{{ .name }} / break;
-
-      {{/* Helm returns a nil pointer error when accessing foo.bar if foo doesn't
+    {{- $apiServiceURL := default "https://kubernetes.default" .apiServiceURL }}
+    {{- $parsed := urlParse $apiServiceURL }}
+    {{- $baseURL := (printf "%s://%s" (get $parsed "scheme") (get $parsed "host")) }}
+    {{- $path := get $parsed "path" }}
```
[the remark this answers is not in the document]
mecampbellsoup: Yea, I am in favor of that idea! I will add a helper function that returns a dict like you suggest. That way the gory details of `urlParse` and concatenating the scheme w/ the host can be hidden away from view 😄 

I will also add a comment at this point in the file explaining why it's necessary.

mecampbellsoup:
@absoludity one more thing - do you know what's up with the e2e tests failing?

absoludity:
> @absoludity one more thing - do you know what's up with the e2e tests failing?

I think it's unrelated to your PR - looks like a recent change has resulted in a test that sometimes passes :/ We'll re-test your PR to see it pass (and eventually find the cause of the flaky test).

chart/kubeapps/templates/kubeapps-frontend-config.yaml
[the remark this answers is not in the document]
mecampbellsoup: @absoludity done in [`ff2ae1c` (#2337)](https://github.com/kubeapps/kubeapps/pull/2337/commits/ff2ae1c778a25cfe2099c54a87cfbf5a908df15d).

Note my `toYaml` / `fromYaml` pattern, where I convert the dict I want into YAML so that I can call `fromYaml` on it in the caller to cast it _back_ to a dict that can then be assigned.

chart/kubeapps/templates/_helpers.tpl
```diff
@@ -6,6 +6,13 @@ Expand the name of the chart.
 {{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
 
+{{- define "kubeapps.parseAPIServiceURL" -}}
+{{- $parsed := urlParse . }}
+{{- $baseURL := (printf "%s://%s" (get $parsed "scheme") (get $parsed "host")) }}
+{{- $path := get $parsed "path" }}
+{{ dict "baseURL" $baseURL "path" $path | toYaml }}
```
mecampbellsoup, line 13: @absoludity this template "returns" a YAML string which can be composed (in the caller) with `| fromYaml` so that I can "return" a dict... 🤯

chart/kubeapps/templates/kubeapps-frontend-config.yaml
```diff
@@ -57,36 +57,40 @@ data:
...
       # used with an auth-proxy where k8s credentials never leave
       # the cluster). See clusters option.
-      {{- range .Values.clusters }}
+    {{- range .Values.clusters }}
       location ~* /api/clusters/{{ .name }} {
-        rewrite /api/clusters/{{ .name }}/(.*) /$1 break;
-        rewrite /api/clusters/{{ .name }} / break;
-
-      {{/* Helm returns a nil pointer error when accessing foo.bar if foo doesn't
+    {{/* We need to split the API service URL(s) into the base url and the path segment so
```
mecampbellsoup, line 62: @absoludity here is the descriptive comment.

chart/kubeapps/templates/kubeapps-frontend-config.yaml
```diff
@@ -57,36 +57,40 @@ data:
...
+           those configurations using a path can be appropriately rewritten below while
+           ensuring the proxy_pass statement is given the base URL only. */}}
+    {{- $apiServiceURLMap := include "kubeapps.parseAPIServiceURL" .apiServiceURL | fromYaml }}
```
mecampbellsoup, line 65: @absoludity and here is the `| fromYaml` usage, whereby I consume the YAML string output of my named template `kubeapps.parseAPIServiceURL` and cast it to a dict which is then assigned for usage/reference in this file, e.g. `$apiServiceURLMap.path`.

absoludity: Yeah, given it's just two values, seems a bit odd to go through the yaml necessities (I forgot the helpers are evaluated as strings only).

We could just do two separate helpers, but at that point, given they're not re-used, it may just be simpler to do it inline as you'd done originally (sorry). To save you the hassle, I've redone it here, and also added sane defaults so it works when an `.apiServiceURL` isn't defined: https://github.com/mecampbellsoup/kubeapps/pull/1

absoludity (changes requested):
Just need the default values restored as in the linked PR, but up to you what you do with the helper. See what you think.

chart/kubeapps/templates/kubeapps-frontend-config.yaml
```diff
@@ -57,36 +57,42 @@ data:
...
+    {{- range .Values.clusters }}
       location ~* /api/clusters/{{ .name }} {
-        rewrite /api/clusters/{{ .name }}/(.*) /$1 break;
-        rewrite /api/clusters/{{ .name }} / break;
-
-      {{/* Helm returns a nil pointer error when accessing foo.bar if foo doesn't
+    {{/* We need to split the API service URL(s) into the base url and the path segment so
+           those configurations using a path can be appropriately rewritten below while
+           ensuring the proxy_pass statement is given the base URL only. */}}
+    {{- $parsed := urlParse (default "" .apiServiceURL) }}
```
mecampbellsoup, line 65: @absoludity only thing I'm unsure about now is why the default value here is `''` instead of `'https://kubernetes.default'` like it [is/was before](https://github.com/kubeapps/kubeapps/pull/2337/files?w=1#diff-122c6e16fc44dcfed4ed0ece01fc53e3d16f46fa3fcd3ba138e61dc3891206a5L82)...

Shouldn't we do:

```
{{- $parsed := urlParse (default "https://kubernetes.default" .apiServiceURL) }}
```
?

chart/kubeapps/templates/kubeapps-frontend-config.yaml
```diff
@@ -57,36 +57,42 @@ data:
...
       location ~* /api/clusters/{{ .name }} {
-        rewrite /api/clusters/{{ .name }}/(.*) /$1 break;
-        rewrite /api/clusters/{{ .name }} / break;
-
-      {{/* Helm returns a nil pointer error when accessing foo.bar if foo doesn't
+    {{/* We need to split the API service URL(s) into the base url and the path segment so
+           those configurations using a path can be appropriately rewritten below while
+           ensuring the proxy_pass statement is given the base URL only. */}}
+    {{- $parsed := urlParse (default "https://kubernetes.default" .apiServiceURL) }}
+    {{- $apiServiceBaseURL := urlJoin (pick $parsed "scheme" "host") }}
```
mecampbellsoup, line 66: @absoludity I changed the `printf` to `urlJoin (pick ...` here as you can see. Seems better to use a stdlib function to generate a [partial] URL instead of doing my own formatting 😄

chart/kubeapps/templates/kubeapps-frontend-config.yaml
```diff
@@ -57,36 +57,42 @@ data:
...
+    {{- range .Values.clusters }}
       location ~* /api/clusters/{{ .name }} {
-        rewrite /api/clusters/{{ .name }}/(.*) /$1 break;
-        rewrite /api/clusters/{{ .name }} / break;
-
-      {{/* Helm returns a nil pointer error when accessing foo.bar if foo doesn't
+    {{/* We need to split the API service URL(s) into the base url and the path segment so
+           those configurations using a path can be appropriately rewritten below while
+           ensuring the proxy_pass statement is given the base URL only. */}}
+    {{- $parsed := urlParse (default "https://kubernetes.default" .apiServiceURL) }}
```
mecampbellsoup, line 65: I made the default value `https://kubernetes.default` here.

chart/kubeapps/templates/kubeapps-frontend-config.yaml
```diff
@@ -57,36 +57,42 @@ data:
...
         proxy_set_header PINNIPED_PROXY_API_SERVER_CERT {{ .certificateAuthorityData }};
-        {{- end }}
+    {{- end }}
         proxy_pass http://kubeapps-internal-pinniped-proxy.{{ $.Release.Namespace }}:{{ $.Values.pinnipedProxy.service.port }};
-      {{- else }}
+    {{- else }}
         # Otherwise we route directly through to the clusters with existing credentials.
-        proxy_pass {{ default "https://kubernetes.default" .apiServiceURL }};
-        {{- if .certificateAuthorityData }}
+        proxy_pass {{ $apiServiceBaseURL }};
```
mecampbellsoup, line 88: No `default` required here now that it happens above one line 65.

mecampbellsoup (commented):
Here is my generated NGINX config (relevant portions):

```nginx
      # Ensure each cluster can be reached (should only be
      # used with an auth-proxy where k8s credentials never leave
      # the cluster). See clusters option.
      location ~* /api/clusters/default {

        rewrite /api/clusters/default/(.*) /k8s/$1 break;
        rewrite /api/clusters/default /k8s/ break;


        # Otherwise we route directly through to the clusters with existing credentials.
        proxy_pass http://api-gateway-kubernetes-ingress.cloud;
      }

      # Forward '/api/assetsvc' to '/assetsvc'
      # but preserving the encoding (eg. '%2F' is not converted to '/')
      # see https://serverfault.com/a/906479
      # Ex: from "/api/assetsvc/what$2Fever?param=value"
      #     it matches as $1="/what$2Fever" and $args="param=value"
      #     downstream services will receive "/assetsvc/what$2Fever?param=value"
      location ~* /api/assetsvc {
        rewrite ^ $request_uri; # pass the encoded url downstream as is,
        rewrite /api/assetsvc([^?]*) /assetsvc$1?$args break;
        proxy_set_header Authorization "Bearer $cookie_sessionid";
        proxy_pass http://kubeapps-internal-kubeops:8080;
      }
```

chart/kubeapps/templates/kubeapps-frontend-config.yaml
mecampbellsoup: [`936ded6` (#2337)](https://github.com/kubeapps/kubeapps/pull/2337/commits/936ded61f610d1ab65efe7076184f0c962b1bdb5)

mecampbellsoup:
@absoludity wow, somehow the 7th check (e2e tests) automagically passes now! 😸

chart/kubeapps/templates/kubeapps-frontend-config.yaml
absoludity: Yep, it works there too. I was keen to leave it where it was for other reasons (I'll need the actual value set `""` to be passed to other services such as pinniped-proxy in this same file, so that they can work with what was actually set by the user and make decisions accordingly, rather than not knowing if the user explicitly set `https://kubernetes.default`). But I can update that if/when we ever need this.

chart/kubeapps/templates/kubeapps-frontend-config.yaml
absoludity: Very nice!

chart/kubeapps/templates/kubeapps-frontend-config.yaml
absoludity: Yep, see above.

absoludity (approved):

absoludity:
> @absoludity wow, somehow the 7th check (e2e tests) automagically passes now! smile_cat

Andres was going to be looking into the failure his Wednesday, so it could have been updated and fixed on master... or we were just lucky :)

PR #2340, opened 2021-02-08
Update pinniped-proxy deps to latest kube and k8s-openapi

absoludity:
### Description of the change

This PR updates the dependencies of pinniped-proxy as I was previously using a differing version of kube-runtime and kube-derive, which can cause compilation issues (as seen on #2286 .

### Benefits

Updated libs, compiles without issue.

### Possible drawbacks

None known.


antgamdia (approved):
Awesome! It's great now we can use the minideb image directly. It will ease the release of the official image, I guess.

PR #2345, opened 2021-02-08
Remove unnecessary code

andresmgot:
### Description of the change

So this old code was used to select the namespace in which Kubeapps is running when using telepresence. It's not needed since this retrieved anyway from the dummy config file in `dashboard/public/config.json`.

### Benefits

No need to include in production code anything related to the dev environment.


antgamdia (approved):
Great! When implementing the i18n thing I wondered why this REACT_APP was still there. Happy to see this old code is removed :)

Issue #2346, opened 2021-02-08
Add support for some kind of translation engine


antgamdia:
I would like to be able to provide my own en.yml to be able to modify the Kubeapps welcome message.

(from Danoo's project note)

PR #2347, opened 2021-02-08
Add support for adding translations

antgamdia:
### Description of the change

This is the first PR aimed to support internationalization and localization in Kubeapps, not only in build time (supported and approved translations in Kubeapps) but also in run-time (by means of customizing `values.yaml`).
Under the hood it is using the `react-intl` library, one of the most supported libs for i18n.

In subsequent PRs actual i18n and l10n will be addressed, this PR is just to add the logic to support so.

The run-time configuration is achieved with the same approach as we are currently doing for the css-injection thing.
Also, note that two formats are allowed here: i) simple key, value json; ii) compiled AST. I've added documentation for that purpose as well.

### Benefits

Kubeapps will be ready to support translations :)

### Possible drawbacks

Once we change the hardcoded strings in another PR, many tests will fail. Nonetheless, the lib has support for testing envs (I have to investigate it a little bit more).

### Applicable issues

  - related  #2346

### Additional information

N/A


dashboard/package.json
```diff
@@ -100,7 +101,9 @@
     "eslint": "eslint '**/*.{ts,tsx}' --ignore-pattern '**/*.d.ts' --max-warnings=0",
     "eslint-fix": "eslint '**/*.{ts,tsx}' --ignore-pattern '**/*.d.ts' --fix --max-warnings=0",
     "lint": "npm-run-all lint-js lint-css eslint",
-    "ts-compile-check": "tsc -p tsconfig.json --noEmit"
+    "ts-compile-check": "tsc -p tsconfig.json --noEmit",
+    "extract-lang": "formatjs extract 'src/**/*.ts*' --out-file lang/en.json --id-interpolation-pattern '[sha512:contenthash:base64:6]' --format simple",
+    "compile-lang": "formatjs compile-folder lang src/locales/ --ast --format simple"
```
andresmgot, line 106: are these scripts used?

dashboard/src/containers/Root.tsx
```diff
@@ -2,23 +2,48 @@ import Header from "components/Header";
...
+    super(props);
+    this.state = getDefaulI18nConfig();
+    this.initLocale();
+  }
+
+  public async initLocale() {
+    const fullLang = (navigator.languages && navigator.languages[0]) || navigator.language;
+    const lang = fullLang.toLowerCase().split(/[_-]+/)[0];
+    this.setState(await getCustomI18nConfig(ISupportedLangs[lang]));
+  }
```
andresmgot, line 29: please use React hooks instead

dashboard/src/shared/I18n.ts
```diff
@@ -0,0 +1,45 @@
...
+import { axiosWithAuth } from "./AxiosInstance";
+
+export interface II18nConfig {
+  locale: string;
+  messages: Record<string, string>;
+}
+
+export enum ISupportedLangs {
+  en = "en",
+  custom = "custom",
```
andresmgot, line 11: isn't it better to leave it just with `en`?

docs/developer/translate-kubeapps.md
```diff
@@ -0,0 +1,33 @@
...
+
+The process is as follows:
+
+  1. Select a base file in the `/dashboard/lang` folder. For instance, `en.json`.
+  2. Translate the strings into your language and save the document as `XX.json`, where `XX` is a two-letter [ISO 639-1 code](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes).
+  3. Execute `yarn compile-lang` to generate the AST-compiled strings, this will improve the performance of the application.
+  4. Add your language to the list of supported ones in the code.
+  5. Feel free to contribute with a new language and send a PR :)
+  
+> Developers should note that when adding a new literal, `yarn extract-lang` must be executed.
```
andresmgot, line 33: thanks for the docs! I would add a note saying this is not ready yet, just in case.

chart/kubeapps/values.yaml
```diff
@@ -496,6 +496,13 @@ dashboard:
   ##      background-color: #991700;
   ##    }
   customStyle: ""
+  ## Custom translations injected to the Dashboard to customize the str
+  ## e.g:
+  ## customLocale: |-
```
andresmgot, line 501: I would add here a link to the docs (with a note saying that it's not ready yet).

andresmgot (commented):
Thanks @antgamdia, I have some minor comments.

dashboard/package.json
antgamdia: Yes, they are used for extracting the messages using the default values. Then these langs should be compiled in AST for performance purposes. See https://formatjs.io/docs/getting-started/message-extraction/ for more info.

dashboard/src/containers/Root.tsx
antgamdia: Oook, got it! I (don't know why) supposed that they weren't allowed for the root component. I will migrate it to hooks. Thanks!

dashboard/src/shared/I18n.ts
antgamdia: We use it when loading the custom one (from values.yaml); don't know why we should remove it

chart/kubeapps/values.yaml
antgamdia: Good idea for the moment. However, once this feature is fully-finished, I'd like to add this info to the chart docs, but, for now, I agree with you

dashboard/src/shared/I18n.ts
andresmgot: I thought this worked differently. I'd assume that you have the default language there (`en`) and you could overwrite it with the one from the chart. Isn't that the case?

antgamdia: Ahh, no no. The current impl is:
(side note: locales != messages; the former is about date, number, time, etc formats)

1. Are the custom messages used? 
  - Yes: use custom: locale=english, messages=custom
    - (runtime): have I the custom message for the string X?
      - Yes: use custom message
      - No: use default harcoded message
  - No:  GOTO 2
2. What is the current language user configuration? Is the ISO code one of the ISupported?
  - Yes: use this lang: locale=lang, messages=lang
  - No: use default: locale=english, messages=english

(This behavior should be also documented, I guess :P)

andresmgot: isn't it easier to just have `en` rather than `en` + `custom`?

antgamdia: I don't get your point here? Why would it be easier? 
I mean, we can remove it and pass the messages directly in [this line](https://github.com/kubeapps/kubeapps/pull/2347/files#diff-f822cbb1da0404c050e7403bee68b227a580cfce308444f6330b57acd38de6c7R34), but what's the benefit?

andresmgot: yep, getting rid of `getCustomI18nConfig` and supporting just one language (but I may be missing something)

dashboard/src/containers/Root.tsx
```diff
@@ -1,55 +1,47 @@
...
-  public async initLocale() {
-    const fullLang = (navigator.languages && navigator.languages[0]) || navigator.language;
-    const lang = fullLang.toLowerCase().split(/[_-]+/)[0];
-    this.setState(await getCustomI18nConfig(ISupportedLangs[lang]));
-  }
+function Root() {
+  const [i18nConfig, setI18nConfig] = useState(getDefaulI18nConfig());
+  useEffect(() => {
+    initLocale().then(customI18nConfig => setI18nConfig(customI18nConfig));
+  }, []);
```
andresmgot, line 23: :+1: 

Now you just need to add a test for this :) You can see an example of how to mock `useEffect` + an async function here: https://github.com/kubeapps/kubeapps/blob/master/dashboard/src/components/Config/AppRepoList/AppRepoList.test.tsx#L75

dashboard/src/shared/I18n.ts
andresmgot: In any case, I am not feeling strong about this so what go with what you prefer.

dashboard/src/shared/I18n.ts
```diff
@@ -2,44 +2,47 @@ import messages_en from "../locales/en.json";
...
+// Load here the compiled messages for each supported language
+messages[ISupportedLangs.en] = messages_en;
 
-export function getI18nConfig(lang: ISupportedLangs): II18nConfig {
-  if (lang && ISupportedLangs[lang]) {
-    return { locale: lang, messages: messages[lang] };
-  } else {
-    return getDefaulI18nConfig();
+export default class I18n {
+  public static getDefaulI18nConfig(): II18nConfig {
```
andresmgot, line 20: I am not sure why you have moved these to a class but if it's already named `I18n` you can remove that from the method names `I18n.getDefaultConfig()`...

andresmgot (approved):
thanks for the tests

PR #2359, opened 2021-02-09
Add registered icon to the MongoDB references

jotamartos:
We need to add the registered icon to all the MongoDB references in our products


antgamdia (approved):
Perhaps we can even get rid of any reference to MongoDB and just say *Since 2.0, the only database supported is PostgreSQL*. Anyway LGTM.

andresmgot (approved):

PR #2360, opened 2021-02-09
Add translatable strings in login pages

### LINKED ISSUE #2325: "Login via OIDC Provider" text not configurable
### Description:

Currently if you configure your Kubeapps installation to use, say VMware Cloud Services for authentication, you don't want users to have to click on the label "Login via OIDC Provider" to log in. Instead, you should be able to configure this text during installation so it can say "Login via VMware Cloud Services" for example.

### LINKED ISSUE #2346: Add support for some kind of translation engine

antgamdia:
## Description of the change

This PR is to simply add translatable strings at the login pages. The texts will be the same, but they can be translated into other languages via `values.yaml`.

### Benefits

Login pages (OIDC and token) will accept custom translations

### Possible drawbacks

The corresponding tests have to be modified since the `react-intl` context is not available when mounting the components. Despite changes herein performed, there are a couple of test cases failing. I still do not why. I have to investigate it a little bit more.

### Applicable issues

  - fixes #2346 
  - fixes #2325

### Additional information

c.f., https://formatjs.io/docs/guides/testing/#helper-function-1


dashboard/lang/en.json
```diff
@@ -1 +1,12 @@
-{}
+{
+  "Kubeapps": "Kubeapps",
+  "Submit": "Submit",
+  "Token": "Token",
+  "error-login-token": "There was an error connecting to the Kubernetes API. Please check that your token is valid.",
+  "login-desc-oidc": "Your cluster operator has enabled login via an authentication provider.",
+  "login-desc-token": "Your cluster operator should provide you with a Kubernetes API token.",
+  "login-oidc": "Login via OIDC Provider",
```
absoludity, line 8: This is the specific one I'm after :P

dashboard/src/components/LoginForm/OauthLogin.tsx
```diff
@@ -1,29 +1,38 @@
...
 }
 
 function OAuthLogin(props: ILoginFormProps) {
+  const intl = useIntl();
   return (
     <section className="title" aria-labelledby="login-title" aria-describedby="login-desc">
       <h3 id="login-title" className="welcome">
-        Welcome to <span>Kubeapps</span>
+        {intl.formatMessage({ id: "login-title-welcome", defaultMessage: "Welcome to" })}
+        <span>{intl.formatMessage({ id: "Kubeapps", defaultMessage: "Kubeapps" })}</span>
```
absoludity, line 15: I don't think we need this translated in any language do we? Does the product name change?

absoludity (approved):
+1 for when you have tests passing :)

dashboard/src/shared/specs/mountWrapperIntl.tsx
```diff
@@ -0,0 +1,41 @@
...
+  );
+
+export const mountWrapperIntl = (store: MockStore, children: React.ReactElement) =>
+  mount(
+    <Provider store={store}>
+      <IntlProvider locale={locale} key={locale} messages={messages} defaultLocale={locale}>
+        <Router>{children}</Router>
+      </IntlProvider>
+    </Provider>,
+  );
```
andresmgot, line 34: I would just use the one with the `<Provider` for simplicity. Also, can we just add the `IntlProvider` to the existing wrapper at `specs/mountWrapper`? (That way we don't need to decide which wrapper to use but I'm not sure if that would break everything. It shouldn't)

dashboard/src/components/LoginForm/OauthLogin.tsx
andresmgot: I believe the goal is to include **all** the strings, not only the ones that may require translation.

dashboard/src/shared/specs/mountWrapperIntl.tsx
antgamdia: The issue is that many tests use the usual `mount` and `swallow` methods and these ones also need to have the intl context.

Regarding adding IntlProvider to the existing wrapper, yep, it is possible and it doesn't break anything (apart from some snapshots)

dashboard/src/locales/en.json
```diff
@@ -1 +1,62 @@
-{}
+{
```
andresmgot, line 1: can we gitignore this file? (to avoid confusion with dashboard/lang/en.json)

dashboard/src/components/LoginForm/LoginForm.test.tsx
```diff
@@ -37,30 +34,23 @@ const defaultProps = {
 
 const authenticationError = "it's a trap";
 
-context("while authenticating", () => {
```
andresmgot, line 40: did you miss this test? (We talked offline to avoid using `itBehavesLike`)

dashboard/src/components/FilterGroup/__snapshots__/FilterGroup.test.tsx.snap
```diff
@@ -13,106 +13,181 @@ exports[`renders a multicheckbox 1`] = `
...
-          "goBack": [Function],
-          "goForward": [Function],
-          "length": 1,
-          "listen": [Function],
-          "location": Object {
-            "hash": "",
-            "pathname": "/",
-            "search": "",
-            "state": undefined,
+  <IntlProvider
```
andresmgot, line 16: snapshots are so unhelpful... I'm thinking on removing all of them

andresmgot (approved):
minor questions here

dashboard/src/locales/en.json
antgamdia: Mmm, I don't really know if it's a good option. If we ignore files under `src/locales/*.json` users cloning the repo will get many console errors (despite the strings will fallback to the default ones) because the compiled AST translation is not present.
If we ignore them, I think we should modify the `start` script from the `package.json` in order to add the language compilation beforehand.

dashboard/src/components/LoginForm/LoginForm.test.tsx
antgamdia: You're absolutely right, thanks for letting me know. Adding it back.

dashboard/src/locales/en.json
andresmgot: yes, I would do so (modify the `start ` method). The same approach we do with the CSS files (compiled from scss source). You may need to update the `build` script as well because that's the one executed in the CI:

```
    "start": "npm-run-all -p watch-css start-js",
    "build": "npm-run-all build-css build-js",
```

andresmgot (approved):
thanks @antgamdia I would generate the lang files on the fly

dashboard/src/locales/en.json
antgamdia: Great, good idea. Do you want me to create a separate PR for that? Or just adding these changes here is fine?

andresmgot: whatever is easier for you

antgamdia: Perhaps, to keep this pr as simple as possible, let's do it on another PR.
If it's OK with you, I'm gonna merge this one and will send the other PR asap.

andresmgot: yep

PR #2374, opened 2021-02-10
Request tags in parallel

andresmgot:
### Description of the change

Similar to #2373 but this time for getting tags. As mentioned in that issue, checking tag per tag is a bit slow (since it requires an API request per tag) so doing it in parallel improves a lot the total time took.

### Benefits

From the test run, I got a 10x faster sync in the part of the sync that checks the tags.

In total (this PR plus the previous one), my test sync went from ~25 to 4 minutes.

### Applicable issues

  - ref #2232

### DIFF
--- a/cmd/asset-syncer/utils.go
+++ b/cmd/asset-syncer/utils.go
@@ -74,6 +74,17 @@ type pullChartResult struct {
 	Error error
 }
 
+type checkTagJob struct {
+	AppName string
+	Tag     string
+}
+
+type checkTagResult struct {
+	checkTagJob
+	isHelmChart bool
+	Error       error
+}
+
 type httpClient interface {
 	Do(req *http.Request) (*http.Response, error)
 }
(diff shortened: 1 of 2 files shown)


cmd/asset-syncer/utils.go
```diff
@@ -318,57 +329,89 @@ func (o *ociAPICli) TagList(appName string) (*TagList, error) {
...
 // Checksum returns the sha256 of the repo by concatenating tags for
 // all repositories within the registry and returning the sha256.
 // Caveat: Mutated image tags won't be detected as new
 func (r *OCIRegistry) Checksum() (string, error) {
-	tagList := map[string]TagList{}
+	r.tags = map[string]TagList{}
+	checktagJobs := make(chan checkTagJob, numWorkers)
+	tagcheckRes := make(chan checkTagResult)
+	var wg sync.WaitGroup
+	wg.Add(1)
```
absoludity, line 368: So similar comment here - this isn't how a wait group should be used. See what you think. The result should be the same - great work on getting it down to 4m!

PR #2377, opened 2021-02-11
fix deploy other clusters

absoludity:
### Description of the change

Currently deploying to an additional cluster fails due to a recent change (see [comment](https://github.com/kubeapps/kubeapps/pull/2310/files#r574249840)) which switched from fetching an AppRepository only ever from the Kubeapps cluster to instead trying to fetch the AppRepository from the *current* cluster.

### Benefits

You can once again deploy to other clusters.

### Possible drawbacks

Not as a result of this, but we currently don't test deploying to other clusters in CI. We should (#2367).

### DIFF
--- a/cmd/kubeops/internal/handler/handler.go
+++ b/cmd/kubeops/internal/handler/handler.go
@@ -175,7 +175,7 @@ func CreateRelease(cfg Config, w http.ResponseWriter, req *http.Request, params 
 		returnErrMessage(err, w)
 		return
 	}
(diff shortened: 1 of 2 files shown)


andresmgot (approved):
Thanks for fixing the issue! (and sorry for causing it)

This requires a (unit) test, no?

andresmgot:
is this blocked by anything @absoludity ?

absoludity:
> is this blocked by anything @absoludity ?

Nope, I was looking briefly at how to unit-test the change but didn't find a solution I was happy with given the underlying support of unit-tests for the functions used. Happy to land it.

PR #2386, opened 2021-02-11
Fix 'All Namespaces' button

### LINKED ISSUE #2289: Show apps in all namespace button keeps checked after changing the namespace
### Description:

Clicking in the button to show all namespace and then changing the namespace causes the toggle button to be active but the Kubeapps shows only the applications of the selected namespace.

Screenshot from 2021-01-28 12-26-41

Refreshing or clicking the button again solves the issue.

1. Click on "Show Apps in All Namespaces"
2. Change namespace in the context selector

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.5.0", GitCommit:"32c22239423b3b4ba6706d450bd044baffdcf9e6", GitTreeState:"clean", GoVersion:"go1.15.6"}
```

```
2.1.0
```

andresmgot:
### Description of the change

Changing the namespace doesn't affect the `all namespaces` selection so the state should not be affected.

### Applicable issues

  - fixes #2289

### DIFF
--- a/dashboard/src/components/AppList/AppList.tsx
+++ b/dashboard/src/components/AppList/AppList.tsx
@@ -9,7 +9,7 @@ import * as qs from "qs";
 import * as React from "react";
 import { useEffect, useState } from "react";
 import { useDispatch, useSelector } from "react-redux";
-import { useLocation } from "react-router";
+import * as ReactRouter from "react-router";
 import { Link } from "react-router-dom";
 import { Kube } from "shared/Kube";
 import { IStoreState } from "../../shared/types";
@@ -20,7 +20,7 @@ import "./AppList.css";
 import AppListGrid from "./AppListGrid";
 
 function AppList() {
-  const location = useLocation();
+  const location = ReactRouter.useLocation();
   const searchQuery = qs.parse(location.search, { ignoreQueryPrefix: true }).q?.toString() || "";
   const allNSQuery = qs.parse(location.search, { ignoreQueryPrefix: true }).allns || "";
   const dispatch = useDispatch();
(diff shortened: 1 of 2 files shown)


dashboard/src/components/AppList/AppList.tsx
```diff
@@ -20,7 +20,7 @@ import "./AppList.css";
 import AppListGrid from "./AppListGrid";
 
 function AppList() {
-  const location = useLocation();
+  const location = ReactRouter.useLocation();
```
antgamdia, line 23: Just out of curiosity: why this change?

antgamdia (approved):

dashboard/src/components/AppList/AppList.tsx
andresmgot: that's an interesting "gotcha". The problem is how `useLocation` is exported on it's package. It's directly exported as a function:

```
export function useLocation<S = H.LocationState>(): H.Location<S>;
```

but for mocking it as a spy, we are importing the whole `ReactRouter`: `jest.spyOn(ReactRouter, "useLocation")`. The problem is that if in the functional file, we only import the function (`import { useLocation } from react-router`) it won't use the mocked version of `ReactRouter` and  will import the real function. (testing things).

The thing is that this doesn't happen with other methods (like `useState` that I'm also using here) because it's exported as a method of a namespace:

```
declare namespace React {
...
    function useState<S>(initialState: S | (() => S)): [S, Dispatch<SetStateAction<S>>];
```

so we will be always mocking the namespace, regardless how you import it. :)

antgamdia: Ahhhh, got it. Thanks a lot for the explanation! It will be useful for my future tests :P

PR #2408, opened 2021-02-16
Allow to skip TLS verification and use a custom CA

andresmgot:
### Description of the change

Allow to use HTTPS for Helm repositories/registries but skip the TLS verification for self-signed certificates.

### Benefits

Since the `Resolver` is now using a custom HTTP client, we can additionally use a custom CA

### Applicable issues

  - ref #2232

### Additional information

As an extra, the `netClient` variable is no longer global but part of the repository `struct`.

Reviewing per commit may be easier.


absoludity (approved):
Nice improvement based on your previous work :)

PR #2300, opened 2021-02-01
Delete unused API endpoints

antgamdia:
### Description of the change

This PR is just to delete a bunch of old endpoints that are no longer used by any part of the code.

### Benefits

We finally will: 1) get rid of the non-cluster aware docs; 2) replace the reference to tiller by kubeops; 3) remove the old non-cluster aware `/kube/` proxypass.

### Possible drawbacks

Can't think of any, but I'm opening this draft PR to see if CI complains

### Applicable issues

  - fixes #2298

### Additional information

Related to  #2291. If this PR becomes merged, `tiller-deploy` should be also renamed there.

PR #2412, opened 2021-02-17
Revert last chart version bump

andresmgot:
### Description of the change

At the moment, `master` builds are broken because we have changed some routes (https://github.com/kubeapps/kubeapps/pull/2300). The last version `v2.1.0` uses `tiller-proxy` URLs while the current NGINX config in the latest chart configures `kubeops`. We have two options: either add backward-compatible config and routes or disable the "latest release" builds knowing that **we cannot release a new version of the chart until we release a new version of Kubeapps**.

I am going for the second option since we plan to release a new version soon anyway. Note that we are only testing the combination of the current chart + the latest release when releasing a new chart version. Because of that, I am reverting the change at #2359 for the version bump.

### Benefits

Fix master

### Possible drawbacks

Block chart releases

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -54,13 +54,14 @@ workflows:
             - test_dashboard
             - build_go_images
             - build_dashboard
-      - GKE_1_15_LATEST_RELEASE:
-          <<: *build_on_master
-          requires:
-            - test_go
-            - test_dashboard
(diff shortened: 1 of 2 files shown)


antgamdia (approved):
Great! Thank you for checking this.
I also agree with the second option since the next release will be soon.

PR #2413, opened 2021-02-17
Add skip TLS verification button

andresmgot:
### Description of the change

Frontend part of #2408.

Adding a button to skip TLS verification if needed.

Screenshot from 2021-02-17 11-33-15

### Applicable issues

  - ref #2232

### DIFF
--- a/dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
+++ b/dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
@@ -1,4 +1,5 @@
 import { CdsButton } from "@clr/react/button";
+import { CdsCheckbox } from "@clr/react/checkbox";
 import actions from "actions";
 import Alert from "components/js/Alert";
 import Column from "components/js/Column";
@@ -22,6 +23,7 @@ interface IAppRepoFormProps {
     syncJobPodTemplate: string,
     registrySecrets: string[],
     ociRepositories: string[],
+    skipTLS: boolean,
   ) => Promise<boolean>;
   onAfterInstall?: () => void;
   namespace: string;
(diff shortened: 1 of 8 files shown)


antgamdia (approved):
Great! Happy to see this functionality :)

dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
```diff
@@ -503,10 +511,22 @@ export function AppRepoForm(props: IAppRepoFormProps) {
...
               placeholder={"-----BEGIN CERTIFICATE-----\n...\n-----END CERTIFICATE-----"}
               value={customCA}
+              disabled={skipTLS}
               onChange={handleCustomCAChange}
             />
           </div>
         </div>
+        <div className="clr-form-control">
+          <CdsCheckbox>
+            <label className="clr-control-label">Skip TLS Verification</label>
```
absoludity, line 521: This is a bit tricky: I don't think we should be exposing an option to skip TLS normally, but I get that here we're doing so because we can't set a custom CA for an OCI repository (due to the underlying library)?

If that's the case, we should be clear about that to the user, at least in a tool-tip or something (but preferably onscreen). What I'd expect (as a user) is if I select the OCI repository type then:

* The customCA is disabled (done, great)
* An short explanation with a link to an issue with more info is displayed: "custom certificate authority certs are [not currently supported for OCI repositories](link to issue). You can instead skip TLS verification for this repository if it requires a custom certificate authority."

Other questions:
* Can people specify non-tls registries? (ie. `http://example.com`) And if so, should we only present both fields (customCA and skip TLS) for URLs that are `https` or have no proto specified (I assume we assume https?).

absoludity (approved):

dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
andresmgot: After #2408 it's also possible to use a custom CA for an OCI repository. This is just to expose the possibility of skipping the TLS verification (that actually works for both OCI and non-OCI repos).

> Can people specify non-tls registries? (ie. http://example.com)

Yes, but I haven't tried for OCI registries. This requires a flag for oras but since we are using the custom net client it may work out of the box as well.

> And if so, should we only present both fields (customCA and skip TLS) for URLs that are https or have no proto specified

:shrug: in any case, we specify that the  custom CA field is optional.

> (I assume we assume https?).

Yep, if not given, `https` is assumed.

PR #2414, opened 2021-02-17
Order tags

andresmgot:
### Description of the change

The order in which tags are received is not assured, making some golang test flaky.

Fixes https://app.circleci.com/pipelines/github/kubeapps/kubeapps/2039/workflows/569b60bb-b6a1-4da2-a6e9-51f44c502985/jobs/37138/parallel-runs/0/steps/0-103


antgamdia (approved):
Great, perhaps it will cause some performance issues when having lots of tags, but for now, it's ok.

andresmgot:
> Great, perhaps it will cause some performance issues when having lots of tags, but for now, it's ok.

Sorting is quite fast (compared to way slower actions like API requests that we are doing :sweat_smile: )

PR #2415, opened 2021-02-17
Encode OCI repository names before storing them

andresmgot:
### Description of the change

When using an OCI repository, there two options:
1. Use the registry URL with the repo and the app name without the repository (e.g. `registry: docker.io/bitnami, app: wordpress`).
2. Use the registry URL without the repo and include that in the app name (e.g. `registry: docker.io, app: bitnami/wordpress`).

The second was not working because the app was stored with `/` while the request to the asset service the name is URL-encoded.

### Applicable issues

  - ref #2232

### DIFF
--- a/cmd/asset-syncer/utils.go
+++ b/cmd/asset-syncer/utils.go
@@ -522,7 +522,7 @@ func pullAndExtract(repoURL *url.URL, appName, tag string, puller helm.ChartPull
 	}
 
 	// Encode repository names to store them in the database.
-	encodedAppName := url.QueryEscape(appName)
+	encodedAppName := url.PathEscape(appName)
 
 	return &models.Chart{
 		ID:            path.Join(r.Name, encodedAppName),
--- a/cmd/asset-syncer/utils.go
+++ b/cmd/asset-syncer/utils.go
@@ -521,9 +521,12 @@ func pullAndExtract(repoURL *url.URL, appName, tag string, puller helm.ChartPull
 		})
 	}
 
+	// Encode repository names to store them in the database.
+	encodedAppName := url.QueryEscape(appName)
+
 	return &models.Chart{
-		ID:            path.Join(r.Name, appName),
-		Name:          appName,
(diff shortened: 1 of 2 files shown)


cmd/asset-syncer/utils.go
```diff
@@ -521,9 +521,12 @@ func pullAndExtract(repoURL *url.URL, appName, tag string, puller helm.ChartPull
 		})
 	}
 
+	// Encode repository names to store them in the database.
+	encodedAppName := url.QueryEscape(appName)
```
antgamdia, line 525: In other escape/unescape things, we are using Path(Un)Escape instead of this one since the request URI was actually invoked without query params (I mean, the chart name was part of the URI).
AFAIK, QueryEscape performs the escape function on a different subset of characters.

If `appName` comes from a query param, everything is fine. Otherwise, perhaps we can use PathEscape. See what you think.

antgamdia (approved):

cmd/asset-syncer/utils.go
andresmgot: right, the repository name will be part of the URL path so it's better to use PathEscape

absoludity:
> 
> When using an OCI repository, there two options:
> 
>     1. Use the registry URL with the repo and the app name without the repository (e.g. `registry: docker.io/bitnami, app: wordpress`).

Where is this first format being used? Seems odd to use `registry: docker.io/bitnami` when that is by definition not a registry aiui? (the server which handles the api requests is `docker.io` - `bitnami/wordpress` identifies an image

andresmgot:
> Where is this first format being used? Seems odd to use `registry: docker.io/bitnami` when that is by definition not a registry aiui? (the server which handles the api requests is `docker.io` - `bitnami/wordpress` identifies an image

Yep, but we call it "Repository URL" in the form (or just `url` internally). For example, in my tests with harbor, I was doing that `url: harbor/my-project`, `app: apache` rather than `url: harbor`, `app: my-project/apache`. In any case, both are forms are valid now.

PR #2419, opened 2021-02-18
Use command for pinniped-proxy to overwrite entrypoint.

absoludity:
### Description of the change

The devel Dockerfile doesn't specify an entrypoint (similar to all Kubeapps devel Dockerfiles) so our deployment worked fine, but since the bitnami Dockerfile for pinniped-proxy *does* define an ENTRYPOINT, our deployment effectively tried to do `./pinniped-proxy pinniped-proxy`.

This change ensures that our deployment specifies the command (which is therefore used instead of the entrypoint).

### Benefits

Our deployment works with the bitnami pinniped-proxy image.

### Possible drawbacks

None.


antgamdia (approved):
Great, thank you for checking it!

andresmgot (approved):

PR #2435, opened 2021-02-19
Replace node-sass-chokidar by sass (dart)

antgamdia:
### Description of the change

This PR simply updates the sass preprocessor implementation since the current one is marked [as deprecated](https://github.com/michaelwayman/node-sass-chokidar):

```
This project had a good run but it looks like https://github.com/sass/dart-sass has now become the better choice. 
I Don't have the time to maintain this project and from a quick glance dart-sass is also using chokidar for fs normalization.
LibSass is Deprecated since October and dart-sass is the new official recommendation. Cheers folks **;)
```

### Benefits

The main reason for updating it (apart from it is deprecated) is that this node-chokidar has some issues [performing imports](https://github.com/michaelwayman/node-sass-chokidar/issues/39).

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

N/A


dashboard/package.json
```diff
@@ -141,7 +142,6 @@
     "lint-staged": "^10.2.11",
     "mock-socket": "^9.0.3",
     "moxios": "^0.4.0",
-    "node-sass-chokidar": "^1.3.3",
```
andresmgot, line 144: you need to add the `sass` devdep, no?

antgamdia: Absolutely right. I added the `--dev` flag when doing the clarity stuff, but totally forgot about it when extracted the change for this PR :S
Thank you!!

andresmgot (approved):

PR #2436, opened 2021-02-19
Store OAuth2 sensitive configuration data in a secret

### LINKED ISSUE #2275: auth-proxy should use secrets for OIDC clientId, clientSecret and cookieSecret
### Description:

The current deployment of oauth2-proxy adds sensitive information to the parameter list of oauth2-proxy.
From `chart/kubeapps/templates/kubeapps-frontend-deployment.yaml´: 
```
- --client-id={{ required "You must fill \".Values.authProxy.clientID\" with the Client ID of the provider" .Values.authProxy.clientID }}
- --client-secret={{ required "You must fill \".Values.authProxy.clientSecret\" with the Client Secret of the provider" .Values.authProxy.clientSecret }}
- --cookie-secret={{ required "You must fill \".Values.authProxy.cookieSecret\" with a 16, 24 or 32 byte base64 encoded seed string for secure cookies" .Values.authProxy.cookieSecret }}       
```

This exposes the data in the process list. It maybe better to put those data in a secret or environment variable as shown nginx-ingress external OAUTH authentication example or kubeprod:
- https://kubernetes.github.io/ingress-nginx/examples/auth/oauth-external-auth/
- https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/docs/examples/auth/oauth-external-auth/oauth2-proxy.yaml
- https://github.com/bitnami/kube-prod-runtime/blob/master/manifests/components/oauth2-proxy.jsonnet

1. Define in helm chart values:
```
authProxy:
  enabled: true
  skipKubeappsLoginPage: true
  provider: "oidc"
  clientID: "myId"
  clientSecret: "myClientSecret"
  cookieSecret: "myCookieSecret"
```

Sensitive data in the process list.

No sensitive data in the process list.

Permanent.

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.3.4", GitCommit:"a61ce5633af99708171414353ed49547cf05013d", GitTreeState:"clean", GoVersion:"go1.14.9"}
```

```
kubeapps        kubeapps        6               2021-01-22 12:27:29.627126572 +0100 CET deployed        kubeapps-5.0.1  2.0.1
```

```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.7", GitCommit:"1dd5338295409edcfff11505e7bb246f0d325d15", GitTreeState:"clean", BuildDate:"2021-01-13T13:23:52Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.7", GitCommit:"1dd5338295409edcfff11505e7bb246f0d325d15", GitTreeState:"clean", BuildDate:"2021-01-13T13:15:20Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}
```

andresmgot:
### Description of the change

Store OAuth2 sensitive configuration data in a secret

### Applicable issues

  - fixes #2275

### Additional information

Tested manually.


chart/kubeapps/templates/_helpers.tpl
```diff
@@ -175,6 +175,13 @@ Create name for the secrets related to an app repository
 apprepo-{{ .name }}-secrets
 {{- end -}}
 
+{{/*
+Create name for the secrets related to oauth2_proxy
+*/}}
+{{- define "kubeapps.oauth2_proxy-secret.name" -}}
+{{ template "kubeapps.fullname" . }}-oauth2
```
antgamdia, line 182: General comment here: we are limited to 53-characters-names  (if I remember correctly). Since we rely on `kubeapps.fullname` we left a variable number of characters available. I mean, the basic things (i.e., deployments, etc) can fit in this 53-char limit but other k8s resource might not.

I think we should guarantee a safe number of characters, so that the users can be sure in advance their name won't silently break anything. 

(but this name is ok, it was just speaking my mind here :P)

chart/kubeapps/templates/oauth2-secret.yaml
```diff
@@ -0,0 +1,15 @@
...
+metadata:
+  name: {{ template "kubeapps.oauth2_proxy-secret.name" . }}
+  {{- if .namespace }}
+  namespace: {{ .namespace }}
+  {{- end }}
+  labels:{{ include "kubeapps.labels" $ | nindent 4 }}
+data:
+  clientID: {{ required "You must fill \".Values.authProxy.clientID\" with the Client ID of the provider" .Values.authProxy.clientID | b64enc }}
+  clientSecret: {{ required "You must fill \".Values.authProxy.clientSecret\" with the Client Secret of the provider" .Values.authProxy.clientSecret | b64enc }}
+  cookieSecret: {{ required "You must fill \".Values.authProxy.cookieSecret\" with a 16, 24 or 32 byte base64 encoded seed string for secure cookies" .Values.authProxy.cookieSecret | b64enc }}
```
antgamdia, line 13: Even if this message is clear, I'd add some example on how to generate this seed, such as `echo "not-good-secret" | base64`

antgamdia (approved):
Great! Thank you for addressing this issue.

chart/kubeapps/templates/_helpers.tpl
andresmgot: > so that the users can be sure in advance their name won't silently break anything.

I won't fail silently, the K8s API fails when trying to install stuff with invalid names:

```
▶ helm install this-is-a-very-long-time-to-ensure ./chart/kubeapps       
Error: ServiceAccount "this-is-a-very-long-time-to-ensure-kubeapps-internal-apprepository-job-postupgrade" is invalid: metadata.labels: Invalid value: "this-is-a-very-long-time-to-ensure-kubeapps-internal-apprepository-controller": must be no more than 63 characters
```

chart/kubeapps/templates/oauth2-secret.yaml
andresmgot: :+1: that's documented in a values.yaml comment but we can add that here too

chart/kubeapps/templates/_helpers.tpl
antgamdia: Yes, the 'helm install` command would complain if we set a very long name. However,  there is a "grey side" here: it is not the install that is failing but the postinstall or other helm hooks. When I face this error again, I'll open an issue and discuss it there

andresmgot: ? I don't fully understand what you mean. If the `postinstall` fails is true that you need to uninstall the chart first before being able to try again.

antgamdia: I didn't explain well, sorry! I'm not being able to reproduce it right now, but the general idea is, more or less, what this issue describes (https://github.com/kubernetes/kubernetes/issues/50850).
Anyway, don't worry, if I could replicate it again, I'll let you know.

chart/kubeapps/templates/oauth2-secret.yaml
absoludity: This have actually improved since that string was created. I don't think auth_proxy requires a specifically sized byte-string any more. Though they still use 16 in their own example. Might be good to point to the docs instead: https://oauth2-proxy.github.io/oauth2-proxy/docs/configuration/overview/#generating-a-cookie-secret

chart/kubeapps/templates/oauth2-secret.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if .Values.authProxy.enabled }}
```
absoludity, line 1: This flag means that the configuration will assume *an* auth proxy is enabled, but not necessarily Kubeapps' managed oauth2-proxy. The help string for `enabled` tries to be clear about the limit of this flag on its own:

```
  ## Set to true if Kubeapps should configure the OAuth login/logout URIs defined below.
```

So I think you want the same condition which is used in the deployment to include the oauth2-proxy:

```
        {{- if and .Values.authProxy.enabled (not .Values.authProxy.external) }}
```

absoludity (approved):
Thanks!

chart/kubeapps/templates/oauth2-secret.yaml
andresmgot: good point, I missed that.

chart/kubeapps/templates/oauth2-secret.yaml
andresmgot: :+1: (actually `not-good-secret` length is 15).

absoludity: It'd be great to also update the example in the values.yaml to just point to the same link.

Issue #2372, opened 2021-02-10
Bump up Clarity version (v4->v5)? Analyze benefits


danoo:

antgamdia:
Some thoughts here:

- AFAIK, the majority of our styles rely on `clr/ui` because it is like the static css style. Other components do rely con `cds/core`. So, for having the whole appearance, we need both styles.
  - Problem: `clr/ui` is not actively maintained now and there are missing features, such as the native support for themes. We have to conditionally load the light/dark mode.
    - In my PoC it is implemented with media queries to the browser native config, but a selector/toggle/whatever is also feasible.
 - Migrating v4->v5 is not that difficult, it is just a change in the priority of some selectors (eg. thenumber of pods deployed was misplaced)
 - There are a number of hardcoded css values (that is, not depending upon sass variables). Some effort on replacing these references by the ones from the clarity docs is required.
   - Once we have it, it becomes [really easy to change the color schema ](https://clarity.design/foundation/themes/usage/#custom-theme)without dealing with low-level css selectors.


There is an ongoing work in #2371 , but nothing PR-ready yet.

absoludity:
> Some thoughts here:

Thanks. I don't have much context for the clarity integration as Andres worked on that, so a couple of questions:
 
>     * AFAIK, the majority of our styles rely on `clr/ui` because it is like the static css style. Other components do rely con `cds/core`. So, for having the whole appearance, we need both styles.

But should we be removing use of `clr/ui`? From the [website](https://clarity.design/get-started/developing/angular/) it looks like `clr/ui` is part of the angular-specific clarity? Or is there no issue to continue to use these?

>       * Problem: `clr/ui` is not actively maintained now and there are missing features, such as the native support for themes. We have to conditionally load the light/dark mode.

So if we were not using `clr/ui` the theming wouldn't require the conditional loading, but the impact isn't big? (ie. if this is a one-line change which does the conditional load for all our components that use `clr/ui`, great, but if not, perhaps it's more reason to invest in moving them to `cds/core`, unless there's other reasons not to?

>         
>         * In my PoC it is implemented with media queries to the browser native config, but a selector/toggle/whatever is also feasible.

[...]

PR #2442, opened 2021-02-22
Prepare css variables to use cds instead of clr

antgamdia:
### Description of the change

This PR changes some css variables from `clr` to `cds`. It also fixes some inconsistencies in the fallback used colors.

### Benefits

This PR paves the way for the upgrading process from clarity v4 to v5. 

### Possible drawbacks

### Applicable issues

  - related #2372

### Additional information

  - Update modals
  - Update layout


dashboard/src/index.scss
```diff
@@ -1,38 +1,44 @@
...
-  min-width: 0;
-}
-
 svg.icon {
   /* Fix positioning */
   position: relative;
   bottom: -0.125em;
 }
 
-/* Override Hex underline text-decoration in buttons in alerts
```
antgamdia, line 18: Unused css selectors

dashboard/src/index.scss
```diff
@@ -1,38 +1,44 @@
...
+  border: 1px solid var(--cds-alias-object-border-color, #ccc);
+  background-color: var(--cds-alias-object-interaction-background), #fff;
   overflow: auto;
   border-radius: 4px;
   outline: none;
   padding: 20px;
   margin-right: -50%;
   transform: translate(-50%, -50%);
+
+  // Copied from the CdsButton style
```
antgamdia, line 27: this is a workaround for an unlikely modal to match the rest of the styles:
image

dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/BasicDeploymentForm.scss
```diff
@@ -43,7 +43,7 @@
 }
 
 .param-separator {
-  border: 1px solid #f1f1f1;
+  border: 1px solid var(#f1f1f1);
```
antgamdia, line 46: Ups, forgot about this one

dashboard/src/components/Header/Menu.scss
```diff
@@ -105,8 +105,14 @@
...
-    color: #fff;
+  a {
+    color: var(--cds-global-color-blue-400, #2ec0ff) !important;
+  }
+  &:visited {
+    color: var(--cds-global-color-lavender-300 #8b8dda) !important;
+  }
+  &:hover {
+    color: var(--cds-global-color-blue-500, #fff) !important;
   }
```
andresmgot, line 116: why this changes? (we should avoid using !important if possible).

dashboard/src/index.scss
andresmgot: what's the behavior without this `.button`?

andresmgot (approved):
thanks for reviewing this!

dashboard/src/index.scss
antgamdia: Something like: 
image

dashboard/src/components/Header/Menu.scss
antgamdia: The `a:link` in the clr-ui css is taking more priority than the css style defined at Menu.scss, so, if not adding the `!important` the bad color won't get overidden.
Pehaps I'm missing something here :S

image

andresmgot: you have the answer :) if you add the `:link` modifier to the line `108`, you don't need the `!important` rule

dashboard/src/index.scss
andresmgot: hum, weird, we should be missing something here but I am not sure what

dashboard/src/components/Header/Menu.scss
antgamdia: Thanks, after some changes we can get rid of the `!important` modifiers :)

dashboard/src/index.scss
antgamdia: image

I've just updated this modal to use the cds button as other components do

PR #2443, opened 2021-02-22
Bump up Clarity to the latest version

antgamdia:
_Depends on #2442_ 

### Description of the change

This PR bumps the clarity versions up, changes every clr->cds import, add some css imports and minor fixes.

### Benefits

This PR paves the way for the upgrading process from clarity v4 to v5. 

### Possible drawbacks

### Applicable issues

  - related #2372

### Additional information

  - Update modals
  - Update layout


dashboard/src/components/Layout/Clarity.tsx
```diff
@@ -29,11 +29,11 @@ import {
...
   trashIcon,
   uploadCloudIcon,
-} from "@clr/core/icon";
-import "@clr/core/icon/register.js";
+} from "@cds/core/icon";
+import "@cds/core/icon/register.js";
 import * as React from "react";
 
-import "@clr/ui/clr-ui.min.css";
+import "@clr/ui/clr-ui.min.css"; // light clarity UI theme
```
andresmgot, line 36: shouldn't we use just one place to import clarity css files? (either here or in the index file, that btw is failing to import the files).

andresmgot (approved):
what's the difference between `@clr` and `@cds` packages? (they seem like mirrors but I guess they are different).

dashboard/src/components/Layout/Clarity.tsx
antgamdia: Yep, this is exactly what we are doing in https://github.com/kubeapps/kubeapps/pull/2450
I just left unchanged this file (apart from clarifying it is the light theme)

antgamdia:
> what's the difference between `@clr` and `@cds` packages? (they seem like mirrors but I guess they are different).

From the [website](https://github.com/vmware/clarity/tree/next): 

`@clr/icons` Contains the custom element icons.
`@clr/ui` Contains the static styles for building HTML components.
`@clr/angular` Contains the Angular components. This package depends on @clr/ui for styles.
`@cds/core` Contains the Web Components that work in any JavaScript framework.

There is also `@cds/react`, which is the official wrapper of these components.

So, we are using all except for the angular one.

dashboard/src/components/AppView/AccessURLTable/__snapshots__/AccessURLTable.test.tsx.snap
```diff
@@ -371,7 +371,20 @@ exports[`when the app contains ingresses should show the table with available in
                                 size="sm"
                                 solid={true}
                               >
-                                <cds-icon />
```
andresmgot, line 374: can you delete the screenshot instead :) ?

antgamdia:
FYI @andresmgot , I had to replace the cds button (the one for adding docker credentials when adding a new repo) by a normal button after hitting the same issue you face in the past. The form was being prematurely submitted. I guess we should replace the current `clr`form with the entire `cds` one, but dunno :s (out of the scope of this PR though)

andresmgot:
> FYI @andresmgot , I had to replace the cds button (the one for adding docker credentials when adding a new repo) by a normal button after hitting the same issue you face in the past. The form was being prematurely submitted. I guess we should replace the current `clr`form with the entire `cds` one, but dunno :s (out of the scope of this PR though)

Weird, maybe the `@clr` package had the fix but the `cds` one not? Anyway, I guess this is fine until we investigate the issue

Issue #2371, opened 2021-02-10
Dark mode


danoo:

antgamdia:
See Clarity theming options: https://clarity.design/foundation/themes/usage/#custom-theme

antgamdia:
Some updates here :) 

![image](https://user-images.githubusercontent.com/11535726/108560521-22bc7300-72fd-11eb-8cb0-0daee16f7871.png)

![image](https://user-images.githubusercontent.com/11535726/108560616-50092100-72fd-11eb-9998-4faa49573439.png)

![image](https://user-images.githubusercontent.com/11535726/108560761-88a8fa80-72fd-11eb-84d7-238ce4043b8f.png)

![image](https://user-images.githubusercontent.com/11535726/108561017-d6256780-72fd-11eb-842a-21a6aa944968.png)

absoludity:
Woah - looks great!

andresmgot:
:rocket: cool!

Minor thing regarding icons: we should use a white-version of the icons we currently have for getting the right contrast, for example, the helm icon here:

![Screenshot from 2021-02-22 10-48-54](https://user-images.githubusercontent.com/4025665/108691316-a8a60d00-74fb-11eb-8e76-c20502bc57c2.png)

antgamdia:
>  cool!
> 
> Minor thing regarding icons: we should use a white-version of the icons we currently have for getting the right contrast, for example, the helm icon here:
> 
> ![Screenshot from 2021-02-22 10-48-54](https://user-images.githubusercontent.com/4025665/108691316-a8a60d00-74fb-11eb-8e76-c20502bc57c2.png)

I did some trick in the 404 image (since it was not that important), a css filter inverting colors.
Here we should change, indeed, the image. I have to look into how the images are being loaded: just css? or do we control them in react?
I'll investigate it and will send a PR soon :)
Thanks!

antgamdia:
Currently, this is the implemented way to decide which dark/light theme should be loaded based on the user browser's preferences.

![darkmode](https://user-images.githubusercontent.com/11535726/108755526-ac5f8100-7547-11eb-9048-3120e6c2f2d5.gif)

antgamdia:
After several PRs, Kubeapps finally has a dark mode in master and it's ready to be released in the upcoming version! 

:sun_with_face: -->  :new_moon_with_face:

[...]

PR #2444, opened 2021-02-22
Detect dark mode and load the proper cds theme

antgamdia:
_Depends on #2443_ 

### Description of the change

This PR changes simply supports dark mode **in cds** (not clr). As a result, very few cds-styled components will get the dark theme when the user's browser is configured in dark mode.

### Benefits

This PR paves the way for the upgrading process from clarity v4 to v5 and dark mode.

### Possible drawbacks

### Applicable issues

  - related #2372
  - related #2371 

### Additional information

  - Update modals
  - Update layout


antgamdia (commented):
It's a temporary change (superseded by #2444), so we can merge it or even close, as you may prefer

andresmgot (approved):
Can we add a test for this?

dashboard/src/containers/Root.tsx
```diff
@@ -18,9 +18,16 @@ async function initLocale() {
 
 function Root() {
   const [i18nConfig, setI18nConfig] = useState(I18n.getDefaultConfig());
+  // Currently we just support 'dark' and 'light' themes
+  // TODO(agamez): add manual override for this setting
+  const theme =
+    window.matchMedia && window.matchMedia("(prefers-color-scheme: dark)").matches
+      ? "dark"
+      : "light";
   useEffect(() => {
```
andresmgot, line 27: also, better to use a different `useEffect` to set this. I believe there is no relation between `initLocale` and setting the cds-theme property.

antgamdia:
These changes are being superseded by https://github.com/kubeapps/kubeapps/pull/2450 (taking into account this PR's comments), so let's close it.

Issue #2169, opened 2020-11-17
Bump CRD from v1beta1 to v1 (k8s 1.16+)


antgamdia:
### Description:

In k8s 1.16+ a deprecation warning of `apiextensions.k8s.io/v1beta1` in favor of `apiextensions.k8s.io/v1` appeared. In 1.22+ it will be fully deprecated.

Note that this change will drop support for <=1.15 versions.

### Steps to reproduce the issue:

Performing `helm install kubeapps` in a 1.16+ cluster

### Describe the results you received:

A deprecation warning is being thrown.

### Describe the results you expected:

No deprecation warning.

### Additional information you deem important (e.g. issue happens only occasionally):

If we bump up the version, k8s clusters prior to 1.15 won't be able to install Kubeapps.
Besides, 1.22+ users aren't able to install it.

### Version of Helm, Kubeapps and Kubernetes:

The warning is being thrown in k8s versions 1.16, 1.17, 1.18, 1.19 (and future 1.20, 1.21).

antgamdia:
Just for the record, this issue is a follow up of the PRs #2164 and #2168.

Also, note that simply change `v1beta1` -> `v1` won't work, as it will throw the following error: (thanks @andresmgot for pointing this out!)

```
Error: failed to install CRD crds/apprepository-crd.yaml: CustomResourceDefinition.apiextensions.k8s.io "apprepositories.kubeapps.com" is invalid: 
[spec.versions: Invalid value: []apiextensions.CustomResourceDefinitionVersion(nil): must have exactly one version marked as storage version, status.storedVersions: Invalid value: []string(nil): must have at least one stored version]
```

https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/

antgamdia:
![image](https://user-images.githubusercontent.com/11535726/99956029-9d6d0500-2d85-11eb-95c4-b336f0679227.png)

Regardless of this lifecycle, we currently plan to support the same versions as GKE: https://cloud.google.com/kubernetes-engine/docs/release-notes#stable-channel

antgamdia:
Just for the record, Ingress objects should also be bumped up from: `extensions/v1beta1` to `networking.k8s.io/v1`. This change is available from 1.14, so there should not be any problem in migrating.
More info at https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/

Note that there is a change in how ports are defined:

```
serviceName: kubeapps 
[...]

PR #2445, opened 2021-02-22
[chart] Use bitnami/common function + provide compatibilty with different K8s versions

juan131:
Signed-off-by: juan131 <juanariza@vmware.com>

### Description of the change

This PR adapts the Kubeapps chart to use the `bitnami/common` library so several common macros are reused. This is important to standardize this chart with the rest of the Bitnami Catalog.

On the other hand, it ensures the `apiVersion` of several objects (e.g. Ingress, RBAC, CRDs, etc.) is compatible with different K8s versions.

### Benefits

- Standardization
- Compatibility with multiple K8s versions.

### Possible drawbacks

None

### Applicable issues

N/A

### Additional information

None

### DIFF
--- a/chart/kubeapps/Chart.yaml
+++ b/chart/kubeapps/Chart.yaml
@@ -3,6 +3,11 @@ annotations:
 apiVersion: v2
 appVersion: DEVEL
 dependencies:
+  - name: common
+    repository: https://charts.bitnami.com/bitnami
+    tags:
+      - bitnami-common
+    version: 1.x.x
   - name: postgresql
     repository: https://charts.bitnami.com/bitnami
     version: '10.X.X'
@@ -20,4 +25,4 @@ maintainers:
 name: kubeapps
 sources:
   - https://github.com/kubeapps/kubeapps
-version: 5.2.2
+version: 5.3.0
--- a/chart/kubeapps/crds/apprepository-crd.yaml
+++ b/chart/kubeapps/crds/apprepository-crd.yaml
@@ -1,4 +1,4 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: {{ include "common.capabilities.crd.apiVersion" . }}
 kind: CustomResourceDefinition
 metadata:
   name: apprepositories.kubeapps.com
@@ -10,4 +10,9 @@ spec:
     plural: apprepositories
     shortNames:
       - apprepos
+  {{- if semverCompare "<1.19-0" (include "common.capabilities.kubeVersion" .) -}}
   version: v1alpha1
+  {{- else }}
+  versions:
+    - name: v1alpha1
+  {{- end }}
--- a/chart/kubeapps/crds/apprepository-crd.yaml
+++ b/chart/kubeapps/crds/apprepository-crd.yaml
@@ -1,4 +1,4 @@
-apiVersion: {{ include "common.capabilities.crd.apiVersion" . }}
+apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   name: apprepositories.kubeapps.com
@@ -10,9 +10,4 @@ spec:
     plural: apprepositories
     shortNames:
       - apprepos
-  {{- if semverCompare "<1.19-0" (include "common.capabilities.kubeVersion" .) -}}
   version: v1alpha1
-  {{- else }}
-  versions:
-    - name: v1alpha1
-  {{- end }}
--- a/chart/kubeapps/templates/NOTES.txt
+++ b/chart/kubeapps/templates/NOTES.txt
@@ -6,7 +6,7 @@ Tip:
 
 Kubeapps can be accessed via port {{ .Values.frontend.service.port }} on the following DNS name from within your cluster:
 
-   {{ template "kubeapps.fullname" . }}.{{ .Release.Namespace }}.svc.cluster.local
+   {{ template "common.names.fullname" . }}.{{ .Release.Namespace }}.svc.cluster.local
 
 {{- $reposWithOrphanSecrets := include "kubeapps.repos-with-orphan-secrets" . }}
 {{- if ne $reposWithOrphanSecrets "" }}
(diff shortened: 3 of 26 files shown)


chart/kubeapps/crds/apprepository-crd.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: {{ include "common.capabilities.crd.apiVersion" . }}
```
absoludity, line 1: Hrm. I thought we (bitnami) were trying to move away from helm functionality which moves away from `constant-input  + chart = constant-output`? That is, using `.Capabilities` (which this does under the hood), or `lookup` functions (not used here, just as example, though I see it is used in the `_secrets.tpl`) means that the output is a function of not only the input and chart, but also the current cluster context.

Not blocking on this as I assume the decision's been made already. We do already have one use of `.Capabilities` but were trying not to add more.

absoludity (commented):
Thanks @juan131 for all the work updating that - and removing the no-longer needed kubeapps helpers!
I'm a little worried that we're (generally) adding more helm specific functionality that moves away from the resulting templates a function of the input and chart only (ie. relying on current k8s environment), but that's not really relevant for this PR.

I'll hold off approving until both @andresmgot 's #2436 is landed (has the same chart version) and the e2e test is passing. It's currently failing with an error that looks like it could be related. I've just re-run the failed test to be sure, but haven't looked more closely.

chart/kubeapps/templates/NOTES.txt
```diff
@@ -6,7 +6,7 @@ Tip:
 
 Kubeapps can be accessed via port {{ .Values.frontend.service.port }} on the following DNS name from within your cluster:
 
-   {{ template "kubeapps.fullname" . }}.{{ .Release.Namespace }}.svc.cluster.local
+   {{ template "common.names.fullname" . }}.{{ .Release.Namespace }}.svc.cluster.local
```
antgamdia, line 9: So, from now on, we shouldn't be worried about the length of the name anymore since this function is automatically truncating it, isn't it?

chart/kubeapps/templates/ingress.yaml
```diff
@@ -32,7 +28,7 @@ spec:
         paths:
         - path: {{ default "/" .path }}
           backend:
-            serviceName: {{ template "kubeapps.fullname" $ }}
+            serviceName: {{ template "common.names.fullname" $ }}
             servicePort: http
```
antgamdia, line 32: Not a chart expert at all. However, I think there are more changes here if we use `networking.k8s.io/v1` instead of `extensions/v1beta1`: https://github.com/kubeapps/kubeapps/issues/2169#issuecomment-733556537

chart/kubeapps/crds/apprepository-crd.yaml
```diff
@@ -10,4 +10,9 @@ spec:
     plural: apprepositories
     shortNames:
       - apprepos
+  {{- if semverCompare "<1.19-0" (include "common.capabilities.kubeVersion" .) -}}
```
antgamdia, line 13: I guess here we are solving this: https://github.com/kubeapps/kubeapps/issues/2169#issuecomment-728910010
No?

chart/kubeapps/Chart.yaml
```diff
@@ -20,4 +25,4 @@ maintainers:
 name: kubeapps
 sources:
   - https://github.com/kubeapps/kubeapps
-version: 5.2.0
+version: 5.2.1
```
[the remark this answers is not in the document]
andresmgot: Actually I am going to release 5.2.2 soon so this should be 5.2.3

antgamdia (commented):
I agree with Michael's comments. We are relying too much on the helm and server context. It may harden the migration from Help to other packaging formats. 
However, from a Bitnami chart perspective, it's a great change since API versions will automagically adapt to the proper ones.

Thanks for the effort and sending this PR :)

chart/kubeapps/crds/apprepository-crd.yaml
andresmgot: Not sure about this, but I guess the goal is to define `.Values.global.kubeVersion` so `.Capabilities` is not called. Can you confirm @juan131 ?

chart/kubeapps/templates/apprepository-rbac.yaml
```diff
@@ -91,8 +91,7 @@ rules:
 # The Kubeapps app repository controller can read and watch its own
 # AppRepository resources cluster-wide. The read and write cluster-roles can
 # also be bound to users in specific namespaces as required.
-apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRole
```
andresmgot, line 95: have you removed the `kind` by accident?

andresmgot: this seems why the CI is failing:

```
Error: UPGRADE FAILED: error validating "": error validating data: kind not set
```

juan131: Correct. I'll fix it

chart/kubeapps/Chart.yaml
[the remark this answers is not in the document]
juan131: No worries. I'll rebase and update it.

chart/kubeapps/crds/apprepository-crd.yaml
juan131: Correct. Let me added the missing values and a couple of refactorizations related to `bitnami/common` more

chart/kubeapps/templates/NOTES.txt
juan131: Yup, it works pretty similar to what you had before.

chart/kubeapps/templates/ingress.yaml
juan131: Yes, let me the rest of refactorization for ingress.

chart/kubeapps/crds/apprepository-crd.yaml
juan131: I guess, but it'd be nice if you give that a try

juan131: Reverting these changes, since only pure YAML (no templates) is accepted for CRDs

chart/kubeapps/templates/tls-secrets.yaml
```diff
@@ -1,13 +1,32 @@
...
+metadata:
+  name: {{ printf "%s-tls" .Values.ingress.hostname }}
+  namespace: {{ .Release.Namespace }}
+  labels: {{ include "kubeapps.labels" . | nindent 4 }}
+type: kubernetes.io/tls
+data:
+  tls.crt: {{ $cert.Cert | b64enc | quote }}
+  tls.key: {{ $cert.Key | b64enc | quote }}
+  ca.crt: {{ $ca.Cert | b64enc | quote }}
 {{- end }}
```
andresmgot, line 31: I think you have an extra `end` here, I only see one `if` opening here

andresmgot (commented):
Can you take a look to the comment?

This PR is too big to be reviewed at once. Can you avoid the version bump? At least we would need to do some manual testing before releasing the version.

chart/kubeapps/templates/tls-secrets.yaml
juan131: You're missing the `range` that's also finished with an `end`. This `end` is necessary.

juan131:
> This PR is too big to be reviewed at once. Can you avoid the version bump? At least we would need to do some manual testing before releasing the version.

Take your time to test it, and ping me every time you consider it necessary. I do consider the version bump necessary since new parameters are included in the chart (to name a few: `ingress.path`, `ingress.pathType`,  `kubeVersion`, etc.`)

chart/kubeapps/templates/tls-secrets.yaml
andresmgot: ah, got it, yes (I miss indentation at least)

andresmgot:
> @andresmgot take your time to test it, and ping me every time you consider it necessary. I do consider the version bump necessary since new parameters are included in the chart (to name a few: `ingress.path`, `ingress.pathType`, `kubeVersion`, etc.)

Yes, I didn't mean that the version bump was wrong but I was thinking on testing master. In any case it's not necessary, so yes, I will give this a try tomorrow :+1: Thanks!

andresmgot (approved):
Sorry for the very late response!

I have tested this a little bit, setting different ingress values and setting the `kubeVersion`. Everything works as expected. Thanks!

PR #2446, opened 2021-02-22
Fix behavior for multiple chart versions for OCI registries

andresmgot:
### Description of the change

I introduced a bug after https://github.com/kubeapps/kubeapps/pull/2373

The variable for the chart versions was scoped to goroutine so only the first one was stored (causing the app to just store the first version received of the chart).

### Additional information

Will need to release a patch version after this :/

### DIFF
--- a/cmd/asset-syncer/utils.go
+++ b/cmd/asset-syncer/utils.go
@@ -550,7 +550,7 @@ func chartImportWorker(repoURL *url.URL, r *OCIRegistry, chartJobs <-chan pullCh
 
 // Charts retrieve the list of charts exposed in the repo
 func (r *OCIRegistry) Charts() ([]models.Chart, error) {
-	result := map[string]models.Chart{}
+	result := map[string]*models.Chart{}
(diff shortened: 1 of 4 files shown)


absoludity:
> The variable for the chart versions was scoped to goroutine so only the first one was stored (causing the app to just store the first version received of the chart).

I couldn't make sense of the above reason in the context - and lost some time trying to see how that was relevant, until I realised it's not (afaict). The issue is simply that the struct value is copied into your temporary var `r`, and we were updating that temporary variable's copy rather than the struct in the map itself.

Here's a simpler example showing the same issue you had: https://play.golang.org/p/WuNjz-4eNzx

Anyway, agree about the solution (since you [can't assign directly to `result[ch.ID].ChartVersions`](https://stackoverflow.com/a/32751792))

absoludity (approved):

andresmgot:
> The issue is simply that the struct value is copied into your temporary var `r`, and we were updating that temporary variable's copy rather than the struct in the map itself.

Yes sorry, that's what I meant that in the goroutine the variable was scoped to that.

PR #2447, opened 2021-02-22
Update CdsModal to cds v5

antgamdia:
_Depends on #2443_ 

### Description of the change

This PR changes the new namespace modal to use the `react-modals` one. This is the modal component being used in other places. However, the reason to update it is just I wasn't able to get it to work after the upgrading (the show/hide based upon a useState variable feature wasn't working).

It also solves an unnoticed problem: the close buttons of the current modals wasn't being rendered. It led us to add another icon even if the modal was trying to show one (but the icon wasn't loaded)

image

### Benefits

This PR paves the way for the upgrading process from clarity v4 to v5 and dark mode.

### Possible drawbacks

### Applicable issues

  - related #2372 

### Additional information

  - Update modals
  - Update layout


dashboard/src/components/Layout/Clarity.tsx
```diff
@@ -37,35 +39,36 @@ import "@clr/ui/clr-ui.min.css"; // light clarity UI theme
 
 Icons.addIcons(
```
antgamdia, line 40: Just sorting alphabetically the icons and adding the missing `windowCloseIcon` one.

dashboard/src/components/js/Modal/Modal.js
```diff
@@ -46,13 +46,13 @@ const Modal = ({
...
               >
-                <clr-icon aria-hidden="true" shape="close" />
+                <cds-icon aria-hidden="true" shape="window-close" />
               </button>
             )}
             <h3 className="modal-title">{title}</h3>
           </div>
           <div className="modal-body">{children}</div>
-          <div className="modal-footer">{footer}</div>
+          {footer && <div className="modal-footer">{footer}</div>}
```
antgamdia, line 55: Hide footer to prevent an extra space from being rendered when no footer is passed

dashboard/src/components/Config/AppRepoList/AppRepoButton.scss
```diff
@@ -1,6 +0,0 @@
-.modal-close {
```
antgamdia, line 1: Delete the modal close handcrafted style since our modal already has one (but the icon wasn't being loaded)

dashboard/src/components/Header/ContextSelector.test.tsx
```diff
@@ -147,8 +147,14 @@ it("shows an error creating a namespace", () => {
   clusters.clusters[clusters.currentCluster].error = { error: new Error("Boom"), action: "create" };
 
   const wrapper = mountWrapper(getStore({ clusters }), <ContextSelector />);
+  const modalButton = wrapper.find(".flat-btn").first();
+  act(() => {
+    (modalButton.prop("onClick") as any)();
+  });
+  wrapper.update();
+
```
antgamdia, line 155: Just for making the modal appear

dashboard/src/components/Header/ContextSelector.tsx
```diff
@@ -167,23 +165,40 @@ function ContextSelector() {
...
+                      <div className="clr-control-container">
+                        <div className="clr-input-wrapper">
+                          <input
+                            type="text"
+                            className="clr-input"
+                            placeholder="my-namespace"
+                            onChange={onChangeNewNS}
+                            required={true}
+                            pattern="[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*"
+                            title="Use lower case alphanumeric characters, '-' or '.'"
```
antgamdia, line 189: I've added this validation (as in other forms), but let me know if I'm wrong here.

andresmgot:
> This PR changes the new namespace modal to use the `react-modals` one. This is the modal component being used in other places. However, the reason to update it is just I wasn't able to get it to work after the upgrading (the show/hide based upon a useState variable feature wasn't working).

mmh, this is going in the wrong direction. I know we are still using `react-modals` in other places but we should use the clarity core one instead. If it's not working, we should verify why is that. Probably we are missing something (and if not, we should notify clarity team).

> It also solves an unnoticed problem: the close buttons of the current modals wasn't being rendered. It led us to add another icon even if the modal was trying to show one (but the icon wasn't loaded)

This was confusing for me in the past. I know there is a property to enable/disable that button, have you seen that?

dashboard/src/components/Header/ContextSelector.tsx
andresmgot: I would not add the pattern validation, we may miss valid cases. I'd let the K8s API to do the validation for us.

antgamdia:
> we should notify clarity team

I've talked (internally) to the Clarity people and they're looking into the issue. Let's see what happens. Demo env here: https://stackblitz.com/edit/clarityv5-react-cdsmodal?file=index.js

Anyway, since the change is simple, I'd recommend accepting the PR and then, once we got a response, decide whether we want to use the cdsModal. If not, note that, currently, in the master branch the modal is broken and cannot be open.

antgamdia:
Shall I change, also, the rest of the modals to use the `cds` ones?

dashboard/src/components/js/Modal/Modal.js
andresmgot: you can remove these changes now?

andresmgot (approved):
Thanks for checking!

andresmgot:
> Shall I change, also, the rest of the modals to use the `cds` ones?

That'd be great :)

dashboard/src/components/js/Modal/Modal.js
antgamdia: Why? Do you mean removing all the changes related to the old modals? Or just this one?

andresmgot: I mean the changes related to this file Modal.js, you are not using it in the PR, no?

PR #2449, opened 2021-02-22
Fix for Operators without owned CRDs

andresmgot:
### Description of the change

If an Operator doesn't contain CRDs, the dashboard fails with: `cannot read owned of undefined`. Adapted the typescript definition to avoid this issue.


antgamdia (approved):
Thanks! I thought every operator had them, good to know it.

PR #2453, opened 2021-02-23
Update oauth2-proxy version to 7.x

absoludity:
### Description of the change

Verified locally that oidc setup works fine with the new version.

I've not updated the chart versions since #2436 hasn't yet landed which also updates the chart version. If you land this first it will be included, otherwise I'll add an update before landing?

### Benefits

Use the current oauth2-proxy with update support.


andresmgot (approved):
:+1: In any case, I plan to do another release for the bug fixes

PR #2455, opened 2021-02-23
Use normalize.css

antgamdia:
### Description of the change

This PR add `normalize.css` to the project, _a small CSS file that provides better cross-browser consistency in the default styling of HTML elements. It’s a modern, HTML5-ready, alternative to the traditional CSS reset._

### Benefits

The rendered style in different browsers should be more similar after adding this dependency.

### Possible drawbacks

Minor tweaks are required (some spaces o margins). I've have detected some of them, but, since we are migrating the clarity version, we can continue fixing them in subsequent PRs.

### Applicable issues

N/A

### Additional information

N/A


andresmgot (approved):
were you looking for something specific when adding this?

antgamdia:
> were you looking for something specific when adding this?

Not really, I just saw some rules setting some initial values and also remembered the huge login button issue we had in the past.
Since we are updating the css, I just thought it could be a good idea to add it

Issue #1583, opened 2020-03-12
filter packages


kfox1111:
It would be good if there was a way to specify patterns of package names to exclude from display. For example, you may not want to expose to your users the bitcoind package from the stable repository.

andresmgot:
Hi @kfox1111,

That's a good point but the thing is that the stable and incubator repositories are deprecated and will be going away in May:

https://github.com/helm/charts#deprecation-timeline

My recommendation for you is to remove the `stable` repository and instead cherry pick the repositories you are interested in. You can browse the different repositories and the charts they contain here:

https://hub.kubeapps.com/charts

andresmgot:
This issue was defined for things to exclude from a repository. It's also a possibility to include only certain things from a repository. Collecting @antgamdia comment:

> Add chart filtering -during- fetching in asset-syncer. For instance, just fetch "category=cms" or "version=latest"

andresmgot:
I am going to start with a quick design for this

andresmgot:
@absoludity @antgamdia @danoo Please take a look whenever you have the chance:

https://docs.google.com/document/d/1dNv7Id0l_3sac2SAtswtn3vj2eq6naFPGf9-YmN-3JE/edit?usp=sharing

PR #2461, opened 2021-02-24
Add filter rules to AppRepository spec

andresmgot:
### Description of the change

Based on the design at #1583 

Add filter rules to the AppRepository specification.

### Applicable issues

  - ref #1583


cmd/apprepository-controller/pkg/apis/apprepository/v1alpha1/types.go
```diff
@@ -70,6 +72,20 @@ type AppRepositoryCustomCA struct {
 	SecretKeyRef corev1.SecretKeySelector `json:"secretKeyRef,omitempty"`
 }
 
+// FilterRulesSpec defines a set of rules and aggreagation logic
```
antgamdia, line 75: FTR: I guess this object will (in the future) be part of a POST HTTP request. If so, we should also model it in the openapi doc. I can take it, it's just for us to remember it.

cmd/apprepository-controller/pkg/apis/apprepository/v1alpha1/types.go
```diff
@@ -47,9 +47,11 @@ type AppRepositorySpec struct {
 	DockerRegistrySecrets []string `json:"dockerRegistrySecrets,omitempty"`
 	// In case of an OCI type, the list of repositories is needed
 	// as there is no API for the index
+	// DEPRECATED: Use FilterRules instead
```
antgamdia, line 50: I guess you're leaving this until all the dependent functionality is implemented (aka not breaking master with this PR), aren't you?

cmd/apprepository-controller/controller.go
```diff
@@ -607,6 +608,15 @@ func apprepoSyncJobArgs(apprepo *apprepov1alpha1.AppRepository, config Config) [
 		args = append(args, "--tls-insecure-skip-verify")
 	}
 
+	if len(apprepo.Spec.FilterRules.Rules) > 0 {
+		rulesJSON, err := json.Marshal(apprepo.Spec.FilterRules)
+		if err != nil {
+			log.Errorf("Unable to parse filter rules for %s: %v", apprepo.Name, err)
+		} else {
+			args = append(args, "--filter-rules", string(rulesJSON))
```
antgamdia, line 616: I was a bit worried about passing an entire json through the command line, but it seems the maximum length for parameters is enough (`MAX_ARG_STRLEN` is 32 times the page size, that is, about 128K), so no need to be worried about :P

antgamdia (approved):

cmd/apprepository-controller/pkg/apis/apprepository/v1alpha1/types.go
andresmgot: right, I can take a look to that later.

cmd/apprepository-controller/pkg/apis/apprepository/v1alpha1/types.go
andresmgot: yes, at the moment, the filters and the ociRepositories are completely different things, I will merge those in a future PR.

cmd/apprepository-controller/controller.go
andresmgot: yes, in this case, the rules were too complex to use simple flags so I took this option. 

Another option would be to get the filter info directly from the `apprepo` resource but for Helm repositories, that's retrieved after the list of charts are resolved.

cmd/apprepository-controller/pkg/apis/apprepository/v1alpha1/types.go
absoludity: See comments on the design doc (sorry for being late on that).

andresmgot: adapted to the new state of the document

cmd/apprepository-controller/controller_test.go
```diff
@@ -1022,6 +1022,91 @@ func Test_newSyncJob(t *testing.T) {
...
+										"sync",
+										"--database-url=postgresql.kubeapps",
+										"--database-user=admin",
+										"--database-name=assets",
+										"--namespace=kubeapps",
+										"my-charts",
+										"https://charts.acme.com/my-charts",
+										"helm",
+										"--filter-rules",
+										`{"anyOf":[{"jq":".name == $var1","variables":{"var1":"wordpress"}}]}`,
```
absoludity, line 1092: I assume `var1` is just arbitrary - set by the form - but I'm confused in that I expected a different format for the `jq` query. Ah, sorry, so I think your complete example is arbitrary (not just the `var1` but also the jq query?). This just serializes whatever query is set on the app repository. OK.

cmd/apprepository-controller/pkg/apis/apprepository/v1alpha1/types.go
```diff
@@ -70,6 +72,18 @@ type AppRepositoryCustomCA struct {
...
+type FilterRulesSpec struct {
+	AnyOf []FilterRule `json:"anyOf,omitempty"`
+	AllOf []FilterRule `json:"allOf,omitempty"`
+}
+
+// FilterRule defines a rule
+type FilterRule struct {
+	JQ        string            `json:"jq"`
+	Variables map[string]string `json:"variables,omitempty"`
+}
```
absoludity, line 85: The struct here is unexpected for me. I assumed we'd just have a single jq query and currently two possible lists of names... something like:
```go
type FilterRulesSpec struct {
    JQ           string                   `json:"jq"`
    Variables map[string]string `json:"variables,omitempty"`
}
```
where for OCI charts, `Variables` would always include a `package_names` var. 

Or why do you think separate `JQ` query would be needed for the all of and anyof cases, given that we can do a simple ` or ` or ` and ` with jq. Something like `select(.name|test("(pkg1,pkg2,pkg3)") or .name|test("!(pkg4,pkg5,pkg6)")`. (or something, I'd need to check).

Normally I assume the Kubeapps ui would not even required this - it would just be setting a very simple query - "include if the name matches any of the following" or "exclude if the name matches any of the following", combining the entered names into the `(pkg1,pkg2,pkg3)` format. But I just mean that even for the most complicated jq queries we might need in the future (or users might set manually on the AppRepo resource), I think just one JQ query should suffice?

andresmgot: yep, with the JQuery format it's true that the `allOf` or `anyOf` formats are less relevant I just added them because it was in the design document but again it can be simplified

absoludity: Great, thanks.

absoludity (approved):

PR #2463, opened 2021-02-24
Compute filter rules in the asset-syncer

andresmgot:
### Description of the change

Follow up of #2461

Adapt the `asset-syncer` to filter assets based on the given input.

### Applicable issues

  - ref #1583

### Additional information

Note that the filter is applied to the latest version of the chart (the one considered generic).


cmd/asset-syncer/utils.go
```diff
@@ -145,6 +148,68 @@ func (r *HelmRepo) Repo() *models.RepoInternal {
 	return r.RepoInternal
 }
 
+func satisfy(chartInput map[string]interface{}, rule *apprepov1alpha1.FilterRuleSpec) (bool, error) {
+	query, err := gojq.Parse(rule.JQ)
+	if err != nil {
+		return false, fmt.Errorf("Unable to parse JQuery: %v", err)
+
```
absoludity, line 155: Not sure why `gofmt` didn't remove the extra space - don't care, just looked odd.

cmd/asset-syncer/utils.go
```diff
@@ -145,6 +148,68 @@ func (r *HelmRepo) Repo() *models.RepoInternal {
...
+		query,
+		gojq.WithVariables(varNames),
+	)
+	if err != nil {
+		return false, fmt.Errorf("Unable to compile JQuery: %v", err)
+	}
+
+	res, _ := code.Run(chartInput, varValues...).Next()
+	if err, ok := res.(error); ok {
+		return false, fmt.Errorf("Unable to compile JQuery: %v", err)
```
absoludity, line 173: Unable to *run* (and not JQuery :P)

cmd/asset-syncer/utils.go
```diff
@@ -145,6 +148,68 @@ func (r *HelmRepo) Repo() *models.RepoInternal {
...
+	for name, val := range rule.Variables {
+		varNames = append(varNames, name)
+		varValues = append(varValues, val)
+	}
+	code, err := gojq.Compile(
+		query,
+		gojq.WithVariables(varNames),
+	)
+	if err != nil {
+		return false, fmt.Errorf("Unable to compile JQuery: %v", err)
```
absoludity, line 168: jq != JQuery

cmd/asset-syncer/utils.go
```diff
@@ -145,6 +148,68 @@ func (r *HelmRepo) Repo() *models.RepoInternal {
...
+	}
+
+	res, _ := code.Run(chartInput, varValues...).Next()
+	if err, ok := res.(error); ok {
+		return false, fmt.Errorf("Unable to compile JQuery: %v", err)
+	}
+
+	satisfied, ok := res.(bool)
+	if !ok {
+		return false, fmt.Errorf("Unable to convert JQuery result to boolean. Got: %v", res)
```
absoludity, line 178: jq

cmd/asset-syncer/utils.go
```diff
@@ -145,6 +148,68 @@ func (r *HelmRepo) Repo() *models.RepoInternal {
...
+	}
+	code, err := gojq.Compile(
+		query,
+		gojq.WithVariables(varNames),
+	)
+	if err != nil {
+		return false, fmt.Errorf("Unable to compile JQuery: %v", err)
+	}
+
+	res, _ := code.Run(chartInput, varValues...).Next()
```
absoludity, line 171: So everything above this line in this function has been compiling the jq query, independent of any chartInput. Yet this function is called during an iteration of charts, so we're doing all of the above to compile/prepare the jq query each time, for each chart. It seems odd that we're not compiling the `code` in a separate function and calling that just once below *before* iterating the charts?

cmd/asset-syncer/utils_test.go
```diff
@@ -1169,3 +1181,133 @@ func Test_extractFilesFromBuffer(t *testing.T) {
...
+				{Name: "bar"},
+			},
+			apprepov1alpha1.FilterRuleSpec{
+				JQ: `.name | test($var1)`, Variables: map[string]string{"$var1": ".*oo.*"},
+			},
+			[]models.Chart{
+				{Name: "foo"},
+			},
+			nil,
+		},
```
absoludity, line 1298: Sweet, nice to see how the gojq lib fits our needs.

absoludity (approved):
I did see this yesterday, but didn't think it was worth looking at until the update for the filter spec format was done. One small possible refactor to consider if you agree, but fine either way. Thanks!

cmd/asset-syncer/utils.go
andresmgot: ookay, thought that `jq query` was redundant but it's true that jquery is something else

cmd/asset-syncer/utils.go
absoludity: Great, thanks for the update to compile it once before the iteration!

PR #2464, opened 2021-02-24
Minor css fixes and tweaks for the dark mode

### LINKED ISSUE #2462: Fix style for the Application Secrets table
### Description:

The overflow properties there are not working properly:

Peek 2021-02-24 11-24

antgamdia:
_Depends on #2450_ 

### Description of the change

This PR introduces some tweaks for the dark mode (rendered when loading the dark clr theme), mainly converts some svg to white for improving the contrast.
It also fixes an issue with the scrollbars in the secret view and a wrong color in a cds-icon.

### Benefits

It adds these changes:

image

image

image

image

image

### Possible drawbacks

N/A

### Applicable issues

  - fixes #2462 
  -  related #2372
  -  related #2371 

### Additional information

N/A

### DIFF
--- a/dashboard/src/components/AppView/AppSecrets/AppSecrets.scss
+++ b/dashboard/src/components/AppView/AppSecrets/AppSecrets.scss
@@ -5,5 +5,4 @@
 
 .app-secrets-section {
   max-height: 10rem;
-  overflow-y: auto;
 }
--- a/dashboard/src/components/Header/Menu.scss
+++ b/dashboard/src/components/Header/Menu.scss
@@ -67,6 +67,7 @@
 
 .dropdown-divider {
   opacity: 0.2;
+  border-bottom-color: var(--cds-global-color-gray-50, #fafafa) !important;
 }
 
 .dropdown-menu-item {
--- a/dashboard/src/components/Header/Menu.scss
+++ b/dashboard/src/components/Header/Menu.scss
@@ -65,9 +65,9 @@
   }
 }
 
-.dropdown-divider {
+div .dropdown-menu .dropdown-divider {
   opacity: 0.2;
-  border-bottom-color: var(--cds-global-color-gray-50, #fafafa) !important;
+  border-bottom-color: var(--cds-global-color-gray-50, #fafafa);
 }
 
 .dropdown-menu-item {
--- a/dashboard/src/index.scss
+++ b/dashboard/src/index.scss
@@ -3,10 +3,32 @@
 @import "~@cds/core/styles/theme.dark.css"; // clarity core dark theme
 @import "~@cds/core/styles/module.shims.css"; // non-evergreen browser shims
 @import "~@cds/city/css/bundles/default.css"; // load base font
-// clarity UI light/dark themes should be conditionalled loaded since they don't support dynamic theming
+// clarity UI light/dark themes are loaded by changing <link href=...>, since they aren't prepared for theming
 
 svg.icon {
   /* Fix positioning */
   position: relative;
   bottom: -0.125em;
+}
+
+// CSS tweaks the dark mode theme theme
+[cds-theme~="dark"] {
+  %convert-to-white {
+    filter: invert(1) brightness(100%) saturate(0%);
+  }
+
+  .section-not-found img {
+    // White 404 logo
+    @extend %convert-to-white;
+  }
+
+  .kubeapps-header-subtitle img {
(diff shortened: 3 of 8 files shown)


dashboard/src/components/ThemeSelector/DarkTheme.scss
```diff
@@ -0,0 +1,16 @@
...
+}
+
+.kubeapps-header-subtitle img {
+  // Convert to white
+  filter: invert(1) brightness(100%) saturate(0%);
+}
+
+// Convert to white
+.bg-img img {
+  filter: invert(1) brightness(100%) saturate(0%);
```
absoludity, line 15: I was just reading about scss to dry this up while still having clear, separate selectors. Not important at all, but see we can do something like:

```
%convert-to-white {
  filter: invert(1) brightness(100%) saturate(0%);
}
.section-not-found img {
  @extend %convert-to-white;
}
...
```

absoludity (approved):

dashboard/src/components/ThemeSelector/DarkTheme.scss
antgamdia: Thanks a lot for the tip. I haven't extensively used scss and wasn't aware of this `@extend`. Really useful :)

antgamdia:
Since you have also reviewed other dark mode PRs, please take a look at this PR @andresmgot and see what you think :)
Thanks!

dashboard/src/components/AppView/AppSecrets/AppSecrets.scss
```diff
@@ -5,5 +5,4 @@
 
 .app-secrets-section {
   max-height: 10rem;
-  overflow-y: auto;
```
andresmgot, line 8: this fixes https://github.com/kubeapps/kubeapps/issues/2462 ?

dashboard/src/components/Header/Menu.scss
```diff
@@ -67,6 +67,7 @@
 
 .dropdown-divider {
   opacity: 0.2;
+  border-bottom-color: var(--cds-global-color-gray-50, #fafafa) !important;
```
andresmgot, line 70: avoid `!important` :)

dashboard/src/index.scss
```diff
@@ -3,10 +3,32 @@
...
+
+  .kubeapps-header-subtitle img {
+    // White header subtitle Helm logo
+    @extend %convert-to-white;
+  }
+
+  .bg-img img {
+    // White Helm logo in card overlay
+    @extend %convert-to-white;
+  }
```
andresmgot, line 33: :+1: 

I wonder if we will miss whenever we add new imgs

andresmgot (approved):

dashboard/src/components/Header/Menu.scss
antgamdia: The point is that the dark.mode clr-ui is loaded by helmet directly as the last element (ie, after the css bundle). So the selector from the `.dropdown-divider` selector from clr-ui-dark has more priority.

We could add an id and apply the style there, but I wanted this PR to be css-only. Would you prefer adding the id selector?

dashboard/src/index.scss
antgamdia: Probably, but, since we have the toggle button, we can test it in runtime :p 
However, what concerns me the most, is branding compliance; changing colors arbitrarily is usually not supported by brand guidelines. However, none of our current icons is a trademark, but we should be aware of that in the future,

dashboard/src/components/AppView/AppSecrets/AppSecrets.scss
antgamdia: Perhaps, I am missing something, but as far as I have checked, I would say so

secretsScroll

dashboard/src/components/Header/Menu.scss
andresmgot: you just need to make the selector more scoped, if you set it as `div .dropdown-menu .dropdown-divider` for example, it should work.

antgamdia: Yep, you're totally right, it works. Thanks!! :)

dashboard/src/components/AppView/AppSecrets/AppSecrets.scss
andresmgot: great, thanks for checking!

PR #2483, opened 2021-03-01
feat(form): Add custom form component integration

### LINKED ISSUE #2352: Feature: Custom Form Components
Kubeapps is an awesome UI solution for launching applications on Kubernetes. However I think that it lacks a lot of customization. One area that I think has a lot of customization potential is the deployment form. I think allowing developers to import custom components would be a great feature and would allow devs to take control of the form components. The component devs can add live validation, custom UI elements, etc and really bring the form building to the next level. 

After looking through the codebase these seem to be the places that would need to change to make this happen:
1. An additional field in `IBasicFormParam`. This would be called something like `customFormComponent` which would exist in the json schema to tell the dashboard that we want to render a custom component instead of one of the provided form components.
2. Some logical condition that checks for `param.customFormComponent` so that we can render the component.
3. A way to programmatically import the component package into the conditional stated above (im not quite sure the best way to do this).

Im very interested in this and am willing to start working on it if you think this is something that would be possible @andresmgot. Im also open to other implementations I just outlined what seemed like the easiest/best way.

aanthonyrizzo:
Signed-off-by: Anthony Rizzo <36711320+aanthonyrizzo@users.noreply.github.com>

### Description of the change

This PR adds the ability to load custom form components from a separate js file by passing the render param and the the handleChange func.

### Benefits

- Allow devs to build custom UI components (radios for example)
- Allow devs to consume external APIs inside of components for rendering logic/validation

### Possible drawbacks

- Unfortunately the external components have to be loaded via a separate js file (im imagining this is a js bundle). As you will see I decided to use a pvc to mount the file into the dashboard nginx so that when pods are deleted the js file is remounted. 

### Applicable issues

Fixes #2352 

### Additional information

Example:
image
Instead of rendering a dropdown it rendered 'Hello world!' from a js bundle file

### DIFF
--- a/chart/kubeapps/templates/dashboard-config.yaml
+++ b/chart/kubeapps/templates/dashboard-config.yaml
@@ -27,6 +27,10 @@ data:
         root /app/custom-locale/;
       }
 
+      location /custom_components.js {
+        root /app/custom-components/;
+      }
+
       location / {
         # Redirects are required to be relative otherwise the internal hostname will be exposed
         absolute_redirect off;
@@ -50,6 +54,8 @@ data:
     }
   custom_style.css: |-
 {{- .Values.dashboard.customStyle | nindent 4 }}
+  custom_components.js: |-
+{{- .Values.dashboard.customComponents | nindent 4 }}
   custom_locale.json: |-
 {{- .Values.dashboard.customLocale | toJson | nindent 4 }}
   config.json: |-
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -508,6 +508,10 @@ dashboard:
   ##      background-color: #991700;
   ##    }
   customStyle: ""
+  ## Custom Form components injected into the BasicDeploymentForm
+  ## Read the reference below for a step-by-step integration
+  ## https://github.com/kubeapps/kubeapps/blob/master/docs/developer/custom-form-component-support.md
+  customComponents: ""
   ## Custom translations injected to the Dashboard to customize the strings used in Kubeapps
   ## Ref: https://github.com/kubeapps/kubeapps/blob/master/docs/developer/translate-kubeapps.md
   ## e.g:
--- a/dashboard/src/RemoteComponent.js
+++ b/dashboard/src/RemoteComponent.js
@@ -1,0 +1,6 @@
+import { createRemoteComponent, createRequires } from "@paciolan/remote-component";
+import { resolve } from "./remote-component.config.js";
+
+const requires = createRequires(resolve);
+
+export const CustomComponent = createRemoteComponent({ requires });
--- a/dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/CustomFormParam.test.tsx
+++ b/dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/CustomFormParam.test.tsx
@@ -1,0 +1,42 @@
+import { mount } from "enzyme";
+import { IBasicFormParam } from "shared/types";
+import { CustomComponent } from "../../../RemoteComponent";
+import CustomFormComponentLoader from "./CustomFormParam";
+
+const param = {
+  path: "enableMetrics",
+  value: true,
+  type: "boolean",
+  customComponent: {
+    className: "test",
+  },
+} as IBasicFormParam;
+
+const defaultProps = {
+  param,
+  handleBasicFormParamChange: jest.fn(),
+};
+
+// Mocking the window so that the injected components are imported correctly
+const { location } = window;
+beforeAll((): void => {
+  // @ts-ignore
+  delete window.location;
+  // @ts-ignore
+  window.location = {
+    origin: "../../../../../docs/developer/examples/CustomComponent.min.js",
+  };
+});
+afterAll((): void => {
+  window.location = location;
+});
+
+it("should render a custom form component", () => {
+  const wrapper = mount(<CustomFormComponentLoader {...defaultProps} />);
+  expect(wrapper.find(CustomFormComponentLoader)).toExist();
+});
+
+it("should render the remote component", () => {
+  const wrapper = mount(<CustomFormComponentLoader {...defaultProps} />);
+  expect(wrapper.find(CustomComponent)).toExist();
+});
--- a/dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/CustomFormParam.tsx
+++ b/dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/CustomFormParam.tsx
@@ -1,0 +1,29 @@
+import React, { useMemo } from "react";
+import { IBasicFormParam } from "shared/types";
+import { CustomComponent } from "../../../RemoteComponent";
+
+export interface ICustomParamProps {
+  param: IBasicFormParam;
+  handleBasicFormParamChange: (
+    p: IBasicFormParam,
+  ) => (e: React.FormEvent<HTMLInputElement>) => void;
+}
+
+export default function CustomFormComponentLoader({
+  param,
+  handleBasicFormParamChange,
+}: ICustomParamProps) {
+  // Fetches the custom-component bundle served by the dashboard nginx
+  const url = `${window.location.origin}/custom_components.js`;
+
+  return useMemo(
+    () => (
+      <CustomComponent
+        url={url}
+        param={param}
+        handleBasicFormParamChange={handleBasicFormParamChange}
+      />
+    ),
+    [handleBasicFormParamChange, param, url],
+  );
+}
--- a/dashboard/src/shared/types.ts
+++ b/dashboard/src/shared/types.ts
@@ -579,6 +579,7 @@ export interface IBasicFormParam {
   maximum?: number;
   render?: string;
   description?: string;
+  customComponent?: object;
   enum?: string[];
   hidden?:
     | {
--- a/docs/developer/custom-form-component-support.md
+++ b/docs/developer/custom-form-component-support.md
@@ -1,0 +1,48 @@
+# Custom Form Component Support
+This is an extension to the [basic form support](https://github.com/kubeapps/kubeapps/blob/master/docs/developer/basic-form-support.md#basic-form-support)
+
+## Possible use cases
+- Custom UI component that are not yet supported by the basic components (e.g: radio selectors)
+- Consuming third party APIs for component values and validation
+
+## Step-by-step integration process
+1. First you will need a react component to render instead of the default Kubeapps form components. You're React components must be compiled into a JS file so that they can be interpreted by the browser since they cannot natively parse `.jsx` or `.tsx` files. You can compile `jsx` or `tsx` into js with tools like webpack, create-react-app, babel, etc. If you just want to try this feature out and you don't have a component yet we provide some test files you can try (Do not try to load the `jsx` file since browsers cannot parse it! We simply include it so that you can see the pre-compiled version of the `.js` files).
+2. The easiest way to add inject the file in is via the command line. You can do it via the following command:
+    ```
+    helm install  bitnami/kubeapps --set-file dashboard.customComponents=*path to file* <other_flags>
+    ```
+    Note: The file can be located anywhere on your file system or even a remote source!
+
+3. Once the deployment is complete you will need a values json that will signal to Kubeapps that we want to render a custom component and not one fo the provided ones. To do that you will need a `values.json.schema` that has a `customComponent` key, more info [here](https://github.com/kubeapps/kubeapps/blob/master/docs/developer/custom-form-component-support.md#render-a-custom-component).
+
+## Render a custom component
+Signaling to the Kubeapps dashboard that you want to render a custom component is pretty straight forward.  Simply add a `customComponent` field to any form parameter defined in the  `values.json.schema` and that will tell the react application to fetch the component from the custom js bundle. An example parameter could look like:
+```
+    "databaseType": {
+      "type": "string",
+      "form": true,
+      "enum": ["mariadb", "postgresql"],
+      "title": "Database Type",
+      "description": "Allowed values: \"mariadb\" and \"postgresql\"",
+      "customComponent": {
+          "type": "radio",
+          "className": "primary-radio"
+      }
+    }
+```
+Note: The `customComponent` field **MUST BE AN OBJECT**. This design decision was made so that developers can pass extra values/properties into their custom components should they require them.
+
+## Updating Helm values with custom components
+Custom form components would be useless without the ability to interact with the YAML state. To do this your custom components should be set up to receive 2 props: `handleBasicFormParamChange` and `param`. `param` is the current json object this is being rendered (denoted by the `customComponent` field) and `handleBasicFormParamChange` which is a function that updates the YAML file. An example of how you use this function can be found in any of the BasicDeploymentForm components such as the [SliderParam](https://github.com/kubeapps/kubeapps/blob/master/dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/SliderParam.tsx#L47-L53).
+```
+  const handleParamChange = (newValue: number) => {
+    handleBasicFormParamChange(param)({
+      currentTarget: {
+        value: newValue,
+      },
+    } as React.FormEvent<HTMLInputElement>);
+  };
+```
+
+## Tips
+To help you get started we provide some examples that you can try [here](https://github.com/kubeapps/kubeapps/blob/master/docs/developer/examples). The three files should give you a good idea about how to start developing and building your own custom components. `CustomComponent.jsx` is a super simple react component that takes the `handleBasicFormParamChange` and `param` props and renders a button that changes the value to 'test'. `CustomComponent.js` is the JavaScript variant of `CustomComponent.jsx` and `CustomComponent.min.js` is a minified js bundle created using [remote-component-starter](https://github.com/Paciolan/remote-component-starter), which is specifically made to help build components that you want to load remotely with the [remote-component](https://github.com/Paciolan/remote-component) tool used by Kubeapps.
--- a/chart/kubeapps/templates/dashboard-deployment.yaml
+++ b/chart/kubeapps/templates/dashboard-deployment.yaml
@@ -57,6 +57,8 @@ spec:
               name: custom-css
             - mountPath: /app/custom-locale
               name: custom-locale
+            - mountPath: /app/custom-components
+              name: custom-components
           ports:
             - name: http
               containerPort: 8080
@@ -88,3 +90,9 @@ spec:
             items:
               - key: custom_locale.json
                 path: custom_locale.json
+        - name: custom-components
+          configMap:
+            name: {{ template "kubeapps.dashboard-config.fullname" . }}
+            items:
+              - key: custom_components.js
+                path: custom_components.js
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -8,6 +8,7 @@
     "@cds/core": "^5.0.3",
     "@cds/react": "^5.0.3",
     "@clr/ui": "^5.0.3",
+    "@paciolan/remote-component": "^2.11.0",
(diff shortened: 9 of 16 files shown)


chart/kubeapps/templates/custom-component.pvc.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
```
absoludity, line 1: Shouldn't this PVC be conditional on the user setting some custom component in the chart?

absoludity (commented):
Thanks @aanthonyrizzo . I'll let Andres go through this in more detail, but a couple of questions:

It looks like you're intending that the operator deploying the chart would need to load the bundle/js onto the pv? Would another approach be to have a `dashboard.customComponentJSFile` chart option that takes a file path and uses helm's [support for files](https://helm.sh/docs/chart_template_guide/accessing_files/) to load the file into a config map?

dashboard/src/shared/types.ts
```diff
@@ -579,6 +579,7 @@ export interface IBasicFormParam {
   maximum?: number;
   render?: string;
   description?: string;
+  custom_component?: object;
```
andresmgot, line 582: this should be `customComponent` (we are using camelCase)

chart/kubeapps/templates/custom-component.pvc.yaml
andresmgot: In fact, I would use a configmap instead. At the moment it seems that we are just defining a single file `custom_components.js` so I would follow the same approach than `custom_style.css` which is adding the code in the `dashboard-config.yaml` file (and mount it from the config file):

```
  custom_components.js: |-
{{- .Values.dashboard.customComponents | nindent 4 }}
```

andresmgot (commented):
Thanks @aanthonyrizzo. I think this is going in the right direction, I would just move the custom code to a configmap to follow the "standard" (it also has the benefit that the code is easier to set).

chart/kubeapps/templates/custom-component.pvc.yaml
aanthonyrizzo: @andresmgot are you sure placing a js bundle into a configmap is the route that you want to go? a bundle can get rather large

absoludity: > In fact, I would use a configmap instead. 

Yep, I'd suggested the same in my [main comment](https://github.com/kubeapps/kubeapps/pull/2483#pullrequestreview-601342624). 

> @andresmgot are you sure placing a js bundle into a configmap is the route that you want to go? a bundle can get rather large

Do you mean that a bundle of custom components for the Kubeapps might be > 1MB ? I'd be pretty horrified if that's the case :) Or do you mean that you think there would be an issue with a 10k bundle in a configmap?

aanthonyrizzo: I've just never used a configMap like this before but since this is the established pattern I should follow it 👍

chart/kubeapps/values.yaml
```diff
@@ -508,6 +508,12 @@ dashboard:
   ##      background-color: #991700;
   ##    }
   customStyle: ""
+  ## Relative path to custom components js bundle
+  ## Note: The bundle file MUST be called 'custom_components.js'
```
andresmgot, line 512: mmh, I think it's not needed to be called `custom_components.js`, the Helm function will just read whatever file you pass it.

dashboard/src/RemoteComponent.js
```diff
@@ -0,0 +1,6 @@
+import { createRemoteComponent, createRequires } from "@paciolan/remote-component";
+import { resolve } from "./remote-component.config.js";
+
+const requires = createRequires(resolve);
+
+export const CustomComponent = createRemoteComponent({ requires });
```
andresmgot, line 6: can you use Typescript here instead? (not sure if it's not possible for some reason)

dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/CustomFormParam.tsx
```diff
@@ -0,0 +1,29 @@
...
+  return useMemo(
+    () => (
+      <CustomComponent
+        url={url}
+        param={param}
+        handleBasicFormParamChange={handleBasicFormParamChange}
+      />
+    ),
+    [handleBasicFormParamChange, param, url],
+  );
```
andresmgot, line 28: can you add some basic testing for this component?

docs/developer/custom-form-component-support.md
```diff
@@ -0,0 +1,42 @@
...
+      "enum": ["mariadb", "postgresql"],
+      "title": "Database Type",
+      "description": "Allowed values: \"mariadb\" and \"postgresql\"",
+      "customComponent": {
+          "type": "radio",
+          "className": "primary-radio"
+      }
+    }
+```
+Note: The `customComponent` field **MUST BE AN OBJECT**. This design decision was made so that developers can pass extra values/properties into their custom components.
```
andresmgot, line 30: is there any mandatory fields in that object? (I assume `className` is required?)

docs/developer/custom-form-component-support.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Updating values with custom components
+Custom form components would be useless without the ability to interact with the YAML state. To do this your custom components should be set up to receive 2 props: `handleBasicFormParamChange` and `param`. `param` is the current json object this is being rendered (denoted by the `customComponent` field) and `handleBasicFormParamChange` which is a function that updates the YAML file. An example of how you use this function can be found in any of the BasicDeploymentForm components such as the [SliderParam](https://github.com/kubeapps/kubeapps/blob/master/dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/SliderParam.tsx#L47-L53).
+```
```
andresmgot, line 34: can you add an example of `custom_component.js` so people can copy-paste and modify it?

docs/developer/custom-form-component-support.md
```diff
@@ -0,0 +1,42 @@
...
+  const handleParamChange = (newValue: number) => {
+    handleBasicFormParamChange(param)({
+      currentTarget: {
+        value: newValue,
+      },
+    } as React.FormEvent<HTMLInputElement>);
+  };
+```
```
andresmgot, line 42: thanks for the docs! Can you add a link to this file from the `basic-form-support.md` file?

andresmgot (commented):
Thanks for the changes @aanthonyrizzo! I just have some minor comments.

docs/developer/custom-form-component-support.md
aanthonyrizzo: I think its up to the person integrating whether they want/need to pass anything extra to the components. If they are only rendering one component then it might be easier for them to just hardcode the className in the component they are importing

docs/developer/custom-form-component-support.md
aanthonyrizzo: Yup!

dashboard/src/RemoteComponent.js
aanthonyrizzo: I spent sometime looking into this and the package doesnt seem to support a typescript implementation yet (they plan on adding it according to their docs). I can add a TODO there for when the package is updated (they seems to update it quite frequently)

dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/CustomFormParam.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+    origin: "./fixtures",
+  };
+});
+afterAll((): void => {
+  window.location = location;
+});
+
+it("should render a custom form component", () => {
+  const wrapper = mount(<CustomFormComponentLoader {...defaultProps} />);
+  expect(wrapper).toMatchSnapshot();
```
andresmgot, line 35: rather than a snapshot I would just check if `CustomComponent` exists.

andresmgot (approved):
Just one very minor comment about the test and we can merge this :) Thanks!

andresmgot:
Also, you need to merge master and fix the conflict with the `package.json`

docs/developer/custom-form-component-support.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Tips
+To try this feature out we offer a pre-minified, bundled react component that you can try out! Note: It does not do much, its a simple button that will set the value of whatever param you render the custom component into as `test`. But it will help you understand how to integrate custom components. That file can be found [here](https://github.com/kubeapps/kubeapps/blob/master/dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/fixtures/custom_components.js).
+
+We also recommend checking out this project https://github.com/Paciolan/remote-component-starter. Its specifically made to help build out components that you want to load remotely into Kubeapps.
```
absoludity, line 47: I'm a little wary of recommending 3rd party packages (especially ones that could be used to load remote code). I'm OK with us using it if @andresmgot has checked it out, but wouldn't send traffic there necessarily unless we're confident in the developer.

dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/fixtures/custom_components.js
```diff
@@ -0,0 +1,220 @@
...
+        onClick={() => handleChange("test")}
+        style={"test" === value ? selectedStyle : null}
+      >
+        Test
+      </button>
+    </>
+  );
+}
+ */
+!(function (t, e) {
```
absoludity, line 43: So a couple of questions:

1. Does the custom component code need to be pre-minified? If not, let's not pre-minify it so users don't need to trust it's not doing something malicious nor go to the effort of verifying it is the minified version of the example code.
2. I understand from the docs that this doesn't need to be under `dashboard/src` at all (where it could at some point easily be bundled in the Kubeapps js by mistake with some unknown minified js doing something bad). I'm strongly against having any chance of pre-minified js code bundled in Kubeapps. Can we instead place this file under `docs/developer/...` next to the documentation (and update the link to it below).

Or let us know why doing so is an issue? Thanks!

docs/developer/custom-form-component-support.md
```diff
@@ -0,0 +1,47 @@
...
+    handleBasicFormParamChange(param)({
+      currentTarget: {
+        value: newValue,
+      },
+    } as React.FormEvent<HTMLInputElement>);
+  };
+```
+
+## Tips
+To try this feature out we offer a pre-minified, bundled react component that you can try out! Note: It does not do much, its a simple button that will set the value of whatever param you render the custom component into as `test`. But it will help you understand how to integrate custom components. That file can be found [here](https://github.com/kubeapps/kubeapps/blob/master/dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/fixtures/custom_components.js).
```
absoludity, line 45: We'd need to update this link after moving the fixture example to the docs directory.

absoludity (changes requested):

dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/fixtures/custom_components.js
aanthonyrizzo: yeah we can easily move it there! i just put it here since i use it in the tests for the component so i figured it would make more sense here. and it don't not have to be minimized, it's the output from the repo that i mentioned earlier

docs/developer/custom-form-component-support.md
aanthonyrizzo: i can def remove it. hopefully if people need their own components they can bundle their own components haha. the only reason i linked it is because this repo is support by the same developer that supports the library used to load the components

absoludity: Perhaps include the link, but just as "The bundled custom component that you can try was created with [remote-component-starter], which is specifically made to help build components that you want to load remotely with the [remote-component] tool used by Kubeapps" ? (ie. just not recommending, but providing info?)

dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/fixtures/custom_components.js
absoludity: Great, thanks.

docs/developer/custom-form-component-support.md
aanthonyrizzo: yeah i totally understand, will make this change

absoludity (approved):
Changing to +1 so that it's ready to land after the agreed change is pushed :)

andresmgot:
> Changing to +1 so that it's ready to land after the agreed change is pushed :)

Thanks for the feedback!

andresmgot:
hey @aanthonyrizzo, just a heads up that we would like to release Kubeapps 2.3 next week and we would like to include this as well. Are you going to be able to apply @absoludity comments during this week? Thanks in advance.

aanthonyrizzo:
> hey @aanthonyrizzo, just a heads up that we would like to release Kubeapps 2.3 next week and we would like to include this as well. Are you going to be able to apply @absoludity comments during this week? Thanks in advance.

Yup! I will have all the changes pushed by EOD today! Thank you!

dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/CustomFormParam.test.tsx
andresmgot: @aanthonyrizzo I think you missed this comment.

docs/developer/examples/custom_components.js
```diff
@@ -0,0 +1,220 @@
...
+        onClick={() => handleChange("test")}
+        style={"test" === value ? selectedStyle : null}
+      >
+        Test
+      </button>
+    </>
+  );
+}
+ */
+(function(e, a) { for(var i in a) e[i] = a[i]; }(exports, /******/ (function(modules) { // webpackBootstrap
```
andresmgot, line 43: can we just use non-minified code here? (or at least tell instructions to generate the minified version from the original code). 

The idea is to allow people to use this code as template for future components, if it's minified it won't be that helpful.

andresmgot (commented):
Hi @aanthonyrizzo, just a couple of minor comments left. Let me know what you think,

dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/CustomFormParam.test.tsx
aanthonyrizzo: yes I did, my bad

docs/developer/examples/custom_components.js
aanthonyrizzo: I added 3 different examples. A `jsx` component, that `jsx` component converted to `js` and then finally the minified version of that file as generated by the remote helper library

docs/developer/custom-form-component-support.md
```diff
@@ -0,0 +1,45 @@
...
+    handleBasicFormParamChange(param)({
+      currentTarget: {
+        value: newValue,
+      },
+    } as React.FormEvent<HTMLInputElement>);
+  };
+```
+
+## Tips
+To help you get started we provide some examples that you can try [here](https://github.com/kubeapps/kubeapps/blob/master/docs/developer/examples). The three files should give you a good idea about how to start developing and building your own custom components. `CustomComponent.jsx` is a super simple react component that takes the `handleBasicFormParamChange` and `param` props and renders a button that changes the value to 'test'. `CustomComponent.js` is the JavaScript variant of `CustomComponent.jsx` and `CustomComponent.min.js` is a minified js bundle created using [remote-component-starter](https://github.com/Paciolan/remote-component-starter), which is specifically made to help build components that you want to load remotely with the [remote-component](https://github.com/Paciolan/remote-component) tool used by Kubeapps.
```
andresmgot, line 45: It would be better if we provide here a step-by-step guide of how to work with `remote-component-starter`. At the beginning, I thought that the minified file was just a performance improvement but giving this a try it's a necessary step to be able to load the component. Also, I was able to use the javascript file (CustomComponent.js) but no the file `CustomComponent.jsx`, how do you transform it? is it manual?

andresmgot (commented):
Hi @aanthonyrizzo, thanks for the changes! I tried this locally but the chart didn't work as expected, check the inline comment.

Also, it was a bit confusing for me to make this work, I think that the documentation can benefit from a step-by-step guide?

chart/kubeapps/templates/dashboard-config.yaml
```diff
@@ -50,6 +54,8 @@ data:
     }
   custom_style.css: |-
 {{- .Values.dashboard.customStyle | nindent 4 }}
+  custom_components.js: |-
+{{- .Files.Get .Values.dashboard.customComponents | nindent 4 }}
```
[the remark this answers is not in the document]
aanthonyrizzo: But dont you think copying the file into the chart directory is still much easier than copying the content of the file into the values file? The reason we really like this approach is so that developers can easily hook their custom components into their own CI/CD for deploying kubeapps as a dependency in their own chart

docs/developer/custom-form-component-support.md
aanthonyrizzo: JSX files cannot be rendering in browsers natively but included that file so that it would be easier to understand the other 2 files.

aanthonyrizzo:
> Hi @aanthonyrizzo, thanks for the changes! I tried this locally but the chart didn't work as expected, check the inline comment.
> 
> Also, it was a bit confusing for me to make this work, I think that the documentation can benefit from a step-by-step guide?

I can def write a more step-by-step guide if you think that would help

chart/kubeapps/templates/dashboard-config.yaml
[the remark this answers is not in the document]
andresmgot: Not really, in that case you need to pull and unpack the chart and then copy the file before you are able to install it. Note that you don't need to manually copy the file. In your use case, you can simply do:

```
helm install bitnami/kubeapps --set-file dashboard.customComponents=./docs/developer/examples/CustomComponent.min.js <other_flags>
```

That would read and properly set the file from any path.

docs/developer/custom-form-component-support.md
andresmgot: Got it. That's why it'd be useful to be able to have a step-by-step guide to know how to do all this.

andresmgot:
> I can def write a more step-by-step guide if you think that would help

That'd be great. Thanks!

chart/kubeapps/templates/dashboard-config.yaml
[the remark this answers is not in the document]
aanthonyrizzo: ohhhhh okay I see, this seems way better. will update and add some docs 👍

andresmgot (approved):
Thanks @aanthonyrizzo for the efforts here :)

dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/CustomFormParam.tsx
```diff
@@ -0,0 +1,29 @@
...
+    p: IBasicFormParam,
+  ) => (e: React.FormEvent<HTMLInputElement>) => void;
+}
+
+export default function CustomFormComponentLoader({
+  param,
+  handleBasicFormParamChange,
+}: ICustomParamProps) {
+  // Fetches the custom-component bundle served by the dashboard nginx
+  const url = `${window.location.origin}/custom_components.js`;
```
mecampbellsoup, line 17: @andresmgot @aanthonyrizzo this request happens in the browser, correct? Doesn't this file need to be versioned/fingerprinted in order to bust browser cache when custom components bundle contents/file changes?

absoludity: I've not explicitly tried, but I wouldn't think so. The loading is using https://github.com/Paciolan/remote-component#src/RemoteComponent.js which uses remote-module-loader which uses fetch (by default) with the default cache option. If you find it is, we can update the config for the loader it seems.

PR #2489, opened 2021-03-02
Use cds modals

antgamdia:
### Description of the change

As suggested in https://github.com/kubeapps/kubeapps/pull/2447#issuecomment-788864141, this PR removes: 1) `react-modal`; 2) `components/js/modal.js`. Instead, we use the `cds-modal` from clarity v5.
To simply the migration, the existing `modal` component has been replaced by the new one using cds.
This `<Modal>` is not used everywhere since there are place with a sliglyly different logic (adding a spinner, using a form, etc)

### Benefits

Use the same cds-modals across the entire project.

### Possible drawbacks

Perhaps, some minor differences in the style, if we compare it to the former modals.

### Applicable issues

  - related #2372

### Additional information

Some examples:

Before:
image

After:
image

Before:
image

After:
image


dashboard/src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,46 @@
...
+  footer?: JSX.Element;
+  hideCloseButton?: boolean;
+  modalSize?: "default" | "sm" | "lg" | "xl" | undefined;
+  showModal: boolean;
+  staticBackdrop?: boolean;
+  title?: string;
+  onModalClose: () => any;
+}
+
+function Modal({
```
andresmgot, line 16: Using sometimes this modal and some other no can be confused, can we just not use the wrapper? (it's not super complex and we avoid having to write tests for this)

dashboard/src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,46 @@
...
+      {showModal && (
+        <CdsModal size={modalSize} closable={!hideCloseButton} onCloseChange={onModalClose}>
+          <CdsModalHeader>{title}</CdsModalHeader>
+          <CdsModalContent>{children}</CdsModalContent>
+          <CdsModalActions>
+            {footer}
+            {!hideCloseButton && (
+              <CdsButton action="outline" type="button" onClick={onModalClose}>
+                Cancel
+              </CdsButton>
```
andresmgot, line 37: I would skip the "Cancel" button if there is already a close button

andresmgot (commented):
Thanks for running the extra mile, let me know what you think about my comment.

dashboard/src/components/Modal/Modal.tsx
antgamdia: Mmm, good idea, since we will only have to deal with one modal. On the other hand, this wrapper is encapsulating some cds-modal code... but, we are using other cds-modals out there... so, yes, let's get rid of the wrapper!

andresmgot:
> After:
> image

Can we reduce the size of the modal in that case? :) Also, I was checking this locally and clicking on the button "Cancel" caused the form to be submitted (to deploy the application), can you check if that is fixed with this?

antgamdia:
> > After:
> > image
> 
> Can we reduce the size of the modal in that case? :) Also, I was checking this locally and clicking on the button "Cancel" caused the form to be submitted (to deploy the application), can you check if that is fixed with this?

I'm gonna check right now, but it seems (again) the issue with the buttons. Somehow, cds-button inside a form are losing the 
`type="button"` attribute as you already reported in the past.

antgamdia:
> Can we reduce the size of the modal in that case?

Currently, it uses de "default" size. The next smaller one is "sm" (but it's a bit ugly, isn't it?)

image

antgamdia:
RollbackDialog also changed: 

Before:
image

After:
image

dashboard/src/components/ConfirmDialog/ConfirmDialog.tsx
```diff
@@ -20,35 +21,49 @@ function ConfirmDialog({
...
+            </>
+          ) : (
+            <>
+              <CdsModalContent>
+                <p>{confirmationText}</p>
+                <p>{extraElem}</p>
+              </CdsModalContent>
+              <CdsModalActions>
+                {/* TODO(andresmgot): CdsButton "type" property doesn't work, so we need to use a normal <button>
+                  https://github.com/vmware/clarity/issues/5038
```
andresmgot, line 53: it'd be good to investigate this...

andresmgot (approved):

andresmgot:
> > Can we reduce the size of the modal in that case?
> 
> Currently, it uses de "default" size. The next smaller one is "sm" (but it's a bit ugly, isn't it?)
> 
> image

The problem was not its width but its height, can we make it smaller?

dashboard/src/components/ConfirmDialog/ConfirmDialog.tsx
antgamdia: Totally agree, perhaps the best option is to create a very simple scenario in a demo environment and pass it through the internal support channel, as I did with the modal issue

antgamdia:
> The problem was not its width but its height, can we make it smaller?

Using just the cds `size` attribute, I don't think so. Here is a version after removing an extra `<p>
Another idea is just to force it to be smaller via css, but it is the last resort option. As you may prefer.

image

Perhaps with a title, it looks a little bit better:

image

andresmgot:
I have already approved this, feel free to merge it whenever you are happy with the result.

antgamdia:
After adding a title and increasing its size:

image

Now it looks better

Issue #2484, opened 2021-03-02
Unable to show the Catalog tab


jiad-vmware:
Just installed KubeApps on Tanzu Kubernetest cluster on vSphere with these steps:
1. helm repo add bitnami https://charts.bitnami.com/bitnami
2. kubectl create namespace kubeapps
3. helm install kubeapps --namespace kubeapps \
  --set assetsvc.service.port=9090 \
  --set frontend.service.type=LoadBalancer \
  --set useHelm3=true \
    bitnami/kubeapps
4. kubectl create --namespace default serviceaccount kubeapps-operator
5. kubectl create clusterrolebinding kubeapps-operator --clusterrole=cluster-admin --serviceaccount=default:kubeapps-operator
6. kubectl get secret $(kubectl get serviceaccount kubeapps-operator -o jsonpath='{range .secrets[*]}{.name}{"\n"}{end}' | grep kubeapps-operator-token) -o jsonpath='{.data.token}' -o go-template='{{.data.token | base64decode}}' && echo
7. login to web UI and click Catalog tab.
8. See an error message "An error occurred while fetching the catalog:"
![kubeappserror](https://user-images.githubusercontent.com/68653329/109598633-7e8db580-7ace-11eb-872b-302f24430b1a.PNG)

absoludity:
Unfortunately we don't have much experience with vSphere k8s clusters yet (but hoping to change that). Can you screenshot the network tab of your browser's developer console (if you can click on the actual request to see more info in the response, that'd be great too.

Can you also paste the output of `kubectl get pods --namespace kubeapps` right after you verify the error still happens? Thanks.

jiad-vmware:
![kubeappserror3](https://user-images.githubusercontent.com/68653329/109601692-efd86300-7ae4-11eb-8d4b-b3408ff91e4e.PNG)
![kubeappserror4](https://user-images.githubusercontent.com/68653329/109601695-f070f980-7ae4-11eb-892d-78867d8b4292.PNG)
![kubeappserror2](https://user-images.githubusercontent.com/68653329/109601696-f1099000-7ae4-11eb-83ac-3b6f6c434017.PNG)


There are a lot of requests like this http://172.29.214.134/api/assetsvc/v1/clusters/default/namespaces/default/charts?page=14838&size=20 . The responses have error messages like : {"code":401,"message":"Authorization token missing"} .

BTW, what is the cluster name should be? I am not sure why it is default here. The name of my cluster should be app01-gc.

[...]

PR #2496, opened 2021-03-03
Avoid endless page requesting if when failing requests + manual retry

antgamdia:
### Description of the change

This PR aims at solving one of the issues detected in https://github.com/kubeapps/kubeapps/issues/2484#issuecomment-788658490. Currently, the pagination is trying to increase the page regardless of the status code.

In this PR we stop requesting more pages if an error is detected. However, solely adding that has a severe impact on the UX: if request happens to fail, all the previously requested items will be discarded and the only way to retry is by performing a hard reload (f5).
Wherefore this PR also adds a mechanism to trigger: i) an increase of the requested page; 2) clearing the current error.

### Benefits

 A live example is depicted below:

tryAgainReq

### Possible drawbacks

This approach just assumes there is a problem on a given page and we force to request the next one. As a result, the failing page will remain unrequested. 

### Applicable issues

  - related https://github.com/kubeapps/kubeapps/issues/2484

### Additional information

Perhaps 'try again' is not the best message (since we do discard the failing page), see what you think.

### DIFF
--- a/dashboard/src/components/Catalog/Catalog.tsx
+++ b/dashboard/src/components/Catalog/Catalog.tsx
@@ -264,7 +264,7 @@ function Catalog(props: ICatalogProps) {
 
   const forceRetry = () => {
     dispatch(actions.charts.clearErrorChart());
-    increaseRequestedPage();
+    fetchCharts(cluster, namespace, reposFilter, page, size, searchFilter);
   };
 
   const increaseRequestedPage = () => {
--- a/dashboard/src/components/Catalog/Catalog.tsx
+++ b/dashboard/src/components/Catalog/Catalog.tsx
@@ -1,5 +1,6 @@
 import { CdsButton } from "@cds/react/button";
 import { CdsIcon } from "@cds/react/icon";
+import actions from "actions";
 import FilterGroup from "components/FilterGroup/FilterGroup";
 import Alert from "components/js/Alert";
 import Column from "components/js/Column";
@@ -256,8 +257,20 @@ function Catalog(props: ICatalogProps) {
 
   // Required to have the latest value of page
   const setPageWithContext = () => {
+    if (!error) {
+      increaseRequestedPage();
+    }
+  };
+
+  const forceRetry = () => {
+    dispatch(actions.charts.clearErrorChart());
+    increaseRequestedPage();
+  };
+
+  const increaseRequestedPage = () => {
     setPage(page + 1);
   };
+
   const observeBorder = (node: any) => {
     // Check if the IntersectionAPI is enabled
     if ("IntersectionObserver" in window && "IntersectionObserverEntry" in window) {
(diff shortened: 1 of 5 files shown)


absoludity:
> In this PR we stop requesting more pages if an error is detected. However, solely adding that has a severe impact on the UX: if request happens to fail, all the previously requested items will be discarded and the only way to retry is by performing a hard reload (f5).

Great that you avoided that. I don't have the full context but is there a reason we need or should be requesting further pages before we've successfully received the results of the previous ones? I didn't think people would scroll that fast - if we only ever requested one page at a time (serially) we wouldn't have "all the previously requested items will be discareded" right? Or maybe I've misunderstood. 

> Wherefore this PR also adds a mechanism to trigger: i) an increase of the requested page; 2) clearing the current error.

Could we just retry the request automatically some threshold of times before failing?

> ### Benefits
> 
> A live example is depicted below:
> 
> tryAgainReq

Thanks for the visual!

> ### Possible drawbacks
> 
> This approach just assumes there is a problem on a given page and we force to request the next one. As a result, the failing page will remain unrequested.

I don't see why we force the next one - why not re-request the current one that failed? (whether or not it's done automatically).


> ### Applicable issues
> 
>     * related #2484
> 
> 
> ### Additional information
> 
> Perhaps 'try again' is not the best message (since we do discard the failing page), see what you think.

Hah - yes, though I think we shouldn't discard the failing page. We know there's not going to be a reason for a particular page to fail and not others.

absoludity (approved):
+1 as I think this is better than what we have, but I do wonder whether we should be requesting the pages serially (looks like it's requesting further pages before a successful response for the current). See what you think.

dashboard/src/components/Catalog/Catalog.tsx
```diff
@@ -256,8 +257,20 @@ function Catalog(props: ICatalogProps) {
 
   // Required to have the latest value of page
   const setPageWithContext = () => {
+    if (!error) {
```
andresmgot, line 260: as far as I can tell, just checking this error is enough to fix this issue.

andresmgot (commented):
Thanks @antgamdia. I think we are fixing an error that doesn't exist here though. You are doing two things:

 - If a page request fails, don't request the next one.
 - Smooth recovery path in one (of many) pages fail.
 
IMO the second one is not really needed. It's highly improbable that if a page success to be retrieved (e.g page 1) the next one can fail. Also, as far as I can tell, we are not re-requesting the same page so we would be leaving the catalog in a inconsistent way.

I would just show the error and let people refresh the whole page (since the most probably page that can fail is the first one) but if you want to keep it this path, I would adapt this code to retry the same page failing.

antgamdia:
> we wouldn't have "all the previously requested items will be discarded" right? 

Yes, currently the redux reducer is setting `items: []` if an error occurs. This PR changes this behavior to be `items: state.items` 

> why not re-request the current one that failed? 

You're right. I make the assumption that just a specific page is failing. However, it doesn't really make sense, as 'it's highly improbable that if a page success to be retrieved (e.g page 1) the next one can fail'.
So, I have changed the implementation to perform a re-fetch of the failing page.


> I would just show the error and let people refresh the whole page (since the most probably page that can fail is the first one) but if you want to keep it this path, I would adapt this code to retry the same page failing.

Mmm, I don't have a strong opinion here. My point on going this way is: supposing a user is fetching a repo with several pages, and, because of an eventual network issue, the last request goes wrong. As this user, I'd like to have a way to recover from this error. But, also, I agree with you, this is very unlikely to happen.

So, to sum up. I would leave this fallback mechanism implemented since users are still able to just refresh the page.

andresmgot (approved):

PR #2515, opened 2021-03-05
Make pinniped-proxy have configurable API group suffixes and global/namespaced API calls

### LINKED ISSUE #2426: pinniped 0.6.0 has backwards incompatible change
### Description:

When running Kubeapps with pinniped-proxy with pinniped 0.6.0 installed on the cluster, the credential exchange fails with:

```
Caused by:
    err creating token exchange: "{\"apiVersion\":\"login.concierge.pinniped.dev/v1alpha1\",\"kind\":\"TokenCredentialRequest\",\"metadata\":{\"name\":\"\",\"namespace\":\"pinniped-concierge\"},\"spec\":{\"token\":\"...\",\"authenticator\":{\"apiGroup\":\"authentication.concierge.pinniped.dev\",\"kind\":\"JWTAuthenticator\",\"name\":\"jwt-authenticator\"}}}"
    ApiError: the server could not find the requested resource: NotFound (ErrorResponse { status: "Failure", message: "the server could not find the requested resource", reason: "NotFound", code: 404 })
 INFO  pinniped_proxy::service > GET https://kubernetes.default/ 500 Internal Server Error
```

It's not yet clear why this fails with 0.6.0 of pinniped while it works fine with 0.5.0, but we'll need to investigate further. Most likely it's related to the [backwards incompatible change mentioned on the releases notes](https://github.com/vmware-tanzu/pinniped/releases/tag/v0.6.0), though I have already tried re-creating the jwtauthenticator (as a cluster-scoped resource) without any change. It could be that the `TokenCredentialRequest` is also cluster-scoped in which case the request from pinniped-proxy will need updating.

antgamdia:
### Description of the change

This PR adds the "PINNIPED_VERSION" and decides with API endpoint should be called: the namespace one or the global.

### Benefits

Pinniped-proxy will work on >= 0.6.0 versions

### Possible drawbacks

N/A

### Applicable issues

  - fixes #2426 

### Additional information

Currently, we have hardcoded "pinniped.dev" API suffix, it will prevent people using `--api-group-suffix` flag to get pinniped-proxy working, I think.

Missing tests, but I still have to figure out how to write them :P 

(creating as a draft since I want to give it an additional try, but,  at the same time, I wanted to share the progress with you)


cmd/pinniped-proxy/src/pinniped.rs
```diff
@@ -131,7 +133,17 @@ async fn call_pinniped_exchange(authorization: &str, k8s_api_server_url: &str, k
         Some(a) => a.to_string(),
         None => authorization.to_string(),
     };
-    let token_creds: Api<TokenCredentialRequest> = Api::namespaced(client.clone(), &pinniped_namespace);
+    
+    let pinniped_version = env::var(PINNIPED_VERSION)?;
```
absoludity, line 137: So this is a short-term issue that we have until TMC updates (it may have already, I haven't checked). I was going to suggest that we default `PINNIPED_VERSION` to `0.6.0` so that if it's not set, we assume the future stable behavior rather than erroring if it's not set?

Although I think that would be fine, it may be clearer to use a more specific flag `NAMESPACED_TOKEN_CREDENTIAL_REQUESTS` and default that to false if not present.

But realistically, I don't think we need to support both. Given pinniped's state and our use of it, I think it'll be fine for us to update to the non-namespaced client and just say that 0.6.0 is the minimum supported version. What do you think?

cmd/pinniped-proxy/src/pinniped.rs
```diff
@@ -79,7 +81,7 @@ fn identity_for_exchange(cred: &ClusterCredential) -> Result<Identity> {
 /// The rust derive macro together with the kube macro creates serializable and deserializable
 /// resources based on the struct. See https://docs.rs/kube/0.43.0/kube/ for more details.
 #[derive(CustomResource, Deserialize, Serialize, Clone, Debug)]
-#[kube(group = "login.concierge.pinniped.dev", version = "v1alpha1", kind = "TokenCredentialRequest", namespaced)]
```
absoludity, line 82: Did you try with 0.5.0 without this `namespaced` option here? I am surprised that it would work without it (in that I assume this macro derives either a namespaced or non-namespaced resource). Ah, or maybe it just determines whether the metadata *must* have a namespace set to be valid or something.

cmd/pinniped-proxy/src/pinniped.rs
```diff
@@ -131,7 +133,17 @@ async fn call_pinniped_exchange(authorization: &str, k8s_api_server_url: &str, k
...
+    
+    let pinniped_version = env::var(PINNIPED_VERSION)?;
+    let token_creds: Api<TokenCredentialRequest>;
+    
+    // Pinniped >= 0.6.0 changed the APIs to be global instead of namespaced
+    if Version::parse(&pinniped_version) >  Version::parse("0.5.0") {
+        token_creds = Api::all(client.clone());
+    }else{
+        token_creds = Api::namespaced(client.clone(), &pinniped_namespace);
+    }
```
absoludity, line 145: If we do go ahead with this change and so it needs testing, I'd recommend just pulling it the functionality that you're adding and just unit-testing that:
```
fn get_token_credential_request() -> ...
```
if there's fields on the result which can be used, as I think I'm only testing the failure cases for this function atm (haven't looked for the equivalent of `httptest` for staging test requests.

absoludity (commented):
Excellent, thanks for investigating Antonio!

cmd/pinniped-proxy/src/pinniped.rs
antgamdia: I have checked it with TMC folks; good and bad news here:
  1. The future stable behavior is the non-namespaced API calls, so we can just require >=0.6.0. Also, TMC is rolling out the new versions automatically, so we should not be worried about it.
  2. But... they will leverage from the `api-group-suffix` already introduced in 0.5.0. And it does have an impact for us.
 
As per 2., we don't have any option but to allow dynamic API groups (TMC will have a special one, whereas pinniped remains to be "pinniped.dev").
In the current update of this PR I'm introducing some logic to support that. However, we should change the way in which we generate the CRD struct. We cannot derive using a runtime var, so the struct creation has to be more complex.

As far as I have read, we have to create a CustomResource manually and implement some traits manually.
Right now I'm a bit stuck here, let's see tomorrow :P 

https://github.com/clux/kube-rs/blob/master/examples/crd_api.rs
https://docs.rs/kube-derive/0.51.0/kube_derive/derive.CustomResource.html


Now, answering your original response: initially, I'd go for the idea of dropping support for <0.6. However, since we have to change the CRD, one more flag is not a big deal and we, therefore, separate the Pinniped-Proxy's lifecycle from the TMC one.

cmd/pinniped-proxy/src/pinniped.rs
antgamdia: I wanted to try, but I don't know why I've been facing the same issue of the missing cookie (#-2505). My requests never reached the pinniped proxy... and I'm using the same files as I was using the last week...

cmd/pinniped-proxy/src/pinniped.rs
antgamdia: I have extracted the logic to a function, but I'm unable to test it properly. I was thinking in something like:

```rust
let actual_tcr = get_token_credential_request(client, pinniped_namespace);
let expected_tcr: Api<TokenCredentialRequest> = Api::all(client.clone());

assert_eq!(actual_tcr, expected_tcr);
```

However, it complains about something related to tokio and I had to introduce `#[cfg(feature = "rt-threaded")]` to pass the test. But, honestly, I dunno what is doing :S

cmd/pinniped-proxy/src/pinniped.rs
absoludity: > As per 2., we don't have any option but to allow dynamic API groups (TMC will have a special one, whereas pinniped remains to be "pinniped.dev").
> In the current update of this PR I'm introducing some logic to support that. However, we should change the way in which we generate the CRD struct. We cannot derive using a runtime var, so the struct creation has to be more complex.
> 
> As far as I have read, we have to create a CustomResource manually and implement some traits manually.
> Right now I'm a bit stuck here, let's see tomorrow :P

Yeah, so it's unfortunate (for us) that they re-compile pinniped for tmc with different options, and that prefix is a build-time var for us. Interested to see what we need here, but also happy to jump in if you want.

> 
> https://github.com/clux/kube-rs/blob/master/examples/crd_api.rs
> https://docs.rs/kube-derive/0.51.0/kube_derive/derive.CustomResource.html
> 
> Now, answering your original response: initially, I'd go for the idea of dropping support for <0.6. However, since we have to change the CRD, one more flag is not a big deal and we, therefore, separate the Pinniped-Proxy's lifecycle from the TMC one.

Yeah, OK.

cmd/pinniped-proxy/src/pinniped.rs
absoludity: So this is only required because in the test you're calling `Client::try_default().await?;` which is an async function (which you're waiting for). I did a bit of a refactor so that you can create a client the same way the code above does (which isn't async), with:

```diff
--- a/cmd/pinniped-proxy/src/pinniped.rs
+++ b/cmd/pinniped-proxy/src/pinniped.rs
@@ -120,24 +120,28 @@ pub struct ClusterCredential {
     client_key_data: String,
 }
 
-/// call_pinniped_exchange returns the resulting TokenCredentialRequest with Status after requesting a token credential exchange.
-async fn call_pinniped_exchange(authorization: &str, k8s_api_server_url: &str, k8s_api_ca_cert_data: &[u8]) -> Result<TokenCredentialRequest> {
-    let pinniped_namespace = env::var(DEFAULT_PINNIPED_NAMESPACE)?;
-
+fn get_client_config(k8s_api_server_url: &str, k8s_api_ca_cert_data: &[u8], pinniped_namespace: String) -> Result<kube::Client> {
     let mut config = Config::new(Url::parse(k8s_api_server_url).context("Failed parsing url for exchange")?);
     config.default_ns = pinniped_namespace.clone();
     let x509 = X509::from_pem(k8s_api_ca_cert_data).context("error creating x509 from pem")?;
     let der = x509.to_der().context("error creating der from x509")?;
     config.root_cert = Some(vec!(der));
-    let client = Client::new(Service::try_from(config)?);
+
+    Ok(Client::new(Service::try_from(config)?))
+}
+
+/// call_pinniped_exchange returns the resulting TokenCredentialRequest with Status after requesting a token credential exchange.
+async fn call_pinniped_exchange(authorization: &str, k8s_api_server_url: &str, k8s_api_ca_cert_data: &[u8]) -> Result<TokenCredentialRequest> {
+    let pinniped_namespace = env::var(DEFAULT_PINNIPED_NAMESPACE)?;
+    let client = get_client_config(k8s_api_server_url, k8s_api_ca_cert_data, pinniped_namespace.clone())?;
 
     let auth_token = match authorization.to_string().strip_prefix("Bearer ") {
         Some(a) => a.to_string(),
         None => authorization.to_string(),
     };
```

With that I could update the test so it wasn't async at all, but unfortunely (for us) the item that we'd need to test in the client is `client.resource` which is a private member on client (sorry, hadn't checked that). The best way to test this, like any other language, would be to use a fixture which starts up a fake api server and records the URLS, but that's not something for this PR or something we should do now, imo. I'd just remove this test.

cmd/pinniped-proxy/src/pinniped.rs
```diff
@@ -152,6 +158,25 @@ async fn call_pinniped_exchange(authorization: &str, k8s_api_server_url: &str, k
         },
     }
 }
+    
+fn get_token_credential_request(client: Client, pinniped_namespace: &str) -> Api<TokenCredentialRequest> {
+    let ns_token_cred_req: bool = env::var(NAMESPACED_TOKEN_CREDENTIAL_REQUESTS).unwrap_or("false".into()).parse().unwrap();
```
absoludity, line 163: Note that this line can cause the app to panic, because you're calling `.unwrap()` which means to unwrap the result and panic if it's an error. More [info in the docs](https://doc.rust-lang.org/edition-guide/rust-2018/error-handling-and-panics/the-question-mark-operator-for-easier-error-handling.html). What you should instead be doing is using the `?` operator which effectively says return an error if there is one, otherwise unwrap the result, and update the function return type to be a result
```
fn get_token_credential_request(client: Client, pinniped_namespace: &str) -> Result<Api<TokenCredentialRequest>> {
    let ns_token_cred_req: bool = env::var(NAMESPACED_TOKEN_CREDENTIAL_REQUESTS).unwrap_or("false".into()).parse()?;
...
    return Ok(Api::...);
}
```

antgamdia: Mmm, interesting. However, I thought it wouldn't be required since `.unwrap_or()` fallback  to `false` and `"false".parse()` should always be `false`.

absoludity: Oh sure, yes if you are certain that the unwrap there can never result in an error. I'd not realised that as the case. Perfect as is then :)

chart/kubeapps/templates/kubeapps-frontend-deployment.yaml
```diff
@@ -113,6 +113,12 @@ spec:
               value: {{ .Values.pinnipedProxy.defaultAuthenticatorType }}
             - name: DEFAULT_PINNIPED_AUTHENTICATOR_NAME
               value: {{ .Values.pinnipedProxy.defaultAuthenticatorName }}
+            - name: DEFAULT_PINNIPED_API_SUFFIX
+              value: {{ .Values.pinnipedProxy.defaultAuthenticatorName }}
+          {{- if and .Values.pinnipedProxy.defaultPinnipedAPISuffix }}
```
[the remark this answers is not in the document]
antgamdia: Wow, I wasn't aware it has been committed like this...I just messed up with the git stash, I guess. Updating a corrected version.

antgamdia:
Some wip changes added:

I used a `cargo-expand`  (https://github.com/dtolnay/cargo-expand) to generate the required "things" so that we can change later the things we need.
See https://github.com/Arnavion/k8s-openapi/blob/master/k8s-openapi-derive/src/lib.rs#L77 

I haven't tested anything, I just committed once it finally compiled. Now it's time to call it a day :P 
If you have any suggestions, they are more than welcomed.

And... uff, it's such a nightmare performing simple things (eg joining strings). It has &str, String, &'static str, and I don't know how many more. These concepts of ownership in variables is kinda weird

absoludity (commented):
> I used a `cargo-expand` (https://github.com/dtolnay/cargo-expand) to generate the required "things" so that we can change later the things we need.
> See https://github.com/Arnavion/k8s-openapi/blob/master/k8s-openapi-derive/src/lib.rs#L77

See my comment inline. Let's see if we can avoid this for now. Originally we thought this change would be small - and maybe it would be if we stuck to the original scope which was just updating the namespaced from 0.5.0 to the non-namespaced of 0.6.0? Let's just keep to that and land that?

We can then consider options for the group and I've the one simpler suggestion here. I'm happy to pick that up if too, if you'd prefer.

> I haven't tested anything, I just committed once it finally compiled. Now it's time to call it a day

For sure!

> If you have any suggestions, they are more than welcomed.

See above :)

> 
> And... uff, it's such a nightmare performing simple things (eg joining strings).

What in particular do you mean? If I want to join any string (whether it's a `String` object or an `str` reference, I just use the `format!` macro and it just works, [like this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b661727c842beb4ca51f38b1e5e805c4).

Or do you mean joining an array/vector of strings with some character?

> It has &str, String, &'static str, and I don't know how many more.

There's only two types there: `str` and `String`. The `'static` is just a lifetime specifier telling the compiler that that variable should live for as long as the program itself (ie. don't free it when it goes out of scope, which is what the compiler will do by default for a stack-allocated var unless ownership is passed on to another piece of code). 

> These concepts of ownership in variables is kinda weird

Yep, thinking about ownership is the difficult part. In this case, an `str` is always passed by reference (ie. an `&str`, since it's just a pointer to a slice of memory somewhere, not a struct). Where as a `String` is an struct that handles pushing extra characters, joining and a bunch of other things, that can be owned or borrowed by others as you decide.

Thinking about ownership is hard at first - no question. It took me a while to get used to. The payoff is the compiler catches a lot of bugs (which can feel frustrating at first too - just compile dammit! :P) that you would otherwise only find at runtime, especially as you start doing concurrent things. Whether it's worth the extra learning and frustration at first is of course the big debate.

I'm assuming you've read [Understanding Ownership](https://doc.rust-lang.org/book/ch04-00-understanding-ownership.html), I had to read it again after playing with some code, until the concept of moving ownership clicked.

cmd/pinniped-proxy/src/pinniped.rs
antgamdia: (just dumping here the logs, later will edit the comment) 

` kubectl -n kube-system logs  -l component=kube-apiserver  -f`

```
E0311 11:51:45.229972       1 controller.go:116] loading OpenAPI spec for "v1alpha1.login.concierge.pinniped.dev" failed with: OpenAPI spec does not exist
I0311 11:51:45.229994       1 controller.go:129] OpenAPI AggregationController: action for item v1alpha1.login.concierge.pinniped.dev: Rate Limited Requeue.
E0311 11:51:51.340896       1 authentication.go:53] Unable to authenticate the request due to an error: invalid bearer token
```

absoludity: > (just dumping here the logs, later will edit the comment)
> 
> ` kubectl -n kube-system logs -l component=kube-apiserver -f`
> 
> ```
> E0311 11:51:45.229972       1 controller.go:116] loading OpenAPI spec for "v1alpha1.login.concierge.pinniped.dev" failed with: OpenAPI spec does not exist
> I0311 11:51:45.229994       1 controller.go:129] OpenAPI AggregationController: action for item v1alpha1.login.concierge.pinniped.dev: Rate Limited Requeue.
> E0311 11:51:51.340896       1 authentication.go:53] Unable to authenticate the request due to an error: invalid bearer token
> ```
> 
> Took 7 min to pull image from projects.registry.vmware.com
> 

Yeah, I found the same - took ages here (although I did suspend in between, perhaps causing problems).

And yes, I see the same error in my cluster api server logs (note you can just export them in kind too, which is sometimes nicer, with `kind export logs --name kubeapps` - gives you all logs in a tmp dir). This error is just because pinniped-concierge apparently isn't defining an OpenAPI spec (which apiservices are meant to do, apparently). ([Here's an example of a similar issue for prometheus operator](https://github.com/prometheus-operator/kube-prometheus/issues/304))

> Pinniped pod
> 
> ```
> E0311 12:48:25.794626       1 reflector.go:138] /cache/gomodcache/k8s.io/client-o@v0.20.1/tools/cache/reflector.go:167: Failed to watch *v1beta1.FlowSchema: failed to list *v1beta1.FlowSchema: flowschemas.flowcontrol.apiserver.k8s.io is forbidden: User "system:serviceaccount:pinniped-concierge:pinniped-concierge" cannot list resource "flowschemas" in API group "flowcontrol.apiserver.k8s.io" at the cluster scope
> ```

Also harmless, I see this too (in my working env). I guess this is because `flowschemas` are a part of the [API priority and fairness feature which is only available in k8s v1.20](https://kubernetes.io/docs/concepts/cluster-administration/flow-control/) and my Kind cluster is currently using 1.18 (yes I should update). But that's just a guess.

> 
> Bad certificate, I copypasted bad
> 
> ```
> E0311 13:09:04.884712       1 oidc.go:224] oidc authenticator: initializing plugin: Get "https://172.19.0.2:32000/.well-known/openid-configuration": x509: certificate signed by unknown authority
> E0311 13:09:12.731575       1 runtime.go:76] Observed a panic: runtime error: invalid memory address or nil pointer dereference
> k8s.io/apiserver/pkg/server/filters.(*timeoutHandler).ServeHTTP.func1.1(0xc0009d7080)
>         /cache/gomodcache/k8s.io/apiserver@v0.20.1/pkg/server/filters/timeout.go:106 +0x113 panic(0x1bd5380, 0x2d74ec0)
> ```

Do you mean that you fixed the above because it was a copy-paste error?

> 
> Namespaced authenticator in 0.6.0,
> 
> ```
> I0311 13:27:15.625383       1 certs_observer.go:66] certsObserverController Sync updated certs in the dynamic cert provider
> I0311 13:27:15.625745       1 certs_expirer.go:88] certsExpirerController Sync found a renew delta of -599h49m34.374266991s
> E0311 13:27:27.132579       1 runtime.go:76] Observed a panic: runtime error: invalid memory address or nil pointer dereference
> ```

Do you mean the above are logs from a pinniped-concierge 0.6.0 pod? Unsure about this one as I see the following in my 0.6.0 pod logs:

```
I0312 04:01:27.976314       1 certs_observer.go:66] certsObserverController Sync updated certs in the dynamic cert provider
I0312 04:01:28.061193       1 certs_expirer.go:88] certsExpirerController Sync found a renew delta of -599h59m48.938814425s
I0312 04:01:28.061526       1 certs_manager.go:136] certsManagerController Sync successfully created secret
```

> 
> Recreate the entire cluster again
> 
> ```
>    1 clientconn.go:948] ClientConn switching balancer to "pick_first"
> E0311 16:00:39.151159       1 controller.go:116] loading OpenAPI spec for "v1alpha1.login.concierge.pinniped.dev" failed with: failed to retrieve openAPI spec, http error: ResponseCode: 503, Body: error trying to reach service: dial tcp 10.96.251.99:443: connect: connection refused
> , Header: map[Content-Type:[text/plain; charset=utf-8] X-Content-Type-Options:[nosniff]]
> I0311 16:00:39.151178       1 controller.go:129] OpenAPI AggregationController: action for item v1alpha1.login.concierge.pinniped.dev: Rate Limited Requeue.
> ...
> E0311 16:23:39.636478       1 controller.go:116] loading OpenAPI spec for "v1alpha1.login.concierge.pinniped.dev" failed with: OpenAPI spec does not exist
> I0311 16:23:39.636517       1 controller.go:129] OpenAPI AggregationController: action for item v1alpha1.login.concierge.pinniped.dev: Rate Limited Requeue.
> ```

Also unrelated, I think? The first one is just the apiserver failing to even reach the `v1alpha1.login.concierge.pinniped.dev` apiservice at first, because the pods wouldn't have been running (you can check how k8s views the api service with `kubectl get apiservices` and see there whether it thinks it's running or not). And the second one is just the same as earlier - the pinniped apiservice apparently doesn't include an OpenAPI spec yet.

antgamdia: > (just dumping here the logs, later will edit the comment)

Whoops, I forgot to update the comments with more descriptive messages! Sorry about that!

> And yes, I see the same error in my cluster api server logs. This error is just because pinniped-concierge apparently isn't defining an OpenAPI spec (which apiservices are meant to do, apparently). ([Here's an example of a similar issue for prometheus operator](https://github.com/prometheus-operator/kube-prometheus/issues/304))

Ahh, ok, I see. I thought the resources weren't being properly installed at all. Thanks for the clarification here.

> (note you can just export them in kind too, which is sometimes nicer, with `kind export logs --name kubeapps` - gives you all logs in a tmp dir). 

Not aware of this feature. Thanks!

> Also harmless, I see this too (in my working env). I guess this is because `flowschemas` are a part of the [API priority and fairness feature which is only available in k8s v1.20](https://kubernetes.io/docs/concepts/cluster-administration/flow-control/) and my Kind cluster is currently using 1.18 (yes I should update). But that's just a guess.

I use the latest Kind version with k8s v1.20.2, but I agree, it should be harmless.

> Do you mean that you fixed the above because it was a copy-paste error?

Yes, I copied  the `client-certificate-data` instead of `certificate-authority-data` :disappointed:  

> Do you mean the above are logs from a pinniped-concierge 0.6.0 pod? Unsure about this one as I see the following in my 0.6.0 pod logs:

Yes, they are; from time to time I continue having some panics even if it is working now.

> Also unrelated, I think? The first one is just the apiserver failing to even reach the `v1alpha1.login.concierge.pinniped.dev` apiservice at first, because the pods wouldn't have been running (you can check how k8s views the api service with `kubectl get apiservices` and see there whether it thinks it's running or not). 

Yep, you're right. Thanks for the `kubectl get apiservices` command, really useful.

So, finally, the problem was an OSI Layer 8 one :P  I was omitting one step, copying the cert from the additional cluster in the Kubeapps values. I noticed it after your changes in the PR, so thanks a lot for the help!

cmd/pinniped-proxy/src/pinniped.rs
```diff
@@ -78,26 +77,199 @@ fn identity_for_exchange(cred: &ClusterCredential) -> Result<Identity> {
 ///
 /// The rust derive macro together with the kube macro creates serializable and deserializable
 /// resources based on the struct. See https://docs.rs/kube/0.43.0/kube/ for more details.
-#[derive(CustomResource, Deserialize, Serialize, Clone, Debug)]
-#[kube(group = "login.concierge.pinniped.dev", version = "v1alpha1", kind = "TokenCredentialRequest", namespaced)]
-#[kube(status = "TokenCredentialRequestStatus")]
+
+
+// From https://github.com/clux/kube-rs/blob/master/kube-derive/src/lib.rs
```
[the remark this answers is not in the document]
antgamdia: > here's only two types there: `str` and `String`. The `'static` is just a lifetime specifier telling the compiler that that variable should live for as long as the program itself (ie. don't free it when it goes out of scope, which is what the compiler will do by default for a stack-allocated var unless ownership is passed on to another piece of code).
> 
> Yep, thinking about ownership is the difficult part. In this case, an `str` is always passed by reference (ie. an `&str`, since it's just a pointer to a slice of memory somewhere, not a struct). Where as a `String` is an struct that handles pushing extra characters, joining and a bunch of other things, that can be owned or borrowed by others as you decide.

I'm getting the point. After having a look at it again, the problem was not a type problem, but a misunderstanding between `static` and `const`. The latter is constant but at compile time! So I was trying to join string, but it is not supported.
Indeed, I think here is where the original problem (use the derive macro) strikes back...

```rust
// It is required to implement this :(
impl k8s_openapi::Resource for TokenCredentialRequest {
    // Not compiling
    const API_VERSION: &'static str = format!(&get_pinniped_login_api_group(), &DEFAULT_VERSION).as_str();
    // Not compiling
    const GROUP: &'static str = &get_pinniped_login_api_group();
    const KIND: &'static str = "TokenCredentialRequest";
    const VERSION: &'static str = &DEFAULT_VERSION;
}
```
`calls in constants are limited to constant functions, tuple structs and tuple variants.`

I have taken a look at https://crates.io/crates/lazy_static, but still not working...
Another idea is to use the cargo build scripts, but the problem is yet the same... not able to change these things in runtime.
I'm running out of new approaches to deal with that; I'm starting to think the best option is just to support a fixed number of API suffixes as in #2536 :S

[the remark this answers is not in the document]
absoludity: > I'm running out of new approaches to deal with that; I'm starting to think the best option is just to support a fixed number of API suffixes as in #2536 :S

Yes, I thinkn the two API suffixes is fine for now. I'd like to have a look at generalising it when I have time, but it's not urgent atm.

absoludity:
So I think this PR can be closed in favour of your other one with static groups?

antgamdia:
> So I think this PR can be closed in favour of your other one with static groups?

Yes, I do agree. At least I've learned a bit about how Rust macros work and the different data types. Let's close it atm.
Thank you for your comments and help!

PR #2521, opened 2021-03-09
Add filter field in AppRepo form

### LINKED ISSUE #1583: filter packages

andresmgot:
### Description of the change

Add the possibility to filter packages from the AppRepo form. Design [here](https://docs.google.com/document/d/1dNv7Id0l_3sac2SAtswtn3vj2eq6naFPGf9-YmN-3JE/edit?ts=604195f5)

Note that, for simplicity, only filtering based on the name is exposed. Because of that, it doesn't make sense to show the filter for the OCI case, since it's already mandatory to add a list of applications there.

Check the docs added for an example.

### Applicable issues

  - fixes #1583 

### Additional information

I noticed some minor things in the backend that were required to send the form with the new field.


cmd/apprepository-controller/pkg/apis/apprepository/v1alpha1/types.go
```diff
@@ -47,12 +47,11 @@ type AppRepositorySpec struct {
 	DockerRegistrySecrets []string `json:"dockerRegistrySecrets,omitempty"`
 	// In case of an OCI type, the list of repositories is needed
 	// as there is no API for the index
-	// DEPRECATED: Use FilterRules instead
 	OCIRepositories []string `json:"ociRepositories,omitempty"`
 	// TLSInsecureSkipVerify skips TLS verification
 	TLSInsecureSkipVerify bool `json:"tlsInsecureSkipVerify,omitempty"`
 	// FilterRule allows to filter packages based on a JQuery
-	FilterRule FilterRuleSpec `json:"filterRules,omitempty"`
+	FilterRule FilterRuleSpec `json:"filterRule,omitempty"`
```
antgamdia, line 54: API docs nitpicky guy here :P reminder: we should also update the openapi accordingly

dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
```diff
@@ -153,6 +162,23 @@ export function AppRepoForm(props: IAppRepoFormProps) {
...
+    if (type === TYPE_HELM && filterNames !== "") {
+      if (filterRegex) {
+        filter = { jq: ".name | test($var)", variables: { $var: filterNames } };
+      } else {
+        const names = filterNames.split(",").map(n => n.trim());
+        const variables = names.reduce((acc, n, i) => {
+          acc[`$var${i}`] = n;
+          return acc;
+        }, {});
+        const jq = names.map((v, i) => `.name == $var${i}`).join(" or ");
```
antgamdia, line 175: Not sure, but perhaps does it worth extracting this logic to a function in charge of generating the `jq` ?

dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
```diff
@@ -219,6 +246,18 @@ export function AppRepoForm(props: IAppRepoFormProps) {
     setSkipTLS(!skipTLS);
     setValidated(undefined);
   };
+  const handleFilterNames = (e: React.ChangeEvent<HTMLTextAreaElement>) => {
+    setFilterNames(e.target.value);
+    setValidated(undefined);
```
antgamdia, line 251: Out of curiosity, why do we need these `setValdiated` ? Which validation is being performed in these cases?

docs/user/private-app-repository.md
```diff
@@ -34,6 +34,35 @@ When Kubeapps deploys any chart from this AppRepository, if a referenced docker
 
 There will be further work to enable private AppRepositories to be available in multiple namespaces. Details about the design can be read on the [design document](https://docs.google.com/document/d/1YEeKC6nPLoq4oaxs9v8_UsmxrRfWxB6KCyqrh2-Q8x0/edit?ts=5e2adf87).
 
+## Filter applications
+
+Since Kubeapps 2.3, it's possible to limit the amount of packages that Kubeapps will expose to the users of an Application Repository. For doing so, edit the filter field of the form:
```
antgamdia, line 39: s/the amount of packages/the number of packages

docs/user/private-app-repository.md
```diff
@@ -34,6 +34,35 @@ When Kubeapps deploys any chart from this AppRepository, if a referenced docker
...
+
+<img src="../img/apprepo-form-filter.png" alt="AppRepository with filter" width="600px">
+
+- If "Exclude Packages" is marked, all the applications matching the given names will be excluded.
+- If "Regular Expression" is marked, rather than treating the input as a comma-separated list of application names, it will be treated as a regex. This can be used when the name of the application is not known or when you want to filter different applications with the same name.
```
antgamdia, line 44: I'd rather say `(...) treated as a javascript regex` since some minor things are not supported in js regex (such as recursion) but they are in PCRE.

docs/user/private-app-repository.md
```diff
@@ -34,6 +34,35 @@ When Kubeapps deploys any chart from this AppRepository, if a referenced docker
...
+
+### Advanced filtering
+
+> **NOTE**: This is not supported from the Kubeapps Dashboard.
```
antgamdia, line 48: s/supported from/supported by

docs/user/private-app-repository.md
```diff
@@ -34,6 +34,35 @@ When Kubeapps deploys any chart from this AppRepository, if a referenced docker
...
+
+In case you want to add a custom filter, based on a metadata field different than the name, it's possible to specify a [jq](https://stedolan.github.io/jq/) query to filter applications. This is only available when manually creating the AppRepository manifest. In this example, we are filtering applications that contains "Bitnami" as one of the maintainers:
```
antgamdia, line 50: s/that contains/that contain

antgamdia (approved):
Great, thank you for your work here (and the AppRepo refactor :P ). I've just added a couple of minor comments, typos and doubts, but LGTM.

cmd/apprepository-controller/pkg/apis/apprepository/v1alpha1/types.go
andresmgot: true, I think this field is not even documented. I am not sure how to work with the new approach, can you document how to modify the API docs?

antgamdia: Sure, I'll do it as soon as I finish the pinniped thing. Thanks for the suggestion.

Briefly, in the meantime, it is just a matter of accessing  https://app.swaggerhub.com/apis/kubeapps/Kubeapps/, modifying the field in the yaml view, click on save, click on sync, click on push and the changes will appear on the `openapi` branch. Then you can just send the PR and that's all.

dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
andresmgot: Right, copy-paste error here, we are not validating anything regarding the filter.

docs/user/private-app-repository.md
andresmgot: TBH, I am not sure what type of regexes jq uses (probably regex, but I would need to check).

andresmgot: so they are PCRE regexes: https://stedolan.github.io/jq/manual/#RegularexpressionsPCRE

antgamdia: Great, thanks for checking it :)

PR #2527, opened 2021-03-10
Adapt e2e test selectors to both versions of kubeapps

andresmgot:
### Description of the change

The current test `create-private-registry` is failing when used against the latest release of kubeapps (2.2) (https://app.circleci.com/pipelines/github/kubeapps/kubeapps/2334/workflows/1ee87892-79a9-42c5-8105-11e765540321/jobs/39920) due to the latest changes in the AppRepo form. This PR adapts the test so it can be used against master (or 2.3) and 2.2.

### Benefits

Fixes master

### Possible drawbacks

We'd need to remember to remove the workaround (try/catch) once we release 2.3.

### DIFF
--- a/integration/use-cases/create-private-registry.js
+++ b/integration/use-cases/create-private-registry.js
@@ -16,31 +16,41 @@ test("Creates a private registry", async () => {
 
   const randomNumber = Math.floor(Math.random() * Math.floor(100));
   const repoName = "my-repo-" + randomNumber;
-  await page.type("cds-form-group > cds-input:nth-child(1) > input", repoName);
+  await page.type("input[placeholder=\"example\"]", repoName);
 
   await page.type(
-    "cds-form-group > cds-input:nth-child(2) > input",
+    "input[placeholder=\"https://charts.example.com/stable\"]",
     "http://chartmuseum-chartmuseum.kubeapps:8080"
   );
 
   await expect(page).toClick("label", { text: "Basic Auth" });
 
   // Credentials from e2e-test.sh
-  await page.type("cds-form-group > div:nth-child(3) > div > div:nth-child(1) > cds-input:nth-child(1) > input", "admin");
-  await page.type("cds-form-group > div:nth-child(3) > div > div:nth-child(1) > cds-input:nth-child(3) > input", "password");
+  await page.type("input[placeholder=\"Username\"]", "admin");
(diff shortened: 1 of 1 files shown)


integration/use-cases/create-private-registry.js
```diff
@@ -16,31 +16,41 @@ test("Creates a private registry", async () => {
...
-  await page.type("cds-form-group > div:nth-child(3) > div > div:nth-child(1) > cds-input:nth-child(3) > input", "password");
+  await page.type("input[placeholder=\"Username\"]", "admin");
+  await page.type("input[placeholder=\"Password\"]", "password");
 
   // Open form to create a new secret
   const secret = "my-repo-secret" + randomNumber;
-  await expect(page).toClick(".btn-info-outline", { text: "Add new credentials" });
-  await page.type("cds-form-group > div.docker-creds-subform > cds-input:nth-child(2) > input", secret);
+  try {
+    // TODO(andresmgot): Remove this line once 2.3 is released
```
antgamdia, line 35: I don't know, but couldn't we do something like: 

`await expect(page).toClick('button', { text: '"Add new credentials' })`

Or 

`await expect(page).toClick({type:'xpath', value:'<xpath to the button>'}, { text: '"Add new credentials' })`

integration/use-cases/create-private-registry.js
```diff
@@ -16,31 +16,41 @@ test("Creates a private registry", async () => {
 
   const randomNumber = Math.floor(Math.random() * Math.floor(100));
   const repoName = "my-repo-" + randomNumber;
-  await page.type("cds-form-group > cds-input:nth-child(1) > input", repoName);
+  await page.type("input[placeholder=\"example\"]", repoName);
```
antgamdia, line 19: Great, much better !

antgamdia (approved):
Thank you for the changes !!

integration/use-cases/create-private-registry.js
andresmgot: yes, that's what I tried but the first doesn't work because:

first option: The cds-button is the one containing the text, and then there is a hidden, empty button. Something like `<cds-button>Add new credentials<button></button><cds-button>`

second option: The path changes from one version to the other, I could not find a path common to those two.

antgamdia: Ah, I see. No problem, the try-catch hotfix is ok though!

PR #2529, opened 2021-03-10
More fixes for the compatibility of e2e tests

andresmgot:
### Description of the change

Follow up of #2527

Add more fixes to be able to run e2e tests with both the latest release and master.

This time I am modifying the CircleCI manifest to ensure that the latest release passes the test before merging.

### Benefits

Fixes master.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -323,6 +323,8 @@ jobs:
       HELM_VERSION: "v3.1.0"
       KUBEAPPS_DB: "postgresql"
       TEST_UPGRADE: "1"
+      # TODO: Revert this once the latest version is fixed
+      TEST_LATEST_RELEASE: 1
     parameters:
       number:
         type: string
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -323,6 +323,8 @@ jobs:
       HELM_VERSION: "v3.1.0"
       KUBEAPPS_DB: "postgresql"
       TEST_UPGRADE: "1"
+      # TODO: Revert this once the latest version is fixed
+      TEST_LATEST_RELEASE: 1
```
antgamdia, line 327: What's the purpose of this env var? I don't see it's being used in this PR, is it?

antgamdia (approved):
Our beloved CI strikes back :P 
Thanks for the changes!

.circleci/config.yml
andresmgot: Yes, this flag is used in the line 163. Basically, if set, we will use the latest tag as image tags for Kubeapps (rather than the image just built). This way we can test the latest release with the current chart (useful for releasing a chart without releasing a new version of kubeapps).

antgamdia: Ahhh, got it! I thought it was a new thing in this PR. Thanks for the explanation :)

PR #2534, opened 2021-03-11
Dark mode for Changes tab

andresmgot:
### Description of the change

5./ at https://github.com/kubeapps/kubeapps/issues/2371#issuecomment-790615176

The changes tab requires to be rendered in a dark mode with the rest of the application. The current library (diff2html) doesn't support a dark-mode so I have migrated this to the package `react-diff-viewer`. This is how it looks like:

Light mode:

Screenshot from 2021-03-11 13-19-52

Dark mode:

Screenshot from 2021-03-11 13-19-40

### Applicable issues

  - ref #2371

### Additional information

I have removed the sub-title of the tab (it was mandatory before) but it was not giving too much information.


dashboard/src/components/DeploymentFormBody/Differential.tsx
```diff
@@ -1,37 +1,57 @@
...
+        gutterBackgroundDark: "#01313f",
+        addedGutterColor: "#d0edf7",
+        removedGutterColor: "#d0edf7",
+        // background
+        diffViewerBackground: "#002B36",
+        emptyLineBackground: "#002B36",
+        // fold text
+        codeFoldContentColor: "white",
+      },
+    },
```
antgamdia, line 36: Mmm, is it required to harcode the colors in the tsx? Isn't there any css approach? I guess no, but wanted to confirm anyway

dashboard/src/components/OperatorInstanceFormBody/OperatorInstanceFormBody.tsx
```diff
@@ -108,11 +108,6 @@ function DeploymentFormBody({
                 key="advanced-deployment-form"
               />,
               <Differential
-                title={
-                  deployedValues
-                    ? "Difference from deployed values"
-                    : "Difference from example defaults"
-                }
```
antgamdia, line 115: Great !

antgamdia (approved):
Thank you once more time for your efforts in the dark mode. The new diff-viewer looks great!

As I use to say in text-editors PRs... Monaco has also this feature :P 
Performing diffs:  https://microsoft.github.io/monaco-editor/playground.html#creating-the-diffeditor-multi-line-example

dashboard/src/components/DeploymentFormBody/Differential.tsx
andresmgot: the documented way of changing the style is through this `styles` object: https://github.com/praneshr/react-diff-viewer#overriding-styles

It may be possible to achieve the same through CSS but it's not documented. Also, while they should maintain the compatibilty of this API (the styles object) they don't need to do so for the CSS which means that if we use CSS only, it may break in future minor releases.

PR #2270, opened 2021-01-22
pinniped proxy dev env playground

absoludity:
### Description of the change

This is just local changes that I have to be able to run kubeapps in a non-oidc multi-cluster env.

```
make multi-cluster-kind

# Update docs/user/manifests/kumeapps-local-dev-additional-kind-cluster
# copying the certificate-authority-data from the additional cluster (~/.kube/kind-config-kubeapps-additional)
# to the certificateAuthorityData of the second cluster.

make deploy-dev
make deploy-pinniped-additional

# Edit docs/user/manifests/kubeapps-pinniped-jwt-authenticator.yaml
# with the certificate-authority-data from the main cluster (~/.kube/kind-config-kubeapps)
make add-pinniped-jwt-authenticator
```

PR #2536, opened 2021-03-11
Make pinniped-proxy compatible with 0.6.0 and TMC

### LINKED ISSUE #2426: pinniped 0.6.0 has backwards incompatible change

antgamdia:
### Description of the change

This PR adds support for Pinniped 0.6.0 (but with a **not-backward-compatible** change). It means Pinniped-proxy will only work in >=0.6.0.

Apart from some very minor refactor, there are two main changes:

1. API calls to `authentication.concierge....` and `login.concierge....` will be global instead of namespaced (this is the breaking change).
2. Shamefully ugly workaround consisting of duplicating code for supporting another API group rather than `pinniped.dev`. Specifically, this PR adds support for the one that will be used in Tanzu Mission Control.

### Benefits

1.  Pinniped >= 0.6.0 will be supported.
2.  Pinniped-proxy can be used in TMC.

### Possible drawbacks

Dropping support for Pinniped < 0.6.0

### Applicable issues

  - fixes #2426 

### Additional information

There is another PR aimed at solving the problem more generically (https://github.com/kubeapps/kubeapps/pull/2515), but it is still in progress.

@absoludity , I'd need your help in setting up the environment. Still not able to get authenticated... see https://github.com/kubeapps/kubeapps/pull/2515#discussion_r592298764, but ping me on Slack if you want.

Edit: thanks a lot!! Now I've been able to deploy the dev scenario for both the default and the TMC API suffixes.

### DIFF
--- a/cmd/pinniped-proxy/src/pinniped.rs
+++ b/cmd/pinniped-proxy/src/pinniped.rs
@@ -241,6 +241,7 @@ fn get_pinniped_login_api_group() ->  String  {
     return format!("{}.{}", "login.concierge", &api_suffix).to_string();
 }
 
+
 #[macro_use]
 #[cfg(test)]
 mod tests {
@@ -293,6 +294,7 @@ mod tests {
     #[test]
     #[serial(envtest)]
     fn test_get_api_group_getters() -> Result<()> {
+        env::remove_var("DEFAULT_DEFAULT_PINNIPED_API_SUFFIX");
         let authenticator_api_group =  get_pinniped_authenticator_api_group();
         assert_eq!(authenticator_api_group, "authentication.concierge.pinniped.dev");
 
@@ -306,5 +308,23 @@ mod tests {
         let login_api_group = get_pinniped_login_api_group();
         assert_eq!(login_api_group, "login.concierge.foo.bar");
         Ok(())
-}
+    }
+
+    #[test]
+    #[serial(envtest)]
+    fn test_get_is_tmc_environment() -> Result<()> {
+        env::remove_var("DEFAULT_PINNIPED_API_SUFFIX");
+        let is_tmc_environment: bool = env::var(DEFAULT_PINNIPED_API_SUFFIX).unwrap_or(DEFAULT_API_SUFFIX.into()) == TMC_PINNIPED_API_SUFFIX;
+        assert_eq!(is_tmc_environment, false);
+        
+        let is_tmc_environment: bool = env::var(DEFAULT_PINNIPED_API_SUFFIX).unwrap_or(DEFAULT_API_SUFFIX.into()) == TMC_PINNIPED_API_SUFFIX;
+        env::set_var(DEFAULT_PINNIPED_API_SUFFIX, "foo.bar");
+        assert_eq!(is_tmc_environment, false);
+
+        let is_tmc_environment: bool = env::var(DEFAULT_PINNIPED_API_SUFFIX).unwrap_or(DEFAULT_API_SUFFIX.into()) == TMC_PINNIPED_API_SUFFIX;
+        env::set_var(DEFAULT_PINNIPED_API_SUFFIX, "pinniped.tmc.cloud.vmware.com");
+        assert_eq!(is_tmc_environment, false);
+
+        Ok(())
+    }
 }
--- a/cmd/pinniped-proxy/src/pinniped.rs
+++ b/cmd/pinniped-proxy/src/pinniped.rs
@@ -294,7 +294,7 @@ mod tests {
     #[test]
     #[serial(envtest)]
     fn test_get_api_group_getters() -> Result<()> {
-        env::remove_var("DEFAULT_DEFAULT_PINNIPED_API_SUFFIX");
+        env::remove_var("DEFAULT_PINNIPED_API_SUFFIX");
         let authenticator_api_group =  get_pinniped_authenticator_api_group();
         assert_eq!(authenticator_api_group, "authentication.concierge.pinniped.dev");
 
--- a/cmd/pinniped-proxy/src/pinniped.rs
+++ b/cmd/pinniped-proxy/src/pinniped.rs
@@ -4,7 +4,6 @@ use std::convert::TryFrom;
 use anyhow::{Context, Result};
 use k8s_openapi::api::core::v1 as corev1;
 use k8s_openapi::apimachinery::pkg::apis::meta::v1 as metav1;
-use k8s_openapi::Metadata;
 use kube::{
     api::{Api, PostParams},
     Client, Config,
@@ -19,9 +18,13 @@ use serde_json;
 use thiserror::Error;
 use url::Url;
 
+const DEFAULT_PINNIPED_API_SUFFIX: &str = "DEFAULT_PINNIPED_API_SUFFIX";
 const DEFAULT_PINNIPED_NAMESPACE: &str = "DEFAULT_PINNIPED_NAMESPACE";
 const DEFAULT_PINNIPED_AUTHENTICATOR_NAME: &str = "DEFAULT_PINNIPED_AUTHENTICATOR_NAME";
 const DEFAULT_PINNIPED_AUTHENTICATOR_TYPE: &str = "DEFAULT_PINNIPED_AUTHENTICATOR_TYPE";
+
+const DEFAULT_API_SUFFIX: &str = "pinniped.dev";
+const TMC_PINNIPED_API_SUFFIX: &str = "pinniped.tmc.cloud.vmware.com";
 
 #[derive(Error, Debug)]
 pub enum PinnipedError {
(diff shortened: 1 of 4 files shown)


absoludity:
> @absoludity , I'd need your help in setting up the environment. Still not able to get authenticated... see [#2515 (comment)](https://github.com/kubeapps/kubeapps/pull/2515#discussion_r592298764), but ping me on Slack if you want.

So I first verified that I can use a local setup with pinniped-proxy and pinniped 0.5.0, which worked fine - I updated the instructions at #2270 (and no worries about the extra commits, thanks for removing them again).

Once I had the known working version, I then did the following to test this PR:

1. Delete pinniped on both clusters manually with:
```
kubectl --kubeconfig=$HOME/.kube/kind-config-kubeapps delete -f https://github.com/vmware-tanzu/pinniped/releases/download/v0.5.0/install-pinniped-concierge.yaml
kubectl --kubeconfig=$HOME/.kube/kind-config-kubeapps-additional delete -f https://github.com/vmware-tanzu/pinniped/releases/download/v0.5.0/install-pinniped-concierge.yaml
```

Updated the makefile targets for deploy-pinniped and deploy-pinniped-additional to use 0.6.0 and used this to installed 0.6.0 on both clusters.

Verified that I get a 500 error in Kubeapps when it now tries to request an authenticated URI (`/api/clusters/mydefaultcluster`) error since pinniped-proxy can no longer create a namespaced TokenCredentialRequest, as per the pinniped-proxy logs:

```
 INFO  pinniped_proxy::service > POST https://172.18.0.3:6443/apis/authorization.k8s.io/v1/selfsubjectaccessreviews 201 Created
 ERROR pinniped_proxy::service > Failed to exchange credentials

Caused by:
    err creating token exchange: "{\"apiVersion\":\"login.concierge.pinniped.dev/v1alpha1\",\"kind\":\"TokenCredentialRequest\",\"metadata\":{\"name\":\"\",\"namespace\":\"pinniped-concierge\"},\"spec\":{\"token\":\"eyJhbGciOiJSUzI1NiIsImtpZCI6IjYwYTc1ZWIwNDk5NmI3MTY4ZjRlZDk2OTcyYjE2Nzk2ZDc0YmVkZmEifQ.eyJpc3MiOiJodHRwczovLzE3Mi4xOC4wLjI6MzIwMDAiLCJzdWIiOiJDaVF3T0dFNE5qZzBZaTFrWWpnNExUUmlOek10T1RCaE9TMHpZMlF4TmpZeFpqVTBOallTQld4dlkyRnMiLCJhdWQiOlsic2Vjb25kLWNsdXN0ZXIiLCJ0aGlyZC1jbHVzdGVyIiwiZGVmYXVsdCJdLCJleHAiOjE2MTU1OTk4MTQsImlhdCI6MTYxNTUxMzQxNCwiYXpwIjoiZGVmYXVsdCIsImF0X2hhc2giOiI0NjE5bnpLWXQ3SEpfT2lwZlhTOW5BIiwiZW1haWwiOiJrdWJlYXBwcy1vcGVyYXRvckBleGFtcGxlLmNvbSIsImVtYWlsX3ZlcmlmaWVkIjp0cnVlfQ.ZkPKP2UFcquLwM2EQy2LHgh6WiUDTEAxrvRUVX9NhSGVBxoZSjG9_9c1CVYDDp9Bd9eZMUaHU-McNsIB6N3bzjzFeOvynNk_wtel2tHmgV-zddCzH1KFnTAfmkAzuw1qsApYWHoBxRBRGj5BfSV-NxNzSVr4A3_XjJGMjsPzkTfwa1NOjObyWB_y95-Qzt24k28MB97jfd1sCkY-6rk_1yErkluGRDTrqRtQZYaGVAVBai_eAJ9GnetnxzHwjDWGYDlMbNgb1xxurHPLzYuBwbxeVdGWd9i_5Wudwh45gN1bbT66KYYQXDKRwb4ldBB1hKNhEQhg44Jo8DsZ3FbkKg\",\"authenticator\":{\"apiGroup\":\"authentication.concierge.pinniped.dev\",\"kind\":\"JWTAuthenticator\",\"name\":\"jwt-authenticator\"}}}"
    ApiError: the server could not find the requested resource: NotFound (ErrorResponse { status: "Failure", message: "the server could not find the requested resource", reason: "NotFound", code: 404 })
 INFO  pinniped_proxy::service > GET https://172.18.0.3:6443/ 500 Internal Server Error
 ```

I then compiled a new pinniped-proxy from this PR, loaded it into my kind node and updated the deployment to use it, then was able to authenticate with both clusters again - yay for me, but I can't reproduce your issue :/

I'll comment separately on your logs on that PR.

**Comment by absoludity** [comment]
>     2. Shamefully ugly workaround consisting of duplicating code for supporting another API group rather than `pinniped.dev`. Specifically, this PR adds support for the one that will be used in Tanzu Mission Control.

I don't think we need to duplicate the code like this - but I'll have a play with that on Monday and propose something.

File: cmd/pinniped-proxy/src/pinniped.rs [hunk 1]
```diff
@@ -201,4 +290,41 @@ mod tests {
             },
         }
     }
+
+    #[test]
+    #[serial(envtest)]
+    fn test_get_api_group_getters() -> Result<()> {
+        env::remove_var("DEFAULT_DEFAULT_PINNIPED_API_SUFFIX");
```
absoludity, line 297: The failure on CI is because the above has `DEFAULT_` twice, afaict locally.

absoludity (approved):
Thanks Antonio. There's a small typo in a test causing a CI failure, but otherwise great.

Since 0.5.0 will still be available in the `pinniped.dev` group when they first upgrade and run 0.6.0 with a different group (if I've understood correctly), I'd recommend we not land this for the next release, waiting until 0.6.0 is available in TMC with the non-namespaced resources.

antgamdia:
Thanks a lot for your help here, @absoludity ;  I finally was able to successfully test my changes in 0.6.0 with both the default and the TMC API groups.!

> So I first verified that I can use a local setup with pinniped-proxy and pinniped 0.5.0, which worked fine - I updated the instructions at #2270.

After these updates, I finally managed to set up the environment. I guess I was missing the first step _# copying the certificate-authority-data from the additional cluster_ :disappointed: 
The remaining steps are straightforward once you have the initial environment working. At least I'm happy my changes were working fine.

> I don't think we need to duplicate the code like this - but I'll have a play with that on Monday and propose something.

Yes, I think at least the `match` part can be extracted to a function, but not sure atm. Thank you!
  ```rust
      match s.credential {
            Some(c) => return identity_for_exchange(&c),
            None => match s.message {
                // A returned status without a credential is unsuccessful authentication so
                // add context to identify this.
                Some(m) => return Err(anyhow::anyhow!(m.clone()).context(PinnipedError::UnsuccessfulAuthentication(m))),
                None => return Err(anyhow::anyhow!("response status neither an error msg or a credential: {:#?}", s)),
```

> Since 0.5.0 will still be available in the `pinniped.dev` group when they first upgrade and run 0.6.0 with a different group (if I've understood correctly),

I have asked again to the team just to confirm, depending on what they answer, let's decide next week what to do.
If you are right, I do agree; better not to land right now.

absoludity:
> After these updates, I finally managed to set up the environment. I guess I was missing the first step _# copying the certificate-authority-data from the additional cluster_ disappointed

I don't think that can be the error, in that that data would only have been used *when* you tried to switch clusters to the additional cluster - unless I'm missing something. But glad it was working.

> 
> > Since 0.5.0 will still be available in the `pinniped.dev` group when they first upgrade and run 0.6.0 with a different group (if I've understood correctly),
> 
> I have asked again to the team just to confirm, depending on what they answer, let's decide next week what to do.
> If you are right, I do agree; better not to land right now.

So sounds like I was wrong - it's a straight upgrade. So best to land this and include it in the release. Can you also include an update to the [user docs](https://github.com/kubeapps/kubeapps/blob/master/docs/user/using-an-OIDC-provider-with-pinniped.md#installing-pinniped) reversing the current statement so that the minimum supported version of pinniped is 0.6.0, and the install URL is for 0.6.0? Thanks.

absoludity (approved):
Perfect - thanks for the extra attention on the doc with the note about the jwt-authenticator group and namespace!

PR #2537, opened 2021-03-12
Fix dangling reference to old template helper.

absoludity:
### Description of the change

Looks like the result of merging both the common chart dependency together with the oauth2 sensitive data PR #2436

As a result, templating when using oauth proxy fails:

```
helm template chart/kubeapps --set authProxy.enabled=true
Error: template: kubeapps/templates/_helpers.tpl:143:12: executing "kubeapps.oauth2_proxy-secret.name" at <{{template "kubeapps.fullname" .}}>: template "kubeapps.fullname" not defined
```

### DIFF
--- a/chart/kubeapps/Chart.yaml
+++ b/chart/kubeapps/Chart.yaml
@@ -25,4 +25,4 @@ maintainers:
 name: kubeapps
 sources:
   - https://github.com/kubeapps/kubeapps
-version: 5.3.0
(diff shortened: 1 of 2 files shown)


antgamdia (approved):
Thank you for the change !

andresmgot (approved):

andresmgot:
we should release a new version since the current released version includes this bug.

PR #2542, opened 2021-03-12
Fix internal port configuration

### LINKED ISSUE #2484: Unable to show the Catalog tab

andresmgot:
### Description of the change

There were several issues if a user wanted to change an internal port (e.g. `assetsvc` port):

 - Several times, the port `8080` was hardcoded in the templates
 - The services require to set the env var `PORT` to change the port
 - (still an "issue") It's necessary to set the liveness and readiness probe port as well. (e.g. `--set assetsvc.service.port=9090 --set assetsvc.livenessProbe.httpGet.port=9090 --set assetsvc.readinessProbe.httpGet.port=9090`).

This PR fixes the first 2 items. Since there is no common use case for wanting to change the port, I haven't addressed the third.

I have also changed the documentation to avoid confussion or people trying incomplete commands.

### Applicable issues

  - fixes #2484

### DIFF
--- a/chart/kubeapps/README.md
+++ b/chart/kubeapps/README.md
@@ -57,11 +57,11 @@ Specify each parameter using the `--set key=value[,key=value]` argument to `helm
 
 ```bash
 helm install kubeapps --namespace kubeapps \
-  --set assetsvc.service.port=9090 \
+  --set ingress.enabled=true \
     bitnami/kubeapps
 ```
 
-The above command sets the port for the assetsvc Service to 9090.
+The above command enables an Ingress Rule to expose Kubeapps.
 
 Alternatively, a YAML file that specifies the values for parameters can be provided while installing the chart. For example,
 
--- a/chart/kubeapps/Chart.yaml
+++ b/chart/kubeapps/Chart.yaml
@@ -25,4 +25,4 @@ maintainers:
 name: kubeapps
 sources:
   - https://github.com/kubeapps/kubeapps
-version: 5.3.1
+version: 5.3.2
--- a/chart/kubeapps/templates/_helpers.tpl
+++ b/chart/kubeapps/templates/_helpers.tpl
@@ -169,7 +169,7 @@ Frontend service port number
 {{- if .Values.authProxy.enabled -}}
 3000
 {{- else -}}
-{{ .Values.frontend.service.port }}
+8080
 {{- end -}}
 {{- end -}}
 
--- a/chart/kubeapps/templates/_helpers.tpl
+++ b/chart/kubeapps/templates/_helpers.tpl
@@ -169,7 +169,7 @@ Frontend service port number
 {{- if .Values.authProxy.enabled -}}
 3000
 {{- else -}}
-8080
+{{ .Values.frontend.service.port }}
 {{- end -}}
 {{- end -}}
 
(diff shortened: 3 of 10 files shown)


antgamdia (approved):
Nice! While it is not that critical, it is great to provide a configurable way to change the hardcoded values. Thanks!

chart/kubeapps/README.md
```diff
@@ -57,11 +57,11 @@ Specify each parameter using the `--set key=value[,key=value]` argument to `helm
 
 ```bash
 helm install kubeapps --namespace kubeapps \
-  --set assetsvc.service.port=9090 \
+  --set ingress.enabled=true \
```
absoludity, line 60: Heh, much better example now :)

absoludity (approved):
> This PR fixes the first 2 items. Since there is no common use case for wanting to change the port, I haven't addressed the third.

I don't understand - if we have a value for `assetsvc.service.port`, then shouldn't the liveness probe use the same value? Why does the chart have a separate `assetsvc.livenessProbe.httpGet.port` option? Or at least, why are we setting that with a default in the values.yaml rather than defaulting to `assetsvc.service.port` ? (same for other deploys). It looks like the only reason may be that we're using `toYaml` to specify the probes from the values in the deployments (probably because that's a bitnami chart thing?).

If we switched the `*.livenessProbe.httpGet.port`'s to all not have a default, we could expand the probes in the deployment container specs and use the value for the `*.livenessProbe.httpGet.port` if set, otherwise default to the port of the actual service in our case?

Anyway, not required for this one, just mentioning in case you think it's worth doing for a follow-up. Currently people can just set the probes explicitly as well if needed.

andresmgot:
> > This PR fixes the first 2 items. Since there is no common use case for wanting to change the port, I haven't addressed the third.
> 
> I don't understand - if we have a value for `assetsvc.service.port`, then shouldn't the liveness probe use the same value? Why does the chart have a separate `assetsvc.livenessProbe.httpGet.port` option? Or at least, why are we setting that with a default in the values.yaml rather than defaulting to `assetsvc.service.port` ? (same for other deploys). It looks like the only reason may be that we're using `toYaml` to specify the probes from the values in the deployments (probably because that's a bitnami chart thing?).
> 
> If we switched the `*.livenessProbe.httpGet.port`'s to all not have a default, we could expand the probes in the deployment container specs and use the value for the `*.livenessProbe.httpGet.port` if set, otherwise default to the port of the actual service in our case?
> 
> Anyway, not required for this one, just mentioning in case you think it's worth doing for a follow-up. Currently people can just set the probes explicitly as well if needed.

Yes, you are right, that would be the implementation I just didn't take the time to update all the probes that we have in the chart because there is no real use-case for changing the port (AFAIK).

PR #2547, opened 2021-03-12
Allow to manually set the PostgreSQL password

### LINKED ISSUE #2409: chart values: one has to provide `existingSecret` to create one on the fly
### Description:

In order to provide passwords so that those do not get re-generated every time, it is now needed to provide `postgresql .existingSecret` which, according to the docs, should point to a pre-deployed secret, when in fact it is being used to create a secret on the fly.

Sample values:

```yaml
authProxy:
  enabled: false
  external: true
postgresql:
  existingSecret: zoomazumzum
  postgresqlPassword: duzzitmadda
  replication:
    password: agennoshomo
securityContext:
  enabled: true
```

1. install the chart with the above values
2. observe the creation of the secret `zoomazumzum` with the intended creds
3. remove the `existingSecret: zoomazumzum` part and install the chart again
4. observe that new credentials get generated

No `existingSecret` to be needed for provisioning precreated passwords.

I addressed this downstream with bitnami but they told me to take it here. That closed issue can be found here: https://github.com/bitnami/charts/issues/5463

andresmgot:
### Description of the change

Fixes #2409 

Allow to manually set the PostgreSQL password. For example: `--set postgresql.existingSecret=null --set postgresql.postgresqlPassword="foo"`

cc/ @Morriz

### DIFF
--- a/chart/kubeapps/templates/apprepository-deployment.yaml
+++ b/chart/kubeapps/templates/apprepository-deployment.yaml
@@ -55,7 +55,7 @@ spec:
             {{- if .Values.postgresql.existingSecret }}
             - --database-secret-name={{ .Values.postgresql.existingSecret }}
             {{- else }}
-            - --database-secret-name={{ template "kubeapps.postgresql.fullname" . }}q
+            - --database-secret-name={{ template "kubeapps.postgresql.fullname" . }}
             {{- end }}
             - --database-secret-key=postgresql-password
             - --database-url={{ template "kubeapps.postgresql.fullname" . }}:5432
--- a/chart/kubeapps/templates/apprepository-deployment.yaml
+++ b/chart/kubeapps/templates/apprepository-deployment.yaml
@@ -52,7 +52,11 @@ spec:
            {{- end }}
             - --repo-sync-cmd=/asset-syncer
             - --namespace={{ .Release.Namespace }}
+            {{- if .Values.postgresql.existingSecret }}
             - --database-secret-name={{ .Values.postgresql.existingSecret }}
+            {{- else }}
+            - --database-secret-name={{ template "kubeapps.postgresql.fullname" . }}q
+            {{- end }}
             - --database-secret-key=postgresql-password
             - --database-url={{ template "kubeapps.postgresql.fullname" . }}:5432
             - --database-user=postgres
--- a/chart/kubeapps/templates/apprepository-jobs-postupgrade.yaml
+++ b/chart/kubeapps/templates/apprepository-jobs-postupgrade.yaml
@@ -49,10 +49,17 @@ spec:
             - --database-name=assets
           env:
             - name: DB_PASSWORD
+              {{- if .Values.postgresql.existingSecret }}
               valueFrom:
                 secretKeyRef:
                   key: postgresql-password
                   name: {{ .Values.postgresql.existingSecret }}
+              {{- else }}
+              valueFrom:
+                secretKeyRef:
+                  name: {{ template "kubeapps.postgresql.fullname" . }}
+                  key: postgresql-password
+              {{- end }}
       {{- if .Values.postgresql.existingSecret }}
       initContainers:
         # Helm 3.1 supports a lookup template tag to create a secret if it does not exist
--- a/chart/kubeapps/templates/apprepository-jobs-postupgrade.yaml
+++ b/chart/kubeapps/templates/apprepository-jobs-postupgrade.yaml
@@ -49,16 +49,13 @@ spec:
             - --database-name=assets
           env:
             - name: DB_PASSWORD
-              {{- if .Values.postgresql.existingSecret }}
               valueFrom:
                 secretKeyRef:
                   key: postgresql-password
+              {{- if .Values.postgresql.existingSecret }}
                   name: {{ .Values.postgresql.existingSecret }}
               {{- else }}
-              valueFrom:
-                secretKeyRef:
                   name: {{ template "kubeapps.postgresql.fullname" . }}
-                  key: postgresql-password
               {{- end }}
       {{- if .Values.postgresql.existingSecret }}
       initContainers:
--- a/chart/kubeapps/templates/db-secret-jobs-cleanup.yaml
+++ b/chart/kubeapps/templates/db-secret-jobs-cleanup.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.postgresql.existingSecret -}}
 apiVersion: batch/v1
 kind: Job
 metadata:
@@ -42,3 +43,4 @@ spec:
           args:
             - -c
             - "kubectl delete secret -n {{ .Release.Namespace }} {{ .Values.postgresql.existingSecret }}|| true"
+{{- end -}}
--- a/chart/kubeapps/Chart.yaml
+++ b/chart/kubeapps/Chart.yaml
@@ -25,4 +25,4 @@ maintainers:
 name: kubeapps
 sources:
   - https://github.com/kubeapps/kubeapps
-version: 5.3.2
+version: 5.3.3
--- a/chart/kubeapps/templates/assetsvc-deployment.yaml
+++ b/chart/kubeapps/templates/assetsvc-deployment.yaml
@@ -45,16 +45,13 @@ spec:
             - --database-url={{ template "kubeapps.postgresql.fullname" . }}-headless:5432
           env:
             - name: DB_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  key: postgresql-password
               {{- if .Values.postgresql.existingSecret }}
-              valueFrom:
-                secretKeyRef:
                   name: {{ .Values.postgresql.existingSecret }}
-                  key: postgresql-password
               {{- else }}
-              valueFrom:
-                secretKeyRef:
                   name: {{ template "kubeapps.postgresql.fullname" . }}
-                  key: postgresql-password
               {{- end }}
             - name: POD_NAMESPACE
               valueFrom:
--- a/chart/kubeapps/templates/assetsvc-deployment.yaml
+++ b/chart/kubeapps/templates/assetsvc-deployment.yaml
@@ -45,10 +45,17 @@ spec:
             - --database-url={{ template "kubeapps.postgresql.fullname" . }}-headless:5432
           env:
             - name: DB_PASSWORD
+              {{- if .Values.postgresql.existingSecret }}
               valueFrom:
                 secretKeyRef:
                   name: {{ .Values.postgresql.existingSecret }}
                   key: postgresql-password
+              {{- else }}
+              valueFrom:
+                secretKeyRef:
+                  name: {{ template "kubeapps.postgresql.fullname" . }}
+                  key: postgresql-password
+              {{- end }}
             - name: POD_NAMESPACE
               valueFrom:
                 fieldRef:


chart/kubeapps/templates/apprepository-deployment.yaml
```diff
@@ -52,7 +52,11 @@ spec:
            {{- end }}
             - --repo-sync-cmd=/asset-syncer
             - --namespace={{ .Release.Namespace }}
+            {{- if .Values.postgresql.existingSecret }}
             - --database-secret-name={{ .Values.postgresql.existingSecret }}
+            {{- else }}
+            - --database-secret-name={{ template "kubeapps.postgresql.fullname" . }}q
```
antgamdia, line 58: is this `q` at the end something intented?

antgamdia (approved):
Thanks for the changes!

chart/kubeapps/templates/apprepository-jobs-postupgrade.yaml
```diff
@@ -49,10 +49,17 @@ spec:
             - --database-name=assets
           env:
             - name: DB_PASSWORD
+              {{- if .Values.postgresql.existingSecret }}
```
absoludity, line 52: Wouldn't this be a bit DRYer if we just put the conditional around the `valueFrom.secretKeyRef.name`? Also in the other resources below.

chart/kubeapps/templates/db-secret-jobs-cleanup.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.postgresql.existingSecret -}}
```
absoludity, line 1: I'm confused about this condition, but I think it's just the name. The problem is that the `.Values.postgresql.existingSecret` gets a default value, which means it's not really telling us if the user is using an existing secret?

In reality, I think the only time we *don't* want to delete this is if the user is using an existing secret that's been created out-of-band. But we don't have a way to know that currently.

Could we not improve this situation here by:

1. remove the current default for `existingSecret` so that it is only defined if explicitly done so by the user,
2. Don't use `kubeapps.postgresql.fullname` when `existingSecret` is not defined, but instead use the current default value of `kubeapps-db` (but perhaps define it as a helper constant so it's just defined once and can be updated with a major release if we want)

? I think with that, existing installs where `existingSecret` hasn't been explicitly defined will still use `kubeapps-db` when they upgrade. Those that have defined `existingSecret` will still continue to use that value. And importantly, we can then update the condition here to only delete the secret if `existingSecret` is undefined (ie. user hasn't explicitly set an existing secret, so we can delete the one kubeapps created).

chart/kubeapps/templates/apprepository-deployment.yaml
andresmgot: mmh, no, it should've slipped after I tested this.

chart/kubeapps/templates/db-secret-jobs-cleanup.yaml
andresmgot: Yes, the name is confusing. The thing is that, by default, postgresql cannot be updated/upgraded. This is because, if left by default, the chart will create a secret with a random password. When upgraded, that secret will be generated again, overriding the password. You need to either, set a password manually or specify a secret. 

In kubeapps we are doing the second. We are generating a secret in the pre-install hook (so it's not modified in the future) and we are filling it with a random password (https://github.com/kubeapps/kubeapps/blob/master/chart/kubeapps/templates/db-secret-bootstrap.yaml). Because of that, we need to "manually" clean up that secret as well when uninstalling. 

We don't allow users to specify their own secret. The two options are: manually setting the password or letting it to be auto-generated.

Morriz:
When using this with the above given `--set postgresql.existingSecret=null --set postgresql.postgresqlPassword="foo"` I get the following output:

```yaml
        env:
            - name: DB_PASSWORD
              valueFrom:
                secretKeyRef:
                  key: postgresql-password
                  name:
```
which leads to errors:
```
cannot patch "kubeapps-internal-assetsvc" with kind Deployment: Deployment.apps "kubeapps-internal-assetsvc" is invalid: spec.template.spec.containers[0].env[0].valueFrom.secretKeyRef.name: Invalid value: "": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```

chart/kubeapps/templates/db-secret-jobs-cleanup.yaml
absoludity: > Yes, the name is confusing. The thing is that, by default, postgresql cannot be updated/upgraded. This is because, if left by default, the chart will create a secret with a random password. When upgraded, that secret will be generated again, overriding the password. You need to either, set a password manually or specify a secret.
> 
> In kubeapps we are doing the second. We are generating a secret in the pre-install hook (so it's not modified in the future) and we are filling it with a random password (https://github.com/kubeapps/kubeapps/blob/master/chart/kubeapps/templates/db-secret-bootstrap.yaml). Because of that, we need to "manually" clean up that secret as well when uninstalling.

Yes, I think all of the above is what I'd understood - this is why the Kubeapps chart currently defaults `postgresql.existingSecret=kubeapps-db` which is confusing because `existingSecret` implies that the secret existed already, wherease Kubeapps is creating it during install.

What wasn't clear to me is that I think you're using `kubeapps.postgresql.fullname` for the secret when `existingSecret` is not defined because that'll be the secret which is created by the postgresql chart when setting `postgresql.postgresqlPassword` I think?

> 
> We don't allow users to specify their own secret. The two options are: manually setting the password or letting it to be auto-generated.

Right, and this is the specific cased raised in the issue - understood. I think it'd be much clearer if Kubeapps didn't use the `existingSecret` option to mean "even though there isn't an existing secret, we'll create one and tell postgres that it existed". An alternative would have been:

1. if (a new) `.Values.userExistingSecret` is unset, do the current behaviour of defaulting `postgresql.existingSecret=kubeapps-db` and clean it up afterwards
2. if `.Values.userExistingSecret` is set, use it for `postgresql.exstingSecret` and *don't* clean it up afterwards

With something like that, I think the issue would be solved with an existing secret, rather than specifying an actual password (though that would still be possible). Anyway, don't want to block on this.

absoludity (approved):
+1 when @Morriz 's point is fixed.

chart/kubeapps/templates/db-secret-jobs-cleanup.yaml
andresmgot: > 1. if (a new) `.Values.userExistingSecret` is unset, do the current behaviour of defaulting `postgresql.existingSecret=kubeapps-db` and clean it up afterwards
> 2. if `.Values.userExistingSecret` is set, use it for `postgresql.exstingSecret` and _don't_ clean it up afterwards

As far as I know, you cannot do this with Helm. You cannot set a value from other. You would need to use helpers for that but we cannot modify the existing helpers of the PostgreSQL chart (which is just reading `.Values.existingSecret`.

andresmgot:
> When using this with the above given `--set postgresql.existingSecret=null --set postgresql.postgresqlPassword="foo"` I get the following output:
> 
> ```yaml
> # Source: kubeapps/templates/apprepository-jobs-postupgrade.yaml
> ...
>           env:
>             - name: DB_PASSWORD
>               valueFrom:
>                 secretKeyRef:
>                   key: postgresql-password
>                   name:
> ```
> 
> which leads to errors:
> 
> ```
> cannot patch "kubeapps-internal-assetsvc" with kind Deployment: Deployment.apps "kubeapps-internal-assetsvc" is invalid: spec.template.spec.containers[0].env[0].valueFrom.secretKeyRef.name: Invalid value: "": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
> ```

I think you may be using the chart from master? It works fine for me here:

```
▶ helm install kubeapps ./chart/kubeapps -n kubeapps --create-namespace --set postgresql.existingSecret=null --set postgresql.postgresqlPassword="foo" --dry-run --debug  | grep 'key: postgresql-password' -B 5 -A 5

            - --database-name=assets
          env:
            - name: DB_PASSWORD
              valueFrom:
                secretKeyRef:
                  key: postgresql-password
                  name: kubeapps-postgresql
---
# Source: kubeapps/templates/apprepositories.yaml
apiVersion: kubeapps.com/v1alpha1
kind: AppRepository
--
            - --database-url=kubeapps-postgresql-headless:5432
          env:
            - name: DB_PASSWORD
              valueFrom:
                secretKeyRef:
                  key: postgresql-password
                  name: kubeapps-postgresql
            - name: POD_NAMESPACE
              valueFrom:
                fieldRef:
                  fieldPath: metadata.namespace
--
              value: "5432"
            - name: POSTGRES_PASSWORD
              valueFrom:
                secretKeyRef:
                  name: kubeapps-postgresql
                  key: postgresql-password
            - name: POSTGRESQL_ENABLE_TLS
              value: "no"
            - name: POSTGRESQL_LOG_HOSTNAME
              value: "false"
            - name: POSTGRESQL_LOG_CONNECTIONS
--
              value: "postgres"
            - name: POSTGRES_PASSWORD
              valueFrom:
                secretKeyRef:
                  name: kubeapps-postgresql
                  key: postgresql-password
            - name: POSTGRES_DB
              value: "assets"
            - name: POSTGRESQL_ENABLE_LDAP
              value: "no"
            - name: POSTGRESQL_ENABLE_TLS
```

Morriz:
No, I just updated the chart repo (`helm repo update`) and fetched it with `helm fetch bitnami/kubeapps --untar`. The version I deployed is:
```yaml
appVersion: 2.2.1
version: 5.3.1
```
Deploying that with these values gives the error above:
```yaml
postgresql:
  existingSecret: null
  postgresqlPassword: duzzitmadda
```

Morriz:
Of course this is not yet released. Pffff. When will it?

absoludity:
Andres will be getting the release out later this week or early next week.

andresmgot:
Actually I bumped the chart version with this PR so a new version was released a bit afterwards. You should be able to see the version `5.3.3` which includes the fix @Morriz .

Morriz:
I just installed the latest chart version which is now at `5.4.0`, but I still get a new `postgresql.postgresqlPassword` generated every time when using `--set postgresql.existingSecret=null --set postgresql.postgresqlPassword="foo"`. This is not solved. Could this be re-opened @andresmgot ?

Morriz:
reproduce it by running this twice and compare output: `helm template bitnami/kubeapps --set postgresql.existingSecret=null --set postgresql.postgresqlPassword="foo"`

Morriz:
nevermind, this points to the fact that `postgresqlPostgresPassword` was not set. When set this also eliminates pass generation for that item.

andresmgot:
just as a heads up @Morriz, we are modifying this behavior in #2595 (and we are going to have a breaking in the next chart release).

Morriz:
Ah, I see that is merged already. I did put my concerns on the bottom there, but want to paste it here as well:

> Guys, I want to strongly suggest that configuration generates idempotent payloads. So no side effects should happen related to a difference in cluster state. We want any apply with same config result in the same resources being deployed.
> 
> So please drop any in-cluster jobs with state dependent logic.
> 
> Imo idempotency should be a very high up principle when developing 12-factor apps.

PR #2551, opened 2021-03-15
Add information about our API docs

antgamdia:
### Description of the change

As a result of a couple of issues asking for information about the Kubeapps API documentation, this PR brings some paragraphs explaining how to access the docs (and a disclaimer, since it is a very early stage).

### Benefits

Users will find our API docs more easily.

### Possible drawbacks

Users will find our API docs more easily :P

### Applicable issues

N/A 

### Additional information

N/A


README.md
```diff
@@ -42,6 +42,14 @@ If you have followed the instructions for [installing Kubeapps](./docs/user/gett
 - [Kubeapps components](./docs/architecture/overview.md)
 - [Roadmap](https://github.com/kubeapps/kubeapps/wiki/Roadmap)
 
+## API documentation
+
+Whereas Kubeapps does not have yet a well-design API, we still want to share the current status of the APIs used across the internal components. Check out the [latest API online documentation](https://app.swaggerhub.com/apis/kubeapps/Kubeapps) or download the Kubeapps [OpenAPI Specification yaml file](./dashboard/public/openapi.yaml) from the repository.
+
+> This API documentation is subject to change in each Kubeapps version. If you are planning to build any solution on the top of the Kubeapps API, please [reach us](https://kubernetes.slack.com/messages/kubeapps) for further information.
```
absoludity, line 49: Normally "please reach out to us for ..." or "you can reach us on slack for ..." rather than "reach us for ..." , but more generally, unless we plan on supporting the current api, I wouldn't want people to plan to build any solution on top of the current api.

README.md
```diff
@@ -42,6 +42,14 @@ If you have followed the instructions for [installing Kubeapps](./docs/user/gett
...
+
+Besides, you install Kubeapps in your cluster, you can find the interactive API documentation portal on the right menu.
```
absoludity, line 51: If you install Kubeapps in your cluster, ...

And "on the right menu" may not help people find it. Perhaps a screenshot?

absoludity (approved):
Thanks Antonio. A couple of small suggestions for language.

README.md
antgamdia: I thought so... but since it is the main readme, I didn't want to make any disruptive UI change there. Maybe we can remove it, no?

antgamdia: Also, unrelated to this PR: should we remove the "Differences from Monocular" section since it is not that relevant? Or even move it to another doc, but not the main readme, no?

andresmgot: > Also, unrelated to this PR: should we remove the "Differences from Monocular" section since it is not that relevant? Or even move it to another doc, but not the main readme, no?

yes, I'd remove it.

README.md
```diff
@@ -42,6 +42,12 @@ If you have followed the instructions for [installing Kubeapps](./docs/user/gett
 - [Kubeapps components](./docs/architecture/overview.md)
 - [Roadmap](https://github.com/kubeapps/kubeapps/wiki/Roadmap)
 
+## API documentation
+
+Whereas Kubeapps does not have yet a well-design API, we still want to share the current status of the APIs used across the internal components. Check out the [latest API online documentation](https://app.swaggerhub.com/apis/kubeapps/Kubeapps) or download the Kubeapps [OpenAPI Specification yaml file](./dashboard/public/openapi.yaml) from the repository.
+
+> This API documentation is subject to change in each Kubeapps version. If you are interested in the Kubeapps API, please [reach out to us](https://kubernetes.slack.com/messages/kubeapps) for further information.
```
andresmgot, line 49: people may want to contribute to the API documenation. Can we document the process? (I know right now it's not possible for external people to modify the api in swaggerhub but at least they can be aware of the environment so the file `./dashboard/public/openapi.yaml` is not manually modified.

antgamdia: Good idea, however, I think the only option for external users is to send us a PR modifying this file. They can use swaggerhub, swagger editor or their preferred editor for doing so, but, in the end, they have to open a PR. 
I don't expect many external changes here, so it is not a big deal to bring manually the upstream changes to swaggerhub, imho.


```suggestion
## API documentation

Whereas Kubeapps does not have yet a well-designed API, we still want to share the current status of the APIs used across the internal components. Check out the [latest API online documentation](https://app.swaggerhub.com/apis/kubeapps/Kubeapps) or download the Kubeapps [OpenAPI Specification yaml file](./dashboard/public/openapi.yaml) from the repository. Please feel free to send us a PR if you find any errors.

> This API documentation is subject to change in each Kubeapps version. If you are interested in the Kubeapps API, please [reach out to us](https://kubernetes.slack.com/messages/kubeapps) for further information.
```

(first time using suggestions, it seems they can be merged in the PR indeed, interesting)

andresmgot: Yes, I wouldn't just point users to send a PR but actually document the process of modifying SwaggerHub (the same content from the gdoc you own). For two reasons:

 - It's easier for us to keep all the dev docs here in the repo.
 - If someone creates a PR against `./dashboard/public/openapi.yaml` and needs to be modified, maybe because it doesn't matches the style that swaggerhub outputs, we can point them to the public docs (even though we would need to re-create the PR on our own, due to the lack of permissions).

If you don't want to re-write the file in markdown, you can just link the google doc as we do with some design docs.

antgamdia: Mmm, ok, not a bad idea. I also prefer having all the docs in this repo, I'll try to reword the docs so that it does not refer to Kubeapps repo, but a user-owned fork. 

However, note that swaggerhub is nothing but a yaml editor on steroids. There is no magic, formatters or autogenerated stuff, so  changes like "use type: number instead of type: string" will be easily achieved with a simple PR.

Anyway, let's add the new doc and discuss it later

andresmgot (approved):
Thanks!

README.md
antgamdia: Thanks for the +1 @andresmgot , but I've just noticed I forgot to update the readme.md (:facepalm: ) adding the link to the guide we discussed about. Doing it right now.
If still +1, we can merge it

PR #2561, opened 2021-03-17
Refactor loading UX

### LINKED ISSUE #2276: Avoid to close the websocket connection when loading the component
### Description:

Follow up of #2267

After https://github.com/kubeapps/kubeapps/pull/2267#issuecomment-765252896, I noticed that the websocket connection to get application resources is opened and close several times during the loading of the component. Causing an error (warning) if the socket is closed before the connection is finished. We should avoid closing the socket if the component is still loading.

### LINKED ISSUE #2539: Use Clarity spinner GIF
### Description:

Clarity already has a solution for the loading animation. We can either use the clr-ui version: https://v2.clarity.design/spinners or the core component (still in preview but it should not be a problem): https://clarity.design/storybook/core/?path=/story/components-preview-circular-progress--page

Also, we usually misplace the loading spinner when loading pages. Rather than at the very top (with no margin), we should center it a little bit. Also some loading behavior can be smoother (like in the ApplicationView):

Peek 2021-03-12 10-06

andresmgot:
### Description of the change

This PR includes several changes/improvements to the loading experience:

 - Use Clarity Core Component for the progress circle.
 - Improve style, adding some loading text explaining what's being loaded.
 - Fix behavior for the Application View where the resources were being requested several times, impacting the performance and the experience.
 - Some other minor CSS fixes.
 
Peek 2021-03-17 11-51

### Applicable issues

  - fixes #2539
  - fixes #2276


dashboard/src/components/AppView/AppSecrets/AppSecrets.scss
```diff
@@ -1,4 +1,6 @@
 .app-secrets-content {
-  margin-top: 1.2rem;
-  margin-right: 1.2rem;
+  margin-top: 1rem;
+  padding-right: 15px;
+  max-height: 15rem;
+  overflow-y: auto;
```
andresmgot, line 5: this was necessary (and related to https://github.com/kubeapps/kubeapps/issues/2462) when the app has too many secrets (e.g. harbor):

Screenshot from 2021-03-17 12-39-05

antgamdia:
Thanks for the gif, much clearer! I want to take a closer look at the PR later, however, just out of curiosity: have you changed the initial loading style (perhaps the one introduced by suspense? )

(This one:)
image

In other corporate websites using clarity I use to see this alternative approach:

image

Something similar to: https://codesandbox.io/s/suspense-progress-bar-mu9lv

Or even this other one:

image



What do you think? Does the top spinner still make sense for the very initial loading?

andresmgot:
> What do you think? Does the top spinner still make sense for the very initial loading?

yes, I think it still makes sense. Let me try to find it and add at least a title like the case for TAC.

andresmgot:
> > What do you think? Does the top spinner still make sense for the very initial loading?
> 
> yes, I think it still makes sense. Let me try to find it and add at least a title like the case for TAC.

The `LoadingWrapper` from the `Suspense` component didn't appear (so I removed it) and the first one loaded was in the `ConfigLoader` so I added a title there:

Screenshot from 2021-03-17 13-21-10

dashboard/src/components/ConfigLoader/ConfigLoader.tsx
```diff
@@ -18,7 +18,9 @@ function ConfigLoader({ getConfig, error, ...otherProps }: IConfigLoaderProps) {
   if (error) {
     return <Alert theme="danger">Unable to load Kubeapps configuration: {error.message}</Alert>;
   }
-  return <LoadingWrapper {...otherProps} />;
+  return (
+    <LoadingWrapper className="margin-t-xxl" loadingText={<h1>Kubeapps</h1>} {...otherProps} />
+  );
```
antgamdia, line 23: > The `LoadingWrapper` from the `Suspense` component didn't appear (so I removed it) and the first one loaded was in the `ConfigLoader` so I added a title the

Oh much better now! Thank you! Minor thing here: since we are allowing modifying Kubeapps by other names, perhaps it should be the following for the sake of consistency:

```suggestion
  return (
    <LoadingWrapper className="margin-t-xxl" loadingText={<h1>{intl.formatMessage({ id: "Kubeapps", defaultMessage: "Kubeapps" })}</h1>} {...otherProps} />
  );
```

andresmgot: The ConfigLoader is out of of the IntlProvider:

```
      <ConfigLoaderContainer>
        <ConnectedRouter history={history}>
          <Suspense fallback={null}>
            <IntlProvider
              locale={i18nConfig.locale}
              key={i18nConfig.locale}
              messages={i18nConfig.messages}
              defaultLocale="en"
            >
```

Causing an issue:

```
Error: [React Intl] Could not find required `intl` object. <IntlProvider> needs to exist in the component ancestry.
```

Should we move the IntlProvider up in the tree or leave the ConfigLoader out?

antgamdia: Mmm, I don't remember right now, but since we are not using anything from the `ConfigLoaderConainer` in the `IntlProvider` I guess we can move it up.

antgamdia (approved):
Just minor comments and some doubts here. I've pulled the PR and it looks much better now, thank you for the UI changes of course, but thanks for the websocket thing! (since I'm with the Pinniped was issue, now it is easier to debug what's happening :)   )

dashboard/src/components/Catalog/Catalog.tsx
```diff
@@ -447,7 +447,7 @@ function Catalog(props: ICatalogProps) {
                       (!filters[filterNames.TYPE].length ||
                         filters[filterNames.TYPE].find((type: string) => type === "Charts")) && (
                         <div className="endPageMessage">
-                          <LoadingWrapper medium={true} loaded={false} />
+                          <LoadingWrapper loaded={false} />
```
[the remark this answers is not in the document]
andresmgot: yes, in this case, is the same loading animation when fetching new apps:

Screenshot from 2021-03-17 17-02-42

And I find it a bit distracting if there is text there that I am not fast enough to read.

dashboard/src/components/AppList/AppList.tsx
```diff
@@ -123,7 +123,11 @@ function AppList() {
           </Link>,
         ]}
       />
-      <LoadingWrapper loaded={!isFetching && !isFetchingResources}>
+      <LoadingWrapper
+        loaded={!isFetching && !isFetchingResources}
+        loadingText="Getting the list of applications..."
```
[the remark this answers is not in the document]
antgamdia: Great, I agree, a full capitalized sentence looks terrible, haha. Thanks for the change :)

dashboard/src/components/Catalog/Catalog.tsx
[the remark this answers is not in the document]
antgamdia: Ok then, thanks for checking!

PR #2562, opened 2021-03-17
Add pinniped-proxy dev env docs (+ config in #2270)

antgamdia:
### Description of the change

This is a follow-up PR that supersedes #2270. I've pulled these changes and write some docs about how to set up the environment.

### Benefits

Newcomer developers might understand better how to configure the environment for developing pinniped-proxy.

### Possible drawbacks

I don't remember if there is any reason to hold #2270 open. If it was just for the lack of documentation, this PR aims at finishing this remaining point so that we can close that draft PR.

### Applicable issues

N/A

### Additional information

(bumped up the Pinniped version to v0.6.0, changed also the jwt authenticator -no longer namespaced-, consequently, this PR won't work out-of-the-box until we release Kubeapps 2.3)

WDYT, @absoludity ?


docs/developer/pinniped-proxy.md
```diff
@@ -55,3 +52,62 @@ Similarly, tests can be run with the cargo tool:
...
+
+### Prerequisite: getting the node IP
+
+First of all, execute `docker network inspect kind | jq '.[0].IPAM.Config[0].Gateway'` (adapt this line as you require, you only will need the Kind Gateway IP), get the IP and use the next one. For example, if you get `172.18.0.1` in your command, you will need `172.18.0.2`.
+
+Next, follow the steps in [/docs/developer/using-makefiles.md](../developer/using-makefiles.md) to modify the proper yaml files by replacing with this value.
+
+### Launching the multi-cluster dev environment
+
+ - Execute `make multi-cluster-kind`.
```
absoludity, line 66: I don't see how we can land this? It's removing the oidc setup which is currently working, replacing it with the pinniped-proxy (no cluster oidc) setup that requires much manual setup.

If we want to land developer support for the pinniped-proxy setup, we'd need to do so without breaking the oidc setup. I also think if we do that we should move away from having all the config in a single docs/user/manifests directory as we do now, and instead have config for each setup (cluster config for kind, values for our chart etc.) in separate directories. Perhaps even separate Makefiles if we're going to continue with them? What do you think?

absoludity (commented):
Not sure how this can land without breaking the oidc setup that currently works with minimal manual interaction? Let me know if I'm missing something.

docs/developer/pinniped-proxy.md
antgamdia: Oh yes, you're totally right. Since I'm about to set up the normal oidc setup now, I'll create separate files for the pinniped env instead.
I was excited about having this Pinniped env seamlessly working, that I forgot about the other, sorry!
I'll eventually update the PR with these changes. Thank you for pointing this out.

antgamdia: Take a look when you have a chance, @absoludity. I already did (tried to, at least) the changes. 
:)

absoludity (approved):
Great, thanks Antonio.

PR #2563, opened 2021-03-17
Fix test to work with the latest eslint rules

antgamdia:
### Description of the change

Since one of the latest PR was being developed at the same point as we were migrating to eslit, there is a minor issue preventing our CI to work.
This PR simply tweaks a little bit the failing file, a test using window.location.

### Benefits

No CI errors.

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

N/A


andresmgot (approved):
I also modified this in #2561 but I like your approach better :)

antgamdia:
> I also modified this in #2561 but I like your approach better :)

Thanks! It is the same we used when we upgrade jest to the latest version and discovered window.* properties are now read-only, so I just used it.

PR #2575, opened 2021-03-18
Parallelize namespace-check requests

### LINKED ISSUE #2506: Very slow response time from kubeops backend requests
### Description:

When kubeapps is deployed to a cluster that has a large number of namespaces (e.g. > 100), at least one endpoint i.e. `/backend/v1/clusters/default/namespaces` becomes very slow:
```
[negroni] 2021-03-03T23:25:59Z | 200 |   23.2154635s | chi-tenant-kubeapps-internal-kubeops:8080 | GET /backend/v1/clusters/default/namespaces
```
In our case this request _always_ takes 23s+ in duration to resolve w/ 200 OK.

1. Create a kubernetes cluster and deploy kubeapps into it.
2. Write a simple bash script to create some large number of namespaces in the cluster.
3. Load the kubeapps dashboard paying particular attention to the proxied request `/api/v1/clusters/default/namespaces`; it will take a very long time to respond.

We have discussed this issue and reproduced it with kubeapps maintainers - for context see the [Slack conversation](https://kubernetes.slack.com/archives/C9D3TSUG4/p1614807084007900).

### Implementation Proposal(s)

@salanki will be updating this issue with an implementation proposal.

The specific spot in the Golang codebase (i.e. kubeops repo) that leads to the latency is where every namespace in the cluster is checked for being accessible by kubeapps or not:

* https://github.com/kubeapps/kubeapps/blob/9356f12b8307f8e3f2764023be461249725a1d43/pkg/kube/kube_handler.go#L803
* https://github.com/kubeapps/kubeapps/blob/9356f12b8307f8e3f2764023be461249725a1d43/pkg/kube/kube_handler.go#L755-L764

In our case, kubeapps will sit behind our API gateway which sets proxy headers including `X-Consumer-Permissions` (a list of permission including *namespaces the user can access*). Ideally we could provide to kubeapps our own definition for "what namespaces can the logged in user/account interact with" - but we aren't too sure how to make such functionality more generic (i.e. a generic way to tell kubeapps what namespaces a user/account can access).

andresmgot:
### Description of the change

Run namespace checks in parallel (the check that verifies if a user can access a namespace). Also, if the user has permissions to list namespaces, this check is skipped.

I did the following tests to verify the new performance:

1. 200 namespaces, without this PR: 39s
~~2. 200 namespaces, 20 parallel workers (the new default): 8.6s~~
2. 200 namespaces, 10 QPS, 15 Burst requests: 19.8s
3. 200 namespaces, 100 QPS, 200 Burst requests: 0.5s

### Possible drawbacks

Using too many workers might saturate the k8s API.

### Applicable issues

  - fixes #2506

### Additional information

cc/ @mecampbellsoup

### DIFF
--- a/chart/kubeapps/README.md
+++ b/chart/kubeapps/README.md
@@ -197,6 +197,7 @@ kubectl delete namespace kubeapps
 - [Is there any API documentation?](#is-there-any-api-documentation)
 - [Why can't I configure global private repositories?](#why-cant-i-configure-global-private-repositories)
 - [Does Kubeapps support Operators?](#does-kubeapps-support-operators)
+- [Slow response when listing namespaces?](#slow-response-when-listing-namespaces)
 - [More questions?](#more-questions)
 
 ### How to install Kubeapps for demo purposes?
@@ -274,6 +275,12 @@ You could alternatively ensure that the `imagePullSecret` is available in all na
 ### Does Kubeapps support Operators?
 
 Yes! You can get started by following the [operators documentation](https://github.com/kubeapps/kubeapps/blob/master/docs/user/operators.md).
+
+### Slow response when listing namespaces
+
+Kubeapps uses the logged user account to retrieve the list of namespaces. If the user doesn't have permission to list namespaces, the fallback mechanism verifies if the user has permissions to get secrets for each namespace (to verify it it should be allowed to use that namespace or not). This can lead to a slow response if the number of namespaces of the cluster is big enough.
+
+To reduce this time, you can increase the number of checks that Kubeapps will perform in parallel setting the value: `kubeops.extraEnvVars[0].name=MAX_REQUESTS` and `kubeops.extraEnvVars[0].value=<desired_number>`. The default value, if not set is 20 concurrent requests.
 
 ### More questions? 
 
--- a/chart/kubeapps/README.md
+++ b/chart/kubeapps/README.md
@@ -278,9 +278,9 @@ Yes! You can get started by following the [operators documentation](https://gith
 
 ### Slow response when listing namespaces
 
-Kubeapps uses the logged user account to retrieve the list of namespaces. If the user doesn't have permission to list namespaces, the fallback mechanism verifies if the user has permissions to get secrets for each namespace (to verify it it should be allowed to use that namespace or not). This can lead to a slow response if the number of namespaces of the cluster is big enough.
+Kubeapps uses the currently logged-in user credential to retrieve the list of all namespaces. If the user doesn't have permission to list namespaces, the backend will try again with its own service account to list all namespaces and then iterate through each namespace to check if the user has permissions to get secrets for each namespace (to verify if they should be allowed to use that namespace or not and hence whether it is included in the selector). This can lead to a slow response if the number of namespaces on the cluster is large.
 
-To reduce this time, you can increase the number of checks that Kubeapps will perform in parallel setting the value: `kubeops.extraEnvVars[0].name=MAX_REQUESTS` and `kubeops.extraEnvVars[0].value=<desired_number>`. The default value, if not set is 20 concurrent requests.
+To reduce this time, you can increase the number of checks that Kubeapps will perform in parallel (per connection) setting the value: `kubeops.burst=<desired_number>` and `kubeops.QPS=<desired_number>`. The default value, if not set, is 15 burst requests and 10 QPS afterwards.
 
 ### More questions? 
 
--- a/pkg/kube/kube_handler.go
+++ b/pkg/kube/kube_handler.go
@@ -22,11 +22,10 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"math"
 	"net/http"
 	"net/url"
-	"os"
 	"path"
-	"strconv"
 	"strings"
 	"sync"
 
@@ -45,22 +44,6 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 )
-
-var (
-	maxReq = 20
-)
-
-func init() {
-	maxReqStr := os.Getenv("MAX_REQUESTS")
-	if maxReqStr != "" {
-		var err error
-		maxReq, err = strconv.Atoi(maxReqStr)
-		if err != nil {
-			log.Errorf("Unable to parse the given MAX_REQUEST: %v", err)
-			maxReq = 20
-		}
-	}
-}
 
 // ClusterConfig contains required info to talk to additional clusters.
 type ClusterConfig struct {
@@ -119,12 +102,6 @@ func NewClusterConfig(inClusterConfig *rest.Config, userToken string, cluster st
 	config := rest.CopyConfig(inClusterConfig)
 	config.BearerToken = userToken
 	config.BearerTokenFile = ""
-	// Modify the default number of requests that the given client can do
-	// This is useful to handle a large number of namespaces which are check in parallel
-	// Burst is the initial number of request made in parallel
-	// Then further requests are performed following the QPS rate
-	config.Burst = maxReq * 2
-	config.QPS = float32(maxReq)
 
 	clusterConfig, ok := clustersConfig.Clusters[cluster]
 	if !ok {
@@ -177,6 +154,7 @@ type combinedClientsetInterface interface {
 	CoreV1() corev1typed.CoreV1Interface
 	AuthorizationV1() authorizationv1.AuthorizationV1Interface
 	RestClient() rest.Interface
+	MaxWorkers() int
 }
 
 // Need to use a type alias to embed the two Clientset's without a name clash.
@@ -189,6 +167,10 @@ type combinedClientset struct {
 
 func (c *combinedClientset) RestClient() rest.Interface {
 	return c.restCli
+}
+
+func (c *combinedClientset) MaxWorkers() int {
+	return int(c.restCli.GetRateLimiter().QPS())
 }
 
 // kubeHandler handles http requests for operating on app repositories and k8s resources
@@ -363,7 +345,7 @@ var ErrEmptyOCIRegistry = fmt.Errorf("You need to specify at least one repositor
 
 // NewHandler returns a handler configured with a service account client set and a config
 // with a blank token to be copied when creating user client sets with specific tokens.
-func NewHandler(kubeappsNamespace string, clustersConfig ClustersConfig) (AuthHandler, error) {
+func NewHandler(kubeappsNamespace string, burst int, qps float32, clustersConfig ClustersConfig) (AuthHandler, error) {
 	clientConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
 		clientcmd.NewDefaultClientConfigLoadingRules(),
 		&clientcmd.ConfigOverrides{
@@ -380,6 +362,16 @@ func NewHandler(kubeappsNamespace string, clustersConfig ClustersConfig) (AuthHa
 	config, err := clientConfig.ClientConfig()
 	if err != nil {
 		return nil, err
+	}
+	// Modify the default number of requests that the given client can do
+	// This is useful to handle a large number of namespaces which are check in parallel
+	// Burst is the initial number of request made in parallel
+	// Then further requests are performed following the QPS rate
+	if burst != 0 {
+		config.Burst = burst
+	}
+	if qps != 0 {
+		config.QPS = qps
 	}
 
 	svcRestConfig, err := rest.InClusterConfig()
@@ -808,11 +800,12 @@ func filterAllowedNamespaces(userClientset combinedClientsetInterface, namespace
 	allowedNamespaces := []corev1.Namespace{}
 
 	var wg sync.WaitGroup
-	checkNSJobs := make(chan checkNSJob, maxReq)
-	nsCheckRes := make(chan checkNSResult, maxReq)
+	workers := int(math.Min(float64(len(namespaces)), float64(userClientset.MaxWorkers())))
+	checkNSJobs := make(chan checkNSJob, workers)
+	nsCheckRes := make(chan checkNSResult, workers)
 
 	// Process maxReq ns at a time
-	for i := 0; i < maxReq; i++ {
+	for i := 0; i < workers; i++ {
 		wg.Add(1)
 		go func() {
 			nsCheckerWorker(userClientset, checkNSJobs, nsCheckRes)
@@ -831,7 +824,7 @@ func filterAllowedNamespaces(userClientset combinedClientsetInterface, namespace
 		close(checkNSJobs)
 	}()
 
-	// Start receiving tags
+	// Start receiving results
 	for res := range nsCheckRes {
 		if res.Error == nil {
 			if res.allowed {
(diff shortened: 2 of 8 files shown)


chart/kubeapps/README.md
```diff
@@ -275,6 +276,12 @@ You could alternatively ensure that the `imagePullSecret` is available in all na
 
 Yes! You can get started by following the [operators documentation](https://github.com/kubeapps/kubeapps/blob/master/docs/user/operators.md).
 
+### Slow response when listing namespaces
+
+Kubeapps uses the logged user account to retrieve the list of namespaces. If the user doesn't have permission to list namespaces, the fallback mechanism verifies if the user has permissions to get secrets for each namespace (to verify it it should be allowed to use that namespace or not). This can lead to a slow response if the number of namespaces of the cluster is big enough.
```
absoludity, line 281:
```suggestion
Kubeapps uses the currently logged-in user credential to retrieve the list of all namespaces. If the user doesn't have permission to list namespaces, the backend will try again with its own service account to list all namespaces and then iterate through each namespace to check if the user has permissions to get secrets for each namespace (to verify if they should be allowed to use that namespace or not and hence whether it is included in the selector). This can lead to a slow response if the number of namespaces on the cluster is large.
```

pkg/kube/kube_handler.go
```diff
@@ -753,24 +778,67 @@ func KubeappsSecretNameForRepo(repoName, namespace string) string {
...
+	go func() {
+		for _, ns := range namespaces {
+			checkNSJobs <- checkNSJob{ns}
 		}
-		if res.Status.Allowed {
-			allowedNamespaces = append(allowedNamespaces, namespace)
+		close(checkNSJobs)
+	}()
+
+	// Start receiving tags
```
absoludity, line 834: Copy-n-pasto :P

pkg/kube/kube_handler.go
```diff
@@ -43,6 +46,22 @@ import (
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 )
 
+var (
+	maxReq = 20
```
absoludity, line 50: Is there a reason you want a package variable here requiring initialization in an `init()` function? There are a bunch of reasons to avoid doing so ([Dave Cheney's thought experiment of removing package vars](https://dave.cheney.net/2017/06/11/go-without-package-scoped-variables) also outlines some of the why's).

Normally the main reason to do something like this (set a package var, use an `init()` fn to reach into the env and set the state, is because it's quicker to code (ie. short-term benefit). The one place where we do this usually is in our `main.go` to parse command-line args (which could be refactored, but not relevant to this PR).

I've got a strong preference to *remove* any precedents of package vars in our codebase when we can, not add to them. I think in this case, that'd not be too hard? Without having tried, it looks like we can update `NewClusterConfig` so that `maxRequests` is an option, and move the setting of this arg to the `main.go` which has the benefit of being both a cmd-line arg that is also set-able via an env var like the others. Then I think we can just pass it through the `handler.Options` to the handler and to `NewClusterConfig`? With that, it's an explicit cmd-line arg, rather than a magic way to reach into the state of the depths of the private handler library.

Is that OK? If you've got a good reason to sprinkle the setting of state in the depths of the app like this, let me know.

pkg/kube/kube_handler.go
```diff
@@ -100,6 +119,12 @@ func NewClusterConfig(inClusterConfig *rest.Config, userToken string, cluster st
 	config := rest.CopyConfig(inClusterConfig)
 	config.BearerToken = userToken
 	config.BearerTokenFile = ""
+	// Modify the default number of requests that the given client can do
+	// This is useful to handle a large number of namespaces which are check in parallel
+	// Burst is the initial number of request made in parallel
+	// Then further requests are performed following the QPS rate
+	config.Burst = maxReq * 2
```
absoludity, line 126: Does this mean we need to give the var a better name? If "Burst is the initial number of requests made in parallel" then it doesn't make sense to me to make it twice the `maxReq` value. Perhaps `queriesPerSecond`?

pkg/kube/kube_handler.go
```diff
@@ -753,24 +778,67 @@ func KubeappsSecretNameForRepo(repoName, namespace string) string {
...
+
+func filterAllowedNamespaces(userClientset combinedClientsetInterface, namespaces []corev1.Namespace) ([]corev1.Namespace, error) {
+	allowedNamespaces := []corev1.Namespace{}
+
+	var wg sync.WaitGroup
+	checkNSJobs := make(chan checkNSJob, maxReq)
+	nsCheckRes := make(chan checkNSResult, maxReq)
+
+	// Process maxReq ns at a time
+	for i := 0; i < maxReq; i++ {
```
absoludity, line 815: A couple of questions here:

1. Shouldn't this be only doing `min(maxReq, len(namespaces))` , rather than always `maxReq`? Or,
2. Given that we've already configured the burst of the client to `2*maxReq` perhaps `min(2*maxReq, len(namespaces))`?

That is, I'd expect the number of workers to match the burst, I thought?

pkg/kube/kube_handler.go
```diff
@@ -801,12 +870,15 @@ func (a *userHandler) GetNamespaces() ([]corev1.Namespace, error) {
...
-	if err != nil {
-		return nil, err
+		// Filter namespaces in which the user has permissions to write (secrets) only
+		namespaceList, err = filterAllowedNamespaces(a.clientset, namespaces.Items)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		// If the user can list namespaces, do not filter them
+		namespaceList = namespaces.Items
```
absoludity, line 881: Ah, so this explains why I thought the FAQ change didn't match the current behaviour - you're changing that behavior. Why? I mean, the fact that a user can list namespaces doesn't mean that it's useful to list all namespaces as if they can use them? I thought that the existing behaviour was like that for that reason.

If the ability to list a namespace was RBAC on an individual namespace, this would make sense, but given it's a cluster perm that just says you can list (all) namespaces regardless of whether you have RBAC to use them, I don't see how the change makes sense? Perhaps you mean that in practise, if people have the cluster RBAC to list namespaces, then they're most likely to have the ability to use them?

OK with me, just want the change to be intentional and understand the reason.

absoludity (commented):
> 
> Run namespace checks in parallel (the check that verifies if a user can access a namespace). Also, if the user has permissions to list namespaces, this check is skipped.

Keen to know the reason for the second change - see inline.

> 
> I did the following tests to verify the new performance:
> 
>     1. 200 namespaces, without this PR: 39s
> 
>     2. 200 namespaces, 20 parallel workers (the new default): 8.6s
> 
>     3. 200 namespaces, 100 parallel workers: 0.5s

Excellent, thanks for the data! Is this with a kind cluster?

> ### Possible drawbacks
> 
> Using too many workers might saturate the k8s API.

Thanks for making it configurable - I think that's a good approach given the possible drawback :)

pkg/kube/kube_handler.go
antgamdia: This is a typical tocket bucket algorithm. According to the k8s docs: 
_The rate limiter allows bursts of up to 'burst' to exceed the QPS, while still maintaining a smoothed qps rate of 'qps'. The bucket is initially filled with 'burst' tokens, and refills at a rate of 'qps'. The maximum number of tokens in the bucket is capped at 'burst'._

So, technically, in a single instant, you can perform up to `burst` requests (that is, it is the `maxReq` indeed).  However, I don't think you can make `burst` requests in the first second. Instead, you will have to wait `burst/qps`  seconds until you are able to perform `burst` requests.

As a result, `burst/qps` would be the time to wait until you can perform another spike of `burst ` requests. That is, by setting this value to `2`, we are allowing to exceed the `qps` each 2 seconds. So a user who has set the "maximum requests" will see that their value is being "ignored".

Here https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/ they use burst=30,qps=20, that is, 1.5s between spikes. However, the spike is just +50% of the qps, not the +100% as we are configuring here.

So, since it is a configuration that really depends upon the cluster hardware, I'd opt for making the qps also configurable and defaulting both params to (burst=30,qps=20).
I think a production-ready cluster would benchmark how much time do the requests take and, given that value, they would set the time between spikes.

pkg/kube/kube_handler.go
antgamdia: Yes, I agree, it should be `min(burst,len(namespaces))` however, I'm thinking right now in a multi-user scenario. Since we have configured our k8s client with these burst/qps params, if N users are trying to fetch their namespaces at the same time, we would have a maximum of `N*burst` requests in a single instant.

I have to check, but, do you know what is the behavior if the requests become throttled? Are they queued or an HTTP 429 is thrown? If the latter, we this function perhaps should also implement a kind of backoff mechanism, no?

pkg/kube/kube_handler.go
absoludity: > However, I don't think you can make `burst` requests in the first second. Instead, you will have to wait `burst/qps` seconds until you are able to perform `burst` requests.

Why? The docs, as you mentioned, say that you can "while still maintaining a smoothed qps rate of 'qps'". If you could only make `burst` requests in `burst/qps` seconds, then you'd not be exceeding the QPS at all, where as it explicitly says you can. My interpretation was that as long as the smoothed QPS rate matches the qps setting (say over 5 s - just as an example), then you can do a burst that exceeds `qps` requests in the first second as long as qps matches over 5 seconds (again, just an example, in reality it's probably based on the gradient getting back towards the set qps).

> As a result, `burst/qps` would be the time to wait until you can perform another spike of `burst ` requests. That is, by setting this value to `2`, we are allowing to exceed the `qps` each 2 seconds. So a user who has set the "maximum requests" will see that their value is being "ignored".

Right, this would be the result if there was no smoothing: if burst was 200, qps was 100, you could do an initial burst of 200 requests, but the qps wouldn't allow you to add any more until 2s. But my assumption is with smoothing this would allow further requests as long as the smoothed qps is going back down to the expected?

> So, since it is a configuration that really depends upon the cluster hardware, I'd opt for making the qps also configurable and defaulting both params to (burst=30,qps=20).

+1 to make both configurable, if Andres agrees (I don't feel strongly - would also be happy with 1.5*qps for burst), but regarding defaults, 30/20 is what `kube-controller-manager` uses as defaults (important part of the k8s control plane). Not sure we should use the same... I'd leave the defaults as they are [in the client-go code (DefaultQPS: 5, DefaultBurst: 10)](https://github.com/kubernetes/client-go/blob/master/rest/config.go#L44), as long as users can configure.

pkg/kube/kube_handler.go
absoludity: > Yes, I agree, it should be `min(burst,len(namespaces))` however, I'm thinking right now in a multi-user scenario. Since we have configured our k8s client with these burst/qps params, if N users are trying to fetch their namespaces at the same time, we would have a maximum of `N*burst` requests in a single instant.

Yep, theoretically possible. More reason to use the client-go defaults of 10/5

> 
> I have to check, but, do you know what is the behavior if the requests become throttled? Are they queued or an HTTP 429 is thrown? If the latter, we this function perhaps should also implement a kind of backoff mechanism, no?

Haven't dug, but see that there is [support for backoff built-in](https://github.com/kubernetes/client-go/blob/f6ce18ae578c8cca64d14ab9687824d9e1305a67/rest/client.go#L138).

pkg/kube/kube_handler.go
andresmgot: No, I believe there was a reason for us doing the filtering even if the user has permissions to list namespaces but I cannot recall why. 

IMO, we don't need to be smarter than the Kubernetes API here. If a user has permissions to list namespaces, why should we hide them? That's how all kubernetes dashboard work: allowing to list namespaces if the user has permissions and then return an error if they try do see/modify something they don't have permissions.

pkg/kube/kube_handler.go
andresmgot: I'm using `min(qps, len(ns))` here but note that it doesn't fix any of the other values (the request will take less than a second). This will just limit the number of  connections opened.

pkg/kube/kube_handler.go
andresmgot: Note that now (10qps/15burst), for 200 namespaces, it takes 19s to load... quite a lot.

antgamdia: > Why? The docs, as you mentioned, say that you can "while still maintaining a smoothed qps rate of 'qps'". If you could only make `burst` requests in `burst/qps` seconds, then you'd not be exceeding the QPS at all, where as it explicitly says you can. My interpretation was that as long as the smoothed QPS rate matches the qps setting (say over 5 s - just as an example), then you can do a burst that exceeds `qps` requests in the first second as long as qps matches over 5 seconds (again, just an example, in reality it's probably based on the gradient getting back towards the set qps).

It is a very interesting topic, I haven't checked the actual implementation, but I have assumed something like:
image
(image credits to Roman Voloboev)

Indeed, see an implementation from the k8s go client:
```go
// NewTokenBucketRateLimiter creates a rate limiter which implements a token bucket approach.
// The rate limiter allows bursts of up to 'burst' to exceed the QPS, while still maintaining a
// smoothed qps rate of 'qps'.
// The bucket is initially filled with 'burst' tokens, and refills at a rate of 'qps'.
// The maximum number of tokens in the bucket is capped at 'burst'.

```
https://github.com/kubernetes/client-go/blob/f6ce18ae578c8cca64d14ab9687824d9e1305a67/util/flowcontrol/throttle.go#L48-L52


> I can leave the default as qps: 10, burst: 15. I think that leaving numbers below that really hurts the UX for larger clusters.

I'm ok with these values as well

pkg/kube/kube_handler.go
```diff
@@ -356,6 +363,16 @@ func NewHandler(kubeappsNamespace string, clustersConfig ClustersConfig) (AuthHa
 	if err != nil {
 		return nil, err
 	}
+	// Modify the default number of requests that the given client can do
+	// This is useful to handle a large number of namespaces which are check in parallel
+	// Burst is the initial number of request made in parallel
+	// Then further requests are performed following the QPS rate
+	if burst != 0 {
```
absoludity, line 370: Not sure why we're handling the case where burst or qps == 0? That would make sense if our chart values defaulted to zero and we were using the k8s defaults, but we're not, so this would only have an effect if the chart was explicitly configured with these values as 0, which would mean we'd not set them in the config and hence use the k8s defaults? Is that what you meant? Otherwise, I'd just set them unconditionally? But fine either way.

pkg/kube/kube_handler.go
absoludity: Yeah, I'm pretty sure it was because people can have the RBAC to list namespaces without being able to use those namespaces. But yep - I don't mind either way, and it makes sense to list them if they have perms to do so, as you say.

absoludity (approved):
Great, thanks for the updates!

pkg/kube/kube_handler.go
absoludity: The graph isn't for a smoothed/rolling 5s window though, hence why you get spikes? I agree that if it's not maintaining a smoothed qps rate, but rather a fixed intervals (non-smoothed) then it would be spiked as you say - I'd just understood (without reading the implementation) that it was a smoothed qps rate in this case. But yep, interesting - I wish we did more basic DS and Algo discussions and learning :)

I'm also fine with the values. Thanks!

pkg/kube/kube_handler.go
andresmgot: yep, not needed, a leftover when it could be a env var.

mecampbellsoup:
@andresmgot @absoludity do you know when this will be released?

andresmgot:
We try to release every two weeks. We can try to cut a new patch version next week.

PR #2581, opened 2021-03-19
Add ability to set annotations to each individual deployment

kristinnardal2:
### Description of the change

The reason I want this supported is so that I can add these deployments
to my service mesh. Most service mesh decide on which deployments/pods
should be part of the mesh using annotations. I would like to be able to
control this via the Helm chart.

Service meshes can also allow you to add a whole namespace to the
service mesh. The reason this doesn't work well is that the sync jobs
created by kubeapps will never complete if they have a sidecar
container, which is what's done to add a pod to a service mesh.

This only adds support for adding annotations to the deployments, not
the pods created by those deployments. If we want to support that as
well, we could add another field called podAnnotations to each of these
components.

### Additional information

I also changed how the labels were added to the deployments, using the
"-" to trim the whitespace.

### DIFF
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -132,6 +132,9 @@ ingress:
 ##
 frontend:
   replicaCount: 2
+  ## Add additional annotations to the frontend deployment
+  ##
+  annotations: {}
   ## Bitnami Nginx image
   ## ref: https://hub.docker.com/r/bitnami/nginx/tags/
   ##
@@ -229,6 +232,9 @@ apprepository:
   ## Running a single controller replica to avoid sync job duplication
   ##
   replicaCount: 1
+  ## Add additional annotations to the controllers deployment
+  ##
+  annotations: {}
   ## Schedule for syncing apprepositories. Every ten minutes by default
   # crontab: "*/10 * * * *"
   ## Bitnami Kubeapps AppRepository Controller image
@@ -348,6 +354,9 @@ hooks:
 # Kubeops is an interface between the Kubeapps Dashboard and Helm 3/Kubernetes.
 kubeops:
   replicaCount: 2
+  ## Add additional annotations to the kubeops deployment
+  ##
+  annotations: {}
   image:
     registry: docker.io
     repository: kubeapps/kubeops
@@ -396,6 +405,9 @@ kubeops:
 ##
 assetsvc:
   replicaCount: 2
+  ## Add additional annotations to the assetsvc deployment
+  ##
+  annotations: {}
   ## Bitnami Kubeapps Assetsvc image
   ## ref: https://hub.docker.com/r/bitnami/kubeapps-assetsvc/tags/
   ##
@@ -461,6 +473,9 @@ assetsvc:
 ##
 dashboard:
   replicaCount: 2
+  ## Add additional annotations to the dashboard deployment
+  ##
+  annotations: {}
   ## Bitnami Kubeapps Dashboard image
   ## ref: https://hub.docker.com/r/bitnami/kubeapps-dashboard/tags/
   ##
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -132,9 +132,9 @@ ingress:
 ##
 frontend:
   replicaCount: 2
-  ## Add additional annotations to the frontend deployment
+  ## Add additional annotations to the frontend pods
   ##
-  annotations: {}
+  podAnnotations: {}
   ## Bitnami Nginx image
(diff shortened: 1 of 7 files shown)


absoludity:
Thanks @kristinnardal2 . I just took a look at the [bitnami postgresql chart](https://hub.kubeapps.com/charts/bitnami/postgresql) to see whether this is consistent with changes there. I notice that in that chart there are `.annotations` for services, PVCs and statefulsets, while there's `.podAnnotations` for pods and `commonAnnotations` that are applied to all resources (not useful for you).

I just want to check with @juan131 whether he thinks it's fine to add an `.annotations` for deployments and if that's done for other charts in the bitnami catalog (even if it's not done for the postgres chart). Otherwise, LGTM :)

kristinnardal2:
I'm happy to change it to `.podAnnotations` for consistency.
For my purposes it's fine to have it on the pod level instead of at the deployment level.

absoludity:
> I'm happy to change it to `.podAnnotations` for consistency.
> For my purposes it's fine to have it on the pod level instead of at the deployment level.

Great. If we don't hear from Juan in the next day or so, let's update to that as it's consistent with other bitnami charts.

juan131:
In other [Bitnami charts](https://github.com/bitnami/charts), we're using two parameters (`commonLabels` and `commonAnnotations`) as a way to provide custom labels/annotations to *every manifest* available in the chart.

Then, there are specific parameters for specific manifests. For instance, check the example below using this approach and adding custom annotations for dashboard service:

- **values.yaml**:

```yaml
## Add labels to all the deployed resources (evaluated as a template)
##
commonLabels: {}
## Add annotations to all the deployed resources (evaluated as a template)
##
commonAnnotations: {}
...
dashboard:
  ...
  ## Service configuration
  ##
  service:
    ...
    ## Provide any additional annotations which may be required (evaluated as a template)
    ##
    annotations: {}
```

- **templates/dashboard-service.yaml**:

```yaml
...
metadata:
  ...
  name: {{ template "kubeapps. dashboard.fullname" . }}
  labels:
    app: {{ template "common.names.name" . }}
    {{- if .Values.commonLabels }}
    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
    {{- end }}
  {{- if or .Values.dashboard.service.annotations .Values.commonAnnotations }}
  annotations:
    {{- if .Values.dashboard.service.annotations }}
    {{- include "common.tplvalues.render" ( dict "value" .Values.dashboard.service.annotations "context" $ ) | nindent 4 }}
    {{- end }}
    {{- if .Values.commonAnnotations }}
    {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
    {{- end }}
  {{- end }}
...
```

absoludity (approved):
I think @kristinnardal2 's use-case meant that they didn't want to apply the annotations to all resources (specifically the jobs and related pods done by the batches, but perhaps others), hence trying to limit it to the deployments. Anyway, I'm giving my +1 as is to the `.annotations` on a deployment, just like the bitname postgres chart already does for other things like services, stateful sets and pvc's. Note that *unlike* the postgresql chart, the value change here is a little ambiguous (ie. is `apprepository.annotations` always going to be just for the deployment, or could it be for the service as well - we could later add `apprepository.service.annotations` I guess). I'll leave it until @andresmgot  is around to add his thoughts and potentially land.

chart/kubeapps/values.yaml
```diff
@@ -132,6 +132,9 @@ ingress:
 ##
 frontend:
   replicaCount: 2
+  ## Add additional annotations to the frontend deployment
+  ##
+  annotations: {}
```
absoludity, line 137: So my only qualm here is that we're using `frontend.annotations` and `apprepository.annotations` to mean something which is really just annotations for the deployment, not the frontend (service, for eg). In this sense, `podAnnotations` would be specific and not need to change in the future.

kristinnardal2: Let's just change this to podAnnotation, and move the annotations to the pod level instead of the deployment level.
This way we conform with what is being done in other charts, and for me it's an equally good solution.

I'll do this change when I have the time, might be able to do it today.

absoludity: Perfect, thanks @kristinnardal2

kristinnardal2: I've done this, I also noticed an error while doing this, adding multiple annotations would have failed because the indentation wasn't correct. I have added a `nindent` to fix this issue.

andresmgot (approved):
LGTM, thanks for the changes!

Issue #2328, opened 2021-02-05
wss requests are 403 when using pinniped


absoludity:
### Description:

It appears that they're not being routed (or there's some issue routing) the wss requests via pinniped. It's not a show-stopper since we fall back to the polls.

Example:
![wss-403s](https://user-images.githubusercontent.com/497518/106996084-57480f00-67d4-11eb-8d80-882848b8135b.png)

I'll need to investigate more, just noting for now.

andresmgot:
why do you say it's not being routed? I guess you checked but being a 403 it may mean that you don't have permissions to `watch` resources.

absoludity:
> why do you say it's not being routed? I guess you checked but being a 403 it may mean that you don't have permissions to `watch` resources.

From memory I didn't even see the requests even hitting pinniped-proxy, but I'll re-check when getting back to this.

absoludity:
Great, thanks Antonio.

antgamdia:
Well, I've been investigating this issue a little bit more, and I'm afraid I don't have a solution right now...

As far as I know, the HTTP connection to `...?watch=true` with requesting moving to websockets is properly being routed (hence the `101 switching protocols` response), however, our pinniped-proxy logic is not handling the messages that may come from the k8s API.

For each connection, we are doing  `make_service_fn(|_conn| { .... ` and `service_fn(move |req| { service::proxy(req, default_ca_data.clone().into_bytes()) }))`, which, after many certs/tls/create clients/etc stuff, it simply does:

https://github.com/kubeapps/kubeapps/blob/32976944d2074a1e1e319b2fb24e4aef12fcee65/cmd/pinniped-proxy/src/service.rs#L83-L89

That is, requesting data (pinniped-proxy -> k8s) with the proper credentials and then, returning back to the actual client (pinniped-proxy->nginx->user's browser).

This is being executed per connection. So, essentially, given each -one- connection it will return -one- response.
If we were to support websockets here, I guess (not sure) we should keep open as many connections as the client  opens (until a close event or timeout), as well as handling the responses from the server.

[This is server logic needed](https://github.com/1tgr/rust-websocket-lite/blob/master/hyper-websocket-lite/src/lib.rs) for accepting the change to websockets. However, this is not necessary at all, since the real k8s is accepting the change (we have the 101 status code)

[...]

PR #2588, opened 2021-03-22
Fail when proxing ws in pinniped proxy

antgamdia:
### Description of the change

As per the discussion in https://github.com/kubeapps/kubeapps/issues/2328#issuecomment-803984871, this PR simply adds an explicit 5xx-error when proxying a websocket connection attempt (aka 101 HTTP status code).

### Benefits

The polling fallback mechanism now will be triggered.

### Possible drawbacks

An error msg will be thrown in the js console.

### Applicable issues

  - rel #2328

### Additional information

N/A

### DIFF
--- a/cmd/pinniped-proxy/src/service.rs
+++ b/cmd/pinniped-proxy/src/service.rs
@@ -83,7 +83,12 @@ pub async fn proxy(mut req: Request<Body>, default_ca_data: Vec<u8>) -> Result<R
     match client.request(req).await {
         Ok(r) => {
             info!("{}", logging::response_log_data(&r, log_data));
-            Ok(r)
+            if r.status() == StatusCode::SWITCHING_PROTOCOLS {
+                // TODO(agamez): implement ws proxing here; see https://github.com/kubeapps/kubeapps/issues/2328
+                Ok(Response::builder().status(StatusCode::NOT_IMPLEMENTED).body(Body::from("pinniped-proxy does not support websockets yet")).unwrap())
+            } else {
+                Ok(r)
+            }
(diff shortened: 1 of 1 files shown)


cmd/pinniped-proxy/src/service.rs
```diff
@@ -83,7 +83,12 @@ pub async fn proxy(mut req: Request<Body>, default_ca_data: Vec<u8>) -> Result<R
     match client.request(req).await {
         Ok(r) => {
             info!("{}", logging::response_log_data(&r, log_data));
-            Ok(r)
+            if r.status() == StatusCode::SWITCHING_PROTOCOLS {
+                // TODO(agamez): implement ws proxing here; see https://github.com/kubeapps/kubeapps/issues/2328
+                Ok(Response::builder().status(StatusCode::NOT_IMPLEMENTED).body(Body::from("pinniped-proxy does not support websockets yet")).unwrap())
```
absoludity, line 88: Great, thanks Antonio. I guess we could do this response without even needing to try proxying the request, but totally fine to do here based on the proxied-response.

absoludity (approved):

cmd/pinniped-proxy/src/service.rs
antgamdia: Yes, this situation can be retrieved from the request as stated by the [RFC 6455](https://tools.ietf.org/html/rfc6455#section-1.2).; it is just a matter of detecting, for example, a header `Connection: Upgrade`. However, it was a little bit easier for me just to compare using the status code from the response.

PR #2589, opened 2021-03-22
Add rustfmt to normalize formatting

antgamdia:
### Description of the change

Similary, as we already to in js and go (I guess), this PR brings the configuration for the rust prettifier, the `rustfmt` tool. It implements the [Rust Style Guide](https://github.com/rust-dev-tools/fmt-rfcs/blob/master/guide/guide.md).

### Benefits

We'll have a consistent code formatting config so that multiple devs have the same linitng rules.

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

I've used the default config for the generation of the `.rustfmt.toml` feel free to suggest other configurations.


cmd/pinniped-proxy/.rustfmt.toml
```diff
@@ -0,0 +1,68 @@
+max_width = 100
```
absoludity, line 1: If these are all defaults, can we not specify them instead? (ie. I think it's much easier to reason about if this file only has the exceptions to the defaults that we want, if any)

absoludity (approved):
Thanks! I actually thought I had rustfmt running in vscode on save, but obviously not :)

cmd/pinniped-proxy/.rustfmt.toml
antgamdia: Yes, I agree. I just wanted to commit all the options to discuss if we should change something else.
However, these options can be retrieved from `rustfmt --print-config default`

> Thanks! I actually thought I had rustfmt running in vscode on save, but obviously not :)

I thought the same, but I broke something in my vscode config and it stopped working. Then I realized I was using the default formatter in vscode and not the rustfmt one.

PR #2590, opened 2021-03-22
Remove helm hooks from apprepositories manifests

andresmgot:
### Description of the change

Remove some of the hooks of the Kubeapps chart:
 
 - chart/kubeapps/templates/apprepositories.yaml: The apprepositories are now installed with the rest of resources. This is okay as far as the CRD is installed before. I have checked the process with Helm v3.5 (the CI will try with v3.1) and the CRD gets installed before the instances.
 - chart/kubeapps/templates/apprepositories-secret.yaml: These are the secrets for the above. Again no issue just removing the the hook.
 - chart/kubeapps/templates/apprepository-jobs-cleanup-*: These are the manifests related to the job that removes global apprepositories. Since the apprepos are now not created in hooks, it's not necessary to clean them up manually. AppRepositories created through the dashboard will persist after uninstalling Kubeapps.

### Possible drawbacks

This change (and the ones to come) cause a breaking change from the chart perspective so I will gather changes in this branch before increasing the major version (and releasing).

### Applicable issues

  - ref #2240


script/e2e-test.sh
```diff
@@ -196,10 +198,9 @@ kubectl create ns kubeapps
 if [[ -n "${TEST_UPGRADE}" ]]; then
   # To test the upgrade, first install the latest version published
   info "Installing latest Kubeapps chart available"
-  installOrUpgradeKubeapps bitnami/kubeapps
-  # Due to a breaking change in PG chart 9.X, we need to delete the statefulset before upgrading
-  # This can be removed after the release 2.0.0
-  kubectl delete statefulset -n kubeapps --all
+  # Breaking change at 6.X, the initialRepos are no longer in a hook
+  installOrUpgradeKubeapps bitnami/kubeapps \
+    --set apprepository.initialRepos=null
```
absoludity, line 203: Why are you setting the initial repos to null rather than using the default bitnami one?

absoludity (approved):
>     * chart/kubeapps/templates/apprepositories.yaml: The apprepositories are now installed with the rest of resources. This is okay as far as the CRD is installed before. I have checked the process with Helm v3.5 (the CI will try with v3.1) and the CRD gets installed before the instances.

Great - I guess this was still from when we needed to support 2.x and 3.x? (since 3.x installs CRDs first)

> ### Possible drawbacks
> 
> This change (and the ones to come) cause a breaking change from the chart perspective so I will gather changes in this branch before increasing the major version (and releasing).

Sounds good, though not sure how this change would be breaking (other than if you uninstall after upgrade, certain resources will remain).

One question inline about why you're trying to set the app repos to null, otherwise +1

script/e2e-test.sh
absoludity: Also, ```Error: unknown flag: --set apprepository.initialRepos

```

Comment (andresmgot) [reply script/e2e-test.sh:203]: this is the breaking change (which I failed to fix in the first try :P). When installing the previous version (`bitnami/kubeapps`), the initialRepos are installed in a hook, which is not handled by helm anymore. When upgrading, now it's a normal resource so it's trying to install it again, which causes a conflict error:

```
Error: UPGRADE FAILED: rendered manifests contain a new resource that already exists. Unable to continue with update: existing resource conflict: namespace: kubeapps, name: bitnami, existing_kind: kubeapps.com/v1alpha1, Kind=AppRepository, new_kind: kubeapps.com/v1alpha1, Kind=AppRepository
```

So people will need to uninstall and install kubeapps again to avoid that issue ^ (or delete the initial repos manually)

absoludity: OK, I think the comment needs some expanding then :) From what I thought helm only sees this as a conflict if the existing AppRepository objects aren't labeled as helm-managed (summarised in the comment at https://github.com/helm/helm/issues/6646#issuecomment-547650430 ), but if they were installed in a hook, I would have thought they'd have the correct labeling etc., so I'm unsure why we're seeing a conflict for them?

andresmgot: the issue you point is to a change in the group version, which is not the case here? I don't know if there is some magic label that would have help us with this but since it's not there in the previous version, there is not much we can do now. I will update the comment.

PR #2595, opened 2021-03-23
Remove DB secret management in hooks

### LINKED ISSUE #2240: Remove Helm exclusive features from the chart

andresmgot:
### Description of the change

Follow up of #2590

This PR removes the workaround used to automatically generate a password for PostgreSQL database and keep it during upgrades.

This is now following the Bitnami standard which is, set a random password when first installing the application and fail if the password is not given when upgrading. Example:

```
▶ helm install kubeapps ./chart/kubeapps -n kubeapps2                                                                                      
NAME: kubeapps
LAST DEPLOYED: Tue Mar 23 16:16:10 2021
NAMESPACE: kubeapps2
STATUS: deployed
REVISION: 1
NOTES:
** Please be patient while the chart is being deployed **

...

#####################################################################################
### WARNING: You did not change the default password for the PostgreSQL root user ###
#####################################################################################

▶ helm upgrade kubeapps ./chart/kubeapps -n kubeapps2                                                                                      
Error: UPGRADE FAILED: template: kubeapps/templates/NOTES.txt:73:4: executing "kubeapps/templates/NOTES.txt" at <include "common.errors.upgrade.passwords.empty" (dict "validationErrors" $passwordValidationErrors "context" $)>: error calling include: template: kubeapps/charts/common/templates/_errors.tpl:18:48: executing "common.errors.upgrade.passwords.empty" at <fail>: error calling fail: 
PASSWORDS ERROR: you must provide your current passwords when upgrade the release
    'postgresql.postgresqlPassword' must not be empty, please add '--set postgresql.postgresqlPassword=$POSTGRESQL_PASSWORD' to the command. To get the current value:

        export POSTGRESQL_PASSWORD=$(kubectl get secret --namespace "kubeapps2" kubeapps-postgresql -o jsonpath="{.data.postgresql-password}" | base64 --decode)

▶ export POSTGRESQL_PASSWORD=$(kubectl get secret --namespace "kubeapps2" kubeapps-postgresql -o jsonpath="{.data.postgresql-password}" | base64 --decode)

▶ helm upgrade kubeapps ./chart/kubeapps -n kubeapps2 --set postgresql.postgresqlPassword=$POSTGRESQL_PASSWORD
# success
```

### Possible drawbacks

Another breaking change. From now on, people will need to manually set the PostgreSQL password (sooner or later).

### Applicable issues

  - ref #2240


antgamdia:
>   set a random password when first installing the application and fail if the password is not given when upgrading. (...) From now on, people will need to manually set the PostgreSQL password (sooner or later).

So (just to confirm), by default, a password will be automatically generated when installing the chart for the very first time. However, if the user wants to upgrade, they will need to know in advance this generated password, isn't it?

In that case, I would add the command to retrieve the secret during the chart installation and a message saying they will need that password if they want to upgrade the chart in the future.

andresmgot:
> So (just to confirm), by default, a password will be automatically generated when installing the chart for the very first time. However, if the user wants to upgrade, they will need to know in advance this generated password, isn't it?

Correct.

> In that case, I would add the command to retrieve the secret during the chart installation and a message saying they will need that password if they want to upgrade the chart in the future.

We can add it but the command is already printed when trying to upgrade. If we print it after the installation it's not that useful because you won't remember that once it's time to upgrade (I guess that's why the bitnami charts don't print it after the installation).

chart/kubeapps/values.yaml
```diff
@@ -542,19 +542,22 @@ dashboard:
...
 ##
 postgresql:
+  # Enable PostgreSQL. This should be true
+  enabled: true
   ## Enable replication for high availability
   replication:
     enabled: true
   ## Create a database for Kubeapps on the first run
   postgresqlDatabase: assets
+  ## PostgreSQL admin password (used when `postgresqlUsername` is not `postgres`)
```
absoludity, line 552: Confused - `postgresqlPassword` is an admin password iff `postgresqlUsername` *is* `postgres` . It is only when `postgresqlUsername` is not `postgres` that `postgresqlPassword` will be the non-admin password and `postgresqlPostgresqlPassword` is used for the admin password. So I'm confused by the above comment, even though I assume you grabbed it from other charts?

chart/kubeapps/values.yaml
```diff
@@ -542,19 +542,22 @@ dashboard:
...
+  ## ref: https://github.com/bitnami/bitnami-docker-postgresql/blob/master/README.md#creating-a-database-user-on-first-run (see note!)
+  ##
+  postgresqlPassword: ""
```
absoludity, line 555: Can we do:
```
postgresqlPassword:
```
instead? (ie. not explicitly put empty quotes, as even though it's functionally the same, it looks like we're overriding it to be an empty string, rather than leaving it unset as it is in the postgresql chart)

absoludity (approved):
> This is now following the Bitnami standard which is, set a random password when first installing the application and fail if the password is not given when upgrading.

I think the Bitnami standard is to support both a random password when first installing, or use an existing secret when installing or upgrading, right?

> > set a random password when first installing the application and fail if the password is not given when upgrading. (...) From now on, people will need to manually set the PostgreSQL password (sooner or later).
> 
> So (just to confirm), by default, a password will be automatically generated when installing the chart for the very first time. However, if the user wants to upgrade, they will need to know in advance this generated password, isn't it?
> 
> In that case, I would add the command to retrieve the secret during the chart installation and a message saying they will need that password if they want to upgrade the chart in the future.

I think we can avoid that and instead encourage people to use `existingSecret` when upgrading. Seems very odd to require people to pull the secret from the cluster to upgrade it again (and unnecessary from the little testing I did, but I'll bring that up internally). Given that this PR just relies on the bitnami/postgresql charts current help text, +1, but I'd like to see that current help text changed.

andresmgot:
> I think the Bitnami standard is to support both a random password when first installing, or use an existing secret when installing or upgrading, right?

Right, both things will be supported (just that the second is less often used).

> > > set a random password when first installing the application and fail if the password is not given when upgrading. (...) From now on, people will need to manually set the PostgreSQL password (sooner or later).
> > 
> > 
> > So (just to confirm), by default, a password will be automatically generated when installing the chart for the very first time. However, if the user wants to upgrade, they will need to know in advance this generated password, isn't it?
> > In that case, I would add the command to retrieve the secret during the chart installation and a message saying they will need that password if they want to upgrade the chart in the future.
> 
> I think we can avoid that and instead encourage people to use `existingSecret` when upgrading. Seems very odd to require people to pull the secret from the cluster to upgrade it again (and unnecessary from the little testing I did, but I'll bring that up internally). Given that this PR just relies on the bitnami/postgresql charts current help text, +1, but I'd like to see that current help text changed.

As discussed offline, setting `existingSecret` only when upgrading causes the secret to be deleted.

chart/kubeapps/templates/NOTES.txt
```diff
@@ -56,3 +58,16 @@ To access Kubeapps from outside your K8s cluster, follow the steps below:
 {{- end }}
 
 2. Open a browser and access Kubeapps using the obtained URL.
+
+{{ if and (not .Values.postgresql.existingSecret) (empty .Values.postgresql.postgresqlPassword) -}}
+#####################################################################################
+### WARNING: You did not change the default password for the PostgreSQL root user ###
```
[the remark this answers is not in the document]
andresmgot: right, I took the text from a chart that included a default password.

chart/kubeapps/values.yaml
andresmgot: right, no, my bad, I copied the comment from a `postgresqlPostgresqlPassword`

Morriz:
Guys, I want to strongly suggest that configuration generates idempotent payloads. So no side effects should happen related to a difference in cluster state. We want any apply with same config result in the same resources being deployed.

So please drop any in-cluster jobs with state dependent logic.

Imo idempotency should be a very high up principle when developing 12-factor apps.

andresmgot:
Yes @Morriz, this PR doesn't affect that. The state of the cluster won't affect the installation. 

The ultimate goal of these PRs (see others linked) is to remove any Helm-specific logic that would differ from a statically generated manifest.

absoludity:
> As discussed offline, setting `existingSecret` only when upgrading causes the secret to be deleted.

Not if the generated secret is explicitly annotated with `helm.sh/resource-policy: keep`. But anyway, this is related to the postgresql chart (and bitnami charts in general), not this change which just accepts whatever those charts do.

> Guys, I want to strongly suggest that configuration generates idempotent payloads. So no side effects should happen related to a difference in cluster state. We want any apply with same config result in the same resources being deployed.

Yep, that's the purpose of removing the helm hooks and other related jobs. Note that helm, being helm, will still do more than just apply your config. That is, what's applied is not just a fn of your input config, but also what the previous helm release included (as it deletes resources that were included in the previous release but are no longer included, for example). Not so bad as using cluster state (and we won't be using `lookup`, I hope).

PR #2596, opened 2021-03-23
fix: missing fsGroup resulted in invalid yaml output

Morriz:
### Description of the change

With securityContext.enabled the output would be invalid due to a missing default value for `securityContext.fsGroup`. This is now set. 

I also added jsonschema for the touched section of the values.

NOTE: these seem to be not maintained, and I suggest to put this in your DoD (if you do scrum that is).

### Benefits

working defaults for `securityContext.enabled: true`

### Possible drawbacks

none


chart/kubeapps/values.schema.json
```diff
@@ -1,135 +1,159 @@
...
         }
-      }
+      },
+      "title": "Ingress configuration"
     },
-    "authProxy": {
-      "type": "object",
-      "title": "OIDC Proxy configuration",
+    "securityContext": {
+      "form": true,
```
andresmgot, line 129: I would not add the `securityContext` to the simple form. We want to keep this form as simple as possible, only including fields that people modify very often. In this case, the securityContext is not that "popular". I am okay leaving this defintion so Helm will validate the values but removing the `form` (and `hidden`) keys.

Just in case you are not aware of what `form: true` does, this is the documentation: https://github.com/kubeapps/kubeapps/blob/master/docs/developer/basic-form-support.md

chart/kubeapps/values.schema.json
```diff
@@ -1,135 +1,159 @@
...
+          "title": "File System User ID",
+          "type": "integer"
         }
-      }
+      },
+      "title": "Frontend configuration"
     }
   }
-}
+}
```
andresmgot, line 159: missing empty line at the end

andresmgot:
> NOTE: these seem to be not maintained, and I suggest to put this in your DoD (if you do scrum that is).

For the record, we only modify this if the items to be shown in the form are modified (that's why there are such few changes there).

chart/kubeapps/values.schema.json
Morriz: Ok. I suggest to include enforcement for formatting, such as prettier.

chart/kubeapps/values.schema.json
```diff
@@ -1,135 +1,147 @@
 {
   "$schema": "http://json-schema.org/schema#",
-  "type": "object",
   "properties": {
-    "frontend": {
-      "type": "object",
```
andresmgot, line 6: I have tried this and while `type: object` is not necessary from a JSONSchema point of view, we are using it to render subsections. Without it the form is not properly rendered:

Screenshot from 2021-03-24 16-39-01

while it should be: 

Screenshot from 2021-03-24 16-39-26

Also, before, a human placed the inputs so it to have a specific order but here it's alphabetically ordered. I prefer not to follow the alphabetical order because some inputs makes more sense in a specific order. Can you just add the new section leaving the rest untouched?

Thanks!

Morriz: Sure.

andresmgot (approved):
Thanks for the changes!

Issue #2078, opened 2020-09-29
Investigate app repositories post 2.0 with hub support etc.


absoludity:
Currently AppRepositories are not only helm-centric but also helm2 centric, where an index.yaml indexed a collection of charts.

With Helm 3, there has been a push towards 'hubs' which can be searched (rather than an index.yaml) without providing a static list of available charts. Furthermore, charts are now installable with a direct URL, which is more inline with charts in an OCI repository.

Finally, if Kubeapps moves to generalise packaging formats (enabling plugins for different formats #2023) we want to plan how an "AppRepository" in the future can model "repositories" of different packaging formats, if at all (it may be more reason to follow helm and move towards searchable hubs generally).

As part of this investigation, we may want to pursue #1178 since we may not need a controller running necessarily anyway.

ppbaena:
To be closed as there is a new approach for app repositories described in #4764

Issue #1560, opened 2020-03-05
Invalidate storage cache only when necessary


absoludity:
Up until recently, mongodb was the only storage option for the assets (charts, files etc.) which Kubeapps indexes. Postgresql was added as a supported backend, but we don't currently support migrations, or even just the ability to know if there's been a schema change.

As a result, and since the storage is just a cache, currently we invalidate the cache and re-create the schema on each upgrade to be sure that the code is working with the correct schema. A less destructive approach would be to only invalidate the cache when a schema change is detected, and the ideal would be to migrate data so that other users don't temporarily see a change in charts etc.

It may be simpler to do this when we only have postgresql as a backend (as even though mongodb doesn't necessarily require resetting the schema, the indexes may also change).

absoludity:
I see the artifacthub folk are using https://github.com/jackc/tern which looks interesting: postgres specific, can be used both as a stand-alone CLI or a go library in your code.

andresmgot:
IMO, running migrations is an ideal solution but maybe is an overkill for our use case. Maintaining the migration seems not trivial. (At the very least we would need to maintain the migration file and configuration and learn how to run these migrations).

If we are able to at least detect schema changes and only delete the database in that case we will be in a better place than right now (and doing that seems more simple).

absoludity:
Maintaining migrations is pretty normal - plenty of tools for this. Running the migrations could simply be running them when the pod starts (one pod will have a transaction and do the migration, others see the migration already done). But I agree - could be more effort than value for this case, as I mentioned in the description.

> If we are able to at least detect schema changes and only delete the database in that case we will be in a better place than right now (and doing that seems more simple).

Yep, as above, it's an improvement happy for us to do that for now, but I do think there are ways of doing that which are a step towards standard migrations. I'll put more context on the PR.

absoludity:
> Yep, as above, it's an improvement happy for us to do that for now, but I do think there are ways of doing that which are a step towards standard migrations. I'll put more context on the PR.

After commenting on the PR, I even wonder if it'd be less effort just to use `tern` either way.

andresmgot:
We are not invalidating the cache after the chart 6.0.0 so we can close this one for the moment (to be re-evaluated if we need to enable the invalidation again).

absoludity:
> We are not invalidating the cache after the chart 6.0.0 

It's more that we are not supporting schema changes to the cache/storage.

[...]

PR #2597, opened 2021-03-24
Move post-upgrade job to the apprepo-controller

andresmgot:
### Description of the change

Follow up of #2590

~~Move the post-upgrade job to an init container in the apprepository-controller so we avoid using helm hooks.~~

Remove the post-upgrade job we avoid using helm hooks. We haven't add any schema change for a long time so it's not needed to run this job for the moment.

### Possible drawbacks

~~If the apprepository-controller gets rescheduled to a different node, the database will be wiped out.~~

If someone updates from a very old version of kubeapps (prior to the chart version 3.4.3 (Feb 2020)) directly to the latest 6.x, it won't work but that's not supported anyway.

### Applicable issues

  - ref #2240


absoludity:
> If the apprepository-controller gets rescheduled to a different node, the database will be wiped out.

Is that the only time? Won't this run for every pod that starts, possibly conflicting. Also if I scale up.

I think we could try a bit harder to improve here. The first thing that comes to mind is that we should be running migrations in a transaction in the init container (or on startup) like most apps with a db (something I've been keen for since we switched from mongo to pg). This would be idempotent and not blow away the db when there is no need. We've avoided that because it's just a(n expensive) cache of data really.

Or can you think of another way that we can ensure the db is updated when required during an upgrade?

andresmgot:
> > If the apprepository-controller gets rescheduled to a different node, the database will be wiped out.
> 
> Is that the only time? Won't this run for every pod that starts, possibly conflicting. Also if I scale up.

Yes, every time the pod starts but this deployment should not be scaled up (since it will process apprepos several times).

> I think we could try a bit harder to improve here. The first thing that comes to mind is that we should be running migrations in a transaction in the init container (or on startup) like most apps with a db (something I've been keen for since we switched from mongo to pg). This would be idempotent and not blow away the db when there is no need. We've avoided that because it's just a(n expensive) cache of data really.
> 
> Or can you think of another way that we can ensure the db is updated when required during an upgrade?

We discussed that in this issue: https://github.com/kubeapps/kubeapps/issues/1560. I don't think we should block this for that? In any case, I am okay prioritizing that issue since it includes "downtime" every time Kubeapps is upgraded.

andresmgot:
Actually... we can just not run the `invalidate-cache` at all? It's been more than a year since the schema last changed. There has been several major version updates since then. If someone is running a really old instance of Kubeapps, they should upgrade `4.X -> 5.x -> 6.x` which won't require that `6.x` run the invalidate-cache job.

I am reluctant to spend more time in the PG integration since I believe we should move away from using our database to store this kind of info (vs using a CR for example). Kind of related to https://github.com/kubeapps/kubeapps/issues/2078 and multicluster support.

absoludity:
> 
> Yes, every time the pod starts but this deployment should not be scaled up (since it will process apprepos several times)

K8s controllers are designed to be able to have multiple controllers watching - they use a workqueue from which only one controller will pull a specific event to handle.

> Actually... we can just not run the `invalidate-cache` job at all? It's been more than a year since the schema last changed. There has been several major version updates since then. If someone is running a really old instance of Kubeapps, they should upgrade `4.X -> 5.x -> 6.x` which won't require that `6.x` run the invalidate-cache job.
> 
> I am reluctant to spend more time in the PG integration since I believe we should move away from using our database to store this kind of info (vs using a CR for example). Kind of related to #2078 and multicluster support.
> 
> In any case, we can re-add the `invalidate-cache` job if we change the schema (or implement proper migrations).

Yep, we can. It just seems that we always push this back. I agree that #2078 may result in not needing a DB (possibly), so worth avoiding that work now.

absoludity (approved):

andresmgot:
> K8s controllers are designed to be able to have multiple controllers watching - they use a workqueue from which only one controller will pull a specific event to handle.
> 

Yes, but our controller is not adapted to do so (as far as I can tell), that's why it is the only deployment in our chart that has 1 replica. I wanted to verify this and got something interesting. When starting, it seems that both replicas process the apprepo (`bitnami`):

```
▶ kubectl logs -n kubeapps kubeapps-internal-apprepository-controller-766d86fd87-x9mrg
...
time="2021-03-26T07:25:50Z" level=info msg="Updating CronJob \"apprepo-kubeapps-sync-bitnami\" in namespace \"kubeapps\" for AppRepository \"bitnami\" in namespace \"kubeapps\""
time="2021-03-26T07:25:50Z" level=info msg="Successfully synced 'kubeapps/bitnami'"

▶ kubectl logs -n kubeapps kubeapps-internal-apprepository-controller-766d86fd87-2p6pn
...
time="2021-03-26T07:25:50Z" level=info msg="Updating CronJob \"apprepo-kubeapps-sync-bitnami\" in namespace \"kubeapps\" for AppRepository \"bitnami\" in namespace \"kubeapps\""
time="2021-03-26T07:25:50Z" level=info msg="Successfully synced 'kubeapps/bitnami'"
``` 

Then I refreshed the repository but got an error in both pods:

```
time="2021-03-26T07:28:23Z" level=info msg="Event(v1.ObjectReference{Kind:\"AppRepository\", Namespace:\"kubeapps\", Name:\"bitnami\", UID:\"307be21f-18b8-40ff-841f-1f82211008cf\", APIVersion:\"kubeapps.com/v1alpha1\", ResourceVersion:\"21770\", FieldPath:\"\"}): type: 'Normal' reason: 'Synced' AppRepository synced successfully"
E0326 07:28:23.271995       1 event.go:264] Server rejected event '&v1.Event{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ObjectMeta:v1.ObjectMeta{Name:"bitnami.166fd3b16bc988f7", GenerateName:"", Namespace:"kubeapps", SelfLink:"", UID:"", ResourceVersion:"21476", Generation:0, CreationTimestamp:v1.Time{Time:time.Time{wall:0x0, ext:0, loc:(*time.Location)(nil)}}, DeletionTimestamp:(*v1.Time)(nil), DeletionGracePeriodSeconds:(*int64)(nil), Labels:map[string]string(nil), Annotations:map[string]string(nil), OwnerReferences:[]v1.OwnerReference(nil), Finalizers:[]string(nil), ClusterName:"", ManagedFields:[]v1.ManagedFieldsEntry(nil)}, InvolvedObject:v1.ObjectReference{Kind:"AppRepository", Namespace:"kubeapps", Name:"bitnami", UID:"307be21f-18b8-40ff-841f-1f82211008cf", APIVersion:"kubeapps.com/v1alpha1", ResourceVersion:"21770", FieldPath:""}, Reason:"Synced", Message:"AppRepository synced successfully", Source:v1.EventSource{Component:"apprepository-controller", Host:""}, FirstTimestamp:v1.Time{Time:time.Time{wall:0x0, ext:63752340350, loc:(*time.Location)(0x22ba300)}}, LastTimestamp:v1.Time{Time:time.Time{wall:0xc00f8125d0123590, ext:153206449513, loc:(*time.Location)(0x22ba300)}}, Count:2, Type:"Normal", EventTime:v1.MicroTime{Time:time.Time{wall:0x0, ext:0, loc:(*time.Location)(nil)}}, Series:(*v1.EventSeries)(nil), Action:"", Related:(*v1.ObjectReference)(nil), ReportingController:"", ReportingInstance:""}': 'events "bitnami.166fd3b16bc988f7" is forbidden: User "system:serviceaccount:kubeapps:kubeapps-internal-apprepository-controller" cannot patch resource "events" in API group "" in the namespace "kubeapps"' (will not retry!)
```

After fixing the RBAC error (will send a PR for this), both pods are handling updates, causing duplicity of the sync jobs:

```
▶ kubectl get jobs -n kubeapps      
NAME                                       COMPLETIONS   DURATION   AGE
apprepo-kubeapps-sync-bitnami-2j5gs        0/1                      10s
apprepo-kubeapps-sync-bitnami-5vr69        0/1                      10s
```

So our controller is not configured for multiple replicas.

absoludity:
> So our controller is not configured for multiple replicas.

Wow - that's worth an issue, it's broken if it can't be scaled. The controller pattern is designed and the go support for it is there so that you can scale the controller without issue. Given no one has ever complained it's not affecting people (as you say, no reason to scale given the low volume of app repositories), so low priority, but it's pretty bad form if we break when scaling.

andresmgot:
I agree.

> it's pretty bad form if we break when scaling

Just for the record, it doesn't break, it's only that repositories are processed twice.

PR #2598, opened 2021-03-24
Document upgrade process for chart 6.0.0

andresmgot:
### Description of the change

Upgrade guide for the upcoming major chart version.

Related to #2590 and #2595.

### Applicable issues

  - ref #2240

### DIFF
--- a/chart/kubeapps/README.md
+++ b/chart/kubeapps/README.md
@@ -324,7 +324,10 @@ It is possible that when upgrading Kubeapps an error appears. That can be caused
 
 > Note: These steps assume that you have installed Kubeapps in the namespace `kubeapps` using the name `kubeapps`. If that is not the case replace the command with your namespace and/or name.
 
+> Note: If you are upgrading from 2.3.1 see the [following section](#upgrading-to-2-3-1).
+
 > Note: If you are upgrading from 1.X to 2.X see the [following section](#upgrading-to-2-0).
+
 
 1.  (Optional) Backup your personal repositories (if you have any):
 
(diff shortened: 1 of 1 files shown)


chart/kubeapps/README.md
```diff
@@ -324,8 +324,11 @@ It is possible that when upgrading Kubeapps an error appears. That can be caused
 
 > Note: These steps assume that you have installed Kubeapps in the namespace `kubeapps` using the name `kubeapps`. If that is not the case replace the command with your namespace and/or name.
 
+> Note: If you are upgrading from 2.3.1 see the [following section](#upgrading-to-2-3-1).
+
 > Note: If you are upgrading from 1.X to 2.X see the [following section](#upgrading-to-2-0).
```
absoludity, line 329: Do you think it's safe to remove the `#upgrading-to-2-0` related content now? Hmm, guess it's only been 5 months.

chart/kubeapps/README.md
```diff
@@ -416,3 +419,40 @@ kubectl delete statefulset -n kubeapps kubeapps-postgresql-master kubeapps-postg
 ```
 
 After that you should be able to upgrade Kubeapps as always and the database will be repopulated.
+
+### Upgrading to 2.3.1
+
+Kubeapps 2.3.1 (Chart version 6.0.0) introduces some breaking changes. Helm specific functionality has been removed in order to support other installation methods (like using YAML manifests, [`kapp`](https://carvel.dev/kapp) or `kustomize`(https://kustomize.io/)). Because of that, there are some steps required before upgrading from a previous version:
+
+1. The secret with the PostgreSQL password is no longer automatically generated and persisted. If you try to upgrade Kubeapps and you installed it without setting a password, you will get the following error:
```
absoludity, line 427:
```suggestion
1. Kubeapps will no longer create a database secret for you automatically but rather will rely on the default behavior of the PostgreSQL chart. If you try to upgrade Kubeapps and you installed it without setting a password, you will get the following error:
```

chart/kubeapps/README.md
```diff
@@ -416,3 +419,40 @@ kubectl delete statefulset -n kubeapps kubeapps-postgresql-master kubeapps-postg
...
+
+> NOTE: Replace the namespace in the command with the namespace in which you have deployed Kubeapps.
+
+After that, you will be able to upgrade Kubeapps to 2.3.1 using the existing database secret:
+
+```
+helm upgrade kubeapps bitnami/kubeapps -n kubeapps --set postgresql.existingSecret=kubeapps-db
+```
+
+> NOTE: Specifing the existingSecret will only work if you are upgrading from 2.3.0 (chart 5.4.0) or before. For newer versions, follow the error instructions to set the PostgreSQL password.
```
absoludity, line 458: This is dangerous is it not? (well, as dangerous as losing a cache of data is). At what point is the `kubeapps-db` secret going to be removed by helm? I would have thought it was on the first upgrade above, which would leave the user unable to follow the error instructions. And if it's only after trying a second upgrade, it can be deleted then accidentally?

I think we should either:

1. Not provide any instruction about the password but rather leave it up to the postgres chart (ie. We keep your 1. above but don't show the error, instead just instruct the user to follow the error. Or,
2. Update your first point about the db secret to just be a command that creates a *copy* the current secret with the name `kubeapps-postgres` (so we have a secret not managed by helm) and use that for `postgresql.existingSecret=kubeapps-postgres` which will work always, I think? (haven't tested).

WDYT? (I think you know where my pref lies here)

absoludity: > Not sure if we are in the same page here. The secret `kubeapps-db` will never be deleted. If someone has ever installed Kubeapps <=v2.3.0, doing `--set postgresql.existingSecret=kubeapps-db` will always work.

You're right that I had thought from something earlier that it would be deleted since it wasn't in the helm manifest anymore, but resources created by hooks are special. That said, it's not guaranteed that it'll never be deleted - see the [helm docs](https://helm.sh/docs/topics/charts_hooks/#hook-resources-are-not-managed-with-corresponding-releases). Yes, currently helm doesn't delete resources that were created during hooks, but it may in the future and they explicitly say that "so any hook resources that must never be deleted should be annotated with `helm.sh/resource-policy: keep`."

It's not as dangerous as I'd thought. If we were already an enterprise application, I would definitely say we need to ensure people copy the secret to something not managed by helm. I personally think we should make decisions as if we were already, but realise this is harder and more effort here - I'll leave it up to you.

> 
> This note was just in case someone in the future reaches this documentation and tries the command but have never installed Kubeapps <=v2.3.0. If someone uses `--set postgresql.existingSecret` and the secret doesn't exists then it's when things get wrong (the managed secret gets deleted). I am not sure if that's what you mean? I guess we can make the command safer if we set the password instead (but I would not copy the secret again, since the `kubeapps-db` secret will persist)

See above.

>. In any case, it would be as destructive to set a wrong password, but it's easier to notice.

So with all of the above in context, perhaps:

```
> NOTE: Specifing the existingSecret will only work if you are upgrading from 2.3.0 (chart 5.4.0) or before, where Kubeapps created a `kubeapps-db` for you. For newer versions, you can create your own secret manually or follow the prompts from the PostgreSQL chart which will error without a password and give you instructions to set one.
```

absoludity (approved):

chart/kubeapps/README.md
andresmgot: Not sure if you have seen the last commit, I have modified the command, rather than using `existingSecret` (that could go wrong if upgrading from a later version) I'm using now `postgresql.postgresqlPassword`. This will work with any version and have the side-effect that we will have a copy of the secret: `kubeapps-db`, the original one and `kubeapps-postgresql`, the new one managed by helm.

I'm going to merge this (in the intermediate branch) as it is at the moment but let me know if you think we should still do something else.

absoludity: > This will work with any version and have the side-effect that we will have a copy of the secret: `kubeapps-db`, the original one and `kubeapps-postgresql`, the new one managed by helm.

Yeah, as I've said elsewhere, I don't think the new one managed by helm is helpful at all: the resource is managed by helm, but the password it contains is not: as a user *I'm* required to manage it from that point on, getting it and setting the env for every upgrade. But yes, it's the current behaviour of the postgresql chart, so let's just do that for now and try to help the team get the chart itself updated.

Issue #2376, opened 2021-02-11
CI doesn't test deployment to other clusters so it breaks.


absoludity:
### Description:

The current CI for Kubeapps does not test deploying to a cluster other than the one on which Kubeapps is installed, so we occasionally get breakage and only find out when trying to create a demo :P.

andresmgot:
I agree but I would much rather cover it with unit tests. Deploying a second cluster and doing a browser test there would be much more slower and prone to flaky environments.

absoludity:
I don't think anyone would argue the related code shouldn't be unit-tested, but this still does not address the situation of "can we deploy to another cluster", in the same way that unit tests for non-multi-cluster code don't obviate the need for CI testing an app release on a single cluster.

We can't test everything during integration, sure, but core functionality like "can I deploy an app to a different cluster from the one on which Kubeapps is installed" is something I think we should consider.

PR #2608, opened 2021-03-26
Upgrade versions used in CI

antgamdia:
### Description of the change

As part of #2376, I needed to bump up the Pupeteer version for better support of the "ignore https errors" flag. Since we had some outdated deps in general, I decided to revisit each one and perform an upgrade.

This PR (at the moment), includes two main changes:

- update the versions used in the integration package.json
- update additional versions used in the CI: helm and kubectl binaries, docker version and postgresql version

### Benefits

There are twofold: i) prepare the env for the follow-up PR; ii) get rid of outdated deps.

### Possible drawbacks

Version incompatibility issues perhaps? However, it's a change in the CI system, if something fails it will complain.

### Applicable issues

  - rel #2376


.circleci/config.yml
```diff
@@ -224,7 +224,7 @@ local_e2e_steps: &local_e2e_steps
     - <<: *exports
     # Install kind
     - run: |
-        curl -LO https://github.com/kubernetes-sigs/kind/releases/download/v0.7.0/kind-Linux-amd64
+        curl -LO https://github.com/kubernetes-sigs/kind/releases/download/v0.10.0/kind-Linux-amd64
```
andresmgot, line 227: I am not sure if we are hardcoding the cluster version elsewher, can you check? I am okay running tests with 1.20 (which is the default of kind v0.10) but we need to be aware of that.

.circleci/config.yml
```diff
@@ -276,13 +276,13 @@ jobs:
...
     docker:
-      - image: circleci/rust
+      - image: circleci/rust:1.50
     steps:
       - checkout
       - run: cargo test --manifest-path ./cmd/pinniped-proxy/Cargo.toml
   test_chart_render:
     environment:
-      HELM_VERSION: "v3.1.0"
+      HELM_VERSION: "v3.5.2"
```
andresmgot, line 285: this version is an old one in purpose. If we set this to the latest we won't know if we are relying in a version that's too recent. This should be the oldest version of Helm that we support.

integration/Makefile
```diff
@@ -2,3 +2,6 @@ default: build
 
 build:
 	docker build -t kubeapps/integration-tests:latest .
+
+push:
+	docker push kubeapps/integration-tests:latest
```
andresmgot, line 7: I think we should use some versioning here, can you use v1.0.0 for example?

andresmgot (commented):
Thanks for splitting the PRs @antgamdia!

.circleci/config.yml
antgamdia: I've checked and I'm not aware of further usages of a raw cluster version. However, I'm changing it to `kind create cluster --image kindest/node:v1.20.2` so that we know which k8s version is being used regardless of the Kind version we decide to use.

absoludity: Just for consistency, since we're using the `v1.20.2` node, why not use the same version to `kubectl` above and DRY it up with a single `K8S_VERSION=v1.20.2` ?

absoludity (approved):

absoludity:
```
unable to recognize "https://github.com/operator-framework/operator-lifecycle-manager/releases/download/v0.17.0/olm.yaml": no matches for kind "OperatorGroup" in version "operators.coreos.com/v1"
```
Odd - not sure why we've not seen this before, but it looks like we need to wait for the APIs installed via the previous `v0.17.0/crds.yaml` are available before installing `v0.17.0/olm.yaml` (0.16.1 was the first release to include the two separate files, it seems).

antgamdia:
> Just for consistency, since we're using the v1.20.2 node, why not use the same version to kubectl above and DRY it up with a single K8S_VERSION=v1.20.2 ?

Yep, it is what I tried to do in the next PR using circleCI params, see how it looks like:
https://github.com/kubeapps/kubeapps/blob/4477be3e9be6e37db68b31a921acc2ebdc2aec72/.circleci/config.yml#L2-L32

> Odd - not sure why we've not seen this before, but it looks like we need to wait for the APIs installed via the previous v0.17.0/crds.yaml are available before installing v0.17.0/olm.yaml 

Same here, I added a simple sleep, but we can try to use a k8s-native wait for a resource.
https://github.com/kubeapps/kubeapps/blob/4477be3e9be6e37db68b31a921acc2ebdc2aec72/.circleci/config.yml#L279-L282

andresmgot (approved):

PR #2618, opened 2021-03-29
Minor DRY changes in CI

antgamdia:
### Description of the change

This PR simply performs some minor refactor, extracting the hardcoded variables to CircleCI parameters. Furthermore, it names some of the steps so that the visualization in the CI dashboard is clearer at a glance.

### Benefits

Better CI visualization and DRY improvements.

### Possible drawbacks

N/A

### Applicable issues

- rel #2376

### Additional information

N/A

Issue #2394, opened 2021-02-11
Update the interface used for syncing chart data from the DB to an assetsvc API that can be used from other namespaces and clusters


danoo:
Update the interface used for syncing chart data from the DB to an assetsvc API that can be used from other namespaces and clusters
Currently the asset sync job requires access to the Kubeapps DB, meaning it can only be securely run in the Kubeapps namespace, see #1647
This same issue stops us from being able to set app repositories on other clusters (#1982)
While there we should update the DB actions to use transactions (or a framework where this comes for free) since many jobs can be posting data to the asset svc.

absoludity:
Thanks - I thought we already had this as an issue, but I see it's only implied from those linked (and described in the multi-cluster doc).

absoludity:
Bringing this into the discussion for next iteration again as it's going to soon be a blocker: If Kubeapps is installed on a specific cluster (such as a shared-services cluster in TKG), then users will only be able to use the app repositories configured with Kubeapps on install. They will be unable to add app repositories and hence unable to use private app repositories etc.

ppbaena:
To be closed as the new approach for package repositories API is in progress in #3496

Issue #1982, opened 2020-08-30
AppRepositories are only availabe on the cluster on which Kubeapps is installed


zhangchunsheng:
http://xxx/#/c/default/ns/kube-system/config/repos

I think it should be done on any cluster.

```
AppRepositories can be created on the default cluster only
kubeapps' multi-cluster support currently enables creation of custom app repositories on the cluster on which kubeapps is installed.

You cannot currently create an app repository on an additional cluster.
```

absoludity:
Yes, this isn't a design decision that we don't want to change, it's an implementation detail of the initial multi-cluster support because we need do quite a bit of work to support app repositories in other clusters.

Note that any app repositories that you make available to all namespaces on your kubeapps cluster are available on all namespaces on all clusters, you just can't create per-namespace app repositories on other clusters yet.

absoludity:
Issues which need to be handled before or while designing and implement this:

At the very least, if we blindly push forward with the current AppRepositories but on other clusters, then #1647 Move app repo sync to run in cluster/namespace will need to be designed and implemented first (as currently the sync jbo assumes DB access to the kubeapps backend). That said, it's not yet clear if we want to do this, imo. We need to re-evaluate the concept of app repositories and how we sync and present data from them.

#1358 Import charts from an OCI registry - before pushing for (helm2) AppRepositories on additional clusters, we should, IMO, decide how we're going to present catalogs moving forward... not just for OCI repos, but also for the `hub` style that helm is using, see #2078

ppbaena:
To be closed as we are not considering this feature in our roadmap. We can reconsider to re-open it if there's more demand on this feature.

PR #2619, opened 2021-03-29
Add CI multicluster + OIDC environment

### LINKED ISSUE #2376: CI doesn't test deployment to other clusters so it breaks.

antgamdia:
### Description of the change

(note: this PR supersedes #2610. It's better to split the multicluster environment and the general improvements)

This PR updates our integration tests to use the multicluster environment so that we can detect issues and errors (as proposed in #2376).

The environment is based on the one we already use for dev purposes, that is: dex/openldap + two kind clusters.
There are, also, a couple of tweaks since the integration tests run inside a container; therefore, the ingress URL is not available and there are some issues related to the cookies. 

### Benefits

Testing Kubeapps in multicluster environments with OIDC.

### Possible drawbacks

As the complexity of the CI increases, the flakiness of the overall CI systems might also increase. 

### Applicable issues

  - fixes #2376

### Additional information

This PR depends upon #2618  to be merged.

As an improvement, we can also consider including the OIDC via Pinniped. It shouldn't be a big deal since all the multicluster effort is already done in this PR.


integration/use-cases/add-multicluster-deployment.js
```diff
@@ -0,0 +1,35 @@
+const utils = require("./lib/utils");
+
+test("Deploys an application with the values by default", async () => {
+  await utils.login(
+    page,
+    process.env.USE_MULTICLUSTER_OIDC_ENV,
+    "/",
+    process.env.ADMIN_TOKEN,
```
andresmgot, line 8: I would not add this `process.env.ADMIN_TOKEN` (just to make clear this test won't work with a token)

integration/use-cases/create-private-registry.js
```diff
@@ -2,53 +2,68 @@ const axios = require("axios");
...
+    page,
+    process.env.USE_MULTICLUSTER_OIDC_ENV,
+    "/#/c/default/ns/default/config/repos",
+    process.env.ADMIN_TOKEN,
+    "kubeapps-operator@example.com",
+    "password",
   );
 
+  // wait for the loading msg to disappear
+  await page.waitForFunction(() => !document.querySelector("cds-progress-circle"));
```
andresmgot, line 21: unless there is a reason that waiting for the button doesn't work I won't add these kind of waits, it just makes the tests more difficult to maintain.

integration/use-cases/create-private-registry.js
```diff
@@ -2,53 +2,68 @@ const axios = require("axios");
...
   try {
     // TODO(andresmgot): Remove this line once 2.3 is released
     await expect(page).toClick("cds-button", { text: "Add new credentials" });
-  } catch(e) {
-    await expect(page).toClick(".btn-info-outline", { text: "Add new credentials" });
+  } catch (e) {
+    await expect(page).toClick(".btn-info-outline", {
+      text: "Add new credentials",
+    });
   }
```
andresmgot, line 50: we can remove this try-catch now

integration/use-cases/create-private-registry.js
```diff
@@ -2,53 +2,68 @@ const axios = require("axios");
...
+
   try {
     // TODO(andresmgot): Remove this line once 2.3 is released
-    await expect(page).toClick(".secondary-input cds-button", { text: "Submit" });
-  } catch(e) {
+    await expect(page).toClick(".secondary-input cds-button", {
+      text: "Submit",
+    });
+  } catch (e) {
     await expect(page).toClick(".btn-info-outline", { text: "Submit" });
```
andresmgot, line 67: same, we can remove the try-catch

integration/use-cases/missing-permissions.js
```diff
@@ -1,21 +1,33 @@
...
+    "password",
   );
 
   await expect(page).toClick("a", { text: "Catalog" });
+  // wait until load
+  await page.evaluate(() => {
+    [...document.querySelectorAll(".kubeapps-dropdown-header")].find(element =>
+      element.outerText.includes("Current Context"),
+    );
+  });
```
andresmgot, line 19: this better waits for the text "apache" to appear

integration/use-cases/missing-permissions.js
```diff
@@ -1,21 +1,33 @@
...
 
   await expect(page).toClick("cds-button", { text: "Deploy" });
 
   await expect(page).toClick("cds-button", { text: "Deploy" });
 
+  // wait for the loading msg to disappear
+  await page.waitForFunction(
+    () => !document.querySelector("#root > section > main > div > div > section > h3"),
+  );
+
```
andresmgot, line 31: this should not be needed (you may want to increase the timeout below)

integration/use-cases/operator-deployment.js
```diff
@@ -4,16 +4,18 @@ const utils = require("./lib/utils");
...
+    page,
+    process.env.USE_MULTICLUSTER_OIDC_ENV,
+    "/#/c/default/ns/kubeapps/operators",
+    process.env.ADMIN_TOKEN,
+    "kubeapps-operator@example.com",
+    "password",
   );
 
+  // wait for the loading msg to disappear
+  await page.waitForFunction(() => !document.querySelector("cds-progress-circle"));
```
andresmgot, line 17: same, wait for "prometheus" (not sure if the timeout below works).

integration/use-cases/operator-deployment.js
```diff
@@ -22,13 +24,21 @@ test("Deploys an Operator", async () => {
...
+    // Deploy the Operator
+    await expect(page).toClick("cds-button", { text: "Deploy" });
+
+    await utils.retryAndRefresh(page, 4, async () => {
+      // The CSV takes a bit to get populated
+      await expect(page).toMatch("Installed", { timeout: 10000 });
+    });
+  } else {
+    console.log("Warning: the operator has already been deployed");
+  }
```
andresmgot, line 41: not sure if you already replied, why the `isAlreadyDeployed`?

integration/use-cases/operator-deployment.js
```diff
@@ -37,24 +47,31 @@ test("Deploys an Operator", async () => {
     await expect(page).toMatch("Operators", { timeout: 10000 });
 
     // Filter out charts to search only for the prometheus operator
-    await expect(page).toClick("label", { text: "Operators" });
+    await expect(page).toClick("label", { text: "Operators", timeout: 10000 });
```
andresmgot, line 50: the only `timeout` option working is for the `toMatch` function:

https://github.com/smooth-code/jest-puppeteer/blob/master/packages/expect-puppeteer/README.md#expectinstancetomatchmatcher-options

script/e2e-test.sh
```diff
@@ -319,8 +346,10 @@ for f in *.js; do
 done
 testsToIgnore=()
 # Operators are not supported in GKE 1.14 and flaky in 1.15, skipping test
+# Also skip the multicluster scenario
 if [[ -n "${GKE_BRANCH-}" ]]; then
   testsToIgnore=("operator-deployment.js" "${testsToIgnore[@]}")
+  testsToIgnore=("add-multicluster-deployment.js" "${testsToIgnore[@]}")
```
andresmgot, line 352: you can just use a single line for this

script/e2e-test.sh
```diff
@@ -339,6 +368,10 @@ kubectl create serviceaccount kubeapps-view -n kubeapps
 kubectl create role view-secrets --verb=get,list,watch --resource=secrets
 kubectl create rolebinding kubeapps-view-secret --role view-secrets --serviceaccount kubeapps:kubeapps-view
 kubectl create clusterrolebinding kubeapps-view --clusterrole=view --serviceaccount kubeapps:kubeapps-view
+## Create view user (oidc)
```
andresmgot, line 371: and the admin user?

andresmgot (commented):
Thanks @antgamdia. This is looking good, just some minor comments.

integration/use-cases/create-private-registry.js
antgamdia: When testing in my fork I didn't build the new CI images (for saving time), and I would still need these lines. Now they can be safely removed, thanks for the reminder!

integration/use-cases/operator-deployment.js
antgamdia: Yep (sorry for dropping the former PR, but I wanted to split the changes for ease the review process)
Check it at https://github.com/kubeapps/kubeapps/pull/2610#discussion_r602470780
Since uninstalling the operator is not that easy, I left this `if` for those cases where we are executing the tests locally. I don't mind removing it if you want, though.

script/e2e-test.sh
antgamdia: It is already created with cluster-admin role with the default RBAC (`kubeapps-operator@example.com`). 
This step is just modifying the `kubeapps-user@example.com`, but we can perform this step in the same place as the operator user for cohesion if you may prefer.

integration/use-cases/operator-deployment.js
andresmgot: it should be fine (if for some reason this is bypassing the operator installation, it should fail when creating an instance).

script/e2e-test.sh
andresmgot: I see, yes, better to have them in the same place.

andresmgot (approved):
LGTM when you fix the current error.

integration/jest.sequencer.js
```diff
@@ -0,0 +1,10 @@
+const Sequencer = require("@jest/test-sequencer").default;
+
+class CustomSequencer extends Sequencer {
+  sort(tests) {
+    const copyTests = Array.from(tests);
+    return copyTests.sort((testA, testB) => (testA.path > testB.path ? 1 : -1));
```
antgamdia, line 6: Despite the tests use to be executed in alphabetical order, there is no actual guarantee of that. It was causing a failure in the multicluster test because the secret created in the secret-repo test wasn't being found.
This sequencer just ensures this alphabetical order (I've pushed the integration img with this dep as well)

script/libtest.sh
```diff
@@ -41,12 +41,21 @@ k8s_wait_for_deployment() {
     
     debug "Waiting for deployment ${deployment} to be successfully rolled out..."
     # Avoid to exit the function if the rollout fails
-    silence kubectl rollout status --namespace "$namespace" deployment "$deployment" || exit_code=$?
+    silence kubectl rollout status --namespace "$namespace" deployment "$deployment" -w --timeout=60s || exit_code=$?
```
antgamdia, line 44: These changes are not strictly necessary (the main error was I didn't pass the db password :facepalm: ), but I think they make the function work as expected, I mean, performing an actual waiting, rather than digesting the exit_code for logging purposes.

antgamdia:
> LGTM when you fix the current error.

Despite my sorrow and regardless of how I want much to press the merge button :P, I would need another +1 just to make sure the new changes are ok with you as well.

absoludity:
> > LGTM when you fix the current error.
> 
> Despite my sorrow and regardless of how I want much to press the merge button :P, I would need another +1 just to make sure the new changes are ok with you as well.

Hah! Andres can merge it if happy with the changes (I thought it might have just been a few, but I see lots of commits, so I'll let Andres continue it). Enjoy your break o/

script/e2e-test.sh
```diff
@@ -229,6 +255,11 @@ if [[ -n "${TEST_UPGRADE}" ]]; then
 
   info "Waiting for Kubeapps components to be ready..."
   k8s_wait_for_deployment kubeapps kubeapps-ci
+
+  # TODO(agamez): Remove these lines in the next version
+  kubectl delete  clusterrole -n kubapps kubeapps:controller:kubeops-ns-discovery-kubeapps kubeapps:controller:kubeops-operators-kubeapps kubeapps:kubeapps:apprepositories-read kubeapps:kubeapps:apprepositories-refresh kubeapps:kubeapps:apprepositories-write || true
+  kubectl delete  clusterrolebinding -n kubapps kubeapps:controller:kubeapps:apprepositories-read kubeapps:controller:kubeops-ns-discovery-kubeapps || true
+  kubectl delete apprepositories.kubeapps.com -n kubeapps bitnami || true
```
andresmgot: I am not sure you should need these?

The clusterrole/clusterrolebinding should be adopted by the new version and the apprepository should not exist because we are using `--set apprepository.initialRepos=null`

script/libtest.sh
```diff
@@ -41,12 +41,21 @@ k8s_wait_for_deployment() {
...
     debug "Rollout exit code: '${exit_code}'"
     if [ ${exit_code} -ne 0 ]
     then
       info "Rollout exit code: '${exit_code}'"
+      exit_code=0;
+      info "Retrying after 60s..."
       kubectl get pods --namespace "$namespace"
+      sleep 60
```
andresmgot, line 52: please don't add sleeps this long. I don't think we need to retry here.

andresmgot (commented):
Thanks @antgamdia. I have some (minor) comments

integration/jest.sequencer.js
absoludity: We don't yet support app repos (let alone secret ones) on other clusters (#1982)  because the interface to sync with the asset svc is via the DB (#2394) (left over from the read-only monocular).

antgamdia: Thanks for the clarification, I always forget about this issue. 
Perhaps we should name the tests like "01-xxxx.js", "00-xxx.js" just to make clear that they are being executed in order.

integration/use-cases/create-private-registry.js
antgamdia: No, I remove them indeed. However, the CI started failing since the selectors weren't been found, so I reverted the change.

https://app.circleci.com/pipelines/github/kubeapps/kubeapps/2614/workflows/dee98a52-3d20-47a6-9a76-a1fa16351eaf/jobs/42596/artifacts

```log
 FAIL  use-cases/operator-deployment.js (108.546 s)

 RUNS  ...
  ● Deploys an Operator

    TimeoutError: Element div.modal-dialog.modal-md > div > div.modal-body > div > div > cds-button:nth-child(2) (text: "Delete") not found

      waiting for function failed: timeout 30000ms exceeded
```

But, since the tests were failing because of other reasons, I'm gonna trigger again a new build and let's see what happens.
If it is still failing I guess it will be something related to the built image.

script/e2e-test.sh
antgamdia: Unfortunately I think you saw a former commit, 7086d6f6d578c43d5c13b50400804d40a71da1f9 should fix it.
Thanks for the heads-up anyway :)

integration/jest.sequencer.js
andresmgot: I know that private repositories are not supported in the second cluster but what I don't understand is why the order of the tests affects the result. I thought that creating a private repo in the default cluster caused an error when trying to deploy in the second cluster but that may not be it.

andresmgot (approved):
LGTM, I just don't understand very well why the tests need to be ordered.

integration/jest.sequencer.js
antgamdia: Ok, ok, I see your point now and it may be an issue as you said:

When creating a private repo first, we are creating the docker credentials. Somehow, when deploying in the second cluster, it tries to retrieve these credentials (not available, though), and fails, therefore.

image

Let me replicate the issue in my dev environment and will file an issue with the details.

However, with regards to this PR, I think (for the sake of replicability) to still define an explicit order during the test suite execution.

antgamdia: I wasn't able to reproduce this issue in my dev environment. I set up the multicluster env and: 1) tried to manually create the private repo and then a new deployment; 2) run the e2e test directly from `yarn start ...`.
In both cases, the credentials are being properly stored and the app deploys seamlessly as well.

That's odd... :S

andresmgot: You probably added a namespaced private repository. If you create it in the `kubeapps` namespace, you are likely able to reproduce the issue.

PR #2621, opened 2021-03-30
Use common environment variables in CI

antgamdia:
### Description of the change

This PRs aims at providing a DRYer approach to define common env vars in the CI environment.

### Benefits

 It, hopefully, prevents from having this kind of errors (kubectl not being downloaded due to a missing `$KUBECTL_VERSION`

```bash
circleci@3ee4bb67d105:~/project$ cat /usr/local/bin/kubectl
<?xml version='1.0' encoding='UTF-8'?><Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message><Details>No such object: kubernetes-release/release//bin/linux/amd64/kubectl</Details></Error>circleci@3ee4bb67d105:~/project$ 
```

### Possible drawbacks

Some vars may not be strictly necessary in some steps, but not a big deal at all.

### Applicable issues

  - related #2376

### Additional information

N/A

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,6 +33,9 @@ parameters:
   POSTGRESQL_VERSION:
     type: "string"
     default: "11.11.0-debian-10-r45"
+  IMAGES_TO_PUSH:
+    type: "string"
+    default: "kubeapps/apprepository-controller kubeapps/dashboard kubeapps/asset-syncer kubeapps/assetsvc kubeapps/kubeops kubeapps/pinniped-proxy"
 
 ## Build conditions
 # Build in any branch or tag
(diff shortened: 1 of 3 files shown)


andresmgot:
> hopefully

Not something that should be in a commit message :P

To verify this, you can add a dummy step to do `kubectl version` (and remove it later) but feel free to merge this as it is if you are confident with the result.

andresmgot (approved):

antgamdia:
> Not something that should be in a commit message :P
> 
> To verify this, you can add a dummy step to do `kubectl version` (and remove it later) but feel free to merge this as it is if you are confident with the result.

Hahaha, you're right :disappointed: . However, since with this approach, all the commons envars are just defined once and reused in every step. As a result of this all-or-nothing way, we would have already noticed it in early steps (for instance, when installing helm: ` https://get.helm.sh/helm-v3.1.0-linux-amd64.tar.gz` - so the envar is being properly set here)

Issue #2425, opened 2021-02-18
OCI Integration Improvements


andresmgot:
### Description:

After #2232, we noticed several points of improvement for the OCI integration. These have a lower priority, that's why this is a separated issue.

Performance improvements:
 - [x] Avoid checking tags before validating the repository checksum. Each tag requires an API call that we can omit if the repository doesn't include any modifications.
 - [x] Introduce chart versions in the database one by one. At the moment, all the versions are fetched, their files extracted and then stored in the database. This means that for a large number of repositories/tags, all of them needs to be fetched before the user is able to use the repository.
   ~~- [ ] Related to the above, since all the version files are stored in memory, this causes a big memory consumption.~~

UX improvements:
 - [x] When a user is introducing some credentials to access the registry, it'd be helpful to allow using the same credentials for pulling images with one click (without the need of introducing the credentials twice).

absoludity:
Regarding the same credential: if we
* Move the repository type radio field up so it's before the repository authorization section, and
* Switch the repository authorization to enable using (or creating) a secret explicitly (like the later image pull secret) so that the user is aware they're referencing a secret rather than just specifying credentials (we'd talked about doing this already when adding the image pulls secrets),

then we can always create a docker-registry type credential in that case (and still pull out the info we need when using oras or similar). That way, when the user gets to the imagepull secret option below they can just choose a simple "[ ] use the same secret" or similar.

andresmgot:
Taking a look to this again. Regarding the UX for the last, you are talking about something like:

![Screenshot from 2021-03-25 17-10-33](https://user-images.githubusercontent.com/4025665/112505651-4572f900-8d8d-11eb-808b-ea78f5c778a9.png)

The other option would be to add a button when selecting the username and password:

![Screenshot from 2021-03-25 17-19-01](https://user-images.githubusercontent.com/4025665/112506624-3d678900-8d8e-11eb-9c88-f760afbd191c.png)

I think the UX of the second is a bit more friendly? (Regardless of how we store the info as a secret). 

For the first option we can also remove the fields for the server and email in the form to make it a bit more simple (so it's only secret name - username - password).

WDYT?

absoludity:
The key point was that if you select an OCI repo (both your screenshots are helm repos?) for Harbor, you have to enter a username/password to access the private repo and then create a credential for the docker pull secret.

[...]

PR #2623, opened 2021-03-30
Delay tag check in OCI scenario

andresmgot:
### Description of the change

When syncing an OCI registry, each tag is checked to verify if it belongs to a Helm chart (or something else like a Docker image). This requires an API call per tag which is an important number of calls for large registries. This PR moves this tag verification to later in the process so if the registry has not changed (the checksum is the same), the tags are not checked and it returns earlier.

### Possible drawbacks

A new Docker tag will be interpreted as a change in the registry, causing a re-sync.

### Applicable issues

  - First item at #2425


absoludity (approved):
Thanks Andres. One point not about this PR, but the existing `Checksum()` functionality: it's changing state as well as calculating the checksum, so now people need to be aware of this when using it (but it's not obvious from the name). That is, if I've called `FilterTags()` already, then call `Checksum()`, it's not doing what I intend since both are modifying the same state.

Now that `Checksum()` itself just does two things - gathers and sets all tags then calculates the sha, perhaps we could avoid trampling state by:

```
// SetRepositoryTags stores the state of all tags for all listed docker repositories.
func (r *OCIRegistry) SetAllTags() {
    r.allTags = map[string]TagList{} 
    ...
    r.allTags[appName] = *tags
    ....
}

// Checksum can now be completely functional
func Checksum(map[string]TagList{}) string

// Depending when and how often the result of FilterTags is used, it can potentially
// be functional too, but if not, not overriding state of anything other than what it explicitly
// declars - the chart tags:
func (r *OCIRegistry) SetChartTags() {
    // Uses r.allTags to set r.chartTags
}

**Comment by andresmgot** [comment]
I agree that working with a global state like this is not ideal. I am a bit limited by the fact that the interface is shared with the chartmuseum (index.yaml) approach but let me take a look and check if I can remove the state.

**Comment by andresmgot** [comment]
Take a look to the last commit for a possible solution ^

File: cmd/asset-syncer/utils.go [hunk 1]
```diff
@@ -124,23 +124,31 @@ func getSha256(src []byte) (string, error) {
...
+	Checksum(interface{}) (string, error)
 	Repo() *models.RepoInternal
-	Charts() ([]models.Chart, error)
+	Charts(interface{}) ([]models.Chart, error)
 	FetchFiles(name string, cv models.ChartVersion) (map[string]string, error)
 }
 
 // HelmRepo implements the Repo interface for chartmuseum-like repositories
 type HelmRepo struct {
-	content []byte
```
absoludity, line 135: I don't think we'd need to get rid of this state - it makes sense to only fetch the index once, even if we include an `Index()` on the interface, it be memoized (only fetch if not already present).

absoludity (commented):
> I agree that working with a global state like this is not ideal. I am a bit limited by the fact that the interface is shared with the chartmuseum (index.yaml) approach but let me take a look and check if I can remove the state.

Yeah, I'd not noticed this was a shared interface, that makes it a bit trickier. I'd revert the commit and if you think it's worthwhile, we can do it separately. FWIW, I don't think we should be using `interface{}` like that to enable functions that can take either a `[]byte` or a `map[string]TagList` - that usually means we need to better think about our interface.

For example, we could just use `Checksum([]byte)` as the interface, the Helm implementation calls it with `.index` where as the oci version calls it with the marshalled bytestring (even if that requires adding a separate `GetChecksumData() []byte` to the interface, where the helm version just returns the `.index` while the oci version marshals the tag list).

As per the inline comment, having some state on an implementation is not bad, it's just when we update state in unexpected ways that it can get tricky later on for people calling the functions. I'm still +1 to land without the extra commit, but not with (let's do that separately if you think it's worthwhile).

Issue #2616, opened 2021-03-29
Update runtime images


antgamdia:
It’s been a long time since we updated the runtime images. Since we are currently updating the CI, it's a good opportunity to revisit the rest of the environments:

As far as I know, we are using:

- Node 12 for the dashboard, but the current LTS is 14.
- Golang 1.13, but the latest stable is 1.16 (indeed, only 1.{14,15,16} are supported right now).
- Rust: we're using the latest version, without sticking to a major version. I'd suggest setting it to a stable version (1.50).

These changes should be performed in:

- [x] - Kubeapps' own dockerfiles
- [x] - Bitnami images
- [x] - CI images

As a side note, should we change to bitnami/xxx build images?

PR #2626, opened 2021-03-30
Update image versions

antgamdia:
### Description of the change

This PR simply bumps the image versions up to the current LTS. This change does not have any direct impact on the final released images since they are created by Bitnami.

I'm still performing some checks before marking this PR as ready (and sending the internal one for also upgrading the Bitnami images).

### Benefits

Kubeapps will be again based on supported versions.

### Possible drawbacks

I don't foresee any error since we are using the latest versions for development. however, let's see if our CI agrees with me.

### Applicable issues

- rel #2616 

### Additional information

N/A


absoludity:
Looks like the CI run was terminated. I'll re-trigger in a bit unless I hear that you intentionally terminated it.

antgamdia:
Yep, re-run it if you want, thank you! I intentionally canceled it for saving time while testing the CI changes (circleci's free plan have no concurrency)

absoludity:
Huh, apparently I don't have write permissions (I think because it's a branch on your own GH org).

andresmgot (approved):

antgamdia:
Once we have updated the CI env, let's bump up the versions in the CI and see how it goes. If everything OK, I'll update the official bitnami images as well.

.circleci/config.yml
```diff
@@ -23,7 +23,7 @@ parameters:
     default: "v1.4.3"
   KUBECTL_VERSION:
     type: "string"
-    default: "v1.20.5"
+    default: "v1.19.9"
```
andresmgot, line 26: a downgrade here?

andresmgot (approved):

.circleci/config.yml
antgamdia: Yes, I noticed we are using 1.19 instead of 1.20 and IO wanted to make sure we are testing the proper version stack https://github.com/kubeapps/kubeapps/blob/master/chart/kubeapps/values.yaml#L335

I'm keen to update it as well, so should I also propagate this update from 1.19 to 1.20 elsewhere?

andresmgot: It doesn't really matter, I was just curious in case there was an error with v1.20.

antgamdia:
Merging; next step: update the official images in the internal pipeline.

PR #2638, opened 2021-04-05
Allow to specify a Docker secret as apprepo credentials

andresmgot:
### Description of the change

Allow to specify a secret with Docker registry credentials in the AppRepository spec. For doing so it's necessary to modify the spec and add the logic to retrieve the credentials from the Docker registry secret on its format.

### Benefits

No need to duplicate auth info.

### Applicable issues

  - ref #2425


cmd/apprepository-controller/pkg/apis/apprepository/v1alpha1/types.go
```diff
@@ -56,18 +56,13 @@ type AppRepositorySpec struct {
 
 // AppRepositoryAuth is the auth for an AppRepository resource
 type AppRepositoryAuth struct {
-	Header   *AppRepositoryAuthHeader `json:"header,omitempty"`
-	CustomCA *AppRepositoryCustomCA   `json:"customCA,omitempty"`
+	Header        *AppRepoAuthSecret `json:"header,omitempty"`
+	CustomCA      *AppRepoAuthSecret `json:"customCA,omitempty"`
+	RegistryCreds *AppRepoAuthSecret `json:"registryCreds,omitEmpty"`
```
antgamdia, line 61: This change will also imply adding the corresponding field in  the API description :sweat_smile::
https://github.com/kubeapps/kubeapps/blob/master/dashboard/public/openapi.yaml#L2104-L2111

pkg/chart/chart.go
```diff
@@ -394,6 +404,13 @@ func (c *OCIClient) InitClient(appRepo *appRepov1.AppRepository, caCertSecret *c
 			return err
 		}
 		headers.Set("Authorization", string(auth))
+	} else if registryCreds != nil && appRepo.Spec.Auth.RegistryCreds != nil {
```
antgamdia, line 407: Just to be sure: there is no case in which both `authSecret` and `registryCreds` are being passed, no? If so, we should say somewhere that the latter takes precedence over the former (and add a test with both parameters).
If not, this ok as is.

pkg/kube/http_utils.go
```diff
@@ -71,6 +74,38 @@ func GetData(key string, s *corev1.Secret) (string, error) {
 	return auth, nil
 }
 
+// GetDataFromRegistrySecret retrieves the given key from the secret as a string
+func GetDataFromRegistrySecret(key string, s *corev1.Secret) (string, error) {
+	dockerConfigJsonEncoded, ok := s.StringData[key]
```
antgamdia, line 79: Will this always be a _Docker_ registry? Shouldn't we be as repo-agnostic as possible?
Edit: I see there is a `Type:  kubernetes.io/dockerconfigjson`, so that's this function is meant for.  Are you aware of other registry credentials rather than that one?

pkg/kube/kube_handler.go
```diff
@@ -567,7 +568,7 @@ func (a *userHandler) DeleteAppRepository(repoName, repoNamespace string) error
 	return err
 }
 
-func getValidationCli(appRepoBody io.ReadCloser, requestNamespace, kubeappsNamespace string) (*v1alpha1.AppRepository, HTTPClient, error) {
+func (a *userHandler) getValidationCli(appRepoBody io.ReadCloser, requestNamespace, kubeappsNamespace string) (*v1alpha1.AppRepository, HTTPClient, error) {
```
antgamdia, line 571: Out of curiosity: why this change?
Edit: ah, for the test!

antgamdia (approved):
Great! It will simplify the UX when adding a new repo. Thank you for the changes!

pkg/kube/kube_handler.go
andresmgot: this is executed when running the apprepo validation. We need now to fetch the secret from the namespace so we need to use the user clientset (so we can retrieve the credentials).

pkg/chart/chart.go
andresmgot: I have changed this to use the same secret and treat the case in which the secret key is `.dockerconfigjson` differently.

PR #2641, opened 2021-04-05
imagePullSecrets: Fix type conversion

### LINKED ISSUE #2634: Deployment template spec with 'imagePullSecrets' fails to deploy
### Description:

When deploying a custom chart that includes `imagePullSecrets` in a Deployment template spec, I get a long stack trace from Kubeapps:

```
An error occurred: PANIC: interface conversion: interface {} is []interface {}, not []map[string]interface {} goroutine 18808 [running]: 
github.com/urfave/negroni.(*Recovery).ServeHTTP.func1(0x7efde834b300, 0xc00087e978, 0xc00062e500, 0xc000924900) 
/bitnami/blacksmith-sandox/kubeapps-kubeops-2.3.0/pkg/mod/github.com/urfave/negroni@v1.0.0/recovery.go:159 +0xcb panic(0x1aa2bc0, 0xc000a36b10) 
/usr/local/go-1.13.5/src/runtime/panic.go:679 +0x1b2 
github.com/kubeapps/kubeapps/pkg/agent.(*DockerSecretsPostRenderer).updatePodSpecWithPullSecrets(0xc00087f2e0, 0xc000a365d0) 
/bitnami/blacksmith-sandox/kubeapps-kubeops-2.3.0/src/github.com/kubeapps/kubeapps/pkg/agent/docker_secrets_postrenderer.go:155
...
```

If I remove the `imagePullSecrets` from the template spec, the error goes away (although of course then I can't pull my images!), so that is definitely the source of the problem. The spec looks like this:

```yaml
...
    spec:
      imagePullSecrets:
        - name: regcred
      containers:
...
```

The error appears to be [this line in docker_secrets_postrenderer.go](https://github.com/kubeapps/kubeapps/blob/master/pkg/agent/docker_secrets_postrenderer.go#L155), which tries to type cast the list of objects (`[]interface {}`) to a `[]map[string]interface{}`, and fails. I'm afraid Go is nowhere near the top of the list of languages I'm comfortable with, so I can't offer much more insight than that. I apologise if I'm misunderstanding something fundamental.

It's important to note that this Helm chart installs perfectly from the command line using `helm install ...`, so I believe my Deployment template spec is correct. The spec is also not templated, so this has nothing to do with parameters specified in `values.yaml` or elsewhere.

1. Create a helm chart, specifying `imagePullSecrets` for a Deployment template spec.
2. Push that chart to a helm registry
3. Add the registry to KubeApps
4. Attempt to deploy the chart.
5. Receive error

Stacktrace above

Deployment to proceed

* If `imagePullSecrets` is not specified in Deployment template spec, KubeApps deployment proceeds normally.
* It makes no difference whether `imagePullSecrets` is templated and loaded from `values.yaml` or not.
* The chart can be deployed via `helm install` on the command line just fine.

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.5.3", GitCommit:"041ce5a2c17a58be0fcd5f5e16fb3e7e95fea622", GitTreeState:"dirty", GoVersion:"go1.15.8"}
```

```
$ helm list -n kubeapps
kubeapps        kubeapps        1               2021-03-27 21:34:45.846049072 +0000 UTC deployed        kubeapps-5.5.0  2.3.0
```

```
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.5", GitCommit:"6b1d87acf3c8253c123756b9e61dac642678305f", GitTreeState:"clean", BuildDate:"2021-03-18T01:10:43Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.5", GitCommit:"6b1d87acf3c8253c123756b9e61dac642678305f", GitTreeState:"clean", BuildDate:"2021-03-18T01:02:01Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"linux/amd64"}
```

andresmgot:
### Description of the change

Fixes #2634

When converting the key `imagePullSecrets`, the object type is `map[interface{}]interface{}` rather than `map[string]interface{}` (even if the object is `- name: foo`).

I'm using the raw YAML in the tests to be able to reproduce the issue.

### DIFF
--- a/pkg/agent/docker_secrets_postrenderer.go
+++ b/pkg/agent/docker_secrets_postrenderer.go
@@ -152,11 +152,12 @@ func (r *DockerSecretsPostRenderer) updatePodSpecWithPullSecrets(podSpec map[int
 	var imagePullSecrets []map[string]interface{}
 	existingNames := map[string]bool{}
 	if existingPullSecrets, ok := podSpec["imagePullSecrets"]; ok {
-		imagePullSecrets = existingPullSecrets.([]map[string]interface{})
-		for _, s := range imagePullSecrets {
-			if name, ok := s["name"]; ok {
+		for _, s := range existingPullSecrets.([]interface{}) {
+			pullSecret := s.(map[interface{}]interface{})
+			if name, ok := pullSecret["name"]; ok {
 				if n, ok := name.(string); ok {
 					existingNames[n] = true
+					imagePullSecrets = append(imagePullSecrets, map[string]interface{}{"name": n})
 				}
 			}
 		}
--- a/pkg/agent/docker_secrets_postrenderer_test.go
+++ b/pkg/agent/docker_secrets_postrenderer_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"gopkg.in/yaml.v2"
 )
 
 func TestNewDockerSecretsPostRenderer(t *testing.T) {
@@ -281,19 +282,14 @@ other: doc
 func TestUpdatePodSpecWithPullSecrets(t *testing.T) {
 	testCases := []struct {
 		name                string
-		podSpec             map[interface{}]interface{}
+		podSpec             string
 		secrets             map[string]string
 		expectedPullSecrets interface{}
 	}{
 		{
 			name: "it does not add image pull secrets when no secret matches",
-			podSpec: map[interface{}]interface{}{
-				"containers": []interface{}{
-					map[interface{}]interface{}{
-						"image": "example.com/foobar:v1",
-					},
-				},
-			},
+			podSpec: `containers:
+- image: "example.com/foobar:v1"`,
 			secrets: map[string]string{
 				"other.com": "secret-1",
 			},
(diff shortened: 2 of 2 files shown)


antgamdia (approved):
Awesome. Just for me to understand it, it has never been a `map[interface{}]interface{}` but a `map[string]interface`, but since in our tests we are setting the type, we didn't notice this issue, right? Thanks for the change!

andresmgot:
> Awesome. Just for me to understand it, it has never been a `map[interface{}]interface{}` but a `map[string]interface`, but since in our tests we are setting the type, we didn't notice this issue, right? Thanks for the change!

Yes (but the other way around, it has never been a `map[string]interface{}`). Apparently when unmarshalling YAML, the key is parsed as a generic `interface{}` not the specific `string`.

pkg/agent/docker_secrets_postrenderer.go
```diff
@@ -152,11 +152,12 @@ func (r *DockerSecretsPostRenderer) updatePodSpecWithPullSecrets(podSpec map[int
 	var imagePullSecrets []map[string]interface{}
 	existingNames := map[string]bool{}
 	if existingPullSecrets, ok := podSpec["imagePullSecrets"]; ok {
-		imagePullSecrets = existingPullSecrets.([]map[string]interface{})
-		for _, s := range imagePullSecrets {
-			if name, ok := s["name"]; ok {
+		for _, s := range existingPullSecrets.([]interface{}) {
```
absoludity, line 155: It's not new in this PR, but shouldn't those type conversions be using the comma, ok idiom, even if we're certain that it'll always convert... if we're wrong, it will [crash with a run-time error](https://golang.org/doc/effective_go#interface_conversions). (Apologies if I wrote the initial type conversion without the comma,ok).

absoludity: Hah - and I see this was the actual issue being addressed (runtime error due to a failed interface conversion), so yes, we really should handle that with the comma-ok, IMO.

PR #2650, opened 2021-04-06
Allow to reuse docker registry credentials

andresmgot:
### Description of the change

Third item at #2425 

After the discussion about what UX would be the best / more simple, I decided to go for a radio button to specify that you are going to reuse some docker creds and change the checkbox list for a `select`:

Screenshot from 2021-04-06 18-04-29

Note that if you select the "Use Docker Registry Credentials" option, the select box becomes required. Also, this is only present for namespaces different than the global one.

### Possible drawbacks

Only one docker registry cred secret can be used. My assumption is that this is okay because one apprepository interacts with a single OCI registry. If that's not the case, we can restore the checkbox list and move the `select` box up to the authorization section (as described [here](https://github.com/kubeapps/kubeapps/issues/2425#issuecomment-809276342))).

### Applicable issues

  - ref #2425 

### Additional Information

Still need to do some testing and write some docs.


dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
```diff
@@ -267,6 +267,11 @@ export function AppRepoForm(props: IAppRepoFormProps) {
     return message;
   };
 
+  /* Only when using a namespace different than the Kubeapps namespace (Global)
+    the repository can be associated with Docker Registry Credentials since
+    the pull secret won't be available in all namespaces */
```
antgamdia, line 272: Could we display this information in a tooltip or similar? Perhaps it's just me, but I felt lost when this button was not appearing until I noticed I was in the Kubeapps namespace.

dashboard/src/components/Config/AppRepoList/AppRepoAddDockerCreds.tsx
```diff
@@ -93,22 +95,17 @@ export function AppRepoAddDockerCreds({
...
-        })}
-      </CdsCheckboxGroup>
+        <select
+          value={selectedImagePullSecret}
+          required={required}
+          onChange={e => selectPullSecret(e.target.value)}
+        >
+          <option />
+          {currentImagePullSecrets.map(secret => {
+            return <option key={`option-${secret.metadata.name}`}>{secret.metadata.name}</option>;
```
antgamdia, line 105: Great to get rid of the the manual checkboxes!

antgamdia (approved):
Great! Thanks for the improvements. Now it is a much simpler UX.
+1 once the e2e test is modified accordingly to match the new selectors.

dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
andresmgot: we can maybe show and disable it in the `kubeapps` namespace, with a tooltip (but I would do that in a different PR).

dashboard/src/components/Config/AppRepoList/AppRepoAddDockerCreds.tsx
```diff
@@ -75,8 +77,8 @@ export function AppRepoAddDockerCreds({
   /* eslint-disable jsx-a11y/label-has-associated-control */
   return (
     <>
-      <CdsCheckboxGroup>
-        <label>Associate Docker Registry Credentials (optional)</label>
+      <CdsSelect>
+        <label>Associate Docker Registry Credentials{required ? "" : " (optional)"}</label>
```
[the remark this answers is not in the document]
andresmgot: exactly

PR #2654, opened 2021-04-07
Disable (rather than hiding) the Docker Credentials section

andresmgot:
### Description of the change

Follow up of https://github.com/kubeapps/kubeapps/pull/2650#discussion_r608603878

Rather than hiding the Docker Credentials section in the AppRepo form (which can be confusing), disable it with a hint:

Screenshot from 2021-04-07 17-05-40


antgamdia (approved):
Awesome! Thank you for this change!
Perhaps, I'd rather add a full stop at the end of the sentence: _"(...) More info here."_

(it would have saved me some time when doing the CI thing :facepalm: )

PR #2661, opened 2021-04-09
Several fixes to reuse docker reg creds

andresmgot:
### Description of the change

When testing #2638 and #2650 in real live, I noticed several issues that are fixed in this PR:

 - The asset-syncer needs to be adapted for the case in which a `.dockerconfigjson` is passed
 - When validating a repository that has a `.dockerconfigjson` linked, it's necessary to specify its namespace (before, the kubeapps namespace was being used)
 - When accessing a secret `.Data`, this data is already converted from base64 (there is no need to do it again).
 - The docker credentials secret was not being properly copied to the `kubeapps` namespace for the sync job to use it.

### DIFF
--- a/cmd/asset-syncer/sync.go
+++ b/cmd/asset-syncer/sync.go
@@ -17,13 +17,16 @@ limitations under the License.
 package main
 
 import (
+	"encoding/json"
 	"os"
 	"time"
 
 	"github.com/kubeapps/common/datastore"
 	"github.com/kubeapps/kubeapps/pkg/chart/models"
+	"github.com/kubeapps/kubeapps/pkg/kube"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	"k8s.io/kubernetes/pkg/credentialprovider"
 )
 
 var syncCmd = &cobra.Command{
@@ -58,6 +61,17 @@ var syncCmd = &cobra.Command{
 		}
 
 		authorizationHeader := os.Getenv("AUTHORIZATION_HEADER")
+		// The auth header may be a dockerconfig that we need to parse
+		dockerConfig := &credentialprovider.DockerConfigJson{}
+		err = json.Unmarshal([]byte(authorizationHeader), dockerConfig)
+		if err != nil {
+			// Not an dockerconfig, skip
+		} else {
+			authorizationHeader, err = kube.GetAuthHeaderFromDockerConfig(dockerConfig)
+			if err != nil {
+				logrus.Fatal(err)
+			}
+		}
 
 		filters, err := parseFilters(filterRules)
 		if err != nil {
--- a/cmd/asset-syncer/sync.go
+++ b/cmd/asset-syncer/sync.go
@@ -62,15 +62,13 @@ var syncCmd = &cobra.Command{
 
 		authorizationHeader := os.Getenv("AUTHORIZATION_HEADER")
 		// The auth header may be a dockerconfig that we need to parse
-		dockerConfig := &credentialprovider.DockerConfigJson{}
-		err = json.Unmarshal([]byte(authorizationHeader), dockerConfig)
-		if err != nil {
-			// Not an dockerconfig, skip
-		} else {
-			authorizationHeader, err = kube.GetAuthHeaderFromDockerConfig(dockerConfig)
+		if os.Getenv("DOCKER_CONFIG_JSON") != "" {
+			dockerConfig := &credentialprovider.DockerConfigJson{}
+			err = json.Unmarshal([]byte(os.Getenv("DOCKER_CONFIG_JSON")), dockerConfig)
 			if err != nil {
 				logrus.Fatal(err)
 			}
+			authorizationHeader, err = kube.GetAuthHeaderFromDockerConfig(dockerConfig)
 		}
 
 		filters, err := parseFilters(filterRules)
--- a/dashboard/src/shared/AppRepository.ts
+++ b/dashboard/src/shared/AppRepository.ts
@@ -108,6 +108,7 @@ export class AppRepository {
 
   public static async validate(
     cluster: string,
+    namespace: string,
     repoURL: string,
     type: string,
     authHeader: string,
@@ -116,17 +117,20 @@ export class AppRepository {
     ociRepositories: string[],
     skipTLS: boolean,
   ) {
-    const { data } = await axiosWithAuth.post<any>(url.backend.apprepositories.validate(cluster), {
-      appRepository: {
-        repoURL,
-        type,
-        authHeader,
-        authRegCreds,
-        customCA,
-        ociRepositories,
-        tlsInsecureSkipVerify: skipTLS,
+    const { data } = await axiosWithAuth.post<any>(
+      url.backend.apprepositories.validate(cluster, namespace),
+      {
+        appRepository: {
+          repoURL,
+          type,
+          authHeader,
+          authRegCreds,
+          customCA,
+          ociRepositories,
+          tlsInsecureSkipVerify: skipTLS,
+        },
       },
-    });
+    );
     return data;
   }
 
--- a/pkg/kube/http_utils.go
+++ b/pkg/kube/http_utils.go
@@ -61,27 +61,7 @@ func (c *clientWithDefaultHeaders) Do(req *http.Request) (*http.Response, error)
 	return c.client.Do(req)
 }
 
-// getDataFromRegistrySecret retrieves the given key from the secret as a string
-func getDataFromRegistrySecret(key string, s *corev1.Secret) (string, error) {
-	dockerConfigJsonEncoded, ok := s.StringData[key]
-	if !ok {
-		authBytes, ok := s.Data[key]
-		if !ok {
-			return "", fmt.Errorf("secret %q did not contain key %q", s.Name, key)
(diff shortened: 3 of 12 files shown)


cmd/asset-syncer/sync.go
```diff
@@ -58,6 +61,17 @@ var syncCmd = &cobra.Command{
 		}
 
 		authorizationHeader := os.Getenv("AUTHORIZATION_HEADER")
+		// The auth header may be a dockerconfig that we need to parse
+		dockerConfig := &credentialprovider.DockerConfigJson{}
+		err = json.Unmarshal([]byte(authorizationHeader), dockerConfig)
+		if err != nil {
+			// Not an dockerconfig, skip
```
absoludity, line 68: In the (distant) future when a new docker config format is enabled, we will error here without any knowledge of why... If possible, can we avoid ignoring an error?

For example, can we reverse the check here and see if `authorizationHeader` is a valid token, if not, check if it's a valid docker config, and if not, actually return the appropriate error? Hmm, though that doesn't work if the normal auth header is just an opaque token (which I think it is, from memory).

In which case, the sync-job *could* have a separate option for the docker config (which would be better than re-using `AUTHORIZATION_HEADER` anyway, I think?). If authorization header is set, it is used. If docker config is set, it is used (or we error if it can't be parsed).

dashboard/src/shared/AppRepository.ts
```diff
@@ -116,17 +117,20 @@ export class AppRepository {
...
-        customCA,
-        ociRepositories,
-        tlsInsecureSkipVerify: skipTLS,
+    const { data } = await axiosWithAuth.post<any>(
+      url.backend.apprepositories.validate(cluster, namespace),
+      {
+        appRepository: {
+          repoURL,
+          type,
+          authHeader,
```
absoludity, line 126: I'm still catching up after being away, but isn't the credential secret already created on the cluster by the time the user submits/validates the repo? If the UX doesn't create this particular secret until after the repo is created, then OK, but otherwise, I'm not sure why we send it over the wire rather than referring to the secret? This leaks the credential to the browser when it potentially never needs to leave the cluster (eg. when using a pre-existing secret).

pkg/kube/http_utils.go
```diff
@@ -61,36 +61,35 @@ func (c *clientWithDefaultHeaders) Do(req *http.Request) (*http.Response, error)
 	return c.client.Do(req)
 }
 
+func GetAuthHeaderFromDockerConfig(dockerConfig *credentialprovider.DockerConfigJson) (string, error) {
+	// This is a simplified handler of a Docker config which only looks for the username:password
+	// of the first entry.
```
absoludity, line 66: To avoid unknown auth errors, can we return an error if there is more than one auth as well until such a time that it's somehow possible to specify which or use all like the client does)?

pkg/kube/http_utils.go
```diff
@@ -61,36 +61,35 @@ func (c *clientWithDefaultHeaders) Do(req *http.Request) (*http.Response, error)
...
+	return "", fmt.Errorf("The given config doesn't include an auth entry")
+}
+
 // getDataFromRegistrySecret retrieves the given key from the secret as a string
 func getDataFromRegistrySecret(key string, s *corev1.Secret) (string, error) {
-	dockerConfigJsonEncoded, ok := s.StringData[key]
+	dockerConfigJson, ok := s.Data[key]
 	if !ok {
-		authBytes, ok := s.Data[key]
+		authBytes, ok := s.StringData[key]
```
absoludity, line 79: So the only reason we check `StringData` here is because we used to be verifying a secret *before* the secret had been created, I assume? Otherwise `StringData` would not be a field at all. If we are now only ever verifying an existing secret (one that has been created already), we shouldn't need to use `StringData` at all - it's just a helper during secret creation.

pkg/kube/kube_handler.go
```diff
@@ -415,19 +415,24 @@ func parseRepoRequest(appRepoBody io.ReadCloser) (*appRepositoryRequest, error)
 
 func (a *userHandler) applyAppRepositorySecret(repoSecret *corev1.Secret, requestNamespace string, appRepo *v1alpha1.AppRepository) error {
 	// TODO: pass request context through from user request to clientset.
-	_, err := a.clientset.CoreV1().Secrets(requestNamespace).Create(context.TODO(), repoSecret, metav1.CreateOptions{})
-	if err != nil && k8sErrors.IsAlreadyExists(err) {
-		_, err = a.clientset.CoreV1().Secrets(requestNamespace).Update(context.TODO(), repoSecret, metav1.UpdateOptions{})
-	}
-	if err != nil {
-		return err
+	if _, ok := repoSecret.Data[".dockerconfigjson"]; !ok {
```
absoludity, line 418: This also looks equivalent to swallowing an error. Why is this now needed? As in, in what situation is this function, `applyAppRepositorySecret`, called without a `.dockerconfigjson`, where this was not the case previously, and why do we ignore it in that case?

pkg/kube/kube_handler.go
```diff
@@ -474,8 +479,15 @@ func (a *userHandler) CreateAppRepository(appRepoBody io.ReadCloser, requestName
 	}
 
 	repoSecret := secretForRequest(appRepoRequest, appRepo)
+	if len(appRepoRequest.AppRepository.AuthRegCreds) > 0 {
+		repoSecret, err = a.GetSecret(appRepoRequest.AppRepository.AuthRegCreds, requestNamespace)
```
absoludity, line 483: This line seems to be clobbering the `repoSecret` var which could already have a value? Should the above condition include `repoSecret == nil && ...` or otherwise be an if/else so that we're not clobbering the value?

It looks like what you're wanting to do is something like: if the request includes a reference to a secret, use that, otherwise use the secret from the request itself (if present). If so, perhaps this should be in `secretForRequest`?

pkg/kube/kube_handler.go
```diff
@@ -517,6 +529,13 @@ func (a *userHandler) UpdateAppRepository(appRepoBody io.ReadCloser, requestName
...
```
absoludity, line 533: Again, the existing var is being clobbered - can you re-arrange, even if it's:
```
var repoSecret ...
if len(...) > 0 {
    repoSecret, err = ...
} else {
    repoSecret = secretForRequest(...)
}
```

absoludity (commented):
Looks like a few subtle changes that aren't documented in tests - it'd be great to test them too.

cmd/asset-syncer/sync.go
andresmgot: > In which case, the sync-job _could_ have a separate option for the docker config (which would be better than re-using `AUTHORIZATION_HEADER` anyway, I think?). If authorization header is set, it is used. If docker config is set, it is used (or we error if it can't be parsed).

Yep, probably it's better to do that (but need some additional changes in the controller).

dashboard/src/shared/AppRepository.ts
andresmgot: yes, `authRegCreds` is the name of the secret not the secret content (like it's the case for `authHeader`)

pkg/kube/http_utils.go
andresmgot: okey

pkg/kube/kube_handler.go
andresmgot: this is the function that creates the secret with the auth info. For the `.dockerconfigjson` case it's not necessary because the secret always exists from before.

pkg/kube/kube_handler.go
andresmgot: > If so, perhaps this should be in secretForRequest?

Yes, I think that's better.

cmd/asset-syncer/sync.go
```diff
@@ -58,6 +61,15 @@ var syncCmd = &cobra.Command{
 		}
 
 		authorizationHeader := os.Getenv("AUTHORIZATION_HEADER")
+		// The auth header may be a dockerconfig that we need to parse
+		if os.Getenv("DOCKER_CONFIG_JSON") != "" {
```
absoludity, line 65: So if both are defined, we're using the docker config? Fine, just wanting to be sure it's intentional (the reason I'd suggested the opposite preference, to use the auth header if it's defined and only check the docker config json if auth header is not defined, was just because there's no error involved... ie. you can't have a situation where the auth header is defined and docker config is defined incorrectly so an error is raised).

dashboard/src/shared/AppRepository.ts
absoludity: But `authHeader` is an actual credential (that will be saved as a secret after validation on submission?)

pkg/kube/http_utils.go
absoludity: You checked all the call-sites for the function to be sure?

absoludity (approved):
Thanks for the improvements. It still looks like we're retro-fitting some code to deal with secrets which may or may not exist (rather than rethinking what is the best way now that secrets may exist already), but I'm not 100%, see what you think. +1

pkg/kube/kube_handler.go
andresmgot: yes, I agree, this `applyAppRepositorySecret` is confusing right now

PR #2677, opened 2021-04-13
Run a shallow sync the first time a repository is processed

andresmgot:
### Description of the change

This is a workaround for the second item at #2425. The goal is to reduce the amount of time that takes for an OCI repository to be available through the UI since the user submits it. Ideally, chart versions should be introduced one by one in the database, first starting with the latest version of each chart. Since that'd require to change the database schema (and I don't have time for that this week), I propose to do a shallow sync of the repository (only adding the latest version of each chart), store that in the database and then do a full sync (including all versions).

### Benefits

Saves a considerable amount of time for the user. In this example, a user of this repository of 55 applications, can start using the repository after 4 min (for a total sync of 10 min):

```
time="2021-04-13T11:49:37Z" level=info msg="Last checksum: "
time="2021-04-13T11:53:57Z" level=info msg="Repository synced" shallow=true
time="2021-04-13T11:59:39Z" level=info msg="Repository synced" shallow=false
time="2021-04-13T11:59:39Z" level=info msg="Stored repository update in cache" url="https://registry.pivotal.io/tac-for-tanzu-advanced/charts/"
time="2021-04-13T11:59:39Z" level=info msg="Successfully added the chart repository tac to database"
```

Note that just retrieving and validating all the tags takes ~2 min so the time to sync a repository is:
 - 2' for the tags retrieval
 - 2' for syncing the latest tag of each app
 - 6' for syncing the rest of tags

### Possible drawbacks

At the moment, the latest version is retrieved twice (once for the shallow run and another one for the full sync), we can improve this a little bit more.

### Applicable issues

  - ref #2425

### Additional information

Note that the shallow run is only performed the first time an apprepo is synced. After that, each loop interval will perform directly a full sync.

### DIFF
--- a/cmd/asset-syncer/postgresql_utils.go
+++ b/cmd/asset-syncer/postgresql_utils.go
@@ -28,7 +28,6 @@ import (
 	"github.com/kubeapps/kubeapps/pkg/chart/models"
 	"github.com/kubeapps/kubeapps/pkg/dbutils"
 	_ "github.com/lib/pq"
-	log "github.com/sirupsen/logrus"
 )
 
 var ErrMultipleRows = fmt.Errorf("more than one row returned in query result")
@@ -72,15 +71,13 @@ func (m *postgresAssetManager) Sync(repo models.Repo, charts []models.Chart) err
 	return m.removeMissingCharts(repo, charts)
 }
 
-func (m *postgresAssetManager) RepoAlreadyProcessed(repo models.Repo, repoChecksum string) bool {
+func (m *postgresAssetManager) LastChecksum(repo models.Repo) string {
 	var lastChecksum string
 	row := m.DB.QueryRow(fmt.Sprintf("SELECT checksum FROM %s WHERE name = $1 AND namespace = $2", dbutils.RepositoryTable), repo.Name, repo.Namespace)
 	if row != nil {
-		err := row.Scan(&lastChecksum)
-		log.Errorf("lastChecksum: %+v", lastChecksum)
-		return err == nil && lastChecksum == repoChecksum
+		row.Scan(&lastChecksum)
 	}
-	return false
+	return lastChecksum
 }
 
 func (m *postgresAssetManager) UpdateLastCheck(repoNamespace, repoName, checksum string, now time.Time) error {
--- a/cmd/asset-syncer/postgresql_utils.go
+++ b/cmd/asset-syncer/postgresql_utils.go
@@ -20,6 +20,7 @@ import (
 	"database/sql"
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strings"
 	"time"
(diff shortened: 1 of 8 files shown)


cmd/asset-syncer/postgresql_utils.go
```diff
@@ -72,15 +71,13 @@ func (m *postgresAssetManager) Sync(repo models.Repo, charts []models.Chart) err
...
 
-func (m *postgresAssetManager) RepoAlreadyProcessed(repo models.Repo, repoChecksum string) bool {
+func (m *postgresAssetManager) LastChecksum(repo models.Repo) string {
 	var lastChecksum string
 	row := m.DB.QueryRow(fmt.Sprintf("SELECT checksum FROM %s WHERE name = $1 AND namespace = $2", dbutils.RepositoryTable), repo.Name, repo.Namespace)
 	if row != nil {
-		err := row.Scan(&lastChecksum)
-		log.Errorf("lastChecksum: %+v", lastChecksum)
-		return err == nil && lastChecksum == repoChecksum
+		row.Scan(&lastChecksum)
```
absoludity, line 78: There's an error return value here being ignored? Looks like the previous code returned false if there was an error, so you're updating that to match by returning "" if there was an error for the select. Is there a reason we don't just do `LastChecksum(repo models.Repo) (string, error)` to return and handle errors? Or does `row.Scan` return an error if the row doesn't exist - Ah ok, if that's the case, fine as is, though a comment to just short-circuit why we're not handling the error would be useful.

cmd/asset-syncer/sync.go
```diff
@@ -92,23 +92,37 @@ var syncCmd = &cobra.Command{
...
+			// If the repo has never been processed, run first a shallow sync to give early feedback
+			// then sync all the repositories
+			shallowSyncs = []bool{true, false}
 		}
 
-		// Fetch and store chart icons
-		fImporter := fileImporter{manager, netClient}
-		fImporter.fetchFiles(charts, repoIface)
+		for _, shallow := range shallowSyncs {
+			charts, err := repoIface.Charts(shallow)
```
absoludity, line 113: For cases like this where the argument name isn't used so it's not obvious what `repoIface.Charts(shallow)` does, I'd re-work so that it reads as `repoIface.Charts(fetch_latest_only)`, for example.

cmd/asset-syncer/utils.go
```diff
@@ -217,7 +224,7 @@ func filterCharts(charts []models.Chart, filterRule *apprepov1alpha1.FilterRuleS
 }
 
 // Charts retrieve the list of charts exposed in the repo
-func (r *HelmRepo) Charts() ([]models.Chart, error) {
+func (r *HelmRepo) Charts(shallow bool) ([]models.Chart, error) {
```
absoludity, line 227: Similar here, I think `latest_charts_only` might better indicate what actually happens, here and below.

cmd/asset-syncer/utils.go
```diff
@@ -769,25 +798,29 @@ func parseRepoIndex(body []byte) (*helmrepo.IndexFile, error) {
...
 
 // Takes an entry from the index and constructs a database representation of the
 // object.
-func newChart(entry helmrepo.ChartVersions, r *models.Repo) models.Chart {
+func newChart(entry helmrepo.ChartVersions, r *models.Repo, shallow bool) models.Chart {
 	var c models.Chart
 	copier.Copy(&c, entry[0])
-	copier.Copy(&c.ChartVersions, entry)
+	if shallow {
+		copier.Copy(&c.ChartVersions, []helmrepo.ChartVersion{*entry[0]})
```
absoludity, line 820: Both the `entry[0]` above and here `*entry[0]` look like they should be checked. but it's been like this for a while without issue (that we've noticed) so... not worth your time right now I assume :)

absoludity (approved):
Thanks Andres - that's a nice improvement. Only naming suggestions really, see what you think.

cmd/asset-syncer/postgresql_utils.go
andresmgot: [edit] At the end, it's not returning a typed error (`ErrNoRows`) as I thought but a different error when the entry doesn't exists: `sql: Scan error on column index 0, name "checksum": converting NULL to string is unsupported` which cannot be cleanly handled.

cmd/asset-syncer/utils.go
andresmgot: :shrug:

PR #2680, opened 2021-04-14
Remove name auto-generation of release names

### LINKED ISSUE #2570: Enforce / customize helm release name
### Description:

Currently we use kubeapps to provide a simple way for creating jenkins instances. This currently leads to the issue, that in one namespace there are like 15 kubeapp apps where 50% have a meaningful release name (team name, product, usecase) and the rest are not (added screenshot which doesn't reflect this ratio but it fluctuates)

We also use the values.yaml in a way that all critical options and values are at the top and we leverage yaml templating to group settings in 'profiles'.

We already have values in your values.yaml you need to configure and it would make sense for us to:
Either enforce entering a release name and not having the option to autogenerate it at all
OR
to build/configure the release name by ourselfs through the helm chart / values.yaml

I just checked out the new forms feature (totally looking forward to this! :D ) and it would be great if that would work with those forms as well.

One thing which makes it also hard is that in the overview of installed apps, you can't see who created which instance but we do use custom tokens for everyone.

Create many instances of one application

A mess as one namespace contains only one application type (jenkins), and the only differentiater besides the helm release name, is the chart version.

An easy overview page which is customer focused and allows a customer to see their applications.

Kubeapps 2 ( 2.0.1 )

Screen Shot 2021-03-18 at 11 38 30

### LINKED ISSUE #2649: Inappropriate random names
Your random name generator sometimes creates inappropriate names. I really appreciate the fun but I was showing a demo to a VP in my company and the generated name was "racial-disgust" which was a bit awkward to say the least...
You can see what I am talking about in your video https://www.youtube.com/watch?v=MHylHSDJqvc&t=150s where you have "poor-wash"

antgamdia:
### Description of the change

After some users reported some inappropriate names being generated by our generator (Moniker), I performed a quick review of replacement alternatives. Even if some of them have dictionaries with appropriate words, there is still a chance of having an unlikely bad combination.

After an internal discussion, we decided to drop the autogeneration feature, as also Helm 3 did. 

### Benefits

No more unfortunate combinations of autogenerated release names.

### Possible drawbacks

Users will no longer have an autogenerated name, so they will have to come up with one.

### Applicable issues

  - fixes #2570
  - fixes #2649

### Additional information

We have the frontend validation:

image

As well as the backend one:

image

### DIFF
--- a/dashboard/src/components/DeploymentForm/DeploymentForm.test.tsx
+++ b/dashboard/src/components/DeploymentForm/DeploymentForm.test.tsx
@@ -168,13 +168,11 @@ it("triggers a deployment when submitting the form", async () => {
     handleValuesChange("foo: bar");
   });
 
-  // Set the release name
-  const releaseNameInput = wrapper.find("#releaseName");
-  releaseNameInput.getDOMNode<HTMLInputElement>().value = releaseName;
-  releaseNameInput.simulate("change");
+  wrapper.find("#releaseName").simulate("change", { target: { value: releaseName } });
 
   wrapper.update();
   expect(wrapper.find(DeploymentFormBody).prop("appValues")).toBe("foo: bar");
+  expect(wrapper.find(DeploymentForm).find("#releaseName").prop("value")).toBe(releaseName);
 
   await act(async () => {
     // Simulating "submit" causes a console.warning
--- a/dashboard/src/components/DeploymentForm/DeploymentForm.test.tsx
+++ b/dashboard/src/components/DeploymentForm/DeploymentForm.test.tsx
@@ -1,5 +1,4 @@
 import { shallow } from "enzyme";
-import * as Moniker from "moniker-native";
 import * as ReactRedux from "react-redux";
 
 import ChartHeader from "components/ChartView/ChartHeader";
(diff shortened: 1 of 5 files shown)


dashboard/src/components/DeploymentForm/DeploymentForm.test.tsx
```diff
@@ -190,6 +167,12 @@ it("triggers a deployment when submitting the form", async () => {
   act(() => {
     handleValuesChange("foo: bar");
   });
+
+  // Set the release name
+  const releaseNameInput = wrapper.find("#releaseName");
+  releaseNameInput.getDOMNode<HTMLInputElement>().value = releaseName;
+  releaseNameInput.simulate("change");
```
andresmgot, line 174: fyi, you can do, `.simulate("change", {target: {value: releaseName}})` (you may need to change `currentTarget` with `target` in the code though.

andresmgot (approved):
thanks!

dashboard/src/components/DeploymentForm/DeploymentForm.test.tsx
antgamdia: Done! I wasn't sure if there were any side effects of changing `currentTarget` with `target` so I implemented this way. But, I've checked it and everything goes fine, so the less wordy the better.
Thanks for the tip!

PR #2683, opened 2021-04-15
Ignore errors when syncing and fix e2e tests

andresmgot:
### Description of the change

Context at https://github.com/kubeapps/kubeapps/pull/2677#discussion_r612942869

So I tried to handle the error when getting the latest checksum but when running the e2e tests, a new error appeared:

```
time="2021-04-15T09:08:58Z" level=fatal msg="pq: relation \"repos\" does not exist"
```

Rather than adding another exception, I left the code as it was before (ignoring the error). In any case, that operation (checking the last checksum) is not critical for the rest of the process.

[EDIT]

Seems that the CI was misbehaving yesterday? #2680 was also merged with a green but the e2e tests require some adaptation to type down the name of the releases. Fixed that as well here.

### Benefits

Fixes master

### Additional information

I merged it because I saw a green in #2677 but it seems that I clicked in the interval before the e2e tests were triggered (weird).

### DIFF
--- a/cmd/asset-syncer/postgresql_db_test.go
+++ b/cmd/asset-syncer/postgresql_db_test.go
@@ -412,20 +412,14 @@ func TestRepoAlreadyProcessed(t *testing.T) {
 	defer cleanup()
 
 	// not processed when it doesn't exist
-	got, err := pam.LastChecksum(repo)
-	if err != nil {
-		t.Errorf("unexpected error %v", err)
-	}
+	got := pam.LastChecksum(repo)
 	if got != "" {
 		t.Errorf("got: %s, want: %s", got, "")
 	}
 
 	// not processed when repo exists but has not been processed
 	pam.EnsureRepoExists(repoNamespace, repoName)
-	got, err = pam.LastChecksum(repo)
-	if err != nil {
-		t.Errorf("unexpected error %v", err)
-	}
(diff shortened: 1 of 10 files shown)


antgamdia:
It seems the e2e tests is not a "required" check in github, as a consequence, we see the green light with 8 checks instead of 9. We both merged our PRs without noticing it. I think we have to change the CircleCI internal config a little bit.

See https://circleci.com/docs/2.0/enable-checks/

antgamdia (approved):
A drastic solution indeed :P 
Couldn't we get the error instead and log it just in case (I mean, instead of throwing the error, printing a warning in the console). In the future, we can just replace the console output with a proper error handling but, at least, this work is not being lost. 

See what you think anyway.  As long as it fixes master, I'm ok with it.

andresmgot:
> Couldn't we get the error instead and log it just in case (I mean, instead of throwing the error, printing a warning in the console). In the future, we can just replace the console output with a proper error handling but, at least, this work is not being lost.

Yes, the thing is that the error is expected here (since the entry won't exists on the first run). The only unexpected error would be something like a failure to execute the query or to connect but that would error in other place (when creating the connexion) anyway.

Let me merge this to fix master and also because it was how it was coded before, but feel free to add more advanced error handling there.

PR #2685, opened 2021-04-15
Improve wait strategy during ci login

antgamdia:
### Description of the change

As some CI checks are still failing, it seems that changing the default `.wait` strategy could benefit the reliability of the tests. 

### Benefits

This PRs should reduce the amount of  `Execution context was destroyed, most likely because of a navigation.` we receive.

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

N/A

### DIFF
--- a/integration/use-cases/lib/utils.js
+++ b/integration/use-cases/lib/utils.js
@@ -39,7 +39,7 @@ module.exports = {
   login: async (page, isOIDC, uri, token, username, password) => {
     await page.goto(getUrl(uri));
     if (isOIDC === "true") {
-      await page.waitForNavigation();
+      await page.waitForNavigation({ waitUntil: "domcontentloaded" });
       await expect(page).toClick("cds-button", {
         text: "Login via OIDC Provider",
       });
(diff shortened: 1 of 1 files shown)


absoludity (approved):
Odd that `waitForNavigation` would return before the document is loaded without this? But I'm assuming you've checked the docs (it'd be great to reference the info you have in the PR next time).

antgamdia:
> (it'd be great to reference the info you have in the PR next time).

Ooops that's true. I recall we talked about it in previous CI PRs, but it's worthwhile to keep it also here for the record.

So `wait` has these strategies: `load` (default), `domcontentloaded`, `networkidle0` and `networkidle2`.
Even if `load` is working usually ok for us, it triggers once the `load` event is fired, the one proposed in this PR waits until `DOMContentLoaded` event is fired.

> The load event is fired when the whole page has loaded, including all dependent resources such as stylesheets and images. This is in contrast to DOMContentLoaded, which is fired as soon as the page DOM has been loaded, without waiting for resources to finish loading.

However, even if it is less restrictive, we are combining this wait with another wait for a selector, so, if a transient issue while downloading a image or css happens to occur, the test won't be affected (unless the selector is not visible, of course).

If this PR doesn't fix the issue, we will have to look at the `promise.all[...]` approach also mentioned in the Puppeteer docs.

Source: 
https://github.com/puppeteer/puppeteer/blob/main/docs/api.md#pagewaitfornavigationoptions
https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event

absoludity:
Excellent, thanks - yes, that makes it much clearer.

andresmgot (commented):
thanks for the PR!

PR #2688, opened 2021-04-16
Limit secret requests when listing app repositories

### LINKED ISSUE #1686: Avoid requesting secrets in a namespace

andresmgot:
### Description of the change

We were requesting all the secrets in a namespace when listing the app repositories. This was for later on, being able to render secrets associated to a repo or to select some docker registry credentials. This PR changes that:

 - The request is now done only when rendering the new/edit repo form.
 - We are listing only docker registry secrets
 - We are getting just the secret associated to a repository (when editing it).

### Applicable issues

  - fixes #1686 

### Additional information

If you request some changes, I will eventually come back to this and address your comments :)

### DIFF
--- a/dashboard/src/components/Config/AppRepoList/AppRepoAddDockerCreds.tsx
+++ b/dashboard/src/components/Config/AppRepoList/AppRepoAddDockerCreds.tsx
@@ -1,4 +1,4 @@
-import { useState } from "react";
+import { useEffect, useState } from "react";
 
 import { CdsControlMessage } from "@cds/react/forms";
 import { CdsInput } from "@cds/react/input";
@@ -49,6 +49,12 @@ export function AppRepoAddDockerCreds({
   const handleEmailChange = (e: React.ChangeEvent<HTMLInputElement>) => setEmail(e.target.value);
   const handleServerChange = (e: React.ChangeEvent<HTMLInputElement>) => setServer(e.target.value);
   const toggleCredSubForm = () => setShowSecretSubForm(!showSecretSubForm);
+
+  useEffect(() => {
+    if (imagePullSecrets.length && !currentImagePullSecrets.length) {
+      setCurrentImagePullSecrets(imagePullSecrets);
+    }
+  }, [imagePullSecrets, currentImagePullSecrets]);
 
   const handleInstallClick = async () => {
     setCreating(true);
--- a/dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
+++ b/dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
@@ -69,6 +69,10 @@ export function AppRepoForm(props: IAppRepoFormProps) {
   const [selectedImagePullSecret, setSelectedImagePullSecret] = useState("");
   const [validated, setValidated] = useState(undefined as undefined | boolean);
 
+  useEffect(() => {
+    dispatch(actions.repos.fetchImagePullSecrets(namespace));
+  }, [dispatch, namespace]);
+
   const {
     repos: {
       imagePullSecrets,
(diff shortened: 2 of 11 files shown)


dashboard/src/components/Config/AppRepoList/AppRepoAddDockerCreds.tsx
```diff
@@ -50,6 +50,12 @@ export function AppRepoAddDockerCreds({
   const handleServerChange = (e: React.ChangeEvent<HTMLInputElement>) => setServer(e.target.value);
   const toggleCredSubForm = () => setShowSecretSubForm(!showSecretSubForm);
 
+  useEffect(() => {
+    if (imagePullSecrets.length && !currentImagePullSecrets.length) {
```
absoludity, line 54: This condition appears to only be true if there were no image pull secrets previously? What if a single secret is changed, or a secret is added (not sure if that's supported, but if not, a comment justifying the simpler condition would help), or if a secret is removed? Seems in those cases we wouldn't be setting the current pull secrets?

dashboard/src/components/Config/AppRepoList/AppRepoList.tsx
```diff
@@ -208,7 +202,7 @@ function AppRepoList() {
               <LoadingWrapper
                 className="margin-t-xxl"
                 loadingText="Fetching Application Repositories..."
-                loaded={!isFetching}
+                loaded={!isFetchingElem.repositories}
```
absoludity, line 205: Not clear to me why this change is here? Looks like we're making the condition for loaded be more specific, to only depend on the repositories?

absoludity (approved):

dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
```diff
@@ -104,30 +108,43 @@ export function AppRepoForm(props: IAppRepoFormProps) {
...
-          } else {
-            setAuthMethod(AUTH_METHOD_CUSTOM);
-            setAuthHeader(atob(secret.data.authorizationHeader));
-          }
+        if (repo.spec?.auth?.header && !secrets.includes(repo.spec.auth.header.secretKeyRef.name)) {
+          secrets.push(repo.spec.auth.header.secretKeyRef.name);
         }
-        if (secret.data[".dockerconfigjson"]) {
-          setAuthMethod(AUTH_METHOD_REGISTRY_SECRET);
+        secrets.forEach(s => dispatch(actions.repos.fetchRepoSecret(namespace, s)));
```
antgamdia, line 119: Mmm, interesting, so in the AppRepoForm, once loaded, we are fetching:

1. The whole list of `imagePullSecrets` un a given namespace
2. The `repoSecret` each time a repo is used/selected/etc

no?

dashboard/src/components/Config/AppRepoList/AppRepoAddDockerCreds.tsx
antgamdia: So perhaps we should compare both objects and merge them, instead of replacing it the `currentImagePullSecrets` is empty?

dashboard/src/components/Config/AppRepoList/AppRepoList.tsx
antgamdia: Are we using this `isFetchingElem` elsewhere? I don't see any `isFetchingElem.secrets` for example in the code, for example.
Just out of curiosity, I'm keen to have more granularity here.

antgamdia (approved):

dashboard/src/components/Config/AppRepoList/AppRepoList.tsx
absoludity: > Looks like we're making the condition for loaded be more specific, to only depend on the repositories?

dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
absoludity: On another tangent, this is a good example where an API for fetching secrets *without* the actual secret data being sent over the wire would be useful, IMO, and only fetching the actual secret data when required for a particular secret. But that's a goal for the new API design, not this PR :)

dashboard/src/components/Config/AppRepoList/AppRepoAddDockerCreds.tsx
andresmgot: no, this is just the `useEffect` of the component, the safe-guard is to avoid re-setting in the component state `currentImagePullSecrets` (and force a re-render) if it's been already set.

dashboard/src/components/Config/AppRepoList/AppRepoList.tsx
andresmgot: Yes, this is because before we were fetching secrets and repos here (now just repos). Without this change, the apprepo form modal is closed just after opening it. This is because we are requesting secrets at that moment and the parent component, this loadingWrapper, was detecting the change and rendering the loading gif instead.

andresmgot (commented):
Had some time to go back to this :)

PR #2696, opened 2021-04-20
Improve Kubeapps cluster detection adding a new optional param

### LINKED ISSUE #2692: Unable to install Kubeapps in single-cluster mode with OIDC
Follow-up of #2662

### Description:

Unable to install Kubeapps in single-cluster mode with OIDC, for instance, when installing Kubeapps with a non-native OIDC provider in managed clusters (e.g., using VMware CSP login in an AKS cluster).

_TBD: I want to test if this issue also applies when not using Pinniped. I'll update this issue in a while._

1. Get access to a managed k8s cluster (we use AKS henceforth)
2. Install Pinniped 0.7 in AKS
3. Retrieve the impersonation proxy URL and CA following our guides
4. Install Kubeapps using a similar config:
```yaml
pinnipedProxy:
  enabled: true

authProxy:
  additionalFlags:
  - --oidc-issuer-url= #your oidc provider
  - --scope=openid email groups
  clientID: <redacted>
  clientSecret: <redacted>
  cookieSecret: <redacted>
  enabled: true
  provider: oidc

clusters:
- apiServiceURL: https://<IMPERSONATION_PROXY_URL>
  certificateAuthorityData: <IMPERSONATION_PROXY_CA>
  name: impersonation-proxy
  pinnipedConfig:
    enable: true
```
3. Install kubeapps as usual
3. Log in and perform actions.

#### Not able to deploy:

image

Using URL:

http://IP/api/kubeops/v1/clusters/impersonation-proxy/namespaces/default/releases

#### Not able to rollback:
image
> This is odd, seems an UI issue?
> 1. the URL is http://IP/api/kubeops/v1/clusters/impersonation-proxy/namespaces/default/releases/nginx-test-to-aks?action=rollback&revision=-1
> 2. The it becomes http://IP/api/kubeops/v1/clusters/impersonation-proxy/namespaces/default/releases/nginx-test-to-aks?action=rollback&revision=NaN

#### Not able to list repositories

Using URL:

http://IP/api/v1/clusters/impersonation-proxy/namespaces

#### Able to upgrade
Using URL: 

http://IP/api/kubeops/v1/clusters/impersonation-proxy/namespaces/default/releases/nginx-test-to-aks

#### Able to create and list namespaces

Using these URLs:

http://IP/api/v1/clusters/impersonation-proxy/can-i
http://IP/api/v1/clusters/impersonation-proxy/namespaces

Kubeapps should use the proper endpoints instead of relying on the `inCluster` k8s API configuration.

Gut feeling: it is related to the usage of `pinnipedConfig.enable=true`

While writing the steps, it seems there is an issue with some URLs (e.g., can-i endpoint). We should address these first and later see if there is also a backend issue.

As per https://github.com/kubeapps/kubeapps/issues/2662#issuecomment-822349783:

> So this is interesting: when I demo'd Kubeapps running in a TKG (1.1) management cluster I had this exact scenario, if I've understood correctly: Kubeapps was installed on the management cluster but was configured to deploy to the workload cluster(s) only (because TKG 1.1 didn't support OIDC for the management cluster, only the workload clusters). That is, it was configured with (both) clusters with explicit api urls and certs. So I'm keen to understand what changed that broke this (though I guess it's not surprising since we hadn't added CI tests). Note that the inCluster config is (from memory) only used to initialise the variable, I thought we overwrite the endpoint based on the config.

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.5.2", GitCommit:"167aac70832d3a384f65f9745335e9fb40169dc2", GitTreeState:"dirty", GoVersion:"go1.15.7"}
```

```
NAME            NAMESPACE       REVISION        UPDATED                                         STATUS          CHART                APP VERSION
kubeapps        kubeapps        2               2021-04-16 15:28:36.759785314 +0200 CEST        deployed        kubeapps-6.0.1       DEVEL      
```

```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.4", GitCommit:"d360454c9bcd1634cf4cc52d1867af5491dc9c5f", GitTreeState:"clean", BuildDate:"2020-11-11T13:17:17Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3+vmware.1", GitCommit:"2ac9e7ea06a1230ca196931def19d2bb67b580c7", GitTreeState:"clean", BuildDate:"2020-10-30T07:25:38Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"linux/amd64"}
```

antgamdia:
### Description of the change

This PRs aims at solving the issue identified in https://github.com/kubeapps/kubeapps/issues/2692. Particularly, a new param `isKubeappsCluster` is introduced in the `values.yaml`. This optional param will allow defining the cluster in which Kubeapps is being installed in case it cannot be automatically inferred. For instance, when every cluster uses an `APIServiceURL` we are not able to know which one is this cluster.
This param is enforced in both Kubeops and the chart template. Therefore, helm will throw an error if the chart does not define properly the cluster in which kubeapps in installed.

### Benefits

Kubeapps will work in OIDC scenarios where it is required to set an APIServiceURL for each cluster (even a single one; for instance, using Pinniped 0.7 impersonation proxy in a managed cluster).

### Possible drawbacks

Can't think right now of any, but not 100% sure.

### Applicable issues

  - fixes #2692

### Additional information

I can confirm this PR solves the initially identified problem in our AKS demo environment (installing Kubepps explicitly there, not just using it as an additional cluster). Both the kubeops and the chart fixes solve every problem I mentioned in the issue,

Marking as a draft since I want to check https://github.com/kubeapps/kubeapps/issues/2692#issuecomment-822940480. Right now, I don't see how `.AsSvc()` is affecting, but I want to look closer at it.


cmd/kubeops/main.go
```diff
@@ -201,12 +201,13 @@ func parseClusterConfig(configPath, caFilesPrefix string) (kube.ClustersConfig,
...
-			} else {
-				return kube.ClustersConfig{}, nil, fmt.Errorf("only one cluster can be configured without an apiServiceURL, two defined: %q, %q", configs.KubeappsClusterName, c.Name)
-			}
+		// Select the cluster in which Kubeapps in installed. We look for either
+		// `isKubeappsCluster: true` or an empty `APIServiceURL`.
+		isKubeappsClusterCandidate := c.IsKubeappsCluster || c.APIServiceURL == ""
+		if isKubeappsClusterCandidate && configs.KubeappsClusterName == "" {
+			configs.KubeappsClusterName = c.Name
+		} else if isKubeappsClusterCandidate && configs.KubeappsClusterName != "" {
+			return kube.ClustersConfig{}, nil, fmt.Errorf("only one cluster can be configured using either 'isKubeappsCluster: true' or without an apiServiceURL to to refer to the cluster on which Kubeapps is installed, two defined: %q, %q", configs.KubeappsClusterName, c.Name)
```
absoludity, line 210: Might be slightly simpler to parse (mentally) as:

```suggestion
		if isKubeappsClusterCandidate {
			if configs.KubeappsClusterName == "" {
				configs.KubeappsClusterName = c.Name
			} else {
				return kube.ClustersConfig{}, nil, fmt.Errorf("only one cluster can be configured using either 'isKubeappsCluster: true' or without an apiServiceURL to to refer to the cluster on which Kubeapps is installed, two defined: %q, %q", configs.KubeappsClusterName, c.Name)
			}
```

but feel free to ignore.

cmd/kubeops/main.go
```diff
@@ -228,5 +229,8 @@ func parseClusterConfig(configPath, caFilesPrefix string) (kube.ClustersConfig,
 		}
 		configs.Clusters[c.Name] = c
 	}
+	if configs.KubeappsClusterName == "" {
+		return kube.ClustersConfig{}, deferFn, fmt.Errorf("unable to determine which cluster Kubeapps is installed in")
+	}
```
absoludity, line 234: Similar here: there is a valid use-case where we don't have a name in the cluster configs for the cluster on which Kubeapps is installed.

cmd/kubeops/main_test.go
```diff
@@ -123,6 +129,30 @@ func TestParseClusterConfig(t *testing.T) {
...
+		{
+			name: "errors if more than one cluster with isKubeappsCluster=true is configured",
+			configJSON: `[
+		       {"name": "cluster-1", isKubeappsCluster: true},
+		       {"name": "cluster-2", isKubeappsCluster: true }
+		]`,
+			expectedErr: true,
+		},
+		{
+			name: "errors if no KubeappsClusterName can be resolved",
```
absoludity, line 141: Similar here: I don't think this should pass :)

pkg/kube/kube_handler.go
```diff
@@ -74,6 +74,12 @@ type ClusterConfig struct {
 	// the pinniped namespace, authenticator type and authenticator name
 	// that should be used for any credential exchange.
 	PinnipedConfig PinnipedConciergeConfig `json:"pinnipedConfig,omitempty"`
+
+	// IsKubeappsCluster is an optional per-cluster configuration specifying
+	// is this cluster is the one in which Kubeapps is being installed.
```
absoludity, line 79:
```suggestion
	// that this cluster is the one in which Kubeapps is being installed.
```

pkg/kube/kube_handler.go
```diff
@@ -74,6 +74,12 @@ type ClusterConfig struct {
...
+	// Even if this info is currently being infered, if every cluster uses
+	// an APIServiceURL, there is no way to distinguish which cluster is.
```
absoludity, line 81:
```suggestion
	// Often this is inferred as the cluster without an explicit APIServiceURL, but 
	// if every cluster defines an APIServiceURL, we can no longer infer the cluster
	// on which Kubeapps is installed.
```

absoludity (commented):
Excellent, thanks Antonio. We just need to catch up to talk through what I think is a valid case where there is no name defined for the cluster on which Kubeapps is installed.

chart/kubeapps/templates/_helpers.tpl
```diff
@@ -183,15 +183,22 @@ a defined apiServiceURL.
         {{- fail "At least one cluster must be defined." }}
     {{- end }}
     {{- range .Values.clusters }}
-        {{- if eq (.apiServiceURL | toString) "<nil>" }}
-            {{- if eq $kubeappsCluster "" }}
+        {{- if eq $kubeappsCluster "" }}
+            {{- if or (eq (.isKubeappsCluster | toString) "true") ( eq (.apiServiceURL | toString) "<nil>") }}
```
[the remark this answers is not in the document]
antgamdia: Oops, indeed I didn't. I assumed it was better to use the `toString` value to cover both `isKubeappsCluster: "false"` and `isKubeappsCluster: false` user inputs. If we don't, "false" will be interpreted as true, afaik.

I'm a little concerned about it, since we don't any json schema to validate our yaml against, we cannot assure the received values are syntactically correct before starting the semantic analysis.

But, yep, for the sake of consistency with the rest of the values.yaml, I'm changing to the proposed suggestion.

chart/kubeapps/templates/_helpers.tpl
```diff
@@ -183,15 +183,22 @@ a defined apiServiceURL.
...
+            {{- end }}
+        {{- else }}
+            {{- if or (eq (.isKubeappsCluster | toString) "true") ( eq (.apiServiceURL | toString) "<nil>") }}
+                {{- fail "Only one cluster can be configured using either 'isKubeappsCluster: true' or without an apiServiceURL to to refer to the cluster on which Kubeapps is installed. Please check the provided 'clusters' configuration." }}
             {{- end }}
         {{- end }}
     {{- end }}
-    {{- $kubeappsCluster }}
+    {{- if empty $kubeappsCluster }}
+        {{- fail "Unable to determine which cluster Kubeapps is installed in. Define 'isKubeappsCluster: true' or add a cluster without an apiServiceURL" }}
```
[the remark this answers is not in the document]
antgamdia: Perhaps I'm missing something (open to chat, let's schedule a call later), but I think we still should handle this scenario there (in the chart, not in kubeops). Let's think about each possible scenario:

- Not setting `clusters` whatsoever:
  - Result: fallback to the default chat values, that is, use `- name: default` --> ergo, has a name.

```yaml
# clusters:
```

- Setting empy`clusters` :
  - Result: fails even in current master: `kubeapps/templates/kubeops-config.yaml:1:11: executing "kubeapps/templates/kubeops-config.yaml" at <len .Values.clusters>: error calling len: len of nil pointer` --> not allowed

```yaml
clusters:
```

- Single cluster (local) in `clusters` :
  - Result: name="kubeapps-cluster"

```yaml
clusters:
- name: kubeapps-cluster
```

- Single cluster (target), flag off, in `clusters` :
  - Result: fails, Unable to determine which cluster Kubeapps is installed in. Given that we are replacing the whole `clusters`, no local default cluster is described, and, consistently with setting an empty `clusters:`, it should fail.

```yaml
clusters:
- name: target-cluster
  apiServiceURL: url
  isKubeappsCluster: false
```

- Single cluster (target), flag on, in `clusters` :
  - Result:  name="target-cluster", it was the use case we aimed at solving in this PR, indeed.

```yaml
clusters:
- name: target-cluster
  apiServiceURL: url
  isKubeappsCluster: true
```
- Two clusters (target), flag on, in `clusters` :
  - Result: fails,  Only one cluster can be configured...

```yaml
clusters:
- name: target-cluster
  apiServiceURL: url
  isKubeappsCluster: on
- name: kubeapps-cluster
```

- Two clusters (target), flag off, in `clusters` :
  - Result:  name="target-cluster"
```yaml
clusters:
- name: target-cluster
  apiServiceURL: url
  isKubeappsCluster: on
- name: kubeapps-cluster
```

Am I missing any other possible situation? Why would we have to allow scenarios like this?

cmd/kubeops/main_test.go
antgamdia: Which one? `: "errors if no KubeappsClusterName can be resolved"` ?
If we only specify:
```
 {"name": "cluster-1", "apiServiceURL": "https://example.com" }
 {"name": "cluster-2", "apiServiceURL": "https://example.com" }
```
We're not able to know which cluster is the correct one, are we?

chart/kubeapps/templates/_helpers.tpl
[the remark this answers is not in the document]
absoludity: > open to chat, let's schedule a call later

Yes, let's chat your Monday on one of the calls.

> Am I missing any other possible situation? Why would we have to allow scenarios like this (D)?

Because that's exactly the configuration required to target a single cluster that is not the cluster on which Kubeapps is installed. Ie. when we installed Kubeapps on a TKG management cluster (to which the user should *not* be able to install) we use a config exactly like D, listing only the cluster(s) available to the user (ie. none of them are the cluster on which Kubeapps is running). You can see this in the [related blog post I wrote](https://liveandletlearn.net/post/kubeapps-on-tkg-management-cluster/), where there are two clusters configured, but neither of which are the cluster on which Kubeapps is installed:

```yaml
# Configure the two workload clusters
clusters:
  - name: my-oidc-cluster
    apiServiceURL: https://DNS_OF_FIRST_WORKLOAD_CLUSTER_API_SERVER:6443
    certificateAuthorityData: CA_DATA_FOR_FIRST_WORKLOAD_CLUSTER_API_SERVER
  - name: second-oidc-cluster
    apiServiceURL: https://DNS_OF_SECOND_WORKLOAD_CLUSTER_API_SERVER:6443
    certificateAuthorityData: CA_DATA_FOR_SECOND_WORKLOAD_CLUSTER_API_SERVER
``` 

This is your (D), just with a second cluster that the user can target, AFAICT.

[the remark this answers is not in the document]
absoludity: Ah, I see below that you did get to this point of understanding from a later comment. Great.

cmd/kubeops/main_test.go
absoludity: > Which one? `: "errors if no KubeappsClusterName can be resolved"` ?

Yes.

> If we only specify:
> 
> ```
>  {"name": "cluster-1", "apiServiceURL": "https://example.com" }
>  {"name": "cluster-2", "apiServiceURL": "https://example.com" }
> ```
> 
> We're not able to know which cluster is the correct one, are we?

Correct, none of the configured clusters for the user to select are the one on which Kubeapps is installed.

chart/kubeapps/templates/_helpers.tpl
```diff
@@ -183,11 +183,11 @@ a defined apiServiceURL.
         {{- fail "At least one cluster must be defined." }}
     {{- end }}
     {{- range .Values.clusters }}
-        {{- if eq (.apiServiceURL | toString) "<nil>" }}
+        {{- if or .isKubeappsCluster ( eq (.apiServiceURL | toString) "<nil>") }}
             {{- if eq $kubeappsCluster "" }}
                 {{- $kubeappsCluster = .name }}
             {{- else }}
-                {{- fail "Only one cluster can be specified without an apiServiceURL to refer to the cluster on which Kubeapps is installed." }}
+                {{- fail "Only one cluster can be configured using either 'isKubeappsCluster: true' or without an apiServiceURL to to refer to the cluster on which Kubeapps is installed. Please check the provided 'clusters' configuration." }}
```
absoludity, line 190:
```suggestion
                {{- fail "Only one cluster can be configured using either 'isKubeappsCluster: true' or without an apiServiceURL to refer to the cluster on which Kubeapps is installed. Please check the provided 'clusters' configuration." }}
```

cmd/kubeops/main.go
```diff
@@ -201,11 +201,14 @@ func parseClusterConfig(configPath, caFilesPrefix string) (kube.ClustersConfig,
...
-		if c.APIServiceURL == "" {
+		// Select the cluster in which Kubeapps in installed. We look for either
+		// `isKubeappsCluster: true` or an empty `APIServiceURL`.
+		isKubeappsClusterCandidate := c.IsKubeappsCluster || c.APIServiceURL == ""
+		if isKubeappsClusterCandidate {
 			if configs.KubeappsClusterName == "" {
 				configs.KubeappsClusterName = c.Name
 			} else {
-				return kube.ClustersConfig{}, nil, fmt.Errorf("only one cluster can be configured without an apiServiceURL, two defined: %q, %q", configs.KubeappsClusterName, c.Name)
+				return kube.ClustersConfig{}, nil, fmt.Errorf("only one cluster can be configured using either 'isKubeappsCluster: true' or without an apiServiceURL to to refer to the cluster on which Kubeapps is installed, two defined: %q, %q", configs.KubeappsClusterName, c.Name)
```
absoludity, line 211:
```suggestion
				return kube.ClustersConfig{}, nil, fmt.Errorf("only one cluster can be configured using either 'isKubeappsCluster: true' or without an apiServiceURL to refer to the cluster on which Kubeapps is installed, two defined: %q, %q", configs.KubeappsClusterName, c.Name)
```

cmd/kubeops/main_test.go
```diff
@@ -86,10 +90,36 @@ func TestParseClusterConfig(t *testing.T) {
 				PinnipedProxyURL: "http://kubeapps-internal-pinniped-proxy.kubeapps:3333",
 			},
 		},
+		{
+			name: "parses config not specifying an explicit Kubeapps cluster",
```
absoludity, line 94: Great, thanks.

absoludity (approved):

PR #2712, opened 2021-04-23
Assume inCluster config when no cluster is passed 

### LINKED ISSUE #2692: Unable to install Kubeapps in single-cluster mode with OIDC

antgamdia:
### Description of the change

As per the discussion in https://github.com/kubeapps/kubeapps/issues/2692#issuecomment-822940480, this PR allows using the inCluster configuration if no clusters have been configured.

Specifically, it will allow configurations like this one, when we prevent users from deploying apps in the cluster in which Kubeapps has been installed on.

```yaml
clusters:
- name: target-cluster
  apiServiceURL: url
  isKubeappsCluster: false
# - name: kubeapps-cluster
```

This PR now simply brings three changes (+ tests)
  - Make `GetAppRepoAndRelatedSecrets` use the asSVC if `isKubeappsCluster && (nonUIKubeappsCluster || isGlobalAppRepository) `. 
  - Make `NewClusterConfig` use the `inCluster` config if the cluster is empty.
  - Make `getSvcClientsetForCluster` use the asSVC approach if the cluster is empty.

Furthermore, as a result of the discussions, two issues have been created: https://github.com/kubeapps/kubeapps/issues/2717 and  https://github.com/kubeapps/kubeapps/issues/2719

### Benefits

Kubeapps will work in scenarios with management/workload clusters.

### Possible drawbacks

~~Currently, this approach is still not valid in case the Kubeapps cluster is a managed one using Pinniped. We still need to declare the impersonation proxy under `clusters:`~~ Not an issue, as per https://github.com/kubeapps/kubeapps/pull/2712#pullrequestreview-644192147

### Applicable issues

  - closes #2692

### Additional information

Related to PR https://github.com/kubeapps/kubeapps/pull/2696

### DIFF
--- a/dashboard/src/shared/url.ts
+++ b/dashboard/src/shared/url.ts
@@ -97,7 +97,7 @@ export const backend = {
     update: (cluster: string, namespace: string, name: string) =>
       `${backend.apprepositories.base(cluster, namespace)}/${name}`,
   },
-  canI: (cluster: string) => `api/v1/clusters${cluster ? "/" + cluster : ""}/can-i`,
+  canI: (cluster: string) => `api/v1/clusters/${cluster}/can-i`,
 };
 
 export const kubeops = {
--- a/dashboard/src/shared/url.ts
+++ b/dashboard/src/shared/url.ts
@@ -97,7 +97,7 @@ export const backend = {
     update: (cluster: string, namespace: string, name: string) =>
       `${backend.apprepositories.base(cluster, namespace)}/${name}`,
   },
-  canI: (cluster: string) => `api/v1/clusters/${cluster}/can-i`,
+  canI: (cluster: string) => `api/v1/clusters${cluster ? "/" + cluster : ""}/can-i`,
 };
 
 export const kubeops = {
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -268,8 +268,9 @@ func ParseDetails(data []byte) (*Details, error) {
 // Depending on the repo namespace and the
 func GetAppRepoAndRelatedSecrets(appRepoName, appRepoNamespace string, handler kube.AuthHandler, userAuthToken, cluster, kubeappsNamespace string) (*appRepov1.AppRepository, *corev1.Secret, *corev1.Secret, error) {
 	client, err := handler.AsUser(userAuthToken, cluster)
-	if kubeappsNamespace == appRepoNamespace {
-		// If we're parsing a global repository (from the kubeappsNamespace), use a service client.
+	if cluster == "" || kubeappsNamespace == appRepoNamespace {
+		// If we're parsing a global repository (from the kubeappsNamespace),
+		// or we don't specify any cluster, use a service client.
 		// AppRepositories are only allowed in the default cluster for the moment
 		client, err = handler.AsSVC(cluster)
 	}
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -271,7 +271,7 @@ func GetAppRepoAndRelatedSecrets(appRepoName, appRepoNamespace string, handler k
 	// If the UI clusters configuration did not include the cluster on which Kubeapps is installed then
 	// we won't know the kubeappsNamespace but it will be empty. In this scenario Kubeapps only supports
 	// global app repositories (#1982 ).
-	nonUIKubeappsCluster = (cluster == "" && kubeappsNamespace == "")
+	nonUIKubeappsCluster := (cluster == "" && kubeappsNamespace == "")
 	if nonUIKubeappsCluster || kubeappsNamespace == appRepoNamespace {
 		// If we're parsing a global repository then use a service client.
 		// AppRepositories are only allowed in the default cluster for the moment
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -268,9 +268,12 @@ func ParseDetails(data []byte) (*Details, error) {
 // Depending on the repo namespace and the
 func GetAppRepoAndRelatedSecrets(appRepoName, appRepoNamespace string, handler kube.AuthHandler, userAuthToken, cluster, kubeappsNamespace string) (*appRepov1.AppRepository, *corev1.Secret, *corev1.Secret, error) {
 	client, err := handler.AsUser(userAuthToken, cluster)
-	if cluster == "" || kubeappsNamespace == appRepoNamespace {
-		// If we're parsing a global repository (from the kubeappsNamespace),
-		// or we don't specify any cluster, use a service client.
+	// If the UI clusters configuration did not include the cluster on which Kubeapps is installed then
+	// we won't know the kubeappsNamespace but it will be empty. In this scenario Kubeapps only supports
+	// global app repositories (#1982 ).
+	nonUIKubeappsCluster = (cluster == "" && kubeappsNamespace == "")
+	if nonUIKubeappsCluster || kubeappsNamespace == appRepoNamespace {
+		// If we're parsing a global repository then use a service client.
 		// AppRepositories are only allowed in the default cluster for the moment
 		client, err = handler.AsSVC(cluster)
 	}
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -266,13 +266,15 @@ func ParseDetails(data []byte) (*Details, error) {
 
 // GetAppRepoAndRelatedSecrets retrieves the given repo from its namespace
 // Depending on the repo namespace and the
-func GetAppRepoAndRelatedSecrets(appRepoName, appRepoNamespace string, handler kube.AuthHandler, userAuthToken, cluster, kubeappsNamespace string) (*appRepov1.AppRepository, *corev1.Secret, *corev1.Secret, error) {
+func GetAppRepoAndRelatedSecrets(appRepoName, appRepoNamespace string, handler kube.AuthHandler, userAuthToken, cluster, kubeappsNamespace string, kubeappsCluster string) (*appRepov1.AppRepository, *corev1.Secret, *corev1.Secret, error) {
 	client, err := handler.AsUser(userAuthToken, cluster)
 	// If the UI clusters configuration did not include the cluster on which Kubeapps is installed then
 	// we won't know the kubeappsNamespace but it will be empty. In this scenario Kubeapps only supports
-	// global app repositories (#1982 ).
-	nonUIKubeappsCluster := (cluster == "" && kubeappsNamespace == "")
-	if nonUIKubeappsCluster || kubeappsNamespace == appRepoNamespace {
+	// global app repositories (#1982).
+	isKubeappsCluster := cluster == kubeappsCluster
+	nonUIKubeappsCluster := cluster == ""
+	isGlobalAppRepository := kubeappsNamespace == appRepoNamespace
+	if isKubeappsCluster && (nonUIKubeappsCluster || isGlobalAppRepository) {
 		// If we're parsing a global repository then use a service client.
 		// AppRepositories are only allowed in the default cluster for the moment
 		client, err = handler.AsSVC(cluster)
--- a/pkg/chart/chart_test.go
+++ b/pkg/chart/chart_test.go
@@ -226,7 +226,7 @@ func TestParseDetailsForHTTPClient(t *testing.T) {
 		kubeappsCluster   string
 	}{
 		{
-			name: "default cert pool without auth (as svc)",
+			name: "default cert pool without auth (as svc because cluster and kubeappsCluster are empty)",
 			details: &Details{
 				AppRepositoryResourceName:      appRepoName,
 				AppRepositoryResourceNamespace: appRepoNamespace,
@@ -237,7 +237,7 @@ func TestParseDetailsForHTTPClient(t *testing.T) {
 			kubeappsCluster:   "",
 		},
 		{
-			name: "default cert pool without auth (as user)",
+			name: "default cert pool without auth (as user because cluster is something other than kubeapps cluster)",
 			details: &Details{
 				AppRepositoryResourceName:      appRepoName,
 				AppRepositoryResourceNamespace: appRepoNamespace,
--- a/pkg/chart/chart_test.go
+++ b/pkg/chart/chart_test.go
@@ -216,19 +216,47 @@ func TestParseDetailsForHTTPClient(t *testing.T) {
 	)
 
 	testCases := []struct {
-		name             string
-		details          *Details
-		appRepoSpec      appRepov1.AppRepositorySpec
-		errorExpected    bool
-		numCertsExpected int
+		name              string
+		details           *Details
+		appRepoSpec       appRepov1.AppRepositorySpec
+		errorExpected     bool
+		numCertsExpected  int
+		cluster           string
+		kubeappsNamespace string
+		kubeappsCluster   string
 	}{
+		{
+			name: "default cert pool without auth (as svc)",
+			details: &Details{
+				AppRepositoryResourceName:      appRepoName,
+				AppRepositoryResourceNamespace: appRepoNamespace,
+			},
+			numCertsExpected:  len(systemCertPool.Subjects()),
+			cluster:           "",
+			kubeappsNamespace: appRepoNamespace,
+			kubeappsCluster:   "",
+		},
+		{
+			name: "default cert pool without auth (as user)",
+			details: &Details{
+				AppRepositoryResourceName:      appRepoName,
+				AppRepositoryResourceNamespace: appRepoNamespace,
+			},
+			numCertsExpected:  len(systemCertPool.Subjects()),
+			cluster:           "target-cluster",
+			kubeappsNamespace: appRepoNamespace,
+			kubeappsCluster:   "",
+		},
 		{
 			name: "default cert pool without auth",
 			details: &Details{
 				AppRepositoryResourceName:      appRepoName,
 				AppRepositoryResourceNamespace: appRepoNamespace,
 			},
-			numCertsExpected: len(systemCertPool.Subjects()),
+			numCertsExpected:  len(systemCertPool.Subjects()),
+			cluster:           "",
+			kubeappsNamespace: "",
+			kubeappsCluster:   "",
 		},
 		{
 			name: "custom CA added when passed an AppRepository CRD",
@@ -246,7 +274,10 @@ func TestParseDetailsForHTTPClient(t *testing.T) {
 					},
 				},
 			},
-			numCertsExpected: len(systemCertPool.Subjects()) + 1,
+			numCertsExpected:  len(systemCertPool.Subjects()) + 1,
+			cluster:           "",
+			kubeappsNamespace: "",
+			kubeappsCluster:   "",
 		},
 		{
 			name: "errors if secret for custom CA secret cannot be found",
@@ -264,7 +295,10 @@ func TestParseDetailsForHTTPClient(t *testing.T) {
 					},
 				},
 			},
-			errorExpected: true,
+			errorExpected:     true,
+			cluster:           "",
+			kubeappsNamespace: "",
+			kubeappsCluster:   "",
 		},
 		{
 			name: "authorization header added when passed an AppRepository CRD",
@@ -282,7 +316,10 @@ func TestParseDetailsForHTTPClient(t *testing.T) {
 					},
 				},
 			},
-			numCertsExpected: len(systemCertPool.Subjects()),
+			numCertsExpected:  len(systemCertPool.Subjects()),
+			cluster:           "",
+			kubeappsNamespace: "",
+			kubeappsCluster:   "",
 		},
 		{
 			name: "errors if auth secret cannot be found",
@@ -300,7 +337,10 @@ func TestParseDetailsForHTTPClient(t *testing.T) {
 					},
 				},
 			},
-			errorExpected: true,
+			errorExpected:     true,
+			cluster:           "",
+			kubeappsNamespace: "",
+			kubeappsCluster:   "",
 		},
 		{
 			name: "authorization header added when passed an AppRepository CRD",
@@ -318,7 +358,10 @@ func TestParseDetailsForHTTPClient(t *testing.T) {
 					},
 				},
 			},
-			numCertsExpected: len(systemCertPool.Subjects()),
+			numCertsExpected:  len(systemCertPool.Subjects()),
+			cluster:           "",
+			kubeappsNamespace: "",
+			kubeappsCluster:   "",
 		},
 	}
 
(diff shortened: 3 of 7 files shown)


pkg/chart/chart.go
```diff
@@ -268,8 +268,9 @@ func ParseDetails(data []byte) (*Details, error) {
 // Depending on the repo namespace and the
 func GetAppRepoAndRelatedSecrets(appRepoName, appRepoNamespace string, handler kube.AuthHandler, userAuthToken, cluster, kubeappsNamespace string) (*appRepov1.AppRepository, *corev1.Secret, *corev1.Secret, error) {
 	client, err := handler.AsUser(userAuthToken, cluster)
-	if kubeappsNamespace == appRepoNamespace {
-		// If we're parsing a global repository (from the kubeappsNamespace), use a service client.
+	if cluster == "" || kubeappsNamespace == appRepoNamespace {
+		// If we're parsing a global repository (from the kubeappsNamespace),
+		// or we don't specify any cluster, use a service client.
```
absoludity, line 273: I think the key point here is that we use a service account if we're fetching a global AppRepository, which is currently one in the Kubeapps namespace. Given that we currently only support app repositories on the cluster on which Kubeapps is installed, is going to be either:
1. The app repository explicitly matches because `cluster` is the kubeapps cluster (unfortunately we don't currently have access to the kubeapps cluster here[1]) and the app repo namespace is the kubeapps namespace, or 
2. The cluster and kubeappsNamespace are both empty because the UI `clusters` config doesn't include the cluster on which kubeapps is installed in which case we must always be fetching global app repos (nothing else is supported).

With this in mind, I'd use variable names to document this a little more clearly with (and to be clear, this is fixing the bug introduced, not a good longer-term solution):

```suggestion
	// If the UI clusters configuration did not include the cluster on which Kubeapps is installed then
	// we won't know the kubeappsNamespace but it will be empty. In this scenario Kubeapps only supports
	// global app repositories (#1982 ).
	nonUIKubeappsCluster = (cluster == "" && kubeappsNamespace == "")
	if nonUIKubeappsCluster || kubeappsNamespace == appRepoNamespace {
		// If we're parsing a global repository then use a service client.
```

[1] We should fix this as there are security implications, in that if I create a namespace on a different cluster with the same name as that in which Kubeapps is installed, we'd currently use the service account. We could possibly just pass `kubeappsCluster` to this function, if it's available at the call-site, not sure.

pkg/kube/kube_handler.go
```diff
@@ -103,6 +103,11 @@ func NewClusterConfig(inClusterConfig *rest.Config, userToken string, cluster st
 	config.BearerToken = userToken
 	config.BearerTokenFile = ""
 
+	// If the cluster is not provided, let's assume inClusterConfig
```
absoludity, line 106:
```suggestion
	// If the cluster is empty, we assume the rest of the inClusterConfig is correct. This can be the case when
	// the cluster on which Kubeapps is installed is not one presented in the UI as a target (hence not in the
	// `clusters` configuration).
```

pkg/kube/kube_handler_test.go
```diff
@@ -1187,6 +1187,30 @@ func TestNewClusterConfig(t *testing.T) {
...
+			},
+			inClusterConfig: &rest.Config{
+				BearerToken:     "something-else",
+				BearerTokenFile: "/foo/bar",
+			},
+			expectedConfig: &rest.Config{
+				BearerToken:     "token-1",
+				BearerTokenFile: "",
+			},
+		},
```
absoludity, line 1213: Great, thanks!

absoludity (approved):
> Currently, this approach is still not valid in case the Kubeapps cluster is a managed one using Pinniped. We still need to declare the impersonation proxy under `clusters:`

Given that in this case (where the Kubeapps cluster is not presented to the user in the UI at all) we do not (should not) have any interactions between the user (and their credential) and the cluster on which Kubeapps is installed, I don't understand why this should be relevant? That is, we shouldn't need to declare the impersonation proxy because it should never be used by Kubeapps for that cluster since the only interaction Kubeapps has with that cluster will be via the internal service account, which shouldn't use pinniped at all?

antgamdia:
> That is, we shouldn't need to declare the impersonation proxy because it should never be used by Kubeapps for that cluster since the only interaction Kubeapps has with that cluster will be via the internal service account, which shouldn't use pinniped at all?

Absolutely right, I needed some fresh air and think again of it. If we don't want to target this cluster, there is no need to explicitly use any impersonation proxy as we were using the service account. Thanks for pointing this out!


> [1] We should fix this as there are security implications, in that if I create a namespace on a different cluster with the same name as that in which Kubeapps is installed, we'd currently use the service account. We could possibly just pass kubeappsCluster to this function, if it's available at the call-site, not sure.

Good catch! I'm adding the kubeapps cluster to the check, so that it becomes:
`IF isKubeappsCluster && (nonUIKubeappsCluster || isGlobalAppRepository) THEN use AsSVC `


I've been testing the suggestion, but I don't see why the `kubeappsNamespace==""` in `nonUIKubeappsCluster` stands for.
For example, testing it out, I realized that `kubeappsNamespace` comes from the config (`kubeappsNamespace := os.Getenv("POD_NAMESPACE")`), so it won't be `""` (otherwise, we throw a fatal):

``` 
	isKubeappsCluster := cluster == kubeappsCluster
	nonUIKubeappsCluster := (cluster == "" && kubeappsNamespace == "")
	isGlobalAppRepository := kubeappsNamespace == appRepoNamespace
---
isKubeappsCluster: true
        cluster: 
        kubeappsCluster: 
nonUIKubeappsCluster: false
        cluster: 
        kubeappsNamespace: kubeapps
isGlobalAppRepository: false
        kubeappsNamespace: kubeapps
        kubeappsNamespace: kubeapps
uses AsSVC: false
``` 

So let's remove this `kubeappsNamespace=""` condition, no?

antgamdia:
(taking an issue apart from the previous comment)

Mmm, I just recalled the issue I mentioned in our AKS scenario:

a) Declaring "aks" (kubeapps cluster) and "tmc" clusters, works fine.
b) Declaring just "tmc" cluster, fails (see below)
c) [local cluster, no AKS] Declaring just "tmc" cluster, works

The problem seems to be a permissions thing:
image

_Side note: while investigating, I discovered the can-i endpoint was failing when cluster="". I've updated the PR with this change, but happy to extract it to a separate one._

It seems the service-account `kubeapps-internal-kubeops` with the default rbac created by kubeapps is not able to retrieve the `appreposiories`. Indeed, performing:

```
kubectl auth can-i get  apprepositories.kubeapps.com --as=system:serviceaccount:kubeapps:kubeapps-internal-kubeops
> no
```

I've (unsuccessfully) tried to fix the root cause for this error. When it works, I guess it is -not- using the inCluster config, but the impersonation proxy way.
However, once I add a ClusterRole+ClusterRoleBinding it works. I was trying to deploy in both `default` and a custom namespace.

I mean, the main purpose of this PR is already achieved... but I'm not happy with the overall experience in the AKS scenario.

Finally, a question just to confirm: in this case (omitting the kubeapps cluster), users cannot modify their repositories. That is, repos can only be changes via yaml. Am I right?

It'd be great if you have a chance to have a look at the cluster with the changes in this PR just to confirm this is the right way to go.

pkg/chart/chart_test.go
```diff
@@ -216,19 +216,47 @@ func TestParseDetailsForHTTPClient(t *testing.T) {
...
+		details           *Details
+		appRepoSpec       appRepov1.AppRepositorySpec
+		errorExpected     bool
+		numCertsExpected  int
+		cluster           string
+		kubeappsNamespace string
+		kubeappsCluster   string
 	}{
+		{
+			name: "default cert pool without auth (as svc)",
```
absoludity, line 229: I think our heads are in this now, so we understand, but in 3 months (or to someone else) more context will be useful?
```suggestion
			name: "default cert pool without auth (as svc because cluster and kubeappsCluster are empty)",
```

pkg/chart/chart_test.go
```diff
@@ -216,19 +216,47 @@ func TestParseDetailsForHTTPClient(t *testing.T) {
...
+				AppRepositoryResourceName:      appRepoName,
+				AppRepositoryResourceNamespace: appRepoNamespace,
+			},
+			numCertsExpected:  len(systemCertPool.Subjects()),
+			cluster:           "",
+			kubeappsNamespace: appRepoNamespace,
+			kubeappsCluster:   "",
+		},
+		{
+			name: "default cert pool without auth (as user)",
```
absoludity, line 240:
```suggestion
			name: "default cert pool without auth (as user because cluster is something other than kubeapps cluster)",
```

pkg/http-handler/http-handler.go
```diff
@@ -302,6 +302,12 @@ func CanI(kubeHandler kube.AuthHandler) func(w http.ResponseWriter, req *http.Re
 		_, requestCluster := getNamespaceAndCluster(req)
 
 		clientset, err := kubeHandler.AsUser(token, requestCluster)
+
+		// If the UI clusters configuration did not include the cluster on which Kubeapps is installed, two
+		// the requestCluster will be empty. We will try using the service account instead.
+		if requestCluster == "" {
+			clientset, err = kubeHandler.AsSVC(requestCluster)
+		}
```
absoludity, line 310: This sounds absolutely dangerous? Given the purpose of the `CanI` handler is to determine whether the *user*  has permission to do something, we should never be switching to the service account, I don't think?

dashboard/src/shared/url.ts
```diff
@@ -97,7 +97,7 @@ export const backend = {
     update: (cluster: string, namespace: string, name: string) =>
       `${backend.apprepositories.base(cluster, namespace)}/${name}`,
   },
-  canI: (cluster: string) => `api/v1/clusters/${cluster}/can-i`,
+  canI: (cluster: string) => `api/v1/clusters${cluster ? "/" + cluster : ""}/can-i`,
```
absoludity, line 100: Hmm, depends what you're doing in the case where `cluster` isn't defined, but sounds like it could be dangerous (EDIT: see below). I'm guessing this is because some code in the frontend is checking whether users can do things with app repositories which are on the kubeapps cluster?

absoludity:
> Absolutely right, I needed some fresh air and think again of it. If we don't want to target this cluster, there is no need to explicitly use any impersonation proxy as we were using the service account. Thanks for pointing this out!

No problem. We may later need to include it (separate from the UI clusters config) for a case you've mentioned (users handling app repositories when not targeting the Kubeapps cluster), but not relevant for now.

> 
> > [1] We should fix this as there are security implications, in that if I create a namespace on a different cluster with the same name as that in which Kubeapps is installed, we'd currently use the service account. We could possibly just pass kubeappsCluster to this function, if it's available at the call-site, not sure.
> 
> Good catch! I'm adding the kubeapps cluster to the check, so that it becomes:
> `IF isKubeappsCluster && (nonUIKubeappsCluster || isGlobalAppRepository) THEN use AsSVC `

Excellent, thanks for that (it's something that's worried me in the past).

> 
> I've been testing the suggestion, but I don't see why the `kubeappsNamespace==""` in `nonUIKubeappsCluster` stands for.
> For example, testing it out, I realized that `kubeappsNamespace` comes from the config (`kubeappsNamespace := os.Getenv("POD_NAMESPACE")`), so it won't be `""` (otherwise, we throw a fatal):

Yes, sorry - I'd not checked and was assuming `kubeappsNamespace` was being set from the config for some reason. Thanks.

> So let's remove this `kubeappsNamespace=""` condition, no?

Yep, you're absolutely right :P

absoludity:
> (taking an issue apart from the previous comment)
> 
> Mmm, I just recalled the issue I mentioned in our AKS scenario:
> 
> a) Declaring "aks" (kubeapps cluster) and "tmc" clusters, works fine.
> b) Declaring just "tmc" cluster, fails (see below)
> c) [local cluster, no AKS] Declaring just "tmc" cluster, works

I don't understand the difference between b and c here? Can you elaborate?
 
> The problem seems to be a permissions thing:
> image
> 
> _Side note: while investigating, I discovered the can-i endpoint was failing when cluster="". I've updated the PR with this change, but happy to extract it to a separate one._

Yeah, I don't think we can include that change and need to think more about that. Let's pull it back out and figure out in what cases the `can-i` endpoint is being called without a target cluster. My guess is for checking whether the user has access to app repos, but we can't switch to the svc account client for that function ever, I don't think.

> 
> It seems the service-account `kubeapps-internal-kubeops` with the default rbac created by kubeapps is not able to retrieve the `appreposiories`. Indeed, performing:
> 
> ```
> kubectl auth can-i get  apprepositories.kubeapps.com --as=system:serviceaccount:kubeapps:kubeapps-internal-kubeops
> > no
> ```

That is unexpected, I'll take a look.

> 
> I've (unsuccessfully) tried to fix the root cause for this error. When it works, I guess it is -not- using the inCluster config, but the impersonation proxy way.
> However, once I add a ClusterRole+ClusterRoleBinding it works. I was trying to deploy in both `default` and a custom namespace.
> 
> I mean, the main purpose of this PR is already achieved... but I'm not happy with the overall experience in the AKS scenario.

Yes, I'd be happy to land it without the `can-i` change, and then look at the specific remaining issues separately.

> 
> Finally, a question just to confirm: in this case (omitting the kubeapps cluster), users cannot modify their repositories. That is, repos can only be changes via yaml. Am I right?

That is correct, currently. "global" repositories are defined in the kubeapps namespace and cluster, and users can't interact with that cluster currently. This is something we could fix (and might give us a chance to find a better solution to the code-smell with multi-cluster configuration).

> It'd be great if you have a chance to have a look at the cluster with the changes in this PR just to confirm this is the right way to go.

I'll try to look today. Thanks!

absoludity (changes requested):
Just making sure we don't accidentally land this with the service account being used in `can-i`.

absoludity:
> > Mmm, I just recalled the issue I mentioned in our AKS scenario:
> > a) Declaring "aks" (kubeapps cluster) and "tmc" clusters, works fine.
> > b) Declaring just "tmc" cluster, fails (see below)
> > c) [local cluster, no AKS] Declaring just "tmc" cluster, works
> 
> I don't understand the difference between b and c here? Can you elaborate?

Just looking at this now, so I think I understand the difference: you mean that (c) is a similar setup but with kubeapps installed  in a local kind cluster but configured with just the "tmc" cluster as the target for installations?

> > It seems the service-account `kubeapps-internal-kubeops` with the default rbac created by kubeapps is not able to retrieve the `appreposiories`. Indeed, performing:
> > ```
> > kubectl auth can-i get  apprepositories.kubeapps.com --as=system:serviceaccount:kubeapps:kubeapps-internal-kubeops
> > > no
> > ```
> 
> That is unexpected, I'll take a look.

I think you might just have missed the namespace? The rbac is for a role, not a cluster-role, so although the service account can't get them cluster-scoped, it can get them in the kubeapps namespace?

```
kubectl auth can-i get  apprepositories.kubeapps.com --as=system:serviceaccount:kubeapps:kubeapps-internal-kubeops
no

kubectl auth can-i get -n kubeapps apprepositories.kubeapps.com --as=system:serviceaccount:kubeapps:kubeapps-internal-kubeops
yes
```

which also may make more sense of the error message you're seeing, which shows it is failing to fetch app repositories in the "default" namespace. Is Kubeapps installed in the default namespace, or is that the issue?

dashboard/src/shared/url.ts
antgamdia: Well, the current behavior is: 
1. performs a POST request to `api/v1/clusters//can-i`
2. got a 302/3 redirect (don't remember) to `api/v1/clusters/can-i`
3. browser (ignoring the RFC) changes POST to GET methdod
4. kubeops receives GET `api/v1/clusters/can-i`, returning a 405 method not allowed

I think we should still handle this route for avoiding odd browser behavior, don't you think?
Regarding the can-i logic, replying separately.

pkg/http-handler/http-handler.go
antgamdia: Mmm, that's absolutely right. I initially thought that it could be good to have fine-grained information about the rbac granted to the kubeops service account, but it may not be a great idea as pointed out.

However, I don't see that danger, once the user is authenticated, all the actions that can be performed using the service account are the same regardless of the user's privileges, aren't they?

I mean, a user with, let's say, view-only permissions, can still manually invoke those actions that eventually will fall back to the serviceaccount, no? I can't foresee any further issues apart from the privilege enumeration (albeit enough to justify removing the changes herein introduced)

antgamdia:
> you mean that (c) is a similar setup but with kubeapps installed in a local kind cluster but configured with just the "tmc" cluster as the target for installations?

Yep, exactly. I should have elaborated it more :S

> so although the service account can't get them cluster-scoped, it can get them in the kubeapps namespace? (...)  which shows it is failing to fetch app repositories in the "default" namespace. Is Kubeapps installed in the default namespace, or is that the issue?

Exactly, that's the odd thing here. Even if I'm declaring the initialRepos (with name and URL, so it should be installed in the global, aka `kubeapps`, namespace) and installing Kubeapps on the `kubeapps` ns... the error is the SA trying to access to the `default` ns 

That's why I promoted this SA to have a clusterrole, but it, clearly, is not the solution.

dashboard/src/shared/url.ts
absoludity: > I think we should still handle this route for avoiding odd browser behavior, don't you think? I mean, I'd rather go for consistent error codes instead of having these redirections regardless of the implemented logic.

What are the call-sites that are calling this without a cluster name? I would think that's the issue - the dashboard should not be trying to check if the user has perms without a cluster name. But that might need some investigation (ie. separate PR).

antgamdia: Ok, I totally agree, let's remove all can-i stuff here and let me file another issue for the future. We better have this PR more focused, you're right

antgamdia:
Well... this PR has been very useful for me to learn more about the different authentication flows (as user / as serviceaccount) 
Thank you for the fruitful comments here!

Just to sum up, let me highlight some aspects here:

This PR now simply brings three changes (+ tests)
  - Make `GetAppRepoAndRelatedSecrets` use the asSVC if `isKubeappsCluster && (nonUIKubeappsCluster || isGlobalAppRepository) `. 
  - Make `NewClusterConfig` use the `inCluster` config if the cluster is empty.
  - Make `getSvcClientsetForCluster` use the asSVC approach if the cluster is empty.

Furthermore, as a result of the discussions, two issues have been created: https://github.com/kubeapps/kubeapps/issues/2717 and  https://github.com/kubeapps/kubeapps/issues/2719

I think that once landed, we can start triggering the new release (as these two issues are not that important/blocker as this one was). Have a look whenever you have a chance to approve the changes @absoludity   (no rush)

absoludity (approved):
Excellent, thanks for the precision there :)

PR #2718, opened 2021-04-27
Update docstring for the ttl-lifetime-afterfinished-job

absoludity:
### Description of the change

Just noticed some inconsistency in the name being documented and the actual var. Fixed that and noted the feature gate requirement.


antgamdia (approved):
Thanks for the clarification!

Issue #2704, opened 2021-04-22
Standardize chart workflow


antgamdia:
Since we are explicitly making the [Kubeapps bitnami chart](https://github.com/bitnami/charts/tree/master/bitnami/kubeapps) the official one so that they can add the proper standardization features and perform improvements on their own, we need first to accomplish:

### Kubeapps end
- [x] Release the upcoming version with the current approach
- [x] Revisit our user docs to explicitly state that we are using the Bitnami chart, so the internal one is just for development purposes, with no support.
- [x] Remove any automagic step with regards to the Bitnami chart 
  - [x] CI steps?
  - [x] ./scripts folder?
- [x] Update our dev docs to match the current release process
- [x] Chart automatic syncing (merged - https://github.com/kubeapps/kubeapps/pull/2754)
- [x] Merge the bitnami changes in our dev chart (merged - https://github.com/kubeapps/kubeapps/pull/2795)

### Bitnami end
- [x] Review the Bitnami chart PR (ongoing - https://github.com/bitnami/charts/pull/6301)
- [x] Check whether the tests in the bitnami pipeline can or should be updated or are fine as they are ( https://github.com/kubeapps/kubeapps/issues/2704#issuecomment-833394288)

antgamdia:
For these early stages, we can perform the PR process manually, but, what if we change the current `sync_chart` CI job to make use of the GitHub CLI to send an automatic PR to the bitnami/chart repo?

https://cli.github.com/manual/gh_pr_create

absoludity:
> For these early stages, we can perform the PR process manually, but, what if we change the current `sync_chart` CI job to make use of the GitHub CLI to send an automatic PR to the bitnami/chart repo?
> 
> https://cli.github.com/manual/gh_pr_create

Possibly. Will we want to create a PR for each change to the chart? IMO there will be some changes that we do want a PR for immediately (together with a chart point release), such as a bug fix in the chart, and others that we don't (such as feature work going on in devel). Up until now, we've always just done a chart release with whatever is in our main branch when we bump the chart version, and perhaps that's OK to continue for now, but I do wonder whether we'll eventually need to use a release branch to better manage changes. Let's see.

antgamdia:
>  but I do wonder whether we'll eventually need to use a release branch to better manage changes

Yep, absolutely! I just mentioned the continuous-PR approach so that we preserve the same behavior as we just had. 
[...]

PR #2725, opened 2021-04-27
Update deps for the upcoming release

antgamdia:
### Description of the change

As per [the docs](https://github.com/kubeapps/kubeapps/blob/master/docs/developer/release-process.md), this PRs upgrades:

- Dockerfile build images
- Chart image versions
- Chart dependencies
- Chart version
- Versions used in CI and integration (to match the others)
- Dashboard dependencies (yarn upgrade) + update types
- Golang dependencies (go mod tidy)
- Rust dependencies (cargo update)

### Benefits

Our upcoming release will use the latest version

### Possible drawbacks

Can't think of any right now, every bumped version should not contain breaking changes.

### Applicable issues

- #2704

### Additional information

Hopefully, last time we have to upgrade the chart for Bitnami.

Let's see if our beloved CI is happy with the changes.

Ah, there are a couple of tests in masters failing; that makes sense since they are using the new css selectors for the new app repo form -but- they're using the latest released image (which does not contain the changes)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/kubeapps/kubeapps
 
-go 1.13
+go 1.16
 
 replace github.com/docker/docker => github.com/docker/docker v0.0.0-20190731150326-928381b2215c
```
antgamdia, line 5: I'm wondering if it is still necessary, I'd rather say it's not, but I wanted to check it with you first

chart/kubeapps/values.yaml
```diff
@@ -335,7 +335,7 @@ hooks:
   image:
     registry: docker.io
     repository: bitnami/kubectl
-    tag: 1.19.9-debian-10-r7
+    tag: 1.19.10-debian-10-r9
```
antgamdia, line 338: Since kubectl 1.21 has been released, kubectl 1.19 will be deprecated sooner or later, perhaps we should think of changing it someday.

absoludity: Yep, let's do it straight after this release in case there's any deprecations is 1.21 apis that cause us pain.

go.mod
absoludity: If what's necessary? Changing the go version in go.mod, or replacing docker/docker ?

absoludity (approved):

go.mod
antgamdia: Oops, I thought I selected the proper line, but this excerpt is confusing indeed. I meant the version replacement, I (want to) suppose that whichever change required in this version is now upstream, no?

go.mod
```diff
@@ -2,27 +2,60 @@ module github.com/kubeapps/kubeapps
...
 
-replace github.com/docker/docker => github.com/docker/docker v0.0.0-20190731150326-928381b2215c
+replace (
+	// required by Oras https://github.com/deislabs/oras/blob/main/go.mod
+	github.com/docker/distribution => github.com/docker/distribution v2.7.1+incompatible
+	github.com/docker/docker => github.com/moby/moby v20.10.6+incompatible
+
+	// k8s.io/kubernetes is not intended to be used as a module, so versions are not being properly resolved.
+	// This replacement is required, see https://github.com/kubernetes/kubernetes/issues/79384
+	// As we support new k8s versions, this replacements should be also updated accordingly.
```
antgamdia, line 12: Having a closer look at the deps, I realized of two issues:

- we were stuck to k8s/kubernetes 0.13 version, which was introducing some security problems (see our Snyk).
- our oras version is conflicting with the one required by helm, so it cannot be upgraded

The solution is far from being elegant, but since the version cannot be resolved automatically, both `go mod -u` and dependabot are unable to update these deps

antgamdia:
Let's merge! However, I'm gonna perform later on a thorough manual review prior the tagging and will revert any change with undesired effect.

go.mod
absoludity: Urg, yeah. And how are these replacements kept up to date?

antgamdia: They don't (automatically), however, whenever we got a new version of the k8s go modules (and we want to upgrade), it is just a matter of changing every dep at the same time.
Fortunately, the versioning is very consistent: 0.xx.yy for the go modules and 1.xx.yy for the main k8s.io/kubernetes.

If they ever publish this dep as a module, we'll be able to remove these replacements.

Regarding the other, when Helm go package got updated, I guess they will be using the same Oras version as we currently use and there will be no need of having the replc.

But, the alternative was not being able to upgrade helm and k8s, a thing that was introducing about 20 vulnerabilities.
However, not a go dep expert at all, if you have any better solution I'm not aware of, I'll happily send another PR for fixing it.

Issue #2615, opened 2021-03-29
Support proxying websocket connections in pinniped-proxy


antgamdia:
This is a follow-up issue of the #2328 one.

Pinniped proxy is not proxying the ws connections; as a result, calls to the k8s API with `?watch=true` will fail. Once the _101 switching protocols_ HTTP status code is received, no more messages are routed by the pinniped-proxy.

The current solution has been just to force the trigger of the existing fallback mechanism in the front-end, that is, explicitly send a >400 status code when a 101 response is received from the server.

Whereas Kubeapps is working as is, the pinniped-proxy is still not able to handle ws connections.

absoludity:
Interestingly, we can close this one now since we no longer use web sockets to poll for changes, but rather server-side events (via grpc streaming).

PR #2728, opened 2021-04-28
Add --proxy-websockets arg to the oauth2-proxy binary

absoludity:
### Description of the change

I was perusing the oauth2-proxy docs looking for something else and noticed that there's an explicit arg for proxying websocket requests. It's not been noticed since we fallback to polling, but I'm guessing without this option the plain OIDC setup would be polling.

### Benefits

Updates for in-progress deployments can now be handled by websocket requests when using OIDC (but not pinniped, we still need to update pinniped-proxy to handle that, see #2615).

### DIFF
--- a/chart/kubeapps/Chart.yaml
+++ b/chart/kubeapps/Chart.yaml
@@ -25,4 +25,4 @@ maintainers:
 name: kubeapps
 sources:
   - https://github.com/kubeapps/kubeapps
-version: 6.1.1
+version: 6.1.2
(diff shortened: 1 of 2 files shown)


antgamdia (approved):
Thanks for the change, I already saw this flag when trying to debug what was happening with the pinniped-proxied requests. It was formerly an "alpha" flag, and then it got promoted to stable, afaik.

However, I'd bet it is unnecessary (unless we just want to make explicit we are using it), since it is `true` is the default value:

I mean, I'm ok not relying on 3rd party defaults, but I just wanted to be sure you knew it :)

image

https://oauth2-proxy.github.io/oauth2-proxy/docs/next/configuration/overview

absoludity:
> I mean, I'm ok not relying on 3rd party defaults, but I just wanted to be sure you knew it :)

Hah, no - I failed to check the default value. Preference for less arguments to be clear what we're actually changing from the defaults, so I'll just close this :) Thanks!

PR #2731, opened 2021-04-28
Enables user-provided tls cert to be used again.

### LINKED ISSUE #2730: Development environment broken with duplicate tls certs
### Description:

A change to the chart (#2445) a couple of months ago broke the kind-cluster dev setup (I'd thought it was a local issue so had just worked around it) so that kubeapps doesn't deploy, but rather helm errors because there is a conflict with tls certs.

1. `make cluster-kind`
2. `make deploy-dev`]

```
helm --kubeconfig=/home/michael/.kube/kind-config-kubeapps install kubeapps ./chart/kubeapps --namespace kubeapps --create-namespace \
        --values ./docs/user/manifests/kubeapps-local-dev-values.yaml \
        --values ./docs/user/manifests/kubeapps-local-dev-auth-proxy-values.yaml \
        --values ./docs/user/manifests/kubeapps-local-dev-additional-kind-cluster.yaml
Error: rendered manifests contain a resource that already exists. Unable to continue with install: Secret "localhost-tls" in namespace "kubeapps" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "kubeapps"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "kubeapps"
```

Kubeapps should be deployed in the local cluster using tls with the pre-generated `localhost-tls` cert.

The problem is that with the change in #2445, our chart template now assumes that if you have tls enabled and don't have cert manager enabled, that it should generate an (untrusted) tls cert for you: https://github.com/kubeapps/kubeapps/blob/master/chart/kubeapps/templates/tls-secrets.yaml#L17

It seems like a valid use case to want to enable tls without cert manager and specify your own (trusted) cert.

absoludity:
### Description of the change

See #2730 for a description of the problem.

Note: needs to be landed after #2728 due to chart version bumps.

### Benefits

Users can again provide their own (trusted) tls cert and have Kubeapps use that.

### Possible drawbacks

Not sure - I've asked for more info on why we'd want the chart to generate an untrusted cert.

### Applicable issues

  - fixes #2730

### DIFF
--- a/chart/kubeapps/templates/tls-secrets.yaml
+++ b/chart/kubeapps/templates/tls-secrets.yaml
@@ -14,7 +14,7 @@ data:
 ---
 {{- end }}
 {{- end }}
-{{- if and .Values.ingress.tls (.Values.ingress.selfSigned) }}
+{{- if and .Values.ingress.tls .Values.ingress.selfSigned }}
 {{- $ca := genCA "kubeapps-ca" 365 }}
 {{- $cert := genSignedCert .Values.ingress.hostname nil (list .Values.ingress.hostname) 365 $ca }}
 apiVersion: v1
--- a/chart/kubeapps/templates/tls-secrets.yaml
+++ b/chart/kubeapps/templates/tls-secrets.yaml
@@ -14,19 +14,4 @@ data:
 ---
 {{- end }}
 {{- end }}
-{{- if and .Values.ingress.tls (not .Values.ingress.certManager) }}
-{{- $ca := genCA "kubeapps-ca" 365 }}
-{{- $cert := genSignedCert .Values.ingress.hostname nil (list .Values.ingress.hostname) 365 $ca }}
-apiVersion: v1
-kind: Secret
-metadata:
-  name: {{ printf "%s-tls" .Values.ingress.hostname }}
-  namespace: {{ .Release.Namespace }}
-  labels: {{ include "kubeapps.labels" . | nindent 4 }}
-type: kubernetes.io/tls
-data:
-  tls.crt: {{ $cert.Cert | b64enc | quote }}
-  tls.key: {{ $cert.Key | b64enc | quote }}
-  ca.crt: {{ $ca.Cert | b64enc | quote }}
-{{- end }}
 {{- end }}
--- a/chart/kubeapps/templates/tls-secrets.yaml
+++ b/chart/kubeapps/templates/tls-secrets.yaml
@@ -14,4 +14,19 @@ data:
 ---
 {{- end }}
 {{- end }}
+{{- if and .Values.ingress.tls (.Values.ingress.selfSigned) }}
+{{- $ca := genCA "kubeapps-ca" 365 }}
+{{- $cert := genSignedCert .Values.ingress.hostname nil (list .Values.ingress.hostname) 365 $ca }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ printf "%s-tls" .Values.ingress.hostname }}
+  namespace: {{ .Release.Namespace }}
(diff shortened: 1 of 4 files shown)


chart/kubeapps/templates/tls-secrets.yaml
```diff
@@ -14,19 +14,4 @@ data:
...
-metadata:
-  name: {{ printf "%s-tls" .Values.ingress.hostname }}
-  namespace: {{ .Release.Namespace }}
-  labels: {{ include "kubeapps.labels" . | nindent 4 }}
-type: kubernetes.io/tls
-data:
-  tls.crt: {{ $cert.Cert | b64enc | quote }}
-  tls.key: {{ $cert.Key | b64enc | quote }}
-  ca.crt: {{ $ca.Cert | b64enc | quote }}
-{{- end }}
```
juan131, line 31: I do find this useful for development purposes.
Instead, I'd recommend to add a `ingress.selfSigned` parameter to explicitly indicate when the self-signed cert should be created.

juan131 (changes requested):

chart/kubeapps/templates/tls-secrets.yaml
absoludity: > I do find this useful for development purposes.

You mean the generation right? In what situation when doing development do you find it useful? That is, if I am doing development and using TLS, I'd prefer to work with a cert that is trusted by my computer, which is exactly what I was doing here before this was added. But yep, perhaps others prefer differently, but I'm not keen on development options like this being in the chart by default if you switch TLS on (and used outside dev envs).

> Instead, I'd recommend to add a `ingress.selfSigned` parameter to explicitly indicate when the self-signed cert should be created.

OK, see what you think, I've updated to that.

absoludity: ping @juan131 ?

antgamdia: Just a side note for you all to know: from now on, the changes performed in this chart -won't - be automatically pushed to the bitnami/chart repository (as required as part of https://github.com/kubeapps/kubeapps/issues/2704)

absoludity: ping @juan131 Is there any reason you see for this not to land? It's now got exactly the same condition on the tls secret generation as you've landed in https://github.com/bitnami/charts/pull/6235/files

> Just a side note for you all to know: from now on, the changes performed in this chart -won't - be automatically pushed to the bitnami/chart repository (as required as part of #2704)

Yep, I've removed the chart bump from this PR because of that, but made me realise, we probably want a `devel` version in our local `Chart.yaml` otherwise it could get confusing when people  use our devel chart and it says it's the actual published version. I'll do that separately ad we can discuss there.

juan131: I'm okay with the current approach! Thanks @absoludity

juan131 (commented):
@absoludity you'll have to adapt the Ingress.yaml as I did at https://github.com/bitnami/charts/blob/master/bitnami/rabbitmq/templates/ingress.yaml

This part:

```yaml
  {{- if or (and .Values.ingress.tls (or .Values.ingress.certManager .Values.ingress.selfSigned)) .Values.ingress.extraTls }}
  tls:
    {{- if and .Values.ingress.tls (or .Values.ingress.certManager .Values.ingress.selfSigned) }}
    - hosts:
(...)
```

I did the same at https://github.com/bitnami/charts/pull/6301

absoludity:
Closing this in favour of integrating the change into https://github.com/bitnami/charts/pull/6301

PR #2735, opened 2021-04-28
Add try/catch in CI test to check the latest released version

antgamdia:
### Description of the change

Add try/catch to prevent CI to fail in master when testing the latest released version (which does not have the UI change)

### Benefits

CI will work on master

### Possible drawbacks

N/A

### Applicable issues

  N/A

### Additional information

This change should be reverted once we got the new release since it's no longer necessary.


antgamdia:
Let me merge this change as it just involves a single CI test that will be reverted soon.

absoludity:
+1

PR #2746, opened 2021-04-30
Run prettier in yaml and md files

antgamdia:
### Description of the change

This PR is simply the result of running the `pretter` tool that we already have configured when committing files under the `dashboard` folder in the rest of *.md and *.yaml files we currently have on the projet.

### Benefits

When sending additional PRs where `prettier` is being run, we won't have so many changes to review. 

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

It has been executed using the default configuration. Also, it could be a good idea to automatically run prettier as we are already doing for the dashboard.
Perhaps we can do it as part of the changes required for https://github.com/kubeapps/kubeapps/pull/2617.


absoludity (approved):
I've not checked all changes, given they're all mechanical.

absoludity:
> Also, it could be a good idea to automatically run prettier as we are already doing for the dashboard.
> Perhaps we can do it as part of the changes required for #2617.

+1 to automatically run prettier and I don't mind if you do that via husky, but let's not rely on that (ie. rely on a local user-setup to ensure the tool is run). In the past, I've tried to ensure that the dev image build will fail if things like the `prettier-check` run here: https://github.com/kubeapps/kubeapps/blob/master/dashboard/Dockerfile#L9 fail. I don't have the husky git hooks enabled (I don't particularly trust npm code to the degree that I want to automatically run it on my local machine as my user), but I want to be sure that I (or any other user) can't land code that contributes to the problem, which is why I did https://github.com/kubeapps/kubeapps/pull/1836 .

antgamdia:
Mmm, I see. That's totally understandable. Even I sometimes commit files using the ` --no-verify` flag.
If we really want to enforce the usage of these tools, they should be at "server-side", for instance, using GitHub Actions: https://github.com/marketplace/actions/prettier-action
Let me file an issue to track and evaluate it.

PR #2754, opened 2021-05-04
Automatic chart syncing in CI

antgamdia:
### Description of the change

This PR brings an automated way to fetch changes from the Bitnami chart and, analouglsy, send our changes back to them,
All the process is being done through pull requests; however, push access is still needed just to avoid creating our own bitnami/chart fork.

### Benefits

The synchronization process will be full-duplex automatic :)

### Possible drawbacks

The process assumes a linear series of events: for instance, multiple ongoing PRs are not supported.
So, what is expected is:

- Perfect sync
- Push to master: no new chart version
- Bitnami update some images
- Push to master: pull changes from bitnami (via PR)
- New PR: kubeapps accepts it
- Push to master: no new chart version
- Kubeapps perform some changes in the chart
- Kubeapps release: push changes to bitnami (via PR)
- Bitnami accepts the PR
- Kubeapps got released

### Applicable issues

- rel #2704

### Additional information

Example PRs:

1. From Bitnami to Kubeapps: https://github.com/antgamdia/kubeapps/pull/205
2. From Kubeapps to Bitnami: https://github.com/antgamdia/charts/pull/4

### DIFF
--- a/docs/developer/ci.md
+++ b/docs/developer/ci.md
@@ -1,8 +1,8 @@
 # Understanding the CircleCI configuration
 
-Kubeapps leverages CircleCI for running the tests (both unit and integration tests), pushing the images and syncing the chart with the official [Bitnami chart](https://github.com/bitnami/charts/tree/master/bitnami/kubeapps). The following image depicts how a successful workflow looks like (after a push to master).
+Kubeapps leverages CircleCI for running the tests (both unit and integration tests), pushing the images and syncing the chart with the official [Bitnami chart](https://github.com/bitnami/charts/tree/master/bitnami/kubeapps). The following image depicts how a successful workflow looks like (after a push to the main branch).
 
-![CircleCI workflow after pushing to master](../img/ci-workflow-master.png "CircleCI workflow after pushing to master")
+![CircleCI workflow after pushing to the main branch](../img/ci-workflow-master.png "CircleCI workflow after pushing to the main branch")
 
 The main configuration is located at this [CircleCI config file](../../.circleci/config.yml)). At a glance, it contains:
 
@@ -21,10 +21,67 @@ The main configuration is located at this [CircleCI config file](../../.circleci
     - Spin up two Kind clusters.
     - Load the CI images into the cluster.
     - Run the integration tests.
-  - `GKE_x_xx_MASTER` and `GKE_X_XX_LATEST_RELEASE` (on master): there is a job for each [Kubernetes version supported by Google Kubernetes Engine](https://cloud.google.com/kubernetes-engine/docs/release-notes) (GKE). It will run the e2e tests in a GKE cluster (version X.XX) using either the code in `master` or in the latest released version. If a change affecting the UI is pushed to master, the e2e test might fail here. Use a try/catch block to temporarily work around this.
+  - `sync_chart_from_bitnami` (on master): each time a new commit is pushed to the main branch, it brings the current changes in the upstream [bitnami/charts repository](https://github.com/bitnami/charts/tree/master/bitnami/kubeapps) and merge the changes. This step involves:
+    - Checking if the chart versions differ.
+    - Deleting the local `chart/kubeapps` folder.may the 4th
+    - Pulling the latest version of the chart provided by Bitnami.
+    - Renaming the production images (`bitnami/kubeapps-xxx`) by the development ones (`kubeapps/xxx`) with the `latest` tag.
+    - Using `DEVEL` as the `appVersion`.
+    - Sending a PR in the Kubeapps repository with these changes.
+  - `GKE_X_XX_MASTER` and `GKE_X_XX_LATEST_RELEASE` (on master): there is a job for each [Kubernetes version supported by Google Kubernetes Engine](https://cloud.google.com/kubernetes-engine/docs/release-notes) (GKE). It will run the e2e tests in a GKE cluster (version X.XX) using either the code in `master` or in the latest released version. If a change affecting the UI is pushed to the main branch, the e2e test might fail here. Use a try/catch block to temporarily work around this.
   - `push_images` (on master): the CI images (which have already been built) get re-tagged and pushed to the `kubeapps` account.
+  - `sync_chart_to_bitnami` (on tag): when releasing, it will synchronize our development chart with the [bitnami/charts repository](https://github.com/bitnami/charts/tree/master/bitnami/kubeapps) and merge the changes. This step involves:
+    - Deleting the local `bitnami/kubeapps` folder.
+    - Pulling the latest version of the chart provided by Kubeapps.
+    - Renaming the development images (`kubeapps/xxx`) by the production ones (`bitnami/kubeapps-xxx`) with the `vX.X.X` tag.
+    - Using `vX.X.X` as the `appVersion`.
+    - Sending a PR in the Bitnami Charts repository with these changes.
   - `release` (on tag): it creates a GitHub release based on the current tag by executing the script [script/create_release.sh](../../script/create_release.sh).
 
 Note that this process is independent of the release of the official Bitnami images and chart. These Bitnami images will be created according to their internal process (so the Golang, Node or Rust versions we define here are not used by them. Manual coordination is expected here if a major version bump happens to occur).
 
-Additionally, currently the Kubeapps team is responsible for sending a PR to the [chart repository](https://github.com/bitnami/charts/tree/master/bitnami/kubeapps) each time a new chart version is pushed to the main branch.
+Also note it is the Kubeapps team that is responsible for sending a PR to the [chart repository](https://github.com/bitnami/charts/tree/master/bitnami/kubeapps) each time a new chart version is pushed to the main branch. Even this process is automatic (using the `sync_chart_to_bitnami` workflow), Kubeapps maintainers may decide to send a manual PR early.
+
+# Credentials
+
+Besides other usual credentials or secrets passed through environment variables via the CircleCI user interface, it is important to highlight how we grant commit and PR access to our robot account `kubernetes-bitnami <kubernetes@bitnami.com>`. The process is twofold:
+
+- Create a [personal access token](https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token) with the robot account, granted, at least, with: `repo:status`, `public_repo` and `read:org`.
+  - That will allow the GitHub CLI to create PRs from the command line on behalf of our robot account.
+- Add deployment keys to the repositories in which the CI will commit. Currently, they are `kubeapps/kubeapps` and `bitnami/charts`.
+  - This step allows pushing branches to the target repositories, otherwise, we would have to create a private fork under the robot account. However, the CI never pushes to the main branch as it tries to create a pull request.
+
+## Generating and configuring the deployment keys
+
+This step is only run once, and it is very unlikely to change. However, it is important to know it in case of secret rotations or further events.
+
+```bash
+# COPY THIS CONTENT TO GITHUB (with write access):
+## https://github.com/kubeapps/kubeapps/settings/keys
+ssh-keygen -t ed25519 -C "user@example.com" -q -N "" -f circleci-kubeapps-deploymentkey
+echo "Kubeapps deployment key (public)"
+cat circleci-kubeapps-deploymentkey.pub
+
+# COPY THIS CONTENT TO GITHUB (with write access):
+## https://github.com/bitnami/charts/settings/keys
+ssh-keygen -t ed25519 -C "user@example.com" -q -N "" -f circleci-charts-deploymentkey
+echo "Charts deployment key (public)"
+cat circleci-charts-deploymentkey.pub
+
+# COPY THIS CONTENT TO CIRCLECI (hostname: github.com):
+## https://app.circleci.com/settings/project/github/kubeapps/kubeapps/ssh
+echo "Kubeapps deployment key (private)"
+cat circleci-kubeapps-deploymentkey
+
+echo "Charts deployment key (private)"
+cat circleci-charts-deploymentkey
+
+# COPY THE FINGERPRINTS TO ".circleci/config.yml"
+## sync_chart_from_bitnami
+echo "Charts deployment key (fingerprint) - edit 'sync_chart_from_bitnami'"
+ssh-keygen -l -E md5 -f circleci-kubeapps-deploymentkey.pub
+
+## sync_chart_to_bitnami
+echo "Charts deployment key (fingerprint) - edit 'sync_chart_to_bitnami'"
+ssh-keygen -l -E md5 -f circleci-charts-deploymentkey.pub
+```
--- a/docs/developer/ci.md
+++ b/docs/developer/ci.md
@@ -21,21 +21,22 @@ The main configuration is located at this [CircleCI config file](../../.circleci
     - Spin up two Kind clusters.
     - Load the CI images into the cluster.
     - Run the integration tests.
-  - `sync_chart_from_bitnami` (on master): each time a new commit is pushed to the main branch, it brings the current changes in the upstream [bitnami/charts repository](https://github.com/bitnami/charts/tree/master/bitnami/kubeapps) and merge the changes. This step involves:
-    - Checking if the chart versions differ.
-    - Deleting the local `chart/kubeapps` folder.
+  - `sync_chart_from_bitnami` (on master): each time a new commit is pushed to the main branch, it brings the current changes in the upstream [bitnami/charts repository](https://github.com/bitnami/charts/tree/master/bitnami/kubeapps) and merges the changes. This step involves:
+    - Checking if the Bitnami chart version is greater than the Kubeapps development chart version. If not, abort.
+    - Deleting the local `chart/kubeapps` folder (note that the changes are already commited in git).
     - Pulling the latest version of the chart provided by Bitnami.
     - Renaming the production images (`bitnami/kubeapps-xxx`) by the development ones (`kubeapps/xxx`) with the `latest` tag.
     - Using `DEVEL` as the `appVersion`.
-    - Sending a PR in the Kubeapps repository with these changes.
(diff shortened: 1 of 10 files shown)


script/PR_external_chart_template.md
```diff
@@ -0,0 +1,27 @@
...
+
+This is an automatic PR for synchronizing the changes performed externally at the [kubeapps development chart](https://github.com/kubeapps/kubeapps/tree/master/chart/kubeapps) official chart.
+
+**Benefits**
+
+The Kubeapps [bitnami/kubeapps chart](https://github.com/bitnami/charts/tree/master/bitnami/kubeapps) will get the newest changes, including image version updates and other updates.
+
+**Possible drawbacks**
+
+Even unlikely, some changes implemented by the Kubeapps team may not be 100% compatible with the current chart version.
```
absoludity, line 13: We should never be pushing new kubeapps application images that change functionality based on chart input without a new chart release I think? I don't understand when this PR would be triggered unless when we're wanting to release a new chart version?
```suggestion
Although unlikely, some changes implemented by the Kubeapps team may not be 100% compatible with the current chart version.
```

script/PR_internal_chart_template.md
```diff
@@ -0,0 +1,19 @@
+### Description of the change
+
+This is an automatic PR for synchronizing the changes performed externally at the [bitnami/kubeapps](https://github.com/bitnami/charts/tree/master/bitnami/kubeapps) official chart.
```
absoludity, line 3:
```suggestion
This is an automatic PR for synchronizing the changes performed externally at the [bitnami/kubeapps official chart](https://github.com/bitnami/charts/tree/master/bitnami/kubeapps) to the Kubeapps development version.
```

script/PR_internal_chart_template.md
```diff
@@ -0,0 +1,19 @@
...
+
+### Benefits
+
+The Kubeapps [development chart](https://github.com/kubeapps/kubeapps/tree/master/chart/kubeapps) will get the newest changes, including image version updates and other updates.
+
+### Possible drawbacks
+
+Even unlikely, some changes implemented by the Bitnami team may not be 100% compatible with the current Kubeapps version.
```
absoludity, line 11: This part is OK since we need to update to cope, I guess.
```suggestion
Although unlikely, some changes implemented by the Bitnami team may not be 100% compatible with the current Kubeapps version.
```

script/chart_sync.sh
```diff
@@ -15,20 +15,19 @@
 
 set -e
 
-CHARTS_REPO="bitnami/charts"
 source $(dirname $0)/chart_sync_utils.sh
 
 user=${1:?}
 email=${2:?}
 if changedVersion; then
```
absoludity, line 22: As mentioned elsewhere, I'm not sure we can rely on `changedVersion`, though maybe we can use semver to have something similar while still allowing local version changes in devel.

script/chart_sync_utils.sh
```diff
@@ -87,15 +109,40 @@ updateRepo() {
...
+    local targetTag=${2:?}
+    local targetTagWithoutV=${targetTag#v}
+    local targetChartPath="${targetRepo}/${CHART_REPO_PATH}"
+    local remoteChartYaml="${targetChartPath}/Chart.yaml"
+    local localChartYaml="${KUBEAPPS_CHART_DIR}/Chart.yaml"
+    if [ ! -f "${remoteChartYaml}" ]; then
+        echo "Wrong repo path. You should provide the root of the repository" > /dev/stderr
+        return 1
+    fi
+    rm -rf "${KUBEAPPS_CHART_DIR}"
```
absoludity, line 131: As mentioned above, not sure how this can work, since we lose any changes that we've made but not yet released.

script/chart_upstream_checker.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+# Copyright (c) 2018 Bitnami
```
absoludity, line 2:
```suggestion
# Copyright (c) 2021 Bitnami
```

absoludity (commented):
So there's a couple of things I'm unsure about here.

1. Why do we need to automate (and hence automatically detect) when we want to release a new chart version, given that that is a manual decision on our part (even if we trigger it by updating something, like a chart version). I think it'd be enough for us to have a script in our repo which generates a PR against bitnami/charts that *we* create, rather than having CI detect that it's what we want and create this for us? When we chatted the other day, I was trying to say that I think there may be a case to automate one direction (bitnami/chart to us), but not the other.
2. I agree we *may* want to automate the sync of the bitnami chart to our repo, depending on the frequency, but I don't think we can do this by deleting and replacing the directory without potentially losing any changes we've not yet released? Given that we don't yet know the frequency with which the bitnami chart will be updated (image updates or otherwise) it may be enough for the moment to automatically generate a patch of the change to the bitnami/kubeapps chart. We can then apply the patch to create a PR manually for now until we see how that works, whether it can be automated (I suspect we'll see conflicts because our devel chart has the tags changed, but we may be able to work around that).
3. AFAICT, the trigger for syncing the remote bitnami chart to our local devel is currently based on `changedVersion`. But I think we may need to do something a little smarter here, but hopefully easy with semver support. Something like `if bitnamiVersion > kubeappsVersion` (but with a semver comparison).

Lastly, I think we could probably simplify things for ourselves here if we didn't bother trying to enforce the chart in our kubeapps repo to use the tag `latest` for our app images. Is there a reason we need to do this now? (or was there in the first place?). Given that it's our local development repository, we can easly deploy the local chart with a `development.yaml` values file to set the repos and tags appropriately. Is there a reason to use different repo names even? Worth thinking through before perpetuating what's already there. If we did have the same repos/tags in the chart in our repo, then the *only* difference between our development chart (without any extra values) and the bitnami one, would be the changes that we intend to release so that patching in either direction should always be trivial. Thoughts? (If, after thinking about it, you agree, let's do that in a prequel PR rather than adding it here, so the PR just tackles that one task independently).

script/chart_upstream_checker.sh
antgamdia: I did change it, trying to be more aligned with other similar projects [1] (i.e., using SPDX and something like _the Kubeapps contributors._), but I noticed it would require a further discussion. This change should propagate across the rest of the files in our project.
But, while undoing the changes, it got changed back to 2018 :facepalm: 
Thanks for catching it!

[1] https://github.com/vmware-tanzu/pinniped/blob/a8bccc5432a1aae7e48cd931dd1b31b66049f3ea/hack/check-copyright-year.sh#L3-L4

antgamdia:
> 1. Why do we need to automate (and hence automatically detect) when we want to release a new chart version, given that that is a manual decision on our part (even if we trigger it by updating something, like a chart version). 

I guess each time we are about to release, we may want to send our latest changes to the _bitnami/charts_ repo. I can't think of an example wherein the official chart is not getting the latest dev changes.
That assumed, given that it will be a periodic thing... why don't we just automate it as part of our CI. Of course, it is not preventing us from manually invoking the script. In the end, what the CI is doing is just loading some env vars and running a certain `.sh` file.

>I think it'd be enough for us to have a script in our repo which generates a PR against bitnami/charts that _we_ create, rather than having CI detect that it's what we want and create this for us? When we chatted the other day, I was trying to say that I think there may be a case to automate one direction (bitnami/chart to us), but not the other.

Then, I totally misunderstood what you tried to say (https://github.com/kubeapps/kubeapps/issues/2704#issuecomment-831094327)

But, well, no strong opinions here. It is just a matter of documenting how to invoke the script as part of the release process instead. 
_Security concern_: depending on how we implement it, the deployment key will be stored in our laptops instead of having it on CircleCI.


> 2. I agree we _may_ want to automate the sync of the bitnami chart to our repo, depending on the frequency, but I don't think we can do this by deleting and replacing the directory without potentially losing any changes we've not yet released?

I'd rather say we won't be deleting unreleased things. Given that everything is committed and pushed upstream to the main branch, the PR will compare two  "stable" versions. Later, an automated PR will be sent back to us, which is great. It is a starting point. I don't mean we have to accept/reject it, but use it as initial work - so we manually can edit the PR and send more commits.

It is just a way to make sure we don't miss any Bitnami updates. Even it's totally fine for us just to ignore the PRs; when releasing, we will sending a PR with our changes anyway.

> Given that we don't yet know the frequency with which the bitnami chart will be updated (image updates or otherwise) it may be enough for the moment to automatically generate a patch of the change to the bitnami/kubeapps chart. 

But, we are not gonna update our chart with the same frequency. The sync will just be triggered whenever a push to the main branch happens. 

> We can then apply the patch to create a PR manually for now until we see how that works, whether it can be automated (I suspect we'll see conflicts because our devel chart has the tags changed, but we may be able to work around that).

But the script is ignoring the kubeapps image tags and the appVersion indeed. So in this case (just updating kubeapps revisions), we only nothing but a chartVersion bump up. We accept the PR and that's all from our side.

So, again, I don't see any reason for not automating it.

> 3. AFAICT, the trigger for syncing the remote bitnami chart to our local devel is currently based on `changedVersion`. But I think we may need to do something a little smarter here, but hopefully easy with semver support. Something like `if bitnamiVersion > kubeappsVersion` (but with a semver comparison).

Yep, here I do agree with you. I guess, using semver comparisons will prevent race conditions. I can look at it, thanks for pointing this out.

> Lastly, I think we could probably simplify things for ourselves here if we didn't bother trying to enforce the chart in our kubeapps repo to use the tag `latest` for our app images. Is there a reason we need to do this now? (or was there in the first place?). Given that it's our local development repository, we can easly deploy the local chart with a `development.yaml` values file to set the repos and tags appropriately. Is there a reason to use different repo names even? Worth thinking through before perpetuating what's already there. If we did have the same repos/tags in the chart in our repo, then the _only_ difference between our development chart (without any extra values) and the bitnami one, would be the changes that we intend to release so that patching in either direction should always be trivial. Thoughts? (If, after thinking about it, you agree, let's do that in a prequel PR rather than adding it here, so the PR just tackles that one task independently).

<TBD> Will edit it in a while

docs/developer/ci.md
```diff
@@ -21,10 +21,67 @@ The main configuration is located at this [CircleCI config file](../../.circleci
     - Spin up two Kind clusters.
     - Load the CI images into the cluster.
     - Run the integration tests.
-  - `GKE_x_xx_MASTER` and `GKE_X_XX_LATEST_RELEASE` (on master): there is a job for each [Kubernetes version supported by Google Kubernetes Engine](https://cloud.google.com/kubernetes-engine/docs/release-notes) (GKE). It will run the e2e tests in a GKE cluster (version X.XX) using either the code in `master` or in the latest released version. If a change affecting the UI is pushed to master, the e2e test might fail here. Use a try/catch block to temporarily work around this.
+  - `sync_chart_from_bitnami` (on master): each time a new commit is pushed to the main branch, it brings the current changes in the upstream [bitnami/charts repository](https://github.com/bitnami/charts/tree/master/bitnami/kubeapps) and merge the changes. This step involves:
+    - Checking if the chart versions differ.
```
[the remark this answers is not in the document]
antgamdia: Yep, I'm kinda ok with that. Having a `-devel` version even would make it more explicit for our users that it is a development version.
However, I still think this version should be the same as is upstream. I mean, `bitnami/kubeapps v6.2.3` and `kubeapps/kubeapps v6.2.3-devel`

docs/developer/ci.md
```diff
@@ -21,10 +21,67 @@ The main configuration is located at this [CircleCI config file](../../.circleci
...
+
+# Credentials
+
+Besides other usual credentials or secrets passed through environment variables via the CircleCI user interface, it is important to highlight how we grant commit and PR access to our robot account `kubernetes-bitnami <kubernetes@bitnami.com>`. The process is twofold:
+
+- Create a [personal access token](https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token) with the robot account, granted, at least, with: `repo:status`, `public_repo` and `read:org`.
+  - That will allow the GitHub CLI to create PRs from the command line on behalf of our robot account.
+- Add deployment keys to the repositories in which the CI will commit. Currently, they are `kubeapps/kubeapps` and `bitnami/charts`.
+  - This step allows pushing branches to the target repositories, otherwise, we would have to create a private fork under the robot account. However, the CI never pushes to the main branch as it tries to create a pull request.
```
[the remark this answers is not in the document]
antgamdia: Yep, it is a good alternative as well. We, therefore, will be totally independent of them, which is great.
I just stuck close to the former approach, but I'm totally ok with creating our fork... and our own robot account. 
`kubernetes-bitnami` doesn't seem a good name if we claim to be independent : P

PS: Pinniped folk have this cute Pinny bot: https://github.com/pinniped-ci-bot

script/PR_external_chart_template.md
```diff
@@ -0,0 +1,27 @@
+Signed-off-by: <USER> <EMAIL>
+
+**Description of the change**
+
+This is an automatic PR for synchronizing the changes performed externally at the [kubeapps development chart](https://github.com/kubeapps/kubeapps/tree/master/chart/kubeapps) official chart.
+
+**Benefits**
+
+The Kubeapps [bitnami/kubeapps chart](https://github.com/bitnami/charts/tree/master/bitnami/kubeapps) will get the newest changes, including image version updates and other updates.
```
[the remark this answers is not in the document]
antgamdia: I think I just copy&pasted wrongly, perhaps:

```suggestion
The Kubeapps [bitnami/kubeapps chart](https://github.com/bitnami/charts/tree/master/bitnami/kubeapps) will get the latest changes used in the Kubeapps development chart.
```

script/chart_sync.sh
antgamdia: +1 to the semver detection to avoid race conditions. But, that said, I still think is approach is valid as long we accept and merge PR sequentially.

script/chart_sync_utils.sh
antgamdia: Not sure about it, could you elaborate  I mean, a simple example in which you think this situation might happen.

antgamdia (commented):
Thanks for your valuable comments, as usual!
So, if I understood correctly, we are agree on:

- [ ]  Add semver to the scripts
- [ ] Create a robot account 
  - [ ] Fork bitnami/charts
- [ ] Add `-devel` suffix to our chart version

(My) Tomorrow I'll start addressing them unless you say otherwise

On the other hand, the open issues are: 

- how periodic this process (inbound and outbound) should be? 
- is removing local changes and applying new ones a good approach?

absoludity (approved):
We chatted through the different points here and are both happy for this to land and we can start trying it, with the caveat that we should not require a credential for the bitnami repo at all (ie. we should create a PR from a repo we control, rather than pushing a branch for a PR to the bitnami one).

Thanks!

absoludity (approved):
Thanks for adding the semver comparison and the gpg support.

antgamdia:
Yesterday I worked on the new changes in my local and today I've dealt with integrating them into the CircleCI pipeline. The main highlights are:

- PGP signatures are working ([example](https://github.com/antgamdia/charts/commit/05a32f5149ab8ba15da6c4e9fd427f15b8efbb73))
- Semver is properly used (otherwise ignore - it would prevent some race conditions in case of multiple PR being recklessly accepted).
  - Sync from local to Bitnami iff KubeappsVersion > BitnamiVersion
  - Sync from Bitnami to local iff BitnamiVersion > KubeappsVersion 
- Now it uses the fork instead of pushing to the Bitnami repo (woah, 100% decoupled now!)
  - It wasn't that easy `gh pr create` has some open issues when dealing with multiple remotes.

Please find below some examples; I've done my best to replicate a realistic scenario but I think we should be ready to expect some issues:

- [From local to Bitnami](https://github.com/antgamdia/charts/pull/5)
- [From Bitnami to local](https://github.com/antgamdia/kubeapps/pull/217)

absoludity:
> Yesterday I worked on the new changes in my local and today I've dealt with integrating them into the CircleCI pipeline. The main highlights are:
> 
>     * PGP signatures are working ([example](https://github.com/antgamdia/charts/commit/05a32f5149ab8ba15da6c4e9fd427f15b8efbb73))
> 
>     * Semver is properly used (otherwise ignore - it would prevent some race conditions in case of multiple PR being recklessly accepted).
>       
>       * Sync from local to Bitnami iff KubeappsVersion > BitnamiVersion
>       * Sync from Bitnami to local iff BitnamiVersion > KubeappsVersion
> 
>     * Now it uses the fork instead of pushing to the Bitnami repo (woah, 100% decoupled now!)
>       
>       * It wasn't that easy `gh pr create` has some open issues when dealing with multiple remotes.
> 

Excellent work!

> 
> Please find below some examples; I've done my best to replicate a realistic scenario but I think we should be ready to expect some issues:
> 
>     * [From local to Bitnami](https://github.com/antgamdia/charts/pull/5)

This one looks clean. I wonder whether we should have it created as a draft so that *we* check it before actually proposing it (for example, it'll include current feature work which we might want to remove before the content team deal with it?)

> 
>     * [From Bitnami to local](https://github.com/antgamdia/kubeapps/pull/217)

This one seems to be including a bunch of unrelated changes - haven't dug to find out why, but perhaps the branch onto which the bitnami chart changes were added was behind or ahead of your main branch?

antgamdia:
> This one looks clean. I wonder whether we should have it created as a draft so that we check it before actually proposing it (for example, it'll include current feature work which we might want to remove before the content team deal with it?)

Yes, totally agree. I'd bet it was already marked as a draft. Pushing the change!

> [From Bitnami to local](https://github.com/antgamdia/kubeapps/pull/217)
This one seems to be including a bunch of unrelated changes - haven't dug to find out why, but perhaps the branch onto which the bitnami chart changes were added was behind or ahead of your main branch?

Exactly, I created a branch in my repo with some changes (modifying the CI config, modifying the scripts and also performing the modifications in the chart). You can safely ignore that changes, since they were sent in the PR cause it was the same branch in which I was testing both the CI changes and the chart changes. Otherwise I would have required to push to master my CI changes, and then, creating another PR with the chart changes.

antgamdia:
Oook, let's land! Let's see if it works when Bitnami merge its PR :crossed_fingers:

PR #2761, opened 2021-05-05
Fix rollback form

### LINKED ISSUE #2760: Unable to rollback app if not clicking on the <select>
### Description:

Unless we click on the select so that the onChange event is being fired, it is impossible to rollback an app since the value becomes `-1`.

1. Install an application
2. Upgrade this application (just once)
3. Click on rollback an application
4. An error will be thrown

image

Get the app successfully rollbacked

Discovered while doing https://github.com/kubeapps/kubeapps/issues/2705

DEV

antgamdia:
### Description of the change

The value was being set to `currentRevision -1`  and later being updated with the value of the `onChange` event. If the user wouldn't change the selected option, the event wouldn't trigger and, therefore, the revision value was totally wrong.
This PR checks the user input and relies on the react hooks to manage the lifecycle.

### Benefits

Rollbacks with a single revision will work again.

### Possible drawbacks

N/A

### Applicable issues

  - fixes #2760

### Additional information

N/A

### DIFF
--- a/dashboard/src/components/AppView/AppControls/RollbackButton/RollbackDialog.test.tsx
+++ b/dashboard/src/components/AppView/AppControls/RollbackButton/RollbackDialog.test.tsx
@@ -23,15 +23,20 @@ it("should render the form if it is not loading", () => {
 });
 
 it("should submit the current revision", () => {
+  const currentRevision = defaultProps.currentRevision;
   const onConfirm = jest.fn();
-  const wrapper = mount(<RollbackDialog {...defaultProps} onConfirm={onConfirm} />);
+  const wrapper = mount(
+    <RollbackDialog {...defaultProps} currentRevision={currentRevision} onConfirm={onConfirm} />,
+  );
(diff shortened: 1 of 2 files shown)


absoludity (approved):
Thanks Antonio, also for the explanation of what caused the regression here (updating to the clarity component, I think you said? Always good to put that kind of info on the PR - ie., not just what's wrong, but why it's wrong, as it's much more useful learning IMO, as long as we do it in a blameless way :) ).

antgamdia:
Thank you for your comment! Yep, it was when upgrading to the `<CdsSelect>` CDS component that it failed.
Hopefully, we always learn from these bugs:

- Nobody complained about the rollback feature (even if it has been kinda broken for 2 releases or so)... is it still a 'core' functionality? 
- Our unit test wasn't able to catch the error: perhaps jest/enzyme not being able to properly simulate the events? perhaps a mistake in writing the test?
- We didn't have any e2e of this functionality

PR #2762, opened 2021-05-05
Add rollback test

### LINKED ISSUE #2705: Add e2e test step that rollbacks an application
As per https://github.com/kubeapps/kubeapps/pull/2698#pullrequestreview-640563664, it would be useful to add an e2e test case performing a rollback so that it gets executed as part of our test suite in the CI.

antgamdia:
### Description of the change

This PR adds a new e2e test that rollbacks an application multiple times, so it tests: i) using the default value w/o interacting with the select; ii) specifying a custom value.

### Benefits

This use case will be periodically tested. Hopefully, we will notice if we break these things again :cry:

### Possible drawbacks

Eventual CI errors will be more likely to happen as we increase the number of tests, but it's worthwhile  :P

### Applicable issues

  - fixes #2705

### Additional information

N/A


antgamdia:
Not passing CI as #2761 hasn't been merged yet :P

antgamdia:
... and now it is passing  :)

absoludity (approved):
Nice. Why not instead update the existing 07-upgrade.js to be 07-upgrade-rollback.js, rather than having two integration tests which both perform a deploy + upgrade?

antgamdia:
> Why not instead update the existing 07-upgrade.js to be 07-upgrade-rollback.js, rather than having two integration tests which both perform a deploy + upgrade?

I did consider it, but I thought it was a good idea to separate both tests as it'll be easier to know which part failed. Also, I didn't want to add much complexity to the existing test. 
Also, if we don't update it, it will be semantically equivalent to the already implemented bitnami pipeline test.

That said, if you still consider it is a good idea to merge them, no strong opinion here, I'm happy to do so.

PR #2763, opened 2021-05-05
Use the credentialissuer in the pinniped docs

antgamdia:
### Description of the change

As pointed out offline by the Pinniped folks, we should use the `CredentialIssuer` in our docs (instead of referring users to svc and secrets directly).

### Benefits

Docs will no longer be coupled to a given `svc` and `secret` names and they will be more Pinniped-compliant.

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

N/A

### DIFF
--- a/docs/user/using-an-OIDC-provider-with-pinniped.md
+++ b/docs/user/using-an-OIDC-provider-with-pinniped.md
@@ -86,15 +86,24 @@ But, what if this `kube-controller-manager` is not a normal pod on a schedulable
 
 In managed clusters, such as AKS, Pinniped cannot read the cluster's certificate and key. In this case, Pinniped will have a fallback mechanism: the [impersonation proxy](https://pinniped.dev/docs/background/architecture/). It simply creates a LoadBalancer service that proxies the actual Kubernetes API. For this reason, when using Kubeapps in managed clusters using Pinniped, you'll need to use the Impersonation Proxy URL (and CA certificate) instead of the usual k8s API server URL.
 
-Assuming you have successfully [installed Pinniped](#installing-pinniped) and configured the [JWTAuthenticator](#configure-pinniped-to-trust-your-oidc-identity-provider), you have to retrieve the Impersonation Proxy IP by executing this command:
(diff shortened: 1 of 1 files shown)


docs/user/using-an-OIDC-provider-with-pinniped.md
```diff
@@ -86,15 +86,24 @@ But, what if this `kube-controller-manager` is not a normal pod on a schedulable
...
+
+```bash
+kubectl get credentialissuer -o json | jq -c '.items[].status.strategies[]  | select( .type | contains("ImpersonationProxy")).frontend.impersonationProxyInfo.endpoint'
 ```
 
-Then, retrieve the Impersonation Proxy CA cert by executing:
+Retrieving the Impersonation Proxy CA:
 
+```bash
+kubectl get credentialissuer -o json | jq -c '.items[].status.strategies[]  | select( .type | contains("ImpersonationProxy")).frontend.impersonationProxyInfo.certificateAuthorityData'
```
absoludity, line 100: Fine as is, just wondering whether it'd be better to just have one command to just query to return all the strategies so that the user has more context if something isn't as expected (like not enabled), and have them copy both fields from that output? See what you think, but fine as is.

absoludity (approved):

docs/user/using-an-OIDC-provider-with-pinniped.md
antgamdia: Mmm, I think users will just have to delete parts of  the jq query. The whole output will be:

 `k get credentialissuer -o json | jq -c '.items[].status.strategies[]  | select( .type | contains("ImpersonationProxy"))`
```json
{
    "frontend": {
        "impersonationProxyInfo": {
            "certificateAuthorityData": "redacted",
            "endpoint": "https://redacted"
        },
        "type": "ImpersonationProxy"
    },
    "lastUpdateTime": "2021-05-05T22:19:45Z",
    "message": "impersonation proxy is ready to accept client connections",
    "reason": "Listening",
    "status": "Success",
    "type": "ImpersonationProxy"
}
``` 

For users who just want to know what should they copy, I think it's ok as is. And, for those interested in the whole output, I'm sure they will figure it out.

Issue #2773, opened 2021-05-10
Add Kubeapps APIs server POC to development chart behind feature flag


absoludity:
### Description:

The current POC for a new API server that can support different package formats as well as other plugin functionality has been up in a PR for a while (#2609) but I'd like to enable working on this together as we push the POC forward, reviewing the code piecemeal as it lands in reasonable chunks, adding it to CI etc.

We can add the related deployment with relevant image building infra to the development chart behind a feature flag, enabling building and testing of the development images in our CI for use in dev envs, easy integration with existing code as well as CI to ensure we're not breaking existing functionality as we go.

Note that we will want to wait until we've synced the [new kubeapps chart standardisation](https://github.com/bitnami/charts/pull/6301) to our local repo given the churn there. Also, given the new way we'll be proposing chart changes upstream to the Bitnami/charts repository, we do not need to even include the featurure flag in the released chart until a later date when we want people to test it more widely.

- [x] Get existing code reviewed and landed piecemeal
- [x] Add Dockerfile for building and update CI to include building the image with default plugins included on the image initially
- [x] Add feature flag together with appropriate deployment and service to the development chart and verify

PR #2805, opened 2021-05-14
Load plugins for apis server.

absoludity:
### Description of the change

Follows #2789 adding two plugins with unimplemented servers for the `packages.GetAvailablePackages` call together with the functionality from the POC to load the plugins.

The main difference from the POC is the file layout, with the proto and generated files separated from the actual plugin implementation (I'll update the README with this info).

Next up:
* Implement the `GetAvailablePlugins` call now that plugins are loaded
* Add Dockerfile and update the dev chart to deploy the service behind a feature flag
* Implement and test `GetAvailablePackages` for kapp-controller.

### Applicable issues

  - ref #2773


cmd/kubeapps-apis/README.md
```diff
@@ -16,6 +16,17 @@ Together, this enables the best of both worlds: a well known and used Interface
...
 Finally, we've also chosen to use the [buf](https://buf.build/) tool for generating the code from the proto files. In the past we've used `protoc` (proto buffer compiler) and its extensions directly, but `buf` allows you to specify a simple yaml config instead, and also provides a `lint` command to ensure that your choice of API structure follows best practise, as well as ensuring you're aware when you break backwards compatability.
 
+## Plug-able
+
+The kubeapps-apis service uses the standard [go plugin package](https://golang.org/pkg/plugin/) to be able to load api plugins at run-time.
+
+Each plugin consists of 2 source files (and some generated files):
+
+* A `.proto` file defining the service that uses the messages defined in relevant part of kubeappsapis.core,
+* A `main.go` that compiles to an .so file for that plugin. This `main.go` has two public functions: one to register the plugin with a GRPC server and one to register the plugin for the http handler as well as the implementation for the server. This may be split into further modules as the complexity of the plugin grows.
```
antgamdia, line 26: Perhaps obvious, but I'd say the `.go` files are is under `./plugins/<pluginName>` whereas the `.proto` are under `./proto/kubeappsapis/plugins/<pluginName>`

cmd/kubeapps-apis/server/server.go
```diff
@@ -88,3 +109,88 @@ type gwHandlerArgs struct {
...
+		}
+
+		log.Infof("Successfully registered plugin %q", pluginPath)
+	}
+	return nil
+}
+
+// registerGRPC finds and calls the required function for registering the plugin for the GRPC server.
+func registerGRPC(p *plugin.Plugin, pluginPath string, registrar grpc.ServiceRegistrar) error {
+	grpcRegFn, err := p.Lookup(grpcRegisterFunction)
```
antgamdia, line 136: Interesting usage of reflection here to be able to register the plugging in both grpc and HTTP server.

cmd/kubeapps-apis/server/server.go
```diff
@@ -88,3 +109,88 @@ type gwHandlerArgs struct {
...
+		var dummyFn gatewayRegisterFunctionType = func(context.Context, *runtime.ServeMux, string, []grpc.DialOption) error { return nil }
+		return fmt.Errorf("unable to use %q in plugin %q due to mismatched signature.\nwant: %T\ngot: %T", gatewayRegisterFunction, pluginPath, dummyFn, gwRegFn)
+	}
+	return gwfn(gwArgs.ctx, gwArgs.mux, gwArgs.addr, gwArgs.dialOptions)
+}
+
+// listSOFiles returns the absolute paths of all .so files found in any of the provided plugin directories.
+//
+// pluginDirs can be relative to the current directory or absolute.
+func listSOFiles(fsys fs.FS, pluginDirs []string) ([]string, error) {
```
antgamdia, line 170: I guess during these early dev stages no, but do you foresee any security concern loading arbitrary so files? Should we allowlisting the plugins in the values.yaml instead?

cmd/kubeapps-apis/server/server_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+			if got, want := got, tc.pluginFilenames; !cmp.Equal(want, got) {
+				t.Errorf("mismatch (-want +got):\n%s", cmp.Diff(want, got))
+			}
+
+		})
+	}
+}
+
+func createTestFS(t *testing.T, filenames []string) fstest.MapFS {
```
antgamdia, line 79: Is this the in-memory file system, that new test feature in Go 1.16, isn't it?

antgamdia (approved):
Interesting dynamic load of the plugins. Perhaps at some point we may think of declaring the plugins in the values-yaml instead of relying on an arbitrary list of so files.
Thank you for this clear PR, very understandable and targeted!

cmd/kubeapps-apis/README.md
absoludity: Done, thanks.

cmd/kubeapps-apis/server/server.go
absoludity: This is how the `plugin` package recommends calling the functions. The only non-standard bit is the `dummyFn` to try to get a useful error message.

cmd/kubeapps-apis/server/server.go
absoludity: I don't think so, in that they won't load and we'll error, unless they have the required functions for the plugin, in which case, it is a plugin in the directory specified by the caller, whether it's evil or not. Not sure, need to think more.

cmd/kubeapps-apis/server/server_test.go
absoludity: Yeah, I was looking to test this and found it. In the past I've had to create actual files in a /tmp dir for tests, so this was nice to try out :)

absoludity (commented):
Thanks Antonio!

PR #2814, opened 2021-05-17
Update dockerfile frontend to 1.2, buildkit support graduated from experimental

absoludity:
### Description of the change

Noticed while adding a Dockerfile that we were still referring to the experimental (unversioned) docker frontend even though the buildkit support (for which it was added) graduated into 1.2.

### Benefits

Stay on a stable API.

### Possible drawbacks

Let's see what CI says.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,4 @@
 dashboard/node_modules
 .git
 docs
-pinniped-proxy/target
+cmd/pinniped-proxy/target
--- a/cmd/apprepository-controller/Dockerfile
+++ b/cmd/apprepository-controller/Dockerfile
@@ -1,4 +1,4 @@
-# syntax = docker/dockerfile:experimental
+# syntax = docker/dockerfile:1.2
 
 FROM bitnami/golang:1.16.3 as builder
 WORKDIR /go/src/github.com/kubeapps/kubeapps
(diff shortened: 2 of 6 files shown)


antgamdia (approved):
Interesting! Thanks for upgrading this!
Just a couple of doubts:

- I guess this is for dockerfiles using a builder image. In that case, should we add the `# syntax...` also to the pinniped-proxy image?
- Since it [uses semver](https://docs.docker.com/engine/reference/builder/#official-releases), do we really want to stick to the minor version, or should we replace it by just `docker/dockerfile:1`?

absoludity:
> Interesting! Thanks for upgrading this!
> Just a couple of doubts:
> 
>     * I guess this is for dockerfiles using a builder image. In that case, should we add the `# syntax...` also to the pinniped-proxy image?

The only reason it was added for the go images was to save re-downloading all the data when building. That said, it'd be useful for the pinniped-proxy image for the same reason, for sure.

> 
>     * Since it [uses semver](https://docs.docker.com/engine/reference/builder/#official-releases), do we really want to stick to the minor version, or should we replace it by just `docker/dockerfile:1`?

It explicitly needs to be >= 1.2 . As long as it'll grab the latest by default, then yep, that should be fine.

absoludity:
FWIW, that simplified the pinniped-proxy Dockerfile considerably, and improved the speed of the dev setup.

PR #2819, opened 2021-05-17
Add dev-only deployment of kubeapps-apis

absoludity:
### Description of the change

Follows #2805. Adds the chart templating to deploy the kubeapps-apis service behind a feature flag.

Note that the local dev env is currently broken due to the upstream Kubeapps chart changes which we synced. I'll deal with that in a separate PR, but with this current change I'm able to deploy Kubeapps, port-forward the service, then curl it with:

```
curl http://localhost:8080/core/plugins/v1alpha1/configured-plugins

{"plugins":["foobar.package.v1"]}
```

### Benefits

Easier testing and further development of the kubeapps-apis service.

### Possible drawbacks

None that I see.

### Applicable issues

  - Ref: #2773

### Additional information

I've not updated the main values.yaml with all the options so as to avoid the auto-documentation of those in the upstream published chart when we merge. Instead, I've put them in a `docs/developer/manifests/values.dev.yaml` that I'll include in the dev env (once I fix the dev env in the separate PR).

### DIFF
--- a/chart/kubeapps/templates/kubeappsapis/rbac.yaml
+++ b/chart/kubeapps/templates/kubeappsapis/rbac.yaml
@@ -1,0 +1,47 @@
+{{- if .Values.featureFlags.kubeappsAPIsServer }}
+{{- if .Values.rbac.create -}}
+# Dev-only RBAC for experimental APIs server until user auth added.
+apiVersion: {{ include "common.capabilities.rbac.apiVersion" . }}
+kind: ClusterRole
+metadata:
+  name: "kubeapps:controller:kubeapps-apis-dev-{{ .Release.Namespace }}"
+  namespace: {{ .Release.Namespace | quote }}
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    app.kubernetes.io/component: kubeappsapis
+    {{- if .Values.commonLabels }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" . ) | nindent 4 }}
+    {{- end }}
+  {{- if .Values.commonAnnotations }}
+  annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+  {{- end }}
+rules:
+  # TODO: Remove in favour of user's own creds used for request.
+  - apiGroups:
+      - ""
+      - "package.carvel.dev"
+    resources: ['*']
+    verbs: ['*']
+---
+apiVersion: {{ include "common.capabilities.rbac.apiVersion" . }}
+kind: ClusterRoleBinding
+metadata:
+  name: "kubeapps:controller:kubeapps-apis-dev-{{ .Release.Namespace }}"
+  namespace: {{ .Release.Namespace | quote }}
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    app.kubernetes.io/component: kubeappsapis
+    {{- if .Values.commonLabels }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" . ) | nindent 4 }}
+    {{- end }}
+  {{- if .Values.commonAnnotations }}
+  annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+  {{- end }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: "kubeapps:controller:kubeapps-apis-dev-{{ .Release.Namespace }}"
+subjects:
+  - kind: ServiceAccount
+    name: {{ template "kubeapps.kubeappsapis.fullname" . }}
+    namespace: {{ .Release.Namespace }}
+{{- end }}
+{{- end }}
--- a/chart/kubeapps/templates/kubeappsapis/service.yaml
+++ b/chart/kubeapps/templates/kubeappsapis/service.yaml
@@ -1,0 +1,30 @@
+{{- if .Values.featureFlags.kubeappsAPIsServer }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ template "kubeapps.kubeappsapis.fullname" . }}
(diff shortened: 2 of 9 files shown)


docs/developer/manifests/values.dev.yaml
```diff
@@ -0,0 +1,197 @@
...
+  readinessProbe:
+    enabled: false
+    httpGet:
+      path: /ready
+      port: 50051
+    initialDelaySeconds: 0
+    periodSeconds: 10
+    timeoutSeconds: 5
+    failureThreshold: 6
+    successThreshold: 1
```
antgamdia, line 126: We are not defining a `/ready` nor a `/live, aren't we?

docs/developer/manifests/values.dev.yaml
```diff
@@ -0,0 +1,197 @@
+## The extra value fields below are currently not included in the chart so as to not document
+## experimental functionality until we're ready to include it in the published chart.
+
```
antgamdia, line 3: Great, good idea! It's too soon to include it as part of the mail values.yaml

cmd/kubeapps-apis/Dockerfile
```diff
@@ -0,0 +1,36 @@
+# syntax = docker/dockerfile:experimental
```
antgamdia, line 1:
```suggestion
# syntax = docker/dockerfile:1
```

cmd/kubeapps-apis/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+    -o /kapp-controller-packages-v1alpha1-plugin.so -buildmode=plugin \
+    ./cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/main.go
+RUN --mount=type=cache,target=/go/pkg/mod \
+    --mount=type=cache,target=/root/.cache/go-build \
+    go build \
+    -o /helm-operator-packages-v1alpha1-plugin.so -buildmode=plugin \
+    ./cmd/kubeapps-apis/plugins/helm_operator/packages/v1alpha1/main.go
+
+# Note: unlike the other docker images for go, we cannot use scratch as the plugins
+# are loaded using the dynamic linker.
```
antgamdia, line 26: Uff, let's see how we will handle it during the official Bitnami image generation

chart/kubeapps/templates/kubeappsapis/serviceaccount.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if .Values.featureFlags.kubeappsAPIsServer }}
+apiVersion: v1
+kind: ServiceAccount
```
antgamdia, line 3: Why do we need a SA at the moment for this service?

chart/kubeapps/templates/kubeappsapis/service.yaml
```diff
@@ -0,0 +1,30 @@
...
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+    {{- end }}
+  {{- end }}
+spec:
+  type: ClusterIP
+  ports:
+    - port: {{ .Values.kubeappsapis.service.port }}
+      targetPort: http
+      protocol: TCP
+      name: http
```
antgamdia, line 27: What about the gRPC port? Are we just serving the HTTP API here?

chart/kubeapps/templates/kubeappsapis/rbac.yaml
```diff
@@ -0,0 +1,47 @@
+{{- if .Values.featureFlags.kubeappsAPIsServer }}
+{{- if .Values.rbac.create -}}
+# Dev-only RBAC for experimental APIs server until user auth added.
```
antgamdia, line 3: Ahh, ok, that answer my question on the serviceaccount, I guess

antgamdia (approved):
Awesome! Happy to see we can start landing this component soon :)

chart/kubeapps/templates/kubeappsapis/service.yaml
absoludity: They're both on the same port. Wasn't sure whether to call it something else, like `grpc_http`. What do you think?

chart/kubeapps/templates/kubeappsapis/serviceaccount.yaml
absoludity: Yep, as you saw, for the dev-only RBAC until I add the user auth.

cmd/kubeapps-apis/Dockerfile
absoludity: Why is that an issue? I don't remember, but is that what caused the difficulties with the pinniped-proxy image? It *shouldn't* be an issue to use a standard base image rather than scratch - keen to understand (maybe not here).

docs/developer/manifests/values.dev.yaml
absoludity: Nope, which is why I've set both `readinessProbe.enabled` and `livenessProbe.enabled` to false for now.

absoludity (commented):
Thanks Antonio!

docs/developer/manifests/values.dev.yaml
antgamdia: Hah - :facepalm: I didn't notice the `enabled: false`

cmd/kubeapps-apis/Dockerfile
antgamdia: Thinking twice... they tend to use `scratch` images but the problem with the pinniped-proxy was caused to the openssl libs (we would have to maintain ourselves these deps, and we don't want to be in charge of updating these deps manually). In this case, as we manage our own deps, it shouldn't be really an issue. I'm not 100% sure, but it is something to recall when creating the production image.

chart/kubeapps/templates/kubeappsapis/service.yaml
antgamdia: Oops, that's true... that's the key of the multiplexing indeed :P
But, yep, I'd rather change it to `grpc_http` to avoid future misunderstandings.

Issue #2827, opened 2021-05-17
logout does not work in oidc/sso setup


dlaloue-vmware:
### Description:

I have a setup with SSO/oidc using the default auth proxy and pinniped proxy/
the logout from the dashboard does not log the user out.

the logout does redirect to the login screen, but clicking the login with oidc provider buttons logs you back in automatically.
this could be reproduces even after closing the browser and opening a new one.

### Steps to reproduce the issue:

setup kubeapps in SSO/oidc configuration, using default auth proxy and pinniped proxy.
open browser and navigate to login screen.
click login with oidc provider, enter username/password. (user is no wlogged in)
click the logout button (users is redirected to login screen)
click on login with oidc provider button again
     ->   users is automatically logged in from previous session.

click logout
close browser
open new browser
click login with oidc provider
     ->   users is automatically logged in from previous session.

(note: i am using firefox)

### Describe the results you received:

users was logged in automatically

### Describe the results you expected:

login screen from oidc provider should have been presented to let the user enter username/password again

absoludity:
I just tried this in the dev env (standard OIDC with pinniped setup) and it is apparently working there:
![login-out](https://user-images.githubusercontent.com/497518/118600076-d1d0f380-b7f3-11eb-878e-4c0f2c86153a.gif)

I'm keen to see your oauth2-proxy configuration... you can copy-n-paste it (sanitized) from the actual frontend deployment yaml. I suspect something is awry, but keen to find out more.

As mentioned on slack, it is actually oauth2-proxy which handles the logout URL (and oauth2-proxy simply deletes the cookie with the request - the cookie isn't even visible to the Kubeapps code since it's http-only). Let's find out :)

dlaloue-vmware:
![logout](https://user-images.githubusercontent.com/74025858/118698409-36288e80-b7c5-11eb-9698-0cffe8570bce.gif)

dlaloue-vmware:
Here is the values.yaml i used (i have removed the pinniped configuration)

frontend:
  service:
    type: LoadBalancer

authProxy:
  enabled: true
  skipKubeappsLoginPage: false
  provider: oidc
  clientID: kubeapps
  clientSecret:   <redacted>
  cookieSecret: Y29va2llLXNlY3JldC0xNg==
  emailDomain: "*"
  additionalFlags:
   - --ssl-insecure-skip-verify
   - --cookie-secure=false
   - --scope=openid email groups
[...]

Issue #2826, opened 2021-05-17
chart install fails when pinnipedProxy is enabled


dlaloue-vmware:
### Description:

helm install fails when pinnipedProxy is enabled, failing on missing pinnipedProxy.service.port value

### Steps to reproduce the issue:

configure the values file to enable pinnipedProxy.
do helm install 

### Describe the results you received:

helm install fails because of missing pinnipedProxy.service.port value

### Describe the results you expected:

successful installation

### Additional information you deem important (e.g. issue happens only occasionally):

3 files are using pinnipedPRoxy.service.port, maybe instead of pinnipedProxy.containerPort:
  frontend configmap.yaml
  frontend service.yaml
  kubeops deployment.yaml

absoludity:
Thanks @dlaloue-vmware . Yes, as per our slack chat, we recently had a major upgrade of the upstream bitnami/kubeapps chart which has broken the pinniped integration (we test oidc on CI but not yet with pinniped). As this is a bug in the upstream chart, we'll need to do the fix there.

For context: the upstream chart standardization was done in this PR: https://github.com/bitnami/charts/pull/6301

dlaloue-vmware:
this issue is dependent on the following bitnami/charts issue: https://github.com/bitnami/charts/issues/6398

antgamdia:
Closing as the changes are already in the production chart

PR #2828, opened 2021-05-18
Fix dev environment after bitnami/kubeapps chart update.

absoludity:
### Description of the change

The recent update of the bitnami/kubeapps chart, which we synced, had a backwards incompatible change with the dev environment.

Specifically, the `extraTls` option now needs to be used to specify an existing TLS secret for ingress.

While there, I updated the target to ensure that it can be run multiple times to continue applying local changes (ie. ensured we create a secret for the postgresql secret also).

UPDATE: I ended up needing a local env with pinniped to investigate #2827 , so fixed the similar issue in the pinniped make targets. I also cleaned them up a bit (no more deleting/creating secrets, instead using the created secret), removed one yaml file which was near identical to the non-pinniped one (so just re-used that and set the one option to enable pinniped). Also fixed the upstream chart issue here for #2826, I'll create a separate PR upstream for that.

### Benefits

Easy dev with OIDC (on linux) works again.

### Possible drawbacks

None.


chart/kubeapps/Chart.yaml
```diff
@@ -25,4 +25,4 @@ maintainers:
 name: kubeapps
 sources:
   - https://github.com/kubeapps/kubeapps
-version: 7.0.0
+version: 7.0.1-dev1
```
absoludity, line 28: @antgamdia I bumped the chart here, so it's a pre-release of 7.0.1 (ie, < 7.0.1 according to semver), but the important thing is that this shouldn't trigger a PR to the upstream chart either right? (as in, the comparison for creating the upstream chart  would need to compare with the current release + 0.0.1 (ie. check if 7.0.1-dev1 > 7.0.1, which it's not... rather than checking if 7.0.1-dev1 > 7.0.0, which it is). Does that make sense? I'll just revert the chart bump here if you don't have time to deal with that for now.

antgamdia: > this shouldn't trigger a PR to the upstream chart either right?
No, our local changes will NOT trigger any update in the bitnami/charts repo, unless we are releasing (i.e., creating a tag)

> (as in, the comparison for creating the upstream chart would need to compare with the current release + 0.0.1 (ie. check if 7.0.1-dev1 > 7.0.1, which it's not... rather than checking if 7.0.1-dev1 > 7.0.0, which it is). 

What we are doing on every push a master is to check:

```bash 
# before bitnami accepts the PR, the remote version will be 7.0.0
semverCompare=$(semver compare "7.0.1-dev1" "7.0.0") # 1
  # Do nothing, our chart is more updated than the remote one, that is OK
...

# after bitnami accepts the PR, the remote version will be 7.0.1
semverCompare=$(semver compare "7.0.1-dev1" "7.0.1") # -1
# The current version is less than the chart external version, then retrieve the changes and send an internal PR with them
# But, given that some changes are already there, we will have a 7.0.1 version with almost no changes.
# In fact, if it is the same set of changes, we only get a chart version update.
```

> Does that make sense? 

Yes, and no. I mean: it is a bit odd sending duplicated PRs in both repos. If we happen to add any other change, we will have a, let's say, version 7.0.1 which does not match the upstream one. In that case, we would have to manually modify the PR and set it to 7.0.2-dev.

antgamdia (approved):

chart/kubeapps/Chart.yaml
absoludity: Perfect. For some reason I thought we were sending a PR the other way if we intentionally made the chart version greater... but what you've described is much saner: we tag a release to trigger a PR to bitnami/charts, while we pull any new chart changes down. Great, thanks.

PR #2838, opened 2021-05-19
Add implementation for kapp-controller GetAvailablePackages.

absoludity:
### Description of the change

Adds the implementation for the kapp_controller plugin's GetAvailablePackages endpoint.

While port-forwarding to the kubeapps-apis service with

```
kubectl -n kubeapps port-forward svc/kubeapps-internal-kubeappsapis 8080:8080
```

http endpoint responds with:
```
curl -s http://localhost:8080/kapp_controller/packages/v1alpha1 | jq .
{
  "packages": [
    {
      "name": "app1.example.com",
      "version": "0.11.3",
      "iconUrl": "",
      "repository": null
    },
    {
      "name": "app2.example.com",
      "version": "3.11.3",
      "iconUrl": "",
      "repository": null
    },
```

while with grpcurl:

```
grpcurl -plaintext localhost:8080 kubeappsapis.plugins.kapp_controller.packages.v1alpha1.PackagesService.GetAvailablePackages
{
  "packages": [
    {
      "name": "app1.example.com",
      "version": "0.11.3"
    },
    {
      "name": "app2.example.com",
      "version": "3.11.3"
    },
    ...
```

**Note**: I still need to check the fake dynamic client to see the form unit-tests will take. I'll those in a subsequent PR.

### Benefits

Demonstrate using the dynamic client to query the plugin backend.

### Applicable issues

  - Ref #2775

PR #2840, opened 2021-05-20
Add GetPackageRepositories implementation for carvel and helm-fluxv2

### LINKED ISSUE #2775: Implement kapp-controller/flux Kubeapps APIs for GetAvailablePackages
### Description:

The [initial POC](#2609) has stubbed responses with fake data for the one basic API response of the kapp-controller plugin.

This should be updated for the kapp-controller plugin to enable getting the actual available packages for install, getting installed packages (both in a namespace and across namespaces where RBAC allows), and if time, installing a package. I would like to also do the same functionality for the helm plugin for comparison, but that may need to wait.

Note that currently the available packages for kapp-controller is a cluster-wide request (ie., unlike our existing Helm support, there is no concept of packages available in a particular namespace), so we'll need to ensure an appropriate response (400) if called with a namespace.

- [X] Implement demo of GetAvailablePackages for kapp-controller plugin
- [X] " GetInstalledPackages and/or GetPackageRepositories for kapp-controller plugin
- [X] " Implement one of the above for another plugin to demo usage (used GetPackageRepositories for flux).
- [ ] ?Implement InstallPackage for kapp-controller plugin

absoludity:
### Description of the change

This was just to rename the `helm_operator` plugin to `helm_fluxv2`, but while there, I've added a demo implementation for the `GetPackageRepositories` method for both.

From the [updated README](https://github.com/kubeapps/kubeapps/blob/add-kubeapps-apis-service-6/cmd/kubeapps-apis/README.md#trying-it-out), with the appropriate resources on your cluster you can now:

```bash
$ curl -s http://localhost:8080/plugins/helm_fluxv2/packages/v1alpha1/packagerepositories | jq .
{
  "repositories": [
    {
      "name": "bitnami",
      "namespace": "flux-system",
    }
  ]
}

$ curl -s http://localhost:8080/plugins/kapp_controller/packages/v1alpha1/packagerepositories | jq .
{
  "repositories": [
    {
      "namespace": "",
    }
  ]
}
```

### Benefits

Provides an example of an API call implemented by two plugins (needed for #2775). Renames helm flux plugin more appropriately (@gfichtenholt found helm_operator is from a pre fluxv2 version). Provides an example so @gfichtenholt can think about GetAvailablePackages for helm-fluxv2 (though as I mentioned on slack, this may be more work than we'd thought, depending whether fluxv2 actually exposes something analogous - afaict so far, it only creates a `HelmChart` object when a user installs from a repo, happy to jump in there to help as needed).

### Possible drawbacks

N/A

### Applicable issues

  - fixes #2775 

### Additional information

Next up for me:
* Start using this to help with the design of the API requests across plugins
* investigating the fake dynamic client so we can write useful unit tests
* potentially try a generic API call above the plugins, such as `/core/packages/v1alpha1/GetPackageRepositories` to show how we could return objects across plugins (chatted with @migmartri about this).
But I'll discuss the above as part of our iteration planning next week.


cmd/kubeapps-apis/plugins/helm_fluxv2/packages/v1alpha1/main.go
```diff
@@ -18,7 +18,7 @@ import (
 	"google.golang.org/grpc"
 
 	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
-	v1alpha1 "github.com/kubeapps/kubeapps/cmd/kubeapps-apis/gen/plugins/helm_operator/packages/v1alpha1"
+	v1alpha1 "github.com/kubeapps/kubeapps/cmd/kubeapps-apis/gen/plugins/helm_fluxv2/packages/v1alpha1"
```
gfichtenholt, line 21: I would probably just go with fluxv2, i.e. drop "helm". Based on their docs Flux supports more than just helm, they also support Git repositories with artifacts, as well as Amazon-style buckets with artifacts.

absoludity: True, I was thinking that we're only interested in the helm support, but we may want to support any repositories/packages. Note that we'd need to then include something like a resource type field (group/version/name) so if we're returning results including both HelmRepositories and GitRepositories (?), they can be recognised as such (I'll just rename in this PR). Let me do the rename...

absoludity: Done. :)

cmd/kubeapps-apis/README.md
```diff
@@ -62,16 +62,53 @@ grpcurl -plaintext localhost:50051 kubeappsapis.core.plugins.v1alpha1.PluginsSer
...
-The packages endpoints for the helm-operator and kapp-controller plugins are unimplemented:
+To test the packages endpoints for the fluxv2 or kapp_controller plugins, you will currently need to build the image from the kubeapps root directory with:
 
 ```bash
-curl http://localhost:50051/kapp_controller/packages/v1alpha1
-{"code":12, "message":"method GetAvailablePackages not implemented", "details":[]}
+IMAGE_TAG=dev1 make kubeapps/kubeapps-apis
+```
+
+and make that imageg available on your cluster somehow. If using kind, you can simply do:
```
gfichtenholt, line 71: spelling and/or wording?

gfichtenholt (approved):
I viewed and I am ok with the changes. I don't have enough expertise on grpc and protobuf to be able to comment constructively at this point, but I will get there someday :-)

gfichtenholt (commented):
sorry missed one question. In earlier version of the PR, I saw a change you made in deploy-dev.mk to deploy kapp-controller to a cluster. I don't see that anymore in this list of files, nor do I see a target to deploy flux v2 Is that deliberate?

absoludity:
> sorry missed one question. In earlier version of the PR, I saw a change you made in deploy-dev.mk to deploy kapp-controller to a cluster. I don't see that anymore in this list of files, nor do I see a target to deploy flux v2 Is that deliberate?

That was in the [previous PR](https://github.com/kubeapps/kubeapps/pull/2838/files#diff-ad67dcc4ae810d5216ae3805d8b531844dfc5e2e01d6eb6ef289275dda431871R53-R57), let me add a similar one with the flux command I used to install the flux CRDs etc., as it's also useful as documentation of what version etc. was being used.

PR #2852, opened 2021-05-21
Flux v2 GetAvailablePackages() implementation

gfichtenholt:
flux v2 implementation of GetAvailablePackages() API


cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
```diff
@@ -15,41 +15,40 @@ package main
...
+	helmrepo "k8s.io/helm/pkg/repo"
+	log "k8s.io/klog/v2"
 )
 
-// Server implements the kapp-controller packages v1alpha1 interface.
+const (
+	// see docs at https://fluxcd.io/docs/components/source/
+	fluxGroup            = "source.toolkit.fluxcd.io"
+	fluxVersion          = "v1beta1"
+	fluxHelmRepoResource = "helmrepositories"
```
absoludity, line 38: I've updated my PR at #2857 with consistent constants, thanks.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
```diff
@@ -81,3 +80,132 @@ func (s *Server) GetPackageRepositories(ctx context.Context, request *corev1.Get
...
+	repos, err := getHelmRepos(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	responsePackages := []*corev1.AvailablePackage{}
+	for _, repoUnstructured := range repos.Items {
+		name, found, err := unstructured.NestedString(repoUnstructured.Object, "metadata", "name")
+		if err != nil || !found {
+			log.Errorf("required field metadata.name not found on HelmRepository: %w:\n%v", err, repoUnstructured.Object)
```
absoludity, line 96: If this ever happened (not sure it's possible, given the field here is `metadata.name`), wouldn't we want an internal error response?

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
```diff
@@ -81,3 +80,132 @@ func (s *Server) GetPackageRepositories(ctx context.Context, request *corev1.Get
...
+							break
+						}
+					}
+				}
+			}
+		}
+
+		if !ready {
+			log.Infof("Skipping packages for repository [%s] because it has not reached 'Ready' state:n%v", name, repoUnstructured.Object)
+		} else {
```
absoludity, line 123: Tiny formatting-only point, but if you `continue` above when `!ready`, we can avoid an else and indentation here. No problem if you prefer it this way.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
```diff
@@ -81,3 +80,132 @@ func (s *Server) GetPackageRepositories(ctx context.Context, request *corev1.Get
...
+				responsePackages = append(responsePackages, repoPackages...)
+			}
+		}
+	}
+	return &corev1.GetAvailablePackagesResponse{
+		Packages: responsePackages,
+	}, nil
+}
+
+func getHelmRepos(ctx context.Context) (*unstructured.UnstructuredList, error) {
```
absoludity, line 145: I was wondering why you don't have this function returning the structured type (ie. `[]core.packages.v1alpha1.PackageRepository`) but of course, you actually need flux-specific data out of here from the status, which won't be in our `PackageRepository` message.

Perhaps it'll prove worthwhile having a `PackageRepository.localIndexYAML` if that becomes a pattern, let's see.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
```diff
@@ -81,3 +80,132 @@ func (s *Server) GetPackageRepositories(ctx context.Context, request *corev1.Get
...
+
+	index.SortEntries()
+	responsePackages := []*corev1.AvailablePackage{}
+	for _, entry := range index.Entries {
+		if entry[0].GetDeprecated() {
+			log.Infof("skipping deprecated chart: [%s]", entry[0].Name)
+			continue
+		}
+		pkg := &corev1.AvailablePackage{
+			Name:    entry[0].Name,
```
absoludity, line 205: `entry[0]` - assuming that's the latest version? Nice. We can always query for more versions when looking at the package detail, I assume.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
```diff
@@ -81,3 +80,132 @@ func (s *Server) GetPackageRepositories(ctx context.Context, request *corev1.Get
...
+			Version: entry[0].Version,
```
absoludity, line 206: You should also have access to the icon URL from the entry, I think?

absoludity (approved):
Nice work Greg! Take a look at the example tests I've added in #2857 , though you'll need a bit extra for the `index.yaml` (should be able to use the httptest for that). See whether you want to include the tests in this PR or add them as a subsequent PR. I don't mind either way.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
gfichtenholt: > Nice work Greg! Take a look at the example tests I've added in #2857 , though you'll need a bit extra for the `index.yaml` (should be able to use the httptest for that). See whether you want to include the tests in this PR or add them as a subsequent PR. I don't mind either way.

Thank you for kind words. Yes, subsequent PR. I will definitely be adding unit tests

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
gfichtenholt: i don't know. i thought i'd log the error and move on to next repo. Kind of mute either way, because it'll probably never happen

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
gfichtenholt: ok, will do that

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
gfichtenholt: sure

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
gfichtenholt: yeah, this is by no means the final iteration what this will look like, just a first take based on looking existing code

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
gfichtenholt: not by looking at the definition of IndexFile. Its there in the index.yaml, but don't see a field for in in IndexFile. Don't worry, I'll figure this out before too long, this is just a first take

Issue #2849, opened 2021-05-21
[kubeapps-apis] Add unit-tests for existing plugin functionality to easy adding further changes


absoludity:
### Description:

The existing demo functionality was added and works with fluxv2 and kapp-controller without unit-tests, but as we build on this, we want to be able to do so using TDD and so need decent examples of doing so with the dynamic client (which will in turn force writing more testable code).

### Additional information you deem important (e.g. issue happens only occasionally):

This is important as more people start to work on the codebase, ensuring further changes are tested and that it is easy to follow existing examples to create new tests before changing/adding functionality, increasing velocity.

absoludity:
Just adding all three of us as assignees since we'll all be adding unit tests as we go :)

gfichtenholt:
I think in addition to unit-tests we should also have a end-2-end tests, where we stand up a cluster deploy carvel/flux on it test out our apis against that. I can open a separate issue for that  if you think its best

antgamdia:
Yes, I agree. We already spin up a cluster as part of the usual CI process, as described in [1]. It's a good idea to continue working on improving our test suite. Let's file a new issue so that we can track the progress we make (eg. installing carvel and flux, etc).
However, I wouldn't start working/landing on those tests yet, since there are a few temporary things waiting to be defined.

Personally, currently, I'm working with a Postman collection where I track the (rest-ish) API calls
![image](https://user-images.githubusercontent.com/11535726/119469886-11459580-bd48-11eb-9a69-5f2f3b4cd93b.png)


[1] https://github.com/kubeapps/kubeapps/blob/master/docs/developer/end-to-end-tests.md#end-to-end-tests-in-the-project.

gfichtenholt:
added https://github.com/kubeapps/kubeapps/issues/2881

gfichtenholt:
I think ideally, everybody contributing to core api server and/or plug-ins, including me, should part-own it, but I can take it if no-one else will

PR #2857, opened 2021-05-21
Example of unit-testing with dynamic client.

absoludity:
### Description of the change

Establish a pattern for testing with the dynamic k8s client in plugin code.

### Benefits

We'll have a pattern to follow and improve on when adding further code.

### Possible drawbacks

N/A

### Applicable issues

  - Ref #2849

### DIFF
--- a/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_test.go
+++ b/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_test.go
@@ -53,7 +53,7 @@ func TestGetAvailablePackagesStatus(t *testing.T) {
 				return fake.NewSimpleDynamicClientWithCustomListKinds(
 					runtime.NewScheme(),
 					map[schema.GroupVersionResource]string{
-						{Group: "package.carvel.dev", Version: "v1alpha1", Resource: "packages"}: "PackageList",
+						{Group: packageGroup, Version: packageVersion, Resource: packagesResource}: "PackageList",
 					},
 					packageFromSpec(map[string]interface{}{
 						"version": "1.2.3",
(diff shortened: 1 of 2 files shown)


cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_test.go
```diff
@@ -0,0 +1,156 @@
...
+			s := Server{
+				clientGetter: func(context.Context) (dynamic.Interface, error) {
+					return fake.NewSimpleDynamicClientWithCustomListKinds(
+						runtime.NewScheme(),
+						map[schema.GroupVersionResource]string{
+							{Group: "package.carvel.dev", Version: "v1alpha1", Resource: "packages"}: "PackageList",
+						},
+						// TODO: Not clear why
+						// pkgs...
+						// won't work.
```
antgamdia, line 134: Mmm, that's true. It has something to do with the `DeepCopyObject` or sth similar.

I've found an example of that kind of testing in the k8s core (autoscaler), but they are using `runtime.Object` instead of `Unstructured`. 
https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller_test.go

absoludity: Thanks Antonio. Note that the issue here was that it worked fine with ` pkgs[0], pkgs[1]` manually added as the arguments, but not when using the `...` to [expand the slice multiple function arguments](https://golang.org/ref/spec#Passing_arguments_to_..._parameters), with `pkgs...`.

`runtime.Object` is the interface, which is implemented by `unstructured.Unstructured` which is why it works with the individual args, but I'm just not sure why it doesn't work when spreading the slice. Anyway, I could work around it by updating the `packagesFromSpecs` to return a slice of the interface type instead. Yay.

antgamdia (approved):
Awesome! Thanks a lot for defining the unit-testing for dynamics resources. So neat and easy to follow!

PR #2869, opened 2021-05-24
add icon url and repo ref to GetAvailablePackages() response

gfichtenholt:


cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
```diff
@@ -130,7 +131,16 @@ func (s *Server) GetAvailablePackages(ctx context.Context, request *corev1.GetAv
 		}
 
 		log.Infof("Found repository: [%s], index URL: [%s]", name, url)
-		repoPackages, err := readPackagesFromRepoIndex(url)
+		repoRef := corev1.AvailablePackage_PackageRepositoryReference{
+			Name: name,
+		}
+		// namespace is optional according to https://kubernetes.io/docs/concepts/overview/working-with-objects/kubernetes-objects/
```
absoludity, line 137: Namespace is optional when creating an object (as per the link), since you might do `kubectl -n foobar -f my-object.yaml` or you can specify the namespace in your object and apply it without the `-n foobar`, but in the case of the code here, we're getting an existing object. If the object is cluster-scoped, then it won't have a namespace, but otherwise I *think* it'd be very unexpected (ie. 500 internal) if we received a namespaced-scoped object without a namespace? Not 100%

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
```diff
@@ -168,20 +178,53 @@ func getHelmRepos(ctx context.Context) (*unstructured.UnstructuredList, error) {
...
 	response, err := http.Get(indexURL)
 	if err != nil {
 		return nil, err
 	}
 	defer response.Body.Close()
 
-	if response.StatusCode != 200 {
-		return nil, fmt.Errorf("received non 200 response code: [%d]", response.StatusCode)
+	if response.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("received non OK response code: [%d]", response.StatusCode)
```
absoludity, line 227: Note, grpc itself will transform any error into a grpc status code (which is then translated into an http status via grpc gateway), so we'll need to be using `status.Errorf` here, rather than `fmt.Errorf`, the former allowing you to specify a status code with the error which can be used for grpc and translated or http.

See https://github.com/kubeapps/kubeapps/blob/master/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server.go#L95 for an example.

Without doing this, the grpc status will be `codes.Unknown` which is translated to a 500.

More about the grpc status codes at https://grpc.github.io/grpc/core/md_doc_statuscodes.html
And their translation to http in grpc-gateway: https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/errors.go#L34

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
```diff
@@ -94,6 +94,7 @@ func (s *Server) GetAvailablePackages(ctx context.Context, request *corev1.GetAv
 		name, found, err := unstructured.NestedString(repoUnstructured.Object, "metadata", "name")
 		if err != nil || !found {
 			log.Errorf("required field metadata.name not found on HelmRepository: %w:\n%v", err, repoUnstructured.Object)
+			// just skip over to the next one
```
absoludity, line 97: Isn't an error here a programming error - in that we expect it not to be possible, but if it is, it's because of an incorrect assumption on our part (ie. than metadata.name should always exist) ? I'm always a bit wary of ignoring any error (from a functionality pov). But hey, it's a plugin and you're the author :P. As you say, it'll make no difference anyway is there will always be a name.

absoludity (approved):
It'd be great to start including tests as we make changes like this. See [some examples](https://github.com/kubeapps/kubeapps/blob/master/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_test.go).

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
gfichtenholt: ok will fix it

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
gfichtenholt: Not a fan of ignoring errors either. But I am not doing that here. Logging an error isn't the same as ignoring it.

absoludity: Yeah, that's what I meant by "from a functionality pov" - I realise you're logging the error. Fine either way.

Issue #2848, opened 2021-05-21
[kubeapps-apis] Demo core.v1alpha1.GetPackageRepositories


absoludity:
### Description:

The current demo functionality, such as `GetPackageRepositories` and `GetAvailablePackages`, are exposed per plugin, for example:

```
$ curl -s http://localhost:8080/plugins/helm_fluxv2/packages/v1alpha1/packagerepositories | jq .
{
  "repositories": [
    {
      "name": "bitnami",
      "namespace": "flux-system",
      "url": "https://charts.bitnami.com/bitnami"
    }
  ]
}

$ curl -s http://localhost:8080/plugins/kapp_controller/packages/v1alpha1/packagerepositories | jq .
{
  "repositories": [
    {
      "name": "repo-name.example.com",
      "namespace": "",
      "url": "foo.registry.example.com/repo-name/main@sha256:cecd9b51b1f29a773a5228fe04faec121c9fbd2969de55b0c3804269a1d57aa5"
    }
  ]
}
```

This allows clients to interact with specific plugins when needed, as we cannot assume that this will not be necessary in some circumstances, such as a plugin-specific UX using extra features of that format.

That said, we also want to ensure it's possible to provide a generic implementation of the packages API implemented in our core, that returns results from all plugins, such as:

```
$ curl -s http://localhost:8080/core/packages/v1alpha1/packagerepositories | jq .
{
  "repositories": [
    {
      "name": "bitnami",
      "namespace": "flux-system",
      "url": "https://charts.bitnami.com/bitnami",
      "plugin": "fluxv2"
    },
    {
      "name": "repo-name.example.com",
      "namespace": "",
      "url": "foo.registry.example.com/repo-name/main@sha256:cecd9b51b1f29a773a5228fe04faec121c9fbd2969de55b0c3804269a1d57aa5",
      "plugin": "kapp-controller"
    }
  ]
}
```

absoludity:
Marked size/M only because there's some uncertainty around calling the grpc endpoints of the plugins from core, but assuming that's straight-forward, I expect this to be 1-2 days.

antgamdia:
So, the idea here is: create a new `core` endpoint and, for each `GetConfiguredPlugins` invoke `GetPackageRepositories` and marshall the response back to the user, no?
I agree, when developing clients for this API, having this endpoint will simplify the UI logic. 
Thanks for filing the issue!

dlaloue-vmware:
[...]

PR #2880, opened 2021-05-25
Implement GetConfiguredPlugins

### LINKED ISSUE #2850: [kubeapps-apis] Add implementation of core.GetConfiguredPlugins
### Description:

Currently this is stubbed as it wasn't yet needed for demoing the plugin behaviour itself, but it will be required for #2848 (and eventually for clients using the apis server).

absoludity:
### Description of the change

This PR refactors the plugin loading so its all handled by the core.plugin.Server itself, which also enables storing the plugins as they're registered when the process starts.

As per the updated examples in the README, you can now:

```
curl -s http://localhost:50051/core/plugins/v1alpha1/configured-plugins | jq .
{
  "plugins": [
    {
      "name": "fluxv2.packages",
      "version": "v1alpha1"
    },
    {
      "name": "kapp_controller.packages",
      "version": "v1alpha1"
    }
  ]
}
```

or

```
grpcurl -plaintext localhost:50051 kubeappsapis.core.plugins.v1alpha1.PluginsService.GetConfiguredPlugins
{
  "plugins": [
    {
      "name": "fluxv2.packages",
      "version": "v1alpha1"
    },
    {
      "name": "kapp_controller.packages",
      "version": "v1alpha1"
    }
  ]
}

```

### Benefits

Enables moving on to try #2848

### Possible drawbacks

We may find we want to separate the plugin name as well, not sure yet.

### Applicable issues

  - fixes #2850

### Additional information

The second commit just fixes the initialization of the kapp-controller plugin so that it uses the `NewServer()` constructor (which sets the config getter).


cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/main.go
```diff
@@ -31,3 +32,11 @@ func RegisterWithGRPCServer(s grpc.ServiceRegistrar) {
 func RegisterHTTPHandlerFromEndpoint(ctx context.Context, mux *runtime.ServeMux, endpoint string, opts []grpc.DialOption) error {
 	return v1alpha1.RegisterPackagesServiceHandlerFromEndpoint(ctx, mux, endpoint, opts)
 }
+
+// GetPluginDetail returns a core.plugins.Plugin describing itself.
+func GetPluginDetail() *plugins.Plugin {
+	return &plugins.Plugin{
+		Name:    "fluxv2.packages",
+		Version: "v1alpha1",
```
antgamdia, line 40: What if a plugin exposes different versions?  Wiil we always load a very specific version or will let users invoke whichever version they want?

If the former, I'm ok with this Plugin signature, but, in case uses can use any version, perhaps `Version` should become `Versions: ["v1alpha1"]` instead, no?

antgamdia (approved):
Great, thanks for implementing this. I only had a minor doubt/thought about versioning the plugins, but after checking the test, I noticed we are registering tuples of name+version.

I don't know if it makes sense a use case where users want to specify which plugin version they want to use in the values.yaml. But right now, it's a good solution as is, imho.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/main.go
absoludity: > What if a plugin exposes different versions? Will we always load a very specific version or will let users invoke whichever version they want?

The latter - we'll load whatever plugins are sitting in the plugin directory, so that could be `fluxv2.packages.v1alpha1` and `fluxv2.packages.v1`, and we'll serve both (not that people need to do that, but it enables supporting both).

> 
> If the former, I'm ok with this Plugin signature, but, in case uses can use any version, perhaps `Version` should become `Versions: ["v1alpha1"]` instead, no?

I did originally have it structured like this (a plugin with a list of versions), but changed to the current format because we're going to want to include the plugin with each package (or repo) that we return. As in, if a client requests packages generically, each package in the response will need to identify the plugin, I believe (I'm playing with this in the next PR).


> 
> Edit: ah, I see, it seems we are registering tuples `<name, version>`, so this `GetPluginDetail()` is returning what is expected. Perhaps it's a bit misleading to use "plugin" to refer to a combination of plugin_name+plugin_version, but I cannot think of a better option right now.

Well each version is a separate plugin and serving independently. You may choose to only ever have one version of fluxv2.packages available, but you may also choose to have two versions available, both independent.

PR #2885, opened 2021-05-25
Add Kubeapps apis docs

### LINKED ISSUE #2859: [kubeapps-apis] coreServer and/or plugins should serve swagger
Exatly how to map gRPC calls to HTTP JSON API is pretty tricky
(described a bit here https://github.com/googleapis/googleapis/blob/master/google/api/http.proto). I think it would be very beneficial if our server and/or plugins would server openapi swagger on a well known endpoint so that can use standard tools like swagger.io and postman so people would not have to remember/know all the tricks to the translation. How to serve openapi swagger is described here https://github.com/grpc-ecosystem/grpc-gateway (it's a pretty long read)

antgamdia:
PR in progress, just for showing the current changes.

### Description of the change

We will have a way to automatically generate our API docs. After importing the OpenAPI file in any tool we will get a ready-to-use client, for instance:

image

### Benefits

We will have automatic API docs!

### Possible drawbacks

N/A, but once integrated as "official" docs, we should think about how to integrate it with the [existing Kubeapps API docs](https://github.com/kubeapps/kubeapps/blob/master/dashboard/public/openapi.yaml)

### Applicable issues

  - fixes #2859

### Additional information

~~I still need to work on defining the proper annotations, like: ~~ DONE

An example of how to use the OAS document to easily consume the API :

image


cmd/kubeapps-apis/buf.lock
```diff
@@ -4,6 +4,13 @@ deps:
     owner: beta
     repository: googleapis
     branch: main
-    commit: aced890b7e944f9ca174481ad810ca5f
-    digest: b1-guLHfQhGOhlEA-QaPyQLoT-ZVFGJkRkluWgkzPT0ma0=
-    create_time: 2021-03-04T14:52:47.094456Z
+    commit: 2e73676eef8642dfba4ed782b7c8d6fe
+    digest: b1-vB11w98W2vFtEP4Veknm56Pi6DU6MpOuocESiOzvbqw=
+    create_time: 2021-04-26T14:55:30.644663Z
```
antgamdia, line 9: I wasn't able to replicate your issue, @absoludity . I did type `buf beta mod updata` and then `buf generate` without any error :S

cmd/kubeapps-apis/buf.gen.yaml
```diff
@@ -12,3 +12,14 @@ plugins:
     out: gen
     opt:
       - module=github.com/kubeapps/kubeapps/cmd/kubeapps-apis/gen
+  # Consider migrating to openapiv3 once ready
+  # https://github.com/grpc-ecosystem/grpc-gateway/pull/2147
+  - name: openapiv2
+    out: docs
+    strategy: all
```
antgamdia, line 19: It was the key point for `allow_merge` to work...
For the record, the default value is  `directory`, so the merge logic, in each folder, was overwriting the file again and again.
Anyway, it was worthwhile as I have learned a bit more about buf :P

cmd/kubeapps-apis/buf.gen.yaml
```diff
@@ -12,3 +12,14 @@ plugins:
...
+    opt:
+      - allow_merge=true
+      - merge_file_name=kubeapps-apis
+      - allow_repeated_fields_in_body=true
+      # - generate_unbound_methods=true # have to re-check it
+      # - json_names_for_fields=false # have to re-check it
```
antgamdia, line 25: I have to see the impact of these options. I have screened the ones that other people are using in other github projects, but I have to re-check when playing around with the generate API docs

cmd/kubeapps-apis/buf.gen.yaml
absoludity: Nice - so we can combine/merge the schema docs without needing the (npm-based) swagger-combine? Excellent if so!

cmd/kubeapps-apis/buf.lock
absoludity: Excellent, I'll recheck locally once this lands and update the issue I created.

cmd/kubeapps-apis/docs/kubeapps-apis.swagger.json
```diff
@@ -0,0 +1,318 @@
+{
+  "swagger": "2.0",
+  "info": {
+    "title": "kubeappsapis/core/packages/v1alpha1/packages.proto",
```
absoludity, line 4: Looks like the info defines this as core.packages.v1alpha1 only, whereas its describing others too (like `core.plugins.v1alpha1` and `plugins.fluxv2.packages.v1alpha1` etc.)?

absoludity (approved):
Excellent, very happy to see a single doc without swagger-combine :) +1'ing now just in case you only need minor changes and I'm not around.

cmd/kubeapps-apis/proto/kubeappsapis/plugins/kapp_controller/packages/v1alpha1/kapp_controller.proto
```diff
@@ -4,15 +4,30 @@ option go_package = "github.com/kubeapps/kubeapps/cmd/kubeapps-apis/gen/plugins/
 
 import "google/api/annotations.proto";
 import "kubeappsapis/core/packages/v1alpha1/packages.proto";
+import "protoc-gen-openapiv2/options/annotations.proto";
 
 service PackagesService {
```
antgamdia, line 9: Question here, @absoludity : is it possible to safely change this service name to something like `kappController_PackagesService` ? 

The generator is not able to deal with duplicated service names and we have to manually disambiguate the operation_id

antgamdia:
Initial version done! I have a minor question (as a comment), just to avoid using things like this to disambiguate names.

```proto
 option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
      operation_id: "xxxxxxx";
...
``` 

Apart from that, I've tried to stick to the comments as much as possible (instead of using `options`) as it is much clearer IMO). The structure, therefore is:

```proto
// Title of the message
//
// Description of the message
message xxxxxxxxxx {
  option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_schema) = {
    example: '{ a json example of the object }'
  };
  // Description of the message field
  XXXX xxx = 1;
}
```

cmd/kubeapps-apis/proto/kubeappsapis/plugins/kapp_controller/packages/v1alpha1/kapp_controller.proto
absoludity: Yes, I think KappControllerPackagesService should be fine. The only restriction `buf lint` puts on is that it needs to end in `Service`, from memory. Feel free to update to that (and Fluxv2PackagesService) and see if that fixes it before landing. Check buf lint too.

absoludity (approved):
Great!

absoludity:
Sorry, probably conflicts (and two extra doc comments) to update when I landed my PR.

PR #2888, opened 2021-05-26
Add plugin field to core package and repository.

absoludity:
### Description of the change

Adds an optional plugin field to the core PackageRepository and AvailablePackage messages in preparation for an aggregated API (#2848).

Updated the gatway config so that plugin (or empty fields in general) are not emitted:

```
curl -s http://localhost:8080/plugins/kapp_controller/packages/v1alpha1/packagerepositories | jq .
{
  "repositories": [
    {
    }
  ]
}
```

This will enable the inclusion of the plugin when called via an aggregate API so repos/packages can be handled appropriately.

While there, I took the opportunity to add tests for the kapp controller GetPackageRepositories and extend the functionality slightly #2849

### Benefits

Prep for #2848 . More test coverage/examples.

### Possible drawbacks

N/:

### Applicable issues

  - Ref #2848, #2849


cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/main.go
```diff
@@ -35,8 +46,5 @@ func RegisterHTTPHandlerFromEndpoint(ctx context.Context, mux *runtime.ServeMux,
 
 // GetPluginDetail returns a core.plugins.Plugin describing itself.
 func GetPluginDetail() *plugins.Plugin {
-	return &plugins.Plugin{
-		Name:    "kapp_controller.packages",
-		Version: "v1alpha1",
-	}
+	return &pluginDetail
```
absoludity, line 49: This change really belongs in the next PR where I'll need the plugin detail for each item when returning package repositories from multiple plugins.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server.go
```diff
@@ -85,15 +85,25 @@ func NewServer() *Server {
 	}
 }
 
-// GetAvailablePackages streams the available packages based on the request.
-func (s *Server) GetAvailablePackages(ctx context.Context, request *corev1.GetAvailablePackagesRequest) (*corev1.GetAvailablePackagesResponse, error) {
+// getClient ensures a client getter is available and uses it to return the client.
+func (s *Server) GetClient(ctx context.Context) (dynamic.Interface, error) {
```
absoludity, line 89: Extracted this code to a function that can be unit-tested separately since the functionality is required for each handler in the module.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server.go
```diff
@@ -104,36 +114,38 @@ func (s *Server) GetAvailablePackages(ctx context.Context, request *corev1.GetAv
...
 	}, nil
 }
 
-// GetPackageRepositories returns the package repositories based on the request.
-func (s *Server) GetPackageRepositories(ctx context.Context, request *corev1.GetPackageRepositoriesRequest) (*corev1.GetPackageRepositoriesResponse, error) {
-	// TODO: replace incluster config with the user config using token from request meta.
-	config, err := rest.InClusterConfig()
-	if err != nil {
-		return nil, fmt.Errorf("unable to create incluster config: %w", err)
+func availablePackageFromUnstructured(ap *unstructured.Unstructured) (*corev1.AvailablePackage, error) {
```
absoludity, line 128: Extracted this so that the functionality can be unit-tested in isolation, rather than as part of the `GetAvailablePackages` unit tests.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server.go
```diff
@@ -146,26 +158,51 @@ func (s *Server) GetPackageRepositories(ctx context.Context, request *corev1.Get
...
-		repo.Url = url
-
 		responseRepos = append(responseRepos, repo)
 	}
 	return &corev1.GetPackageRepositoriesResponse{
 		Repositories: responseRepos,
 	}, nil
 }
+
+func packageRepositoryFromUnstructured(pr *unstructured.Unstructured) (*corev1.PackageRepository, error) {
```
absoludity, line 172: Similarly here (though I've actually written the unit tests for this one :) ).

gfichtenholt (commented):
I don't really understand why we are checking in generated files (such as *.pb.go and *.pb.gw.go)? Aren't those produced as part of the build? I could be wrong, just need a clarification

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_test.go
```diff
@@ -47,6 +48,53 @@ func TestGetAvailablePackagesStatus(t *testing.T) {
 			},
 			statusCode: codes.FailedPrecondition,
 		},
+		{
+			name: "returns failed-precondition when configGetter itself errors",
+			clientGetter: func(context.Context) (dynamic.Interface, error) {
+				return nil, fmt.Errorf("Bang!")
+			},
+			statusCode: codes.FailedPrecondition,
```
antgamdia, line 56: Perhaps I'm missing something, but, isn't it the same tests as above? :P

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_test.go
```diff
@@ -185,14 +233,216 @@ func TestGetAvailablePackages(t *testing.T) {
 				t.Fatalf("%+v", err)
 			}
 
-			if got, want := response.Packages, tc.expectedPackages; !cmp.Equal(got, want, cmp.Comparer(pkgEqual)) {
-				t.Errorf("mismatch (-want +got):\n%s", cmp.Diff(want, got, cmp.Comparer(pkgEqual)))
+			if got, want := response.Packages, tc.expectedPackages; !cmp.Equal(got, want, cmpopts.IgnoreUnexported(corev1.AvailablePackage{})) {
```
antgamdia, line 236: I guess this is for ignoring the optional fields and just checking if the required fields are present, no?

cmd/kubeapps-apis/server/server.go
```diff
@@ -95,3 +96,17 @@ type gwHandlerArgs struct {
 	addr        string
 	dialOptions []grpc.DialOption
 }
+
+// Create a gateway mux that does not emit unpopulated fields.
+func gatewayMux() *runtime.ServeMux {
```
antgamdia, line 101: Ah, nice, so here is where we should put something like [1] if we wanna support real HTTP status code when serving the rest-ish API

[1] https://grpc-ecosystem.github.io/grpc-gateway/docs/mapping/customizing_your_gateway/#customizing-routing-errors

antgamdia (approved):
Awesome, this PR paves the way to have the core `GetAvailablePackages` and `GetPackageRepositories`. Really interesting. Thank you for the unit tests!!

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_test.go
absoludity: Woops, thanks.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_test.go
absoludity: No, it's because `cmp.Equal` (and `cmp.Diff`) will by default error if they can't compare an unexported field, like this:
```
    --- FAIL: TestGetAvailablePackages/it_returns_carvel_packages_from_the_cluster (0.00s)
panic: cannot handle unexported field at {[]*v1alpha1.AvailablePackage}[1].state:
        "github.com/kubeapps/kubeapps/cmd/kubeapps-apis/gen/core/packages/v1alpha1".AvailablePackage
consider using a custom Comparer; if you control the implementation of type, you can also consider using an Exporter, AllowUnexported, or cmpopts.IgnoreUnexported [recovered]
        panic: cannot handle unexported field at {[]*v1alpha1.AvailablePackage}[1].state:
        "github.com/kubeapps/kubeapps/cmd/kubeapps-apis/gen/core/packages/v1alpha1".AvailablePackage
consider using a custom Comparer; if you control the implementation of type, you can also consider using an Exporter, AllowUnexported, or cmpopts.IgnoreUnexported

```

cmd/kubeapps-apis/server/server.go
absoludity: Perhaps, if we need to customise them, but the default conversions done by grpc-gateway should be what we need though, right? https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/errors.go#L36

I think the doc you linked to is just for being able to customise the *routing* errors, the default implementation for which is at https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/errors.go#L165 ?

absoludity:
> I don't really understand why we are checking in generated files (such as *.pb.go and *.pb.gw.go)? Aren't those produced as part of the build? I could be wrong, just need a clarification

I think the consensus in the go community is that If the generated code is used only as part of the build process when building your project, then yes, it should always just be produced as part of the build and not be checked in.

protobuf/grpc differs in that the generated code is not only used as part of the build process, it generates clients that external projects use to be able to build *their* projects. So people should be able to include kubeapps (currently, though in the future we may split out the apis server) as a dependency for their project, which will `go get` our file, and their project source uses the clients. For this reason, most go projects that generate grpc clients (and other artifacts) include them in the source (for example, run `find . -name "*.pb.go"` on a clone of the k8s repo, or other grpc-using projects).

Of course there are other ways to solve this, and some projects do that instead (eg. publish the generated clients only as part of a build process to a different go-gettable repository), but I've just gone with the norm. Happy to revisit it later.

gfichtenholt (approved):
Thanks for clarification

cmd/kubeapps-apis/server/server.go
antgamdia: Ahh, great, so we already have default conversions even for the routing. Good to know!  When I saw the response bodies (`{ "code": 5, "message": "Not Found" }`) I initially thought they weren't being converted, but, yep, having a look at the whole HTTP response, the status codes are correct.
Thank you for letting me know!

Issue #2686, opened 2021-04-16
Authenticated session expires when initial jwt token expires


absoludity:
### Description:

Noticed while testing on TKG that when we login via OIDC (of any type), the session lasts only until the token expires, whereas oauth2-proxy should be refreshing the token.

### Steps to reproduce the issue:

1. Login via OIDC
2. Leave the browser tab open for 5-10 mins (depending on the token time set in your oauth2 client settings)
3. Try to use Kubeapps

### Describe the results you received:

A helpful notice about the session expiring and logging out.

### Describe the results you expected:

oauth2-proxy should be renewing the token a certain number of times.

### Additional information you deem important (e.g. issue happens only occasionally):

It may be we just need to test using the [--cookie-refresh option](https://oauth2-proxy.github.io/oauth2-proxy/docs/configuration/overview/#command-line-options) to oauth2-proxy, not sure.

dlaloue-vmware:
it looks like the refresh issue is a know issue in oauth2 proxy and is being worked on:
    https://github.com/oauth2-proxy/oauth2-proxy/issues/490
I will monitor that thread to know when a fix is available and try to verify it locally.

absoludity:
I think that's a different issue? That describes the situation where once the id_token expires, oauth2-proxy does refresh to get a new token, but doesn't update the session age, and so continues to refresh the token unnecessarily from that point on (until the access token is also expired).

The situation we want to check here is whether the refresh even happens without logging the user out: my experience (which I just dumped into this issue) was that I was logged out as soon as the first id_token expired, rather than being able to continue with a refreshed token. It could just be the configuration, not sure, but we need to ensure that when the id token expires, oauth2-proxy refreshes and provides a new one transparently to Kubeapps, without kubeapps logging out etc.

antgamdia:
I've been having a look at the issue as well:

## Configurations:

### Case 1: access token expires with a proper refresh. Expected result: auto-refresh, no logout :ok: 

```
Access token TTL: 1 min
Refresh token TTL: 60 days
--cookie-refresh=10s
```

``` 
[...]

PR #2900, opened 2021-05-27
updated doc with troubleshooting section for console logout issue

dlaloue-vmware:
### Description of the change

Added troubleshooting section for the console logout issue reported in https://github.com/kubeapps/kubeapps/issues/2686
Updated a few sample sections as well to include the new option.

### Benefits

Users may encounter the issue where they are unexpectedly logged out of the console. The new section helps identify the issue and provide a way to configure Kubeapps to prevent/fix the issue.


docs/user/using-an-OIDC-provider.md
```diff
@@ -359,3 +361,11 @@ Another common point of confusion is the `--oidc-username-prefix` option specifi
 ### Checking the logs of your Kubernetes API server
 
 Finally, if none of the above are relevant to your issue, you can check the logs of the Kubernetes API server deployment for OIDC-related lines at the time of your login attempt. These may show a configuration issue with the API server itself.
+
+### User automatically logged out from Kubeapps Console
+
+When using the default auth proxy, some users may experience the behavior where they are automatically logged out from the console.
+The default auth proxy is not configured to refresh the access/openid token and the console will logout once the token expires. In the case of Keycloak for example, this can happen quickly as the default access token expiration is 5mn.
```
absoludity, line 368: Isn't this statement going to be incorrect once your chart change lands (as it's setting a sensible default, which ever way it does so)?

docs/user/using-an-OIDC-provider.md
```diff
@@ -359,3 +361,11 @@ Another common point of confusion is the `--oidc-username-prefix` option specifi
...
+
+To avoid this issue, add the option `--cookie-refresh=2m` to `authProxy.additionalFlags` in your values file. The duration for the refresh must be lesser than the access/openid expiration time configured in the OAuth2/OIDC provider.
```
absoludity, line 370: Similar to my comment on the chart, I think this would be a better UX to be able to just set a chart value directly. Guess we'll want to wait until the chart change is landed before updating this.

docs/user/using-an-OIDC-provider.md
dlaloue-vmware: this will be obsolete for new installation / upgrades once we have the new chart.
but this is an existing problem for users, so i think it is useful to have a troubleshooting section for it.
I agree that i will have to reword it once we have the change in the chart.

docs/user/using-an-OIDC-provider.md
dlaloue-vmware: yes, this was worded with assumption the new chart is not yet available.
if we wait for the chart, i will reword accordingly (reason why i marked the PR as draft :-)

docs/user/using-an-OIDC-provider.md
```diff
@@ -359,3 +361,16 @@ Another common point of confusion is the `--oidc-username-prefix` option specifi
 ### Checking the logs of your Kubernetes API server
 
 Finally, if none of the above are relevant to your issue, you can check the logs of the Kubernetes API server deployment for OIDC-related lines at the time of your login attempt. These may show a configuration issue with the API server itself.
+
+### User automatically logged out from Kubeapps Console
+
+When using the default auth proxy, some users may experience the behavior where they are automatically logged out from the console.
+The default auth proxy (prior to version 7.0.5) is not configured to refresh the access/openid token and the console will logout once the token expires. In the case of Keycloak for example, this can happen quickly as the default access token expiration is 5mn.
```
absoludity, line 368:
```suggestion
Prior to Kubeapps 7.0.5, the auth proxy configuration did not include a default `--cookie-refresh` value to refresh the access/openid token and so the console will logout once the token expires. In the case of Keycloak for example, this can happen quickly as the default access token expiration is 5m.
```

docs/user/using-an-OIDC-provider.md
```diff
@@ -359,3 +361,16 @@ Another common point of confusion is the `--oidc-username-prefix` option specifi
...
+
+To avoid this issue, you can do one of the following:
+ - upgrade Kubeapps to version 7.0.5+
```
absoludity, line 371:
```suggestion
 - upgrade Kubeapps to version 7.0.5+ which sets a default of `--cookie-refresh=2m` and exposes the value in the chart values as `authProxy.cookieRefresh`.
```

docs/user/using-an-OIDC-provider.md
```diff
@@ -359,3 +361,16 @@ Another common point of confusion is the `--oidc-username-prefix` option specifi
...
+ - update Kubeapps by adding the option `--cookie-refresh=2m` to `authProxy.additionalFlags`.
+
+The duration for the refresh must be lesser than the access/openid expiration time configured in the OAuth2/OIDC provider.
+
+**Note**: The issue may still occur even after upgrading or updating Kubeapps as OAuth2 Proxy does not support cookie-refresh for all providers. See [OAuth2 Proxy Configuration Overview](https://oauth2-proxy.github.io/oauth2-proxy/docs/configuration/overview/#footnote1) for the list of supported providers.
```
absoludity, line 376: Not sure how relevant this is given that we recommend the `oidc` provider which is supported. But yep, people aren't restricted to that (eg. I've used the google provider in certain cases). Perhaps:
```suggestion
**Note**: If you have configured a provider other than `oidc` for oauth2-proxy, the issue may still occur even after upgrading or updating Kubeapps as OAuth2 Proxy does not support cookie-refresh for all providers. See [OAuth2 Proxy Configuration Overview](https://oauth2-proxy.github.io/oauth2-proxy/docs/configuration/overview/#footnote1) for the list of supported providers.
```

absoludity (approved):
A couple of suggestions but feel free to ignore them if you think otherwise.

docs/user/using-an-OIDC-provider.md
```diff
@@ -359,3 +361,16 @@ Another common point of confusion is the `--oidc-username-prefix` option specifi
 ### Checking the logs of your Kubernetes API server
 
 Finally, if none of the above are relevant to your issue, you can check the logs of the Kubernetes API server deployment for OIDC-related lines at the time of your login attempt. These may show a configuration issue with the API server itself.
+
+### User automatically logged out from Kubeapps Console
+
+When using the default auth proxy, some users may experience the behavior where they are automatically logged out from the console.
+Prior to Kubeapps 7.0.5, the auth proxy configuration did not include a default `--cookie-refresh` value to refresh the access/openid token and so the console will logout once the token expires. In the case of Keycloak for example, this can happen quickly as the default access token expiration is 5m.
```
antgamdia, line 368:
```suggestion
Prior to the Kubeapps chart version 7.1.0, the auth proxy configuration did not include a default `--cookie-refresh` value to refresh the access/openid token and so the console will logout once the token expires. In the case of Keycloak for example, this can happen quickly as the default access token expiration is 5m.
```

docs/user/using-an-OIDC-provider.md
```diff
@@ -359,3 +361,16 @@ Another common point of confusion is the `--oidc-username-prefix` option specifi
...
+
+To avoid this issue, you can do one of the following:
+ - upgrade Kubeapps to version 7.0.5+ which sets a default of `--cookie-refresh=2m` and exposes the value in the chart values as `authProxy.cookieRefresh`.
```
antgamdia, line 371:
```suggestion
 - upgrade the Kubeapps chart to version 7.1.0+ which sets a default of `--cookie-refresh=2m` and exposes the value in the chart values as `authProxy.cookieRefresh`.
```

antgamdia (approved):
Great, thank you for performing these changes.
I've also noticed a couple of (already existing minor typos),  if you want, check them; but feel free to ignore them as they are not from this PR.

- Remove duplicated "accept" in: `accept accept access_tokens from their identity provider as bearer tokens (see GKE below).`
- Replace "prodcution" by "production" in `the prodcution VMware cloud services issuer`

absoludity:
You'll need to switch this from a draft PR to be able to land it, I think, if you think it's ready. It's got +1s.

antgamdia:
@dlaloue-vmware, feel free to click "squash and merge" whenever you want. Looking forward to have the docs up to date with these changes  :)

Issue #2851, opened 2021-05-21
[kubeapps-apis] Switch to use user credentials from request


absoludity:
### Description:

For demoing initial behaviour, we've just used a service account so the endpoints can just be curl'd. The privileges of the service account should be completely removed and the plugins updated to always use the requesting user's own token (sent with the request).

### Additional information you deem important (e.g. issue happens only occasionally):

Should just be pulling the token out of the header (which in grpc, is in the metadata of the context, as [shown in the example](https://github.com/grpc/grpc-go/blob/master/examples/features/authentication/server/main.go#L100)), and using that with the k8s client config, rather than using the in cluster config.

PR #2901, opened 2021-05-27
Add kubeappsapis behind nginx

antgamdia:
### Description of the change

This PR simply adds (if the feature flag is enabled) the current kubeapps-apis behind the nginx reverse proxy. This way, we can pass through the authentication token as in the rest of the services.

### Benefits

Besides paving the way for #2851, it also allows using the API without any port-forwarding. 

### Possible drawbacks

Note that -currently- it uses a privileged SA to list everything, but it is not a production-ready thing and we soon change are changing it to use the user's credentials, it's not a big deal.

### Applicable issues

N/A

### Additional information

I hope to be sending the auth PR soon :S

### DIFF
--- a/chart/kubeapps/templates/frontend/configmap.yaml
+++ b/chart/kubeapps/templates/frontend/configmap.yaml
@@ -140,9 +140,9 @@ data:
       }
 
     {{- if .Values.featureFlags.kubeappsAPIsServer }}
-      location ~* /api/kubeappsapis {
+      location ~* /apis {
         rewrite ^ $request_uri; # pass the encoded url downstream as is,
-        rewrite /api/kubeappsapis([^?]*) $1?$args break;
+        rewrite /apis([^?]*) $1?$args break;
     {{- if .Values.frontend.proxypassExtraSetHeader }}
         proxy_set_header {{ .Values.frontend.proxypassExtraSetHeader }};
     {{- end }}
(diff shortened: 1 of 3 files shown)


chart/kubeapps/templates/frontend/configmap.yaml
```diff
@@ -139,6 +139,22 @@ data:
     {{- end }}
       }
 
+    {{- if .Values.featureFlags.kubeappsAPIsServer }}
+      location ~* /api/kubeappsapis {
```
absoludity, line 143: Could we just use `/kubeappsapis` or even `/apis` given that the full paths would then be things like `/apis/core/packages/v1alpha1/...`

What do you think? It's still behind a featureflag so I've no problem changing it later.

absoludity (approved):
> Note that -currently- it uses a privileged SA to list everything, but it is not a production-ready thing and we soon change are changing it to use the user's credentials, it's not a big deal.

Right, I'll feel better about this once the `kubeappsapis` service requires a `--unsafe-use-demo-svc-account` or similar to use the service account. Currently it uses it by default for easy testing and because it's not (wasn't) exposed, but as long as that's one of the first changes you'll make to the kubeappsapis service, I'm happy :)

chart/kubeapps/templates/frontend/configmap.yaml
antgamdia: Good idea indeed, shorter and more concise. Changing right away!
Regarding the SA flag, yep, totally agree. Of course, we are not gonna release a new official version as it is right now,
However, let's see how this flag can be passed through, I'm struggling with passing information towards the plugins. I Will let you know in another PR.

PR #2904, opened 2021-05-28
Demo aggregate core apis.

### LINKED ISSUE #2848: [kubeapps-apis] Demo core.v1alpha1.GetPackageRepositories

absoludity:
### Description of the change

This was an investigative PR that adds the ability to recognise when plugins satisfy the interface for core services (well, currently only core.packaging.v1alpha1) and register them for aggregation in the core api, so that you can:

```
curl -s http://localhost:8080/core/packages/v1alpha1/packagerepositories | jq .
{
  "repositories": [
    {
      "name": "bitnami",
      "namespace": "flux-system",
      "url": "https://charts.bitnami.com/bitnami",
      "plugin": {
        "name": "fluxv2.packages",
        "version": "v1alpha1"
      }
    },
    {
      "name": "demo-package-repository",
      "url": "k8slt/corp-com-pkg-repo:1.0.0",
      "plugin": {
        "name": "kapp_controller.packages",
        "version": "v1alpha1"
      }
    }
  ]
}
```
and similar for available packages.

It demonstrates only those two calls at the moment, but the idea is that a client could then refer to such an available package and hit the core.packages api to install it, and the core api can call the relevant plugin implementation etc., enabling a client to potentially treat all plugins in a homogeneous way, where possible, while still allowing each plugin to have custom extra functionality for clients to use as needed.

### Benefits

The nice thing is that we can do this via go's interfaces in the same way that grpc does so.

### Applicable issues

  - fixes #2848

### Additional information

I have an idea how I can test the core API functions, but will do that in a follow-up PR if that's OK.

I'll add some notes in context inline.


cmd/kubeapps-apis/buf.gen.yaml
```diff
@@ -8,6 +8,22 @@ plugins:
...
+      # added explicitly. But a plugin can still embed the
+      # `Unimplemented**Server` struct, it's just not enforced. The
+      # unintentional downside of enforcing it via the
+      # `mustEmbedUnimplemented**Server` for us is that the
+      # `mustEmbedUnimplemented**Server` method is not exposed, which means that
+      # *only* implementations in the same package (ie. with access to the same
+      # unexposed method) can satisfy the interface. We explicitly want plugins
+      # defined in other packages to be able to satisfy a generic gRPC server
+      # interface.
+      # TODO: create upstream issue about this.
```
absoludity, line 25: I'll add an upstream issue and link here on Monday.

cmd/kubeapps-apis/server/packages.go
```diff
@@ -0,0 +1,82 @@
...
+
+import (
+	"context"
+
+	packages "github.com/kubeapps/kubeapps/cmd/kubeapps-apis/gen/core/packages/v1alpha1"
+)
+
+// packagesServer implements the API defined in proto/kubeappsapis/core/packages/v1alpha1/packages.proto
+type packagesServer struct {
+	packages.UnimplementedPackagesServiceServer
```
absoludity, line 23: This is the specially generated struct which we embed as always, we're just not forced to do so with the `require_unimplemented_servers=false`, and since the required interface is then generated *without* the private method, plugin servers are able to satisfy the interface as long as they implement the actual functions (eg. `GetPackageRepositories`, `GetAvailablePackages`, ... whatever else we add).

cmd/kubeapps-apis/server/packages.go
```diff
@@ -0,0 +1,82 @@
...
+	return &packagesServer{
+		plugins: plugins,
+	}
+}
+
+// GetAvailablePackages returns the packages based on the request.
+func (s packagesServer) GetAvailablePackages(ctx context.Context, request *packages.GetAvailablePackagesRequest) (*packages.GetAvailablePackagesResponse, error) {
+	pkgs := []*packages.AvailablePackage{}
+	for _, p := range s.plugins {
+		response, err := p.server.GetAvailablePackages(ctx, request)
```
absoludity, line 40: So there were a number of different ways to call the separate plugin versions. The simplest would have been to use the client and actually do the grpc call to localhost, but that then requires serializing/deserializing for every request. Instead, because we recorded at startup which plugins satisfy the core packaging server interface and store a reference to the implementation, we can just use the implementations directly here without going over the wire at all, so no serialization/deserialization required.

cmd/kubeapps-apis/server/plugins.go
```diff
@@ -54,16 +70,18 @@ func NewPluginsServer(pluginDirs []string, registrar grpc.ServiceRegistrar, gwAr
 		log.Fatalf("failed to check for plugins: %v", err)
 	}
 
-	pluginDetails, err := registerPlugins(pluginPaths, registrar, gwArgs)
+	ps := &pluginsServer{}
+
+	pluginDetails, err := ps.registerPlugins(pluginPaths, registrar, gwArgs)
```
absoludity, line 74: This file is refactored a little so that the plugin detail can be passed down the chain of calls and included together with the implementation server to use in the core server.

cmd/kubeapps-apis/server/plugins.go
```diff
@@ -109,19 +128,43 @@ func registerPlugins(pluginPaths []string, grpcReg grpc.ServiceRegistrar, gwArgs
...
 	if !ok {
-		var dummyFn grpcRegisterFunctionType = func(grpc.ServiceRegistrar) {}
-		return fmt.Errorf("unable to use %q in plugin %q due to mismatched signature.\nwant: %T\ngot: %T", grpcRegisterFunction, pluginPath, dummyFn, grpcRegFn)
+		var dummyFn grpcRegisterFunctionType = func(grpc.ServiceRegistrar) interface{} { return nil }
+		return fmt.Errorf("unable to use %q in plugin %v due to mismatched signature.\nwant: %T\ngot: %T", grpcRegisterFunction, pluginDetail, dummyFn, grpcRegFn)
 	}
-	grpcFn(registrar)
+	server := grpcFn(registrar)
 
+	return s.registerPluginsSatisfyingCoreAPIs(server, pluginDetail)
```
absoludity, line 143: This is the new piece that is called when registering a grpc function ^^^ , defined below.

cmd/kubeapps-apis/server/plugins.go
```diff
@@ -109,19 +128,43 @@ func registerPlugins(pluginPaths []string, grpcReg grpc.ServiceRegistrar, gwArgs
...
+	// The following heck if the service implements an interface is what
+	// grpc-go itself does, see:
+	// https://github.com/grpc/grpc-go/blob/v1.38.0/server.go#L621
+	serverType := reflect.TypeOf(pluginSrv)
+	corePackagesType := reflect.TypeOf((*packages.PackagesServiceServer)(nil)).Elem()
+
+	if serverType.Implements(corePackagesType) {
+		pkgsSrv, ok := pluginSrv.(packages.PackagesServiceServer)
+		if !ok {
+			return fmt.Errorf("Unable to convert plugin %v to core PackagesServicesServer although it implements the same.", pluginDetail)
```
absoludity, line 159: This error should not be possible, since we've already checked that the pluginSrv does indeed implement the core packages interface, but I can't ignore it :)

cmd/kubeapps-apis/buf.gen.yaml
antgamdia: So, just for clarification: there is nothing wrong in continue using `require_unimplemented_servers=true` if the `mustEmbedUnimplemented**Server` was exposed when generating the code.
So, the upstream issue is for them to expose it, am I getting it right?

antgamdia (approved):
Thank you for this interesting approach for registering the plugins satisfying the interface!

cmd/kubeapps-apis/buf.gen.yaml
absoludity: Woops, still need to create the upstream issue. Yes, that's what I'll be suggesting upstream, though there may be a reason I'm not aware of as to why they want it private.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/main.go
```diff
@@ -22,9 +22,12 @@ import (
 	v1alpha1 "github.com/kubeapps/kubeapps/cmd/kubeapps-apis/gen/plugins/fluxv2/packages/v1alpha1"
 )
 
-// RegisterWithGRPCServer enables a plugin to register with a gRPC server.
-func RegisterWithGRPCServer(s grpc.ServiceRegistrar) {
+// RegisterWithGRPCServer enables a plugin to register with a gRPC server
+// returning the server implementation.
+func RegisterWithGRPCServer(s grpc.ServiceRegistrar) interface{} {
+	svr := NewServer()
 	v1alpha1.RegisterPackagesServiceServer(s, NewServer())
```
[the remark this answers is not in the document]
absoludity: Indeed, thanks.

cmd/kubeapps-apis/server/packages.go
absoludity: Yeah - especially since it's stored as (a pointer to) the interface, makes it trivial here.

cmd/kubeapps-apis/server/plugins.go
```diff
@@ -81,26 +98,27 @@ func (s *pluginsServer) GetConfiguredPlugins(ctx context.Context, in *plugins.Ge
...
+			pluginDetails = append(pluginDetails, pluginDetail)
 		}
 
-		if err = registerHTTP(p, pluginPath, gwArgs); err != nil {
+		if err = s.registerGRPC(p, pluginDetail, grpcReg); err != nil {
 			return nil, err
 		}
 
-		if pluginDetail, err := getPluginDetail(p, pluginPath); err != nil {
+		if err = registerHTTP(p, pluginDetail, gwArgs); err != nil {
```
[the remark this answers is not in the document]
absoludity: `registerGRPC` was only updated to a method of the `pluginServer` because I needed to store state while registering each plugin. Yep, I could update `registerHTTP` to be a method as well for consistency, but my preference is for stateless functions where possible (easier to reason about and obvious that there's no state of pluginServer being used).

cmd/kubeapps-apis/server/server.go
```diff
@@ -66,6 +67,13 @@ func Serve(port int, pluginDirs []string) {
 		log.Fatalf("failed to register core.plugins handler for gateway: %v", err)
 	}
 
+	// Create the core.packages server and register it for both grpc and http.
```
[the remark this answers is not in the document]
absoludity: Yes, as in, the kubeapps-apis service manually registers both the `core.plugins` and `core.packages` packages, the only two core packages currently. Yes, the `core.plugins` package also loads and registers all the loaded plugins.

I actually want to do another PR to move the `core.plugins` and `core.packages` into their own (versioned) packages too, but I'll wait until your auth PR has landed before doing that refactor.

PR #2908, opened 2021-05-30
Utilize user credentials to invoke the new kubeapps-api

### LINKED ISSUE #2851: [kubeapps-apis] Switch to use user credentials from request

antgamdia:
### Description of the change

This PR tries to replace the current hardcoded SA with the actual user's credentials, including the proper configuration for the Pinniped scenario.
Currently, it also adds a new chart flag for dev purposes (`kubeappsapis.UnsafeUseDemoSA: true`) that allow having the same behavior prior to this PR.

### Benefits

A consistent auth approach with the rest of the Kubeapps components.

### Possible drawbacks

~No SA for quick testing, but it will be get solved with a flag soon.~

### Applicable issues

  - fixes #2851

### Additional information

WIP, there is a couple of things I wanna re-check and thoughts:

 - (_outdated_) As it is a sort of transversal functionality, we should provide a unified auth experience for plugin developers. Can plugins invoke main program functions? Another way round: can the main program share info with the plugins?
Currently, I just came with using envars to share the config so that it can be accessed by the plugins.
 - (_outdated_) Now we are passing the parsed config back to the plugins. Ideally, I'd like to pass the whole k8s client, but we still the token in each request (plus ... caching? we are creating the config each time the method is invoked :S). 
 - (_outdated_) Testing is still pending, I'll do it tomorrow

For subsequent PRs

- [x] Use shared directory for manifests: #2926 
- [ ] Move the parser under `pkg/..` to avoid duplicated code: https://github.com/kubeapps/kubeapps/pull/2908#discussion_r644476352
- (future) Improve the clientGetter testing: https://github.com/kubeapps/kubeapps/pull/2908#discussion_r644528421


cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
```diff
@@ -293,3 +331,82 @@ func getHelmIndexFileFromURL(indexURL string) (*helmrepo.IndexFile, error) {
...
+	if err != nil {
+		return nil, status.Errorf(codes.FailedPrecondition, fmt.Sprintf("unable to get client : %v", err))
+	}
+	return client, nil
+}
+
+// extractToken returns the token passed through the gRPC request in the "authorization" metadata
+// It is equivalent to the A"uthorization" usual HTTP 1 header
+// For instance: authorization="Bearer abc" will return "abc"
+func extractToken(ctx context.Context) (string, error) {
```
absoludity, line 351: Great - this is what I expected the crux of the initial PR to be - testing that we get the token correctly.

absoludity (commented):
> WIP, there is a couple of things I wanna re-check and thoughts:
> 
>     * As it is a sort of transversal functionality, we should provide a unified auth experience for plugin developers. Can plugins invoke main program functions?

No - I don't think they should be able to do so, but open to suggestions.

> Another way round: can the main program share info with the plugins?

Yes, we can definitely do this. For example, we could include configuration data in the signature of the `RegisterWithGRPCServer` function, or we could define a set of env vars that are available. Or, it may be useful (as I think you suggested) to even pass a function or something to the plugins when registering that returns a k8s config (so each plugin doesn't have to do this, or care about pinniped etc.)

>
>       Currently, I just came with using envars to share the config so that it can be accessed by the plugins.

One thing: I didn't expect this initial PR to require cluster config or anything to do with pinniped etc. Is that just because your local dev server is setup with pinniped? I'd recommend first just using a normal non-pinniped setup and verifying the way to ensure plugins have a valid k8s config. Then we can update to support pinniped etc. But up to you.

Glad the crux of it is straight-forward (extracting the token from the context, which is translated from the http header).

gfichtenholt:
I am not 100% sure about this, but I think I am hitting some RBAC-related issues, possibly related to this PR 
with the current plugin code in fluxv2 server.go. I notice that 
`  func clientForRequestContext(ctx context.Context) (dynamic.Interface, error) `
currently isn't making use of ctx argument with the "to be updated" notice
The issue I am hitting is I can programmatically list existing flux CRDs with something like this:
   `chartList, err := client.Resource(chartsResource).List(ctx, metav1.ListOptions{})`
just fine. But creating a CRD via the same resource
   `newChart, err := client.Resource(chartsResource).Create(ctx, &unstructuredChart, metav1.CreateOptions{})`
results in 
`0531 06:36:34.695099       1 server.go:235] error creating chart: the server could not find the requested resource`
 I have done some digging and its looking like the issue is likely to be due to some RBAC settings that somehow 
 allow read-only operations but not write operations for the user identity on behalf of which the code is being executed.
BTW, related question - is there a call in go I can make to log the identity? 
Also, interestingly, just as an experiment I added a similar piece of code into kapp_controller plug in
 (i.e. use the resource to programmatically create a packageRepository)
  and it worked there. I can't quite explain that one yet. To be continued...

chart/kubeapps/templates/kubeappsapis/config.yaml
```diff
@@ -0,0 +1,18 @@
...
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    app.kubernetes.io/component: kubeappsapis
+    {{- if .Values.commonLabels }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" . ) | nindent 4 }}
+    {{- end }}
+  {{- if .Values.commonAnnotations }}
+  annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+  {{- end }}
+data:
+  clusters.conf: |-
```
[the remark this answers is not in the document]
antgamdia: Exactly, I just went this way because of the current structure of the repo, but I certainly don't like having two identical manifests. The point here is: will kubeops be entirely replaced by kubeapps-apis? if so, we can bear with it for the moment. If not, perhaps a good option is just creating a `shared` folder with common manifests like this one.

chart/kubeapps/templates/kubeappsapis/deployment.yaml
```diff
@@ -74,6 +74,14 @@ spec:
...
             - name: PORT
               value: {{ .Values.kubeappsapis.containerPort | quote }}
+            {{- if .Values.clusters }}
+            - name: CLUSTERS_CONFIG_PATH
+              value: /config/clusters.conf
+            {{- end }}
+            {{- if .Values.pinnipedProxy.enabled }}
+            - name: PINNIPED_PROXY_URL
+              value: http://{{ template "kubeapps.pinniped-proxy.fullname" . }}.{{ .Release.Namespace }}:{{ .Values.pinnipedProxy.containerPort }}
+            {{- end }}
```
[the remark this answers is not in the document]
antgamdia: Yep, I did because of my dev setup. I tend to use use the full setup (pinniped + 2 clusters), in fact, sometimes I also use the AKS cluster for testing specific things. 
Perhaps it was simpler to just get the token, but, for me, the crux was on how to pass the authentication from the main program back to the plugins.
Now I have everything more clear, I can split it into two PRs if you want.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
```diff
@@ -169,13 +177,46 @@ func (s *Server) GetAvailablePackages(ctx context.Context, request *corev1.GetAv
...
+	clustersConfigPath := os.Getenv("CLUSTERS_CONFIG_PATH")
+	if clustersConfigPath == "" {
+		log.Fatal("CLUSTERS_CONFIG_PATH should be defined")
+	}
+
+	const clustersCAFilesPrefix = "/etc/additional-clusters-cafiles"
+
+	pinnipedProxyURL := os.Getenv("PINNIPED_PROXY_URL")
+	if pinnipedProxyURL == "" {
+		log.Fatal("PINNIPED_PROXY_URL should be defined")
```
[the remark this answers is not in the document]
antgamdia: Absolutely agree, I just copy&pasted the same structure. Once I send today's updates, this config is retrieved as a Cobra argument, not an issue anymore.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
```diff
@@ -293,3 +331,82 @@ func getHelmIndexFileFromURL(indexURL string) (*helmrepo.IndexFile, error) {
...
+	return client, nil
+}
+
+// extractToken returns the token passed through the gRPC request in the "authorization" metadata
+// It is equivalent to the A"uthorization" usual HTTP 1 header
+// For instance: authorization="Bearer abc" will return "abc"
+func extractToken(ctx context.Context) (string, error) {
+	md, ok := metadata.FromIncomingContext(ctx)
+	if !ok {
+		return "", status.Errorf(codes.Unauthenticated, "no authorization metadata found")
```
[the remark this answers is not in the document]
antgamdia: Ooops, that's true. I've updated that function accordingly. 

However, I still have a doubt here: do anonymous endpoints (eg `getConfiguredPlugins`) still require the k8s authentication? In this PR I just added that logic in those which are not anonymous.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
```diff
@@ -49,14 +52,18 @@ type Server struct {
...
 
 // NewServer returns a Server automatically configured with a function to obtain
 // the k8s client config.
-func NewServer() *Server {
+func NewServer(config kube.ClustersConfig, useSA bool) *Server {
 	return &Server{
-		clientGetter: clientForRequestContext,
+		clientGetter:    clientForRequestContext,
+		clustersConfig:  config,
+		unsafeUseDemoSA: useSA,
```
antgamdia, line 66: I don't like that part, I need the `useSA` and `config` in the server to pass the information via `v1alpha1.RegisterKappControllerPackagesServiceServer(s, NewServer(config, unsafeUseDemoSA))` in the  `RegisterWithGRPCServer` function. Currently, it's the only way I came up with, but I feel I'm missing something

absoludity:
> I am not 100% sure about this, but I think I am hitting some RBAC-related issues, possibly related to this PR
> with the current plugin code in fluxv2 server.go. I notice that
> ` func clientForRequestContext(ctx context.Context) (dynamic.Interface, error)`
> currently isn't making use of ctx argument with the "to be updated" notice

Correct, yes that's the bit that Antonio is working on in this branch, but it won't be affecting your experience yet (nor when it lands, I hope :) ).

> The issue I am hitting is I can programmatically list existing flux CRDs with something like this:
> `chartList, err := client.Resource(chartsResource).List(ctx, metav1.ListOptions{})`
> just fine. But creating a CRD via the same resource
> `newChart, err := client.Resource(chartsResource).Create(ctx, &unstructuredChart, metav1.CreateOptions{})`
> results in
> `0531 06:36:34.695099 1 server.go:235] error creating chart: the server could not find the requested resource`

My first thought was that our [development RBAC](https://github.com/kubeapps/kubeapps/blob/master/chart/kubeapps/templates/kubeappsapis/rbac.yaml) may not have you covered, but it seems to (since all the flux CRDs appear to be for `source.toolkit.fluxcd.io/v1beta1` which that RBAC covers).

My next guess is that it's a namespace-scoped  object (a HelmChart)? So perhaps you need to do `client.Resource(chartsResource).Namespace('foo').Create(...)`, otherwise the client may be looking for a cluster-scoped resource (which doesn't exist). But that's just a guess.

> BTW, related question - is there a call in go I can make to log the identity?

Currently for the ease of development, all calls are using the in-cluster config which is the service account linked to the above RBAC. Once Antonio finishes, we'll additionally have the option of switching off the development-only RBAC so the user (JWT) identity is used, which I guess we could parse and use the username. The `auth-proxy` logs on the frontend already do something like this for us, but if you think it's needed for this service, it should be straight forward to add once Antonio is finished.

> Also, interestingly, just as an experiment I added a similar piece of code into kapp_controller plug in
> (i.e. use the resource to programmatically create a packageRepository)
> and it worked there. I can't quite explain that one yet. To be continued...

Right, that might be more evidence for my hypothesis above, given the `PackageRepository` is a cluster-scoped object currently.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/main.go
```diff
@@ -20,11 +20,12 @@ import (
 	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
 	plugins "github.com/kubeapps/kubeapps/cmd/kubeapps-apis/gen/core/plugins/v1alpha1"
 	v1alpha1 "github.com/kubeapps/kubeapps/cmd/kubeapps-apis/gen/plugins/fluxv2/packages/v1alpha1"
+	"github.com/kubeapps/kubeapps/pkg/kube"
 )
 
 // RegisterWithGRPCServer enables a plugin to register with a gRPC server.
-func RegisterWithGRPCServer(s grpc.ServiceRegistrar) {
-	v1alpha1.RegisterFluxV2PackagesServiceServer(s, NewServer())
+func RegisterWithGRPCServer(s grpc.ServiceRegistrar, config kube.ClustersConfig, unsafeUseDemoSA bool) {
```
absoludity, line 27: So, what I'd recommend here is to use the fact that functions are first-class citizens in go, and so can be passed around as parameters. You'll see that I've already used this to be able to test the plugins, but it looks to me like it'd be worth extending that here so that the signature was:

```suggestion
func RegisterWithGRPCServer(s grpc.ServiceRegistrar, clientGetter func(context.Context) (dynamic.Interface, error)) {
```
or more simply by defining the type alias:
```
type DynamicClientGetter func(context.Context) (dynamic.Interface, error)
func RegisterWithGRPCServer(s grpc.ServiceRegistrar, clientGetter DynamicClientGetter) {
```

(see https://golang.org/doc/codewalk/functions/ for more about using functions like this).

That way, we can define the function in the kubeappsapis once, and once only, then pass that function to each plugin when registering. That way, plugins literally don't have to care about the details, they just call the function with the context to get a client (and/or handle the error).

Quite neatly, the only change would then be that the `NewServer(...)` functions would now just accept the client rather than creating it, and we can remove the function for creating clients from both plugins?

Let me know if something doesn't make sense there, or you get stuck. Happy to jump in.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
absoludity: Right, and more importantly, then each plugin is responsible for creating the client. Have a play with my suggestion above, if you get a chance.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
absoludity: So this functionality would be included as part of the `clientGetter` function passed to the plugin, so the plugin doesn't need to know or care how the auth is done.

gfichtenholt:
> > I am not 100% sure about this, but I think I am hitting some RBAC-related issues, possibly related to this PR
> > with the current plugin code in fluxv2 server.go. I notice that
> > ` func clientForRequestContext(ctx context.Context) (dynamic.Interface, error)`
> > currently isn't making use of ctx argument with the "to be updated" notice
> 
> Correct, yes that's the bit that Antonio is working on in this branch, but it won't be affecting your experience yet (nor when it lands, I hope :) ).
> 
> > The issue I am hitting is I can programmatically list existing flux CRDs with something like this:
> > `chartList, err := client.Resource(chartsResource).List(ctx, metav1.ListOptions{})`
> > just fine. But creating a CRD via the same resource
> > `newChart, err := client.Resource(chartsResource).Create(ctx, &unstructuredChart, metav1.CreateOptions{})`
> > results in
> > `0531 06:36:34.695099 1 server.go:235] error creating chart: the server could not find the requested resource`
> 
> My first thought was that our [development RBAC](https://github.com/kubeapps/kubeapps/blob/master/chart/kubeapps/templates/kubeappsapis/rbac.yaml) may not have you covered, but it seems to (since all the flux CRDs appear to be for `source.toolkit.fluxcd.io/v1beta1` which that RBAC covers).
> 
> My next guess is that it's a namespace-scoped object (a HelmChart)? So perhaps you need to do `client.Resource(chartsResource).Namespace('foo').Create(...)`, otherwise the client may be looking for a cluster-scoped resource (which doesn't exist). But that's just a guess.
> 
> > BTW, related question - is there a call in go I can make to log the identity?
> 
> Currently for the ease of development, all calls are using the in-cluster config which is the service account linked to the above RBAC. Once Antonio finishes, we'll additionally have the option of switching off the development-only RBAC so the user (JWT) identity is used, which I guess we could parse and use the username. The `auth-proxy` logs on the frontend already do something like this for us, but if you think it's needed for this service, it should be straight forward to add once Antonio is finished.
> 
> > Also, interestingly, just as an experiment I added a similar piece of code into kapp_controller plug in
> > (i.e. use the resource to programmatically create a packageRepository)
> > and it worked there. I can't quite explain that one yet. To be continued...
> 
> Right, that might be more evidence for my hypothesis above, given the `PackageRepository` is a cluster-scoped object currently.

Yep, that was exactly it, I needed to set the namespace, as I was dealing with Namespaced-scoped object. **Thanks, Michael!**. Your help is always spot on and very much appreciated. Still would like to see an example of piece of go code that logs the user identity on behalf of which the code is executed. I know how to do this in Java (using thread-locals) in spring or EE container but no clue in go

chart/kubeapps/templates/kubeappsapis/config.yaml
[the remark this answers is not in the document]
antgamdia: Great, let's go with the `shared` folder. However, in this PR, I'd opt not to modify kubeops, just creating the `cm` in the shared folder. In a follow-up PR I'll do the change in kubeops (to avoid more uncertainty/instability here)

chart/kubeapps/templates/kubeappsapis/deployment.yaml
```diff
@@ -71,6 +73,15 @@ spec:
...
             - /plugins/
+            {{- if .Values.clusters }}
+            - --clusters-config-path=/config/clusters.conf
+            {{- end }}
+            {{- if .Values.pinnipedProxy.enabled }}
+            - --pinniped-proxy-url=http://{{ template "kubeapps.pinniped-proxy.fullname" . }}.{{ .Release.Namespace }}:{{ .Values.pinnipedProxy.containerPort }}
+            {{- end }}
+            {{- if .Values.kubeappsapis.unsafeUseDemoSA }}
+            - --unsafe-use-demo-sa=true
+            {{- end }}
```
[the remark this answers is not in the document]
antgamdia: I haven't added it (because I thought we may want to keep it not that visible), but I've just included it. Thanks!

chart/kubeapps/templates/kubeappsapis/rbac.yaml
```diff
@@ -24,6 +24,7 @@ rules:
       - "source.toolkit.fluxcd.io"
     resources: ['*']
     verbs: ['*']
+{{- if .Values.kubeappsapis.unsafeUseDemoSA }}
```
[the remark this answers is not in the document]
antgamdia: Exactly, that was my idea. I don't really know how the RBAC of flux/kapp is being set by default, so I just wanted to have the ClusterRole in case if we need to create the binding for ourselves.

Once we get rid of the `unsafeUseDemoSA`, I guess we can remove the whole RBAC, as it is not something that Kubeapps should care about (as you mentioned).

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/main.go
antgamdia: > You'll see in ./cmd/root.go that it's sole responsibility is parsing all the args/config, then starting the server by calling server.Serve(...). Currently this has only two args, the port and the plugin dirs, but it is exactly the point where any info required to run the server should go.

> Does the above solve this? ie. if you pass the data you need into the Serve() function, we should have everything we need? My assumption is that that data will be used in the function you create... I'll take a look soon.

Yep, I got it. After refactoring a little bit for writing the tests, I just improved the b) option, creating two functions w/ and w/o the required params (still don't know why golang is against overloading functions...)

> Yeah, though right now you can just update the plugin functions to use the same signature? ...

I copied the example wrongly. I meant `main.DynamicClientGetter` and  `plugin.DynamicClientGetter`. So even if they are the same ` DynamicClientGetter`  since they're coming from different packages and the signatures don't match  :(

antgamdia:
I've tried to add as many unit tests as possible, but some functions are really coupled to a k8s cluster, so, even if I tried to mock some objects, the coverage is far from being ideal.

fwiw, I've tested it locally with a single cluster using pinniped and csp, so... let's mark it as ready for review (finally... :grin: )

cmd/kubeapps-apis/server/plugins.go
```diff
@@ -232,3 +253,140 @@ func listSOFiles(fsys fs.FS, pluginDirs []string) ([]string, error) {
...
+			return nil, fmt.Errorf("unable to create dynamic client: %w", err)
+		}
+	}
+	return client, nil
+}
+
+// dynClientGetterForContext returns a k8s client for use during interactions with the cluster.
+// It utilizes the user credential from the request context. The plugins just have to call this function
+// passing the context in order to retrieve the configured k8s client
+func dynClientGetterForContext(ctx context.Context) (dynamic.Interface, error) {
```
absoludity, line 342: As above, I think this is the reason you've needed global vars - so you can reference `pluginsServeOpts` inside this static function. But using a function closure should mean you don't need the global var at all.

absoludity (commented):
> Yep, I got it. After refactoring a little bit for writing the tests, I just improved the b) option, creating two functions w/ and w/o the required params (still don't know why golang is against overloading functions...)

The (b) option is using a global var I think, see inline comments about using function closures to avoid the globals. Regarding the two functions w/ and w/o the required params, I didn't see them? Perhaps you figured that out? (as I see you're using the raw signature now rather than a type alias defined in a package).

cmd/kubeapps-apis/server/plugins_test.go
```diff
@@ -240,3 +243,120 @@ func createTestFS(t *testing.T, filenames []string) fstest.MapFS {
...
 	return fs
 }
+
+func TestExtractToken(t *testing.T) {
+	testCases := []struct {
+		name          string
+		contextKey    string
+		contextValue  string
+		expectedToken string
+		expectedErr   string
```
absoludity, line 253: You may also be able to use the `errors` package to instead define an error type here (ie. nil, for no error, or an actual error if it is, then use [errors.Is](https://golang.org/pkg/errors/#Is). But fine as is.

cmd/kubeapps-apis/server/plugins_test.go
```diff
@@ -240,3 +243,120 @@ func createTestFS(t *testing.T, filenames []string) fstest.MapFS {
...
+	}{
+		{
+			name:          "Good token",
```
absoludity, line 256: So this function is an easy one to test - great. I try to make the name of the test follow the test spec, like "it returns the expected token without error for a valid context value", but that's just a personal preference.

cmd/kubeapps-apis/server/plugins.go
```diff
@@ -232,3 +253,140 @@ func listSOFiles(fsys fs.FS, pluginDirs []string) ([]string, error) {
 	}
 	return matches, nil
 }
+
+// parseClusterConfig returns a kube.ClustersConfig struct after parsing the raw `clusters` object provided by the user
+// TODO(agamez): this fn is the same as in kubeapps/cmd/kubeops/main.go, export it and use it instead (unit test available at: cmd/kubeops/main_test.go)
```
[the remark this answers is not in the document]
antgamdia: But let's do it in another PR, no? I already have it planned once we got this PR landed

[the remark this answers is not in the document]
absoludity: Yes, I didn't mean it needs to be moved here.

cmd/kubeapps-apis/server/plugins_test.go
antgamdia: Yup, totally right. I also like having meaningfulñ names for the tests, but I was too focused on writing them that I forgot to edit them before pushing. Thanks for the heads up!

cmd/kubeapps-apis/server/plugins_test.go
antgamdia: I've changed it to `fmt.Errorf(...)` and `status.Errorf` (passing the expected `codes.Unauthenticated` instead of relying on the string msg). Thanks!

cmd/kubeapps-apis/server/plugins_test.go
```diff
@@ -240,3 +243,120 @@ func createTestFS(t *testing.T, filenames []string) fstest.MapFS {
...
+						PinnipedConfig: kube.PinnipedConciergeConfig{
+							Enable: true,
+						},
+						IsKubeappsCluster: true,
+					},
+				},
+			}
+			dynamicInterface, err := dynClientGetterForContextWithConfig(context, inClusterConfig, serveOpts, config)
+
+			if tc.shouldCreate != (dynamicInterface != nil) {
```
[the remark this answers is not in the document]
antgamdia: Exactly, I initally was trying to -use- the dynamic interface, but it was more tricky than expected, so I just checked if the creation was being performed. 

I've added a //TODO there, but I'd rather leave the current test. At least we are checking the closure fn is being called with the proper params so that it doesn't throw any error.

antgamdia:
I don't know why, but the e2e tests are not passing because of a CI issue :sob: . Why it's happening here? Dunno, the changes herein are totally unrelated... I've sent a fix in #2950

cmd/kubeapps-apis/server/plugins.go
```diff
@@ -232,3 +253,150 @@ func listSOFiles(fsys fs.FS, pluginDirs []string) ([]string, error) {
...
+	return func(ctx context.Context) (dynamic.Interface, error) {
+		var err error
+		token, err := extractToken(ctx)
+		if err != nil {
+			return nil, status.Errorf(codes.Unauthenticated, "invalid authorization metadata: %v", err)
+		}
+
+		var client dynamic.Interface
+		if !serveOpts.UnsafeUseDemoSA {
+			restConfig, err := kube.NewClusterConfig(inClusterConfig, token, "default", config)
```
absoludity, line 347: Guessing you didn't mean to hard-code the "default" cluster here?

absoludity (approved):
Great work Antonio! Small note inline about hard-coding the "default" cluster, but otherwise +1

cmd/kubeapps-apis/server/plugins.go
antgamdia: Oops, leftover I didn't notice! It should retrieve the cluster from `ClustersConfig.KubeappsClusterName` (since we are not using cluster-scoped endpoints yet, afaik). Changing it straight away. Thanks!

antgamdia:
> I don't know why, but the e2e tests are not passing because of a CI issue  . Why it's happening here? Dunno, the changes herein are totally unrelated... I've sent a fix in #2950

Well, it was not the CI who blame 100%, but a minor kind-of syntax bug (always forget Helm doesn't have lazy eval...). The issue is that it should have been caught by the `test chart render` CI step.
I've sent another PR addressing it #2957 , but now this one is passing finally. So... let's merge! (and create the one for extracting the shared function to `/pkg`)

Thanks again for your valuable suggestions and ideas!

Issue #2864, opened 2021-05-24
[kubeapps-apis] Investigate provision of chart Readme and values


absoludity:
### Description:

The fluxv2 backend that we'll be using for the plugin does cache the index.yaml for a chart repository nicely, but given its gitops use-case, does not extract or otherwise provide access to other metadata usually presented to users when viewing details of a chart for potential installation. We want to investigate possible options here, such as extracting and caching when a specific chart (version) details are requested, rather than having to extract all and store in a db like is currently done for the existing helm support.

@gfichtenholt Feel free to edit and fill in the details here, I'm just creating this so we can include it in the iteration plan (assuming it's going to be something you'll be working on).

gfichtenholt:
yes, right on. I'll take it

gfichtenholt:
Here is what I found out so far. Fluxv2 has two different CRDs: HelmChart and HelmRelease. One can create an instance of HelmChart in order to fetch (or pull) the chart contents locally to the cluster. E.g. having previously deployed 'bitnami' HelmRepository, I can create this object 
apiVersion: source.toolkit.fluxcd.io/v1beta1
kind: HelmChart
metadata:
  name: redis
  namespace: default
spec:
  chart: redis
  version: 10.5.x
  sourceRef:
    name: bitnami
    kind: HelmRepository
  interval: 10m
and I will end up with 
$ kubectl get helmchart
NAME    CHART   VERSION   SOURCE KIND      SOURCE NAME   READY   STATUS                      AGE
redis   redis   10.5.x    HelmRepository   bitnami       True    Fetched revision: 10.5.14   2m50s
and similarly to how the helmrepository object caches the index.yaml, the .tgz file for the chart will be cached (and updated at specified interval) locally in the cluster 
$ kubectl get helmchart/redis -o yaml
... skipping a lot of yaml
status:
  artifact:
    checksum: 440b4dd25710c3299bac0def9fd694ddfe684188
    lastUpdateTime: "2021-05-27T04:39:27Z"
    path: helmchart/default/redis/redis-10.5.14.tgz
    revision: 10.5.14
    url: http://source-controller.flux-system.svc.cluster.local./helmchart/default/redis/redis-10.5.14.tgz
  conditions:
  - lastTransitionTime: "2021-05-27T04:39:25Z"
    message: 'Fetched revision: 10.5.14'
    reason: ChartPullSucceeded
    status: "True"
    type: Ready
  observedGeneration: 2
[...]

PR #2915, opened 2021-05-31
Get package details prototype

gfichtenholt:
### Description of the change

A prototype for fluxv2 plugin's feature that returns package details

### Applicable issues

[kubeapps-apis] Investigate provision of chart Readme and values #2864


cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
```diff
@@ -161,3 +166,29 @@ message GetPackageRepositoriesResponse {
...
+  PackageRepositoryReference repository = 2;
+}
+
+message GetPackageMetaRequest {
+  PackageReference package = 1;
+}
+
+message GetPackageMetaResponse {
+  message PackageMeta {
+    string readme = 1;
```
absoludity, line 191: I was thinking about this field - Helm charts have a readme, but a more general option might be `longDescription`, noting that it can be markdown, but we can discuss that in the design doc soon.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
```diff
@@ -172,6 +186,61 @@ func (s *Server) GetAvailablePackages(ctx context.Context, request *corev1.GetAv
...
+
+	newChart, err := client.Resource(chartsResource).Create(ctx, &unstructuredChart, metav1.CreateOptions{})
+	if err != nil {
+		log.Errorf("error creating chart: %v\n%v", err, unstructuredChart)
+		return nil, err
+	}
+
+	log.Infof("created chart: [%v]", newChart)
+
+	return nil, status.Errorf(codes.Unimplemented, "not implemented yet")
```
absoludity, line 241: Guessing you plan at this point to wait for the chart to reconcile before returning the metadata, though I don't see it on the CR (https://fluxcd.io/docs/components/source/helmcharts/#specification) so we'd need to parse the chart itself, right? (which is cached by fluxv2?)

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
gfichtenholt: Michael, this is not ready for review in any shape. Its a draft just to see what it costs to make the "electrons flow". When I am ready for review the interface will be much more generic and not look like that. So no need to worry

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
gfichtenholt: please hold comments until I mark it ready for review :-)

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
absoludity: Ah sorry... I thought you'd created a draft PR because it wasn't ready for review but wanted some feedback (something I often do). No probs, I'll not look at something until you request it :)

gfichtenholt:
I marked it ready for review. To be clear on what this is - just a prototype piece of
code that shows a few things like creating a HelmChart CRD, waiting until its fully ready, unpacking the helm chart etc, i.e. the *general* direction in which I will implement GetAvailablePackageDetail() API
I wanted to push this code before I start a re-factor to conform to the latest design spec
because in addition to the above, it also has some general cleanup/improvements and I wanted 
to make sure those are not lost. So don't worry about the API name, etc - that will all be 
changed to conform to spec as soon as we finalize it

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
```diff
@@ -40,6 +37,24 @@ const (
 	fluxHelmRepository     = "helmrepository"
 	fluxHelmRepositories   = "helmrepositories"
 	fluxHelmRepositoryList = "HelmRepositoryList"
+	fluxHelmChart          = "HelmChart"
+	fluxHelmCharts         = "helmcharts"
+)
+
+// these should really be constants, rather than global vars
+// but alas, go does not allow const structs (why???)
```
absoludity, line 45: Because an instance of a struct is not a compile-time object, therefore, not really constant. I assume the reason for that is for the speed of the compiler (it does not need to check whether code is updating things that are declared as constants).

I'd personally prefer not to set a precedent of using global vars at all, and instead just instantiate the GroupVersionResource using the const strings that you already have.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
```diff
@@ -172,6 +201,130 @@ func (s *Server) GetAvailablePackages(ctx context.Context, request *corev1.GetAv
 	}, nil
 }
 
+// GetPackageMeta streams the package metadata based on the request.
+func (s *Server) GetPackageMeta(ctx context.Context, request *corev1.GetPackageMetaRequest) (*corev1.GetPackageMetaResponse, error) {
```
absoludity, line 205: I think this is what you were referring to, that I should ignore? As in, this will later be changed to `GetAvailablePackageDetail` or something? If so, why not just add it as we plan it to be (just this one call)?

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
```diff
@@ -172,6 +201,130 @@ func (s *Server) GetAvailablePackages(ctx context.Context, request *corev1.GetAv
...
+	}, nil
+}
+
+func (s *Server) pullChartTarball(ctx context.Context, packageRef *corev1.PackageReference) (*string, error) {
+	client, err := s.GetClient(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	resourceIfc := client.Resource(chartsResource).Namespace("default")
```
absoludity, line 233: Why is the namespace hard-coded here? Fine if it's just for demoning right now.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
```diff
@@ -172,6 +201,130 @@ func (s *Server) GetAvailablePackages(ctx context.Context, request *corev1.GetAv
...
+
+	// see if we the chart already exists
+	chartList, err := resourceIfc.List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO it'd be better if we could filter on server-side
```
absoludity, line 241: You should be able to use the `metav1.ListOptions{}` above to specify filtering using the [`FieldSelector`](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#ListOptions). More info at https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/ . You may even be able to specify that the pull should be complete to be included (ie. that status conditions ready is true), not sure, but that'd be nice.

Again, at this point it's fine if you just want to update your comment to link to that and do it in a later PR.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
```diff
@@ -172,6 +201,130 @@ func (s *Server) GetAvailablePackages(ctx context.Context, request *corev1.GetAv
...
+	if err != nil {
+		log.Errorf("error creating watch: %v\n%v", err, unstructuredChart)
+		return nil, err
+	}
+
+	return waitUntilChartPullComplete(watcher)
+}
+
+func waitUntilChartPullComplete(watcher watch.Interface) (*string, error) {
+	ch := watcher.ResultChan()
```
absoludity, line 301: Nice - great use of both channel and the watcher to handle the async request. Note that this effectively makes `pullChartTarball` a blocking call, which is fine in it's current use (where it's only called in a context of one chart). Not for now, but in the future you might instead want to consider something like passing a results channel (of string urls) to pullChartTarball, so it returns immediately and you wait on the results channel at the call-site, which would mean you could call it for 20 different charts and just wait for the results to come in whatever order they happen to take, rather than serially.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
```diff
@@ -247,6 +392,34 @@ func isRepoReady(repoUnstructured *unstructured.Unstructured) (bool, error) {
 	return false, nil
 }
 
+//
+// TODO the semantics of this really should be do we need to keep polling or not
+//
+func isChartPullComplete(unstructuredChart *unstructured.Unstructured) (bool, error) {
```
absoludity, line 398: As above, hopefully this fn isn't required if we can only list charts that we know are ready.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/utils.go
```diff
@@ -0,0 +1,134 @@
+/*
```
absoludity, line 1: Is the functionality in this utils file already available in the existing codebase? If so, I'd be keen to import and re-use that?

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
```diff
@@ -22,6 +22,11 @@ service PackagesService {
       get: "/core/packages/v1alpha1/packagerepositories"
     };
   }
+  rpc GetPackageMeta(kubeappsapis.core.packages.v1alpha1.GetPackageMetaRequest) returns (kubeappsapis.core.packages.v1alpha1.GetPackageMetaResponse) {
```
absoludity, line 25: Can we just use `GetInstalledPackageDetails`? Fine if not, we'll be updating it soon.

absoludity (approved):
Nice work Greg. I'm OK if you want to land this as is, slight preference for at least using the endpoint name that we're going to move to, but you may have reasons to prefer to land and continue. A few comments in line.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
gfichtenholt: ok, sounds good. I'll change this. 
FWIW, there was a lot of debate on the subject of const structs in go. What I got from it, it is technically possible just requires a lot of work, and there wasn't enough "bang for the buck" (literally an expression someone at google used) to make it happen. bleh

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
gfichtenholt: yeah, this will happen, don't worry about that

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
gfichtenholt: just a placeholder, like I said, a lot of refactoring to be done

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
gfichtenholt: As I remember I did try a few things here. The problem is the set of supported fields in FieldSelector is very small. things like spec.chart are certainly not supported. see https://github.com/kubernetes/client-go/issues/713 and https://github.com/flant/shell-operator/blob/8fa3c3b8cfeb1ddb37b070b7a871561fdffe788b/HOOKS.md#fieldselector. Nevertheless, I made a TODO comment here  just as you suggested to see if I can improve later

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/utils.go
gfichtenholt: yeah, kind of, but its private to asset-syncer. now. Needs to be re-packaged for re-used. I will put a TODO item to see if I can clean it up

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
gfichtenholt: one thing at a time :-). It will probably be in my next PR

PR #2926, opened 2021-06-02
Move cm used by kubeops and kubeappsapis to a shared folder

antgamdia:
### Description of the change

This PR  just moves the ConfigMap used by kubeops to a shared folder and performs the proper renaming.

### Benefits

Since this cm holds the information regarding the `clusters` section in the values.yaml, it is susceptible to be shared among other components (such as the upcoming kubeappsapis).

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

See https://github.com/kubeapps/kubeapps/pull/2908#discussion_r642717781

(Holding the merge until we got https://github.com/kubeapps/kubeapps/pull/2908 merged)


absoludity (commented):
Not sure this will even be necessary, as you're adding the config to `shared/` in your other PR, so it'll just be updating the kubeops deployment to use that instead? Anyway, let's see once your other PR lands.

antgamdia:
> Not sure this will even be necessary, as you're adding the config to `shared/` in your other PR, so it'll just be updating the kubeops deployment to use that instead? Anyway, let's see once your other PR lands.

Yes, you're right. I just wanted to make them explicit, but, as you pointed out, many of them are already superseded by the #2851. I've just pushed the new changes here (modifying kubeops and bumping the chart version).

absoludity (approved):

PR #2957, opened 2021-06-08
Fix Helm template CI check

antgamdia:
### Description of the change

Follow up of #2950. After analyzing the root cause (besides the lack of enough time to terminate/create the pod), it was a silly issue with the template (I always forget Helm doesn't have lazy evaluation...). The problem is that our CI was unable to catch this error since Helm 3.1 wasn't returning any information or errors.  

### Benefits

This PR performs two changes: a) minor change in the logic introduced in #2950 (silly typo in the counter variable name...); b) installing a new helm version (latest stable) available at `helm-stable` just for performing an additional `helm template` check to avoid these errors. 

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

Example of the former bug being caught: https://app.circleci.com/pipelines/github/antgamdia/kubeapps/590/workflows/ee637b9f-8426-4db2-b87c-78eb1ef184d5/jobs/4006

### DIFF
--- a/script/libtest.sh
+++ b/script/libtest.sh
@@ -56,8 +56,8 @@ k8s_wait_for_deployment() {
       exit_code=0;
       until $rolloutCmd || [ $attempts -eq $extraAttempts ]; do
         $rolloutCmd 
-        extraAttempts=$((attempts + 1))
-        info "Retrying after $extraAttemptsWait..."
+        attempts=$((attempts + 1))
+        info "$attempts/$extraAttempts - Retrying after $extraAttemptsWait s..."
         sleep $extraAttemptsWait
       done
     fi
--- a/script/libtest.sh
+++ b/script/libtest.sh
@@ -51,11 +51,11 @@ k8s_wait_for_deployment() {
         sleep $retriesWait
         retries=$((retries - 1))
     done
-  if [ $retries == 0 ]; then
-      info "Error while rolling out deployment ${deployment} in ns ${namespace}"
-      exit 1
-  fi
-   return $exit_code
+    if [ $retries == 0 ]; then
+        info "Error while rolling out deployment ${deployment} in ns ${namespace}"
+        exit 1
+    fi
+    return $exit_code
 }
 
 ########################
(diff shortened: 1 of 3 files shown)


script/libtest.sh
```diff
@@ -38,30 +38,24 @@ k8s_wait_for_deployment() {
...
+        fi
+        info "Attempt failed, retrying after ${retriesWait}... (remaining attempts: ${retries})"
+        sleep $retriesWait
+        retries=$((retries - 1))
+    done
+  if [ $retries == 0 ]; then
+      info "Error while rolling out deployment ${deployment} in ns ${namespace}"
+      exit 1
+  fi
+   return $exit_code
```
absoludity, line 58:
```suggestion
    if [ $retries == 0 ]; then
        info "Error while rolling out deployment ${deployment} in ns ${namespace}"
        exit 1
    fi
    return $exit_code
```

Just alignment with the existing code? I assume the misalignment wasn't intentional?

absoludity (approved):
>  installing a new helm version (latest stable) available at `helm-stable` just for performing an additional `helm template` check to avoid these errors.

I'm a little confused in that I see you now additionally template with `helm-stable`, but the `helm template` command  using the minimum version is still there, which was what was erroring no? How does this avoid the error? (I see it does pass below).

antgamdia:
> I'm a little confused in that I see you now additionally template with `helm-stable`, but the `helm template` command using the minimum version is still there, which was what was erroring no? How does this avoid the error? (I see it does pass below).

Well, let me explain myself better: the error in the template was -not- being detected by the Helm version we are using in the rest of the CI steps (that is, the minimum supported Helm version). Once we use the newest stable version, the error does get detected.
The root cause is that this old version just ignored the wrong file and simply exited with "no files to template" with a 0 exit code, so CircleCI assumed the check was passing.

So, in order to preserve the former behavior (and since I'm not aware of the changes between helm versions), I just decided to render the template twice (with the old version and the stable one).

An alternative is just using the latest stable version, but I cannot guarantee we are ensuring compatibility with the Helm minimum version.

Thx for the formatting suggestion!

absoludity:
> Well, let me explain myself better: the error in the template was -not- being detected by the Helm version we are using in the rest of the CI steps

Thanks, yes - when we chatted about this in the standup it then made sense - I meant to come back and say so here :) Still +1.

PR #2964, opened 2021-06-09
Minor apidocs update

antgamdia:
### Description of the change

After #2908, the API doc must reflect the need for user credentials and 401 status code (at least). Also, I've noticed that running `buf lint` we got some warnings, I've also addressed them (as they are related to the apidocs)

### Benefits

API docs a little bit more consistent with the (eventual) production config.

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

N/A


cmd/kubeapps-apis/proto/kubeappsapis/apidocs/v1/apidocs.proto
```diff
@@ -1,5 +1,5 @@
 syntax = "proto3";
-package kubeappsapis.apidocs;
+package kubeappsapis.apidocs.v1;
```
absoludity, line 2: We should similarly be using v1alpha1 here so we don't find ourselves caught out later (which means moving the directory).

absoludity (approved):

cmd/kubeapps-apis/proto/kubeappsapis/apidocs/v1/apidocs.proto
antgamdia: Ah, that's totally true... I just used the suggested one by the linter, but, you're right, better just use the same `v1alpha1`. Will update it shortly!

absoludity: Just note, I think you only renamed the package name to `v1alpha1`, so `buf lint` will probably complain that the file should be in the directory `cmd/kubeapps-apis/proto/kubeappsapis/apidocs/v1alpha1/apidocs.proto` . But we can just squeeze that in with another PR. In the future we can ensure that `buf lint` is required to exit 0 when building the image so will be caught by CI.

antgamdia: >  In the future we can ensure that `buf lint` is required to exit 0 when building the image so will be caught by CI.

Oops, that's true. We certainly need adding at list a linter step in the CI, I've added the card for the next iteration discussion. Thanks for pointing this out!

Issue #2945, opened 2021-06-07
[kubeapps-apis] Update the messages in each operation 


antgamdia:
Update the existing kubeapps-apis demo code to use the messages from the design doc (ie. Context, AvailablePackageReference etc. etc.).

Note that the message definition is an ongoing task, so the updates in the code will appear after we reach an agreement on the design doc.

absoludity:
@antgamdia We've just agreed on the naming of `AvailablePackage` and `InstalledPackage`, so feel free to start this. I'd suggest the following (so that no-one is blocked):

1. Update the existing messages and server in the core.packaging.v1alpha1 package and generate. Propose and land this even though it breaks the two existing plugins (they won't build)
2. Then kapp-controller and fluxv2 can be updated independently, whether you do both or whether greg updates fluxv2.

Just note that the kapp-controller plugin is still on a previous alpha version (so may be tricky to test IRL unless we've access to a demo repo from the same alpha version), but just updating the tests is fine for now, imo, until we later updated to something more stable (if it's worthwhile).

antgamdia:
> Then kapp-controller and fluxv2 can be updated independently, whether you do both or whether greg updates fluxv2.

Initial draft at #2945

antgamdia:
Moving to "committed" to continue working on that once we agree on the new names, but I'm also ok closing it and creating a new issue for each batch of renames

antgamdia:
Closing as we already performed the changes in the already implemented logic. Remaining things in the design doc haven't been implemented yet.

PR #2965, opened 2021-06-09
[kubeapps-apis] Rename proto messages

antgamdia:
### Description of the change

This PR changes the `proto` files with the current version of the different messages and rpc calls. It also performs the rename on the `core` apis, but not in the plugins (it will be addressed in an additional PR or by the plugin developers)

### Benefits

We adapt our messages to the agreed ones.

### Possible drawbacks

The current code for the plugins will stop working/compiling

### Applicable issues

  - related #2945

### Additional information

More info at https://github.com/kubeapps/kubeapps/issues/2945#issuecomment-856384692

Also, FWIW, I've been doing, locally, the renaming thing in the actual code: https://github.com/antgamdia/kubeapps/commit/55c78d69a468aca0e636002301cb4f08dff5b8df


cmd/kubeapps-apis/proto/kubeappsapis/core/plugins/v1alpha1/plugins.proto
```diff
@@ -11,24 +11,17 @@ import "protoc-gen-openapiv2/options/annotations.proto";
...
-  // Plugin name
-  //
-  // The name of the plugin, such as `fluxv2.packages` or `kapp_controller.packages`.
-  string name = 1;
-
-  // Plugin version
-  // 
-  // The version of the plugin, such as v1alpha1
-  string version = 2;
+service PluginsService {
```
absoludity, line 14: I thought the doc was only the core packaging api currently, but I guess you're re-ordering things here for consistency?

cmd/kubeapps-apis/proto/kubeappsapis/plugins/fluxv2/packages/v1alpha1/fluxv2.proto
```diff
@@ -7,10 +7,16 @@ import "kubeappsapis/core/packages/v1alpha1/packages.proto";
 import "protoc-gen-openapiv2/options/annotations.proto";
 
 service FluxV2PackagesService {
-  // GetAvailablePackages returns the available packages managed by the 'fluxv2' plugin
-  rpc GetAvailablePackages(kubeappsapis.core.packages.v1alpha1.GetAvailablePackagesRequest) returns (kubeappsapis.core.packages.v1alpha1.GetAvailablePackagesResponse) {
+  // GetAvailablePackageSummaries returns the available packages managed by the 'fluxv2' plugin
+  rpc GetAvailablePackageSummaries(kubeappsapis.core.packages.v1alpha1.GetAvailablePackageSummariesRequest) returns (kubeappsapis.core.packages.v1alpha1.GetAvailablePackageSummariesResponse) {
     option (google.api.http) = {
-      get: "/plugins/fluxv2/packages/v1alpha1/packages"
+      get: "/plugins/fluxv2/packages/v1alpha1/availablepackagesummaries"
```
absoludity, line 13: Sorry, yes I had a TODO to add http endpoints to the doc but hadn't yet addressed it yet. They're easy to change though for now (while we don't have a client :) ). These are fine for now.

cmd/kubeapps-apis/proto/kubeappsapis/plugins/fluxv2/packages/v1alpha1/fluxv2.proto
```diff
@@ -7,10 +7,16 @@ import "kubeappsapis/core/packages/v1alpha1/packages.proto";
...
+    };
+  };
+  // GetAvailablePackageDetail returns the package metadata managed by the 'fluxv2' plugin
+  rpc GetAvailablePackageDetail(kubeappsapis.core.packages.v1alpha1.GetAvailablePackageDetailRequest) returns (kubeappsapis.core.packages.v1alpha1.GetAvailablePackageDetailResponse) {
+    option (google.api.http) = {
+      get: "/plugins/fluxv2/packages/v1alpha1/availablepackagedetails"
```
absoludity, line 19:
```suggestion
      get: "/plugins/fluxv2/packages/v1alpha1/availablepackagedetail"
```
maybe?

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
```diff
@@ -12,124 +12,249 @@ import "protoc-gen-openapiv2/options/annotations.proto";
...
+  //   };
+  // }
 }
-//
+
 // Standard request and response messages for each required function are defined
 // below for import and re-use.
 
-// GetAvailablePackages
+// -- Start definitions of the request messages --
```
absoludity, line 68: Wonder if it would be simpler to have the request and response together for each?

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
```diff
@@ -12,124 +12,249 @@ import "protoc-gen-openapiv2/options/annotations.proto";
...
+  // Optional specific version (or version reference) to request.
+  // By default the latest version (or latest version matching the reference)
+  // will be returned.
+  string version = 2;
+}
+
+// GetPackageRepositories
+//
+// Request for GetPackageRepositories
+message GetPackageRepositoriesRequest {
```
absoludity, line 96: I'm not sure if we actually want/need any CRUD API for repositories right now? We implemented this to check a few things, but I'm wanting to focus solely on the packaging API for now.

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
```diff
@@ -12,124 +12,249 @@ import "protoc-gen-openapiv2/options/annotations.proto";
...
+     };
+   }
+  rpc GetAvailablePackageDetail(GetAvailablePackageDetailRequest) returns (GetAvailablePackageDetailResponse) {
     option (google.api.http) = {
-      get: "/core/packages/v1alpha1/packages"
+      get: "/core/packages/v1alpha1/availablepackagedetails"
     };
   }
-  rpc GetPackageRepositories(kubeappsapis.core.packages.v1alpha1.GetPackageRepositoriesRequest) returns (kubeappsapis.core.packages.v1alpha1.GetPackageRepositoriesResponse) {
+  rpc GetPackageRepositories(GetPackageRepositoriesRequest) returns (GetPackageRepositoriesResponse) {
```
absoludity, line 25: I'm keen to remove this from the core packages API (that is, any CRUD for repositories). It's no problem if a plugin defines it (we did add them to test things the other week), but when we finish the packages API, I expect we'll design and add a separate one for CRUD actions on repositories.

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
```diff
@@ -12,124 +12,249 @@ import "protoc-gen-openapiv2/options/annotations.proto";
...
+  // A short description of the app provided by the package
+  string short_description = 5;
+}
+
+// AvailablePackageDetail
+//
+// An AvailablePackageDetail provides additional details required when
+// inspecting an individual package.
+message AvailablePackageDetail {
+  // TODO: add example for API docs
```
absoludity, line 192: Does the example for API docs make sense when the entity is a read-only returned entity (ie. the user never constructs this message).

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
```diff
@@ -12,124 +12,249 @@ import "protoc-gen-openapiv2/options/annotations.proto";
...
+  //
+  // A plugin can define further key-values for data which does not (yet) fit
+  // with the core.packaging.AvailablePackageDetail.
+  map<string, bytes> opaque_data = 10;
 }
 
 // PackageRepository
 //
 // A PackageRepository defines a repository of packages for installation.
 message PackageRepository {
```
absoludity, line 253: Hmm, we need this for references do we? Fine to leave this one if so... we can move it later as needed. Actually, there's a PackageRepositoryReference below, so perhaps not?

absoludity (commented):
Thanks Antonio. Some comments in line.

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
absoludity: Just confirming after having gone through the doc (and updating a little from some discussion), let's delete `GetPackageRepositories` from this file, as well as the `PackageRepository` message (which will no longer be needed, more below).

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
absoludity: Yep, delete this and the related response.

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
```diff
@@ -12,124 +12,249 @@ import "protoc-gen-openapiv2/options/annotations.proto";
...
+  AvailablePackageDetail available_package_detail = 1;
 }
 
-// GetAvailablePackages
+// GetPackageRepositories
 //
-// Response for GetAvailablePackages
-message GetAvailablePackagesResponse {
+// Response for GetPackageRepositories
+message GetPackageRepositoriesResponse {
```
absoludity, line 137: Delete this one too pls.

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
```diff
@@ -12,124 +12,249 @@ import "protoc-gen-openapiv2/options/annotations.proto";
...
+  // };
 
-  // A namespace can be provided if the package repositories for a specific namespace
-  // are requested, when supported by the packaging format.
-  string namespace = 2;
+  // Available package reference
+  //
+  // A reference uniquely identifying the package.
+  AvailablePackageReference available_package_ref = 1;
+
```
absoludity, line 164: We'll need to add a:
```
string name = 1;
```
here since the available package ref no longer (necessarily) stores the name, but rather an identifier.

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
```diff
@@ -12,124 +12,249 @@ import "protoc-gen-openapiv2/options/annotations.proto";
...
+  // option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_schema) = {
+  //   example: '{}'
+  // };
 
-  // TODO: Add standard filters. Potentially filter by plugin.
+  // Available package reference
+  //
+  // A reference uniquely identifying the package.
+  AvailablePackageReference package_ref = 1;
+
```
absoludity, line 201: Pls add a
```
string name = 2;
```
here for the same reason.

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
absoludity: Delete this one too pls.

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
```diff
@@ -154,45 +279,100 @@ message PackageRepository {
...
 }
 
-message PackageReference {
-  // required package name should be unique within the scope of specified repository, if specified
-  string name = 1; 
+// PackageRepositoryReference
+//
+// A PackageRepository is identified by a fully qualified name and
+// optionally a specific namespace.
+message PackageRepositoryReference {
```
absoludity, line 309: Delete this one too pls.

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
```diff
@@ -154,45 +279,100 @@ message PackageRepository {
...
+
+  // Available package context
+  //
+  // The context (cluster/namespace) for the package.
+  Context context = 1;
+
+  // Available package name
+  //
+  // The fully qualified name of the package (ie. a unique name in context).
+  string name = 2;
```
absoludity, line 345: so this should now be
```
string identifier = 2;
```
See design doc for example comment (feel free to improve it :) )

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
```diff
@@ -154,45 +279,100 @@ message PackageRepository {
...
+
+  // Available package repository reference
+  //
+  // An optional package repository reference where this package is located. Not
+  // all plugins may support this back reference (eg. kapp-controller)
+  PackageRepositoryReference repository = 3;
```
absoludity, line 351: This also should be removed.

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
```diff
@@ -154,45 +279,100 @@ message PackageRepository {
...
+
+  // Available package plugin
```
absoludity, line 353: It's just a plugin, not a plugin specifically for available packages - but maybe you've another reason, not sure.

absoludity (approved):
So, after going through the doc again with a bit of discussion with Dimitri, we've updated slightly in a way that affects this. I've tried to go through and identify each point explicitly, but basically the available package reference has an "identifier" field instead of a "name" field, which can be unique in the context even when there's no CR (whereas a name field will not necessarily be unique in context). As a result, we no longer need the PackageRepositoryReference on the reference and can remove that entity (as well as anything related to package repositories), and we need to add a "name" field to the available package summary and detail (since before we were relying on the name being available in the reference). Hopefully I've identified all the spots here inline.

I'm leaving a +1 here so you can land when ready, as otherwise it'll begin to block others. Sorry for the changes after you'd started, some churn is going to be hard to avoid.

cmd/kubeapps-apis/proto/kubeappsapis/core/plugins/v1alpha1/plugins.proto
antgamdia: Yep, that's precisely was my intention. Since they aren't functional changes, I assumed it was ok to send them along with this PR

cmd/kubeapps-apis/proto/kubeappsapis/plugins/fluxv2/packages/v1alpha1/fluxv2.proto
antgamdia: Well, neither is really correct indeed, we should move to something more rest-ish, like `/plugins/fluxv2/packages/v1alpha1/availablepackagedetails/{packageID}`
or even `/plugins/fluxv2/packages/v1alpha1/availablepackages/{packageID}/detail`, I don't think?

But, given that the details refer to a specific package, the package id should be part of the URL somehow.

However, we can discuss it in the design doc and change all of them in batch once we reach an agreement.

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
antgamdia: I initially tried that order, but there is a minor downside: the annotations for building the API docs differ depending on whether they are for the request or the response (has a title and an example). So, when mixing responses and requests, one can easily omit an annotation (if using, for instance, the above definition as 'inspiration').

So, the reason for this ordering was just trying to avoid manual errors or omissions,. but, yep, semantically it makes more sense to have them together.

See what you think, no strong opinion on my side :P

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
antgamdia: Just to confirm: you mean removing every reference to `GetPackageRepositories` (in both proto and go code), do you?

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
antgamdia: Well, good point: it is not required or indispensable whatsoever, but, from the API client perspective, I find it really useful to have examples of the returned values. This way is even simpler for clients to mock the returned values when building the API client. 
We are providing the schema, which should be more than enough, but I personally like to provide real examples.

See what you think here, as usual, no strong opinion though.

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
antgamdia: Great, thanks, I'm changing, for consistency with other msg, it to

```proto
  AvailablePackageReference available_package_ref = 1;
  string name = 2;
```

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
antgamdia: It is just used for setting a title in the OAS document. I just thought it was clearer to have something like "_the plugin that is being used for interacting with this available package_" so I just used that name. If it is misleading, I'm ok changing it back to `plugin` or any other name. 

```json
      "plugin": {
          "$ref": "#/definitions/v1alpha1Plugin",
          "description": "The plugin used to interact with this available package.\nThis field should be omitted when the request is in the context of a specific plugin.",
          "title": "Available package plugin"
        }
```

antgamdia:
> (...) Hopefully I've identified all the spots here inline.

Thanks a lot for such a detailed review, really useful. Nonetheless, having already added all the changes in the design doc perhaps is too much going through each change here as well. I'm totally ok and thankful, but, if it tedious, please feel free to just refer me to the doc :P 

> I'm leaving a +1 here so you can land when ready, as otherwise it'll begin to block others. Sorry for the changes after you'd started, some churn is going to be hard to avoid.

No problem! I'm aware there will be (more) changes as we still in a very early stage. Happy to add whichever new change is required. 

WRT landing the PR, there are a couple of comments marked as unresolved (I've resolved a bunch of them to reduce the noise), but I think they are not blockers (or they can be addressed in future PRs)

So, my point here is: I'm not merging atm just in case we wanna change something else. 
But if anyone requires these changes, please feel free to hit the merge button or even jump in, pull the PR branch and add your changes in the same PR, I don't mind at all.

absoludity:
> > (...) Hopefully I've identified all the spots here inline.
> 
> Thanks a lot for such a detailed review, really useful. Nonetheless, having already added all the changes in the design doc perhaps is too much going through each change here as well. I'm totally ok and thankful, but, if it tedious, please feel free to just refer me to the doc :P

Yeah, I would normally, I just felt bad because I'd told you yesterday it was safe to go ahead :/

> WRT landing the PR, there are a couple of comments marked as unresolved (I've resolved a bunch of them to reduce the noise), but I think they are not blockers (or they can be addressed in future PRs)

I'll take a look at the unresolved ones now.

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
absoludity: Yeah, same - no strong opinion. If you've found it safer in this order to ensure the correct annotations, let's do that.

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
absoludity: Yep, if the example is also useful as a response, not just an input, great.

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
absoludity: Yes, in the context of the description, the title makes sense as is. I read "Available package plugin" as "the plugin for available packages" rather than "the plugin that is being used for interacting with this available package". Perhaps "Plugin for the available package" would avoid that? See what you think, fine either way.

cmd/kubeapps-apis/proto/kubeappsapis/core/plugins/v1alpha1/plugins.proto
absoludity: Yep, totally fine, just unexpected at first until I realised you'd just re-ordered.

cmd/kubeapps-apis/proto/kubeappsapis/plugins/fluxv2/packages/v1alpha1/fluxv2.proto
absoludity: > Well, neither is really correct indeed, we should move to something more rest-ish, like `/plugins/fluxv2/packages/v1alpha1/availablepackagedetails/{packageID}`
> or even `/plugins/fluxv2/packages/v1alpha1/availablepackages/{packageID}/detail`, I don't think?

Yes, I'd love to (in another PR) use the latter, but it's going to be tricky as that isn't a unique URL on its own without also including the context in the URL, like:

```
/plugins/fluxv2/packages/v1alpha1/c/default/ns/my-namespace/availablepackages/{packageID}/detail
```

but both the cluster and the namespace can be empty. Actually, we can ensure the cluster is always present easily enough (after the recent work you did to provide the cluster config), so it's just the namespace. I don't know if the protobuf annotations for specifying the URL can cope to additionally serve
;

```
/plugins/fluxv2/packages/v1alpha1/c/default/availablepackages/{packageID}/detail
```
so we can support both namespaced and cluster-scoped packages. Need to check. Otherwise it might need some sucky illegal namespace name (`.../ns/_cluster-scoped/availablepackages/...`). 

> 
> But, given that the details refer to a specific package, the package id should be part of the URL somehow.
> 
> However, we can discuss it in the design doc and change all of them in batch once we reach an agreement.

Yes - sounds good. So back to the suggestion here... I'd only balked at the plural because it read as "available package (details)" (as in details of an available package), but I see now you meant the rest-ish "(available package detail)s" (as in many of the "available package detail"). Let's stick with yours for now as it's at least more correct than the suggestion :)

cmd/kubeapps-apis/proto/kubeappsapis/plugins/kapp_controller/packages/v1alpha1/kapp_controller.proto
```diff
@@ -4,20 +4,78 @@ option go_package = "github.com/kubeapps/kubeapps/cmd/kubeapps-apis/gen/plugins/
...
   // GetPackageRepositories returns the repositories managed by the 'kapp_controller' plugin
-  rpc GetPackageRepositories(kubeappsapis.core.packages.v1alpha1.GetPackageRepositoriesRequest) returns (kubeappsapis.core.packages.v1alpha1.GetPackageRepositoriesResponse) {
+  rpc GetPackageRepositories(GetPackageRepositoriesRequest) returns (GetPackageRepositoriesResponse) {
     option (google.api.http) = {
       get: "/plugins/kapp_controller/packages/v1alpha1/packagerepositories"
     };
   };
 }
+
+// Specific messages used by the 'kapp_controller' plugin
```
absoludity, line 26: Thanks for moving them in here. Even if we delete them soon it's nice for that to happen when the implementation is deleted, rather than with this PR.

absoludity (approved):
See inline. Thanks for all the thought!

cmd/kubeapps-apis/proto/kubeappsapis/plugins/fluxv2/packages/v1alpha1/fluxv2.proto
antgamdia: I'd go with something like `/plugins/fluxv2/packages/v1alpha1/c/{namespace}/availablepackages/{packageID}/detail`  (adding the cluster as a path param, but not the namespace).

The gRPC docs says: 
> Any fields in the request message which are not bound by the path template automatically become HTTP query parameters if there is no HTTP request body.

So I don't foresee any problems in not using the namespace as it can be empty.

> Let's stick with yours for now as it's at least more correct than the suggestion :)

Great, thank you!   Naming is always really tricky; besides, we are `AvailablePackageDetails` is not an easy word. When using RPC naming is just a matter of thinking of operations/actions, but back to the rest world, we should move to resources and use nouns instead of verbs... but sometime these nouns are tricky haha.

antgamdia:
> I'm leaving a +1 here so you can land when ready, as otherwise it'll begin to block others

So, let's land this change now not to block others, but note the `test_go` will fail as the golang code is not compiling. It will be addressed in a follow-up PR

PR #2970, opened 2021-06-10
[kubeapps-apis] Add fallback mechanism to use the local kubeconfig file

### LINKED ISSUE #2945: [kubeapps-apis] Update the messages in each operation

antgamdia:
### Description of the change

This PR adds a fallback mechanism in case that: 1. getting an inCluster fails AND 2. the `--unsafe-use-demo-sa=true` is passed. In that case, it will try to retrieve the kubeconfig file from the KUBECONFIG env var.

### Benefits

Local development will be easier since it isn't required to deploy kubeapps-apis in the cluster, but just connecting to it using a kubeconfig file.

### Possible drawbacks

As it is just a change intended for dev purposes we probably want to remove it in the mid-term.

### Applicable issues

  - fixes #2945

### Additional information

Since it was easy to implement (without considering edge cases, of course, just  a rough impl to easen development), I just go with this fallback mechanism. But also happy to consider other alternatives.

Adding a change for https://github.com/kubeapps/kubeapps/pull/2964#discussion_r649120053


absoludity (approved):
Nice - this not only avoids requiring the clusters config in development, but also makes it trivial to run locally talking to your development k8s api server. Only suggestion I'd have is to simplify the condition by adding a `--unsafe-local-dev-kubeconfig` flag instead of relying on a combination of missing clusters config and a related `--unsafe-user-demo-sa` which is for a different purpose, though related. But up to you.

antgamdia:
> Only suggestion I'd have is to simplify the condition by adding a `--unsafe-local-dev-kubeconfig` flag instead of relying on a combination of missing clusters config and a related `--unsafe-user-demo-sa` which is for a different purpose, though related. But up to you.

Great, I've just added the flag :) 
I just went quickly performing the changes so that others didn't get blocked because of this, but, while then I realized it was easier to add this kubeconfig feature rather than just ignoring the error. Glad to know it can be useful :)

PR #2976, opened 2021-06-13
Move code from asset syncer to pkg

gfichtenholt:
The main goal of this PR is to address one feedback from the last PR I pushed https://github.com/kubeapps/kubeapps/pull/2915, where Michael asked:
"Is the functionality in this utils file already available in the existing codebase? If so, I'd be keen to import and re-use that?" 
So this PR is addressing this by pulling out some (private) code from asset-syncer into a pkg directory where it can be used by both asset-syncer and the fluxv2 plug in, namely:
*  tar utils
* http-client 
* helm/index 
In the future, as proceed with the implementation of fluxv2 plug-in,  I expect to pull out even more code out of asset-syncer into these or new modules. 
Another small change is some code cleanup and an addition of a unit test


pkg/tarutil/tarutil.go
```diff
@@ -10,78 +10,52 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
-
-//
-// TODO some of this functionality already exists in asset-syncer but is private
-// so it needs to be re-packaged so that it can be re-used
-//
-func getHelmIndexFileFromURL(indexURL string) (*helmrepo.IndexFile, error) {
-	log.Infof("+getHelmIndexFileFromURL(%s) 1", indexURL)
-	// Get the response bytes from the url
-	response, err := http.Get(indexURL)
+func FetchDetailFromTarball(name string, chartTarballURL string, userAgent string, authz string, netClient httpclient.Client) (map[string]string, error) {
```
antgamdia, line 30: I'd rename this fn to sth like `FetchChartDetailFromTarball` (to highlight it's just for chart tarballs, not any tarball) since it is now publicly available under the `tarutil` package

pkg/helm/testdata/valid-index.yaml
```diff
@@ -0,0 +1,69 @@
+entries:
```
antgamdia, line 1: Question: are we using files for testing elsewehere? Probably it will make sense due to the nature of the test, but I'm not 100% sure if it's the way to go.

pkg/helm/index.go
```diff
@@ -0,0 +1,77 @@
...
+		copier.Copy(&c.ChartVersions, entry)
+	}
+	c.Repo = r
+	c.Name = url.PathEscape(c.Name) // escaped chart name eg. foo/bar becomes foo%2Fbar
+	c.ID = fmt.Sprintf("%s/%s", r.Name, c.Name)
+	c.Category = entry[0].Annotations["category"]
+	return c
+}
+
+func ChartsFromIndex(contents []byte, r *models.Repo, shallow bool) ([]models.Chart, error) {
```
antgamdia, line 54: In exported (at least) functions I'd like to have some documentation around it. Perhaps this PR is just about moving the existing code, but since it's now public, I think it might be worthwhile to add a couple of lines, no?

https://blog.golang.org/godoc

pkg/http-client/httpclient.go
```diff
@@ -0,0 +1,52 @@
...
+	"fmt"
+	"io/ioutil"
+	"net/http"
+)
+
+type Client interface {
+	Do(req *http.Request) (*http.Response, error)
+}
+
+func Get(url string, cli Client, headers map[string]string) ([]byte, error) {
```
antgamdia, line 25: Idem here about documenting

pkg/tarutil/tarutil.go
antgamdia: Idem here about documenting

pkg/tarutil/test/utils.go
```diff
@@ -0,0 +1,48 @@
...
+	"io"
+
+	log "github.com/sirupsen/logrus"
+)
+
+type TarballFile struct {
+	Name, Body string
+}
+
+func CreateTestTarball(w io.Writer, files []TarballFile) {
```
antgamdia, line 26: Idem here about documenting

antgamdia (commented):
Thank you for splitting up the code! Now the fluxv2 looks simpler :)

pkg/tarutil/tarutil.go
gfichtenholt: will do. Thanks

pkg/helm/testdata/valid-index.yaml
gfichtenholt: yes, this file is also under asset syncer testdata and is used there for a bunch of tests to test asset syncer-specific features that only exist there. If/when I end up moving more code out of asset syncer into pkg utils there maybe a point where there no longer any tests left there using this file, at which point I will delete it. But not yet

pkg/helm/index.go
gfichtenholt: sure will do

pkg/http-client/httpclient.go
gfichtenholt: ok

pkg/tarutil/test/utils.go
gfichtenholt: will do

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
```diff
@@ -214,14 +232,16 @@ func (s *Server) GetPackageMeta(ctx context.Context, request *corev1.GetAvailabl
 	log.Infof("Found chart url: [%s]", *url)
 
 	// unzip and untar .tgz file
-	meta, err := fetchMetaFromChartTarball(request.AvailablePackageRef.Identifier, *url)
+	// TODO: userAgent, authz and netClient w/TLS config similar to asset-syncer utils.initNetClient(),
```
absoludity, line 235: If you can add your gh nick to the todo (ie. `// TODO (gfichtenholz): ...`) it'd be helpful (as the obvious committer from git is a pain to dig for as soon as this is refactored and copied elsewhere). Same for other TODOs in this change.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
```diff
@@ -238,14 +258,21 @@ func (s *Server) pullChartTarball(ctx context.Context, packageRef *corev1.Availa
 		Resource: fluxHelmCharts,
 	}
 
+	// TODO: use the namespace from the request
```
absoludity, line 261: Is this a TODO only because it's a pain to construct a request currently (as we're not yet mapping the namespace to the URL)? Otherwise why not just use request.Namespace or whatever the field is?

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
```diff
@@ -238,14 +258,21 @@ func (s *Server) pullChartTarball(ctx context.Context, packageRef *corev1.Availa
...
+	// https://github.com/kubeapps/kubeapps/pull/2915
+	// @minelson says: You should be able to use the metav1.ListOptions{} above
+	// to specify filtering using the FieldSelector.
+	// More info at https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/ .
 	// You may even be able to specify that the pull should be complete to be included (ie.
 	// that status conditions ready is true), not sure, but that'd be nice.
+	// @gfichtenholt As I remember I did try a few things here. The problem is the set of supported
+	// fields in FieldSelector is very small. things like spec.chart are certainly not supported.
+	// see kubernetes/client-go#713 and https://github.com/flant/shell-operator/blob/8fa3c3b8cfeb1ddb37b070b7a871561fdffe788b/// HOOKS.md#fieldselector. Nevertheless, I made a TODO comment here just as you suggested to see
+	// if I can improve later
```
absoludity, line 275: Personally don't think we need the history of the discussion there, I think it's enough to just state the current state - something along the lines of the ListOptions FieldSelector not supporting the fields required (and link to the relevant issue/pr comment for more context, if it's helpful).

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
```diff
@@ -310,6 +338,16 @@ func (s *Server) pullChartTarball(ctx context.Context, packageRef *corev1.Availa
...
+// TODO:
+// https://github.com/kubeapps/kubeapps/pull/2915
+// @minelson says Nice - great use of both channel and the watcher to handle the
+// async request. Note that this effectively makes pullChartTarball a blocking call,
+// which is fine in it's current use (where it's only called in a context of one chart).
+// Not for now, but in the future you might instead want to consider something like
+// passing a results channel (of string urls) to pullChartTarball, so it returns
+// immediately and you wait on the results channel at the call-site, which would mean
+// you could call it for 20 different charts and just wait for the results to come in
+//  whatever order they happen to take, rather than serially.
```
absoludity, line 350: Same here. For example, state the TODO and link to more context (the gh issue comment).

pkg/helm/index.go
```diff
@@ -0,0 +1,82 @@
...
+	c.ID = fmt.Sprintf("%s/%s", r.Name, c.Name)
+	c.Category = entry[0].Annotations["category"]
+	return c
+}
+
+//
+// given a an array of bytes containg the contents of index.yaml from a helm repo it will return
+// all Chart models. shallow flag controls whether only the latest version of the charts is kept
+// or all versions
+//
```
absoludity, line 58: See the link Antonio provided... it's fine as is, but the normal convention would be for something like

```suggestion
// ChartsFromIndex receives an array of bytes containing the contents of index.yaml from a helm repo and returns
// all Chart models from that index. The shallow flag controls whether only the latest version of the charts is returned
// or all versions.
```

pkg/helm/index.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+	for _, entry := range index.Entries {
+		// note that 'entry' itself is an array of chart versions
+		// after index.SortEntires() call, it looks like there is only one entry per package,
+		// and entry[0] should be the most recent chart version, e.g. Name: "mariadb" Version: "9.3.12"
+		// while the rest of the elements in the entry array keep track of previous chart versions, e.g.
+		// "mariadb" version "9.3.11", "9.3.10", etc. For entry "mariadb", bitnami catalog has
+		// almost 200 chart versions going all the way back many years to version "2.1.4".
+		// So for now, let's just keep track of the latest, not to overwhelm the caller with
+		// all these outdated versions
```
absoludity, line 73: This comment doesn't appear to be related to the `GetDeprecated` below, but perhaps the following expression appending the new chart? But even there, `newChart` receives a `helm.ChartVersions` so not sure which you correctly pass, so unsure where you're using `entry[0]` instead, or whether you can do so here without breaking the current helm support (which expects all versions).

pkg/helm/index.go
```diff
@@ -0,0 +1,82 @@
...
+	var index helmrepo.IndexFile
+	err := yaml.Unmarshal(contents, &index)
+	if err != nil {
+		return nil, err
+	}
+	index.SortEntries()
+	return &index, nil
+}
+
+// Takes an entry from the index and constructs a database representation of the
```
absoludity, line 37:
```suggestion
// Takes an entry from the index and constructs a model representation of the
```

pkg/helm/testdata/valid-index.yaml
absoludity: Yep, until recently it was necessary to create test files like this and ensure your code accepts a director so you can point it to your test data during tests. More recently go added the `fstest` package (https://golang.org/pkg/testing/fstest/) which I've used in cmd/kubeapps-apis/server/plugins_test.go . But agree, not worth changing this at all, as part of the benefit of this actual file test data is that it's an exact copy of a subset of a real index.yaml (rather than something we construct in memory to match).

absoludity (approved):
Thanks Greg. A few comments in line, see what you think but +1 either way (assuming it's not changing current functionality for the existing helm support - wasn't 100% with part of the refactor where you mention just returning the latest entry).

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
gfichtenholt: will do

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
gfichtenholt: Nothing hard about it. Just didn't get to it yet :-) Will happen soon

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
gfichtenholt: ok you got it

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
gfichtenholt: ok

pkg/helm/index.go
gfichtenholt: np, will make that change. Thanks

pkg/helm/index.go
gfichtenholt: disagree on this one. Comment is right after the line 
`for _, entry := range index.Entries {
`. and really applies to the entire block on code, including the if statement` if entry[0].GetDeprecated() {I` immediately following it. I don't believe I broke anything, except possibly the format of the log message. The rest of the code should be exactly as it is today in asset syncer. https://github.com/kubeapps/kubeapps/blob/master/cmd/asset-syncer/utils.go#L801

pkg/helm/index.go
gfichtenholt: ok

pkg/helm/index.go
absoludity: Great. Yes, I was looking at the following line (the one at the same indent), not the previous. The last sentence of the comment sounded as though you'd changed something, but I couldn't identify the change. +1

PR #2977, opened 2021-06-15
Update port name for kubeapps-apis to be clear it includes both grpc and http.

absoludity:
### Description of the change

Just a follow-up from a [discussion on a previous PR](https://github.com/kubeapps/kubeapps/pull/2819#discussion_r634193991).

### Benefits

It is clear from the k8s config that the port is used for both http and grpc.


antgamdia (approved):
Great! Thank you for the change (I've almost forgotten it, haha). Much clearer now.

absoludity:
Yeah, I only remembered because it was in the backlog of my personal todo :P

PR #2982, opened 2021-06-16
Add Azure managed AAD guide

antgamdia:
### Description of the change

We noticed some users were struggling with the configuration of the Azure managed AAD (not the legacy one). Our guides just supported this legacy method, so this PR is to add a clear step by step guide on how to perform this configuration

### Benefits

People will be able to use kubeapps with the existing managed Azure Active Directory they are already using for managing the RBAC in an AKS cluster.

### Possible drawbacks

The guide has been developed after my personal investigation on Azure AKS/ADD, but I haven't much experience with them, so I cannot 100% guarantee it's the right way to go, but, at least, it's a configuration that worked and could help some people.

### Applicable issues

  - fixes #2943

### Additional information

As this guide is growing I wonder if it's better to split it up into different files.

PR #2989, opened 2021-06-16
Add missing step in AAD guide

antgamdia:
### Description of the change

This PR simply adds a missing step I forgot to document (the email claim is not generated by default even in v2 tokens, so you have to manually ask for it).

As suggested by @latiif, I also improved the placeholders to make it clearer what should be replaced. Tack så mycket!

### Benefits

No more ` Error creating session during OAuth2 callback: id_token did not contain an email and profileURL is not defined`

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

Follow-up of https://github.com/kubeapps/kubeapps/pull/2982


absoludity (approved):
Why do we need access to the graphql API?

antgamdia:
> Why do we need access to the graphql API?

We need permissions in the Microsoft Graph API for retrieving the user email. It is not coming along with the id_token v2.
But, let me re-check it again... I noticed I did not include `email` as part of the scope, so perhaps it's just a matter of including it (but no adding additional claims, will see)

PR #2991, opened 2021-06-16
Add dev apis UI with swagger-ui

### LINKED ISSUE #2990: Add swaggerui in kubeapps-apis
As suggested by Greg, it should speed up the development  process

antgamdia:
### Description of the change

This PR brings an easy way to test the API changes just using the generated OpenAPI v2 document. For doing so, this PR is embedding and serving a tuned up swagger-ui index.html.

https://user-images.githubusercontent.com/11535726/122264676-17a6d780-ced8-11eb-9196-816c451b9456.mp4

### Benefits

Make developers happier :) 

### Possible drawbacks

N/A

### Applicable issues

  - fixes #2990

### Additional information

N/A


cmd/kubeapps-apis/server/server.go
```diff
@@ -131,4 +131,23 @@ func gatewayMux() *runtime.ServeMux {
 			},
 		}),
 	)
+
+	// TODO(agamez): remove these '/openapi.json' and '/docs' paths. They are serving a
+	// static 'swagger-ui' dashboard with hardcoded values just intended for develoment purposes.
+	// This docs will eventually converge into the docs already (properly) served by the dashboard
```
absoludity, line 137: It's not clear to me how this `openapi.json` has hard-coded values... isn't it the same `openapi.json` that gets server by the frontend? More generally, why don't we just serve it like this permanently so it works both in dev as well as when deployed on the cluster? (I think a go binary is pretty efficient at serving a static file, but check).

Or is the issue rather the `index.html` (haven't checked the content, too large for the we UI so I assume it's just static webui for openapi)? If so, couldn't we still leave this permanently, but have the frontend handle the same path with the proper docs? So in devel, you get this one by default, but when configured on the cluster, this handler (for `/docs`) is never executed because the frontend already handles the path or similar?

absoludity (approved):

cmd/kubeapps-apis/server/server.go
antgamdia: Thanks for your comments! Summarizing what we discussed offline:

- The `openapi.json` is not an additional file but simply the path in which the usual file (the one generated with `buf generate`) is served. There is no problem in continue using this path in the future. In fact, it even makes more sense (since this component is responsible for its own docs, more cohesion)

- The issue is the `index.html`: I wasn't able to simply use [FileServer](https://golang.org/pkg/net/http/#FileServer)  from `gwmux.Handle` function (which takes a [*runtime.HandlerFunc](https://pkg.go.dev/github.com/grpc-ecosystem/grpc-gateway/v2/runtime#HandlerFunc)) without refactoring a little bit the code (see https://github.com/lstoll/grpc-rest-example/blob/master/go/server/main.go#L63)

So, IMHO this is a reasonable change considering the scope of this PR (just a temporary UI for playing around while this comp is under heavy development)

antgamdia: Merging since the [checks passed](https://app.circleci.com/pipelines/github/antgamdia/kubeapps/658/workflows/06aeb736-5c8d-4891-91b9-c9f992774c00) but wasn't properly notified back to github

PR #3001, opened 2021-06-17
Add latest CR renames in KappController plugin

### LINKED ISSUE #2992: [kubeapps-apis] Implement changes after the renaming in the kapp-controller CRs
The latest CRs in https://github.com/vmware-tanzu/carvel-kapp-controller/blob/develop/alpha-releases/v0.20.0-rc.1.yml
are currently unsupported by the current kubeapps-apis code. We need to adapt the code to make it work again with the latest version.

antgamdia:
### Description of the change

This PR brings the changes in the kappcontroller CRs (v20-rc1) into our plugin code, namely: 1) straightforward CR renames; 2) use namespace as package repositories are now namespace.

### Benefits

kappcontroller plugin will now work with the latest version. Future releases are not expected to include big renaming changes.

### Possible drawbacks

N/A

### Applicable issues

  - fixes #2992 

### Additional information

I've noticed that kapp plugin does not currently satisfy the core packaging API (that's why this PR also adds some changes in the log traces, to clearly show where these traces are from)

~Edit: oops, PR in progress, I totally forgot to update the tests~ :facepalm:


cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server.go
```diff
@@ -40,14 +40,18 @@ import (
...
-	packageGroup     = "package.carvel.dev"
+	packageGroup     = "packaging.carvel.dev"
 	packageVersion   = "v1alpha1"
-	packagesResource = "packages"
+	packageResource  = "PackageInstall"
+	packagesResource = "packageinstalls"
 
 	// See https://carvel.dev/kapp-controller/docs/latest/packaging/#packagerepository-cr
-	installPackageGroup   = "install.package.carvel.dev"
+	installPackageGroup   = "packaging.carvel.dev"
```
absoludity, line 49: We should no longer need the separate `installPackageGroup` and `installPackageVersion`? These were only required because previously the CRs were on a different group.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_test.go
```diff
@@ -155,23 +158,26 @@ func TestGetAvailablePackagesStatus(t *testing.T) {
...
-func packageFromSpec(spec map[string]interface{}) *unstructured.Unstructured {
+func packageFromSpec(version interface{}, spec map[string]interface{}) *unstructured.Unstructured {
 	return &unstructured.Unstructured{
 		Object: map[string]interface{}{
 			"apiVersion": fmt.Sprintf("%s/%s", packageGroup, packageVersion),
-			"kind":       "Package",
+			"kind":       packageResource,
 			"metadata": map[string]interface{}{
-				"name": fmt.Sprintf("%s.%s", spec["publicName"], spec["version"]),
+				"name": fmt.Sprintf("%s.%s", spec["packageRef"].(map[string]interface{})["refName"], version),
```
absoludity, line 167: Hmm, this could panic now right? I see from the way this function is used in setting up test data in structs, that we don't want to return an error here (it'd make it painful to use), but what we can do is update the function to accept a `t *testing.T` as the first arge, then rather than assuming in the line above that the conversion is successful, use the "comma, ok" idiom (see https://golang.org/doc/effective_go#interface_conversions ) to avoid an unnecessary panic in the test code, with something like:

```
pkgRef, ok := spec["packageREf"].(map[string]interface{})
if !ok {
    t.Fatalf("unable to convert %+v to a map[string]interface{}", spec["packageRef"])
}
```

It's a little more verbose, but just good to avoid situations that can panic generally.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_test.go
```diff
@@ -155,23 +158,26 @@ func TestGetAvailablePackagesStatus(t *testing.T) {
...
+			},
 		},
 	}
 }
 
 func packagesFromSpecs(specs []map[string]interface{}) []runtime.Object {
 	pkgs := []runtime.Object{}
 	for _, s := range specs {
-		pkgs = append(pkgs, packageFromSpec(s))
+		pkgs = append(pkgs, packageFromSpec(s["version"], s["spec"].(map[string]interface{})))
```
absoludity, line 180: Same here.

absoludity (approved):
Great, thanks! I'll comment about the interface being satisfied separately.

absoludity:
> I've noticed that kapp plugin does not currently satisfy the core packaging API (that's why this PR also adds some changes in the log traces, to clearly show where these traces are from)

So if we want to ensure that the kapp service implementation satisfies the core packaging API, we can put in a compile-time statement that does exactly that. Try adding the following line to `plugins/kapp_controller/packages/v1alpha1/server.go`:

```go
var _ corev1.PackagesServiceServer = (*Server)(nil)
```

and run your tests again (or anything that causes a compile) to see the error msg:

```
plugins/kapp_controller/packages/v1alpha1/server.go:41:5: cannot use (*Server)(nil) (type *Server) as type "github.com/kubeapps/kubeapps/cmd/kubeapps-apis/gen/core/packages/v1alpha1".PackagesServiceServer in assignment:
        *Server does not implement "github.com/kubeapps/kubeapps/cmd/kubeapps-apis/gen/core/packages/v1alpha1".PackagesServiceServer (missing GetAvailablePackageDetail method)

```

(This was actually by discussed recently on the internal #golang channel, if you want to take a look). The statement just causes the compiler to verify that it can convert a `*Server` to a `corev1.PackagesServiceServer`, without allocating a variable (because we've used the throw-away underscore).

To fix this, without yet having to implement the actual functionality, just update the `KappControllerPackagesService` in `proto/kubeappsapis/plugins/kapp_controller/packages/v1alpha1/kapp_controller.proto` with that, then regenerate, and it'll generate a stub for it (on the `UnimplementedKappControllerPackagesServiceServer` struct which is embedded in our `plugins/kapp_controller/packages/v1alpha1/server.go:Server`) so that the interface is satisfied (what a mouthful, but trying to be explicit so you can understand the details :) ).

Once it's satisfied, I think it's fine to include the above compile-time check since we do want to be sure that the plugin satisfies the core packages interface.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_test.go
antgamdia: Thanks for the tip here! I assumed that, given that we control the tests, these situations wouldn't appear. But, yep, you're right, we better avoid these panic situations,

antgamdia:
> The statement just causes the compiler to verify that it can convert a *Server to a corev1.PackagesServiceServer, without allocating a variable (because we've used the throw-away underscore).

What an interesting approach! I've added this check in both plugins (this way we avoid generating non-compliant plugins while redesigning the messages) and fixed a minor issue preventing the +`AvailablePackageReference` to appear in the core response.

I don't mind splitting these changes into a different PR if you want.
 
> but trying to be explicit so you can understand the details :) 

Thank you for your explanations! Really useful as always :) :hugs:

absoludity:
> I don't mind splitting these changes into a different PR if you want.

No need - happy to land this branch with those changes.

absoludity:
Landing this as I want to try some things dependent on the changes here.

PR #3018, opened 2021-06-21
Fix non-alpha label for port

absoludity:
### Description of the change

I didn't check the label change with a deploy, can't use underscores.

I'm testing this in a deploy now and won't land it unless it works (since we don't have CI testing kubeappsapis yet).

### Benefits

Can deploy in dev again.

### Possible drawbacks

None (assuming it works as advertised).


antgamdia (approved):
I also forgot about the DNS syntax here :S

Issue #2993, opened 2021-06-16
[kubeapps-apis] Implement "direct-Helm" plugin based on the existing features in kubeapps


antgamdia:
In the same way that we already have `kapp-controller` and `fluxv2` plugins (more or less) implemented, it is required to also support the legacy/direct/current Helm approach that kubeapps is using.
Otherwise, existing OCI support would get compromised (AFAIK) unless another plugin does support it.

Edit: FWIW, it will leverage the existing AppRepositories/postgresql logic

Related to https://github.com/kubeapps/kubeapps/issues/2023

#### Transversal aspects
- [x] Add proto and generated files: https://github.com/kubeapps/kubeapps/pull/3020, https://github.com/kubeapps/kubeapps/pull/3039
- [x] Expose the required logic directly from the assetsvc: https://github.com/kubeapps/kubeapps/pull/3036 and https://github.com/kubeapps/kubeapps/pull/3040
- [x] Add filters options: https://github.com/kubeapps/kubeapps/pull/3038
- [x] Expose a normal clientSet instead of the dynamic interface: https://github.com/kubeapps/kubeapps/pull/3043
- [x] Modify proto messages to implement [gRPC recommended pagination](https://cloud.google.com/apis/design/design_patterns#list_pagination): https://github.com/kubeapps/kubeapps/pull/3074

#### Operations already "agreed"
- [x] GetAvailablePackagesSummaries: https://github.com/kubeapps/kubeapps/pull/3022
- [x]  GetAvailablePackageDetail: https://github.com/kubeapps/kubeapps/pull/3034
- [x] Pagination for GetAvailablePackageSummaries #3091 
- [ ]  GetAvailablePackageVersions: _in progress_ 

#### Operations in active discussion
- [ ]  GetInstalledPackageSummaries
- [ ]  GetInstalledPackageDetail
- [ ]  CreateInstalledPackage
- [ ]  UpdateInstalledPackage
- [ ]  DeleteInstalledPackage

#### UI

Tracked at https://github.com/kubeapps/kubeapps/issues/3085

antgamdia:
_Originally posted at https://github.com/kubeapps/kubeapps/pull/3022#issuecomment-870735907_

#### Additional fields for this method

Currently, the client requires the following params  for displaying the catalog: `icon, name, repo, version, description, namespace, id`

A response of this APIs is something like:

```json
{
  "availablePackagesSummaries": [
    {
      "availablePackageRef": {
        "context": {
[...]

Issue #2946, opened 2021-06-07
[kubeapps-apis] Implement packaging related functionality for fluxv2 plugin


absoludity:
Once #2944 is ready, we want to implement the packaging API (only, ie. not yet api functionality for repositories) for the fluxv2 plugin (as well as one other plugin, but perhaps that should be a separate issue - see #2993 for the direct helm plugin).

This will involve something similar to (subject to name changes):

- [x] GetAvailablePackagesSummaries
- [x] GetAvailablePackageDetail
- [x] GetAvailablePackageVersions
- [x] GetInstalledPackageSummaries
- [x] GetInstalledPackageDetail
- [x] CreateInstalledPackage
- [x] UpdateInstalledPackage
- [ ] DeleteInstalledPackage

absoludity:
I think we've agreed here that we'll just do the first 3 initially. Also, as part of this Greg will be looking at providing a caching layer initially specifically to the flux plugin, but generalising it to provide caching to plugins. It may be worth splitting that to a separate issue.

gfichtenholt:
yes, that's correct, that was the agreement. As far as splitting into separate issue  is concerned, doesn't matter to me, if you think that's best, go ahead or if you prefer let me know and I will open one

absoludity:
IMO it's a matter of clearly communicating what we're currently working on. If you think that the cache investigation and work is going to be significant in its own right (as in, something to report progress on at the end of the iteration), then yep, I'd recommend splitting it out, briefly outlining what you expect to be involved, so you can updated it to make it easier for Pepe at the end of the iteration to summarise what's been happening in the world of Kubeapps. You could also just do that here if you want, but the board (which some people use to see what we're doing) won't reflect that of course.

gfichtenholt:
created https://github.com/kubeapps/kubeapps/issues/3032

antgamdia:
I'd have a question (perhaps it was already said, but I don't remember now): is the Fluxv2 plugin going to support rolling back releases (even if it is not natively supported by flux)? Or, given that Flux does not support it, we won't support it?

gfichtenholt:
yes, that should be doable. If I remember correctly, manual rollback of a release was an example given to me by flux authors in a slack discussion in that one has do invoke the underlying helm client to do it.
https://cloud-native.slack.com/archives/CLAJ40HV3/p1629245848190300?thread_ts=1629244025.187100&cid=CLAJ40HV3

antgamdia:
[...]

PR #3022, opened 2021-06-21
[kubeapps-apis] Add direct-helm GetAvailablePackageSummaries

antgamdia:
### Description of the change

This PR just implements the `GetAvailablePackageSummaries` method in the direct-helm plugin.

AFAIK, this method just calls the assetsvc db, no Helm logic is required yet.

~Test pending.~

### Benefits

We will have a brand new `GetAvailablePackageSummaries` operation up and running. 

### Possible drawbacks

N/A

### Applicable issues

  - related #2993
  - depends on https://github.com/kubeapps/kubeapps/pull/3020

### Additional information

~If this approach is good, we should move some files under `pkg`~ No, we shouldn't; we better expose them directly from their current package.


chart/kubeapps/templates/kubeappsapis/deployment.yaml
```diff
@@ -85,6 +85,21 @@ spec:
...
+              value: {{ template "kubeapps.postgresql.fullname" . }}-headless:{{ default "5432" .Values.postgresql.service.port }}
+            - name: ASSET_SYNCER_DB_NAME
+              value: {{ .Values.postgresql.postgresqlDatabase }}
+            - name: ASSET_SYNCER_DB_USERNAME
+              value: postgres
+            - name: ASSET_SYNCER_DB_USERPASSWORD
+              valueFrom:
+                secretKeyRef:
+                  key: postgresql-password
+                  name: {{ include "kubeapps.postgresql.secretName" . }}
```
absoludity, line 102: We should somehow indicate that this config should be temporary only (ie. we don't want to be providing such config for plugins) until we can switch to the caching provided to plugins that will eventually come out of #2946 . You've said something similar when pulling in the utils.go temporarily below.

absoludity (commented):
Sorry, realised this is draft after first comment, so stopped :)

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server_test.go
```diff
@@ -12,3 +12,292 @@ limitations under the License.
...
+				if got, want := availablePackageSummary, tc.expected; !cmp.Equal(got, want, opt1) {
+					t.Errorf("mismatch (-want +got):\n%s", cmp.Diff(want, got, opt1))
+				}
+			}
+		})
+	}
+}
+
+func TestGetAvailablePackageSummaries(t *testing.T) {
+	getter := func(context.Context) (dynamic.Interface, error) {
```
absoludity, line 209: Not sure the k8s client is used currently, but it will be when we add a test ensuring that it 403s for namespaces to which the user doesn't have access, etc.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server_test.go
```diff
@@ -12,3 +12,292 @@ limitations under the License.
...
+				}
+				rows.AddRow(string(chartJSON))
+			}
+			mock.ExpectQuery("SELECT info FROM").
+				WillReturnRows(rows)
+
+			mock.ExpectQuery("^SELECT count(.+) FROM").
+				WillReturnRows(rowCount)
+
+			availablePackageSummaries, err := s.GetAvailablePackageSummaries(context.Background(), &corev1.GetAvailablePackageSummariesRequest{Context: &corev1.Context{}})
```
absoludity, line 288: Currently this is just testing only one case where the namespace is empty? We'd want to ensure that we test the cases:

- specific namespace results a 403 if the user doesn't have access to the ns
- specific namespace results in a query for charts in that ns only otherwise,
- blank namespace results in a query for charts in the kubeapps namespace and any namespace to which the user has access

I think?

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/utils.go
```diff
@@ -0,0 +1,299 @@
...
+	"github.com/kubeapps/kubeapps/pkg/chart/models"
+	"github.com/kubeapps/kubeapps/pkg/dbutils"
+	_ "github.com/lib/pq"
+	log "k8s.io/klog/v2"
+)
+
+// TODO(agamez): logic temporaryly extracted from:
+// /cmd/assetsvc/postgresql_utils.go
+// /cmd/assetsvc/utils.go
+// we can either: export this logic under pkg or re-implement the data access layer with an ORM
```
absoludity, line 33: Why not ensure this functionality is exported in its current location? If that's an issue, keen to know why. If it's not an issue, I'd recommend just doing it as a separate, pre-quel branch, that is:

-  Public interface `assetManager` -> `AssetManager`
-  Public NewPGManager
-  Public methods

In fact, let me see if I can do that for you to save you the time: #3036 (if you agree, just land that and update).

absoludity (commented):
Great work Antonio. I think there's two things missing here:

- existing checks to ensure the user has access to the namespace
- correct handing of a request without the namespace

absoludity:
> Great work Antonio. I think there's two things missing here:
> 
>     * existing checks to ensure the user has access to the namespace
> 
>     * correct handing of a request without the namespace

Actually, thinking more about the last point, perhaps it's enough to just return Unimplemented for now for GetAvailablePackageSummaries without a namespace, given that the current UI doesn't even do this (it was part of ensuring a future possibility with a different workflow). Thoughts?

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -56,3 +77,94 @@ func (s *Server) GetClient(ctx context.Context) (dynamic.Interface, error) {
...
+	}
+
+	log.Infof("+helm GetAvailablePackageSummaries %s", contextMsg)
+
+	namespace := ""
+	if request.Context != nil {
+		if request.Context.Cluster != "" {
+			return nil, status.Errorf(codes.Unimplemented, "Not supported yet: request.Context.Cluster: [%v]", request.Context.Cluster)
+		}
+		if request.Context.Namespace != "" {
```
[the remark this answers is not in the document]
antgamdia: Ooops, just a leftover after copy-n-pasting from the kapp_controller plugin (there, it was initialized to `namespace := globalPackagingNamespace`  instead of  `""`). Thanks for pointing this out

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/utils.go
antgamdia: Thank you for performing that changes. I initially opted to copy-n-paste the required logic just to have a clear idea of what was really necessary and what's not. 
Also, I didn't want to mix things up, so I just thought, ok let's do it in another PR, but I didn't manage to do it yet, haha.

However, after yours, I've created this another PR https://github.com/kubeapps/kubeapps/pull/3040. More details there, but I had to move the files to a different folder and go package so that I was able to import them. Perhaps I'm missing something.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -56,3 +77,94 @@ func (s *Server) GetClient(ctx context.Context) (dynamic.Interface, error) {
...
+			namespace = request.Context.Namespace
+		}
+	}
```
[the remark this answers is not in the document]
antgamdia: Great, good point here. Thanks for pointing to the auth code (I will comment on my code about it). I've added a check that will prevent users from retrieving packages from ns they can't read.
As you mentioned later, the `ns==""` case won't be implemented yet.

antgamdia:
> Actually, thinking more about the last point, perhaps it's enough to just return Unimplemented for now for GetAvailablePackageSummaries without a namespace, given that the current UI doesn't even do this (it was part of ensuring a future possibility with a different workflow). Thoughts?

That's what I was about to say: let's replicate the current behavior as much as possible, so that we speed up the dev process. Later on, we can start thinking about implementing the TODOs.
I've pushed an initial version of the ns check using the existing logic for the can-i method.

Problem: the client we are passing back to the plugins (via clientGetter) is just for dynamic interfaces, That is, we can't query structured/typed data like a `selfsubjectaccessreviews`. 
Here I've just implemented a temporary workaround, but, my question is: should we also pass to the plugin a normal k8s client besides the existing one?

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -56,3 +85,160 @@ func (s *Server) GetClient(ctx context.Context) (dynamic.Interface, error) {
...
+	})
+	if err != nil {
+		return false, fmt.Errorf("Error parsing the selfSubjectAccessReviews request: %s", err)
+	}
+	selfSubjectAccessReviewsResponse, err := client.Resource(selfSubjectAccessReviews).Create(ctx, &unstructured.Unstructured{Object: unstructuredData}, metav1.CreateOptions{})
+	if err != nil {
+		return false, fmt.Errorf("Error creating the selfSubjectAccessReviews request: %s", err)
+	}
+	res := &authorizationapi.SelfSubjectAccessReview{}
+	err = runtime.DefaultUnstructuredConverter.FromUnstructured(selfSubjectAccessReviewsResponse.Object, res)
```
antgamdia, line 239: This is what I meant by temporary ugly workaround :P . I just needed to use the dynamic client, so I converted it with `ToUnstructured` and with `FromUnstructured`

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
```diff
@@ -70,12 +70,21 @@ message GetAvailablePackageSummariesRequest {
   // The context (cluster/namespace) for the request
   Context context = 1;
 
+  // FilterOptions
+  //
   // FilterOptions available when requesting summaries
   message FilterOptions {
```
antgamdia, line 76: Adding these comments for a better swagger document. 
When embedding these messages inside other msgs we often forgot to add the proper docs, that's why I tend not to embed them. But, on the other side, it's clearer to know which mgs they belong to, so... no strong opinion though.

absoludity: Yep, and better if we then move the embedded message out. But again, if it's OK with you, I'll add those in the separate PR which hasn't landed yet which adds the filter options. It'll be easier if you don't merge that here nor add functionality depending on them in this PR, but rather keep this one focused on getting the available package summaries (securely, checking the namespace) without filter options. Just so we can work together without depending on unlanded branches.

absoludity: Or if you want to update once it's landed and still include the filters, that's fine too, but just keen to avoid depending on other people's unlanded code unless it's necessary (ie. if this PR was dependent on filtering options, but I don't think it is).

absoludity: Hah - so I just found the source of confusion here. Sorry, it was my fault. I'd thought I'd created a PR with the filter options yesterday... but it turns out I'd created the branch before our standup but didn't get back to push it and create the PR. I only found out because I just went to update my branch and pushed the changes to find that it created a new branch on GH. So from your POV you obviously weren't working on something that I'd done at all. Sorry for the confusion :)

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -56,3 +85,160 @@ func (s *Server) GetClient(ctx context.Context) (dynamic.Interface, error) {
...
+
+func (s *Server) hasAccessToNamespace(ctx context.Context, namespace string) (bool, error) {
+	client, err := s.GetClient(ctx)
+	if err != nil {
+		return false, err
+	}
+
+	// TODO(agamez): this is a temporary workaround since the client getter we (the plugin) get is just
+	// for dynamic interfaces. I guess we should also pass a normal k8s client in case the plugins want to
+	// interact with typed kubernetes resources.
```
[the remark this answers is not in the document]
absoludity: Did this in #3043 . Feel free to land that and update, or just review it and continue your PR without it (and I'll update later).

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
absoludity: Yep, see #3043 .

absoludity (commented):
> I've pushed an initial version of the ns check using the existing logic for the can-i method.
> 
> Problem: the client we are passing back to the plugins (via clientGetter) is just for dynamic interfaces, That is, we can't query structured/typed data like a `selfsubjectaccessreviews`.
> Here I've just implemented a temporary workaround, but, my question is: should we also pass to the plugin a normal k8s client besides the existing one?

Yep. I've jumped in to do that separately for you in #3043 .

> 
> btw: tests are still pending (I'll do it tomorrow)

Yep, let me know when it's ready and I'll finish the review. Thanks!

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
antgamdia: Hahah, I wasn't getting the point here; I added the FilterOptions in my branch so, as you said, I didn't see you were also working on that. It totally makes sense now.
I'll add the remaining changes to the filter (version and app version) here so that we have the whole filtering functionality implemented.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -56,3 +85,160 @@ func (s *Server) GetClient(ctx context.Context) (dynamic.Interface, error) {
...
+			return nil, status.Errorf(codes.Unimplemented, "Not supported yet: request.Context.Cluster: [%v]", request.Context.Cluster)
+		}
+		namespace = request.Context.Namespace
+	}
+	// Check the requested namespace: if any, return "everything a user can read";
+	// otherwise, first check if the user can access the requested ns
+	if namespace == "" {
+		// TODO(agamez): not including a namespace means that it returns everything a user can read
+		return nil, status.Errorf(codes.Unimplemented, "Not supported yet: not including a namespace means that it returns everything a user can read")
+	} else {
```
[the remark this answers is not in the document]
antgamdia: I do agree (I'll change it). However, I decided to leave the `else` statement since, at some point, the `if` branch will have some logic. This way, fewer git changes (aka indentation) when we add the "no namespace" case.
Anyway, happy to remove it.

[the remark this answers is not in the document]
absoludity: Yep, up to you (I don't mind, just a general note for clarity).

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server_test.go
```diff
@@ -12,3 +12,395 @@ limitations under the License.
...
+			request: &corev1.GetAvailablePackageSummariesRequest{
+				Context: &corev1.Context{
+					Namespace: "my-ns",
+				},
+			},
+			charts:           []*models.Chart{{Name: "foo"}},
+			expectedPackages: []*corev1.AvailablePackageSummary{},
+			statusCode:       codes.Unauthenticated,
+		},
+	}
```
absoludity, line 373: Excellent :)

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -56,3 +84,144 @@ func (s *Server) GetClients(ctx context.Context) (kubernetes.Interface, dynamic.
...
+	res, err := client.AuthorizationV1().SelfSubjectAccessReviews().Create(context.TODO(), &authorizationv1.SelfSubjectAccessReview{
+		Spec: authorizationv1.SelfSubjectAccessReviewSpec{
+			ResourceAttributes: &authorizationv1.ResourceAttributes{
+				Group:     "",
+				Resource:  "secrets",
+				Verb:      "get",
+				Namespace: namespace,
+			},
+		},
+	}, metav1.CreateOptions{})
```
absoludity, line 222: Much neater.

absoludity (approved):
Great, +1. Thanks Antonio. Is it worth implementing the pagination support in the request for this now? What else is required for this method to be used in the client?

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
```diff
@@ -380,3 +375,38 @@ message Maintainer {
...
+
+    // Repositories
+    //
+    // Collection of repositories where the packages belong to
+    repeated string repositories = 3;
+
+    // Package version
+    //
+    // Package version for the request
+    string version = 4;
```
absoludity, line 406: I think elsewhere we've stared using `pkg_version` to be specific together with `app_version` as you've used below?

antgamdia: Good idea indeed. I've added some comments in the doc for unifying the names here. Using just `version` could be misleading in some cases. I'd change each version by either `pkg_` or `app_` prefixes.

antgamdia:
> Is it worth implementing the pagination support in the request for this now? What else is required for this method to be used in the client?

#### Additonal fields for this method

Currently, the client requires the following params  for displaying the catalog: `icon, name, repo, version, description, namespace, id`

A response of this APIs is something like:

```json
{
  "availablePackagesSummaries": [
    {
      "availablePackageRef": {
        "context": {
          "namespace": "kubeapps" namespace
        },
        "identifier": "marketplace/airflow" id
      },
      "latestVersion": "10.2.1", version
      "iconUrl": "https://bitnami.com/assets/stacks/airflow/img/airflow-stack-220x234.png", icon
      "displayName": "airflow", name
      "shortDescription": "Apache Airflow is a platform to programmatically author, schedule and monitor workflows." description
    }
  ]
};
```

So the only missing field is `repo` (it could eventually be retrieved using the `identifier`). 


#### Pagination

Well, this is interesting. Currently, the API is returning something like:

```json
{"data":[ charts ],"meta":{"totalPages": pages }}
```

Perhaps, we could modify the messages to use this `data`/`meta`  approach. Or even go with some of these options:
1) GitHub is using the [`Links` header](https://datatracker.ietf.org/doc/html/rfc5988): https://docs.github.com/en/rest/guides/traversing-with-pagination
2) Google suggests including the pagination as part of the proto msg: https://cloud.google.com/apis/design/design_patterns#list_pagination

However, I'd rather land this PR, continue to work on the other operation and, later on, look into these two additional things in a separate PR. Specially pagination... we should come to an agreement before start changing the UI.
_(I'm gonna crosspost it in the parent issue)_

absoludity:
> However, I'd rather land this PR, continue to work on the other operation and, later on, look into these two additional things in a separate PR. Specially pagination... we should come to an agreement before start changing the UI.

Great, yes - I didn't mean to imply it should be in this PR, just trying to ensure we work on the most relevant things next.

> _(I'm gonna crosspost it in the parent issue)_

Great, I'll reply there :)

PR #3030, opened 2021-06-23
Misc todo cleanup in fluxv2 plugin

gfichtenholt:
I wanted to clean up and merge some todo items leftover from the previous PR before I embark on a new big task to add caching layer to fluxv2 plug-in


cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
```diff
@@ -194,11 +197,14 @@ func (s *Server) GetAvailablePackageSummaries(ctx context.Context, request *core
...
-			repo.Namespace = namespace
+		if err != nil || !found {
+			// should not happen in reality
+			log.Errorf("field metadata.namespace not found on HelmRepository: %w:\n%v", err, obj)
+			// just skip over to the next one
+			continue
 		}
+		repo.Namespace = namespace
 
 		repoPackages, err := readPackagesFromRepoIndex(&repo, url)
```
absoludity, line 209: Probably worth a TODO at the start of the surrounding iteration that each repo should be read in a separate go routine, if this code is used in the future.

pkg/http-client/httpclient.go
```diff
@@ -13,15 +13,85 @@ limitations under the License.
...
+	defaultTimeoutSeconds = 10
 )
 
 type Client interface {
 	Do(req *http.Request) (*http.Response, error)
 }
 
+//
+// creates a new instance of Client, given a path to addtional certificates
+// certFile may be empty string, which means no additional certs will be used
```
absoludity, line 35: I think this copy-n-paste needs updating :)

pkg/http-client/httpclient.go
```diff
@@ -13,15 +13,85 @@ limitations under the License.
...
+		Transport: &http.Transport{
+			Proxy: http.ProxyFromEnvironment,
+		},
+	}
+}
+
+//
+// creates a new instance of Client, given a path to addtional certificates
+// certFile may be empty string, which means no additional certs will be used
+//
```
absoludity, line 50: Just note that generally golang docstrings are not preceded with an empty `//\n` line (nor followed by one), and usually begin with the function/struct/constant name, such as:
```
// Foo does stuff before returning nil
// Extra details about Foo here.
func Foo() nil
```
See https://blog.golang.org/godoc for the background, but it just plays well with the `godoc` binary.

absoludity (approved):

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
gfichtenholt: will do

pkg/http-client/httpclient.go
gfichtenholt: indeed

pkg/http-client/httpclient.go
gfichtenholt: mmm, maybe, but it makes code easier to read. Oh well, when in Rome 👍 I will change it

PR #3036, opened 2021-06-24
Update the postgresql asset manager to be publicly exported.

absoludity:
### Description of the change

Potential prequel for #3022, so that the plugin can just export the required manager and functions from the existing code.

Just does:
-  Public interface `assetManager` -> `AssetManager` (interface is just because we transition from mongo and supported both implementations for a while).
-  Public NewPGManager
-  Public PostgresAssetManager. (woops, I'll add a commit with this in a bit).
-  Public methods on the postgres asset manager

### Benefits

Let's Antonio focus on 3022

### Possible drawbacks

?

### Applicable issues

  - Ref #2993


antgamdia (approved):
Great! Thanks a lot for handling this!
(I recall my first days in the project when I even didn't know how to export a method in Go.... haha)

I'll update accordingly my PR once it got landed

PR #3039, opened 2021-06-24
Add GetAvailablePackageVersions messages

absoludity:
### Description of the change

Adds an initial draft for GetAvailablePackageVersions.

May change if there's feedback on the doc, but fine to review.

### Benefits

Adds ability to get a summary of versions for a package or optionally
the complete history for a specific major or major.minor version.

### Applicable issues

  - Ref #2993


antgamdia (approved):
Great - thanks for adding it. Later on, we can add some modifications, but it's a great starting point to start defining the actual implementation for our direct Helm plugin.

absoludity:
> Great - thanks for adding it. Later on, we can add some modifications, but it's a great starting point to start defining the actual implementation for our direct Helm plugin.

Yep, but note, I added this (after working on the doc) because I assumed it *wasn't* going to interfere with your current branch (as in, you weren't yet taking filter options into account as they weren't present, so would be added later in another PR by whoever). Pulling them into your current branch before they're landed etc. can make things tricky... what I'd recommend moving forward is that we don't add extra functionality to a PR/branch unless it's required to land.

absoludity:
> Great - thanks for adding it. Later on, we can add some modifications, but it's a great starting point to start defining the actual implementation for our direct Helm plugin.

Sorry, deleted previous comment as it was related to a confusion about another PR. Yep, we can modify it moving forward, but keen to know what modifications you think might be required?

PR #3040, opened 2021-06-24
Extract assetsvc shared logic to assetsvc/pkg/utils

antgamdia:
### Description of the change

Follow up of https://github.com/kubeapps/kubeapps/pull/3036

I wasn't able to use the exported functions and interfaces since they belong to different pkg. Even if uppercased, they still belong to the `main` pkg (used for created the binary). So, for actually using them, I had to move them to another subfolder. They still are under `assetsvc` though.

### Benefits

Re-use code
### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

Perhaps there is a way to do so, but I've tried importing them and the compiler never was able to retrieve the dependency, so I just went this other way. Happy to discover any alternative solution though.

### DIFF
--- a/cmd/assetsvc/handler.go
+++ b/cmd/assetsvc/handler.go
@@ -25,7 +25,7 @@ import (
 
 	"github.com/gorilla/mux"
 	"github.com/kubeapps/common/response"
-	"github.com/kubeapps/kubeapps/cmd/assetsvc/pkg/assetsvc_utils"
+	"github.com/kubeapps/kubeapps/cmd/assetsvc/pkg/utils"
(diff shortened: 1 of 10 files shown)


absoludity (approved):
LGTM. Given that the pkg is under assetsvc, I don't see a need for it to be `assetsvc/pkg/assetsvc_utlis` rather than simply `assetsvc/pkg/utils`. When importing, if people want to use a named import to be specific they can do that (eg. `import assetsvc_utils "github.com/.../.../assetsvc/pkg/utils"`).

And yep, I'd missed that they were all in a main package.

Issue #1850, opened 2020-07-07
Upgrade action should be exclusive


lingsamuel:
### Description:

Trigger a new `upgrade` action before the previous one finishes, the newer one will probably fail because some hook jobs name conflict, which makes the release status become `failed`. And some jobs remain in the namespace.

### Steps to reproduce the issue:

1. Create a chart contains a lot hooks
2. Upgrade it multiple times
3. The newer one will failed because name conflict, but the previous one is executing normally
4. Release status changes to `failed`

### Describe the results you expected:

The second `upgrade` action should be rejected if there is another upgrade action is running.

### Version of Helm, Kubeapps and Kubernetes:

Kubeapps version 3.7.2 with Helm 3.

andresmgot:
Hi @lingsamuel, we could indeed check the release status and disable the upgrade button if the status is `PENDING_UPGRADE`. Would that solve your issue?

lingsamuel:
I think that's okay.

ppbaena:
Close as it is merged in #3041

PR #3041, opened 2021-06-24
initial fix for issue 1850, disable upgrade button

dlaloue-vmware:
### Description of the change

1 - refactored the upgrade button (from AppView) into its own file, similar to delete/rollback, so it can now have its own logic
2 - made so the button is disabled if the release has a pending action (delete, upgrade, rollback...)
3 - added a tooltip for cases when the button is disabled

### Benefits

Users have reported issues when doing an upgrade while an action was pending may result in conflicts. Preventing users from doing an upgrade should  avoid such conflicts.

### Possible drawbacks

none

### Applicable issues

issue #1850 

### Additional information

none


antgamdia:
Thanks for fixing this issue! Now our buttons will be 'safer'  :P 
Haven't had a look at the draft PR yet, but just two questions: 
- are you planning to also modify the operators view?
- are you also gonna change the rest of the buttons (delete and rollback)? I foresee they're affected by the same issue, don't you think?

dlaloue-vmware:
I haven't create a test for the new file. I will work on that next.

but i haven't been able to reproduce the issue using the UI.
even if i have 2 web browsers open, and do an upgrade in one, if i refresh the other browser, the status is never one of the pending status.
it seems that the backend is actually sending back the deployed version, not the new one being deployed until it is ready.

dlaloue-vmware:
> Thanks for fixing this issue! Now our buttons will be 'safer' :P
> Haven't had a look at the draft PR yet, but just two questions:
> 
>     * are you planning to also modify the operators view?
> 
>     * are you also gonna change the rest of the buttons (delete and rollback)? I foresee they're affected by the same issue, don't you think?

dlaloue-vmware:
sorry, mistake, did not mean to close :-(

dlaloue-vmware:
> Thanks for fixing this issue! Now our buttons will be 'safer' :P

see my other comment, i don't think this is safe, yet, as i could not get to reproduce a case where the button would actually be disabled.


> Haven't had a look at the draft PR yet, but just two questions:
>     * are you planning to also modify the operators view?

I tried to look at operators, but it does not seem enabled by default in the UI. I need to install something to make it work (operator lifecycle manager), so i will have to look into it later.
also, the code is slightly different, i did not spend too much time on it yet, but i did not find the same "status" data, and the buttons have much less logic in them.
I will only look at operators once i am sure this fix actually works.

> 
>     * are you also gonna change the rest of the buttons (delete and rollback)? I foresee they're affected by the same issue, don't you think?

yes, they would most likely have the same issue. once i can verify this is the right fix, and the code looks ok, i can expand to the other 2 buttons.

antgamdia:
> see my other comment, i don't think this is safe, yet, as i could not get to reproduce a case where the button would actually be disabled.

There's a pretty simple way to check this issue. Deploy or upgrade an app and, while it says "loading bla bla", click "back" in your browser and then click "next". If the fix is correct, you'll see your upgrade button disabled and the rest, enabled.

 > yes, they would most likely have the same issue. once i can verify this is the right fix, and the code looks ok, i can expand to the other 2 buttons.

Great, thanks!



FWIW, here is my personal script I use for deploying the OLM. Run this and you will get everything up and running in Kubeapps:

```bash
kubectl apply -f "https://github.com/operator-framework/operator-lifecycle-manager/releases/download/v0.17.0/crds.yaml"
kubectl wait --for=condition=Established -f "https://github.com/operator-framework/operator-lifecycle-manager/releases/download/v0.17.0/crds.yaml"
kubectl apply -f "https://github.com/operator-framework/operator-lifecycle-manager/releases/download/v0.17.0/olm.yaml"

# wait for deployments to be ready
kubectl rollout status -w deployment/olm-operator --namespace="olm"
kubectl rollout status -w deployment/catalog-operator --namespace="olm"

retries=30
until [[ $retries == 0 ]]; do
    new_csv_phase=$(kubectl get csv -n "olm" packageserver -o jsonpath='{.status.phase}' 2>/dev/null || echo "Waiting for CSV to appear")
    if [[ $new_csv_phase != "$csv_phase" ]]; then
        csv_phase=$new_csv_phase
        echo "Package server phase: $csv_phase"
    fi
    if [[ "$new_csv_phase" == "Succeeded" ]]; then
	break
    fi
    sleep 10
    retries=$((retries - 1))
done

if [ $retries == 0 ]; then
    echo "CSV \"packageserver\" failed to reach phase succeeded"
    exit 1
fi

kubectl rollout status -w deployment/packageserver --namespace="olm"
```

dashboard/src/components/AppView/AppControls/DeleteButton/DeleteButton.test.tsx
```diff
@@ -67,3 +69,16 @@ it("renders an error", async () => {
 
   expect(wrapper.find(Alert)).toIncludeText("Boom!");
 });
+
+it("renders disabled", async () => {
```
antgamdia, line 73:
```suggestion
it("should render a disabled button if when passing an in-progress status", async () => {
```

dashboard/src/components/AppView/AppControls/RollbackButton/RollbackButton.test.tsx
```diff
@@ -71,3 +73,16 @@ it("renders an error", async () => {
 
   expect(wrapper.find(Alert)).toIncludeText("Boom!");
 });
+
+it("renders disabled", async () => {
```
antgamdia, line 77:
```suggestion
it("should render a disabled button if when passing an in-progress status", async () => {
```

dashboard/src/components/AppView/AppControls/StatusAwareButton.tsx
```diff
@@ -0,0 +1,32 @@
+import { CdsButton } from "@cds/react/button";
+import ReactTooltip from "react-tooltip";
+import { hapi } from "../../../shared/hapi/release";
+
+export interface IStatusAwareButtonProps {
+  id: string;
+  releaseStatus: hapi.release.IStatus | undefined | null;
```
antgamdia, line 7: In other places, you are using `status`; I would suggest renaming `status` to `releaseStatus` instead. IMO, just using `status` reminds me of a true/false value instead of a `hapi.release.IStatus` object.

dashboard/src/components/AppView/AppControls/StatusAwareButton.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const { id, releaseStatus, ...otherProps } = props;
+  const disabled: boolean = releaseStatus?.code
+    ? releaseStatus.code >= 5 && releaseStatus.code <= 8
+    : false;
+  const tooltips = {
+    5: "The applicatixon is being deleted.",
+    6: "The application is pending installation.",
+    7: "The application is pending upgrade.",
+    8: "The application is pending rollback.",
+  };
```
antgamdia, line 20: I'd suggest extracting this logic to a separate function instead as it is really coupled to the helm status codes (of course, it should be aware of that, but I'd rather encapsulate it). See what you think.

dashboard/src/components/AppView/AppControls/StatusAwareButton.tsx
```diff
@@ -0,0 +1,32 @@
...
+  releaseStatus: hapi.release.IStatus | undefined | null;
+}
+
+export default function StatusAwareButton<T extends IStatusAwareButtonProps>(props: T) {
+  const { id, releaseStatus, ...otherProps } = props;
+  const disabled: boolean = releaseStatus?.code
+    ? releaseStatus.code >= 5 && releaseStatus.code <= 8
+    : false;
+  const tooltips = {
+    5: "The applicatixon is being deleted.",
```
antgamdia, line 16: Minor typo:  s/applicatixon/application

dashboard/src/components/AppView/AppControls/StatusAwareButton.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { hapi } from "../../../shared/hapi/release";
+
+export interface IStatusAwareButtonProps {
+  id: string;
+  releaseStatus: hapi.release.IStatus | undefined | null;
+}
+
+export default function StatusAwareButton<T extends IStatusAwareButtonProps>(props: T) {
+  const { id, releaseStatus, ...otherProps } = props;
+  const disabled: boolean = releaseStatus?.code
```
antgamdia, line 12: Currently, it isn't handling when it is loading and no status code has been received:

https://user-images.githubusercontent.com/11535726/123978094-de826300-d9bf-11eb-8c48-2b3e02a7e56a.mp4


https://user-images.githubusercontent.com/11535726/123978096-df1af980-d9bf-11eb-8df9-bbb05916a7c3.mp4

antgamdia: I'd suggest sth like:

```js
  // Disable the button if: the code is undefined or null OR the status code is one of [5,6,7,8]
  // See https://github.com/kubeapps/kubeapps/blob/master/dashboard/src/shared/hapi/release.d.ts#L559
  const disabled = releaseStatus?.code ? _.inRange(releaseStatus?.code, 5, 9) : true;
```
See what you think

dashboard/src/components/AppView/AppControls/UpgradeButton/UpgradeButton.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const mockDispatch = jest.fn();
+  spyOnUseDispatch = jest.spyOn(ReactRedux, "useDispatch").mockReturnValue(mockDispatch);
+});
+
+afterEach(() => {
+  actions.kube = { ...kubeaActions };
+  spyOnUseDispatch.mockRestore();
+});
+
+it("renders disabled", async () => {
```
antgamdia, line 32:
```suggestion
it("should render a disabled button if when passing an in-progress status", async () => {
```

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -195,23 +192,27 @@ export default function AppView() {
         helm={true}
         icon={icon}
         buttons={[
-          <Link to={url.app.apps.upgrade(cluster, namespace, releaseName)} key="upgrade-button">
-            <CdsButton status="primary">
-              <CdsIcon shape="upload-cloud" /> Upgrade
-            </CdsButton>
-          </Link>,
+          <UpgradeButton
```
antgamdia, line 195: Thanks for adding the RollbackButton component. I don't know if, at this moment, it worths just having the `StatusAwareButton` with an additional property `type="Upgrade|Rollback|Delete". This way we would be DRYer no?

Just an idea, I'm totally fine leaving this PR as is

dashboard/src/components/AppView/AppControls/StatusAwareButton.tsx
```diff
@@ -0,0 +1,32 @@
...
+    ? releaseStatus.code >= 5 && releaseStatus.code <= 8
+    : false;
+  const tooltips = {
+    5: "The applicatixon is being deleted.",
+    6: "The application is pending installation.",
+    7: "The application is pending upgrade.",
+    8: "The application is pending rollback.",
+  };
+  const tooltip = releaseStatus?.code ? tooltips[releaseStatus.code] : undefined;
```
antgamdia, line 21: WDYT? 

```suggestion
  const tooltip = releaseStatus?.code ? tooltips[releaseStatus.code] : "Loading application...";
```

antgamdia (commented):
Thanks for the refactoring effort! Apart from the in-line comments, two general things:

- What do you think about just having a `StatusAwareButton` with an additional property that deals with the "Upgrading" or "xxxx" texts?
- I'm missing the unit test for the `StatusAwareButton` component itself. There I'd like to see each mode and case is properly handled (codes 1 to 8, null and undefined values)

dashboard/src/components/AppView/AppControls/StatusAwareButton.tsx
dlaloue-vmware: That is the exact purpose of the StatusAwareButton. I don't see a need to add one more layer of indirection.
All the button does is encapsulate the logic of computing the disabled flag and tooltip.
what would be the reason for another indirection?

dashboard/src/components/AppView/AppControls/StatusAwareButton.tsx
dlaloue-vmware: in both videos, there is a status of code 1 (deployed) that is returned, that is why the buttons are not disabled. This is the issue i mentioned int he standup, the backend is not returning the proper code.
I was able to confirm that by always showing a tooltip with the code returned, and it is always "deployed".

unless we fix the backend, the UI will not be able to work as expected. I believe we agreed to wait until we move over to the new apis.

dlaloue-vmware: regarding the second comment, getting a status code that is undefined is an edge case.
I think it is safer to not disable to avoid preventing the user from doing a compensating/fixing action.
but if you feel it is better to disable when the status is undefined, i can make the change.

dashboard/src/components/AppView/AppView.tsx
dlaloue-vmware: I am not familiar with what you mean by "DRYer"?
the StatusAwareButton only deals with disabling and adding a tooltip. Adding a type would inject more state than its needs to have.
what would be the benefit of adding type to it?

dashboard/src/components/AppView/AppControls/StatusAwareButton.tsx
dlaloue-vmware: as stated above, the button does not handle the case of "undefined" in purpose.
and also, i believe that if the status is undefined, the whole page displays a "loading" screen before rendering the details, meaning that the buttons are not displayed

dashboard/src/components/AppView/AppControls/StatusAwareButton.tsx
antgamdia: Ok, sounds reasonable. I was thinking about what if Helm changes their codes. But, the condition is rather simple, so there is no need to adding an extra layer as you mention.

dashboard/src/components/AppView/AppControls/StatusAwareButton.tsx
antgamdia: Exactly, the code: 1 is the one being returned usually. So there is no way for the frontend to be aware of that changes.
We can think of alternatives like adding a redux state action triggered when performing an upgrade, that would disable the button until a new status:1 event is returned again. This way we prevent users from clicking when the upgrade operation hasn't begun yet.
It's a bit more complicated and, given we are gonna refactor the backed, I don't think it worths it.

Regarding the `undefined` edge case: I don't see it as an edge case in that this status is returned every time the component is mounted (write a `console.log` statement and see how many times it gets called).
I think it would prevent users from clicking multiple times the buttons going back and forth on the browser's history (as I use to do when testing out an app haha).
(btw, doing so makes an error like this appear: https://stackoverflow.com/questions/53949393/cant-perform-a-react-state-update-on-an-unmounted-component)

In which situations do you think having the button enabled could be useful when the status is `undefined` ?

dashboard/src/components/AppView/AppView.tsx
antgamdia: Hah - too many acronyms sometimes. I meant DRY as in https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

I just felt we are replicating too much code in each `UpgradeButton`, `RollbackButton` and `DeleteButton` (eg. unit tests). Perhaps it's better to remain them as separated components as a single "generic" status aware button would deal with too much logic and state.
Just wanted to open up the discussion, but I'm ok with the current implementation!

dashboard/src/components/AppView/AppControls/StatusAwareButton.tsx
antgamdia: Ok, even if we consider the `undefined` case, I agree the tooltip is not really necessary.

dashboard/src/components/AppView/AppControls/StatusAwareButton.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+const kubeaActions = { ...actions.kube };
+beforeEach(() => {
+  const mockDispatch = jest.fn();
+  spyOnUseDispatch = jest.spyOn(ReactRedux, "useDispatch").mockReturnValue(mockDispatch);
+});
+
+afterEach(() => {
+  actions.kube = { ...kubeaActions };
+  spyOnUseDispatch.mockRestore();
+});
```
antgamdia, line 20: Do you really need this?  I think you can get rid of it ?

dashboard/src/components/AppView/AppControls/StatusAwareButton.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+    { code: 6, disabled: true, tooltip: "installation" },
+    { code: 7, disabled: true, tooltip: "upgrade" },
+    { code: 8, disabled: true, tooltip: "rollback" },
+    { code: undefined, disabled: false, tooltip: undefined },
+    { code: null, disabled: false, tooltip: undefined },
+  ];
+
+  for (const testProps of testsProps) {
+    let releaseStatus;
+    switch (testProps.code) {
```
antgamdia, line 46: Why not using just `testProps.code` directly?

antgamdia:
Thanks for adding the test.  The only open aspect is the `disabled` condition:

```js
const disabled: boolean = releaseStatus?.code ? releaseStatus.code >= 5 && releaseStatus.code <= 8 : false;
``` 

vs the one I suggested:


```js
// Disable the button if: the code is undefined or null OR the status code is one of [5,6,7,8]
// See https://github.com/kubeapps/kubeapps/blob/master/dashboard/src/shared/hapi/release.d.ts#L559
const disabled = releaseStatus?.code ? _.inRange(releaseStatus?.code, 5, 9) : true;
```

CCing @absoludity . What do you think about considering the `undefined` case to also disabling the button (see my second video at https://github.com/kubeapps/kubeapps/pull/3041#discussion_r661536855)

absoludity:
> CCing @absoludity . What do you think about considering the `undefined` case to also disabling the button (see my second video at [#3041 (comment)](https://github.com/kubeapps/kubeapps/pull/3041#discussion_r661536855))

We chatted about this on slack, Dimitri will include this with his update.

dashboard/src/components/AppView/AppControls/StatusAwareButton.tsx
dlaloue-vmware: i will disable the buttons when the status is undefined

dashboard/src/components/AppView/AppControls/StatusAwareButton.test.tsx
dlaloue-vmware: i may not need this. I have added it to be consistent with other test files.
but in my scenario, i am not making any backend calls, so i agree this should not be necessary. I will remove it and see that the tests are still passing fine.

dashboard/src/components/AppView/AppControls/StatusAwareButton.test.tsx
dlaloue-vmware: the reason is that the variable releaseStatus is not the code, it is a object that contains the code, they are of different types

antgamdia (approved):
Awesome! Thanks for the changes! LGTM now :) 
Land whenever you want

PR #3043, opened 2021-06-25
Add typed client to plugin client getter.

absoludity:
### Description of the change

As per discussion on #3022, we originally tried the typed client for interacting with packaging CRs but hit the issue that we'd need the same version of client-go used by those packaging systems, which leads to dependency hell. So we switched to use a dynamic client for the plugins.

That said, now that certain plugins also need to interact with the core k8s api resources, a typed client(set) would also be useful (and nicer to work with) for interacting with those core resources.

This PR updates so that plugins are provided with both - a type clientset and a dynamic clientset.

### Benefits

Plugins don't need to use unstructured objects when interacting with the core k8s resources.

### Possible drawbacks

None (assuming the client works IRL, which we'll test in a subsequent PR that uses the typed client for the direct helm plugin).

### Applicable issues

  - ref #2993

### Additional information

Antonio, if you want to land this so you can use it in your PR, that's fine, or if you'd rather not depend on it but move forward with your unstructured work-around and switch later, that's fine too :)

### DIFF
--- a/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_test.go
+++ b/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_test.go
@@ -22,19 +22,23 @@ import (
 	"github.com/google/go-cmp/cmp/cmpopts"
 	corev1 "github.com/kubeapps/kubeapps/cmd/kubeapps-apis/gen/core/packages/v1alpha1"
 	"github.com/kubeapps/kubeapps/cmd/kubeapps-apis/gen/plugins/kapp_controller/packages/v1alpha1"
+	"github.com/kubeapps/kubeapps/cmd/kubeapps-apis/server"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
-	"k8s.io/client-go/dynamic/fake"
+	dynfake "k8s.io/client-go/dynamic/fake"
+	"k8s.io/client-go/kubernetes"
+	typfake "k8s.io/client-go/kubernetes/fake"
 )
 
 func TestGetClient(t *testing.T) {
+
 	testCases := []struct {
 		name         string
-		clientGetter func(context.Context) (dynamic.Interface, error)
+		clientGetter server.KubernetesClientGetter
 		statusCode   codes.Code
 	}{
 		{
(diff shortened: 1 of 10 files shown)


cmd/kubeapps-apis/server/plugins.go
```diff
@@ -298,37 +298,38 @@ func createClientGetter(serveOpts ServeOptions) (func(context.Context) (dynamic.
...
-			}
+			config = inClusterConfig
+		}
+		dynamicClient, err := dynamic.NewForConfig(config)
+		if err != nil {
+			return nil, nil, fmt.Errorf("unable to create dynamic client: %w", err)
+		}
+		typedClient, err := kubernetes.NewForConfig(config)
+		if err != nil {
+			return nil, nil, fmt.Errorf("unable to create typed client: %w", err)
```
absoludity, line 330: Main change here (other than adding the new client) was just to DRY up the code for creating a client (or clients).

cmd/kubeapps-apis/server/plugins.go
```diff
@@ -51,6 +48,9 @@ const (
 	clustersCAFilesPrefix   = "/etc/additional-clusters-cafiles"
 )
 
+// KubernetesClientGetter is a function type used by plugins to get a k8s client
+type KubernetesClientGetter func(context.Context) (kubernetes.Interface, dynamic.Interface, error)
```
absoludity, line 52: Created this type alias so its easier to update the signature (before I added the typed interface).

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_test.go
```diff
@@ -66,16 +70,19 @@ func TestGetClient(t *testing.T) {
...
+			typedClient, dynamicClient, err := s.GetClients(context.Background())
 
 			if got, want := status.Code(err), tc.statusCode; got != want {
 				t.Errorf("got: %+v, want: %+v", got, want)
 			}
 
-			// If there is no error, the client should be a dynamic.Interface implementation.
+			// If there is no error, the clients should not be nil.
 			if tc.statusCode == codes.OK {
-				if _, ok := client.(dynamic.Interface); !ok {
```
absoludity, line 77: Not sure why I initially used a type conversion here, since if it's non-nil, it's going to be the correct type (or it wouldn't compile).

cmd/kubeapps-apis/server/plugins.go
antgamdia: Great, thanks for adding it. I initially thought of adding a clientGetter for each type, sth like "clientGetter" and "dynamicClientGetter", but returning both clients is better (assuming there are no more clients out there)

antgamdia (approved):
Thanks for jumping in adding the client getter. It's interesting the amount of PRs that are required for a simple listing packages operation. Step by step :)

I'll land this PR and modify mine accordingly (happy to get rid of the from/to structured)

PR #3057, opened 2021-06-29
Improve credentials handling when fetching repo resources

### LINKED ISSUE #2136: Avoid using given auth credentials when syncing charts outside index.yaml hostname
### Description:

At the moment, if some credentials are given when creating an Application Repository, these credentials are used for both, downloading the `index.yaml` file of the repository AND downloading the chart tarballs of that repository. The issue is that the hosts for those two files may be different so reusing the same credentials may be a security concern. We should avoid using them if the host name differs.

antgamdia:
### Description of the change

This PR adds a similar change as the Helm folks did in https://github.com/helm/helm/commit/179f90151d5ecb4aa3d35ada35e82b5c1e791752

Now, the credentials will be passed to an external URL if: 1) a user adds `pass-credentials` to the repo OR 2) the domain in the chart index.yaml file matches the icon/tarball domain.

### Benefits

Not passing credentials to external domains (such as icons and tarballs)

### Possible drawbacks

If users don't explicitly decide to allow passing the credentials, in scenarios when the index.yaml URL doesn't match the icon/tarball URL domain, the credentials won't be passed through anymore.

### Applicable issues

- fixes #2136

### Additional information

image

More testing is required, I feel the mock clients we use here are not enough to make sure we are doing it right.

### DIFF
--- a/cmd/apprepository-controller/pkg/apis/apprepository/v1alpha1/types.go
+++ b/cmd/apprepository-controller/pkg/apis/apprepository/v1alpha1/types.go
@@ -54,7 +54,7 @@ type AppRepositorySpec struct {
 	FilterRule FilterRuleSpec `json:"filterRule,omitempty"`
 	// (optional) description
 	Description string `json:"description,omitempty"`
-	// PassCredentials allows passing credentials to all domains
+	// PassCredentials allows passing credentials with requests to other domains linked from the repository
 	PassCredentials bool `json:"passCredentials,omitempty"`
 }
 
--- a/cmd/apprepository-controller/pkg/apis/apprepository/v1alpha1/types.go
+++ b/cmd/apprepository-controller/pkg/apis/apprepository/v1alpha1/types.go
@@ -54,6 +54,8 @@ type AppRepositorySpec struct {
 	FilterRule FilterRuleSpec `json:"filterRule,omitempty"`
 	// (optional) description
 	Description string `json:"description,omitempty"`
+	// PassCredentials allows passing credentials to all domains
+	PassCredentials bool `json:"passCredentials,omitempty"`
 }
 
 // AppRepositoryAuth is the auth for an AppRepository resource
--- a/cmd/asset-syncer/utils.go
+++ b/cmd/asset-syncer/utils.go
@@ -236,11 +236,18 @@ func (r *HelmRepo) Charts(fetchLatestOnly bool) ([]models.Chart, error) {
 
 // FetchFiles retrieves the important files of a chart and version from the repo
 func (r *HelmRepo) FetchFiles(name string, cv models.ChartVersion) (map[string]string, error) {
+	authorizationHeader := ""
+	chartTarballURL := chartTarballURL(r.RepoInternal, cv)
+
+	if (passCredentials || compareURLDomains(chartTarballURL, r.URL)) && len(r.AuthorizationHeader) > 0 {
+		authorizationHeader = r.AuthorizationHeader
+	}
+
 	return tarutil.FetchChartDetailFromTarball(
 		name,
-		chartTarballURL(r.RepoInternal, cv),
+		chartTarballURL,
 		userAgent(),
-		r.AuthorizationHeader,
+		authorizationHeader,
 		r.netClient)
 }
 
@@ -783,7 +790,7 @@ func (f *fileImporter) fetchAndImportIcon(c models.Chart, r *models.RepoInternal
 		return err
 	}
 	req.Header.Set("User-Agent", userAgent())
-	if len(r.AuthorizationHeader) > 0 {
+	if (passCredentials || compareURLDomains(c.Icon, r.URL)) && len(r.AuthorizationHeader) > 0 {
 		req.Header.Set("Authorization", r.AuthorizationHeader)
 	}
 
@@ -826,7 +833,11 @@ func (f *fileImporter) fetchAndImportIcon(c models.Chart, r *models.RepoInternal
 	// TODO: make this configurable?
 	resizedImg := imaging.Fit(img, 160, 160, imaging.Lanczos)
 	var buf bytes.Buffer
-	imaging.Encode(&buf, resizedImg, imaging.PNG)
+	err = imaging.Encode(&buf, resizedImg, imaging.PNG)
+	if err != nil {
+		log.WithFields(log.Fields{"name": c.Name}).WithError(err).Error("failed to encode icon")
+		return err
+	}
 	b = buf.Bytes()
 	contentType = "image/png"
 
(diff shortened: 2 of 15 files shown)


cmd/apprepository-controller/pkg/apis/apprepository/v1alpha1/types.go
```diff
@@ -54,6 +54,8 @@ type AppRepositorySpec struct {
 	FilterRule FilterRuleSpec `json:"filterRule,omitempty"`
 	// (optional) description
 	Description string `json:"description,omitempty"`
+	// PassCredentials allows passing credentials to all domains
```
absoludity, line 57:
```suggestion
	// PassCredentials allows passing credentials with requests to other domains linked from the repository
```

cmd/asset-syncer/utils.go
```diff
@@ -236,11 +236,18 @@ func (r *HelmRepo) Charts(fetchLatestOnly bool) ([]models.Chart, error) {
 
 // FetchFiles retrieves the important files of a chart and version from the repo
 func (r *HelmRepo) FetchFiles(name string, cv models.ChartVersion) (map[string]string, error) {
+	authorizationHeader := ""
+	chartTarballURL := chartTarballURL(r.RepoInternal, cv)
+
+	if (passCredentials || compareURLDomains(chartTarballURL, r.URL)) && len(r.AuthorizationHeader) > 0 {
```
absoludity, line 242: Tiny point, but I'd generally have the `len(r.AuthorizationHeader) > 0` condition first so the rest is only evaluated if there's authorization?

cmd/asset-syncer/utils.go
```diff
@@ -783,7 +790,7 @@ func (f *fileImporter) fetchAndImportIcon(c models.Chart, r *models.RepoInternal
 		return err
 	}
 	req.Header.Set("User-Agent", userAgent())
-	if len(r.AuthorizationHeader) > 0 {
+	if (passCredentials || compareURLDomains(c.Icon, r.URL)) && len(r.AuthorizationHeader) > 0 {
```
absoludity, line 793: ditto here.

cmd/asset-syncer/utils.go
```diff
@@ -826,7 +833,11 @@ func (f *fileImporter) fetchAndImportIcon(c models.Chart, r *models.RepoInternal
 	// TODO: make this configurable?
 	resizedImg := imaging.Fit(img, 160, 160, imaging.Lanczos)
 	var buf bytes.Buffer
-	imaging.Encode(&buf, resizedImg, imaging.PNG)
+	err = imaging.Encode(&buf, resizedImg, imaging.PNG)
+	if err != nil {
+		log.WithFields(log.Fields{"name": c.Name}).WithError(err).Error("failed to encode icon")
+		return err
+	}
```
absoludity, line 840: Nice catch for the missing error.

cmd/asset-syncer/utils.go
```diff
@@ -871,3 +882,21 @@ func (f *fileImporter) fetchAndImportFiles(name string, repo Repo, cv models.Cha
 	// entry if digest has changed
 	return f.manager.insertFiles(chartID, chartFiles)
 }
+
+// Check if two URL strings are in the same domain.
+// Return true if so, and false otherwise or when an error occurs
+func compareURLDomains(url1Str, url2Str string) bool {
```
absoludity, line 888: Slight preference for a name that indicates the boolean nature of this, such as `isURLDomainEqual` or `hostsEqual`.

cmd/asset-syncer/utils.go
```diff
@@ -871,3 +882,21 @@ func (f *fileImporter) fetchAndImportFiles(name string, repo Repo, cv models.Cha
...
+		return false
+	}
+	url2, err := url.ParseRequestURI(url2Str)
+	if err != nil {
+		return false
+	}
+
+	a := url1.Hostname()
+	b := url2.Hostname()
+	// return url1.Hostname() == url2.Hostname()
```
absoludity, line 900: Guessing this is what you intended to return rather than using the `a` and `b` unnecessarily?

But related to this, I reckon we should also ensure that the proto is the same (ie. both https). If the index.yaml was listed as https://example.com/index.yaml but referred to an icon at http://example.com/icons/foo.png, I'd hope we wouldn't assume it's the same host (ie. same host name but different proto). Thoughts?

cmd/asset-syncer/utils_test.go
```diff
@@ -430,6 +450,52 @@ func Test_fetchAndImportIcon(t *testing.T) {
...
+		pgManager, _, cleanup := getMockManager(t)
+		defer cleanup()
+		netClient := &goodAuthenticatedHTTPClient{}
+
+		fImporter := fileImporter{pgManager, netClient}
+		passCredentials = false
+		assert.Err(t, fmt.Errorf("401 %s", charts[0].Icon), fImporter.fetchAndImportIcon(charts[0], repo))
+	})
+
+	// t.Run("valid icon (passing through if same domain)", func(t *testing.T) {
```
absoludity, line 473: Not sure why this is commented out?

cmd/asset-syncer/utils_test.go
```diff
@@ -1235,3 +1301,45 @@ func Test_filterCharts(t *testing.T) {
...
+			"https://bitnami.com/bitnami/index.yaml",
+			"https://charts.bitnami.com/assets/stacks/airflow/img/airflow-stack-220x234.png",
+			false,
+		},
+		{
+			"it returns true if they are under the same",
+			"https://charts.bitnami.com/bitnami/airflow-10.2.0.tgz",
+			"https://charts.bitnami.com/assets/stacks/airflow/img/airflow-stack-220x234.png",
+			true,
+		},
```
absoludity, line 1335: possibly tests for different proto?

cmd/asset-syncer/utils_test.go
```diff
@@ -1235,3 +1301,45 @@ func Test_filterCharts(t *testing.T) {
...
+			"https://charts.bitnami.com/bitnami/index.yaml",
+			"https://bitnami.com/assets/stacks/airflow/img/airflow-stack-220x234.png",
+			false,
+		},
+		{
+			"it returns false if they are under different subdomains",
+			"https://bitnami.com/bitnami/index.yaml",
+			"https://charts.bitnami.com/assets/stacks/airflow/img/airflow-stack-220x234.png",
+			false,
+		},
```
absoludity, line 1329: These two tests appear to be testing the same thing? Also, this is interesting - the subdomain. Is this the same behaviour in the helm patch? Asking because it's much more likely that I'd have my assets on a subdomain of the repo... but fine either way.

absoludity (approved):
Excellent neat change from CR through to UX :)

cmd/asset-syncer/utils.go
antgamdia: I can't take the credit here :P  the merit is on vscode and https://github.com/golangci/golangci-lint this time:

image

antgamdia: Mmm I'm dropping a card in the next interaction discussion for considering addressing all linter issues at once and adding it as part of the build process of our go dockerfiles

cmd/asset-syncer/utils.go
antgamdia: Yep, just a leftover when running the debugger :P 
Good catch! We have to verify the URL scheme as well, of course! 

By the way, there is an interesting go testing feature that will get released in 1.17: Fuzzing tests: https://blog.golang.org/fuzz-beta
Testing this method would be an interesting use case indeed.

cmd/asset-syncer/utils_test.go
antgamdia: Sorry, I commented all of them out so that running the test suite was quicker. I forgot to enable this one again :grimacing: 
Thanks for pointing out!

cmd/asset-syncer/utils_test.go
antgamdia: They are using the same approach (below), so I think we should stick close to it. If a user happens to really need to pass the credentials down to a different subdomain, we may consider adding a `--allowed-pass-credentials-subdomains` flag

```
(u1.Scheme == u2.Scheme && u1.Host == u2.Host)
```

Side note: we were using `.Hostname()`, which yields the `Host` value without ports. Having noticed they also compare the ports, I've changed the PR's code to match their appraoch.

PR #3066, opened 2021-07-01
Update to support k8s 1.21 requiring dex update.

absoludity:
### Description of the change

I updated my local kind installation (to v0.11.1, so supporting k8s 1.21) only to find that it breaks the dev env.

The reason is a change in 1.21 which results in the version of dex we were using crash-looping.

Since we were using the deprecated (and no longer updated) stable repo for dex, I updated to the current dex chart repo and got things working with the latest version of that chart instead.

### Benefits

Dev environment works with k8s 1.21.

Updated dex version (only visible difference is that it now lets the user choose to grant the email scope).

### Possible drawbacks

None that I can think of.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -422,6 +422,7 @@ install_multicluster_deps: &install_multicluster_deps
       helm install dex dex/dex --version 0.5.0 --namespace dex --values ./docs/user/manifests/kubeapps-local-dev-dex-values.yaml
 
       # Install openldap
+      helm repo add stable https://charts.helm.sh/stable
       kubectl create namespace ldap
       helm install ldap stable/openldap --namespace ldap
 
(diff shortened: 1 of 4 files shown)


antgamdia:
Oh, thanks for the change! I would have liked to address it when revamping the CO config flor the multicluster env, but I didn't.

AFAIK, you probably need to also change the circleci config. In fact, a CI check is failing. 

+1 when circle ci wants

antgamdia (approved):

.circleci/config.yml
```diff
@@ -415,11 +415,11 @@ install_multicluster_deps: &install_multicluster_deps
     name: "Install multicluster deps"
     command: |
       sed -i -e "s/172.18.0.2/$DEFAULT_DEX_IP/g;s/localhost/kubeapps-ci.kubeapps/g" ./docs/user/manifests/kubeapps-local-dev-dex-values.yaml
-      helm repo add stable https://charts.helm.sh/stable
+      helm repo add dex https://charts.dexidp.io
```
antgamdia, line 418: After removing `stable`, we should also move `openldap`:
`error: failed to download "stable/openldap" (hint: running `helm repo update` may help)`

absoludity: Yeah, I didn't want to try to change ldap. Just updated and fired off yesterday before moving onto something else, and forgot that we'll need stable still. It'll still fail after updating I expect as there's a new UI step in dex when logging in (giving consent to email being passed to 3rd party).

Issue #2695, opened 2021-04-20
Allow configuring pinniped-proxy with per-cluster namespace and suffix


antgamdia:
Follow-up from https://github.com/kubeapps/kubeapps/issues/2662#issuecomment-822879943

### Description:

Currently, Pinniped-Proxy is being configured globally in the values.yaml (see excerpt below). This config allows editing the API suffixes and the namespace in which Pinniped has been installed. This approach will work as long as every cluster is configured identically. 

```yaml
pinnipedProxy:
  defaultPinnipedAPISuffix: pinniped.tmc.cloud.vmware.com
  defaultPinnipedNamespace: vmware-system-tmc
  enabled: true
```

What if a cluster is using the TMC Pinniped version and we want to attach another cluster that uses de default one? Currently. it is not possible unless we make configurable this suffix and namespace per-cluster.

Edit: here, actually, we have two problems:

- [ ] Use any API suffix in Pinniped proxy. It was the main goal of my former PR https://github.com/kubeapps/kubeapps/pull/2515, but we weren't able to pass the at runtime, but at compile time. This ended up in two hardcoded API groups (the usual one and the TMC one). The crux is here: https://github.com/kubeapps/kubeapps/blob/master/cmd/pinniped-proxy/src/pinniped.rs#L47
- [ ] Make this information configurable per-cluster. This is the main problem this issue is about.

absoludity:
Thanks. Note that this is already planned in the code in that the config structs already contain placeholders, per-cluster, for pinniped-namespace, authenticator type and authenticator name (but not yet api suffix), they're just not being used per-cluster: https://github.com/kubeapps/kubeapps/blob/v2.3.1/pkg/kube/kube_handler.go#L85-L89

antgamdia:
Have a look at the new features in 0.55 and 0.53 
Some refs:

- https://github.com/clux/kube-rs/blob/master/CHANGELOG.md
- https://github.com/clux/kube-rs/blob/master/examples/dynamic_pod.rs
- https://github.com/clux/kube-rs/blob/master/examples/dynamic_api.rs

PR #3067, opened 2021-07-01
bumped kube version

dlaloue-vmware:
### Description of the change

upgraded kube crate from version 0.55.0 to 0.57.0.
refactored code for backward incompatible changes

### Benefits

upgraded to new version to use dynamic kube objects

### Possible drawbacks

none

### Applicable issues

issue #2695: Allow configuring pinniped-proxy with per-cluster namespace and suffix


antgamdia (approved):
Great! +1 if the unit tests pass and a, in a manual setup, it works as usual.
Thanks for upgrading!

absoludity (approved):
Looks like a unit-test is failing after the update? (Also odd that we didnt' get an auto-bump PR like #2863 ). But yes, if the test failure is trivial to fix and you've verified it works in real life, +1.

PR #3074, opened 2021-07-02
Add pagination options for GetAvailablePackageSummariesRequest

absoludity:
### Description of the change

In addition to adding the proto msg, this also does some small changes which I noticed while preparing to test and add the pagination.

I'll comment in-line with explanations.

### Benefits

Just a prequel to the actual pagination change (which I was aiming to do but ran out of road on Friday afternoon :) ).

### Applicable issues

  - Ref #2993


cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -186,9 +186,7 @@ func AvailablePackageSummaryFromChart(chart *models.Chart) (*corev1.AvailablePac
 		Identifier: chart.ID,
 		Plugin:     GetPluginDetail(),
 	}
-	if chart.Repo != nil {
-		pkg.AvailablePackageRef.Context = &corev1.Context{Namespace: chart.Repo.Namespace}
-	}
+	pkg.AvailablePackageRef.Context = &corev1.Context{Namespace: chart.Repo.Namespace}
```
absoludity, line 189: `chart.Repo` is now required for a valid chart... see below.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -326,23 +324,30 @@ func (s *Server) hasAccessToNamespace(ctx context.Context, namespace string) (bo
 // isValidChart returns true if the chart model passed defines a value
 // for each required field described at the Helm website:
 // https://helm.sh/docs/topics/charts/#the-chartyaml-file
+// together with required fields for our model.
```
absoludity, line 327: This function is checking that we have all the data expected and required to be able to identify the chart. So yes, the helm spec for the fields is great for things like description and name etc., but there are other fields on the model, specific to Kubeapps' model of a chart, such as ID and Repo, which are required to be able to identify/reference the chart.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server_test.go
```diff
@@ -199,33 +199,80 @@ func TestIsValidChart(t *testing.T) {
...
+				Repo: &models.Repo{
+					Name:      "bar",
+					Namespace: "my-ns",
+				},
+				ChartVersions: []models.ChartVersion{
+					{
+						Version: "3.0.0",
+					},
+				},
+			},
```
absoludity, line 215: I *think* this is more than a personal preference: unit tests are the one place where DRY is not always better. Having to scroll up and down the file to find what `chartOK` is reduces the readability of the tests. So throughout the tests I was looking at, I've updated with more verbose but hopefully more obvious input/output.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server_test.go
```diff
@@ -366,17 +471,56 @@ func TestGetAvailablePackageSummaries(t *testing.T) {
 					Cluster:   "",
 					Namespace: globalPackagingNamespace,
 				},
-				FilterOptions: &corev1.FilterOptions{
-					Query:        "",
-					AppVersion:   "",
-					PkgVersion:   "",
-					Categories:   nil,
-					Repositories: nil,
```
absoludity, line 374: Not sure why the FilterOptions are being specified here if we're not testing them at all (yet).

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server_test.go
```diff
@@ -366,17 +471,56 @@ func TestGetAvailablePackageSummaries(t *testing.T) {
...
+					ID:   "repo-1/chart-2",
+					Repo: &models.Repo{
+						Name:      "repo-1",
+						Namespace: "my-ns",
+					},
+					ChartVersions: []models.ChartVersion{
+						{
+							Version: "2.0.0",
+						},
+					},
```
absoludity, line 500: Updated with data so the set has more than one member :)

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server_test.go
antgamdia: Mmm, I see... yep, I kind of agree with it, but still not 100% convinced :-P
The point is: what if we need and use the "concept of a perfect chart model" across different tests? Do we really need to write it again and again? It could be more readable, that's for sure... but, when one has to perform a change.. it tedious to replicate it over every test suite.

However, no strong opinion whatsoever. Let's go with the verbose way for consistency.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server_test.go
antgamdia: I just tried to initialize every field for having a clear picture of what this model is comprised of. Then I deleted those unused fields... but not this one it seems. It's just a leftover, thanks for removing it!

antgamdia (approved):
Great! The first step on having our API paginated by default. Thanks!

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server_test.go
absoludity: What I've done in the past in that situation, when you begin to have the same large data struct in lots of places, is to instead define a function like `createValidChart(options...)` which can be called without options, or specific options to set specific things. Not sure if that helps here, but I'll take a look in the next PR and we can discuss there.

PR #3083, opened 2021-07-05
Add makeChart test helper for explicit yet DRY tests.

### LINKED ISSUE #2993: [kubeapps-apis] Implement "direct-Helm" plugin based on the existing features in kubeapps

absoludity:
### Description of the change

Just addressing a comment on the previous PR at:

https://github.com/kubeapps/kubeapps/pull/3074#discussion_r662823041

### Benefits

Tests now are explicit (you can see exactly what details are used for input) while being DRY.

### Applicable issues

  - fixes #2993

### DIFF
--- a/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server_test.go
+++ b/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server_test.go
@@ -376,6 +376,7 @@ func TestAvailablePackageSummaryFromChart(t *testing.T) {
 	}
 }
 
+// makeChart makes a chart with specific input used in the test and default constants for other relevant data.
 func makeChart(chart_name, repo_name, namespace string, chart_versions []string) *models.Chart {
 	ch := &models.Chart{
 		Name:        chart_name,
--- a/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server_test.go
+++ b/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server_test.go
@@ -43,56 +43,11 @@ import (
 	log "k8s.io/klog/v2"
 )
 
-const globalPackagingNamespace = "kubeapps"
-
-var chartOK = &models.Chart{
-	Name:        "foo",
-	ID:          "foo/bar",
-	Category:    "cat1",
-	Description: "best chart",
-	Icon:        "foo.bar/icon.svg",
-	Repo: &models.Repo{
-		Name:      "bar",
-		Namespace: "my-ns",
-	},
-	Maintainers: []chart.Maintainer{{Name: "me", Email: "me@me.me"}},
-	ChartVersions: []models.ChartVersion{
-		{Version: "3.0.0", AppVersion: "1.0.0", Readme: "chart readme", Values: "chart values", Schema: "chart schema"},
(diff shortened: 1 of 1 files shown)


cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server_test.go
```diff
@@ -597,9 +543,9 @@ func TestGetAvailablePackageSummaries(t *testing.T) {
 					Namespace: "",
 				},
 			},
-			charts:           []*models.Chart{{Name: "foo"}},
-			expectedPackages: []*corev1.AvailablePackageSummary{},
-			statusCode:       codes.Unimplemented,
+			expectDBQuery: false,
```
absoludity, line 546: This is the test that was previously setting a mock expectation that was never met until the subsequent test tried to run (it just happened to match as an internal error is returned if the expectation doesn't match).

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server_test.go
```diff
@@ -845,9 +844,6 @@ func TestGetAvailablePackageDetail(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			currentExpectedPackage := tc.expectedPackage
-			currentExpectedPackage.PkgVersion = tc.expectedVersion
```
absoludity, line 849: The line above seemed to be added to *update* the expected package, rather than just having the expected package with the correct version (I assume because a pre-created expected detail package was used).

antgamdia (approved):
Thanks a lot for the changes. Much better now!

PR #3090, opened 2021-07-05
Pre release updates

antgamdia:
### Description of the change

According to the process described at https://github.com/kubeapps/kubeapps/blob/master/docs/developer/release-process.md, this PR brings several updates; namely:

- Upgrade base image versions (in CI, Dockerfiles and chart): Go (1.165,) Node (14.17.2), Rust (1.53)
- Upgrade UI dependencies (**note** this step wasn't fully achieved due to errors that will be addressed in a separated issue)
- Upgrade go dependencies
- Upgrade rust dependencies
- Upgrade chart dependencies
- Upgrade binaries used in CI:
    - Kubectl, OLM, Kind, Github CLI
- Upgrade versions used in the integration image (bumped it up to 1.0.1)

### Benefits

Everything up to date for the upcoming release.

### Possible drawbacks

Unexpected incompatibilities? Anyway, we perform manual tests prior to the release, so, if any, it should be caught by our test suite or the manual test.

### Applicable issues

N/A

### Additional information

N/A


chart/kubeapps/Chart.yaml
```diff
@@ -25,4 +25,4 @@ maintainers:
 name: kubeapps
 sources:
   - https://github.com/kubeapps/kubeapps
-version: 7.1.3-dev1
+version: 7.1.4-dev0
```
absoludity, line 28: I expected that we'd do this manual change as part of the auto-generated PR triggered when the chart change lands at bitnami/charts and is synced back to our repo? But no issue, it's only dev related.

docs/developer/release-process.md
```diff
@@ -109,6 +109,7 @@ yarn upgrade
 ```bash
 go mod tidy
 go list -u -f '{{if (and (not (or .Main .Indirect)) .Update)}}{{.Path}}: {{.Version}} -> {{.Update.Version}}{{end}}' -m all 2> /dev/null
+go get -u ./...
```
absoludity, line 112: This conflicts with the next line, I think? Assuming the above automatically updates all go dependencies? Perhaps if it works as you've documented (this time), include it in the following sentence as an option if there are no issues. See what you think.

absoludity (approved):
Thanks Antonio.

> ### Possible drawbacks
> 
> Unexpected incompatibilities? Anyway, we perform manual tests prior to the release, so, if any, it should be caught by our test suite or the manual test.

I wonder if we should rather do this early *after* a release, rather than right before a release? WDYT? Might be less stressful. Can also be handed as a task to a person who *didn't* do the release :P .

chart/kubeapps/Chart.yaml
antgamdia: Yes, generally we would change the chart version in the bitnami/chart repo, however, as we have perform some updates in the local chart (nginx 1.19 to 1.20, or re-generate the readme based on the lastest `redminator` tool version), I assumed we should also modify the local version (regardless of the change we also would need to to in the bitnami repo)

docs/developer/release-process.md
antgamdia: I misread it when I wrote the guide, the `go list -u...` is just for _To view available minor and patch upgrades only for the direct dependencies_

And:

`go get -u ./... `to use the latest minor or patch releases (and add -t to also upgrade test dependencies)
`go get -u=patch ./... `to use the latest patch releases (and add -t to also upgrade test dependencies)

So let me reword the text to make it more explicit. Thanks for pointing it out.

PR #3094, opened 2021-07-06
CI - Fix rollback test

antgamdia:
### Description of the change

In the rollback test, we were performing the rollback action prior to check if the app was fully deployed. It caused some inconsistencies; especially, the rollback dialog showed "loading, please wait". Wherefore the expected text never appeared.

This PR is for checking the app is fully deployed and the rollback dialog is loaded before checking the revision version.

Every time the tests passed... it was like planet conjunction :P

### Benefits

The CI (should) pass again.

### Possible drawbacks

N/A

### Applicable issues

N/A
### Additional information

N/A


antgamdia:
Rubber-stamping this PR since it just implies required changes on the CI

PR #3111, opened 2021-07-08
A little extra help since we're requiring signed-off commits

absoludity:
### Description of the change

Since we say in the CONTRIBUTING.md guidelines that we require DCO sign-off on all commits, I've added the DCO bot to check PRs.

This update just has a little extra info to help when people find they need to do sign-off retrospectively, as well as ensuring it's in all commits for the project if they want.

### Benefits

Avoid frustration with DCO sign-off.

### Possible drawbacks

None that I can see.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -70,3 +70,14 @@ By making a contribution to this project, I certify that:
     maintained indefinitely and may be redistributed consistent with
     this project or the open source license(s) involved.
 ```
+
+If you find yourself in the position where you've created a pull-request and it's failing the DCO check because some of your commits are not signed off, you can just follow the details of the DCO check failure to sign-off on those commits with a single command, such as:
+
+```bash
+git rebase HEAD~N --signoff
+git push --force-with-lease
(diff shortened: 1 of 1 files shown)


antgamdia (approved):
Great, thanks for the information. I don't know if we should directly move to CLA instead of DCO, but, since we are requiring DCO atm, it's a great addition.

antgamdia:
I think there's an easier way to achieve it by appending `--signoff`

`git commit -m "whatever" --signoff`

https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---signoff

absoludity:
> I think there's an easier way to achieve it by appending `--signoff`
> 
> `git commit -m "whatever" --signoff`

Yes, that's already mentioned earlier. The new section is about (1) how to deal with a situation where you've already got a bunch of commits and (2) how to avoid needing to remember.

As mentioned, I've switched the DCO bot off again to avoid churn of processes until next iteration when we discuss, but will land this just in case we decide to re-enable it.

PR #3118, opened 2021-07-08
Httpclient

dlaloue-vmware:
### Description of the change

Issue 3045 reported that the files http_utils.go and httpclient.go might need to be merged as they share similar code and logic.
the change is keeping both files but makes httpclient.go specific to HTTP Client and http_utils.go focus on the kube aspect of configuration (secrets, app repository...).

changes include:
 - moving Client interface/struct from http_utils.go to httpclient.go
 - refactoring some fo the logic into more granular functions for better composability
 - fixed a few places where http Transport defaults were lost 

notes:
 - http_utils.go still need to be renamed as it is not http focused anymore
 - will add tests in httpclient.go in next commit

### Benefits

removed duplication of code
better separation of responsibilities

### Possible drawbacks

none

### Applicable issues

Issue 3045: combine pkg/kube/http_utils.go and pkg/http-client/httpclient.go into one


pkg/http-client/httpclient.go
```diff
@@ -54,36 +75,60 @@ func NewWithCertFile(certFile string, skipTLS bool) (*http.Client, error) {
...
-			InsecureSkipVerify: skipTLS,
-			RootCAs:            caCertPool,
-		}}
-	return client, nil
+	return caCertPool, nil
+}
+
+// configure the given proxy on the given client
+func SetClientProxy(client *http.Client, proxy func(*http.Request) (*url.URL, error)) {
+	client.Transport.(*http.Transport).Proxy = proxy
```
absoludity, line 123: Generally, it's always best to use the "comma, ok" idiom when doing an interface conversion. I get that here your assumption is that the transport is *always* an http transport, but code changes in the future can be unexpected. In this case, I don't see any issue with updating `SetClientProxy` to return an error and doing something like:
```suggestion
	transport, ok := client.Transport.(*http.Transport)
	if !ok {
		return fmt.Errorf("transport was not an http.Transport")
	}
	transport.Proxy = proxy
	return nil
```

Or, perhaps even better as another option: afaict every call to this and SetClientTLS and SetClientProxy is directly after calling `New()`, both here and in the other packages. In which case, the TLS and proxy data could just be part of the New() constructor? (as in, it doesn't seem to be required for updates)

Probably easiest to just use the transport as the arg, something like `New(transport *http.Transport)`. That might be neater if there's no reason against it? Thoughts?

absoludity (approved):
Generally looks great Dimitri. Question about whether we even need interface conversions inline. Thanks!

pkg/http-client/httpclient.go
dlaloue-vmware: i agree with making the code more explicit in term of type checking. I am also writing tests to ensure the type is as expected.

SetClientTLS and SetClientProxy are currently always after a New in the code, though i had a version of the code where it was not. Users should be free to structure their code flow as they see fit. So we cannot assume this will always be the case.

The main reason i added methods is to help ensure that the defaults are not lost. In the previous code, methods creating a client would always create their own Transport from scratch (equivalent to the suggestion of passing it as a constructor arg) and defaults would often be lost, or one would have to replicate the entire default configuration (not maintainable).

dlaloue-vmware: i have added code to to check for the type conversion. but checking for err each time the methods are called seemed verbose.
let me know if this is as expected or whether there is a better way to write the code.

absoludity: Yep, up to you. I agree that it's much better (obviously) to only update the part you want to update, I'm just not sure there's a use-case currently for updating parts of a transport separately so was keen to avoid the type assertion. But fine either way. Still +1.

absoludity: > let me know if this is as expected or whether there is a better way to write the code.

Only avoiding type assertions :P . Otherwise, fine with the error check (yes, go is verbose in this way generally... preferring explicit error handling with the cost of verbosity).

Issue #3146, opened 2021-07-12
Implement further direct helm plugin features


absoludity:
### Description:

In #2993 we implemented GetAvailablePackageSummaries, GetAvailablePackageDetail, and GetAvailablePackageVersions. The remaining tasks are:

- [x] GetInstalledPackageSummaries
- [x] Update GetAvailablePackageDetail to include readme and values etc.
- [x] GetInstalledPackageDetail
- [x] CreateInstalledPackage
- [x] UpdateInstalledPackage
- [x] DeleteInstalledPackage

which we'll add to the direct helm backend plugin while agreeing on the messages in the doc (#2944).

antgamdia:
Let me reopen this issue as I think it was accidentally closed via PR (if only we already implemented every operation... :P)

antgamdia:
Closed in favor of https://github.com/kubeapps/kubeapps/issues/3340

absoludity:
> Closed in favor of #3340

Not sure I want to close this in favour of #3340. This one here is a unit of work that we can effectively complete in the next iteration (I would think). Whereas #3340 is a summary of a whole bunch of work that we are in no place to complete in the next iteration. I think it's important to be able to break things down and assign smaller parts to get done?

antgamdia:
> Whereas #3340 is a summary of a whole bunch of work that we are in no place to complete in the next iteration. I think it's important to be able to break things down and assign smaller parts to get done?

I do agree here. I just thought we might want to create two issues for tracking each operation during this iteration and left the #3340 as the only high-level issue. But, yep, if we want this issue to be the unit of work for this iteration, I'm totally ok with it!

PR #3147, opened 2021-07-12
Add messages for GetInstalledPackageSummaries and generate unimplemented client methods.

absoludity:
### Description of the change

Adds the messages from the design doc for the GetInstalledPackageSummaries call. I've noted clearly with comments on the PR where I differed and updated the messages (both here and on the design doc) from our last discussion about them.

I'm planning to do each addition as a separate PR so we can review the fields piecemeal here and if there's need for a larger discussion I'll organise a call and work through it together on the design doc. I expect that this will be necessary only for the later CreateInstalledPackage work, though perhaps also for the GetInstalledPackageDetail.

### Benefits

Enables implementation of the direct helm implementation of GetInstalledPackageSummaries.

### Applicable issues

  - ref #3146

### Additional information

I'll comment inline on the couple of changes I made to the design doc while creating this PR.


cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
```diff
@@ -108,6 +109,19 @@ message GetAvailablePackageVersionsRequest {
   string pkg_version = 2;
 }
 
+// GetInstalledPackageSummariesRequest
+//
+// Request for GetInstalledPackageSummeraies
+message GetInstalledPackageSummariesRequest {
+  // The context (cluster/namespace) for the request.
+  Context context = 1;
+
```
absoludity, line 118: We don't currently support filtering of installed packages at all, so I've removed the TODO from the design doc for here in the request (and in the doc).

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
```diff
@@ -306,6 +333,67 @@ message AvailablePackageDetail {
...
+  // InstalledPackageReference
+  //
+  // A reference uniquely identifying the package.
+  InstalledPackageReference installed_package_ref = 1;
+
+  // Name
+  //
+  // A name given to the installation of the package.
+  string name = 2;
```
absoludity, line 354: I added `name` here - the name of the installation summary (and in the design doc), because we'd forgotten to add it after we switched the `name` field in the installed package ref to a more opaque identifier a week or two ago (which we did to allow plugins to avoid ambiguous references based on the name).

gfichtenholt (approved):
I don't see anything glaringly wrong, so I approve. I am a little confused about the sequence of things though. We agreed on 3-4 APIs having to do with available packages. At the time we punted on other stuff  (install-related), as in didn't even discuss it, because we said (a) we were tired cuz (b) it takes a long time for all to agree on the perfect API and (c) the 3 -4 APIs we did agree on were sufficient to get some of the UX work started, like the catalog. So I thought that's what would come next. Instead you seemed to have gone ahead and added APIs we haven't discussed and agreed upon. Seems a little out of order, but I don't want to block you from doing this so go ahead

antgamdia (approved):
LGTM as well. I'm stating feeling that the current design doc is becoming quite messy and something not really useful.
I don't know if, perhaps, we can have a different approach for the upcoming operations: rather than discussing in gdocs, we could bring the discussion to a PR (that allows suggestions and conversations). 
Perhaps, opening a PR per operation (without any autogeneration, just the proto), could be an option. This way, we could resove the comments and add suggestions directly on the code (removing, thus, an extra step)
See what you think. Anyway, +1 in this PR, as Greg said

absoludity:
> I don't see anything glaringly wrong, so I approve. I am a little confused about the sequence of things though. We agreed on 3-4 APIs having to do with available packages.

Correct, and they are now implemented and landed for the direct helm plugin, so we can start there in the UI now.

I'm keen to be able to continue, so started looking at the next one which from memory we did actually get through (GetInstalledPackageSummaries) though we did indeed then realise we weren't going to get through the rest of the installed stuff, so we stopped and agreed to go ahead only with the subset - the available package API calls. So when I started looking at this one, it did indeed seem quite straight forward, so I thought a PR may suffice to identify anything for this one, though clearly the other calls (GetInstalledPackageDetail and especially the CreateInstalledPackage) will require higher bandwidth discussion.

> At the time we punted on other stuff (install-related), as in didn't even discuss it, because we said (a) we were tired cuz (b) it takes a long time for all to agree on the perfect API and (c) the 3 -4 APIs we did agree on were sufficient to get some of the UX work started, like the catalog. So I thought that's what would come next.

Yes, those API calls are implemented now for direct helm, which enables moving that part forward in the UX, which Antonio is looking at. While he's looking at the UI side, I plan to start moving the remaining APIs forward.

> Instead you seemed to have gone ahead and added APIs we haven't discussed and agreed upon. Seems a little out of order, but I don't want to block you from doing this so go ahead

I've not added them yet - this PR was to discuss the first ones before I'd go ahead and implement them in direct helm. I'm more than happy to have a call to go through the messages for GetInstalledPackageSummaries if it's needed. Let's organise one.

absoludity:
> LGTM as well. I'm stating feeling that the current design doc is becoming quite messy and something not really useful.

Please do try to help improve the doc. The only part I'm currently focusing on is the messages as it is what's blocking us. I think it is useful to first think about the bigger picture of all the messages, but it is a lot of context to hold. I'll go through the doc today and try to resolve any outstanding conversations. Or what in particular, about the messages for the API, are you finding messy if it's not the comments/conversations? (I agree other sections, some of which are us dumping thoughts for unresolved aspects specific to particular plugins, are tricky to clean).

> I don't know if, perhaps, we can have a different approach for the upcoming operations: rather than discussing in gdocs, we could bring the discussion to a PR (that allows suggestions and conversations).

That's exactly what I was trying to do here for `GetInstalledPackageSummaries`. Do you have any discussion points for this one?

> Perhaps, opening a PR per operation (without any autogeneration, just the proto), could be an option. This way, we could resove the comments and add suggestions directly on the code (removing, thus, an extra step)

This change here is just the proto and the related autogeneration (which is usually hidden in the diff anyway). I'm happy to remove the generated changes if that helps.

> See what you think. Anyway, +1 in this PR, as Greg said

Thanks.

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
```diff
@@ -306,6 +333,67 @@ message AvailablePackageDetail {
...
+  //
+  // A reference uniquely identifying the package.
+  InstalledPackageReference installed_package_ref = 1;
+
+  // Name
+  //
+  // A name given to the installation of the package (eg. "my-postgresql-for-testing").
+  string name = 2;
+
+  // PkgVersionReference
```
gfichtenholt, line 356: PkgVersionReference => VersionReference?

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
```diff
@@ -306,6 +333,67 @@ message AvailablePackageDetail {
...
+
+  // CurrentPkgVersion
+  //
+  // The version of the package which is currently installed.
+  string current_pkg_version = 4;
+
+  // Installed package icon URL
+  //
+  // A url for an icon.
+  string icon_url = 5;
```
gfichtenholt, line 370: fluxv2 HelmRelease doesn't have a field for this. Would the plug-in leave that empty or have some logic to derive it from available package?

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
```diff
@@ -306,6 +333,67 @@ message AvailablePackageDetail {
...
+
+  // PackageDisplayName
+  //
+  // The package name as displayed to users (provided by the package, eg. "PostgreSQL")
+  string pkg_display_name = 6;
+
+  // ShortDescription
+  //
+  // A short description of the package (provided by the package)
+  string short_description = 7;
```
gfichtenholt, line 380: flux helmrelease doesn't provide these, so I'll need to pull these 2 off the corresponding helmchart? Just making sure

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
```diff
@@ -306,6 +333,67 @@ message AvailablePackageDetail {
...
+  // For example, if the pkg_version_reference is ">10.3.0 < 10.4.0" and 10.3.1
+  // is installed, then:
+  //   * if 10.3.2 is available, latest_matching_version should be 10.3.2, but
+  //   * if 10.4 is available while >10.3.1 is not, this should remain empty.
+  string latest_matching_pkg_version = 8;
+
+  // LatestPkgVersion
+  //
+  // The latest version available for this package, regardless of the pkg_version_reference.
+  string latest_pkg_version = 9;
```
gfichtenholt, line 394: Again, flux doesn't seem to provide these 2 in helmrelease.status field, which is where I'd expect to find them. Are these necessary to be able to display whether or not there is an upgrade available? If so, it *SEEMS* to me that the UX or core server can figure these out based on getAvailablePackageVersions. In fact, these two fields seem like they're about "available" more than "installed"

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
absoludity: This is the public name of the field on the go struct, not the type. It's confusing, but since the name in the protobuf below is `pkg_version_reference`, the corresponding Go struct is generated with go-naming conventions as `PkgVersionReference`. You can see that in the generated changes to cmd/kubeapps-apis/gen/core/packages/v1alpha1/packages.pb.go above.

Aside: So far, we have been documenting each field using the name of the field in the go struct, but as you'd expect, for the generated JS/Swagger, it's generated with JS naming conventions as `pkgVersionReference`, which you can see in the swagger changes above. The title for the swagger documentation is taken from this comment, so appears as `PkgVersionReference`. Perhaps we want to switch all our titles here to use the JS names for the swagger docs. Anyway, not something for here, imo.

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
absoludity: Yes the latter. The client (such as Kubeapps dashboard) wants to present to the user all the installed apps, including the icons (which come from the package, or in your specific case, the `HelmChart`). Similar for some other fields. Earlier on we looked at whether we wanted to instead include the available package nested in the installed package, but decided to use redundant fields on the installed package.

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
absoludity: Yes.

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
absoludity: > Again, flux doesn't seem to provide these 2 in helmrelease.status field, which is where I'd expect to find them. Are these necessary to be able to display whether or not there is an upgrade available?

Yes, whether an upgrade is available (whether that's in the current version specification, ie. should auto-upgrade unless auto-reconcilliation is paused, or outside the current version specification, in which case a manual upgrade is required if desired.

> If so, it _SEEMS_ to me that the UX or core server can figure these out based on getAvailablePackageVersions. In fact, these two fields seem like they're about "available" more than "installed"

They're about both, at least, the first one is: latest matching pkg version is a function of the available versions that match the specific package version reference used for this installation. Yes, the UX can indeed figure these out based on fetching the available versions for each item in the paginated list and doing the semver work, but it can be quite complex, while (a) it's part of what a client needs to display a list of installed packages to the user, and (b) the data is all available to the plugin. If we did want to remove these from here, I think we'd at least need an API function to get latest matching pkg versions for an input list of (available_pkg_summary_ref, pkg_version_ref) tuples, so only one request is required to get the data for N installed packages that are being displayed.

Thoughts?

absoludity (commented):
Thanks Greg. Comments below.

PR #3152, opened 2021-07-14
Add check for empty charts

### LINKED ISSUE #3150: Handle empty chart from index
### Description:
In some cases the index returns empty charts (i.e. https://chart.testarchitect.com/index.yaml) and the asset-syncer is unable to get the "deprecated" property so the execution goes in error.

1. Add https://chart.testarchitect.com as repo in kubeapps
2. Click on refresh for getting the latest charts
3. Check the log on the asset-syncer microservice

### Additional info
I've solved in my local enviroment adding a check for the empty entry in kubeapps/pkg/helm/index.go.

ram-pi:
### Description of the change
Avoid errors when a repo contains empty charts.

### Benefits
When an empty chart is present in a repo it will be skipped.

### Applicable issues
fixes #3150 

### Additional information
Tested with:
`go run ./cmd/asset-syncer/ sync --database-user=postgres --database-name=assets selenium https://chart.testarchitect.com helm`


pkg/helm/index.go
```diff
@@ -71,6 +71,12 @@ func ChartsFromIndex(contents []byte, r *models.Repo, shallow bool) ([]models.Ch
 		// almost 200 chart versions going all the way back many years to version "2.1.4".
 		// So for now, let's just keep track of the latest, not to overwhelm the caller with
 		// all these outdated versions
+
+		// skip if the entry is empty
+		if len(entry) < 1 {
+			continue
+		}
```
antgamdia, line 78: Perhaps we can add here a log trace like:
```go
		if len(entry) < 1 {
			log.Infof("skipping empty chart: [%s]", key)
			continue
		}
```

The `for` should include the `key` in this case

antgamdia (commented):
Thanks for the PR, it's great you have tested it locally. However, a unit test would be really useful to make avoif future regressions.

For instance, in the `index_test.go` you could add something like:

```golang
func Test_emptyChart(t *testing.T) {
	r := &models.Repo{Name: "test", URL: "http://testrepo.com"}
	charts, err := ChartsFromIndex([]byte(validRepoIndexYAML), r, false)
	assert.NoErr(t, err)
	assert.Equal(t, len(charts), 2, "number of charts")

	indexWithEmptyChart := validRepoIndexYAML + `
  emptyChart: []`
	charts, err = ChartsFromIndex([]byte(indexWithEmptyChart), r, false)
	assert.NoErr(t, err)
	assert.Equal(t, len(charts), 2, "number of charts")
	assert.Equal(t, len(charts[1].ChartVersions), 2, "number of versions")
}
```

pkg/helm/index.go
ram-pi: Yep, I'll add the unit test.
I'll also add the log message but we have not any info about the entry, if it is ok I could log only the skipping string :)

antgamdia: Great, thanks! I just tried replacing the `for` statement by `for key, entry := range index.Entries {`. In your example (below), I'm getting the key name (in this case, `selenium`). Anyway, no strong opinion here, it's just an info log message.

```yaml
entries:
  selenium: []
  selenium3:
```

ram-pi: Thanks for the hint.
It should be done with the last commit :)

antgamdia (approved):
LGTM :) Thanks for the contribution! 
Let's wait, though, until another maintainer (cc @absoludity) have the chance to look at it too.

pkg/helm/index_test.go
```diff
@@ -79,3 +82,11 @@ func Test_newChart(t *testing.T) {
 	assert.Equal(t, c.Repo, r, "repo set")
 	assert.Equal(t, c.ID, "test/wordpress", "id set")
 }
+
+func Test_loadRepoWithEmptyCharts(t *testing.T) {
```
antgamdia, line 86: Just wondering if adding the extra yaml file really worths it or we can just go with appending the empty chart (as in https://github.com/kubeapps/kubeapps/pull/3152#pullrequestreview-706194220) in an already valid yaml file.

ram-pi: It looked more clean to me but if you prefer I could rewrite the test using the same file, as you suggest.
I'm not such a good golang programmer so advices are welcome :)

ram-pi:
It seems that the pipeline goes in timeout on the use-cases/07-upgrade.js step. Any suggestions?

antgamdia:
> It seems that the pipeline goes in timeout on the use-cases/07-upgrade.js step. Any suggestions?

Don't worry, it's a known issue in our CI pipeline; sometimes these e2e tests eventually fail. Let me just re-run the Ci job and it should be green in a while :)

absoludity (approved):
Great, thanks @ram-pi

PR #3154, opened 2021-07-15
Re-instate the userAgent config when fetching charts.

### LINKED ISSUE #3076: Add userAgent to requests from kubeops service
### Description:

The `kubeops` service is currently setting an empty user agent. The old `tiller-proxy` service used to include a `version.go` which set a userAgent string to be used in the chart client:

https://github.com/kubeapps/kubeapps/blob/v1.11.3/cmd/tiller-proxy/version.go

but it seems this was never moved across to the kubeops service. As a result, the [`UserAgent` field of the `Options`](https://github.com/kubeapps/kubeapps/blob/master/cmd/kubeops/internal/handler/handler.go#L43) is not being set during [kubeops startup](https://github.com/kubeapps/kubeapps/blob/master/cmd/kubeops/main.go#L80), and the existing `userAgentComment` which is in the code is left unused. We should add a version.go similar to the old tiller-proxy one or the [current asset-syncer one](https://github.com/kubeapps/kubeapps/blob/master/cmd/asset-syncer/version.go).

Check the charts.bitnami.com logs and see only `tiller-proxy/v...` from old installations of Kubeapps. I suspect there will also be ones with empty userAgent strings.

absoludity:
See #3076 for more info.

### Applicable issues

  - fixes #3076

### DIFF
--- a/cmd/kubeops/main.go
+++ b/cmd/kubeops/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"net/http/httputil"
 	"net/url"
@@ -38,6 +39,9 @@ var (
 	userAgentComment       string
 	namespaceHeaderName    string
 	namespaceHeaderPattern string
+	// This version var is updated during the build (see the -ldflags option
+	// in the cmd/kubeops/Dockerfile)
+	version = "devel"
 )
 
 func init() {
@@ -86,6 +90,7 @@ func main() {
 		QPS:                    qps,
 		NamespaceHeaderName:    namespaceHeaderName,
 		NamespaceHeaderPattern: namespaceHeaderPattern,
+		UserAgent:              userAgent(),
 	}
 
 	storageForDriver := agent.StorageForSecrets
@@ -181,4 +186,16 @@ func main() {
 	srv.Shutdown(ctx)
 	log.Info("All requests have been served. Exiting")
 	os.Exit(0)
+}
+
+// Returns the user agent to be used during calls to the chart repositories
+// Examples:
+// kubeops/devel
+// kubeops/2.3.4 (kubeapps v2.3.4-beta4)
+func userAgent() string {
+	ua := "kubeops/" + version
+	if userAgentComment != "" {
+		ua = fmt.Sprintf("%s (%s)", ua, userAgentComment)
+	}
+	return ua
 }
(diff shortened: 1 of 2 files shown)


cmd/kubeops/main.go
```diff
@@ -182,3 +187,15 @@ func main() {
 	log.Info("All requests have been served. Exiting")
 	os.Exit(0)
 }
+
+// Returns the user agent to be used during calls to the chart repositories
+// Examples:
+// kubeops/devel
+// kubeops/2.3.4 (kubeapps v2.3.4-beta4)
+func userAgent() string {
```
antgamdia, line 195: Ah, it is the same approach as we are currently doing in other components. Great!

cmd/kubeops/main.go
```diff
@@ -38,6 +39,9 @@ var (
 	userAgentComment       string
 	namespaceHeaderName    string
 	namespaceHeaderPattern string
+	// This version var is updated during the build (see the -ldflags option
+	// in the cmd/kubeops/Dockerfile)
+	version = "devel"
```
antgamdia, line 44: Thanks for the comment here, I wasn't aware of what the -ldflags was doing actually

antgamdia (approved):
Some tests seem to be failing, +1 once they passed.
Besides, I wonder if we should add also a `main_test.go` as we already have in other components.

cmd/kubeops/main.go
absoludity: Yeah, I had to dig when checking how it was being updated for the images, so thought I'd save myself time when I look again in a year :)

cmd/kubeops/main.go
absoludity: Yep, though I've now updated so it's a function with args, rather than relying on the global params (so it's simple to test).

absoludity (commented):
> Some tests seem to be failing, +1 once they passed.

Hmm, the failing test is unrelated - it's in the fluxv2 plugin

```
FAIL	github.com/kubeapps/kubeapps/cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1	600.208s
```

though I can't reproduce it locally (tried a loop 100 times running those tests without cache) and it has passed when re-run now.  @gfichtenholt might be interested to [look at the error](https://app.circleci.com/pipelines/github/kubeapps/kubeapps/3637/workflows/0b918464-f2aa-466d-958b-6ac7d4470b6f/jobs/52748?invite=true#step-103-602), not sure if you're aware of it? (and it may disappear when your latest branch lands, so may not be worth looking at yet). Actually, it looks like the tests are hanging, which is something you've noted on your current branch as well.

> Besides, I wonder if we should add also a `main_test.go` as we already have in other components.

Yep, added and tested the new `getUserAgent` function.

gfichtenholt:
yeah that's an odd one. I haven't seen it myself. But I am re-writing the way cache watcher is done. Hopefully after the re-write this goes away for good

Issue #3085, opened 2021-07-05
Use direct-Helm plugin in the UI


antgamdia:
This issue is tracking the progress on making the current UI use the new grpc KubeappsAPIs.
It involves changes to migrate the current UI to use the new API (just displaying the catalog) plus changes in the operations and fields.

## Pending

- [x] [priority/higest] Get the tests to work again. I'd suggest a twofold approach:
    - [x] Ensure every e2e is passing ASAP
    - [x] Start fixing the failing unit tests, if too complex, skip. Prioritize coverage over test quality
- [ ] [priority/high] Implement operations for the Helm actions, namely:
    - [ ] CreateInstalledPackage
    - [ ] UpdateInstalledPackage
    - [ ] DeleteInstalledPackage
- [ ] [priority/medium] Implement the corresponding core operations for each op already implemented in the Helm direct plugin (I'd suggest something quick and then think about the pagination and dups handling later)
- [ ] [priority/low] Rename all the references to "charts" in the code to "package" or sth similar. Do it in a separate PR for reducing noise.

## Done

- [x] Add new operations (tracked at https://github.com/kubeapps/kubeapps/issues/3146)
- [x] Add minor changes "on the fly" (tracked at https://github.com/kubeapps/kubeapps/issues/3165)
  - Missing fields tracked at https://github.com/kubeapps/kubeapps/issues/3165#issuecomment-882943900
  - Missing secondary operations tracked at https://github.com/kubeapps/kubeapps/issues/3165#issuecomment-882944035
- [x] Perform the UI migration 
  - Required extra changes (eg, testing, refactor, etc) tracked at https://github.com/kubeapps/kubeapps/issues/3165#issuecomment-882945858
  - Actual UI migration divided per use case tracked at https://github.com/kubeapps/kubeapps/issues/3165#issuecomment-882950182

- [x]  Check, quickly, the current UI requirements (in terms of field required in the data model) using the non-grpc calls.
  - https://github.com/kubeapps/kubeapps/issues/3085#issuecomment-876196248
- [x] Investigate and add support for performing grpc calls from the client
  - https://github.com/kubeapps/kubeapps/pull/3155 + discussion in this issue

_(updated july 20th)_

antgamdia:
(Reposting it here for completeness from https://github.com/kubeapps/kubeapps/issues/2993#issuecomment-872414596)

I'm now addressing the UI changes. Since it is a complex endeavor, let's break this task down into three different milestones:

[...]

PR #3155, opened 2021-07-15
Add grpcweb stack

antgamdia:
### Description of the change

This PR brings a number of changes aimed at making it possible to consume grpc via frontend in a really easy way (by means of some autogenerated code from a proto file). Mainly:

- Add support for the grpcweb protocol via `grpcweb` go proxy/wrapper.
- Add a new `ts-generator` protoc plugin for autogenerating the grpc client code.
- Add an initial configured grpcClient class in the frontend code.

### Benefits

Hopefully, we will able to consume grpc services via UI.

### Possible drawbacks

The chosen solution is good enough for our use case, but it has been extensively discussed at the parent issue (https://github.com/kubeapps/kubeapps/issues/3085)

### Applicable issues

  - related #3085

### Additional information

This is just the foundation of the upcoming PRs which will perform actual changes in the fronted code (just the catalog in short-term).

Here is an example (with some extra code not included in this PR):

image


chart/kubeapps/templates/frontend/configmap.yaml
```diff
@@ -142,7 +142,7 @@ data:
     {{- if .Values.featureFlags.kubeappsAPIsServer }}
       location ~* /apis {
         rewrite ^ $request_uri; # pass the encoded url downstream as is,
-        rewrite /apis([^?]*) $1?$args break;
+        rewrite /apis([^?]*) $1 break;
```
absoludity, line 145: Why this change?

Also, I was originally wondering why this is relevant at all, since it's the url for handling the gateway rendered API endpoints. I was expecting to see something for forwarding requests via content-type for the actual grpc/grpc-web requests, but I see down further that the client in the dashboard is configured to talk to `/apis` as a base URL. Nice, that's a bit neater in that we don't need nginx to do the proxying specifically.

cmd/kubeapps-apis/server/server.go
```diff
@@ -95,9 +96,23 @@ func Serve(serveOpts ServeOptions) {
...
+		grpcweb.WithOriginFunc(func(origin string) bool { return true }),
+		grpcweb.WithWebsockets(true),
+		grpcweb.WithWebsocketOriginFunc(func(req *http.Request) bool { return true }),
+	)
+
+	httpSrv.Handler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if webrpcProxy.IsGrpcWebRequest(r) || webrpcProxy.IsAcceptableGrpcCorsRequest(r) || webrpcProxy.IsGrpcWebSocketRequest(r) {
+			webrpcProxy.ServeHTTP(w, r)
+		}
+	})
```
absoludity, line 112: I'm confused by this line: it appears to *overwrite* the existing `httpSrv.Handler` (which is the gateway mux), with a handler that handles the grpc-web requests, but doesn't fall back to the existing gateway mux for other requests? Maybe it was forgotten, or maybe I'm missing something?

dashboard/src/shared/GrpcClient.ts
```diff
@@ -0,0 +1,22 @@
...
+import * as URL from "./url";
+
+export class GrpcClient {
+  private baseUrl = URL.api.kubeappsapis;
+  private grpcClient!: GrpcWebImpl;
+
+  private GetClientConfig = () => {
+    return {
+      // https://github.com/improbable-eng/grpc-web/blob/master/client/grpc-web/docs/transport.md#specifying-transports.
+      transport: grpc.CrossBrowserHttpTransport({ withCredentials: true }),
```
absoludity, line 12: Did you try without the `withCredentials: true`? From the docs:

```
  // send browser cookies along with cross-origin requests (CORS), defaults to `false`.
  withCredentials?: boolean
```

It's not clear to me from the docs there whether when its false, we'd still get credentials sent since the request isn't cross origin, or whether it means when false it won't send any credentials with any requests whether they satisfy CORS or not.

absoludity (approved):
Leaving a +1, but please check the mux comment for the new handler and ensure that we still have gateway URLs being handled. Thanks!

antgamdia (commented):
I've added the suggestions you made. Thanks a lot, especially for the http-handler one! Let me know if you have any further comment before merging :)

dashboard/src/shared/GrpcClient.ts
absoludity: Great, thanks.

cmd/kubeapps-apis/server/server.go
```diff
@@ -95,9 +96,23 @@ func Serve(serveOpts ServeOptions) {
...
+
 	go grpcSrv.Serve(grpcLis)
+	go grpcSrv.Serve(grpcwebLis)
```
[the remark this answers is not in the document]
absoludity: That's an improvement, in that we'll at least see a log message if something goes wrong and the binary will stop and be restarted (by k8s). Thanks.

absoludity:
LGTM, +1 and thanks for the changes!

PR #3161, opened 2021-07-19
Update so plugins are passed a rest.Config getter rather than a client getter

absoludity:
### Description of the change

Originally we passed plugins a client getter that return a dynamic
client. We then found that the helm plugin additionally needs a typed
client, so updated the client getter to return both. For work that I'm
currently doing instantiating a helm API action configuration, I also
need the plugin to have access to a genericclioptions.RESTClientGetter.

Rather than updating the existing client getter to return yet another
type, and given that it's a single line to create a client from the rest
config, I'm instead updating so that plugins are initialised with an
initialized rest.Config getter so that they can create any client they
need while still not needing to know any details of the config (eg.
pinniped setup or similar).

### Benefits

I can continue with the helm work, using the config to create a RESTClientGetter for the Helm API.

### Applicable issues

  - Ref #3146

### Additional information

Since neither kapp_controller nor fluxv2 plugins use the typed client, I updated those to only create the client that they use (the dynamic one).

cc @gfichtenholt JFYI - this shouldn't affect your branch, unless you've also updated how the kubernetes clients are created, but if there's a conflict, I'm happy to merge this after your PR and deal with the conflicts.

### DIFF
--- a/cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/cache.go
+++ b/cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/cache.go
@@ -21,7 +21,6 @@ import (
 	"sync"
 
 	"github.com/go-redis/redis/v8"
-	"github.com/kubeapps/kubeapps/cmd/kubeapps-apis/server"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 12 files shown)


gfichtenholt:
> cc @gfichtenholt JFYI - this shouldn't affect your branch, unless you've also updated how the kubernetes clients are created, but if there's a conflict, I'm happy to merge this after your PR and deal with the conflicts.

yeah, i am sure there will be something to resolve. but my thing is going to take some doing and I am in the thick of it, so it'll be a couple of days at least. I don't want to hold you up so go ahead. come what may :-)

antgamdia (approved):
Ah, great. Much better approach as we are increasing the variability of the returned`clientGetter`s. This way, we delegate in each plugin the responsibility of creating its own config.
Thank you for adding the required changes in every plugin.

PR #2387, opened 2021-02-11
Allow to change kubeops TZ

andresmgot:
### Description of the change

More info at #2288 

The timezone used to store apps info can be changed with an environment variable but it requires some extra files in the image.

cc/ @juan131 can you replicate this change in the Bitnami image?

### Applicable issues

  - fixes #2288

PR #3169, opened 2021-07-20
Add implementation for GetInstalledPackageSummaries.

absoludity:
### Description of the change

Adds the implementation for GetInstalledPackageSummaries for the Helm plugin.

I'll do a follow-up PR that adds the LatestPkgVersion field as well.

IRL test:
```
grpcurl -plaintext -d '{"context": {"namespace": "kubeapps"}}' localhost:8080 kubeappsapis.plugins.helm.packages.v1alpha1.HelmPackagesService.GetInstalledPackageSummaries
{
  "installedPackagesSummaries": [
    {
      "installedPackageRef": {
        "context": {
          "namespace": "kubeapps"
        },
        "identifier": "chartmuseum"
      },
      "name": "chartmuseum",
      "pkgVersionReference": {
        "version": "2.14.2"
      },
      "currentPkgVersion": "2.14.2",
      "iconUrl": "https://raw.githubusercontent.com/helm/chartmuseum/master/logo2.png",
      "pkgDisplayName": "chartmuseum",
      "shortDescription": "Host your own Helm Chart Repository"
    },
    {
      "installedPackageRef": {
        "context": {
          "namespace": "kubeapps"
        },
        "identifier": "kubeapps"
      },
      "name": "kubeapps",
      "pkgVersionReference": {
        "version": "7.1.3-dev1"
      },
      "currentPkgVersion": "7.1.3-dev1",
      "iconUrl": "https://raw.githubusercontent.com/kubeapps/kubeapps/master/docs/img/logo.png",
      "pkgDisplayName": "kubeapps",
      "shortDescription": "Kubeapps is a dashboard for your Kubernetes cluster that makes it easy to deploy and manage applications in your cluster using Helm"
    }
  ]
}
```

### Applicable issues

  - Ref #3146

### DIFF
--- a/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
+++ b/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
@@ -26,9 +26,13 @@ import (
 	corev1 "github.com/kubeapps/kubeapps/cmd/kubeapps-apis/gen/core/packages/v1alpha1"
 	"github.com/kubeapps/kubeapps/cmd/kubeapps-apis/gen/plugins/helm/packages/v1alpha1"
 	"github.com/kubeapps/kubeapps/cmd/kubeapps-apis/server"
+	"github.com/kubeapps/kubeapps/pkg/agent"
 	"github.com/kubeapps/kubeapps/pkg/chart/models"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+	"helm.sh/helm/v3/pkg/action"
+	"helm.sh/helm/v3/pkg/kube"
+	"helm.sh/helm/v3/pkg/release"
 	authorizationv1 "k8s.io/api/authorization/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/dynamic"
@@ -37,6 +41,7 @@ import (
 )
 
 type clientGetter func(context.Context) (kubernetes.Interface, dynamic.Interface, error)
+type helmActionConfigGetter func(ctx context.Context, namespace string) (*action.Configuration, error)
 
 // Compile-time statement to ensure this service implementation satisfies the core packaging API
 var _ corev1.PackagesServiceServer = (*Server)(nil)
@@ -56,6 +61,7 @@ type Server struct {
 	clientGetter             clientGetter
 	globalPackagingNamespace string
 	manager                  utils.AssetManager
+	actionConfigGetter       helmActionConfigGetter
 }
 
 // NewServer returns a Server automatically configured with a function to obtain
@@ -96,6 +102,29 @@ func NewServer(configGetter server.KubernetesConfigGetter) *Server {
 				return nil, nil, status.Errorf(codes.FailedPrecondition, fmt.Sprintf("unable to get typed client : %v", err))
 			}
 			return typedClient, dynamicClient, nil
+		},
+		actionConfigGetter: func(ctx context.Context, namespace string) (*action.Configuration, error) {
+			if configGetter == nil {
+				return nil, status.Errorf(codes.Internal, "configGetter arg required")
+			}
+			config, err := configGetter(ctx)
+			if err != nil {
+				return nil, status.Errorf(codes.FailedPrecondition, fmt.Sprintf("unable to get config : %v", err))
+			}
+
+			restClientGetter := agent.NewConfigFlagsFromCluster(namespace, config)
+			clientSet, err := kubernetes.NewForConfig(config)
+			if err != nil {
+				return nil, status.Errorf(codes.FailedPrecondition, fmt.Sprintf("unable to create kubernetes client : %v", err))
+			}
+			// TODO(mnelson): Update to allow different helm storage options.
+			storage := agent.StorageForSecrets(namespace, clientSet)
+			return &action.Configuration{
+				RESTClientGetter: restClientGetter,
+				KubeClient:       kube.New(restClientGetter),
+				Releases:         storage,
+				Log:              log.Infof,
+			}, nil
 		},
 		manager:                  manager,
 		globalPackagingNamespace: kubeappsNamespace,
@@ -170,7 +199,10 @@ func (s *Server) GetAvailablePackageSummaries(ctx context.Context, request *core
 	if err != nil {
 		return nil, status.Errorf(codes.InvalidArgument, "Unable to intepret page token %q: %v", request.GetPaginationOptions().GetPageToken(), err)
 	}
-	charts, _, err := s.manager.GetPaginatedChartListWithFilters(cq, pageOffset, int(pageSize))
+	// The current assetsvc manager works on a page number (ie. 1 for the first page),
+	// rather than an offset.
+	pageNumber := pageOffset + 1
+	charts, _, err := s.manager.GetPaginatedChartListWithFilters(cq, pageNumber, int(pageSize))
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "Unable to retrieve charts: %v", err)
 	}
@@ -205,7 +237,7 @@ func (s *Server) GetAvailablePackageSummaries(ctx context.Context, request *core
 // for a specific plugin when combining).
 func pageOffsetFromPageToken(pageToken string) (int, error) {
 	if pageToken == "" {
-		return 1, nil
+		return 0, nil
 	}
 	offset, err := strconv.ParseUint(pageToken, 10, 0)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -97,6 +103,29 @@ func NewServer(configGetter server.KubernetesConfigGetter) *Server {
...
+			if configGetter == nil {
+				return nil, status.Errorf(codes.Internal, "configGetter arg required")
+			}
+			config, err := configGetter(ctx)
+			if err != nil {
+				return nil, status.Errorf(codes.FailedPrecondition, fmt.Sprintf("unable to get config : %v", err))
+			}
+
+			restClientGetter := agent.NewConfigFlagsFromCluster(namespace, config)
+			clientSet, err := kubernetes.NewForConfig(config)
```
antgamdia, line 116: Just a thought: are we creating yet another config here? Shouldn't we use the `typedClient` which already is doing ` `kubernetes.NewForConfig(config)` I mean, atm I don't know if it returns the same instance or it creates an additional one.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -97,6 +103,29 @@ func NewServer(configGetter server.KubernetesConfigGetter) *Server {
...
+			if err != nil {
+				return nil, status.Errorf(codes.FailedPrecondition, fmt.Sprintf("unable to create kubernetes client : %v", err))
+			}
+			// TODO(mnelson): Update to allow different helm storage options.
```
antgamdia, line 120: Currently is just either `configmap`, `secret` or  `sql (PostgreSQL)`, no? If we ever continue using Postgresql as the shared cache layer, we could also offer out-of-the-box support for it. Interesting

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -481,3 +512,67 @@ func isValidChart(chart *models.Chart) (bool, error) {
...
+	response := &corev1.GetInstalledPackageSummariesResponse{
+		InstalledPackagesSummaries: installedPkgSummaries,
+	}
+	if len(releases) == cmd.Limit {
+		response.NextPageToken = fmt.Sprintf("%d", cmd.Limit+1)
+	}
+	return response, nil
+}
+
+func installedPkgSummaryFromRelease(r *release.Release) *corev1.InstalledPackageSummary {
```
antgamdia, line 555: Looking forward to starting using this backend to get rid of the helm2 to helm3 converters we currently have out there :P

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server_test.go
```diff
@@ -38,7 +46,10 @@ import (
 	"k8s.io/client-go/kubernetes"
 	typfake "k8s.io/client-go/kubernetes/fake"
 	k8stesting "k8s.io/client-go/testing"
-	"k8s.io/helm/pkg/proto/hapi/chart"
+
+	// TODO(mnelson): models.Chart.Maintainers is depending on the old v1 chart
+	// code. I don't expect there is any reason other than a historical one.
+	chartv1 "k8s.io/helm/pkg/proto/hapi/chart"
```
antgamdia, line 52: But, I guess it will be removed in Helm 3.6.2+ versions (though we are unable to upgrade due to the k8s 1.22 upgrade, no?)

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
antgamdia: Also, totally unrelated to this PR, but I've just recalled one issue with TZ we had in the past.
https://github.com/kubeapps/kubeapps/pull/2387/files
Perhaps we should add this envar to the kubeappsapis binary as well at some point.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server_test.go
```diff
@@ -1222,3 +1236,336 @@ func TestPackageAppVersionsSummary(t *testing.T) {
...
+			if got, want := response, tc.expectedResponse; !cmp.Equal(want, got, opts) {
+				t.Errorf("mismatch (-want +got):\n%s", cmp.Diff(want, got, opts))
+			}
+		})
+	}
+}
+
+// newActionConfigFixture returns an action.Configuration with fake clients
+// and memory storage.
+func newActionConfigFixture(t *testing.T, namespace string, rels []releaseStub) *action.Configuration {
```
antgamdia, line 1525: Great! Neat way to create the releases!

antgamdia (approved):
Awesome! FInally actual helm support here! Thanks a lot for the effort!
Further operations should be easier to implement now.

antgamdia:
> I'll do a follow-up PR that adds the LatestPkgVersion field as well.

Forgot to ask: do you mean in this operation or in every helm-direct op ?

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
absoludity: There is no additional client created here because `GetInstalledPackageDetails` doesn't even use `clientGetter` (since it doesn't require a k8s client outside of the helm client). But yes, if a handler did need both, the helm action config would have a separate client (in memory) from the typed client used directly by the handler. But I don't see an issue with this?

We could refactor things so that the `actionConfigGetter` is passed a typed client, but then it'd be additionally creating the dynamic client unnecessarily. I think the `clientGetter` is only returning a tuple of clients because of the history here (it was the interface originally between the plugins and the core), so we could update to just provide separate getters `typedClientGetter` and `dynamicClientGetter` that can be used by our functions.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
absoludity: > Currently is just either `configmap`, `secret` or `sql (PostgreSQL)`, no?

The current helm support defaults to secret but allows `configmap` as an option. I wasn't aware of the newer PostgresSQL support as a Helm storage driver, but we don't support this currently.

> If we ever continue using Postgresql as the shared cache layer, we could also offer out-of-the-box support for it. Interesting

Possibly, though it would have other implications too (eg. we use the postgresql more like a cache and don't worry about guarantees currently, or another, currently access to releases in secrets/configmaps is limited by RBAC, but access to releases in the DB would need to be manually checked per namespace like we do assets etc.)

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -170,7 +199,10 @@ func (s *Server) GetAvailablePackageSummaries(ctx context.Context, request *core
 	if err != nil {
 		return nil, status.Errorf(codes.InvalidArgument, "Unable to intepret page token %q: %v", request.GetPaginationOptions().GetPageToken(), err)
 	}
-	charts, _, err := s.manager.GetPaginatedChartListWithFilters(cq, pageOffset, int(pageSize))
+	// The current assetsvc manager works on a page number (ie. 1 for the first page),
+	// rather than an offset.
+	pageNumber := pageOffset + 1
+	charts, _, err := s.manager.GetPaginatedChartListWithFilters(cq, pageNumber, int(pageSize))
```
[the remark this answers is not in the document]
absoludity: This wasn't so much caching a bug, as refactoring to be clear about the current use of pageNumber (so I could update `pageOffsetFromPageToken` below to return 0 by default, as it should, rather than 1.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server_test.go
absoludity: We could probably update `models.Chart.Maintainers` now, regardless of the later helm version. Just didn't want to do that here.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
antgamdia: I agree, I don't see any issue with it, given that both use the same config so, even they are different objects in memory, they'll be the same thing in practice.
Mmm, yep, perhaps adding separate getters could be a good refactor option.
Anyway, it just was a thought, nothing important or critical.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
antgamdia: I don't know when Helm added support for Postgresql, but I have assumed it also was a reason to migrate to pg indeed.
Yep, Helm also mentions the permissions thing: https://helm.sh/docs/topics/permissions_sql_storage_backend/

Issue #3165, opened 2021-07-19
Changes required for using the Helm-direct plugin 


antgamdia:
As #3085 is increasingly growing, for the sake of readability, I'm splitting it into a separate issue, so that we can keep track of the required changes.

### Quick summary (sep 1st):

Triggering the release

### Quick summary (aug 31th):

Bugs detected after CI, working on that.

- [x] Multicluster mode failing, unknown issues, check later.
- [x] Upgrading a chart. Twofold scenario:  chart w/ repo; chart w/o repo.
- [ ] Getting duplicates charts after searching for chart names in input

### Quick summary (aug 30th):

Getting CI passing, not luck

### Quick summary (aug 27th):

- Kubeapps v2.3.4 released!
- Issues with the CI
- Minor PRs
- Fixing tests: first round completed! 

However, 20 tests from 5 test suites are still being skipped, (below). The reasons range from: 1) tradeoff complexity vs velocity; 2) something requiring extra checks because could be actual errors; 3) candidate to be removed, but unsure.

If you can continue from here, @absoludity , it'd be great. Otherwise, I'll continue working next week

1. - src/components/Catalog/Catalog.test.tsx (9)
2. - src/UpgradeForm.test.tsx (6)
3. - src/actions/apps.test.tsx (3)
4. - src/shared/App.test.ts (1)
5. - src/components/AppList/AppListItem.test.tsx (1)

Edit: I'm also running the e2e tests to see what is failing (**important**: we should also check the Bitnami tests, otherwise, a failing test there will block the release). I'll re-run them to ensure they're not false positives.

```console
 
FAIL  use-cases/07-upgrade.js (7.508 s)

 RUNS  ...
  ● Upgrades an application

    Option not found "select[name="chart-versions"]" ("7.3.2")


FAIL  use-cases/08-rollback.js (67.947 s)

 RUNS  ...
  ● Upgrades an application

    TimeoutError: Text not found "(current: 3)"
```

### Quick summary (aug 26th):

- Fixing issues when releasing v2.3.4
- Fixing tests

There are 2 tests suites (56 tests) which are failing (via `yarn react-scripts test -- --watchAll=false 2>&1 | grep 'FAIL' | sort --unique`) + 7 tests commented out so far:

1. FAIL src/components/Catalog/Catalog.test.tsx (30 tests)
2. FAIL src/components/ChartView/ChartView.test.tsx (10 tests)
[...]

PR #3170, opened 2021-07-20
[kubeappsapis] Add missing fields + initial core GetAvailablePackageDetail

antgamdia:
### Description of the change

This PR adds different changes in the new API:

- ~Minor documentation fixes (missing title or description, nothing critical)~ (moved to https://github.com/kubeapps/kubeapps/pull/3174)
- Add `LatestAppVersion` in the `AvailablePackageSummary`
- Add `Category` in the `AvailablePackageSummary` and `AvailablePackageDetail`
- Add an initial implementation of the core `GetAvailablePackageDetail` (current rationale: return the first match, which should be enough if we just deploy the direct-helm plugin, but I wonder what would happen in the flux+helm case)

### Benefits

This PR addresses the missing fields discovered in https://github.com/kubeapps/kubeapps/issues/3165#issuecomment-882943900 (at the time of writing).

### Possible drawbacks

Perhaps too many changes at once. I'm a bit unsure whether `GetAvailablePackageDetail` should come in this PR...but I finally decided to include it to centralize the discussions (besides, the current alternative is returning a "not implemented" error)

### Applicable issues

  - related #3165

### Additional information

N/A


cmd/kubeapps-apis/server/packages.go
```diff
@@ -68,3 +68,41 @@ func (s packagesServer) GetAvailablePackageSummaries(ctx context.Context, reques
...
+	if request.AvailablePackageRef != nil && request.AvailablePackageRef.Context != nil {
+		contextMsg = fmt.Sprintf("(cluster=[%s], namespace=[%s])", request.AvailablePackageRef.Context.Cluster, request.AvailablePackageRef.Context.Namespace)
+	}
+
+	log.Infof("+core GetAvailablePackageDetail %s", contextMsg)
+
+	pkg := &packages.AvailablePackageDetail{}
+
+	// TODO: We can do these in parallel in separate go routines.
+	for _, p := range s.plugins {
```
absoludity, line 84: For this call (singular - GetAvailablePackageDetail) it doesn't make sense to iterate the plugins, as it only makes sense to call the particular plugin identified in the request context? The request must include the plugin in the context and only that plugin should be used.

cmd/kubeapps-apis/server/packages.go
```diff
@@ -68,3 +68,41 @@ func (s packagesServer) GetAvailablePackageSummaries(ctx context.Context, reques
...
+
+		response, err := p.server.GetAvailablePackageDetail(ctx, request)
+		if err != nil {
+			return nil, err
+		}
+
+		if response.AvailablePackageDetail != nil {
+			pkg = response.AvailablePackageDetail
+
+			if response.AvailablePackageDetail.AvailablePackageRef == nil {
```
absoludity, line 94: This condition should be an error. A plugin must return the detail with the required data.

cmd/kubeapps-apis/server/packages.go
```diff
@@ -68,3 +68,41 @@ func (s packagesServer) GetAvailablePackageSummaries(ctx context.Context, reques
...
+				pkg.AvailablePackageRef = &packages.AvailablePackageReference{}
+			}
+			pkg.AvailablePackageRef.Plugin = p.plugin
```
absoludity, line 97: This should no longer be necessary. After discussion, we've updated to always filling in the plugin in the reference. If you check `GetAvailablePackageDetail` for the helm plugin, you'll see it's already including it.

cmd/kubeapps-apis/server/packages.go
```diff
@@ -68,3 +68,41 @@ func (s packagesServer) GetAvailablePackageSummaries(ctx context.Context, reques
...
+
+			// TODO: handle multiple matches for a package
```
absoludity, line 99: This is not possible as above. A valid AvailablePackageRef here must include the plugin.

absoludity (commented):
A couple of questions to help me understand things:

> 
> This PR adds different changes in the new API:
> 
>     * Minor documentation fixes (missing title or description, nothing critical)

Great - I need to check the openapi schema output before committing new docstrings.
 
>     * Add `LatestAppVersion` in the `AvailablePackageSummary`

+1 - can you also add that on the doc with a comment linking to this PR, just saying that while implementing the direct helm plugin we found the `LatestAppVersion` field was missing from the summary.

> 
>     * Add `Category` in the `AvailablePackageSummary` and `AvailablePackageDetail`

So I'm not sure that we should just add `Category` as a field (as per the [discussion on the doc](https://docs.google.com/document/d/1z03msZRGsRvcRaom-yrvEwIWcEfNy6fSA5Zg28gjYvA/edit?pli=1&disco=AAAAM6dYEOc)). Category is a custom piece of information with which Bitnami is annotating its charts, but is not standard in Helm itself, let alone in other package formats. But yes, our UI uses it when available. Also, although Bitnami has chosen to assign only a single category per package, other providers/packaging systems may allow multiple similar categories/tags per package. In fact, I just checked the current [carvel PackageMetadata](https://carvel.dev/kapp-controller/docs/latest/packaging/#packagemetadata) and it does exactly that with its `spec.categories` field.

I had originally thought it may be better to investigate a more generic `Annotations` field, but since Carvel validates that `categories` is something different packaging formats are thinking about, I'm +1 to include it as long as it's plural/repeated (no problem for the Bitnami charts to only have a single category). If that's what you do, can you please update the message on the doc similarly (painful, I know, but just so those that don't watch PRs get the updated info there).

> 
>     * Add an initial implementation of the core `GetAvailablePackageDetail` (current rationale: return the first match, which should be enough if we just deploy the direct-helm plugin, but I wonder what would happen in the flux+helm case)

Right, I was at first confused by your words here, but realised in the code that you're assuming you'd have to handle the core `GetAvailablePackageDetail` with an `AvailablePackageRef` that didn't include the plugin, I think? This is not the case. That is, a request to the core `GetAvailablePackageDetail()` *must* have a valid reference (including the plugin). The plural handler (`GetAvailablePackageSummaries`) returns them with the plugin specified. Clients don't need to parse or understand the reference, they just include it in a request for detail with the knowledge that it identifies a specific package.

Question from me: I was suprised because I didn't think you'd be using the core API in the UI yet? (As in, I thought you were using the helm plugin client right now).

absoludity:
Note, not for this PR, but we'll also need to update the `GetAvailablePackageSummariesResponse` message to include a `Categories` field that has all available categories (as per Dimitri's comment in the discussion on the design doc), so the UI can display the selection of available categories for the complete (not paginated) result set to choose from.

antgamdia:
> +1 - can you also add that on the doc with a comment linking to this PR, just saying that while implementing the direct helm plugin we found the `LatestAppVersion` field was missing from the summary.

Ok! I didn't update the design doc as I didn't want to change anything until having confirmed the changes in the PR (well, this explanation sounds exactly like an antipattern haha).

> (...) I'm +1 to include it as long as it's plural/repeated (no problem for the Bitnami charts to only have a single category). If that's what you do, can you please update the message on the doc similarly (painful, I know, but just so those that don't watch PRs get the updated info there).

Ok, sure, I will do!  Indeed, I thought about it (but didn't check the doc, my bad!), but I finally implemented the "bitnami" way (a single cat) trying not to diverge so much with the current implementation, But, yeah, sure, I do agree with the list of categories.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -237,6 +238,9 @@ func AvailablePackageSummaryFromChart(chart *models.Chart) (*corev1.AvailablePac
 	if chart.ChartVersions != nil || len(chart.ChartVersions) != 0 {
 		pkg.LatestPkgVersion = chart.ChartVersions[0].Version
 	}
+	if chart.ChartVersions != nil || len(chart.ChartVersions) != 0 {
+		pkg.LatestAppVersion = chart.ChartVersions[0].AppVersion
+	}
```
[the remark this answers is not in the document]
antgamdia: Of course... I copy-n-pasted it too quickly :grimacing:

antgamdia:
> Right, I was at first confused by your words here, but realised in the code that you're assuming you'd have to handle the core GetAvailablePackageDetail with an AvailablePackageRef that didn't include the plugin, I think? 

Yep, I assumed it was similar to the existing `GetAvailablePackageSummaries`, so I assumed I would have to iterate over each plugin and collect the results back somehow.

> This is not the case. That is, a request to the core GetAvailablePackageDetail() must have a valid reference (including the plugin).

Great, got it! 

> Question from me: I was suprised because I didn't think you'd be using the core API in the UI yet? (As in, I thought you were using the helm plugin client right now).

Well, I initially thought (and implemented) that we should invoke the Helm plugin in the catalog. However, later, I wondered: why  coupling the UI (the catalog) to the helm endpoint if we are using the very same model in the core API?
Then, I realized we didn't have this operation implemented and, since this PR was adding future required changes, I just said: why don't adding an initial implementation and see if I can use this core endpoint when addressing the "get chart details" view?

My idea is: shouldn't be equivalent 1) using the helm endpoints instead of the core ones when deploying more than 1 plugin; 2) configuring kubeapps to be deployed just with the helm plugin and using the core endpoints?

antgamdia:
> Note, not for this PR, but we'll also need to update the `GetAvailablePackageSummariesResponse` message to include a `Categories` field that has all available categories (as per Dimitri's comment in the discussion on the design doc), so the UI can display the selection of available categories for the complete (not paginated) result set to choose from.

I'm a bit lost here (I don't really remember what we all agreed here).

As per the docs: "(...) Perhaps we could request the annotations to be returned in the `GetAvailablePackageSummariesResponse` as an aggregation (ie. next to the repeated summaries, a separate field). The 'Category' annotation as an example"

So, should we change this field from `[]string` to `map<string>string`, so `categories` becomes `annotations` ? 

If so: the changes would be twofold: 1) edit this PR with the proto (and related) changes; 2) a new PR implementing the logic in the design doc for the helm plugin. Wouldn't they?

absoludity:
> > Note, not for this PR, but we'll also need to update the `GetAvailablePackageSummariesResponse` message to include a `Categories` field that has all available categories (as per Dimitri's comment in the discussion on the design doc), so the UI can display the selection of available categories for the complete (not paginated) result set to choose from.
> 
> I'm a bit lost here (I don't really remember what we all agreed here).

Right, two things are being confused, I think.

> 
> As per the docs: "(...) Perhaps we could request the annotations to be returned in the `GetAvailablePackageSummariesResponse` as an aggregation (ie. next to the repeated summaries, a separate field). The 'Category' annotation as an example"
> 
> So, should we change this field from `[]string` to `map<string>string`, so `categories` becomes `annotations` ?

Now not for now. That was what I had originally thought, but as mentioned above, given that carvel also has `categories`, i think your choice of specifically adding `categories` is a good one for now. We may want to add something more generic later, but let's not do so prematurely.

The comment about an *extra* Categories field at a higher level, the response message itself, is so the UX can display *all* categories that are relevant for the given query (ie. not just the results being paginated). But let's chat about this in our standup, it's something for this PR nor requiring any changes here, I don't think.

absoludity (approved):
Great, thanks for the updates.

absoludity:
Landing so I can build on this.

antgamdia:
>  Categories field at a higher level, the response message itself, is so the UX can display all categories that are relevant for the given query (ie. not just the results being paginated). 

Mmm, I think this behavior was discarded when doing the pagination stuff (perhaps https://via.vmw.com/2020-kubeapps-ui-proposal), but I can't find the link to the comment atm.

It was argued that displaying the relevant categories would result in a different UX as we currently provide. Right now, the filers are built kinda statically, I mean, an update in the query (eg, get charts like "apache")  won't result in a different category list.
One of the reasons (off the top of my head) is that if a user starts filtering, they eventually will see some left filters are disappearing and, the only way to remove them is using the list at the top.

In order to keep the UX as currently is, I'd suggest going with the Dimitri's proposal: `size=0` to query everything but also implement the return of the relevant categories, of course (though the UI only will use the size=0).

Thanks for the clarification here! Regarding the categories/annotations, good, I also agree; let's generalize in the future if really necessary.

absoludity:
> Mmm, I think this behavior was discarded when doing the pagination stuff (perhaps https://via.vmw.com/2020-kubeapps-ui-proposal), but I can't find the link to the comment atm.

Yes, we clarified that on slack. My mistake (I'd thought the current UI showed categories for the complete result set, but it's just all categories).

PR #3172, opened 2021-07-21
Populate AvailablePackageSummary.Name

absoludity:
Although we'd added AvailablePackageSummary.Name to the messages (after
switching to use an Identifier in the reference), we hadn't yet populated
it.

Related to discussion at https://github.com/kubeapps/kubeapps/issues/3165#issuecomment-883812172


antgamdia (approved):
Great, thanks!

PR #3179, opened 2021-07-22
Include the latest package version in the installed pkg summaries.

absoludity:
### Description of the change

```
grpcurl -plaintext localhost:8080 kubeappsapis.plugins.helm.packages.v1alpha1.HelmPackagesService.GetInstalledPackageSummaries
...
    {
      "installedPackageRef": {
        "context": {
          "namespace": "default"
        },
        "identifier": "test-apache"
      },
      "name": "test-apache",
      "pkgVersionReference": {
        "version": "8.5.8"
      },
      "currentPkgVersion": "8.5.8",
      "iconUrl": "https://bitnami.com/assets/stacks/apache/img/apache-stack-220x234.png",
      "pkgDisplayName": "apache",
      "shortDescription": "Chart for Apache HTTP Server",
      "latestPkgVersion": "8.5.9"
    }
```

### Benefits

Includes the latest package version, so that the UI can present when upgrades are available.

### Applicable issues

  - Ref #3146

### DIFF
--- a/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
+++ b/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
@@ -543,6 +543,30 @@ func (s *Server) GetInstalledPackageSummaries(ctx context.Context, request *core
 		installedPkgSummaries[i] = installedPkgSummaryFromRelease(r)
 	}
 
+	// Fill in the latest package version for each.
+	// TODO(mnelson): Update to do this with a single query rather than iterating and
+	// querying per release.
+	for i, rel := range releases {
+		// Helm does not store a back-reference to the chart used to create a
+		// release (https://github.com/helm/helm/issues/6464), so for each
+		// release, we look up a chart with than name and version available in
+		// the release namespace, and if one is found, pull out the latest chart
+		// version.
+		cq := utils.ChartQuery{
+			Namespace:  rel.Namespace,
+			ChartName:  rel.Chart.Metadata.Name,
+			Version:    rel.Chart.Metadata.Version,
+			AppVersion: rel.Chart.Metadata.AppVersion,
+		}
+		charts, _, err := s.manager.GetPaginatedChartListWithFilters(cq, 1, 0)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "Error while fetching related charts: %v", err)
+		}
+		if len(charts) == 1 && len(charts[0].ChartVersions) > 0 {
+			installedPkgSummaries[i].LatestPkgVersion = charts[0].ChartVersions[0].Version
+		}
+	}
+
 	response := &corev1.GetInstalledPackageSummariesResponse{
 		InstalledPackagesSummaries: installedPkgSummaries,
 	}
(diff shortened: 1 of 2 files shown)


cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -543,6 +543,30 @@ func (s *Server) GetInstalledPackageSummaries(ctx context.Context, request *core
 		installedPkgSummaries[i] = installedPkgSummaryFromRelease(r)
 	}
 
+	// Fill in the latest package version for each.
+	// TODO(mnelson): Update to do this with a single query rather than iterating and
+	// querying per release.
```
antgamdia, line 548: You mean adding a new method in the `PostgresAssetManager` and delegate the complexity of an actual SQL query, don't you?
If so, yep, I agree. We better perform this kind of operations on the DB side.

BTW, we should refactor it at some point (using an ORM or sth)

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -543,6 +543,30 @@ func (s *Server) GetInstalledPackageSummaries(ctx context.Context, request *core
...
+			ChartName:  rel.Chart.Metadata.Name,
+			Version:    rel.Chart.Metadata.Version,
+			AppVersion: rel.Chart.Metadata.AppVersion,
+		}
+		charts, _, err := s.manager.GetPaginatedChartListWithFilters(cq, 1, 0)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "Error while fetching related charts: %v", err)
+		}
+		if len(charts) == 1 && len(charts[0].ChartVersions) > 0 {
+			installedPkgSummaries[i].LatestPkgVersion = charts[0].ChartVersions[0].Version
```
antgamdia, line 566: Otherwise the field won't get populated... but is it an error we should handle? I mean, should the UI assume that always this pkg version will be available?
Well, I'm not sure right now (I haven't checked yet the "installed*" part in the UI.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server_test.go
```diff
@@ -1484,23 +1491,59 @@ func TestGetInstalledPackageSummaries(t *testing.T) {
...
-			server, _, cleanup := makeServer(t, authorized, actionConfig)
-			// It is the namespace of the the driver which determines the results. In the prod code,
-			// the actionConfigGetter sets this using StorageForSecrets(namespace, clientset).
-			// actionConfig.Releases.Driver.(*driver.Memory).SetNamespace(tc.request.GetContext().GetNamespace())
+			server, mock, cleanup := makeServer(t, authorized, actionConfig)
 			defer cleanup()
 
+			if tc.expectedStatusCode == codes.OK {
+				populateAssetDB(t, mock, tc.expectedResponse.InstalledPackagesSummaries)
+			}
```
antgamdia, line 1545: Thanks for updating the test!

antgamdia (approved):
Awesome! Thanks for the changes!!

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
absoludity: > You mean adding a new method in the `PostgresAssetManager` and delegate the complexity of an actual SQL query, don't you?
> If so, yep, I agree. We better perform this kind of operations on the DB side.
> 
> BTW, we should refactor it at some point (using an ORM or sth)

If we end up using postgresql long term, then yes, it'd be good to abstract back a bit to an ORM, but that depends on the outcome of the plugin caching etc. So I see no use to invest in that until we know.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
absoludity: No, it cannot always be available. If you add an app repository, install a package, then remove an app repository, kubeapps will correctly display the installed release, but has no way of determining this value (currently also).

absoludity (commented):
Thanks for the review!

PR #3185, opened 2021-07-23
Update to enable individual plugins for apis server

absoludity:
### Description of the change

I noticed that currently we were always loading all plugins with no option to enable individual ones.

Given that we're planning to release initially with only the helm plugin enabled, this change updates to allow enabling specific plugins, defaulting to the helm one.

I've also updated the dependencies in the Chart.yaml, so that we don't accidentally update the upstream chart with the redis dependency.

Finally, I've set the replicaCount for redis in dev to zero (similar to what we do for postgres). Let me know if this is an issue Greg.

### Benefits

Upstream changes to our chart (whether by us or the content team) should include only the helm plugin for now.

### Possible drawbacks

None

### Additional information

I'll annotate in line for anything I missed here.

### DIFF
--- a/chart/kubeapps/Chart.yaml
+++ b/chart/kubeapps/Chart.yaml
@@ -11,9 +11,14 @@ dependencies:
   - name: postgresql
     repository: https://charts.bitnami.com/bitnami
     version: 10.x.x
+  # Currently redis is only used for an in-progress plugin for flux support.
+  # Our upstream bitnami/kubeapps chart should not include redis as a
+  # dependency yet, and in development we can set redis.enabled if developing
+  # other plugins only.
   - name: redis
     repository: https://charts.bitnami.com/bitnami
     version: 14.x.x
+    condition: redis.enabled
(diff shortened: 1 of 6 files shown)


gfichtenholt (approved):
looks good to me. Thanks

chart/kubeapps/Chart.yaml
```diff
@@ -11,9 +11,14 @@ dependencies:
...
     repository: https://charts.bitnami.com/bitnami
     version: 10.x.x
+  # Currently redis is only used for an in-progress plugin for flux support.
+  # Our upstream bitnami/kubeapps chart should not include redis as a
+  # dependency yet, and in development we can set redis.enabled if developing
+  # other plugins only.
   - name: redis
     repository: https://charts.bitnami.com/bitnami
     version: 14.x.x
+    condition: redis.enabled
```
antgamdia, line 21: Wow, I wasn't aware of this helm feature. Good to know!

antgamdia (approved):
Great! It's something I wanted to do too. In my local setup I was just commenting the build lines :P, but it's much better.

PR #3186, opened 2021-07-23
Further fluxv2 plugin features

gfichtenholt:
Here I added all the missing fields to GetAvailablePackageDetail response, except LongDescription. Not sure what to set that one to (helm plug-in doesn't do it either). 
Also the server.go and server_test.go were getting too big so I separated out functionality and tests into separate repo and chart-specific .go files, hopefully it makes it a bit easier to follow.


absoludity (approved):
Looks fine to me. I expect we may have some extra fields added here that we hadn't thought of as we start using the data in the UX (for helm), let's see. Thanks!

gfichtenholt:
Thanks Michael

PR #3187, opened 2021-07-23
Remove unnecessary and incorrect check on maintainers.

absoludity:
### Description of the change

Dimitri pointed out that the check on these seemed odd (why check ChartVersions at all?).

There's actually no need for checking whether chart.Maintainers is nil or not since nil is the zero value for arrays, a range of it results in nothing to iterate.

The existing tests in TestIsValidChart demo this already (examples with nil maintainers).

I'm not yet sure about sharing code between plugins, so for now haven't dealt with that (but more than happy if you want to have a go Dimitri).

### Benefits

Removes confusing code :)

### Possible drawbacks

None.

### DIFF
--- a/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
+++ b/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
@@ -526,11 +526,9 @@ func isValidChart(chart *models.Chart) (bool, error) {
 			}
 		}
 	}
-	if chart.Maintainers != nil || len(chart.ChartVersions) != 0 {
-		for _, maintainer := range chart.Maintainers {
-			if maintainer.Name == "" {
-				return false, status.Errorf(codes.Internal, "required field .Maintainers[i].Name not found on helm chart: %v", chart)
-			}
+	for _, maintainer := range chart.Maintainers {
+		if maintainer.Name == "" {
+			return false, status.Errorf(codes.Internal, "required field .Maintainers[i].Name not found on helm chart: %v", chart)
 		}
 	}
 	return true, nil


dlaloue-vmware (approved):
looks good

gfichtenholt (commented):
no need to fix flux plug in. I re-organized the code and will fix it myself. Thanks

antgamdia:
> There's actually no need for checking whether chart.Maintainers is nil or not since nil is the zero value for arrays, a range of it results in nothing to iterate

Ooops, I read the (offline) conversation and I agree, it's odd how go manages `nil` references. I wrote that check (my bad!) I still have to assimilate and get used to :p  
Thanks for the PR +1

gfichtenholt:
Well, Michael had explained that you have to change your way of looking at this. Go is NOT an object-oriented language. Technically, we are not calling methods on objects. It's this subtle difference between objects and function receivers, which isn't covered that well in the go intro or FAQs, I looked. This is where experience plays a big part. Not that I agree with these decisions, they definitely look odd for a guy coming from C/C++/Java background. Oh well. It is what it is, as they say here.

gfichtenholt (changes requested):
yeah, fluxv2 files are way of out date. Please exclude them from this PR

absoludity:
> yeah, fluxv2 files are way of out date. Please exclude them from this PR

I did already? You'll see above your message that I'd updated a second time after I realised the same thing :)

absoludity:
> > yeah, fluxv2 files are way of out date. Please exclude them from this PR
> 
> I did already? You'll see above your message that I'd updated a second time after I realised the same thing :)

Ping @gfichtenholt  ^^ I'd already removed all changes to the flux2 plugin before you finished your comment as you can see above (though you would not have known that at the time you commented, but it would be great to unblock the PR).

gfichtenholt:
> > > yeah, fluxv2 files are way of out date. Please exclude them from this PR
> > 
> > 
> > I did already? You'll see above your message that I'd updated a second time after I realised the same thing :)
> 
> Ping @gfichtenholt ^^ I'd already removed all changes to the flux2 plugin before you finished your comment as you can see above (though you would not have known that at the time you commented, but it would be great to unblock the PR).

yeah, ok, good. Sorry didn't realize you were waiting for me to respond

gfichtenholt (approved):

PR #2573, opened 2021-03-18
Css login fix

antgamdia:
### Description of the change

After the css changes, the login page has a couple of issues (described in #2572 ). This PR aims at solving both.

### Benefits

Login page will look better.

### Possible drawbacks

N/A

### Applicable issues

  - fixes #2572

### Additional information

 N/A

PR #3191, opened 2021-07-23
feat(dashboard): Add default theme value

### LINKED ISSUE #3176: Define default dashboard theme from helm
## Ask
Instead of letting `window.matchMedia("(prefers-color-scheme: dark)")` determine the default theme on first load I was wondering if it would be possible to set a preferred default in the `values.yaml` file and pass it to the frontend in `config.json`. Behavior would be if `Values.dashboard.theme` is set, default to that, otherwise utilize `window.matchMedia`

## Implementation Question
I am willing to take ownership of this feature but looking through the codebase I haven't identified an ideal place to make this work. It appears that the theme is set before the request for the config is fired. Any help would be greatly appreciated

aanthonyrizzo:
Signed-off-by: Anthony Rizzo <36711320+aanthonyrizzo@users.noreply.github.com>

### Description of the change
This PR adds the ability to overwrite the initial default theme for Kubeapps via a config value.

### Benefits
Allows for more control for theming and UX

### Possible drawbacks
If this value is set. Anytime a user reloads or refreshes, the theme is changed back to the default value, even if the user sets a different one. We can potentially add another value to local storage called `user_theme`, which is set by the theme toggle switch. And add logic there so if that value is present, ignore the config.theme value.

### Applicable issues
Fixes #3176

### DIFF
--- a/chart/kubeapps/README.md
+++ b/chart/kubeapps/README.md
@@ -178,6 +178,7 @@ Once you have installed Kubeapps follow the [Getting Started Guide](https://gith
 | `dashboard.customStyle`                           | Custom CSS injected to the Dashboard to customize Kubeapps look and feel                  | `""`                 |
 | `dashboard.customComponents`                      | Custom Form components injected into the BasicDeploymentForm                              | `""`                 |
 | `dashboard.customLocale`                          | Custom translations injected to the Dashboard to customize the strings used in Kubeapps   | `""`                 |
+| `dashboard.defaultTheme`                          | Default theme used in the Dashboard if the user has not selected any theme yet.           | `""`                 |
 | `dashboard.replicaCount`                          | Number of Dashboard replicas to deploy                                                    | `2`                  |
 | `dashboard.extraEnvVars`                          | Array with extra environment variables to add to the Dashboard container                  | `[]`                 |
 | `dashboard.extraEnvVarsCM`                        | Name of existing ConfigMap containing extra env vars for the Dashboard container          | `nil`                |
--- a/chart/kubeapps/templates/dashboard/configmap.yaml
+++ b/chart/kubeapps/templates/dashboard/configmap.yaml
@@ -76,5 +76,5 @@ data:
       "authProxySkipLoginPage": {{ .Values.authProxy.skipKubeappsLoginPage }},
       "featureFlags": {{ .Values.featureFlags | toJson }},
       "clusters": {{ template "kubeapps.clusterNames" . }},
-      "defaultTheme": "{{ .Values.dashboard.defaultTheme }}"
+      "theme": "{{ .Values.dashboard.defaultTheme }}"
     }
--- a/chart/kubeapps/templates/dashboard/configmap.yaml
+++ b/chart/kubeapps/templates/dashboard/configmap.yaml
@@ -76,5 +76,5 @@ data:
       "authProxySkipLoginPage": {{ .Values.authProxy.skipKubeappsLoginPage }},
       "featureFlags": {{ .Values.featureFlags | toJson }},
       "clusters": {{ template "kubeapps.clusterNames" . }},
-      "theme": "{{ .Values.dashboard.theme }}"
+      "defaultTheme": "{{ .Values.dashboard.defaultTheme }}"
     }
--- a/chart/kubeapps/templates/dashboard/configmap.yaml
+++ b/chart/kubeapps/templates/dashboard/configmap.yaml
@@ -75,5 +75,6 @@ data:
       "oauthLogoutURI": {{ .Values.authProxy.oauthLogoutURI | quote }},
       "authProxySkipLoginPage": {{ .Values.authProxy.skipKubeappsLoginPage }},
       "featureFlags": {{ .Values.featureFlags | toJson }},
-      "clusters": {{ template "kubeapps.clusterNames" . }}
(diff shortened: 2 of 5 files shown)


antgamdia:
Thanks for the PR! I haven't had a chance yet to reply to the issue and have a look into it. 
It seems OK;  the logic you suggest is adding the config value as the default one, but enabling users to change it using the toggle, no? However, the current implementation is just overriding the user's preferences on each reload.

What would be the desired behavior? Should this parameter prevent users to change the theme (e.g., enforce a custom corporate theme) or it is just a default value in case the user hasn't selected anything yet?

aanthonyrizzo:
The original scenario that I was hoping to address w/ this feature was to ensure that everytime a user first visits our instance of Kubeapps they are greeted w/ the dark theme. However when I was testing the feature I noticed that when a default value was set, and a user wanted to change it via the UI, on next reload it would go back to the default. 

> What would be the desired behavior? Should this parameter prevent users to change the theme (e.g., enforce a custom corporate theme) or it is just a default value in case the user hasn't selected anything yet?

In theory all I wanted was on first load, the theme would be dark. However, because of the drawback mention I think we should either:
1. Add another local storage variable to know when the theme has been overridden by the user
2. As u mentioned, disable the toggle in the UI when this is set

Im biased to number 2 (since I was going to do this with css). SO if you think that is a reasonable compromise I can patch this PR and include that functionality in the documentation.

absoludity:
> 
> In theory all I wanted was on first load, the theme would be dark.

Right, that's what I understood on the discussion on the issue too, when I suggested:

> when a user loads Kubeapps in their browser, if they have not previously explicitly set a theme themselves, it may switch from the browser detected theme used when the HeaderManager loads, to the one set from the config (if you update so that the theme gets set when the config loads).

I do think we need to ensure it doesn't override a user setting, somehow.

> However, because of the drawback mention I think we should either:
> 
>     1. Add another local storage variable to know when the theme has been overridden by the user

The existing one is exactly that already, I think? (it allows the user to specify the theme and stores that value).
> 
>     2. As u mentioned, disable the toggle in the UI when this is set

This could be useful, but separate I think (ie. something like allowUserThemeChange). First we need to ensure that a default theme can be configured, while users can choose one too.

> 
> Im biased to number 2 (since I was going to do this with css). SO if you think that is a reasonable compromise I can patch this PR and include that functionality in the documentation.

That would mean that if a theme is set in the config that it's not a default theme but an enforced theme right? It sounds different from what you originally proposed. Let's see what Antonio thinks. From my pov, I would have thought we want to:

1. If the user has not set a theme and no default theme is known, use the window.match as done currently,
2. If when the config is loaded, a default theme has been configured and the user has not set a custom theme, switch to the default theme and store that in local storage (so there's no flicker next time between the HeaderManager loading and the config being loaded
3. If the user explicitly sets a theme, we store that choice in local config and always use that (overriding any defaults from the browser or config).

Just my 2c.

aanthonyrizzo:
The issue here is figuring out how local storage was set. Since `localStorage.getItem("theme")` is always set (either by `window.matchMedia` or persisting on load) when the `config.json` is loaded I have no good way of determining if this was set by the user or by `matchMedia`. Hence my proposal to  

> 1. Add another local storage variable to know when the theme has been overridden by the user

but there can definitely be another way that I'm not seeing since the codebase is quite large

absoludity:
Hmm, I had thought that the local storage is only ever set if and only if the user explicitly makes a decision, but I just checked and see you're correct. Looks like that changed with https://github.com/kubeapps/kubeapps/pull/2573 . Let me check something about that change with @antgamdia , as it's not clear to me why we need to set the local storage there.

antgamdia:
> but there can definitely be another way that I'm not seeing since the codebase is quite large

Hey ya! Sorry for jumping in so late (I've been very busy with other issues). First of all, thanks again for the contributions and the insights you provided.
You are right, the current theming implementation is quite a mess in terms of how we are setting the theme. I've tried to have a look... but I ended up changing a bunch of things using this PR as a base.

I've sent this (yet) another PR aimed at solving it: https://github.com/kubeapps/kubeapps/pull/3209, please have a look and tell us if it works for you :)

antgamdia:
I've jumped in this PR for adding a couple of minor changes. This way, this PR will add the changes in the chart whereas mine adds the logic in the Kubeapps dashboard :)
Thanks again!

chart/kubeapps/README.md
```diff
@@ -178,6 +178,7 @@ Once you have installed Kubeapps follow the [Getting Started Guide](https://gith
 | `dashboard.customStyle`                           | Custom CSS injected to the Dashboard to customize Kubeapps look and feel                  | `""`                 |
 | `dashboard.customComponents`                      | Custom Form components injected into the BasicDeploymentForm                              | `""`                 |
 | `dashboard.customLocale`                          | Custom translations injected to the Dashboard to customize the strings used in Kubeapps   | `""`                 |
+| `dashboard.defaultTheme`                          | Default theme used in the Dashboard if the user has not selected any theme yet.           | `""`                 |
```
antgamdia, line 181: Just for letting you know, we are using this tool for generating the readme automatically: https://github.com/bitnami-labs/readme-generator-for-helm

antgamdia (approved):

chart/kubeapps/README.md
aanthonyrizzo: Ah I missed that, thanks!

PR #3195, opened 2021-07-26
Add chart files to response for AvailablePackageDetail

### LINKED ISSUE #3146: Implement further direct helm plugin features

absoludity:
### Description of the change

As Greg pointed out, the existing code was using what appear to be no-longer-used fields on the model.

This PR updates to request the chart files from the DB and use those instead.

I'll follow-up with a PR to remove the code which is no longer being used.

IRL:
```
grpcurl -plaintext -d '{"available_package_ref": {"context": {"namespace": "kubeapps"},"identifier": "bitnami/wordpress"}}' localhost:8080 kubeappsapis.plugins.helm.packages.v1alpha1.HelmPackagesService.GetAvailablePackageDetail
{
  "availablePackageDetail": {
    "availablePackageRef": {
      "context": {
        "namespace": "kubeapps"
      },
      "identifier": "bitnami/wordpress",
      "plugin": {
        "name": "helm.packages",
        "version": "v1alpha1"
      }
    },
    "name": "wordpress",
    "pkgVersion": "11.1.5",
    "appVersion": "5.7.2",
    "iconUrl": "https://bitnami.com/assets/stacks/wordpress/img/wordpress-stack-220x234.png",
    "displayName": "wordpress",
    "shortDescription": "Web publishing platform for building blogs and websites.",
    "readme": "# WordPress\n\n[WordPress](https://wordpress.org/) is one of the most versatile open source content management systems on the market. A publishing platform for building blogs and websites.\n\n## TL;DR\n\n```console\n$ helm repo add bitnami https://charts.bitnami.com/bitnami\n$ helm install my-release bitnami/wordpress\n```\n\n## Introduction\n\nThis chart bootstraps a [WordPress](https://github.com/bitnami/bitnami-docker-wordpress) deployment on a [Kubernetes](http://kubernetes.io) 
...
```

### Benefits

We actually see Readme and other data required for the details.

### Applicable issues

  - fixes #3146

Other info:

- Fixed the tests to ensure they're sending url-encoded repos
- Fixed the chart to not expect a redis password if redis isn't being used.

### DIFF
--- a/chart/kubeapps/templates/kubeappsapis/deployment.yaml
+++ b/chart/kubeapps/templates/kubeappsapis/deployment.yaml
@@ -87,6 +87,7 @@ spec:
           env:
             - name: PORT
               value: {{ .Values.kubeappsapis.containerPort | quote }}
+            {{- if .Values.redis.enabled }}
             # REDIS-* vars are required by the plugins for caching functionality
             # TODO (gfichtenolt) this as required by the kubeapps apis service (which will
             # longer-term pass something to the plugins so that the plugins won't need to
@@ -100,6 +101,7 @@ spec:
                   name: {{ include "kubeapps.redis.secretName" . }}
             - name: REDIS_DB
               value: "0"
+            {{- end }}
             # TODO(agamez): pass this configuration using a separated config file
             # These env vars are currently (and temporarily) required by the 'helm' plugin
             - name: POD_NAMESPACE
--- a/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
+++ b/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
@@ -328,7 +328,19 @@ func (s *Server) GetAvailablePackageDetail(ctx context.Context, request *corev1.
 		return nil, status.Errorf(codes.Internal, "Unable to retrieve chart: %v", err)
 	}
 
-	availablePackageDetail, err := AvailablePackageDetailFromChart(&chart)
+	if len(chart.ChartVersions) == 0 {
+		return nil, status.Errorf(codes.Internal, "Chart returned without any versions: %+v", chart)
+	}
+	if version == "" {
+		version = chart.ChartVersions[0].Version
+	}
+	fileID := fileIDForChart(unescapedChartID, chart.ChartVersions[0].Version)
+	chartFiles, err := s.manager.GetChartFiles(namespace, fileID)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "Unable to retrieve chart files: %v", err)
+	}
+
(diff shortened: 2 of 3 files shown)


chart/kubeapps/templates/kubeappsapis/deployment.yaml
```diff
@@ -87,6 +87,7 @@ spec:
           env:
             - name: PORT
               value: {{ .Values.kubeappsapis.containerPort | quote }}
+            {{- if .Values.redis.enabled }}
```
antgamdia, line 90: Haha, we ran into the same issue. I've added these changes at https://github.com/kubeapps/kubeapps/pull/3204/files

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -328,7 +328,19 @@ func (s *Server) GetAvailablePackageDetail(ctx context.Context, request *corev1.
 		return nil, status.Errorf(codes.Internal, "Unable to retrieve chart: %v", err)
 	}
 
-	availablePackageDetail, err := AvailablePackageDetailFromChart(&chart)
+	if len(chart.ChartVersions) == 0 {
```
antgamdia, line 331: Is it really an error (I mean, `codes.Internal`), or perhaps it is just a temporary 404-alike error because the assets have not been fetched yet?

antgamdia (approved):
Great! Thanks for adding it! Really useful and necessary for my next step in the UI :P

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
absoludity: Yeah, I was wondering when adding this. The reason I chose an internal error is because the internal state is incorrect if this condition is true (since we have the chart in the DB, but without any chart versions - it should never be reached, since the `info` json blob is added atomically).

chart/kubeapps/templates/kubeappsapis/deployment.yaml
absoludity: Yeah, I noticed your commit, which was why I wondered if you'd missed my PR :) Thanks for the review!

gfichtenholt (commented):
looks fine

Issue #2944, opened 2021-06-07
[kubeapps-apis] Agreement for methods and messages for packaging API in design doc.


absoludity:
We're currently discussing and documenting the required methods and request/response messages for the packaging API on the related design doc.

These need to be ratified before we begin #.... , implementing them for two plugins (fluxv2 and one more).

ppbaena:
I think we can move this task to done, as we have an agreement for methods and messages for packaging API in the design doc. What do you think @absoludity ?

PR #3203, opened 2021-07-26
[kubeappsapis] Return categories as part of the GetAvailablePackageSummariesResponse

antgamdia:
### Description of the change

The current UI requires an endpoint to fetch the distinct categories to build the filters. There are mainly two implementation options: 1) overload an existing operation or 2) create a new one.

Since #2944 is still open but #helm #3085 should be finished soon, I just went with 1) as I overloaded the `GetAvailablePackageSummaries` operation. If it is invoked with `pageSize=-1` it will return all the distinct categories matching a given set of `FilterOptions`.

### Benefits

We can continue working on the UI revamping.

### Possible drawbacks

It is not a nice solution whatsoever... probably. if we do need this kind of results (district categories, or any other metadata), perhaps we should go implementing a new `getAbstract/projection/whatever(fields, scope, filters)` operation.

### Applicable issues

  - related #3165

### Additional information

N/A


absoludity (commented):
Sorry Antonio, I think there's been some mis-communication here. We should not be abusing the existing `response.AvailablePackageSummaries` slice to return the categories here, but rather adding a `Categories` field to the response message so that all `GetAvailablePackageSummariesResponse` messages can include the categories.

That is, when [@dlaloue-vmware said on the relevant design doc comment:](https://docs.google.com/document/d/1z03msZRGsRvcRaom-yrvEwIWcEfNy6fSA5Zg28gjYvA/edit?pli=1&disco=AAAAM6dYEOc)

> adding it to the package summaries makes sense.
> UI could, if needed, even do a dry-run query by passing "size=0" (pagination) and thus get only the categories."

I'm pretty sure from context that what he meant is that in addition to the `AvailablePackageSummary.Categories` field that we already have for each individual pkg, that:

- Adding a `Categories` field to the parent `GetAvailablePackageSummariesResponse` message makes sense, so that in addition to each package summary in the response defining its own category, the response includes all categories for available packages as well.
- If the UI just wanted to get all the categories and didn't even need any of the individual package summaries, it could call the function with a pagination option indicating a limit of zero [1], so that the response was effectively:
```
GetAvailablePackageSummariesResponse{
    AvailablePackageSummaries: [], // or nil or whatever the zero value is :P 
    Categories: ["aardvark", ..., "workflow"],    
}
```

Does that make sense? ie. I don't think anyone was suggesting overloading the function so that it *doesn't* do what it says it does (ie. return `AvailablePackageSummaries`)... that would be pretty awful indeed (I imagine you've been debating internally whether you should say something, given that's what you understood :P)

[1] I see you went with -1 because 0 is the zero value which means none was set (ie. no limit for the pagination). That makes sense and is fine from my POV, but tbh, we don't even need to implement that optional extra if it confuses things. As long as the response message includes all the categories, the UX has all the info it needs for now.

Let me know if that's not clear, and we can chat about it later.

dlaloue-vmware:
I agree with Michael said.

typically, page size = -1 is to indicates that no pagination is required and all results should be returned. page size = 0 should return an empty list (this is often used to get other info like the total count etc...).

also, i believe we have not yet finalized whether we really want the categories returned with the summaries or as an another cal (see comment in [spec](https://docs.google.com/document/d/1z03msZRGsRvcRaom-yrvEwIWcEfNy6fSA5Zg28gjYvA/edit?disco=AAAANGrmDNA)).
there are 2 issues i believe we need to discuss:
 - performance issues, can we ask this info on every page request
 - should the categories be scoped only to the packages the user can see, or be all categories regardless
the answer to those 2 questions will affect the implementation.

absoludity:
> I agree with what Michael said.
> 
> typically, page size = -1 is to indicate that no pagination is required and all results should be returned. page size = 0 should return an empty list (this is often used to get other info like the total count etc...). we are not trying to overload the semantics.

That depends whether you think users expect by default (ie. when they don't specify any pagination options) to get 0 results back, or some results back :) I'd stick to the pattern we have from the [protobuf pagination design pattern we're following](https://cloud.google.com/apis/design/design_patterns#list_pagination):

> If the `page_size` is 0, the server will decide the number of results to be returned.

Given than 0 is the default (zero) value for an integer field, I as a user would expect to get *something* back if I didn't specify it. I think this is what the recommendation means by "the server will decide the number of results to be returned" rather than just returning 0 results. Whether we choose to go with a default of all results (our current default) or some max size we're happy with, I don't care, but I don't think we should return zero results when pagination is not specified (which is implied by your point I think, Dimitri).

> 
> also, i believe we have not yet finalized whether we really want the categories returned with the summaries or as another call (see comment in [spec](https://docs.google.com/document/d/1z03msZRGsRvcRaom-yrvEwIWcEfNy6fSA5Zg28gjYvA/edit?disco=AAAANGrmDNA)).

Right, we assumed from your comment at the end of the doc which  Iinked to and quoted above, that you were +1 for it being returned with the summaries.

> there are 2 issues i believe we need to discuss:
> 
>     * performance issues, can we ask this info on every page request

This is no issue for the helm plugin and should be no different for any *cached* result (when we have generic caching).

> 
>     * should the categories be scoped only to the packages the user can see, or be all categories regardless
>       the answer to those 2 questions will affect the implementation.

I had pushed in the past to have them scoped only to the packages in the users current filter, but it was rightly pointed out that this does not provide a good UX experience, and so we had already decided to return all categories for the unfiltered available apps (you'll see this is what the existing Kubeapps UX does).

There's a third point to consider (which Antonio and I did consider): in the long run we may find there are other "annotations" or metadata that we want to fetch aggregated across packages, so in the long-term we may find a need for a general GRPC function that returns an aggregation of a provided annotation (eg. 'category', or 'maintainers' or 'licenses'). But right now I think it'd be premature to do so.

So I'm +1 for this PR being updated so that the Response includes `Categories` which is populated and returned with the categories for available packages for all requests, but am OK if people would rather do it as a separate-yet-simple `GetAvailablePackageCategories()` method.

dlaloue-vmware:
> Right, we assumed from your comment at the end of the doc which Iinked to and quoted above, that you were +1 for it being returned with the summaries.

In the comment, i was merely pointing out the categories were missing. You replied with a comment raising the question of performance, and that comment is still not resolved. that is why i have mentioned it here as i thought we still needed to discuss your comment. If you think performance is not an issue, then we can update the doc, resolve the spec comment and proceed with the implementation.

> we had already decided to return all categories for the unfiltered available apps

can you clarify if you mean that it is for available packages visible to the user or for all available packages regardless of permissions?
if we were not doing caching for example, we would have to scan all packages the user can see and aggregate the categories. caching should normally not change permissions rights. i am trying to clarify what we mean and will implement.

absoludity:
> > Right, we assumed from your comment at the end of the doc which Iinked to and quoted above, that you were +1 for it being returned with the summaries.
> 
> In the comment, i was merely pointing out the categories were missing. You replied with a comment raising the question of performance, and that comment is still not resolved. that is why i have mentioned it here as i thought we still needed to discuss your comment. If you think performance is not an issue, then i am ok to go ahead and update the doc, resolve the spec comment and proceed with the implementation.

Again, *different* comment (confusing I know, that's why I tried to link above to the specific but different comment). I'm trying to refer to your comment [here](https://docs.google.com/document/d/1z03msZRGsRvcRaom-yrvEwIWcEfNy6fSA5Zg28gjYvA/edit?pli=1&disco=AAAAM6dYEOc) where you said: "adding it to the package summaries makes sense. UI could, if needed, even do a dry-run query by passing "size=0" (pagination) and thus get only the categories."

 
> 
> > we had already decided to return all categories for the unfiltered available apps
> 
> can you clarify if you mean that it is for available packages visible to the user or for all available packages regardless of permissions?
> if we were not doing caching for example, we would have to scan all packages the user can see and aggregate the categories. caching should normally not change permissions rights. i am trying to clarify what we mean and will implement.

The current Kubeapps UX fetches all categories once from the cache, regardless of permissions, I believe (Antonio can correct me if wrong). The backend function allows only returning categories for a chart query, but afaict from the behaviour, we don't use that. This means that if a private `AppRepository` is on the cluster and has a chart with a category of `Category: my-password-is-foo`, that this data would be visible to all users. That said, I see no reason why we couldn't limit the categories to those from app repositories to which the user has access (though that will indeed make it more expensive which influences other decisions).

Keen to hear what Antonio thinks.

antgamdia:
Woah, what an interesting discussion! 

> Sorry Antonio, I think there's been some mis-communication here. We should not be abusing the existing response.AvailablePackageSummaries slice to return the categories here, but rather adding a Categories field to the response message so that all GetAvailablePackageSummariesResponse messages can include the categories.

Yes, I totally misunderstood it; my bad! Fortunately, it's a simple change to perform; I'll do it in a while and will see what you all think.

> If the page_size is 0, the server will decide the number of results to be returned.

I'll remain unaltered this logic in this PR, but +1 to return everything (with a configurable upper threshold perhaps for performance). 

> can you clarify if you mean that it is for available packages visible to the user or for all available packages regardless of permissions?

Under the hood what this `GetAllChartCategories` is doing is:

```sql
SELECT (info ->> 'category') AS name, COUNT( (info ->> 'category')) AS count FROM charts WHERE .........  GROUP BY (info ->> 'category') ORDER BY (info ->> 'category') ASC
```

We are just passing the Namespace, so there `where` clause becomes:   `(repo_namespace = cq.Namespace OR repo_namespace = m.GetKubeappsNamespace())`

So, we aren't fetching the categories of the entire catalog, but the ones in either the current namespace or the global namespace.
That said, as there are no restrictions on this operation, a user can possibly (and manually) query the categories in any namespace-

> That said, I see no reason why we couldn't limit the categories to those from app repositories to which the user has access 

Yep, I agree. Even if it is possible to query them, the UI will just display the ones the user has access to (because we are querying by namespace)


So, let me change the impl to the one you suggested, Michael (and sorry for the misunderstanding :S)

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -200,54 +200,47 @@ func (s *Server) GetAvailablePackageSummaries(ctx context.Context, request *core
...
-	// A GetAvailablePackageSummaries request with pageSize=-1 will return a GetAvailablePackageSummariesResponse
-	// with a single empty AvailablePackageSummary containing only the distinct category names considering
-	// the FilterOptions. Otherwise, return the usual response after querying all the available packages
-	if pageSize == -1 {
-		chartCategories, err := s.manager.GetAllChartCategories(cq)
-		if err != nil {
-			return nil, status.Errorf(codes.Internal, "Unable to fetch chart categories: %v", err)
-		}
+	// This plugin will include, as part of the GetAvailablePackageSummariesResponse,
+	// a "Categories" field containing only the distinct category names considering the FilterOptions
```
absoludity, line 204: So I know that the existing code accepts a chart query argument, that you've used below. But in the UX, I see that we always show all categories for the namespace, regardless of any filtering. Which at first left me confused as to why we'd want to return just the categories for the packages (charts) matching the query here. But then I realised that the UX can just use the categories from the initial response when the user views the catalog? Is that what you intend?

absoludity (approved):
LGTM, thanks Antonio.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
antgamdia: As it is currently implemented in the UI, we do make two separate calls. In fact, this could be an issue in that a size=0 query will return every availablePackageSummary... . We might work around it by setting size=1, but still a non-elegant solution.
I can try reducing it to a single call in https://github.com/kubeapps/kubeapps/pull/3203, once we got the first UI PR merged.

Allowing passing the FilterOptions through is just for not limiting any future use case; I mean, what if we decide to make the categories dynamic? Another reason is that we are already using a "chartQuery" when passing the namespace, so why not passing also the rest of the query?

absoludity: > We might work around it by setting size=1, but still a non-elegant solution.

But what I'm saying is that I don't think we'd need to, in that, when a user first navigates to the "App Catalog", the UI will be calling `GetAvailablePackageSummaries()` with some pagination options only (ie. no filtering). The `utils.ChartQuery` struct only will only have the namespace set. So this initial response will contain all the categories that you need, I think? Or why do you think a second query would be required with size=1 or similar to get them? The only can I can imagine that being necessary is if the user filters by name, then hits re-load, so the UX redux state does not already contain all the categories for the namespace or similar. Or maybe we don't (yet) store them in the redux state.

Issue #2572, opened 2021-03-18
CSS issues in login page


antgamdia:
Note the white background color on the left and the wrong min-height:

![image](https://user-images.githubusercontent.com/11535726/111631871-3374e180-87f4-11eb-8991-414ccf12c9fd.png)

Also, the `cds-theme` property is not being set since the migration of the theme set to the redux state.

PR #3205, opened 2021-07-27
Don't set theme in local storage unless user sets.

absoludity:
### Description of the change

In #2573 a couple of CSS issues were fixed, one of which was to update the config reducer so that when a `receiveTheme` action is triggered:

- it calls setTheme(), explicitly setting the theme in the local config, and
- it updates the redux state to include the new theme value.

As a result of this, even if a user doesn't explicitly choose a theme, as soon as `getConfig()` is called, the local storage is set with a choice the user did not make.

I suspect (but am not 100%), that we don't need to be calling `setTheme()` to set the local storage in this case.

I've built an image from this and tested it, and it appears to function as expected, with the local storage only being set if I, the user, explicitly choose a theme.

@antgamdia Can you build and verify that this doesn't re-introduce any of the issues you were fixing in the related issue #2572? I cannot see any of them resurfacing, and I see the `cds-theme` set correctly both before and after an explicit selection is made.

### Benefits

It'll enable @aanthonyrizzo to continue thinking about a solution in #3191, with the assumption that the theme is only ever set in the local storage if the user has explicitly chosen a theme.

### Possible drawbacks

None that I can see, but let's see what Antonio knows about the original issues.

### Applicable issues

  - ref #2572 #3176

### DIFF
--- a/dashboard/src/reducers/config.ts
+++ b/dashboard/src/reducers/config.ts
@@ -31,7 +31,6 @@ const configReducer = (state: IConfigState = initialState, action: ConfigAction)
         ...action.payload,
       };
     case getType(actions.config.receiveTheme):
-      Config.setTheme(action.payload);
       return {
         ...state,
         theme: action.payload,


antgamdia:
I've just pulled your changes and it seems the issue pops up again (note the white gap):

image

antgamdia (commented):
I've just pulled your changes and it seems the issue pops up again (note the white gap):

image

I think it needs a deeper refactor. Have a look at my proposal here: https://github.com/kubeapps/kubeapps/pull/3209

absoludity:
> I've just pulled your changes and it seems the issue pops up again (note the white gap):
> 

Wonder why I couldn't reproduce the white border? I tried to (FF, had OIDC logon, not sure if that made a difference) and couldn't see it, and verified that the class was set on the body etc.

> 
> I think it needs a deeper refactor. Have a look at my proposal here: #3209

Yep, I'll close this, but as mentioned on your PR, I still think we should not be setting the local storage unless the user chooses a theme (rather than whenever we evaluate the theme).

PR #3206, opened 2021-07-27
Add AppVersion and Status to GetInstalledPackageSummaries.

absoludity:
### Description of the change

While looking at the current Kubeapps UI for browsing installed packages, I noticed each tile:

has additional info that wasn't currently included, including:

- Status info
- app version

This PR fixes that.

### Benefits

The API call should now include all info required for the UX.

### Possible drawbacks

None

### Applicable issues

  - Ref #3146


cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
```diff
@@ -605,3 +615,24 @@ message VersionReference {
...
+  // Ready
+  //
+  // An indication of whether the installation is ready or not
+  bool ready = 1;
+
+  // Reason
+  //
+  // A single word indicating the reason for the current status. For example, "Pending",
+  // "InstallSucceeded", or "InstallFailed"
+  string reason = 2;
```
antgamdia, line 632: I guess this will be (in the helm-direct) something like this: https://github.com/helm/helm/blob/main/pkg/release/status.go#L23

Just thinking off the top of my mind: would including a list of every possible state somewhere be useful? I don't really think so. Just perhaps adding it as an `enum` in the code implementation, but not a user-facing thing.

chart/kubeapps/templates/kubeappsapis/rbac.yaml
```diff
@@ -24,7 +24,7 @@ rules:
   # We'll need to add rbac for our dev user to install later as well.
   - apiGroups: [""]
     resources: ["secrets"]
-    verbs: ['get']
+    verbs: ["get", "list"]
```
antgamdia, line 27: I guess this is for listing the details of an installed release?

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
```diff
@@ -415,12 +420,17 @@ message InstalledPackageSummary {
...
   // LatestPkgVersion
   //
   // The latest version available for this package, regardless of the pkg_version_reference.
-  string latest_pkg_version = 9;
+  string latest_pkg_version = 10;
+
+  // Status
+  //
+  // The current status of the installed package.
+  InstalledPackageStatus status = 11;
```
antgamdia, line 433: So, right now we should go with the polling approach, don't we?
If I understood it properly. the alternative to our current ws impl is using grpc streams like:

`rpc getInstalledPackageSummaryStatus(InstalledPackageSummaryRequest) returns (stream Status);` 

?  (but totally out of scope right now, I know)

antgamdia (approved):
Awesome! Looking forward to starting using these operations in the UI. Thanks!

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
antgamdia: A second thought: perhaps a kind of standardization on the statuses is, indeed, required, no? If not, how can we know if a package is OK/pending/failed/etc from the core API ?

PR #3208, opened 2021-07-27
Add uniform bash error handling opts

antgamdia:
### Description of the change

I just noticed we are using different bash error handling flags in a couple of sh files. This PR is just for making every file use the same. 
See https://elder.dev/posts/safer-bash/#enable-error-handling-options for further reference. Happy to change it to the shorter format `set -eu -o pipefail`

### Benefits

Every script we use will behave the same way.

### Possible drawbacks

Perhaps we'll unveil some errors? Let's wait to our beloved CI to discover

### Applicable issues

N/A

### Additional information

N/A


antgamdia:
Mmm, some unset variables did pop up... making as a draft til we have some bandwidth to have a look at it.

absoludity (approved):
+1 when the unset vars are fixed.

Preference for the longer (more readable versions) :)

PR #3209, opened 2021-07-27
Improve theme switcher mechanism

### LINKED ISSUE #3176: Define default dashboard theme from helm

antgamdia:
### Description of the change

After #3205 and #3191, we noticed we were overriding the theme config. It was the result of small improvements towards the dark mode and bug fixes along the way.
This PR is the first step at having a clearer logic on which theme preference takes precedence over another.

Currently: `user (via toggle)>system (via values.yaml)>browser (via browser config)>fallback (hardcoded light theme)`

### Benefits

It (should) paves the way to create other config parameters like "allowChangingTheme" or similar.

### Possible drawbacks

I've checked and I don't foresee any error... but UI can always fail OSI L8 actors are involved :P 

### Applicable issues

  - fixes #3176

### Additional information

I've used the changes from #3191 as my base branch, all the credits should go for @aanthonyrizzo. I'm thinking of merging that PR and then this one? cc @absoludity 

Here is a demo:

https://user-images.githubusercontent.com/11535726/127175624-2252d77a-541b-46a5-8fb5-c4aed5f4ddd6.mp4

### DIFF
--- a/dashboard/src/actions/config.ts
+++ b/dashboard/src/actions/config.ts
@@ -13,14 +13,11 @@ export const receiveConfig = createAction("RECEIVE_CONFIG", resolve => {
 export const receiveTheme = createAction("RECEIVE_THEME", resolve => {
   return (theme: SupportedThemes) => resolve(theme);
 });
-export const setThemeState = createAction("SET_THEME", resolve => {
-  return (theme: SupportedThemes) => resolve(theme);
-});
 export const errorConfig = createAction("ERROR_CONFIG", resolve => {
   return (err: Error) => resolve(err);
 });
 
-const allActions = [requestConfig, receiveConfig, receiveTheme, setThemeState, errorConfig];
+const allActions = [requestConfig, receiveConfig, receiveTheme, errorConfig];
 export type ConfigAction = ActionType<typeof allActions[number]>;
 
 export function getConfig(): ThunkAction<Promise<void>, IStoreState, null, ConfigAction> {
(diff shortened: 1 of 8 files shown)


aanthonyrizzo:
@antgamdia This looks perfect! Thank you guys for the support on this feature. Please let me know if there is anything that you guys need from me in regard to this PR or the one that I opened. You can merge #3191 at will or just include any changes u need from it into this PR. Thanks again!

dashboard/src/actions/config.ts
```diff
@@ -35,18 +32,28 @@ export function getConfig(): ThunkAction<Promise<void>, IStoreState, null, Confi
   };
 }
 
+// getTheme retrieves the current server configuration, calculates the proper theme
+// and stores it in both the redux state and the user's localstorage
```
absoludity, line 36: I don't understand why `getTheme` is additionally storing the gotten theme in the local storage. This is exactly what I thought the problem was: it should only ever be set in local storage iff the user explicitly makes a decision, I think? (given that it is treated as the user's preference)

For example, suppose kubeapps is installed and used by a user for the first time, and their window prefs say light. Now when this `getTheme` action is run, it determines the light theme from the browser (given no other indicators) and stores the theme in local storage. At that point, it's no longer possible to determine whether that was a user-preference or browser or system configuration. So if the installation is updated with a specific theme in the values, it won't override the previous preference of the users browser since the code determines that the light theme recorded in the local storage is now a user preference.

Everything else about this PR looks great (and clear!), just wondering if we need to store the it in local storage here, or can instead only do so when the user changes the theme explicitly.

absoludity (approved):
+1, but please see below comment before merging.

dashboard/src/actions/config.ts
antgamdia: Yeah! You are right, we can definitely do that. I've added some changes aimed at that: I've split  the `setTheme` and `setUserTheme`. The former updates the redux state but the latter also save it in the local storage.

This other video should showcase what you meant in this comment:
https://user-images.githubusercontent.com/11535726/127335777-3e7cdef0-75e9-4af5-8fb4-1b716f13ea30.mp4

absoludity: Great thanks.

PR #3210, opened 2021-07-27
Fetch categories using grpc in UI

antgamdia:
### Description of the change

This PR simply changes the current fetchChartCategories endpoint by the grpc one + changes in tests.

### Benefits

Moving ahead towards the UI revamping

### Possible drawbacks

N/A

### Applicable issues

  - related #3165

### Additional information

Marking as a draft until #3164 gets merged. Jus creating the PR is someone wants to have a look, but I have to force-push to clean the commit mess I currently have :P


antgamdia:
Take https://github.com/kubeapps/kubeapps/pull/3203#discussion_r678689686 into account.

antgamdia:
> when a user first navigates to the "App Catalog", the UI will be calling GetAvailablePackageSummaries() with some pagination options only (ie. no filtering). 

That's what I thought also... but there is another use case: "ey buddy, install one of these charts `http://localhost:3000/#/c/default/ns/default/catalog?Search=nginx&Repository=bitnami`"
In this scenario the initial query is not "clean", therefore we cannot blindly rely on it.

If we fetch the categories there, we will show an incomplete view (in fact, we don't reload the categories until the context changes).

So, I foresee a couple of options:

1. We go with the `size=1` or any other tricky combination of page,size params to reduce the network data transference
2. We remove the filters in the backed, so they `categories` field just returns those categories in the required namespace.
3. We are more flexible with the "dynamic filters" and make them dynamic in each request. Alternatives:
    1. Including pagination: new categories will always refer to something the users can view and select
    2. Omitting pagination: we will display the whole list of categories even though the users hasn't reached the page in which it appears.

I'm holding any further changes here until we decide what to do.

absoludity:
> > when a user first navigates to the "App Catalog", the UI will be calling GetAvailablePackageSummaries() with some pagination options only (ie. no filtering).
> 
> That's what I thought also... but there is another use case: "ey buddy, install one of these charts `http://localhost:3000/#/c/default/ns/default/catalog?Search=nginx&Repository=bitnami`"
> In this scenario the initial query is not "clean", therefore we cannot blindly rely on it.

Yeah, that's what I meant by refreshing the page (when the redux state isn't already populated with the categories). But that's easy to detect (if the redux state doesn't have the categories for the namespace ).

> If we fetch the categories there, we will show an incomplete view (in fact, we don't reload the categories until the context changes).
> 
> So, I foresee a couple of options:
> 
>     1. We go with the `size=1` or any other tricky combination of page,size params to reduce the network data transference
> 
>     2. We remove the filters in the backed, so they `categories` field just returns those categories in the required namespace.

I'm fine with 2. When we figure out whether we can provide a more general, separate, grpc method to get not only categories, but other aggregated data, we can deprecate this one and remove it.

> 
>     3. We are more flexible with the "dynamic filters" and make them dynamic in each request. Alternatives:
>        
>        1. Including pagination: new categories will always refer to something the users can view and select
>        2. Omitting pagination: we will display the whole list of categories even though the users hasn't reached the page in which it appears.

I don't think we can make the UX categories displayed dynamic (ie. update based on current filtering), as we've discussed previously. If I want to show packages that are in either the Workflow category or the LMS category, I can't - because as soon as I select one category, the result set is filtered by that removing any option of the other (unless packages have multiple categories, which they don't with the bitnami  catalog).
> 
> 
> I'm holding any further changes here until we decide what to do.

My vote is for your 2. for now (ie. we just pass an empty filter set to get just categories for the namespace, always).

antgamdia:
> My vote is for your 2. for now (ie. we just pass an empty filter set to get just categories for the namespace, always).

Great, I also think it is the way to go right now. I've sent a separate PR for the API change at https://github.com/kubeapps/kubeapps/pull/3225

dashboard/src/actions/charts.ts
```diff
@@ -119,29 +115,13 @@ export function fetchCharts(
           nextPageToken: response.nextPageToken,
         }),
       );
+      dispatch(receiveChartCategories(response.categories));
```
antgamdia, line 118: This is the key change, once we don't consider the FilterOptions (https://github.com/kubeapps/kubeapps/pull/3225) we can safely rely on the same fetchCharts calls to update the categories.

absoludity (approved):
+1 once you've verified with #3225 merged.

dashboard/src/shared/types.ts
```diff
@@ -77,9 +75,8 @@ export interface IChart {
 }
 
 export interface IReceiveChartsActionPayload {
-  items: AvailablePackageSummary[];
+  response: GetAvailablePackageSummariesResponse;
   page: number;
-  nextPageToken: string;
 }
```
antgamdia, line 80: I've simplified this payload to avoid having unnecessary extra fields (especially, now we have autogenerated models)

PR #3219, opened 2021-07-29
Add some logging when creating a screenshot.

absoludity:
Also ensure a screenshot is created when logging in so the directory is
non-empty.

### Description of the change

While trying to find out why CI isn't saving screenshots as artifacts anymore, I checked a build from a few months ago, and there I see the following in the Uploading Artifacts step log:
```
Uploading /home/circleci/project/integration/reports to integration/reports
Uploading /home/circleci/project/integration/reports/.notempty: SKIPPED, file is empty
Uploading /home/circleci/project/integration/reports/screenshots/upgrades-an-application.png (20 kB): DONE
```

whereas today, we see only the first two lines, implying that circle ci thinks there's nothing to upload.

This PR just adds some logging so we can test whether the directory we think we're writing to is actually the one we're expecting to write to.

### Benefits

Hopefully shed light on what's happening with screenshots.

### Applicable issues

  Just trying to help Antonio with the CI issues for the UX integration.

### DIFF
--- a/integration/jest.environment.js
+++ b/integration/jest.environment.js
@@ -25,7 +25,7 @@ class ScreenshotOnFailureEnvironment extends PuppeteerEnvironment {
     try {
       // Check the server is up before running the test suite
       console.log(
-        `Waiting ${endpoint} to be ready before running the tests 
+        `Waiting ${endpoint} to be ready before running the tests
         (${waitTimeout / 1000}s)`,
       );
       await waitOn({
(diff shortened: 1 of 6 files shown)


absoludity:
Bingo: so the failure in this test run is because we're calling the retry helper with a blank test name it seems: https://app.circleci.com/pipelines/github/kubeapps/kubeapps/3783/workflows/8674ab14-1687-4a73-b4bc-5a09b51e7b5a/jobs/54036/artifacts

absoludity:
Updated to set the test name for each call of the retry.

antgamdia (approved):
Awesome! Thanks for catching this bug!

PR #3228, opened 2021-07-31
GetInstalledPackageSummaries in fluxv2 plug-in

gfichtenholt:
Adds GetInstalledPackageSummaries() functionality to fluxv2 plug-in. Unit tests included


cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/chart.go
```diff
@@ -26,73 +26,179 @@ import (
...
+	} else if url != "" {
+		return url, nil, nil
+	}
+
+	// did not find the chart, need to create
+	// see https://fluxcd.io/docs/components/source/helmcharts/
+	// TODO (gfichtenholt)
+	// 1. HelmChart object needs to be co-located in the same namespace as the HelmRepository it is referencing.
+	// 2. flux impersonates a "super" user when doing this (see fluxv2 plug-in specific notes at the end of
+	//	design doc). We should probably be doing simething similar to avoid RBAC-related problems
```
absoludity, line 105: FWIW, I don't think we'll ever include a `cluster-admin` service account with Kubeapps or a plugin - that's just too dangerous, as you can see from the [flux proposal](https://github.com/fluxcd/flux2/pull/582/files#diff-e377d0286cdc254b5541414d2db6f291be9d92a669ccb2d4e7fb14a8645dfa57) (that hasn't been accepted yet) which you linked to in the doc, about how they're trying to get out of that situation.

Helm itself used to do something similar, having `tiller` running in the cluster with crazy privileges that could be used to do exactly what they describe in the flux proposal... I'm quite surprised that they have `cluster-admin` as the default for the helm controller TBH, given the history.
Anyway, fine as is, as long as we're not wanting to deploy a kubeapps component with super privs.
So back to here: The user carrying out the request will need to have RBAC for whatever is done - currently it looks like you're creating the flux HelmChart CR explicitly. If that is necessary (see previous comment above as I'm not sure why it is if the HelmRelease already exists) we can document that as a requirement.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/chart.go
```diff
@@ -26,73 +26,179 @@ import (
...
+		log.Errorf("Error creating chart: %v\n%v", err, unstructuredChart)
+		return "", err, nil
+	}
+
+	log.Infof("Created chart: [%v]", prettyPrintMap(newChart.Object))
+
+	// Delete the created helm chart regardless of success or failure. At the end of
+	// GetAvailablePackageDetail(), we've already collected the information we need,
+	// so why leave a flux chart chart object hanging around?
+	// Over time, they could accumulate to a very large number...
```
absoludity, line 124: Yes. More generally it seems a bit icky to be creating them when they should exist if a HelmRelease already exists (and if not, I'm not sure we should be trying to provide the info). It seems analogous to the situation with normal helm, where if the user has deleted the AppRepository (ie. so kubeapps no longer has info about the related charts), we can't provide that info (such as upgrade info or latest version) and that's OK.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/chart.go
```diff
@@ -26,73 +26,179 @@ import (
...
-						}
-						// chart pull is done and it's a failure
-						return true, status.Errorf(codes.Internal, msg)
-					}
+		return "", err, nil
+	} else if url != "" {
+		return url, nil, nil
+	}
+
+	// did not find the chart, need to create
```
absoludity, line 100: I don't understand why we need to be creating a chart here at all. We're wanting to return information about the installed packages (ie. the HelmReleases). From what I've understood, if a user has created a `HelmRelease` CR, then the helm-controller would have created the related `HelmChart` for that release, so it should exist. If it doesn't, because the user has manually deleted it or whatever, then the best we can do is return the info about the installed package that we have (ie. the info from the HelmRelease).

Or what reason do you see here for a HelmRelease without the corresponding HelmChart?

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/chart.go
```diff
@@ -26,73 +26,179 @@ import (
...
-						if msg, ok = conditionAsMap["message"].(string); !ok {
-							msg = fmt.Sprintf("No message available in condition: %v", conditionAsMap)
-						}
-						// chart pull is done and it's a failure
-						return true, status.Errorf(codes.Internal, msg)
-					}
+		return "", err, nil
+	} else if url != "" {
+		return url, nil, nil
+	}
```
absoludity, line 98: I'm probably missing something, but I think we should just be doing
```
return url, nil, nil
```
here, regardless. More below.

absoludity (commented):
Hi Greg.

I started going through this, but there's something fundamental that I'm not understanding which probably needs clarifying before I keep going: This is servicing a request for info about the installed packages (`HelmRelease` CRs). Now with Flux , AIUI, when a user installs a package (ie. creates a `HelmRelease`), flux creates the related `HelmChart` CR. So when we're servicing this request, we're returning info about the `HelmReleases` with possibly some extra info from the related `HelmChart`. But a lot of the code here seems to be catering for a scenario where the related `HelmChart` doesn't exist and creating it (temporarily), and I don't yet understand why we'd want to do that?

Thanks for any info.

gfichtenholt:
> Hi Greg.
> 
> I started going through this, but there's something fundamental that I'm not understanding which probably needs clarifying before I keep going: This is servicing a request for info about the installed packages (`HelmRelease` CRs). Now with Flux , AIUI, when a user installs a package (ie. creates a `HelmRelease`), flux creates the related `HelmChart` CR. So when we're servicing this request, we're returning info about the `HelmReleases` with possibly some extra info from the related `HelmChart`. But a lot of the code here seems to be catering for a scenario where the related `HelmChart` doesn't exist and creating it (temporarily), and I don't yet understand why we'd want to do that?
> 
> Thanks for any info.

Hi Michael, 
I did move code around a bit. Looks like your comment is about lines 78 or so, in other words, getChartTarball() func. This isn't used in GetInstalledPackageDetail() at all. Only in GetAvailablePackageDetail(). Otherwise, your understanding is correct.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/chart.go
gfichtenholt: +1. If you'd like me to word the comment a certain way, just say so.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/chart.go
gfichtenholt: GetAvailablePackageDetail() requires me to provide some info not available otherwise, like the readme. For a chart that may not be installed at all. I can only get that if I have the chart tgz file, so that's the only reason I am doing it. I am open to other proposals, please.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/chart.go
gfichtenholt: no, this is a different use case - GetAvailablePackageDetail()

absoludity: Ah - ok. I assumed this PR was just about GetInstalledPackageSummaries . Great, I'll finish going through the PR tomorrow.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/chart.go
absoludity: I think the comment is fine as a note - word it however you like :) I just don't think option 2 is something we should do (even flux seems to be backing away from it).

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/chart.go
absoludity: Yep, it makes sense for GetAvailablePackageDetail() - and given that's a singular request, the efficiency isn't an issue either. As above, I'd thought this PR was about GetInstalledPackageSummaries only as per the title, so when I saw these changes, assumed it must be for them.

absoludity (commented):
Great, thanks for the info (that lots of the changes in this PR are about GetAvailablePackageDetail, not GetInstalledPackageSummaries). I'll finish it tomorrow then :)

absoludity (approved):
Great, thanks Greg.

gfichtenholt:
> Great, thanks Greg.

thank you

PR #3242, opened 2021-08-05
feat(form): Allow remote loading for form components

aanthonyrizzo:
### Description of the change

This PR adds the ability for developers to load their custom components from a remote URL as an alternative to loading from the configmap. This is an optional value that can be set for the custom components feature. If the value is not set the `<CustomComponent />` will be pulled from the `custom_components.js` file.

### Benefits

- No longer worry about 1 MiB max configmap size
- Allows for easier version control of the form components (should that be needed)

### Possible drawbacks

- Bundle will have to be served from a remote host and will be the responsibility of the developer

### DIFF
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -437,6 +437,9 @@ dashboard:
   ## ref: https://github.com/kubeapps/kubeapps/blob/master/docs/developer/custom-form-component-support.md
   ##
   customComponents: ""
+  ## @param dashboard.remoteComponentsUrl Remote URL that can be used to load custom components vs loading from the local filesystem
+  ##
+  remoteComponentsUrl: ""
   ## @param dashboard.customLocale Custom translations injected to the Dashboard to customize the strings used in Kubeapps
   ## ref: https://github.com/kubeapps/kubeapps/blob/master/docs/developer/translate-kubeapps.md
   ## e.g:
--- a/dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/CustomFormParam.test.tsx
+++ b/dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/CustomFormParam.test.tsx
@@ -1,4 +1,4 @@
-import { mount } from "enzyme";
+import { getStore, mountWrapper } from "shared/specs/mountWrapper";
 import { IBasicFormParam } from "shared/types";
 import { CustomComponent } from "../../../RemoteComponent";
 import CustomFormComponentLoader from "./CustomFormParam";
@@ -17,6 +17,10 @@ const defaultProps = {
   handleBasicFormParamChange: jest.fn(),
 };
 
+const defaultState = {
+  config: { remoteComponentsUrl: "" },
+};
+
 // Mocking the window so that the injected components are imported correctly
 const location = window.location;
 beforeAll((): void => {
@@ -32,11 +36,36 @@ afterAll((): void => {
 });
 
 it("should render a custom form component", () => {
-  const wrapper = mount(<CustomFormComponentLoader {...defaultProps} />);
+  const wrapper = mountWrapper(
+    getStore(defaultState),
+    <CustomFormComponentLoader {...defaultProps} />,
+  );
   expect(wrapper.find(CustomFormComponentLoader)).toExist();
 });
 
 it("should render the remote component", () => {
-  const wrapper = mount(<CustomFormComponentLoader {...defaultProps} />);
+  const wrapper = mountWrapper(
+    getStore(defaultState),
+    <CustomFormComponentLoader {...defaultProps} />,
+  );
   expect(wrapper.find(CustomComponent)).toExist();
+});
+
+it("should render the remote component with the default URL", () => {
+  const wrapper = mountWrapper(
+    getStore(defaultState),
+    <CustomFormComponentLoader {...defaultProps} />,
+  );
+  expect(wrapper.find(CustomComponent)).toExist();
+  expect(wrapper.find(CustomComponent).prop("url")).toContain("custom_components.js");
+});
+
+it("should render the remote component with the URL if set in the config", () => {
+  const wrapper = mountWrapper(
+    getStore({
+      config: { remoteComponentsUrl: "www.thiswebsite.com" },
+    }),
+    <CustomFormComponentLoader {...defaultProps} />,
+  );
+  expect(wrapper.find(CustomComponent).prop("url")).toBe("www.thiswebsite.com");
 });
--- a/dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/CustomFormParam.tsx
+++ b/dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/CustomFormParam.tsx
@@ -1,7 +1,7 @@
 import React, { useMemo } from "react";
-import { IBasicFormParam } from "shared/types";
+import { IBasicFormParam, IStoreState } from "shared/types";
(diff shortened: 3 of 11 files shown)


absoludity:
Thanks @aanthonyrizzo , this looks really interesting, and to some degree, solves an up-coming issue that we'll have (enabling plugins for different packaging formats to provide their own bundle with custom form stuff).

Sorry I haven't had a chance to take a close look at the details (and am away for a few days to a week - but hopefully someone else on the team can take a look in the mean-time). I'm guessing the solution here just allows a single bundle for a single custom form (not a problem - I mean, that's what the current one does too, just verifying, as for our plugin support, we'll be checking for a bundle for each plugin eventually).

Thanks for the extra effort to get these improvements landed here too :)

aanthonyrizzo:
Yes exactly. The bundle contains a bunch of different form components that I load into the deployment form. Its basically does a match on a value that I set in the `values.schema.json` and the bundle just returns the necessary component

dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/CustomFormParam.test.tsx
```diff
@@ -32,11 +36,35 @@ afterAll((): void => {
...
+  );
   expect(wrapper.find(CustomComponent)).toExist();
 });
+
+it("should render the remote component with the default URL", () => {
+  const wrapper = mountWrapper(
+    getStore(defaultState),
+    <CustomFormComponentLoader {...defaultProps} />,
+  );
+  expect(wrapper.find(CustomComponent)).toExist();
```
antgamdia, line 59: Should it also check the URL is the expected one (`....custom_components.js`) ?

chart/kubeapps/values.yaml
```diff
@@ -437,6 +437,9 @@ dashboard:
   ## ref: https://github.com/kubeapps/kubeapps/blob/master/docs/developer/custom-form-component-support.md
   ##
   customComponents: ""
+  ## @param dashboard.remoteComponentsUrl Remote URL that can be used to load custom components vs loading from the local filesystem
```
antgamdia, line 440: This param should also appear in the readme, not a big deal as we always re-generate the chart readme before releasing.

If curious, we use a bitnami tool for that. Here are the steps I use to follow:

1. `git clone https://github.com/bitnami-labs/readme-generator-for-helm.git`
2. ` cd readme-generator-for-helm/`
3. ` npm install`
4. `node bin/index.js  -r ../kubeapps/chart/kubeapps/README.md -v ../kubeapps/chart/kubeapps/values.yaml `

dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/CustomFormParam.tsx
```diff
@@ -14,7 +14,14 @@ export default function CustomFormComponentLoader({
   handleBasicFormParamChange,
 }: ICustomParamProps) {
   // Fetches the custom-component bundle served by the dashboard nginx
-  const url = `${window.location.origin}/custom_components.js`;
+
+  const {
+    config: { remoteComponentsUrl },
+  } = useSelector((state: IStoreState) => state);
+
+  const url = remoteComponentsUrl
```
antgamdia, line 22: As this URL is a Kubeapps' admin responsibility, it's ok accepting any value, but, have you consider these scenarios?

- Kubeapps running with HTTPS trying to load a custom component served via HTTP.
- Kubeapps running in a domain totally different from the one where the custom component is served (i.e., cross-origin requests).

I'm just wondering if adding this remote URL loading would also imply some changes in the headers (CSP, CORS, etc) used in the nginx server.

dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/CustomFormParam.tsx
```diff
@@ -14,7 +14,14 @@ export default function CustomFormComponentLoader({
...
+    ? remoteComponentsUrl
+    : `${window.location.origin}/custom_components.js`;
```
antgamdia, line 24: If the url is unreachable, which error is thrown to the user? I'm guessing it's an internal 404 that simply results in this component not being loaded, but I want to double-check it.

antgamdia (approved):
Thanks for the changes, it is a nice improvement for supporting any dynamic or remotely defined component. 
It also adds some security concerns, but it is the ultimate responsibility of the cluster-admin to render components "from the wild". 

I'm +1ing, but I really like you, @aanthonyrizzo , to check what would happen in the cases I mention in the comments. If some extra header it's required, you could also add them as a chart value (e.g., which external domains are allowed in CSP/CORS, etc).

Once checked, I'm ok with merging it

dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/CustomFormParam.tsx
aanthonyrizzo: So funny enough I ran into this! I was doing some testing with a WDS that I would load some components from, I needed to add these headers to the WDS to allow the request from my kubeapps frontend
```
headers: {
      'Access-Control-Allow-Origin': '*',
      'Access-Control-Allow-Methods': 'GET, POST, PUT, DELETE, PATCH, OPTIONS',
      'Access-Control-Allow-Headers':
        'X-Requested-With, content-type, Authorization'
    }
```

dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/CustomFormParam.test.tsx
aanthonyrizzo: I will amend the test. Although it is assumed, an explicit test is a good idea 👍

chart/kubeapps/values.yaml
aanthonyrizzo: Im sorry u mentioned this last time as well, will generate when I adjust test

dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/CustomFormParam.tsx
aanthonyrizzo: Yes correct there is a 404 and `param` object that specified a custom component will not load and will generate an section with `Error` in plain text

aanthonyrizzo:
Let me know if there's anything else I need to do for this. Do you think that we will be able to include this PR in the n cat release?

absoludity:
> Let me know if there's anything else I need to do for this. Do you think that we will be able to include this PR in the n cat release?

Sorry for the delay. Antonio's been off but will be back today. I'll land it now based on his +1.

antgamdia:
Great, thanks for the double-check and Michael for merging it! I've been a week off (vacation time, finally! :P )

> Do you think that we will be able to include this PR in the n cat release? 

As it is already merged, definitely, it will be included in the next week release!

mecampbellsoup:
@antgamdia @absoludity will some git bot tag this PR with the release version in which it's included automatically so we can check here to know when it's been released? For some reason that is my recollection... 😄

antgamdia:
Hi! Currently, we are [manually triggering the releases](https://github.com/kubeapps/kubeapps/blob/master/docs/developer/release-process.md) with everything already included in the main branch (like your PR :P). We are planning to create `v2.3.4` really soon (today or tomorrow)  :smile: . It's a process that takes some time (tagging, ensuring everything works, create the official Bitnami images, etc...)
By the way, feel free to join one of our bi-weekly community meetings if you want (link in [Slack](https://kubernetes.slack.com/messages/kubeapps))

PR #3244, opened 2021-08-08
docs: add namespace flag for retrieve token example

hugocortes:
### Description of the change

Updates "Getting Started" docs retrieve token example to use `--namespace default` flag. 

### Benefits

When creating a serviceaccount in previous section, it's also done with `--namespace default` flag in case user is not in the `default` namespace context. By adding the namespace flag, users in different context can copy and paste the shell example otherwise they'll receive this error:
```shell
» kubectl get serviceaccount kubeapps-operator -o jsonpath='{range .secrets[*]}{.name}{"\n"}{end}' | grep kubeapps-operator-token                                                                                                          
Error from server (NotFound): serviceaccounts "kubeapps-operator" not found

```

### Applicable issues

n/a

### DIFF
--- a/docs/user/getting-started.md
+++ b/docs/user/getting-started.md
@@ -38,7 +38,7 @@ To retrieve the token,
 ### On Linux/macOS:
 
 ```bash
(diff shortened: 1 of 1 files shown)


antgamdia:
Hi @hugocortes, good gotcha, thanks for the contribution!
Let me retrigger the CI as it seems it's not reporting the green light and then I'll merge it :)

antgamdia (approved):

PR #3261, opened 2021-08-12
add GetInstalledPackageDetail() for fluxv2 plugin

gfichtenholt:


antgamdia:
Just using this PR for a question: do you know if the `GetInstalledPackageDetail` operation is including the resources created because of a release install?  (like in the current implementation?)

gfichtenholt:
> Just using this PR for a question: do you know if the `GetInstalledPackageDetail` operation is including the resources created because of a release install? (like in the current implementation?)

don't think so. There isn't anywhere to hang it yet in InstalledPackageDetail, 
Michael has a TODO in the spec
  // TODO: Install notes and resources are missing
  KubernetesResources map[string] strings
but its just a comment in the doc, not even in the proto files, much less implementation

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/release.go
```diff
@@ -185,3 +259,44 @@ func installedSummaryStatusFromUnstructured(unstructuredRelease map[string]inter
...
+func installedPackageAvailablePackageRefFromUnstructured(unstructuredRelease map[string]interface{}) (*corev1.AvailablePackageReference, error) {
+	repoName, found, err := unstructured.NestedString(unstructuredRelease, "spec", "chart", "spec", "sourceRef", "name")
+	if !found || err != nil {
+		return nil, status.Errorf(codes.Internal, "missing required field spec.chart.spec.sourceRef.name")
+	}
+	chartName, found, err := unstructured.NestedString(unstructuredRelease, "spec", "chart", "spec", "chart")
+	if !found || err != nil {
+		return nil, status.Errorf(codes.Internal, "missing required field spec.chart.spec.chart")
+	}
+	repoNamespace, _, _ := unstructured.NestedString(unstructuredRelease, "spec", "chart", "spec", "sourceRef", "namespace")
```
absoludity, line 288: Just wondering why the err is ignored here?

absoludity (approved):
LGTM, thanks Greg. Small comment in line.

absoludity:
> > Just using this PR for a question: do you know if the `GetInstalledPackageDetail` operation is including the resources created because of a release install? (like in the current implementation?)
> 
> don't think so. There isn't anywhere to hang it yet in InstalledPackageDetail,
> Michael has a TODO in the spec
> // TODO: Install notes and resources are missing
> KubernetesResources map[string] strings
> but its just a comment in the doc, not even in the proto files, much less implementation

Yep, as per the comment there, I'm keen that we *don't* stuff those into the `InstalledPackageDetail` but rather design a separate `GetK8sResourcesForInstalledPackage` call. Until then, I'm hoping we can continue with the existing API to get them? If that's a problem, I'll try to do that ASAP.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/release.go
gfichtenholt: my bad, will fix. Thanks for reviewing

Issue #3278, opened 2021-08-17
[kubeapps-apis] Implement support for `cluster` in packaging API


absoludity:
### Description:

Currently we've left support for `cluster` in the message context, but have not yet implemented the support. This is required for us to release, since we need to support the existing functionality.

Splitting off from the [Missing Operations comment on #3165](https://github.com/kubeapps/kubeapps/issues/3165#issuecomment-882944035) so we can focus the iteration.

PR #3280, opened 2021-08-18
Update server.KubernetesConfigGetter to support cluster arg

absoludity:
### Description of the change

Updates the KubernetesConfigGetter function alias to accept the cluster as an arg.

It uses this to return a config for the specified cluster, but defaults to the current behaviour of the cluster on which Kubeapps is running.

Tests that the correct config is created depending on the clusterConfig.

### Benefits

This is step 1 in being able to match the existing cluster support but with the new APIs server (for Helm).

### Applicable issues

  - Ref #3278

### Additional information

The following PR will add support for the Helm plugin to reply sanely for requests to other clusters, while leaving the other plugins using the default behaviour for now.


gfichtenholt (approved):
I don't see any problems. +1 if it works as expected
Thanks

PR #3283, opened 2021-08-19
Proposal to ensure that we always have both the package and app version available

absoludity:
### Description of the change

In [Antonio's comment](https://github.com/kubeapps/kubeapps/issues/3165#issuecomment-882943900) he mentioned that when working on the UX he was missing the latest app version in a couple of places where he needed it.

In some places we have just the package version, in others we're listing two separate fields for the package and app versions. In the response for the `GetAvailablePackageVersions` we use a combined `PackageAppVersion` message to contain both.

The proposal here is to pull out the `PackageAppVersion` message, and use that in all the places where we have a field for the package version so that we always include both the package and app version information without explicitly needing two fields.

Edit: After the conversation below, I've updated the PR with a mostly mechanical change to use the `PackageAppVersion` message as proposed.

### Benefits

We will have both available wherever we have a field for the package version.

### Applicable issues

  - Ref #3165

### Additional information

If there's agreement, I'll

 - [ ] update the design doc to match and 
 - [x] take care of updating both the helm and flux plugins to this.


gfichtenholt:
Just so I understand, in this PR you plan just to update the .proto file? Wouldn't that break the build and/or unit tests?

absoludity:
> Just so I understand, in this PR you plan just to update the .proto file?

No - if there was agreement, I'd go ahead and do the actual changes too as part of this branch. I just didn't want to go ahead and do those changes without first getting your thoughts. The proto file change just shows very cleanly what I'm proposing, that's all.

gfichtenholt:
> The proposal here is to pull out the PackageAppVersion message, and use that in all the places where we 
> have a field for the package version so that we always include both the package and app version 
> information without explicitly needing two fields.

Eh, you mean in responses only? Requests still have places where only pkg_version is used?
Also, what about the part where Antonio is asking 'add currentAppVersion to InstalledPacakgeDetail'. Or is that a separate PR?

absoludity:
> Eh, you mean in responses only? Requests still have places where only pkg_version is used?

Yes, I've only proposed changing the actual entities (which are returned in the response). We don't currently have any need for requesting an available package at a particular app version, but yes, I see no reason why not. Happy to update both `GetAvailablePackageDetailRequest` and `GetAvailablePackageVersionRequest` to use `PackageAppVersion` instead (even if for now we return unimplemented if the app version is specified? Let me know.

> Also, what about the part where Antonio is asking 'add currentAppVersion to InstalledPacakgeDetail'. Or is that a separate PR?

No - you're right. I'd missed that one. I've updated the proposal so that `InstalledPackageDetail` now uses a `PackageAppVersion` for the `current_version` field, and also added the `latest_version` and `latest_matching_version` fields to the detail (they were on the summary but are also needed on the detail).

Let me know if you think this proposal is a good way to solve the problem of the missing app versions, or if you have a different suggestion/proposal. Thanks.

absoludity:
> Also, I am looking at InstalledPackageSummary and it has
> 
> ```
> / CurrentVersion
>   //
>   // The version of the package which is currently installed.
>   PackageAppVersion current_version = 4;
> 
>   // CurrentAppVersion
>   //
>   // The version of the app included in the current package version.
>   string current_app_version = 5;
> ```
> 
> That seems wrong.

Yep, I saw the same when editing and removed the `current_app_version` which I'd missed.

> 
> One last thing is you might want to merge the latest upstream master and resolve conflicts in this proto file. Things have been moved around quite a bit (by me in a previous PR)

Yeah, if we decide to go ahead with this proposal, I'll do that (and renumber the fields etc.) when doing the implementation.

gfichtenholt:
> > Eh, you mean in responses only? Requests still have places where only pkg_version is used?
> 
> Yes, I've only proposed changing the actual entities (which are returned in the response). We don't currently have any need for requesting an available package at a particular app version, but yes, I see no reason why not. 

Nah, let's stick with what's being asked now. We can always iterate later

gfichtenholt:
a small favor to ask as you're editing this file. There are a few occurences that look like this

// Optional version reference for which ....
  string pkg_version = 2;
  
which makes it a bit confusing as we also have a VersionReference message type. So it will be cleaner if the comment just said 'version string' instead.

gfichtenholt:
I am ok with this proposal. Not sure if you need me to +1 now as there are more changes to follow? Let me know

absoludity:
> I am ok with this proposal. Not sure if you need me to +1 now as there are more changes to follow? Let me know

Great. No that's all I'm after at the moment, just before investing in the actual change. Thanks.

absoludity:
> I am ok with this proposal. Not sure if you need me to +1 now as there are more changes to follow? Let me know

I've updated now with the mostly mechanical changes to use the `PackageAppVersion` message.

absoludity:
Closing this in favour of the identical PR #3290 from a branch in the kubeapps repository (as Github doesn't allow me to stack branches on my own repo).

PR #3286, opened 2021-08-20
add PostInstallationNotes to fluxv2 GetInstalledPackageDetail

gfichtenholt:


cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/release.go
```diff
@@ -249,17 +255,42 @@ func (s *Server) installedPackageDetail(ctx context.Context, name types.Namespac
...
+func (s *Server) postInstallationNotesFromUnstructured(ctx context.Context, name types.NamespacedName, unstructuredRelease map[string]interface{}) (string, error) {
+	// post installation notes can only be retrieved via helm APIs, flux doesn't do it
+	// see discussion in https://cloud-native.slack.com/archives/CLAJ40HV3/p1629244025187100
+	if s.actionConfigGetter == nil {
+		return "", status.Errorf(codes.FailedPrecondition, "Server is not configured with actionConfigGetter")
+	}
+
+	helmReleaseName, found, err := unstructured.NestedString(unstructuredRelease, "spec", "ReleaseName")
+	// according to docs ReleaseName is optional and defaults to a composition of
+	// '[TargetNamespace-]Name'.
```
absoludity, line 272: In both this comment and the one below, I think you mean that it's optional when creating a `HelmRelease` via flux, but for the purposes here, won't it always be set? (As in, if it wasn't provided by the user when creating the `HelmRelease`, flux will have set it). In which case, I think we should returning with an error at this point if err != nil?

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/release.go
```diff
@@ -249,17 +255,42 @@ func (s *Server) installedPackageDetail(ctx context.Context, name types.Namespac
...
+	if !found || err != nil || helmReleaseName == "" {
+		targetNamespace, found, err := unstructured.NestedString(unstructuredRelease, "spec", "targetNamespace")
+		// according to docs targetNamespace is optional and defaults to the namespace of the HelmRelease
+		if !found || err != nil || targetNamespace == "" {
+			targetNamespace = name.Namespace
+		}
+		helmReleaseName = fmt.Sprintf("%s-%s", targetNamespace, name.Name)
```
absoludity, line 279: As above, I don't see why we'd need to construct the release name, unless you've found that what I mention above is incorrect and a `HelmRelease` CRD can exist without a `ReleaseName`, but that would be very odd?

absoludity (approved):
LGTM, though as inline, I'm not sure we need to calculate the helm release name. Worth checking on the CRD itself (I may well be wrong :) ).

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/release.go
gfichtenholt: No, that's not correct. Optional really means optional. As in unset when you do a get
https://fluxcd.io/docs/components/helm/helmreleases/
See example:
```
$ kubectl get hr/my-podinfo -o yaml
apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
metadata:
  creationTimestamp: "2021-08-18T01:01:39Z"
  finalizers:
  - finalizers.fluxcd.io
  generation: 1
  managedFields:
  - apiVersion: helm.toolkit.fluxcd.io/v2beta1
    fieldsType: FieldsV1
    fieldsV1:
      f:spec:
        .: {}
        f:chart:
          .: {}
          f:spec:
            .: {}
            f:chart: {}
            f:sourceRef:
              .: {}
              f:kind: {}
              f:name: {}
              f:namespace: {}
            f:version: {}
        f:install:
          .: {}
          f:createNamespace: {}
        f:interval: {}
        f:targetNamespace: {}
        f:values: {}
    manager: flux
    operation: Update
    time: "2021-08-18T01:01:39Z"
  - apiVersion: helm.toolkit.fluxcd.io/v2beta1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:finalizers:
          .: {}
          v:"finalizers.fluxcd.io": {}
      f:status:
        .: {}
        f:conditions: {}
        f:helmChart: {}
        f:lastAppliedRevision: {}
        f:lastAttemptedRevision: {}
        f:lastAttemptedValuesChecksum: {}
        f:lastReleaseRevision: {}
        f:observedGeneration: {}
    manager: helm-controller
    operation: Update
    time: "2021-08-19T21:32:51Z"
  name: my-podinfo
  namespace: default
  resourceVersion: "378861"
  selfLink: /apis/helm.toolkit.fluxcd.io/v2beta1/namespaces/default/helmreleases/my-podinfo
  uid: 054ff881-0ff6-43d3-ba81-8df43a780806
spec:
  chart:
    spec:
      chart: podinfo
      sourceRef:
        kind: HelmRepository
        name: podinfo
        namespace: default
      version: '*'
  install:
    createNamespace: true
  interval: 1m0s
  targetNamespace: test
  values:
    ui:
      color: '#24578c'
      message: what we do in the shadows
status:
  conditions:
  - lastTransitionTime: "2021-08-19T21:32:51Z"
    message: Release reconciliation succeeded
    reason: ReconciliationSucceeded
    status: "True"
    type: Ready
  - lastTransitionTime: "2021-08-18T01:01:41Z"
    message: Helm install succeeded
    reason: InstallSucceeded
    status: "True"
    type: Released
  helmChart: default/default-my-podinfo
  lastAppliedRevision: 6.0.0
  lastAttemptedRevision: 6.0.0
  lastAttemptedValuesChecksum: 5d9b9ed51dba5144c174a4f11d003c973a698f1f
  lastReleaseRevision: 1
  observedGeneration: 1
```

absoludity: Yep, ok. I mis-understood what they meant by "Defaults to ...." (in that, I thought it meant if you don't set it, it gets default to..., where as it seems they mean the actual name of the helm release will get the default). Thanks for pointing to the detail.

absoludity: I'm assuming you saw that I'd already +1'd it earlier anyway?

gfichtenholt: yes, thank you. just making minor last minute cleanup and its done

PR #3287, opened 2021-08-20
replicated fix previously made for create to upgrade as well

dlaloue-vmware:
### Description of the change

The Create operation was fixed in a previous commit to allow creation of apps on a remote cluster from a global repository.
The fix was not applied to the Upgrade operation and as such upgrading an app deployed on a remote cluster was failing.

The change is applying the same fix made for Create onto Upgrade as well.

### Benefits

Upgrade of remote applications will not be possible.

### Possible drawbacks

none

### Applicable issues

Issue 3213

### Additional information

N/A


absoludity (approved):
Great, thanks Dimitri. Glad it was so simple. It's a shame the tests there aren't written to be able to test the calls to the fn, but not worth changing here.

PR #3298, opened 2021-08-24
Add CreateInstalledPackage rpc and messages for initial Helm support.

absoludity:
### Description of the change

Adds the CreateInstalledPackage rpc call to the core API as well as the 3 plugins, so that they each have an automatic unimplemented version.

Also, while updating the design doc, just noticed that I'd missed renumbering fields on the InstalledPackageSummary.

The only relevant changes for review here are the four .proto files (core packages then each plugin). The rest are generated.

The next PR will add the implementation for the Helm version.

### Benefits

Support creating package installs with helm.

### Possible drawbacks

We've initially simplified the request based on the needs of helm, but the design doc still includes the other fields which we will expand as we add support for other plugins.


absoludity:
Setting as draft as I'll include other changes here.

absoludity:
Sorry, closing this as I need to stack a followup PR (for the implementation). Moved to https://github.com/kubeapps/kubeapps/pull/3299

PR #3302, opened 2021-08-24
Update the InstalledPackageReference.Identifier for Helm.

absoludity:
Updates the identifier to include the release version.

### Description of the change

As noted in the [missing fields comment](https://github.com/kubeapps/kubeapps/issues/3165#issuecomment-882943900), the dashboard needs to be able to present release versions for the rollback action.

Initially we thought we'd need to update the `InstalledPackageDetail` with a list of rollback versions, but this doesn't translate well to any other packaging systems (Flux allows rolling back, but not to a specific release version, just the previous one, so doesn't expose any release version, kapp_controller doesn't appear to have any concept of release revisions/versions yet).

After realising that the current dashboard simply [generates the list of rollback targets based on the current revision](https://github.com/kubeapps/kubeapps/blob/master/dashboard/src/components/AppView/AppControls/RollbackButton/RollbackDialog.tsx#L39-L42), we just need to ensure that the helm rollback action knows the current release version. And given that it's likely we'll follow suite and later update to simply allow rolling back (without specifying which release version, but just rolling back to the most recent successful one as flux does), I don't want to add it unnecessarily to the `InstalledPackageDetail`.

So, instead, this PR updates the `InstalledPackageReference` so that the plugin-specific identifier can include the release version. This is a good idea regardless since otherwise the helm plugin has no way to fetch anything other than the latest release version.

A query with an identifier without the release version will return the latest release version (with the correct identifier).

### Benefits

Two-fold:

* The dashboard can for now pull out the helm release version (though longer-term I'd suggest we support rolling back to the latest successful release version so the dashboard does not need this info at all),
* The helm plugin can now be used to fetch different release versions of a release.

### Possible drawbacks

I didn't want to change the proto here, so you'll notice in the tests that when I fetch a previous release version (that has been superseded), the status says unknown. We may want to add a more specific enum for superseded.

### Applicable issues

  - Ref #3165

### DIFF
--- a/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
+++ b/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
@@ -677,7 +677,7 @@ func installedPkgSummaryFromRelease(r *release.Release) *corev1.InstalledPackage
 			Context: &corev1.Context{
 				Namespace: r.Namespace,
 			},
-			Identifier: r.Name,
+			Identifier: fmt.Sprintf("%s/%d", r.Name, r.Version),
 		},
 		Name: r.Name,
 		PkgVersionReference: &corev1.VersionReference{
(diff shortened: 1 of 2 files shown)


antgamdia (approved):
Great! Tested in my branch and working :)

antgamdia:
> We may want to add a more specific enum for superseded.

Currently, this is my mapping in the UI, we're missing status for pending upgrade and pending rollback:

https://github.com/kubeapps/kubeapps/blob/55a137161b72529e8c1c17b7f3a10eab7ad88484/dashboard/src/components/AppView/AppControls/StatusAwareButton.tsx#L25-L33

absoludity:
So we agreed that for a temporary communication of the helm specific release revision, the custom details field that we'd originally planned but removed would be a better option than stuffing it into the identifier. I'll not land the PR, but go ahead and record, then afterwards we can update the PR and dashboard.

PR #3306, opened 2021-08-25
Prepare deps and chart for release

antgamdia:
### Description of the change

This PR is mostly executing the release process flow in terms of upgrading deps, but just the easiest ones. Also, and most importantly, it brings the kubeapps-apis component to foreground: the feature flag is enabled by default and the `values.yaml` contains everything needed to work.
In the next release, we could potentially remove the feature flag itself.

### Benefits

Everything prepared for the upcoming (intermediate) release!

### Possible drawbacks

N/A (an unnoticed incompatibly  perhaps?)

### Applicable issues

N/A

### Additional information

Still performing manual checks, but I wanted to trigger the CI anyway asap. 
Edit: well, some tricks are still required for the CI to work... it's a pity we have to deal with this state transitions every time haha


chart/kubeapps/values.yaml
```diff
@@ -1538,3 +1538,226 @@ postgresql:
...
+    ## @param kubeappsapis.service.port KubeappsAPIs service HTTP port
+    ##
+    port: 8080
+    ## @param kubeappsapis.service.annotations Additional custom annotations for KubeappsAPIs service
+    ##
+    annotations: {}
+## Redis&trade; chart configuration
+## ref: https://github.com/bitnami/charts/blob/master/bitnami/redis/values.yaml
+##
+redis:
```
absoludity, line 1749: Not sure why we want to promote the options for redis in the values.yaml of the release, rather than just leaving them in the development values.kubeappsapis.yaml (as you've done) for now?

absoludity (approved):

chart/kubeapps/values.yaml
antgamdia: Honestly, I didn't manually verify it but I assumed, given that many redis values are used in different places (notes.txt, chart.yaml for conditionally load redis, that I had to define the redis values in the chart even they are not used at all.
Indeed, we have to thoroughly review them when enabling it by default; I'm not sure the password is being set properly.

Another reason for doing so is that this way we can just use the values.yaml as is, without any more conflicts when syncing dev-prod charts.

Anyway, happy to spend some cycles in removing redis -though it would imply still having two separate chart versions; not sure if's the best option for us right now-. See what you think.

antgamdia:
A couple of extra changes after the +1, but we got CI green light, so I guess I can merge. These changes are:

- Sort the extra headers and proxypass directives (in some cases we were setting the header after the proxypass)
- Revert the CI e2e script change where we used the bitnami Kubeapps APIs img. It fails since it is not in the released chart, so for the next release we should be able to finally remove the TODOs, but not now.
- Updated the nginx-ingress controller used in dev, as we were using an old version with some deprecation warnings in the newer k8s versions.
- Updated the latest k8s release from 20 to 21; note we still use the 19 and 20 in GKE
- Updated GKE versions to 19 and 20 (according to https://cloud.google.com/kubernetes-engine/docs/release-notes)

Note, I'm merging since we [got the green light](https://app.circleci.com/pipelines/github/antgamdia/kubeapps/1048/workflows/e6446276-e3e5-4b22-8f56-fe91f2dca73c), but I later realized we should upgrade the GKE versions, but there are no relevant changes for merging (and I'd like to trigger the release soon :P)

PR #3321, opened 2021-08-28
Update to Go 1.17

antgamdia:
### Description of the change

This PR updates to Go 1.17, the main change is that the indirect dependencies have now their own block (https://golang.org/doc/go1.17#go-command).

### Benefits

Use the latest version

### Possible drawbacks

N/A (go doesn't break backward compatibility) 

### Applicable issues

N/A

### Additional information

The official (Bitnami) images should also be updated. Marking as draft PR in the meanwhile.

Edit: also getting this CI error:

```console
------
 > [builder 11/12] RUN --mount=type=cache,target=/go/pkg/mod     --mount=type=cache,target=/root/.cache/go-build     go build     -o /fluxv2-packages-v1alpha1-plugin.so -buildmode=plugin     ./cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/*.go:
#21 1.896 # github.com/cespare/xxhash/v2
#21 1.896 asm: xxhash_amd64.s:120: when dynamic linking, R15 is clobbered by a global variable access and is used here: 00092 (/go/pkg/mod/github.com/cespare/xxhash/v2@v2.1.1/xxhash_amd64.s:120)	ADDQ	R15, AX
#21 1.896 asm: assembly failed
------
executor failed running [/bin/sh -c go build     -o /fluxv2-packages-v1alpha1-plugin.so -buildmode=plugin     ./cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/*.go]: exit code: 2
make: *** [Makefile:26: kubeapps/kubeapps-apis] Error 1
```


antgamdia:
More updates here (at some point we might want to move to Go 1.17, especially for the new dependency format -explicit differentiation direct vs indirect ones, plus "_the performance improvement due to the new register-based calling convention. Your CPU-bound programs will magically get 5% faster when compiled with 1.17_")

Due to this register-based calling, when dynamically linking files and error popped up (fixed at https://github.com/golang/go/commit/c870e863298a44ead08e0b797d62ea74d7b7c971). However, one of the libraries, xxhash, still relied on the R15 registry (https://github.com/cespare/xxhash/issues/54). Fortunately, they released a fix for that https://github.com/cespare/xxhash/commit/b2091fed9204874217a72109cd119fbf9ca0ac75 and got released  in https://github.com/cespare/xxhash/releases/tag/v2.1.2

Checking if I can upgrade this dependency to the proper version

absoludity (approved):
Sweet, looks good to go!

PR #3326, opened 2021-08-30
Update further tests for UI using new API

absoludity:
### Description of the change

Goes through the list at #3165, updating the skipped tests (or in some cases, removing as they are no longer releavent).


dashboard/src/actions/apps.test.tsx
```diff
@@ -75,98 +75,6 @@ describe("fetches applications", () => {
     expect(store.getActions()).toEqual(expectedActions);
     expect(listAppsMock.mock.calls[0]).toBeUndefined();
   });
-
-  // TODO(agamez): Check the current implemented logic, now it's no longer required to
-  // perform additional calls for fetching the latest versions. Perhaps delete some tests here, not sure yet.
```
absoludity, line 80: Correct - deleted as they are all ensuring that *after* the installed apps are fetched, the available versions for those apps are also fetched to present things like whether updates are available. This info is all contained in the installed packages summaries.

dashboard/src/components/AppList/AppListItem.test.tsx
```diff
@@ -17,6 +17,7 @@ import AppListItem, { IAppListItemProps } from "./AppListItem";
 const defaultProps = {
   app: {
     name: "foo",
+    pkgDisplayName: "foo",
```
absoludity, line 20: All that was missing was this, since it's used to display the name in the span (it was blank otherwise and the span was not found).

dashboard/src/components/Catalog/Catalog.test.tsx
```diff
@@ -88,7 +88,7 @@ const defaultState = {
   charts: defaultChartState,
   operators: { csvs: [] } as Partial<IOperatorsState>,
   repos: { repos: [] } as Partial<IAppRepositoryState>,
-  config: { kubeappsCluster: "default", kubeappsNamespace: "kubeapps" } as IConfigState,
+  config: { kubeappsCluster: defaultProps.cluster, kubeappsNamespace: defaultProps.kubeappsNamespace } as IConfigState,
```
absoludity, line 91: Tests were depending on "default-cluster", but getting "default" because of this config difference.

dashboard/src/components/Catalog/Catalog.test.tsx
```diff
@@ -310,6 +310,18 @@ describe("filters by the searched item", () => {
...
+
+  beforeEach(() => {
+    spyOnUseDispatch = jest.spyOn(ReactRedux, "useDispatch").mockReturnValue(mockDispatch);
+  });
+
+  afterEach(() => {
+    spyOnUseDispatch.mockRestore();
+    mockDispatch.mockReset();
+  });
+
```
absoludity, line 324: This was common for a few components which you'd converted to functions: they no longer had a `dispatch` prop, so we have to spy on the one used in the new function (via `useDispatch`).

dashboard/src/components/UpgradeForm/UpgradeForm.test.tsx
```diff
@@ -96,14 +98,14 @@ const populatedProps = {
     appVersion: testVersion.appVersion,
     readme: "readme",
     readmeError: undefined,
-    values: "values",
+    values: "initial: values",
```
absoludity, line 101: The string value used for the `values` field needs to be something that can be parsed as a yaml hash or array. So I've updated it here to `initial: values`, which has also shown that these will be included in upgrade tests below.

dashboard/src/components/UpgradeForm/UpgradeForm.test.tsx
```diff
@@ -350,21 +346,23 @@ describe("when receiving new props", () => {
...
 `,
-      result: `foo:
-  - foo1: 
-    bar1: value1
-  - foo2: 
-    bar2: value2
-`,
+      result: [
+        `foo:`,
+        `  - foo1: `,
```
absoludity, line 360: Apparently the space is in the results that we're comparing, but our (or my) editor removes trailing spaces if I format it as a string as previously. This just allows including the trailing spaces (could also just do as a normal string "foo: \n  - foo1: \n...")

dashboard/src/shared/App.test.ts
```diff
@@ -37,34 +37,6 @@ describe("App", () => {
     jest.resetAllMocks();
   });
 
-  describe("listApps", () => {
```
absoludity, line 40: The functionality that this was testing (different endpoint if namespace was present) is no longer relevant at all (since the apis server just receives it either way at the same endpoint now).

dashboard/src/components/UpgradeForm/UpgradeForm.test.tsx
antgamdia: I just ended up temporarily disabling this vscode feature to save this test :P, but your solution is much better

antgamdia (approved):
Thanks a lot for the changes! I was kinda exhausted last week fixing so many test suites.

Good to see you came up with the `expect(mockDispatch)...`. Reading the `react-router` docs, it seemed like I should inspect the `history` object... 
I've run prettier and commit the changes here directly.

PR #3331, opened 2021-08-30
Fix development ingress to be valid with ingress-nginx 1.0.0

absoludity:
### Description of the change

Tried to setup a dev env and it was not working because the ingress-nginx controller was claiming that the kubeaps ingress was invalid.

Turned out to be related to the update to use 1.0.0 of the ingress-nginx controller in our dev env, which requires the class to be set for the ingress to pass validation.

### Benefits

Dev env works again.

### Applicable issues

  - fixes #


antgamdia (approved):
Great, forgot yo check the local setup files. Sorry! Thanks for adding it

PR #3333, opened 2021-08-31
Check ci e2e tests with new UI

absoludity:
### Applicable issues

  - fixes #

### DIFF
--- a/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
+++ b/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
@@ -336,7 +336,7 @@ func (s *Server) GetAvailablePackageDetail(ctx context.Context, request *corev1.
 
 	// Currently we support available packages on the kubeapps cluster only.
 	if cluster != "" && cluster != s.globalPackagingCluster {
-		return nil, status.Errorf(codes.InvalidArgument, "Requests for available packages on clusters other than %q not supported.", s.globalPackagingCluster)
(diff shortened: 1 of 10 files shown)


antgamdia (approved):
Mmm interesting the increase in the expiration. I assumed the error while retrieving the chart was caused because it was not available, but a key expiration also make sense

antgamdia:
At some point we have to think about moving to https://playwright.dev (or any other) more seriously... current e2e tests are such a pain sometimes :disappointed:

absoludity:
> Mmm interesting the increase in the expiration. I assumed the error while retrieving the chart was caused because it was not available, but a key expiration also make sense

Oh - that change is unrelated to the e2e tests. Rather, it's because when setting up a dev environment, if dex rotates the signing keys (for the tokens), my kind cluster then starts 401ing all requests (k8s api-server doesn't seem to check and update from the oidc .well-known-configuration).

absoludity:
So there's one remaining failing test here, that I can reproduce locally. It's 08-rollback, which upgrades a deployment setting the number of replicas to 2. Before clicking upgrade, the test checks the "Changes" tab for the new setting ("replicaCount: 2"), but it doesn't find this because:

08-rollback-0

the changes tab appears to think *everything* has changed and doesn't find the "replicaCount: 2" text at all. I suspect the Changes tab component isn't getting the previous release's recorded values, but haven't checked yet. Right now, since it's a minor issue, I've just commented out those lines with a TODO.

antgamdia:
As I said offline, THANKS for making the test pass. I'm gonna work on a fix for this issue, so that I can uncomment these lines. 
Merging to reduce the number of ongoing branches!

PR #3337, opened 2021-09-01
Further fluxv2 plugin features: CreateInstalledPackage initial implementation

gfichtenholt:
A few things in this PR:
1) initial implementation and unit tests of CreateInstalledPackage for flux v2 plug-in. Much more is expected to come
2) small fix to .proto files to generate correct swagger for (1)
3) check whether flux CRDs have been installed during startup to fail gracefully if not
4) update flux version to latest 0.17.0


chart/kubeapps/templates/kubeappsapis/rbac.yaml
```diff
@@ -26,6 +27,11 @@ rules:
   - apiGroups: [""]
     resources: ["secrets"]
     verbs: ["get", "list"]
+  # needed by fluxv2 plug-in to check whether flux CRDs have been installed
+  # TODO (gfichtenholt) possibly surround this with "if fluxv2 plugin is enabled"
```
absoludity, line 31: Yes, but not as a condition on this ClusterRole... this cluster role is currently intended to be used in development only (ie. the binding below is only done when `unsafeUseDemoSA` is set to true), hence the name being `kubeapps-apis-dev` above. The helm (direct) and kapp-controller plugins are able to use the users' credential for all interactions when requests come from the client.

But in your case, because you're caching data and interacting with the cluster outside of user requests, you need to ensure that extra RBAC required by your plugin is available to the service account. So longer term (not for this PR), I'd separate it out into a different cluster-role (perhaps in a different file) and put the condition there.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/cache.go
```diff
@@ -160,6 +162,33 @@ func newCacheWithRedisClient(config cacheConfig, redisCli *redis.Client, waitGro
 	return &c, nil
 }
 
+func (c NamespacedResourceWatcherCache) isGvrValid() error {
+	if c.config.gvr.Empty() {
+		return status.Errorf(codes.FailedPrecondition, "server configured with empty GVR")
```
absoludity, line 167: Not sure why grpc status errors are used in this function... the call-site isn't returning a response I don't think? If it's just on startup, normal errors are fine (though I don't see any issue using grpc ones).

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/release.go
```diff
@@ -312,6 +315,51 @@ func (s *Server) helmReleaseFromUnstructured(ctx context.Context, name types.Nam
 	return release, nil
 }
 
+func (s *Server) newRelease(ctx context.Context, packageRef *corev1.AvailablePackageReference, targetName types.NamespacedName) (*corev1.InstalledPackageReference, error) {
+	// HACK: just for now assume HelmRelease CRD will live in the kubeapps namespace
+	kubeappsNamespace := os.Getenv("POD_NAMESPACE")
+	if kubeappsNamespace == "" {
+		// unit tests don't have this env var set
```
absoludity, line 322: You can set an env var for the duration of a test - see https://stackoverflow.com/a/66061355 (nice, I didn't know about the new `testing.T.SetEnv`)

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/release.go
```diff
@@ -312,6 +315,51 @@ func (s *Server) helmReleaseFromUnstructured(ctx context.Context, name types.Nam
...
+
+	name, err := namespacedName(newRelease.Object)
+	if err != nil {
+		return nil, err
+	}
+	return &corev1.InstalledPackageReference{
+		Context:    &corev1.Context{Namespace: name.Namespace},
+		Identifier: name.Name,
+	}, nil
+}
```
absoludity, line 361: Nice - that function is simpler than I expected it to be, due to the work you've done previously for `fetchChartFromCache` etc. :)

absoludity (approved):
Great!

chart/kubeapps/templates/kubeappsapis/rbac.yaml
gfichtenholt: +1

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/cache.go
gfichtenholt: it had something to do with unit tests, I think. Not important enough now to go into. Will change if/when needed

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/release.go
gfichtenholt: oh, cool. Will use that, thanks

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/release.go
gfichtenholt: thanks for the review.

PR #3346, opened 2021-09-04
feat(UI): Add ability to load custom App Views

### LINKED ISSUE #3273: Feature Request: Allow custom application views
### Ask
The current Kubeapps UI provides a very nice UI for managing standard helm deployments. However there are certain deployments where the interface and functionality can leave something to be desired. To address this I think a good step to take would be allowing developers to load their own UIs for specific helm charts (similar to what we do for the form components).

### Design/Implementation Proposal
The implementation for this should be pretty similar to the way that we load custom components except we also need to design a way to actually determine which charts to load the custom UIs.

1. Add a variable to the configmap that the frontend will use to identify which chart/s that a custom UI should be used for. I think an array would work well for this so that multiple charts can be defined. Something like:
```
{
  "kubeappsCluster": "default",
  "kubeappsNamespace": "kubeapps",
  "appVersion": "DEVEL",
  "authProxyEnabled": false,
  "authProxySkipLoginPage": false,
  "oauthLoginURI": "/oauth2/start",
  "oauthLogoutURI": "/oauth2/sign_out",
  "clusters": ["default"],
  "theme": "light",
  "remoteComponentsUrl": "",
  "customAppViews": ["example-chart-1", "example-chart-2"],
}
```
2. In the `<AppView/>` we can just add a logic statement that checks if `app.chart.metadata.name` is included in `config.customAppViews`. Then we can follow the same pattern thats used in `<CustomFormParam />` and load the UI from a separate bundle.

One thing to consider. If someone is using the custom components feature, should it use the same bundle or a completely different one?

And as before, I'm willing to take ownership of this feature :)

aanthonyrizzo:
### Description of the change

This PR adds the ability to load in custom AppViews following the same pattern as the custom form components. It will allow developers to create custom views for specific helm charts. It uses the customAppViews array to logically load in the custom view.

### Benefits

Gives developers more power and freedom w/ releases. A good use case is a deployment that leverages helm and possibly a custom CRD to manage certain resources in a deployment.

### Possible drawbacks

Will require development knowledge and a little work to build the page.

### Applicable issues

  - fixes #3273


dashboard/src/components/AppView/AppView.tsx
```diff
@@ -191,6 +193,16 @@ export default function AppView() {
     resourceRefs;
   const revision = app?.revision ?? 0;
   const icon = appDetails?.iconUrl ?? placeholder;
+
+  // If the package identifier matches the current list of loaded customAppViews,
+  // then load the custom view from external bundle instead of the default one.
+  if (
+    app?.availablePackageRef?.identifier &&
+    customAppViews.includes(app.availablePackageRef.identifier)
```
antgamdia, line 201: I'm doubtful about the proper way to match a given package to its associated view. Not sure if the identifier is the best way.

antgamdia (approved):
@aanthonyrizzo , I've jumped in myself in this PR, as we released just 6 days ago the UI refactor, and there were some required changes to adapt your PR to our new backend API (tl;dr: now we use `package` instead of  `chart` as we are about to support multiple packaging formats)

As for me, +1 to this PR and, as always, thanks for your contributions!!
Edit: CI is not passing, we have to figure out what's happening.

Let's see what @absoludity also thinks about it before merging.

dashboard/src/components/AppView/AppView.tsx
absoludity: +1. Yes, sorry for not providing input earlier (we had a crazy couple of weeks), but I'm a little concerned about adding this in a way that won't be compatible when we're enabling support for pluggable packaging formats. At the very least, we'd need the plugin being used (`helm`) as well as the (plugin-specific) identifier proposed above.

That is, to future-proof this, we need to ensure the `customAppViews` data isn't just string, but structured data with a couple of fields: `{ "plugin": "helm", "name": "apache" }`. Note that this would be the `apache` chart from any repository. That would also enable later adding extra fields for more granular support (such as `{ "plugin": "helm", "name": "apache", "repository": "bitnami", "namespace": "kubeapps"}` etc.)

Is that an option? (for now, each custom app view defining both the `plugin` (just `helm`) and the name?

Thanks for the updates based on the recent changes Antonio, and thanks Anthony for the contributions.

dashboard/src/components/AppView/CustomAppView/CustomAppView.tsx
```diff
@@ -0,0 +1,28 @@
...
+}
+
+function CustomAppView({ resourceRefs, app }: ICustomAppViewProps) {
+  const {
+    config: { remoteComponentsUrl },
+  } = useSelector((state: IStoreState) => state);
+
+  const url = remoteComponentsUrl
+    ? remoteComponentsUrl
+    : `${window.location.origin}/custom_components.js`;
```
absoludity, line 20: The naming here implies that the remote URL (and the default local url) defines custom component*s* (plural), so the bundle defines multiple custom components to be used by different parts of the code (app views, forms for deployment etc.), but I can't see anything being passed to the CustomComponent below to identify a specific custom component from the bundle? What am I missing?

dashboard/src/components/AppView/AppView.tsx
aanthonyrizzo: With the new changes and functionality that you guys want to add this is a good point. At the moment I'm getting antiquated w/ some of the interface changes and I'll figure out a way to target more specific deployments

dashboard/src/components/AppView/CustomAppView/CustomAppView.tsx
aanthonyrizzo: Yeah I'm glad you pointed this out. I definitely need to add some sort of identifier to make it easier to determine what to render.

aanthonyrizzo:
@antgamdia I might need some help w/ the CI. https://app.circleci.com/pipelines/github/kubeapps/kubeapps/4203/workflows/0092154f-6d45-48d6-9edc-14f7c004ec98/jobs/57631?invite=true&invite=true#step-114-1096 I seem to have somehow broken the e2e tests (and I haven't quite figured out how yet). Also seems that the script is exiting w/ a `1` but circle CI is receiving a `0`

antgamdia:
> @antgamdia I might need some help w/ the CI. https://app.circleci.com/pipelines/github/kubeapps/kubeapps/4203/workflows/0092154f-6d45-48d6-9edc-14f7c004ec98/jobs/57631?invite=true&invite=true#step-114-1096 I seem to have somehow broken the e2e tests (and I haven't quite figured out how yet). Also seems that the script is exiting w/ a `1` but circle CI is receiving a `0` (not sure if that is by design)

Hi! I haven't had a chance to pull the changes and see what's happening, but certainly, the changes in the PR are breaking the CI. 

The CI is launching Kubeapps not with a simple demo token but with an OIDC setup. According to CircleCI, it is failing in the page load, so probably something is preventing Kubeapps to load. Perhaps it is a missing default value somewhere that is breaking things, but just supposing. I hope to have more time tomorrow to have a closer look at it.
Anyway, I'm manually re-triggering CI just in case :P

```console
Rolls back an application

    TimeoutError: Navigation timeout of 30000 ms exceeded

      41 |     await page.goto(getUrl(uri));
      42 |     if (isOIDC === "true") {
    > 43 |       await page.waitForNavigation({ waitUntil: "domcontentloaded" });
         |       ^
      44 |       await expect(page).toClick("cds-button", {
      45 |         text: "Login via OIDC Provider",
      46 |       });
```

chart/kubeapps/templates/dashboard/configmap.yaml
```diff
@@ -77,5 +77,6 @@ data:
       "featureFlags": {{ .Values.featureFlags | toJson }},
       "clusters": {{ template "kubeapps.clusterNames" . }},
       "theme": "{{ .Values.dashboard.defaultTheme }}",
-      "remoteComponentsUrl": "{{ .Values.dashboard.remoteComponentsUrl }}"
+      "remoteComponentsUrl": "{{ .Values.dashboard.remoteComponentsUrl }}",
+      "customAppViews": {{ .Values.dashboard.customAppViews | toJson }},
```
absoludity, line 81: There's a trailing comma that was added on the lastline above, which I assume is resulting in a json parsing error when the dashboard loads the config, but haven't verified.

absoludity (commented):
> @antgamdia I might need some help w/ the CI. https://app.circleci.com/pipelines/github/kubeapps/kubeapps/4203/workflows/0092154f-6d45-48d6-9edc-14f7c004ec98/jobs/57631?invite=true&invite=true#step-114-1096 I seem to have somehow broken the e2e tests (and I haven't quite figured out how yet). Also seems that the script is exiting w/ a `1` but circle CI is receiving a `0` (not sure if that is by design)

See inline, but my guess is that the `config.json` is invalid. Odd though, you should see the same error when deploying the chart locally?

antgamdia:
Yesterday we found an issue in one of our CI tests (https://github.com/kubeapps/kubeapps/pull/3377#issuecomment-920310967), I'd suggest merging the latest changes in the main branch to get this fix in yours. Sorry!

aanthonyrizzo:
> Yesterday we found an issue in one of our CI tests ([#3377 (comment)](https://github.com/kubeapps/kubeapps/pull/3377#issuecomment-920310967)), I'd suggest merging the latest changes in the main branch to get this fix in yours. Sorry!

Ah ok cool, Ill rebase w/ master, thanks

aanthonyrizzo:
Thank for you fixing the tests @antgamdia! Is there anything else that needs to be addressed here?

antgamdia:
Anytime! We are having some bad days with our CI :P 

Before merging, as I discussed with Michael offline, we would really like to have a working example of this remote app view. This way, we can add it to de documentation and make it easier for other people to build their own custom views.
Also, it'd be useful for us to verify the PR + our code + our future changes still work.

So, @aanthonyrizzo, would you mind adding a small working example of this remote app view? 
Thanks!

absoludity:
> Okay I can def get that for you. Im assuming just a very basic view that displays the props and how to use the handlers for the actions will be enough? And what is the timeline for the 7.4.1 release? Id like to get this included if possible so I want to to budget my time accordingly. I will try to make the bi-weekly meeting but it is 5:30AM my time disappointed

Great, thanks @aanthonyrizzo . The main thing from my pov is that it's easy for us (the reviewers) to test the functionality ourselves before landing, but it'd also be useful for others wanting to try it out. I'll take another look at the code (sorry ,didn't yet check since you pushed the updates - crazy time).

dashboard/src/components/AppView/CustomAppView/CustomAppView.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+it("should render the remote component with the URL if set in the config", () => {
+  const wrapper = mountWrapper(
+    getStore({
+      config: { remoteComponentsUrl: "www.thiswebsite.com" },
+    }),
+    <CustomAppView {...defaultProps} />,
+  );
+  expect(wrapper.find(CustomComponent).prop("url")).toBe("www.thiswebsite.com");
+});
```
absoludity, line 86: Great.

dashboard/src/components/AppView/CustomAppView/CustomAppView.tsx
absoludity: It's still not clear to me how the custom component below, which is given only the url to the bundle of remote component, will be able to identify and use the specific remote component from the bundle. Or I've completely misunderstood what the intent was. That's why I'm keen to see an example bundle with multiple components, to understand how you're planning to use multiple remote components for different packages/charts, before we land this.

Looking at the README at https://github.com/Paciolan/remote-component makes it look like it's a single remote component per url, but I've not dug into the documentation.

Given that we'll soon be requiring a way to load and use multiple components from a remote bundle, I'm keen to have one solution here that works for the future rather than one solution here that works for a single remote component only, that we need to support while we add a separate one later (or have to break this support when updating).

Does that make sense?

dashboard/src/components/AppView/CustomAppView/CustomAppView.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+        apiVersion: "v1",
+        kind: "Service",
+        plural: "services",
+        namespaced: true,
+        name: "ssh-server-example",
+        namespace: "demo-namespace",
+      },
+    ],
+    secrets: [],
+  } as unknown as IAppViewResourceRefs,
```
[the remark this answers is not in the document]
aanthonyrizzo: Ill look into this. I cannot remember if there is a reason for this but Ill look. It might have been a weird case where it wasn't similar enough to the defined object.

dashboard/src/components/AppView/CustomAppView/CustomAppView.tsx
aanthonyrizzo: Ahhh okay I did not understand your concern at first but now I think I understand. We have all of our components compiled into 1 bundle, this includes all of the custom form components that we currently use in production today AND the custom app views that I've been developing while working on this feature. We use a logical case to determine what component we want to render. The entry point looks like:
```
import React from 'react';
import PropTypes from 'prop-types';
import DeploymentFormRenderer from './DeploymentForm/DeploymentFormRenderer';
import AppViewRenderer from './AppView/AppViewRenderer';

const componentIds = {
  AppViewRenderer,
  DeploymentFormRenderer,
};

const LocalComponent = props => {
  const Component = componentIds[props.componentType || 'DeploymentFormRenderer'];
  return <Component {...props} />;
};

LocalComponent.propTypes = {
  componentType: PropTypes.string,
};

export default LocalComponent;
```
And then in each renderer we determine what to render based on the props that it receives.

Keeping and rendering all components from a single bundle is possible and I think the cleanest/best way to do it. I want to add a `componentType` prop to the custom form components as well but I was going to do that in a separate PR to prevent scope creep.

absoludity (approved):
Excellent, thanks for the extra info. It'd be great to include in the example how you're choosing the specific component as you did in your comment, but fine as is.

dashboard/src/components/AppView/CustomAppView/CustomAppView.test.tsx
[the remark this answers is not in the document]
antgamdia: In the past, I've also used `Partial<MyType>`  (e.g., `as Partial<IAppViewResourceRefs>`) when I just wanted to define a subset of the properties of a type for a specific instance. 
More info: https://www.typescriptlang.org/docs/handbook/utility-types.html

antgamdia (approved):
+1 to this, but I do agree with Michael in delegating the decision to the bundle, which could be more scalable and useful for the rest of the views in the future.
Again, thanks a lot for your contributions!

> . will try to make the bi-weekly meeting but it is 5:30AM my time

wrt that, I guess we can arrange another time it fits both of us, we are happy to chat about Kubeapps and better understand your needs (cc'ing our EM, @ppbaena )

dashboard/src/components/AppView/CustomAppView/CustomAppView.test.tsx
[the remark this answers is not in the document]
aanthonyrizzo: The TS error is kind of annoying since I straight up copied the object and I still got the error. I will try this solution instead of the `unknown`

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -121,7 +122,7 @@ function parseResources(
   return result;
 }
 
-interface IRouteParams {
+export interface IRouteParams {
   cluster: string;
   namespace: string;
   releaseName: string;
```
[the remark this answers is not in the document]
aanthonyrizzo: Thats a really good question. Im not sure what is best practice in regards to exporting interfaces from components vs keeping their definitions separate.

dashboard/src/components/AppView/CustomAppView/CustomAppView.test.tsx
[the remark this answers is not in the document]
aanthonyrizzo: So I tried to leverage `Partial<>` and the compiler still did not like it so I left is as is :(

dashboard/src/components/AppView/CustomAppView/CustomAppView.tsx
aanthonyrizzo: > Excellent - thanks Anthony, that's exactly what I was after. So the componentType prop is used in the developer provided code to return the component. So this means in effect that a class of views (AppView or DeploymentForm) can have a specific component. And as you say, the developer can do what they want here, so if you want a specific component for a specific chart, you can update the component selection to be dependent not only on the componentType but also the package identifier etc.

Yup! Exactly

> Which makes me then wonder, why do we need the customAppViews config at all... why not leave the decision wholly within the bundle? That is, what if you want to use your custom AppView for all packages? Currently the decision to use a custom app view is in both places: in the Kubeapps code depending on customAppViews, and also in the user provided bundle to decide which component to use. If a custom bundle had been provided, could we instead always check whether the bundle wants to provide the component and just default to the Kubeapps component when not? This way, the same code could be used in all parts of the UX (eg. in addition to AppView, we'd have RepositoryView, AppUpgrade, and for each, the bundle decides whether to return the default component or its own.

I decided to only have specific charts load the AppView for the time being since a developer might want to only inject these for certain deployments that require it. In my use case for example we leverage kubevirt to run VMs on our Kubernetes Cluster. The issue that we run into a lot is the delineation that they use for their resources (VirtualMachine and VirtualMachineInstance). We wanted to add information about the both resources. 

And I think there are some improvements that we can make for this sort of "plugin" architecture in the future to maybe make it simpler and the experience more streamlined. And I'm sure ill be able to come up with some cool ideas and places I will want to add some custom code :)

aanthonyrizzo:
> +1 to this, but I do agree with Michael in delegating the decision to the bundle, which could be more scalable and useful for the rest of the views in the future.
> Again, thanks a lot for your contributions!
> 
> > . will try to make the bi-weekly meeting but it is 5:30AM my time
> 
> wrt that, I guess we can arrange another time it fits both of us, we are happy to chat about Kubeapps and better understand your needs (cc'ing our EM, @ppbaena )

Yeah for sure! I think we plan on leveraging kubeapps for the foreseeable future and I'd love to show you some of the things we do with it as well

absoludity:
Cool - I'll land when CI passes then. Thanks again @aanthonyrizzo

aanthonyrizzo:
> Cool - I'll land when CI passes then. Thanks again @aanthonyrizzo

I have been really good at breaking these tests. I rebased master and pushed some docs changes and broke it again lol

absoludity:
> > Cool - I'll land when CI passes then. Thanks again @aanthonyrizzo
> 
> I have been really good at breaking these tests. I rebased master and pushed some docs changes and broke it again lol

Heh, yes - it's frustrating. I'd really like to prioritise improving the transparency and debugability of the CI tests (moving away from puppeteer in particular).

mecampbellsoup:
Good job Anthony!

PR #3348, opened 2021-09-06
Fixes #3347 by not assuming semver for app version.

### LINKED ISSUE #3347: Invalid Version: latest
Hello

Im using kubeapps 2.4.0 and im getting this error if the docker image tag is "latest":
```
Invalid Version: latest
```

Screenshot :

ksnip_20210905-040944

absoludity:
### Description of the change

As per the [Helm docs for the Chart.yaml](https://helm.sh/docs/topics/charts/#the-chartyaml-file), the appVersion field is not required to be semver at all.

I wonder whether we should assume this for package version as well. For Helm it's safe, but we can't base our decisions on Helm. Furthermore, it's already enough that the backend has indicated that the latest version is different to the current version, so I can't see why we need the semver check here in the frontend even for the package version.

Let me know and I'll update to remove it there also.

### Applicable issues

  - fixes #3347

### DIFF
--- a/dashboard/src/components/AppList/AppListItem.test.tsx
+++ b/dashboard/src/components/AppList/AppListItem.test.tsx
@@ -67,7 +67,7 @@ it("should add a tooltip with the chart update available", () => {
   expect(tooltip.text()).toBe("A new package version is available: 1.1.0");
 });
 
-it("should add a second label with the app update available", () => {
+it("should add a tooltip with the app update available", () => {
   const props = {
     ...defaultProps,
     app: {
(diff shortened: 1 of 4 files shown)


antgamdia (approved):
Awesome! Thanks for the fix. Yep, we assumed appVersion should be semver... 

+1, but please add the fix also in this other view (`ChartUpdateInfo`):

https://github.com/kubeapps/kubeapps/blob/95064092e372a08002f8d8e92934e9796b9e25d9/dashboard/src/components/AppView/ChartInfo/ChartUpdateInfo.tsx#L16-L26

absoludity:
> +1, but please add the fix also in this other view (`ChartUpdateInfo`):
> 

Done. Also, did you see my comment above:

> I wonder whether we should assume this for package version as well. For Helm it's safe, but we can't base our decisions on Helm. Furthermore, it's already enough that the backend has indicated that the latest version is different to the current version, so I can't see why we need the semver check here in the frontend even for the package version.

?

antgamdia:
Ooops, I missed it, sorry!

 > I wonder whether we should assume this for package version as well. For Helm it's safe, but we can't base our decisions on Helm. Furthermore, it's already enough that the backend has indicated that the latest version is different to the current version, so I can't see why we need the semver check here in the frontend even for the package version.

Mmm, good point here... yep, it makes sense. Given that we are, actually, passing the `lastest_version` as part of the response, we could just rely on `version != latest_version` for this check. Besides, it is less Helm-centric, which is great. I'll send a PR fixing that. Thanks for pointing it out.

PR #3362, opened 2021-09-07
Update chart.RegistrySecretsPerDomain to depend on kube clientset only.

absoludity:
### Description of the change

This is a spin-off change for supporting CreateInstalledPackage via the new kubeapps apis. I don't want the new kubeapps-apis to need to deal with the `kube.handler` abstraction, since our plugins don't even see the user token, nor should they need to know about them.

### Benefits

The RegistrySecretsPerDomain function depends on less code.

### Possible drawbacks

None known.

### Applicable issues

  - ref #3146


antgamdia (approved):
Neat! Thanks for the change, happy to see we are just passing the ctx instead instead of manually pass the cluster, token etc...

Issue #3368, opened 2021-09-07
Unable to use Kubeapps under a subpath


antgamdia:
### Description:

Eventually, we get this odd error in the `GetInstalledPackagesSummaries` operation:

![image](https://user-images.githubusercontent.com/11535726/132399757-60e87dfc-bb6c-4882-ac1f-42d676e21057.png)

### Steps to reproduce the issue:

1. Install Kubeapps under a sub-path, for example, using the local dev environment, with:

```diff
--- a/docs/user/manifests/kubeapps-local-dev-values.yaml
+++ b/docs/user/manifests/kubeapps-local-dev-values.yaml
@@ -18,6 +18,8 @@ kubeappsapis:
 ingress:
   enabled: true
   hostname: localhost
+  path: "/kubeapps(/|$)(.*)"
+  pathType: Prefix
   tls: true
   extraTls:
   - hosts:
@@ -27,3 +29,5 @@ ingress:
     nginx.ingress.kubernetes.io/proxy-read-timeout: "600"
     # Required for ingress-nginx 1.0.0 for a valid ingress.
     kubernetes.io/ingress.class: nginx
+    nginx.ingress.kubernetes.io/configuration-snippet: "rewrite ^(/kubeapps)$ $1/ redirect;"
+    nginx.ingress.kubernetes.io/rewrite-target: "/$2"
```

2. Login and try to view applications

### Describe the results you received:

`Response closed without grpc-status (Headers only) in GetInstaledPackageSummaries`.

Specifically, the grpc request to the APIs server does not include the prefix:
![GetInstalledPackageSummaries-no-prefix](https://user-images.githubusercontent.com/497518/132426382-e2aefe4a-aca9-4938-bb2d-8b795eaa0d3c.png)

### Describe the results you expected:

The grpc request should be prefixed just like all other requests.

### Additional information you deem important (e.g. issue happens only occasionally):

Slack thread discussing this issue: https://kubernetes.slack.com/archives/C9D3TSUG4/p1631011129013300

### Version of Helm, Kubeapps and Kubernetes:

Kubeapps  2.4.0

absoludity:
Actually, this one is reproducible, but isn't the same cause as the one we saw during release. The cause here is that there's a subpath is being used for the Kubeapps appliication, while the new client in the dashboard uses an absolute url `/apis`. You can verify that with a local kind cluster that it doesn't work.

So the request is never even getting to the apis server at all. I'm fixing that (one-liner).

PR #3369, opened 2021-09-08
Ensure relative path is used for kubeappsapis so any subpath is included.

absoludity:
### Description of the change

Updates the kubeappsapis URL in the dashboard to be relative like all other endpoints.

See the related #3368  for details (and the related slack conversation). Herve found that if you use a subpath, then requests are not sent to the backend apis server. This can itself be reproduced in the local dev env with the diff on the issue repro steps, and is fixed by the one-line change here (to use a relative URL path).

Whether or not there is still an issue in Herve's environment once the paths are corrected in the dashboard remains to be seen.

### Benefits

Kubeapps can be deployed on a subpath again.

### Possible drawbacks

None that I can see.

### Applicable issues

  - ref #3368

### DIFF
--- a/dashboard/src/shared/url.ts
+++ b/dashboard/src/shared/url.ts
@@ -181,5 +181,5 @@ export const api = {
       `api/v1/clusters/${cluster}/namespaces/${namespace}/operator/${name}/logo`,
   },
 
-  kubeappsapis: "/apis",
+  kubeappsapis: "apis",
 };


dud225 (approved):
I've just built the image locally and I confirm that it addresses the issue I faced.

absoludity:
Excellent, thanks for the confirmation @dud225

antgamdia (approved):
I hope it wasn't just a slash preventing us from releasing this crazy day... :laughing: 
But happy to know it solves the issue with subpaths!

absoludity:
> I hope it wasn't just a slash preventing us from releasing that crazy day... laughing

I'm pretty sure it'll be something else, since in that case, other paths were working which do grpc requests.

PR #3371, opened 2021-09-08
Kubeapps-APIs implementation of helm create installed pkg 

absoludity:
### Description of the change

Adds implementation for `CreateInstalledPackage` for the helm plugin.

```
$ grpcurl -plaintext -d '{"available_package_ref": {"context": {"namespace": "kubeapps"},"identifier": "bitnami/apache"}, "target_context": {"namespace": "default"}, "name": "apache-via-api"}' localhost:8080 kubeappsapis.plugins.helm.packages.v1alpha1.HelmPackagesService.CreateInstalledPackage
{
  "installedPackageRef": {
    "context": {
      "cluster": "default",
      "namespace": "default"
    },
    "identifier": "apache-via-api",
    "plugin": {
      "name": "helm.packages",
      "version": "v1alpha1"
    }
  }
}

$ k -n default get po
NAME                              READY   STATUS    RESTARTS   AGE
apache-via-api-6dbc85b48b-cdd56   0/1     Running   0          11s
```

### Benefits

One more piece we can switch over with the current helm support.

### Possible drawbacks

None that I'm aware of. We really should do some profiling of this function - similar to the non-kubeapps-apis version it takes a while to complete, mostly I assume because it's downloading the chart from the index, but perhaps there are other causes which we can get around (such as, it may be first downloading the index, even though we should have the URL).

### Applicable issues

  - Ref #3146

### Additional information

I'll follow with a PR to ensure we can set the user-agent appropriately so we don't lose stats again :)

### DIFF
--- a/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
+++ b/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
@@ -73,6 +73,7 @@ type Server struct {
 	globalPackagingCluster   string
 	manager                  utils.AssetManager
 	actionConfigGetter       helmActionConfigGetter
+	chartClientFactory       chart.ChartClientFactoryInterface
 }
 
 // NewServer returns a Server automatically configured with a function to obtain
@@ -140,6 +141,7 @@ func NewServer(configGetter server.KubernetesConfigGetter, globalPackagingCluste
 		manager:                  manager,
 		globalPackagingNamespace: kubeappsNamespace,
 		globalPackagingCluster:   globalPackagingCluster,
+		chartClientFactory:       &chart.ChartClientFactory{},
 	}
 }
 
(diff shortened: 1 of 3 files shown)


cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -797,3 +805,136 @@ func installedPkgDetailFromRelease(r *release.Release, ref *corev1.InstalledPack
...
+			AppRepositoryResourceName:      appRepo.Name,
+			AppRepositoryResourceNamespace: appRepo.Namespace,
+			ChartName:                      chartName,
+			Version:                        chartVersion,
+		},
+		appRepo,
+		caCertSecret, authSecret,
+		// TODO(minelson): add a useragent comment to kubeapps APIs and ensure it is passed
+		// through to be used here.
+		s.chartClientFactory.New(appRepo.Spec.Type, "kubeapps-apis/devel"),
```
antgamdia, line 850: Weren't we already using the proper user-agent?

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -797,3 +805,136 @@ func installedPkgDetailFromRelease(r *release.Release, ref *corev1.InstalledPack
...
+	// TODO: currently app repositories are only supported on the cluster on
+	// which Kubeapps is installed. #1982
+	chartID := request.GetAvailablePackageRef().GetIdentifier()
+	repoNamespace := request.GetAvailablePackageRef().GetContext().GetNamespace()
+	repoName, chartName, err := splitChartIdentifier(chartID)
+	chartVersion := request.GetPkgVersionReference().GetVersion()
+	if err != nil {
+		return nil, err
+	}
+
```
antgamdia, line 831: Perhaps adding a console message to log this fn has been invoked? We are doing so in many fn, but not in every one. Should we remove them (or set other log level), or leave them? 
I personally like having a log entry for each operation, but see what you think.

antgamdia (approved):
Great!!! Really happy to have this operation finally implemented. Looking forward to using it in the UI :)

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
absoludity: You reviewed that branch first, but it was done after this one and targeted this branch (when you reviewed it... I'll be rebasing it after landing this one).

PR #3373, opened 2021-09-08
Add plugin-specific user-agent to kubeapps-apis helm plugin

absoludity:
### Description of the change

Ensures that the user-agent is set when fetching the chart from the bitnami repository with with something that can be used to query access logs for all requests from the kubeapps-apis server or from specific plugins or specific versions and builds of specific plugins.

As an example, after building the image with:

```bash
$ IMAGE_TAG=dev4 VERSION=test-version-from-env make kubeapps/kubeapps-apis
```

and loading and running, I see the following in the logs when creating a new installed package:

```
I0908 03:50:27.512904       1 server.go:841] +helm CreateInstalledPackage fetching chart "bitnami/
apache" with user-agent "kubeapps-apis/plugins/helm.packages/v1alpha1/test-version-from-env"
I0908 03:50:28.960310       1 plugins.go:114] +core GetConfiguredPlugins
I0908 03:50:38.960270       1 plugins.go:114] +core GetConfiguredPlugins
2021/09/08 03:50:48 Downloading https://charts.bitnami.com/bitnami/apache-8.6.3.tgz ...
I0908 03:50:48.900042       1 plugins.go:309] +clientGetter.GetClient
```

### Benefits

We can track requests to the bitnami repository for charts using our new api server.

### Applicable issues

  - Ref #3146 

### Additional information

Confirmed that more than 50% of the request time is actually fetching the index and then parsing it to get the specific chart URL in the index. We should instead be storing the URL for specific chart (versions) in the cache when syncing repositories, so that installing a chart doesn't first require fetching the index. This would halve this lengthy endpoint.

### DIFF
--- a/cmd/kubeapps-apis/Dockerfile
+++ b/cmd/kubeapps-apis/Dockerfile
@@ -5,6 +5,7 @@ WORKDIR /go/src/github.com/kubeapps/kubeapps
 COPY go.mod go.sum ./
 COPY pkg pkg
 COPY cmd cmd
+ARG VERSION="devel"
 
 ARG BUF_VERSION="0.52.0"
 RUN curl -sSL "https://github.com/bufbuild/buf/releases/download/v$BUF_VERSION/buf-Linux-x86_64" -o "/tmp/buf" && chmod +x "/tmp/buf"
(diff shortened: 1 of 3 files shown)


antgamdia (approved):
Great! Thanks for adding it (I always forget about the user-agent thing...)

PR #3374, opened 2021-09-08
Chart: update the cert-manager annotations

dud225:
### Description of the change

According to the [documentation](https://cert-manager.io/docs/usage/ingress/#optional-configuration), the `kubernetes.io/tls-acme` annotations pertains to the https://github.com/jetstack/kube-lego project which is no longer maintained and has been displaced by cert-manager.
As a result the cert-manager annotations should be used instead.


antgamdia:
Thanks for the contribution! We are currently analyzing the change and synchronizing with the people maintaining the  `bitnami/charts` repository as we need to keep our charts as uniform as possible in certain aspects. We

dud225:
Closed in favor of https://github.com/bitnami/charts/pull/7434

Issue #3340, opened 2021-09-01
Implement remaining parts for the Helm-direct plugin


antgamdia:
As https://github.com/kubeapps/kubeapps/issues/3085 and https://github.com/kubeapps/kubeapps/issues/3165 became really big issues (we tracked the status there), I'm opening this new one just to track the remaining or unfinished aspects preventing us from having a fully-working Helm-direct plugin.

- [x]  https://github.com/kubeapps/kubeapps/issues/2944
    - [x]  https://github.com/kubeapps/kubeapps/pull/3427
    - [x] https://github.com/kubeapps/kubeapps/pull/3487
- [x]  Implement already agreed operations (Helm plugin)
    - [x] https://github.com/kubeapps/kubeapps/pull/3371
    - [x] https://github.com/kubeapps/kubeapps/pull/3430
    - [x] https://github.com/kubeapps/kubeapps/pull/3459
    - [x] https://github.com/kubeapps/kubeapps/pull/3487
- [x] Use core APIs instead of the Helm ones
    - [x]  https://github.com/kubeapps/kubeapps/pull/3376
    - [x]  https://github.com/kubeapps/kubeapps/pull/3377
- [x] Use new core API instead of Kubeops for managing Helm:
    - [x] https://github.com/kubeapps/kubeapps/pull/3463
    - [x] https://github.com/kubeapps/kubeapps/pull/3476
    - [x] https://github.com/kubeapps/kubeapps/pull/3475
    - [x] https://github.com/kubeapps/kubeapps/pull/3490
- [x] Bug-hunting:
    - [x] https://github.com/kubeapps/kubeapps/pull/3338
    - [x] https://github.com/kubeapps/kubeapps/issues/3368
    - [x] https://github.com/kubeapps/kubeapps/pull/3369
- [x] On-the-fly refactoring:
    - [x] https://github.com/kubeapps/kubeapps/pull/3373
    - [x] https://github.com/kubeapps/kubeapps/pull/3370
    - [x] https://github.com/kubeapps/kubeapps/pull/3362
    - [x] https://github.com/kubeapps/kubeapps/pull/3439
- [x] https://github.com/kubeapps/kubeapps/pull/3533

Feel free to edit this issue to add something I might have missed.

_Updated on September 30th_

antgamdia:
[...]

Issue #3339, opened 2021-09-01
Handle multiple matching charts in the kubeapps-api Helm plugin


antgamdia:
### Description:

When there are multiple charts (packages) matching a certain application the UX could be pretty misleading as we just return the first result (see below) and it might not meet the user expectations.

Scenario: company C add "bitnami/charts" as global chart. Later, user U add "tac/charts" in their own namespace. As a result, user u would get their chart upgraded with the values of one of these two repos (we're not sorting results, so non-deterministic results are expected). 

A possible solution would be changing proto to AvailablePackageRef be an array, so that we return any matching package (and deal with it in the UI, as we currently are, with the disambiguation page). 

Here are the problematic //TODO lines:

https://github.com/kubeapps/kubeapps/blob/b063daa3ffdb50624c6b3c6f791051338cff1e56/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go#L637-L653

https://github.com/kubeapps/kubeapps/blob/b063daa3ffdb50624c6b3c6f791051338cff1e56/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go#L747-L767

### Steps to reproduce the issue:

1. Add bitnami/charts as a repo
2. Add bitnami/charts or TAC or jfrog or any other repo containing a subset overlap
3. Upgrade an app

### Describe the results you received:

The values will come from one of the multiple matching packages in a non-deterministic way.

### Describe the results you expected:

I'd expect having the exact match as part of the installedPackage, but, if not possible, perhaps we should allow multiple matching references.

### Additional information you deem important (e.g. issue happens only occasionally):

If (and only if) the two matching charts contain the same values, this issue won't have an impact. However, we can't guarantee it.

PR #3377, opened 2021-09-08
Use the core apis in the UI

### LINKED ISSUE #3404: Make the UI use the core APIs instead of the Helm ones

antgamdia:
### Description of the change

This PR transitions from the direct usage of the Helm plugin to the generic Core APIs. Although it was initially said that it wouldn't have major implications, we do need to uniquely define which plugin a certain package refers to. 
.
Note we can't always rely on the state, because users are expected to reload the page at any moment. As a consequence, this information ought to be stored as part of the URL, wherefore so many changes in this PR.

It is not something negative, though, but rather good in that we will be able to render different components based on the plugin. Currently, it is just the background image in the catalog, but in the future, we can even use a similar approach as Anthony implemented in https://github.com/kubeapps/kubeapps/pull/3346.

### Benefits

Ui will use the core APIs, which will allow adding any other package format -more or less- seamlessly.

### Possible drawbacks

As of now, the tests are failing and the are some issues during some component render (a wrong dep in a `userEffect` probably), but I haven't had a chance yet to look at it. 

### Applicable issues

  - related #3340
  - closes #3404


absoludity:
> Although it was initially said that it wouldn't have major implications, we do need to uniquely define which plugin a certain package refers to.

Yes, we were thinking that every package returned from the API does uniquely identify the plugin, which it does, but ...

> .
> Note we can't always rely on the state, because users are expected to reload the page at any moment. As a consequence, this information ought to be stored as part of the URL, wherefore so many changes in this PR.

Right, we missed that, that  *fetching* the package needs to be possible based on the SPA URL in case of reload.

> 
> It is not something negative, though, but rather good in that we will be able to render different components based on the plugin.

Yep, neat!

antgamdia:
Marking a ready, but it requires the core operations to be implemented before merging this one (I didn't push upstream my changes, but there are just two or three extra go files, the rest is the UI thing)

dashboard/src/shared/utils.ts
```diff
@@ -57,3 +61,26 @@ export function flattenResources(
   });
   return result;
 }
+
+export function getPluginIcon(plugin: Plugin | string) {
```
antgamdia, line 65: Initial change towards having custom UIs based upon the plugin. I just have split removed each imported icon in favor of a centralized way to decide which icon we should render.

dashboard/src/actions/repos.ts
```diff
@@ -359,13 +359,18 @@ export function findPackageInRepo(
 ): ThunkAction<Promise<boolean>, IStoreState, null, AppReposAction> {
   return async dispatch => {
     dispatch(requestRepo());
-    if (app?.availablePackageRef?.context?.namespace && app?.availablePackageRef?.identifier) {
+    if (
+      app?.availablePackageRef?.context?.namespace &&
+      app?.availablePackageRef?.identifier &&
+      app?.availablePackageRef?.plugin
```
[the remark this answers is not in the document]
absoludity: Also, not sure if I'm reading the indentation correctly, but looks like we dispatch an error saying that the chart could not be found, rather than that not enough info was provided below, for the `else` of this condition?

dashboard/src/components/AppList/AppListItem.tsx
```diff
@@ -58,7 +60,12 @@ function AppListItem(props: IAppListItemProps) {
   return (
     <InfoCard
       key={app.installedPackageRef?.identifier}
-      link={url.app.apps.get(cluster, app.installedPackageRef?.context?.namespace || "", app.name)}
+      link={url.app.apps.get(
+        cluster,
+        app.installedPackageRef?.context?.namespace || "",
+        app.name,
+        pluginObj,
```
absoludity, line 67: Similar here in that I wonder if it'd be cleaner for us to use an `InstalledPackageRef` as the arg here?

Also, related to that, isn't it `app.installedPackageRef?.identifier` rather than `app.name` that we should be using here? (for helm they'll be identical, but that's up to the plugin). Hmm, not sure if that applies given this is not an api url, but a frontend hash url. We still want them to be unique, right? (and a different plugin may define the identifier differently from the name to maintain uniqueness).

dashboard/src/components/AppUpgrade/AppUpgrade.tsx
```diff
@@ -48,17 +50,28 @@ function AppUpgrade() {
...
+    dispatch(actions.apps.getApp(cluster, namespace, releaseName, pluginObj));
+  }, [dispatch, cluster, namespace, releaseName, pluginObj]);
 
   useEffect(() => {
-    if (repoNamespace && app?.availablePackageRef?.identifier && app?.currentVersion?.pkgVersion) {
+    if (
+      repoNamespace &&
+      app?.availablePackageRef?.identifier &&
+      app?.currentVersion?.pkgVersion &&
+      app?.availablePackageRef?.plugin
```
absoludity, line 67: Possible `.isValid()` again (not something needed, just wondering what you think, rather than checking multiple fields in multiple places).

dashboard/src/components/AppView/AppControls/UpgradeButton/UpgradeButton.tsx
```diff
@@ -9,16 +10,18 @@ interface IUpgradeButtonProps {
...
 export default function UpgradeButton({
   cluster,
   namespace,
   releaseName,
   releaseStatus,
+  plugin,
 }: IUpgradeButtonProps) {
   return (
-    <Link to={url.app.apps.upgrade(cluster, namespace, releaseName)}>
+    <Link to={url.app.apps.upgrade(cluster, namespace, releaseName, plugin)}>
```
absoludity, line 24: Similar point regarding uniqueness (identifier instead of releaseName), and whether it's worth thinking about passing an `InstalledPackageRef` instead of many fields.

dashboard/src/components/ChartView/ChartView.tsx
```diff
@@ -50,18 +53,34 @@ export default function ChartView() {
...
-  }, [dispatch, packageId, chartNamespace, chartCluster, queryVersion]);
+  }, [dispatch, packageId, chartNamespace, chartCluster, queryVersion, pluginObj]);
 
   // Fetch all versions
   useEffect(() => {
-    dispatch(actions.charts.fetchChartVersions(chartCluster, chartNamespace, packageId));
-  }, [dispatch, packageId, chartNamespace, chartCluster]);
+    dispatch(
+      actions.charts.fetchChartVersions(chartCluster, chartNamespace, packageId, {
+        name: plugin?.split("-")[0],
```
absoludity, line 79: Don't you already have `pluginObj.name` here? (and `.version`)

dashboard/src/components/DeploymentForm/DeploymentForm.tsx
```diff
@@ -50,10 +53,14 @@ export default function DeploymentForm() {
   const [releaseName, setReleaseName] = useState("");
   const [appValues, setAppValues] = useState(values || "");
   const [valuesModified, setValuesModified] = useState(false);
+  const [pluginObj] = useState(
+    selected.availablePackageDetail?.availablePackageRef?.plugin ??
+      ({ name: plugin?.split("-")[0], version: plugin?.split("-")[1] } as Plugin),
```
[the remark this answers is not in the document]
absoludity: Is there ever a case that the plugin is not defined in the state? There shouldn't be, it should be required in the ref.

RE using a `-`, yeah, it's tricky, given that a `-` could be in both the plugin name and/or the version. Perhaps check the regex used for pod/deployment names in k8s. Not 100%, but from memory `-` is fine, but `.` isn't, in which case we could split on the first period (eg. `helm.v1alpha1` or `helm.v1.2` would both work). See what you think.

dashboard/src/components/SelectRepoForm/SelectRepoForm.test.tsx
```diff
@@ -16,6 +17,7 @@ const installedPackageDetail = {
   availablePackageRef: {
     context: { cluster: "default", namespace: "default" },
     identifier: "bitnami/my-chart",
+    plugin: { name: "my.plugin", version: "0.0.1" } as Plugin,
```
absoludity, line 20: Heh - yeah, so with my suggestion above, `my.plugin` would not be allowed as a name, but again, let's look for precedents in the k8s names.

dashboard/src/containers/RoutesContainer/Routes.tsx
```diff
@@ -25,17 +25,17 @@ type IRouteComponentPropsAndRouteProps = RouteProps & RouteComponentProps<any>;
...
   "/c/:cluster/ns/:namespace/catalog": Catalog,
   "/c/:cluster/ns/:namespace/catalog/:repo": Catalog,
-  "/c/:cluster/ns/:namespace/charts/:repo/:id": ChartView,
-  "/c/:cluster/ns/:namespace/:global(global)-charts/:repo/:id": ChartView,
-  "/c/:cluster/ns/:namespace/charts/:repo/:id/versions/:version": ChartView,
-  "/c/:cluster/ns/:namespace/:global(global)-charts/:repo/:id/versions/:version": ChartView,
+  "/c/:cluster/ns/:namespace/charts/:repo/:plugin/:id": ChartView,
+  "/c/:cluster/ns/:namespace/charts/:repo/:plugin/:id/versions/:version": ChartView,
+  "/c/:cluster/ns/:namespace/:global(global)-charts/:repo/:plugin/:id": ChartView,
+  "/c/:cluster/ns/:namespace/:global(global)-charts/:repo/:plugin/:id/versions/:version": ChartView,
```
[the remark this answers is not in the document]
absoludity: I think it's not a big enough issue to warrant it, but if you want to, sure :)

dashboard/src/shared/utils.ts
absoludity: Excellent.

absoludity (approved):
Excellent work (as always) Antonio!

dashboard/src/components/DeploymentForm/DeploymentForm.tsx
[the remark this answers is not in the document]
antgamdia: > Is there ever a case that the plugin is not defined in the state? There shouldn't be, it should be required in the ref.

Not usually, but if refreshing the page or accessing it directly (eg a bookmarked link), the state won't be populated. 

Yep, I used `-` but quickly realized it wasn't a good idea. I'm changing it to `+` instead, as it isn't allowed in k8s [1] but is a legal URI character [2] that is, also, unescaped when is not part of a query param [3].
After this change, we will have sth like `helm.packaging+v1alpha1`. 
If `+` doesn't fit you, feel free to propose any other from [3]  (if other is chose, we'd have to deal with encoding)


[1] https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L179

[2] https://datatracker.ietf.org/doc/html/rfc3986#section-3.1

[3] This list includes: `#`,`&`,`+`,`,`,`/`,`:`,`;`,`=`,`@`

dashboard/src/actions/charts.ts
```diff
@@ -106,11 +108,18 @@ export function fetchChartVersion(
...
   version?: string,
 ): ThunkAction<Promise<void>, IStoreState, null, ChartsAction> {
   return async dispatch => {
     try {
-      const response = await Chart.getAvailablePackageDetail(cluster, namespace, id, version);
+      const response = await Chart.getAvailablePackageDetail(
+        cluster,
+        namespace,
+        id,
+        plugin,
```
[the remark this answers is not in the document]
antgamdia: Yep, good idea. Done. 
However, at some points, it is not just the `namespace` field that is used, but instead something like "global namespace or xxx" or even "route param namespace or xxx". I've left this logic as is for now, but we'll want to revisit it when refactoring the AppRepos or even when adding Carvel pkgs.

dashboard/src/actions/repos.ts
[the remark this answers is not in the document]
antgamdia: wrt the first comment, yep, fixed. Once we use the `...packageRef` we no longer need this check (which was mainly required in typescript, given that some field could eventually be undefined - although we could have just appended `!` at the end to force the type assertion-)

This function is used to, given an InstalledPackage, trace it back and retrieve the proper details. However, take into account that, we are here (in this function) because we don't have enough data to populate the upgrade form (usually pkg values, pkg id or pkg version).

Even if a manual repo disambiguation should be enough, we are not always on the ideal case. Let's see:

1. Ideal scenario: an app has been installed from a repo that still exists, still contains the package and it has been installed with a plugin, but we have to disambiguate manually. 
2. Bad, but recoverable, scenario: the repo doesn't exist or it no longer contains this app. Once the user adds a new repo containing the item, everything backs to normal.
3. Worst case, unknown package: there is an unknown package (aka, w/o any availablePackage reference or plugin). In that case, just show "ey, this app cannot be upgraded"

However, the underlying problem was that we were actually allowing those weird cases. I mean, why would the UI allow  upgrading something we know is not upgradeable whatsoever. So, in this PR (I can do it separately if needed), I'm disabling the upgrade/rollback buttons and rendering the AppView, so that users don't see "Error: xxx", but a handy warning.

Demo here:


https://user-images.githubusercontent.com/11535726/133310239-1f6d4798-24b7-41ac-b3d0-86840cf660f8.mp4

dashboard/src/components/AppList/AppListItem.tsx
antgamdia: Yes, I missed that one. Thanks!

dashboard/src/components/AppView/AppControls/UpgradeButton/UpgradeButton.tsx
antgamdia: Done!

dashboard/src/components/AppUpgrade/AppUpgrade.tsx
antgamdia: As in other comment
> wrt the first comment, yep, fixed. Once we use the ...packageRef we no longer need this check

dashboard/src/components/ChartView/ChartView.tsx
antgamdia: That's was the issue I was facing at when I started this PR, if `pluginObj` is added to dependency array of this useEffect it ended up in an endless loop :D However, now it's fixed with `plugin: getPluginFromString(plugin),`

dashboard/src/components/SelectRepoForm/SelectRepoForm.test.tsx
antgamdia: Using now `+`, but I've refactored this logic to just change the char in a single place, just in case.

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -195,82 +203,96 @@ export default function AppView() {
...
-        ) : (
-          <Alert theme="danger">An error occurred: {error.message}</Alert>
-        ))}
-      {!app || !app?.status?.userReason ? (
-        <LoadingWrapper loadingText={`Loading ${releaseName}...`} />
+    <LoadingWrapper loaded={!!app} loadingText="Retrieving application..." className="margin-t-xl">
+      {!app || !app?.installedPackageRef ? (
+        <Alert theme="danger">There is a problem with this package</Alert>
       ) : (
-        <Row>
```
antgamdia, line 235: This was allowing wrong errors during a normal load. Indeed, in the video I recorded, there's a red message complaining about a plugin not being set or sth like that. Moved to avoid this behavior

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -195,82 +203,96 @@ export default function AppView() {
...
-      {error &&
-        (error.constructor === DeleteError ? (
-          <Alert theme="danger">Unable to delete the application. Received: {error.message}</Alert>
-        ) : (
-          <Alert theme="danger">An error occurred: {error.message}</Alert>
-        ))}
-      {!app || !app?.status?.userReason ? (
-        <LoadingWrapper loadingText={`Loading ${releaseName}...`} />
+    <LoadingWrapper loaded={!!app} loadingText="Retrieving application..." className="margin-t-xl">
+      {!app || !app?.installedPackageRef ? (
```
antgamdia, line 209: Redundant, but otherwise the typescript transpiler complains about possible undefined. If forcing type assertion `!`, the linter complains :P

dashboard/src/components/AppView/AppControls/RollbackButton/RollbackDialog.tsx
```diff
@@ -53,7 +54,18 @@ function RollbackDialog({
         <CdsModal closable={false} onCloseChange={closeModal}>
           <CdsModalHeader>Rollback application</CdsModalHeader>
           <CdsModalContent>
-            {error && <Alert theme="danger">An error occurred: {error.message}</Alert>}
+            {error &&
+              (error.constructor === FetchWarning ? (
+                <Alert theme="warning">
+                  There is a problem with this package: {error["message"]}
```
antgamdia, line 60: Added `FetchWarning` to allow viewing partially-missing packages and degraded interactions (ie, just delete the package, not deleting or rolling back)

dashboard/src/components/DeploymentForm/DeploymentForm.tsx
```diff
@@ -58,7 +59,13 @@ export default function DeploymentForm() {
   );
 
   useEffect(() => {
-    dispatch(actions.charts.fetchChartVersions(chartCluster, chartNamespace, packageId, pluginObj));
+    dispatch(
+      actions.charts.fetchChartVersions({
+        context: { cluster: chartCluster, namespace: chartNamespace },
+        plugin: pluginObj,
+        identifier: packageId,
+      } as AvailablePackageReference),
```
antgamdia, line 67: This is what I meant in another comment: in this PR I'm just using the same variables we already had to build the `xxReference`... however, I think we should move to an approach like:

Is this component a view with route params? 
- Yes: merge (with a util function) these params with the ones in the state (if any).
- No, that means it is just a child component used like `<Blablabla props={props}...>`. Rely on the data passed by the parent component.

I can do it here or maybe in a separate PR, I don't mind.

dashboard/src/components/UpgradeForm/UpgradeForm.tsx
```diff
@@ -163,102 +175,96 @@ function UpgradeForm({
...
+        </h3>
+      )}
+      {!isFetching && error && <Alert theme="danger">An error occurred: {error?.message}</Alert>}
+      <LoadingWrapper
+        loaded={!isDeploying && !isFetching && versions?.length > 0 && !!availablePackageDetail}
+      >
+        {(!isFetching && versions?.length === 0) || !availablePackageDetail ? (
+          <></>
+        ) : (
+          <>
```
antgamdia, line 219: Similar thing as already commented herein: avoiding wrong transient error messages due to an incomplete page load,

dashboard/src/components/DeploymentForm/DeploymentForm.tsx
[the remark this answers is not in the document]
absoludity: Great - that's a good solution.

dashboard/src/actions/repos.ts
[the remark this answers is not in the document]
absoludity: Thanks for the demo.

>     3\. Worst case, unknown package: there is an unknown package (aka, w/o any availablePackage reference or plugin).

I'm still unsure why we'd ever be missing a plugin here (regardless of whether the page reloads). We fetch the installed package via a plugin - it should be an error on our side if there's no plugin. But yes, if the plugin returns the installed package without an available package reference, then it was unable to find one.

> In that case, just show "ey, this app cannot be upgraded"

Rather than "There is a problem with this package: this package has missing information, some actions might not be available", perhaps we can give a way forward... something like: "Kubeapps was unable to find the source from which "kubo" was installed. You may need to re-add the source to enable some actions." Hmm, but it is tricky keeping that general.


> 
> 
> However, the underlying problem was that we were actually allowing those weird cases. I mean, why would the UI allow upgrading something we know is not upgradeable whatsoever.

Yeah, similar to displaying a selection of releases which may not even exist to roll-back to.

> So, in this PR (I can do it separately if needed), I'm disabling the upgrade/rollback buttons and rendering the AppView, so that users don't see "Error: xxx", but a handy warning.

Great.

absoludity:
Still +1 once CI passes :)

antgamdia:
> Still +1 once CI passes :)

It's funny... since https://github.com/bitnami/charts/commit/4903073af146c2358eccd8d742e50fc6de0f92a8 (10h ago), our tests fail because of https://github.com/kubeapps/kubeapps/issues/3339. Since it gets the first matching version from the bitnami repo instead of the charmusseum one... from now on, "latest" is no longer 2.6.3 but 2.3.4.
I've added a warning with a TODO in the test.

absoludity:
> > Still +1 once CI passes :)
> 
> It's funny... since [bitnami/charts@4903073](https://github.com/bitnami/charts/commit/4903073af146c2358eccd8d742e50fc6de0f92a8) (10h ago), our tests fail because of #3339. Since it gets the first matching version from the bitnami repo instead of the charmusseum one... from now on, "latest" is no longer 2.6.3 but 2.3.4.
> I've added a warning with a TODO in the test.

Yeah, so we (Kubeapps, and Helm for that matter) do not know which repo was used to install something. We can check for the exact current version, but multiple repos may match that too.

PR #3388, opened 2021-09-09
Update to React 17

antgamdia:
### Description of the change

This PR simply upgrades the react version from 16 to 17, which is mostly an internal refactor with no new features (they will add them in React 18 instead).
In fact, they say:
> We’ve only had to change fewer than twenty components out of 100,000+ in the Facebook product code to work with these changes, so we expect that most apps can upgrade to React 17 without too much trouble. Please tell us if you run into problems.

However, the enzyme adaptor we are using for the tests has to be updated accordingly (see drawbacks).

### Benefits

We will be ready for the upcoming react versions.

### Possible drawbacks

There is no official enzyme adapter yet (although there is an open PR https://github.com/enzymejs/enzyme/pull/2430). In the meanwhile, there is an unofficial module published that is working (working=our tests are passing, since enzyme is just for testing) (more info https://github.com/enzymejs/enzyme/issues/2429#issuecomment-679265564)

### Applicable issues

N/A

### Additional information

There is no reason to merge this PR now, as it is just an update I wanted to perform some time ago, but, as I said, no new features so far. We can hold it until we release 2.4.1, for instance. Up to you, I don't foresee any errors, but let's see what our beloved CI thinks.


dashboard/src/components/js/hooks/useOutsideClick/useOutsideClick.test.js
```diff
@@ -30,21 +30,27 @@ describe(useOutsideClick, () => {
...
+    document.addEventListener = jest.fn(
+      (event, cb) => {
+        listeners[event] = cb;
+      },
+      { capture: true },
+    );
 
     mount(<TestComponent />);
-    expect(Object.keys(listeners).length).toBe(1);
+    expect(Object.keys(listeners).length).toBe(2);
```
absoludity, line 41: Why did this change?

absoludity (approved):
CI is happy. I'm happy :)

dashboard/src/components/js/hooks/useOutsideClick/useOutsideClick.test.js
antgamdia: I'm not 100% sure, but it has something to do with the changes in the event delegation [1], but I just manually confirmed the hook was working as expected (that is, clicking outside the context switcher and right menu is working as expected), but I performed this change for the tests to pass.

[1] https://reactjs.org/blog/2020/10/20/react-v17.html#changes-to-event-delegation

PR #3405, opened 2021-09-13
Update ROADMAP.md

antgamdia:
### Description of the change

I've noticed the table in the Roadmap file wasn't being properly rendered; therefore this PR is to fix this problem as well as other minor format and syntax issues.

I've also run `npx prettier --write  ROADMAP.md`

### Benefits

The roadmap table will render again

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

Longer-term I'd also propose creating an issue for tracking each roadmap item (that is, an epic to link the rest of the issues). Let's talk about it during the iteration.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -4,7 +4,7 @@
 
 **About this document**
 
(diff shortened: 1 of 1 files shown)


absoludity (approved):
> Longer-term I'd also propose creating an issue for tracking each roadmap item (that is, an epic to link the rest of the issues). Let's talk about it during the iteration.

Yeah, that'd be great to link from the roadmap directly to something with a burn-down task list.

Issue #3385, opened 2021-09-09
Add proper request logging middleware or similar for kubeapps-apis


antgamdia:
Currently nothing is written for each request itself.

_Originally posted by @ absoludity_

Edit:

This issue is rather wide, depending on what we focus on. I don't know what Michael's idea is, but as far as I see it:

- The error messages that are displayed to the client is not the best. We are just aggregating the error chain like: "there is an error: error while doing X: error calling Y: error whatever", instead of "error: the user is not allowed to perform the action X". This is not a matter of adding a logging interceptor, but just making the top-level error messages (those returned to the users) a little bit clearer.
- The plugins are defining, right now, their own login system (klog, logrus, etc), but, for the sake of standardizing these common aspects for the plugins, the main server should pass a pre-configured logger (as we are passing the k8s clients or the token), so the plugins only have to use it for printing mgs.
    - An idea is using this project: https://github.com/grpc-ecosystem/go-grpc-middleware  as it provides middleware for not only logging, but any other future thing we might want to tackle in the future (like rate limiting).

absoludity:
@satya-dillikar

satya-dillikar:
> Currently nothing is written for each request itself.
> 
> _Originally posted by @ absoludity_
> 
> Edit:
> 
> This issue is rather wide, depending on what we focus on. I don't know what Michael's idea is, but as far as I see it:
> 
> * The error messages that are displayed to the client is not the best. We are just aggregating the error chain like: "there is an error: error while doing X: error calling Y: error whatever", instead of "error: the user is not allowed to perform the action X". This is not a matter of adding a logging interceptor, but just making the top-level error messages (those returned to the users) a little bit clearer.

@antgamdia, Please share an example of the real error, so that I can recreate what you are saying here with X & Y. and evaluate if we can make it better.

> * The plugins are defining, right now, their own login system (klog, logrus, etc), but, for the sake of standardizing these common aspects for the plugins, the main server should pass a pre-configured logger (as we are passing the k8s clients or the token), so the plugins only have to use it for printing mgs.
@antgamdia, Please share the piece of code where we are passing the k8s clients or the token. I would like to see what you thinking here.

>   
>   * An idea is using this project: https://github.com/grpc-ecosystem/go-grpc-middleware  as it provides middleware for not only logging, but any other future thing we might want to tackle in the future (like rate limiting).

[...]

PR #3407, opened 2021-09-13
Use getters to avoid verbose nil checking

antgamdia:
### Description of the change

Follow-up PR based on https://github.com/kubeapps/kubeapps/pull/3376#discussion_r706962795

### Benefits

Less verbosity for just building a log msg

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

N/A

### DIFF
--- a/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
+++ b/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
@@ -169,20 +169,12 @@ func (s *Server) GetManager() (utils.AssetManager, error) {
 
 // GetAvailablePackageSummaries returns the available packages based on the request.
 func (s *Server) GetAvailablePackageSummaries(ctx context.Context, request *corev1.GetAvailablePackageSummariesRequest) (*corev1.GetAvailablePackageSummariesResponse, error) {
-	contextMsg := ""
-	if request.Context != nil {
-		contextMsg = fmt.Sprintf("(cluster=[%s], namespace=[%s])", request.Context.Cluster, request.Context.Namespace)
-	}
-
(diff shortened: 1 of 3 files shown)


cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -169,20 +169,12 @@ func (s *Server) GetManager() (utils.AssetManager, error) {
 
 // GetAvailablePackageSummaries returns the available packages based on the request.
 func (s *Server) GetAvailablePackageSummaries(ctx context.Context, request *corev1.GetAvailablePackageSummariesRequest) (*corev1.GetAvailablePackageSummariesResponse, error) {
-	contextMsg := ""
-	if request.Context != nil {
-		contextMsg = fmt.Sprintf("(cluster=[%s], namespace=[%s])", request.Context.Cluster, request.Context.Namespace)
-	}
-
+	contextMsg := fmt.Sprintf("(cluster=[%s], namespace=[%s])", request.GetContext().GetCluster(), request.GetContext().GetNamespace())
```
absoludity, line 172: +1. I wondered when I initially saw these why we don't instead use the go fmt built in support for quoting things to do something like:

```suggestion
	contextMsg := fmt.Sprintf("(cluster=%q, namespace=%q)", request.GetContext().GetCluster(), request.GetContext().GetNamespace())
```

But not necessary - just if you agree that'd be nicer (main reason for me is that it's clear that cluster and namespace a string values in the log output, as opposed to being arrays/slices of length 1).

absoludity (approved):

antgamdia:
I didn't change anything in the Fluxv2 plugin, since it is under heavy development and I don't want to interfere. 
It is using different log messages (like`log.Infof("+fluxv2 GetInstalledPackageSummaries [%v]", request)`), which isn't sth bad, but opens the discussion for ulterior issues like https://github.com/kubeapps/kubeapps/issues/3385

PR #3427, opened 2021-09-16
Proposal for DeleteInstalledPackage messages.

absoludity:
### Description of the change

The actual messages and RPC for DeleteInstalledPackage isn't very controversial.

But I took the opportunity to start looking at the transcoding to http URLs more seriously. The delete url is interesting in that by default, grpc-gateway and openapiv2 don't allow a body with the request. Anyway, I'll include more comments inline.

Note: I'll not land this until I can update once Greg's branch has landed.

### Benefits

We can implement the delete method. We can follow-up with a PR for consistent URL mappings (using the inbuilt transcoding from grpc to http).

### Applicable issues

- Ref #3146

### Additional information

See https://cloud.google.com/endpoints/docs/grpc-service-config/reference/rpc/google.api#httprule for the grpc transcoding to http URL specification.

If people are happy with the general approach for the URL here for the Delete, I'll do a separate PR with similar transcoding for the other http urls.


cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
```diff
@@ -54,11 +54,17 @@ service PackagesService {
...
   // }
-  // rpc DeleteInstalledPackage(DeleteInstalledPackageRequest) returns (DeleteInstalledPackageResponse) {
-  //   option (google.api.http) = {
-  //     delete: "/core/packages/v1alpha1/installedpackages"
-  //   };
-  // }
+  rpc DeleteInstalledPackage(DeleteInstalledPackageRequest) returns (DeleteInstalledPackageResponse) {
+    option (google.api.http) = {
+      // Example:
+      // /core/packages/v1alpha1/installedpackages/plugin/helm.packages/v1alpha1/ns/my-ns/release-identifier
```
absoludity, line 60: @antgamdia : I tried to use the same format that you'd used in the frontend URLs for the plugin: `helm.packages+v1alpha1` but the grpc transcoding basically enforces that one (path) segment maps to one variable. That's why I've done `{plugin.name}/{plugin.version}` . If you're OK with that, I might update the UI to use the same too so we're consistent.

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
```diff
@@ -54,11 +54,17 @@ service PackagesService {
...
+      delete: "/core/packages/v1alpha1/installedpackages/plugin/{installed_package_ref.plugin.name}/{installed_package_ref.plugin.version}/ns/{installed_package_ref.context.namespace}/{installed_package_ref.identifier}"
```
absoludity, line 61: What I wanted to do here was just `/core/packages/v1alpha1/installedpackages/{identifier built by b64 of the installed package ref}` but it's probably not a good idea (wouldn't be unique) and it's not possible with the grpc transcoding.

So instead of `/installedpackages/{identifier of some sort}` we have a textual identifier built from the plugin/namespace/installed package identifier.

For the plugins (as opposed to the core), the identifier is just the namespace and identifier (since we know the plugin).

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
```diff
@@ -54,11 +54,17 @@ service PackagesService {
...
+      // We allow a body for the delete request so that we can support multi-cluster
+      // environments, where the context.cluster is also set, without requiring the
+      // cluster in the URL.
```
absoludity, line 64: Given that Kubeapps may not be wanted as a multi-cluster packaging solution, I'm a little hesitant to include the cluster in the URL. Instead, it'll be optional as it is already for the grpc calls, but can be specified in the http body.

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
antgamdia: I'd rather go with the "long" identifier solution, albeit more prone to enumeration attacks, it follows the REST spirit. 

I also was going to mention we should add the `cluster` to the path. Reading your comment below... mmm, you're right in that we may not want to make the cluster explicit... but I don't know if it's a good idea to use the message body in a DELETE msg.

Do you know if we can have two HTTP paths linked to the same grpc operation? If so, we could have one with cluster and another one without.

If not... adding a body to a DELETE method is really against the protocol. Whereas is not disallowed... the msg body of a DELETE should be ignored.
That's why there are some API clients (Java Jersey I think?, some Android httpclients and several SaaS systems) that just don't support bodies in DELETEs. 
So... even if technically possible... it could translate into UX issues for developers (which wouldn't be a big deal if it weren't for it being aimed at being consumed by 3rd party people)

Let's see what Greg thinks, but my -1 to use bodies for now :S

antgamdia (approved):
+1 to the general aspects, but I do have concerns with adding bodies to a DELETE request. Let's continue the discussion, but approving for not block.

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
absoludity: > Do you know if we can have two HTTP paths linked to the same grpc operation? If so, we could have one with cluster and another one without.

Let me try - that'd be a much better solution if we can.

> 
> If not... adding a body to a DELETE method is really against the protocol. Whereas is not disallowed... the msg body of a DELETE should be ignored.
> That's why there are some API clients (Java Jersey I think?, some Android httpclients and several SaaS systems) that just don't support bodies in DELETEs.
> So... even if technically possible... it could translate into UX issues for developers (which wouldn't be a big deal if it weren't for it being aimed at being consumed by 3rd party people)

Yeah - I'm totally happy to update to remove it. My only concern was having the cluster in the URL endpoints when certain scenarios of how our API could be used in the future might actively not want multi-cluster support. But it sounds like it'd be worse to have the body in delete than to have a redundant `/c/default/` in the URL. I'll check as above first if we can define multiple paths (I know we can define multiple paths, just unsure whether that can be for the same method, doubt it), but if not, will update removing the body option.

Thanks Antonio.

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
absoludity: Yeah ok, I'm guessing you'll be touching that code soon anyway, thanks.

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
absoludity: Tried with:

```diff
--- a/cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
+++ b/cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
@@ -60,6 +60,7 @@ service PackagesService {
       // Example:
       // /core/packages/v1alpha1/installedpackages/plugin/helm.packages/v1alpha1/c/default/ns/my-ns/release-identifier
       delete: "/core/packages/v1alpha1/installedpackages/plugin/{installed_package_ref.plugin.name}/{installed_package_ref.plugin.version}/c/{installed_package_ref.context.cluster}/ns/{installed_package_ref.context.namespace}/{installed_package_ref.identifier}"
+      delete: "/core/packages/v1alpha1/installedpackages/plugin/{installed_package_ref.plugin.name}/{installed_package_ref.plugin.version}/ns/{installed_package_ref.context.namespace}/{installed_package_ref.identifier}"
     };
   }
 }
```

but

```bash
buf generate

proto/kubeappsapis/core/packages/v1alpha1/packages.proto:63:7:method kubeappsapis.core.packages.v1alpha1.PackagesService.DeleteInstalledPackage: option (google.api.http): non-repeated option field delete already set
```

So for now I've just pushed with the url including the cluster (and no more body on delete :) ).

gfichtenholt: I don't quite understand. DELETE is not so unique. We have several GET operations which cannot have bodies either. So buf happily generates request query parameters for everything. Why would DELETE be so different? See attachment
Screen Shot 2021-09-16 at 6 35 02 PM
is it the case you're trying to accomplish 2 things in one PR:
1) add stubs for DELETE and
2) change the way DELETE is transcoded for grpc?

If so, I would break this into 2 separate PRs:
1) add stubs for DELETE and
2) change the way ALL operations that need so are transcoded so we have a consistent pattern. Not just one sore thumb sticking out

gfichtenholt (commented):
Will give a +1 once I understand why this complexity needs to be done in this PR

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
absoludity: GET params can be transcoded from URI params without issue to populate the mesages. DELETE is different enough that the tooling does not allow DELETE with a body by default. See Antonio's comment for more info.



> Is it the case you're trying to accomplish 2 things in one PR:
> 
>     1. add stubs for DELETE and
> 
>     2. change the way DELETE is transcoded for grpc?

I didn't change it, we weren't doing it at all. I was planning to just add the stubs for DELETE but both grpc-gateway *and* openapi refused to allow a body without overriding explicitly (as I'd done originally, you can see it in a previous commit). So I wanted to discuss what would be best for DELETE specifically here and be able to follow up with a PR to propose adding more REST-like transcoding for the other msgs

> 
> 
> If so, may I suggest you split this into 2 separate PRs:
> 
>     1. add stubs for DELETE consistent with the rest of the APIs. So DELETE isn't one sore thumb sticking out and once you land that then

Happy to do so if that makes things easier. I thought it was worthwhile looking at DELETE in particular due to the above warnings that the tooling adds about using a body with DELETE... it seems it is special, but I'll let Antonio (he has a PhD's worth of REST/OpenAPI knowledge) chime in.
 
> 
>     2. change the way ALL operations that need so are transcoded so we have a consistent pattern. We can have a productive discussion then that covers all use cases

Yes, that's the plan (as mentioned in the description), to follow up with a PR proposing the rest if we agree here on the structure for DELETE (cluster/namespace/etc.).

gfichtenholt: > GET params can be transcoded from URI params without issue to populate the mesages. DELETE is different enough that the tooling does not allow DELETE with a body by default. See Antonio's comment for more info.
> 
> > Is it the case you're trying to accomplish 2 things in one PR:
> > ```
> > 1. add stubs for DELETE and
> > 
> > 2. change the way DELETE is transcoded for grpc?
> > ```
> 
> I didn't change it, we weren't doing it at all. I was planning to just add the stubs for DELETE but both grpc-gateway _and_ openapi refused to allow a body without overriding explicitly (as I'd done originally, you can see it in a previous commit). So I wanted to discuss what would be best for DELETE specifically here and be able to follow up with a PR to propose adding more REST-like transcoding for the other msgs
> 
> > If so, may I suggest you split this into 2 separate PRs:
> > ```
> > 1. add stubs for DELETE consistent with the rest of the APIs. So DELETE isn't one sore thumb sticking out and once you land that then
> > ```
> 
> Happy to do so if that makes things easier. I thought it was worthwhile looking at DELETE in particular due to the above warnings that the tooling adds about using a body with DELETE... it seems it is special, but I'll let Antonio (he has a PhD's worth of REST/OpenAPI knowledge) chime in.
> 
> > ```
> > 2. change the way ALL operations that need so are transcoded so we have a consistent pattern. We can have a productive discussion then that covers all use cases
> > ```
> 
> Yes, that's the plan (as mentioned in the description), to follow up with a PR proposing the rest if we agree here on the structure for DELETE (cluster/namespace/etc.).

I read Antonio's comment. Still confused why DELETE is different than a GET in that respect. DELETE with a body is a non-starter. Same as GET. I have known this and operated with this knowledge for possibly at least 8-10 years, ever since my introduction to REST. Having any kind of tooling that allows a body with DELETE should have been a non-starter. I propose we treat DELETE same as GET in this PR, that is have the tooling generate query params for inputs. Then in a different PR, lets discuss how to make the world better :-)

absoludity: Sorry, I'd misread your comment and thought you wanted it consistent with post etc., but see it was GET. Great, so everyone is in agreement about not using body, and Greg is keen that we don't propose a better URL for delete in this PR, but instead use the same and update all in one PR, which makes sense. I've pushed that.

gfichtenholt (approved):
I am good with this. Curious what you plan to do after this: helm implementation of Delete(...) or changing the grpc transcoding of the URLs in APIs. If we are really trying to push get the existing functionality up and running I would go with former. Leave the latter for dessert :-)

absoludity:
> I am good with this. Curious what you plan to do after this: helm implementation of Delete(...) or changing the grpc transcoding of the URLs in APIs. If we are really trying to push get the existing functionality up and running I would go with former. Leave the latter for dessert :-)

I'd already done Delete yesterday based of this PR. So yes :)

PR #3425, opened 2021-09-16
Add stubs for UpdateInstallPackage

gfichtenholt:

PR #3430, opened 2021-09-16
Add UpdateInstalledPackage for Helm plugin.

absoludity:
### Description of the change

Adds the implementation for UpdateInstalledPackage for the Helm plugin, then refactors to DRY up similar code in Create/Update methods.

### Benefits

Can switch UI to use update from new API as well.

### Possible drawbacks

Haven't done an IRL test yet, but will do so before landing.

### Applicable issues

- ref: #3146

### DIFF
--- a/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
+++ b/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
@@ -30,13 +30,14 @@ import (
 	helmv1 "github.com/kubeapps/kubeapps/cmd/kubeapps-apis/gen/plugins/helm/packages/v1alpha1"
 	"github.com/kubeapps/kubeapps/cmd/kubeapps-apis/server"
 	"github.com/kubeapps/kubeapps/pkg/agent"
-	"github.com/kubeapps/kubeapps/pkg/chart"
+	chartutils "github.com/kubeapps/kubeapps/pkg/chart"
 	"github.com/kubeapps/kubeapps/pkg/chart/models"
 	"github.com/kubeapps/kubeapps/pkg/handlerutil"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/anypb"
 	"helm.sh/helm/v3/pkg/action"
+	"helm.sh/helm/v3/pkg/chart"
 	"helm.sh/helm/v3/pkg/kube"
 	"helm.sh/helm/v3/pkg/release"
 	"helm.sh/helm/v3/pkg/storage/driver"
@@ -51,7 +52,7 @@ import (
 )
 
 type clientGetter func(context.Context, string) (kubernetes.Interface, dynamic.Interface, error)
-type helmActionConfigGetter func(ctx context.Context, cluster, namespace string) (*action.Configuration, error)
+type helmActionConfigGetter func(ctx context.Context, pkgContext *corev1.Context) (*action.Configuration, error)
 
 // Compile-time statement to ensure this service implementation satisfies the core packaging API
 var _ corev1.PackagesServiceServer = (*Server)(nil)
@@ -74,7 +75,7 @@ type Server struct {
 	globalPackagingCluster   string
 	manager                  utils.AssetManager
 	actionConfigGetter       helmActionConfigGetter
-	chartClientFactory       chart.ChartClientFactoryInterface
+	chartClientFactory       chartutils.ChartClientFactoryInterface
 }
 
 // NewServer returns a Server automatically configured with a function to obtain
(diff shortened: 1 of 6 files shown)


cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -708,10 +713,9 @@ func (s *Server) GetInstalledPackageDetail(ctx context.Context, request *corev1.
 	log.Infof("+helm GetInstalledPackageDetail %s", contextMsg)
 
 	namespace := request.GetInstalledPackageRef().GetContext().GetNamespace()
-	cluster := request.GetInstalledPackageRef().GetContext().GetCluster()
 	identifier := request.GetInstalledPackageRef().GetIdentifier()
 
-	actionConfig, err := s.actionConfigGetter(ctx, cluster, namespace)
+	actionConfig, err := s.actionConfigGetter(ctx, request.GetInstalledPackageRef().GetContext())
```
antgamdia, line 718: Nice! We have `cluster, namespace` in plenty of different places. Good to see how we progressively are evolving to use the Context object (front/back-end)

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -822,70 +826,106 @@ func (s *Server) CreateInstalledPackage(ctx context.Context, request *corev1.Cre
...
+	contextMsg := fmt.Sprintf("(cluster=%q, namespace=%q)", installedRef.GetContext().GetCluster(), installedRef.GetContext().GetNamespace())
+	log.Infof("+helm UpdateInstalledPackage %s", contextMsg)
+
+	// Determine the chart used for this installed package.
+	// We may want to include the AvailablePackageRef in the request, given
+	// that it can be ambiguous, but we dont yet have a UI that allows the
+	// user to select which chart to use, so until then.
+	detailResponse, err := s.GetInstalledPackageDetail(ctx, &corev1.GetInstalledPackageDetailRequest{
+		InstalledPackageRef: installedRef,
+	})
```
antgamdia, line 887: Don't we? I guess we can extend the existing `dashboard/src/components/SelectRepoForm/SelectRepoForm.tsx` to support more scenarios like this.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -822,70 +826,106 @@ func (s *Server) CreateInstalledPackage(ctx context.Context, request *corev1.Cre
...
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "Unable to upgrade helm release %q in the namespace %q: %v", releaseName, installedRef.GetContext().GetNamespace(), err)
+	}
+
+	return &corev1.UpdateInstalledPackageResponse{
 		InstalledPackageRef: &corev1.InstalledPackageReference{
 			Context: &corev1.Context{
-				Cluster:   cluster,
+				Cluster:   installedRef.GetContext().GetCluster(),
 				Namespace: release.Namespace,
```
antgamdia, line 929: Doubt: shouldn't cluster and namespace both be extracted from the `installedRef` ? Well, I mean, `release.Namespace` should match `installedRef.GetContext().GetNamespace()` shouldn't it?

antgamdia (approved):
Great! We are really close to having every operation implemented for the Helm-direct case!

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
absoludity: Yeah, it's been nice to gradually simplify calls, not just with context, but also with available/installed package references. Can't do it everywhere in the backend as we're often sharing code with previously working utils, but the closer we get, the less we need those to support both the api and kubeops.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
absoludity: Yeah, it'd also be nice to send it explicitly (given it's already returned to the client with the detail) since we then wouldn't need to fetch the detail here again (took that as a shortcut, could have just fetched the related available pkg ref, and probably should... I'll add a TODO). Greg and I weren't 100% whether we should include the available pkg reference... see the [thread on his PR adding the proto messages](https://github.com/kubeapps/kubeapps/pull/3425#discussion_r709712844) for more context.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
absoludity: Yes, the namespaces should definitely match, which is why, in case of any error in logic, I go with the release namespace. But since the cluster isn't part of the release so I have no option but to use the requested one (which is what is passed to the helm action config). That said, I should be similarly ensuring that cluster is set to the default cluster if it was blank in the request here... I'll add a test and fix that too.

PR #3423, opened 2021-09-15
Inherit labels & annotations in jobs/cronjobs from a apprepository

antgamdia:
### Description of the change

This PR aims at passing labels and annotations to the jobs and cronjobs that the AppRepositoryController is creating. This is being done by inheriting the existing labels/annotations of the AppRepository object.

If a user adds some labels/annotations, we will merge them with the inherited ones.

### Benefits

Currently, these jobs do not inherit any label/annotation, which could be something positive, or not. I'm not 100% sure here to be honest. With this PR, every annotation will be passed through.

### Possible drawbacks

There are two problematic scenarios:

- Cleanup jobs: since they don't belong to any repo, they won't get these labels/annotations.
- User-created repositories: from the UI, since we just POST to the k8s API, these objects haven't any of these labels/annotations (neither we can easily pass via env var, see https://github.com/kubeapps/kubeapps/pull/3423#issuecomment-920973671)

### Applicable issues

- related #3400
- related #3431

### Additional information

This PR won't solve the issue, but could still be useful if we want to inherit the labels. Not sure if it is useful... if not, happy to close the PR.

PR #3431, opened 2021-09-16
Set custom labels and annotations jobs/cronjobs/pods

### LINKED ISSUE #3400: Allow passing custom label/annotation for auto-deleting jobs
Using k8s 1.21 annotations will allow limiting the TTL of each job, which is useful for running Kubeapps.

Using kbld labels will prevent having orphan jobs once Kubeapps is uninstalled 

More context: 
> (...), we also need to add the label `kapp.k14s.io/app`. These Jobs are not created directly by the chart, instead, they are created by the app-repository-controller kubeapps component.
>
> It seems we have two different options:
>
> - Configure Jobs to be deleted automatically https://kubernetes.io/docs/concepts/workloads/controllers/job/#clean-up-finished-jobs-automatically 
> - Allow creating Jobs with custom labels and annotations. (...) we don't really need a custom label for the Jobs. We can just use the ones present in the parent component. For annotation, yes, we should be able to specify it from the outside.

antgamdia:
### Description of the change

This PR aims at setting custom labels/annotaitons to the jobs and cronjobs that the AppRepositoryController. This is being done by adding a new command line argument, so it will be available even for cleanup jobs.

### Benefits

We will be able to set custom labels/annotations for these objects. This approach is valid for cleanup jobs as well as user-created repos.

### Possible drawbacks

There is no granularity as these labels/annotations will be passed to every object. However, I don't foresee any issues here.

### Applicable issues

- fixes #3400
- related #3423

### Additional information

N/A


cmd/apprepository-controller/main.go
```diff
@@ -85,6 +90,8 @@ func parseFlags(progname string, args []string) (config *Config, output string,
 	// The support for this is currently alpha in K8s itself, requiring a feature gate being set to enable
 	// it. See https://kubernetes.io/docs/concepts/workloads/controllers/job/#clean-up-finished-jobs-automatically
 	flagSet.StringVar(&conf.TTLSecondsAfterFinished, "ttl-lifetime-afterfinished-job", "3600", "Lifetime limit after which the resource Jobs are deleted expressed in seconds by default is 3600 (1h) ")
+	flagSet.StringSliceVar(&conf.CustomAnnotations, "custom-annotations", []string{""}, "xxxxxxxxxxxxx")
+	flagSet.StringSliceVar(&conf.CustomLabels, "custom-labels", []string{""}, "xxxxxxxxxxxxx")
```
absoludity, line 94: Guessing the help text was left as a todo?

absoludity (approved):
Much prefer this over the drawbacks of #3423 that you mention.

chart/kubeapps/templates/apprepository/deployment.yaml
```diff
@@ -89,6 +89,16 @@ spec:
...
+              {{- if.Values.apprepository.customAnnotations }}
+                {{- range $key, $value := .Values.apprepository.customAnnotations }}
+            -  --custom-annotations={{ (print $key "=" $value) | quote }}
+                {{- end }}
+              {{- end }}
+              {{- if.Values.apprepository.customLabels }}
+                {{- range $key, $value := .Values.apprepository.customLabels }}
+            -  --custom-labels={{ (print $key "=" $value) | quote }}
+                {{- end }}
+              {{- end }}
```
antgamdia, line 101: I'm passing `Values.apprepository.customAnnotations` instead of `Values.commonAnnotations`, not sure if we should support both for consistency with the expected `commonXXXX` semantics.
I went this way not to struggle with the chart templating if the PR wasn't to be accepted, but no it's has been +1, I wonder what should we do.

cmd/apprepository-controller/main.go
```diff
@@ -69,7 +74,7 @@ func parseFlags(progname string, args []string) (config *Config, output string,
 
 	flagSet.StringVar(&conf.Kubeconfig, "kubeconfig", "", "Path to a kubeconfig. Only required if out-of-cluster.")
 	flagSet.StringVar(&conf.MasterURL, "master", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
-	flagSet.StringVar(&conf.RepoSyncImage, "repo-sync-image", "quay.io/helmpack/chart-repo:latest", "container repo/image to use in CronJobs")
+	flagSet.StringVar(&conf.RepoSyncImage, "repo-sync-image", "docker.io/kubeapps/asset-syncer:latest", "container repo/image to use in CronJobs")
```
antgamdia, line 77: Deleted an old reference in the default value, with no actual impact

chart/kubeapps/templates/apprepository/deployment.yaml
absoludity: > I'm passing `Values.apprepository.customAnnotations` instead of `Values.commonAnnotations`, not sure if we should support both for consistency with the expected `commonXXXX` semantics.

So `custom*` is used so that these can be specific to the apprepository controller right? Yet, you're wondering if we should also add the common ones (which the chart says will be added to every deployed resource). Yes, we probably should, but not an issue if it's not done as part of this. Fine either way.

> 
> I went this way not to struggle with the chart templating if the PR wasn't to be accepted,

What do you mean?

> but now it's has been +1, I wonder what should we do, cc @absoludity

Still +1 from me, confused if you mean you were hesitant to land given it'd changed since the review.

antgamdia: Yes, maybe I didn't explain myself. I'm passing just specific labels/annotations in a dedicated field in the `values.yaml`, which is ok, as we might want these to be specific to the apprepository. However, as you mention, we probably want -also- (not instead) to have the global commons* being passed.
Since it is solely a change in the chart, let me merge the changes in this PR (since it involves coding on our side), and perform this other proposed change in another one (or even in the bitnami/charts repo).

PR #3432, opened 2021-09-16
Try dynamic latest and previous version for upgrade.

absoludity:
### Description of the change

Just updates to not depend on specific apache versions.

Depending on a specific version for the first install used to work since we always displayed all versions available.
We're now displaying the latest 3. So instead this PR aims to just use the latest and previous to test the upgrade.

### Benefits

Don't need to update when new versions published.

### Possible drawbacks

Let's see.

### Applicable issues

- fixes #

### Additional information

I'm not sure about the comment that I've removed here. It looks like you (Antonio) thought this was related to an extra repo, but I don't see why.


absoludity:
Woops, need to update it to grab the next version rather than the next part of the string split of "chartver / appver"

PR #3437, opened 2021-09-17
Use dynamic current and latest versions in test

antgamdia:
### Description of the change

Same purpose as https://github.com/kubeapps/kubeapps/pull/3432, but relying on selectors and pseudoselectors instead. 
Triggering the CI to see if it works, if not, I'll debug in circleci cia ssh

### Benefits

The CI test will no longer depend upon a fixed set of versions, but instead, it will get the `latest` and `latest - 1` to perform the upgrade.

### Possible drawbacks

It will mask issues like #3339, but we'd better go with this approach so that our CI doesn't get stuck in a permanent failure.

### Applicable issues

- related #3339
- closes #3432

### Additional information

N/A

### DIFF
--- a/integration/use-cases/07-upgrade.js
+++ b/integration/use-cases/07-upgrade.js
@@ -16,41 +16,40 @@ test("Upgrades an application", async () => {
 
   await expect(page).toClick("cds-button", { text: "Deploy" });
 
+  let initialPackageVersion = "";
   let currentPackageVersion = "";
 
   await utils.retryAndRefresh(
     page,
     3,
     async () => {
-      // wait to load every pkg version and get the current version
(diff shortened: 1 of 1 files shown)


antgamdia:
Rubber-stamping PR to unblock our ci pipeline

antgamdia:
Perhaps we should either increase the timeouts or add a waitFor .... "current". We are getting some erratic failures and I guess that's why they're due to.

absoludity:
Thanks for picking that up when i left.

> Perhaps we should either increase the timeouts or add a waitFor .... "current". We are getting some erratic failures and I guess that's why they're due to.

Yeah, short-term. We'd all prefer I'm sure to not have to guess and have deterministic tests, but we need to invest in that.

PR #3439, opened 2021-09-17
Add plugin version to the UI paths

antgamdia:
### Description of the change

As suggested in XXXX, we should pass the plugin version as a URL subpath instead of dealing with `xxxx+yyyy`. This PR is to perform the required changes on the UI side.

### Benefits

The UI URLs will be more aligned to the API endpoints.

### Possible drawbacks

N/A, but let the CI decide :P

### Applicable issues

N/A

### Additional information

N/A


absoludity (approved):
Excellent, thanks.

Thought that can be ignored: it'd be nice if we could just pass a `plugin` rather than two strings, and then if `plugin.AsString()` or whatever was just what we need: name/version). Not sure whether the react route would encode the slash. Probably not worth it.

antgamdia:
> Thought that can be ignored: it'd be nice if we could just pass a `plugin` rather than two strings, and then if `plugin.AsString()` or whatever was just what we need: name/version). Not sure whether the react route would encode the slash. Probably not worth it.

Yep, it is easy to override the default `toString()` fn in js, 

```js
class Plugin {name; version; toString(){ return this.name+"/"+this.version; }}
let plugin = new Plugin();plugin.name="helm.packages"; plugin.version="v1beta1"; plugin.toString()
// 'helm.packages/v1beta1'
``` 

but given it is autogenerated code, we would need to extend the `Plugin` class to prevent our changes to be removed. I'll think about it next time working on the UI, though.

Issue #3447, opened 2021-09-20
front: edition of an apprepository filter clears it


dud225:
### Description:

When creating a new Application Repository, all works as expected. But when editing it to filter out some charts, there is no error on the UI but the resulting `apprepositories` filter is reset:
```
  filterRule:
    jq: ""
```

### Steps to reproduce the issue:

1. Create an App Repository, e.g. [Bitnami](https://charts.bitnami.com/bitnami)
2. Once the App Repository is creted, edit it to modify the filter, eg `nginx, jenkins`
3. Check the resulting `apprepositories`, the filter will be blank even if one was set up at the time of creation

### Version of Helm, Kubeapps and Kubernetes:

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.5.0", GitCommit:"32c22239423b3b4ba6706d450bd044baffdcf9e6", GitTreeState:"clean", GoVersion:"go1.15.6"}
```

- Output of `helm list -n <kubeapps-namespace> <kubeapps-release-name>`:

```
NAME            NAMESPACE       REVISION        UPDATED                                 STATUS          CHART           APP VERSION
kubeapps        kubeapps        5               2021-09-08 10:10:27.536393583 +0000 UTC deployed        kubeapps-7.4.0  2.4.0
```

- Output of `kubectl version`:

```
Client Version: version.Info{Major:"1", Minor:"21+", GitVersion:"v1.21.4-3+e5758f73ed2a04", GitCommit:"e5758f73ed2a04898a6bba130459e17b95402dcd", GitTreeState:"clean", BuildDate:"2021-08-12T15:05:00Z", GoVersion:"go1.16.6", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"21+", GitVersion:"v1.21.4-3+e5758f73ed2a04", GitCommit:"e5758f73ed2a04898a6bba130459e17b95402dcd", GitTreeState:"clean", BuildDate:"2021-08-12T15:00:31Z", GoVersion:"go1.16.6", Compiler:"gc", Platform:"linux/amd64"}
```

absoludity:
Thanks @dud225 . Easy to repro. Haven't dug, but from the network request, it looks like the dashboard is simply not setting the filters when updating the app repo. For example, I see the following when setting the filters to "apache,jenkins"
[...]

PR #3448, opened 2021-09-20
Allow filter rules to be set on App Repositories

dud225:
### Description of the change

Allow filter rules to specified when creating Application Repositories from Helm.

### DIFF
--- a/chart/kubeapps/templates/apprepository/apprepositories.yaml
+++ b/chart/kubeapps/templates/apprepository/apprepositories.yaml
@@ -60,5 +60,9 @@ spec:
         name: {{ printf "apprepo-%s-secrets" .name }}
     {{- end }}
   {{- end }}
+  {{- if .filterRule }}
+  filterRule:
+    {{- toYaml .filterRule | nindent 4 }}
+  {{- end }}
 ---
 {{ end -}}
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -729,8 +729,9 @@ apprepository:
   ##     ociRepositories:
   ##       - nginx
   ##       - jenkins
-  ##     # Optionally filter out some charts
-  ##     # Use variables so that the UI is able parse the expression
+  ##     # Optionally filter out some charts.
+  ##     # The jq query format is not exposed in the UI, so care needs to be taken to use the format which the UI expects to parse, 
+  ##     # which is why variables are used in the example below.
   ##     filterRule:
   ##       jq: .name == $var0 or .name == $var1
   ##       variables:
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -729,6 +729,13 @@ apprepository:
   ##     ociRepositories:
   ##       - nginx
   ##       - jenkins
+  ##     # Optionally filter out some charts
+  ##     # Use variables so that the UI is able parse the expression
+  ##     filterRule:
+  ##       jq: .name == $var0 or .name == $var1
+  ##       variables:
+  ##         $var0: nginx
+  ##         $var1: jenkins
   ##
   initialRepos:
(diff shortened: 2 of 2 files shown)


chart/kubeapps/values.yaml
```diff
@@ -729,6 +729,13 @@ apprepository:
   ##     ociRepositories:
   ##       - nginx
   ##       - jenkins
+  ##     # Optionally filter out some charts
+  ##     # Use variables so that the UI is able parse the expression
```
absoludity, line 733: Thanks @dud225 . Can you explain what you mean here? Haven't checked, but didn't think the UI was presenting the filter rules atm. This was just if you manually create/update an `AppReository` via the CLI, and you're adding here the ability to specify a filter rule in the config (great). Does the [example in the docs](https://github.com/kubeapps/kubeapps/blob/master/docs/user/private-app-repository.md#advanced-filtering) (which does not use vars) cause the UI to bail with a parsing error? (`    jq: .maintainers | any(.name == "Bitnami") `. ie. does `jq: .name == "nginx" or .name == "jenkins"` cause a parsing error?

Otherwise, +1 from me (sorry about CI, I'm hoping to get to that today).

absoludity: I should have read your related issue first. But yes, that's a bug if the previously defined filter is cleared. Still unsure about the above comment regarding variables.

dud225: Hello
Yes sorry my comment is a bit blurry and doesn't actually relate to https://github.com/kubeapps/kubeapps/issues/3447.

I noticed that creating an AppRepository with filters set from the UI (I remember that the UI already presented them in v2.3.3):
Capture d’écran de 2021-09-21 09-09-26
yields the following rule:
```yaml
  filterRule:
    jq: .name == $var0 or .name == $var1
    variables:
      $var0: nginx
      $var1: jenkins
```
but if I edit it to just look as follows:
```yaml
  filterRule:
    jq: .name == "nginx" or .name == "jenkins"
```
then after refreshing Kubeapps the filter vanishes from the UI:
Capture d’écran de 2021-09-21 09-13-12
even though the filter is sent by the server:
Capture d’écran de 2021-09-21 09-21-50

Reverting the change makes the filter to appear again, so Kubeapps is definitely expecting some particular format in order for the UI to present the filter. 

So I may be missing something or this may be due to the fact that the UI can't support all variations of the filter rule, however I believe that some insights should be provided to the users to allow them to keep the UI in-sync with the CRD.

absoludity: Right, thanks. That gives me more context of the comment. Could you pls update the comment to something like: "The jq query format is not exposed in the UI, so care needs to be taken to use the format which the UI expects to parse, which is why variables are used in the example below."

Also, can you please merge or rebase with master (there was an issue with CI) then this should pass and I'm happy to land. Thanks for the improvement!

absoludity (approved):
Thanks!

chart/kubeapps/values.yaml
dud225: Done. Thanks for your suggestion that clarifies my comment.

chart/kubeapps/templates/apprepository/apprepositories.yaml
```diff
@@ -60,5 +60,9 @@ spec:
         name: {{ printf "apprepo-%s-secrets" .name }}
     {{- end }}
   {{- end }}
+  {{- if .filterRule }}
+  filterRule:
+    {{- toYaml .filterRule | nindent 4 }}
+  {{- end }}
```
antgamdia, line 66: Just an internal node, it would require an increase in the patch version of the chart, but we will do when syncing with the official Bitnami repo.

antgamdia (approved):
Great! Thanks for the contribution :)

Issue #3429, opened 2021-09-16
Add `--version` to all of our binaries


antgamdia:
Currently, our command line-based binaries do support `--version` command. However, the version is not being properly passed.
As required by https://github.com/kubeapps/kubeapps/issues/3143 (which is being executed by some internal colleagues on their side), we need to add this command to ensure every image is released properly.

Ongoing: 
- [x]  https://github.com/kubeapps/kubeapps/pull/3440
- [x]  https://github.com/kubeapps/kubeapps/pull/3441
- [x]  https://github.com/kubeapps/kubeapps/pull/3449
- [x]  https://github.com/kubeapps/kubeapps/pull/3450
- [x]  https://github.com/kubeapps/kubeapps/pull/3451
- [x]  https://github.com/kubeapps/kubeapps/pull/3452
- [x]  kubeapps-dashboard (not a binary, but it's available under dashboard/public/config.json, `appVersion`)
 

Pending: 

**Important -for content team-**: the changes made in the dockerfiles have to be extrapolated to the bitnami pipeline.

antgamdia:
Everything is done on our side, however, as stated above, the build flags thing is something yet to be implemented in the Bitnami end.

PR #3452, opened 2021-09-20
Add versions - pinniped-proxy

antgamdia:
### Description of the change

This PR adds the `--version` command to `pinniped-proxy`

### Benefits

Our binaries will properly display their version

### Possible drawbacks

N/A

### Applicable issues

- related #3429

### Additional information

I have been also exploring [vergen](https://docs.rs/vergen/5.1.15/vergen/index.html), but it gives us much more build/git/etc env information than we need.

```
IMAGE_TAG=dev make kubeapps/pinniped-proxy && docker run  docker.io/kubeapps/pinniped-proxy:dev   --version
```

### DIFF
--- a/cmd/pinniped-proxy/build.rs
+++ b/cmd/pinniped-proxy/build.rs
@@ -1,0 +1,9 @@
+use std::env;
+
+fn main() {
+    // If the binary is built with the ENV var "PINNIPED_PROXY_VERSION",
+    // the value will be available at buildime. Otherwise, it becomes "devel"
+    // We use this ENV var to display a custom version when passing the "-- version" flag
+    let foo = env::var("PINNIPED_PROXY_VERSION").unwrap_or("devel".to_string());
+    println!("cargo:rustc-env=PINNIPED_PROXY_VERSION={}", foo);
+}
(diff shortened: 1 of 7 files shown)


cmd/pinniped-proxy/build.rs
```diff
@@ -0,0 +1,9 @@
+use std::env;
+
+fn main() {
+    // If the binary is built with the ENV var "PINNIPED_PROXY_VERSION",
+    // the value will be available at buildime. Otherwise, it becomes "devel"
+    // We use this ENV var to display a custom version when passing the "-- version" flag
+    let foo = env::var("PINNIPED_PROXY_VERSION").unwrap_or("devel".to_string());
```
absoludity, line 7: s/foo/version/ ?

And nice, hadn't read about `bulid.rs` - interesting interface to have for the build environment.

absoludity (approved):
Guessed you intended a different var name, but otherwise +1.

cmd/pinniped-proxy/build.rs
antgamdia: Yep, right.

It is really interesting (https://doc.rust-lang.org/cargo/reference/build-scripts.html), it seems to be very promising. I still have to take a closer look a the docs, just to see if there is something else we can leverage from.

PR #3463, opened 2021-09-21
Add initial createInstalledPkg from UI

antgamdia:
### Description of the change

In this PR we switch from the kubeops endpoint to the new API for installing new packages.

### Benefits

We will use the new service and will be able to sunset (at some point) kubeops

### Possible drawbacks

N/A, but let see if CI thinks otherwise

### Applicable issues

- related #3340

### Additional information

https://user-images.githubusercontent.com/11535726/134319758-453f5238-130e-4e07-8037-5ef55adc6546.mp4


dashboard/src/actions/apps.ts
```diff
@@ -173,9 +175,20 @@ export function deployChart(
           );
         }
       }
-
-      await App.create(targetCluster, targetNamespace, releaseName, availablePackageDetail, values);
-      dispatch(receiveDeployApp());
+      if (
+        availablePackageDetail?.availablePackageRef &&
+        availablePackageDetail?.version?.pkgVersion
+      ) {
```
absoludity, line 181: Looks like this is silently doing nothing if either the availablePackageRef or the pkgVersion are unset? The first seems like an error condition, in that we should never have an `availablePackageDetail` without the reference - and if we do, it's an error. The second similarly seems like an error in that the version for an `availablePackageDetail` should always be specified (and if that's not true, we could default to the latest if blank, in the backend, if we wanted).

So why have this condition at all?  If there is an error, it'll be caught and dispatched below?

absoludity (approved):
Great. One small comment, but +1 either way.

dashboard/src/actions/apps.ts
antgamdia: You're right, it wasn't being handled, by bad. 
It is what I told you (offline) about: https://github.com/stephenh/ts-proto/issues/288

Since some of the objects can be undefined, the typescript type check complains, so, we either: i) add this `if` condition properly handled or ii) we force the type with `!`. The latter will fail in runtime, but it will be caught by the try catch block.

I'm going with i) since it forces us to make more meaningful error messages and the code is more self-explanatory that just relying on a `!` symbol.

antgamdia:
I'm re-targeting the PR to hit a new upstream branch (ui-newapi-ops) where I plan to land the UI changes before going to the main branch.

PR #3468, opened 2021-09-22
fix flux plugin UpdateInstalledPackage() according to agreed upon semantics

gfichtenholt:


cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/integration_utils_test.go
```diff
@@ -0,0 +1,337 @@
...
+		var fluxPluginClient fluxplugin.FluxV2PackagesServiceClient
+		var err error
+		if fluxPluginClient, err = getFluxPluginClient(t); err != nil {
+			t.Fatalf("Failed to get fluxv2 plugin due to: [%v]", err)
+		}
+		rand.Seed(time.Now().UnixNano())
+		return fluxPluginClient
+	}
+	return nil
+}
```
absoludity, line 67: Thanks - yes, now it explicitly fails if someone tries to enable them but hasn't got the kind cluster running etc.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/integration_utils_test.go
```diff
@@ -0,0 +1,337 @@
...
+	for i := range b {
+		b[i] = letters[rand.Intn(len(letters))]
+	}
+	return string(b)
+}
+
+// global vars
+var letters = []rune("abcdefghijklmnopqrstuvwxyz0123456789")
+var typedClient kubernetes.Interface
+var dynamicClient dynamic.Interface
```
absoludity, line 337: Do we need global vars? Normally it should be possible to avoid this type of state (your kubeGet* fns above already return the clients).

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/release_test.go
```diff
@@ -800,6 +629,208 @@ func TestCreateInstalledPackage(t *testing.T) {
...
+				redis_existing_spec_completed,
+			},
+			expectedStatusCode: codes.OK,
+			expectedResponse: &corev1.UpdateInstalledPackageResponse{
+				InstalledPackageRef: my_redis_ref,
+			},
+			expectedRelease: flux_helm_release_updated_1,
+		},
+		{
+			name: "returns invalid if installed package doesn't exist",
```
absoludity, line 659: s/invalid/not found/ ?

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
```diff
@@ -50,7 +50,7 @@ service PackagesService {
 
   rpc UpdateInstalledPackage(UpdateInstalledPackageRequest) returns (UpdateInstalledPackageResponse) {
      option (google.api.http) = {
-       patch: "/core/packages/v1alpha1/installedpackages"
+       put: "/core/packages/v1alpha1/installedpackages"
```
absoludity, line 53: Nice, so there will be no problem with us later providing a `patch` method here either (though I understand the message formats will need references throughout).

absoludity (approved):
Great, thanks Greg.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/integration_utils_test.go
gfichtenholt: not sure I get what you're asking. What's the alternative to global vars if I want to keep some state in between calls (other than to have to declare a new receiver struct or explicitly pass in the state back and forth, yuck)? Must be some language feature I am not aware of. Please tell me.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/release_test.go
gfichtenholt: yes, indeed. I copied that case from your test, so you might want to correct that one. hehe

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
gfichtenholt: yes, right on

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/integration_utils_test.go
absoludity: I'm not asking you to change anything here, but worth reading: https://dave.cheney.net/2017/06/11/go-without-package-scoped-variables

gfichtenholt: Sure, I'll give a read. Just so we are on the same page, this is test code we are talking about, not production. Furthermore, this is test code that you've pointed out will become obsolete at some point not too far away. So going an extra mile (or in this case an extra 10 miles based on what I read of the article so far) and taking a shortcut seemed like an acceptable thing to do for that case. If you find any cases where I am doing it in production code, please point them out

absoludity: Yep, as above, nothing to change here. Just worth being aware, that's all. I like the last line ("the bottom line") of Dave's post.

PR #3471, opened 2021-09-22
Add core implementations for Create, Update, Delete.

absoludity:
### Description of the change

Adds core implementations for Create, Update and Delete installed packages.

### Benefits

Dashboard can use the core client instead of the specific helm one (so we can test other plugins like flux).

### Applicable issues

- fixes #


antgamdia (approved):
Great!! Thank you!

PR #3473, opened 2021-09-22
Update kubeapps-apis http endpoints to more restful URIs.

absoludity:
### Description of the change

Follow-on from a discussion [on a previous PR](https://github.com/kubeapps/kubeapps/pull/3427#discussion_r709837195) where we discussed more restful URIs for the http endpoints.

Since we're (Kubeapps) not using the http endpoints, it's not been a priority to do the URI endpoints, but as we release, people will start looking at the swagger doc, hence doing this now.

The change updates so that we have "availablepackages" and "installedpackages" as resources for plugins (or core), and updates the detail request for each so the URI is unique.

And that is all :)

### Benefits

URI endpoints are unique so can sit behind caches as needed.

### Possible drawbacks

We've gone with including the cluster in the URL (see previous discussion linked above) which is perfect for Kubeapps, but potential other clients which may not be interested in multi-cluster setups may prefer it as a qparam. But it'd be premature to optimized for potential other clients over Kubeapps right now.

### Applicable issues

- fixes #


cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
```diff
@@ -15,29 +15,29 @@ import "protoc-gen-openapiv2/options/annotations.proto";
 service PackagesService {
   rpc GetAvailablePackageSummaries(GetAvailablePackageSummariesRequest) returns (GetAvailablePackageSummariesResponse) {
      option (google.api.http) = {
-       get: "/core/packages/v1alpha1/availablepackagesummaries"
+       get: "/core/packages/v1alpha1/availablepackages"
```
antgamdia, line 18: Shouldn't this URL also be cluster/ns-dependant? Or we are just gonna encode them via query params?
I mean, it's okay in that the rest of the URIs have the `c` and `ns` under the `plugin`, so there isn't any semantic issue.
Besides, they are optional, so not passing as query params would imply multiplexing the endpoint in: `.../availablepackages`, `.../availablepackages/c/<cluster>` and `.../availablepackages/c/<cluster>/ns/<ns> and I don't know if it is possible.

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
```diff
@@ -15,29 +15,29 @@ import "protoc-gen-openapiv2/options/annotations.proto";
...
     option (google.api.http) = {
-      get: "/core/packages/v1alpha1/availablepackageversions"
+      get: "/core/packages/v1alpha1/availablepackages/plugin/{available_package_ref.plugin.name}/{available_package_ref.plugin.version}/c/{available_package_ref.context.cluster}/ns/{available_package_ref.context.namespace}/{available_package_ref.identifier}/versions"
     };
   }
 
   rpc GetInstalledPackageSummaries(GetInstalledPackageSummariesRequest) returns (GetInstalledPackageSummariesResponse) {
     option (google.api.http) = {
-      get: "/core/packages/v1alpha1/installedpackagesummaries"
+      get: "/core/packages/v1alpha1/installedpackages"
```
antgamdia, line 34: Same comment as in `availablepackages`

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
```diff
@@ -50,14 +50,14 @@ service PackagesService {
...
-       put: "/core/packages/v1alpha1/installedpackages"
+       put: "/core/packages/v1alpha1/installedpackages/plugin/{installed_package_ref.plugin.name}/{installed_package_ref.plugin.version}/c/{installed_package_ref.context.cluster}/ns/{installed_package_ref.context.namespace}/{installed_package_ref.identifier}"
        body: "*"
       };
   }
 
   rpc DeleteInstalledPackage(DeleteInstalledPackageRequest) returns (DeleteInstalledPackageResponse) {
     option (google.api.http) = {
-      delete: "/core/packages/v1alpha1/installedpackages"
+      delete: "/core/packages/v1alpha1/installedpackages/plugin/{installed_package_ref.plugin.name}/{installed_package_ref.plugin.version}/c/{installed_package_ref.context.cluster}/ns/{installed_package_ref.context.namespace}/{installed_package_ref.identifier}"
```
antgamdia, line 60: Great, thanks! :smile:

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
```diff
@@ -50,14 +50,14 @@ service PackagesService {
 
   rpc UpdateInstalledPackage(UpdateInstalledPackageRequest) returns (UpdateInstalledPackageResponse) {
      option (google.api.http) = {
-       put: "/core/packages/v1alpha1/installedpackages"
+       put: "/core/packages/v1alpha1/installedpackages/plugin/{installed_package_ref.plugin.name}/{installed_package_ref.plugin.version}/c/{installed_package_ref.context.cluster}/ns/{installed_package_ref.context.namespace}/{installed_package_ref.identifier}"
```
antgamdia, line 53: wrt the discussion in the former PR, here we are sending the full resource representation through the wire, so `PUT` is fine. I'm just wondering, do you know if grpc-gateway also supports `PATCH` requests?

antgamdia (approved):
Great! Thanks for performing the changes. Happy to finally see kubeappsapis becoming more RESTish :P 
Some minor questions, but +1ing anyway.

Also, at some point, we have to think about integrating these swagger docs into the file being served in the frontend.

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto
absoludity: Yes, for the list endpoints I've left them as qparams since they are optional. The only way I see for multiplexing different URLs for the same functionality is by using different RPC function names (ie. `GetAvailablePackageSummariesPerNamespace` with its own get, with the grpc function just calling the other one... not pretty).

gfichtenholt (commented):
`get: "/core/packages/v1alpha1/availablepackages/plugin/{available_package_ref.plugin.name}/{available_package_ref.plugin.version}/c/{available_package_ref.context.cluster}/ns/{available_package_ref.context.namespace}/{available_package_ref.identifier}"
`
ugh. gru said it [best](https://www.youtube.com/watch?v=GmZiNLRVRL4)
not super constructive, I know. Give me a few hours or a day to come to terms with this

Just so I am clear a concrete example for flux might look like this:
`GET "http://localhost:8080/core/packages/v1alpha1/availablepackages/plugin/fluxv2.packages/v1alpha1/c//ns/default/bitnami%2Fredis"`

Note empty string for cluster parameter. Is this correct?

absoludity:
> `get: "/core/packages/v1alpha1/availablepackages/plugin/{available_package_ref.plugin.name}/{available_package_ref.plugin.version}/c/{available_package_ref.context.cluster}/ns/{available_package_ref.context.namespace}/{available_package_ref.identifier}" `
> ugh. gru said it [best](https://www.youtube.com/watch?v=GmZiNLRVRL4)
> not super constructive, I know. Give me a few hours or a day to come to terms with this

Heh - yeah. As mentioned in earlier conversations, ideally we'd just have `/core/packages/v1alpha1/availablepackages/{uid}` but we don't have a unique id for an available or installed package. By design we don't have separate storage, so we need to uniquely identify them by their fields. I did think about using an encoding of the same info, but then you've got semantic IDs (and not sure if we can pull the info out easily in the handlers - certainly can't auto-populate the fields.

> 
> Just so I am clear a concrete example for flux might look like this:
> `GET "http://localhost:8080/core/packages/v1alpha1/availablepackages/plugin/fluxv2.packages/v1alpha1/c//ns/default/bitnami/redis"`
> 
> Note empty string for cluster parameter. Is this correct?

Yes, a call to the core API for a specific flux `redis` available package in the default namespace, without specifying the cluster, would look like that. But for our purposes (the Kubeapps UI), the cluster field will always be present in the request. We could also just make the cluster field required in our RPC implementations instead (since it will be there for all requests from Kubeapps itself).

gfichtenholt:
> > `get: "/core/packages/v1alpha1/availablepackages/plugin/{available_package_ref.plugin.name}/{available_package_ref.plugin.version}/c/{available_package_ref.context.cluster}/ns/{available_package_ref.context.namespace}/{available_package_ref.identifier}" `
> > ugh. gru said it [best](https://www.youtube.com/watch?v=GmZiNLRVRL4)
> > not super constructive, I know. Give me a few hours or a day to come to terms with this
> 
> Heh - yeah. As mentioned in earlier conversations, ideally we'd just have `/core/packages/v1alpha1/availablepackages/{uid}` but we don't have a unique id for an available or installed package. By design we don't have separate storage, so we need to uniquely identify them by their fields. I did think about using an encoding of the same info, but then you've got semantic IDs (and not sure if we can pull the info out easily in the handlers - certainly can't auto-populate the fields.
> 
> > Just so I am clear a concrete example for flux might look like this:
> > `GET "http://localhost:8080/core/packages/v1alpha1/availablepackages/plugin/fluxv2.packages/v1alpha1/c//ns/default/bitnami/redis"`
> > Note empty string for cluster parameter. Is this correct?
> 
> Yes, a call to the core API for a specific flux `redis` available package in the default namespace, without specifying the cluster, would look like that. But for our purposes (the Kubeapps UI), the cluster field will always be present in the request. We could also just make the cluster field required in our RPC implementations instead (since it will be there for all requests from Kubeapps itself).

Well, shucks. 
Ideally, this would have been perfect for flux:
`GET "http://localhost:8080/core/packages/v1alpha1/plugin/fluxv2.packages/namespaces/default/available/bitnami%Fredis"`
Notice:
1. Only one version 
2. No cluster path parameter (does not apply to the plug-in)
3. re-arranged where "availablepackages" appears in the path
This would be consistent with k8s API, like https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#read-job-v1-batch

Now, I understand perfectly well the reasons why we have 1, 2 and 3. If we really can't do better, that so be it. Its only aesthetics, not functionality, after all (but so is most of REST, I feel). So I will give it a +1 and let you make that call :-)

gfichtenholt (approved):
+1. Comments already left

absoludity:
Great, thanks Greg.

PR #3487, opened 2021-09-23
Add helm-only implementation of RollbackInstalledPackage.

absoludity:
### Description of the change

Adds a `RollbackInstalledPackage` rpc to the helm plugin only.

This is one of the remaining functions missing to switch the helm functionality in the current UI.

It provides a nice example of extending a plugin with custom functionality not available in the core API without hindering the plugins ability to satisfy the core API.

### Benefits

The helm client can be used in the UI to do rollbacks.

### Possible drawbacks

If, in the future, other packaging systems support a rollback, then we'd need to consider moving this to the core API perhaps, or just having both plugins with custom functionality.

### Applicable issues

- rel #3340


antgamdia (approved):
Nice! Thanks for implementing this!! Can't wait to have all the operations being invoked by the new API server instead of kubeops!

PR #3488, opened 2021-09-23
Document pinnipedConfig in values.yaml. Minor fix in Pinniped docs

antgamdia:
### Description of the change

This minor PR just documents the `pinnipedConfig: true` values under the `clusters` section in our chart and removes a no longer useful warning about the required pinniped version.

### Benefits

Better docs

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

N/A

### DIFF
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -1464,7 +1464,7 @@ allowNamespaceDiscovery: true
 ## - isKubeappsCluster is an optional parameter that allows defining the cluster in which Kubeapps is installed;
 ##   this param is useful when every cluster is using an apiServiceURL (e.g., when using the Pinniped Impersonation Proxy)
 ##   as the chart cannot infer the cluster on which Kubeapps is installed in that case.
-## - pinnipedConfig is an optional parameter that allows defining the cluster running Pinniped
+## - pinnipedConfig is an optional parameter that contains configuration options specific to a cluster running the pinniped concierge service.
 ## e.g.:
 ## clusters:
 ## - name: default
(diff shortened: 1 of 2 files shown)


chart/kubeapps/values.yaml
```diff
@@ -1475,6 +1475,8 @@ allowNamespaceDiscovery: true
 ##   certificateAuthorityData: LS0tLS1CRUdJ...
 ##   serviceToken: ...
 ##   isKubeappsCluster: true
+##   pinnipedConfig: true
```
absoludity, line 1478: But it's not a boolean value, rather an object?
```suggestion
##   pinnipedConfig:
##     enable: true
```

chart/kubeapps/values.yaml
```diff
@@ -1465,6 +1464,7 @@ allowNamespaceDiscovery: true
 ## - isKubeappsCluster is an optional parameter that allows defining the cluster in which Kubeapps is installed;
 ##   this param is useful when every cluster is using an apiServiceURL (e.g., when using the Pinniped Impersonation Proxy)
 ##   as the chart cannot infer the cluster on which Kubeapps is installed in that case.
+## - pinnipedConfig is an optional parameter that allows defining the cluster running Pinniped
```
absoludity, line 1467:
```suggestion
## - pinnipedConfig is an optional parameter that contains configuration options specific to a cluster running the pinniped concierge service.
```

absoludity (approved):
Small change required as the `pinnipedConfig` field is an object, not boolean, but otherwise +1

chart/kubeapps/values.yaml
antgamdia: Oops, my bad, I did it quickly and overlapped `pinnipedConfig` and `enable: true` in my mind. Thanks for fixing it!

PR #3493, opened 2021-09-24
Don't rely on randomness for test names.

absoludity:
### Description of the change

Saw the following failure in CI:
07-upgrade-0

We should not be relying on random names to avoid conflict between the tests. Updates to use a release name based on the running test. I've left the random suffix so that if we ever have the same tests running concurrently in the same cluster for different Kubeapps versions, we won't immediately fail.

### Benefits

CI won't fail due to name clashes.

### Possible drawbacks

None.

### Applicable issues

- fixes #

### DIFF
--- a/integration/use-cases/01-add-multicluster-deployment.js
+++ b/integration/use-cases/01-add-multicluster-deployment.js
@@ -33,7 +33,7 @@ test("Deploys an application with the values by default", async () => {
   );
 
   await expect(page).toMatchElement("#releaseName", { text: "" });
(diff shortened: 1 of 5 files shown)


antgamdia (approved):
Great! Thanks for doing it. Sometimes I saw some related errors, but given that the collision chances weren't too high, I never did it, but we should.
Let's see if we decrease the ci errors, otherwise, perhaps it worths rewriting the tests for a new framework, it's frustrating so many failures haha

antgamdia:
Merging. Let's see if CI fail rate improves :)

PR #3483, opened 2021-09-22
Update go deps. Clean up unused code

antgamdia:
### Description of the change

This PR is to update the `helm` dependency to the latest 3.7.0 version. It implies, we no longer have conflicts with related deps like `containerd` and `oras`.
This helm dep has played a crucial role in kubeapps historically, wherefore the existence of some old code, still dependant from the helmv2 code. In this PR we also address that problem and removing the old code and migrating the old references to Helmv2 to the Helm3 equivalent (mainly: the `Maintainers` field or some minor changes in how to access to some properties).

### Benefits

We will no longer depend upon old dependencies. Specially, is remarkable that with this change we will get rid of a number of securities advisories, namely: `GHSA-56hp-xqp3-w2jf` (this one addressed in our code in the past) and `GHSA-c72p-9xmj-rx3w` from containerd (with no impact on our code, though, afaik).

Note: there is still the k8s CVEs (CVE-2020-8561 and CVE-2021-25740, but they do not affect Kubeapps - one is related to the AdmissionController's webhooks and the other was just a wrong script in their repo, but it hasn't been shipped in any artifact)

### Possible drawbacks

Let's see how it goes when the CI executes its tests... I don't foresee any problem since all our tests are passing, but I haven't performed yet a manual check.

### Applicable issues

- related #3489
- closes #3086

### Additional information

There is still yet another PR (depending on this one) for just performing a typical update of patch versions as part of our release process.

PR #3498, opened 2021-09-24
Remove unused code

antgamdia:
### Description of the change

Due to the memory issues faced in https://github.com/kubeapps/kubeapps/pull/3483#issuecomment-926699100, this PR aims at solving just a single: removing the old references to Helm2 as well as other unused code. 
This action is required before we update the go deps to avoid deps conflicts.

### Benefits

The root of the memory issue perhaps will arise.

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

This is not the same code as in #3483; I've replaced `response.NewDataResponse(rel).Write(w)` with `response.NewDataResponse(*rel).Write(w)`, since I'm thinking it could be the cause. Not verified, though, just sending the PR to see if it is OOMKilled.


absoludity (approved):
Excellent. Glad to see references to proxy gone.

Issue #3489, opened 2021-09-23
Update dependencies prior to the release


antgamdia:
As part of our release process, we usually bump the versions up. However, in this case, we have some major upgrades in the golang deps.

PR #3479, opened 2021-09-22
Use new kubeappsapis to perform the package actions 

antgamdia:
### Description of the change

In this PR we switch from the kubeops endpoint to the new API performing the actions (create, upgrade, delete and rollback)

### Benefits

We will use the new service and will be closer to be able to sunset (at some point) kubeops

### Possible drawbacks

N/A

### Applicable issues

- closes #3477

### Additional information

https://user-images.githubusercontent.com/11535726/134525785-f77ef68b-3888-4c09-b74a-1537755331f6.mp4

This PR is the aggregation of:

- [x] https://github.com/kubeapps/kubeapps/pull/3463
- [x] https://github.com/kubeapps/kubeapps/pull/3475
- [x] https://github.com/kubeapps/kubeapps/pull/3476
- [x] https://github.com/kubeapps/kubeapps/pull/3490

PR #3501, opened 2021-09-25
Update go dependencies

### LINKED ISSUE #3086: Upgrade to use k8s 1.21 client-go
There are numerous dependencies (eg. Helm) that we're not updating as they depend on 1.21.

antgamdia:
### Description of the change

This PR is to update the `helm` dependency to the latest 3.7.0 version. It implies, we no longer have conflicts with related deps like `containerd` and `oras`.
This helm dep has played a crucial role in kubeapps historically, wherefore the existence of some old code, still dependant from the helmv2 code. In this PR we also address that problem and removing the old code and migrating the old references to Helmv2 to the Helm3 equivalent (mainly: the `Maintainers` field or some minor changes in how to access to some properties).

### Benefits

We will no longer depend upon old dependencies. Specially, is remarkable that with this change we will get rid of a number of securities advisories, namely: `GHSA-56hp-xqp3-w2jf` (this one addressed in our code in the past) and `GHSA-c72p-9xmj-rx3w` from containerd (with no impact on our code, though, afaik).

Note: there is still the k8s CVEs (CVE-2020-8561 and CVE-2021-25740, but they do not affect Kubeapps - one is related to the AdmissionController's webhooks and the other was just a wrong script in their repo, but it hasn't been shipped in any artifact)

### Possible drawbacks

Let's see how it goes when the CI executes its tests... I don't foresee any problem since all our tests are passing, but I haven't performed yet a manual check.

### Applicable issues

- related #3489
- closes #3086

### Additional information

This PR supersedes https://github.com/kubeapps/kubeapps/pull/3483, as the cleanup process was performed separately to identify the cause of a memory issue in https://github.com/kubeapps/kubeapps/pull/3498


antgamdia:
Still getting a bad gateway, probably because of https://github.com/kubeapps/kubeapps/pull/3483#issuecomment-926699100. Further investigation is required.

antgamdia:
> Still getting a bad gateway, probably because of [#3483 (comment)](https://github.com/kubeapps/kubeapps/pull/3483#issuecomment-926699100). Further investigation is required.

We've got reports telling kubeops is already hitting memory limits (v2.3.2). I've temporarily increased it (https://github.com/kubeapps/kubeapps/pull/3501/commits/f5f5b83ed96f05627e44aca314cd2285aeb27d3f) to see what CI thinks, but let's see once  https://github.com/kubeapps/kubeapps/pull/3479 gets merged.

absoludity:
We may need to progressively update and see where the memory issue is. I assume we can repro this locally so aren't dependent on CI? Given it's already happening in released versions (though not so drastically), perhaps it'd be best not to yet land this change (guessing that's what you're already doing since the release has started) and instead add this as a task for the next iteration?

Issue #3480, opened 2021-09-22
Check old frontend deps


antgamdia:
While upgrading our frontend deps I've noticed there are several not being upgraded in quite a long time.
Perhaps they are deprecated or have been replaced by new ones, so they stopped publishing under this npm module nada.

In a preliminary selection of those whose latest release date is 1y ahead, I've extracted:

### More than a year

- [ ] `fstream`
- [ ] `jsonwebtoken`
- [ ] `normalize.css`
- [ ] `prop-types`
- [ ] `raf`
- [ ] `redux-thunk`
- [ ] `seamless-immutable`

### Under a year

- [ ] `react-diff-viewer`
- [ ] `react-helmet`
- [x] `react-jsonschema-form`:
    - **Action**: replace with `@rjsf/core`
    - **Explaination**: they stopped publishing after v1 and it got renamed to `@rjsf/core`
- [ ] `typesafe-actions`

Perhaps some of them are so simple that they don't require any frequent upgrade, but I think it worths having a look at each of them.

PR #3521, opened 2021-09-29
Remove `react-jsonschema-form` UI dependency

antgamdia:
### Description of the change

As part of the analysis done in https://github.com/kubeapps/kubeapps/issues/3480, this PR is to remove the `react-jsonschema-form` dependency as it appears to be old and unused in our current code.

### Benefits

### Possible drawbacks

### Applicable issues

- related #3480

### Additional information


absoludity (approved):
Conflict, and the same addition of a proxy to your package.json - not sure that's intended for the PR.

PR #3522, opened 2021-09-29
Remove `immutable` and `seamless-immutable` UI dependencies

antgamdia:
### Description of the change

As part of the analysis done in https://github.com/kubeapps/kubeapps/issues/3480, this PR is to remove the `immutable` and `seamless-immutable` dependency as they appear to be old and unused in our current code.

### Benefits

### Possible drawbacks

### Applicable issues

- related #3480

### Additional information

### DIFF
--- a/dashboard/yarn.lock
+++ b/dashboard/yarn.lock
@@ -7343,7 +7343,7 @@ immutable@^3.8.1, immutable@^3.x.x:
   resolved "https://registry.yarnpkg.com/immutable/-/immutable-3.8.2.tgz#c2439951455bb39913daf281376f1530e104adf3"
   integrity sha1-wkOZUUVbs5kT2vKBN28VMOEErfM=
 
-"immutable@^3.8.1 || ^4.0.0-rc.1", immutable@^4.0.0-rc.15:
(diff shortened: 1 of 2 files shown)


dashboard/yarn.lock
```diff
@@ -7343,7 +7343,7 @@ immutable@^3.8.1, immutable@^3.x.x:
   resolved "https://registry.yarnpkg.com/immutable/-/immutable-3.8.2.tgz#c2439951455bb39913daf281376f1530e104adf3"
   integrity sha1-wkOZUUVbs5kT2vKBN28VMOEErfM=
 
-"immutable@^3.8.1 || ^4.0.0-rc.1", immutable@^4.0.0-rc.15:
+"immutable@^3.8.1 || ^4.0.0-rc.1":
```
absoludity, line 7346: So they're both still being used via a dependency rather than directly, but older version.

absoludity (approved):

dashboard/yarn.lock
antgamdia: Yep, transitive deps... :disappointed:

PR #3523, opened 2021-09-29
Remove `raf` UI dependency

antgamdia:
### Description of the change

As part of the analysis done in https://github.com/kubeapps/kubeapps/issues/3480, this PR is to remove the `raf` dependency as it appears to be old and unused in our current code.

### Benefits

### Possible drawbacks

### Applicable issues

- related #3480

### Additional information

### DIFF
--- a/dashboard/src/setupTests.ts
+++ b/dashboard/src/setupTests.ts
@@ -2,7 +2,6 @@ import Adapter from "@wojtekmaj/enzyme-adapter-react-17";
 import Enzyme from "enzyme";
 import "jest-enzyme";
 import { WebSocket } from "mock-socket";
-import "raf/polyfill"; // polyfill for requestAnimationFrame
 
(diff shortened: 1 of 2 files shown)


dashboard/src/setupTests.ts
```diff
@@ -2,7 +2,6 @@ import Adapter from "@wojtekmaj/enzyme-adapter-react-17";
 import Enzyme from "enzyme";
 import "jest-enzyme";
 import { WebSocket } from "mock-socket";
-import "raf/polyfill"; // polyfill for requestAnimationFrame
```
absoludity, line 5: Odd that it was only used in tests for a polyfill. Guess browsers handle this properly now. https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame#browser_compatibility - seems fully supported.

absoludity (approved):

PR #3524, opened 2021-09-29
Replace `normalize` with `modern-normalize` UI dependency

antgamdia:
### Description of the change

As part of the analysis done in https://github.com/kubeapps/kubeapps/issues/3480, this PR is to replace the `normalize` dependency as it appears to be old and clarity moved to `modern-normalize`

### Benefits

### Possible drawbacks

### Applicable issues

- related #3480

### Additional information

### DIFF
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -22,7 +22,7 @@
     "jsonwebtoken": "^8.5.1",
     "lodash": "^4.17.21",
     "lodash-es": "^4.17.21",
-    "normalize.css": "^8.0.1",
+    "modern-normalize": "^1.1.0",
     "prop-types": "^15.7.2",
     "protobufjs": "^6.11.2",
     "qs": "^6.10.1",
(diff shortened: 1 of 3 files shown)


dashboard/package.json
```diff
@@ -22,7 +22,7 @@
     "jsonwebtoken": "^8.5.1",
     "lodash": "^4.17.21",
     "lodash-es": "^4.17.21",
-    "normalize.css": "^8.0.1",
+    "modern-normalize": "^1.1.0",
```
absoludity, line 25: Do we need this direct dependency because of the css inclusion? Otherwise it looks like this is already pulled in as a dependency from react.

absoludity (approved):

dashboard/package.json
antgamdia: Not required, but I think it worths explicitly declare it since we are using it for resetting the style. I mean, what if react drops this dep as we did with normalize.css

PR #3525, opened 2021-09-29
fix(upgrade): Fetch package from its installed cluster

aanthonyrizzo:
### Description of the change

This PR addresses (what I think is a bug) regarding the fetching of Packages when upgrading.

#### Current behavior
When navigating to the upgrade form for deployments that are installed in the `"default"` cluster. The upgrade form fails to fetch the applicable package details if it is looking in the wrong cluster.
image

#### New behavior
When fetching the package details, we use the cluster that is defined in the `AvailablePackageDetail` property when fetching the package. This should lead to correct fetching behavior.

### Benefits

Global packages that are installed in the `""` cluster but are deployed in the `"default"` can now be upgraded.

### Possible drawbacks

I don't believe there are any.

### Applicable issues

I did not create an issue since I was experimenting w/ solutions locally and seemed to have solved it.

### Additional information

I believe this is a relatively impactful bug that is present in https://github.com/kubeapps/kubeapps/commit/05775ff847dcd3bc320f898d9c5963caf4e00d57 and should be addressed before the next major release. I have not pinpointed the exact commit that broke this behavior.

@antgamdia @absoludity for visibility

### DIFF
--- a/dashboard/src/components/AppUpgrade/AppUpgrade.tsx
+++ b/dashboard/src/components/AppUpgrade/AppUpgrade.tsx
@@ -75,7 +75,10 @@ function AppUpgrade() {
     dispatch(
       actions.charts.getDeployedChartVersion(
         {
-          context: { cluster: cluster, namespace: repoNamespace ?? "" },
+          context: {
+            cluster: app?.availablePackageRef?.context?.cluster ?? cluster,
+            namespace: repoNamespace ?? "",
+          },
           identifier: app?.availablePackageRef?.identifier ?? "",
           plugin: app?.availablePackageRef?.plugin,
         } as AvailablePackageReference,
--- a/dashboard/src/components/UpgradeForm/UpgradeForm.tsx
+++ b/dashboard/src/components/UpgradeForm/UpgradeForm.tsx
@@ -81,6 +81,8 @@ function UpgradeForm({
 
   const { availablePackageDetail, versions, schema, values, pkgVersion } = selected;
 
+  const packageCluster = availablePackageDetail?.availablePackageRef?.context?.cluster;
+
   const {
     apps: { isFetching: appsFetching },
     charts: { isFetching: chartsFetching },
@@ -91,12 +93,15 @@ function UpgradeForm({
   useEffect(() => {
     dispatch(
       actions.charts.fetchChartVersions({
-        context: { cluster: cluster, namespace: repoNamespace },
+        context: {
+          cluster: packageCluster ?? cluster,
+          namespace: repoNamespace,
+        },
         plugin: pluginObj,
         identifier: packageId,
       } as AvailablePackageReference),
     );
-  }, [dispatch, cluster, repoNamespace, packageId, pluginObj]);
+  }, [dispatch, packageCluster, repoNamespace, packageId, cluster, pluginObj]);
 
   useEffect(() => {
     if (deployed.values && !modifications) {
@@ -122,7 +127,7 @@ function UpgradeForm({
     dispatch(
       actions.charts.fetchChartVersion(
         {
-          context: { cluster: cluster, namespace: repoNamespace },
+          context: { cluster: packageCluster, namespace: repoNamespace },
           plugin: pluginObj,
           identifier: packageId,
         } as AvailablePackageReference,
@@ -131,7 +136,7 @@ function UpgradeForm({
     );
   }, [
     dispatch,
-    cluster,
+    packageCluster,
     repoNamespace,
     packageId,
     deployed.chartVersion?.version?.pkgVersion,
@@ -160,7 +165,7 @@ function UpgradeForm({
     dispatch(
       actions.charts.fetchChartVersion(
         {
-          context: { cluster: cluster, namespace: repoNamespace },
+          context: { cluster: packageCluster, namespace: repoNamespace },
           plugin: pluginObj,
           identifier: packageId,
         } as AvailablePackageReference,
--- a/dashboard/src/components/UpgradeForm/UpgradeForm.test.tsx
+++ b/dashboard/src/components/UpgradeForm/UpgradeForm.test.tsx
@@ -164,7 +164,10 @@ it("fetches the available versions", () => {
   Chart.getAvailablePackageVersions = getAvailablePackageVersions;
   mountWrapper(defaultStore, <UpgradeForm {...defaultProps} />);
   expect(getAvailablePackageVersions).toHaveBeenCalledWith({
-    context: { cluster: defaultProps.cluster, namespace: defaultProps.repoNamespace },
+    context: {
+      cluster: defaultProps.cluster,
+      namespace: defaultProps.repoNamespace,
+    },
     identifier: defaultProps.packageId,
     plugin: defaultProps.plugin,
   } as AvailablePackageReference);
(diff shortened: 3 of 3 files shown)


dashboard/src/components/UpgradeForm/UpgradeForm.tsx
```diff
@@ -81,6 +81,8 @@ function UpgradeForm({
 
   const { availablePackageDetail, versions, schema, values, pkgVersion } = selected;
 
+  const packageCluster = availablePackageDetail?.availablePackageRef?.context?.cluster;
```
antgamdia, line 84: Since it is a child component depending upon the `AppUpgrade` view, I wonder whether we should get perform the changes in the view instead. 

Given that a `UpgradeForm` will always be rendered if the `app` is defined, why not reducing the number of attributes being passed to this component?

But, perhaps we can just add this simple change and later on, review all the namespace/cluster being passed through. (I've added a note in our project board for having a discussion about it)

Also, the tests are failing, I think (not the e2e, but the dashboard_test).

As usual, thanks a lot for contributing to Kubeapps!

aanthonyrizzo: Ah okay interesting, yeah maybe putting it in the parent component would be best. I can give it another one over when I fix the tests.

mecampbellsoup:
@aanthonyrizzo @antgamdia so looking at the `config.json` response, does "global cluster" correspond to `kubeappsCluster` while our "default" cluster corresponds to `clusters[0]`?

```
{
  "kubeappsCluster": "",
  "kubeappsNamespace": "kubeapps",
  "appVersion": "v2.3.4",
  "authProxyEnabled": true,
  "oauthLoginURI": "/",
  "oauthLogoutURI": "/",
  "authProxySkipLoginPage": true,
  "featureFlags": {"invalidateCache":true,"kubeappsAPIsServer":true},
  "clusters": ["default"],
  "theme": "dark",
  "remoteComponentsUrl": "https://cloud.coreweave.com/static/6befb660ea1eeca2b5e78d452e4a52d9b51aca5e.kubeapps.bundle.js"
}
```

antgamdia:
> @aanthonyrizzo @antgamdia so looking at our `config.json` response, does "global cluster" correspond to `kubeappsCluster` while our "default" cluster corresponds to `clusters[0]`?

You're right. "Global cluster" (perhaps not the best word for it) stands for "the cluster on which Kubeapps is installed". This value is calculated here [1] (that is, the cluster configured without a cluster URL or the one marked with `isKubeappsCluster: true`), and should correspond with one of the `clusters[]`.
I don't think `""` is a proper value for this `kubeappsCluster` property. Perhaps the `"{{ template "kubeapps.kubeappsCluster" . -}}",` in the chart is not working as expected? I'd try using `--dry-run` or ' --debug` when installing the helm chart to ensure this value is being set.

This is my local config, for example:

```json

appVersion: "DEVEL"
authProxyEnabled: false
authProxySkipLoginPage: false
clusters: ["default"]
customAppViews: []
kubeappsCluster: "default"
kubeappsNamespace: "kubeapps"
oauthLoginURI: "/oauth2/start"
oauthLogoutURI: "/oauth2/sign_out"
remoteComponentsUrl: ""
theme: "light"
```

[1] https://github.com/kubeapps/kubeapps/blob/master/chart/kubeapps/templates/_helpers.tpl#L139

aanthonyrizzo:
hey @antgamdia just FYI [`6838601` (#3525)](https://github.com/kubeapps/kubeapps/pull/3525/commits/68386019d43dd9f475712568af1649aebde77a66) fixes a diff detection bug between deployed and the actual chart. This was caused by `deployed` always being `{}`, causing https://github.com/kubeapps/kubeapps/blob/master/dashboard/src/components/UpgradeForm/UpgradeForm.tsx#L102 to never evaluate to true

mecampbellsoup:
Nice so https://github.com/kubeapps/kubeapps/pull/3525 should be good to go now, hopefully that can get merged quickly @antgamdia - thanks for chiming in here!

aanthonyrizzo:
> Awesome! Thanks for fixing it. Even if we try our best to test and ensure everything is working as expected, with this UI revamp to support multiple plugins we have still to polish up some edge cases. I'll try to squeeze a thorough review of the `cluster` and `namespace` params usage in the next iteration.

Yeah I've been poking around and I think there are a few places were some things can be refactored/cleaned up (possibly just renaming some variables and such). It also seems like there is some duplicated data which might be muddying the source of truth and causing some of these little edge case issues. 

> Also FYI, our CI ("just" the e2e tests) is failing right now. This matter needs to be discussed (more info at #3526), meaning the PR merge process (and the release 2.4.1) has to be postponed to the next week (Michael is on PTO this week).

Okay awesome thank you for the update. Let me know if theres anything else I can do

dashboard/src/components/AppUpgrade/AppUpgrade.tsx
```diff
@@ -75,7 +75,10 @@ function AppUpgrade() {
     dispatch(
       actions.charts.getDeployedChartVersion(
         {
-          context: { cluster: cluster, namespace: repoNamespace ?? "" },
+          context: {
+            cluster: app?.availablePackageRef?.context?.cluster ?? cluster,
```
absoludity, line 79: Thanks Anthony. Yes, this should definitely be defaulting to the cluster of the available package, rather than the cluster of the installed app. +1 from me. Is the other discussion below resolved Antonio?

absoludity (approved):

dashboard/src/components/AppUpgrade/AppUpgrade.tsx
antgamdia: Yep, everything is ready to go, I think. Indeed, given how messy the upgrade view was becoming I gave it a shot and try to perform some improvements (https://github.com/kubeapps/kubeapps/pull/3540).
However, the scope of the change is much bigger and I don't think we should release with my PR, but with this simpler and laser-focus one.

Let me jump in and merge the latest changes in the main branch (which hopefully should fix the CI tests...)

aanthonyrizzo: @antgamdia that PR is def gonna be a good improvement. There are a lot of rerenders that end up going all the way down the chain and they _could_ be optimized and improve the performance of the page for sure

antgamdia: > There are a lot of rerenders that end up going all the way down the chain

Yeah... in fact we have been facing some CI issues because of that (the dropdown was being re-rendered multiple times making our e2e engine chose the wrong version)

> Let me jump in and merge the latest changes in the main branch (which hopefully should fix the CI tests...)

I've tried but no permissions to edit this PR, I think. Please, @aanthonyrizzo, pull the latest changes in `master` and merge them into your branch. Let's see if tomorrow we can start tagging a release :P

aanthonyrizzo: AH. Sorry we changed the ownership of the repo from my user to our organization. I just rebased so should be up to date.

antgamdia: Mmm, perhaps it is why the CI is failing when building the images? Not sure, but maybe we should create another PR from this repo?

aanthonyrizzo: Ah that could be. I'm going to be away for a bit but feel free to cherry pick the commits into another branch and merge :)

antgamdia: Mmm, don't know, re-triggering it made the trick. CI things...
Merging! Thanks again for the contributions :)

PR #3527, opened 2021-09-30
Add isLegacySchema to support both Kubeapps schema and openapi schema

### LINKED ISSUE #3526: Kubeapps forms no longer compatible with "OpenAPI Schema Object" schemas
Since a few hours ago, our upgrade/rollback tests started failing for no specific reason. It is probably an upstream issue related to the latest Apache chart used in our test.

antgamdia:
### Description of the change

As per the investigation in #3526, new charts using the [readme-generator](https://github.com/bitnami-labs/readme-generator-for-helm) tool will get an autogenerated version of the schema which is no longer compatible with the kubeapps forms.

This PR addresses (or workaround) this issue by just identifying which schema are we using (right now: "having $schema means the old one")  and decide the logic to render the form elements,
Whereas in the legacy mode only properties having `form: true` will be displayed, in a normal JSONSchema this prop does not exist, so the default behavior is showing every property (unless manually marked as `form: false`).

Note this is a really naive approach and perhaps other solutions might come up but: i) we need to discuss internally which is the roadmap and the horizon of changes ahead; ii) one of the first chart updates was Apache... which the one used in our e2e tests, wherefore they are failing since yesterday (even the main branch)

### Benefits

Ci will work again. Charts using plain jsonschemas will get a form in kubeapps for each of the properties it defines.

### Possible drawbacks

We might not want to display every property, but I don't know the proper way to add this `form:false`, `hidden, for want of a better word, to a canonic JSONSchema file while ensuring its validity (that is, ideally we would want this property to be added by the readme-generator tool just using an annotation).

### Applicable issues

- fixes #3526

### Additional information

As already said, we need to have an internal discussion on what do to in the future. Because... if JSONSchema is the way to go, perhaps we can delegate all the current logic to a 3rd party dep. Note that the loading and UI interaction times are becoming annoying as the number of fields increases.

Let's see if CI agrees and passes with these changes

### DIFF
--- a/dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/Param.tsx
+++ b/dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/Param.tsx
@@ -162,7 +162,7 @@ export default function Param({
   } else {
     const label = param.title || param.path;
     let inputType = "string";
-    if (type === "integer") {
+    if (type === "integer" || type === "number") {
       inputType = "number";
     }
     if (
--- a/dashboard/src/shared/schema.ts
+++ b/dashboard/src/shared/schema.ts
@@ -16,6 +16,7 @@ nullOptions.nullStr = "";
 // It uses the raw yaml to setup default values.
 // It returns a key:value map for easier handling.
 export function retrieveBasicFormParams(
+  isLegacySchema: boolean,
   defaultValues: string,
   schema?: JSONSchemaType<any>,
   parentPath?: string,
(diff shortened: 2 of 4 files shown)


dashboard/src/components/DeploymentFormBody/BasicDeploymentForm/Param.tsx
```diff
@@ -162,7 +162,7 @@ export default function Param({
   } else {
     const label = param.title || param.path;
     let inputType = "string";
-    if (type === "integer") {
+    if (type === "integer" || type === "number") {
```
antgamdia, line 165: https://spec.openapis.org/oas/v3.0.2#dataTypeFormat

dashboard/src/shared/schema.ts
```diff
@@ -156,3 +200,7 @@ export function validate(
   const valid = ajv.validate(schema, yaml.load(values));
   return { valid: !!valid, errors: ajv.errors };
 }
+
+export function isLegacySchema(schema?: JSONSchemaType<any> | any): boolean {
+  return !!schema?.["$schema"] && schema["$schema"] !== "";
```
antgamdia, line 205: Since the new format is an OpenAPI Schema Object (as per v3.0.X, not 3.1.x) instead of a JSON schema (well, it is a subset of it), it does not include `$schema`, but perhaps it's not the best way to distinguish between them.
Open to suggestions here.

dashboard/src/shared/schema.ts
```diff
@@ -28,29 +29,42 @@ export function retrieveBasicFormParams(
...
+          params = params.concat(
+            retrieveBasicFormParams(
+              isLegacySchema,
+              defaultValues,
+              properties[propertyKey],
+              `${itemPath}/`,
+            ),
+          );
+          // Unless there is a form property with a false value, we will render every field in the schema
+        } else if (form !== false) {
```
antgamdia, line 65: We're supposing way too much... an OpenAPI Schema won't have any `form` property. But it does support extensions, so it could eventually become `x-kubeapps-form: true` and it would be openapi-compliant.

antgamdia: More details at https://github.com/bitnami-labs/readme-generator-for-helm/issues/15

absoludity:
> As per the investigation in #3526, new charts using the [readme-generator](https://github.com/bitnami-labs/readme-generator-for-helm) tool will get an autogenerated version of the schema which is no longer compatible with the kubeapps forms.

Not only not compatible, but it removed information (rendering options, conditional display etc.) Would have been great to work together on transitioning this :/

> 
> This PR addresses (or workaround) this issue by just identifying which schema are we using (right now: "having $schema means the old one") and decide the logic to render the form elements,

Excellent. If that's a reliable determinant, then this allows us to continue supporting existing users who have created their own charts with simple form support (which we need to ensure).

> Whereas in the legacy mode only properties having `form: true` will be displayed, in a normal JSONSchema this prop does not exist, so the default behavior is showing every property (unless manually marked as `form: false`).

Not sure about this. It seems that without a transition, we've lost most of the data that makes this feature useful. Displaying all options from the values without any rendering hints (eg. "textArea" or "slider") is, imo, less useful than the straight values.yaml (which has lots of annotated info and help).

> 
> Note this is a really naive approach and perhaps other solutions might come up but: i) we need to discuss internally which is the roadmap and the horizon of changes ahead; ii) one of the first chart updates was Apache... which the one used in our e2e tests, wherefore they are failing since yesterday (even the main branch)

I think there are two other options here to get CI passing:

1. Update to no longer check the simple form when deploying/upgrading the apache chart, or,
2. Update our CI tests to *include* the tarball of the apache (or otherwise) chart that we test.

The second option has the advantage that we continue to test the current simple form support, even though the bitnami apache chart no longer supports it, but has the disadvantage that we no longer test the latest bitnami apache chart.

We could have the best of both worlds by switching to (1) for our current CI tests and adding a single 09-test-simple-form test for verifying the simple form rendering in CI.

> ### Benefits
> 
> Ci will work again.

Excellent.

> Charts using plain jsonschemas will get a form in kubeapps for each of the properties it defines.

Not sure as above.

> ### Possible drawbacks
> 
> We might not want to display every property, but I don't know the proper way to add this `form:false`, `hidden, for want of a better word, to a canonic JSONSchema file while ensuring its validity (that is, ideally we would want this property to be added by the readme-generator tool just using an annotation).

Exactly. So right now, I think we should just ensure *we* don't break the existing support (by perhaps adding a CI test with a custom chart rather than the bitnami apache), and not render a form at all for the new charts in the catalog until we have a better mid-term solution together with the bitnami content team.

absoludity:
Ah - I see. But that would mean releasing with a backwards step (no more simple form support for key charts in the bitnami catalog). So perhaps we want to further delay the release to ensure we have in place support for the new solution too. Let's talk about it in our standup.

absoludity:
> Ah - I see. But that would mean releasing with a backwards step (no more simple form support for key charts in the bitnami catalog). So perhaps we want to further delay the release to ensure we have in place support for the new solution too. Let's talk about it in our standup.

Actually, that point is moot... the metadata was removed from the catalog, and is independent of our release cycle. The simple form support still works in Kubeapps as advertised for charts that define it (ie. peoples' custom charts and older versions of the chart(s) in the bitnami catalog), but the current apache (and postgres) charts have dropped that data already, so even the existing releases will not be displaying a form, but rather just presenting the values.yaml. I really think that's the correct behaviour (rather than us updating to display a form with all options without any simplicity, custom rendering of widgets or hiding fields etc.)

antgamdia:
Great. As discussed during our standup, yep, we should invest some time in coordinating the actions with the Bitnami Application Catalog people. As I said, perhaps adding an "advanced form" with every param appearing in the (new) values.schema could be worthwhile.

> Update our CI tests to include the tarball of the apache (or otherwise) chart that we test.

I went with this option, this way we no longer depend on what BAC is doing, but rather we manually download our charts manually and just use them in our CI tests.

https://github.com/kubeapps/kubeapps/pull/3541

I'm gonna close this PR, but happy to come back to it once we start looking into the forms in the future.

Issue #3531, opened 2021-09-30
Release 2.4.1


antgamdia:
Creating an issue given it will take some time and there are some blockers we may want to track here.

antgamdia:
Triggering the pre-release flow to check everything is working on the CI side. However, I expect more minor PRs  (like the chart upgrade, etc)  to be merged before the release.

Job here: https://app.circleci.com/pipelines/github/kubeapps/kubeapps/4459/workflows/6ca46184-9087-434e-b17f-3799a77e4cd2

Reminder: pushing to the branch `prerelease` is enough to trigger this job, which is just a check, no write actions are carried out.

PR #3541, opened 2021-10-05
Mutate chart name when pushing to charmusseum

antgamdia:
### Description of the change

This PR is intended to mitigate https://github.com/kubeapps/kubeapps/issues/3339 by replacing the chart `chart` with `prefix-chart`. This way, we will get a valid chart from the bitnami repo (as we were doing), but it cannot be retrieved by looking up the bitnami catalog (since `prefix-chart` does not exist anywhere except for our chartmusseum instance).
Also, it changes the description to ensure we are selecting the mutated chart.

Besides, instead of using the "bitnami" repo, this PR also installs a "bitnami" repo, but points to the local chartmusseum (which holds the mutated charts).

### Benefits

I really hope this PR would fix the CI issues we're having, as well as protect us from further upstream changes...

### Possible drawbacks

Perhaps the e2e tests will have to be modified. Let's see what our beloved CI has to say first. Edit: it passed!

### Applicable issues

-  related #3531

### Additional information

N/A

### DIFF
--- a/integration/use-cases/01-add-multicluster-deployment.js
+++ b/integration/use-cases/01-add-multicluster-deployment.js
@@ -20,8 +20,8 @@ test("Deploys an application with the values by default", async () => {
 
   await expect(page).toClick("a", { text: "Catalog" });
 
-  await expect(page).toMatchElement("a", { text: "Apache HTTP Server", timeout: 60000 });
-  await expect(page).toClick("a", { text: "Apache HTTP Server" });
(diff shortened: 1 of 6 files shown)


antgamdia:
It passed! :sweat_smile: Let me rubber-stamp this PR as it only contains CI changes (and CI is currently failing anyway...)

ccing @absoludity for visibility (there's a lot of ongoing PRs haha)

absoludity (commented):
Excellent!

Issue #3494, opened 2021-09-24
Configure flux-specific RBAC for kubeapps-apis


absoludity:
### Description:

Currently the flux plugin requires either running the kubeapps-apis server with the unsafe, dev-only RBAC for testing, or manually configuring extra RBAC for the service, so that the linked service account has the access control required by the plugin.

The extra RBAC is required because the flux plugin starts its own background processes that watch resources in the cluster asynchronously (outside of the user interactions, which we usually use to limit required RBAC).

This RBAC should be moved to a separate file such as `chart/kubeapps/templates/kubeappsapis/rbac_fluxv2.yaml` that is only installed if `enabledPlugins` contains `fluxv2`, defining a cluster role and role binding for the service account.

It'd be great to minimize the RBAC required too, of course (currently verb: * etc.)

gfichtenholt:
First, I'd like to make sure I understand the problem correctly. I think the gist of what you're saying is we should not need to use "--set kubeappsapis.unsafeUseDemoSA=true" when doing helm install of kubeapps to be able use flux plug-in. Today we have to. If so, sounds reasonable. Hopefully I got that right :-)

Secondly, looks like there is a little more to do than that is described above. Based on looking at current code, it looks like the service account "kubeapps-internal-kubeappsapis" is not even created unless one uses "--set kubeappsapis.unsafeUseDemoSA=true" flag. Is that intentional? If so, this is a little more than adding some RBAC to an existing service account. I think I'd need to:
1) create a service account along with clusterrole and binding. Would that be a flux plug-in specific service account? 
2) if we don't want to use any service account in production for running kubeapps-api deployment (is that true?), then I need to programmatically run the flux plug-in code as (1)

Or maybe the service account "kubeapps-internal-kubeappsapis" was really meant to be created regardless of "--set kubeappsapis.unsafeUseDemoSA=true" flag and kubeappsapis deployment is meant to be using this service account? That would follow the same pattern with apprepository controller and kubeops? 

In short, I am a little confused and need clarification before I proceed. Thanks

antgamdia:
My two cents here:

> I think the gist of what you're saying is we should not need to use "--set kubeappsapis.unsafeUseDemoSA=true" when doing helm install of kubeapps to be able use flux plug-in. 

Yep, I also think so. I'm confident we shouldn't require using any `unsafeXXX` option for deploying Kubeapps whatsoever.

[...]

PR #3551, opened 2021-10-06
 Configure flux-specific RBAC for kubeapps-apis #3494 

gfichtenholt:
In addition to addressing
 Configure flux-specific RBAC for kubeapps-apis #3494 
I have also done some clean up to existing templates, where ClusterRole and ClusterRoleBinding were defined with a namespace, even though they are cluster-wide resources and are not associated with any particular namespace. I mentioned this on slack a few days ago and decided to just clean it up


antgamdia:
CI is failing, very likely due to the latest CI patch we added not being in your branch. Just merging master into your branch should solve the issue.

chart/kubeapps/templates/kubeappsapis/rbac_fluxv2.yaml
```diff
@@ -0,0 +1,47 @@
+{{- if .Values.featureFlags.kubeappsAPIsServer }}
```
absoludity, line 1: I'll do a separate PR to remove that option... it should no longer be an option since the UI is using it now.

chart/kubeapps/templates/kubeappsapis/rbac_fluxv2.yaml
```diff
@@ -0,0 +1,47 @@
...
+    {{- if .Values.commonLabels }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" . ) | nindent 4 }}
+    {{- end }}
+  {{- if .Values.commonAnnotations }}
+  annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+  {{- end }}
+rules:
+  - apiGroups: ["source.toolkit.fluxcd.io", "helm.toolkit.fluxcd.io"]
+    resources: ["helmrepositories", "helmcharts", "helmreleases"]
+    verbs: ["get", "list", "watch", "create", "update", "delete"]
```
absoludity, line 19: Just want to verify that the plugin really does need all 6 verbs? I'm surprised only because I thought the user is creating the HelmRelease CR.

chart/kubeapps/templates/kubeops/rbac.yaml
```diff
@@ -55,8 +55,7 @@ subjects:
 apiVersion: {{ include "common.capabilities.rbac.apiVersion" . }}
 kind: ClusterRole
 metadata:
-  name: "kubeapps:controller:kubeops-ns-discovery-{{ .Release.Namespace }}"
-  namespace: {{ .Release.Namespace | quote }}
+  name: "kubeapps:controller:kubeops-ns-discovery"
```
absoludity, line 58: Again, great to remove the namespace metadata, but the intent of the name is to identify the ClusterRole created by the kubeapps installation in the `.Release.Namespace` namespace.

absoludity (approved):
> I have also done some clean up to existing templates, where ClusterRole and ClusterRoleBinding were defined with a namespace, even though they are cluster-wide resources and are not associated with any particular namespace.

+1 to clean up where those resources specified a `.metadata.namespace` (as that's confusing and would be ignored), but we included the kubeapps installation namespace in the name of those cluster-level resources so that they don't clash when installing multiple copies of Kubeapps in different namespaces (ie. the name of each cluster role is unique per Kubeapps installation)

This used to be possible, but since Helm 3 updated the way CRDs are installed, it may be that its no longer possible (haven't checked) in which case, I don't mind either way, but have a slight preference for not changing the names as part of this PR.

chart/kubeapps/templates/kubeappsapis/rbac.yaml
```diff
@@ -3,8 +3,7 @@
 apiVersion: {{ include "common.capabilities.rbac.apiVersion" . }}
 kind: ClusterRole
 metadata:
-  name: "kubeapps:controller:kubeapps-apis-dev-{{ .Release.Namespace }}"
-  namespace: {{ .Release.Namespace | quote }}
+  name: "kubeapps:controller:kubeapps-apis-dev"
```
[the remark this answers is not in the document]
gfichtenholt: OK, will do that as long as I am changing the file anyway

chart/kubeapps/templates/kubeappsapis/rbac_fluxv2.yaml
gfichtenholt: sounds good

chart/kubeapps/templates/kubeappsapis/rbac_fluxv2.yaml
gfichtenholt: Nope, the plug-in creates HelmRelease CRD when the user calls CreateInstalledPackage(...)

chart/kubeapps/templates/kubeops/rbac.yaml
gfichtenholt: gotcha. will do

Issue #3403, opened 2021-09-13
Design and implement `GetK8sResourcesForInstalledPackage`


ppbaena:
Currently when an app is installed on the cluster, our AppView gets the data about the related k8s resources from the k8s api server.

We want to replace that with an API call which allows getting the (relevant) resources for a particular installed package (only).

This will be a huge step forward in removing the requirement to expose the k8s api server to the dashboard.

antgamdia:
In order to avoid a never-ending issue, let's move some of the remaining open items from https://github.com/kubeapps/kubeapps/issues/3340 to this issue:

- [ ]  https://github.com/kubeapps/kubeapps/issues/3403
- [ ]  Use `GetK8sResources` in the UI (blocked)
    - [ ] Remove old kubeops references for retrieving k8s resources?

_Updated on September 28th_

absoludity:
After successfully testing out some ideas to enable a separate "resources" plugin with a single streaming endpoint for watching multiple/many resources, I've added a [section in the design doc for the GetK8sResources endpoint](https://docs.google.com/document/d/1z03msZRGsRvcRaom-yrvEwIWcEfNy6fSA5Zg28gjYvA/edit#heading=h.2pgw6nn9xing), as well as updating the core packaging API with an extra endpoint for [GetInstalledPackageResourceRefs](https://docs.google.com/document/d/1z03msZRGsRvcRaom-yrvEwIWcEfNy6fSA5Zg28gjYvA/edit#heading=h.5e95xks1mhfz) (which is specific for each packaging backend).

If you can take a look when you have a chance, that'd be great. (cc @gfichtenholt as I don't think you're subscribed to all issue email :) ).

gfichtenholt:
I will take a close look when I get a chance. How soon do you need it? Are you ready to starting adding code for this and and just waiting for me to review? Right now I am have like 3 non-trivial things on my plate that needed to be done yesterday :-) But I will drop everything, as usual, to unblock you as soon as you need it

Just a couple of superficial comments based on cursory examination:
1) you need to update InstalledPackageDetail to fix the TODO
[...]

PR #3581, opened 2021-10-12
Add GetInstalledPackageResourceRefs to packages API and implement in Helm.

absoludity:
### Description of the change

Draft.

Adds the definition for a `GetInstalledPackageResourceRefs` endpoint to the core.packaging API, with an implementation for the helm plugin and the aggregated core plugin. It's the more obvious part for the work in #3403.

See the [relevant section in the design doc](https://docs.google.com/document/d/1z03msZRGsRvcRaom-yrvEwIWcEfNy6fSA5Zg28gjYvA/edit#heading=h.5e95xks1mhfz) which also links to the resources plugin function for getting the resources themselves.

### Benefits

Each packaging plugin is able to return a Kubernetes GroupVersionKind and a name, uniquely identifying all resources created by an installed package.

Enables the further work so that the resources for an installed package can be fetched by clients (without requiring re-implementation by each packaging plugin, since it's not packaging specific).

### Applicable issues

- Ref #3403


gfichtenholt (approved):
I skipped generated files, tests and typescript (front end) code. What's left looks good. +1

Issue #3560, opened 2021-10-08
Flux plugin requires an additional config getter


absoludity:
### Description:

For Helm (and Carvel, fwiw), all requests to the kubeapps-apis service are synchronous requests where we avoid privilege escalation by carrying out the underlying request to the k8s API server with the users credentials.

This is done by providing each plugin (upon initialization) with a kubernetes config getter function which takes the request context and returns a k8s config to use.

This isn't suffecient for the flux plugin which additionally runs out-of-request interactions with the Kubernetes API server. Although we've already ensured that if the flux plugin is selected, that the service account will be granted additional read privileges, we also need to ensure that the plugin can get a config based on the service account rather than the request context.

As [outlined by Greg in a comment](https://github.com/kubeapps/kubeapps/blob/master/cmd/kubeapps-apis/server/plugins.go#L325-L342), the only way to ensure the flux plugin can interact with the required privileges is by patching the code to use the in cluster config (ie. configured service account).

FWIW, I think the solution here is just to update the flux plugin so that for those out-of-band requests to the API server, the plugin simply uses the in cluster config rather than the provided configGetter function. The plugin can already grab this directly with `rest.InClusterConfig()` since it doesn't support clusters other than the one one which Kubeapps (and the apisserver/plugin) is running. We may later want to make this available as a helper to plugins with a cluster-aware function (ie. `getServiceAccountConfig(cluster string)`, but until that is certain, it's simpler for the flux plugin to use `rest.InClusterConfig()`.

cc @gfichtenholt Documenting here as an issue so we can discuss and fix :)

gfichtenholt:
don't forget it also has to work in unit tests with the fake k8s  client

gfichtenholt:
ok, so it looks like the current state after #3559 got merged is that flux plug in is broken and prevents kubeappsapis server from starting (obviously if enabled) due to stuff like this at startup

Error: failed to initialize plugins server: failed to register plugins: plug-in "name:\"fluxv2.packages\" version:\"v1alpha1\"" failed to register due to: customresourcedefinitions.apiextensions.k8s.io "helmrepositories.source.toolkit.fluxcd.io" is forbidden: User "system:anonymous" cannot get resource "customresourcedefinitions" in API group "apiextensions.k8s.io" at the cluster scope

So this needs to be fixed ASAP. Is this PR something you wanted me to implement or were you intending to make these changes yourself?

gfichtenholt:
No problem, I will take of it

[...]

PR #3582, opened 2021-10-12
fix for  Flux plugin requires an additional config getter #3560 

gfichtenholt:
+ a few integration tests for flux v2 plug in were added/refactored


gfichtenholt:
So there is quite a bit of "noise" in this PR, which for the most part feel free to ignore. The main part that's important is in utils.go, called newBackgroundClientGetter(). That's really the only important piece

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
```diff
@@ -71,7 +70,7 @@ func NewServer(configGetter server.KubernetesConfigGetter) (*Server, error) {
 		return nil, err
 	} else {
 		return &Server{
-			clientGetter:       clientGetter,
+			clientGetter:       newClientGetter(configGetter),
```
absoludity, line 73: Great, so the cache is configured with the background client while the server (handling user requests) still always uses the provided config getter (that gets authz from the context).

cmd/kubeapps-apis/server/plugins.go
```diff
@@ -352,11 +334,12 @@ func createConfigGetterWithParams(inClusterConfig *rest.Config, serveOpts ServeO
...
 func extractToken(ctx context.Context) (string, error) {
-	// per https://github.com/kubeapps/kubeapps/pull/3044
-	// extractToken() to return an empty token with a nil error if there is no metadata with the context.
+	// per https://github.com/kubeapps/kubeapps/issues/3560
+	// extractToken() to raise an error if there is no metadata with the context.
+	// note, the caller will wrap this as a codes.Unauthenticated status
 	md, ok := metadata.FromIncomingContext(ctx)
 	if !ok {
-		return "", nil
+		return "", fmt.Errorf("missing authorization metadata")
```
absoludity, line 342: Excellent, thanks.

absoludity (approved):
Excellent, thanks Greg.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
gfichtenholt: correct

PR #3595, opened 2021-10-18
Show http or https when deploying with LoadBalancer

antgamdia:
### Description of the change

I've noticed I had this uncommitted minor change: displaying httpS if the port is 443 in the installation notes. Even if it is a stop-gap solution (not every HTTP over TLS is under 443 port), it will cover a wide range of deployments using the std port.

### Benefits

Navigating to the generated link will no longer result in a 404 page if using a loadbalancer that only exposes the :443 socket (and no redirection has been configured)

### Possible drawbacks

A user defining a plain HTTP protocol listening at :443 will get the error page... but, it's their fault for using an IANA-reserved port :P

### Applicable issues

N/A

### Additional information

N/A


absoludity:
> A user defining a plain HTTP protocol listening at :443 will get the error page... but, it's their fault for using an IANA-reserved port :P

Haha - yes, agree.

Issue #3587, opened 2021-10-13
UI: InternalError too much recursion when the ns exceeds some length


dud225:
Hello

when deploying a chart on a namespace that exceeds some length, e.g. `test1test2test3test4`, Kubeapps' UI fails and the user looses complete control:
![Capture d’écran de 2021-10-13 15-05-02](https://user-images.githubusercontent.com/17550475/137138849-089d26ad-ccf4-4d66-ae8c-eb1c821dca5b.png)
![Capture d’écran de 2021-10-13 16-27-26](https://user-images.githubusercontent.com/17550475/137153450-f9e0bb7e-192e-42b1-abd7-4836801c3e81.png)

There is no error outputted in the dashboard pods.

absoludity:
That depends whether you can reproduce it reliably in some context? Are you able to reproduce it with your chart in the `test1test2test3test4` namespace?

Assuming your chart is private and can't be shared, the next best thing would be to create a public example of a minimal chart that exhibits the issue.

So it could just be that the combination of long namespace name and long chart or release name is causing the issue. Or something your chart in particular is doing with names that it creates for k8s resources? If you can find that, then create and attach a public chart tarball that we can use to trigger the issue, we can then work on improving the handling there.

PR #3602, opened 2021-10-18
Add ReDoS linter

antgamdia:
### Description of the change

After an open issue, we discovered we are adding some ReDoS vulnerable regexes, meaning some malformed (or even legal) inputs could make Kubeapps crash. Since it is mainly client code, the worst-case scenario is a browser tab crash, but, of course, it is sth we should prevent from happening.

This PR is just to add a linter checking the existence of regex that could result in a problem (aka having quadratic or exponential complexity).

### Benefits

We will detect problematic regex at build time.

### Possible drawbacks

Perhaps the lining step will take a bit longer, but I think it worths it.

### Applicable issues

- rel #3587 

### Additional information

This PR won't pass the tests since currently, we have some vulnerable regexes. At some point it will, though (hopefully).

### DIFF
--- a/dashboard/.eslintrc.json
+++ b/dashboard/.eslintrc.json
@@ -30,13 +30,14 @@
     "plugin:react/recommended",
     "plugin:prettier/recommended"
   ],
-  "plugins": ["@typescript-eslint", "import", "jest", "jsx-a11y", "react", "react-hooks"],
+  "plugins": ["@typescript-eslint", "import", "jest", "jsx-a11y", "react", "react-hooks", "redos"],
   "env": {
     "browser": true,
     "es6": true,
     "jest": true
   },
   "rules": {
+    "redos/no-vulnerable": "error", // avoid ReDoS vulnerable regex
     "quotes": [
       "error", // enforce double quotes
       "double",
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -113,6 +113,7 @@
     "eslint-config-prettier": "^8.3.0",
     "eslint-import-resolver-typescript": "^2.5.0",
     "eslint-plugin-prettier": "^4.0.0",
+    "eslint-plugin-redos": "^1.2.0",
     "husky": "^4.3.8",
     "jest-enzyme": "^7.1.2",
     "jest-plugin-context": "^2.9.0",
--- a/dashboard/src/components/AppView/AppNotes/AppNotes.test.tsx
+++ b/dashboard/src/components/AppView/AppNotes/AppNotes.test.tsx
@@ -4,5 +4,6 @@ import AppNotes from "./AppNotes";
 it("renders AppNotes", () => {
   // Basic content testing
   expect(shallow(<AppNotes notes="" />).text()).toBe("");
+  // eslint-disable-next-line redos/no-vulnerable
   expect(shallow(<AppNotes notes="Foo" />).text()).toMatch(/Notes.*Foo/);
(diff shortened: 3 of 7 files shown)


dashboard/src/components/AppView/AccessURLTable/AccessURLItem/AccessURLIngressHelper.tsx
```diff
@@ -48,6 +45,29 @@ export function GetURLItemFromIngress(ingress: IResource) {
...
+export function ShouldGenerateLink(url: string): boolean {
+  // Check if it is a valid URL, delegating the check to the browser API
+  let builtURL;
+  try {
+    // Note this browser API constructor will accept many invalid URLs (not well-encoded, wrong IP octets, etc.)
+    builtURL = new URL(url);
+  } catch (_) {
+    return false;
+  }
+  if (!(builtURL.protocol === "http:" || builtURL.protocol === "https:")) {
```
absoludity, line 62: I assume we don't need to worry about `data:` urls, given they weren't matched by the previous regex.

dashboard/src/components/AppView/AccessURLTable/AccessURLItem/AccessURLIngressHelper.tsx
```diff
@@ -48,6 +45,29 @@ export function GetURLItemFromIngress(ingress: IResource) {
...
+    return false;
+  }
+
+  // Check if any url fragment includes a character likely used in a regex, that is, is not encoded
+  // c.f. https://datatracker.ietf.org/doc/html/rfc3986#section-2.2
+  // even if it won't catch every possible case, we avoid having an O(x^n) regex
+  const regex = /^[^:/?#[\]@!$&'*+,;=]*$/;
+
+  // If it is an URL but it includes some "ilegal" characters, then return false
+  return !builtURL.pathname.split("/")?.some(p => !regex.test(p));
```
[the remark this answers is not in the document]
absoludity: Yep, sounds good.

dashboard/src/shared/Auth.ts
```diff
@@ -42,7 +42,9 @@ export class Auth {
       return [];
     }
     return [
-      "base64url.bearer.authorization.k8s.io." + btoa(token).replace(/=*$/g, ""),
+      // Trimming the b64 padding character ("=") as it is not accepted by k8s
+      // https://github.com/kubernetes/apiserver/blob/master/pkg/authentication/request/websocket/protocol.go#L38
```
absoludity, line 46: Since code changes in main branches, best to link to the code in the latest tag imo. (much less likely to change).

dashboard/src/shared/Auth.ts
```diff
@@ -42,7 +42,9 @@ export class Auth {
...
+      "base64url.bearer.authorization.k8s.io." + btoa(token).replaceAll("=", ""),
```
absoludity, line 47: I was wondering why you weren't just removing them from the end, but given that `=` isn't part of radix-64, they can only be present as padding.

antgamdia: Yep, it won't appear if the string is a multiple of a 3 character number since b64 encodes each 3 bytes. There is a good answer here: https://stackoverflow.com/questions/6916805/why-does-a-base64-encoded-string-have-an-sign-at-the-end
As you said, "=" can only be padding, so it is safe to be wiped out.

Issue #3496, opened 2021-09-24
Investigate and propose package repositories API with similar core interface to packages API.


absoludity:
As we close in on the basic functionality for the packages API, we'll need to begin investigating and proposing a similar core API for package repositories, so that as a user I can use the Kubeapps UI to interact with `AppRepositories` (helm and oci) as well as flux sources.

We've recently had contributions for remote components that are conditionally used in different views, which will be something we'll want to begin using (though not remote, restricted to our internal services) both in the packages API as well as the package repositories API, where I assume the UX will be quite different for the various repository types (more so than packages).

absoludity:
I'm not sure why this is committed: I'd expect it to be prioritised and scheduled to come after #3403 which I'm starting now?

gfichtenholt:
I am starting work on this now. My plan of action is this: for each operation (CRUD) 
1) update design doc https://docs.google.com/document/d/1z03msZRGsRvcRaom-yrvEwIWcEfNy6fSA5Zg28gjYvA/edit?usp=sharing with corresponding messages and rpc func
2) once (1) is reviewed, add corresponding stubs in core packages.proto and each plug-in .proto
3) add implementation for flux plugin and unit tests

Let me know it sounds good or if I missed anything. Thanks

absoludity:
Great, thanks Greg.

> I am starting work on this now. My plan of action is this: for each operation (CRUD)
> 
>     1. update design doc https://docs.google.com/document/d/1z03msZRGsRvcRaom-yrvEwIWcEfNy6fSA5Zg28gjYvA/edit?usp=sharing with corresponding messages and rpc func
> 
>     2. once (1) is reviewed, add corresponding stubs in core packages.proto and each plug-in .proto

Sounds good. I wasn't sure whether we'd want this is the packages API or as a separate `repositories` API which can also be implemented by plugins - that was part of what needs to be thought through. Perhaps it is simpler to just do it in the packages API as you've thought, if each packaging system has its own take on an app repository anyway.

> 
>     3. add implementation for flux plugin and unit tests

I'm not sure, but thought that implementing the helm plugin may be higher priority since it allows us to switch over the existing frontend code, but no worries if you'd rather leave that for someone else.

absoludity:
Yep, fine either way, I more thought that it's important to ensure that the new API satisfies the requirements of the existing helm functionality. It doesn't bother me who does it as long as we work with this in mind.

gfichtenholt:
more work to do

[...]

PR #3609, opened 2021-10-19
Enable UX to request fluxv2 available packages

absoludity:
### Description of the change

This will be the first in a list of PRs required to use the flux plugin from the UI (I'll add and link an issue to track them).

Running the Kubeapps UX with the fluxv2 plugin enabled results in:

kubeapps-with-fluxv2-unimplemented

This is because currently the plugin handles the case where the `request.Context.Cluster == ""` (ie. cluster isn't specified) only, in which case it correctly defaults to the cluster on which Kubeapps is installed. Since the UI will not know which requests plugins can handle (ie. which plugins support requests for different clusters etc.), it will always set the full context and rely on the plugin to respond with unimplemented or a valid response accordingly.

With this change, the response correctly comes back as empty rather than unimplemented (before I've added any flux `HelmRepositories`):

kubeapps-with-fluxv2-empty

### Benefits

Don't see an error when browsing for available apps using the flux plugin.

### Possible drawbacks

Can't think of any.

### Applicable issues

Ref: #3610

### DIFF
--- a/cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/main.go
+++ b/cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/main.go
@@ -40,7 +40,7 @@ func init() {
 // returning the server implementation.
 func RegisterWithGRPCServer(s grpc.ServiceRegistrar, configGetter server.KubernetesConfigGetter, clustersConfig kube.ClustersConfig) (interface{}, error) {
 	log.Infof("+fluxv2 RegisterWithGRPCServer")
-	svr, err := NewServer(configGetter)
+	svr, err := NewServer(configGetter, clustersConfig.KubeappsClusterName)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


gfichtenholt (commented):
What about `newHelmActionConfigGetter()` in utils.go? Should that be also modified to be consistent?

absoludity:
> What about `newHelmActionConfigGetter()` in utils.go? Should that be also modified to be consistent?

Yes, and all the other operations will need to be updated. I've just written the tests and updated the one operation here because it enables me to continue focusing on the UX (assuming that's what you'd prefer), and have added an item to the related issue #3610  for doing the other operations. Let me know if you're able to do that item (no UX, plugin only) and we'll be able to get more UX work done.

gfichtenholt:
> > What about `newHelmActionConfigGetter()` in utils.go? Should that be also modified to be consistent?
> 
> Yes, and all the other operations will need to be updated. I've just written the tests and updated the one operation here because it enables me to continue focusing on the UX (assuming that's what you'd prefer), and have added an item to the related issue #3610 for doing the other operations. Let me know if you're able to do that item (no UX, plugin only) and we'll be able to get more UX work done.

ok, fine with me. Regarding other work. Generally, I am currently knee deep in #3496 (it is very hard and going slow), but if there a simple change I can make to unblock someone else, I am down

gfichtenholt (approved):
+1

Issue #3610, opened 2021-10-19
Enable flux plugin to be used from Kubeapps UX


absoludity:
### Description:

While Greg's done a great job on the fluxv2 plugin itself, we have not yet used it from the Kubeapps UI. This issue is to identify and fix those issues so that others can use Kubeapps with flux.

I expect this list to grow a little as we work through the functionality, but feel free to put your initials next to an item if you want to grab it. To run Kubeapps with the flux plugin only, I'm using the `deploy-dev-kubeapps` make target with the following local diff:

```diff
--- a/docs/user/manifests/kubeapps-local-dev-values.yaml
+++ b/docs/user/manifests/kubeapps-local-dev-values.yaml
@@ -15,6 +15,8 @@ postgresql:
   existingSecret: postgresql-db
 kubeappsapis:
   replicaCount: 1
+  enabledPlugins:
+    - fluxv2
 ingress:
   enabled: true
   hostname: localhost
@@ -27,3 +29,7 @@ ingress:
     nginx.ingress.kubernetes.io/proxy-read-timeout: "600"
     # Required for ingress-nginx 1.0.0 for a valid ingress.
     kubernetes.io/ingress.class: nginx
+redis:
+  enabled: true
+  auth:
+    password: "testing"
```

## Issues

- [x] Plugin's GetAvailablePackageSummaries needs to handle UX request with cluster (MN: #3609 )
- [x] Available apps don't display due to non-empty repo filter requested by UX (MN: https://github.com/kubeapps/kubeapps/pull/3619)
- [x] Getting available package detail: `Unable to get the available package detail for the package "bitnami/apache" using the plugin "fluxv2.packages": rpc error: code = NotFound desc = no fully indexed repository [default/bitnami] has been found`. Issue is that currently dashboard assumes repo is either in the current namespace, or the global namespace. This is not the case with flux (or generally). Need to update URLs to always encode the namespace? (MN: #3628)
- [x] All other fluxv2 plugin operations must also support UX request with cluster (similar to previous, just tests and one-liner) GF: #3620 
- [x] Unable to deploy package via Helm due to incorrect namespace used for available package (MN: #3640)
- [x] `HelmRelease` resource created when deploying via UI is in the `kubeapps` namespace, though the helm release and resources are correctly in the target namespace (see additional info on #3640) 
- [x] UX tries to grab a helm release via kubeops directly after deploying (which it can't and shouldn't do either way, see additional info on #3640) (MN: #3686)

PR #3619, opened 2021-10-20
Don't send repository filter [""] by default.

absoludity:
- REFACTOR: Update tests to check actual client call.
- RED: Don't add filter options when there aren't any.
- GREEN: Don't add filter options when there aren't any.

### Description of the change

I added a fluxv2 HelmRepository and watched (in the logs) as it populated the cache, but the dashboard was still getting zero available packages.

After some investigation, it turned out the reason was that the dashboard was requesting the available packages with `filter.repositories` set to a non-empty list `[""]`.

I found I couldn't write a test for this since the existing test wasn't checking the args sent to the actual grpc client (our boundary here), so I first refactored the test to do that, then updated the tests so that filters are only expected if they are requested then the code.

We could instead just always send the empty defaults (in retrospect, I see that the issue was just that `repos.split(",")` doesn't return an empty list which was probably expected, but I prefer to only set them if they are set by the call-site.

### Benefits

Available apps are now displayed when using the flux plugin.

### Possible drawbacks

None.

### Applicable issues

- Ref #3610

### Additional information

If you now click on an available app, it does not yet display but renders the error: "Unable to fetch package: Unable to get the available package versions for the package "bitnami/apache" using the plugin "fluxv2.packages": rpc error: code = Internal desc = unable to retrieve versions for chart: [bitnami/apache]", but this may be fixed by ensuring all flux rpc's handle the cluster being set.

The request for the package detail also fails, UX seems to be incorrectly looking for the repo data in the current namespace (default) rather than where the repo is (flux-system), but not sure yet, needs investigation:
flux-get-available-package-detail-error

cc @gfichtenholt , just FYI .

### DIFF
--- a/dashboard/src/shared/PackagesService.test.ts
+++ b/dashboard/src/shared/PackagesService.test.ts
@@ -116,7 +116,7 @@ describe("App", () => {
           } as GetAvailablePackageSummariesResponse),
         );
         // Create a real client, but we'll stub out the function we're interested in.
-        let mockClient = new KubeappsGrpcClient().getPackagesServiceClientImpl();
+        const mockClient = new KubeappsGrpcClient().getPackagesServiceClientImpl();
         jest
           .spyOn(mockClient, "GetAvailablePackageSummaries")
           .mockImplementation(mockClientGetAvailablePackageSummaries);


absoludity:
> Just FWIW I checked my code: in order for the error message to say what it says, it MUST be the case that `request.AvailablePackageRef.Context.Namespace == "default"`

Yep, that's what I suspected (assuming you mean the next issue that I've highlighted in the additional info). Thanks for checking Greg.

dashboard/src/shared/PackagesService.ts
```diff
@@ -17,14 +17,35 @@ export default class PackagesService {
...
       context: { cluster: cluster, namespace: namespace },
-      filterOptions: {
-        query: query,
-        repositories: repos.split(","),
-      },
       paginationOptions: { pageSize: size, pageToken: page.toString() },
-    });
+    };
+
+    // Add filter options only as needed.
```
antgamdia, line 25: Great, thanks for this change. I assumed it was ok with passing `""` and `[""]` to the plugin, so I didn't care about it so much.

dashboard/src/shared/PackagesService.test.ts
```diff
@@ -32,59 +33,88 @@ describe("App", () => {
...
           } as GetAvailablePackageSummariesResponse),
         );
+        // Create a real client, but we'll stub out the function we're interested in.
+        const mockClient = new KubeappsGrpcClient().getPackagesServiceClientImpl();
         jest
-          .spyOn(PackagesService, "getAvailablePackageSummaries")
-          .mockImplementation(mockGetAvailablePackageSummaries);
+          .spyOn(mockClient, "GetAvailablePackageSummaries")
+          .mockImplementation(mockClientGetAvailablePackageSummaries);
+        jest.spyOn(PackagesService, "client").mockImplementation(() => mockClient);
```
antgamdia, line 117: Nice. This way we can really inspect the options being used!

Issue #3032, opened 2021-06-23
Investigate caching layer for kubeapps plug-ins


gfichtenholt:
Initially caching layer initially specifically to the flux plugin, but the end goal is to generalise it to provide caching to all plugins as needed

Here is the plan 
1) install redis as part of kubeapps helm chart
2) write code for new deployment/operator that watches for changes to certain flux HelmRepository CRD and updates the contents of the redis cache, e.g. loads the index.yaml
3) modify flux plug-in code to use said cache as needed

absoludity:
> Initially caching layer initially specifically to the flux plugin, but the end goal is to generalise it to provide caching to all plugins as needed
> 
> Here is the plan
> 
>     1. install redis as part of kubeapps helm chart
> 
>     2. write code for new deployment/operator that watches for changes to certain flux HelmRepository CRD and updates the contents of the redis cache, e.g. loads the index.yaml

Just a thought: I don't know that we need to go that far, and even if we did, it doesn't solve the fact that such a CRD won't necessarily change when the corresponding index.yaml is updated. With flux it will, aiui, flux itself will handle keeping the cached index.yaml up to date with some set frequency, and I guess you're planning to watch that resource for changes and update the extracted content of the redis cache when that happens.

But I wonder if a more general approach, which will work for any requirements, would be for a plugin to be able to provide a function that gets called on some condition, whether that's a time or a watch.

Note also that I don't think a separate controller/deployment is required... any goroutine could run on startup as part of the existing kubeappsaips service to watch resources and/or do things at certain times... might be worth starting as a goroutine there, but see what you think.

I guess one other thing that may become essential here (if it's a general service provided to all plugins) is that the process that watches is going to need explicit RBAC to whatever it's watching (unlike our API endpoints which can run synchronously as the user). If it's a goroutine running in the kubeappsapis service, then I'm OK with the burden of responsibility being on the person configuring (or the chart) to bind the read-only RBAC for the configured plugins to the kubeappsapis service account.

Looks like an interesting problem!

> 
>     3. modify flux plug-in code to use said cache as needed

gfichtenholt:
the first step was pushed. My next item is to improve the efficiency and robustness of cache based on watches. 
at the moment, the code completely relies on k8s events firing and notifying of any changes to make sure cache is up to date. Watches have known limitations (e.g. limited history)  and in my experience also are sometimes flaky, meaning dying and/or missing events. I plan to follow
[...]

PR #3620, opened 2021-10-20
Fix for #3610 All other fluxv2 plugin operations must also support UX request with cluster 

gfichtenholt:
Plus previously discussed performance optimization for data flux plug stores in the cache
https://github.com/kubeapps/kubeapps/issues/3032#issuecomment-942924026


cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
```diff
@@ -191,10 +193,14 @@ func (s *Server) GetAvailablePackageSummaries(ctx context.Context, request *core
...
 		nextPageToken = fmt.Sprintf("%d", pageOffset+1)
 	}
+
 	return &corev1.GetAvailablePackageSummariesResponse{
 		AvailablePackageSummaries: packageSummaries,
 		NextPageToken:             nextPageToken,
 		// TODO (gfichtenholt) Categories?
+		// Just happened to notice that helm plug-in returning this.
+		// Never discussed this and the design doc appears to have a lot of back-and-forth comments
+		// about this, semantics aren't very clear
```
absoludity, line 203: Yes, we decided to return the categories so that we could finish the helm integration in the UX. It's not the best solution long-term but I think is OK right now. It's not so important, given that only the Bitnami repository defines categories like this, though if the bitnami repo becomes a main use-case for the flux plugin, we may want to include it.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/release_integration_test.go
```diff
@@ -1056,4 +1078,13 @@ var (
 			"kubectl -n @TARGET_NS@ port-forward deploy/@TARGET_NS@-my-podinfo-13 8080:9898\n",
 		AvailablePackageRef: availableRef("podinfo-13/podinfo", "default"),
 	}
+
+	create_request_wrong_cluster = &corev1.CreateInstalledPackageRequest{
+		AvailablePackageRef: availableRef("podinfo-14/podinfo", "default"),
+		Name:                "my-podinfo",
+		TargetContext: &corev1.Context{
+			Namespace: "test-14",
+			Cluster:   "this is not the cluster you're looking for",
```
absoludity, line 1087: Haha - nice.

absoludity (approved):
Thanks Greg.

Issue #3622, opened 2021-10-20
UI: "new package available" link should pre-fill the latest version


dud225:
Hello

When Kubeapps reports that a new release is available it displays a link to the upgrade page:
![Capture d’écran de 2021-10-20 15-34-29](https://user-images.githubusercontent.com/17550475/138110486-8292f55b-d1bc-4ceb-90db-afa063d132b3.png)

From there the current release is selected, but it would be nice to have the latest version automatically selected instead:
![Capture d’écran de 2021-10-20 15-34-44](https://user-images.githubusercontent.com/17550475/138110582-4105e07d-2372-4c46-8741-e458adb19898.png)

If one doesn't pay attention to this, it can easily lead to some confusion.

Issue #3563, opened 2021-10-08
Review cluster/namespace usage in the UI


antgamdia:
After the UI migration we have been accomplishing, there are some missing aspects we should verify more carefully.

Some UI components are views that take directly the URL params to fetch some data, however, other components rely on some state properties to be populated.

Now, our data model does have enough information about the context or each object (aka pagacke), so we no longer need to pass the namespace and cluster again and again.

So currently ve have a combination of "repoNamespace", "namespace", "availablePkgRef.context.namespace", instaledPkgRef.context.namespace", etc, which is far from being an ideal situation.

I guess only those components that are directly mapped to an URL should be responsible for handling URL params and populate the state. Next, other components should only rely on this state.

This issue is to perform a review of the usage of "namespace" and "cluster" across the different places of our UI codebase and take the corresponding actions.

PR #3623, opened 2021-10-20
Check cluster/ns in views + refactor + ui improvements

antgamdia:
### Description of the change

This PR adds a bunch of improvements in the UI side, namely: 

Bugs:
- Avoid duplicated key in AppList view
  - No user-faced mgs, but errors in the console if two apps in different ns have the same name.
- Clear the state when fetching apps
  - When navigating from the catalog, attempting to deploy sth and then, going back to the installed apps and trying to upgrade something, the values were the ones from the first visited package.
- Prevent async errors when unmounting some components
  - No user-faced msg, but when navigating, some errors appeared in the console because of a wrong unmount of the component.

Features:
- Generate and use upgrade links with a fixed version (cf https://github.com/kubeapps/kubeapps/issues/3622)

Refactor:
- Use the package id instead of splitting the name by (`/`) to get the repo name.
- Use the availablePkgSummary in the catalog item instead of using a custom data model.
- Refactor PackageView and DeploymentForm to ensure proper usage of cluster/ns
- Delete an unused route 

### Benefits

Avoid aforementioned bugs, new feature and cleaning-up of our code.

### Applicable issues

- fixes https://github.com/kubeapps/kubeapps/issues/3563
- fixes https://github.com/kubeapps/kubeapps/issues/3622

### Additional information

Sending the PR as a draft for triggering the CI. Some tests are still pending (like the one for the named versioned upgrade link).
Edit: CI is failing? I still have to check why.

### DIFF
--- a/dashboard/src/components/UpgradeForm/UpgradeForm.test.tsx
+++ b/dashboard/src/components/UpgradeForm/UpgradeForm.test.tsx
@@ -305,6 +305,11 @@ it("defaults the upgrade version to the current version", () => {
 });
 
 it("uses the selected version passed in the component's props", () => {
+  const mockDispatch = jest.fn().mockReturnValue(true);
+  jest.spyOn(ReactRedux, "useDispatch").mockReturnValue(mockDispatch);
+  const fetchAndSelectAvailablePackageDetail = jest.fn();
+  actions.packages.fetchAndSelectAvailablePackageDetail = fetchAndSelectAvailablePackageDetail;
+
   const state = {
     ...defaultStore,
     apps: {
@@ -313,19 +318,33 @@ it("uses the selected version passed in the component's props", () => {
       isFetching: false,
     } as IAppState,
     packages: {
-      selected: selectedPkg,
+      selected: {
+        ...selectedPkg,
+        versions: [] as PackageAppVersion[],
+      },
     } as IPackageState,
   };
 
-  const wrapper = mountWrapper(
+  mountWrapper(
     getStore({ ...state }),
-    <MemoryRouter initialEntries={[routePathParam]}>
-      <Route path={routePath}>
-        <UpgradeForm version={"0.0.1"} />,
+    <MemoryRouter initialEntries={[routePathParam + "/1.5.0"]}>
+      <Route path={routePath + "/:version"}>
+        <UpgradeForm version={"1.5.0"} />,
       </Route>
     </MemoryRouter>,
   );
-  expect(wrapper.find(DeploymentFormBody).prop("packageVersion")).toBe("0.0.1");
+
+  expect(fetchAndSelectAvailablePackageDetail).toHaveBeenCalledWith(
+    {
+      context: {
+        cluster: defaultProps.cluster,
+        namespace: defaultProps.repoNamespace,
+      },
+      identifier: defaultProps.packageId,
+      plugin: defaultProps.plugin,
+    },
+    "1.5.0",
+  );
 });
 
 it("forwards the appValues when modified", () => {


dashboard/src/components/Catalog/CatalogItems.tsx
```diff
@@ -28,25 +30,16 @@ export default function CatalogItems({
...
             name: c.displayName,
-            icon: c.iconUrl ?? placeholder,
-            version: c.latestVersion?.pkgVersion ?? "",
-            description: c.shortDescription,
-            // TODO(agamez): get the repo name once available
-            // https://github.com/kubeapps/kubeapps/issues/3165#issuecomment-884574732
-            repo: {
-              name: c.availablePackageRef?.identifier.split("/")[0],
-              namespace: c.availablePackageRef?.context?.namespace,
-            } as IRepo,
```
absoludity, line 45: It's not clear to me from context how all this data can be removed (or replaced with availablePackageSummary) - unless you've updated elsewhere to populate that.

dashboard/src/components/PackageHeader/PackageHeader.test.tsx
```diff
@@ -42,7 +42,7 @@ const testProps: IPackageHeaderProps = {
 
 it("renders a header for the package", () => {
   const wrapper = mount(<PackageHeader {...testProps} />);
-  expect(wrapper.text()).toContain("testrepo/test");
+  expect(wrapper.text()).toContain("testrepo/foo");
```
absoludity, line 45: What's triggering an expectation change here?

dashboard/src/components/UpgradeForm/UpgradeForm.tsx
```diff
@@ -7,28 +7,20 @@ import PackageHeader from "components/PackageHeader/PackageHeader";
...
 import DeploymentFormBody from "../DeploymentFormBody/DeploymentFormBody";
 import LoadingWrapper from "../LoadingWrapper/LoadingWrapper";
 import "./UpgradeForm.css";
 
 export interface IUpgradeFormProps {
-  installedAppAvailablePackageDetail: AvailablePackageDetail;
-  installedAppInstalledPackageDetail: InstalledPackageDetail;
-  selectedPackage: IPackageState["selected"];
-  chartsIsFetching: boolean;
-  error?: Error;
```
absoludity, line 31: So these props stopped being used or moved into state in a previous commit?

absoludity (approved):
+1 when ci passes.

dashboard/src/components/Catalog/CatalogItems.tsx
antgamdia: I didn't have the chance to write some comments to ease the review, but, yep, of course, I had to update it elsewhere. Particulary, this info is being used in the `<InfoCard>` that is rendered at `PackageCatalogItem.tsx`.

Ideally, these components would get simplified to just an `AvailablePackageSummaryItem`, but currently we sill to differentiate between a package and an operator.
Wherefore I just simply extracted the operator things (namespace, description, etc) back to the `OperatorCatalogItem` and simplify as much as I could the package item. However, I had to leave some fields (like the `name: c.displayName` as they were being used in the `CatalogItem`.

But, at least, we are converging, step by step, in simplifying the UI logic in favor of the new package data model.

dashboard/src/components/Catalog/PackageCatalogItem.tsx
```diff
@@ -1,38 +1,53 @@
...
+      title={decodeURIComponent(availablePackageSummary.displayName)}
+      link={packageViewLink}
+      info={availablePackageSummary?.latestVersion?.pkgVersion || ""}
+      icon={availablePackageSummary.iconUrl || placeholder}
+      description={trimDescription(availablePackageSummary.shortDescription)}
+      tag1Content={<span>{repositoryName}</span>}
+      bgIcon={getPluginIcon(availablePackageSummary.availablePackageRef?.plugin)}
     />
   );
 }
```
antgamdia, line 53: > It's not clear to me from context how all this data can be removed (or replaced with availablePackageSummary) - unless you've updated elsewhere to populate that.

Here are the changes indeed. I've used this way to extract the repo name; not sure if we would want to include it as part of the API, though.

dashboard/src/components/PackageHeader/PackageHeader.test.tsx
antgamdia: It is a problem with the availablePackageDetail being passed. In fact, I'm fixing it instead of changing the assertion.

Let  `availablePackageDetail.name` be `test` and `availablePackageDetail.availablePackageRef.identifie` be `testrepo/foo`,
before, we were concatening repo+name, that is `testrepo/test` and, now, we're just using the id, that is `testrepo/foo`

We just should change the availablePackageDetail.name from "test" to "foo".

dashboard/src/components/UpgradeForm/UpgradeForm.tsx
antgamdia: Yeah, I forgot to remove them when I moved them to the state. Now I'm just passing the version for the fixed upgrade.

dashboard/src/components/Catalog/CatalogItems.tsx
absoludity: Excellent. Thanks for the explanation :)

PR #3628, opened 2021-10-21
Enable viewing AvailablePackageDetail with flux.

absoludity:
### Description of the change

The current Kubeapps dashboard UX, in particular the urls, support only two options for where to find an available package: the current namespace or the global one.

This doesn't work for flux, where (currently) a repo can be in any namespace and is effectively public.

This PR updates the URLs for displaying an available package so that the package namespace (and cluster) is always included in the URL (ie. we're representing the full available package reference in the URL now) and enables viewing available packages for a target namespace (where those packages are from a repo in a different namespace).

CI will tell if this breaks current helm support :)

### Benefits

I can view the available package detail when using the flux plugin.

### Applicable issues

Ref: #3610 

### Additional information

Just noted that fetching the available package detail appears to be taking around 2.5s (see below). Not a big deal, but just checking in case @gfichtenholt can make it quicker (he's good at that :P )

flux-apache

Note also: the screenshot shows that the chart title is being rendered without being decoded, but I'm guessing that's from your PR @antgamdia ? Let me know if not, but don't think I've changed the chart id here.

### DIFF
--- a/dashboard/src/components/PackageHeader/PackageView.tsx
+++ b/dashboard/src/components/PackageHeader/PackageView.tsx
@@ -52,7 +52,7 @@ export default function PackageView() {
   const [pluginObj] = useState({ name: pluginName, version: pluginVersion } as Plugin);
 
   const isGlobal =
-    packageCluster == config.kubeappsCluster && packageNamespace == config.kubeappsNamespace;
+    packageCluster === config.kubeappsCluster && packageNamespace === config.kubeappsNamespace;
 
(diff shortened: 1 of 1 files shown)


gfichtenholt:
> 
> Just noted that fetching the available package detail appears to be taking around 2.5s (see below). Not a big deal, but just checking in case @gfichtenholt can make it quicker (he's good at that :P )

do you happen to know if is it taking this long on the first call (for a given package) or subsequent calls too? My magic does have limits :-)

Issue #2962, opened 2021-06-09
Need graceful error handling when adding an OCI repository that has no charts


gfichtenholt:
While experimenting with kubeapps OCI support I went to demo.goharbor.io, created a project 'greg' and pushed an image 'hello-world'. I then tried to create a OCI repository in kubeapps, not realizing that the code expected to see helm charts, not images. There was no error in the UX, but it turns out on the back-end the sync job fails with 
$ kubectl logs pod/apprepo-default-sync-demo.goharbor.io-1623213600-m8ncv -n kubeapps
time="2021-06-09T04:43:15Z" level=info msg="Last checksum: "
panic: runtime error: index out of range [0] with length 0

goroutine 75 [running]:
main.(*OCIRegistry).Charts.func3(0xc000110000, 0x1, 0xc0000892c0)
	/go/src/github.com/kubeapps/kubeapps/cmd/asset-syncer/utils.go:682 +0x235
created by main.(*OCIRegistry).Charts
	/go/src/github.com/kubeapps/kubeapps/cmd/asset-syncer/utils.go:679 +0x2a7

@absoludity suggested I file an issue. "Yeah, it'd certainly be worth seeing something like "the specified repository is not a helm chart" or similar.

PR #3656, opened 2021-10-25
Add tests for successful OCI repo validation.

absoludity:

### Description of the change

Just a follow-up of #3605 to add some unit-tests for the OCI validation as well as the handler itself.

It changes the code in a few places as a result of those tests (see my inline comments), but the main change is adding a small test double for an upstream OCI registry that handles the `tag/list` and `manifest` endpoints so we can test the code against it. Not perfect, but better than not testing.

### Benefits

Found some small errors in the responses that we could fix.
No longer send 1MB of data back for a successful validation request.

### Applicable issues

- Ref #2962 .

### Additional information

I want to follow-up with a PR to use a HEAD request for validating non-OCI repos where possible (falling back for a method-not-allowed) as this will speed up the validation of bitnami hugely.

### DIFF
--- a/pkg/kube/kube_handler.go
+++ b/pkg/kube/kube_handler.go
@@ -49,6 +49,8 @@ import (
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 )
 
+const OCIImageManifestMediaType = "application/vnd.oci.image.manifest.v1+json"
+
 // ClusterConfig contains required info to talk to additional clusters.
 type ClusterConfig struct {
 	Name                     string `json:"name"`
@@ -751,7 +753,7 @@ func getOCIAppRepositoryTag(cli httpclient.Client, repoURL string, repoName stri
 	}
 
 	//This header is required for a successful request
-	req.Header.Set("Accept", "application/vnd.oci.image.manifest.v1+json")
+	req.Header.Set("Accept", OCIImageManifestMediaType)
 
 	resp, err := cli.Do(req)
 	if err != nil {
@@ -807,7 +809,7 @@ func getOCIAppRepositoryMediaType(cli httpclient.Client, repoURL string, repoNam
 	}
 
 	//This header is required for a successful request
-	req.Header.Set("Accept", "application/vnd.oci.image.manifest.v1+json")
+	req.Header.Set("Accept", OCIImageManifestMediaType)
 
 	resp, err := cli.Do(req)
 


pkg/kube/kube_handler.go
```diff
@@ -421,20 +421,6 @@ type appRepositoryRequestDetails struct {
...
-}
-
-type HelmNonOCIValidator struct {
-	req *http.Request
-}
-
-type HelmOCIValidator struct {
-	appRepo *v1alpha1.AppRepository
-}
-
```
absoludity, line 437: I've just moved the interface and related structs down to where the implementation is.

pkg/kube/kube_handler.go
```diff
@@ -733,11 +719,13 @@ type repoTagsList struct {
...
-	Config struct {
-		MediaType string `json:"mediaType"`
-	} `json:"config"`
+type repoConfig struct {
+	MediaType string `json:"mediaType"`
+}
+
+// repoManifest stores the mediatype for an OCI repository.
+type repoManifest struct {
+	Config repoConfig `json:"config"`
```
absoludity, line 728: Needed to split out the inner struct so that I can create that data in tests.

pkg/kube/kube_handler.go
```diff
@@ -775,6 +759,10 @@ func getOCIAppRepositoryTag(cli httpclient.Client, repoURL string, repoName stri
 	}
 	defer resp.Body.Close()
 
+	if resp.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("Unexpected status code when querying %q: %d", repoName, resp.StatusCode)
+	}
+
```
absoludity, line 765: This was one change which the new tests spotted as a requirement: if a non-OK status code is returned, we should not continue on and in particular, should not assume the body is parsable json.

pkg/kube/kube_handler.go
```diff
@@ -881,56 +861,57 @@ func ValidateOCIAppRepository(appRepo *v1alpha1.AppRepository, cli httpclient.Cl
...
 	if err != nil {
 		// If the request fail, it's not an internal error
 		return &ValidationResponse{Code: 400, Message: err.Error()}, nil
 	}
-	body, err := ioutil.ReadAll(res.Body)
-	if err != nil {
-		return nil, fmt.Errorf("Unable to parse validation response. Got: %w", err)
-	}
-	response = &ValidationResponse{Code: res.StatusCode, Message: string(body)}
+	response = &ValidationResponse{Code: res.StatusCode, Message: "OK"}
```
absoludity, line 884: This was another change that came out of the tests (checking the response). Turns out we're sending over a MB of data back with the validation request of a large repo (like bitnami) since we were putting the body in the message, the body being the index.yaml in this case. So updated to only read the body for non-200 responses from the upstream repo.

pkg/kube/kube_handler.go
```diff
@@ -881,56 +861,57 @@ func ValidateOCIAppRepository(appRepo *v1alpha1.AppRepository, cli httpclient.Cl
...
+	response = &ValidationResponse{Code: 200, Message: "OK"}
 
-	isValidRepo, err := ValidateOCIAppRepository(r.appRepo, cli)
-	if !isValidRepo {
+	// If there was an error validating the OCI repository, it's not an internal error.
+	isValidRepo, err := ValidateOCIAppRepository(r.AppRepo, cli)
+	if err != nil || !isValidRepo {
 		response = &ValidationResponse{Code: 400, Message: err.Error()}
 	}
-	return response, err
```
absoludity, line 921: Generally in Go, if you have an error, you return `nil, err`, if you don't have an error, you return `response, nil`, keeping them mutually exclusive. Again, the test brought this out so that we now return a validation response here. We could potentially remove the error from the signature, if it was only the OCI validator, but the non-OCI validator still has cases where it returns an error and I didn't want to change that here.

pkg/kube/kube_handler.go
```diff
@@ -963,13 +944,16 @@ func (a *userHandler) ValidateAppRepository(appRepoBody io.ReadCloser, requestNa
 	// Split body parsing to a different function for ease testing
 	appRepo, cli, err := a.getValidationCli(appRepoBody, requestNamespace, a.kubeappsNamespace)
 	if err != nil {
-		return nil, err
+		return &ValidationResponse{
+			Code:    400,
+			Message: err.Error(),
+		}, nil
```
absoludity, line 950: This is the last change that the tests brought up: if the request is for a global app repo but has a docker secret, an error is returned that is most definitely a validation error (ie. should be returned as the validation response). As it was the only error returned by `getValidationCli` I've just updated to return the validation response here on error.

pkg/kube/kube_handler.go
```diff
@@ -881,56 +861,57 @@ func ValidateOCIAppRepository(appRepo *v1alpha1.AppRepository, cli httpclient.Cl
...
-	return r.req
+type HelmOCIValidator struct {
+	AppRepo *v1alpha1.AppRepository
 }
 
-func (r HelmNonOCIValidator) getAppRepo() *v1alpha1.AppRepository {
-	return nil
-}
-func (r HelmOCIValidator) IsValid(cli httpclient.Client) (*ValidationResponse, error) {
+func (r HelmOCIValidator) Validate(cli httpclient.Client) (*ValidationResponse, error) {
```
absoludity, line 900: Since the interface method has a signature that isn't returning a bool, but rather returning a validation response, I've renamed the method (as `IsValid` implies a single bool return).

pkg/kube/kube_handler_test.go
```diff
@@ -58,11 +61,13 @@ type secretStub struct {
...
 
 type fakeHTTPCli struct {
+	request  *http.Request
 	response *http.Response
 	err      error
 }
 
-func (f *fakeHTTPCli) Do(*http.Request) (*http.Response, error) {
+func (f *fakeHTTPCli) Do(r *http.Request) (*http.Response, error) {
+	f.request = r
```
absoludity, line 70: Just updated to record the request, helpful for the initial tests.

pkg/kube/kube_handler_test.go
```diff
@@ -1088,137 +1093,523 @@ func setClientsetData(cs fakeCombinedClientset, namespaceNames []existingNs, err
...
-		requestNamespace           string
-		expectedURLs               []string
-		expectedHeaders            http.Header
-		expectedError              error
-		expectedReqResolutionError error
+	testCases := []struct {
+		name              string
+		appRepo           *v1alpha1.AppRepository
+		expectedValidator HttpValidator
+		expectedError     error
```
absoludity, line 1102: Updated this test case to focus on unit-testing `getValidator` only, just input vs output and error. Note that we can use the interface as the field type so the test cases can expect both the OCI and non-OCI result.

pkg/kube/kube_handler_test.go
```diff
@@ -1088,137 +1093,523 @@ func setClientsetData(cs fakeCombinedClientset, namespaceNames []existingNs, err
...
+	if err != nil {
+		t.Fatalf("%+v", err)
+	}
+
+	testCases := []struct {
+		name             string
+		httpValidator    HelmNonOCIValidator
+		fakeHttpError    error
+		fakeRepoResponse *http.Response
+		expectedResponse *ValidationResponse
```
absoludity, line 1199: This one for testing the non-OCI validation. It's just unit-testing the `Validate` method only: for a given upstream response, what's the validation response to the client.

pkg/kube/kube_handler_test.go
```diff
@@ -1088,137 +1093,523 @@ func setClientsetData(cs fakeCombinedClientset, namespaceNames []existingNs, err
...
-					t.Errorf("mismatch (-want +got):\n%s", cmp.Diff(tc.expectedHeaders, cli.(*httpclient.ClientWithDefaults).DefaultHeaders))
-				}
-			} else {
-				httpValidatorAppRepo := httpValidator.getAppRepo()
-				if httpValidatorAppRepo != appRepo {
-					t.Errorf("Unexpected Repos: %v", cmp.Diff(appRepo, httpValidatorAppRepo))
-				}
+// makeTestOCIServer returns a small test double for an OCI server that handles requests
+// for tags/list and manifest only.
+func makeTestOCIServer(t *testing.T, registryName string, repos map[string]fakeOCIRepo, requiredAuthHeader string) *httptest.Server {
```
absoludity, line 1260: Reality is, it's not great to only unit-test the internal functions, we want to be able to unit test the actual exposed handler as well. This wasn't possible for the OCI success case (which I guess is why you didn't get to add them :) ) without either a massive refactoring to enable injecting fake clients, *or* by creating a test double for the upstream OCI server, which is the path I've gone for here. For a given registry and set of repos, it responds to the `tag/list` and `manifest` endpoints only:

pkg/kube/kube_handler_test.go
```diff
@@ -1088,137 +1093,523 @@ func setClientsetData(cs fakeCombinedClientset, namespaceNames []existingNs, err
...
+	}))
+}
+
+func TestOCIValidate(t *testing.T) {
+	registryName := "bitnami"
+	testCases := []struct {
+		name             string
+		repos            map[string]fakeOCIRepo
+		validator        HelmOCIValidator
+		expectedResponse *ValidationResponse
```
absoludity, line 1303: With the test double, we can now unit-test the specific OCI Validate() method here, as well as the general handler (including authn) below.

PR #3677, opened 2021-10-27
Reduce memory consumption `GetCharts`

antgamdia:
### Description of the change

TBD: it's WIP: i'm using this PR to trigger the e2e tests easily.

now I've just replaced the cache key, instead of reading the whole repo content again and again, I use the contentLength as the key. This PR is to see if it results in an improvement.
Note that just using that length may cause some false positive cache hits and false negative cache misses. It is

Mem reduced from 512 to 128 to trigger more failures.

https://github.com/ReneKroon/ttlcache

### Applicable issues

- fixes #

### DIFF
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -1657,7 +1657,7 @@ kubeappsapis:
   resources:
     limits:
       cpu: 250m
-      memory: 128Mi
+      memory: 256Mi
     requests:
       cpu: 25m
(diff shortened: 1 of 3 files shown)


antgamdia:
Closing in favor of https://github.com/kubeapps/kubeapps/pull/3699 and https://github.com/kubeapps/kubeapps/pull/3700, as I wanted to split this PR into two, so that I could explain better what each one addresses.

PR #3689, opened 2021-10-28
Avoid requesting repo in getChart 

antgamdia:
### Description of the change

WIP: I think we can avoid fetching the whole repo if we pass when URL beforehand.
Checking if CI agrees with me

### Applicable issues

- fixes #


antgamdia:
Closing in favor of https://github.com/kubeapps/kubeapps/pull/3699 and https://github.com/kubeapps/kubeapps/pull/3700, as I wanted to split this PR into two, so that I could explain better what each one addresses.

Issue #3567, opened 2021-10-08
Memory issues in kubeappsapis/helm-related logic: `getChart` fetches the whole repo each time it's invoked


antgamdia:
We have had some reports of the `kubeops` being OOMKilled in the current version (not after performing go updates like in the https://github.com/kubeapps/kubeapps/pull/3483#issuecomment-926699100 issue).

Furthermore, we noticed some transient errors while releasing that are consistent with a pod being killed arbitrarily, so this possibility should be considered and investigated.

PR #3699, opened 2021-10-29
Improve Kubeappsapis default config

antgamdia:
### Description of the change

As outlined in https://github.com/kubeapps/kubeapps/issues/3567#issuecomment-954311431, this PR addresses one of the proposals to mitigate the effect of some memory peeks we don't control (that is, instant allocated memory prior to a GC pass). It mostly limits the number of concurrent operations. For instance, with 256MB only ~3-4 upgrades operations can be performed, due to we (well, the Helm library in this case) need to parse the data, validate it and other stuff. 

Also, this PR increases the number of default replicas (as in the original Kubeops pod) from `1` to  `2`. This way, in case of an OOMKilled event, the user won't notice any error (unless they were doing many concurrent actions and they continue doing so).

Finally, it decreases the temporary memory increase we did for merging the deps PR. The value decreases from `512` to `256` MB.

### Benefits

This PR tries to increase the velocity the GC is called (it has a "price", which is a slightly higher CPU consumption, but since we don't have any CPU-intensive process, I think the tradeoff is worth it. Nevertheless. the chosen value `GOGC` is rather arbitrary so a more thoughtful analysis would be required in order to determine the value that maximizes the overall performance. Note this is not a trivial thing and ever there are some commercial solutions for doing so ([example](https://concertio.com/blog/optimizing-the-go-garbage-collector-and-concurrency/)).

### Possible drawbacks

Even though this PR does its best to improve the overall performance and error rate, some changes in the code are still required. Also, note these changes just mitigates the problem, but do not solve it completely.

### Applicable issues

- fixes https://github.com/kubeapps/kubeapps/issues/3567

### Additional information

N/A


chart/kubeapps/Chart.yaml
```diff
@@ -33,4 +33,4 @@ maintainers:
 name: kubeapps
 sources:
   - https://github.com/kubeapps/kubeapps
-version: 7.5.11-dev0
+version: 7.5.11-dev1
```
absoludity, line 36: Small update after chart changes resynced.
```suggestion
version: 7.5.12-dev1
```

chart/kubeapps/values.yaml
```diff
@@ -1626,7 +1626,7 @@ kubeappsapis:
     pullSecrets: []
   ## @param kubeappsapis.replicaCount Number of frontend replicas to deploy
   ##
-  replicaCount: 1
+  replicaCount: 2
```
absoludity, line 1629: Thanks, this should have been updated to 2 once we started using the kubeapps-apis in released code. (Or even better, I should have made it 2 from the start even when we weren't using it since it wasn't deployed in that case, zero cost without having to remember).

absoludity (approved):

Issue #2363, opened 2021-02-10
Request to install or upgrade helm chart times out


absoludity:
### Description:

Most requests within Kubeapps return in a reasonable time, but the ones which standout (and can cause timeouts in slower environments) are the PUT to `/api/tiller-deploy/v1/clusters/tmc-cluster-1/namespaces/default/releases/<release-name>` when installing (or similar when upgrading).

This is because the chart is downloaded, checks are done to ensure the user can create all the resources (which can take time) before the actual install takes place.

### Steps to reproduce the issue:

1. Choose a chart from the catalog
2. Open the network tab of the debugging tools of the browser
3. Click deploy and watch the timing of the request

### Describe the results you received:

![deploy-timing](https://user-images.githubusercontent.com/497518/107461092-613f8880-6bad-11eb-99cd-9c180b020149.png)

### Describe the results you expected:

Under 2s, preferably under 1s.

### Additional information you deem important (e.g. issue happens only occasionally):

It may be worth investigating whether we can avoid doing the checks serially as part of this request: are they done to provide useful feedback before the attempt is made? Could we trigger this (and wait for the result) before the user clicks deploy (ie. as soon as the page to deploy something is displayed?

Or can we or otherwise do the checks async so that the request is service directly, the UI updates indicating that the Kubeapps is checking to ensure they can deploy the chart etc., before updating to do the actual install?

andresmgot:
> This is because the chart is downloaded, checks are done to ensure the user can create all the resources (which can take time) before the actual install takes place.

The checks are not actually done (I said that, but forgot that was just for tiller-proxy). What we do is: 1. get the apprepo 2.  download the chart. 3. Get the registry secrets (if needed) 4. install the chart. I assume 2. is what's getting you the slow response but I don't think we can do much about it.

BTW, we are configuring NGINX to avoid timing out: https://github.com/kubeapps/kubeapps/blob/master/chart/kubeapps/templates/kubeapps-frontend-config.yaml#L16 I guess there is some extra jump (pinniped-prpoxy?) that requires the same config.

doddatpivotal:
[...]

PR #3700, opened 2021-10-29
Pass the tarball file in the `getChart` function to avoid re-fetching the repo

antgamdia:
### Description of the change

As outlined in https://github.com/kubeapps/kubeapps/issues/3567#issuecomment-954311431, this PR addresses another proposal to mitigate the effect of some memory peeks. In particular, this PR aims at solving https://github.com/kubeapps/kubeapps/issues/2363.

We simply pass the tarball URL (which we retrieve beforehand from cache), so we skip the expensive fetchRepo calls.

### Benefits

Even if the memory peaks can still occur, they won't be caused by our code, so to speak. The deploy/update chart actions are expensive in that they required some ulterior validation performed in the Helm/k8s code. About 50 MB or so, meaning 2-3 concurrent actions in a 256MB system. 

### Possible drawbacks

I'd really like to have removed the expensive code, but it is still used in Kubeops (just in the create/update actions). However, since it is not deprecated yet, I've just remained it untouched.

### Applicable issues

- fixes https://github.com/kubeapps/kubeapps/issues/3567

### Additional information

N/A

### DIFF
--- a/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
+++ b/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
@@ -1014,6 +1014,16 @@ func (s *Server) fetchChartWithRegistrySecrets(ctx context.Context, chartDetails
 	chartID := fmt.Sprintf("%s/%s", appRepo.Name, chartDetails.ChartName)
 	log.Infof("fetching chart %q with user-agent %q", chartID, userAgentString)
 
+	// Look up the cachedChart cached in our DB to populate the tarball URL
+	cachedChart, err := s.manager.GetChartVersion(chartDetails.AppRepositoryResourceNamespace, chartID, chartDetails.Version)
+	if err != nil {
+		return nil, nil, status.Errorf(codes.Internal, "Unable to fetch the chart %s (version %s) from the namespace %q: %v", chartID, chartDetails.Version, chartDetails.AppRepositoryResourceNamespace, err)
+	}
+	var tarballURL string
+	if cachedChart.ChartVersions != nil && len(cachedChart.ChartVersions) > 0 && cachedChart.ChartVersions[0].URLs != nil {
+		tarballURL = cachedChart.ChartVersions[0].URLs[0]
+	}
+
 	// Grab the chart itself
 	ch, err := handlerutil.GetChart(
 		&chartutils.Details{
@@ -1021,6 +1031,7 @@ func (s *Server) fetchChartWithRegistrySecrets(ctx context.Context, chartDetails
 			AppRepositoryResourceNamespace: appRepo.Namespace,
 			ChartName:                      chartDetails.ChartName,
 			Version:                        chartDetails.Version,
+			TarballURL:                     tarballURL,
 		},
 		appRepo,
 		caCertSecret, authSecret,
--- a/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
+++ b/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
@@ -1020,7 +1020,10 @@ func (s *Server) fetchChartWithRegistrySecrets(ctx context.Context, chartDetails
 		return nil, nil, status.Errorf(codes.Internal, "Unable to fetch the chart %s (version %s) from the namespace %q: %v", chartID, chartDetails.Version, chartDetails.AppRepositoryResourceNamespace, err)
 	}
 	var tarballURL string
-	if cachedChart.ChartVersions != nil && len(cachedChart.ChartVersions) > 0 && cachedChart.ChartVersions[0].URLs != nil {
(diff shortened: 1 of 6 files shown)


cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -1014,13 +1014,24 @@ func (s *Server) fetchChartWithRegistrySecrets(ctx context.Context, chartDetails
 	chartID := fmt.Sprintf("%s/%s", appRepo.Name, chartDetails.ChartName)
 	log.Infof("fetching chart %q with user-agent %q", chartID, userAgentString)
 
+	// Look up the cachedChart cached in our DB to populate the tarball URL
+	cachedChart, err := s.manager.GetChartVersion(chartDetails.AppRepositoryResourceNamespace, chartID, chartDetails.Version)
+	if err != nil {
+		return nil, nil, status.Errorf(codes.Internal, "Unable to fetch the chart %s (version %s) from the namespace %q: %v", chartID, chartDetails.Version, chartDetails.AppRepositoryResourceNamespace, err)
+	}
+	var tarballURL string
+	if cachedChart.ChartVersions != nil && len(cachedChart.ChartVersions) > 0 && cachedChart.ChartVersions[0].URLs != nil {
```
absoludity, line 1023: Given that we're requesting a particular version from the cache, why do we not check here that the `cachedChart.ChartVersions == 1` , rather than ` > 0`? Leaves me wondering whether we can get different values here?

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -1014,13 +1014,24 @@ func (s *Server) fetchChartWithRegistrySecrets(ctx context.Context, chartDetails
...
+		tarballURL = cachedChart.ChartVersions[0].URLs[0]
```
absoludity, line 1024: Similarly here. I just checked the bitnami index.yaml and see that the first URL is always the tarball, but worth a comment since it's not obvious when looking at this why the first URL can just be used. Perhaps just with a link to https://helm.sh/docs/topics/chart_repository/#the-index-file

pkg/chart/chart.go
```diff
@@ -323,24 +331,32 @@ func (c *HelmRepoClient) Init(appRepo *appRepov1.AppRepository, caCertSecret *co
 
 // GetChart retrieves and loads a Chart from a registry in both
 // v2 and v3 formats.
```
absoludity, line 333: Thanks for cleaning up the imports etc. The above comment is probably now untrue too.

pkg/chart/chart.go
```diff
@@ -323,24 +331,32 @@ func (c *HelmRepoClient) Init(appRepo *appRepov1.AppRepository, caCertSecret *co
...
-	if err != nil {
-		return nil, err
+	var chartURL string
+	var err error
+	if details.TarballURL != "" {
+		chartURL, err = resolveChartURL(repoURL, details.TarballURL)
+		if err != nil {
+			return nil, err
+		}
+	} else {
```
absoludity, line 345: Excellent, so this is the main change: don't fetch the index if we already have the tarball URL.

What are the kubeops cases where details.TarballURL is "", out of interest?

absoludity (approved):
Excellent, looking forward to that landing :)

pkg/chart/chart.go
antgamdia: It is just the same scenario as in the Helm-direct plugin: creating and upgrading releases. Since these operations are no longer used, it is safe to delete the expensive if-branch once we deprecate this service.

https://github.com/kubeapps/kubeapps/blob/f9f24694b1fb5717ba10d1a6101cbce6d5aad41e/cmd/kubeops/internal/handler/handler.go#L188

https://github.com/kubeapps/kubeapps/blob/f9f24694b1fb5717ba10d1a6101cbce6d5aad41e/cmd/kubeops/internal/handler/handler.go#L241

pkg/chart/chart_test.go
```diff
@@ -512,8 +512,14 @@ func TestGetChart(t *testing.T) {
...
 		chartVersion  string
 		userAgent     string
+		tarballURL    string
 		errorExpected bool
 	}{
+		{
+			name:         "gets the chart with tarballURL",
+			chartVersion: "5.1.1-apiVersionV1",
+			tarballURL:   "http://example.com/nginx-5.1.1-apiVersionV1.tgz",
+		},
```
antgamdia, line 522: Added missed test case for the `getChart`  function itself

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
antgamdia: Yep, I wasn't totally sure the query returns always a single chartVersion (albeit it should), that's why relaxed the condition. But, yep, we better set it to `1` so that we can detect if more than one result is returned.

PR #3704, opened 2021-10-29
Fix errata in CI config

antgamdia:
### Description of the change

As expected, after the changes in https://github.com/kubeapps/kubeapps/pull/3676, there were some minor issues. In this case, a wrong variable.

### Benefits

Master builds will pass again

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

N/A


antgamdia:
Rubber-stamping it as it is just a trivial change.

PR #3710, opened 2021-11-02
Two documentation uses of `enable` that I missed in #3590

absoludity:

### Description of the change

Just two cases of `enable` that I missed, in the documentation for pinniped-proxy setup.

Just noticed while pointing someone else at the docs.

### Benefits

When we do the next release, people will use the consistent key rather than the deprecated key..

### Possible drawbacks

Until we release, people need to use the old key (the new release will support both for now).

For this reason, I'll not land this change until we release.

### Applicable issues

- Ref #3590


antgamdia (approved):
Thanks for catching it!

Issue #3636, opened 2021-10-22
Charts themselves may need caching in fluxv2 plugin


absoludity:
Just moving from the [PR discussion](https://github.com/kubeapps/kubeapps/pull/3628#issuecomment-948340352) to an actual issue, as requested by @gfichtenholt , where Greg said:

> I know why that is. I'm not caching any specific charts (only parsed repo index). (Some of the) chart details have to come from a tarball, so if one isn't already there I create a HelmChart on-the-fly, wait while until it reconciles, get the tarball, and then clean up by deleting the HelmChart. I'm a little surprised that it takes that long, I have seen much shorter times, bit I digress. So 2nd call wouldn't be any faster than the 1st. If I didn't do the clean up, it obviously would be, but I didn't want potentially unbound number of HelmCharts laying around after a while.
> 
> One obvious solution would be to cache package details (i.e. introduce another resource-based watcher cache)...
> 
> Another solution might be to make the cleanup process a bit more sophisticated, to make it kind of a garbage collector, that keeps around a around , say, 5 of LRU HelmCharts. That way the number of HelmCharts left around is still bounded. Back in school they taught me if a user clicks on one package, they're likely to click there again, so this would solve it too.
> 
> Neither of these approaches would not speed up the 1st call, but either one would make the cost of 2nd and subsequent calls to `GetAvailablePackageDetail` insignificant.

This was in response to my comment that:
> Just noted that fetching the available package detail appears to be taking around 2.5s (see below). Not a big deal, but just checking in case @gfichtenholt can make it quicker (he's good at that :P )
>
> ![flux-apache](https://user-images.githubusercontent.com/497518/138218705-c3629e7a-2936-4ad6-ab18-90236dae4974.png)

absoludity:
> Neither of these approaches would not speed up the 1st call, but either one would make the cost of 2nd and subsequent calls to `GetAvailablePackageDetail` insignificant.

Yep. Another possibility is to do something similar to helm where a job is started and caches the files for the latest few versions of each chart. That way clicking on an available app summary would show the details very quickly, where as if the user then switches to an older version they get a 2s delay which is OK.

gfichtenholt:
[...]

PR #3715, opened 2021-11-03
flux plugin: add cluster to responses and introduce maxmemory setting on redis to limit size of the cache

gfichtenholt:
Two unrelated changes:
1) to address https://github.com/kubeapps/kubeapps/pull/3686#issue-1038093832 which adds cluster name to responses from flux plug in
2) introduce "maxmemory" setting on redis configuration to limit the size of the cache which was discussed in https://github.com/kubeapps/kubeapps/issues/3636#issuecomment-953742312. The trick was to change the back-end not to treat the cache as the "source-of-truth" but rather as a "best effort" kind of "source" of information
There are more changes coming down the pipeline related to this. Some unit tests I want to write which may take a little which and lots of little TODOs to be addressed in the code


absoludity (approved):
Thanks for the cluster update.

Issue #3221, opened 2021-07-29
Carvel package support in Kubeapps


absoludity:
### Description:

We're currently working on plugin-based support for different packaging systems, moving across our existing Helm support as well as adding fluxv2 support.

We included Carvel support in the original POC, but due to it changing constantly at the time, as well as a higher priority for flux support, we paused the carvel work to focus on flux and porting the existing helm support to a plugin.

We'd love to come back and add a Carvel/kapp-controller plugin once other priorities are completed (especially as it's now stable).

antgamdia:
Just adding a list to track the remaining operations:

- [x] List available packages
  - [x] GetAvailablePackageSummaries: https://github.com/kubeapps/kubeapps/pull/3728
  - [x] GetAvailablePackageDetail: https://github.com/kubeapps/kubeapps/pull/3735
  - [x] GetAvailablePackageVersions: https://github.com/kubeapps/kubeapps/pull/3734
- [ ] List installed packages
  - [ ] GetInstalledPackageSummaries
  - [ ] GetInstalledPackageDetail
- [ ] Action "buttons"
  - [ ] CreateInstalledPackage
  - [ ] UpdateInstalledPackage
  - [ ] DeleteInstalledPackage

Additionally, we should be fixing some minor UI issues and features we encounter (e.g., icons, filtering by package plugin, etc...) [will update this comment to track them]

antgamdia:
I've created a bunch of PRs aimed at having initial Carvel support. However, the final result is far from being seamless and smooth. 

https://user-images.githubusercontent.com/11535726/142704788-5be9d832-52d7-4dde-ab33-ecd683c56e82.mp4

For instance, just in this demo I'd remark a couple of issues:

- The UI redirects when not every resource is created. Currently,  we are throwing errors when sth is not found. Not the best UX
- The loading times for the catalog are way too long... I'd foresee we're gonna need a cache (even if I tried using goroutines... it seems it's not enough)
  - Especially if you add the Bitnami repo as Carvel packages... 
- The demo package gets installed, but I' haven't been able to get to work another pkg but this one. I mean, not having a default values.yaml hinders a quick

PR #3724, opened 2021-11-04
Update kapp-controller plugin to current version.

absoludity:

### Description of the change

This PR just updates the (very) outdated kapp-controller plugin to match the current carvel k8s API.

The nice thing is that since the Kubeapps dashboard is using the aggregated API, running Kubeapps with both helm and carvel plugins:

```diff
--- a/docs/user/manifests/kubeapps-local-dev-values.yaml
+++ b/docs/user/manifests/kubeapps-local-dev-values.yaml
@@ -15,6 +15,11 @@ postgresql:
   existingSecret: postgresql-db
 kubeappsapis:
   replicaCount: 1
+  enabledPlugins:
+    - helm
+    - kapp-controller
+  image:
+    tag: dev5
 ingress:
   enabled: true
   hostname: localhost
```

enables viewing a combined catalog:

kubeapps-carvel

### Benefits

Carvel plugin works (only get available packages, and even that is missing data).

### Possible drawbacks

None.

### Applicable issues

- Ref #3221

### Additional information

cc @migmartri 

I'll follow-up with another PR that adds more metadata (icon, description etc.) to the available package summaries.

### DIFF
--- a/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server.go
+++ b/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server.go
@@ -39,20 +39,21 @@ import (
 	"google.golang.org/grpc/status"
 )
 
-type clientGetter func(context.Context) (dynamic.Interface, error)
+type clientGetter func(context.Context, string) (dynamic.Interface, error)
 
 const (
-	packagingGroup = "packaging.carvel.dev"
 
 	// See https://carvel.dev/kapp-controller/docs/latest/packaging/#package-cr
+	packageGroup     = "data.packaging.carvel.dev"
 	packageVersion   = "v1alpha1"
-	packageResource  = "PackageInstall"
-	packagesResource = "packageinstalls"
+	packageResource  = "Package"
+	packagesResource = "packages"
 
 	// See https://carvel.dev/kapp-controller/docs/latest/packaging/#packagerepository-cr
-	installPackageVersion = "v1alpha1"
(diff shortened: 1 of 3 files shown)


cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server.go
```diff
@@ -73,13 +74,10 @@ type Server struct {
 // the k8s client config.
 func NewServer(configGetter server.KubernetesConfigGetter) *Server {
 	return &Server{
-		clientGetter: func(ctx context.Context) (dynamic.Interface, error) {
+		clientGetter: func(ctx context.Context, cluster string) (dynamic.Interface, error) {
 			if configGetter == nil {
 				return nil, status.Errorf(codes.Internal, "configGetter arg required")
 			}
-			// The Kapp Controller plugin currently supports interactions with
```
migmartri, line 80: why is it that now we support other clusters?

migmartri (approved):
Thanks @absoludity!

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server.go
absoludity: I'm not seeing any reason why we wouldn't... we use the users cred to whatever cluster is in the context. As long as kapp-controller is installed there, it should work fine (but haven't tested it). I needed to update this because the UI always sends the cluster with the request anyway. If there's a problem we can make it conditional (I'll check it at some point).

PR #3725, opened 2021-11-04
Expose Plugin configuration errors 

### LINKED ISSUE #3722: Plugin configuration errors are hidden rather than exposed
### Description:

Just as a follow-on from #3703, currently if I provide incorrect configuration for the plugins config, [any error is ignored and the defaults are used](https://github.com/kubeapps/kubeapps/blob/master/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go#L110-L121). In practise, swallowing errors like this makes bugs hard to find.

We should always be returning an error (which in turn, causes the service not to start, exposing the error).

1. An a more obvious example, edit the values.yaml with

```diff
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -1607,9 +1607,9 @@ kubeappsapis:
       packages:
         v1alpha1:
           versionsInSummary:
-            major: 3
-            minor: 3
-            patch: 3
+            major: 7
+            minor: "bananas"
+            patch: 2
```
2. Deploy/Upgrade Kubeapps
3. Check the versions in the dropdown

The dropdown has the last 3 patch versions of the last 3 minor versions of the last 3 major versions.

The kubeapps-apis server should fail when calling the Helm plugin's `NewServer`, [just like any other part of the initialization in that function](https://github.com/kubeapps/kubeapps/blob/master/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go#L139-L149).

The test could be updated to expect an error in both cases where it is currently swallowed (config file fails to be opened, config file fails to be parsed as json, config file does not match expected struct fields)

satya-dillikar:
### Description of the change

 kubeapps-apis server fails on parsing invalid Plugin configuration.

### Benefits

Stop the user from feeding invalid plugin config.

### Possible drawbacks

N/A
### Applicable issues

- fixes #3722

### Additional information

N/A

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -86,7 +86,7 @@ parameters:
     default: "master"
   KUBEAPPS_REPO:
     type: "string"
-    default: "kubeapp/kubeapps"
+    default: "kubeapps/kubeapps"
   BRANCH_KUBEAPPS_REPO:
     type: "string"
     default: "master"
--- a/chart/kubeapps/Chart.lock
+++ b/chart/kubeapps/Chart.lock
@@ -1,12 +1,12 @@
 dependencies:
 - name: common
   repository: https://charts.bitnami.com/bitnami
-  version: 1.10.0
+  version: 1.10.1
 - name: postgresql
   repository: https://charts.bitnami.com/bitnami
-  version: 10.12.8
+  version: 10.13.2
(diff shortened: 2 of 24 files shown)


satya-dillikar:
wrong commit.

PR #3726, opened 2021-11-04
expose invalid config errors

### LINKED ISSUE #3722: Plugin configuration errors are hidden rather than exposed

satya-dillikar:
### Description of the change

kubeapps-apis server fails on parsing invalid Plugin configuration.

### Benefits

Stop the user from feeding invalid plugin config.
### Applicable issues

- fixes #3722

### Additional information

N/A


cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -111,18 +111,15 @@ func parsePluginConfig(pluginConfigPath string) VersionsInSummary {
 	if err != nil {
 		// return default value of VersionsInSummary
 		return VersionsInSummary{MajorVersionsInSummary,
-			MinorVersionsInSummary, PatchVersionsInSummary}
+			MinorVersionsInSummary, PatchVersionsInSummary}, nil
```
absoludity, line 114: Isn't this still swallowing the error (in this case, that the file could not be read)? We should be returning an error when we have one.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -111,18 +111,15 @@ func parsePluginConfig(pluginConfigPath string) VersionsInSummary {
...
 	}
 	err = json.Unmarshal([]byte(pluginConfig), &config)
 	if err != nil {
-		// return default value of VersionsInSummary
-		return VersionsInSummary{MajorVersionsInSummary,
-			MinorVersionsInSummary, PatchVersionsInSummary}
+		return VersionsInSummary{}, status.Errorf(codes.InvalidArgument, "plugin-config-path: %s error: %v", string(pluginConfig), err)
```
absoludity, line 118: Note that we parse configs during startup, not when handling grpc requests, so we have no need for grpc related status codes via `status.Errorf` (it is specifically for associating an error code with a grpc request/response cycle). You can simply use `fmt.Errorf("failed to unmarshal plugin config: %w", err)` which wraps the returned error with the specific message that's relevant here (which you then correctly log with `log.Fatalf` below).

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server_test.go
```diff
@@ -1706,11 +1706,13 @@ func TestParsePluginConfig(t *testing.T) {
 		name                    string
 		pluginYAMLConf          string
 		exp_versions_in_summary VersionsInSummary
+		exp_status_code         codes.Code
```
absoludity, line 1709: So this will be `expectedError   error` and the check can use `errors.Is(err, tc.expectedError)`. You'll need to check what kind of error is returned for `json.Unmarshal` in that case etc..

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server_test.go
```diff
@@ -1735,6 +1738,21 @@ core:
...
+core:
+  packages:
+    v1alpha1:
+      versionsInSummary:
+        major: 4
+        minor: 2
+        patch: 1-IFC-123
+      `,
+			exp_versions_in_summary: VersionsInSummary{},
+			exp_status_code:         codes.InvalidArgument,
```
absoludity, line 1755: Great.

absoludity (commented):
In addition, just check that it works if no plugin config has been set (I'm not sure, but guess from the code that you'll currently try to `json.Unmarshal` an empty string), as this is not an invalid setting.

Thanks!

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
satya-dillikar: Initially, I thought not to break any backward compatibility with users who have not provided values.yaml plugin config. As this plugin config is present in the config file, we can now handle it as an error. I have corrected this now.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
satya-dillikar: used 'errors' and removed Code.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server_test.go
satya-dillikar: thanks

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server_test.go
satya-dillikar: Added two new error codes.

satya-dillikar:
Please review my [commit](https://github.com/kubeapps/kubeapps/pull/3726/commits/086a15f21395a83712128dd1d027758b43e46b00)

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
absoludity: Just generally, no need to resolve conversations... usually the person who starts the conversation will resolve it if they want to, but they really help to regain context when switching back and forth between reviews.

> Initially, I thought not to break any backward compatibility with users who have not provided values.yaml plugin config.
> As this plugin config is present in the config file, we can now handle it as an error. I have corrected this now.

Great. Yes, the volume was always mounted, so the error would have only occurred if there was some failure at the OS level. We just shouldn't be swallowing errors at all. There's usually a better way.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server_test.go
```diff
@@ -1704,58 +1705,85 @@ func TestPackageAppVersionsSummary(t *testing.T) {
...
 		exp_versions_in_summary VersionsInSummary
+		exp_error               error
 	}{
 		{
-			name:                    "empty plugin config",
-			pluginYAMLConf:          ``,
+			name:                    "non existing plugin-config file",
+			pluginYAMLConf:          nil,
 			exp_versions_in_summary: VersionsInSummary{0, 0, 0},
+			exp_error:               ErrFileNotFound,
```
absoludity, line 1716: ...which means you won't be able to just check for a canned error that you've created, but rather you need to check for the actual error returned by the underlying call. But that's fine, because the errors package gives you [errors.Is](https://pkg.go.dev/github.com/pkg/errors#Is), which keeps unwrapping an error and checking. So in this case, the exp_error will be whatever the underlying `ioutil.Readfile` returns, which you'll need to find out. Could be an `fs.ErrNotExist` or `fs.PathError`, not sure.

absoludity (commented):
Just a small change needed here so we aren't still throwing away the original error and replacing it with one defined here.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server_test.go
absoludity: After chatting on slack, I took a look and although the above (using `errors.Is`) works fine for the missing file, it doesn't for the `json.Unmarshal` error. We can find the error easily enough by grepping for "cannot unmarshal":

```
➜  v1alpha1 git:(versionInSummary_3722) ✗ rg -i "cannot unmarshal" /usr/local/go/src/encoding/json
/usr/local/go/src/encoding/json/decode_test.go
2182:   {new(net.IP), "json: cannot unmarshal number into Go value of type *net.IP"},

/usr/local/go/src/encoding/json/decode.go
134:            return "json: cannot unmarshal " + e.Value + " into Go struct field " + e.Struct + "." + e.Field + " of type " + e.Type.String()
136:    return "json: cannot unmarshal " + e.Value + " into Go value of type " + e.Type.String()
150:    return "json: cannot unmarshal object key " + strconv.Quote(e.Key) + " into unexported field " + e.Field.Name + " of type " + e.Type.String()
```

Checking `decode.go` [shows that it's an `UnmarshalTypeError`](https://cs.opensource.google/go/go/+/refs/tags/go1.17.3:src/encoding/json/decode.go;l=134), which you'd think would solve the problem, but it doesn't, because the `UnmarshalTypeError` has a pointer receiver, which means the use of `errors.Is` will effectively be doing: `errors.Is(err, &json.UnmarshalTypeError{})`, which, looking at the implementation of `errors.Is`, repeatedly unwraps the `err`, checking each time for equality with the target. But because the target is a pointer, [equality means that they point to the same address](https://golang.org/ref/spec#Comparison_operators), so `errors.Is` always returns false. It appears that `errors.Is` is only intended to work with non-pointer-receivers.

So yes, rather than working around that, as much as we dislike comparing error strings, probably a good balance of pragmatism and effort to just compare the error string for the test :)

satya-dillikar: Thanks for understanding. As discussed on slack, I shall use string.Contains() for comparing and update the PR.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -75,6 +75,11 @@ type VersionsInSummary struct {
 	Patch int `json:"patch"`
 }
 
+var (
+	ErrFileNotFound = fmt.Errorf("file not found")
+	ErrUnmarshal    = fmt.Errorf("json unmarshal failed")
+)
```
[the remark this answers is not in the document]
satya-dillikar: fine, we can move this to test code.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -111,20 +116,15 @@ func parsePluginConfig(pluginConfigPath string) VersionsInSummary {
 
 	pluginConfig, err := ioutil.ReadFile(pluginConfigPath)
 	if err != nil {
-		// return default value of VersionsInSummary
-		return VersionsInSummary{MajorVersionsInSummary,
-			MinorVersionsInSummary, PatchVersionsInSummary}
+		return VersionsInSummary{}, fmt.Errorf("plugin-config-path: %s error: %w", pluginConfigPath, ErrFileNotFound)
```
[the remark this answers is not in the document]
satya-dillikar: ok

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -111,20 +116,15 @@ func parsePluginConfig(pluginConfigPath string) VersionsInSummary {
...
 	}
 	err = json.Unmarshal([]byte(pluginConfig), &config)
 	if err != nil {
-		// return default value of VersionsInSummary
-		return VersionsInSummary{MajorVersionsInSummary,
-			MinorVersionsInSummary, PatchVersionsInSummary}
+		return VersionsInSummary{}, fmt.Errorf("plugin-config: %s error: %w", string(pluginConfig), ErrUnmarshal)
```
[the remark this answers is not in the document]
satya-dillikar: ok

satya-dillikar:
Please review and help to close this.

absoludity (approved):
Thanks Satya. I've retriggered the CI job (seemed to fail setting up the pod from which the CI tests are run).

absoludity:
> Please review and help to close this.

Heh - yes, Antonio and I are subscribed to every issue and PR change, so we'll always see the updates and review ASAP :)

absoludity:
> In addition, just check that it works if no plugin config has been set (I'm not sure, but guess from the code that you'll currently try to `json.Unmarshal` an empty string), as this is not an invalid setting.
> 
> Thanks!

Hi Satya, did you check that the above works (ie. that the helm plugin uses default values if the config is empty)? I haven't looked at the details, but CI is consistently failing with:

```
helm upgrade --install kubeapps-ci --namespace kubeapps bitnami/kubeapps ... --set kubeappsapis.image.tag=build-3348590a91d45653242f552d742e46affbedce84 ...
Release "kubeapps-ci" does not exist. Installing it now.
Error: timed out waiting for the condition
```

I'll check locally later if I get a chance, but my guess is that CI is failing at that point, since it's testing the install of the latest kubeapps chart (values.yaml) from the bitnami repository but with the newly built image, which would mean that an empty config is being unmarshalled resulting in an error. We should be ensuring that the plugin uses the default values if no config was specified in the values (ie. if the config file content is blank, use defaults), rather than depending on it being specified. Again, I'm not 100% that this is the cause, but we should be able to replicate the error locally to see.

satya-dillikar:
> > In addition, just check that it works if no plugin config has been set (I'm not sure, but guess from the code that you'll currently try to `json.Unmarshal` an empty string), as this is not an invalid setting.
> > Thanks!
> 
> Hi Satya, did you check that the above works (ie. that the helm plugin uses default values if the config is empty)? I haven't looked at the details, but CI is consistently failing with:
> 
> ```
> helm upgrade --install kubeapps-ci --namespace kubeapps bitnami/kubeapps ... --set kubeappsapis.image.tag=build-3348590a91d45653242f552d742e46affbedce84 ...
> Release "kubeapps-ci" does not exist. Installing it now.
> Error: timed out waiting for the condition
> ```
> 
> I'll check locally later if I get a chance, but my guess is that CI is failing at that point, since it's testing the install of the latest kubeapps chart (values.yaml) from the bitnami repository but with the newly built image, which would mean that an empty config is being unmarshalled resulting in an error. We should be ensuring that the plugin uses the default values if no config was specified in the values (ie. if the config file content is blank, use defaults), rather than depending on it being specified. Again, I'm not 100% that this is the cause, but we should be able to replicate the error locally to see.

If the plugin config is empty, then we throw the error. See the first if-statement.  I thought you wanted it this way.  As said, I swallowed that error earlier and returned defaults for backward compatibility.

As we have already updated the values.yaml file in the previous PR with default plugin config, I don’t think the error is because of empty plugin config unless the CI is using a very old bitnami image.

Is it possible to see the CI detail logs (kubeapps-apis.log ) or update the bitnami repo with the latest code which has the plugin config in values.yaml

absoludity:
> > > In addition, just check that it works if no plugin config has been set (I'm not sure, but guess from the code that you'll currently try to `json.Unmarshal` an empty string), as this is not an invalid setting.
> > > Thanks!
> > 
> > 
> > Hi Satya, did you check that the above works (ie. that the helm plugin uses default values if the config is empty)? I haven't looked at the details, but CI is consistently failing with:
> > ```
> > helm upgrade --install kubeapps-ci --namespace kubeapps bitnami/kubeapps ... --set kubeappsapis.image.tag=build-3348590a91d45653242f552d742e46affbedce84 ...
> > Release "kubeapps-ci" does not exist. Installing it now.
> > Error: timed out waiting for the condition
> > ```
> >   
> > I'll check locally later if I get a chance, but my guess is that CI is failing at that point, since it's testing the install of the latest kubeapps chart (values.yaml) from the bitnami repository but with the newly built image, which would mean that an empty config is being unmarshalled resulting in an error. We should be ensuring that the plugin uses the default values if no config was specified in the values (ie. if the config file content is blank, use defaults), rather than depending on it being specified. Again, I'm not 100% that this is the cause, but we should be able to replicate the error locally to see.
> 
> If the plugin config is empty, then we throw the error.

So I tested this in my local environment and my guess was slightly wrong: the service is failing to start at that particular point of the CI test because the config path itself is empty, since the previously released chart does not (and should not) set it.

> I thought you wanted it this way.

Not quite, no: we need to raise an error when there is an error with the config value, but that is different to ensuring sensible defaults if the config is not provided at all (which can happen in different scenarios, as it did in CI).

> As said, I swallowed that error earlier and returned defaults for backward compatibility.

We can provide backward compatibility without swallowing any errors.

> 
> As we have already updated the values.yaml file in the previous PR with default plugin config, I don’t think the error is because of empty plugin config unless the CI is using a very old bitnami image.

The step that is failing in CI is when it tries first to install the latest released chart. I'm not sure if it's intentional, but it currently first installs the previously release Kubeapps version but using the freshly-built images (and has done so for a while). The implication being that the new images should be able to run with the previously released chart.

> 
> Is it possible to see the CI detail logs (kubeapps-apis.log )

Yes, but it's even simpler to just reproduce it locally based on the output shown in CI. If I build the image from this branch at the current commit (tagged as dev1) and run the command from CI that is failing to deploy the latest kubeapps released chart but with the new image, I see the same failure:

```
$ helm upgrade --install kubeapps --namespace kubeapps bitnami/kubeapps --set kubeappsapis.image.tag=dev1 --set kubeappsapis.image.repository=kubeapps/kubeapps-apis

k -n kubeapps logs kubeapps-internal-kubeappsapis-56f859965b-rmrv4  | head -n 2
I1115 00:32:31.991124       1 root.go:51] kubeapps-apis has been configured with: server.ServeOptions{Port:50051, PluginDirs:[]string{"/plugins/helm"}, ClustersConfigPath:"/config/clusters.conf", PluginConfigPath:"", PinnipedProxyURL:"http://kubeapps-internal-pinniped-proxy.kubeapps:3333", UnsafeLocalDevKubeconfig:false}
F1115 00:32:32.480345       1 server.go:147] unable to open plugin config at "": open : no such file or directory
```

> or update the bitnami repo with the latest code which has the plugin config in values.yaml

We can't do that without the change to the image itself (ie. we need to wait for the next kubeapps release), since this change sends an extra argument `--config-path` to the binary, which will error for the previous image which doesn't support that option.

But it's trivial to fix in this PR, without swallowing any errors, but simply using a sensible default if the `--plugin-config-path` is not passed to the binary. I've added a commit with this so we can land without any more trouble (hopefully, let's see what CI says :P )

satya-dillikar:
Thanks Michael for your help.

PR #3728, opened 2021-11-05
kapp-controller plugin metadata

absoludity:
- Basic meta, but need to refactor.
- Switch to use PackageMetadata as available package.

### Description of the change

Carvel's kapp-controller changed their API quite significantly since we'd last looked at this plugin.

This PR switches to use the `PackageMetadata` as the "AvailablePackage", as it is the data which is common across versions, and the Package CR as the specific version.

It then adds the remaining metadata so that carvel and helm packages display side-by side without differences as shown:

carvel-2-summaries

Note: it looks like something is incorrect in the UI with the icon being much smaller, but it's just because of the SVG icon in the carvel test package, which has a size much bigger than the actual content, so in its original format looks like:

carvel-test-logo-size

### Benefits

We can browser available packages with the kapp-controller plugin enabled.
(Can't yet look at the details).

### Applicable issues

- Ref #3221

### Additional information

cc @migmartri (for when you're back).

Next up, available package detail.

### DIFF
--- a/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server.go
+++ b/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server.go
@@ -44,10 +44,12 @@ type clientGetter func(context.Context, string) (dynamic.Interface, error)
 const (
 
 	// See https://carvel.dev/kapp-controller/docs/latest/packaging/#package-cr
-	packageGroup     = "data.packaging.carvel.dev"
-	packageVersion   = "v1alpha1"
-	packageResource  = "Package"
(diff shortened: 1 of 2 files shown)


antgamdia (approved):
Ahh, it was the base PR. I think I left comments in others, my bad. 
As I said, we already tested it up in the live demo, so good to go!

absoludity:
> Ahh, it was the base PR. I think I left comments in others, my bad. As I said, we already tested it up in the live demo, so good to go!

That's perfect. They're three separate PRs, each needing a review. I've just stacked them as I hadn't yet landed them (without approval). Thanks.

PR #3735, opened 2021-11-08
Add implementation for GetAvailablePackageDetail for kapp-controller.

absoludity:

### Description of the change

Adds an initial implementation for GetAvailablePackageDetail for the kapp-controller plugin.

### Benefits

Can now click on a tile to see (a somewhat sparse) available package detail:

carvel-available-package-detail

It's sparse because there's no concept of a readme for the cavel package, so currently it's just the long description and support description.

### Applicable issues

- Ref #3221

### Additional information

Next is CreateInstalledPackage, which may also require fetching the config values (as part of getting the details) the way the tanzu cli does (need to find out what this does via kubectl cli/api).

### DIFF
--- a/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server.go
+++ b/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server.go
@@ -298,6 +298,174 @@ func (s *Server) GetAvailablePackageVersions(ctx context.Context, request *corev
 	}, nil
 }
 
+// GetAvailablePackageDetail returns the package metadata managed by the 'helm' plugin
+func (s *Server) GetAvailablePackageDetail(ctx context.Context, request *corev1.GetAvailablePackageDetailRequest) (*corev1.GetAvailablePackageDetailResponse, error) {
+	namespace := request.GetAvailablePackageRef().GetContext().GetNamespace()
+	cluster := request.GetAvailablePackageRef().GetContext().GetCluster()
+	log.Infof("+kapp-controller GetAvailablePackageVersions (cluster=%q, namespace=%q)", cluster, namespace)
+
+	if request.GetAvailablePackageRef().GetContext() == nil {
+		return nil, status.Errorf(codes.InvalidArgument, "No request AvailablePackageRef.Context provided")
+	}
+
+	client, err := s.getDynamicClient(ctx, cluster)
+	if err != nil {
+		return nil, err
+	}
+
+	// Get the metadata first.
+	identifier := request.GetAvailablePackageRef().GetIdentifier()
+	metaGVR := schema.GroupVersionResource{Group: packageGroup, Version: packageVersion, Resource: packageMetadataResources}
+	pkgMetadata, err := client.Resource(metaGVR).Namespace(namespace).Get(ctx, identifier, metav1.GetOptions{})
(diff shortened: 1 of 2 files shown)


cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server.go
```diff
@@ -297,6 +297,174 @@ func (s *Server) GetAvailablePackageVersions(ctx context.Context, request *corev
...
+		detail.LongDescription = longDescription
+	}
+	detail.Readme = longDescription
+
+	supportDescription, found, err := unstructured.NestedString(pkgMeta.Object, "spec", "supportDescription")
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "required field spec.supportDescription not found on kapp-controller packagemetadata: %v\n%v", err, pkgMeta.Object)
+	}
+	if found {
+		detail.Readme = detail.Readme + "\n\n" + supportDescription
```
antgamdia, line 429: Not sure, but instead of just concatenating here, have you tried using some markdown (like creating two sections or so) to make it look less sparse ? Perhaps the final rendered result is worse, haha, not sure.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server.go
```diff
@@ -297,6 +297,174 @@ func (s *Server) GetAvailablePackageVersions(ctx context.Context, request *corev
...
+		return nil, status.Errorf(codes.Internal, "required field metadata.namespace not found on kapp-controller packagemetadata: %v\n%v", err, pkgMeta.Object)
+	}
+	detail.Name = name
+	displayName, found, err := unstructured.NestedString(pkgMeta.Object, "spec", "displayName")
+	if err != nil || !found {
+		return nil, status.Errorf(codes.Internal, "required field spec.displayName not found on kapp-controller packagemetadata: %v\n%v", err, pkgMeta.Object)
+	}
+	detail.DisplayName = displayName
+
+	// Carvel uses base64-encoded SVG data for IconSVGBase64, whereas we need
```
antgamdia, line 385: I guess it is always like that? Or someone as a package maintainer can opt to just put there a URL ?   (I need to refresh my carvel knowledge...)

antgamdia (approved):
Excellent, thank you! In fact, we did test it live in the Tanzu Tuesday demo and run seamlessly haha.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server.go
absoludity: Yes, I did consider it, but given that there's no standard for the format of carvel long description or support description yet, I didn't want to add any markdown. We can consider it in the future once we've some actual examples of carvel packages (interested to see the bitnami ones).

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server.go
absoludity: The actual field name on the API is `IconSVGBase64`, so looks like it is always required in that form (which makes sense if you want to support on-prem use with limited egress).

PR #3736, opened 2021-11-08
flux plugin: add integration test for low memory env and check repo checksum before indexing

gfichtenholt:
This is mostly an incremental PR which isn't meant to address any big issue, but more specifically
1) adds an integration test for scenarios where redis cache is limited in size, has to evict entries in order to make room for new ones and flux plug-in functions as expected . This took a bit of an effort to make work
2) add a checksum comparison for repos before performing expensive operation of indexing 
3) tidy up/fix a few minor issues in unit tests

### DIFF
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -1818,9 +1818,12 @@ redis:
       ## allkeys-lru: evict keys by trying to remove the less recently used (LRU) keys first, in order 
       ## to make space for the new data added
       - "--maxmemory-policy allkeys-lru"
+      ## ref https://stackoverflow.com/questions/22815364/flushall-and-flushdb-commands-on-redis-return-unk-command
+      ## Redis official Helm chart by default disables FLUSHDB and FLUSHALL commands
+    disableCommands: []
   replica:
     ## @param redis.replica.replicaCount Number of Redis&trade; replicas to deploy
-    replicaCount: 0
+    replicaCount: 1
     ## @param master.extraFlags Array with additional command line flags for Redis&trade; replicas
     extraFlags:
       ## The maxmemory configuration directive is used in order to configure Redis to use a specified
@@ -1831,3 +1834,6 @@ redis:
       ## allkeys-lru: evict keys by trying to remove the less recently used (LRU) keys first, in order 
       ## to make space for the new data added
       - "--maxmemory-policy allkeys-lru"
+      ## ref https://stackoverflow.com/questions/22815364/flushall-and-flushdb-commands-on-redis-return-unk-command
+      ## Redis official Helm chart by default disables FLUSHDB and FLUSHALL commands
+    disableCommands: []
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -1834,6 +1834,6 @@ redis:
       ## allkeys-lru: evict keys by trying to remove the less recently used (LRU) keys first, in order 
       ## to make space for the new data added
       - "--maxmemory-policy allkeys-lru"
-      ## ref https://stackoverflow.com/questions/22815364/flushall-and-flushdb-commands-on-redis-return-unk-command
-      ## Redis official Helm chart by default disables FLUSHDB and FLUSHALL commands
+    ## ref https://stackoverflow.com/questions/22815364/flushall-and-flushdb-commands-on-redis-return-unk-command
+    ## Redis official Helm chart by default disables FLUSHDB and FLUSHALL commands
     disableCommands: []
--- a/cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/cache.go
+++ b/cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/cache.go
@@ -345,7 +345,10 @@ func (c NamespacedResourceWatcherCache) onAddOrModify(add bool, unstructuredObj 
 
 	var oldValue []byte
 	if !add {
-		oldValue, _ = c.redisCli.Get(c.redisCli.Context(), key).Bytes()
+		if oldValue, err = c.redisCli.Get(c.redisCli.Context(), key).Bytes(); err != nil {
+			log.Errorf("Failed to get value for object with key [%s] in cache due to: %v", key, err)
+			return err
+		}
 	}
 
 	var newValue interface{}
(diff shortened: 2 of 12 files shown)


chart/kubeapps/values.yaml
```diff
@@ -1818,9 +1818,12 @@ redis:
       ## allkeys-lru: evict keys by trying to remove the less recently used (LRU) keys first, in order 
       ## to make space for the new data added
       - "--maxmemory-policy allkeys-lru"
+      ## ref https://stackoverflow.com/questions/22815364/flushall-and-flushdb-commands-on-redis-return-unk-command
+      ## Redis official Helm chart by default disables FLUSHDB and FLUSHALL commands
+    disableCommands: []
```
absoludity, line 1823: Did you mean to disable those? Or just indicate here that it can be used? Unless we're populating/using it here explicitly, not sure why it's added to our chart.

Also note the comment indentation will need to line up with the field for the readme generator, I think (not 100%). (for the replica below too).

chart/kubeapps/values.yaml
```diff
@@ -1818,9 +1818,12 @@ redis:
...
   replica:
     ## @param redis.replica.replicaCount Number of Redis&trade; replicas to deploy
-    replicaCount: 0
+    replicaCount: 1
```
absoludity, line 1826: Hmm, did we use 0 by default before we had the enable/disable switch? If so, +1, and should it default to 2?

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/cache.go
```diff
@@ -342,36 +343,39 @@ func (c NamespacedResourceWatcherCache) onAddOrModify(add bool, unstructuredObj
 		return nil
 	}
 
+	var oldValue []byte
+	if !add {
+		oldValue, _ = c.redisCli.Get(c.redisCli.Context(), key).Bytes()
```
absoludity, line 348: Not sure if this is ignoring any error, or something else?

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/cache.go
```diff
@@ -398,42 +402,47 @@ func (c NamespacedResourceWatcherCache) onDelete(unstructuredObj map[string]inte
 	}
 
 	if delete {
-		err = c.redisCli.Del(c.redisCli.Context(), key).Err()
+		keysremoved, err := c.redisCli.Del(c.redisCli.Context(), key).Result()
 		if err != nil {
 			log.Errorf("Failed to delete value for object [%s] from cache due to: %v", key, err)
 			return err
+		} else {
+			// debugging an intermittent failure
```
absoludity, line 410: Not sure if you intend to leave this debugging statement, but probably not?

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/cache.go
```diff
@@ -398,42 +402,47 @@ func (c NamespacedResourceWatcherCache) onDelete(unstructuredObj map[string]inte
...
+	//  - Redis.Nil if the key does  not exist or has been evicted due to memory pressure/TTL expiry
+	//
 	// TODO (gfichtenholt) See if there might be a cleaner way than to have onGet() take []byte as
 	// a 2nd argument. In theory, I would have liked to pass in an interface{}, just like onAdd/onModify.
 	// The limitation here is caused by the fact that redis go client does not offer a
 	// generic Get() method that would work with interface{}. Instead, all results are returned as
 	// strings which can be converted to desired types as needed, e.g.
 	// redisCli.Get(ctx, key).Bytes() first gets the string and then converts it to bytes.
 	bytes, err := c.redisCli.Get(c.redisCli.Context(), key).Bytes()
+	// debugging an intermittent issue
```
absoludity, line 431: Same here.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/chart_test.go
```diff
@@ -463,24 +463,7 @@ func TestGetAvailablePackageVersions(t *testing.T) {
...
-					{PkgVersion: "12.10.0", AppVersion: "6.0.12"},
-					{PkgVersion: "12.9.2", AppVersion: "6.0.12"},
-					{PkgVersion: "12.9.1", AppVersion: "6.0.12"},
-					{PkgVersion: "12.9.0", AppVersion: "6.0.12"},
-					{PkgVersion: "12.8.3", AppVersion: "6.0.12"},
-					{PkgVersion: "12.8.2", AppVersion: "6.0.12"},
-					{PkgVersion: "12.8.1", AppVersion: "6.0.12"},
-				},
-			},
+			expectedResponse:   expected_versions_redis,
```
absoludity, line 466: Up to you, but test cases are the one place where I personally think it's worth being verbose, so when you come back, you don't need to scroll through the file to find out what the expectation is, but it's clearly readable as part of each test case.

absoludity (approved):

chart/kubeapps/values.yaml
gfichtenholt: Yes, this is intentional. Bitnami disables FLUSHDB command by default (WTF?) and I need it, so this is what its here for. I will fix the indent

chart/kubeapps/values.yaml
gfichtenholt: looks that way. I was testing with 1 for now. Can switch to 2 later

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/cache.go
gfichtenholt: good point. thanks. Will fix

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/cache.go
gfichtenholt: I will take care of it in due time, it's on my list. I want it there for a little while I am actively working on it

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/cache.go
gfichtenholt: same as above

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/chart_test.go
gfichtenholt: well, then let's agree to disagree on this one. I do see your point. However, while I am actively developing I find that when a test case has a dozen or so scenarios and I only need to debug/look in depth at one or two of them, and need to comment the other ones out. It's a lot easier to do so when the cases are small code line-wise, and a painful to do when there is a lot of scrolling to do and hard to get the open/close comments into the right place. So this style is a trade-off I chose: there are pluses and minutes, and it's just a matter of more pluses than minuses

absoludity: Yep, makes sense :)

PR #3741, opened 2021-11-09
Minor fix in api docs

antgamdia:
### Description of the change

This PR just addresses two minor bugs: a CSS color issue (in the light theme, text inputs color and background were the same) and a wrong path preventing users from using the swagger-ui API client in the new KubeappsAPI.

### Benefits

API docs section won't have those bugs.

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

N/A

### DIFF
--- a/dashboard/public/openapi.yaml
+++ b/dashboard/public/openapi.yaml
@@ -905,7 +905,7 @@ paths:
               schema:
                 type: string
   # Temporarily manually extracted from /cmd/kubeapps-apis/docs/kubeapps-apis.swagger.json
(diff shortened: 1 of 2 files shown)


absoludity (approved):
Sorry, should've checked after adding the upgrade swagger-react.

antgamdia:
> Sorry, should've checked after adding the upgrade swagger-react.

No, no, I think the CSS thing has nothing to do with the update. I noticed it some time ago, but forgot either to fix it or to file an issue... and we hit it when preparing the demo.

PR #3746, opened 2021-11-10
Bump default OLM version up

antgamdia:
### Description of the change

When installing the OLM for testing the icons I noticed there is a newer version of the OLM, so this PR is to update our default version to this one.

### Benefits

Using (and testing in CI) the latest OLM version

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

N/A


absoludity (approved):
+1 assuming that CI passes.

PR #3747, opened 2021-11-10
Tidy up the used icons. Add new pkg icons

antgamdia:
### Description of the change

As new plugins are becoming a real thing, this PR is aimed at adding some icons for a better UX. However, I noticed a couple of things I couldn't resist to address: 1) there were some unused icons; 2) the svg icons were a mess wrt to sizes, metadata, etc.
Therefore, is PR is to:
  - remove unused icons
  - add new carvel and flux icons
  - ensure we are using the official icons (either from vmware brand or from cncf branding portal)
  - manually edit every svg to ensure they are 100x100, so that the ulterior CSS calculations regarding the placing is consistent with the rest of the icons
  - compress the svg to save some bytes

### Benefits

Our plugins icons will look better :P

### Possible drawbacks

N/A

### Applicable issues

- related-ish https://github.com/kubeapps/kubeapps/issues/3221

### Additional information

Demo time:

icons_helm
icons_flux
icons_carvel
icons_olm

icons_helm_min
icons_flux_min
icons_carvel_min
icons_olm_min


dashboard/src/shared/utils.ts
```diff
@@ -85,6 +87,10 @@ export function getPluginIcon(plugin?: Plugin | string) {
     switch (plugin?.name) {
       case PluginNames.PACKAGES_HELM:
         return helmIcon;
+      case PluginNames.PACKAGES_FLUX:
+        return fluxIcon;
+      case PluginNames.PACKAGES_KAPP:
+        return carvelIcon;
```
absoludity, line 93: Created #3748

absoludity (approved):
Thanks Antonio. I love your attention to details like this :) (selfish, I know :P)

For the Helm icon, it's really obvious, but we can't count on the icons being obvious and using words. Do we provide any tool-tip or alt text for these?

dashboard/src/shared/utils.ts
antgamdia: Well, right now we aren't displaying anything, but, yep I agree we should. 
Out of curiosity (and vocabulary in the advertising world), for a background thing like this, I prefer using "isologos" (like helm, text is not separable) and "isotypes" (just the icon, not text) rather than the whole "imagotype" (icon+text, like flux or the OLM).

Some alternatives:

- A1: Use the tag for those non-helm pkgs
image

- A2: add a tooltip
image

- A3: add a tag, same style with the plugin name
image

- A4:  add a tag, different style with the plugin name (simplified)
image

WDYT, @absoludity ?  I'd go with A4, not sure about the color yet (bc we should also see how it looks like in dark mode)

absoludity: Yep, I agree with the simplified additional tag, probably capitalized? (`Helm` or `Flux` or `Carvel`) of A4. Not sure about capitalization given it's not normally used for tags, but in this case it's for a brand/name.

RE color/style: It may be enough if it's always in the same position (bottom right)? (but don't mind either way).

PR #3749, opened 2021-11-11
Add plugin badge in the catalog + minor UI fixes for multiplugin

antgamdia:
### Description of the change

Following up the https://github.com/kubeapps/kubeapps/pull/3747#discussion_r747393066, this is a small PR implementing one of the proposals

### Benefits

It'll be easier to identify which plugin a pkg belongs to?

### Possible drawbacks

N/A

### Applicable issues

- related-ish https://github.com/kubeapps/kubeapps/issues/3221

### Additional information

Before:
image

After: 
image

image


dashboard/src/components/Catalog/PackageCatalogItem.tsx
```diff
@@ -15,17 +15,24 @@ export default function PackageCatalogItem(props: IPackageCatalogItem) {
...
-      repositoryName = availablePackageSummary.availablePackageRef?.identifier.split("/")[0];
+      pkgRepository = availablePackageSummary.availablePackageRef?.identifier.split("/")[0];
+      pkgPlugin = "helm";
+      break;
+    case PluginNames.PACKAGES_FLUX:
+      // TODO: get repo from flux
+      pkgPlugin = "flux";
+      break;
+    case PluginNames.PACKAGES_KAPP:
+      pkgPlugin = "carvel";
```
absoludity, line 31: Perhaps another update to the plugin API, which currently has the plugin details just with the machine-readable name (fluxv2.packages or kapp-controller.packages)... we could think about adding a displayname to avoid the dashboard needing to know about packaging plugins here.

absoludity (approved):
Nice. I'm slightly colourblind, so not the best person to comment on colours, but the dark-mode colour almost looks to my eyes like an alert.

Only other thought, as I mentioned on our other chat, I wonder if capitalization of the plugin brand/name would make it easier to identify, even though it's a tag. But fine either way. +1 from me. Thanks!

dashboard/src/components/Catalog/PackageCatalogItem.tsx
antgamdia: Yep, good idea indeed. I've added a card not to forget. Thanks!

antgamdia:
Oops, sorry, didn't remember :sweat_smile: btw, do you know if can I simulate it in chrome with any of these options? 
image
I'm really willing to ensure the UX is the best for everyone!

> Only other thought, as I mentioned on our other chat, I wonder if capitalization of the plugin brand/name would make it easier to identify, even though it's a tag. But fine either way. +1 from me. Thanks!

Mmm, interesting suggestion. Let me give you two proposals: small-caps and capitalize:

0) current

image

1) small-caps

image

2) capitalize

image


I'd go with the current one, everything lowercased, but no strong opinion here

cc @ppbaena if you want to give your opinion as well

dashboard/src/components/AppList/AppListItem.tsx
```diff
@@ -75,7 +76,7 @@ function AppListItem(props: IAppListItemProps) {
       }
       description={app.shortDescription}
       tag1Content={appStatus}
-      tag1Class={appStatus === "deployed" ? "label-success" : "label-warning"}
+      tag1Class={appReady ? "label-success" : "label-warning"}
```
antgamdia, line 79: Squeezing this change in this PR, just for having a pkg-agnostic CSS here as well :P

antgamdia: And... squeezing also this other change (https://github.com/kubeapps/kubeapps/pull/3749/commits/cdeead47baf509ed83f27b56bda61a5ebdec3994) for a multi-plugin header subtitles:

image

absoludity:
> Oops, sorry, didn't remember sweat_smile btw, do you know if can I simulate it in chrome with any of these options? image I'm really willing to ensure the UX is the best for everyone!

I don't. And it's no problem, I was just trying to say that I'm not a good person to comment on colours, that's all.

> 
> > Only other thought, as I mentioned on our other chat, I wonder if capitalization of the plugin brand/name would make it easier to identify, even though it's a tag. But fine either way. +1 from me. Thanks!

> 
> Mmm, interesting suggestion. Let me give you two proposals: small-caps and capitalize:
> 
[snip]
> 
> I'd go with the current one, everything lowercased (0) or even (3), but no strong opinion here.

Yeah, I'd go with the current one or (4) (first letter capitalized), but don't care either. Note, the suggestion wasn't because of how the word "helm" looked on its own, but rather how it compares to the other tags, in particular, whether it stands out as the package type when next to other tags such as "marketplace". But the colour difference that you did is good for that, I'm sure.

dashboard/src/components/AppList/AppListItem.tsx
absoludity: That looks much clearer... ah, but this is just on the page header for a package detail? At first I thought you'd added the subtitle on the summary badge. But either way, this is excellent, much clearer.

dashboard/src/shared/utils.ts
```diff
@@ -96,3 +98,57 @@ export function getPluginIcon(plugin?: Plugin | string) {
...
+        return "unknown plugin";
+    }
+  } else {
+    switch (plugin?.name) {
+      case PluginNames.PACKAGES_HELM:
+        return "Helm";
+      case PluginNames.PACKAGES_FLUX:
+        return "Flux";
+      case PluginNames.PACKAGES_KAPP:
+        return "Carvel";
```
antgamdia, line 122: > (4) (first letter capitalized),

Done :)

absoludity: > > I don't see any numbers in the ishihara 9 test
> 
> I can replicate it with protanopia (no red) and deuteranopia (no green), so next time I'll double-check all the UI decisions with (at least) these two options.

Heh - Thanks Antonio, but I've never noticed any disadvantage caused by it.

PR #3756, opened 2021-11-15
Add initial GetResources implementation.

absoludity:
Signed-off-by: Michael Nelson <minelson@vmware.com>
### Description of the change

Adds and tests an initial implementation for the GetResources call of the resources plugin.

It'll be followed by further PRs to:
* Filter by the requested resource refs,
* Enable watching for further changes of the requested resource refs
* Replace the use of UnsafeGuessKindToResource.

### Benefits

Enables requesting resources for an installed package.

IRL example:
```
grpcurl -plaintext -d '{"installed_package_ref": {"context": {"cluster": "default", "namespace": "default"}, "plugin": {"name": "helm.packages", "version": "v1alpha1"}, "identifier": "default-kafka-from-flux"}, "watch": true }' -H "Authorization: $token" localhost:8080 kubeappsapis.plugins.resources.v1alpha1.ResourcesService.GetResources{  "resourceRef": {    "apiVersion": "v1",    "kind": "ServiceAccount",    "name": "default-kafka-from-flux"  },
  "manifest": {
    ...
    "@value": "eyJhcGlWZXJzaW9uIjoidjEiLCJhdXRvbW91bnRTZXJ2aWNlQWNjb3VudFRva2VuIjp0cnVlLCJraW5kIjoiU2VydmljZUFjY291bnQiLCJtZXRhZGF0YSI6eyJhbm5vdGF0aW9ucyI6eyJtZXRhLmhlbG0..."
```

where the manifest decodes to:
```
echo "eyJhcGlWZXJzaW9uIjoidjEiLCJhdXRvbW91bnRTZXJ2aWNlQWNjb3VudFRva2VuIjp0cnVlLCJraW5kIjoiU2VydmljZUFjY291bnQiLCJtZXRhZGF0YSI6eyJhbm5vdGF0aW9ucyI6eyJtZXRhLmhlbG0..." | base64 -d | jq .
{
  "apiVersion": "v1",
  "automountServiceAccountToken": true,
  "kind": "ServiceAccount",
  "metadata": {
    "annotations": {
      "meta.helm.sh/release-name": "default-kafka-from-flux",
      "meta.helm.sh/release-namespace": "default"
    },
    "creationTimestamp": "2021-11-10T03:17:31Z",
    "labels": {
      "app.kubernetes.io/component": "kafka",
      "app.kubernetes.io/instance": "default-kafka-from-flux",
      "app.kubernetes.io/managed-by": "Helm",
      "app.kubernetes.io/name": "kafka",
[...]

PR #3757, opened 2021-11-15
Ensure only requested refs are returned.

absoludity:
### Description of the change

Following on from #3756 , this PR adds and tests the functionality to:

* Ensure only requested refs are returned (ie. filter by the requested refs)
* Ensure only requested refs belonging to installed pkg are returned.

### Benefits

We don't return unnecessary resources.
We don't expose arbitrary resources in the installed package namespace.

### Applicable issues

- ref #3403

### DIFF
--- a/cmd/kubeapps-apis/plugins/resources/v1alpha1/server.go
+++ b/cmd/kubeapps-apis/plugins/resources/v1alpha1/server.go
@@ -119,7 +119,7 @@ func (s *Server) GetResources(r *v1alpha1.GetResourcesRequest, stream v1alpha1.R
 				}
 			}
 			if !found {
-				return status.Errorf(codes.InvalidArgument, "requested resource %+v does not belonge to installed package %+v", requestedRef, r.GetInstalledPackageRef())
+				return status.Errorf(codes.InvalidArgument, "requested resource %+v does not belong to installed package %+v", requestedRef, r.GetInstalledPackageRef())
 			}
 		}
 		resourcesToReturn = r.GetResourceRefs()
--- a/cmd/kubeapps-apis/plugins/resources/v1alpha1/server.go
+++ b/cmd/kubeapps-apis/plugins/resources/v1alpha1/server.go
@@ -103,14 +103,34 @@ func (s *Server) GetResources(r *v1alpha1.GetResourcesRequest, stream v1alpha1.R
 		return err
 	}
 
+	var resourcesToReturn []*pkgsGRPCv1alpha1.ResourceRef
+	// If the request didn't specify a filter of resource refs,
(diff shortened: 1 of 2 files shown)


cmd/kubeapps-apis/plugins/resources/v1alpha1/server.go
```diff
@@ -103,14 +103,34 @@ func (s *Server) GetResources(r *v1alpha1.GetResourcesRequest, stream v1alpha1.R
...
+		for _, requestedRef := range r.GetResourceRefs() {
+			found := false
+			for _, pkgRef := range refsResponse.GetResourceRefs() {
+				if resourceRefsEqual(pkgRef, requestedRef) {
+					found = true
+					break
+				}
+			}
+			if !found {
+				return status.Errorf(codes.InvalidArgument, "requested resource %+v does not belonge to installed package %+v", requestedRef, r.GetInstalledPackageRef())
```
antgamdia, line 122:
```suggestion
				return status.Errorf(codes.InvalidArgument, "requested resource %+v does not belong to installed package %+v", requestedRef, r.GetInstalledPackageRef())
```

cmd/kubeapps-apis/plugins/resources/v1alpha1/server.go
```diff
@@ -103,14 +103,34 @@ func (s *Server) GetResources(r *v1alpha1.GetResourcesRequest, stream v1alpha1.R
 		return err
 	}
 
+	var resourcesToReturn []*pkgsGRPCv1alpha1.ResourceRef
+	// If the request didn't specify a filter of resource refs,
+	// we return all those found for the installed package. Otherwise
+	// we only return the requested ones.
+	if len(r.GetResourceRefs()) == 0 {
+		resourcesToReturn = refsResponse.GetResourceRefs()
+	} else {
```
antgamdia, line 112: wrt my previous PR comment - Ah, ok, got it, this 400 is thrown if filtering and nothing is found. 
However... why not a 404-like? I mean, a 400 would assume the client is doing sth wrong, but it might haven caused by querying a transient state, no?

antgamdia (approved):
Thank you! Just wondering if the 400 should be a 404, but fine either way.

cmd/kubeapps-apis/plugins/resources/v1alpha1/server.go
absoludity: Yes, currently it's intended to catch a bad request, but I may have to update the behaviour if we do have the transient situations where the request is good but the refs aren't yet available. In that situation, if wait=true, I'd hold off and keep waiting. But I'll update that when testing IRL from the UI. Thanks!

PR #3766, opened 2021-11-16
Add logging middleware for kubeapps-apis 

satya-dillikar:
### Description of the change

There are two commits addressing the below points:

1)Define a Logger interface in the kubeapps-apis server. A default logger (Klog) is implemented in the server and passed as an input to all plugins. Each plugin can use the given logger interface or create a new one if required.
It is recommended to use the default logger so that any future changes can be done in one place and plugins code need not implement any logging actions.
 See [commit](https://github.com/kubeapps/kubeapps/commit/e04698d3842b1266e841a1386ff85aa26c9f492d)
2) Use an interceptor to automatically log every RPC call to the kubeapps-apis server.
If required an additional client-id can be added as an authorization to further restrict the APIs access from any code. A sample client implementation is added to show this functionality.
 See [commit](https://github.com/kubeapps/kubeapps/commit/1f5109cffbdee7d1f75c917c20a860926b51947f)

### Benefits

1. Common logging infra for all kubeapps-apis plugins.
2. intercepting all RPC calls in one location.

### Applicable issues

- This tries to address the requests in #3385 

### Additional information

N/A

### DIFF
--- a/cmd/kubeapps-apis/client/client.go
+++ b/cmd/kubeapps-apis/client/client.go
@@ -1,0 +1,39 @@
+package main
+
+import (
+	"context"
+	"log"
+
+	packagesGRPCv1alpha1 "github.com/kubeapps/kubeapps/cmd/kubeapps-apis/gen/core/packages/v1alpha1"
+	"github.com/kubeapps/kubeapps/cmd/kubeapps-apis/interceptors"
+	"google.golang.org/grpc"
+)
+
+func main() {
+	ctx := context.Background()
+
+	// Create a connection and add our ClientPingCounter interceptor as a UnaryInterceptor to the connection
+	conn, err := grpc.DialContext(ctx, "localhost:50051", grpc.WithInsecure(), grpc.WithBlock(), grpc.WithChainUnaryInterceptor(
+		//interceptors.Identity{ID: "kubeapps-apis-client-junk"}.UnaryClient,
+		interceptors.Identity{ID: "kubeapps-apis-client"}.UnaryClient,
+	))
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer conn.Close()
+	// A new GRPC client to use
+	client := packagesGRPCv1alpha1.NewPackagesServiceClient(conn)
+
+	request := &packagesGRPCv1alpha1.GetAvailablePackageSummariesRequest{
+		Context: &packagesGRPCv1alpha1.Context{
+			Cluster:   "",
+			Namespace: "kubeapps",
+		},
+	}
+
+	availablePackageSummaries, err := client.GetAvailablePackageSummaries(ctx, request)
+	if err != nil {
+		log.Fatal(err)
+	}
+	log.Println(availablePackageSummaries)
+}
--- a/cmd/kubeapps-apis/core/plugins/v1alpha1/plugins.go
+++ b/cmd/kubeapps-apis/core/plugins/v1alpha1/plugins.go
@@ -160,17 +160,17 @@ func (s *pluginsServer) registerGRPC(p *plugin.Plugin, pluginDetail *plugins.Plu
 	if err != nil {
 		return nil, fmt.Errorf("unable to lookup %q for %v: %w", grpcRegisterFunction, pluginDetail, err)
 	}
-	type grpcRegisterFunctionType = func(grpc.ServiceRegistrar, core.KubernetesConfigGetter, kube.ClustersConfig, string) (interface{}, error)
+	type grpcRegisterFunctionType = func(grpc.ServiceRegistrar, core.KubernetesConfigGetter, kube.ClustersConfig, string, core.Logger) (interface{}, error)
 
 	grpcFn, ok := grpcRegFn.(grpcRegisterFunctionType)
 	if !ok {
-		var dummyFn grpcRegisterFunctionType = func(grpc.ServiceRegistrar, core.KubernetesConfigGetter, kube.ClustersConfig, string) (interface{}, error) {
+		var dummyFn grpcRegisterFunctionType = func(grpc.ServiceRegistrar, core.KubernetesConfigGetter, kube.ClustersConfig, string, core.Logger) (interface{}, error) {
 			return nil, nil
 		}
 		return nil, fmt.Errorf("unable to use %q in plugin %v due to mismatched signature.\nwant: %T\ngot: %T", grpcRegisterFunction, pluginDetail, dummyFn, grpcRegFn)
 	}
 
-	server, err := grpcFn(registrar, clientGetter, s.clustersConfig, pluginConfigPath)
+	server, err := grpcFn(registrar, clientGetter, s.clustersConfig, pluginConfigPath, core.NewBuiltinKlogger())
 	if err != nil {
 		return nil, fmt.Errorf("plug-in %q failed to register due to: %v", pluginDetail, err)
 	} else if server == nil {
--- a/cmd/kubeapps-apis/core/serveopts.go
+++ b/cmd/kubeapps-apis/core/serveopts.go
@@ -19,9 +19,12 @@ package core
 import (
 	"context"
 
+	"flag"
+
 	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
 	"google.golang.org/grpc"
 	"k8s.io/client-go/rest"
+	klogv2 "k8s.io/klog/v2"
 )
 
 // ServeOptions encapsulates the available command-line options.
@@ -47,3 +50,45 @@ type GatewayHandlerArgs struct {
 // that call-sites don't need to know how to obtain an authenticated client, but
 // rather can just pass the request context and the cluster to get one.
 type KubernetesConfigGetter func(ctx context.Context, cluster string) (*rest.Config, error)
+
+type Logger interface {
+	Fatal(args ...interface{})
+	Fatalf(format string, args ...interface{})
+	Error(args ...interface{})
+	Errorf(format string, args ...interface{})
+	Info(args ...interface{})
+	Infof(format string, args ...interface{})
+}
+
+type KubeappsApisLogger struct {
+	//use "k8s.io/klog/v2" as default
+}
+
+func NewBuiltinKlogger() KubeappsApisLogger {
+	klogv2.InitFlags(nil) // initializing the flags
+	flag.Set("v", "5")
+	flag.Parse()
+	//defer klogv2.Flush() // flushes all pending log I/O
+	return KubeappsApisLogger{}
+}
+func (l KubeappsApisLogger) Error(args ...interface{}) {
+	klogv2.Error(args...)
+}
+
+func (l KubeappsApisLogger) Errorf(format string, args ...interface{}) {
+	klogv2.Errorf(format, args...)
+}
+func (l KubeappsApisLogger) Info(args ...interface{}) {
+	klogv2.Info(args...)
+}
+
+func (l KubeappsApisLogger) Infof(format string, args ...interface{}) {
+	klogv2.Infof(format, args...)
+}
+func (l KubeappsApisLogger) Fatal(args ...interface{}) {
+	klogv2.Fatal(args...)
+}
+
+func (l KubeappsApisLogger) Fatalf(format string, args ...interface{}) {
+	klogv2.Fatalf(format, args...)
+}
--- a/cmd/kubeapps-apis/interceptors/interceptors.go
+++ b/cmd/kubeapps-apis/interceptors/interceptors.go
@@ -1,0 +1,65 @@
+package interceptors
+
+import (
+	"context"
+
+	log "k8s.io/klog/v2"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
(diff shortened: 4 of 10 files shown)


absoludity:
Thanks Satya. I'll take a look first thing tomorrow.

> THIS IS JUST A PROPOSAL TO SEEK FEEDBACK. NOTHING IS FINAL!!!

You can mark a PR as a draft PR to indicate this too (saves the all-caps :P )

satya-dillikar:
> Thanks Satya. I'll take a look first thing tomorrow.
> 
> > THIS IS JUST A PROPOSAL TO SEEK FEEDBACK. NOTHING IS FINAL!!!
> 
> You can mark a PR as a draft PR to indicate this too (saves the all-caps :P )

Thanks converted to Draft.

absoludity:
Thanks for the outline Satya!

> 
> There are two commits addressing the below points:
> 
> 1)Define a Logger interface in the kubeapps-apis server. A default logger (Klog) is implemented in the server and passed as an input to all plugins. Each plugin can use the given logger interface or create a new one if required. It is recommended to use the default logger so that any future changes can be done in one place and plugins code need not implement any logging actions. See [commit](https://github.com/kubeapps/kubeapps/commit/e04698d3842b1266e841a1386ff85aa26c9f492d)

Nice. Note that I don't think we benefit from defining a logging interface here right now (since we're always passing in a klog instance, and there is no standard logging interface yet in go). So I think it's fine for us to just decide on which logger we'll provide, and pass that type?

If you do need the interface for some reason, you may be able to just embed the klog logger and not need to repeat all the functions to call the embedded implementation. See [embedding in effective go](https://golang.org/doc/effective_go#embedding) for more info.


> 2) Use an interceptor to automatically log every RPC call to the kubeapps-apis server.

Great. It'd be worth thinking about what the interceptor can usefully log for us automatically. Perhaps for unary requests, something like

```
[timestamp] [duration] [status code] [full path]
```

looking like:
```
2021-11-17Z0721 33ms OK kubeappsapis.plugins.resources.v1alpha1.ResourcesService.GetResources
```

Have a play and see what you think. We'll probably need to leave the individual RPCs to log the context (cluster, namespace) as they do currently, since that info is specific to the packages API and not generalisable to an interceptor.

> If required an additional client-id can be added as an authorization to further restrict the APIs access from any code. A sample client implementation is added to show this functionality. See [commit](https://github.com/kubeapps/kubeapps/commit/1f5109cffbdee7d1f75c917c20a860926b51947f)

Yeah, although client interceptors are available, I'm not sure that we need this, given that grpc-web is already including the authz metadata in requests from the standard authz header (I assume it does so itself using a client interceptor, but haven't checked).

Thanks for demoing the possibilities!

satya-dillikar:
@absoludity, Thanks, Michael for your valuable inputs.

I shall drop the Logger interface and funcs UnaryClient() & VerifyUnaryServer().
I shall enhance the LogRequest interceptors as you requested.

Is this sufficient to close issue#[3385](https://github.com/kubeapps/kubeapps/issues/3385)? Let me know should I create a new PR or commit to this PR.

Thanks

antgamdia:
Yeah, thank you for showing us all the possibilities; it is really interesting.

I agree with Michael's point: just passing through the logger we want... but, yep, I don't know if it's gonna work for "any" logger (as they may have different functions signatures), so yep, I'd go with deciding (from the main server) which log we want to pass to the plugins. There is no actual need for the plugins to use different loggers.
I think just this change would be a great PR already.

Secondly, thanks for the interceptors demo. The main short-term benefit for us is defining the chosen log format for the unary requests. However, as we were discussing offline, there are a couple of built-in interceptos that we may want to add in mid-term: the rate limiter and the input validator.
So, I'd go with the interceptor just for the log in another PR and, perhaps, decide from there if we should continue adding more interceptors as follow-up PRs or we just create the corresponding issues haha.

satya-dillikar:
> embedding in effective go

klog does not provide an interface. If we agree on using klog package for all logging purposes in plugins then we can get rid of adding a new interface "Logger". Let me know what do you want.

satya-dillikar:
> Yeah, thank you for showing us all the possibilities; it is really interesting.
> 
> I agree with Michael's point: just passing through the logger we want... but, yep, I don't know if it's gonna work for "any" logger (as they may have different functions signatures), so yep, I'd go with deciding (from the main server) which log we want to pass to the plugins. There is no actual need for the plugins to use different loggers. I think just this change would be a great PR already.
> 
> @antgamdia, thanks for looking into this. If we want the 'main server' to decide on the logger, then I guess we need the interface "Logger" as defined here. But if we agree on a common logging package as @absoludity is alluding then we don't need this new code. I think here we all have to agree on what we want to do next. I am happy to go either way. Please let me know.
> 
> Secondly, thanks for the interceptors demo. The main short-term benefit for us is defining the chosen log format for the unary requests. However, as we were discussing offline, there are a couple of built-in interceptos that we may want to add in mid-term: the rate limiter and the input validator. So, I'd go with the interceptor just for the log in another PR and, perhaps, decide from there if we should continue adding more interceptors as follow-up PRs or we just create the corresponding issues haha.

satya-dillikar:
reopne

absoludity:
> Is this sufficient to close issue#[3385](https://github.com/kubeapps/kubeapps/issues/3385)? Let me know should I create a new PR or commit to this PR.

Fine to commit to this PR. Based on Antonio's comment:

> So, I'd go with the interceptor just for the log in another PR and, perhaps, decide from there if we should continue adding more interceptors as follow-up PRs or we just create the corresponding issues haha.

it sounds like a good strategy would be to update this PR (or create a new one, up to you) with functionality to pass the common logger through to plugins (you may want to include configuring this logger, eg, with a timestamp, not sure), and create a separate PR to add your interceptor with the standard request log showing the duration of an RPC as I outlined earlier. (This just allows each piece of functionality to be reviewed / discussed in isolation).


> @antgamdia, thanks for looking into this. If we want the 'main server' to decide on the logger, then I guess we need the interface "Logger" as defined here. But if we agree on a common logging package as @absoludity is alluding then we don't need this new code. I think here we all have to agree on what we want to do next. I am happy to go either way. Please let me know.

I may have misunderstood, but I thought Antonio was agreeing that the main server would just pass the chosen logger through, rather than suggesting that the main server will require an interface to support different loggers (given there is no standard interface in go, yet).

But I'm fine either way. Let's see what Antonio clarifies.

antgamdia:
> I may have misunderstood, but I thought Antonio was agreeing that the main server would just pass the chosen logger through, rather than suggesting that the main server will require an interface to support different loggers (given there is no standard interface in go, yet).
> 
> But I'm fine either way. Let's see what Antonio clarifies.


Sorry I wasn't clear (yesterday I only slept 2h haha) ! Yep, what I wanted to suggest is just passing the actual logger (like klog) through, and avoiding adding "fancy" stuff that we might not need right now.

satya-dillikar:
> > I may have misunderstood, but I thought Antonio was agreeing that the main server would just pass the chosen logger through, rather than suggesting that the main server will require an interface to support different loggers (given there is no standard interface in go, yet).
> > But I'm fine either way. Let's see what Antonio clarifies.
> 
> Sorry I wasn't clear (yesterday I only slept 2h haha) ! Yep, what I wanted to suggest is just passing the actual logger (like klog) through, and avoiding adding "fancy" stuff that we might not need right now.

I understand we don't want any fancy stuff.
But unfortunately, klog("k8s.io/klog/v2") does not have a func like "New" to create a logger instance.

whereas "log" module, we can use `
var logger = log.New(os.Stdout, "TEST-LOG: ", 5)`.

I thought we want to stick with klog, so I don't have a better option other than the interface. Please let me know if you have any other ideas. 
thanks

absoludity:
@antgamdia I just chatted with Satya about this PR. The actual logging middleware is in #3780, so this PR here is specifically about passing a configured logger through for plugins to use.

What Satya is describing here is that klog doesn't give us a logger object that can be configured and then passed to others (unlike other logging systems, it defines a [global `logging` var](https://github.com/kubernetes/klog/blob/9ad246211af1ed84621ee94a26fcce0038b69cd1/klog.go#L528) in the module (?!) which all call-sites use). So if we want to use klog, it already provides what we want in that it is configured globally (via command-line flags) and everyone who imports klog/v2 uses it with that configuration. It doesn't provide options to modify the log format, but does provide options for handling verbosity and different outputs for different verbosity etc., as well as structured logging. I think when you and I originally discussed the option of passing in a logger, it was on the assumption that we'd instantiate and configure it in the core, then pass the instance to the plugins.

So if we agree that klog is what we want to use, then we don't need to be passing a logger in to the various plugins. Let us know if you agree - if so we'll just close this PR, but otherwise we could look at other loggers that are created as an instance that can be passed. I'd be +1 to just use klog.

satya-dillikar:
> * Add proper request logging middleware or similar for kubeapps-apis #3385

Let me know if we can close this PR and mark #3385 as resolved.

absoludity:
> > * Add proper request logging middleware or similar for kubeapps-apis [Add proper request logging middleware or similar for kubeapps-apis #3385](https://github.com/kubeapps/kubeapps/issues/3385)
> 
> Let me know if we can close this PR and mark #3385 as resolved.

So I was just thinking about this PR while looking at the other logging one. Earlier, I said:

> What Satya is describing here is that klog doesn't give us a logger object that can be configured and then passed to others (unlike other logging systems, it defines a [global `logging` var](https://github.com/kubernetes/klog/blob/9ad246211af1ed84621ee94a26fcce0038b69cd1/klog.go#L528) in the module (?!) which all call-sites use). So if we want to use klog, it already provides what we want in that it is configured globally (via command-line flags) and everyone who imports klog/v2 uses it with that configuration.

But I don't think this is quite true (yet). There are a couple of things that will need a bit of investigation and updating here:

1. We currently don't expose any of the klog command-line flags in our binary. So if we are going to move forward and want the klog logger to be able to be configured globally by command-line flags, we'll need to include these flags (just tested [these instructions](https://flowerinthenight.com/blog/2019/02/05/golang-cobra-klog) and seems to work, though it'd be nice if we just had our nice cobra help with `-h`, needs more investigation). Importantly, this will allow us as devs to run the server with a given log verbosity.
2. For a normal golang binary, it would be true that every module that imports `klog/v2` would then get the same configuration, but we are loading plugins dynamically, so I suspect each plugin will just be loading their own klog module and running with the default klog options, regardless of what is set on the command-line. That is, we need to check whether when we run the kubeapps-apis binary with the (for example) `-v 4` option, that the plugins also log with the same verbosity... I suspect they will not (but will be happy to be surprised). Which will mean, given we're standardising on klog, we may need each plugin to receive the klog flags that we're interested in during plugin registration. (It'd be nice if klog fell back to check env vars, but it doesn't seem to, at least from the quick check I had).

Keen to hear what you find!

satya-dillikar:
> > > * Add proper request logging middleware or similar for kubeapps-apis [Add proper request logging middleware or similar for kubeapps-apis #3385](https://github.com/kubeapps/kubeapps/issues/3385)
> > 
> > 
> > Let me know if we can close this PR and mark #3385 as resolved.
> 
> So I was just thinking about this PR while looking at the other logging one. Earlier, I said:
> 
> > What Satya is describing here is that klog doesn't give us a logger object that can be configured and then passed to others (unlike other logging systems, it defines a [global `logging` var](https://github.com/kubernetes/klog/blob/9ad246211af1ed84621ee94a26fcce0038b69cd1/klog.go#L528) in the module (?!) which all call-sites use). So if we want to use klog, it already provides what we want in that it is configured globally (via command-line flags) and everyone who imports klog/v2 uses it with that configuration.
> 
> But I don't think this is quite true (yet). There are a couple of things that will need a bit of investigation and updating here:
> 
> 1. We currently don't expose any of the klog command-line flags in our binary. So if we are going to move forward and want the klog logger to be able to be configured globally by command-line flags, we'll need to include these flags (just tested [these instructions](https://flowerinthenight.com/blog/2019/02/05/golang-cobra-klog) and seems to work, though it'd be nice if we just had our nice cobra help with `-h`, needs more investigation). Importantly, this will allow us as devs to run the server with a given log verbosity.
> 2. For a normal golang binary, it would be true that every module that imports `klog/v2` would then get the same configuration, but we are loading plugins dynamically, so I suspect each plugin will just be loading their own klog module and running with the default klog options, regardless of what is set on the command-line. That is, we need to check whether when we run the kubeapps-apis binary with the (for example) `-v 4` option, that the plugins also log with the same verbosity... I suspect they will not (but will be happy to be surprised). Which will mean, given we're standardising on klog, we may need each plugin to receive the klog flags that we're interested in during plugin registration. (It'd be nice if klog fell back to check env vars, but it doesn't seem to, at least from the quick check I had).
> 
> Keen to hear what you find!

1. Passing verbosity level (--v=3) as a command-line argument to the binary is a good idea. I shall explore that.
2.  I guess all plugins get the same configuration of klog. Hopefully, all plugins do log with the same verbosity. I shall verify and update here.
Thanks
Satya

satya-dillikar:
@absoludity , please review the Draft PR [3884](https://github.com/kubeapps/kubeapps/pull/3884)

satya-dillikar:
Addressed via PR#3780 and PR#3884

Closing it for now.

PR #3767, opened 2021-11-16
Add support for watching resources via GetResources.

absoludity:

### Description of the change

Adds support for watching resources for changes.

As noted in the comments, I tried to unit-test the watch requests but am failing to trigger the watch events when using the fake client. But the IRL test works fine.

### Benefits

Can begin the implementation to watch many resources from the dashboard with a single request.

### Applicable issues

- ref #3403

### Additional information

When testing IRL, I notice that there are several responses (for the same resource) sent initially. I plan to check the type of the event and see if we can get it to a single response initially. Subsequent responses (when I manually update the resource being watched) are just one per change, as expected.

### DIFF
--- a/cmd/kubeapps-apis/plugins/resources/v1alpha1/server.go
+++ b/cmd/kubeapps-apis/plugins/resources/v1alpha1/server.go
@@ -16,6 +16,7 @@ import (
 	"context"
 	"encoding/json"
 	"os"
+	"sync"
 
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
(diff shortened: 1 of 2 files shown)


cmd/kubeapps-apis/plugins/resources/v1alpha1/server.go
```diff
@@ -130,34 +135,164 @@ func (s *Server) GetResources(r *v1alpha1.GetResourcesRequest, stream v1alpha1.R
...
+
+// ResourceEvent embeds a watch.Event and adds the resource ref.
+type ResourceEvent struct {
+	watch.Event
+	ResourceRef *pkgsGRPCv1alpha1.ResourceRef
+}
+
+// mergeWatchers returns a single watcher for many.
+// Inspired by the fan-in merge at https://go.dev/blog/pipelines
+func mergeWatchers(watchers []*ResourceWatcher) *MultiResourceWatcher {
```
antgamdia, line 214: This is really interesting! Thanks for the link; I'll take a closer look at it (channels and mastering goroutines are still in my todo list of GO things to learn :P)

antgamdia (approved):
Thanks for the details here. Looking forward to giving it a try!

PR #3775, opened 2021-11-17
Replace the unsafe gvk to gvr conversion with restmapper.

absoludity:

### Description of the change

Who would have thought converting from the GroupVersionKind of a yaml resource to the GroupVersionResource required by the API server would be so arduous?!

This PR replaces the unsafe conversion that was used temporarily, instead initializing a RESTMapper from the APIserver (using the service account) at startup and switching to use that at run-time for the conversion.

Follows on from #3767 

### Benefits

We don't depend on known-to-be-unsafe code.

### Applicable issues

- Ref #3403

### DIFF
--- a/cmd/kubeapps-apis/plugins/resources/v1alpha1/server.go
+++ b/cmd/kubeapps-apis/plugins/resources/v1alpha1/server.go
@@ -26,11 +26,16 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/restmapper"
 	log "k8s.io/klog/v2"
 
+	"github.com/kubeapps/kubeapps/cmd/apprepository-controller/pkg/client/clientset/versioned/scheme"
 	"github.com/kubeapps/kubeapps/cmd/kubeapps-apis/core"
 	pkgsGRPCv1alpha1 "github.com/kubeapps/kubeapps/cmd/kubeapps-apis/gen/core/packages/v1alpha1"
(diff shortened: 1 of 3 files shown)


cmd/kubeapps-apis/plugins/resources/v1alpha1/server.go
```diff
@@ -49,9 +54,45 @@ type Server struct {
...
+	// kindToResource is a function to convert a GVK to GVR. Can be replaced
+	// in tests with a dummy version using the unsafe helpers while the real
+	// implementation queries the k8s API for a REST mapper.
+	kindToResource func(meta.RESTMapper, schema.GroupVersionKind) (schema.GroupVersionResource, error)
+}
+
+// createRESTMapper returns a rest mapper configured with the APIs of the
+// local k8s API server. This is used to convert between the GroupVersionKinds
+// of the resource references to the GroupVersionResource used by the API server.
+func createRESTMapper() (meta.RESTMapper, error) {
```
antgamdia, line 70: This is nuts... what an extra amount of coding for performing just a conversion...I guess I lack some K8s internal knowledge here, but I don't know why we need this kind of runtime information.
Thanks a lot for coming up with this solution (I don't know if the restmapper might be useful for additional things as well).

cmd/kubeapps-apis/plugins/resources/v1alpha1/server_test.go
```diff
@@ -105,6 +107,12 @@ func getResourcesClient(t *testing.T, objects ...runtime.Object) (v1alpha1.Resou
 		corePackagesClientGetter: func() (pkgsGRPCv1alpha1.PackagesServiceClient, error) {
 			return pkgsGRPCv1alpha1.NewPackagesServiceClient(conn), nil
 		},
+		// For testing, define a kindToResource converter that doesn't require
+		// a rest mapper.
+		kindToResource: func(mapper meta.RESTMapper, gvk schema.GroupVersionKind) (schema.GroupVersionResource, error) {
+			gvr, _ := meta.UnsafeGuessKindToResource(gvk)
```
antgamdia, line 113: Great, at least the unsafe thing is still useful :P

antgamdia (approved):

cmd/kubeapps-apis/plugins/resources/v1alpha1/server.go
absoludity: It comes down to the fact that a client (`kubectl` or in our case our `client-go` client) cannot safely make assumptions about the mapping between an object (the yaml manifest with the GroupVersionKind) and the resource endpoint in the API. The only safe way to determine the mapping is to query the api server for the API information, which is what this now does.

Issue #3779, opened 2021-11-18
Update dashboard to use new k8s resources endpoint of api server


absoludity:
### Description:

Following on from #3403, which designed and implemented the GRPC api service to fetch (and watch) k8s resources for an installed package, we now need to update the dashboard, in particular the `AppView` and friends, to use this instead connecting directly to the k8s api server.

This will be a major step forward in removing the direct access to the k8s API server.

I'll break down the task with a list as I start looking at the dashboard code.

- [x] Switch dashboard to use the packaging API to get resource refs rather than the Helm only fetching of the release to parse the manifest. (In progress)
- [x] Update the packaging API to include `namespace` as a field on ResourceRef (since current dashboard code supports a different namespace from the release and I don't want to remove that support)
- [x] Update the watch requests to only be for interesting resources (ie. not secrets, config maps etc.)
- [x] Switch the many watch requests for a single grpc call on the new resources API

PR #3781, opened 2021-11-18
Update the ResourceRef message with namespace.

absoludity:
### Description of the change

When beginning the UI/Dashboard work for getting resources, it was apparent that the existing code already handles the case where a resource for a helm deploy may not be in the same namespace as the helm release.

We had talked about this in the design doc and intended to update to support the same with the resource refs, but I'd missed this when implementing the backend.

This PR just updates to include the namespace in the resource ref for an installed package.

There will be one other small change to the backend implementation of the resources plugin which will follow.

### Benefits

We don't remove support for this situation in the dashboard when switching to the new resources API.

### Applicable issues

- ref #3779


antgamdia (approved):
Nice! Thanks for the change! When performing UI changes we notice a lot of use cases and edge cases we weren't aware of.
When implementing the carvel thing... each time I run it with the UI a new problem popped up haha

PR #3782, opened 2021-11-18
Update resources plugin to support resource refs with namespace.

absoludity:
### Description of the change

Follows #3781 and updates the resources plugin itself to support the namespace field in a resource ref.

This means the plugin is now able to get or watch resources from a namespace other than that of the installed package - as long as it belongs to the installed package.

It also makes it much cleaner to handle non-namespaced resources since it's simply depends on whether the resource ref has a namespace (hmm, actually, we've had non-namespaced resources in the kubeapps chart where we've specified the namespace, so we'll still need to ensure, server-side, that the resource is namespaced or not).

### Benefits

See benefits of #3781

### Applicable issues

- ref #3779


cmd/kubeapps-apis/plugins/resources/v1alpha1/server.go
```diff
@@ -198,7 +198,12 @@ func (s *Server) GetResources(r *v1alpha1.GetResourcesRequest, stream v1alpha1.R
 		}
 
 		if !r.GetWatch() {
-			resource, err := dynamicClient.Resource(gvr).Namespace(namespace).Get(stream.Context(), ref.GetName(), metav1.GetOptions{})
+			var resource interface{}
+			if ref.Namespace != "" {
```
antgamdia, line 202: Neat!

antgamdia (approved):
Thanks!!

PR #3784, opened 2021-11-19
Add 'GetAvailablePackageSummaries' impl

antgamdia:
### Description of the change

Initial implementation of the `GetAvailablePackageSummaries` operation

### Benefits

### Possible drawbacks

N/A

### Applicable issues

### Additional information

N/A


cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
```diff
@@ -11,3 +11,119 @@ See the License for the specific language governing permissions and
...
+	namespace := request.GetContext().GetNamespace()
+	cluster := request.GetContext().GetCluster()
+	pageSize := request.GetPaginationOptions().GetPageSize()
+	pageOffset, err := pageOffsetFromPageToken(request.GetPaginationOptions().GetPageToken())
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "unable to intepret page token %q: %v", request.GetPaginationOptions().GetPageToken(), err)
+	}
+	// Assume the default cluster if none is specified
+	if cluster == "" {
+		cluster = s.globalPackagingCluster
```
absoludity, line 41: Same thought as earlier... why do we not just require cluster in the request now?

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
```diff
@@ -11,3 +11,119 @@ See the License for the specific language governing permissions and
...
+
+	if len(pkgMetadatas) > 0 {
+		startAt := -1
+		if pageSize > 0 {
+			startAt = int(pageSize) * pageOffset
+		}
+		for i, pkgMetadata := range pkgMetadatas {
+			wg.Add(1)
+			if startAt <= i {
+				go func(i int, pkgMetadata *datapackagingv1alpha1.PackageMetadata) error {
```
absoludity, line 63: I think this is the same/similar method you're using elsewhere? (sorry, it was hours ago in a different review, not 100%), in which case, perhaps we can DRY it up (eventually, no problem if it's just a comment for now).

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
```diff
@@ -11,3 +11,119 @@ See the License for the specific language governing permissions and
...
+	pageOffset, err := pageOffsetFromPageToken(request.GetPaginationOptions().GetPageToken())
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "unable to intepret page token %q: %v", request.GetPaginationOptions().GetPageToken(), err)
+	}
+	// Assume the default cluster if none is specified
+	if cluster == "" {
+		cluster = s.globalPackagingCluster
+	}
+	// fetch all the package metadatas
+	pkgMetadatas, err := s.getPkgMetadatas(ctx, cluster, namespace)
```
absoludity, line 44: Given that the `PackageMetadata` contains the in-lined SVG icon, this call could itself be causing performance/mem issues for a repo like Bitnami? Worth checking and if so, only request the page size number of results.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
```diff
@@ -11,3 +11,119 @@ See the License for the specific language governing permissions and
...
+		}
+		for i, pkgMetadata := range pkgMetadatas {
+			wg.Add(1)
+			if startAt <= i {
+				go func(i int, pkgMetadata *datapackagingv1alpha1.PackageMetadata) error {
+					defer wg.Done()
+					// fetch the associated packages
+					// Use the field selector to return only Package CRs that match on the spec.refName.
+					fieldSelector := fmt.Sprintf("spec.refName=%s", pkgMetadata.Name)
+					pkgs, err := s.getPkgsWithFieldSelector(ctx, cluster, namespace, fieldSelector)
```
absoludity, line 68: Makes me wonder, given that the package resources are pretty lean (no inlined SVG!), whether it'd be faster here to do a single request for all packages and just filter it down based on the metadata that we're dealing with. If you've got your cluster setup with the bitnami (carvel) repo, you can just try `time kubectl get packages --all-namespaces`.

That'd mean no need for go-routines here, if so.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
```diff
@@ -11,3 +11,119 @@ See the License for the specific language governing permissions and
...
+		}
+	}
+	wg.Wait() // Wait until each goroutine has finished
+
+	// TODO(agamez): the slice with make is filled with <nil>, in case of an error in the
+	// i goroutine, the i-th <nil> stub will remain. Check if 'errgroup' works here, but I haven't
+	// been able so far.
+	// An alternative is using channels to perform a fine-grained control... but not sure if it worths
+
+	// filter out <nil> values
```
absoludity, line 101: Similar comment to the related PR (for getinstalled? not sure): why are we not returning an error if there was an error while fetching the result, rather than returning incomplete results (which it sounds like the case for <nil> values is).

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_adapters.go
```diff
@@ -11,3 +11,46 @@ See the License for the specific language governing permissions and
...
 package main
+
+import (
+	"fmt"
+
+	corev1 "github.com/kubeapps/kubeapps/cmd/kubeapps-apis/gen/core/packages/v1alpha1"
+	datapackagingv1alpha1 "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apiserver/apis/datapackaging/v1alpha1"
+)
+
+func (s *Server) getAvailablePackageSummary(pkgMetadata *datapackagingv1alpha1.PackageMetadata, pkgVersionsMap map[string][]pkgSemver, cluster string) (*corev1.AvailablePackageSummary, error) {
```
absoludity, line 22: Similar comment here: slight preference for a name that doesn't communicate getting/fetching over a network.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_adapters.go
```diff
@@ -11,3 +11,46 @@ See the License for the specific language governing permissions and
...
+	versions := pkgVersionsMap[pkgMetadata.Name]
+	if len(versions) == 0 {
+		return nil, fmt.Errorf("no package versions for the package %q", pkgMetadata.Name)
+	}
+
+	// Carvel uses base64-encoded SVG data for IconSVGBase64, whereas we need
+	// a url, so convert to a data-url.
+	iconUrl := ""
+	if pkgMetadata.Spec.IconSVGBase64 != "" {
+		iconUrl = fmt.Sprintf("data:image/svg+xml;base64,%s", pkgMetadata.Spec.IconSVGBase64)
```
absoludity, line 33: Same comment about Strings.Builder.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_test.go
```diff
@@ -120,3 +123,295 @@ func TestGetClient(t *testing.T) {
...
+			if tc.expectedStatusCode != codes.OK {
+				return
+			}
+
+			if got, want := response.AvailablePackageSummaries, tc.expectedPackages; !cmp.Equal(got, want, ignoreUnexported) {
+				t.Errorf("mismatch (-want +got):\n%s", cmp.Diff(want, got, ignoreUnexported))
+			}
+		})
+	}
+}
```
absoludity, line 417: Similar comment: not sure why the tests were removed to be re-added, rather than just editing them so it's clear what changed (just for next time).

absoludity (approved):

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
antgamdia: I guess

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
antgamdia: Yep, I've used this pagination logic (inspired by Greg's code :P) also in the GetInstalledPackageSummaries. Adding a comment in both places for future improvement. Thanks!

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
antgamdia: I've quickly analyzed the elapsed times of each part: 

- Get all package metadatas: 0.2 seconds, it's ok
- getPkgsWithFieldSelector: **2 seconds avg**
  - The "problem" is in the `resource.List(ctx, listOptions)` code, which is taking from 0.8 to 2.4 seconds in retrieving all the packages whose FieldSelector is foo. Some examples:
    - `resource.List function took 2.245123407s for (3) items for spec.refName=contour.community.tanzu.vmware.com`
    -  `resource.List function took 1.550673668s for (3) items for spec.refName=pkg.test.carvel.dev`
- getPkgVersionsMap function: microseconds, not relevant
- Build the  AvailablePackageSummary object: microseconds, not relevant


Adding a TODO linked to this comment.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
antgamdia: I assumed that delegating the pkg search thing back to kubernetes was quicker than doing it by ourselves, but you are possibly right... we better fetch the whole list -in memory: future issues??- and then we filter.

```
> time kubectl get packages --all-namespaces
real    0m1,464s
```

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
antgamdia: I tend to do it as fail-tolerant as possible; especially, when developing I had some inconsistencies (like in manually deleting CRs) so I opted for being lenient and letting packages have errors.

A reason for that is we currently propagate that error through the chain of calls, so, a maybe temporary inconsistency or error in a single package will prevent the whole packages core API from working until the problem is manually fixed. Is that the intended behavior? I don't know, but, as a user, I don't want to break Kubeapps for a single corrupted package.

Let me leave this TODO as is right now so that we can discuss it in the future. Adding a link to the comment anyway.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
absoludity: No no, all good. I'd mistakenly thought that we'd updated already elsewhere, but what it was was that we updated the flux plugin to accept and use it when set (since the dashboard always sends it). Better to have it consistent with the other plugins. Sorry for the noise.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
absoludity: Excellent - thanks for the info.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
absoludity: Yeah, though if the inline SVG isn't the issue, my original thought above is moot.

PR #3734, opened 2021-11-08
Add implementation of GetAvailablePackageVersions for kapp-controller.

absoludity:
### Description of the change

Adds implementation of GetAvailablePackageVersions for kapp-controller
### Benefits

Step 1 for displaying available pkg detail.

### Applicable issues

- Ref #3221

PR #3785, opened 2021-11-19
Add 'GetAvailablePackageVersions' impl

antgamdia:
### Description of the change

Initial implementation of the `GetAvailablePackageVersions` operation

### Benefits

### Possible drawbacks

N/A

### Applicable issues

### Additional information

N/A

### DIFF
--- a/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
+++ b/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
@@ -127,3 +127,46 @@ func (s *Server) GetAvailablePackageSummaries(ctx context.Context, request *core
 	}
 	return response, nil
 }
+
+// GetAvailablePackageVersions returns the package versions managed by the 'kapp_controller' plugin
+func (s *Server) GetAvailablePackageVersions(ctx context.Context, request *corev1.GetAvailablePackageVersionsRequest) (*corev1.GetAvailablePackageVersionsResponse, error) {
+	log.Infof("+kapp-controller GetAvailablePackageVersions")
+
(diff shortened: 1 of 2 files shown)


cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
```diff
@@ -127,3 +127,46 @@ func (s *Server) GetAvailablePackageSummaries(ctx context.Context, request *core
...
+	cluster := request.GetAvailablePackageRef().GetContext().GetCluster()
+	identifier := request.GetAvailablePackageRef().GetIdentifier()
+
+	// Validate the request
+	if namespace == "" || identifier == "" {
+		return nil, status.Errorf(codes.InvalidArgument, "Required context or identifier not provided")
+	}
+
+	if cluster == "" {
+		cluster = s.globalPackagingCluster
```
absoludity, line 146: As per the previous PR, I see no reason (yet) to not assume invalid argument if the cluster is not set.

absoludity (approved):
Much neater without the dynamic client version. The tests are the same as those added in #3734 I think right? Were the original ones deleted? (I was unsure why I didn't see them deleted here since you're adding them again, but maybe they were removed in a previous branch? Same with the original `GetAvailablePackageVersions` function?

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
antgamdia: See https://github.com/kubeapps/kubeapps/pull/3784#discussion_r755995552

PR #3786, opened 2021-11-19
Add 'GetAvailablePackageDetail' impl

antgamdia:
### Description of the change

Initial implementation of the `GetAvailablePackageDetail` operation

### Benefits

### Possible drawbacks

N/A

### Applicable issues

### Additional information

N/A


cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_adapters.go
```diff
@@ -54,3 +54,83 @@ func (s *Server) getAvailablePackageSummary(pkgMetadata *datapackagingv1alpha1.P
...
+	}
+
+	maintainers := []*corev1.Maintainer{}
+	for _, maintainer := range pkgMetadata.Spec.Maintainers {
+		maintainers = append(maintainers, &corev1.Maintainer{
+			Name: maintainer.Name,
+		})
+	}
+
+	readme := fmt.Sprintf(`## Details
```
antgamdia, line 74: I've changed the previous readme we had to also include some fields in the `package` object.
Still not sure... but here's how it looks like now:

image

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_adapters.go
```diff
@@ -54,3 +54,83 @@ func (s *Server) getAvailablePackageSummary(pkgMetadata *datapackagingv1alpha1.P
...
 	return availablePackageSummary, nil
 }
+
+func (s *Server) getAvailablePackageDetail(pkgMetadata *datapackagingv1alpha1.PackageMetadata, requestedPkgVersion string, foundPkgSemver *pkgSemver, cluster string) (*corev1.AvailablePackageDetail, error) {
+
+	// Carvel uses base64-encoded SVG data for IconSVGBase64, whereas we need
+	// a url, so convert to a data-url.
+	iconUrl := ""
+	if pkgMetadata.Spec.IconSVGBase64 != "" {
+		iconUrl = fmt.Sprintf("data:image/svg+xml;base64,%s", pkgMetadata.Spec.IconSVGBase64)
```
absoludity, line 64: Same thought about Strings.Builder.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_adapters.go
absoludity: Certainly looks better like that. I'd still personally include the long description there somewhere, if set?

Note: related to another PR about skipping available package details when no readme is set, this means that for this plugin the new setting would do nothing even when set, I think?

absoludity (approved):

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_adapters.go
antgamdia: Yes, sure! I just used the information from the package and totally forgot about the pkg metadata object! I've added the long description as well as some support description that is also available.

Re skipping the view, since the latest PR changes, now we don't depend on the readme anymore, so it's not a problem. Anyway, thanks for pointing this out!

PR #3787, opened 2021-11-19
Add 'GetInstalledPackageSummaries' impl

antgamdia:
### Description of the change

Initial implementation of the `GetInstalledPackageSummaries` operation

### Benefits

### Possible drawbacks

N/A

### Applicable issues

### Additional information

N/A


cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
```diff
@@ -238,3 +239,107 @@ func (s *Server) GetAvailablePackageDetail(ctx context.Context, request *corev1.
...
+			}
+		}
+	}
+	// Wait until each goroutine has finished
+	wg.Wait()
+
+	// TODO(agamez): the slice with make is filled with <nil>, in case of an error in the
+	// i goroutine, the i-th <nil> stub will remain. Check if 'errgroup' works here, but I haven't
+	// been able so far.
+	// An alternative is using channels to perform a fine-grained control... but not sure if it worths
```
absoludity, line 324: Yeah. You'd normally avoid multiple go-routines writing to the same slice and instead collect the processed data via a channel as you've noted. But I know you're doing it carefully here so that each go routine is responsible for one index only. Stepping back, I'm not sure why the nil's here would be an issue for you anyway given that we should return an error (and a nil value) if an error has been returned, I'd think?

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_adapters.go
```diff
@@ -134,3 +136,51 @@ func (s *Server) getAvailablePackageDetail(pkgMetadata *datapackagingv1alpha1.Pa
 
 	return availablePackageDetail, nil
 }
+
+func (s *Server) getInstalledPackageSummary(pkgInstall *packagingv1alpha1.PackageInstall, pkgMetadata *datapackagingv1alpha1.PackageMetadata, pkgVersionsMap map[string][]pkgSemver, cluster string) (*corev1.InstalledPackageSummary, error) {
```
absoludity, line 140: I'd normally reserve the `get` prefix for a function if it is getting/fetching something... ie something that is network constrained. What you're doing here is creating an installed package summary based on data that's already been fetched (ie. no network here). Maybe `installedPackageSummaryFromData` or something without `get` would be clearer.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_adapters.go
```diff
@@ -134,3 +136,51 @@ func (s *Server) getAvailablePackageDetail(pkgMetadata *datapackagingv1alpha1.Pa
...
+	versions := pkgVersionsMap[pkgInstall.Spec.PackageRef.RefName]
+	if len(versions) == 0 {
+		return nil, fmt.Errorf("no package versions for the package %q", pkgInstall.Spec.PackageRef.RefName)
+	}
+
+	// Carvel uses base64-encoded SVG data for IconSVGBase64, whereas we need
+	// a url, so convert to a data-url.
+	iconUrl := ""
+	if pkgMetadata.Spec.IconSVGBase64 != "" {
+		iconUrl = fmt.Sprintf("data:image/svg+xml;base64,%s", pkgMetadata.Spec.IconSVGBase64)
```
absoludity, line 151: I wonder if this is one of the things causing the slowness... some of the SVG's are quite big (300k) and we're having to copy them to each and return via the network (rather than a URL to a resource). If so, definitely a place where caching may help us.

In the mean-time, it might be worth replacing the `fmt.Sprintf` here with a https://pkg.go.dev/strings#Builder so that we're not copying the string when assigned below. Instead we can just using `builder.String()` in the final struct below?

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_adapters.go
```diff
@@ -134,3 +136,51 @@ func (s *Server) getAvailablePackageDetail(pkgMetadata *datapackagingv1alpha1.Pa
...
+		InstalledPackageRef: &corev1.InstalledPackageReference{
+			Context: &corev1.Context{
+				Namespace: pkgMetadata.Namespace,
+				Cluster:   cluster,
+			},
+			Plugin:     &pluginDetail,
+			Identifier: pkgInstall.Name,
+		},
+		LatestMatchingVersion: &corev1.PackageAppVersion{
+			PkgVersion: versions[0].version.String(),
```
absoludity, line 168: I don't think this is necessarily true, is it? (Maybe leave at least a comment). I mean, if the `PackageInstall` has a `versionSelection.constraint` of `2.0.0`, and versions has a 3.0.0, 2.1.0 and 2.0.0, then the latest matching version is still 2.0.0, isn't it?

absoludity (approved):

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_adapters.go
absoludity: Another thought: if we find that it is the collection of all the svg data being fetched across the network for each installed package detail and sent all together in the one response that is the issue, we could even add an extra endpoint just to this kapp-controller plugin for `AvailablePackageIcon`, expose it via the gateway, then populate the IconURL of the summary request with the gateway endpoint. This would mean that the response to `GetInstalledPackageSummaries` (and `GetAvailablePackageSummaries`) would just have a normal URL for the icon which would be fetched separately by the browser.

We'd still want to avoid pulling the icon data over the network from the k8s server when getting the metadata for other fields though. Have to see if that's possible.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
```diff
@@ -238,3 +239,107 @@ func (s *Server) GetAvailablePackageDetail(ctx context.Context, request *corev1.
 		AvailablePackageDetail: availablePackageDetail,
 	}, nil
 }
+
+// GetInstalledPackageSummaries returns the installed packagesmanaged by the 'kapp_controller' plugin
```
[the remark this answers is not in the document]
antgamdia: Thank you for catching the typo! I also use this feature the same way; it's a great manner of showing others what you suggest. When it's an isolated change, I rather apply the suggestion, but in this case, as I have to move through different parts and branches it feels more natural doing the changes myself.
As usual, thanks for your kind and detailed explanations :P

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
```diff
@@ -238,3 +239,107 @@ func (s *Server) GetAvailablePackageDetail(ctx context.Context, request *corev1.
...
+	// Retrieve the proper parameters from the request
+	namespace := request.GetContext().GetNamespace()
+	cluster := request.GetContext().GetCluster()
+	pageSize := request.GetPaginationOptions().GetPageSize()
+	pageOffset, err := pageOffsetFromPageToken(request.GetPaginationOptions().GetPageToken())
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "unable to intepret page token %q: %v", request.GetPaginationOptions().GetPageToken(), err)
+	}
+
+	// Assume the default cluster if none is specified
```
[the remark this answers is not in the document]
antgamdia: As I mentioned in https://github.com/kubeapps/kubeapps/pull/3784#discussion_r755995552, let me leave it as is right now. I'm adding a card/issue to discuss what should be doing regarding this. I mean, I'm ok throwing an error, but I'd like to verify first this choose is uniform and consistent with the rest of the functions/plugins.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
```diff
@@ -238,3 +239,107 @@ func (s *Server) GetAvailablePackageDetail(ctx context.Context, request *corev1.
...
+		if pageSize > 0 {
+			startAt = int(pageSize) * pageOffset
+		}
+		for i, pkgInstall := range pkgInstalls {
+			wg.Add(1)
+			if startAt <= i {
+				go func(i int, pkgInstall *packagingv1alpha1.PackageInstall) error {
+					defer wg.Done()
+					// fetch additional information from the package metadata
+					// TODO(agamez): if the repository where the package belongs to is not installed, it will throw an error;
```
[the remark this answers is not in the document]
antgamdia: Ah, it was me that removed a `PackageRepository` (therefore deleting the associated `Package` and `PackageMetadata` ) and noticed that behavior. From a UX point of view, if a user decides to delete the, let's say, the Bitnami repo, but they still have a package installed, if we throw an error here... the package won't be manageable by kubeapps anymore; forcing the user to manually delete it.

What we did in the past for Helm is return the data we possibly can retrieve and then, the UI will decide which actions will be displayed to users. Currently, we say sth like: "ey, we don't have enough info about this pkg, you only can delete it"

So the point of this ToDo is just to keep in mind this possible use case

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
```diff
@@ -238,3 +239,107 @@ func (s *Server) GetAvailablePackageDetail(ctx context.Context, request *corev1.
...
+					// decide whether we should ignore it or it is ok with returning an error
```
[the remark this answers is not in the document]
antgamdia: Same as in https://github.com/kubeapps/kubeapps/pull/3787#discussion_r756370794, currently returning error though. It is something we can decide after when we polish up the UI/UX.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
antgamdia: Yes, I'd have liked to dig into the goroutines+channels (it's one of my pending topics in Go; btw, now I have a lot of interesting Greg's code to learn from, so thank you!) but I didn't want to spend more time here (I mean, before landing something we can start iterating from).

> given that we should return an error (and a nil value) if an error has been returned, I'd think?

That's the point, I'm not so sure we should return a "nil, error" if a goroutine (=single package) fails.
See https://github.com/kubeapps/kubeapps/pull/3784#discussion_r756281729

I'm ok reverting back to the hard-error way, but also happy to discuss longer-term ideas like letting users decide or at least returning "values,error", so that the client/UI was able to decide.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_adapters.go
antgamdia: Yep, I'm with the Java folks who like using getters, but sounds fair :P  However, I feel like I do need a verb for a function name, so... what about `buildInstalledPackageSummary(...` ?

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_adapters.go
antgamdia: Yes, absolutely; I added a temporary value for populating the field, but just forgot to add a comment explaining it :sweat_smile: 
Adding a TODO for now.

PR #3790, opened 2021-11-19
Add 'UpdateInstalledPackage' impl

antgamdia:
### Description of the change

Initial implementation of the `UpdateInstalledPackage` operation

### Benefits

### Possible drawbacks

N/A

### Applicable issues

### Additional information

N/A


cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
```diff
@@ -509,3 +511,96 @@ func (s *Server) CreateInstalledPackage(ctx context.Context, request *corev1.Cre
...
+		return nil, errorByStatus("get", "PackageInstall", installedPackageName, err)
+	}
+
+	// Update the values.yaml values file if any is passed, otherwise, delete the values
+	if values != "" {
+		secret, err := s.newSecret(installedPackageName, values, targetNamespace)
+		if err != nil {
+			return nil, errorByStatus("upsate", "Secret", secret.Name, err)
+		}
+		updatedSecret, err := typedClient.CoreV1().Secrets(targetNamespace).Update(ctx, secret, metav1.UpdateOptions{})
```
absoludity, line 559: Similar to the previous PR, I'd consider only updating the secret once the pkg install itself has successfully updated (if it is updating). We want to minimize the possibility of leaving an inconsistent state (and without transactions, we can't guarantee it).

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
```diff
@@ -509,3 +511,96 @@ func (s *Server) CreateInstalledPackage(ctx context.Context, request *corev1.Cre
...
+		if updatedSecret == nil || err != nil {
+			return nil, errorByStatus("update", "Secret", secret.Name, err)
+		}
+	} else {
+		// Delete all the associated secrets
```
absoludity, line 564: Hmm. Unsure about this but don't see a better option. We don't want to orphan secrets created by us, but neither would I want to delete secrets that are not created by us. Perhaps eventually we can label secrets created by Kubeapps and only delete them if it has the label.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
```diff
@@ -509,3 +511,96 @@ func (s *Server) CreateInstalledPackage(ctx context.Context, request *corev1.Cre
...
+		if reconciliationOptions.Interval > 0 {
+			pkgInstall.Spec.SyncPeriod = &metav1.Duration{Duration: time.Duration(reconciliationOptions.Interval) * time.Second}
+		}
+		if reconciliationOptions.ServiceAccountName != "" {
+			pkgInstall.Spec.ServiceAccountName = reconciliationOptions.ServiceAccountName
+		}
+		pkgInstall.Spec.Paused = reconciliationOptions.Suspend
+	}
+
+	// get rid of the status field, since now there will be a new reconciliation process
```
absoludity, line 585: Is that necessary? I had thought the status section of a resource was only updated by the controller... At least, if I edit the status of a resource via kubectl, it appears to just be ignored (or is very quickly set back, not sure).

absoludity (approved):

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
antgamdia: I'm not sure, I must reckon... I just assumed I should do it as per the snippet from Flux's code (below) to force a reconciliation trigger, but I guess that logic does not make sense here, as kapp-ctrl should be handling every state change and Kubeapps shouldn't be messing around with it.
Deleting it right away!

https://github.com/kubeapps/kubeapps/blob/260fb34fd70cc5d00faf158b517820c8cc34e845/cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/release.go#L458

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
antgamdia: Yes, you're right. Moving the secret handling logic right after we perform the `PackageInstall` update.

I don't think we have transactions in k8s, but we might want to set some finalizers over the resources we create, so that everything got cleaned up automagically? This leads me to think... shouldn't be kapp-ctrl already handling it? should we add some special labels/annotations?

absoludity: Heh, yes, no transactions here.

> This leads me to think... shouldn't be kapp-ctrl already handling it? should we add some special labels/annotations?

I don't think we should add any kapp-ctrler annotations or such, but yep, perhaps (later) we can ensure that we know when a secret was created by Kubeapps on behalf of the user. Nothing for now though.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
absoludity: Yeah - as above, I'm not sure. I just know what happens via the kubectl cli if I try to modify the status. YMMV

PR #3792, opened 2021-11-19
Add 'GetPackageRepositories' impl

antgamdia:
### Description of the change

Initial implementation of the `GetPackageRepositories` operation

### Benefits

### Possible drawbacks

N/A

### Applicable issues

### Additional information

N/A

### DIFF
--- a/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_repositories.go
+++ b/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_repositories.go
@@ -11,3 +11,42 @@
 // limitations under the License.
 // */
 package main
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/kubeapps/kubeapps/cmd/kubeapps-apis/gen/plugins/kapp_controller/packages/v1alpha1"
+	log "k8s.io/klog/v2"
+)
+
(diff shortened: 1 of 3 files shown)


cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_repositories.go
```diff
@@ -11,3 +11,42 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/kubeapps/kubeapps/cmd/kubeapps-apis/gen/plugins/kapp_controller/packages/v1alpha1"
+	log "k8s.io/klog/v2"
+)
+
+// GetPackageRepositories returns the package repositories based on the request managed by the 'kapp_controller' plugin
```
absoludity, line 23: Note: we don't necessarily need this yet. We're not yet using it in the dashboard and the API is something Greg's working on in the background. We just had it initially with the kapp-controller plugin as an example, I think.

absoludity (approved):

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_repositories.go
antgamdia: Yep, I know, but I didn't want to remove the existing code and, given that we are gonna work on that soon, it worths the effort on using the same from/to unstructured approach, for example.

PR #3794, opened 2021-11-22
Update dashboard to fetch resource refs via API.

absoludity:
### Description of the change

This change updates the dashboard so that fetching the list of resource references for an installed package is no longer based on parsing the helm manifest, but instead a call to the apis server.

### Benefits

Other plugins can provide the installed package references for an installed app.

### Applicable issues

- ref #3779

### Additional information

While removing some of the dashboard code here, I noticed two cases that the backend may not yet support, so need to update that in the helm plugin.

### DIFF
--- a/dashboard/src/components/AppView/AppView.test.tsx
+++ b/dashboard/src/components/AppView/AppView.test.tsx
@@ -11,15 +11,15 @@ import {
   InstalledPackageStatus,
   InstalledPackageStatus_StatusReason,
   PackageAppVersion,
+  ResourceRef as APIResourceRef,
   VersionReference,
 } from "gen/kubeappsapis/core/packages/v1alpha1/packages";
 import { Plugin } from "gen/kubeappsapis/core/plugins/v1alpha1/plugins";
-import * as yaml from "js-yaml";
 import * as ReactRedux from "react-redux";
 import { MemoryRouter, Route } from "react-router";
 import ResourceRef from "shared/ResourceRef";
 import { defaultStore, getStore, mountWrapper } from "shared/specs/mountWrapper";
-import { DeleteError, FetchError, IResource } from "shared/types";
+import { DeleteError, FetchError } from "shared/types";
 import { PluginNames } from "shared/utils";
 import AccessURLTable from "./AccessURLTable/AccessURLTable";
 import DeleteButton from "./AppControls/DeleteButton/DeleteButton";
@@ -64,15 +64,6 @@ afterEach(() => {
 });
 
 describe("AppView", () => {
-  // Generates a Yaml file separated by --- containing every object passed.
-  const generateYamlManifest = (items: any[]): string => {
-    let yamlManifest = "";
-    items.forEach(i => {
-      yamlManifest += "---\n" + yaml.dump(i);
-    });
-    return yamlManifest;
-  };
-
   const installedPackage = {
     name: "test",
     postInstallationNotes: "test",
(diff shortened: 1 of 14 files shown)


dashboard/src/components/AppView/AppView.test.tsx
```diff
@@ -472,108 +400,10 @@ describe("AppView", () => {
     });
   });
 
-  it("renders a list of resources", () => {
```
antgamdia, line 475: Just curious, why aren't these tests longer useful?

dashboard/src/shared/types.ts
```diff
@@ -479,6 +480,6 @@ export interface IBasicFormSliderParam extends IBasicFormParam {
 }
 
 export interface CustomInstalledPackageDetail extends InstalledPackageDetail {
-  manifest: any;
+  apiResourceRefs: ResourceRef[];
```
antgamdia, line 483: I don't think it really belongs in here but in the `InstalledPackageDetail` intrerface, I think?

antgamdia (approved):
Nice! Thanks for this last (hope so) step in having the resources view calling the new API. Can't wait to start pruning Kubeops :P

dashboard/src/shared/types.ts
absoludity: Ah - ok. I'll move it and delete the CustomInstalledPackageDetail then.

dashboard/src/components/AppView/AppView.test.tsx
absoludity: These tests are ensuring that given a yaml manifest, that the expected resources (from the refs) are displayed. Since we no longer use a manifest at all, they're no longer relevant. I did move them into the equivalent code in the helm plugin which is returning the resource refs based on the manifest.

absoludity (commented):
> Nice! Thanks for this last (hope so) step in having the resources view calling the new API. Can't wait to start pruning Kubeops :P

Unfortunately not, but it is a big step forward nonetheless. With this change, we can now have installed package details for other plugins (carvel and/or fluxv2) as long as they implement the method to get the refs (I might jump in and do for flux so we get that data.

Once the UI has the refs, it's still using the k8s API server to watch/fetch those resources right now. That'll be the next PRs.

dashboard/src/shared/types.ts
absoludity: Hmm, so just checked: we can't put it on `InstalledPackageDetail` given that `InstalledPackageDetail` is generated from the proto. I'd added it here simply because the manifest was there (and this field replaced the manifest). I think we'll need to instead pass it as a separate value, with the InstalledPackageDetail.

absoludity:
> > Nice! Thanks for this last (hope so) step in having the resources view calling the new API. Can't wait to start pruning Kubeops :P
> 
> Unfortunately not, but it is a big step forward nonetheless. With this change, we can now have installed package details for other plugins (carvel and/or fluxv2) as long as they implement the method to get the refs (I might jump in and do for flux so we get that data.
> 
> Once the UI has the refs, it's still using the k8s API server to watch/fetch those resources right now. That'll be the next PRs.

But we can prune kubeops now, I think... I'll give it a try...

dashboard/src/shared/types.ts
antgamdia: Ahh, I thought the `InstalledPackageDetail` (proto) already has this `ResourceRef` field! Got it then, makes sense to keep it here, temporarily.
Thanks for the clarification!

absoludity: Cool. Just to be clear though, the `InstalledPackageDetail` proto won't include the resource refs, since it's a separate request. We were originally thinking about just including it on the detail, but after discussion, went for a separate call.

PR #3227, opened 2021-07-30
fix(chart): Skip README when it cannot be found

aanthonyrizzo:
Signed-off-by: Anthony Rizzo <36711320+aanthonyrizzo@users.noreply.github.com>

### Description of the change

This PR adds a logical statement to redirect and not render the `<ChartReadme />` when there is a README error. Since I only want to redirect when the file is not found, I check the error as such `.includes("not found")`. I cant see why having this as the default behavior would cause issues but this can easily be controlled w/ a config variable if we wanted.

### Benefits

This change will circumvent the no README error and direct the user to the deployment form

### Possible drawbacks

I don't see any but I could be wrong

### Applicable issues

Fixes #3226

PR #3799, opened 2021-11-22
Skip README when it cannot be found

### LINKED ISSUE #3226: Redirect to DeploymentForm when chart is missing README
Hi guys, me again :)

### Proposal
When there are  helm charts w/out a README users a greeted w/ a blank page that doesnt seem to provide much value (just some 'no README' text and the version controller dropdown). 
image
I was wondering how you guys would feel about just doing a direct link to the deployment form when there is no README for the associated chart.

As of right now I don't see any real drawbacks from making this the default behavior but feedback is greatly appreciated. As per all of my issues/feat requests, Im willing to take ownership of this feature.

antgamdia:
This PR is just to merge the latest upstream changes (after the UI revamp) onto the existing work performed at https://github.com/kubeapps/kubeapps/pull/3227

Edit: the logic has been slightly adapted (see https://github.com/kubeapps/kubeapps/pull/3799#issuecomment-977773790)

Copying below the original PR:

---- 

### Description of the change

This PR adds a logical statement to redirect and not render the `<ChartReadme />` when there is a README error. Since I only want to redirect when the file is not found, I check the error as such `.includes("not found")`. I cant see why having this as the default behavior would cause issues but this can easily be controlled w/ a config variable if we wanted.

### Benefits

This change will circumvent the no README error and direct the user to the deployment form

### Possible drawbacks

I don't see any but I could be wrong

### Applicable issues

Fixes #3226


chart/kubeapps/README.md
```diff
@@ -178,6 +178,7 @@ Once you have installed Kubeapps follow the [Getting Started Guide](https://gith
 | `dashboard.customAppViews`                        | Package names to signal a custom app view                                                  | `[]`                 |
 | `dashboard.customComponents`                      | Custom Form components injected into the BasicDeploymentForm                               | `""`                 |
 | `dashboard.remoteComponentsUrl`                   | Remote URL that can be used to load custom components vs loading from the local filesystem | `""`                 |
+| `dashboard.skipAvailablePackageDetails`           | Skip package details when they are not available                                           | `false`              |
```
absoludity, line 181: Fine as is, just confused me at first because it's not skipping when the details are not available, but rather, skipping when the readme is not available.

dashboard/src/components/PackageHeader/PackageView.test.tsx
```diff
@@ -231,6 +232,39 @@ it("renders the home link when set", () => {
...
+        config: { skipAvailablePackageDetails: false },
+      }),
+      <Router history={history}>
+        <Route path={routePath}>
+          <PackageView />
+        </Route>
+      </Router>,
+    );
+    expect(wrapper.containsMatchingElement(<PackageReadme />)).toBe(true);
+  });
```
absoludity, line 265: Shouldn't we also have a case here something like:
```
it("does not redirect when skipAvailablePackageDetails is true but the package has a readme", () => {...})
```

absoludity (approved):
Great, thanks for the update.

chart/kubeapps/README.md
antgamdia: Yep, it's kinda misleading... perhaps something like `skipPackageDetailsWhenNoReadme` ? WDYT?
I guess the name came from the bulk rename from Chart->AvailablePackage

absoludity: Yep, sounds good.

absoludity: Actually, the reason it stuck out to me was that I was wondering whether the original use-case might be more that you want to skip it regardless? Ie. always skipp package details, rather than it being dependent on whether the readme is populated. Don't remember but might be worth checking.

antgamdia: You're totally right, I missed that context (just thought it was a matter of update the PR to include the latest changes):

> So yes, I'm keen that we don't land #3227 as is (ie. don't depend on the README being present or not), but rather do:
>> ... 
> >Add a config option, skipAvailablePackageDetails which defaults to false to keep the current behaviour, but when true, the tiles presented in the catalog link directly to the deploy form.

Let me implement this way straight away.

antgamdia:
As per the discussion, I've changed the proposed logic to: "if enabled, skip regardless of the readme or any other property, just go to the latest version"

Demo video:

https://user-images.githubusercontent.com/11535726/143227837-498e401d-27ad-4a21-b4d0-2ce6b97c5e70.mp4

chart/kubeapps/README.md
absoludity: Ah - great, we had recorded the discussion on the issue :) Excellent, thanks for the update!

PR #3812, opened 2021-11-24
Remove last remaining connection to kubeops from dashboard.

absoludity:

### Description of the change

Now that the dashboard is using the helm plugin for all helm-related functionality, we have no need for the connection to the kubeops service.

This PR just removes the two small references which are unused afaict.

With this, I believe we can indeed remove the kubeops (and already assetsvc) services? Not sure if we want to do that as part of this release. We'd have to update the internal content team tests I assume. What do you think Antonio?

I've additionally removed the config from the frontend so that no requests are forwarded to the kubeops or assetsvc services to see if anything breaks.

### Benefits

Less moving parts :)

### Possible drawbacks

None here, but if we remove the kubeops/assetsvc from the code and chart, we'll need to update the content teams tests or similar.

### Applicable issues

- Ref #3779

### DIFF
--- a/chart/kubeapps/templates/frontend/configmap.yaml
+++ b/chart/kubeapps/templates/frontend/configmap.yaml
@@ -120,6 +120,49 @@ data:
         proxy_pass {{ include "kubeapps.kubeappsapis.proxy_pass" . -}};
       }
 
+  {{- if .Values.assetsvc.enabled }}
+      # Forward '/api/assetsvc' to '/assetsvc'
+      # but preserving the encoding (eg. '%2F' is not converted to '/')
+      # see https://serverfault.com/a/906479
+      # Ex: from "/api/assetsvc/what$2Fever?param=value"
+      #     it matches as $1="/what$2Fever" and $args="param=value"
+      #     downstream services will receive "/assetsvc/what$2Fever?param=value"
+      location ~* /api/assetsvc {
+        rewrite ^ $request_uri; # pass the encoded url downstream as is,
+        rewrite /api/assetsvc([^?]*) /assetsvc$1?$args break;
+
+    {{- if .Values.frontend.proxypassExtraSetHeader }}
+        proxy_set_header {{ .Values.frontend.proxypassExtraSetHeader }};
+    {{- end }}
+
+    {{- if .Values.frontend.proxypassAccessTokenAsBearer }}
+        # Google Kubernetes Engine requires the access_token as the Bearer when talking to the k8s api server.
+        proxy_set_header Authorization "Bearer $http_x_forwarded_access_token";
+    {{- end }}
+        proxy_pass {{ include "kubeapps.frontend-config.proxy_pass" . -}};
+      }
+  {{- end }}
+
+  {{- if .Values.kubeops.enabled }}
+      location ~* /api/kubeops {
+        # Keep the connection open with the API server even if idle (the default is 60 seconds)
+        # Setting it to 10 minutes which should be enough for our current use case of deploying/upgrading/deleting apps
+        proxy_read_timeout 10m;
+        rewrite /api/kubeops/(.*) /$1 break;
+        rewrite /api/kubeops / break;
+
+    {{- if .Values.frontend.proxypassExtraSetHeader }}
+        proxy_set_header {{ .Values.frontend.proxypassExtraSetHeader }};
+    {{- end }}
+
+    {{- if .Values.frontend.proxypassAccessTokenAsBearer }}
+        # Google Kubernetes Engine requires the access_token as the Bearer when talking to the k8s api server.
+        proxy_set_header Authorization "Bearer $http_x_forwarded_access_token";
+    {{- end }}
+
+        proxy_pass {{ include "kubeapps.frontend-config.proxy_pass" . -}};
+      }
+
       # The route for the Kubeapps backend API is not prefixed.
       location ~* /api/ {
         rewrite /api/(.*) /backend/$1 break;
@@ -136,6 +179,7 @@ data:
 
         proxy_pass {{ include "kubeapps.frontend-config.proxy_pass" . -}};
       }
+  {{- end }}
 
       location / {
         # Add the Authorization header if exists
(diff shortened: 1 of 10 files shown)


chart/kubeapps/templates/frontend/configmap.yaml
```diff
@@ -103,47 +103,6 @@ data:
       }
     {{- end }}
 
-      # Forward '/api/assetsvc' to '/assetsvc'
-      # but preserving the encoding (eg. '%2F' is not converted to '/')
-      # see https://serverfault.com/a/906479
-      # Ex: from "/api/assetsvc/what$2Fever?param=value"
-      #     it matches as $1="/what$2Fever" and $args="param=value"
-      #     downstream services will receive "/assetsvc/what$2Fever?param=value"
-      location ~* /api/assetsvc {
```
antgamdia, line 112: Although, in theory, it is safe to remove this piece of code, I wouldn't do it in a soon-to-be _patch_ release. What if we just render this excerpt if assetsvc.enabled=true ? And it is in the next _minor_ that we effectively remove it?

chart/kubeapps/templates/frontend/configmap.yaml
```diff
@@ -103,47 +103,6 @@ data:
...
-
-    {{- if .Values.frontend.proxypassAccessTokenAsBearer }}
-        # Google Kubernetes Engine requires the access_token as the Bearer when talking to the k8s api server.
-        proxy_set_header Authorization "Bearer $http_x_forwarded_access_token";
-    {{- end }}
-
-        proxy_pass {{ include "kubeapps.frontend-config.proxy_pass" . -}};
-      }
-
-      location ~* /api/kubeops {
```
antgamdia, line 128: Idem here. I'd conditionally render that part based on a kubeops.enabled thing rather than removing it.

chart/kubeapps/templates/frontend/configmap.yaml
```diff
@@ -103,47 +103,6 @@ data:
...
-
-    {{- if .Values.frontend.proxypassAccessTokenAsBearer }}
-        # Google Kubernetes Engine requires the access_token as the Bearer when talking to the k8s api server.
-        proxy_set_header Authorization "Bearer $http_x_forwarded_access_token";
-    {{- end }}
-
-        proxy_pass {{ include "kubeapps.frontend-config.proxy_pass" . -}};
-      }
-
       location ~* /apis {
```
antgamdia, line 106: Well, this is also a reminiscence of our beloved Kubeops service, I guess, but it is mostly aimed at handling `can-i` and other auth stuff mainly, I think?

dashboard/src/shared/App.ts
```diff
@@ -100,16 +98,4 @@ export class App {
       installedPackageRef,
     } as DeleteInstalledPackageRequest);
   }
-
-  // TODO(agamez): remove it once we return the generated resources as part of the InstalledPackageDetail.
```
antgamdia, line 104: Great :tada:

antgamdia (approved):
I have a slight preference for avoiding "breaking" changes in a patch release. Considering a possible scenario where someone is using the kubeops API directly, dropping support for it in the upcoming version is too heavy, IMO. 
I'd just conditionally render the nginx rules based on the `enabled` property of the kubeops deployment (or sth similar). See what you think!

Re: removing it from the UI code, 100% agree

chart/kubeapps/templates/frontend/configmap.yaml
absoludity: Yeah true. I'd forgotten that people could potentially be using the current kubeops outside of our dashboard usage. +1 for just putting it behind a flag like you've done for assetsvc.

chart/kubeapps/templates/frontend/configmap.yaml
absoludity: Actually, I think it's the one further down (`/api/` not `/apis/`, the latter being kubeapps-apis)? But yes, you're right - that one has a proxy_pass of:
```
        proxy_pass {{ include "kubeapps.frontend-config.proxy_pass" . -}};
```
which I missed because I was grepping for kubeops in the config. But this helper does indeed just proxy to the kubeops service (not sure why it's labeled frontend-config).

absoludity:
> I have a slight preference for avoiding "breaking" changes in a patch release. Considering a possible scenario where someone is using the kubeops API directly, dropping support for it in the upcoming version is too heavy, IMO. I'd just conditionally render the nginx rules based on the `enabled` property of the kubeops deployment (or sth similar). See what you think!

PTAL, I've done this instead (as well as updated the assetsvc frontend config stanza to be conditional on its similar enabled flag). Also put all the kubeops resources within a similar condition (matching assetsvc). Only difference is that kubeops.enabled of course currently defaults to true.

Thanks Antonio.

chart/kubeapps/values.yaml
```diff
@@ -874,6 +874,10 @@ apprepository:
 ## Kubeops parameters
 ##
 kubeops:
+  ## @param kubeops.enabled Specifies whether this component should be installed.
+  ## In a future release we will be setting this to default to false.
+  ##
+  enabled: true
```
antgamdia, line 880: Oops, new param... let me re-run the readme generator to also add it here

antgamdia:
> PTAL, I've done this instead (as well as updated the assetsvc frontend config stanza to be conditional on its similar enabled flag). Also put all the kubeops resources within a similar condition (matching assetsvc). Only difference is that kubeops.enabled of course currently defaults to true.

Awesome, thank you for taking it into account. I mean, I really want to get rid of the old kubeops code (and some logic it calls under `pkg/...`), but we have to be good citizens :P 
That said... next release... :fire: 

I've added a commit after executing the readminator .

mecampbellsoup:
Are there docs anywhere on migrating from old versions where we relied on kubeops, to newer versions that use kubeappasapis?

absoludity:
The migration at the time was transparent to the user... we'd gradually moved API calls from the Kubeapps dashboard client across to the new apis service during a few releases until this PR where we removed the last API call.

Unless you're calling APIs on the kubeops service yourself explicitly from your own client or something, there should be no migration needed? If you are doing so, let us know your usage, happy to point you in the right direction to use the new API?

PR #3813, opened 2021-11-24
Add eslint warning if console.log are left

antgamdia:
### Description of the change

Minor PR adding just an eslint warning if the committed code includes unnoticed `console.log`. I thought it was enabled by default, but some previous updates in the eslint config must have disabled it.

### Benefits

I won't commit again (or at least, our CI won't let it pass) code with console.log eventually useful for debugging a unit test :P

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

I'd say it was marked (in the past) as an `error`, which prevented eventual console statements when developing, Now it is a warning, we should be able to run `yarn start` without any issues.

### DIFF
--- a/dashboard/.eslintrc.json
+++ b/dashboard/.eslintrc.json
@@ -37,6 +37,7 @@
     "jest": true
   },
   "rules": {
+    "no-console": "warn", // warns if a console.log is left in the code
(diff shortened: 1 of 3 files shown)


absoludity (approved):
Great, thanks.

Issue #3604, opened 2021-10-18
Timeout Helm deployment > 300 seconds failed


misteio:
### Description:
I want to install an Helm Chart that takes more than 300 seconds to deploy.

### Steps to reproduce the issue:

1. Deploy a personal Helm Chart through KubeApps.
2. Wait around 300 seconds
3. It failed

### Describe the results you received:

![Error](https://i.ibb.co/W5B5Y42/screen-kubeapps.png )

### Describe the results you expected:
It could be great if we can change this default behavior.

### Additional information you deem important (e.g. issue happens only occasionally):

I use a custom Helm Chart I've made. When  I add the following annotations, it won't wait until the end of deployment.
```yaml
annotations:
    "helm.sh/hook": post-install,post-upgrade
    "helm.sh/hook-weight": "-5"
    "helm.sh/hook-delete-policy": hook-succeeded,before-hook-creation
```
### Version of Helm, Kubeapps and Kubernetes:

- Output of `helm version`:
```
version.BuildInfo{Version:"v3.6.3", GitCommit:"d506314abfb5d21419df8c7e7e68012379db2354", GitTreeState:"clean", GoVersion:"go1.16.5"}
```

- Output of `helm list -n <kubeapps-namespace> <kubeapps-release-name>`:

```
kubeapps        shared          2               2021-10-18 18:08:32.619996037 +0000 UTC deployed        kubeapps-7.5.8          2.4.1  
```

- Output of `kubectl version`:

```
Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.0", GitCommit:"c2b5237ccd9c0f1d600d3072634ca66cefdf272f", GitTreeState:"clean", BuildDate:"2021-08-04T18:03:20Z", GoVersion:"go1.16.6", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"20+", GitVersion:"v1.20.7-eks-d88609", GitCommit:"d886092805d5cc3a47ed5cf0c43de38ce442dfcb", GitTreeState:"clean", BuildDate:"2021-07-31T00:29:12Z", GoVersion:"go1.15.12", Compiler:"gc", Platform:"linux/amd64"}
```

castelblanque:
I'm trying to come up with an implementation for this :)

- Handling of the input flag in the Cobra command is clear
- On the other end, usage of the value for e.g. Helm's `Install` command is clear as well

[...]

PR #3822, opened 2021-11-25
Added timeout parameter for Helm operations (#3604)

castelblanque:

### Description of the change

Helm has a default timeout for operations set to 300 seconds.
When installing a Helm chart, if there is any pre/post installation hook taking more than 300 seconds, the process will fail.

This PR allows to set a timeout for Helm operations from the Kubeapps chart values.

### Benefits

Helm charts installation can now take more than 300 seconds overall :)

### Possible drawbacks

Allowing to increase timeout for Helm operations might lead to scenarios where other timeouts kick in before, e.g. Helm timeout = 200s and Nginx timeout set to 60s ?

### Applicable issues

- Ref #3604

### Additional information


castelblanque:
Writing the tests is a bit tricky, as the timeout is handled by Helm's kube client, and usually in tests the kube client is entirely mocked using `FailingKubeClient` and no logic for timeouts exists there.
Any ideas on how to test it? Some work [done here](https://github.com/kubeapps/kubeapps/blob/3604-helm-timeout/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/create_test.go#L202).

absoludity:
Thanks Rafa. Great work so far. Sorry for the essay here:

> Writing the tests is a bit tricky, as the timeout is handled by Helm's kube client, and usually in tests the kube client is entirely mocked using `FailingKubeClient` and no logic for timeouts exists there. Any ideas on how to test it? Some work [done here](https://github.com/kubeapps/kubeapps/blob/3604-helm-timeout/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/create_test.go#L202).

I think you already mentioned, but the [fake Helm kube client completely ignores the timeout](https://github.com/helm/helm/blob/98c4ccf29b6e13568349fe57f5bfd3607ac286e8/pkg/kube/fake/printer.go#L77). You've created your own fake which can actually use the timeout when set, but it looks like you're doing an integration test which is actually going to wait for seconds. I think this road is tricky because we'd really need a proper integration test with a chart that has a hook which will run for a configurable time so that we can cause it to timeout?

But in terms of a unit test here, I think (personally, others may disagree) that we'd only want to test to the border of our contract with helm. With this in mind, you've tested the config parsing already (great), we just want to ensure that a server configured with the timeout results in a helm client that is passed that same timeout, I think? (The helm project is responsible for the contract of what happens with that timeout).

But easier said than done. I was then a bit stumped as to how we could do that, because the agent just returns the release, we have no way to test the internal state of the helm command there. I went down the rabbit-hole that you may have already traveled of seeing if we could have a hook defined for the test chart so we could trigger the preinstall hook and use that to record the timeout used, but it's quite complicated to do so.

The only solution I can see to verify this is to split it into two separate unit tests, one for the plugin's CreateRelease, and one for the agent code, to verify that the value is passed through to helm.

To first test the plugin, we could update the server struct so that it has a field for the CreateRelease function which is initialised in `NewServer` to the `agent.CreateRelease` function, but in tests we explicitly set it to a dummy function that just records the timeout used. Then test here that calling the server (configured with a timeout) `CreateInstalledPackage` function results in a call to the `CreateRelease` function with the expected arguments (including the timeout).

Then secondly to unit-test the code in the `agent` module, extract the lines which create the `cmd` object to a private function of the module that returns the `cmd`, and unit-test that to show that the `cmd.Timeout` is set with the value from the argument. Not perfect.

All in all, it's not a great situation testing-wise that you've found here (and the solution outlined isn't that great either, just testing the arguments passed). So I'm OK with your branch just having the parsing test that you've done, or if you want to give the above a go, that's fine too, but totally up to you.

castelblanque:
The way to approach testing is always an open discussion in projects 😄 
When possible, I like to create integration tests to be sure that the whole flow works as expected end to end.
For this case, probably it is not worth the effort to create complete integration tests.

>The only solution I can see to verify this is to split it into two separate unit tests, one for the plugin's CreateRelease, and one for the agent code, to verify that the value is passed through to helm.

I totally agree with your approach for testing here, makes sense.
Let's stick to the unit tests with extracted Helm command creation and CreateRelease function variable.

Many thanks Michael!

absoludity (approved):
> When possible, I like to create integration tests to be sure that the whole flow works as expected end to end.

Yeah, that'd be ideal... I just can't see how to do that with the provided testing fakes from Helm (that ignore the timeout). The only option would be to add an actual e2e test specifically for this, but that becomes very heavy-weight.

Thanks for the updates!

PR #3824, opened 2021-11-25
Update deps and submodule for apprepo-ctrl

antgamdia:
### Description of the change

Given that we finally weren't able to start the release process today, I'm squeezing one more PR for updating our deps :innocent: 

Also, I noticed something undocumented in our docs: upgrading the git submodule for the apprepository generator. I haven't added anything in the docs yet, mainly because I guess ideally we will adapting the generated code somehow, but haven't checked it. I just wanted to update the submodule's go.mod file to avoid conflicts with this upgrade.

### Benefits

Up to date deps

### Possible drawbacks

I don't foresee any, but in case the CI or even ourselves detect something wrong, we can postpone this PR as much as needed.

### Applicable issues

N/A

### Additional information

N/A


chart/kubeapps/values.yaml
```diff
@@ -1606,6 +1606,7 @@ kubeappsapis:
   ##
   enabledPlugins:
     - helm
+    - kapp-controller
```
absoludity: Did you intend to enable the kapp-controller plugin by default? I wouldn't have thought to do this, especially in this first release, but even in others when users need to install kapp controller etc. It may not be an issue for the kapp-controller plugin, as it may not try to query any carvel resources until a request comes in, but the flux plugin will correctly cause the kubeappsapis server to fail to start if it doesn't find the required resources.

absoludity (approved):

chart/kubeapps/values.yaml
absoludity: Hah, I think you saw this yourself before I submitted and removed. Sorry for the noise.

PR #3825, opened 2021-11-25
Use the same assert library to avoid extra deps

antgamdia:
### Description of the change

I assumed the dependabot error (below) was somethig transient, but it is not. It is due to an old dep causing some problems when auto-upgrading deps. Therefore, this PR is to replace it with the one we have been extensibily using (the other one is from the Monocular times).

image

### Benefits

Dependabot will be able, hopefully, to keep our go deps up to date... and I will be sleep happy knowing everything is smoothly being upgraded :P

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

N/A

### DIFF
--- a/cmd/asset-syncer/server/utils_test.go
+++ b/cmd/asset-syncer/server/utils_test.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"compress/gzip"
 	"encoding/base64"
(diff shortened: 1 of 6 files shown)


absoludity (approved):
Great housekeeping :)

PR #3832, opened 2021-11-26
Revert change in crobjob API. Add kubeops.enabled in CI

antgamdia:
### Description of the change

This PR aims at fixing some issues found when preparing the release, namely: 1) enabling kubeops via --set flag (only needed now we added this new param and it is not available in the current-1 chart version yet), 2) reverting the change in the cronjob API.
 
### Benefits

Kubeapps will continue to [work in k8s 1.20](https://app.circleci.com/pipelines/github/kubeapps/kubeapps/5106/workflows/78c1abb8-7a40-442f-9927-6c1c1c74b224/jobs/65517)

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

The issues found arise two questions to me:

1) Should we start using the same approach for selecting the API group version as we did in the past (you know, the fix Michael did when removing the unsafe lookup function) to spin up the cronjobs?

2) Should we clearly state in the project readme the current supported k8s versions?  And, what's more, perhaps also adding the tested stack: like kapp-ctrl 0.30, OLM 1.17, Helm 3.x.x-3.7.1, pinniped 10,etc ?

If you all agree, happy to create a couple of issues to track them

### DIFF
--- a/cmd/apprepository-controller/server/controller.go
+++ b/cmd/apprepository-controller/server/controller.go
@@ -32,6 +32,7 @@ import (
 	"github.com/kubeapps/kubeapps/pkg/kube"
 	log "github.com/sirupsen/logrus"
 	batchv1 "k8s.io/api/batch/v1"
+	batchv1beta1 "k8s.io/api/batch/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


absoludity (approved):
> The issues found arise two questions to me:
> 
>     1. Should we start using the same approach for selecting the API group version as we did in the past (you know, the fix Michael did when removing the unsafe lookup function) to spin up the cronjobs?

I don't remember what we did when removing the unsafe lookup function, but thought here we'd want the golang equivalent of the [common.capabilities.cronJob.apiVersion chart helper](https://github.com/bitnami/charts/blob/master/bitnami/common/templates/_capabilities.tpl#L40-L49).

> 
>     2. Should we clearly state in the project readme the current supported k8s versions?  And, what's more, perhaps also adding the tested stack: like kapp-ctrl 0.30, OLM 1.17, Helm 3.x.x-3.7.1, pinniped 10,etc ?
> 
> 
> If you all agree, happy to create a couple of issues to track them

+1 to both (I'm assuming GKE will continue to support 1.20 for a while yet). My fail for not checking when `batch/v1` was supported from.

PR #3838, opened 2021-11-29
Watch specific resources rather than list of all gvrs.

absoludity:

### Description of the change

While testing the Dashboard using the new resources plugin to watch k8s resources, I found a few issues in the plugin itself.

The main one being that the watch request was not limited to a particular resource but rather all gvr's in the namespace. Only noticed because I had two deployments in the default namespace with different replicas and was seeing the wrong number of pods in the UI.

Other small changes which I'll comment on inline.

### Benefits

Dashboard UX won't get confused when displaying k8s resources for an installed package (when I update to use the resources plugin in a following PR).

### Applicable issues

- Ref #3779

### DIFF
--- a/cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/release.go
+++ b/cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/release.go
@@ -678,7 +678,8 @@ func newFluxHelmRelease(chart *models.Chart, targetName types.NamespacedName, ve
 }
 
 // resourceRefsFromManifest returns the resource refs for a given yaml manifest.
-func resourceRefsFromManifest(m string) ([]*corev1.ResourceRef, error) {
+// TODO(minelson): share common functionality between plugins.
+func resourceRefsFromManifest(m, pkgNamespace string) ([]*corev1.ResourceRef, error) {
 	decoder := yaml.NewYAMLToJSONDecoder(strings.NewReader(m))
 	refs := []*corev1.ResourceRef{}
 	doc := yamlResource{}
@@ -695,20 +696,33 @@ func resourceRefsFromManifest(m string) ([]*corev1.ResourceRef, error) {
 		}
 		if doc.Kind == "List" || doc.Kind == "RoleList" || doc.Kind == "ClusterRoleList" {
 			for _, i := range doc.Items {
+				namespace := i.Metadata.Namespace
+				if namespace == "" {
+					namespace = pkgNamespace
+				}
 				refs = append(refs, &corev1.ResourceRef{
 					ApiVersion: i.APIVersion,
 					Kind:       i.Kind,
 					Name:       i.Metadata.Name,
-					Namespace:  i.Metadata.Namespace,
+					Namespace:  namespace,
 				})
 			}
 			continue
+		}
+		// Helm does not require that the rendered manifest specifies the
+		// resource namespace so some charts do not do so (ldap).  We explicitly
+		// set the namespace for the resource ref so that it can be used as part
+		// of the key for the resource ref.  At the moment we do not distinguish
+		// between cluster-scoped and namespace-scoped resources for the refs.
+		namespace := doc.Metadata.Namespace
+		if namespace == "" {
+			namespace = pkgNamespace
 		}
 		refs = append(refs, &corev1.ResourceRef{
 			ApiVersion: doc.APIVersion,
 			Kind:       doc.Kind,
 			Name:       doc.Metadata.Name,
-			Namespace:  doc.Metadata.Namespace,
+			Namespace:  namespace,
 		})
 	}
 
--- a/cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/release.go
+++ b/cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/release.go
@@ -712,8 +712,13 @@ func resourceRefsFromManifest(m, pkgNamespace string) ([]*corev1.ResourceRef, er
 		// Helm does not require that the rendered manifest specifies the
 		// resource namespace so some charts do not do so (ldap).  We explicitly
 		// set the namespace for the resource ref so that it can be used as part
-		// of the key for the resource ref.  At the moment we do not distinguish
-		// between cluster-scoped and namespace-scoped resources for the refs.
+		// of the key for the resource ref.
+		// TODO(minelson): At the moment we do not distinguish between
+		// cluster-scoped and namespace-scoped resources for the refs.  This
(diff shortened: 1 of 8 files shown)


cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/release.go
```diff
@@ -695,20 +696,33 @@ func resourceRefsFromManifest(m string) ([]*corev1.ResourceRef, error) {
...
+					Namespace:  namespace,
 				})
 			}
 			continue
 		}
+		// Helm does not require that the rendered manifest specifies the
+		// resource namespace so some charts do not do so (ldap).  We explicitly
+		// set the namespace for the resource ref so that it can be used as part
+		// of the key for the resource ref.  At the moment we do not distinguish
+		// between cluster-scoped and namespace-scoped resources for the refs.
```
absoludity, line 716: This came up while using the new plugin from the dashboard: the LDAP chart just happens to not specify the namespace metadata in the rendered templates of the helm manifest, which meant that all the resource refs were missing this info, which caused problems on two levels:
* The resource refs are meant to additionally check that a specific resource belongs to a package... if it does not include the namespace, it could be used to read a similarly named resource (even a secret) in a different namespace.
* In the dashboard code, we need to store resources with a unique key in the redux state, which of course needs to include the resource. This was what found the issue for me since the corresponding code which was pulling data out of the redux state wouldn't find it (since the original key did not have the namespace).

cmd/kubeapps-apis/plugins/resources/v1alpha1/server.go
```diff
@@ -58,10 +59,11 @@ type Server struct {
...
 
-	// kindToResource is a function to convert a GVK to GVR. Can be replaced
-	// in tests with a dummy version using the unsafe helpers while the real
-	// implementation queries the k8s API for a REST mapper.
-	kindToResource func(meta.RESTMapper, schema.GroupVersionKind) (schema.GroupVersionResource, error)
+	// kindToResource is a function to convert a GVK to GVR with
+	// namespace/cluster scope information. Can be replaced in tests with a
+	// dummy version using the unsafe helpers while the real implementation
+	// queries the k8s API for a REST mapper.
+	kindToResource func(meta.RESTMapper, schema.GroupVersionKind) (schema.GroupVersionResource, meta.RESTScopeName, error)
```
absoludity, line 66: This update was required to be able to get/watch resources correctly. The current code only works for namespaced resources as, without this extra info, it assumes a namespace scope.

cmd/kubeapps-apis/plugins/resources/v1alpha1/server.go
```diff
@@ -216,8 +220,17 @@ func (s *Server) GetResources(r *v1alpha1.GetResourcesRequest, stream v1alpha1.R
 			continue
 		}
 
-		watcher, err := dynamicClient.Resource(gvr).Namespace(namespace).Watch(stream.Context(), metav1.ListOptions{})
+		listOptions := metav1.ListOptions{
+			FieldSelector: fmt.Sprintf("metadata.name=%s", ref.GetName()),
```
absoludity, line 224: This is the only way I can see to watch a particular resource using the client-go's Watch.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/release.go
antgamdia: So, just to confirm I'm getting it right: even if packages can generate cluster-scoped resources, we are not supporting it?  That is, are we assuming a "no namespace" = "namespace where the pkg in installed" ? 
If so, how are we retrieving those cluster-scoped resources? Perhaps it's just me missing something

cmd/kubeapps-apis/plugins/resources/v1alpha1/server.go
```diff
@@ -192,14 +194,16 @@ func (s *Server) GetResources(r *v1alpha1.GetResourcesRequest, stream v1alpha1.R
...
+		// the scope of the resource (namespaced or not).
+		gvr, scopeName, err := s.kindToResource(s.restMapper, gvk)
 		if err != nil {
 			return status.Errorf(codes.Internal, "unable to map group-kind %v to resource: %s", gvk.GroupKind(), err.Error())
 		}
 
 		if !r.GetWatch() {
 			var resource interface{}
-			if ref.Namespace != "" {
+			if scopeName == meta.RESTScopeNameNamespace {
```
antgamdia, line 206: Question: I guess `scopeName` refers to the namespace of the resource, but what 
 does `meta.RESTScopeNameNamespace` stand for?

antgamdia (approved):
Good catch! Thank you for the changes; I just have some questions to ensure I'm getting it properly.
I guess something similar should have to be implemented for the Carvel plugin as well, so thanks also for shedding some light here :P

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/release.go
absoludity: Cluster-scoped resources are fine, in that when fetching the referenced resource in the resources plugin, it checks whether the resource is namespaced via the k8s rest mapper and does the right thing (ie. it does not rely on the namespace to determine the scope)... rather than just relying on `ref.Namespace` (this is a change below).

The issue is that, as we're using the helm manifest to get the references, it is decided by the chart author, so you get (for example) both `ClusterRole`s with a `metadata.namespace` set (which, when applied, just results in the cluster-scoped resource) and `Role`s without a namespace (as Helm automatically applies the resources in the context of the release namespace). I'm just standardising here so the generated ResourceRef(erence) will always have the ResourceRef.Namespace set, regardless of the scope of the resource so that (a) we can define a unique key/identifier for a resource ref, and (b) the reference cannot be used to get/watch resources in other namespaces. 

Ideally here we'd only be setting the namespace only when it is not set *and* the resource is a namespace-scoped resource, which is what I'm referring to in the last sentence of the above comment, but that means providing both plugins with a RESTMapper, which I didn't do here. I'll update that to an explicit TODO.

cmd/kubeapps-apis/plugins/resources/v1alpha1/server.go
absoludity: `scopeName` is either `meta.RESTScopeNameNamespace` or `meta.RESTScopeNameRoot`, both of which are [defined in the meta package](https://pkg.go.dev/k8s.io/apimachinery@v0.22.4/pkg/api/meta#RESTScopeName) as string constants corresponding to "namespace" and "root" respectively.

PR #3862, opened 2021-11-30
Add actions for getting resources from resources plugin.

absoludity:
### Description of the change

This PR adds the new required actions (and handling of those actions in the reducer) for interacting with the new resources plugin to get and watch resources.

This allows a single request to the resources plugin to watch many resources. The streaming rpc call uses [server-side events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events) rather than web sockets, with the grpc-web client using rxjs observables which makes the implementation quite tidy.

### Benefits

This PR doesn't change the AppView yet, so the only benefit is preparation for the next PR where the AppView will be updated to use the new action.

### Applicable issues

- Ref #3779

### DIFF
--- a/dashboard/src/actions/kube.tsx
+++ b/dashboard/src/actions/kube.tsx
@@ -58,11 +58,16 @@ export const requestResources = createAction("REQUEST_RESOURCES", resolve => {
     watch: boolean,
     handler: (r: GetResourcesResponse) => void,
     onError: (e: Event) => void,
-  ) => resolve({ pkg, refs, watch, handler, onError });
+    onComplete: (pkg: InstalledPackageReference) => void,
+  ) => resolve({ pkg, refs, watch, handler, onError, onComplete });
 });
 
 export const receiveResourcesError = createAction("RECEIVE_RESOURCES_ERROR", resolve => {
   return (err: Error) => resolve(err);
+});
+
+export const closeRequestResources = createAction("CLOSE_REQUEST_RESOURCES", resolve => {
+  return (pkg: InstalledPackageReference) => resolve(pkg);
 });
 
 export const addTimer = createAction("ADD_TIMER", resolve => {
@@ -81,6 +86,7 @@ const allActions = [
   closeWatchResource,
   requestResources,
   receiveResourcesError,
+  closeRequestResources,
   receiveResourceFromList,
   requestResourceKinds,
   receiveResourceKinds,
@@ -188,6 +194,9 @@ export function getResources(
         },
         (e: any) => {
           dispatch(receiveResourcesError(e));
+        },
+        (pkg: InstalledPackageReference) => {
+          dispatch(closeRequestResources(pkg));
         },
       ),
     );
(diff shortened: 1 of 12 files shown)


dashboard/src/actions/kube.tsx
```diff
@@ -44,6 +49,27 @@ export const closeWatchResource = createAction("CLOSE_WATCH_RESOURCE", resolve =
   return (ref: ResourceRef) => resolve(ref);
 });
 
+// requestResources takes a ResourceRef[] and subscribes to an observable to
+// process the responses as they arrive.
+export const requestResources = createAction("REQUEST_RESOURCES", resolve => {
```
antgamdia, line 54: So, it is in this _request_ event that we also perform the actual subscription. Other actions (which are not meant to return any msg-subscription thing) split the _request_ and _receive_ actions, but here, given that the concept of requesting resources implies requesting a subscription, maybe makes sense to leave it this way.

dashboard/src/actions/kube.tsx
```diff
@@ -146,3 +175,45 @@ export function getAndWatchResource(
...
+
+export function processGetResourcesResponse(
+  r: GetResourcesResponse,
+  dispatch: ThunkDispatch<IStoreState, null, KubeAction>,
+) {
+  const key = keyForResourceRef(
+    r.resourceRef!.apiVersion,
+    r.resourceRef!.kind,
+    r.resourceRef!.namespace,
+    r.resourceRef!.name,
```
antgamdia, line 214: Are we 100% sure that `r` does have a `resourceRef` ? If not, we can just dispatch an error instead of forcing the null check? No strong opinion, though, happy as it is.

dashboard/src/shared/ResourceRef.ts
```diff
@@ -21,6 +21,14 @@ export function fromCRD(
   return ref;
 }
 
+// TODO: (minelson) Update to use API resourceRef type once old model removed.
```
antgamdia, line 24: Nitpick: I'd use the same comment format to ease the search term when we want to find for them.

dashboard/src/shared/types.ts
```diff
@@ -1,5 +1,9 @@
 import { JSONSchemaType } from "ajv";
 import { RouterState } from "connected-react-router";
+// TODO(minelson): Why is lint complaining:
+// error  Unable to resolve path to module 'rxjs'  import/no-unresolved
```
antgamdia, line 4: I guess is something missing in the eslintrc.json (perhaps in `import/resolver` ?) not sure, I'd have to check

antgamdia (approved):
Thank you!! This is way simpler than the manual webhook+fallback polling mechanism! Really happy to move the complexity to the existing protocols and backend instead of a heavy frontend!

dashboard/src/shared/types.ts
absoludity: So finally found the right incantation here, from https://stackoverflow.com/a/55280867 . It looks like it was because `rxjs` is itself written in typescript directly, but node was (?) only looking for js files in `node_modules`.

PR #3863, opened 2021-11-30
Allow to sync and store charts with spaces in name (#3758)

### LINKED ISSUE #3758: Problems with charts with spaces in their name
### Description:

Packages including spaces as part of their names cannot be displayed in Kubeapps (well, the pkg details view at least).

1. Add a package with spaces in its name. For example, using Helm and Harbor:
  1. Go to https://demo.goharbor.io/harbor/projects an create a new project, e.g., `kubeapps`
  2. Create a new Helm chart, e.g., running ` helm create demo`, then edit the `Chart.yaml` file to sth like `name: foo bar`
  3. Package the chart with `helm package demo` 
  4. Upload the `foo bar.tgz` to the Harbor instance 
  5. In Kubeapps, add a new repository: `https://demo.goharbor.io/chartrepo/kubeapps`
2. Verify the package is being listed (pic below)
  1. image
3. Click on the pkg to access its details and here's the error (next pic)

image

Details should be displayed, as usual.

It is probably due to a missing encoding of the spaces (`%20` instead of ` `), as we are actually doing with the `/` (`%2F`).
I did face an issue with encoding slashes in the past, so do not hesitate to ping me for further details.

 N/A (current devel version)

castelblanque:

### Description of the change

This PR allows to sync and store in DB charts that contain spaces in name. End user will then be allowed to reach to the package details page, and when deploying, the error thrown by Helm will be shown.
Applied only to Helm repos, not to OCI.

### Benefits

Kubeapps is transparent with regards to charts containing spaces in name. Those charts are synced and stored properly, and end user can navigate to package details now. However Helm installation will still fail due to its restrictions. [See here](https://github.com/kubeapps/kubeapps/issues/3758#issuecomment-981748752)

### Possible drawbacks

Packages that can not be installed are listed.
This fix applies only to spaces in name: should we do the same exercise for other special characters? Helm only allows alphanumeric, dash, dot and underscore.

### Applicable issues

- fixes #3758

### DIFF
--- a/cmd/asset-syncer/server/utils_test.go
+++ b/cmd/asset-syncer/server/utils_test.go
@@ -1296,6 +1296,101 @@ func Test_filterCharts(t *testing.T) {
 	}
 }
 
+func TestUnescapeCharsData(t *testing.T) {
+	tests := []struct {
+		description string
+		input       []models.Chart
+		expected    []models.Chart
+	}{
+		{
+			"chart with encoded spaces in id",
+			[]models.Chart{
+				{ID: "foo%20bar"},
+			},
+			[]models.Chart{
+				{ID: "foo bar"},
+			},
+		},
+		{
+			"chart with encoded spaces in name",
+			[]models.Chart{
+				{Name: "foo%20bar"},
+			},
+			[]models.Chart{
+				{Name: "foo bar"},
+			},
+		},
+		{
+			"chart with mixed encoding in name",
+			[]models.Chart{
+				{Name: "test/foo%20bar"},
+			},
+			[]models.Chart{
+				{Name: "test/foo bar"},
+			},
+		},
+		{
+			"chart with no encoding nor spaces",
+			[]models.Chart{
+				{Name: "test/foobar"},
+			},
+			[]models.Chart{
+				{Name: "test/foobar"},
+			},
+		},
+		{
+			"chart with unencoded spaces",
+			[]models.Chart{
+				{Name: "test/foo bar"},
+			},
+			[]models.Chart{
+				{Name: "test/foo bar"},
+			},
+		},
+	}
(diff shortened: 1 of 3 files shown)


antgamdia (approved):
LGTM! Thanks for this fix!

However, I wonder what would happen if the chart also contains encoded `/` (`%2F`) as part of the name.
Have you also checked if a chart whose id is, for instance, `my-repo/bitnami/cool wordpress`   is also stored properly? I guess so, because (from memory I recall) we were handling the `%2F` when retrieving data from the database, but not sure atm.

Have a look also at: https://github.com/kubeapps/kubeapps/blob/master/cmd/assetsvc/pkg/utils/postgresql_utils.go

btw, I don't know why Go doesn't have a "safe" version of `url.PathUnescape(value)`, it is always so verbose...

castelblanque:
That scenario is not tested, but I didn't want to mix things up, due to the issue referencing spaces only.
That is why I raise the question in "Possible drawbacks" section above: the number of special characters that may appear in a chart name is unknown (e.g. `{`, `&`, etc.). We only know what Helm allows.
Probably this need further discussion.

cmd/asset-syncer/server/utils_test.go
```diff
@@ -1296,6 +1296,101 @@ func Test_filterCharts(t *testing.T) {
 	}
 }
 
+func TestUnescapeCharsData(t *testing.T) {
```
absoludity, line 1299: Just a typo above.

absoludity (approved):
Excellent :)

cmd/asset-syncer/server/utils_test.go
castelblanque: Thanks for pointing it out!

PR #3872, opened 2021-12-01
Switch AppView to use new getResources.

absoludity:

### Description of the change

Switches the AppView to use the new `getResources` action (using the resources plugin), with minimal changes to other code.

Checking the application detail for the Kubeapps app now only shows a single request for api server's `getResources` rather than 16 or so watch requests directly to the k8s api server:

kubeapps-watch-resources

And creating a new installation watches for updates so that the UI gets updated when pods are ready:

apache-install

### Benefits

* Single request for watching all resources.
* No longer hitting the k8s API server to do so
* Can now remove *lots* of code and simplify other code.

### Possible drawbacks

See what CI says.

### Applicable issues

- ref #3779

### Additional information

There will be a few PRs to follow removing lots of code, extracting the current `ResourceRef` model and all its dependencies, removing the request for all resource kinds from the server, and other indicated improvements like not watching all resources etc.

BTW: This is the third PR in a series, I've not yet landed the first until I hear that we're good to do so (ie., when the current release tag is final). Let me know :)

### DIFF
--- a/chart/kubeapps/Chart.yaml
+++ b/chart/kubeapps/Chart.yaml
@@ -33,4 +33,4 @@ maintainers:
 name: kubeapps
 sources:
   - https://github.com/kubeapps/kubeapps
-version: 7.6.1-dev1
+version: 7.6.2-dev1
--- a/chart/kubeapps/Chart.yaml
+++ b/chart/kubeapps/Chart.yaml
@@ -33,4 +33,4 @@ maintainers:
 name: kubeapps
 sources:
   - https://github.com/kubeapps/kubeapps
-version: 7.6.2-dev0
+version: 7.6.1-dev1
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -1605,6 +1605,7 @@ kubeappsapis:
   ##
   enabledPlugins:
     - helm
+    - resources
   pluginConfig:
     core:
       packages:
--- a/dashboard/src/components/AppView/AppView.test.tsx
+++ b/dashboard/src/components/AppView/AppView.test.tsx
@@ -1,4 +1,5 @@
 import actions from "actions";
+import { getType } from "typesafe-actions";
 import Alert from "components/js/Alert";
 import LoadingWrapper from "components/LoadingWrapper/LoadingWrapper";
 import PageHeader from "components/PageHeader";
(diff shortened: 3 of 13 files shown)


absoludity:
Hmm, CI disagrees with my local experience - with [app details displayed without the resources](https://app.circleci.com/pipelines/github/kubeapps/kubeapps/5175/workflows/0c50c03c-db9c-4ce5-9108-20adb7f4994e/jobs/66174/artifacts). I'll check in the morning.

dashboard/src/components/AppView/AppView.test.tsx
```diff
@@ -1,4 +1,5 @@
 import actions from "actions";
+import { getType } from "typesafe-actions";
```
absoludity, line 2: I've refactored a little in this file because the existing tests aren't testing that actions are dispatched by the view, but rather that the actions themselves are called... this meant that my test passed even though I'd missed `dispatch`ing the action. So I've left the existing tests as they are but updated the ones for the functionality that I'm changing so that they actually assert that the correct action is dispatched. The other issue I had from what I'd included in the diff yesterday was that I hadn't enabled the resources plugin :)

absoludity:
This PR is the next to merge into master, when reviewed - so if someone has time (I've added you antonio, but no problem if it waits for @castelblanque ... in fact I'll add you both ;) )

dashboard/src/components/AppView/AppView.test.tsx
antgamdia: Thanks for the info and refactoring the test. I've just hit the same issue... the error message wasn't clear at all so, almost running out of ideas I decided to check if the resources plugin was enabled... :facepalm:

chart/kubeapps/Chart.yaml
```diff
@@ -33,4 +33,4 @@ maintainers:
 name: kubeapps
 sources:
   - https://github.com/kubeapps/kubeapps
-version: 7.6.2-dev0
+version: 7.6.1-dev1
```
antgamdia, line 36: Not sure if decreasing the version is intentional?

chart/kubeapps/values.yaml
```diff
@@ -1605,6 +1605,7 @@ kubeappsapis:
   ##
   enabledPlugins:
     - helm
+    - resources
```
antgamdia, line 1608: :grin:

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -204,20 +204,22 @@ export default function AppView() {
...
-    Object.values(resourceRefs).forEach((refs: ResourceRef[]) => {
-      refs.forEach(ref => dispatch(actions.kube.getAndWatchResource(ref)));
-    });
+    if (!app || !app.installedPackageRef || !app.apiResourceRefs) {
+      return () => {};
+    }
+    // TODO(minelson): Update to watch only those resources that
+    // we're interested in watching change (deployments, statefulsets etc.)
+    // and get the rest.
+    dispatch(actions.kube.getResources(app!.installedPackageRef!, app!.apiResourceRefs, true));
```
antgamdia, line 218: I guess typescript transpiler is not smart enough to know `app` is != undefined at this point

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -204,20 +204,22 @@ export default function AppView() {
...
     return function cleanup() {
```
antgamdia, line 219: Nice! btw, we are not usually using cleanup functions. Perhaps at some point, we should revisit it to ensure we don't have any resource left (I recall to have seen some unmount warnings at the console)

antgamdia (approved):
Awesome! Thanks! 
Sorry for the delay in taking a look at it... with the release I totally forgot about it.

chart/kubeapps/Chart.yaml
absoludity: Nope, must have missed the patch when resolving the conflict. Updated.

dashboard/src/components/AppView/AppView.tsx
absoludity: Not sure of the sequence of events, but it appears to be happy without them now.

dashboard/src/components/AppView/AppView.tsx
absoludity: Yeah, it's required here, because unmounting this component triggers the action to finish the request (and remove the subscription from the state, which reminds me, I want to remove it from the state as it's no longer necessary to track there. I'll do that in my current branch).

PR #3878, opened 2021-12-01
Update integration image

antgamdia:
### Description of the change

As part of the release process, we ought to update the integration image with the latest changes (just bumping up deps), so this PR is to perform this change.
I've pushed upstream so that the CI gets the changes.

### Benefits

Perhaps new versions bring more CI stability? 

### Possible drawbacks

Perhaps new versions bring more CI **in**stability? 

### Applicable issues

N/A

### Additional information

N/A

### DIFF
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -1,5 +1,6 @@
 {
   "name": "dashboard",
+  "proxy": "https://localhost",
   "version": "0.1.0",
   "private": true,
   "homepage": "./",
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -1,6 +1,5 @@
 {
   "name": "dashboard",
-  "proxy": "https://localhost",
   "version": "0.1.0",
   "private": true,
   "homepage": "./",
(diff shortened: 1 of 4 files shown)


dashboard/package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "dashboard",
+  "proxy": "https://localhost",
```
castelblanque, line 3: Just for my understanding, what is the purpose of adding the proxy here? I don't mean [technically](https://create-react-app.dev/docs/proxying-api-requests-in-development/), but more the reasoning behind. Thanks! 😄

antgamdia: Hahaha, it is just me clicking the wrong "commit all" button instead of "commit staged" one :facepalm: . Thanks for pointing this out.

castelblanque: Great!
If it helps, in order to avoid having to modify the `package.json` for this purpose, I add in my local a `src/setupProxy.js` file like this
```
const createProxyMiddleware = require('http-proxy-middleware');

module.exports = function(app) {
  app.use(
    ['/apis', '/api'],
    createProxyMiddleware({
      target: 'http://localhost',
      changeOrigin: true,
    })
  );
};
```
It is also a file that could be committed by mistake, but maybe we could add it to `.gitignore` 😅

castelblanque (approved):

PR #3879, opened 2021-12-01
Fix css overflow issues

antgamdia:
### Description of the change

Since I'm working on the deployment/upgrade views, I couldn't resist fixing something that really annoys me: a wrong overflow when reducing the size of the window:sweat_smile: 

I also removed some <ul> bullets, but happy to revert back if you like them.

Before:

image

image

After:

image

image

### Benefits

Better UI?

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

N/A


absoludity (approved):
Looks much nicer at the smaller width, thanks!

PR #3880, opened 2021-12-01
Use cds-select in the deployment/upgrade views

antgamdia:
### Description of the change

As part of the slow and continuous (?) process to migrate our UI to use the new Clarity CDS core components, I've moved the pkg version selects to the new ones. 

Before:

image

image

After:

image

image

### Benefits

We will be able to add more components (like the ServiceAccount selector) without mixing `clr` and `cds` components.

Also, the `cds` component looks much better, it has a responsive layout already built-in, so less thing to worry about.

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

I don't know if there is an actual reason for keeping two package version selectors... what do you think about removing the one in the header (just for the upgrade view, I mean). If you agree, I'll do it in a separate PR (perhaps it has some e2e tests implications and I want to land this one soon for the service account selector thing)

### DIFF
--- a/dashboard/src/components/DeploymentForm/DeploymentForm.tsx
+++ b/dashboard/src/components/DeploymentForm/DeploymentForm.tsx
@@ -1,3 +1,5 @@
+import { CdsFormGroup } from "@cds/react/forms";
+import { CdsInput } from "@cds/react/input";
 import actions from "actions";
 import AvailablePackageDetailExcerpt from "components/Catalog/AvailablePackageDetailExcerpt";
 import Alert from "components/js/Alert";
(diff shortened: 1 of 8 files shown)


absoludity (approved):

absoludity:
> I don't know if there is an actual reason for keeping two package version selectors... what do you think about removing the one in the header (just for the upgrade view, I mean).

I'm not 100%, but guess that the redundancy is there because the detail view used to display the package version in the header (rather than the left panel), so when people clicked on upgrade, they may expect to edit it there also. As it is, that's no longer the case.

> If you agree, I'll do it in a separate PR (perhaps it has some e2e tests implications and I want to land this one soon for the service account selector thing)

Yep, sounds good.

castelblanque:
Looks better organized, cool!

PR #3883, opened 2021-12-01
Allow selecting which SA should be used to install a Carvel pkg

### LINKED ISSUE #3850: Add a mechanism to select a ServiceAccount with which the Carvel bundle will get installed
Parent issue: https://github.com/kubeapps/kubeapps/issues/3849

Some alternatives:

1.  Kubeapps creating SAs (therefore needing a privileged user to install carvel stuff)
    1. Before installing, autogenerate a SA with the proper RBA, or
    2. Letting users select which SA and RBAC they want to grant
2.  Kubeapps not creating SA:
    1. Before installing, a user would select a SA from a list of available SA within a given namespace. Throw any RBAC errors back, or
    2. Letting users write the name of the SA (not listing perms required). Throw any RBAC errors back.

Personally, I don't think Kubeapps should be that privilege piece creating SAs and RBACs... so I'd rather discard option 1). 
Instead, I'd go with 2.1 and 2.2, I mean:

In the deployment form, we could query if the user can list SAs; if so, I'd render a dropdown with all the available SAs in the current namespace. Otherwise, if the listing fails, I'd fall back to rendering a text input so that the user can write which SA they want.

image

image

antgamdia:
### Description of the change

This PR (~still a draft~, ~tests remaining~), aims at adding a new service account selector field when deploying a package (iff it is a carvel package, atm).
It requires:
1. Adding a new backend endpoint for retrieving the SA; not the `corev1.ServiceAccount`  object, but just their names (as we don't need the SA secret names at all).
    - ~I implemented it in Kubeops (as we did for the can-i endpoint); notwithstanding, this code ought to be moved to a separate plugin at some point.~
    - I implemented it in the new Resources plugin :)
2. Consuming this new API endpoint from the backend
3. Creating a selector with its logic to i) fetch the whole list; ii) pass the selected SA to the createInstalledPkg op.

The list of changes is not that big, but I'm willing to split each of those ones out to separate PRs.

### Benefits

Installing Carvel packages will work as the Carvel folks intended, I guess :P

image

image

### Possible drawbacks

N/A

### Applicable issues

- fixes #3850

### Additional information

~I'll work on the tests tomorrow~
~This PR depends on https://github.com/kubeapps/kubeapps/pull/3894 (yeah, I could have stacked it, but it would imply recreating the branch upstream and I don't foresee much work on the other, so once merged, it will get simplified (ie, not so many changes as it is atm).~


absoludity:
> This PR (still a draft, tests remaining),

k, won't look at the code yet :)

> aims at adding a new service account selector field when deploying a package (iff it is a carvel package, atm).

I think we want this for both carvel and flux plugins, no?

> It requires:
> 
>     1. Adding a new backend endpoint for retrieving the SA; not the `corev1.ServiceAccount`  object, but just their names (as we don't need the SA secret names at all).
>        
>        * I implemented it in Kubeops (as we did for the can-i endpoint); notwithstanding, this code ought to be moved to a separate plugin at some point.

Yeah, I'd be keen not to add more code to Kubeops right now. This is exactly the thing for the resources plugin, imo, since it's not specific to any packaging format. `GetServiceAccountNames` passing the context, would fit? (And it'd still be doing so only if the users RBAC allows).

> 
> The list of changes is not that big, but I'm willing to split each of those ones out to separate PRs.

at +246, -83 (the current modified/deleted lines) it's absolutely no issue. If it grows significantly then it may be worth splitting to help reviews, just see what you think.

antgamdia:
> Yeah, I'd be keen not to add more code to Kubeops right now. This is exactly the thing for the resources plugin, imo, since it's not specific to any packaging format. GetServiceAccountNames passing the context, would fit? (And it'd still be doing so only if the users RBAC allows).

Great; I wasn't sure if the purpose of these resources plugins was solely giving those resources generated by an installation or if it was aimed at covering a wide range of scenarios (like this one or even the can-i).
So, I'll do a prequel PR adding the new stuff for this operation (proto+autogenerated code)

dashboard/src/shared/utils.ts
```diff
@@ -125,3 +125,11 @@ export function getPluginPackageName(plugin?: Plugin | string) {
...
   }
 }
+
+export function getPluginsRequiringSA(): string[] {
+  return [PluginNames.PACKAGES_FLUX, PluginNames.PACKAGES_KAPP];
+}
+
+export function getPluginsSupportingRollback(): string[] {
+  return [PluginNames.PACKAGES_HELM];
+}
```
antgamdia, line 135: Extracting the logic of which plugins are for certain capabilities, to DRY-up the code.

dashboard/src/components/DeploymentForm/DeploymentForm.tsx
```diff
@@ -52,11 +57,20 @@ export default function DeploymentForm() {
...
+  const [reconciliationOptions, setReconciliationOptions] = useState({} as ReconciliationOptions);
 
   const error = apps.error || selectedPackage.error;
 
   const [pluginObj] = useState({ name: pluginName, version: pluginVersion } as Plugin);
 
+  const onChangeSA = (e: React.FormEvent<HTMLSelectElement>) => {
+    setReconciliationOptions({
+      ...reconciliationOptions,
+      serviceAccountName: e.currentTarget.value,
```
antgamdia, line 70: In the future, we may want to allow modifying other parameters, like the reconciliation interval, that's why I'm storing this obj instead of just the SA

dashboard/src/components/DeploymentForm/DeploymentForm.tsx
```diff
@@ -76,6 +90,17 @@ export default function DeploymentForm() {
     return () => {};
   }, [dispatch, packageReference, packageVersion]);
 
+  useEffect(() => {
+    // Populate the service account list if the plugin requires it
+    if (getPluginsRequiringSA().includes(pluginObj.name)) {
+      // We assume the user has enough permissions to do that. Fallback to a simple input maybe?
+      Kube.getServiceAccountNames(targetCluster, targetNamespace).then(saList =>
+        setServiceAccountList(saList.serviceaccountNames),
```
antgamdia, line 98: I'm storing the SA list in the component's state solely; as it is not shared by other components.

dashboard/src/components/DeploymentForm/DeploymentForm.tsx
```diff
@@ -191,7 +217,37 @@ export default function DeploymentForm() {
...
+                          <option key=""></option>
+                          {serviceAccountList?.map(o => (
+                            <option key={o} value={o}>
+                              {o}
+                            </option>
+                          ))}
+                        </select>
+                        <CdsControlMessage error="valueMissing">
+                          The Service Account name this application will be installed with.
+                        </CdsControlMessage>
```
antgamdia, line 244: Using CDS components brings some form validation capabilities out-of-the-box (idem with inputs).

dashboard/src/components/DeploymentForm/DeploymentForm.tsx
absoludity: Yep, sound good.

dashboard/src/components/DeploymentForm/DeploymentForm.tsx
absoludity: Yes, and we need to fetch it whenever the target namespace changes (as you've indicated below with the dependencies), so yep, component state sounds perfect.

dashboard/src/shared/utils.ts
absoludity: Excellent - much better to have it all in the one place for when we want to replace it with the plugin-provided items themselves.

pkg/http-handler/http-handler.go
```diff
@@ -377,6 +377,7 @@ func SetupDefaultRoutes(r *mux.Router, namespaceHeaderName, namespaceHeaderPatte
 	if err != nil {
 		return err
 	}
+	//TODO(agamez): move these endpoints to a separate plugin when possible
```
absoludity, line 380: Yes, though hopefully we won't just move them. For example, we don't want to provide generic endpoints that can return all app repositories. Let's see.

absoludity (approved):
LGTM. I ignored the generated code and backend changes (which I assume are because you didn't get to rebase on master and remove the commits from #3894).

castelblanque (approved):

PR #3887, opened 2021-12-02
Remove most references to shared/ResourceRef

absoludity:
### Description of the change

When the dashboard was handling the fetching of resources directly from the k8s api server, it needed to also keep state about whether resources were namespaced, url endpoints for the resources and a bunch of other book-keeping. This is no longer necessary for the dashboard.

This PR begins the work of extracting the ResourceRef model, by removing it from the AppView and corresponding components/containers, and doing a drop-in replacement with the simpler generated ResourceRef from our protobuf message.

### Benefits

Less code to maintain, simpler dashboard.

### Applicable issues

- Ref #3779


dashboard/src/containers/AccessURLTableContainer/AccessURLTableContainer.tsx
```diff
@@ -22,10 +19,4 @@ function mapStateToProps({ kube }: IStoreState, props: IAccessURLTableContainerP
   };
 }
```
antgamdia, line 20: Not necessarily in this PR, but we should also get rid of the xxxxContainer.tsx in favor of just using react hooks, especially, now we have removed more logic

antgamdia (approved):
Excellent!! It's great we are removing so much logic from the UI. Now it looks much better! Thanks for the clean-up!

PR #3888, opened 2021-12-02
Extract more unused actions from actions/Kube

absoludity:
### Description of the change

Just removes more actions and related tests that are no longer used.

I had hoped to then delete the old `ResourceRef` class, but alas, it's used by the operator support (directly in the dashboard) to interact with the k8s API server. So it needs to remain for now.

I will remove further actions here, but ran out of road before dinner.

### Benefits

Less code, simpler dashboard

### Applicable issues

- ref #3779


antgamdia (approved):
+1 when CI also agrees :P
Yep, it's a pity... supporting operators as is right now is disappointing. When changing the catalog view, I had to keep several types, functions, etc to maintain both the "package" and the "operator".

PR #3891, opened 2021-12-02
Fix install repo firing twice (#3678)

### LINKED ISSUE #3678: UI: app repository Install Repo button fires twice (sometimes)
### Description:

I ran several times into the use case where clicking the "Install Repo" button to add a new app repository is being executed twice, resulting in an error message being displayed.
The first execution seems to properly create the app repository, and does not seem to be affected by the second execution.

Any submit button should be guarded from firing multiple times.

castelblanque:

### Description of the change

Fixes the problem of "Install Repo" button being fired multiple times.

### Benefits

Repo installation process is not triggered more than once.

### Possible drawbacks

N/A

### Applicable issues

- fixes #3678

### Additional information

Makes use of [React's useRef](https://reactjs.org/docs/hooks-reference.html#useref)

### DIFF
--- a/dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
+++ b/dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
@@ -7,7 +7,7 @@ import { CdsTextarea } from "@cds/react/textarea";
 import actions from "actions";
 import Alert from "components/js/Alert";
 import * as yaml from "js-yaml";
-import { useEffect, useState } from "react";
+import { useEffect, useRef, useState } from "react";
 import { useDispatch, useSelector } from "react-redux";
 import { Action } from "redux";
 import { ThunkDispatch } from "redux-thunk";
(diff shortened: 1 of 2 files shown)


dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
```diff
@@ -49,6 +49,7 @@ const TYPE_OCI = "oci";
 
 export function AppRepoForm(props: IAppRepoFormProps) {
   const { onSubmit, onAfterInstall, namespace, kubeappsNamespace, repo, secret } = props;
+  const workingRef = useRef(false);
```
antgamdia, line 52: I'd maybe choose another name, a more self-explanatory one like  `isInstallingRef` ?

antgamdia (approved):
Thanks for fixing that issue! Great work :)

dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
```diff
@@ -620,7 +627,7 @@ export function AppRepoForm(props: IAppRepoFormProps) {
         </Alert>
       )}
       <div className="margin-t-xl">
-        <CdsButton disabled={validating} onClick={install}>
+        <CdsButton type="submit" disabled={validating}>
```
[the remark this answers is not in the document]
castelblanque: U nailed it! `install` function was being called twice: here and inside form's onSubmit logic.

dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
castelblanque: You are right, will rename it. Initially I wanted to make the solution something general for all forms+buttons :)

PR #3894, opened 2021-12-02
Add proto and implementations for GetServiceAccountNames

antgamdia:
### Description of the change

This is a prequel PR for https://github.com/kubeapps/kubeapps/pull/3883, mainly intended for generating all the resources, proto, implementation and API docs for the GetServiceAccountNames.
This operation just returns a list of service account names for a given authenticated user.

### Benefits

We will be able to use this operation for retrieving the list of SAs when installing a package.

### Possible drawbacks

N/A

### Applicable issues

- rel https://github.com/kubeapps/kubeapps/issues/3850

### Additional information

I've also updated our API docs; as it is still a manual process... it forgot to do it in the previous release :S  

Example:

image


cmd/kubeapps-apis/plugins/resources/v1alpha1/server.go
```diff
@@ -254,6 +255,34 @@ func (s *Server) GetResources(r *v1alpha1.GetResourcesRequest, stream v1alpha1.R
 	return nil
 }
 
+// GetResources returns the resources for an installed package.
```
absoludity, line 258: Copy-n-paste-o...

cmd/kubeapps-apis/plugins/resources/v1alpha1/server.go
```diff
@@ -254,6 +255,34 @@ func (s *Server) GetResources(r *v1alpha1.GetResourcesRequest, stream v1alpha1.R
...
+	namespace := r.GetContext().GetNamespace()
+	cluster := r.GetContext().GetCluster()
+	log.Infof("+resources GetServiceAccountNames (cluster: %q, namespace=%q)", cluster, namespace)
+
+	typedClient, _, err := s.clientGetter(ctx, cluster)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "unable to get the k8s client: '%v'", err)
+	}
+
+	saList, err := typedClient.CoreV1().ServiceAccounts(namespace).List(context.TODO(), metav1.ListOptions{})
```
absoludity, line 269: We should be using the `ctx` that was passed in above, rather than creating a new context here?

absoludity (approved):

PR #3897, opened 2021-12-03
Remove watchResource and associated code.

absoludity:
### Description of the change

Following on from the other #3779 PRs, this PR removes more now-unused code (after switching to the new resources plugin endpoint) and simplifies the `keyForResourceRef` helper now that it is used solely with our protobuf-generated resourceref.

### Applicable issues

- ref #3779

### DIFF
--- a/dashboard/src/containers/helpers/index.ts
+++ b/dashboard/src/containers/helpers/index.ts
@@ -7,7 +7,5 @@ import { IKubeState } from "shared/types";
 // found are filtered out and the response will only contain resources that are
 // available in the state.
 export function filterByResourceRefs(refs: ResourceRef[], resources: IKubeState["items"]) {
-  return refs
-    .map(r => resources[keyForResourceRef(r.apiVersion, r.kind, r.namespace, r.name)])
-    .filter(r => r !== undefined);
+  return refs.map(r => resources[keyForResourceRef(r)]).filter(r => r !== undefined);
 }
--- a/dashboard/src/reducers/kube.ts
+++ b/dashboard/src/reducers/kube.ts
@@ -122,7 +122,9 @@ export const initialState: IKubeState = {
   items: {},
   kinds: initialKinds,
   // We book keep on subscriptions, keyed by the installed package ref,
-  // so that we can
+  // so that we can unsubscribe when the closeRequestResources action is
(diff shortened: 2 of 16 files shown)


dashboard/src/containers/helpers/index.ts
```diff
@@ -7,7 +7,5 @@ import { IKubeState } from "shared/types";
 // found are filtered out and the response will only contain resources that are
 // available in the state.
 export function filterByResourceRefs(refs: ResourceRef[], resources: IKubeState["items"]) {
-  return refs
-    .map(r => resources[keyForResourceRef(r.apiVersion, r.kind, r.namespace, r.name)])
-    .filter(r => r !== undefined);
+  return refs.map(r => resources[keyForResourceRef(r)]).filter(r => r !== undefined);
```
antgamdia, line 10: I don't know about our preference of using `lodash` vs plain js, but I guess you can also do sth like `_.omitBy`. Anyway, it is clear enough and easy to understand, so that's totally fine with me.

dashboard/src/reducers/kube.ts
```diff
@@ -121,9 +121,9 @@ export const initialKinds = {
 export const initialState: IKubeState = {
   items: {},
   kinds: initialKinds,
+  // We book keep on subscriptions, keyed by the installed package ref,
+  // so that we can
```
antgamdia, line 125: Maybe this comment is stripped out? I think I don't get the sentence

antgamdia (approved):
Great! Happy to be removing plenty of complex code in the UI (and tests....). Thanks for the clean-up!

dashboard/src/containers/helpers/index.ts
absoludity: Yeah, this was the existing code that needed to handle undefined resources in the state. I'm not sure how the state could previously get into this situation, but I'm pretty confident that it cannot do so now anyway (as we only add resources to the state in the first place if they are sent by the resources plugin), but lots of test data currently depends on it. I may remove it in another PR and update the tests.

dashboard/src/reducers/kube.ts
absoludity: Sorry - probably got distracted half way through writing it :P. Fixed.

PR #3902, opened 2021-12-06
Only watch resources that are updated in UX.

### LINKED ISSUE #3779: Update dashboard to use new k8s resources endpoint of api server

absoludity:

### Description of the change

Rather than watching all resources for an installed package (as Kubeapps has done for quite some time), this PR updates so that we only watch the resources which are updated in the UX and just fetch the rest once only when the AppView is mounted.

### Benefits

Viewing Kubeapps, for example, now results in only 13 watches on the k8s API server (from the plugin), rather than ~40.

### Applicable issues

- fixes #3779

### DIFF
--- a/dashboard/src/actions/kube.tsx
+++ b/dashboard/src/actions/kube.tsx
@@ -99,6 +99,10 @@ export function getResources(
           dispatch(receiveResourcesError(e));
         },
         () => {
+          // The onComplete handler should only dispatch a closeRequestResources
+          // action if this call to `getResources` is for watching. If it is not
+          // watching resources, the server will close the request automatically
+          // (and we have no book-keeping in the redux state).
           if (watch) {
             dispatch(closeRequestResources(pkg));
           }
--- a/dashboard/src/actions/kube.tsx
+++ b/dashboard/src/actions/kube.tsx
@@ -39,7 +39,7 @@ export const requestResources = createAction("REQUEST_RESOURCES", resolve => {
     watch: boolean,
     handler: (r: GetResourcesResponse) => void,
     onError: (e: Event) => void,
-    onComplete: (pkg: InstalledPackageReference) => void,
+    onComplete: () => void,
   ) => resolve({ pkg, refs, watch, handler, onError, onComplete });
 });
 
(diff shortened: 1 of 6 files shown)


dashboard/src/actions/kube.tsx
```diff
@@ -39,7 +39,7 @@ export const requestResources = createAction("REQUEST_RESOURCES", resolve => {
     watch: boolean,
     handler: (r: GetResourcesResponse) => void,
     onError: (e: Event) => void,
-    onComplete: (pkg: InstalledPackageReference) => void,
+    onComplete: () => void,
```
absoludity, line 42: There was no need for the arg since it can be included via closure.

dashboard/src/components/AppView/AppView.tsx
```diff
@@ -166,33 +165,40 @@ export default function AppView() {
...
-    if (Object.values(appViewResourceRefs).some(ref => ref.length)) {
+    if (!app?.apiResourceRefs) {
       return () => {};
     }
 
     const parsedRefs = parseResources(app.apiResourceRefs);
     setAppViewResourceRefs(parsedRefs);
     return () => {};
-  }, [app, cluster, kinds, appViewResourceRefs]);
+  }, [app?.apiResourceRefs]);
```
absoludity, line 175: Just cleaned up the dependencies (ie. we don't want to fire this effect whenever the app changes, only when the resource refs change).

castelblanque:
Awesome! From 40 to 13 watches :)
Thank you!

castelblanque (approved):

PR #3906, opened 2021-12-06
Fix the typescript checks to work with current ts-proto gen.

absoludity:

### Description of the change

The current version of `ts-proto` which we use generates typescript that fails our typescript lint checks. This means that any new PR which needs to generate, fails to pass CI.

I've created an issue upstream to see if it's something that may be fixed upstream, in https://github.com/stephenh/ts-proto/issues/432

For now I've just excluded `src/gen` from our tsconfig so that it's not checked. Longer term we should remove this and also generate these files during our build process so that this is caught immediately when the ts-proto package was upgraded.

### Benefits

CI can pass on any new PRs that modify the protobuf.

### Applicable issues

- fixes #


absoludity:
Of course I can't exclude the generated content from the build... I want to exclude it only from the type check. I'll try a different approach.

Issue #3908, opened 2021-12-07
Generate protobuf content during build


absoludity:
### Description:

While we were developing the kubeapps-apis service, we didn't update the bulid process to generate the protobuf content, but should do so now that we're using and depending on this service.

I had the issue yesterday that running `buf generate` on master leads to code that won't pass our ts-compile checks. I've documented that separately upstream at https://github.com/stephenh/ts-proto/issues/432 , but had we been generating during our build, we would not have updated the `ts-proto` module without noticing.

We could now do something like git ignoring the generated content in our local checkouts and generating it in the build, or instead failing the build if the output is different. Either way, we'll want to ensure we use the same version of buf locally as in CI etc. Anyway, needs more discussion, just highlighting for inclusion soon.

castelblanque:
My experience in other projects is that generated source code is never added to Git. This is, removed from repo and added to gitignore.
It will be generated in build time, either locally or in CI.
If developers can not modify it at all, why having it in the repo? If any change needs to be done to the API, the definition file is changed.

absoludity:
> My experience from other projects is that generated source code is never added to Git. This is, removed from repo and added to gitignore.

Yeah, that's the normal approach with generated source code and it is something we can do in Kubeapps as above. GRPC-generated code in go is sometimes (but not always) treated differently because the normal tooling is to `go get` a go module and expect to be able to use it in your code without having to run any third-party tooling. But I don't think we need to cater for this case anyway, so +1 to remove it from the repo and add locally to git ignore as outlined above. The APIs service is no longer just a proof-of-concept.

antgamdia:
I do agree. It's important to ensure we are using the same version in every environment. 

For the record, currently, when running `buf generate` we are generating: 1) openapi v2 docs; 2) go backend code; 3) ts frontend code
Besides, we manually perform some ulterior manual (and potentially undocumented) tasks:
1) convert the openapi v2 json file to an openapi v3 in yaml, copy the `paths` and `schema` sections to the `openapi.yaml` file under `dashboard/public` folder
2) _no action required_
3) go to the dashboard directory and run `yarn run prettier` (or any equivalent way)

During the generation process, we depend on 1) the `buf` binary already installed in our systems (which embed `protoc`) and is out of the scope of our VCS system. and 2) the `buf mod` deps, which are tracked in git.

So, if we plan to untrack the autogenerated code, I foresee two potential issues:

[...]

Issue #3896, opened 2021-12-03
Replace remaining Dashboard calls to K8s API server


absoludity:
### Description:

With #3779 nearing completion, the main use-case for the Kubeapps dashboard to interact with the K8s API server (fetch and watch resources related to an install) has been replaced with the new `resources` plugin.

There is still the Operator support which requires talking directly to the K8s API server (from `shared/Operators`), which will be handled in the short-term with a flag to enable/disable it (so that Kubeapps can run without proxying to the K8s API server) as documented in #3252 .

The other remaining call-sites to the K8s API are:

- [x] `shared/Namespace`, both `Namespaces.create` and `Namespace.get` use the k8s API directly and need to switch to use the resources plugin.
- [x] `shared/Secret`, used to get and list secrets as well as create pull secrets when interacting with Helm AppRepositories.
- [x] `shared/Auth`, used to validate a token credential when not using OIDC, as well as to determine if the app is authenticated with an http-only cookie (by simply doing a request to check the response, as by design there's no way to determine from javascript if there's an http-only cookie present). We should be able to replace these with calls to the resources plugin, since it uses the same user credential.
- [x] `shared/AppRepository` has a `.getSelfLink` which uses the k8s API and is used in `actions/repos`. Replace with list or a backend get. Edit: Actually, this code path (for SelectForm) is no longer used afaict, since we don't enable the Upgrade button if we don't have an available/installed pkg.
- [x] `integration/02-private-repository` uses the k8s API to verify that a secret is created and used with a deployment.

These are the only three remaining call-sites found by grepping for `url.api.k8s` (it's possible this grep misses files that may import specific parts of the `url` module. EDIT: verified with `rg "import .* from.*url\"" dashboard/src` that the only place this happens is in the operator modules).

With those three call-sites taken care of, we should then be able to finally:
- [x] Update the frontend config so that we are no longer proxying to the K8s API server, unless the operator support has been enabled (for now).

PR #3909, opened 2021-12-07
Upgrade ts-proto with fix for noImplicitReturns.

absoludity:

### Description of the change

Replaces #3906 . The current version of `ts-proto` which we use generates typescript that fails our typescript lint checks. This means that any new PR which needs to generate, fails to pass CI.

cc @antgamdia I'm not 100%, but it looks like you may have manually updated the generated content in [this commit](https://github.com/kubeapps/kubeapps/pull/3894/commits/dcf07cc2b4b2ea4c91267da647d21a9b0d5cd273) so that it would pass? I thought of doing the same but we need to be able to work with the generated content.

I've created an issue upstream to see if it's something that may be fixed upstream, in https://github.com/stephenh/ts-proto/issues/432

For now I've just disabled the `noImplicitReturns` check. Longer term we should re-enable this and also generate these files during our build process so that this is caught immediately when the ts-proto package was upgraded (see #3908)

### Benefits

We can continue to modify protobuf files without breaking the build.

### Applicable issues

- ref #3896


boukeversteegh:
Hey @absoludity , thanks for opening an issue in ts-proto, we will fix the issue there.

In the meantime, I thought the following might be helpful. When generated code doesn't match your project's linting rules, you could also disable linting on the generated code, or exclude a violated rule for a particular folder or file. That way you won't be blocked and don't need to disable the rule for your whole project. It's also handy for cases where the rules of the two projects contradict each other, and it cannot be fixed upstream.

For example, you could do this:

```js
// .eslintrc.js
module.exports = {
  // exclude entire directory from linting
  ignorePatterns: ['src/lib/**'],
  overrides: [
    // disable specific rule in directory
    {
      files: ['somedirectory/**/*.ts'],
      rules: {
        '@typescript-eslint/no-undef': 'off',
      },
    },
  ],
}
```

I'm not an expert on eslint, but this configuration worked for me.

boukeversteegh:
PR ready https://github.com/stephenh/ts-proto/pull/436

absoludity:
Excellent, thanks @boukeversteegh ... now I can just update ts-proto instead :) And thanks for the info about the eslintrc..., I did briefly look for a way to supress it for a particular file, but missed the overrides.

absoludity:
Rubber-stamping this one to unblock changes to proto files.

castelblanque:
@absoludity After this change (containing the upgrade of ts-proto): could we get rid of the generated TS source code for Dashboard?

absoludity:
> @absoludity After this change (containing the upgrade of ts-proto): could we get rid of the generated TS source code for Dashboard?

It's unrelated to this change/upgrade. We may want to do so but that'll be part of the discussion at #3908 .

antgamdia:
Sorry, I've been OOO for a while.

> cc antgamdia I'm not 100%, but it looks like you may have manually updated the generated content in this commit so that it would pass? I thought of doing the same but we need to be able to work with the generated content.

Yep, you're totally right. [I added](https://github.com/kubeapps/kubeapps/pull/3894/commits/dcf07cc2b4b2ea4c91267da647d21a9b0d5cd273#diff-3794f941f1e38cef30a3b20d95a1d241818d2dc31375455fb103b78cfcc7406dR375-R376) a `return undefined` to pass the linter issue (albeit being consistent with the method's signature) after the ts-proto dep upgrade while developing the service account thing. I forgot to look into the issue in-depth and I totally miss that change when committing and sending the PR. My bad. I'm sorry :(

Thanks for catching the issue, Michael, and special thanks to the ts-proto team; what a quick solution!!

PR #3911, opened 2021-12-07
Switch dashboard to use checkNamespaceExists.

absoludity:

### Description of the change

Updates the dashboard to use the resources plugin to check if a namespace exists (rather than the more generic getNamespace which was hitting the k8s API server directly from the dashboard).

Also updates the `Namespace.create` shared helper to use the resources client.

### Benefits

See #3896 

### Applicable issues

- ref #3896

### DIFF
--- a/dashboard/src/shared/Namespace.ts
+++ b/dashboard/src/shared/Namespace.ts
@@ -3,8 +3,11 @@ import { Auth } from "./Auth";
 import { axiosWithAuth } from "./AxiosInstance";
 import { ForbiddenError, IResource, NotFoundError } from "./types";
 import * as url from "./url";
+import { KubeappsGrpcClient } from "./KubeappsGrpcClient";
 
 export default class Namespace {
+  private static resourcesClient = () => new KubeappsGrpcClient().getResourcesServiceClientImpl();
+
   public static async list(cluster: string) {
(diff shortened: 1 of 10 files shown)


dashboard/src/shared/Namespace.ts
```diff
@@ -3,35 +3,40 @@ import { Auth } from "./Auth";
...
   public static async create(cluster: string, namespace: string) {
-    const { data } = await axiosWithAuth.post<IResource>(url.api.k8s.namespaces(cluster), {
-      apiVersion: "v1",
-      kind: "Namespace",
-      metadata: {
-        name: namespace,
+    await this.resourcesClient().CreateNamespace({
+      context: {
+        cluster,
+        namespace,
```
castelblanque, line 25: It looks cleaner than before :)

castelblanque:
One more step to abstract away to direct K8s API calls. Thanks!

castelblanque (approved):

absoludity:
> One more step to abstract away from direct K8s API calls. Thanks!

Thanks for the approval. Just note that I can't land this as the previous PR isn't yet approved. If it's easier, you can see the whole series of PRs for this work on #3896

castelblanque (approved):

PR #3916, opened 2021-12-08
Switch dashboard's createDockerRegistrySecret to use resources API.

absoludity:

### Description of the change

Switches the `Secret.createDockerRegistrySecret` helper from using the k8s API directly to using the resources plugin.

Tested IRL to ensure both the dashboard change and the backend grpc function work together as expected (which they do).

### Benefits

* No longer receive the secret data over the network when creating a secret,
* One less call-site using the k8s API

### Applicable issues

- ref #3896 

### Additional information

Next up is switching the dashboard's `Secret.list` to use the plugin's `GetSecretNames` which will be trickier, as it also involves updating the dashboard to no longer require the secret data for the UX interaction.

### DIFF
--- a/dashboard/src/actions/repos.ts
+++ b/dashboard/src/actions/repos.ts
@@ -73,7 +73,7 @@ export const receiveImagePullSecrets = createAction("RECEIVE_IMAGE_PULL_SECRETS"
 });
 
 export const createImagePullSecret = createAction("CREATE_IMAGE_PULL_SECRET", resolve => {
-  return (secret: ISecret) => resolve(secret);
+  return (secretName: string) => resolve(secretName);
 });
 
 const allActions = [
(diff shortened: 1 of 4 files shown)


castelblanque (approved):

dashboard/src/actions/repos.ts
```diff
@@ -433,16 +433,8 @@ export function createDockerRegistrySecret(
...
-        currentCluster,
-        name,
-        user,
-        password,
-        email,
-        server,
-        namespace,
-      );
-      dispatch(createImagePullSecret(secret));
+      await Secret.createPullSecret(currentCluster, name, user, password, email, server, namespace);
```
antgamdia, line 436: Not sure if we can/want use the `Context` here instead of passing the cluster/ns directly

antgamdia (approved):
Great, thanks!

antgamdia:
We probably want to keep this PR open.

dashboard/src/actions/repos.ts
absoludity: Yep, we probably can now, but will need to update the form props and call-sites etc.

PR #3780, opened 2021-11-18
add gRPC UnaryServerInterceptor for kubeapps-apis 

satya-dillikar:
### Description of the change

Use an interceptor to automatically log every RPC call to the kubeapps-apis server.

### Benefits

intercepting all RPC calls in one location.
### Possible drawbacks

The kubectl logs for pod/kubeapps-internal-kubeappsapis will be flooded with more events.

### Applicable issues

This tries to address  #3385

### Additional information

N/A

PR #3917, opened 2021-12-08
suppress noisy logs

satya-dillikar:
### Description of the change

Kubeapps-api is using "httpGet probe" for livenessProbe & readinessProbe every 10secs.
values.yaml
```
  livenessProbe:
    enabled: true
    httpGet:
      path: /core/plugins/v1alpha1/configured-plugins
      port: 50051
    initialDelaySeconds: 60
    periodSeconds: 10

  readinessProbe:
    enabled: true
    httpGet:
      path: /core/plugins/v1alpha1/configured-plugins
      port: 50051
    initialDelaySeconds: 0
    periodSeconds: 10
```

This is causing noisy logs of probe endpoint
`/kubeappsapis.core.plugins.v1alpha1.PluginsService/GetConfiguredPlugins`

as mentioned in PR#[3780](https://github.com/kubeapps/kubeapps/pull/3780)

 
Solution:

Option-1:
We can probably use "GRPC Health Checking Protocol" as mentioned in  [health-checking-grpc-servers-on-kubernetes](https://kubernetes.io/blog/2018/10/01/health-checking-grpc-servers-on-kubernetes/)

Adding “grpc-health-probe” needs to implement standard health check "protocol" as in [health-checking](https://grpc.github.io/grpc/core/md_doc_health-checking.html)  and bundle health check utility "tool" with the kubeapps-api binary.

But this still does not solve the noisy logging of RPC calls.

Option-2:
Suppress the endpoints in the grpc interceptor call.

The commit implements 
1) Suppress the noisy endpoints
2) uses consistent logging format (same as klog)
### Benefits

no noisy endpoints logs 

### Possible drawbacks

N/A

### Applicable issues

- fixes #

### DIFF
--- a/cmd/kubeapps-apis/server/server.go
+++ b/cmd/kubeapps-apis/server/server.go
@@ -22,6 +22,7 @@ import (
 	"net"
 	"net/http"
 	"reflect"
+	"strings"
 	"time"
 
 	"github.com/improbable-eng/grpc-web/go/grpcweb"
@@ -65,13 +66,48 @@ func CreateRequestLogger() func(ctx context.Context, req interface{}, info *grpc
 	}
 }
 
+func getLogLevelOfEndpoint(endpoint string) klogv2.Level {
+
+	// Add all endpoint function names which you want to suppress in interceptor logging
+	supressLoggingOfEndpoints := []string{"GetConfiguredPlugins"}
+	var level klogv2.Level
+
+	// level=3 is default logging level
+	level = 3
+	for i := 0; i < len(supressLoggingOfEndpoints); i++ {
+		if strings.Contains(endpoint, supressLoggingOfEndpoints[i]) {
+			level = 4
+			break
+		}
+	}
+
+	return level
+}
+
+func LogRequest(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (response interface{}, err error) {
+
+	start := time.Now()
+	res, err := handler(ctx, req)
+
+	level := getLogLevelOfEndpoint(info.FullMethod)
+
+	// Format string : [status code] [duration] [full path]
+	// OK 97.752µs /kubeappsapis.core.packages.v1alpha1.PackagesService/GetAvailablePackageSummaries
+	klogv2.V(level).Infof("%v %s %s\n",
+		status.Code(err),
+		time.Since(start),
+		info.FullMethod)
+
+	return res, err
+}
+
 // Serve is the root command that is run when no other sub-commands are present.
 // It runs the gRPC service, registering the configured plugins.
 func Serve(serveOpts core.ServeOptions) error {
 	// Create the grpc server and register the reflection server (for now, useful for discovery
 	// using grpcurl) or similar.
-	logRequest := CreateRequestLogger()
-	grpcSrv := grpc.NewServer(grpc.ChainUnaryInterceptor(logRequest))
+
+	grpcSrv := grpc.NewServer(grpc.ChainUnaryInterceptor(LogRequest))
 	reflection.Register(grpcSrv)
 
 	// Create the http server, register our core service followed by any plugins.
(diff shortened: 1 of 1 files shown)


cmd/kubeapps-apis/server/server.go
```diff
@@ -65,13 +66,48 @@ func CreateRequestLogger() func(ctx context.Context, req interface{}, info *grpc
 	}
 }
 
+func getLogLevelOfEndpoint(endpoint string) klogv2.Level {
+
+	// Add all endpoint function names which you want to suppress in interceptor logging
+	supressLoggingOfEndpoints := []string{"GetConfiguredPlugins"}
+	var level klogv2.Level
+
+	// level=3 is default logging level
```
absoludity, line 75: The default log level is 3 if the user hasn't set a `-v` option on the cli... don't we need to default to whatever the user has specified (which will be 3 if nothing was specified)?

cmd/kubeapps-apis/server/server.go
```diff
@@ -65,13 +66,48 @@ func CreateRequestLogger() func(ctx context.Context, req interface{}, info *grpc
...
+	level = 3
+	for i := 0; i < len(supressLoggingOfEndpoints); i++ {
```
absoludity, line 77: It's useful having `suppressLoggingOfEndpoints` as a string slice if it's something configurable or we have more than one endpoint being hit for liveness/readiness... but otherwise, why not just do a single string comparison here?

cmd/kubeapps-apis/server/server.go
```diff
@@ -65,13 +66,48 @@ func CreateRequestLogger() func(ctx context.Context, req interface{}, info *grpc
...
+
+	return res, err
+}
+
 // Serve is the root command that is run when no other sub-commands are present.
 // It runs the gRPC service, registering the configured plugins.
 func Serve(serveOpts core.ServeOptions) error {
 	// Create the grpc server and register the reflection server (for now, useful for discovery
 	// using grpcurl) or similar.
-	logRequest := CreateRequestLogger()
```
absoludity, line 73: We need to delete the `CreateRequestLogger()` function to, don't we?

absoludity:
Thanks Satya! Just a couple of thoughts - otherwise +1

cmd/kubeapps-apis/server/server.go
satya-dillikar: I don't think we should complicate more by using the user settings (--v).

cmd/kubeapps-apis/server/server.go
satya-dillikar: Today, we have only one endpoint to suppress. I guess as we test the code more, we might have a few more endpoints.

I kept it as a string slice so that In the future, we add more names to suppress. (This will be a one-time change and I don't think we need any configurable option for now.

cmd/kubeapps-apis/server/server.go
satya-dillikar: Yes. We can delete the CreateRequestLogger()  as it is not used. 

I left it unused. If you want we can remove it.
Please let me know.

cmd/kubeapps-apis/server/server.go
absoludity: Sorry, I confused this in my head with something it is not... I was thinking we were deciding here what the default log level would be for the app, but it's not, it's just the level that will be used for the interceptor. Carry on :)

cmd/kubeapps-apis/server/server.go
absoludity: Yep, fine either way.

cmd/kubeapps-apis/server/server.go
absoludity: Yes please... we generally remove any unused code (it's just more code to maintain).

absoludity (approved):

PR #3960, opened 2021-12-15
Fix for malformed can-i URL (#2719)

### LINKED ISSUE #2719: Malformed can-i URL on non-kubeapps cluster
### Description:

405 errors are thrown when hitting the can-i URL in the cluster in which Kubeapps in installed on, but without declaring it under the `clusters:` object in the values.yaml.
It results in a `currentCluster == ""`, so when building the URL, the empty is passed through.

There is -no- UX impact, but it generates unexpected calls in Kubeops depending upon the browser.

1. performs a POST request to `api/v1/clusters//can-i`
2. got a 301 redirect to `api/v1/clusters/can-i`
3. browser (ignoring the RFC) changes the POST to a GET method
4. kubeops receives GET `api/v1/clusters/can-i`, returning a 405 method not allowed

1. Configure Kubeapps so that users can target a second cluster and not target the cluster on which Kubeapps is installed,
2. Click on the AppRepositories menu item
3. See the network inspector

405 status code response.

Either:

I) a 200 OK with `authorized: no` (since it using the serviceaccount, not the user account) (I'd go for this one)
ii) another meaningful and handled HTTP status code.

However, note that not generating a proper response for this case (`requestedCluster==""`) it will generate a 401 response, so users will get logged out each time they click on app-repositories. 

devel Kubeapps

_Originally posted by @antgamdia in https://github.com/kubeapps/kubeapps/pull/2712#discussion_r621000615_

castelblanque:

### Description of the change

This PR prevents FE from placing can-i requests with malformed URLs that otherwise end up in 404 or 405 HTTP errors.
Thought that it was more correct formally not placing the requests instead of returning `200 OK` with body `authorized: no`. Reason being that the URL was incorrect, and not the parameters.

### Benefits

Less useless requests from browser to backend for can-i checks.

### Possible drawbacks

N/A

### Applicable issues

- fixes #2719

### DIFF
--- a/dashboard/src/shared/Kube.test.ts
+++ b/dashboard/src/shared/Kube.test.ts
@@ -264,5 +264,9 @@ describe("App", () => {
       const allowed = await Kube.canI("cluster", "v1", "namespaces", "create", "");
       expect(allowed).toBe(true);
     });
+    it("should ignore empty clusters", async () => {
+      const allowed = await Kube.canI("", "v1", "namespaces", "create", "");
+      expect(allowed).toBe(false);
+    });
   });
 });
--- a/dashboard/src/shared/Kube.ts
+++ b/dashboard/src/shared/Kube.ts
@@ -120,6 +120,9 @@ export class Kube {
     namespace: string,
   ) {
     try {
+      if (!cluster) {
+        return false;
+      }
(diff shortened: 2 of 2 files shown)


antgamdia (approved):
Great, thanks for the fix! We are about to move this endpoint to the new resources API, but we better fix it anyway.

PR #3970, opened 2021-12-16
Check App CR creation when CreateInstalledPackage

### LINKED ISSUE #3852: Installed app not found right after installing a Carvel package
Parent issue: https://github.com/kubeapps/kubeapps/issues/3849

When a package is installed, we got redirected to the InstalledPackage view, which is great. The problem is that Kapp hasn't created the `App` resource yet. Alternatives (I'd go with 2)

1. Auto-retry from the UI in N attempts
    -  follow-up idea: should we add exp. backoff retries to every get-alike request to avoid transient API connectivity issues?
2. Check the `App` resource exists in the backend before sending the OK response back.

antgamdia:
### Description of the change

This PR simply adds a polling mechanism (interval/timeouts are harcoded, but could be split out into separate params if required) before sending OK when creating an installed package.

### Benefits

Since the App CR has to be created by kapp automatically before something can be considered as "installed", this active wait will ensure the App is present in the cluster; that is, no more UI errors after accessing to the installed view.

### Possible drawbacks

Can't think of; I mean, if the App CR is not created... that means something is wrong, so we better catch this error when executing the creation operation.

### Applicable issues

- fixes #3852

### Additional information

I want to DRY the tests up. Now we have the whole test suite and know what we need, we can refactor it and reduce the number of lines for creating the data. But it should be a follow-up PR (I've added it as a pending thing in the parent task).

It is possibly my last PR of the year before going on PTO :palm_tree:  :smile:

### DIFF
--- a/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
+++ b/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
@@ -27,6 +27,7 @@ import (
 	"google.golang.org/grpc/status"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 	log "k8s.io/klog/v2"
 )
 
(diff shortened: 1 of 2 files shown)


cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
```diff
@@ -526,6 +527,25 @@ func (s *Server) CreateInstalledPackage(ctx context.Context, request *corev1.Cre
 		return nil, errorByStatus("create", "PackageInstall", newPkgInstall.Name, err)
 	}
 
+	// TODO(agamez): some seconds should be enough, however, make it configurable
+	err = wait.PollImmediateWithContext(ctx, time.Second*1, time.Second*5, func(ctx context.Context) (bool, error) {
```
castelblanque, line 531: Awesome! Let's create a new issue to make the seconds configurable.

castelblanque (approved):

castelblanque:
> It is possibly my last PR of the year before going on PTO 🌴 😄

BTW enjoy your well-deserved holidays!!!

Issue #3680, opened 2021-10-27
Inconsistent handling of namespace-scoped users between view/deploy


dlaloue-vmware:
Create a namespace-scoped user to namespace X by granting the standard cluster role 'edit'.

The user can login to Kubeapps, confirm that only namespace X is available in the context, view the catalog (from global repositories, e.g. bitnami), select for example nginx from bitnami to install, configure everything and click deploy.

On clicking deploy, the action is not allowed because of not having access to the global repositories.
Access was not required by the kubeapps apis when looking at the catalog or app repository page.
why is it required for deploying?

![Screen Shot 2021-10-27 at 12 08 34 PM](https://user-images.githubusercontent.com/74025858/139130990-d253d837-60a1-439f-8790-10ade3c4f36d.png)

antgamdia:
Sorry for the long read, TL;DR: I think is necessary to create the rolebinding in a user's namespace to clusterrole `kubeapps:kubeapps:apprepositories-read`. Otherwise it won't work as expected at all.

---

I'm trying to repro this issue, although I thought https://github.com/kubeapps/kubeapps/pull/3700 would have addressed it, I think hasn't.

> Create a namespace-scoped user to namespace X by granting the standard cluster role 'edit'.

What do you mean by this? Aren't you granting any permissions to apprepostories at all? If so, I can repro it by creating a SA to get the token, for the sake of simplicity, which has rolebinding "edit" on the "default" namespace, that is:

```bash
kubectl create serviceaccount kubeapps-edit -n kubeapps
kubectl create rolebinding kubeapps-edit -n default --clusterrole=edit --serviceaccount kubeapps:kubeapps-editrolebinding.rbac.authorization.k8s.io/kubeapps-edit created
```

Assuming no permissions whatsoever on the `apprepositories` CR, the very first line of this function will fail:

https://github.com/kubeapps/kubeapps/blob/597de8d06b663172a462e9cd3cccb0973da27036/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go#L1058-L1063

However, I'd bet an "edit user" should, at least, have read permissions on the apprepos. Any reason not to grant them?

I mean, when no apprepos perms, the UI is not really happy about it:

You can't see apps from your private repo nor filter them
![image](https://user-images.githubusercontent.com/11535726/141171755-372c7fc5-d795-4a56-ac83-8581e37eed7a.png)

You can't perform any interactions whatsoever with the App Repositories tab:
[...]

PR #3990, opened 2021-12-23
Moved global repos to their own namespace and outside the kubeapps one

### LINKED ISSUE #3989: Update Kubeapps to keep global repositories in a separate namespace
### Description:

Modify Kubeapps so that global repositories CRD (accessible to all users anywhere) are kept in a new namespace specifically meant for global access.
This brings the benefit of being able to control the permissions to access those CRD on that namespace without affecting the Kubeapps namespace. Therefore we have a better separation of concerns: global repos on one namespace and Kubeapps on its own.

Refers to [this section of design documentation](https://docs.google.com/document/d/1YEeKC6nPLoq4oaxs9v8_UsmxrRfWxB6KCyqrh2-Q8x0/edit#heading=h.sogxgchn6bm2) and [this comment](https://github.com/kubeapps/kubeapps/issues/3680#issuecomment-965791778).

With this change in place, #3680 can be tackled to grant users with the clusterrole `kubeapps:{{ KUBEAPPS_NAMESPACE }}:apprepositories-read`  if they have the OOB `view` role. Same will apply for `edit` role and the `:apprepositories-write` clusterrole.

This will not affect the syncing jobs for indexing the packages, and two secrets will still be kept for each "per-namespace" repository: one in the namespace of the repo and one in the Kubeapps namespace. Moreover, I hope it helps on the implementation of [Move app repo sync to run in repo namespace (and cluster)](https://github.com/kubeapps/kubeapps/issues/1647).

Tightly related to #3680

castelblanque:
### Description of the change

Sorry for the long PR, but in order to have the feature working it is all or nothing 😄 

This PR moves the global repos from the Kubeapps namespace to its own namespace.

Based on the problem stated by @dlaloue-vmware  and the discussion in #3680, this allows to give read access to users to the global repos namespace, and therefore not having to grant them with access to the kubeapps namespace.

Changes are done e.g. to Helm templates, `repositories-controller`, `asset-syncer` and `kubeapps-apis`, so it is large change, but basically it is adding the global namespace parameter and substituting the kubeapps namespace where needed.

No changes done to the double secrets kept for non-global repos.
Modifications in Kubeops are done for compatibility reasons.

### Benefits

Global repos can be placed into their own namespace, and therefore better RBAC can be applied separated from the kubeapps namespace. Hopefully this helps somehow to implement #1647.

### Possible drawbacks

Assets syncing should not be affected, but we could keep an eye on it.

### Applicable issues

- fixes #3989


pkg/kube/kube_handler.go
```diff
@@ -106,9 +106,10 @@ type PinnipedConciergeConfig struct {
 
 // ClustersConfig is an alias for a map of additional cluster configs.
 type ClustersConfig struct {
-	KubeappsClusterName string
-	PinnipedProxyURL    string
-	Clusters            map[string]ClusterConfig
+	KubeappsClusterName  string
+	GlobalReposNamespace string
```
castelblanque, line 110: Not sure if this is the best place to add the global repos namespace...otherwise we might need to change the register RPC method signature.

chart/kubeapps/templates/apprepository/deployment.yaml
```diff
@@ -76,6 +76,7 @@ spec:
             {{- end }}
             - --repo-sync-cmd=/asset-syncer
             - --namespace={{ .Release.Namespace }}
+            - --global-repos-namespace={{ .Release.Namespace }}-{{ .Values.apprepository.globalReposNamespaceSuffix }}
```
castelblanque, line 79: Preferred to add a proper flag instead of passing the value through env variables

chart/kubeapps/templates/shared/global-repos-namespace.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: "{{ .Release.Namespace }}-{{ .Values.apprepository.globalReposNamespaceSuffix }}"
```
castelblanque, line 4: Creates the global repos namespace

chart/kubeapps/templates/apprepository/apprepositories-secret.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
   {{- if .namespace }}
   namespace: {{ .namespace | quote }}
   {{- else }}
-  namespace: {{ $.Release.Namespace | quote }}
+  namespace: {{ printf "%s-%s" $.Release.Namespace $.Values.apprepository.globalReposNamespaceSuffix | quote }}
```
absoludity, line 10: Regarding making this a non-breaking change, could we instead do:

```suggestion
  namespace: {{ printf "%s%s" $.Release.Namespace $.Values.apprepository.globalReposNamespaceSuffix | quote }}
```
so that if `globalReposNamespaceSuffix` is empty (which can be the default, initially), we will have the existing behavior? We can then update the default value to `-repos-global` when we do a major version change, breaking the backwards compatibility?

It'd require a conditional around the namespace creation later too, etc.

chart/kubeapps/templates/apprepository/deployment.yaml
absoludity: +100.

pkg/kube/kube_handler.go
absoludity: Yeah, I think that's fine for now.

absoludity (commented):
> Sorry for the long PR,

Heh - at `+161 −91` this isn't a long PR, the focused change just touches a lot of files :) Start to worry about that if it's over +400 (IMO).

> but in order to have the feature working it is all or nothing smile

Hmm, I'm not sure this is the case... in particular, have you checked how this code handles upgrading an existing deployment, where there are existing global repos? Won't they disappear with this change (anything other than the default bitnami repo, that is). I think it'd be easy to update so this is a non-breaking change by. See what you think (more context inline).

chart/kubeapps/templates/apprepository/apprepositories-secret.yaml
castelblanque: Thanks for the suggestion Michael, it makes sense. So, defaulting now to the same `kubeapps` namespace, would mean leaving without effect the purpose of this change until next major version 3.x?
As-is in a Kubeapps upgrade scenario, new global repos will go to the new ns and old global repos will stay in the `kubeapps` ns, just not being considered global anymore.

castelblanque:
> Hmm, I'm not sure this is the case... in particular, have you checked how this code handles upgrading an existing deployment, where there are existing global repos? Won't they disappear with this change (anything other than the default bitnami repo, that is). I think it'd be easy to update so this is a non-breaking change by. See what you think (more context inline).

I don't remember from the top of my head if I tested that scenario before Christmas 😅 , will give it a try. We are talking about:
- Install Kubeapps v2.4.2 (global repos in ns `kubeapps`)
- Deploy app from one of the global repos
- Upgrade Kubeapps to this snapshot (global repos in new global ns)
- Upgrade app?

Old global repos will stay in `kubeapps` and repos coming from Helm values will go to the new ns.

chart/kubeapps/templates/apprepository/apprepositories-secret.yaml
absoludity: > Thanks for the suggestion Michael, it makes sense. So, defaulting now to the same `kubeapps` namespace, would mean leaving without effect the purpose of this change until next major version 3.x?

By default, yes, but we can verify it works as expected, even update CI to use it (by setting a value for the suffix) if we want, to ensure that switching the default later is a non-event (other than the instructions for users).

> As-is in a Kubeapps upgrade scenario, new global repos will go to the new ns and old global repos will stay in the `kubeapps` ns, just not being considered global anymore.

Yep, we'll just ensure we document that in the release notes so people can move their repos as needed.

absoludity:
> > Hmm, I'm not sure this is the case... in particular, have you checked how this code handles upgrading an existing deployment, where there are existing global repos? Won't they disappear with this change (anything other than the default bitnami repo, that is). I think it'd be easy to update so this is a non-breaking change by. See what you think (more context inline).
> 
> I don't remember from the top of my head if I tested that scenario before Christmas sweat_smile , will give it a try. We are talking about:
> 
>     * Install Kubeapps v2.4.2 (global repos in ns `kubeapps`)
> 
>     * Deploy app from one of the global repos
> 
>     * Upgrade Kubeapps to this snapshot (global repos in new global ns)
> 
>     * Upgrade app?
> 
> 
> Old global repos will stay in `kubeapps` and repos coming from Helm values will go to the new ns.

Yeah - in particular, does that affect existing installations (ie. if I'd installed apache in `my-namespace`, because apache was from the bitnami repo which was in the global repo before the upgrade, after the upgrade I may not be able to upgrade the app, as kubeapps may have lost the link to the available package?)

chart/kubeapps/values.yaml
```diff
@@ -719,6 +719,9 @@ apprepository:
     ##   - myRegistryKeySecretName
     ##
     pullSecrets: []
+  ## @param apprepository.globalReposNamespaceSuffix Suffix for the namespace of global repos
+  ##
+  globalReposNamespaceSuffix: repos-global
```
antgamdia, line 724: Agree with the rest of the comments in that the suffix should also include the `-` to prevent backward-incompatible changes.

Also, in order to stick close to similar names within our organization, why not go with a similar suffix as Carvel people does: `-packaging-global` ?

castelblanque: I find using the word `repos` more intuitive 😄 
Is there any accepted terminology for this?

castelblanque:
> Yeah - in particular, does that affect existing installations (ie. if I'd installed apache in `my-namespace`, because apache was from the bitnami repo which was in the global repo before the upgrade, after the upgrade I may not be able to upgrade the app, as kubeapps may have lost the link to the available package?)

@absoludity I tested the scenario of updating Kubeapps to the new global repos ns, and as expected:

- Custom global repos created prior to Kubeapps update were kept in `kubeapps` ns, but they are not global anymore.
- When retrieving details of already installed packages from old custom global repos, this error appears:
image
- When switching context to ns `kubeapps`, old custom global repos packages appear and they can be installed normally, only in `kubeapps` ns.
- If the old global repo is created/moved into the new global repos namespace, existing installations can be updated. Maybe in the future we could provide a mechanism to move repositories from one namespace to another? This way admins could move their repos to other namespaces after updating.

By now we are safe as we are defaulting to `kubeapps` ns (no global suffix).
When a proper global namespace will be used, existing installations from custom global repositories will be affected.

absoludity:
Excellent, thanks for testing Rafa.

> * If the old global repo is created/moved into the new global repos namespace, existing installations can be updated. Maybe in the future we could provide a mechanism to move repositories from one namespace to another? This way admins could move their repos to other namespaces after updating.

Great, yes, at the very least, instructions in the release notes when the default changes.

+1, thanks!

absoludity (approved):

PR #3998, opened 2021-12-24
Added some more cases to the tests for chars encoding (#3758)

### LINKED ISSUE #3758: Problems with charts with spaces in their name

castelblanque:

### Description of the change

Extension of tests introduced in #3758 to cover characters other than space in chart names, although Helm specifies that `chart names must contain only lower case letters and numbers. Words may be separated with dashes (-)`. See [Helm general conventions](https://helm.sh/docs/chart_best_practices/conventions/#chart-names).

Therefore, the special characters tested in this PR won't have effect for Helm, as it will fail if chart contains them. Maybe for systems other than Helm is useful.

### Benefits

Bit more test coverage for special characters encoding on chart names

### Possible drawbacks

N/A

### Applicable issues

- fixes #3758

### Additional information

Maybe we should limit the possible characters available for package names.

### DIFF
--- a/cmd/asset-syncer/server/testdata/helm-index-spaces.yaml
+++ b/cmd/asset-syncer/server/testdata/helm-index-spaces.yaml
@@ -5,3 +5,9 @@ entries:
   chart-without-spaces:
     - appVersion: v2
       name: "chart-without-spaces"
+  chart-with-special-chars-1:
+    - appVersion: v3
+      name: "chart\x23with\x23hashes"
+  chart-with-special-chars-4:
+    - appVersion: v1
+      name: "chart\x24with\x24chars"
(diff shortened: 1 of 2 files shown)


antgamdia (approved):
Great, thank you!

Issue #3890, opened 2021-12-02
Remove duplicated pkg version selector in the `PackageHeader` when upgrading


antgamdia:
As per https://github.com/kubeapps/kubeapps/pull/3880#issuecomment-984092580, we noticed we are duplicating the package version selector in the upgrade view, which could be eventually misleading for users (although currently, both dropdowns will perform the same action).

Since it is a custom component, adding an attribute to the `PackageHeader` to conditionally render it or not would make the trick.

However, although the implementation is easy, I'm not 100% sure about the implications on the CI, hence creating a separate issue for tracking it.

PR #4013, opened 2022-01-03
Disable the version selector when upgrading pkgs

antgamdia:
### Description of the change

In this PR, we add an attribute to conditionally render the version selector in the header, so that we can choose where we should render it.

### Benefits

No duplicated version selectors. In the picture, install vs upgrade views:

image

### Possible drawbacks

If the CI fails, we should also add some fixes there.

### Applicable issues

- fixes https://github.com/kubeapps/kubeapps/issues/3890

### Additional information

(it's my first PR of 2022 :P)


absoludity (approved):
Great, thanks Antonio.

PR #4019, opened 2022-01-03
Set variable for Kind cluster name in makefile

castelblanque:

### Description of the change

This PR brings a small adjustment for the Makefile in order to use the right cluster name when accessing the Kind node.

Kind creates nodes in Docker based on the `--name` parameter.
```
kind create cluster \
		[...]
		--name ${CLUSTER_NAME} \
                [...]
```

So when accessing those nodes, the corresponding name from variable is now used.

### Benefits

Previously the key+cert files were being retrieved from a hardcoded Kind node name. Now it is in line with the cluster name.

### Possible drawbacks

N/A

### Applicable issues

None


antgamdia (approved):
Great! Thanks

Issue #3853, opened 2021-11-29
Provide initial values when installing a Carvel package


antgamdia:
Parent issue: https://github.com/kubeapps/kubeapps/issues/3849

Carvel packages do not have any "default values.yaml" file, or, at least, there is no standard/recommended Carvel-way to do that. As a result, users will see an empty textarea waiting for them to add all the values. Ideas (I'd go with 1)

1. Auto-infer a values.yaml file using the OpenAPI Schema (follow-up idea: we are gonna need to parse this schema at some point... should we just implement this "advanced form" now -two birds one stone - ? 
2. Do nothing until Carvel folks recommend something.

antgamdia:
As per @vrabbi's suggestion (thanks for the pointer!) , le also consider the work done in https://github.com/vmware-tanzu/tanzu-framework/pull/1422/

antgamdia:
Confirming this approach is doable :)

![image](https://user-images.githubusercontent.com/11535726/147957546-33386615-2323-4b67-92ab-a40bd5db7a77.png)

Still have to check a couple of things (marshaling issues, checking why Tanzu Framework folks decided to adapt the k8s code, add tests, etc.).

antgamdia:
#### Quick update (Jan 4th):

After a few tweaks in the code, the values can be auto-inferred (using the `default` in the Schema and fallback to `""`, `0`, `false`, `[]` and `{}`).

![image](https://user-images.githubusercontent.com/11535726/148054463-81e158d3-a64f-45fd-966c-94f23d7fc390.png)

Just out of curiosity, I've pulled the changes in https://github.com/kubeapps/kubeapps/pull/3527. It works pretty decently (though some performance issues appear when the schema is large). Perhaps we may want to add these changes in a follow-up PR or even work on the "advanced form" tab,

![image](https://user-images.githubusercontent.com/11535726/148055055-15c0e0e6-a0a6-4460-b603-5e7cfea4092f.png)

vrabbi:
this looks awesome! the form would also be great to have!

PR #4023, opened 2022-01-04
Set the `ValuesSchema` and `DefaultValues` in Carvel pkgs

antgamdia:
### Description of the change

This PR is aimed at providing a set of initial values for users when installing/upgrading a Carvel package. Particularly, in this PR:

* The raw schema is passed through the wire (for the custom form, if ever implemented - see https://github.com/kubeapps/kubeapps/issues/3853#issuecomment-1004746080)
* The default values are calculated as follows: if a `.Default` property exists, use it; otherwise, fall back to each data type default value (eg., `""` or `0`)
* Several tests have been written since I noticed the `server_utils_test` wasn't being tested. 

### Benefits

Users will get more information about the values that a package requires, being a similar UX as in other pkg formats (Helm).

### Possible drawbacks

Some problems arise:

* I've noticed some schemas (from our usual Carvel repos) have problems when performing the yaml un/marshaling operation. For example, `Fluent Bit` package fails. In these cases, I've added a commented-out warning message.
  * Edit: the root cause seems to be an `additionalProperties: string` (where it should be either an object or a boolean; [reported to the TCE pkgs team](https://kubernetes.slack.com/archives/C02GY94A8KT/p1641389173476200)) https://github.com/vmware-tanzu/community-edition/issues/2772
* The logic behind retrieving the defaults from is heavily inspired by https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apiserver/schema/defaulting/algorithm.go; I don't like the idea of re-writing k8s code, but I haven't been able to find any other dep for the same purpose. Happy to hear any alternatives.

### Applicable issues

- fixes  https://github.com/kubeapps/kubeapps/issues/3853

### Additional information

IRL example: 
image

### DIFF
--- a/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_utils.go
+++ b/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_utils.go
@@ -14,6 +14,7 @@ package main
 
 import (
 	"fmt"
+	"reflect"
 	"regexp"
 	"sort"
 	"strconv"
@@ -25,7 +26,11 @@ import (
 	datapackagingv1alpha1 "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apiserver/apis/datapackaging/v1alpha1"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
+	structuralschema "k8s.io/apiextensions-apiserver/pkg/apiserver/schema"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/yaml"
 )
 
 type pkgSemver struct {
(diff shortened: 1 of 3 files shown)


cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_utils.go
```diff
@@ -126,5 +131,126 @@ func extractValue(body string, key string) string {
...
+	return string(yaml), nil
+}
+
+// Default does defaulting of x depending on default values in s.
+// Based upon https://github.com/kubernetes/apiextensions-apiserver/blob/release-1.21/pkg/apiserver/schema/defaulting/algorithm.go
+// Plus modifications from https://github.com/vmware-tanzu/tanzu-framework/pull/1422
+// In short, it differs from upstream in that:
+// -- 1. Prevent deep copy of int as it panics
+// -- 2. For type object scan the first level properties for any defaults to create an empty map to populate
+// -- 3. If the property does not have a default, add one based on the type ("", false,)
```
absoludity, line 170: Thanks for the details there.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_utils_test.go
```diff
@@ -0,0 +1,524 @@
...
+    type: object`,
+		),
+			`missingDefaultArray: []
+missingDefaultBoolean: false
+missingDefaultInteger: 0
+missingDefaultNumber: 0
+missingDefaultObject: {}
+missingDefaultString: ""
+valueWithDefault: 80
+`},
```
absoludity, line 249: Looks like this should be 7 test cases (you've even added the infra for iterating the test cases) that you've put in one test-case? (ie. it will fail if any of the parts fail). Not a big deal and fine as is, just unexpected.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_utils_test.go
```diff
@@ -0,0 +1,524 @@
...
+				t.Fatal(err)
+			}
+
+			var expected interface{}
+			if err := json.Unmarshal([]byte(tt.expected), &expected); err != nil {
+				t.Fatal(err)
+			}
+
+			defaultValues(in, tt.schema)
+			if !reflect.DeepEqual(in, expected) {
```
absoludity, line 510: I'm guessing there's a reason that the upstream change doesn't use `cmp.Equal` instead of `reflect.DeepEqual`? Either way we probably want to keep the code as similar as possible.

absoludity (approved):
LGTM, though is it not possible for the rendered default values to be commented out as we'd discussed before the break? The Carvel team was pretty keen to be able to differentiate between values that are explicitly set by the user and default values, whereas generating them like this removes that distinction. I guess there are issues with rendering the default values commented out so the user can just uncomment any parts they want to set?

(I realise this would be different to helm, where we include the provided values, but it is a different situation - Carvel doesn't provide values.yaml, and the carvel team seem pretty particular about this, so might be worth investigating).

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_utils_test.go
antgamdia: Same... I don't know, but I tried to keep it as aligned as possible to the upstream version :S

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_utils_test.go
antgamdia: haha, you're right; I started writing the different cases at the same time, but I forgot to split them apart. Done (+ few additional test cases more)

antgamdia:
> LGTM, though is it not possible for the rendered default values to be commented out as we'd discussed before the break? 

I did try to comment them out and even add some comments with the `.description` or even `.examples`, however, I wasn't able to add them in in the yaml. Comments and yaml has [always been a pain](https://stackoverflow.com/questions/60891174/i-want-to-load-a-yaml-file-possibly-edit-the-data-and-then-dump-it-again-how), but there are some implementations supporting them (using Nodes in `go-yaml` v3) but I don't see any quick way to access every node to add the comments there.

An alternative would be re-implementing the default function to directly write the yaml nodes instead of relying on an ulterior marshaling process. However, it seemed non-trivial and given that we may want to add the "advanced form tab", it wouldn't be that useful then.

What I have done, instead, is to comment out the whole generated string afterward (appending the `#` direcly`). 
Note that our ACE text editor has the `ctrl + /` keybind enabled, so we can uncomment in block easily.

image. 
See what you think.

absoludity:
Looks awesome. +1 from me for providing them commented out.

Issue #3922, opened 2021-12-09
AppRepo form currently requires fetching secrets from k8s API server


absoludity:
(I'm using this issue as a bit of a rubber-duck, but am keen for any other input too).

### Description:

I'm currently working on #3896 to remove the remaining dashboard calls to the K8s API server, in particular, where the Dashboard fetches secrets.

The only place where the dashboard fetches secrets is when editing an AppRepository via the AppRepo form, populating any credential field that had been set with the parsed secret value, such as:

![edit-repo-with-basic-auth](https://user-images.githubusercontent.com/497518/145339271-f61503cd-7c3c-41fc-be25-90d92ed8959e.png)

All of basic, bearer and custom options result in an opaque K8s secret with the data:

```
data:
  authorizationHeader: <b64-encoded-data-for-authorization-header>
```

rather than using secrets with a specific type ([such as `kubernetes.io/basic`](https://kubernetes.io/docs/concepts/configuration/secret/#basic-authentication-secret)). So once the secret is created, there is nothing relating the type of authorization that it represents for the app repo other than parsing the actual secret data, which is [what the dashboard currently does :/ ](https://github.com/kubeapps/kubeapps/blob/v2.4.2/dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx#L128-L151).

The constraints that I believe I have are:

- We don't want to add a GRPC endpoint to the resources plugin that just allows fetching secrets using the users credential (not far from proxying to the API server)
- We don't want to make too much change on the existing app repository API endpoint in the current backend, since replacing it with the apis server is in the works (#3496 )
- We do want to remove the dashboard's access to the k8s API server.

The main options I can see to move forward here are:

1. Update the dashboard to use the `resources.GetSecretNames` endpoint as I had planned, which includes the secret type for each name, then
2. Potentially change the UX experience so that you can't see previously entered secret values and are instead only able to replace the value with a new one when editing an AppRepository,

Or

3. Keep the same UX so that you can view existing credentials, by returning the secret data (both `authorizationHeader` and `ca.crt` when they exist) as part of the app repository from the backend for now.

AFAICT, if I implement 3 now, we will still need the `GetSecretNames` to populate the dropdown for the docker-config json's for image pull secrets. Anyway, I'll sleep on this and pick it up tomorrow.

castelblanque:
[...]

PR #4027, opened 2022-01-05
Add GetAppRepository to kubeops backend

absoludity:
### Description of the change

This PR primarily adds an endpoint to the current kubeops backend handler to get an app repository, importantly, that includes the related auth header/ca.crt secret in the response. The dashboard currently relies on fetching app repo secrets directly from the k8s API server, which we want to remove. This is an initial work-around to avoid requiring a generic get secret endpoint (see #3922 for discussion of options).

While there, I've fixed the `returnK8sError` helper which was logging incorrect and hence confusing information.

### Benefits

It'll enable the followup PR which updates the dashboard to no longer fetch secrets via the k8s API.

### Applicable issues

- Ref #3922

### DIFF
--- a/pkg/http-handler/http-handler.go
+++ b/pkg/http-handler/http-handler.go
@@ -60,13 +60,13 @@ func JSONError(w http.ResponseWriter, err interface{}, code int) {
 	json.NewEncoder(w).Encode(err)
 }
 
-func returnK8sError(err error, w http.ResponseWriter) {
+func returnK8sError(err error, action string, resource string, w http.ResponseWriter) {
 	if statusErr, ok := err.(*k8sErrors.StatusError); ok {
 		status := statusErr.ErrStatus
-		log.Infof("unable to create app repo: %v", status.Reason)
+		log.Infof("unable to %s %s: %v", action, resource, status.Reason)
 		JSONError(w, statusErr.ErrStatus, int(status.Code))
 	} else {
-		log.Errorf("unable to create app repo: %v", err)
+		log.Errorf("unable to %s %s: %v", action, resource, err)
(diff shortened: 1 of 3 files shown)


pkg/http-handler/http-handler.go
```diff
@@ -60,13 +61,13 @@ func JSONError(w http.ResponseWriter, err interface{}, code int) {
 	json.NewEncoder(w).Encode(err)
 }
 
-func returnK8sError(err error, w http.ResponseWriter) {
+func returnK8sError(err error, action string, resource string, w http.ResponseWriter) {
```
antgamdia, line 64: Great, more meaningful error messages

pkg/http-handler/http-handler.go
```diff
@@ -261,17 +262,67 @@ func DeleteAppRepository(kubeHandler kube.AuthHandler) func(w http.ResponseWrite
...
 		err = clientset.DeleteAppRepository(repoName, requestNamespace)
 		if err != nil {
-			returnK8sError(err, w)
+			returnK8sError(err, "delete", "AppRepository", w)
 		}
 	}
 }
 
+// GetAppRepository gets an App Repository with a related secret if present.
+func GetAppRepository(kubeHandler kube.AuthHandler) func(w http.ResponseWriter, req *http.Request) {
```
antgamdia, line 277: I guess this function will get eventually moved to the new repositories API

antgamdia (approved):
Great, thanks!

pkg/http-handler/http-handler.go
absoludity: Yep, indeed, but that needs to be properly designed for a more generic approach etc.

PR #4028, opened 2022-01-05
Remove repo secrets from redux state, fetch on mount

absoludity:
### Description of the change

Following on from #4027, this PR updates the AppRepoForm component so that it no longer receives a related secret as a prop (from the redux state?), but rather fetches the secret when mounting and uses the local component state.

### Benefits

The new endpoint for fetching a secret for a repo's secret is used (rather than the k8s API), and it avoids the need to update the redux state when a repo is updated (since we always fetch the secret when loading the form). Also avoids storing all repo secrets in the redux state (and removes the related actions/reducers).

### Applicable issues

- ref #3922 

### Additional information

As always, testing the combination of an async react hook (useEffect) turned out to be painful to realise, but got there after a lot of head scratching.

I still need to test this IRL locally before landing (once CI passes etc.)

### DIFF
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -75,6 +75,7 @@
   "devDependencies": {
     "@formatjs/cli": "^4.7.0",
     "@improbable-eng/grpc-web-fake-transport": "^0.15.0",
+    "@testing-library/react": "^12.1.2",
     "@types/enzyme": "^3.10.11",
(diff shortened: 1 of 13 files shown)


antgamdia (approved):
Great, happy to see we're getting rid of unexpected direct calls to the k8s API.
Certainly, we should explore the capabilities of the "testing library" when implementing future tests.

PR #4034, opened 2022-01-06
No more secret requests to K8s API

absoludity:
### Description of the change

This PR follows on from #4028, updating the AppRepoForm so that it fetches the image pull secret *names* only (from the resources API), rather than fetching (and storing in global state) the image pull secrets for a namespace. Also updated, while there, to no longer use global state for these, but rather just local state on the form.

It also does some cleaning up, so that `shared/Secret.ts` no longer has any references to the k8s API, as well as removing the urls from `shared/urls` etc.

Tested IRL.

### Benefits

Kubeapps no longer hits the k8s API for secrets, nor does it expose any method for fetching secrets generally.

### Applicable issues

- ref #3896

### DIFF
--- a/dashboard/src/actions/repos.ts
+++ b/dashboard/src/actions/repos.ts
@@ -61,13 +61,6 @@ export const errorRepos = createAction("ERROR_REPOS", resolve => {
     resolve({ err, op });
 });
 
-export const requestImagePullSecrets = createAction("REQUEST_IMAGE_PULL_SECRETS", resolve => {
-  return (namespace: string) => resolve(namespace);
-});
-export const receiveImagePullSecrets = createAction("RECEIVE_IMAGE_PULL_SECRETS", resolve => {
-  return (secrets: ISecret[]) => resolve(secrets);
-});
-
 export const createImagePullSecret = createAction("CREATE_IMAGE_PULL_SECRET", resolve => {
   return (secretName: string) => resolve(secretName);
 });
@@ -87,8 +80,6 @@ const allActions = [
   requestRepo,
   redirect,
   redirected,
-  requestImagePullSecrets,
-  receiveImagePullSecrets,
   createImagePullSecret,
 ];
 export type AppReposAction = ActionType<typeof allActions[number]>;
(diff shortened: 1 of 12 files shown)


dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
```diff
@@ -71,31 +72,34 @@ export function AppRepoForm(props: IAppRepoFormProps) {
...
     // Select the pull secrets if they are already selected in the existing repo
-    imagePullSecrets.forEach(pullSecret => {
-      const secretName = pullSecret.metadata.name;
+    imagePullSecrets.forEach(secretName => {
       if (repo?.spec?.dockerRegistrySecrets?.some(s => s === secretName)) {
         setSelectedImagePullSecret(secretName);
       }
     });
-  }, [imagePullSecrets, repo, selectedImagePullSecret]);
+  }, [imagePullSecrets, repo]);
```
absoludity, line 102: I removed `selectedImagePullSecret` from the dependencies here, as it was causing a bug: every time you selected a different image pull secret, it would trigger this effect, which would set the selected image pull secret back to the one that matches the repos current reference, effectively stopping you from ever changing it.

dashboard/src/shared/Secret.ts
```diff
@@ -9,16 +6,22 @@ import {
...
+  public static async getDockerConfigSecretNames(cluster: string, namespace: string) {
+    const result = await this.resourcesClient().GetSecretNames({
+      context: {
+        cluster,
+        namespace,
+      },
+    });
+
+    const secretNames = [];
+    for (const [name, type] of Object.entries(result.secretNames)) {
```
antgamdia, line 19: Perhaps less wordy with a [for..in](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in) statement but fine either way

dashboard/src/actions/repos.ts
```diff
@@ -358,30 +348,6 @@ export function findPackageInRepo(
...
-  namespace: string,
-): ThunkAction<Promise<void>, IStoreState, null, AppReposAction> {
-  return async (dispatch, getState) => {
-    const {
-      clusters: { currentCluster },
-    } = getState();
-    try {
-      dispatch(requestImagePullSecrets(namespace));
-      // TODO(andresmgot): Create an endpoint for returning just the list of secret names
-      // to avoid listing all the secrets with protected information
```
antgamdia, line 371: We are finally doing it :P

antgamdia (approved):
Great! From now on, the browser's console will look way better without all these secret requests!

dashboard/src/actions/repos.ts
absoludity: Yeah, I was sad I couldn't close the related issue (because it was already closed, not sure why).

PR #4043, opened 2022-01-10
Update dashboard's validateToken not to use k8s api server.

absoludity:

### Description of the change

Part one of removing `shared/Auth`'s use of the k8s API server, this PR updates the `validateToken` check, used when Kubeapps is configured with token authentication, so that it hits the resources API (`checkNamespace`), rather than hitting the k8s API server directly.

To do so, I had to correct the error handling in the resources plugin to differentiate between 401s and 403s (and will later follow up as per the TODO there).

I also fixed an issue where the dashboard was checking cookie auth based only on the loginURL being set, which it is by default anyway (which resulted in an extra error being displayed when token login failed).

### Benefits

One more action no longer hitting k8s API.

### Applicable issues

- Ref #3896

### DIFF
--- a/cmd/kubeapps-apis/plugins/resources/v1alpha1/server.go
+++ b/cmd/kubeapps-apis/plugins/resources/v1alpha1/server.go
@@ -415,14 +415,18 @@ func resourceRefsEqual(r1, r2 *pkgsGRPCv1alpha1.ResourceRef) bool {
 }
 
 // errorByStatus generates a meaningful error message
+// TODO(minelson): Move to a shared helper for plugins interacting
+// with the k8s cluster.
 func errorByStatus(verb, resource, identifier string, err error) error {
 	if identifier == "" {
 		identifier = "all"
 	}
 	if errors.IsNotFound(err) {
 		return status.Errorf(codes.NotFound, "unable to %s the %s '%s' due to '%v'", verb, resource, identifier, err)
-	} else if errors.IsForbidden(err) || errors.IsUnauthorized(err) {
-		return status.Errorf(codes.PermissionDenied, "Unauthorized to %s the %s '%s' due to '%v'", verb, resource, identifier, err)
+	} else if errors.IsForbidden(err) {
+		return status.Errorf(codes.PermissionDenied, "Forbidden to %s the %s '%s' due to '%v'", verb, resource, identifier, err)
+	} else if errors.IsUnauthorized(err) {
+		return status.Errorf(codes.Unauthenticated, "Authorization required to %s the %s '%s' due to '%v'", verb, resource, identifier, err)
 	} else if errors.IsAlreadyExists(err) {
 		return status.Errorf(codes.AlreadyExists, "Cannot %s the %s '%s' due to '%v' as it already exists", verb, resource, identifier, err)
 	}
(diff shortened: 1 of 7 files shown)


dashboard/src/shared/Auth.ts
```diff
@@ -60,21 +65,27 @@ export class Auth {
   // Throws an error if the token is invalid
   public static async validateToken(cluster: string, token: string) {
     try {
-      await Axios.get(url.api.k8s.base(cluster) + "/", {
-        headers: { Authorization: `Bearer ${token}` },
+      await this.resourcesClient(token).CheckNamespaceExists({
```
antgamdia, line 68: +1 to this approach; however, I wonder whether we should use an alternative function here. I mean, at first glance, calling `CheckNamespaceExists` for the `validateToken` function seems a bit misleading.

Perhaps we could use the `can-i` operation there or, even better, directly use the k8s object aimed at validating tokens, the `TokenReview` one ([more info](https://kubernetes.io/docs/reference/access-authn-authz/authentication/)):


```json
{
  "apiVersion": "authentication.k8s.io/v1",
  "kind": "TokenReview",
  "spec": {
    "token": "014fbff9a07c...", 
  }
}
```

dashboard/src/shared/Auth.test.ts
```diff
@@ -1,53 +1,71 @@
...
-        });
-        Axios.get = mock;
+        mockClientCheckNamespaceExists = jest
+          .fn()
+          .mockImplementation(() => Promise.reject({ code: testCase.grpcCode }));
+        jest
+          .spyOn(client, "CheckNamespaceExists")
+          .mockImplementation(mockClientCheckNamespaceExists);
         // TODO(absoludity): tried using `expect(fn()).rejects.toThrow()` but it seems we need
         // to upgrade jest for `toThrow()` to work with async.
```
antgamdia, line 70: Just wondering if this ToDo still applies after having applied several upgrades to the jest deps.

cmd/kubeapps-apis/plugins/resources/v1alpha1/server.go
```diff
@@ -415,14 +415,18 @@ func resourceRefsEqual(r1, r2 *pkgsGRPCv1alpha1.ResourceRef) bool {
 }
 
 // errorByStatus generates a meaningful error message
+// TODO(minelson): Move to a shared helper for plugins interacting
+// with the k8s cluster.
```
antgamdia, line 419: +1, we are using this function also in the flux and carvel plugins

antgamdia (approved):
Thanks for the changes, +1 anyway, but have a look at the `TokenReview` comment

dashboard/src/shared/Auth.ts
absoludity: Yeah, I'd initially gone along this direction, adding a new endpoint to the resources plugin (didn't think of using the TokenReview - that would have been a good match), but the reason I stopped was that this functionality (for `validateToken`) is relevant only for the "demo-only" functionality of using service account tokens to authenticate with Kubeapps.

The existing functionality just tries to get the root URL of the k8s API server and I realised that the endpoint is actually arbitrary - the check just needs to see a 200 or 403 - either is fine.

+1 to switch it over to a `can-i` endpoint, once we have one available in the resources plugin.

dashboard/src/shared/Auth.test.ts
absoludity: Nope, it works with the `.toThrow` now in an async function. Thanks for the nudge, updated.

PR #4045, opened 2022-01-10
Update cookie auth check to no longer use axios/k8s api server

absoludity:

### Description of the change

Follows on from #4043, this time updating the cookie-based authentication so that it no longer uses a request to the k8s API server to determine the authentication status.

I'd originally planned to remove axios from `shared/Auth` (with the `isAnonymous` and `is403FromAuthProxy` functions), but then found they are still used in `shared/AxiosInstance` which is still used by the operator support (which will continue for now to query the k8s API directly).

As with the previous PR the pain here was testing IRL locally the various possibilities to find the best solution.

### Benefits

Hopefully no more requests to the k8s api server, outside of the operator support (though I'll do a proper audit and test tomorrow).

### Applicable issues

- ref #3896

### DIFF
--- a/dashboard/src/shared/Auth.ts
+++ b/dashboard/src/shared/Auth.ts
@@ -1,7 +1,6 @@
-import Axios, { AxiosResponse } from "axios";
+import { AxiosResponse } from "axios";
 import * as jwt from "jsonwebtoken";
 import { get } from "lodash";
-import * as url from "shared/url";
 import { IConfig } from "./Config";
 import { KubeappsGrpcClient } from "./KubeappsGrpcClient";
 import { grpc } from "@improbable-eng/grpc-web";
@@ -90,6 +89,16 @@ export class Auth {
     }
   }
 
+  // is403FromAPIsServer returns true if the response is a 403 determined to have originated
+  // from the grpc-web APIs server, rather than the auth proxy.
+  public static is403FromAPIsServer(e: any): boolean {
+    const contentType = e.metadata?.headersMap["content-type"] as string[];
+    if (contentType.some(v => v === "application/grpc-web+proto")) {
+      return true;
+    }
+    return false;
+  }
+
   // is403FromAuthProxy returns true if the response is a 403 determined to have originated
   // from the auth proxy itself, rather than upstream.
   //
@@ -99,6 +108,8 @@ export class Auth {
   // upstream result). Hence encapsulating this ugliness here so we can fix
   // it in the one spot. We may need to query `/oauth2/info` to avoid potential
   // false positives.
+  // Note: This function is only used now by AxiosInstance which
+  // in turn is only used for operators support.
   public static is403FromAuthProxy(r: AxiosResponse<any>): boolean {
     if (r.data && typeof r.data === "string" && r.data.match("system:serviceaccount")) {
       // If the error message is related to a service account is not from the auth proxy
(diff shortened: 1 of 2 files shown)


dashboard/src/shared/Auth.ts
```diff
@@ -112,39 +123,43 @@ export class Auth {
...
+      // auth proxy, means we are authenticated and did get
+      // through to the API server but were rejected by RBAC.
+      if (this.is403FromAPIsServer(e)) {
+        return true;
       }
-      // Finally, the k8s api server nowadays defaults to allowing anonymous
-      // requests, so that rather than returning a 401, a 403 is returned if
-      // RBAC does not allow the anonymous user access. An http-only cookie
-      // will not result in an anonymous request, so...
-      return !this.isAnonymous(response);
```
absoludity, line 147: I think this assumption that I'd made a year ago is actually incorrect, in that the K8s API server will never receive an anonymous request when the auth proxy is being used, since the auth proxy itself redirects to login when unauthenticated. Hence simply removing this condition. It would be required if this function was called when not using the auth proxy, but this is exactly what I fixed (in the LoginForm) in the previous PR, so that it's only used when `config.authProxyEnabled` is true.

dashboard/src/shared/Auth.ts
```diff
@@ -90,6 +89,16 @@ export class Auth {
     }
   }
 
+  // is403FromAPIsServer returns true if the response is a 403 determined to have originated
+  // from the grpc-web APIs server, rather than the auth proxy.
+  public static is403FromAPIsServer(e: any): boolean {
+    const contentType = e.metadata?.headersMap["content-type"] as string[];
+    if (contentType.some(v => v === "application/grpc-web+proto")) {
```
castelblanque, line 96: is it always the case that if the exception contains this content type is a 403?

castelblanque (approved):

dashboard/src/shared/Auth.ts
antgamdia: Just FTR, the grpc-gw fn to determine which is a grpcwerb request is: 

```go
func (w *WrappedGrpcServer) IsGrpcWebRequest(req *http.Request) bool {
	return req.Method == http.MethodPost && strings.HasPrefix(req.Header.Get("content-type"), "application/grpc-web")
}
```

so  perhaps, instead of `===` we could check the prefix to cover all the possible cases of the [grpc-web protocol](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md#protocol-differences-vs-grpc-over-http2), namely:

- application/grpc-web+proto
- application/grpc-web+json
- application/grpc-web+thrift
- application/grpc-web-text+proto
- application/grpc-web-text+thrift

absoludity: It's not an assumption that if the exception contains this content type, that it is a 403 error. The function itself (`is403FromAPIsServer`) assumes that the argument *is* a 403 already, so is checking the content type of the given 403 error. Though it doesn't need to assume that, I suppose?

Maybe that's what you mean? We could (and perhaps should) rename this function as `isErrorFromAPIsServer()`, since the fact that it's currently being used to check 403 errors only isn't relevant to the function itself? I'll update to that.

absoludity: > so perhaps, instead of `===` we could check the prefix to cover all the possible cases of the [grpc-web protocol](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md#protocol-differences-vs-grpc-over-http2), 

Yep, makes sense, will update to that. Thanks.

castelblanque: Yeah, that was my question. Thanks for the name change!

Issue #3848, opened 2021-11-29
Reduce same-purpose Go dependencies 


antgamdia:
### Description:

While upgrading our deps in the past, I noticed we are using a bunch of dependencies for the same purpose. Unless there is an actual reason for doing so, I think we should keep our dependencies as tight as possible.

#### Duplicated YAML deps:

- github.com/ghodss/yaml v1.0.0
- gopkg.in/yaml.v2 v2.4.0
- sigs.k8s.io/yaml v1.3.0
- k8s.io/apimachinery/pkg/util/yaml

#### Duplicated logging deps:

- k8s.io/klog/v2 v2.30.0
- github.com/pkg/errors v0.9.1
- github.com/sirupsen/logrus v1.8.1

#### Dupplicated deps

- github.com/Masterminds/semver v1.5.0
- github.com/Masterminds/semver/v3 v3.1.1

absoludity:
+1 for the more obvious ones (semver v3, yaml). With the logging ones, +1 to use klog/v2, but how is the errors package related to logging? (It should be superseded by the standard errors pkg now, I tihnk?)

antgamdia:
WRT Duplicated logging deps, citing this [Michael's comment](https://github.com/kubeapps/kubeapps/discussions/4087#discussioncomment-2029801) for the record:

> We can (and should) update in the kubeapps apisserver to using klog.InfoS and klog.ErrorS which are drop-in replacements, then users can choose to add the --logging-format option to the deployment (or better, we enable from the chart). At least that would have us enabling structured logging in the main service into the future.

antgamdia:
Solved via:

- [ ] https://github.com/kubeapps/kubeapps/pull/4177
- [ ] https://github.com/kubeapps/kubeapps/pull/4178
- [ ] https://github.com/kubeapps/kubeapps/pull/4179
- [ ] https://github.com/kubeapps/kubeapps/pull/4180

Using named import is something still under discussion at https://github.com/kubeapps/kubeapps/pull/4176 (although it doesn't affect to this issue, directly)

PR #4050, opened 2022-01-10
Use another yaml dep to unmarshall the schema

antgamdia:
### Description of the change

Follow-up PR of #4023. 

After giving a try to the repo that the TCE folks so kindly created (https://github.com/vmware-tanzu/community-edition/pull/2775#issuecomment-1006784759)), I noticed it still failed in Kubeapps. The unmarshaller was returning an error even if the `addtionalProperties` prop was properly being set (according to the JSONSchema Spec). 
I guess it has something to do with the `JSONSchemaPropsOrBool` (from k8s) field it tries to unmarshall to, but I don't know the root cause yet.

Anyway, it gets sorted out when using another `yaml` dependency, which in case of failures, it seems to be using the default value in `JSONSchemaPropsOrBool` (which is `true`).

This PR replaces this yaml dep and updates the test cases accordingly.

### Benefits

Kubeapps will be able to set default values for all the TCE packages (and any other with the same issue in the schema)

### Possible drawbacks

Perhaps it can influence #3848.

### Applicable issues

- rel #3853

### Additional information

image

### DIFF
--- a/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_utils.go
+++ b/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_utils.go
@@ -27,11 +27,11 @@ import (
 	datapackagingv1alpha1 "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apiserver/apis/datapackaging/v1alpha1"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+	"gopkg.in/yaml.v3"
 	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
 	structuralschema "k8s.io/apiextensions-apiserver/pkg/apiserver/schema"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
-	"sigs.k8s.io/yaml"
 )
 
 type pkgSemver struct {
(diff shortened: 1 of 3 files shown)


cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_utils.go
```diff
@@ -27,11 +27,11 @@ import (
 	datapackagingv1alpha1 "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apiserver/apis/datapackaging/v1alpha1"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+	"gopkg.in/yaml.v3"
```
castelblanque, line 30: Shouldn't we unify the versions we use?
[Here](https://github.com/kubeapps/kubeapps/blob/master/pkg/agent/docker_secrets_postrenderer.go#L11) we are using v2.

castelblanque:
Thanks! Kubeapps is one step closer to TCE :)

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_utils.go
antgamdia: Absolutely agree, that's indeed the purpose of https://github.com/kubeapps/kubeapps/issues/3848.
We have several duplicated deps we ought to simplify sooner or later. Regarding this `v2` vs `v3` I'd clearly go with `v3` since it allows editing the Yaml parsed nodes directly, meaning one is eventually able to create custom nodes manually (e.g., adding commented-out values)

absoludity (approved):
Excellent!

PR #4051, opened 2022-01-10
Add switch to enable/disable operators support

### LINKED ISSUE #3252: Add switch to enable/disable operators support
The Operator support which was added to Kubeapps was done so rather urgently and has the Dashboard communicating directly with the K8s API server.

We're currently removing all call-sites in the dashboard which talk directly to the K8s API server and instead replacing them with package-specific plugins.

We need to ensure that Kubeapps can be run without operator support (which will be the default) so that the frontend is not proxying requests to the k8s API server.  As we will want CI to be running our normal tests without any proxying to the K8s API server, the normal CI tests will run with operator support disabled (the default). We need to evaluate whether we should also run a separate CI job with the operator support to ensure it stays working.

We should also evaluate how much effort would be involved to replace the Operator functionality in the dashboard with a packaging plugin, together with the community use of the operator support.

castelblanque:
### Description of the change

Adds a flag in the Kubeapps Helm chart to control if Operators support must be enabled or disabled.
By default is set to disabled. Hopefully with this changes + the huge job that @absoludity has done (#3896), Kubeapps shouldn't be directly calling to K8s APIs.

Changes here prevent both dashboard and Nginx from interacting with Operators. But still the OLM framework can live and function in the cluster as usual, of course.

### Benefits

Operators can be ignored from the Kubeapps perspective, and no direct calls to K8s apis are done.

### Possible drawbacks

Existing Kubeapps installation that work with Operators will need to set the new flag to true when updating in order to keep working. Breaking change?

### Applicable issues

- fixes #3252

### Additional information

N/A


chart/kubeapps/templates/frontend/configmap.yaml
```diff
@@ -65,6 +65,16 @@ data:
       # used with an auth-proxy where k8s credentials never leave
       # the cluster). See clusters option.
     {{- range .Values.clusters }}
+      
+    {{- if not $.Values.featureFlags.operators }}
+
+      # Operators support disabled
+      location ~* /api/clusters/{{ .name }}/.*operators.coreos.com.* {
```
castelblanque, line 72: With this, should be enough for Nginx to avoid calls to URLs related to operators mgmt.

dashboard/src/actions/kube.tsx
```diff
@@ -67,10 +68,17 @@ export type KubeAction = ActionType<typeof allActions[number]>;
...
-  return async dispatch => {
+  return async (dispatch, getStore) => {
     dispatch(requestResourceKinds());
     try {
-      const groups = await Kube.getAPIGroups(cluster);
+      let groups = await Kube.getAPIGroups(cluster);
+      // Ignore APIs for operators if specified
+      const operatorsEnabled = getStore().config.featureFlags?.operators === true;
+      if (!operatorsEnabled) {
+        groups = groups.filter(
```
castelblanque, line 78: When operators disabled and OLM is existing in the cluster, groups are still containing stuff related to operators.

dashboard/src/containers/RoutesContainer/Routes.tsx
```diff
@@ -47,8 +61,10 @@ const privateRoutes = {
     OperatorInstanceViewContainer,
   "/c/:cluster/ns/:namespace/operators-instances/:csv/:crd/:instanceName/update":
     OperatorInstanceUpdateContainer,
-  "/c/:cluster/ns/:namespace/config/repos": AppRepoList,
-  "/docs": ApiDocs,
+} as const;
+
+const unsupportedRoutes = {
+  "/c/:cluster/ns/:namespace/operators*": "Operators support has been disabled for Kubeapps.",
```
absoludity, line 67: Perhaps: "Operators support has been disabled by default for Kubeapps. Enable it in your values.yaml" (or something better, that sounds awkward, but something communicating that it can be enabled).

dashboard/src/containers/RoutesContainer/Routes.tsx
```diff
@@ -73,6 +90,14 @@ class Routes extends React.Component<IRoutesProps> {
         {Object.entries(privateRoutes).map(([route, component]) => (
           <PrivateRouteContainer key={route} exact={true} path={route} component={component} />
         ))}
+        {this.props.featureFlags?.operators &&
+          Object.entries(operatorsRoutes).map(([route, component]) => (
+            <PrivateRouteContainer key={route} exact={true} path={route} component={component} />
+          ))}
+        {!this.props.featureFlags?.operators &&
+          Object.entries(unsupportedRoutes).map(([route]) => (
```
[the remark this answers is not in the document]
absoludity: Nice!

absoludity (approved):
Great, thanks Rafa! A few thoughts inline, but up to you - see what you think.

script/e2e-test.sh
```diff
@@ -198,6 +198,7 @@ installOrUpgradeKubeapps() {
     --set apprepository.initialRepos[0].url=http://chartmuseum-chartmuseum.kubeapps:8080
     --set apprepository.initialRepos[0].basicAuth.user=admin
     --set apprepository.initialRepos[0].basicAuth.password=password
+    --set featureFlags.operators=true
```
absoludity, line 201: We will probably need to have a switch (env var) for CI so that we can run CI with or without the operator support and tests, but doesn't need to be done here and now, just noting.

absoludity (commented):
FWIW, I've gone ahead and done two of the small points I've mentioned here:
- updated the ci script so that operators can be enabled or disabled via env,
- no fetching of resources happens at all from the dashboard if operators are not enabled
in my own PR at #4054, so I'd be able to test running on CI without the api proxying in the frontend.

Feel free to land this PR here as is and I'll update my PR to target master, or fine if you want to change anything here (I'll just wait until your PR lands before I take my PR out of draft).

chart/kubeapps/templates/frontend/configmap.yaml
castelblanque: Exactly, this piece should be temporary until we get rid of the direct calls to k8s API.

dashboard/src/actions/kube.tsx
castelblanque: Didn't try to not calling `getResourceKinds`, as I assumed if it was there, it was for some reason other than operators.
I'm totally fine with giving it a try and removing it. Is everything related to resources plugin already in `master` or should I create an issue to track this later?

script/e2e-test.sh
castelblanque: Absolutely, by now we will have CI running with operators to check that we keep on supporting it 😄

castelblanque: Wow, only now I see your comments below, forget my comment above this one 😃

dashboard/src/components/Header/Menu.test.tsx
```diff
@@ -35,15 +38,39 @@ afterEach(() => {
...
+  const state = deepClone(initialState) as IStoreState;
+  state.config.featureFlags = { operators: false };
+  const store = getStore(state);
+  const wrapper = mountWrapper(store, <Menu {...defaultProps} />);
+  expect(wrapper.find(".dropdown")).not.toHaveClassName("open");
+  const menu = wrapper.find("button");
+  menu.simulate("click");
+  wrapper.update();
+  expect(wrapper.find(".dropdown")).toHaveClassName("open");
+  // It render links for AppRepositories and operators
```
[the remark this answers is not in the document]
castelblanque: Well, actually, when operators are enabled, it renders both links. Will rewrite it, though.

dashboard/src/actions/kube.tsx
castelblanque: I believe you already tackled this [here](https://github.com/kubeapps/kubeapps/pull/4051#pullrequestreview-848600122)?

PR #4056, opened 2022-01-11
DRY up use of errorByStatus in two plugins.

absoludity:
### Description of the change

Follow-up PR to DRY up the use of `errorByStatus` in two plugins.

### Benefits

Only one implementation.

### Applicable issues

- ref #3896


castelblanque:
Thanks for removing duplicate code!

castelblanque (approved):

Issue #3849, opened 2021-11-29
Remaining work for running Kubeapps with Carvel


antgamdia:
### Description:

After https://github.com/kubeapps/kubeapps/issues/3221 we got initial and limited support for managing Carvel bundles in Kubeapps
However, there is still a (growing) bunch of pending things before we are able to announce "Carvel support in Kubeapps".

_Whereas trivial tasks will be tracked here (marked checkbox means done), others requiring further effort or discussions are split out to separated issues_ 

#### UI improvements

- [x] Check everything works fine working in different namespaces
   1. Case 1: available package in A; install it in A; available package in A; install it in B
   2. Case 2: get installed packages in A; get installed packages in every namespace
- [x] #3852
- [x] #3850 
- [x] https://github.com/kubeapps/kubeapps/pull/4138
- [x] https://github.com/kubeapps/kubeapps/pull/4144

#### UX improvements

- [x] #3853
  - [x]  https://github.com/kubeapps/kubeapps/pull/4050
- [x] https://github.com/kubeapps/kubeapps/pull/4100

#### Backend improvements

- [x] https://github.com/kubeapps/kubeapps/pull/4068
- [x] https://github.com/kubeapps/kubeapps/pull/4060
- [x] https://github.com/kubeapps/kubeapps/pull/4067
  - [x] https://github.com/kubeapps/kubeapps/pull/4085
- [x] https://github.com/kubeapps/kubeapps/pull/4059
- [x] https://github.com/kubeapps/kubeapps/pull/4057
- [x] https://github.com/kubeapps/kubeapps/pull/4074
- [x] https://github.com/kubeapps/kubeapps/pull/4139
- [x] https://github.com/kubeapps/kubeapps/pull/4165
- [x] https://github.com/kubeapps/kubeapps/pull/4192
- [x] https://github.com/kubeapps/kubeapps/pull/4222

#### Response time issues
- [x] https://github.com/kubeapps/kubeapps/pull/4069

See also #4099

#### Docs/CI improvements

- [x] https://github.com/kubeapps/kubeapps/pull/4127

See also #4093

#### Open questions

1. Think about error handling: aggressive early error return vs ignoring errors (ie: a single failure in a package should break/stop the whole core API call?)
2. Should we default to globalCluster if none is specified? Or just throw an error?
3. Should we implement multicluster support here?

antgamdia:
Some items have been extracted apart to separate issues, namely:

[...]

PR #4060, opened 2022-01-11
Use the package's namespace instead of the requested one in carvel plugin

antgamdia:
### Description of the change

This small PR just uses the package's namespace instead of the requested one for retrieving the rest of the data for building the  `GetInstalledPackageSummaries` response. 

### Benefits

The "allNs" toggle will work for carvel packages.

### Possible drawbacks

N/A

### Applicable issues

- rel #3849

### Additional information

image


castelblanque:
Great, thanks!

castelblanque (approved):

absoludity (approved):
Nice! (having the all namespaces option, that is). Might be worth a unit-test here to stop an accidental reversion or otherwise make it more maintainable.

PR #4062, opened 2022-01-11
3854 get resources

### LINKED ISSUE #3854: Add a proper implementation for

antgamdia:
### Description of the change

This PR removes the workaround we implemented when we released it last time and implements a generic way to retrieve all the kubernetes resources matching a certain label. This way we can fetch every k8s resource created as a result of a Carvel PackageInstall.

### Benefits

The retrieved resources for a carvel package will now include every possible one, not a fixed list of pods, svc and deployments.

### Possible drawbacks

Hitting the k8s APIs for querying every possible available resource is expensive in time. I guess it is also due to my Kubelet config in kind with limited resources, but I'm still worried about it. 

* Example: an app with 97 resources (in a cluster with  88 different resource types) takes around 15 seconds to retrieve the resource refs.
  * With plenty of ` Waited for <some time> ms due to client-side throttling, not priority and fairness, request: GET:https://10.96.0.1:443/apis/<api resource>?labelSelector=kapp.k14s.io%2Fapp%3D<my-id>` console messages 

Note that the `kapp` CLI is already is able to fetch its generated resources; however, solely using a Kubernetes client (not the `kapp` binary), there is no CR holding this information. 
See how the `kapp inspect` command under the hood here: https://github.com/vmware-tanzu/carvel-kapp/blob/v0.44.0/pkg/kapp/cmd/app/inspect.go#L76

### Applicable issues

- fixes #3854
- requires https://github.com/kubeapps/kubeapps/pull/4057 for the UI to be fully working.

### Additional information

Some resources from a Habor installation: 
image

Created as a draft as the tests are still misisng.

### DIFF
--- a/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server.go
+++ b/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server.go
@@ -16,13 +16,20 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/kubeapps/kubeapps/cmd/apprepository-controller/pkg/client/clientset/versioned/scheme"
 	"github.com/kubeapps/kubeapps/cmd/kubeapps-apis/core"
 	corev1 "github.com/kubeapps/kubeapps/cmd/kubeapps-apis/gen/core/packages/v1alpha1"
 	"github.com/kubeapps/kubeapps/cmd/kubeapps-apis/gen/plugins/kapp_controller/packages/v1alpha1"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/restmapper"
 )
 
 type clientGetter func(context.Context, string) (kubernetes.Interface, dynamic.Interface, error)
@@ -43,11 +50,48 @@ type Server struct {
 	clientGetter             clientGetter
 	globalPackagingNamespace string
 	globalPackagingCluster   string
+
+	// We keep a restmapper to cache discovery of REST mappings from GVK->GVR.
+	restMapper meta.RESTMapper
+
+	// kindToResource is a function to convert a GVK to GVR with
+	// namespace/cluster scope information. Can be replaced in tests with a
+	// dummy version using the unsafe helpers while the real implementation
+	// queries the k8s API for a REST mapper.
+	kindToResource func(meta.RESTMapper, schema.GroupVersionKind) (schema.GroupVersionResource, meta.RESTScopeName, error)
+}
+
+// createRESTMapper returns a rest mapper configured with the APIs of the
+// local k8s API server. This is used to convert between the GroupVersionKinds
+// of the resource references to the GroupVersionResource used by the API server.
+func createRESTMapper() (meta.RESTMapper, error) {
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		return nil, err
+	}
+	// To use the config with RESTClientFor, extra fields are required.
+	// See https://github.com/kubernetes/client-go/issues/657#issuecomment-842960258
+	config.GroupVersion = &schema.GroupVersion{Group: "", Version: "v1"}
+	config.NegotiatedSerializer = serializer.WithoutConversionCodecFactory{CodecFactory: scheme.Codecs}
+	client, err := rest.RESTClientFor(config)
+	if err != nil {
+		return nil, err
+	}
+	discoveryClient := discovery.NewDiscoveryClient(client)
+	groupResources, err := restmapper.GetAPIGroupResources(discoveryClient)
+	if err != nil {
+		return nil, err
+	}
+	return restmapper.NewDiscoveryRESTMapper(groupResources), nil
 }
 
 // NewServer returns a Server automatically configured with a function to obtain
 // the k8s client config.
 func NewServer(configGetter core.KubernetesConfigGetter, globalPackagingCluster string) *Server {
+	mapper, err := createRESTMapper()
+	if err != nil {
+		return nil
+	}
 	return &Server{
 		clientGetter: func(ctx context.Context, cluster string) (kubernetes.Interface, dynamic.Interface, error) {
 			if configGetter == nil {
@@ -69,6 +113,14 @@ func NewServer(configGetter core.KubernetesConfigGetter, globalPackagingCluster 
 		},
 		globalPackagingNamespace: globalPackagingNamespace,
 		globalPackagingCluster:   globalPackagingCluster,
+		restMapper:               mapper,
+		kindToResource: func(mapper meta.RESTMapper, gvk schema.GroupVersionKind) (schema.GroupVersionResource, meta.RESTScopeName, error) {
+			mapping, err := mapper.RESTMapping(gvk.GroupKind())
+			if err != nil {
+				return schema.GroupVersionResource{}, "", err
+			}
+			return mapping.Resource, mapping.Scope.Name(), nil
+		},
 	}
 }
 
--- a/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_resources.go
+++ b/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_resources.go
@@ -14,15 +14,22 @@ package main
 
 import (
 	"context"
+	"fmt"
+	"sync"
 
+	corev1 "github.com/kubeapps/kubeapps/cmd/kubeapps-apis/gen/core/packages/v1alpha1"
 	kappctrlv1alpha1 "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/kappctrl/v1alpha1"
 	packagingv1alpha1 "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/packaging/v1alpha1"
 	datapackagingv1alpha1 "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apiserver/apis/datapackaging/v1alpha1"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/dynamic"
+	log "k8s.io/klog/v2"
 )
 
 const (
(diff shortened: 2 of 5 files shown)


cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server.go
```diff
@@ -43,43 +50,95 @@ type Server struct {
...
 			if err != nil {
-				return nil, nil, status.Errorf(codes.FailedPrecondition, fmt.Sprintf("unable to get typed client : %v", err))
+				return nil, nil, nil, status.Errorf(codes.FailedPrecondition, fmt.Sprintf("unable to get typed client : %v", err))
 			}
-			return typedClient, dynamicClient, nil
+			// TODO(agamez): this call may be expensive and subject to be cached;
+			// have a look at the CachedDiscoveryClient in k8s
+			// https://github.com/kubernetes/client-go/blob/release-1.22/discovery/cached/disk/cached_discovery.go
+			discoveryClient := typedClient.Discovery()
+			if discoveryClient == nil {
```
antgamdia, line 116: Adding the `discoveryClient` as part of the clientGetter so that we can swap it and mock it later for testing

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server.go
```diff
@@ -43,43 +50,95 @@ type Server struct {
...
+	}
+	// To use the config with RESTClientFor, extra fields are required.
+	// See https://github.com/kubernetes/client-go/issues/657#issuecomment-842960258
+	config.GroupVersion = &schema.GroupVersion{Group: "", Version: "v1"}
+	config.NegotiatedSerializer = serializer.WithoutConversionCodecFactory{CodecFactory: scheme.Codecs}
+	client, err := rest.RESTClientFor(config)
+	if err != nil {
+		return nil, err
+	}
+	discoveryClient := discovery.NewDiscoveryClient(client)
```
antgamdia, line 80: Here we still create a new discovery, but as Michael did in the past, we are using the RESTMapper just in production. For testing, we mock the entire `kindToResource` function

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_resources.go
```diff
@@ -399,3 +406,117 @@ func (s *Server) updatePkgInstall(ctx context.Context, cluster, namespace string
...
+			if len(resource.Verbs) == 0 {
+				continue
+			}
+			// filter to resources that support the specified "List" verb
+			if !sets.NewString(resource.Verbs...).HasAll([]string{"list"}...) {
+				continue
+			}
+			wg.Add(1)
+			// spawn a goroutine to fetch the matching k8s resources
+			go func(resource metav1.APIResource, ch chan resourceAndGvk, wg *sync.WaitGroup) {
```
antgamdia, line 478: I've tried to make it as parallel as possible, but the bottleneck seems to be in the k8s apiserver itself (in my local environment, haven't tested in a real cluster yet).

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_resources.go
```diff
@@ -399,3 +406,117 @@ func (s *Server) updatePkgInstall(ctx context.Context, cluster, namespace string
...
+
+	// fetch every possible kubernetes resource list that is available in the cluster
+	apiResourceLists, err := discoveryClient.ServerPreferredResources()
+	if err != nil {
+		return nil, err
+	}
+
+	// create channel to collect the matching k8s resources thus avoid blocking the waitgroup
+	// ussing a buffered channel with size 100; assumption: 60% native k8s resources + 40 CRDs
+	ch := make(chan resourceAndGvk, 100)
```
antgamdia, line 454: I'm not used to channels yet, I've made this assumption, but please feel free to tell me more about the buffer sizes here if it is wrong.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_resources.go
```diff
@@ -399,3 +406,117 @@ func (s *Server) updatePkgInstall(ctx context.Context, cluster, namespace string
...
+		}
+
+	}
+	// wait every goroutine to finish and close the channel
+	go func() {
+		wg.Wait()
+		close(ch)
+	}()
+
+	// once closed, iterate over the closed channel and push the objects to the response
```
absoludity, line 512: I don't think this comment is correct? That is, the channel will be closed when the `wg.Wait()` returns, but that's correctly happening inside a go routine. So the code below will start processing the data on the channel as it arrives (as it should) and will only complete (the `range ch`) when the channel is closed and all data has been drained.

Which still leaves me wondering why you need a large buffer on your channel? The work of the consumer here should be very quick compared to the producer (which is doing network fetches), so I'd not think you'd need a buffer at all.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_resources.go
absoludity: Yes, unless I'm missing something, this code is calling `List` on *every* listable resource in all namespaces on the cluster? I'm skeptical that we should be doing this?

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_resources.go
```diff
@@ -399,3 +406,117 @@ func (s *Server) updatePkgInstall(ctx context.Context, cluster, namespace string
...
+			go func(resource metav1.APIResource, ch chan resourceAndGvk, wg *sync.WaitGroup) {
+				// convert the gvk to a gvr
+				gvk := gv.WithKind(resource.Kind)
+				gvr, _, err := s.kindToResource(s.restMapper, gvk)
+				if err != nil {
+					ch <- resourceAndGvk{nil, err}
+				}
+
+				// ignore the namespace as the resources may appear in different ns from which the PackageInstall is created
+				resources, err := dynamicClient.Resource(gvr).List(ctx, listOptions)
```
absoludity, line 487: Interesting, I'd thought the above syntax was for non-namespaced (ie. cluster-level) resources, rather than resources across all namespaces, but see that the dynamic client apparently supports both (`NamespaceableResourceInterface` which is [implemented to do both](https://github.com/kubernetes/client-go/blob/v0.23.1/dynamic/dynamiclister/lister.go))

absoludity (commented):
I've gone through the channel code that you mentioned you were uncertain about and left a few thoughts, as well as taken a look through the rest of the code. My main question or concern is that this appears to be getting *all* resources across all namespaces of the cluster to filter them by whatever the kapp selector is. If that's the case, then yep, I think your concern about the timing is a valid one. Is that the only option we've got?

I had a brief look at the kapp code that you pointed to, I didn't think it appeared to be doing so, but rather doing something much more specific (it appears to be listing identified resources only, but could be reading it incorrectly). Then I read the docs and tried the following on my dev cluster:

```
kapp inspect -a 'label:'
...
878 resources
```

Wow. So it's possible to use kapp to do that, but is that what it's doing by default when you inspect an app? (Haven't checked but assume you have).

Either way, what I'd hope we can do here is use kapp itself to get the resources for an app, so we're not implementing our own version. Why can we not create a `NewInspectionCommand` and run it, similar to how we run Helm commands? That way, we'd not be re-implementing kapp functionality (listing all resources for an app), in the same way we try to run the Helm CLI commands where we can?

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_resources.go
```diff
@@ -399,3 +406,117 @@ func (s *Server) updatePkgInstall(ctx context.Context, cluster, namespace string
...
+			continue
+		}
+
+		// for each kubernetes resource, spwan a goroutine to check if it matches the listOptions
+		for _, resource := range apiResourceList.APIResources {
+			if len(resource.Verbs) == 0 {
+				continue
+			}
+			// filter to resources that support the specified "List" verb
+			if !sets.NewString(resource.Verbs...).HasAll([]string{"list"}...) {
```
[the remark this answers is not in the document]
antgamdia: Yep, I just used the same code as in the kubectl client, which is aimed at supporting any set of verbs. In our scenario, your suggestion is way simpler. Thanks for pointing it out

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_resources.go
```diff
@@ -399,3 +406,117 @@ func (s *Server) updatePkgInstall(ctx context.Context, cluster, namespace string
...
+	ch := make(chan resourceAndGvk, 100)
+	wg := sync.WaitGroup{}
+
+	// for each kubernetes resource list, filter out those not being having the verb "list"
+	for _, apiResourceList := range apiResourceLists {
+		if len(apiResourceList.APIResources) == 0 {
+			continue
+		}
+		gv, err := schema.ParseGroupVersion(apiResourceList.GroupVersion)
+		if err != nil {
```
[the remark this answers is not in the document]
antgamdia: Idem as in other comment, just using the k8s code directly (https://github.com/kubernetes/kubectl/blob/release-1.22/pkg/cmd/apiresources/apiresources.go#L172) :S

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_resources.go
```diff
@@ -399,3 +406,117 @@ func (s *Server) updatePkgInstall(ctx context.Context, cluster, namespace string
...
 	return &pkgInstall, nil
 }
+
+// appLabelIdentifier returns the app label identifier for the given Kapp app
+//
+// Apparently, App CRs not being created by a "kapp deploy" command don't have the proper annotations.
+// So, in order to retrieve the annotation value,
+// we have to get the ConfigMap <AppName>-ctrl and, then, fetch the value of the key "labelValue" in data.spec.
+// See https://kubernetes.slack.com/archives/CH8KCCKA5/p1637842398026700
+// https://github.com/vmware-tanzu/carvel-kapp-controller/issues/430
```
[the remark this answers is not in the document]
antgamdia: I don't really know. The issue is not very active and I haven't heard about it on the community slack recently.
As of today (v 0.44 got released), it still needs the `-ctrl` suffix.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server.go
```diff
@@ -43,43 +50,95 @@ type Server struct {
...
+				return schema.GroupVersionResource{}, "", err
+			}
+			return mapping.Resource, mapping.Scope.Name(), nil
+		},
 	}
 }
 
 // GetClients ensures a client getter is available and uses it to return both a typed and dynamic k8s client.
-func (s *Server) GetClients(ctx context.Context, cluster string) (kubernetes.Interface, dynamic.Interface, error) {
+func (s *Server) GetClients(ctx context.Context, cluster string) (kubernetes.Interface, dynamic.Interface, discovery.DiscoveryInterface, error) {
```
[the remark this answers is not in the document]
antgamdia: Mmm, that's true... we can just use the `typedClient` and later on, call `typedClient.Discovery()`. However, I wasn't able to swap it by the `fake.FakeDiscovery` in the tests (that's why https://github.com/kubeapps/kubeapps/pull/4062/commits/c7004f89faefc00859489458b567b26acceab233).
If we go this way (ie not using the carvel code), I'll give it another thought anyway

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_resources.go
antgamdia: Mmm, yes, it's exactly what the code is doing :S
However, I can't think of another way of "getting an unknown number of resources matching a label in an unknown number of different namespaces". Perhaps by banning some resources we certainly know they can't be created (like `Nodes` ?)

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_resources.go
antgamdia: Ah, right! I mixed it up...  initially, I had an unbuffered channel, so I needed the goroutine encapsulating the `wg.wait()` and `close(ch)`. Then I moved to a buffered one, but left this goroutine back.

Yes.. the consumer is just aggregating data, which should be really fast; not really needing a buffer here.

antgamdia:
> Either way, what I'd hope we can do here is use kapp itself to get the resources for an app, so we're not implementing our own version. Why can we not create a `NewInspectionCommand` and run it, similar to how we run Helm commands? That way, we'd not be re-implementing kapp functionality (listing all resources for an app), in the same way we try to run the Helm CLI commands where we can?

Quick notes, I'm in a rush; will try to explain myself better later:

TLDR: I don't see any performance increase when using Kapp's code directly (I'd say it's slower)

- I have successfully used the Kapp code as in the Helm scenario. 
- There is no way to configure it with a custom rest.Config for generating the actions. You have to copy some parts (creating the clients: NewResourceTypesImpl, NewResourcesImpl, NewIdentifiedResources, NewApps)
- Then using the same code as in the cmd `inspect` action 

Perhaps it's a matter of our Kubernetes client default config; the QPS and other params. I'll have a look later.

Also, if finally going this way, we should create an issue for the carvel team to allow configuring the client (or even a PR, shouldn't be that difficult)

absoludity:
Excellent, nice work Antonio! I'm +1 whichever way you decide, though I personally lean the same way (use Kapp and get upstream fixed so it can be configured, so we don't maintain more code unnecessarily).

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_resources.go
absoludity: Yeah, OK. If this is what the kapp code does as well, I assume they've thought about it more than us (and would prefer to just use their code, thanks for the second version).

absoludity:
> Excellent, nice work Antonio! I'm +1 whichever way you decide, though I personally lean the same way (use Kapp and get upstream fixed so it can be configured, so we don't maintain more code unnecessarily).

Looks like you haven't yet pushed your changes here (no rush, I just wanted to leave a +1 after seeing those, but will do so tonight / tomorrow)

antgamdia:
> Looks like you haven't yet pushed your changes here (no rush, I just wanted to leave a +1 after seeing those, but will do so tonight / tomorrow)


Not yet, I wanted to discuss both alternatives; that's why I pushed the changes in a separate PR (https://github.com/kubeapps/kubeapps/pull/4068). 

The current situation is: my approach is fully working (with the QPS fix) but the alternative is not yet (not getting resources from other namespaces yet). 

My plan: working today on having the carvel approach fully working and tested and benchmark both alternatives.

antgamdia:
Closing this PR in favor of https://github.com/kubeapps/kubeapps/pull/4068.
Thanks for your comments and feedback on this version!

PR #4065, opened 2022-01-12
Update check in e2e test to use resources api.

absoludity:

### Description of the change

This is a prequel PR to fix the issue blocking CI in #4054.

Specifically, it updates the 02-create-private-registry test so that it no longer tries to hit the k8s API to verify that the created secret was used with the deployment.

Instead, it uses the resources API to fetch the same deployment data.

There was a bit of a yack-shave, in that it turned out that I was incorrectly using an `Any` proto field for the `GetInstalledPackageResourceRefsResponse`, which meant that the http/gateway endpoint didn't work for this URL (complaining that the Any field didn't have a TypeURL). An Any field should only ever be used for any *message*, not any data. So instead, the manifest field is now more correctly a string field. This required a small update to the dashboard also.

### Benefits

#4054 can land, which will complete #3896

### Applicable issues

- ref #3896


castelblanque:
Nice to see that there was only a single point in CI tests (apart from operators maybe?) where K8s API was hit!

castelblanque:
Looking forward to getting rid of generated files. I believe it would remove noise in the dev process 😄

castelblanque (approved):

PR #4068, opened 2022-01-12
Retrieve every k8s resource generated by a Carvel package installation (alt)

antgamdia:
See main PR https://github.com/kubeapps/kubeapps/pull/4062

An alternative approach using code from Kapp


cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_resources.go
```diff
@@ -399,3 +401,55 @@ func (s *Server) updatePkgInstall(ctx context.Context, cluster, namespace string
 	}
 	return &pkgInstall, nil
 }
+
+// inspectKappK8sResources returns the list of k8s resources matching the given listOptions
+func (s *Server) inspectKappK8sResources(ctx context.Context, cluster, namespace, packageId string) ([]*corev1.ResourceRef, error) {
```
antgamdia, line 406: This is mostly the code that kapp CLI itself uses for inspecting an app, but adapted to our needs.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_test.go
```diff
@@ -3004,6 +3006,207 @@ func TestDeleteInstalledPackage(t *testing.T) {
...
+
+			s := Server{
+				clientGetter: func(context.Context, string) (kubernetes.Interface, dynamic.Interface, error) {
+					return typedClient, dynClient, nil
+				},
+				kappClientsGetter: func(ctx context.Context, cluster, namespace string) (ctlapp.Apps, ctlres.IdentifiedResources, *kappcmdapp.FailingAPIServicesPolicy, ctlres.ResourceFilter, error) {
+					// Create a fake DepsFactory and pass it the fake k8s clients
+					depsFactory := NewFakeDepsFactoryImpl()
+					depsFactory.SetCoreClient(typedClient)
+					depsFactory.SetDynamicClient(dynClient)
```
antgamdia, line 3170: This is the approach I've been able to come up with for faking the kapp internal clients. Even if it _kinda_ works, it returns an empty list of resources found :(
If you have any suggestions or can give it a try, I'd really appreciate it!

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server.go
```diff
@@ -69,6 +78,50 @@ func NewServer(configGetter core.KubernetesConfigGetter, globalPackagingCluster
...
+				return ctlapp.Apps{}, ctlres.IdentifiedResources{}, nil, ctlres.ResourceFilter{}, status.Errorf(codes.Internal, "configGetter arg required")
+			}
+			// Retrieve the k8s REST client from the configGetter
+			config, err := configGetter(ctx, cluster)
+			if err != nil {
+				return ctlapp.Apps{}, ctlres.IdentifiedResources{}, nil, ctlres.ResourceFilter{}, status.Errorf(codes.FailedPrecondition, "unable to get config due to: %v", err)
+			}
+			// Pass the REST client to the (custom) kapp factory
+			configFactory := NewConfigurableConfigFactoryImpl()
+			configFactory.ConfigureRESTConfig(config)
```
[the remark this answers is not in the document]
absoludity: Seems like a simple and useful change, so not sure why not.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server.go
```diff
@@ -69,6 +78,50 @@ func NewServer(configGetter core.KubernetesConfigGetter, globalPackagingCluster
...
+			supportingNsObjs, err := kappcmdapp.FactoryClients(depsFactory, kappcmdcore.NamespaceFlags{Name: namespace}, resourceTypesFlags, logger.NewNoopLogger())
+			if err != nil {
+				return ctlapp.Apps{}, ctlres.IdentifiedResources{}, nil, ctlres.ResourceFilter{}, status.Errorf(codes.FailedPrecondition, "unable to get config due to: %v", err)
+			}
+
+			// Getting non-namespaced clients (e.g., for fetching every k8s object in the cluster)
+			supportingObjs, err := kappcmdapp.FactoryClients(depsFactory, kappcmdcore.NamespaceFlags{Name: ""}, resourceTypesFlags, logger.NewNoopLogger())
+			if err != nil {
+				return ctlapp.Apps{}, ctlres.IdentifiedResources{}, nil, ctlres.ResourceFilter{}, status.Errorf(codes.FailedPrecondition, "unable to get config due to: %v", err)
+			}
```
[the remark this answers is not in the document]
absoludity: Sorry, I'm missing some info here? Does the harbor package create resources in the target namespace (of the package install) as well as another namespace? And why can't we see the client that the Kapp CLI creates to do the same command?

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_resources.go
```diff
@@ -399,3 +401,55 @@ func (s *Server) updatePkgInstall(ctx context.Context, cluster, namespace string
...
+	for _, resource := range resources {
+		refs = append(refs, &corev1.ResourceRef{
+			ApiVersion: resource.GroupVersion().String(),
+			Kind:       resource.Kind(),
+			Name:       resource.Name(),
+			Namespace:  resource.Namespace(),
+		})
+	}
+	return refs, nil
+}
```
absoludity, line 455: Excellent, thanks for the *very* clear code in this part :)

absoludity (approved):
+1ing in case I don't get back to it today, as it looks great. Hopefully I'll get to look at the issue for the tests later too.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server.go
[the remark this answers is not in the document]
antgamdia: Yep, sorry, let me explain myself better.

When we install a package in Kubeapps in a certain namespace (e.g., Harbor in `default` ns), we are creating: a `PackageInstall`  in `default` and a `ConfigMap` in `default`. 
Additionally, kapp, under the scenes, is gonna create an `App` in `default`. Neither the `PackageInstall` nor the `App` has references to anything different from the namespace `default`.

However, when reconciling, this `App` starts creating the resources, which happen to be a namespace `harbor` and a set of several resources that are also created in this `harbor` namespace.

Perhaps there's something I missed in the Kapp CLI code. In fact, they have a `--dangerous-scope-to-fallback-allowed-namespaces` flag, so I guess, for the name it has, they are not allowing querying everything.
If only I had a reference to this `harbor` namespace somewhere, I'd be able to query the current `default` ns as well as `harbor` as this kind of "fallback allowed namespace".

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_test.go
absoludity: So I setup the vscode debugger (which was pretty simple - hadn't used it much before) and traced this down from your call to `	resources, err := resourcesClient.List(labelSelector, nil)` in `server_data_resources.go`. From there:

- kapp first tries to [list all the resource types](https://github.com/vmware-tanzu/carvel-kapp/blob/v0.44.0/pkg/kapp/resources/identified_resources_list.go#L17) on the cluster, which, for your test client, results in an empty slice, so that
- kapp [goes on to call `resources.All(resTypes, allOpts)`](https://github.com/vmware-tanzu/carvel-kapp/blob/v0.44.0/pkg/kapp/resources/identified_resources_list.go#L45), but this just iterates the slice of `resTypes`, so also returns an empty slice.

Looking at the implementation, the core issue is the [function which collects the resource types](https://github.com/vmware-tanzu/carvel-kapp/blob/v0.44.0/pkg/kapp/resources/resource_types.go#L111) is using the discovery api, so in your test it's not returning any APIs. Might be worth re-checking the kapp test code to see how they setup the discovery API in tests or otherwise ensure certain resource types (pods, in your case) are returned.

HTH!

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_test.go
```diff
@@ -3004,6 +3009,200 @@ func TestDeleteInstalledPackage(t *testing.T) {
...
+						{Name: "pods", Namespaced: true, Kind: "Pod", Verbs: []string{"list", "get"}},
+					},
+				},
+			}
+
+			typedClient := typfake.NewSimpleClientset(tc.existingTypedObjects...)
+
+			// We cast the dynamic client to a fake client, so we can set the response
+			fakeDiscovery, _ := typedClient.Discovery().(*disfake.FakeDiscovery)
+			fakeDiscovery.Fake.Resources = apiResources
```
antgamdia, line 3144: After creating a bunch of custom types... it just came down to this approach.
The crux was just in casting the Discovery() result so that we can manually leverage from the builtin `.Fake.Resources`

Thanks for the pointer, Michael. Best solution when you're stuck is have some rest and start over the next day with a fresh mind :P

PR #4069, opened 2022-01-12
Allow customizing QPS/Burst in the kubeapps-api service

antgamdia:
### Description of the change

When talking to the k8s API it is critical to configure the rate limit accordingly rather than relying on the default (5 requests per second, AFAIK). This PR allows changing the QPS/Burst values as we already did in Kubeops a long time ago.

### Benefits

More customization.

### Possible drawbacks

N/A

### Applicable issues

- rel https://github.com/kubeapps/kubeapps/issues/3849

### Additional information

Required by https://github.com/kubeapps/kubeapps/pull/4062 to work decently (setting qps to ~100 is enough)


chart/kubeapps/values.yaml
```diff
@@ -1658,6 +1660,12 @@ kubeappsapis:
   ## @param kubeappsapis.replicaCount Number of frontend replicas to deploy
   ##
   replicaCount: 2
+  ## @param kubeappsapis.qps KubeappsAPIs Kubernetes API client QPS limit
+  ##
+  qps: "10.0"
```
absoludity, line 1665: Out of interest, I notice that for kubeops, the chart defaults are "" (ie. not set by the chart) with the kubeops binary setting defaults of 10 and 15 when the args aren't set, but here for kubeapps apis you're explicitly setting the defaults here?

absoludity (approved):
> When talking to the k8s API it is critical to configure the rate limit accordingly rather than relying on the default (5 requests per second, AFAIK)

Heh, yes, 5 per second is a sensible default for a human client serving one keyboard, not a service :) Good find.

> Required by #4062 to work decently (setting qps to ~100 is enough)

Hmm, why default to 10.0 then?

chart/kubeapps/values.yaml
antgamdia: Exactly,  that's precisely the intention: in kubeops, we were modifying the default rate limit values directly in the code, so users wouldn't notice the actual value being used.

If we set the values in the values.yaml, it will appear directly in the readme, so users will immediately know which is the default value without digging into the code.


> Hmm, why default to 10.0 then?

In this PR I'm just using the same values as we were defining in Kubeops. The proper value for the kubeappsapis service to work decently is yet to be analyzed and defined (100 was just a random feasible value, but we have to give it a thought)

PR #4079, opened 2022-01-14
Move operator integration test to its own directory and test separately.

absoludity:

### Description of the change

I wanted to ensure that we are testing Kubeapps in the default setup (ie. without any k8s API proxying) while also ensuring we still have the confidence of CI for basic operator support. Initially I was thinking we'd need to either run two jobs (not great, given the failures), but then realised we can simply run the main tests first, then update Kubeapps with the operator support and run the operator test at the end.

This enables the best of both worlds, and we can still switch off the operator test if we like (when a decision is made).

### Benefits

All the main CI tests for Kubeapps run without any access to the k8s API.

### Possible drawbacks

None that I can see.

### Applicable issues

- ref #3896


script/e2e-test.sh
```diff
@@ -432,12 +410,39 @@ admin_token="$(kubectl get -n kubeapps secret "$(kubectl get -n kubeapps service
...
+## Upgrade and run operator test
+# Operators are not supported in GKE 1.14 and flaky in 1.15, skipping test
+if [[ -z "${GKE_BRANCH-}" ]] && [[ -n "${TEST_OPERATORS-}" ]]; then
+  installOLM "${OLM_VERSION}"
+
+  # Update Kubeapps settings to enable operators and hence proxying
+  # to k8s API server.
+  info "Installing latest Kubeapps chart available"
+  installOrUpgradeKubeapps bitnami/kubeapps \
+    "--set" "featureFlags.operators=true"
```
castelblanque, line 431: That is a brilliant idea Michael!
Kubeapps resources are recreated to enable operators without the need to deploy the whole cluster again.

script/e2e-test.sh
```diff
@@ -432,12 +410,39 @@ admin_token="$(kubectl get -n kubeapps secret "$(kubectl get -n kubeapps service
...
   warn "PODS status on failure"
   kubectl cp "${pod}:/app/reports" ./reports
   exit 1
 fi
-info "Integration tests succeeded!!"
+info "Main integration tests succeeded!!"
+
+## Upgrade and run operator test
+# Operators are not supported in GKE 1.14 and flaky in 1.15, skipping test
+if [[ -z "${GKE_BRANCH-}" ]] && [[ -n "${TEST_OPERATORS-}" ]]; then
```
antgamdia, line 424: Just wondering if skipping this test in GKE is still required (as now we are in more recent versions). Anyway, perhaps worths a separate PR not to be mixing up CI potentially breaking changes :P

antgamdia (approved):

Issue #3723, opened 2021-11-04
User-friendly error handling for failed kubeapps-apis server rpc calls


absoludity:
### Description:

In #3385, @antgamdia said:

> The error messages that are displayed to the client is not the best. We are just aggregating the error chain like: "there is an error: error while doing X: error calling Y: error whatever", instead of "error: the user is not allowed to perform the action X". This is not a matter of adding a logging interceptor, but just making the top-level error messages (those returned to the users) a little bit clearer.

I'm splitting the error handling out from #3385 so that it can focus on the logging.

absoludity:
Regarding the errors from the grpc service, I don't think the string grpc-error-message should ever be used to convey a message to users. For now, the UI or client should be using the status code only to generate a simple error, such as "Unable to create an installed package: you are not allowed to create an installed package in the namespace 'foo'. " for a `codes.PERMISSION_DENIED`.

The actual grpc error message should be detailed - it's an error from a remote procedure call - just not exposed in the UX (though it shouldn't include stack traces, which I think I may have seen :P )

Longer-term we may want to look at using [google.rpc.status](https://stackoverflow.com/a/48750825) or otherwise [encoding a more specific enum into error responses](https://www.vinsguru.com/grpc-error-handling/).

satya-dillikar:
I have written a sample example using googleapis/rpc/errdetails [here](https://play.golang.org/p/nP8FRF4eIzu). Maybe something like this we might use.

absoludity:
Thanks Satya. Yes, I think using errdetails is an excellent idea for providing richer errors for our API, but I don't (personally) think we should mix the UI user error messages into the API errors. Those details can help the dashboard client decide whether to say simply "Unable to create installed package: there was a permission error when creating the package", or "Unable to create installed package: an internal error occurred", but that should be the client's decision not our API servers (IMO). But happy to hear/discuss otherwise.

satya-dillikar:
I agree with you, we don't want to mix UI and API server errors. 
Anything which API-servers don't like may create errdetails,  if that is helpful. Otherwise, we can continue with the existing error wrapping mechanism. 

So, do you think we have the issue to fix in UI/ dashboard?

absoludity:
Sorry Satya, I'm not sure what you mean with the last comment? I'll try to catch you on your Monday afternoon to check about this and other items, if you haven't gotten the input you need by then.

absoludity:
[...]

PR #4084, opened 2022-01-14
Error alert component

castelblanque:

### Description of the change

Introduces a React `ErrorAlert` component aimed at improving readability of UI errors and that could replace existing `<Alert>` usages. 

It is able to turn this kind of UI error:

image

into this:

image

It can detect RPC errors and show them indented while not affecting existing plain text errors.
The change also allows to specify a stack of errors when creating a custom error (or an extension of it).

### Benefits

Better error readability on UI

### Possible drawbacks

None that I can think of.

### DIFF
--- a/dashboard/src/actions/apps.ts
+++ b/dashboard/src/actions/apps.ts
@@ -159,7 +159,13 @@ export function fetchApps(
       dispatch(receiveAppList(installedPackageSummaries));
       return installedPackageSummaries;
     } catch (e: any) {
-      dispatch(errorApp(new FetchError(e.message)));
+      dispatch(
+        errorApp(
+          e instanceof Error
+            ? new FetchError("Unable to list apps", [e])
+            : new FetchError("Unable to list apps: " + e.message),
+        ),
+      );
       return [];
     }
   };
--- a/dashboard/src/components/AppList/AppList.tsx
+++ b/dashboard/src/components/AppList/AppList.tsx
@@ -2,7 +2,7 @@ import { CdsButton } from "@cds/react/button";
 import { CdsIcon } from "@cds/react/icon";
 import { CdsToggle, CdsToggleGroup } from "@cds/react/toggle";
 import actions from "actions";
-import Alert from "components/js/Alert";
+import ErrorAlert from "components/ErrorAlert";
 import LoadingWrapper from "components/LoadingWrapper/LoadingWrapper";
 import { push } from "connected-react-router";
 import qs from "qs";
@@ -131,7 +131,7 @@ function AppList() {
         className="margin-t-xl"
       >
         {error ? (
-          <Alert theme="danger">Unable to list apps: {error.message}</Alert>
+          <ErrorAlert>{error}</ErrorAlert>
         ) : (
           <AppListGrid
             appList={listOverview}
--- a/dashboard/src/components/ErrorAlert/ErrorAlert.tsx
+++ b/dashboard/src/components/ErrorAlert/ErrorAlert.tsx
@@ -5,7 +5,7 @@ import { CustomError } from "shared/types";
 import "./ErrorAlert.css";
 
 export interface IErrorAlert {
-  children: CustomError | string;
+  children: CustomError | Error | string;
 }
 
 function createWrap(message: any, index: number, indented: boolean): JSX.Element {
@@ -34,6 +34,8 @@ export default function ErrorAlert({ children }: IErrorAlert) {
     if (children.causes) {
       messages.push(buildMessages(children.causes));
     }
+  } else if (children instanceof Error) {
+    messages = [createWrap(children.message, 0, false)];
   } else {
     messages = [children];
   }
--- a/dashboard/src/components/ErrorAlert/ErrorAlert.tsx
+++ b/dashboard/src/components/ErrorAlert/ErrorAlert.tsx
@@ -1,0 +1,41 @@
+import Alert from "components/js/Alert";
+import RpcErrorMessage from "components/RpcErrorMessage";
+import { RpcError } from "shared/RpcError";
+import { CustomError } from "shared/types";
+import "./ErrorAlert.css";
+
+export interface IErrorAlert {
+  children: CustomError | string;
+}
+
+function createWrap(message: any, index: number, indented: boolean): JSX.Element {
+  return (
+    <div className={indented ? "error-alert-indent" : "error-alert"} key={index}>
+      {message}
+    </div>
+  );
+}
+
+function buildMessages(errors: Error[]): JSX.Element[] {
+  return errors.map((cause, index) => {
+    if (cause instanceof RpcError) {
+      return createWrap(<RpcErrorMessage>{cause}</RpcErrorMessage>, index + 1, true);
+    } else {
+      return createWrap(cause.message, index + 1, true);
+    }
+  });
+}
+
+// Extension of Alert component for showing more meaningful Errors
+export default function ErrorAlert({ children }: IErrorAlert) {
+  let messages: any[];
+  if (children instanceof CustomError) {
+    messages = [createWrap(children.message, 0, false)];
+    if (children.causes) {
+      messages.push(buildMessages(children.causes));
+    }
+  } else {
+    messages = [children];
+  }
+  return <Alert theme="danger">{messages}</Alert>;
+}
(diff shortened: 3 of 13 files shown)


dashboard/src/actions/apps.ts
```diff
@@ -159,7 +159,13 @@ export function fetchApps(
       dispatch(receiveAppList(installedPackageSummaries));
       return installedPackageSummaries;
     } catch (e: any) {
-      dispatch(errorApp(new FetchError(e.message)));
+      dispatch(
+        errorApp(
+          e instanceof Error
+            ? new FetchError("Unable to list apps", [e])
```
castelblanque, line 165: Only place where the new constructor for errors is used by now.

dashboard/src/components/AppList/AppList.tsx
```diff
@@ -131,7 +131,7 @@ function AppList() {
         className="margin-t-xl"
       >
         {error ? (
-          <Alert theme="danger">Unable to list apps: {error.message}</Alert>
+          <ErrorAlert>{error}</ErrorAlert>
```
castelblanque, line 134: Only usage of the new `ErrorAlert` component by now.

dashboard/src/shared/RpcError.ts
```diff
@@ -0,0 +1,24 @@
+export class RpcError extends Error {
+  // Following https://github.com/grpc/grpc-go/blob/master/internal/status/status.go#L140
+  static RPC_ERROR_REGEX = /rpc error: code = (\w+) desc = (.*)/;
```
castelblanque, line 3: Until we can have GRPC code to properly throw a dedicated error object, we can only guess if it is an rpc error by checking the error message string.

dashboard/src/shared/types.ts
```diff
@@ -15,13 +15,28 @@ import { IAuthState } from "../reducers/auth";
...
+import { RpcError } from "./RpcError";
 
-class CustomError extends Error {
+export class CustomError extends Error {
+  public causes: Error[] | undefined;
   // The constructor is defined so we can later on compare the returned object
-  // via err.contructor  == FOO
-  constructor(message?: string) {
+  // via err.constructor  == FOO
+  constructor(message?: string, causes?: Error[]) {
```
castelblanque, line 24: Allows to optionally specify a stack of errors.

dashboard/src/components/ErrorAlert/ErrorAlert.tsx
```diff
@@ -0,0 +1,43 @@
+import Alert from "components/js/Alert";
+import RpcErrorMessage from "components/RpcErrorMessage";
+import { RpcError } from "shared/RpcError";
+import { CustomError } from "shared/types";
+import "./ErrorAlert.css";
+
+export interface IErrorAlert {
+  children: CustomError | Error | string;
```
castelblanque, line 8: Component supports rendering this three types.

castelblanque:
This can be a complimentary work to #3723

dashboard/src/components/AppList/AppList.tsx
antgamdia: Great, just to confirm: you mean there are other errors out there (like `there is a problem with this package...`) which are not being handled by this novel `ErrorAlert`, don't you? 
So this is just a kind of proof of concept, no?

dashboard/src/components/ErrorAlert/ErrorAlert.tsx
antgamdia: What if there are no children?

dashboard/src/components/ErrorAlert/ErrorAlert.tsx
```diff
@@ -0,0 +1,43 @@
...
+  let messages: any[];
+  if (children instanceof CustomError) {
+    messages = [createWrap(children.message, 0, false)];
+    if (children.causes) {
+      messages.push(buildMessages(children.causes));
+    }
+  } else if (children instanceof Error) {
+    messages = [createWrap(children.message, 0, false)];
+  } else {
+    messages = [children];
```
antgamdia, line 40: As above, what if we have no response from the other side of the wire. For example, the API server being OOMkilled by k8s and the gRPC call just returns the nginx error 502 page ? - meaning no explicit error message being sent.

dashboard/src/components/RpcErrorMessage/RpcErrorMessage.scss
```diff
@@ -0,0 +1,9 @@
+.error-alert-rpc {
+  ul > li {
+    font-weight: bolder;
```
antgamdia, line 3: We are using explicit numbers in the rest of the CSS files, usually 400, 500 or 600. See what you think

dashboard/src/components/RpcErrorMessage/RpcErrorMessage.scss
```diff
@@ -0,0 +1,9 @@
...
+
+    span.rpc-value {
+      font-weight: normal;
```
antgamdia, line 6: Idem here

dashboard/src/shared/RpcError.ts
```diff
@@ -0,0 +1,24 @@
+export class RpcError extends Error {
```
antgamdia, line 1: I wonder if we should rename `Rpc` to `GRpc` instead

dashboard/src/shared/RpcError.ts
```diff
@@ -0,0 +1,24 @@
...
+    Object.setPrototypeOf(this, new.target.prototype);
+    [this.code, this.desc, this.message] = this.extractData(error.message);
+  }
+  public static isRpcError(error: Error) {
+    return error.message.trim().indexOf(this.RPC_ERROR_CHECK) > -1;
+  }
+  private extractData(message: string): [string, string, string] {
+    const found = message.trim().match(RpcError.RPC_ERROR_REGEX);
+    return found
+      ? [found[1], found[2], message.substring(0, message.indexOf(found[0]))]
```
antgamdia, line 21: I'd rather check found.length to avoid undesired undefined

antgamdia (approved):
Great. It seems a great UI improvement; I only have some minor comments and questions but in general LGTM! 
Thank you!

dashboard/src/components/AppList/AppList.tsx
castelblanque: Exactly, I first wanted the community to review it. If accepted, we can start using it from now on and eventually replace existing `<Alert>` with it.

dashboard/src/components/ErrorAlert/ErrorAlert.tsx
castelblanque: Children is a mandatory item it seems. In this case doesn't make sense to show the error block without any content, does it?

dashboard/src/components/ErrorAlert/ErrorAlert.tsx
castelblanque: Same as commented above, it does not make sense to have the error block with empty data. For the error codes, it will show exactly the same as current `<Alert>` component, [just the passed children.](https://github.com/kubeapps/kubeapps/blob/master/dashboard/src/components/js/Alert/Alert.js#L56)

dashboard/src/components/RpcErrorMessage/RpcErrorMessage.scss
castelblanque: Sure, will change it to be aligned 😃

dashboard/src/shared/RpcError.ts
castelblanque: Yeah, I thought about it, but error message states `rpc error:` so I went ahead with a generic `rpc` naming for everything.

dashboard/src/shared/RpcError.ts
castelblanque: Added tests for empty strings and undefined and it seems ok.

castelblanque: I was curious, so check the `match` return for empty strings. It is undefined, for a positive case, the data is in the array. So we are good.

dashboard/src/shared/RpcError.ts
absoludity: You should be able to check the grpc error code? At least, the gRPC dashboard code *should* be throwing an error object with a `.code` property. That's what we're using here already: https://github.com/kubeapps/kubeapps/blob/master/dashboard/src/shared/Auth.ts#L71-L88

castelblanque: Thanks, will definitely take a look. It would be simpler.

PR #4085, opened 2022-01-14
Configurable auto-upgrades policy for Carvel pkgs

antgamdia:
### Description of the change

Following up https://github.com/kubeapps/kubeapps/pull/4067#discussion_r784688310, this PR adds a new `upgradePolicy: none|patch|minor|major` field in the plugin config section in the chart to set the default constraint being generated when installing a Carvel package.

### Benefits

The `LatestMatchingVersion` field will now notify of any update matching this constraint.

### Possible drawbacks

Kapp will auto-update the package to the latest version satisfying the constraint. 

### Applicable issues

- rel https://github.com/kubeapps/kubeapps/issues/3849
- rel https://github.com/kubeapps/kubeapps/pull/4067#discussion_r784688310

### Additional information

N/A


chart/kubeapps/README.md
```diff
@@ -491,6 +491,7 @@ Once you have installed Kubeapps follow the [Getting Started Guide](https://gith
 | `kubeappsapis.pluginConfig.core.packages.v1alpha1.versionsInSummary.minor` | Number of minor versions to display in the summary                                        | `3`                      |
 | `kubeappsapis.pluginConfig.core.packages.v1alpha1.versionsInSummary.patch` | Number of patch versions to display in the summary                                        | `3`                      |
 | `kubeappsapis.pluginConfig.core.packages.v1alpha1.timeoutSeconds`          | Value to wait for Kubernetes commands to complete                                         | `300`                    |
+| `kubeappsapis.pluginConfig.kappController.packages.v1alpha1.upgradePolicy` | Default upgrade policy generating version constraints ()                                  | `patch`                  |
```
castelblanque, line 494: Should we reference this in Carvel documentation created in #4127 ?

antgamdia: Absolutely yes, I think so. However, I was waiting to confirm this is the chosen solution before writing about it, that's why I haven't added it yet.

castelblanque (approved):

chart/kubeapps/values.yaml
```diff
@@ -1632,6 +1632,12 @@ kubeappsapis:
             patch: 3
           ## @param kubeappsapis.pluginConfig.core.packages.v1alpha1.timeoutSeconds Value to wait for Kubernetes commands to complete
           timeoutSeconds: 300
+    kappController:
+      packages:
+        v1alpha1:
+          ## @param kubeappsapis.pluginConfig.kappController.packages.v1alpha1.upgradePolicy Default upgrade policy generating version constraints
+          ## enum: [ "major", "minor", "patch", "none" ]
+          upgradePolicy: patch
```
absoludity, line 1640: Given that we'll (at some point) enable people to set the upgrade policy via the UX, can we name this `defaultUpgradePolicy`? (seems to be how it's being used below already anyway).

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
```diff
@@ -612,8 +612,18 @@ func (s *Server) UpdateInstalledPackage(ctx context.Context, request *corev1.Upd
...
+	if err != nil {
+		return nil, err
+	}
+
 	// Update the rest of the fields
-	pkgInstall.Spec.PackageRef.VersionSelection = &vendirVersions.VersionSelectionSemver{Constraints: pkgVersion}
+	pkgInstall.Spec.PackageRef.VersionSelection = &vendirversions.VersionSelectionSemver{
+		Constraints: versionConstraints,
+		// https://github.com/vmware-tanzu/carvel-kapp-controller/issues/116
+		// This is to allow prereleases to be also installed
```
absoludity, line 624: I'm not certain from the comment exactly what the decision is here? The issue has been closed (fixed almost a year ago), but from the description seems to only be relevant *if* you want to include prereleases in your version selection? Also, see the later https://github.com/vmware-tanzu/carvel-kapp-controller/issues/300 . So wouldn't that mean that we're making that decision here to include pre-releases, as opposed to not doing so by default?

Looks to me like another config option for `defaultIncludePrereleases` or similar?

absoludity (approved):

chart/kubeapps/README.md
antgamdia: Mmm... true, perhaps this behavior is too distant from what an average kubeapps user would expect. Defaulting it back to `none`. Thanks!

BTW, I've added some docs here now the other PR got merged

chart/kubeapps/values.yaml
antgamdia: Totally right! Naming is often the most difficult part. I've renamed it to`defaultUpgradePolicy`.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
antgamdia: Right... I missed this pending task when defining the list of pending actions. I initially add this `VersionSelectionSemverPrereleases{}` to explicitly allow prereleases by default, which is the opposite behavior in kapp-ctrl, just to ensure we supported everything (it's easier to notice something is there when it shouldn't than the other way round :P )

Since it's a simple and similar change as the one I did here, let me merge this PR and create a follow-up one straight away.
Thanks for pointing this out!

antgamdia: Done: https://github.com/kubeapps/kubeapps/pull/4165

Issue #4097, opened 2022-01-17
Add a "filter by plugin" in the Catalog view


antgamdia:
If Kubeapps is configured with multiple plugins, a user may want to filter by the plugin type (eg, "list all the carvel packages").

Adding a new filter in the Catalog view (ie, just frontend) should be enough. Long-term, we might consider adding backed filters to the core API to query just a subset of the configured plugins.

absoludity:
> Adding a new filter in the Catalog view (ie, just frontend) should be enough. Long-term, we might consider adding backed filters to the core API to query just a subset of the configured plugins.

Why wouldn't we do the backend filter from the start? Seems odd to add functionality to the frontend that is doing it in a way that we'd want to later remove, with no gain other than looking like it's already done (when it's not)? I usually push back on this kind of thing because it gives the impression that we're further ahead than we are which skews expectations. But if you think it's necessary, OK.

ppbaena:
Close as it is duplicated in #3889

PR #4100, opened 2022-01-17
Add a "filter by plugin" in the catalog view

antgamdia:
### Description of the change

This quick PR just passes the same `enabledPlugins` array to the UI so that it can be aware of the current config (in deploy-time, not run-time though). Next, a simple Catalog filter using each plugin's name has been added.

### Benefits

Users will be able to filter out packages based upon the plugin name.

### Possible drawbacks

This simplistic approach is not aware of: i) the plugin's purpose (e.g., "packages API" vs "resources" or "repositories"); ii) the configuration in runtime (e.g., a plugin that failed and is not fully loaded/registered properly).

Nonetheless, i) has been mitigated ignoring every plugin name returning `unknown...` and ii) can't happen as of today (a failed plugin will result in an error in the whole API server).

### Applicable issues

- fixes https://github.com/kubeapps/kubeapps/issues/4097

### Additional information

https://user-images.githubusercontent.com/11535726/149788139-6275648a-da5d-4afd-8f31-e862e2dd78fc.mp4

PS: the lack of this filter was really annoying me when testing a specific plugin, that's why I decided to implement this STAT :sweat_smile:


castelblanque:
Awesome! It is great to have this filter. Wouldn't it be faster to do the filtering in the backend?

castelblanque (approved):

absoludity:
> PS: the lack of this filter was really annoying me when testing a specific plugin, that's why I decided to implement this STAT sweat_smile

OK, if it's needed immediately AND you think that adding the filtering to the backend from the start would be a significant amount of effort. Otherwise, as I mentioned on the issue, I'd be -1 to adding tech-debt like this.

antgamdia:
> OK, if it's needed immediately AND you think that adding the filtering to the backend from the start would be a significant amount of effort. Otherwise, as I mentioned on the issue, I'd be -1 to adding tech-debt like this.

No, it's not really "urgent", so to speak. In fact, it was something I quickly squeezed at the beginning of the iteration for avoiding typing "carvel" in the search box again and again (or even disabling the helm plugin).

I'm sure we can just use the `GetConfiguredPlugins` endpoint for populating the react state and use the same filter; shouldn't be a big deal.

+1 not to merge this PR, and can add it to the "next interaction discussion".

absoludity:
Closing this PR for now and I'll move the related issue back into the next iteration discussion as you mentioned above.

PR #4121, opened 2022-01-19
Upgrade dependencies

antgamdia:
### Description of the change

As part of the upcoming release, this PR is aimed at performing a general update on the dependencies we use.

### Benefits

Up to date deps.

### Possible drawbacks

Some unnoticed breaking changes? Let our beloved CI speak up.

### Applicable issues

N/A

### Additional information

If you, locally, face some weird issues when running the `yarn test` command, just run `yarn test --clearCache` and then `yarn test` 

This PR depends on https://github.com/kubeapps/kubeapps/pull/4137

PR #4137, opened 2022-01-20
Use react-router-dom when possible

antgamdia:
### Description of the change

When upgrading some deps at https://github.com/kubeapps/kubeapps/pull/4121, I noticed there were a few errors related to react-router.

This PR is replacing _some_ imports from `react-router` to the library we extensively use across the project `react-router-dom`. Whereas they are both really similar, the latter includes some defaults values for web applications (instead of being generic so that react-native applications can also use the routing features). 

### Benefits

This change makes the tests pass after bump up the versions in https://github.com/kubeapps/kubeapps/pull/4121.

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

Unfortunately, I don't know what is the root cause. Nevertheless, we are using an old version of react-router. In v6 there are plenty of breaking changes. However, the required effort isn't worthwhile atm (and they are still developing some adapters to ease the transitions, so we better keep waiting).


castelblanque:
Excellent, an easy but effective fix. Thanks!

castelblanque (approved):

Issue #4142, opened 2022-01-21
Click any application error: "An error occurred: btoa(...).replaceAll is not a function."


ygqygq2:
kubeapps version: v2.4.2
Like that
![image](https://user-images.githubusercontent.com/17672440/150472493-c9aa6a92-d12f-421e-a908-b8d1a0f31ab4.png)

![image](https://user-images.githubusercontent.com/17672440/150472412-e04dfa45-0a77-434c-8d04-e01ae99212e4.png)

antgamdia:
Thanks for reporting the issue, however, I'm unable to reproduce it locally as I don't really know which are the steps to repro it.
However, it's true the `atob` and `btoa` functions have been deprecated in favor of `Buffer.from`, so, perhaps it's can be related to.

I'm sending a PR addressing this deprecation.

ygqygq2:
Thanks. It’s not just google chrome

![image](https://user-images.githubusercontent.com/17672440/150630093-825f9f69-58a2-4996-940a-be09ca860787.png)

PR #4145, opened 2022-01-21
Use `Buffer.from` instead of deprecated `atob` and `btoa`

antgamdia:
### Description of the change

This PR uses `Buffer.from` instead of deprecated `atob` and `btoa`functions for converting from/to base64.

### Benefits

New browsers not having the currently deprecated `atob` and `btoa` won't error anymore.

### Possible drawbacks

N/A [(as current browsers already have support for `Buffer` functions)](https://caniuse.com/?search=Buffer)

### Applicable issues

- fixes https://github.com/kubeapps/kubeapps/issues/4142


absoludity (approved):
Thanks

PR #4165, opened 2022-01-25
Add `defaultIncludePrereleases`

antgamdia:
### Description of the change

Having considered https://github.com/kubeapps/kubeapps/pull/4085#discussion_r791835168, the purpose of this PR is twofold:

1. Add a `defaultIncludePrereleases` param in the plugin-specific config to decide whether or not we should add the `VersionSelectionSemverPrereleases{}` when creating the installed pkg.
2. Refactor a little bit the way the plugin config is created (mainly creating a type) and improve the tests.
    1. We weren't checking the actual `InstalledPackage` being created in a test case, so this PR also includes some minor changes in the test case

### Benefits

Until we have a way to configure some options from the UI, this PR introduces a way to configure some plugin-specific options, specifically `defaultIncludePrereleases` for allowing installing pre-releases by kapp-controller.

### Possible drawbacks

N/A

### Applicable issues

- rel #3849

### Additional information

N/A


cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_test.go
```diff
@@ -2603,28 +2612,24 @@ func TestCreateInstalledPackage(t *testing.T) {
...
-						UsefulErrorMessage:  "Deployed",
+						ObservedGeneration:  0,
+						Conditions:          nil,
+						FriendlyDescription: "",
+						UsefulErrorMessage:  "",
 					},
-					Version:              "1.2.3",
-					LastAttemptedVersion: "1.2.3",
+					Version:              "",
+					LastAttemptedVersion: "",
```
antgamdia, line 2632: Minor changes in the test cases to match the actual `PackageInstall` being created.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_utils.go
```diff
@@ -318,6 +318,28 @@ func isKindInt(src interface{}) bool {
...
+				V1alpha1 struct {
+					DefaultUpgradePolicy      string `json:"defaultUpgradePolicy"`
+					DefaultIncludePrereleases bool   `json:"defaultIncludePrereleases"`
+				} `json:"v1alpha1"`
+			} `json:"packages"`
+		} `json:"kappController"`
+	}
+	kappControllerPluginParsedConfig struct {
+		defaultUpgradePolicy      upgradePolicy
+		defaultIncludePrereleases bool
```
antgamdia, line 334: Struct holding the plugin's config.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server.go
```diff
@@ -47,55 +49,58 @@ type Server struct {
...
+	err = json.Unmarshal([]byte(pluginConfigFile), &pluginConfig)
 	if err != nil {
-		return none, fmt.Errorf("unable to open plugin config at %q: %w", pluginConfigPath, err)
+		return config, fmt.Errorf("unable to unmarshal pluginconfig: %q error: %w", string(pluginConfigFile), err)
 	}
-	err = json.Unmarshal([]byte(pluginConfig), &config)
+
+	// parse the configuration options
+	defaultUpgradePolicy := upgradePolicyMapping[pluginConfig.KappController.Packages.V1alpha1.DefaultUpgradePolicy]
+	defaultIncludePrereleases, err := strconv.ParseBool(pluginConfig.KappController.Packages.V1alpha1.DefaultUpgradePolicy)
```
castelblanque, line 73: Is `pluginConfig.KappController.Packages.V1alpha1.DefaultUpgradePolicy` the right param to parse for `defaultIncludePrereleases`?

antgamdia: Absolutely right! I didn't notice, thanks!

chart/kubeapps/values.yaml
```diff
@@ -1638,6 +1638,8 @@ kubeappsapis:
           ## @param kubeappsapis.pluginConfig.kappController.packages.v1alpha1.defaultUpgradePolicy Default upgrade policy generating version constraints
           ## enum: [ "major", "minor", "patch", "none" ]
           defaultUpgradePolicy: none
+          ## @param kubeappsapis.pluginConfig.kappController.packages.v1alpha1.defaultIncludePrereleases Default policy for allowing pre-releases to be installed
+          defaultIncludePrereleases: false
```
absoludity, line 1642: Sorry, just a thought: I know I'd suggested a boolean when reviewing your other PR - but it might be just as simple to actually have a value here (ie. rename `defaultPrereleasesVersionSelection` that is empty (nil) by default, but can be set to `{}` or whatever? This way, we don't have to later change this config option to support actually setting version selections for pre releases? I'm not suggesting we implement support for them yet, it'd still effectively be a boolean option, but one that doesn't need to change later.

Fine either way.

absoludity (approved):

chart/kubeapps/values.yaml
antgamdia: No, it's fine. In fact, after re-reading the carvel issues and docs I've started having some doubts/questions, so I asked them at: https://kubernetes.slack.com/archives/CH8KCCKA5/p1643376802571959

In short, `prereleases` can have three values:

1. `prereleases: {}`: allow any prerelease if the version constraint allows it.
2. `prereleases: nil`: disallow any prerelease 
    * Exception: if the version constraint is a prerelease itself:
          1. Current behavior: error, it won't install a prerelease if `prereleases: nil`:
          2. Future behavior: install it, it won't be required to set `prereleases:  {}` if the version constraint is a prerelease
3. `prereleases: {Identifiers: []string{"rc"}}`: allow only prerelease with "rc" as part of the name if the version constraint allows it.

> I'm not suggesting we implement support for them yet, it'd still effectively be a boolean option for now, but one that doesn't need to change later.

It is mostly the same effort, I mean, it is just a matter of appending strings or not. I'll update this PR with the full implementation soon. Thanks for the suggestion, much better this way!

antgamdia: Also, I'd like to add (in a separate PR) another default config like `allowDowngrading` or sth, to add the `packaging.carvel.dev/downgradable: "` annotation to the PackageInstall objs to allow them to be downgraded: https://carvel.dev/kapp-controller/docs/latest/package-consumer-concepts/#downgrading

absoludity: Nice, yes, that'd be useful (as a kubeapps config default).

antgamdia: Done: https://github.com/kubeapps/kubeapps/pull/4192  :)

PR #4170, opened 2022-01-26
Make ci resource-class explicit

antgamdia:
### Description of the change

In order to make explicit the resource class we are using in circleci, this PR simply adds a default param (set to `medium`, the current one), so that we can later override it when necessary.

### Benefits

It will be easier to override any class if needed.

### Possible drawbacks

Let see if CI thinks otherwise...

### Applicable issues

N/A

### Additional information

Resource classes available: https://circleci.com/product/features/resource-classes/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,6 +51,9 @@ parameters:
   GKE_REGULAR_VERSION:
     type: "string"
     default: "1.21"
+  DEFAULT_RESOURCE_CLASS:
+    type: "string"
+    default: "medium"
   IMAGES_TO_PUSH:
     type: "string"
     default: "kubeapps/apprepository-controller kubeapps/dashboard kubeapps/asset-syncer kubeapps/assetsvc kubeapps/kubeops kubeapps/pinniped-proxy kubeapps/kubeapps-apis"
@@ -497,6 +500,7 @@ run_e2e_tests: &run_e2e_tests
 gke_test: &gke_test
   docker:
     - image: circleci/golang:<< pipeline.parameters.GOLANG_VERSION >>
+  resource_class: << pipeline.parameters.DEFAULT_RESOURCE_CLASS >>
   steps:
     - checkout
     - run:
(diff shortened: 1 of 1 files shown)


castelblanque:
Mmm not sure if we need this. Each CI step has different needs: `test_chart_render` runs in 28s whereas e2e tests take ~10m. Given that the free tier has a limited total amount of minutes , we can better balance the consumption of free minutes if we just adjust each step's needs for resources (where needed) instead of harmonizing all to the same machine.
Also, if CircleCi decides to increase the default `resource_class` we would need manual intervention to update it.
I mean, I'm ok with the change, but I would prefer to keep the Yaml simpler 😄

antgamdia:
Yep, I kind of agree as well. I mean, this PR is just intended to be a prequel for the next one in which we just use the custom resource class according to our need to minimize the credit consumption. That is, just making explicit what we currently have nowadays.

Also, there are two types of resource classes, ones for the _docker executors_ (the vast majority of our jobs) and a reduced list for the _machine_ executors (which, for instance, does not have the `medium+`).

There is no actual need for this PR to be merged; I just did it when upgrading the images (in a separate PR) just in case it was useful for what you're doing. Feel free to cherry-pick any changes you like and close the PR :P

Reference: 
https://circleci.com/docs/2.0/configuration-reference/#machine-executor-linux
https://circleci.com/docs/2.0/configuration-reference/#docker-executor

.circleci/config.yml
```diff
@@ -684,7 +699,9 @@ jobs:
           command: |
             ./script/create_release.sh ${CIRCLE_TAG} << pipeline.parameters.KUBEAPPS_REPO >>
   local_e2e_tests:
-    machine: true
```
antgamdia, line 687: It was deprecated as well, they recommend using a specific image, so I've added a param and also included it in the release process docs.

.circleci/config.yml
```diff
@@ -684,7 +699,9 @@ jobs:
...
+    machine:
+      image: << pipeline.parameters.DEFAULT_MACHINE_IMG >>
+    resource_class: << pipeline.parameters.DEFAULT_RESOURCE_CLASS >>
```
antgamdia, line 704: In fact, this resource_class is different, as we only can choose from this list: https://circleci.com/docs/2.0/configuration-reference/#machine-executor-linux

Issue #3695, opened 2021-10-29
Remaining work for running Kubeapps with Flux


absoludity:
### Description:

With #3610 we're able to use flux from the Kubeapps UI with some limitations and missing documentation.

This issue is capturing the remaining work that can be picked up when other dependencies are completed.

- [x] Currently after deploying, the installed package details view does not yet include any of the resources (dependent on #3403) 
- [x] When creating an installed pkg, we try to fetch the resources before flux has created the actual helm release, resulting in a not-found error
- [x] After helm release has been created, but before flux finishes reconciling, we're displaying a plain warning about reconciliation in progress, even though we already have the resources and could be showing the progress chart. #4247 
- [x] After the reconciliation succeeds, the UI doesn't update the app info status, so the upgrade/delete buttons remain disabled. #4256 
- [x] The catalog of available packages for flux2 doesn't currently display the repository #4231
- [x] The status for installed packages on the application tiles is inconsistent (and leads to formatting issues) between flux and carvel. Fixed in #4326  
- [x] There is no documentation for configuring and using Kubeapps with Flux. Fixed with #4317 
- [x] When the catalog (available packages) is empty, the UX should not display "Manage App Repositories" unless either the helm plugin is enabled (for now) or the AppRepository API is implemented (#3496) and supported in the UX. For now it'd be better to display a link to the documentation (above) so that we can (for now) have instructions about manually adding to the catalog when using flux (ie. adding a `HelmRepository`). #4331 
- [x] error if both helm and flux are selected for now
- [x] Update chart so that redis is deployed automatically if flux plugin enabled. Created subtask #4264 to discuss options.
- [x] Update chart so that postgres is only deployed if helm plugin is selected - https://github.com/bitnami/charts/pull/9204
- [x] Trying to deploy with unprivileged service account has unhelpful message. #4333
- [x] When flux cannot deploy helm chart, helmrelease CR remains pending but getresourcerefs eventually 404s as the actual helm release isn't created. Fixed in #4338 
- [ ] Use more user-faced `UserReason` fields (eg. currently we use `InstallFailed`, but we should at least add a space) [added by agamez]

PR #4172, opened 2022-01-27
Don't set explicit targetNamespace for HelmRelease.

absoludity:

### Description of the change

When creating a fluxv2 helm release via Kubeapps, the plugin explicitly sets both the HelmRelease's `metadata.namespace` and the `spec.targetNamespace` field to the same target namespace value. The first is obviously correct and necessary, the second is redundant because Flux automatically defaults to using the HelmRelease's `metadata.namespace` as the target namespace when it is not set.

As a result, [as per the docs](https://fluxcd.io/docs/components/helm/helmreleases/#specification) when the `targetNamespace` is set, the actual name of the helm release (not the HelmRelease) defaults to `[TargetNamespace-]Name`:

```
helm ls -n kubeapps-user-namespace
NAME                                    NAMESPACE               REVISION        UPDATED                                 STATUS  CHART           APP VERSION
kubeapps-user-namespace-test-apache-2   kubeapps-user-namespace 1               2022-01-27 00:23:28.749271732 +0000 UTC failed  apache-9.0.1    2.4.52     

```

So when the UX then calls `GetInstalledPackageDetail`, it fails with:

```
Application not found: Unable to get the resource refs for the package "test-apache-2" using the plugin "fluxv2.packages": rpc error: code = NotFound desc = Unable to find Helm release "test-apache-2" in namespace "kubeapps-user-namespace": release: not found
```
since that code assumes the helm release name is the same as the `HelmRelease` name.

The solution here for now is to simply not set the `targetNamespace` redundantly with the same value as the `HelmRelease.metadata.namespace`, which means that flux itself defaults to using the `HelmRelease.metadata.namespace` for the actual helm release (ie. no change), but importantly, flux creates the helm release with the same name as the flux `HelmRelease`, and the UX just works.

It's not the only solution. Another would be to instead additionally explicitly set the `HelmRelease.Spec.ReleaseName` with the same value used for `HelmRelease.metadata.name`, or, later when we actually need support for helm releases in different target contexts, update the code to get (or derive) the name based on the flux2 rules.

But for now, given that we don't need support, I'm not sure there's a reason to not just use the flux default by removing the redundant `targetNamespace`.

Thoughts?

### Applicable issues

- Ref #3695

### DIFF
--- a/cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/release.go
+++ b/cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/release.go
@@ -520,7 +520,6 @@ func (s *Server) newFluxHelmRelease(chart *models.Chart, targetName types.Namesp
 						},
 					},
 				},
-				"targetNamespace": targetName.Namespace,
 			},
 		},
 	}
--- a/cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/release_test.go
+++ b/cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/release_test.go
@@ -1675,8 +1675,7 @@ var (
 					},
 				},
 			},
-			"interval":        "1m",
-			"targetNamespace": "test",
+			"interval": "1m",
 		},
 	}
 
@@ -1699,8 +1698,7 @@ var (
 					"version": "> 5",
 				},
 			},
-			"interval":        "1m",
-			"targetNamespace": "test",
+			"interval": "1m",
 		},
 	}
 
@@ -1725,7 +1723,6 @@ var (
 			"interval":           "1m0s",
 			"serviceAccountName": "foo",
 			"suspend":            false,
-			"targetNamespace":    "test",
 		},
 	}
 
@@ -1747,8 +1744,7 @@ var (
 					},
 				},
 			},
-			"interval":        "1m",
-			"targetNamespace": "test",
+			"interval": "1m",
 			"values": map[string]interface{}{
 				"ui": map[string]interface{}{"message": "what we do in the shadows"},
 			},


gfichtenholt:
My opinion: I don't like this. 
We already had a discussion where the current behavior was decided 
https://github.com/kubeapps/kubeapps/pull/3640#issuecomment-949315105
Not to rub it in or anything, but it was "obvious" to you then that field should be set.
I feel if the field is there in the spec, we should set it to *avoid* confusion and having the users to go though flux docs too see what happens when the field is missing. 
I kind of want to own this. If there is a bug in the flux plug implementation, like GetInstalledPackageDetail () fails for whatever reason when it should work, file an issue, assign it to me. You're kind of putting the cart before the horse with   "Don't set explicit targetNamespace for HelmRelease".  I'd rather start at the root cause and look at the problem holistically, not at the solution. Thanks

absoludity:
> My opinion: I don't like this.

That's fine, I just tested this to be able to continue in the front end, thought I'd propose it, if you agree, great, if you don't, no issue - I can still continue while you come up with a better fix.

> We already had a discussion where the current behavior was decided [#3640 (comment)](https://github.com/kubeapps/kubeapps/pull/3640#issuecomment-949315105) Not to rub it in or anything, but it was "obvious" to you then that field should be set.

Huh? I think you're misreading or misinterpreting there. I said "So yes, 3 is obvious - the resources should be in the target namespace"... and they should be. I wasn't meaning to imply that we should be redundantly setting the `HelmRelease.spec.targetnamespace` and didn't realise that's what you meant by (3), rather thought you just outlined the 3 namespaces that Flux works with. I wasn't against the idea of setting it either (from memory I'd not realised it would affect the name of the actual helm release).

> I feel if the field is there in the spec, we should set it to _avoid_ confusion and having the users to go though flux docs too see what happens when the field is missing.

That's fine. With that in mind, the other option I mentioned in the description above might fit your expectations: additionally explicitly setting the `ReleaseName` field to avoid the same confusion.

> I kind of want to own this. If there is a bug in the flux plug implementation, like GetInstalledPackageDetail () fails for whatever reason when it should work, file an issue, assign it to me. You're kind of putting the cart before the horse with "Don't set explicit targetNamespace for HelmRelease". I'd rather start at the root cause and look at the problem holistically, not at the solution. Thanks

It was just a few lines implementing one of three suggestions I outlined in the PR description to be able to start a discussion to hear what you think (that's why I asked for you to review). Happy to create an issue instead.

gfichtenholt:
Good, thanks I will get to it as soon as I am finished with what've been in the middle of last 5 days. Very tedious work, and I am keen to switch gears for a while :-)

PR #4175, opened 2022-01-27
First step for  flux plugin: switch to using flux native data types instead of unstructured #4112 

gfichtenholt:
This is the first step toward implementing  flux plugin: switch to using flux native data types instead of unstructured #4112 . In this PR I am using Flux's native data types up until the point I need to interact with k8s API server, which is done via `dynamic.Interface` so requires me to convert to/from unstructured. This should bring flux plug-in up-to parity with kapp-controller plugin in terms of handling native types and conversion to/from unstructured.

Next step for me will be getting rid of the dependency on` dynamic.Interface` probably in favor of  https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.11.0/pkg/client#Client so I don't have to convert back and forth between unstructured and typed and just pass the typed objects all the way through the entire stack. 
I am a little exhausted at the moment. This first step was very arduous and pain-staking, so I would like to take a little break, switch gears for a little while and work on other outstanding issues before coming back to do the next step

I am not sure who to ask to review this. Antonio can do this, since he's done similar work in kapp-controlller plugin. Michael is welcome to review as well. I always look forward to his feedback

PR #4178, opened 2022-01-27
Migrate from go yaml.v2 to yaml.v3

antgamdia:
### Description of the change

This PR simply updates a piece of code using the yaml version 2 instead of v3. Since it has some breaking changes, it is a separate PR.

### Benefits

Avoid multiple dependency versions in our codebase.

### Possible drawbacks

Some additional and unnoticed breaking changes? Let's see what the CI thinks.

### Applicable issues

- rel #3848

### Additional information

This is one of the few places we are using the go yaml library instead of the k8s one. In this piece of code, we don't have any typed model to marshal from/to, so I guess that's the reason.
However, I wonder about this decision, I mean, why not rely on a PodSpec since it is a well-established k8s API (corev1)?
Happy to discuss it offline in our standup, though.

### DIFF
--- a/pkg/agent/docker_secrets_postrenderer.go
+++ b/pkg/agent/docker_secrets_postrenderer.go
@@ -11,7 +11,9 @@ import (
 
 	"github.com/distribution/distribution/reference"
 	log "github.com/sirupsen/logrus"
-	"gopkg.in/yaml.v2"
+
+	// The usual "sigs.k8s.io/yaml" is not used because we're dealing with unstructured yaml directly
+	yaml "gopkg.in/yaml.v3"
 )
 
 const (
@@ -57,7 +59,7 @@ func NewDockerSecretsPostRenderer(secrets map[string]string) (*DockerSecretsPost
 
 func (r *DockerSecretsPostRenderer) processResourceList(resourceList []interface{}) {
 	for _, resourceItem := range resourceList {
-		resource, ok := resourceItem.(map[interface{}]interface{})
+		resource, ok := resourceItem.(map[string]interface{})
 		if !ok {
 			continue
 		}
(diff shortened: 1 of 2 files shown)


absoludity:
> This is one of the few places we are using the go yaml library instead of the k8s one. In this piece of code, we don't have any typed model to marshal from/to, so I guess that's the reason.
> However, I wonder about this decision, I mean, why not rely on a PodSpec since it is a well-established k8s API (corev1)?

I'm not 100% about "the" reason (it's been a while), but a post renderer takes a byte array of the generated manifests and returns a byte array of the mutated manifests. These are arbitrary yaml manifests, not specific pod specs or similar. But if there's a way to parse that with a typed parser and pull out just the pod specs, that'd be excellent. Perhaps a parser that just parses the standard metadata, then you can just deal with the know (typed) resources that you're interested in.

> Happy to discuss it offline in our standup, though.

+1

absoludity (approved):

antgamdia:
> But if there's a way to parse that with a typed parser and pull out just the pod specs, that'd be excellent. Perhaps a parser that just parses the standard metadata, then you can just deal with the know (typed) resources that you're interested in.

Thanks for the explanation!

As I mentioned in https://github.com/kubeapps/kubeapps/pull/4175#discussion_r794376862, I wonder if using the built-in decoder in k8s `UniversalDecoder()` + `DecodeInto(...)`  [1] would be useful here, haven't tried though.

[1] https://stackoverflow.com/a/63338697

absoludity:
> > But if there's a way to parse that with a typed parser and pull out just the pod specs, that'd be excellent. Perhaps a parser that just parses the standard metadata, then you can just deal with the know (typed) resources that you're interested in.
> 
> Thanks for the explanation!
> 
> As I mentioned in [#4175 (comment)](https://github.com/kubeapps/kubeapps/pull/4175#discussion_r794376862), I wonder if using the built-in decoder in k8s `UniversalDecoder()` + `DecodeInto(...)` [1] would be useful here, haven't tried though.
> 
> [1] https://stackoverflow.com/a/63338697

Hmm, maybe? Haven't played with it, but `DecodeInto` is the typed version, which assumes you know the type which the post-processor doesn't. There is also there `Object.Decode`, but then you're back with a non-typed object. Universal decoder might be removed, but could help - not sure. But yep, happy if there's a better way to do so with these.

PR #4176, opened 2022-01-27
Unify import aliases

antgamdia:
### Description of the change

First of all, apologies in advance for this series of endless PRs I'm sending this iteration :smile:  Except for some clean-up PRs once we finally deprecate asstesvc, kubeops, etc, I'd bet this is one of the last huge-PR.

While working on https://github.com/kubeapps/kubeapps/issues/3848, I noticed there were several dependencies being imported with no rationale underlying, leading to situations like: 1) a dep being imported multiple times but with different names; 2) inconsistent names for the same dependency across the codebase.

I tried to look for widespread good practices, but, to the best of my knowledge, there isn't any. Later on,  I came across the [code autogenerated by k8s for CRD controllers](https://github.com/kubernetes/sample-controller/blob/master/pkg/generated/clientset/versioned/clientset.go) and noticed they were adding an alias regardless of the actual dependency name (ie, no matter if it's the same).

So I've decided to go this way, and that's what this PR is for: adding an alias to every dependency in our codebase.

### Benefits

Even if the IDEs are smart enough to follow imports, find references, etc, sometimes (IMHO), nothing is more effective but performing a text search when developing. 
If each dep is imported each time with a different name, it is simply impossible. 
A couple of examples:

- Unequivocally way to discern if a `v1alpha1.xxx` is the package core API or the fluxv2 plugin one.
- Easy bulk renames with F2 in the IDE
- Easy way to identify dependencies that don't use the last fragment of their path as the pkg name.

### Possible drawbacks

I know go defaults to the package name of the imported dep, so doing the opposite could be seen like going against the "go-way", but apart from that, don't see any cons.

### Applicable issues

- rel #3848

### Additional information

The upcoming PRs use this one as a prequel.

#### Appendix

Find below the list of unique deps and the chosen alias. Happy to rename if you come up with a better name!

```Dockerfile
anypb "google.golang.org/protobuf/types/known/anypb"
apiscmd "github.com/kubeapps/kubeapps/cmd/kubeapps-apis/cmd"
apiscore "github.com/kubeapps/kubeapps/cmd/kubeapps-apis/core"
apisserver "github.com/kubeapps/kubeapps/cmd/kubeapps-apis/server"
apprepo "github.com/kubeapps/kubeapps/cmd/apprepository-controller/pkg/apis/apprepository"
apprepoclient "github.com/kubeapps/kubeapps/cmd/apprepository-controller/pkg/client/clientset/versioned"
[...]

PR #4179, opened 2022-01-27
Migrate from ghodss/yaml to sigs.k8s.io/yaml

antgamdia:
### Description of the change

This PR just replaces `ghodss/yaml` with Kubernetes fork of the same library, `sigs.k8s.io/yaml`, as it was the most used along our codebase.

### Benefits

Reduce same-purpose deps.

### Possible drawbacks

I don't know we are ever relying on a very specific feature of the original dep whose behavior differs from the k8s-fork one. I can't think of it, but let's our CI decide.

### Applicable issues

- rel #3848

### Additional information

N/A

### DIFF
--- a/cmd/asset-syncer/server/utils.go
+++ b/cmd/asset-syncer/server/utils.go
@@ -24,7 +24,6 @@ import (
 	semver "github.com/Masterminds/semver/v3"
 	"github.com/containerd/containerd/remotes/docker"
 	"github.com/disintegration/imaging"
-	"github.com/ghodss/yaml"
 	"github.com/itchyny/gojq"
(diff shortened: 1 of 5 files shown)


absoludity (approved):
Not sure we need to use custom names for the imports? I'd do that only where it needs disambiguation (ie. if one file imported both for some reason), otherwise, I'd avoid naming `k8syaml` and `goyaml` when there's no ambiguity in the file.

absoludity:
> Not sure we need to use custom names for the imports? I'd do that only where it needs disambiguation (ie. if one file imported both for some reason), otherwise, I'd avoid naming `k8syaml` and `goyaml` when there's no ambiguity in the file.

Oh, I should have read #4176 first, I'll reply there.

PR #4180, opened 2022-01-27
Replace `logrus` by `klog`

antgamdia:
### Description of the change

This PR continues the effort started some time ago in unifying the logging libraries. We opted for `klog`, the one used in Kubernetes. Some minor changes have been introduced in order to adapt the log operations to the ones available in klog (no `WithParams` or `WithError` are available here).

### Benefits

Reduce same-purpose deps. It also simplifies the task of moving to structured logging as we were asked in the past.

### Possible drawbacks

N/A

### Applicable issues

- rel #3848

### Additional information

N/A

### DIFF
--- a/cmd/asset-syncer/server/utils.go
+++ b/cmd/asset-syncer/server/utils.go
@@ -32,11 +32,11 @@ import (
 	"github.com/kubeapps/kubeapps/pkg/helm"
 	httpclient "github.com/kubeapps/kubeapps/pkg/http-client"
 	"github.com/kubeapps/kubeapps/pkg/tarutil"
-	log "github.com/sirupsen/logrus"
 	"github.com/srwiley/oksvg"
 	"github.com/srwiley/rasterx"
 	"helm.sh/helm/v3/pkg/chart"
 	helmregistry "helm.sh/helm/v3/pkg/registry"
+	log "k8s.io/klog/v2"
 )
 
 const (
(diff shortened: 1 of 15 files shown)


cmd/asset-syncer/server/utils.go
```diff
@@ -518,7 +518,7 @@ func pullAndExtract(repoURL *url.URL, appName, tag string, puller helmutils.Char
 
 func chartImportWorker(repoURL *url.URL, r *OCIRegistry, chartJobs <-chan pullChartJob, resultChan chan pullChartResult) {
 	for j := range chartJobs {
-		log.WithFields(log.Fields{"name": j.AppName, "tag": j.Tag}).Debug("pulling chart")
+		log.V(4).Infof("pulling chart, name=%s, tag=%s", j.AppName, j.Tag)
```
absoludity, line 521: Why not use klog's support for structured logging here and above, with something like:
```suggestion
		log.V(4).InfoS("pulling chart", "name", j.AppName, "tag", j.Tag)
```

or even

```suggestion
		log.V(4).InfoS("pulling chart", "chartJob", j)
```

Better eg showing the difference: https://go.dev/play/p/EW5ZQh7MDXI

Or maybe you want to do that separately so it's consistent across the codebase (personally I'd think it OK to do it as we update, like here, but up to you).

absoludity (approved):

cmd/asset-syncer/server/utils.go
antgamdia: Thanks for the suggestion and the playground example. I haven't given it a try yet. Really interesting.

As per https://github.com/kubeapps/kubeapps/issues/3848#issuecomment-1019959653, I already have planned to do so in a separate PR once this stack of PRs became merged. I've added a card to our board not to forget about it.

PR #4186, opened 2022-01-28
Remove wrongly added annotations in carvel plugin

antgamdia:
### Description of the change

As per the discussion in https://github.com/kubeapps/kubeapps/pull/4074#discussion_r794704956, this PR simply reverts back some annotations that were wrongly added in a previous PR.

### Benefits

We will follow the recommnetarions made by the carvel team.

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

N/A


absoludity (approved):

antgamdia:
Also, FTR, at some point, they might be supporting auto-deleting the associated secrets by adding an annotation.
But, we better remove it right now and let's see in the future once it is officially supported.

> i could see how that annotation would be useful to apply (...) (we might want to come up with a plan here if we want this to be an API)

[Source](https://kubernetes.slack.com/archives/CH8KCCKA5/p1643389969628739)

PR #4192, opened 2022-01-31
Add default allow downgrades in carvel pkgs 

antgamdia:
### Description of the change

This PR simply adds support for another default plugin-wide configuration as I commented at https://github.com/kubeapps/kubeapps/pull/4165#discussion_r794583005. 

More specifically, it enables users to decide whether or not the packages can be upgraded to a prior version, that is, a downgrade. This is done by adding (or not) an annotation in the `PackageInstall` objects.

### Benefits

More customization options for Carvel plugin users.

### Possible drawbacks

At the moment, this is a plugin-wide configuration, meaning it cannot be modified (yet) for specific packages.
Nevertheless, it is a good way to identify which config options we may want to add in the UI in the future,

### Applicable issues

- rel #3849 

### Additional information

N/A


docs/user/managing-carvel-packages.md
```diff
@@ -96,6 +96,9 @@ Additionally, you can pass the following configuration values to the `kapp-contr
   - `null`: no prereleases will be installed.
   - `[]`: any prerelease version will be installed (if the version `defaultUpgradePolicy` allows it).
   - `["rc", "foo"]`: only prerelease versions with `rc` or `foo` as part of their names will be installed (if the version `defaultUpgradePolicy` allows it). For instance, `1.0.0-rc1` will be installed, but `1.0.0-bar` will not.
+- `defaultAllowDowngrades`: is a boolean value that determines whether the `kapp-controller` will allow downgrades of packages.
+  - `true`: adds an annotation to every `PackageInstall` resource created by Kubeaoos that will allow them to be downgraded to previous versions of a `Package`.
```
absoludity, line 100:
```suggestion
  - `true`: adds an annotation to every `PackageInstall` resource created by Kubeapps that will allow them to be downgraded to previous versions of a `Package`.
```

absoludity (approved):
When you first mentioned this option, I'd not realised it would be setting an annotation on the installed packages (thought it would be affecting how we calculate the version selector). Not sure about setting the annotation as a blanket, but assume you've thought through the consequence (and it's a default that we may later enable changing per package etc.)

PR #4199, opened 2022-02-01
Update error in AppView to use ErrorAlert.

### LINKED ISSUE #4190: Update `ErrorAlert` component to accept other children for rendering (in Appview)
Currently the `ErrorAlert` is a great improvement for many api errors that are currently exposed in the UX, but it assumes the error is the only child, whereas views like the `AppView` have other children in the alert (such as a button to retry).

absoludity:

### Description of the change

Initially was just to update the error displayed in the AppView to use the `ErrorAlert` component, but the `ErrorAlert` component didn't currently support child elements like the old one did, which the AppView uses. So after chatting with Rafa, I've updated that component to take the error as a prop allowing the children to be normal react nodes.

### Benefits

Before:
appview-error-before

After:
appview-error-after

Clicking on the "Try again" does what it always did, re-issuing the request and rendering the view when it succeeds.

### Applicable issues

- fixes #4190 
- Ref #3695

### DIFF
--- a/dashboard/src/actions/apps.ts
+++ b/dashboard/src/actions/apps.ts
@@ -126,7 +126,13 @@ export function getApp(
       }
       dispatch(selectApp(installedPackageDetail!, resourceRefs, availablePackageDetail));
     } catch (e: any) {
-      dispatch(errorApp(new FetchError(e.message)));
+      dispatch(
+        errorApp(
+          e instanceof Error
+            ? new FetchError("Unable to get installed package", [e])
+            : new FetchError(`Unable to get installed package: ${e.message}`),
+        ),
+      );
     }
   };
 }
--- a/dashboard/src/actions/apps.ts
+++ b/dashboard/src/actions/apps.ts
@@ -126,13 +126,7 @@ export function getApp(
       }
       dispatch(selectApp(installedPackageDetail!, resourceRefs, availablePackageDetail));
     } catch (e: any) {
-      dispatch(
-        errorApp(
-          e instanceof Error
-            ? new FetchError("Unable to get installed package", [e])
-            : new FetchError(`Unable to get installed package: ${e.message}`),
-        ),
-      );
+      dispatch(errorApp(new FetchError("Unable to get installed package", [e])));
     }
   };
 }
@@ -168,13 +162,7 @@ export function fetchApps(
       dispatch(receiveAppList(installedPackageSummaries));
       return installedPackageSummaries;
     } catch (e: any) {
-      dispatch(
-        errorApp(
-          e instanceof Error
-            ? new FetchError("Unable to list apps", [e])
-            : new FetchError("Unable to list apps: " + e.message),
-        ),
-      );
+      dispatch(errorApp(new FetchError("Unable to list apps", [e])));
       return [];
     }
   };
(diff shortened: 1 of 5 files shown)


dashboard/src/actions/apps.ts
```diff
@@ -126,7 +126,13 @@ export function getApp(
       }
       dispatch(selectApp(installedPackageDetail!, resourceRefs, availablePackageDetail));
     } catch (e: any) {
-      dispatch(errorApp(new FetchError(e.message)));
+      dispatch(
+        errorApp(
+          e instanceof Error
```
absoludity, line 131: I've just followed the pattern which you'd done elsewhere Rafa, but wondering when `e` would not be an instance of `Error`? Guessing you saw something else, but it still had an `e.message`, so confused?

dashboard/src/components/ErrorAlert/ErrorAlert.test.tsx
```diff
@@ -7,52 +7,39 @@ import ErrorAlert from "./ErrorAlert";
 import { CustomError } from "shared/types";
 
 describe("Error Alert", () => {
-  it("should render string messages", () => {
-    const wrapper = mount(<ErrorAlert>foo</ErrorAlert>);
-    expect(wrapper.text()).toContain("foo");
-  });
```
absoludity, line 13: See comment below on the interface for more info why I thought to remove this.

dashboard/src/components/ErrorAlert/ErrorAlert.tsx
```diff
@@ -8,7 +8,8 @@ import { CustomError } from "shared/types";
 import "./ErrorAlert.css";
 
 export interface IErrorAlert {
-  children: CustomError | Error | string;
+  error: CustomError | Error;
```
absoludity, line 11: When changing to an `error` prop, given that this is an `ErrorAlert` component (ie. something that should be used when you have an error), it no longer seemed to make sense to support a plain string here since (a) why use an `ErrorAlert` in that case, and (b) even if you wanted to, you could put the string as the child.

dashboard/src/actions/apps.ts
castelblanque: Don't really remember, probably due to the any in `} catch (e: any) {`. But given that it is a catch, I'm with you, can't think of a situation where `e` would not be an Error.

dashboard/src/components/ErrorAlert/ErrorAlert.tsx
castelblanque: There could be use cases where we need to show a plain text message as an error on UI (red, etc.), but agree with you that it can be done with `(b)`.

castelblanque:
Awesome, thanks!

castelblanque (approved):

dashboard/src/actions/apps.ts
```diff
@@ -126,7 +126,13 @@ export function getApp(
...
+            ? new FetchError("Unable to get installed package", [e])
+            : new FetchError(`Unable to get installed package: ${e.message}`),
```
castelblanque, line 133:
```suggestion
          new FetchError("Unable to get installed package", [e])
```

dashboard/src/actions/apps.ts
absoludity: Great, I'll update both this one and the one you'd added before landing.

Issue #4173, opened 2022-01-27
Cannot view newly created flux 2 package from UX


absoludity:
### Description:

Kubeapps UX fails to display a newly installed flux2 package

### Steps to reproduce the issue:

1. Configure Kubeapps with the flux2 plugin
2. Add a `HelmRepository` manually via `kubectl` (I used bitnami's repo)
3. Create a service account with admin role in your chosen namespace
4. Deploy apache (or other chart) from the catalog

Expected result:

Created package can be viewed

Actual result:

The UI displays an error returned from the plugin:

```Application not found: Unable to get the resource refs for the package "test-apache" using the plugin "fluxv2.packages": rpc error: code = NotFound desc = Unable to find Helm release "test-apache" in namespace "kubeapps-user-namespace": release: not found```

### Additional information you deem important (e.g. issue happens only occasionally):

The issue appears to be that the plugin code currently assumes that the flux HelmRelease and the actual helm release object have the same name, but they do not:

```
$ k -n kubeapps-user-namespace get helmrelease
NAME          READY   STATUS                             AGE
test-apache   True    Release reconciliation succeeded   104s

$ helm -n kubeapps-user-namespace ls  
NAME                                    NAMESPACE               REVISION        UPDATED                                 STATUS          CHART           APP VERSION
kubeapps-user-namespace-test-apache     kubeapps-user-namespace 1               2022-01-27 03:11:42.567993337 +0000 UTC deployed        apache-9.0.1    2.4.52     

```

The reason for the different names is due to the defaults that flux uses depending whether the `targetNamespace` is set on the release as outlined on #4172 .

gfichtenholt:
reproduced the problem in the UX
<img width="1440" alt="Screen Shot 2022-01-27 at 9 30 43 PM" src="https://user-images.githubusercontent.com/74032303/151492671-55c8b030-ca17-47c0-87e2-4155391df23c.png">
.
However, with the command line grpcurl there is no problem:
`$ grpcurl -plaintext -d '{"installed_package_ref": {"context": {"cluster": "default", "namespace": "default"}, "plugin": {"name": "fluxv2.packages", "version": "v1alpha1"}, "identifier": "my-podinfo"}}' -H "Authorization: $token" localhost:8080 kubeappsapis.core.packages.v1alpha1.PackagesService.GetInstalledPackageDetail
{
  "installedPackageDetail": {
    "installedPackageRef": {
      "context": {
        "cluster": "default",
[...]

Issue #4189, opened 2022-01-31
Customer requests that kubeapps do not set TargetNamespace field on CRs created by flux plugin


gfichtenholt:
Discussion carried over from: https://github.com/kubeapps/kubeapps/issues/4173
In short:
Customer is requesting when kubeapps is creating new Flux HelmRelease CRs:
1) we do not set the TargetNamespace field
2) we set the ReleaseName field

FWIW, this is what Flux CLI does _by default_: 
```
$ flux create source helm podinfo \
>     --url https://stefanprodan.github.io/podinfo \
>     --namespace default
✚ generating HelmRepository source
► applying HelmRepository source
✔ source created
◎ waiting for HelmRepository source reconciliation
✔ HelmRepository source reconciliation completed
✔ fetched revision: 83a3c595163a6ff0333e0154c790383b5be441b9db632cb36da11db1c4ece111
$ flux create hr my-podinfo \
>     --namespace=default \
>     --source=HelmRepository/podinfo.default \
>     --chart=podinfo
✚ generating HelmRelease
► applying HelmRelease
✔ HelmRelease created
◎ waiting for HelmRelease reconciliation
✔ HelmRelease my-podinfo is ready
✔ applied revision 6.0.3
$ kubectl get helmrelease/my-podinfo -o yaml
apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
metadata:
creationTimestamp: "2022-01-31T07:50:56Z"
finalizers:
- finalizers.fluxcd.io
generation: 1
name: my-podinfo
namespace: default
resourceVersion: "2363"
uid: 24731219-6d27-422b-8f9a-2f71da306882
spec:
chart:
  spec:
    chart: podinfo
    reconcileStrategy: ChartVersion
    sourceRef:
      kind: HelmRepository
      name: podinfo
      namespace: default
    version: '*'
interval: 1m0s
status:
conditions:
- lastTransitionTime: "2022-01-31T07:50:57Z"
  message: Release reconciliation succeeded
  reason: ReconciliationSucceeded
  status: "True"
  type: Ready
- lastTransitionTime: "2022-01-31T07:50:57Z"
  message: Helm install succeeded
  reason: InstallSucceeded
  status: "True"
  type: Released
helmChart: default/default-my-podinfo
lastAppliedRevision: 6.0.3
lastAttemptedRevision: 6.0.3
[...]

PR #4204, opened 2022-02-01
Fix for Cannot view newly created flux 2 package from UX #4173 

gfichtenholt:


cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/release.go
```diff
@@ -676,3 +665,24 @@ func installedPackageAvailablePackageRef(rel *helmv2.HelmRelease) (*corev1.Avail
...
 }
+
+// ref https://fluxcd.io/docs/components/helm/helmreleases/
+func helmReleaseName(key types.NamespacedName, rel *helmv2.HelmRelease) types.NamespacedName {
+	helmReleaseName := rel.Spec.ReleaseName
+	// according to docs ReleaseName is optional and defaults to a composition of
+	// '[TargetNamespace-]Name'.
+	if helmReleaseName == "" {
+		targetNamespace := rel.Spec.TargetNamespace
+		// according to docs targetNamespace is optional and defaults to the namespace of the HelmRelease
```
absoludity, line 676: Just noting, not necessary for this fix, but for the related #4189, that the behaviour flux shows is slightly different to below (see my notes from updating the flux CR on #4172, or your notes on #4189) but still consistent with the comment above: the namespace targeted by the `HelmRelease` defaults to the namespace of the `HelmRelease`, but flux doesn't set the `targetNamespace` field with that value, it remains empty as specified. And when the `targetNamespace` is empty/unset, the helm release name is just the `HelmRelease.metadata.name` (ie. not prefixed by the namespace at all, and so different to below). This is why I was unsure why we were explicitly setting the `targetNamespace` given flux's default behaviour.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/release.go
```diff
@@ -676,3 +665,24 @@ func installedPackageAvailablePackageRef(rel *helmv2.HelmRelease) (*corev1.Avail
...
+		if targetNamespace == "" {
+			targetNamespace = key.Namespace
+		}
+		helmReleaseName = fmt.Sprintf("%s-%s", targetNamespace, key.Name)
```
absoludity, line 680: Just leaving this here so I can refer to it on your related issue #4189, not suggesting the change for this PR

```suggestion
		helmReleaseName = key.Name
		if targetNamespace != "" {
			helmReleaseName = fmt.Sprintf("%s-%s", targetNamespace, key.Name)
		}
```
which would match the behaviour that I saw by default from flux, I think.

absoludity (approved):
Excellent, Greg, thanks for the quick fix!

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/release.go
gfichtenholt: > Just noting, not necessary for this fix, but for the related #4189, that the behaviour flux shows is slightly different to below (see my notes from updating the flux CR on #4172, or your notes on #4189) but still consistent with the comment above: the namespace targeted by the `HelmRelease` defaults to the namespace of the `HelmRelease`, but flux doesn't set the `targetNamespace` field with that value, it remains empty as specified. And when the `targetNamespace` is empty/unset, the helm release name is just the `HelmRelease.metadata.name` (ie. not prefixed by the namespace at all, and so different to below). This is why I was unsure why we were explicitly setting the `targetNamespace` given flux's default behaviour.

noted

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/release.go
gfichtenholt: ok, sounds good

PR #4211, opened 2022-02-02
Separate action for requesting resource refs.

absoludity:

### Description of the change

Separates out the action for fetching resource refs for an installed application.

### Benefits

When the resource refs were first added to the `SelectedApp` state, I think Antonio and I both agreed that it didn't belong there (in the `CustomInstalledPackageDetail` which was temporarily created to accommodate the helm version field), but at the time I wasn't keen to refactor it.

Now we have an unrelated issue that for flux (and carvel), the call to fetch the resource refs fails with a 404 initially (within 1 second of creating the installed package) because the underlying reconciliation creates the helm (for flux) release very soon after creating the installed package, but not immediately.

I'm going to create an issue for the plugin backend so that the request for resource refs (a valid one, for which there's already an installed package) doesn't return until the underlying helm release is created (within a second), but am also planning here in the dashboard to work around the issue for now by requesting the resource refs a few times with a fallback (next branch). But to do that, I needed to fix this tech-debt to separate out the call for the resource refs.

### Possible drawbacks

Behaviour is the same.

### Applicable issues

- ref #3695

### DIFF
--- a/dashboard/src/actions/apps.ts
+++ b/dashboard/src/actions/apps.ts
@@ -31,7 +31,13 @@ import { validate } from "../shared/schema";
 
 const { createAction } = deprecated;
 
-export const requestApps = createAction("REQUEST_APPS");
+export const requestApp = createAction("REQUEST_APP");
+
+export const requestAppResourceRefs = createAction("REQUEST_APP_RESOURCE_REFS");
+
+export const receiveAppResourceRefs = createAction("RECEIVE_APP_RESOURCE_REFS", resolve => {
+  return (refs: ResourceRef[]) => resolve(refs);
+});
 
 export const listApps = createAction("REQUEST_APP_LIST");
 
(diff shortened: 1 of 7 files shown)


castelblanque:
Great, thanks for the refactor!

dashboard/src/actions/apps.ts
```diff
@@ -99,11 +104,10 @@ export function getApp(
   installedPackageRef?: InstalledPackageReference,
 ): ThunkAction<Promise<void>, IStoreState, null, AppsAction> {
   return async dispatch => {
-    dispatch(requestApps());
+    dispatch(requestApp());
```
castelblanque, line 107: Are we going to unify terminology regarding "app" vs "package", etc? I mean, for example this function is called `getApp`, actions are `requestApp` and `selectApp` but it is dealing with `*PackageDetail` and outputting  `installedPackageDetail` and `availablePackageDetail`. Or am I not getting the distinction here (which could perfectly be 😅 ) ?

antgamdia: I guess this is for historical reasons. I did perform a rename from `chart` to `package` back in time; however, the `apps` thing is something still pending.
+1 to move from `App` to `InstalledPackage` instead, of course.

antgamdia (approved):
Great, awesome. This (and the subsequent PRs) is gonna ensure a good UX once we move any big delays to get Refs operation, as the minimal UI still will be rendering and we could possibly add some spinner or whatever to let users know something else is still being fetched.
Thanks!

dashboard/src/actions/apps.ts
absoludity: Yes, exactly. I did actually have it with package initially, but then saw the other actions related to `App` and `Apps`, and didn't want to create an inconsistency here nor rename all of them here. Happy to do a separate PR with such.

Issue #4096, opened 2022-01-17
Investigate CI improvements


castelblanque:
### Description:

Current CI solution makes use of CircleCI and GKE, but it is failing at an increasing rate.
Investigate what actions can be taken in order to improve/fix the problems when running CI.
In order to do so, check what other Tanzu projects are doing regarding CI.

Some aspects to take a look at:

- Pipeline tool for running CIs and its potential limitations
- Usage for K8s clusters other than GKE
- Which actions to perform for each push, PR and release
- Possible improvements to the current features of Kubeapps CI
- Independent of external resources such as repos and charts.

### Describe the results you expected:

Having a proper CI system that runs seamlessly and allows extension of features.

castelblanque:
Please find below a list of proposed actions to do for Kubeapps CI. There is no particular order, it is just some initial ideas to discuss and extend/reduce if needed. List is a mix of improvements and fixes.
Goal of this is having a better quality of the product but also making contributors life easier.

1. Run E2E tests for all three browsers available in Playwright when doing a release.
    * Benefit: Covering more browsers to find potential UI issues
    * CI event: On release
2. Add new CI step to check that all files contain the right copyright header (see [Cartographer](https://github.com/vmware-tanzu/cartographer/blob/main/.github/workflows/release.yaml#L33)).
    * Benefit: No file in project will be merged in main without Copyright header by mistake
    * CI event: Always
3. Add static code analysis
    * Benefit: Adds continuous inspection, code coverage, code smells detection, code quality gates, etc.
    * CI event: On PRs
4. Clean up   of `page.waitForTimeout()` usage in tests
    * Benefit: Reducing CI tests total time and simplifying tests source code
5. Use [pre-commit](https://pre-commit.com/) CLI on dev machines. [See example from Pinniped.](https://github.com/vmware-tanzu/pinniped/blob/main/.pre-commit-config.yaml)
    * Benefit: Possibility to add validations even before code is pushed to Git repo (e.g. check Copyright year)
6. (Fix) Install ChartMuseum before Kubeapps in CI, so that initial repos syncing does not fail due to ChartMuseum not being available at the time of Kubeapps installation. Install it in its own namespace, instead of Kubeapps.
    * Benefit: Initial repos syncing is successful and therefore less chance for tests to fail
    * CI event: Always
7. Add additional ChartMuseum instance with public access, so that we can get rid of Gitlab repo and Prometheus repo. 
[...]

PR #4221, opened 2022-02-02
Default user positive and negative test for deploying

### LINKED ISSUE #4081: CI does not test install with a default user
### Description:

Our current CI tests work with two roles: `kubeapps-user` and `kubeapps-operator`, but only use the `kubeapps-user` role for the check that an install fails (when installed in a namespace to which the user does not have access).

We should update our positive test-cases to use the more limited `kubeapps-user` role where we can, for a simple deployment. This would have caught #4040 .

castelblanque:
### Description of the change

This PR adds a positive test-case to use the more limited user `kubeapps-user` (deploy & delete).
It also adds a new negative test for that user trying to deploy from a repo with secrets to which user does not have access.

### Benefits

We are better covering the "limited" user deploy/undeploy actions in CI.

### Possible drawbacks

N/A

### Applicable issues

- fixes #4081

### DIFF
--- a/integration/tests/main/09-user-positive-installation.spec.js
+++ b/integration/tests/main/09-user-positive-installation.spec.js
@@ -1,0 +1,81 @@
+// Copyright 2022 the Kubeapps contributors.
+// SPDX-License-Identifier: Apache-2.0
+
+const { test, expect } = require("@playwright/test");
+const { KubeappsLogin } = require("../utils/kubeapps-login");
+const utils = require("../utils/util-functions");
+
+test.describe("Limited user simple deployments", () => {
+  test("Regular user can deploy and delete packages in its own namespace from global repo without secrets", async ({
+    page,
+  }) => {
+    test.setTimeout(120000);
+
+    // Log in as admin to create a repo without password
+    const k = new KubeappsLogin(page);
+    await k.doLogin("kubeapps-operator@example.com", "password", process.env.ADMIN_TOKEN);
+
+    // Change namespace using UI
+    await page.click(".kubeapps-dropdown .kubeapps-nav-link");
+    await page.selectOption('select[name="namespaces"]', "kubeapps");
+    await page.click('cds-button:has-text("Change Context")');
+
+    // Go to repos page
+    await page.click(".dropdown.kubeapps-menu button.kubeapps-nav-link");
+    await page.click('a.dropdown-menu-link:has-text("App Repositories")');
+    await page.waitForTimeout(3000);
+
+    // Add new repo
+    const repoName = utils.getRandomName("repo-test-09");
+    console.log(`Creating repository "${repoName}"`);
+    await page.click('cds-button:has-text("Add App Repository")');
+    await page.type("input#kubeapps-repo-name", repoName);
+    await page.type(
+      "input#kubeapps-repo-url",
+      "https://prometheus-community.github.io/helm-charts",
+    );
+    await page.click('cds-button:has-text("Install Repo")');
+    await page.waitForLoadState("networkidle");
+
+    // Log out admin and log in regular user
+    await k.doLogout();
+    await k.doLogin("kubeapps-user@example.com", "password", process.env.VIEW_TOKEN);
+
+    // Switch to user's namespace using UI
+    await page.click(".kubeapps-dropdown .kubeapps-nav-link");
+    await page.selectOption('select[name="namespaces"]', "kubeapps-user-namespace");
+    await page.click('cds-button:has-text("Change Context")');
+
+    // Select package to deploy
+    await page.click('a.nav-link:has-text("Catalog")');
+    await page.locator("input#search").type("alertmanager");
+    await page.waitForTimeout(3000);
+    await page.click('a:has-text("alertmanager")');
+    await page.click('cds-button:has-text("Deploy") >> nth=0');
+
+    // Deploy package
+    const releaseNameLocator = page.locator("#releaseName");
+    await releaseNameLocator.waitFor();
+    await expect(releaseNameLocator).toHaveText("");
+    const releaseName = utils.getRandomName("test-09-release");
+    console.log(`Creating release "${releaseName}"`);
+    await releaseNameLocator.type(releaseName);
+    await page.locator('cds-button:has-text("Deploy")').click();
+
+    // Check that package is deployed
+    await page.waitForSelector("css=.application-status-pie-chart-number >> text=1");
+    await page.waitForSelector("css=.application-status-pie-chart-title >> text=Ready");
+
+    // Delete deployment
+    await page.locator('cds-button:has-text("Delete")').click();
+    await page.locator('cds-modal-actions button:has-text("Delete")').click();
+    await page.waitForTimeout(10000);
+
+    // Search for package deployed
+    await page.click('a.nav-link:has-text("Applications")');
+    await page.locator("input#search").type("alertmanager");
+    await page.waitForTimeout(3000);
+    const packageLocator = page.locator('a:has-text("alertmanager")');
+    await expect(packageLocator).toHaveCount(0);
+  });
+});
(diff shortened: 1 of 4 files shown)


integration/tests/main/09-user-positive-installation.spec.js
```diff
@@ -0,0 +1,81 @@
+// Copyright 2022 the Kubeapps contributors.
```
absoludity, line 1: FWIW, I wasn't thinking that we needed a new test for this, but rather just updating the existing default deployment test to use the `kubeapps-user`... we know if it passes with that that it will have no issues to pass with a higher-privileged account. Adding redundant tests will just lengthen the test run, won't it? Or is there a reason you think we should test both with the `kubeapps-user` as well as with `kubeapps-operator`?

absoludity (commented):
Love the new output with playwright. Especially marking flaky tests etc.

absoludity:
From the errors, it looks like the kubeaps-user account can't read secrets in the `kubeapps-user-namespace`. I think what's happened is that previously these specific tests were done with token auth (not OIDC) and so were using the `view` and/or `edit` token which has view/edit access in the `default` namespace. To use the OIDC user (which is better, excellent), we'll need to add similar rolebindings (in the `e2e-test.sh`) for the `oidc:kubeapps-user@example.com` user.

absoludity:
In the dev env, this is done with https://github.com/kubeapps/kubeapps/blob/main/docs/user/manifests/kubeapps-local-dev-users-rbac.yaml#L30-L52

castelblanque:
> To use the OIDC user (which is better, excellent), we'll need to add similar rolebindings (in the `e2e-test.sh`) for the `oidc:kubeapps-user@example.com` user.

I noticed that yesterday, wanted to give it a try first with the user "as-is" so that I'm sure I don't break up something else. For some reason the `kubeapps-user` rolebinding is applied first...and deleted [in this line](https://github.com/kubeapps/kubeapps/blob/main/.circleci/config.yml#L317).

>  I think what's happened is that previously these specific tests were done with token auth (not OIDC)

That explains the deletion, thanks!
Now I'll try removing the deletion of the rolebinding.

integration/tests/main/09-user-positive-installation.spec.js
castelblanque: In the new `09` test, repository is created by admin user and consumed by another user with less privileges. I don't think we are covering that in other tests? I mean, in `03` test, admin does everything (add repo, deploy, etc.) and in test `04` admin deploys to `default` namespace using an initial repo from Kubeapps installation.
I think each test covers a slightly different scenario, it's a bit of the "how far do we want to get with e2e tests?" question I would say.
However, happy to remove any test if you think that is not needed!

castelblanque:
@absoludity Update: removing the rolebinding deletion from `.circleci/config.yml` works as expected.

integration/tests/main/09-user-positive-installation.spec.js
absoludity: OK - I hadn't realised it was adding a repo. I'm OK with having a separate test, but it's not clear to me why we'd add a repo as part of this test (given we already have a repo with charts we can use for this test, and have other tests that add repositories). Also, I'm not sure about adding a non-bitnami repo so that our CI is now dependent on the prometheus community helm charts?

absoludity: > @absoludity Update: removing the rolebinding deletion from `.circleci/config.yml` works as expected.

Great, though it might be worth checking with @antgamdia - looks like he added the deletion. I can only guess that the intent was to use the `default` namespace for the kubeapps user and so extra RBAC for other namespaces was removed, at the time.

absoludity (approved):
Just check with Antonio about the deletion, in case there's a reason we don't know.

Also, I don't think we should make our CI test 09 dependent on the prometheus community helm chart repo? Can we consider either not including adding a repo in the 09 test (another test already tests that), or adding the bitnami repo with a different name and navigating to that as the other user, if it's necessary?

castelblanque:
> Also, I don't think we should make our CI test 09 dependent on the prometheus community helm chart repo? Can we consider either not including adding a repo in the 09 test (another test already tests that), or adding the bitnami repo with a different name and navigating to that as the other user, if it's necessary?

The repo added in test 09 is a public one, whereas the chartmuseum we are using in other tests has user/pwd. That allows to test both scenarios.
As I saw we are using a repo from Gitlab in another test, I thought it could be ok. But Gitlab's repo didn't allow to deploy correctly (it was failing due to needing to create statefulsets, secrets, etc.).
I wanted to avoid the Bitnami repo so that we don't overload the CI environment + the UI. Speed of the test clicking through the UI is a bit variable and sometimes it gets to the catalog before packages are there.
We can look at alternatives like adding Bitnami repo with filtering, or any other known small VMWare repo (if any)? Another posibilty is to add an additional, public instance of chartmuseum.

castelblanque:
> Just check with Antonio about the deletion, in case there's a reason we don't know.

@antgamdia Do you know if putting back the `kubeapps-user` role binding for the view user could bring any issues?
Tests are passing successfully.

integration/tests/main/09-user-positive-installation.spec.js
antgamdia: That's right, though I can't 100% remember, the idea was just using the default namespace (perhaps because of the problems you mentioned about the test selecting the wrong context?), so I just deleted this extra RBAC (which was granting `edit` ClusterRole in the `kubeapps-user-namespace` namespace
Now the tests are handling better how to switch contexts, +1 to remove any deletion that might veil any errors on the current rbac dev file we're using.

absoludity:
> > Also, I don't think we should make our CI test 09 dependent on the prometheus community helm chart repo? Can we consider either not including adding a repo in the 09 test (another test already tests that), or adding the bitnami repo with a different name and navigating to that as the other user, if it's necessary?
> 
> The repo added in test 09 is a public one, whereas the chartmuseum we are using in other tests has user/pwd. That allows to test both scenarios.

The test is to ensure that a "Regular user can deploy and delete packages in its own namespace from global repo without secrets", which I think is a scenario we could have done with the default deployment (already using the public bitnami repo). Ah, or maybe you mean testing the *adding* of both a public and private repo (we currently test deploying from both, but didn't need to add a public one, since bitnami was there already). Not sure it's necessary, but yep, that's a valid reason, but different to the stated intent of the test. Fine either way.

> As I saw we are using a repo from Gitlab in another test, I thought it could be ok. But Gitlab's repo didn't allow to deploy correctly (it was failing due to needing to create statefulsets, secrets, etc.). I wanted to avoid the Bitnami repo so that we don't overload the CI environment + the UI. Speed of the test clicking through the UI is a bit variable and sometimes it gets to the catalog before packages are there, or performing a search gets stuck. We can look at alternatives like adding Bitnami repo with filtering, or any other known small VMWare repo (if any)? Another posibilty is to add an additional, public instance of chartmuseum.

I'd generally avoid as many external dependencies as possible to reduce CI issues unrelated to actual tests. So my preference would be for the latter if we really wanted to. Or simply generate and serve a trivial index.yaml of our own with one chart (using whatever the latest bitnami apache chart is or similar). As it is, I think our CI is now dependent on the external prometheus chart deploying successfully? Or just update the 04-default-deployment to use the `kubeapps-user` user (and yes, we wouldn't be testing adding a public repo in e2e tests, just adding a private one, which is OK imo). Up to you - I don't feel strongly, just don't understand why we would choose to add this dependency on both an external helm repo and chart.

castelblanque:
> we currently don't actually add the normal bitnami repository at all

Yeah, it was like that previously, so I preferred not to add it. Public Gitlab repo was being added instead. See my proposal below.

> I'd generally avoid as many external dependencies as possible to reduce CI issues unrelated to actual tests. 

Totally agree, but the priority was to unblock CI. There are many tasks to be done in order to tidy up CI, let's gather proposals in #4096.
In order to be independent from external sources (even Bitnami) we should have IMHO:
- Chartmuseum instance (or similar) with no credentials (we already have the one with creds)
- Images registry serving repos with and without credentials (in GKE we might be able to use Google's)

Not rocket science, just takes some time.

PR #4222, opened 2022-02-02
Check the versionSelection before install/update

antgamdia:
### Description of the change

I've noticed I've been having some issues when installing some carvel packages: they got "installed" but never reconcillied.
The problem was that we were requesting a version that should not be selected. For instance, a prerelease when they are disabled by default.
Idem when updating: nothing prevents one from selecting a previous version, regardless of the default config.

### Benefits

The version will be checked before installing, this way the user will get a ~friendly~ message.

### Possible drawbacks

Although IMO, this assertion belongs in the backend world, we should ensure we only display "valid" versions in the user-faced UI selectors.

### Applicable issues

- rel #3849 

### Additional information

Marked as a draft as tests are still missing.

image


cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_adapters.go
```diff
@@ -324,6 +326,17 @@ func (s *Server) buildPkgInstall(installedPackageName, targetCluster, targetName
 	}
 	prereleases := prereleasesVersionSelection(s.pluginConfig.defaultPrereleasesVersionSelection)
 
+	versionSelection := &vendirversions.VersionSelectionSemver{
+		Constraints: versionConstraints,
+		Prereleases: prereleases,
+	}
+
+	// Ensure the selected version can be, actually installed to let the user know before installing
+	elegibleVersion, err := versions.HighestConstrainedVersion([]string{pkgVersion}, vendirversions.VersionSelection{Semver: versionSelection})
```
antgamdia, line 335: This is how kapp-ctrl is actually doing that; which makes me wonder: should we use `vendir` from carvel instead of other `semver` dependencies?

absoludity: If it suits our needs, sure.

absoludity (approved):
> Although IMO, this assertion belongs in the backend world, we should ensure we only display "valid" versions in the user-faced UI selectors.

I agree (that we should only present the valid versions for upgrading), but that wouldn't obviate the need for this change would it? (given that it's an API, so a request could come in to install an invalid version regardless of what we present in the UX).

PR #4223, opened 2022-02-02
3695 check installed pkg details 4

absoludity:
### Description of the change

Following on from a [comment in the previous PR](https://github.com/kubeapps/kubeapps/pull/4211#discussion_r797477016), this one is just mechanical renaming/refactoring. I was initially just going to make the `actions/apps.ts` content consistent, but then also did the related reducers, the shared helper (`shared/App.ts`), and moved the existing `packages.ts` to `avaliablepackages.ts` to make room for the `installedpackages.ts`.

### Benefits

The code (in those files at least), now consistently uses the current terms.

### Applicable issues

- ref #3695


antgamdia (approved):
Great! Thanks for the effort! Now it'd be way easier to identify which concept are we referring to.

PR #4224, opened 2022-02-03
fix for  Customer requests that kubeapps do not set TargetNamespace field on CRs created by flux plugin #4189 

gfichtenholt:


cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
```diff
@@ -514,8 +514,16 @@ func (s *Server) GetInstalledPackageResourceRefs(ctx context.Context, request *c
...
+				Cluster: s.kubeappsCluster,
+				// TODO (gfichtenholt) it is not specifically called out in the spec why there is a
+				// need for a Context in the response and MORE imporantly what the value of Namespace
+				// field should be. In particular, there is use case when Flux Helm Release in
+				// installed in ns1 but specifies targetNamespace as test2. Should we:
+				//  (a) return ns1 (the namespace where CRs are installed) OR
+				//  (b) return ns2 (the namespace where flux installs the resources specified by the
+				//    release).
+				// For now lets use (a)
+				Namespace: key.Namespace,
```
absoludity, line 526: Yep - agree. In the dashboard UX we're just pulling out the resourcerefs only (ie. not even using the context). I cannot see any reason that a client would need this (since they include the context in the request).

absoludity (approved):
Thanks Greg.

gfichtenholt:
Thank you Michael

PR #4225, opened 2022-02-03
Update tests for installed package to test call to client.

absoludity:

### Description of the change

I was getting ready to update the `GetInstalledPackageResourceRefs` shared util, went to write a test for the existing code, but found that the existing tests for that module were mocking the actual function being tested, rather than mocking the underlying client so that we can verify that the client is called with the arguments.

This PR just updates those tests so that we are not mocking the actual functions that we're testing (guess we did this when rushing to switch to the new grpc client). The tests aren't particularly valuable here, since the util is just calling the client with the same args, but the test for GetInstalledPackageResourceRefs will be extended to be more useful in the following PR (where I'll add retries).

Only stopping and pushing this now because I don't have time to do the second change today.

### Benefits

Just preparing for the next PR which will add the retries for GetInstalledPackageResourceRefs.

### Applicable issues

- ref #3695

### DIFF
--- a/dashboard/src/shared/InstalledPackage.test.ts
+++ b/dashboard/src/shared/InstalledPackage.test.ts
@@ -6,33 +6,24 @@ import {
   Context,
   CreateInstalledPackageResponse,
   DeleteInstalledPackageResponse,
+  GetInstalledPackageResourceRefsResponse,
   InstalledPackageReference,
   UpdateInstalledPackageResponse,
   VersionReference,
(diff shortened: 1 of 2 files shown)


antgamdia (approved):
Thanks for this change! Yep, we were in a rush and didn't properly mock the client.
In fact, I had a look (trying to mock the grpc client itself by intercepting the transport with a library), but I finally wasn't able to do it. This neat approach of `setMockCoreClient` is simple and effective :)

PR #4232, opened 2022-02-03
Extract some functions to be used in other plugins

antgamdia:
### Description of the change

This PR simply extracts the "generate default values from a schema" and the "wait for a k8s resource to be present" apart. 

Also, the `semver` dep version is updated from v2 to v3 (to make it consistent with the rest of our codebase)

### Benefits

Other plugins can re-use the same functions.

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

N/A


absoludity (approved):
Great, thanks!

Issue #4213, opened 2022-02-02
Fluxv2: GetInstalledPackageResourceRefs() returns not-found initially before flux creates helm release


absoludity:
### Description:

When creating an installed package via flux, flux creates the `HelmRelease` and returns, so that the Kubeapps UI redirects to display the installed package details.

The UI requests `GetInstalledPackageResourceRefs()`, but the plugin initially returns a not-found error, because it tries to find the actual helm release which the flux controller hasn't yet created (it's quick, but obviously not synchronous).

I plan to work around this in the UI simply by having the request re-issued a few times in this specific situation, but I think ideally, given that the plugin can easily determine that the `HelmRelease` exists and that the actual helm release will be created (probably in under a second from local testing), that it should in this situation delay returning until the actual helm release has been created and it can continue to return the resource references. This means that clients of the API don't need to handle this. Of course it would still have a context deadline or similar to fail if the actual helm release wasn't created in some reasonable time.

Anyway, that's just my thoughts - let me know what you think.

### Steps to reproduce the issue:

1. Install Kubeapps with fluxv2 enabled
2. Install apache

### Describe the results you received:

UX displays the not-found error, complaining that the actual helm release was not found (because flux is about to create it). Clicking Try again then succeeds and displays the installed package details, including the resources.

### Describe the results you expected:

The UI displays the installed package details without error. (I'll add a work-around to the UX as mentioned above so this is achieved for now).

gfichtenholt:
No problem. I will take a look tomorrow. Thanks

gfichtenholt:
Sorry, just to clarify, are you suggesting I make `CreateInstalledPackage()` into a blocking call? Or introduce some kind of delay mechanism into `GetInstalledPackageResourceRefs()` such that it'll block until the newly created  release appears in helm? Either way, that doesn't feel right. Any artificial blocking I introduce will basically be polling for some condition. So, why poll on the server side? That _feels_ off. Our UX is just one kind of client for kubeapps. We also have (or could have) an REST/gRPC API clients. If I add the polling on the server-side, every client will be affected. Some clients might not care. To me it seems client side should be the one polling. 
Also, we have an API that returns the installed package detail, including status. So, to me it seems logical for UX to poll on status until it shows SUCCESS or FAILURE, and only after that ask for `GetInstalledPackageResourceRefs()`
Having said that, I may have missed something and I am listening and willing to be convinced otherwise.

gfichtenholt:
[...]

PR #4233, opened 2022-02-04
Call GetInstalledPackageResourceRefs with exp backoff.

absoludity:

### Description of the change

See #4213 for description of the problem.

This change just ensures that the call will retry a number of times, backing off exponentially. For Flux it only needs to be a second, but carvel has other needs.

### Benefits

Error no longer shown when first deploying flux package.

### Applicable issues

- Ref #4213 
- Ref #3695

### DIFF
--- a/dashboard/src/shared/InstalledPackage.ts
+++ b/dashboard/src/shared/InstalledPackage.ts
@@ -43,10 +43,20 @@ export class InstalledPackage {
 
   public static async GetInstalledPackageResourceRefs(
     installedPackageRef?: InstalledPackageReference,
+    initialWait: number = 500,
   ) {
-    return await this.coreClient().GetInstalledPackageResourceRefs({
-      installedPackageRef: installedPackageRef,
-    });
+    // TODO(minelson): The backend plugin may take care of waiting
+    // for the required data to become available in which case this
+    // can be removed.
+    // See https://github.com/kubeapps/kubeapps/issues/4213
+    // Note: initialWait is set with a default value so it can be
+    // tested with a value of 0 (because I couldn't get jest's mock
+    // timers to work with promises here.)
+    const fn = async () =>
+      this.coreClient().GetInstalledPackageResourceRefs({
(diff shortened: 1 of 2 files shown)


dashboard/src/shared/InstalledPackage.ts
```diff
@@ -105,3 +115,21 @@ export class InstalledPackage {
     } as DeleteInstalledPackageRequest);
   }
 }
+
+// Helpers to be able to call with a retry backing off exponentially
+// with 500ms, 1000ms, 2000ms, 4000ms etc.
+const wait = (ms: number) => new Promise(res => setTimeout(res, ms));
+
+const callWithRetry = async (fn: any, initialWait: number, depth = 0): Promise<any> => {
```
antgamdia, line 123: What if we just use a library instead for also [adding jitter](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/) to this exp backoff, like https://github.com/lifeomic/attempt ?

antgamdia (approved):
Great! Thanks! +1 once CI passes

dashboard/src/shared/InstalledPackage.ts
castelblanque: @antgamdia thanks for the article about jitter, superinteresting!
That library looks easy to implement 😄

antgamdia:
By the way, do you think we should add some kind of spinner when waiting? 

image

If so, (probably in another PR) we should let the `ResourceTabs` component know if the fetching attempts have been exhausted or not, I guess.

dashboard/src/shared/InstalledPackage.ts
absoludity: Why would we need jitter here? It's guaranteed to be a single client fetching the resource refs after creating the package... unless we coordinate a DDOS on a Kubeapps installation :P. (ie. we don't have the situation described in the article that leads to adding jitter to an exponential backoff, nor could we, unless I'm missing something?)

PR #4234, opened 2022-02-04
Preparation for release v2.4.3

castelblanque:
### Description of the change

This PR adds:

- Dependency updates for source code
- Updates from Bitnami chart (mainly a change in the ingress certManagerRequest)

### Benefits

We can have prepared release v2.4.3

### Possible drawbacks

E2E tests 03 and 07 seem to be failing consistently now. I'm preparing another PR that should fix that.


chart/kubeapps/Chart.yaml
```diff
@@ -33,4 +33,4 @@ maintainers:
 name: kubeapps
 sources:
   - https://github.com/kubeapps/kubeapps
-version: 7.8.0-dev3
+version: 7.7.4
```
castelblanque, line 36: @antgamdia Is this version correct?
[Bitnami chart version](https://github.com/bitnami/charts/blob/master/bitnami/kubeapps/Chart.yaml#L36)

antgamdia: Nope, I don't think so. Since we added new params in the chart (in the kapp-ctrl plugin config), this should be a `minor` release according to the Content team. So I just bumped it to 7.8.0 in the first change, and then I continue modifying the devXX prerelease tag.
In short, I don't think we should modify the chart version here.

What we do have to change, though, is this version once our CI creates the PR in the bitnami/charts repo. I mean, in their repo, not in ours.

dashboard/package.json
```diff
@@ -26,6 +26,7 @@
     "watch-css": "yarn run build-css && yarn run compile-lang && sass src/ --watch --no-source-map"
   },
   "dependencies": {
+    "@babel/parser": "^7.15.3",
```
antgamdia, line 29: I don't think this one is required. Sometimes it becomes added when doing `yarn upgrade`, but it is already present as part of the react-scripts deps. If you remove it and then run yarn install, you'll notice everything works fine.

chart/kubeapps/Chart.yaml
castelblanque: Ok, got it! Thanks

dashboard/package.json
castelblanque: Done!

chart/kubeapps/Chart.yaml
castelblanque: Done!

antgamdia (approved):
Awesome! Thanks for the upgrades :)

PR #4235, opened 2022-02-04
Add initial tpl validation for plugins

antgamdia:
### Description of the change

While I was enabling/disabling plugins in my local env, I've spent some time wondering what was wrong in my setup... it turned to be just a typo in the plugin name :facepalm: 

This PR adds some validation to the plugin names we are using, at least, for the mistakes I use to make :P
Feel free to suggest more.

### Benefits

No "OSI layer 8"-derived problems, haha

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information
```bash
    kubeapps: kubeappsapis.enabledPlugins 
        You enter "flux", perhaps you meant "fluxv2"?
    kubeapps: kubeappsapis.enabledPlugins 
        You enter "kapp_controller", perhaps you meant "kapp-controller"?
    kubeapps: kubeappsapis.enabledPlugins 
        If you enable the "fluxv2" plugin, you must also set redis.enabled=true
```

### DIFF
--- a/chart/kubeapps/templates/_helpers.tpl
+++ b/chart/kubeapps/templates/_helpers.tpl
@@ -203,6 +203,7 @@ Compile all warnings into a single message, and call fail.
 {{- define "kubeapps.validateValues" -}}
 {{- $messages := list -}}
 {{- $messages := append $messages (include "kubeapps.validateValues.ingress.tls" .) -}}
+{{- $messages := append $messages (include "kubeapps.validateValues.kubeappsapis.enabledPlugins" .) -}}
 {{- $messages := without $messages "" -}}
 {{- $message := join "\n" $messages -}}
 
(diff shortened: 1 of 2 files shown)


chart/kubeapps/templates/_helpers.tpl
```diff
@@ -227,6 +228,25 @@ kubeapps: ingress.tls
...
+    {{- if has "kapp_controller" .Values.kubeappsapis.enabledPlugins }}
+    kubeapps: kubeappsapis.enabledPlugins 
+        You enter "kapp_controller", perhaps you meant "kapp-controller"?
+    {{- end -}}
+    {{- if and (has "fluxv2" .Values.kubeappsapis.enabledPlugins) (not .Values.redis.enabled) }}
+    kubeapps: kubeappsapis.enabledPlugins 
+        If you enable the "fluxv2" plugin, you must also set redis.enabled=true
+    {{- end -}}
+{{- end -}}
+
```
absoludity, line 249: Heh - I like dealing with the typo, just not sure we should do so here in the chart in an ad-hoc way. Could we not just update the kubeapps-apis binary so that if any plugin fails or is not found, we print out the names of all the plugins available in the plugin dirs? (We already have code to list those, I tihnk). That would deal with everything except the hint to also enable redis, but that's a temporary thing anyway (in that, we  should update so that it is enabled if you are using fluxv2).

Anyway, leaving a +1 in case you have a strong reason to prefer it here in the chart.

absoludity (approved):

chart/kubeapps/templates/_helpers.tpl
absoludity: Also realised your validation here will trigger *before* the chart deploys, which is a bit nicer from a UX POV.

antgamdia: Currently, we are just adding the plugin if it exists, but doing nothing if it does not. I agree the kubeapps-apis bin should deal with this situation; likely listing the available plugin list could make the trick.

However, as you pointed out later, this requires the chart to be deployed and someone to have a look at the logs. So, in practice, I would have faced the same issue (wondering why it isn't working! what have I broken?) until I would have realized it was an issue with a plugin.
So, even if doing these checks in the templating step is kind of dodgy, I still believe it is useful for our clumsy days :P

PR #4236, opened 2022-02-04
Add same workaround to show repo in flux

### LINKED ISSUE #4231: [fluxv2] Tag HelmRepository name in Catalog
### Description:

I am planing to use only fluxv2 plugin.
I have two different Airflow chart in two different `HelmRepository`(let's say one for prod, one for dev).
Is it possible to show `HelmRepository` name which chart is coming from which?
Just like helm plugin that shows the name of the repo.

스크린샷 2022-02-03 오후 10 35 24

Is there any option that I missed to enable this feature?
Please guide me if so.

I really appreciate your effort!

antgamdia:
### Description of the change

This simple PR just adds the same workaround to get the repository for those packages coming from the flux plugin.
Until we came up with a better solution, I think it solves the short-term problem for some users.

### Benefits

Repo will be displayed in pkgs from Fluxv2

### Possible drawbacks

It's just a short-term solution, but so was the Helm one as well.

### Applicable issues

- fixes #4231

### Additional information

image

### DIFF
--- a/dashboard/src/components/Catalog/PackageCatalogItem.tsx
+++ b/dashboard/src/components/Catalog/PackageCatalogItem.tsx
@@ -23,11 +23,8 @@ export default function PackageCatalogItem(props: IPackageCatalogItem) {
 
   // Get the pkg repository for the plugins that have one.
   switch (availablePackageSummary.availablePackageRef?.plugin?.name) {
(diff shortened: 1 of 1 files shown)


absoludity (approved):
Sweet, thanks Antonio.

Issue #4119, opened 2022-01-19
httpclient library SetClientTLS cleanup + unit tests


gfichtenholt:
Documenting leftover items for flux plugin per Pepe's request so we can prioritize it against other work

This is a (small) leftover item from #3899
I had to enhance some functionality in our httpclient library: 
https://github.com/kubeapps/kubeapps/blob/master/pkg/http-client/httpclient.go
to do with adding new ways to use TLS config. It was done in a hurry because I was rushing to get a much larger change merged. Now is a good time to go back and clean up a couple of items:
1) fix up https://github.com/kubeapps/kubeapps/blob/50d979f26975c54d953e36edffd2fcf2a36a21e8/pkg/http-client/httpclient.go#L138 to fix up the  func signature to be cleaner (marked by a TODO comment)
2) add unit tests for new functionality

Issue #4112, opened 2022-01-19
flux plugin: switch to using flux native data types instead of unstructured 


gfichtenholt:
Documenting leftover items for flux plugin per Pepe's request so we can prioritize it against other work

This is a leftover item from #3899

@antgamdia  said:
Not now, but perhaps we can use the typed struct instead of dealing with the unstructured data directly (as I'm doing in the kapp plugin).

This has been on my TODO list for a while. To be clear. The benefit to end user is none. But the benefit to developers/maintainers of this code is huge. Code based on unstructured types is difficult to work with/maintain and is error prone. Not to mention very verbose. 
Ideally, I'd want to switch to use APIs like https://github.com/fluxcd/source-controller/tree/main/api/v1beta1 which should make the code much more concise and easier to read.

antgamdia:
Totally agree, there is no direct user-faced benefit, but from a developer pov, the difference is huge. For the record, find below a screenshot comparing the two approaches:

![image](https://user-images.githubusercontent.com/11535726/150106058-f8becc07-191b-4eff-bfeb-9c03034d7451.png)

gfichtenholt:
Just FYI, I did see what Antonio did with kapp_controller and I am hoping to take things a bit further and least in flux plugin. In kapp_controller there are constant conversions from unstructured to typed (`runtime.DefaultUnstructuredConverter.FromUnstructured`) and back (`runtime.DefaultUnstructuredConverter.ToUnstructured`)  due to having to work with `dynamic.Interface` behind the scenes to access server resources. dynamic.Interface forces one to use unstructured, therefore constant back and forth. Yes, what is there now a reasonable first step. But, for flux, I am hoping to eventually get rid of dependency on `dynamic.Interface` in production code and tests and replace it with something like https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.11.0/pkg/client#Client which will let me work with typed objects all the way from server to caller and back. This is a bit of work though. A lot of existing code (production and tests) will need to be refactored, but I feel it's worth the effort.

absoludity:
Excellent, that'd be great to be able to use a typed client without having to use the same client-go version as the upstream go lib (carvel/flux - that was what stopped us initially using the typed client for carvel, from memory, needing to update our own go.mod deps to match their version).

gfichtenholt:
still work to do...

PR #4238, opened 2022-02-05
fix for httpclient library SetClientTLS cleanup + unit tests #4119 & finish  flux plugin: switch to using flux native data types instead of unstructured #4112 

gfichtenholt:
plus a couple of un-related minor TODO clean up items in flux plugin unit tests leftover from previous PRs.
Update: had some extra time while CI was being repaired so I also finished the remainder of  flux plugin: switch to using flux native data types instead of unstructured #4112


cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/release.go
```diff
@@ -68,6 +68,13 @@ func (s *Server) listReleasesInCluster(ctx context.Context, namespace string) ([
 		return nil, err
 	}
 
+	// TODO (gfichtenholt)
+	// 1) we need to make sure that .List() always returns results in the
+	// same order, otherwise pagination is broken (duplicates maybe returned and some entries
+	// missing).
```
absoludity, line 74: I'm still not clear on this: isn't your point (1) fixed by point (2)? That is, AIUI, if we update to pass the token used for pagination from the k8s api server, then we shouldn't be getting any duplicates or missing entries even when there is churn on the data level (ie. new HelmReleases or deleted ones).

Ah, or is the issue that the plugin is currently fetching all the objects and then returning just a subset for the paginated results of our server (in which case, we won't have the snapshot consistency, since each request to the kubeapps API will be getting results from a different snapshot?

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/release_test.go
```diff
@@ -329,14 +250,144 @@ func TestGetInstalledPackageSummaries(t *testing.T) {
...
 				t.Errorf("there were unfulfilled expectations: %s", err)
 			}
 		})
 	}
 }
 
+func TestGetInstalledPackageSummariesWithPagination(t *testing.T) {
+	// one big test case that can't really be broken down to smaller cases because
+	// the tests aren't independent/idempotent: there is state that needs to be
+	// kept track from one call to the next
```
absoludity, line 263: Yep, makes sense, and the test itself is nice and clear.

pkg/http-client/httpclient_test.go
```diff
@@ -102,6 +111,31 @@ func TestSetClientTls(t *testing.T) {
...
+		if resp.StatusCode != 200 {
+			t.Fatalf("expected OK, got: %d", resp.StatusCode)
+		}
+
+		payload, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			t.Fatal(err)
+		} else if got, want := payload, expectedPayload; !cmp.Equal(got, want) {
+			t.Errorf("mismatch (-want +got):\n%s", cmp.Diff(want, got))
+		}
```
absoludity, line 138: Nice, testing the actual server configured with the TLS.

absoludity (approved):
Excellent, thanks Greg.

PR #4243, opened 2022-02-07
E2e tests

castelblanque:
### Description of the change

This PR tries to fix problems with CI E2E tests in both local Kind and GKE environments.
It adds some logging of the parameters used in the `e2e-test.sh script`.
One of the problems was in resources created by test `03` being unable to have images pulled due to a `401 Unauthorized` error. Now the existing Docker credentials are used, but still the secret is created in the test.

More comments below.

### Benefits

E2E tests should pass successfully now in both local Kind and GKE.

### Possible drawbacks

No one knows what is the next surprise with E2E tests...

### Applicable issues

- Related to #2898

PR #4246, opened 2022-02-08
CI: Don't overwrite `/app/reports` directory with html reporter.

absoludity:

### Description of the change

This PR doesn't fix any failing tests (see Rafa's #4243 for that), but rather solves an issue of the disappearing local screenshots and videos.

In the end, it was simply that the 1.0.0 image had the html reporter enabled and set to write to the *same* directory as the test reports. Turns out the reporter *overwrites* the directory. So the screenshots and videos were being written and saved, then at the end the html report was generated and overwrote the directory.

This doesn't happen by default, because the [default output directory used for automatic screenshots etc. is `test-results`](https://playwright.dev/docs/test-configuration#automatic-screenshots) and the [default output for the html reporter is `playwright-report`](https://playwright.dev/docs/test-reporters#html-reporter). But on the 1.0.0 image, both were set to the same `reports/`.

I've enabled the reporter, assuming that is the intention given it's enabled on the 1.0.0 image (though I noticed it's commented out on the 1.0.1 image, but perhaps it was for this reason). I'll let you push the 1.0.2 image if you approve Rafa.

### Benefits

We can view the screenshots and videos as well as the html report.

### Possible drawbacks

None.

### Applicable issues

- fixes #

### Additional information

How did I find the cause? I wrote a new test which created explicit screenshots to different paths, and noted that even then, all would be written fine except the ones I added to the `/app/reports` directory, which were missing. I then touched some files in that directory and re-ran my test, noting that those files disappeared when the test ran, so something was wiping the directory clean. Then it was just figuring out what was causing it.

### DIFF
--- a/docs/developer/end-to-end-tests.md
+++ b/docs/developer/end-to-end-tests.md
@@ -72,7 +72,7 @@ kubectl apply -f manifests/executor.yaml
 pod=$(kubectl get po -l run=integration -o jsonpath="{.items[0].metadata.name}")
 
 # Copy latest tests
-kubectl cp ./use-cases ${pod}:/app/
+kubectl cp ./tests ${pod}:/app/
 
 # If you also modify the test configuration, you will need to update the files
 # for f in *.js; do   kubectl cp "./${f}" "${pod}:/app/"; done
--- a/integration/playwright.config.js
+++ b/integration/playwright.config.js
@@ -25,7 +25,7 @@ const config = {
       use: { ...devices["Desktop Chrome"] },
     },
   ],
-  //reporter: [ ['list'], ['html', { open: 'never', outputFolder: 'reports/' }] ],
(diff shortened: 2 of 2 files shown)


castelblanque:
Awesome Michael, thanks!

> I noticed it's commented out on the 1.0.1 image

Yeah, don't remember very well, but I disabled it due to causing some problems. Hadn't had the time to properly tackle it.

castelblanque:
BTW if we want the HTML report in CircleCI we should add the `integration/reports-html` in the step of storing artifacts. [Here](https://github.com/kubeapps/kubeapps/blob/main/.circleci/config.yml#L584) and [here](https://github.com/kubeapps/kubeapps/blob/main/.circleci/config.yml#L549).

castelblanque:
> I'll let you push the 1.0.2 image if you approve

Oki, will do.

castelblanque (approved):

PR #4247, opened 2022-02-08
Enable showing the progression of a pending installation.

absoludity:
### Description of the change

Currently, when a flux2 package is installed, the installed package is created but remains in the pending state until the reconciliation completes. As it was, the ApplicationStatus component did not display the pie chart while the app was in the pending state and the app.status never updates in the UX, yet we have all the data required to display the pie chart at this point (ie. we can fetch the resources already etc.)

There is a second issue which I'll deal with in a fdlow-up PR, in that the app status never updates in the view. It seems we stream the resources and so get all the updates there, but for the straight helm case, there was never a need to re-fetch the app (the helm release). The only effect of this in the UX currently is that for this flux case, where a newly installed package is initially in the pending status, the buttons (upgrade/delete) never re-enable.

### Benefits

When installing a package with flux, we now see the pie chart as soon as the resources have been fetched and can see it progress to completion.

### Possible drawbacks

I need to check whether we use this pending state for the straight helm case. From memory we may see it in the UX when there are pre-install hooks, although, if that's the case, I'd expect you'd need to refresh the page (due to the issue above of the app status not updating in the UX). Even if so, I then suspect we'd simply display "No resources found" until the state changed in helm itself and the resources were created. Hopefully CI will show whether it still works.

### Applicable issues

- Ref #3695


dashboard/src/components/ApplicationStatus/ApplicationStatus.tsx
```diff
@@ -134,10 +134,15 @@ export default function ApplicationStatus({
       </div>
     );
   }
-  if (info && info.status) {
+  if (info?.status?.reason) {
     // If the status code is different than "Deployed", display that status
```
castelblanque, line 138: Shouldn't we update this comment to include the pending status added, or it is not related to the `if` condition modified?

castelblanque:
Great to have better UI feedback, thanks!

castelblanque (approved):

dashboard/src/components/ApplicationStatus/ApplicationStatus.tsx
absoludity: Woops, yep, I'll update it. Thanks.

PR #4256, opened 2022-02-10
Trigger fetch of package status after resources are updated.

absoludity:

### Description of the change

Fixes an issue whereby a package installs successfully but the upgrade/delete buttons remain disabled.

With this change, we refresh the package status whenever a bunch of resources are processed.

### Benefits

When using Flux plugin to install a package, the buttons enable once the package is reconciled.

### Applicable issues

- ref #3695

### DIFF
--- a/dashboard/src/actions/installedpackages.test.tsx
+++ b/dashboard/src/actions/installedpackages.test.tsx
@@ -7,9 +7,11 @@ import {
   InstalledPackageSummary,
   GetInstalledPackageSummariesResponse,
   InstalledPackageReference,
+  InstalledPackageStatus,
   VersionReference,
   ReconciliationOptions,
   ResourceRef,
+  InstalledPackageStatus_StatusReason,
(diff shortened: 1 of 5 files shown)


antgamdia (approved):
Awesome, given that we are already fetching the info periodically, it's a great opportunity to also update the pkg status. Thanks!

Also, I'd say we may want something similar in the "Applications" view. Once they are loaded, we won't get refreshed anymore (so they could have failed or reconciled without, but the displayed status will be still the same)

absoludity:
> Awesome, given that we are already fetching the info periodically, it's a great opportunity to also update the pkg status. Thanks!
> 
> Also, I'd say we may want something similar in the "Applications" view. Once they are loaded, we won't get refreshed anymore (so they could have failed or reconciled without, but the displayed status will be still the same)

Hmm... less of a priority, but yes, we could update tiles so that those in non-final states can refresh (though even then, there's no "final" state for flux/carvel where they can continually update the installed app etc.)

PR #4257, opened 2022-02-10
Ensure only one of helm or fluxv2 plugins are configured.

absoludity:

### Description of the change

Currently it's possible to configure the kubeapps apis service with both the fluxv2 and helm plugins running. This is not something we are able to support right now, since it would enable interacting with the helm releases directly when they are being managed by flux (as well as confusingly display apps twice).

I've used the new chart validation that @antgamdia added recently, since it's nice to catch the configuration issue *before* doing the deployment.

### Benefits

We don't allow a configuration which we don't intend to support.

### Applicable issues

- ref #3695 

### Additional information

Next up, removing the need to enable redis (and removing postgres when helm plugin is not used? - possibly, depends on effects, may need to also remove assetsvc, apprepo controller etc.).

### DIFF
--- a/chart/kubeapps/Chart.lock
+++ b/chart/kubeapps/Chart.lock
@@ -1,12 +1,12 @@
 dependencies:
 - name: common
   repository: https://charts.bitnami.com/bitnami
-  version: 1.11.0
+  version: 1.11.1
 - name: postgresql
   repository: https://charts.bitnami.com/bitnami
   version: 10.16.2
 - name: redis
   repository: https://charts.bitnami.com/bitnami
   version: 15.7.6
-digest: sha256:4102035f8d115bb5ea6d4bc51767abbee940de90425077245ecc8117567561bf
-generated: "2022-02-01T17:58:10.326513+01:00"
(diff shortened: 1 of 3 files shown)


chart/kubeapps/Chart.lock
```diff
@@ -1,12 +1,12 @@
 dependencies:
 - name: common
   repository: https://charts.bitnami.com/bitnami
-  version: 1.11.0
+  version: 1.11.1
```
absoludity, line 4: I was getting the following error until I updated the deps:


> Error: UPGRADE FAILED: template: kubeapps/templates/ingress.yaml:54:42: executing "kubeapps/templates/ingress.yaml" at <include "common.ingress.certManagerRequest" (dict "annotations" .Values.ingress.annotations)>: error calling include: template: no template "common.ingress.certManagerRequest" associated with template "gotpl"

antgamdia (approved):
Great, it's good to prevent some undesired combinations before actually installing the release.
(btw, I've to check again my IDE config wrt the trailing spaces :P)

PR #4258, opened 2022-02-10
Fix for test 01 aimed at GKE

castelblanque:

### Description of the change

Fixes the E2E test number 01 that is failing in GKE but not on Kind.

### Benefits

Test should not fail now in GKE.

### Possible drawbacks

Hopefully none

### DIFF
--- a/integration/tests/main/01-missing-permissions.spec.js
+++ b/integration/tests/main/01-missing-permissions.spec.js
@@ -28,6 +28,8 @@ test.describe("Limited user simple deployments", () => {
     await releaseNameLocator.fill(utils.getRandomName("test-01-release"));
     await page.locator('cds-button:has-text("Deploy")').click();
 
+    // Assertions
+    await page.waitForSelector(".alert-items .alert-text");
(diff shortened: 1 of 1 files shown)


integration/tests/main/01-missing-permissions.spec.js
```diff
@@ -28,6 +28,8 @@ test.describe("Limited user simple deployments", () => {
     await releaseNameLocator.fill(utils.getRandomName("test-01-release"));
     await page.locator('cds-button:has-text("Deploy")').click();
 
+    // Assertions
+    await page.waitForSelector(".alert-items .alert-text");
```
antgamdia, line 32: So it was just a kind of race condition, that's why retrying made it to work?

castelblanque: Yeah, I would say that if the error message was appearing almost immediately, the problem was hidden. It makes total sense to have this `waitForSelector`, as neither `page.locator` nor `expect(errorLocator)` wait for the error to be there.

PR #4265, opened 2022-02-11
Enable specifying plugins that determine chart dependencies.

### LINKED ISSUE #4264: Better handling of configured plugins and their dependencies
### Description:

Our Kubeapps-APIs service can be configured to run with different plugins (currently `fluxv2`, `kapp-controller`, `helm` and `resources`), which enables those plugins to be loaded at runtime. The fluxv2 plugin depends on having redis available, as the helm plugin depends on postgres (as do other parts). The resources plugin is always required for Kubeapps.

Currently the kubeapps-apis service plugins are explicitly enabled by setting the list of `kubeappsapis.enabledPlugins`, while their dependencies, such as redis for fluxv2, need to be enabled separately. Or related to this, as the resources plugin is always required , users can currently mis-configure Kubeapps (eg. #4259 ). This should not be possible.

I'd like to update this so that instead, as a user, we can enable fluxv2 support for Kubeapps (I don't care about some kubeappsapis sub-service), and both the kubeapps-apis plugins as well as the redis enablement will be set automatically.

Helm subcharts can be conditionally included based on either a valid value in the `values.yaml` (eg. `featureFlags.fluxv2.enabled` or similar), or a tag. See [conditions and tags](https://helm.sh/docs/chart_best_practices/dependencies/#conditions-and-tags). The condition cannot be anything calculated (such as a helper).

As such, one solution would be that we remove the ability to configure the kubeapps-apis plugins directly, and instead expose something like the following at the top level of the chart values:

```
plugins:
  helm:
    enabled: true
  flux:
    enabled: false
  carvel:
    enabled: false
```
(note that this means we can expose the kapp-controller plugin simply as `carvel` as we previously wanted to... so users can "enable carvel support in Kubeapps", or similarly, flux support and we enable the fluxv2 plugin). 

This would enable us to:
1. automatically include any required subcharts conditionally,
2. calculate the actual plugins required for kubeapps-apis, in addition to the required resources plugin,
3. provide simple user feature names (eg. "carvel" or "flux").
4. we could also move the configuration to this top level also?

Thoughts?

absoludity:

@batiati If approved, feel free to land this and include in the release if you want.

### Description of the change

This PR has two related changes:
1. Create a user-centric `packaging` option in the values for specifying whether Kubeapps should run with helm, flux or carvel packaging. Defaulting to helm of course.
2. Uses these new boolean values to control whether the Redis chart dependency is installed, so that we no longer need to manually specify `redis.enabled` (removed that from the chart, since it's no longer used in the chart dependency condition).

```
# Default plugins are helm and resources:
➜  kubeapps git:(3695-dont-require-explicit-enable-redis-2) ✗ git --no-pager diff docs/user/manifests/kubeapps-local-dev-values.yaml
➜  kubeapps git:(3695-dont-require-explicit-enable-redis-2) ✗ make deploy-dev-kubeapps | grep "\- /kubeapps-apis" -A 5              
            - /kubeapps-apis
          args:
            - --plugin-dir
            - /plugins/helm
            - --plugin-dir
            - /plugins/resources

# Enabling carvel updates the enabled plugins.
➜  kubeapps git:(3695-dont-require-explicit-enable-redis-2) ✗ git --no-pager diff docs/user/manifests/kubeapps-local-dev-values.yaml
diff --git a/docs/user/manifests/kubeapps-local-dev-values.yaml b/docs/user/manifests/kubeapps-local-dev-values.yaml
index 3b7518cc..ff73074b 100644
--- a/docs/user/manifests/kubeapps-local-dev-values.yaml
+++ b/docs/user/manifests/kubeapps-local-dev-values.yaml
@@ -30,3 +30,6 @@ ingress:
     # Required for ingress-nginx 1.0.0 for a valid ingress.
     kubernetes.io/ingress.class: nginx
+packaging:
+  carvel:
+    enabled: true
➜  kubeapps git:(3695-dont-require-explicit-enable-redis-2) ✗ make deploy-dev-kubeapps | grep "\- /kubeapps-apis" -A 7
            - /kubeapps-apis
          args:
            - --plugin-dir
            - /plugins/kapp-controller
            - --plugin-dir
            - /plugins/helm
            - --plugin-dir
            - /plugins/resources

# Enabling flux results in an error as helm is also enabled:
➜  kubeapps git:(3695-dont-require-explicit-enable-redis-2) ✗ git --no-pager diff docs/user/manifests/kubeapps-local-dev-values.yaml
diff --git a/docs/user/manifests/kubeapps-local-dev-values.yaml b/docs/user/manifests/kubeapps-local-dev-values.yaml
index 3b7518cc..7117cbaa 100644
--- a/docs/user/manifests/kubeapps-local-dev-values.yaml
+++ b/docs/user/manifests/kubeapps-local-dev-values.yaml
@@ -30,3 +30,8 @@ ingress:
     # Required for ingress-nginx 1.0.0 for a valid ingress.
     kubernetes.io/ingress.class: nginx
+packaging:
+  carvel:
+    enabled: true
+  flux:
+    enabled: true
➜  kubeapps git:(3695-dont-require-explicit-enable-redis-2) ✗ make deploy-dev-kubeapps | grep "\- /kubeapps-apis" -A 7              
Error: execution error at (kubeapps/templates/kubeappsapis/deployment.yaml:71:35): packaging: Please enable only one of the flux and helm plugins, since they both operate on Helm releases.

Use --debug flag to render out invalid YAML
make: *** [script/makefiles/deploy-dev.mk:40: deploy-dev-kubeapps] Error 1

# Disabling helm gets things going as expected:
➜  kubeapps git:(3695-dont-require-explicit-enable-redis-2) ✗ git --no-pager diff docs/user/manifests/kubeapps-local-dev-values.yaml
diff --git a/docs/user/manifests/kubeapps-local-dev-values.yaml b/docs/user/manifests/kubeapps-local-dev-values.yaml
index 3b7518cc..ba928fcc 100644
--- a/docs/user/manifests/kubeapps-local-dev-values.yaml
+++ b/docs/user/manifests/kubeapps-local-dev-values.yaml
@@ -30,3 +30,10 @@ ingress:
     # Required for ingress-nginx 1.0.0 for a valid ingress.
     kubernetes.io/ingress.class: nginx
+packaging:
+  carvel:
+    enabled: true
+  flux:
+    enabled: true
+  helm:
+    enabled: false
➜  kubeapps git:(3695-dont-require-explicit-enable-redis-2) ✗ make deploy-dev-kubeapps | grep "\- /kubeapps-apis" -A 7              
            - /kubeapps-apis
          args:
            - --plugin-dir
            - /plugins/kapp-controller
            - --plugin-dir
            - /plugins/fluxv2
            - --plugin-dir
            - /plugins/resources

# Adding some other key to the packaging value causes a fail:
Use --debug flag to render out invalid YAML
make: *** [script/makefiles/deploy-dev.mk:40: deploy-dev-kubeapps] Error 1
➜  kubeapps git:(3695-dont-require-explicit-enable-redis-2) ✗ git --no-pager diff docs/user/manifests/kubeapps-local-dev-values.yaml
diff --git a/docs/user/manifests/kubeapps-local-dev-values.yaml b/docs/user/manifests/kubeapps-local-dev-values.yaml
index 3b7518cc..c3a43886 100644
--- a/docs/user/manifests/kubeapps-local-dev-values.yaml
+++ b/docs/user/manifests/kubeapps-local-dev-values.yaml
@@ -30,3 +30,12 @@ ingress:
     # Required for ingress-nginx 1.0.0 for a valid ingress.
     kubernetes.io/ingress.class: nginx
+packaging:
+  carvel:
+    enabled: true
+  flux:
+    enabled: true
+  helm:
+    enabled: false
+  unknownpkg:
+    enabled: true
➜  kubeapps git:(3695-dont-require-explicit-enable-redis-2) ✗ make deploy-dev-kubeapps | grep "\- /kubeapps-apis" -A 7              
Error: execution error at (kubeapps/templates/kubeappsapis/deployment.yaml:71:35): packaging: Unsupported packaging option: unknownpkg

Use --debug flag to render out invalid YAML
make: *** [script/makefiles/deploy-dev.mk:40: deploy-dev-kubeapps] Error 1

# Manually overriding the enabled plugins is still possible:
➜  kubeapps git:(3695-dont-require-explicit-enable-redis-2) ✗ git --no-pager diff docs/user/manifests/kubeapps-local-dev-values.yaml
diff --git a/docs/user/manifests/kubeapps-local-dev-values.yaml b/docs/user/manifests/kubeapps-local-dev-values.yaml
index 3b7518cc..e1fe4b91 100644
--- a/docs/user/manifests/kubeapps-local-dev-values.yaml
+++ b/docs/user/manifests/kubeapps-local-dev-values.yaml
@@ -18,6 +18,9 @@ postgresql:
   existingSecret: postgresql-db
 kubeappsapis:
   replicaCount: 1
+  enabledPlugins:
+  - unknownpkg
+  - anotherunknownpkg
 ingress:
   enabled: true
   hostname: localhost
@@ -30,3 +33,12 @@ ingress:
     # Required for ingress-nginx 1.0.0 for a valid ingress.
     kubernetes.io/ingress.class: nginx
+packaging:
+  carvel:
+    enabled: true
+  flux:
+    enabled: true
+  helm:
+    enabled: false
+  unknownpkg:
+    enabled: true
➜  kubeapps git:(3695-dont-require-explicit-enable-redis-2) ✗ make deploy-dev-kubeapps | grep "\- /kubeapps-apis" -A 7              
            - /kubeapps-apis
          args:
            - --plugin-dir
            - /plugins/unknownpkg
            - --plugin-dir
            - /plugins/anotherunknownpkg
            - --clusters-config-path=/config/clusters.conf
            - --plugin-config-path=/config/kubeapps-apis/plugins.conf
```
### Benefits

Users don't need to care about "plugins" and ensuring the names are correct nor ensuring the resources plugin has been enabled etc.

Checks automatically that at least 1 packaging plugin has been enabled. Also continues to ensure that you can't run with flux and helm together.

Still retains the option to explicitly set the `kubeappsapis.enabledPlugins` if the user really wants to (in case of new or in-development plugins etc).

### Applicable issues

- fixes #4264

### DIFF
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -1867,18 +1867,12 @@ kubeappsapis:
 ## @section Redis&trade; chart configuration
 ## ref: https://github.com/bitnami/charts/blob/master/bitnami/redis/values.yaml
 ##
+## Redis will be enabled and installed if `packages.flux.enabled` is true.
 redis:
   ## @param redis.redisPassword Password used in Redis&trade;
   ## ref: https://github.com/bitnami/bitnami-docker-redis/blob/master/README.md#setting-the-server-password-on-first-run
   ##
   redisPassword: ""
-  ## @param redis.enabled Enable the Redis&trade; deployment when deploying Kubeapps APIs.
-  ## We currently have the situation that Redis is required for the fluxv2 plugin only.
-  ## Until such a point that we're releasing with the fluxv2 plugin enabled, or the
-  ## plugin cache support has been generalised so all plugins use Redis, we'll need
-  ## to manually enable this in dev while ensuring it is false for releases (as it
-  ## is a conditional dependency in the Chart.yaml).
-  enabled: false
   master:
     ## @param redis.master.extraFlags Array with additional command line flags for Redis&trade; master
     extraFlags:
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -136,10 +136,13 @@ ingress:
   ##
   secrets: []
 
-## @section Kubeapps plugin options
+## @section Kubeapps packaging options
 ## Note: the helm and flux plugins are mutually exclusive, you can only
 ## enable one or the other since they both operate on Helm release objects.
-plugins:
+## Enabling carvel or flux does *not* install the required related Carvel or
+## Flux controllers on your cluster. Please read the documentation for running
+## Kubeapps with Carvel or Flux support.
+packaging:
   ## Default helm plugin
   ## @param plugins.helm.enabled Enable the standard helm plugin.
   helm:
@@ -1626,8 +1629,13 @@ postgresql:
 
 ## @section kubeappsapis parameters
 kubeappsapis:
-  ## @param kubeappsapis.enabledPlugins Enabled plugins for the Kubeapps-APIs service
-  ## DEPRECATED: use the top-level `plugins` value instead.
+  ## @param kubeappsapis.enabledPlugins Manually override which plugins are
+  ## enabled for the Kubeapps-APIs service
+  ##
+  ## NOTE: normally this should remain blank, with the top-level `packaging`
+  ## value automatically determining which plugins should be enabled. Only
+  ## set this value if you want to manually override the list of plugins
+  ## enabled for the service.
   ##
   enabledPlugins:
   pluginConfig:
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -143,16 +143,16 @@ ingress:
 ## Flux controllers on your cluster. Please read the documentation for running
 ## Kubeapps with Carvel or Flux support.
 packaging:
-  ## Default helm plugin
-  ## @param plugins.helm.enabled Enable the standard helm plugin.
+  ## Default helm packaging
+  ## @param packaging.helm.enabled Enable the standard Helm packaging.
   helm:
     enabled: true
-  ## Carvel plugin
-  ## @param plugins.carvel.enabled Enable the carvel plugin.
+  ## Carvel packaging
+  ## @param packaging.carvel.enabled Enable support for the Carvel (kapp-controller) packaging.
   carvel:
     enabled: false
-  ## Flux (v2) plugin
-  ## @param plugins.flux.enabled Enable the Flux plugin.
+  ## Flux (v2) packaging
+  ## @param packaging.flux.enabled Enable support for Flux (v2) packaging.
   flux:
     enabled: false
 
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -136,6 +136,23 @@ ingress:
   ##
   secrets: []
 
+## @section Kubeapps plugin options
+## Note: the helm and flux plugins are mutually exclusive, you can only
+## enable one or the other since they both operate on Helm release objects.
+plugins:
+  ## Default helm plugin
+  ## @param plugins.helm.enabled Enable the standard helm plugin.
+  helm:
+    enabled: true
+  ## Carvel plugin
+  ## @param plugins.carvel.enabled Enable the carvel plugin.
+  carvel:
+    enabled: false
+  ## Flux (v2) plugin
+  ## @param plugins.flux.enabled Enable the Flux plugin.
+  flux:
+    enabled: false
+
 ## @section Frontend parameters
 
 ## Frontend parameters
@@ -1610,15 +1627,9 @@ postgresql:
 ## @section kubeappsapis parameters
 kubeappsapis:
   ## @param kubeappsapis.enabledPlugins Enabled plugins for the Kubeapps-APIs service
-  ## e.g:
-  ## enabledPlugins:
-  ## - helm
-  ## - fluxv2
-  ## - kapp-controller
+  ## DEPRECATED: use the top-level `plugins` value instead.
   ##
   enabledPlugins:
-    - helm
-    - resources
   pluginConfig:
     core:
       packages:
@@ -1641,7 +1652,7 @@ kubeappsapis:
           ## @param kubeappsapis.pluginConfig.kappController.packages.v1alpha1.defaultPrereleasesVersionSelection Default policy for allowing prereleases containing one of the identifiers
           ## ref: https://carvel.dev/kapp-controller/docs/latest/package-consumer-concepts/#prereleases
           ## e.g:
-          # defaultPrereleasesVersionSelection: 
+          # defaultPrereleasesVersionSelection:
           # - rc
           defaultPrereleasesVersionSelection: null
           ## @param kubeappsapis.pluginConfig.kappController.packages.v1alpha1.defaultAllowDowngrades Default policy for allowing applications to be downgraded to previous versions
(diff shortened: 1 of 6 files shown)


chart/kubeapps/values.yaml
```diff
@@ -136,10 +136,13 @@ ingress:
...
+## @section Kubeapps packaging options
 ## Note: the helm and flux plugins are mutually exclusive, you can only
 ## enable one or the other since they both operate on Helm release objects.
-plugins:
+## Enabling carvel or flux does *not* install the required related Carvel or
+## Flux controllers on your cluster. Please read the documentation for running
+## Kubeapps with Carvel or Flux support.
+packaging:
   ## Default helm plugin
   ## @param plugins.helm.enabled Enable the standard helm plugin.
```
antgamdia, line 147:
```suggestion
  ## @param packaging.helm.enabled Enable the standard Helm plugin.
```

chart/kubeapps/values.yaml
```diff
@@ -136,6 +136,26 @@ ingress:
...
+## Enabling carvel or flux does *not* install the required related Carvel or
+## Flux controllers on your cluster. Please read the documentation for running
+## Kubeapps with Carvel or Flux support.
+packaging:
+  ## Default helm plugin
+  ## @param plugins.helm.enabled Enable the standard helm plugin.
+  helm:
+    enabled: true
+  ## Carvel plugin
+  ## @param plugins.carvel.enabled Enable the carvel plugin.
```
antgamdia, line 151:
```suggestion
  ## @param packaging.carvel.enabled Enable the Carvel (kapp-controller) plugin.
```

chart/kubeapps/values.yaml
```diff
@@ -136,6 +136,26 @@ ingress:
...
+  carvel:
+    enabled: false
+  ## Flux (v2) plugin
+  ## @param plugins.flux.enabled Enable the Flux plugin.
```
antgamdia, line 155:
```suggestion
  ## @param packaging.flux.enabled Enable the Flux (v2) plugin.
```

docs/user/managing-carvel-packages.md
```diff
@@ -72,19 +72,17 @@ This `kapp-controller` plugin is currently being built by default in the Kubeapp
 
 > **TIP**: Please refer to the [getting started documentation](./getting-started.md) for more information on how to install Kubeapps and pass custom configuration values.
 
-In the [values.yaml](../../chart/kubeapps/values.yaml) file, under `kubeappsapis.enabledPlugins` add
-`kapp-controller` to the list of enabled plugins. For example:
+In the [values.yaml](../../chart/kubeapps/values.yaml) file, enable the `packaging.carvel` option (and disable helm, depending on your needs):
```
antgamdia, line 75: Not sure if we should disable helm in this example, as newcomer users could think both plugins aren't getting along with each other.
I'd suggest just enabling carvel, but no strong opinion though

antgamdia (approved):
Great! Much better now, this is a less error-prone UX for enabling the plugins. Thanks!

chart/kubeapps/values.yaml
absoludity: With the above 3 changes, I additionally updated to avoid the word "plugin" at this level. ie. "Enable support for Flux (v2) packaging."

docs/user/managing-carvel-packages.md
absoludity: Yeah, I wasn't saying people had to disable helm, just that it may be their intention. But yep, perhaps unnecessary. Removed.

PR #4277, opened 2022-02-14
Autologin if a token is passed as queryParam

antgamdia:
### Description of the change

This PR just allows passing a `?token=xxxxxx` in the token login view so that we can bypass manually entering a value.

### Benefits

No human interaction for accessing Kubeapps with a token, if desired.

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

x

Animation1


absoludity (approved):
Nice.

PR #4290, opened 2022-02-16
Add step in the release docs

antgamdia:
### Description of the change

Due to our recent problems with the release, this PR just adds another step to ensure we double-check our new dockerfiles against the latest release's ones. 

### Benefits

We (hopefully) won't forget this step anymore when releasing.

### Possible drawbacks

Still manual and therefore error-prone, but... nothing more we can do :S

### Applicable issues

N/A

### Additional information

N/A

### DIFF
--- a/docs/developer/release-process.md
+++ b/docs/developer/release-process.md
@@ -9,7 +9,7 @@ It consists of four main stages: update the development images, update the CI, u
 
 ### 0.1 - Development images
 
-For building the [development container images](https://hub.docker.com/u/kubeapps), a number of base images are used in the build stage Specifically:
+For building the [development container images](https://hub.docker.com/u/kubeapps), a number of base images are used in the build stage. Specifically:
 
 - The [dashboard/Dockerfile](../../dashboard/Dockerfile) uses:
   - [bitnami/node](https://hub.docker.com/r/bitnami/node/tags) for building the static files for production.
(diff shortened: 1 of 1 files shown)


docs/developer/release-process.md
```diff
@@ -244,12 +244,18 @@ Once the PR has been created, have a look at it (eg. remove any development chan
 
 > Some issues can arise here, so please check the app versions are being properly updated at once and ensure you have the latest changes in the PR branch. Note that the [bitnami-bot](https://github.com/bitnami-bot) usually performs some automated commits to the main branch that might collide with the changes in our PR. In particular, it will release a new version of the chart with the updated images.
 
-## 5 - Publish the GitHub release
+## 5 - Check Dockerfiles and notify the proper teams
+
+Eventually, as the Kubeapps code continues to evolve, some changes are often introduced in our own [development container images](https://hub.docker.com/u/kubeapps). However, those changes won't get released in the official Bitnami repository unless we manually notify the proper team to also include those changes in their building system.
+
+> As part of this release process, each Kubeapps component's Dockerfile _must_ compared against the one in the previous release. If they functionally differ each other, the Bitnami Content team _must_ be notified.
```
absoludity, line 251: Thanks Antonio. Can we even compare them, or is this something we need to ask the Bitnami content team to do? I tried yesterday to find the Dockerfile, but only found a Dockerfile which pulls a tar.gz and expands it... so quite opaque.

absoludity (approved):

docs/developer/release-process.md
antgamdia: We can compare them with the internal `gitlist` tool, but yep, we wouldn't find a `Dockerfile` as is, but a kind of "recipe" with their way to build the image. Then we would have to figure out if these recipe steps match our Dockerfile; so it is really a manual process likely to be performed by the proper team.

PR #4295, opened 2022-02-17
Update to use the same plugin directories that bitnami Dockerfile uses.

absoludity:

### Description of the change

Small follow-up to #4264 that updates the location that kubeapps-apis plugins are located (to match the Bitnami containers).

@castelblanque If you do want this change, just be sure to merge it first (as it's currently targeting #4264 and so will be merged into there, but if that PR is merged first to main, this will be automatically updated by GH to target main and may need rebasing, depending).

### Benefits

Consistency. Nice to explicitly identify packaging plugins.

### Applicable issues

- ref #3695

### DIFF
--- a/chart/kubeapps/templates/_helpers.tpl
+++ b/chart/kubeapps/templates/_helpers.tpl
@@ -230,11 +230,11 @@ kubeapps: ingress.tls
       {{- range $plugin, $options := .Values.packaging }}
         {{- if $options.enabled }}
           {{- if eq $plugin "carvel" }}
-            {{- $enabledPlugins = append $enabledPlugins "kapp-controller" }}
+            {{- $enabledPlugins = append $enabledPlugins "kapp-controller-packages" }}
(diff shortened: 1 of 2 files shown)


castelblanque:
@absoludity To fix the current issue with release v2.4.3 on Bitnami side, exactly the opposite changes are needed. Doing them in here would mean Bitnami would need to undo them for the next release. Is that what we want?
Otherwise we would need to restart all over the tagging of the release, etc.

castelblanque:
We agreed on going forward with the current state for v2.4.3, which means that Bitnami team will do the changes on their side for now.
Merging this PR now so it is in `main` for next release. We'll have to remember to notify Bitnami to undo their changes to match this PR's changes.
Thanks @absoludity!

absoludity:
Yep, sounds good .I'd missed that this would obviously require tagging a new image release for the kubeapps-apis server.

PR #4308, opened 2022-02-18
first step for  Investigate and propose package repositories API with similar core interface to packages API. #3496 

gfichtenholt:
First step for  Investigate and propose package repositories API with similar core interface to packages API. #3496 
Much work remains, I am aware. I just wanted to push the code before I am gone for a week on PTO


cmd/kubeapps-apis/proto/kubeappsapis/plugins/fluxv2/packages/v1alpha1/fluxv2.proto
```diff
@@ -82,59 +79,13 @@ service FluxV2PackagesService {
...
-  // The plugin used to interact with this package repository.
-  kubeappsapis.core.plugins.v1alpha1.Plugin plugin = 4;
-
-  // TODO: Other fields such as type, status... TBD.
+service FluxV2RepositoriesService {
+  // AddPackageRepository add an existing package repository to the set of ones already managed by the 
+  // 'fluxv2' plugin
+  rpc AddPackageRepository(kubeappsapis.core.packages.v1alpha1.AddPackageRepositoryRequest) returns (kubeappsapis.core.packages.v1alpha1.AddPackageRepositoryResponse) {
+    option (google.api.http) = {
+      post: "/core/packages/v1alpha1/repositories"
```
absoludity, line 87: The url here needs to be specific to the flux plugin like the others above?

cmd/kubeapps-apis/server/server.go
```diff
@@ -94,28 +94,17 @@ func Serve(serveOpts core.ServeOptions) error {
...
-	err = pluginsGRPCv1alpha1.RegisterPluginsServiceHandlerFromEndpoint(gwArgs.Ctx, gwArgs.Mux, gwArgs.Addr, gwArgs.DialOptions)
-	if err != nil {
-		return fmt.Errorf("failed to register core.plugins handler for gateway: %v", err)
-	}
-
 	// Ask the plugins server for plugins with GRPC servers that fulfil the core
 	// packaging v1alpha1 API, then pass to the constructor below.
 	// The argument for the reflect.TypeOf is based on what grpc-go
 	// does itself at:
 	// https://github.com/grpc/grpc-go/blob/v1.38.0/server.go#L621
```
absoludity, line 101: This comment needs to be moved with the line below into the function (it's documenting where the crazy reflect.TypeOf arg comes from)

absoludity (approved):
> micahel, if you get a chance pls review in the next couple of days. A lot of this is auto-generated. If not, no problem, i'll just get back to it after I am back from PTO. Thanks

Sorry - I don't generally check work stuff over the weekend unless there's an emergency situation or similar.

Great start! The proto stuff looks mostly according to spec, just see the note about the URL for the flux service RPC. Other tiny comment about a comment staying with the line to which it refers.

Thanks!

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/repositories.proto
```diff
@@ -0,0 +1,216 @@
...
+
+  // A user-provided name for the package repository (e.g. bitnami)
+  string name = 2;
+
+  // A user-provided description. Optional
+  string description = 3;
+
+  // Whether this repository is global or namespace-scoped. Optional.
+  // By default, the value is false, i.e. the repository is global
+  bool namespace_scoped = 4;
```
castelblanque, line 49: Currently in UI there is no such flag.
If the repo is created in the context of the global namespace, then it will be global. Why not using the same convention here? It might lead to confusion, e.g. repo is created with `namespace_scoped=false` but namespace in context is other than the global one. I guess logic in controller should throw some kind of error when `context` and `namespace_scoped` have inconsistent values?

absoludity: > Currently in UI there is no such flag. If the repo is created in the context of the global namespace, then it will be global. Why not using the same convention here?

I'll do my best to answer for Greg in his absence (PTO this week): Helm (and Carvel, for that matter) have the concept of a "global namespace", but it's not something that Flux has at all AIUI. So I think Greg is keen to design the API not to match the current UX assumptions necessarily, while still enabling the current functionality.

> It might lead to confusion, e.g. repo is created with `namespace_scoped=false` but namespace in context is other than the global one.

We (the kubeapps dashboard authors) can choose how to use this field (ie. not expose it to the user at all, but set it appropriately based on the namespace in context etc.) I think?

> I guess logic in controller should throw some kind of error when `context` and `namespace_scoped` have inconsistent values?

Yes, the helm plugin can handle this to match its expectations. Note, you can read more about the [decisions and questions for this API in the design doc](https://docs.google.com/document/d/1z03msZRGsRvcRaom-yrvEwIWcEfNy6fSA5Zg28gjYvA/edit#heading=h.u3lj7whdof7t).

gfichtenholt: Couldn't have put it better myself. Thanks Michael

castelblanque: ok, understood. Thanks!

cmd/kubeapps-apis/proto/kubeappsapis/plugins/fluxv2/packages/v1alpha1/fluxv2.proto
gfichtenholt: indeed. will fix

cmd/kubeapps-apis/server/server.go
gfichtenholt: yep. you got it

PR #4309, opened 2022-02-18
Fix Flux deployment for local dev env.

absoludity:

### Description of the change

I broke the Flux deployment in my recent commit because I focused on the enabling of the correct plugins and installation of the chart dependency, but didn't check it running IRL.

Turns out we had two places in the chart still referring to the old `redis.enabled`.

This fixes that,  and I just verified IRL that with this I can deploy flux locally :)

I'll update the upstream PR later (just going AFK).

### Applicable issues

- fixes #


antgamdia (approved):
Oops, when dealing with chart changes it's really easy to miss some params. Thanks for the fix!

PR #4326, opened 2022-02-22
Ensure consistent status in tile for different plugins.

absoludity:

### Description of the change

The status displayed in the application list was using the plugin-specific `status.userReason` field, resulting in the following when displaying both flux and carvel packages together:

status-reason-before

 rather than the `status.reason` field which is consistent across plugins, which results in:

status-reason-after

### Benefits

We avoid formatting issues with the plugin specific user reason on the tiles. Get standard status's across plugins for apps.

### Possible drawbacks

Changes the default from "deployed" to "installed". We could put a condition in there, but it seems appropriate to use `installed` (given our generic packaging wording) so left it. Let me know.

### Applicable issues

- ref #3695

### DIFF
--- a/dashboard/src/components/AppList/AppListItem.tsx
+++ b/dashboard/src/components/AppList/AppListItem.tsx
@@ -2,7 +2,11 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import Tooltip from "components/js/Tooltip";
-import { InstalledPackageSummary } from "gen/kubeappsapis/core/packages/v1alpha1/packages";
+import {
+  InstalledPackageSummary,
+  InstalledPackageStatus_StatusReason,
+  installedPackageStatus_StatusReasonToJSON,
+} from "gen/kubeappsapis/core/packages/v1alpha1/packages";
 import { getPluginIcon } from "shared/utils";
 import placeholder from "../../placeholder.png";
 import * as url from "../../shared/url";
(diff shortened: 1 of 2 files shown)


castelblanque:
Nice, now it looks cleaner. Thanks!

> it seems appropriate to use installed (given our generic packaging wording)

Agree, though at some points Kubeapps still shows the "Deploy" term, like
image
We'll need to change that to be in line.

castelblanque (approved):

dashboard/src/components/AppList/AppListItem.tsx
```diff
@@ -14,10 +18,19 @@ export interface IAppListItemProps {
   cluster: string;
 }
 
+function getAppStatusLabel(
+  statusReason: InstalledPackageStatus_StatusReason = InstalledPackageStatus_StatusReason.STATUS_REASON_UNSPECIFIED,
+): string {
+  // The JSON versions of the reasons are forced to follow the standard
+  // pattern STATUS_REASON_<reason> by buf.
+  const jsonReason = installedPackageStatus_StatusReasonToJSON(statusReason);
+  return jsonReason.replace("STATUS_REASON_", "").toLowerCase();
```
antgamdia, line 27: Neat approach! Thanks for the comment, good to know!

antgamdia (approved):

PR #4331, opened 2022-02-23
Update empty catalog message to include docs for flux/carvel

absoludity:

### Description of the change

Just adds an extra paragraph to the empty catalog display page, linking directly to our docs:

empty-catalog

### Benefits

A user has more information about where to go if they see an empty catalog.

### Applicable issues

- ref #3695 

### Additional information

I was originally planning to check which plugins are configured and just show info related to that plugin, but it looks to be more work than it is worth given that we're currently investing in the app repository API (which will be aggregated etc. and replace the current AppRepository handling).

I had also planned to link to specific documentation for flux and carvel, but given the current doc restructuring that @ppbaena is working on, I instead just pointed to the root of the docs.

### DIFF
--- a/dashboard/src/components/Catalog/Catalog.tsx
+++ b/dashboard/src/components/Catalog/Catalog.tsx
@@ -87,7 +87,7 @@ export default function Catalog() {
     },
     operators,
     repos: { repos },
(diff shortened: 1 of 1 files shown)


castelblanque:
That's a nice help for the users. Thank you!

castelblanque (approved):

PR #4333, opened 2022-02-23
Include installed pkg status and user reason on app detail.

absoludity:

### Description of the change

Currently we haven't included the installed package status on the detail page at all (only on the list page, as a tag). For Helm this wasn't a problem, as we only ever got to the detail page if the helm chart was installed, but for Flux (and others) the semantics are slightly different: The `HelmRelease` is created, but the helm chart may not be installed.

This happens most obviously when the service account chosen for the install doesn't have the required perms. The install details page shows zero information about the state of the install.

With this change, we update to display the status (`installed`, `pending`, etc.), and for the case where the status isn't `installed`, we also display the user reason:

flux-pending-failed

In the above example, the user reason isn't great: it should ideally be saying:
> secrets is forbidden: User "system:serviceaccount:kubeapps-user-namespace:default" cannot list resource "secrets" in API group "" in the namespace "kubeapps-user-namespace"

or even

> Selected service account does not have the required RBAC.

but that's the plugin's responsibility (I may take a look at that separately).

### Benefits

In the case of a pending install, the user has some information about what is wrong.

### Applicable issues

- ref #3695 

### Additional information

When a package is installed successfully, the status is still currently displayed (just not any user reason):

apache-success

### DIFF
--- a/dashboard/src/components/AppList/AppListItem.tsx
+++ b/dashboard/src/components/AppList/AppListItem.tsx
@@ -2,12 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import Tooltip from "components/js/Tooltip";
-import {
-  InstalledPackageSummary,
-  InstalledPackageStatus_StatusReason,
-  installedPackageStatus_StatusReasonToJSON,
-} from "gen/kubeappsapis/core/packages/v1alpha1/packages";
-import { getPluginIcon } from "shared/utils";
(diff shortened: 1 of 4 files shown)


antgamdia (approved):
Great, thanks for adding this additional information, really useful.

What we did in the past was just to show the status if it was! != installed/pending ([code](https://github.com/kubeapps/kubeapps/blob/main/dashboard/src/components/ApplicationStatus/ApplicationStatus.tsx#L137)), but adding the status as a "permanent" field is a much better option!

Issue #4337, opened 2022-02-24
Clients cannot differentiate between temporary and permanent "PENDING" status


absoludity:
### Description:

For discussion. I've hit an issue when integrating the flux work into the dashboard, in that a flux installed package can be PENDING in two very different ways:

1. pending installation - a state which will progress without any user interaction. This happens when a package is first installed (ie. the `HelmRelease` is first created) before and during the reconciliation process.
2. pending some required change - a state which can happen, for example, when the service account does not have the required RBAC, so the UI requests the package installation, the backend creates the `HelmRelease`, but the reconciliation never progresses as the RBAC associated with the service account needs to be updated for the reconciliation to move forward.

```
k get helmrelease --all-namespaces
NAMESPACE                 NAME             READY   STATUS                                AGE
kubeapps-user-namespace   test-apache      False   failed to get last release revision   28h
```

In the first case, the client will want to request the resource references for the installed package and expect to get them shortly. In the second case, the backend won't be able to provide any resource references until something changes.

Currently the client cannot differentiate between these, since both have an app status of simply PENDING. As a result, we need to keep checking for the resource references (link).

If we had two separate status' which could be provided by the backend, the client could know when to expect the resource references etc.

That said, it may not be possible even for the plugin to distinguish between these states (checking the underlying reason for the pending status may not be maintainable).

Thoughts?

castelblanque:
Shared my thoughts on [this related thread](https://github.com/vmware-tanzu/kubeapps/pull/4833#issuecomment-1148452780).
That is with regards to how to communicate to the consumers the different states when getting the resourcerefs.

Regarding when consumers should continue polling or not, as you mention in that other thread:

> So yes, there will always be resource refs for an install package or app, eventually.

So we could do: if returned status = ok and resources found =0 => consumers could continue polling, as it is assumed that application is existing but not yet complete/ready. Otherwise the responded status would be != OK.

PR #4338, opened 2022-02-24
Continue to poll for resource references when not found

absoludity:
### Description of the change

This PR updates the AppView so it handles the fetching of the resource references directly (rather than dispatching and going through redux) so that it can decide itself when to poll and importantly, can abort the polling when the component is unmounted.

This fixes a couple of issues that I found recently while testing deploying a flux package with a service account that doesn't have the required RBAC:

1. The existing code would retry the request up to 4 times, then display the not found error. This was find for the case where we expect the resources to exist in a short time-frame, but not for the case where they will never exist until RBAC is updated (see #4337 ),
2. The existing code would continue to retry the request even when you navigated away from the app view. This was because there was not a way currently to abort the requests when they are abstracted away through the redux dispatch/state-update lifecycle.

Since the direction of React's function components with `useEffect` etc. seems to be encouraging local state (rather than requiring redux for global state), I updated the component to request the refs directly and use local state. 

### Benefits

People won't be confronted with a NotFound error when viewing the detail of an installed app that is pending due to insufficient RBAC.

### Possible drawbacks

Testing components that manage local state via async network requests is more challenging (had to re-read about `async act()` to understand why components weren't rendered in tests etc.).

### Applicable issues

- ref #3695

### DIFF
--- a/dashboard/src/actions/installedpackages.test.tsx
+++ b/dashboard/src/actions/installedpackages.test.tsx
@@ -16,11 +16,7 @@ import { Plugin } from "gen/kubeappsapis/core/plugins/v1alpha1/plugins";
 import configureMockStore from "redux-mock-store";
 import thunk from "redux-thunk";
 import { InstalledPackage } from "shared/InstalledPackage";
-import {
-  IInstalledPackageState,
-  UnprocessableEntity,
-  UpgradeError,
-} from "shared/types";
+import { IInstalledPackageState, UnprocessableEntity, UpgradeError } from "shared/types";
 import { PluginNames } from "shared/utils";
 import { getType } from "typesafe-actions";
 import actions from ".";
--- a/dashboard/src/actions/installedpackages.test.tsx
+++ b/dashboard/src/actions/installedpackages.test.tsx
@@ -10,7 +10,6 @@ import {
   InstalledPackageStatus,
   VersionReference,
   ReconciliationOptions,
-  ResourceRef,
   InstalledPackageStatus_StatusReason,
 } from "gen/kubeappsapis/core/packages/v1alpha1/packages";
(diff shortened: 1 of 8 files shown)


castelblanque:
Excellent!, it made me reread about React's `act` too -> `This makes your test run closer to how React works in the browser.` 🤔 
Just a question: why not using one of the many retry libraries out there?

castelblanque (approved):

absoludity:
Thanks for the review :)

> Just a question: why not using one of the many retry libraries out there?

I did check if lodash already has something I can use (since we already import lodash), but didn't find one. Otherwise, I'm a bit hesitant to import something like https://www.npmjs.com/package/async-retry without seeing a benefit (try re-writing the small change in AppView.ts to use it). If we were using retries in multiple places, I'd definitely see the worth in pulling out the shared functionality and perhaps using async-retry. Let me know if there's a benefit I'm missing for a single case like this?

castelblanque:
> Let me know if there's a benefit I'm missing for a single case like this?

Once the code is written, no benefit in switching back to a library.
The way I see it, sometimes even for a single case it is worth using one library: your project's code focuses on the core business, might take advantage of additional features (e.g. exponential backoff), or even increase readability (e.g. code block should be shorter).

absoludity:
> > Let me know if there's a benefit I'm missing for a single case like this?
> 
> Once the code is written, no benefit in switching back to a library. The way I see it, sometimes even for a single case it is worth using one library: your project's code focuses on the core business, might take advantage of additional features (e.g. exponential backoff), or even increase readability (e.g. code block should be shorter).

Yep, agree if there's increased readability and/or additional features that are needed it'd be worth it, just don't see that here at all (tried re-writing as above, harder to see, because it's more generic, and we don't want exp backoff or anything similar as it needs to respond quickly when the installed package is created).

PR #4349, opened 2022-02-25
Use env var instead of git config for setting the ssh key

antgamdia:
### Description of the change

Probably after the update of the images used in CircleCI, I don't know, but the sync chart from bitnami step stopped working.  This PR just changes the way the ssh key is passed back to git: before, we were setting the local git config, which was working nicely, but now this same info is passed through an env var.

### Benefits

CI step will work again

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

Tested while accessing via ssh to the circle ci machine with a successful output: https://github.com/kubeapps/kubeapps/pull/4348

### DIFF
--- a/script/chart_sync_utils.sh
+++ b/script/chart_sync_utils.sh
@@ -166,8 +166,7 @@ updateRepoWithRemoteChanges() {
     git -C "${TARGET_REPO}" pull upstream "${BRANCH_CHARTS_REPO_ORIGINAL}"
 
     # https://superuser.com/questions/232373/how-to-tell-git-which-private-key-to-use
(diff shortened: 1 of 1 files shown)


script/chart_sync_utils.sh
```diff
@@ -166,8 +166,7 @@ updateRepoWithRemoteChanges() {
     git -C "${TARGET_REPO}" pull upstream "${BRANCH_CHARTS_REPO_ORIGINAL}"
 
     # https://superuser.com/questions/232373/how-to-tell-git-which-private-key-to-use
```
absoludity, line 168: I see the link is still relevant (also has answers for the env var method).

absoludity (approved):

PR #4353, opened 2022-02-28
2nd step for  Investigate and propose package repositories API with similar core interface to packages API. #3496 

gfichtenholt:
This is just a small incremental check-in that adds a few things toward #3496


cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo.go
```diff
@@ -215,50 +215,58 @@ func (s *Server) newRepo(ctx context.Context, targetName types.NamespacedName, u
...
 		if typedClient, err := s.clientGetter.Typed(ctx, s.kubeappsCluster); err != nil {
-			return err
+			return nil, err
 		} else if _, err = typedClient.CoreV1().Secrets(targetName.Namespace).Get(ctx, secretRef, metav1.GetOptions{}); err != nil {
-			return statuserror.FromK8sError("get", "secret", secretRef, err)
+			return nil, statuserror.FromK8sError("get", "secret", secretRef, err)
 		}
+		// TODO (gfichtenholt) also check that the secret type corresponds to specified auth type,
+		// e.g. if AuthType is PackageRepositoryAuth_PACKAGE_REPOSITORY_AUTH_TYPE_BASIC_AUTH,
+		// check that the secret has "username" and "password" fields, etc.
```
absoludity, line 251: Not sure if it's useful, but kubernetes also has actual secret types that - such as `kubernetes.io/basic-auth` - that cause certain validation to be done on the data (ie. ensuring that the "username" and "password" fields are present). I think we already chatted about these a while back when looking at the design, but worth checking so we don't double-up. https://kubernetes.io/docs/concepts/configuration/secret/#secret-types

absoludity (approved):
Great. Thanks for the small diffs Greg (when the generated stuff is ignored :) )

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo.go
gfichtenholt: yes, I know about those. I don't know if flux authors do. Their examples 
https://fluxcd.io/docs/components/source/helmrepositories/#spec-examples
only show opaque secrets. That doesn't necessarily mean they don't support typed secrets, I just need to test it out. It feels  kind of weird for them not to use typed secrets in examples - would have been easier to understand, but I digress. I will add a TODO for this

absoludity: Yep, sounds good. I think the typed secrets are a more recent and unenforced, so many code bases (including ours) haven't updated to use them yet.

PR #4359, opened 2022-03-01
Enable an empty response from carvel.GetAvailablePackageSummaries

### LINKED ISSUE #4328: [carvel] Error when no repositories available
### Description:

If carvel support is enabled but there are no `packagerepository` resources in the global namespace, an error is returned:

carvel-no-packages

rather than showing the normal empty catalog.

1. Run Kubeapps configured with carvel support
2. Go to the catalog

See above screenshot

The normal empty catalog page

Just came across this by chance after having recorded a demo, then wanting to work on a related flux issue for what to display when the catalog is empty. I may jump in and fix it along the way, but just recording in case I don't get to do so.

absoludity:

### Description of the change

See #4328 for description of the visible problem.

In the code, AFAICT, within the Carvel plugin, there was no difference in functionality for an empty state vs an incorrect state. So if no repo was setup, an error was returned as per #4328 in the same way that the error was returned if a carvel package metadata did not have a corresponding package (version).

This change separates errors from an empty result, returning a specific Internal error when it happens, but not returning an error for an empty catalog.

### Benefits

Users can see the empty catalog view with links to documentation for setting up repositories etc.

empty-catalog

### Applicable issues

- fixes #4328 

### Additional information

Since I needed a channel to gather any errors, it seemed convenient to use that channel for synchronization rather than a separate waitgroup.

I'll probably do a follow-up PR that also updates the channel to include the result (not just error), so that we're not doing write operations on a shared slice concurrently, but it's not necessary, just a suggestion.

### DIFF
--- a/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
+++ b/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
@@ -53,8 +53,10 @@ func (s *Server) GetAvailablePackageSummaries(ctx context.Context, request *core
 	// paginate the list of results
 	availablePackageSummaries := make([]*corev1.AvailablePackageSummary, len(pkgMetadatas))
 
-	// create the waiting group for processing each item aynchronously
-	var wg sync.WaitGroup
+	// Create a channel to receive any errors. The channel is also used as a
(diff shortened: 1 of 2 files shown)


antgamdia (approved):
Great! As you suggested some time ago, we better use channels as they are more versatile for controlling the results and errors.
Thanks for the fix!

PR #4360, opened 2022-03-01
Avoid writing to shared memory in go-routine.

absoludity:

### Description of the change

This change is a refactor only. It updates to:
1. Move the functionality of the go-routine into a typed function returning both the available pkg summary and error,
2. Update so that the go-routine communicates via the channel, sending the available pkg summary and error, rather than writing to shared memory.

As a result, we no longer need to worry about `nil` entries in a separate slice etc.

Note:
- I've also include the index in the channel message, just so the current order is maintained.
- I've added a test for fetching a non-initial page of results (as when I refactored I knew I'd broken it at first, but tests passed).

### Benefits

Main benefit is that we're preferring communication via channels (and using that channel for syncing results) rather than using shared memory with an extra wait group. Not a big deal, but good to have an example to follow in other parts of the code, imo.

### Applicable issues

- fixes #

### Additional information

I don't mind if you prefer not to change this.

Up next: I have a situation I want to investigate, where I've added the TCE repo, I can see the packages (and meta) in the cluster, but Kubeapps shows an empty catalog (both on main and in this branch). Probably something I'm doing wrong, but I'll investigate first thing tomorrow.

### DIFF
--- a/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
+++ b/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
@@ -50,95 +50,87 @@ func (s *Server) GetAvailablePackageSummaries(ctx context.Context, request *core
 		return nil, statuserror.FromK8sError("get", "PackageMetadata", "", err)
 	}
 
-	// paginate the list of results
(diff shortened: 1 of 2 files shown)


antgamdia (approved):
Interesting! Yep, a much better approach for avoiding dealing with `nil` values.  Thanks for improving it!

PR #4368, opened 2022-03-02
Carvel docs to specify namespace for repository.

absoludity:

### Description of the change

While testing yesterday I'd added the TCE carvel repo as per the docs, but noticed I still had an empty catalog. Turned out to be because the TCE `packagerepository` was created in the `default` namespace, so the packages were only visible in the catalog when I switched context to the default namespace.

### Benefits

Avoid confusion when trying out the carvel support.

### Applicable issues

- fixes #


antgamdia (approved):
Great, thanks for investigating and documenting it!

PR #4370, opened 2022-03-02
Set reasonable QPS/Burst for resources RESTMapper client

absoludity:
### Description of the change

Sets reasonable QPS and Burst for the client config used only in the discovery of the REST API (by the RESTMapper).

Before this change, registering the resources plugin took around 8 seconds due to client throttling:
```
I0302 00:40:47.539843       1 plugins.go:137] Successfully registered plugin "/plugins/fluxv2-packages/fluxv2-packages-v1alpha1-plugin.so"
I0302 00:40:48.318594       1 request.go:597] Waited for 109.310691ms due to client-side throttling,
 not priority and fairness, request: GET:https://10.96.0.1:443/apis/networking.k8s.io/v1beta1
...
I0302 00:40:56.919228       1 request.go:665] Waited for 8.611529015s due to client-side throttling,
 not priority and fairness, request: GET:https://10.96.0.1:443/apis/kustomize.toolkit.fluxcd.io/v1beta1

I0302 00:40:56.923634       1 plugins.go:137] Successfully registered plugin "/plugins/resources/resources-v1alpha1-plugin.so"
...
I0302 00:40:56.924955       1 server.go:158] Starting server on :50051
```

With this change, it's under 500ms (with no warnings of throttling):

```
I0302 05:39:33.515837       1 plugins.go:137] Successfully registered plugin "/plugins/fluxv2-packages/fluxv2-packages-v1alpha1-plugin.so"
I0302 05:39:34.010957       1 plugins.go:137] Successfully registered plugin "/plugins/resources/resources-v1alpha1-plugin.so"
```

### Benefits

Faster startup time for the kubeapps-apis server (see #4329).

### Possible drawbacks

Initial hammer of the k8s API server when the rest mapper initialises, but don't see this as an issue.

### Applicable issues

- ref #4329 

### Additional information

I did investigate whether we could use the client-getter to create a client using the chart configured defaults, but the client-getter is very careful to restrict itself to user requests, enforcing a user token, ensuring any service account token from the service is not used, etc., which I do not think we should change. Given that this client is used only for generating the rest mapper, I ended up going for simple constants.

### DIFF
--- a/cmd/kubeapps-apis/core/plugins/v1alpha1/plugins.go
+++ b/cmd/kubeapps-apis/core/plugins/v1alpha1/plugins.go
@@ -45,6 +45,10 @@ type GRPCPluginRegistrationOptions struct {
 	ConfigGetter     core.KubernetesConfigGetter
 	ClustersConfig   kube.ClustersConfig
 	PluginConfigPath string
+	// The QPS and Burst options that have been configured for any
+	// clients of the K8s API server created by plugins.
+	ClientQPS   float32
+	ClientBurst int
 }
 
 // PluginWithServer keeps a record of a GRPC server and its plugin detail.
@@ -131,7 +135,7 @@ func (s *PluginsServer) registerPlugins(pluginPaths []string, grpcReg grpc.Servi
 			return err
 		}
 
-		if grpcServer, err := s.registerGRPC(p, pluginDetail, grpcReg, configGetter, serveOpts.PluginConfigPath); err != nil {
+		if grpcServer, err := s.registerGRPC(p, pluginDetail, grpcReg, configGetter, serveOpts); err != nil {
 			return err
 		} else {
 			pluginsWithServers = append(pluginsWithServers, PluginWithServer{
@@ -156,7 +160,7 @@ func (s *PluginsServer) registerPlugins(pluginPaths []string, grpcReg grpc.Servi
 
 // registerGRPC finds and calls the required function for registering the plugin for the GRPC server.
 func (s *PluginsServer) registerGRPC(p *plugin.Plugin, pluginDetail *plugins.Plugin, registrar grpc.ServiceRegistrar,
-	clientGetter core.KubernetesConfigGetter, pluginConfigPath string) (interface{}, error) {
+	configGetter core.KubernetesConfigGetter, serveOpts core.ServeOptions) (interface{}, error) {
 	grpcRegFn, err := p.Lookup(grpcRegisterFunction)
 	if err != nil {
 		return nil, fmt.Errorf("unable to lookup %q for %v: %w", grpcRegisterFunction, pluginDetail, err)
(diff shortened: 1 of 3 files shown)


cmd/kubeapps-apis/plugins/resources/v1alpha1/server.go
```diff
@@ -33,6 +33,11 @@ import (
 	"github.com/kubeapps/kubeapps/cmd/kubeapps-apis/plugins/pkg/statuserror"
 )
 
+const (
+	DISCOVERY_CLIENT_QPS   = 50.0
+	DISCOVERY_CLIENT_BURST = 100
```
castelblanque, line 38: Good catch! Would it make sense to have them as environment variables, defaulted to those values if not present? Don't remember if we are using env variables for this kind of setup. I was thinking in the cases of environments where K8s resources are scarce (e.g. Docker desktop extension or small clusters).

castelblanque (approved):

cmd/kubeapps-apis/plugins/resources/v1alpha1/server.go
antgamdia: +1 to provide these vars externally. I mean, we're already passing `QPS` and `burst` in the chart, so it is just a matter of using it, no?

castelblanque: > It might even be a chance to replace the (growing) args to the grpc registration function with a struct so that we don't need to update the function signature each time.

+1, that would be nice and give flexibility to the registration process.

absoludity: OK, I've updated this PR, rebasing on #4372 so it's clear as an example how we can update the args to the plugin registration without having to modify other plugins necessarily. PTAL :)

cmd/kubeapps-apis/core/plugins/v1alpha1/plugins.go
```diff
@@ -45,6 +45,10 @@ type GRPCPluginRegistrationOptions struct {
 	ConfigGetter     core.KubernetesConfigGetter
 	ClustersConfig   kube.ClustersConfig
 	PluginConfigPath string
+	// The QPS and Burst options that have been configured for any
+	// clients of the K8s API server created by plugins.
+	ClientQPS   float32
+	ClientBurst int
```
antgamdia, line 51: Ah, ok, so these values are already been set by the `kube-api-qps` and `kube-api-burst` CLI flags we added in the past, aren't they?

antgamdia (approved):
Great, thank you! I do like this approach rather than hardcoding the defaults as constants. Especially, when it comes to rate limitations. This way we ensure Kubeapps can be adjusted for fitting each cluster's performance needs.

cmd/kubeapps-apis/core/plugins/v1alpha1/plugins.go
absoludity: Yes, if we're passing them in, I didn't see a reason to use different flags for this client than those already passed in for the user client, so populate the values from the existing flags.

absoludity:
Landed at #4386

PR #4372, opened 2022-03-02
Replace args with struct for grpc registration.

absoludity:

### Description of the change

This is a prequel to #4370 . It doesn't make any functional change, but just pulls out the arguments to the dynamically loaded grpc registration function into a struct defined in the one place. This makes it easier moving forward for us to modify it when needed in the one place. Please see #4370 for an example where we ensure plugins have access to QPS and Burst options without having to change other plugins necessarily.

### Benefits

Easier to modify grpc registration args in the future.

### Applicable issues

- ref #4329 

### Additional information

I again hit an import naming issue here, because some plugins import both the generated plugins/v1alpha1 as well as our implementation module for plugins/v1alpha1. So  @antgamdia , I looked up your earlier #4176 where you'd put some thought into this and found in the list there, where I think you found options for this same situation:

```
pluginsgrpcv1alpha1 "github.com/kubeapps/kubeapps/cmd/kubeapps-apis/gen/core/plugins/v1alpha1"
pluginsgrpcv1alpha1 "github.com/kubeapps/kubeapps/cmd/kubeapps-apis/gen/plugins/fluxv2/packages/v1alpha1"
pluginsv1alpha1 "github.com/kubeapps/kubeapps/cmd/kubeapps-apis/core/plugins/v1alpha1"
pluginsv1alpha1 "github.com/kubeapps/kubeapps/cmd/kubeapps-apis/gen/core/plugins/v1alpha1"
```

Now I assume the last entry there is incorrect (probably before you'd realised that some modules import both) as it duplicates the alias of the 3rd and the import path of the first. Similarly, I think the 2nd is misnamed (it's a flux plugin), and so I have used here the examples from the 1st and 3rd entries in the above snippet for the generated module and our implementation module.

Let me know if that's not what you intended.

### DIFF
--- a/cmd/kubeapps-apis/core/plugins/v1alpha1/plugins.go
+++ b/cmd/kubeapps-apis/core/plugins/v1alpha1/plugins.go
@@ -37,6 +37,16 @@ const (
 	clustersCAFilesPrefix   = "/etc/additional-clusters-cafiles"
 )
 
+// GRPCPluginRegistrationOptions defines the single argument that
+// a plugin's RegisterWithGRPCServer function must accept. This allows
+// the arguments to be defined (or modified) in the one place.
+type GRPCPluginRegistrationOptions struct {
+	Registrar        grpc.ServiceRegistrar
+	ConfigGetter     core.KubernetesConfigGetter
+	ClustersConfig   kube.ClustersConfig
+	PluginConfigPath string
+}
+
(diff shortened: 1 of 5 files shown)


antgamdia (approved):
Great, I'm a big fan of the struct args, otherwise adding new params is always a mess. Thanks for the change!

PR #4373, opened 2022-03-02
Switch to standalone cobra-cli dependency

liggitt:
### Description of the change

Switches to use the extracted `cobra-cli` command since it is being removed from the `github.com/spf13/cobra` module (xref https://github.com/spf13/cobra/issues/1597)

### Benefits

Makefile and docs will continue to be accurate after https://github.com/spf13/cobra/issues/1597 is completed

### Possible drawbacks

-

### Applicable issues

xref https://github.com/spf13/cobra/issues/1597


absoludity (approved):
Thanks for the reference (wow, quite an organised update!).

As mentioned on the referenced issue, we also don't really use the cobra cli except in rare cases of adding another command, so we may later remove it. But happy to see this landed for now and unblock your work. Thanks!

Issue #4099, opened 2022-01-17
Reduce response times when using large repos in Carvel plugin


antgamdia:
The current implementation is unusable (40-50 seconds) when it comes to large repositories. Since Carvel stores a `Package` for each version, we have `#packages * #versions` different packages.

For retrieving the package details, we need to query this object, so, for each package, we need to perform this O(n*m) search. Current implementation's response times are up to 10 seconds...

An initial general improvement should be tried and, if not improving, we might have to consider adding a cache...

More info at: https://github.com/kubeapps/kubeapps/pull/3784

Increasing the QPS (eg 100), does not seem to be working (still resp. times up to: 40-50 seconds)

antgamdia:
I'd bet we're gonna need a similar approach as the one implemented at https://github.com/kubeapps/kubeapps/issues/3636 :sweat_smile:

absoludity:
I took a look at the issue and possible solutions to this problem today, specifically at the `GetAvailablePackageSummaries` call.

I added the TCE repo (12 package metadatas with 17 package versions). Opening the catalog in Kubeapps three times resulted in:

```
I0303 05:14:17.418209       1 server.go:59] OK 5.691394379s /kubeappsapis.core.packages.v1alpha1.PackagesService/GetAvailablePackageSummaries
...
I0303 05:14:40.599514       1 server.go:59] OK 5.765965114s /kubeappsapis.core.packages.v1alpha1.PackagesService/GetAvailablePackageSummaries
...
I0303 05:14:48.814321       1 server.go:59] OK 5.128993047s /kubeappsapis.core.packages.v1alpha1.PackagesService/GetAvailablePackageSummaries
```

Similar to your thoughts above @antgamdia , it looks like the main issue is that currently we're doing a separate request for every package metadata to get the related package (version) resource. The data which is fetched with this extra request per item is simply the latest version.

Three options I can think of (ordered by worthiness :P ) are:

## Option 1 - fetch all package (versions) with a single request

Using a single request for all package (version) resources and then filtering this for the data we need while iterating the results (rather than converting it to an in-memory slice) would be much more efficient than N queries in most situations.

It'd be nice if we could just select all *relevant* package (version) resources, but the k8s API server only allows a limited set of operators for field selectors, so it's only possible to fetch either all or for *one* specific metadata name (the latter being what the code does currently)

## Option 2 - Don't display the version in the package summaries for carvel

[...]

PR #4378, opened 2022-03-03
Reduce Carvel response times for GetAvailablePackageSummaries

absoludity:
### Description of the change

See the [comment on the related issue](https://github.com/kubeapps/kubeapps/issues/4099#issuecomment-1057687768) for the background investigation.

This PR reduces the response time for the GetAvailablePackageSummaries RPC call of the Carvel plugin with the TCE repository (12 package metas, 18 packages) from over 5s to around 280-500ms on my local computer, without changing the functionality (~~although I've left a TODO related to one assumption which may not be correct, but it's an easy update if the assumption is false~~ Edit: I've now removed this assumption).

Timing with TCE repo after updating without assumption of version order:
```
I0304 01:28:41.068310       1 server.go:59] OK 603.570424ms /kubeappsapis.core.packages.v1alpha1.PackagesService/GetAvailablePackageSummaries
...
I0304 01:31:19.465106       1 server.go:59] OK 365.881391ms /kubeappsapis.core.packages.v1alpha1.PackagesService/GetAvailablePackageSummaries
...
I0304 01:32:00.670037       1 server.go:59] OK 557.427017ms /kubeappsapis.core.packages.v1alpha1.PackagesService/GetAvailablePackageSummaries
```

~~I'm keen to test it with a much larger repo.~~ Edit: Improvements are also 10-fold for my local system with over 100 packages, from 40-50s down to 4-5s. See [results on the issue](https://github.com/kubeapps/kubeapps/issues/4099#issuecomment-1058743438).

### Benefits

Faster catalog when using the Carvel support.

### Applicable issues

- ref #4099 

### Additional information

If we're happy with this approach, we can apply the same to other parts of the carvel support (installed package summaries).

I'll leave some notes inline too.

### DIFF
--- a/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
+++ b/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
@@ -57,55 +57,68 @@ func (s *Server) GetAvailablePackageSummaries(ctx context.Context, request *core
 	if pageSize > 0 {
 		startAt = int(pageSize) * pageOffset
 		if startAt > len(pkgMetadatas) {
-			// bad request.
+			return nil, status.Errorf(codes.InvalidArgument, "invalid pagination arguments %v", request.GetPaginationOptions())
 		}
 		pkgMetadatas = pkgMetadatas[startAt:]
 		if len(pkgMetadatas) > int(pageSize) {
 			pkgMetadatas = pkgMetadatas[:pageSize]
 		}
 	}
-	// TODO: update so pkgMetadatas is just the page of results and get
-	// rid of checks below. Then update to fetch all packages for those
-	// metas and work from there.
 
-	// Create a channel to receive any results. The channel is also used as a
-	// natural waitgroup to synchronize the results.
-	type fetchResult struct {
-		index                   int
-		availablePackageSummary *corev1.AvailablePackageSummary
-		err                     error
+	// Create a channel to receive all packages available in the namespace
+	getPkgsChannel := make(chan *datapackagingv1alpha1.Package)
+	var getPkgsError error
+	go func() {
+		getPkgsError = s.getPkgs(ctx, cluster, namespace, getPkgsChannel)
+	}()
+
+	// Skip through the packages until we get to the first item in our
+	// paginated results.
+	currentPkg := <-getPkgsChannel
+	for currentPkg != nil && currentPkg.Spec.RefName != pkgMetadatas[0].Name {
+		currentPkg = <-getPkgsChannel
 	}
-	fetchResults := make(chan fetchResult)
-	numFetched := 0
 
-	// TODO(agamez): DRY up this logic (cf GetInstalledPackageSummaries)
-	for i, pkgMetadata := range pkgMetadatas {
-		numFetched++
-		go func(i int, pkgMetadata *datapackagingv1alpha1.PackageMetadata) {
-			availablePackageSummary, err := s.fetchPackageSummaryForMeta(ctx, cluster, namespace, pkgMetadata)
-
-			// The index of this result is relative to the page.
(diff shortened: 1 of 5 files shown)


cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_adapters.go
```diff
@@ -25,15 +25,9 @@ import (
 	log "k8s.io/klog/v2"
 )
 
-func (s *Server) buildAvailablePackageSummary(pkgMetadata *datapackagingv1alpha1.PackageMetadata, pkgVersionsMap map[string][]pkgSemver, cluster string) (*corev1.AvailablePackageSummary, error) {
+func (s *Server) buildAvailablePackageSummary(pkgMetadata *datapackagingv1alpha1.PackageMetadata, latestVersion string, cluster string) *corev1.AvailablePackageSummary {
```
absoludity, line 28: There was no requirement for having the package map, just the latest version, which also means we don't need the ability to return an error here.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_resources.go
```diff
@@ -203,9 +203,31 @@ func (s *Server) getApp(ctx context.Context, cluster, namespace, identifier stri
 
 //  List of resources getters
 
-// getPkgs returns the list of packages for the given cluster and namespace
-func (s *Server) getPkgs(ctx context.Context, cluster, namespace string) ([]*datapackagingv1alpha1.Package, error) {
-	return s.getPkgsWithFieldSelector(ctx, cluster, namespace, "")
+// getPkgs requests the packages for the given cluster and namespace and sends
+// them to the channel to be processed immediately, closing the channel
+// when finished or when an error is returned.
+func (s *Server) getPkgs(ctx context.Context, cluster, namespace string, ch chan<- *datapackagingv1alpha1.Package) error {
```
absoludity, line 209: This function wasn't actually used at all, so I updated it to do what I want. Importantly, it does not iterate all the results into an in-memory slice, but rather sends each package down the return channel as it iterates (so we can process them while it continues iterating).

The other important point here is, as mentioned in the comment, that this function guarantees to close the channel, so we can be sure when it has completed (or returned with an error).

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
```diff
@@ -45,61 +47,90 @@ func (s *Server) GetAvailablePackageSummaries(ctx context.Context, request *core
...
-		categories = append(categories, fetchResult.availablePackageSummary.Categories...)
+		// The kapp-controller returns both packages and package metadata
+		// in order.
+		if currentPkg.Spec.RefName != pkgMetadata.Name {
+			return nil, status.Errorf(codes.Internal, fmt.Sprintf("unexpected order for kapp-controller packages, expected %q, found %q", pkgMetadata.Name, currentPkg.Spec.RefName))
+		}
+		// Collect the packages for a particular refName to be able to send the
+		// latest semver version. For the moment, kapp-controller just returns
+		// CRs with the default alpha sorting of the CR name.
+		// Ref https://kubernetes.slack.com/archives/CH8KCCKA5/p1646285201181119
```
antgamdia, line 103: Great, thanks for mentioning it. Interesting discussion!

antgamdia (approved):
Excelsior! Thanks for the changes, the single request+ulterior processing in channel is a great approach. 
We should release soon including this change, so that users can effectively/painlessly use carvel packages :P

castelblanque:
> Improvements are also 10-fold for my local system with over 100 packages, from 40-50s down to 4-5s.

That is mind-blowing!!!

absoludity:
Great, thanks. I'll try to update the get install pkg function soon too.

Issue #4390, opened 2022-03-08
[fluxv2] check RBAC settings of the currently logged in user when listing packages/repositories


gfichtenholt:
See related discussion in https://github.com/kubeapps/kubeapps/issues/4298#issuecomment-1061296805
Flux plugin currently has an issue. For the purpose of this issue, consider 2 "personas" involved:

1) an out-of-band background process running in the context of `kubeapps-internal-kubeappsapis` service account that watches all helm repositories added/updated/delete in the cluster and sync's them to redis cache

2) a user who logs into the UX dashboard or (gRPC API server) and is wishing to 
list available packages or maybe install a package. This user may have very different 
RBAC settings from those of `kubeapps-internal-kubeappsapis` service account, which may or may not 
allow them to list a given repository (e.g. by not having read access to the namespace in which the repository exists)

Flux plugin currently ignores (2). That is, any user that can login to the dashboard will 
be allowed to read any repository from redis cache thus bypassing the RBAC settings for this user.

This needs to be fixed. As a start, I may want to do something along the lines of what 
direct helm plugin does in `hasAccessToNamespace()` https://github.com/kubeapps/kubeapps/blob/5d4ea934808f8258d44d7edee7f71d5a6c1494e0/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go#L439. To be clear this is not a 100% accurate solution because the RBAC access may be more fine-grained than the namespace-level but its a start

gfichtenholt:
So one complication that arose is this:
unlike direct helm, flux allows one to read/use repositories from any namespace, unless explictly configured RBAC says otherwise. So, when the user wants to see the Catalog, i.e. GetAvailablePackageSummaries() is called, the first thing I do is call k8s API server List(...) for all namespaces to get a cluster-wide list of all repos. If the currently logged in user does not have such cluster-wide access, an error is raised 'rpc error: code = PermissionDenied desc = Forbidden to list the HelmRepository '/*' due to '[helmrepositories.source.toolkit.fluxcd.io](http://helmrepositories.source.toolkit.fluxcd.io/) is forbidden: User "system:serviceaccount:test-b3ot:test-caller-ctx-loser" cannot list resource "helmrepositories" in API group "[source.toolkit.fluxcd.io](http://source.toolkit.fluxcd.io/)" at the cluster scope'. This is a problem. My API should not be raising an error, but rather should be returning an empty list.

absoludity:
[...]

PR #4401, opened 2022-03-09
first step toward fixing  [fluxv2] check RBAC settings of the currently logged in user when listing packages/repositories #4390 

gfichtenholt:
This is just the first check-in toward fixing [[fluxv2] check RBAC settings of the currently logged in user when listing packages/repositories](https://github.com/kubeapps/kubeapps/issues/4390#)
#4390

I only took care of one API: GetAvailablePackageSummaries(). I need to go through all the APIs with a fine tooth comb and apply similar logic where needed

I also fixed a couple of minor bugs in unit tests and moved some util code around where it made sense


cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo.go
```diff
@@ -48,21 +48,40 @@ var (
 )
 
 // namespace maybe apiv1.NamespaceAll, in which case repositories from all namespaces are returned
+// the repositories to which the caller context has no access are filtered out
 func (s *Server) listReposInNamespace(ctx context.Context, namespace string) ([]sourcev1.HelmRepository, error) {
-	client, err := s.getClient(ctx, namespace)
```
absoludity, line 52: I assume this would return a client that was either for the specific namespace, or for all namespaces? But the new code below seems to just ignore `namespace`... fetching all repos in all namespaces and returning those the user has access to. Don't we only want to do that if `namespace = apiv1.NamespaceAll`, as per the function comment above?

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo.go
```diff
@@ -48,21 +48,40 @@ var (
...
+	// the actual List(...) call will be executed in the context of
+	// kubeapps-internal-kubeappsapis service account
+	// ref https://github.com/kubeapps/kubeapps/issues/4390 for explanation
```
absoludity, line 55: I would expect this to only be if the request is across all namespaces, otherwise it's going to be simpler/safer to request the repos in a single namespace with the user context?

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo.go
```diff
@@ -48,21 +48,40 @@ var (
...
 
 	var repoList sourcev1.HelmRepositoryList
-	if err := client.List(ctx, &repoList); err != nil {
+	if err := client.List(backgroundCtx, &repoList); err != nil {
 		return nil, statuserror.FromK8sError("list", "HelmRepository", namespace+"/*", err)
 	} else {
-		return repoList.Items, nil
+		// filter out those repos the caller has no access to
+		items := []sourcev1.HelmRepository{}
+		for _, item := range repoList.Items {
```
absoludity, line 68: Wouldn't it be better to iterate the namespaces (given that we can have N repos in a single namespace)? Iterating the repos we'd execute `s.hasAccessToNamespace()` N times for the one namespace, wouldn't we?

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
```diff
@@ -45,14 +45,19 @@ type Server struct {
 	// clientGetter is a field so that it can be switched in tests for
 	// a fake client. NewServer() below sets this automatically with the
 	// non-test implementation.
-	clientGetter       clientgetter.ClientGetterFunc
+	// It is meant for in-band interactions (i.e. in the context of a caller)
+	// with k8s API server
+	clientGetter clientgetter.ClientGetterFunc
+	// for out-of-band interactions (in the context of kubeapps-internal-kubeappsapis
+	// service account with k8s API server
+	backgroundClientGetter clientgetter.BackgroundClientGetterFunc
```
absoludity, line 53: It's not being used for just background/out-of-band interactions is it? Not sure, but serviceAccountClientGetter might be less misleading?

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
```diff
@@ -573,41 +578,33 @@ func (s *Server) getClient(ctx context.Context, namespace string) (ctrlclient.Cl
...
+	typedCli, err := s.clientGetter.Typed(ctx, s.kubeappsCluster)
 	if err != nil {
-		return pkgutils.VersionsInSummary{}, 0, fmt.Errorf("unable to open plugin config at %q: %w", pluginConfigPath, err)
-	}
-	err = json.Unmarshal([]byte(pluginConfig), &config)
+		return false, err
+	}
+
+	res, err := typedCli.AuthorizationV1().SelfSubjectAccessReviews().Create(
+		context.TODO(),
```
absoludity, line 589: Why not the original `ctx`? (it may have deadlines or cancellations etc. associated with it).

absoludity (commented):
I was wondering at first how all the config parsing code was relevant, but I think you've just updated something unrelated right? (making the timeout configurable)

gfichtenholt:
> I was wondering at first how all the config parsing code was relevant, but I think you've just updated something unrelated right? (making the timeout configurable)

correct. Its something unrelated. Just tidying up some code

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo.go
gfichtenholt: np. I will make that change

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
gfichtenholt: ok, will rename it

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
gfichtenholt: it was a copy and paste of code from helm plugin. I will change it in mine. I assume you'll take care of helm

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo.go
absoludity: Not sure if you misread my comment, but I wasn't meaning that this function should change to list all repos, but rather that we'd only be listing all repos if the request was for `apiv1.NamespaceAll`, otherwise if the function was meant to list repos in a specific namespace, we should use the user client (not service account). But that said, maybe you have a reason for changing it like this (maybe there's no need to list repos just in a specific namespace, not sure).

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo.go
```diff
@@ -47,22 +47,69 @@ var (
...
 )
 
-// namespace maybe apiv1.NamespaceAll, in which case repositories from all namespaces are returned
-func (s *Server) listReposInNamespace(ctx context.Context, namespace string) ([]sourcev1.HelmRepository, error) {
-	client, err := s.getClient(ctx, namespace)
+func (s *Server) listAllRepos(ctx context.Context) ([]sourcev1.HelmRepository, error) {
+	// the actual List(...) call will be executed in the context of
+	// kubeapps-internal-kubeappsapis service account
+	// ref https://github.com/kubeapps/kubeapps/issues/4390 for explanation
+	backgroundCtx := context.Background()
```
absoludity, line 54: Similar question about the context here - why not pass `ctx` into the function below? Context's are meant to be passed right through the request chain normally (though there are some exceptions).

absoludity (approved):

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go
absoludity: I don't see those boundaries the way you do - I don't mind if you update it in helm too (in fact, it'd seem normal). But no problem if you don't want to.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo.go
gfichtenholt: Eeeh, on the 2nd look, that's not right. Even when the caller asks `GetAvailablePackagesSummaries` for namespace-A, flux plugin should return packages from repos regardless which namespace they are in (of course minus repos in namespaces caller has no access to). These are all packages available for install as far as flux is concerned. So I think the `List() `request should always be cluster-wide. There is no use case I can see where the `List()` call is done w.r.t. to a single namespace.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo.go
gfichtenholt: Well, this isn't the right context for the List(..) call. I don't want caller credentials in this call. There maybe some information from the caller ctx that I could use, like timeout, if set, but I didn't want to use it "AS IS"

PR #4414, opened 2022-03-11
added keycloak doc

dlaloue-vmware:
### Description of the change

added document about keycloak installation/configuration for OIDC support

### DIFF
--- a/docs/user/OIDC/OAuth2OIDC-keycloak.md
+++ b/docs/user/OIDC/OAuth2OIDC-keycloak.md
@@ -170,7 +170,7 @@ For each of the two AWS clusters, we will create a client as follows:
 Once created, configure the authentication as follows:
  - Ensure the protocol is set to “openid-connect”
  - Configure the “Access Type” to be “confidential”. This will add a new “Credentials” tab from which you can get the client secret
- - For my case, as i just wanted to use tokens to access the cluster, so i disabled “Standard Flow Enabled”. Keep this option if you plan to use a local browser login screen (e.g. if using pinniped cli).
+ - If you just want to use tokens to access the cluster, you can disable the “Standard Flow Enabled”. Keep this option if you plan to use a local browser login screen (e.g. if using pinniped cli).
  - Ensure “Direct Access Grants Enabled” is enabled, as this is how we can get the tokens via API.
  - Save
 
--- a/docs/user/OIDC/OAuth2OIDC-keycloak.md
+++ b/docs/user/OIDC/OAuth2OIDC-keycloak.md
@@ -158,7 +158,7 @@ Note: if you navigate to “Client Scopes” and then select the tab “Default 
 
 ## Clients
 
-In probably a very simplified view, Clients represent the application to be protected and accessed via SSO and OIDC. In my use case, the environment consisted in the Kubeapps webapp and two AWS clusters. So we need to create three clients.
+In probably a very simplified view, Clients represent the application to be protected and accessed via SSO and OIDC. In my use case, the environment consisted of the Kubeapps web app and two AWS clusters. So we need to create three clients.
 
 ### Cluster clients
 
--- a/docs/user/OIDC/OAuth2OIDC-keycloak.md
+++ b/docs/user/OIDC/OAuth2OIDC-keycloak.md
@@ -250,7 +250,7 @@ In order to test multiple users with different levels of authorization, it is us
 ## Helm Install
 
 Few changes are required to values.yaml for the helm installation:
- - I changed the frontend service type to LoadBalancer so we can have a public hostname for the callback
+ -The `frontend` service type was changed to LoadBalancer so we can have a public hostname for the callback. Using an ingress could be another alternative.
  - The auth proxy must be configured. Here we will be using the default one.
 
 Here is an example of values.yaml:
--- a/docs/user/OIDC/OAuth2OIDC-keycloak.md
+++ b/docs/user/OIDC/OAuth2OIDC-keycloak.md
@@ -10,7 +10,7 @@ The base installation is done using the bitnami chart for Keycloak.
 
 ## SSL
 
-In order to support OIDC or OAuth, most servers and proxies require HTTPS. By default, the certificate created by the helm chart / keycloak server is both invalid (error with notBefore attribute) and also based on a deprecated certificate version making it incompatible to use (i.e. is it based on Common Name instead of SAN and is rejected).
+In order to support OIDC or OAuth, most servers and proxies require HTTPS. By default, the certificate created by the helm chart / keycloak server is both invalid (error with `notBefore` attribute) and also based on a deprecated certificate version making it incompatible to use (i.e. is it based on Common Name instead of SAN and is rejected).
 
 In this section we will see how to configure Keycloak with its “auth.tls” enabled. Note that there are other options to configure TLS, for example via ingress TLS, either manually or using a cert manager. This section will focus on TLS at the server level.
 
--- a/docs/user/OIDC/OAuth2OIDC-keycloak.md
+++ b/docs/user/OIDC/OAuth2OIDC-keycloak.md
@@ -1,10 +1,9 @@
 # Configuring Keycloak as an OIDC provider
-(created May 17th 2021)
 
 This document explains how to configure Keycloak as an IDP + OIDC provider (check general information and pre-requisites for [using an OAuth2/OIDC Provider with Kubeapps](../using-an-OIDC-provider.md)).
-It covers the installation and documentation for Kubeapps interacting with two AWS clusters. 
+It covers the installation and documentation for Kubeapps interacting with two Kubernetes clusters. 
 
-The base installation is done using the bitnami chart for Keycloak.
+The installation used the [bitnami chart for Keycloak](https://github.com/bitnami/charts/tree/master/bitnami/keycloak) (version 12.0.4/2.4.8) and [bitnami chart for Kubeapps](https://github.com/bitnami/charts/tree/master/bitnami/kubeapps) (version 7.0.0/2.3.2) 
 
 # Installation
 
@@ -46,7 +45,7 @@ subjectAltName      = @alt_names
 DNS.1               = *.us-east-2.elb.amazonaws.com
 ```
 
-The important section of the config is the `[req_ext]` extension section with alternate names. As the load balancer URL is created during installation (here on an AWS cluster), we use the wildcard mechanism so the final hostname can be verified by the certificates. Other hostnames, e.g. localhost, can also be added following the numbers pattern (`DNS.x`).
+The important section of the config is the `[req_ext]` extension section with alternate names. As the load balancer URL is created during installation, we use the wildcard mechanism so the final hostname can be verified by the certificates (here `DNS.1` illustrates a cluster on AWS). Other hostnames, e.g. localhost, can also be added following the numbers pattern (`DNS.x`).
 
 The following command will generate the private key and certificate files:
 
@@ -85,7 +84,7 @@ kubectl create secret generic keycloak-tls --from-file=./keycloak-0.keystore.jks
 
 To provide a default install, not many values must be provided in values.yaml - this is mostly some default passwords and the name of the secret created in Step 3 above.
 
-Here is the values.yaml file that i used:
+Here is a the values.yaml file that was applied:
 
 ```yaml
 ## Keycloak authentication parameters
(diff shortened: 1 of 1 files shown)


docs/user/OIDC/OAuth2OIDC-keycloak.md
```diff
@@ -1,9 +1,310 @@
 # Configuring Keycloak as an OIDC provider
+(created May 17th 2021)
```
antgamdia, line 2: Not sure if we should add the date here, as it is kind of redundant (this info is already in git) and we don't have it in the rest of the docs.

docs/user/OIDC/OAuth2OIDC-keycloak.md
```diff
@@ -1,9 +1,310 @@
...
 
-This document explains how to configure Keycloak as an OIDC provider (check general information and pre-requisites for [using an OAuth2/OIDC Provider with Kubeapps](../using-an-OIDC-provider.md)).
+This document explains how to configure Keycloak as an IDP + OIDC provider (check general information and pre-requisites for [using an OAuth2/OIDC Provider with Kubeapps](../using-an-OIDC-provider.md)).
+It covers the installation and documentation for Kubeapps interacting with two AWS clusters. 
 
-In the case of Keycloak, you can find the parameters in the Keycloak admin console:
+The base installation is done using the bitnami chart for Keycloak.
```
antgamdia, line 7:
```suggestion
The base installation is done using the [bitnami chart for Keycloak](https://github.com/bitnami/charts/tree/master/bitnami/keycloak).
```

docs/user/OIDC/OAuth2OIDC-keycloak.md
```diff
@@ -1,9 +1,310 @@
...
+
+# Installation
+
+## SSL
+
+In order to support OIDC or OAuth, most servers and proxies require HTTPS. By default, the certificate created by the helm chart / keycloak server is both invalid (error with notBefore attribute) and also based on a deprecated certificate version making it incompatible to use (i.e. is it based on Common Name instead of SAN and is rejected).
```
antgamdia, line 13:
```suggestion
In order to support OIDC or OAuth, most servers and proxies require HTTPS. By default, the certificate created by the helm chart / keycloak server is both invalid (error with `notBefore` attribute) and also based on a deprecated certificate version making it incompatible to use (i.e. is it based on Common Name instead of SAN and is rejected).
```

docs/user/OIDC/OAuth2OIDC-keycloak.md
```diff
@@ -1,9 +1,310 @@
...
+
+In this section we will see how to configure Keycloak with its “auth.tls” enabled. Note that there are other options to configure TLS, for example via ingress TLS, either manually or using a cert manager. This section will focus on TLS at the server level.
```
antgamdia, line 15:
```suggestion
In this section, we will see how to configure Keycloak with its `auth.tls` enabled. Note that there are other options to configure TLS, for example via ingress TLS, either manually or using a cert manager. This section will focus on TLS at the server level.
```

docs/user/OIDC/OAuth2OIDC-keycloak.md
```diff
@@ -1,9 +1,310 @@
...
+commonName_default          = keycloak
+
+[req_ext]
+subjectAltName      = @alt_names
+
+[alt_names]
+DNS.1               = *.us-east-2.elb.amazonaws.com
+```
+
+The important section of the config is the [req_ext] extension section with alternate names. As the load balancer URL is created during installation (here on an AWS cluster), we use the wildcard mechanism so the final hostname can be verified by the certificates. Other hostnames, e.g. localhost, can also be added following the numbers pattern (DNS.x).
```
antgamdia, line 49:
```suggestion
The important section of the config is the `[req_ext]` extension section with alternate names. As the load balancer URL is created during installation (here on an AWS cluster), we use the wildcard mechanism so the final hostname can be verified by the certificates. Other hostnames, e.g. localhost, can also be added following the numbers pattern (`DNS.x`).
```

docs/user/OIDC/OAuth2OIDC-keycloak.md
antgamdia, line 13:
```suggestion
In order to support OIDC or OAuth, most servers and proxies require HTTPS. By default, the certificate created by the helm chart / Keycloak server is both invalid (error with notBefore attribute) and also based on a deprecated certificate version making it incompatible to use (i.e. is it based on Common Name instead of SAN and is rejected).
```

docs/user/OIDC/OAuth2OIDC-keycloak.md
```diff
@@ -1,9 +1,310 @@
...
+
+```shell
+kubectl create secret generic keycloak-tls --from-file=./keycloak-0.keystore.jks  --from-file=./keycloak.truststore.jks
+```
+
+## Helm Install
+
+To provide a default install, not many values must be provided in values.yaml - this is mostly some default passwords and the name of the secret created in Step 3 above.
+
+Here is the values.yaml file that i used:
```
antgamdia, line 88:
```suggestion
The following excerpt shows the relevant values used in the `values.yaml` when installing the Helm Chart:
```

docs/user/OIDC/OAuth2OIDC-keycloak.md
```diff
@@ -1,9 +1,310 @@
...
+
+   ## Password to access the keystore when it's password-protected.
+   ##
+   keystorePassword: Vmware!23
+   ## Password to access the truststore when it's password-protected.
+   ##
+   truststorePassword: Vmware!23
+```
+
+Then just deploy Keycloak either using Kubeapps or helm cli.
```
antgamdia, line 128: I'd add the command like:

```shell
  helm install keycloak bitnami/keycloak --values my-values.yaml
````

File: docs/user/OIDC/OAuth2OIDC-keycloak.md [hunk 8]
```diff
@@ -1,9 +1,310 @@
...
+
+
+# Configuration
+
+Follow the Keycloak documentation to create and configure a new Realm to work with.
```
antgamdia, line 133:
```suggestion
Follow the [Keycloak documentation](https://www.keycloak.org/documentation) to create and configure a new Realm to work with.
```

docs/user/OIDC/OAuth2OIDC-keycloak.md
```diff
@@ -1,9 +1,310 @@
...
+
+This section will focus on a few aspects to configure for the SSO scenario to work.
+
+## Groups Claim
+
+By default, there is no “groups” scope/claim. We will create a global client scope for groups.
```
antgamdia, line 139:
```suggestion
By default, there is no "groups" scope/claim. We will create a global client scope for groups.
```

antgamdia: This suggestion also applies to the rest of the right/left quotation marks`“` (unicode 201C and 201D) ---> `"` (unicode 0022)

docs/user/OIDC/OAuth2OIDC-keycloak.md
```diff
@@ -1,9 +1,310 @@
...
+     - Enter “groups” as the token claim name
+     - Ensure the “Full group path” is OFF
+     - Keep the other knobs as ON
+ - Click ‘Save’
+
+Note: if you navigate to “Client Scopes” and then select the tab “Default Client Scopes” you should be able to see the newly created “groups” scope in the “available client scopes” lists.
+
+## Clients
+
+In probably a very simplified view, Clients represent the application to be protected and accessed via SSO and OIDC. In my use case, the environment consisted in the Kubeapps webapp and two AWS clusters. So we need to create three clients.
```
antgamdia, line 161:
```suggestion
In probably a very simplified view, Clients represent the application to be protected and accessed via SSO and OIDC. In my use case, the environment consisted of the Kubeapps web app and two AWS clusters. So we need to create three clients.
```

docs/user/OIDC/OAuth2OIDC-keycloak.md
```diff
@@ -1,9 +1,310 @@
...
+
+### Cluster clients
+
+For each of the two AWS clusters, we will create a client as follows:
+ - Click “Clients” from the left navigator
+ - Click “Create” from the table
+ - Enter an “id” and Save (i used aws1 and aws2 respectively)
```
antgamdia, line 168:
```suggestion
 - Enter an “id” and Save (for example, `aws1`and `aws2` respectively)
```

docs/user/OIDC/OAuth2OIDC-keycloak.md
```diff
@@ -1,9 +1,310 @@
...
+
+Once created, configure the authentication as follows:
+ - Ensure the protocol is set to “openid-connect”
+ - Configure the “Access Type” to be “confidential”. This will add a new “Credentials” tab from which you can get the client secret
+ - For my case, as i just wanted to use tokens to access the cluster, so i disabled “Standard Flow Enabled”. Keep this option if you plan to use a local browser login screen (e.g. if using pinniped cli).
```
antgamdia, line 173:
```suggestion
 - If you just want to use tokens to access the cluster, you can disable the “Standard Flow Enabled”. Keep this option if you plan to use a local browser login screen (e.g. if using pinniped cli).
```

docs/user/OIDC/OAuth2OIDC-keycloak.md
```diff
@@ -1,9 +1,310 @@
...
+ - In the “Valid Redirect URIs” field, enter “http://localhost:8000/*” as a placeholder. We will need to revisit this field once we know the public hostname of kubeapps
+ - Save
+
+As for the cluster clients, we need to configure the client scopes:
+ - Click the “Client Scopes” tab
+ - Ensure the “email” scope is available either in the “Assigned Default Client Scopes” list or the “Assigned Optional Client Scopes” list
+ - The “groups” client scope should be available in the lists on the left. Add it either to the “Assigned Default Client Scopes” list or the “Assigned Optional Client Scopes” list.
+
+The last step is to configure the kubeapps client to be aware of the two cluster clients and be allowed to invoke them. There are different ways to configure Keycloak:
+ - Using automatic audience resolution. I did not understand how to use this method, so it is not explained here.
```
antgamdia, line 203:
```suggestion
 - Using automatic audience resolution. We haven't explored this method yet, therefore it won't be covered in this guide.
```

docs/user/OIDC/OAuth2OIDC-keycloak.md
```diff
@@ -1,9 +1,310 @@
...
+The oauth proxy used in kubeapps requires email as the username. Furthermore, if the email is not marked as verified, JWT validation will fail and authentication will fail.
+
+In order to test multiple users with different levels of authorization, it is useful to create them with multiple dummy email addresses. This can be done by ensuring that when the user is created, the field “email verified” is ON (skipping an actual email verification workflow).
+
+# Kubeapps
+
+## Helm Install
+
+Few changes are required to values.yaml for the helm installation:
+ - I changed the frontend service type to LoadBalancer so we can have a public hostname for the callback
```
antgamdia, line 253:
```suggestion
 -The `frontend` service type was changed to LoadBalancer so we can have a public hostname for the callback. Using an ingress could be another alternative.
```

antgamdia (approved):
Thanks for adding the docs. Just mainly minor formatting issues.

+1 as it improves the current docs and we have users asking for it; however, I think we should continue to iterate over the docs to make the style more consistent and clearer for users.

docs/user/OIDC/OAuth2OIDC-keycloak.md
dlaloue-vmware: i will remove the date, and instead include the versions of the charts i used (easier for users)

docs/user/OIDC/OAuth2OIDC-keycloak.md
dlaloue-vmware: i will make a change locally to include the chart versions

docs/user/OIDC/OAuth2OIDC-keycloak.md
dlaloue-vmware: will fix

docs/user/OIDC/OAuth2OIDC-keycloak.md
dlaloue-vmware: will add in next commit

docs/user/OIDC/OAuth2OIDC-keycloak.md
dlaloue-vmware: will add in next commit

Issue #4371, opened 2022-03-02
Generate the readme in the CI sync chart


antgamdia:
During the sync chart process, the readme is left as is. However, given that the readme-generator-for-helm project is increasingly evolving, our CI should run the generation command to avoid tedious manual steps.

_Originally posted by @absoludity in https://github.com/kubeapps/kubeapps/pull/4369#discussion_r817345829_

PR #4429, opened 2022-03-15
Add readme genenator on each chart sync

### LINKED ISSUE #371: can't install kubeapps components: Error updating roles kubeapps.apprepository-controller: roles.rbac.authorization.k8s.io "apprepository-controller" is forbidden: unable to create new content in namespace kubeapps because it is being terminated
when I install kubeapps, I type: kubeapps up
then I got the following error:

can't install kubeapps components: Error updating roles kubeapps.apprepository-controller: roles.rbac.authorization.k8s.io "apprepository-controller" is forbidden: unable to create new content in namespace kubeapps because it is being terminated

My Environment:
OS: Ubuntu Server 16.04.01 LTS
Kubernetes:  1.8.13
Helm: 2.9.1

antgamdia:
### Description of the change

This PR adds an extra step in the chart sync CI script to also sync the chart's readme based upon the latest changes.

### Benefits

Readme in sync

### Possible drawbacks

N/A

### Applicable issues

- fixes https://github.com/kubeapps/kubeapps/issues/4371

### Additional information

Marking as a draft as I'm launching some debugging circleci instances to double-check it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -163,8 +163,7 @@ workflows:
             - build_pinniped_proxy
             - sync_chart_from_bitnami
       - sync_chart_from_bitnami:
-          <<: *build_always # TEMPORARY CHANGE
-          # <<: *build_on_main
+          <<: *build_on_main
       - GKE_REGULAR_VERSION_MAIN:
           <<: *build_on_tag_or_prerelease
           requires:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -528,7 +528,7 @@ gke_test: &gke_test
     - <<: *exports
     - <<: *install_gcloud_sdk
     - setup_remote_docker:
-          version: << pipeline.parameters.DOCKER_VERSION >>
+        version: << pipeline.parameters.DOCKER_VERSION >>
     - run:
         name: Configure Google Cloud
         command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -96,6 +96,9 @@ parameters:
   BRANCH_KUBEAPPS_REPO:
     type: "string"
     default: "main"
+  README_GENERATOR_REPO:
+    type: "string"
+    default: "bitnami-labs/readme-generator-for-helm"
 
 ## Build conditions
 # Build on any branch or tag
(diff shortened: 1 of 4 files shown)


absoludity:
I'm +1 for landing so that we have the readme step documented in code, but:

> Confirming it works:
> 
> A new chart version has been pushed to the bitnami/charts repo (see [bitnami/charts@de1763a](https://github.com/bitnami/charts/commit/de1763a86f765f76d5c8ea5e505381d1159e370f)), but it just updates the values.yaml .

So the upstream chart (bitnami/charts kubeapps) had its values updated but not the readme...

> However, our CI has successfully pushed the changes including the readme sync (see [aebeb51](https://github.com/kubeapps/kubeapps/commit/aebeb51400a4c471146219532803eadb2308d3b5))

...and our development branch gets the readme updated (but upstream doesn't).

For this scenario, I think it's something that the upstream content change should be doing, so that when the values.yaml is updated with bumped versions, the readme is regenerated. That way the upstream chart is correct *and* we still benefit.

Either way, if we have (dev-only) changes to kubeapps/kubeapps/chart, we'll still need to revert those as normal and then regenerate the README right?

I think this step will be more beneficial when we merge the upstream changes into our chart directory (perhaps just resolving the chart version conflict manually, or having a strategy for that), and *then* automatically generating the readme.

But anyway, +1 as it documents the usage which is great.

absoludity (approved):

antgamdia:
>So the upstream chart (bitnami/charts kubeapps) had its values updated but not the readme...
>...and our development branch gets the readme updated (but upstream doesn't).

>For this scenario, I think it's something that the upstream content change should be doing, so that when the values.yaml is updated with bumped versions, the readme is regenerated. 

Yep, currently, when automatic image bumps occur, the readme is not being updated upstream. This is an issue and I believe it will get addressed by the proper team eventually.

> Either way, if we have (dev-only) changes to kubeapps/kubeapps/chart, we'll still need to revert those as normal and then regenerate the README right?

Yes, even the readme generation step is done after adding our local changes (like substituting the images with the kubeapps') we still might have some dev changes (a new param in the chart, etc) that we would have to manually edit.

I think this step will be more beneficial when we merge the upstream changes into our chart directory (perhaps just resolving the chart version conflict manually, or having a strategy for that), and then automatically generating the readme.

> We eventually could take this generation step apart to a separate CI action (even a gh action) so that it becomes more flexible to trigger it whenever we want. We can discuss this in further CI improvements. 

But anyway, +1 as it documents the usage which is great.

Issue #4424, opened 2022-03-15
[fluxv2] Add `defaultUpgradePolicy` in the Fluxv2


antgamdia:
### Description:

As per the offline discussion we had, a short-term solution for allowing passing the version selection in Flux v2 would be just allowing the Fluxv2 to be configured with a plugin-wide config as we were doing in Carvel (see https://github.com/kubeapps/kubeapps/issues/4346#issuecomment-1054819538)

Later on, we can start discussing long-term solutions like https://github.com/kubeapps/kubeapps/issues/4365, but it is not a priority right now, as the community's feedback is that this version selection is not extensively being used out there, so just adding a plugin-wide option would be enough for now.

gfichtenholt:
Hi, I have a follow-up question:
Flux plugin implements a core API, like `CreateInstalledPackage`, where the caller may pass, among other things, a version reference. ref https://github.com/kubeapps/kubeapps/blob/90f135c82868a9c2599469eef9e3a0b03ac718d3/cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/packages.proto#L826.
Let's say for the sake of argument the caller sends in a version reference that is not a specific version but rather looks like a semver expression: `">5 <9"`. That is a perfectly legal version semver expression. Let's also suppose the `defaultUpgradePolicy` for the plugin happens to be set to `"minor"`. Based on looking at what carvel plugin currently does and what we'd expect from flux I think the result persisted in the CR will be something like `">=5.0.0 <6.0.0"` which is very different from what the caller intended. If I have misunderstood something please point it out. Thanks

absoludity:
Curerntly the Kubeapps UI does not allow specifying a semver expression at all - how we deal with that when it does is more related to #4346 . Rather, the UX currently always sends the exact version. The carvel plugin currently assumes this and uses it together with the `defaultUpgradePolicy` to set the result in the CR (as in [versionConstraintWithUpgradePolicy](https://github.com/kubeapps/kubeapps/blob/v2.4.3/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_utils.go#L237-L260) )

I think it's fine for the flux plugin to also assume this for now (ie. that it will currently receive a specific version). In fact, I'd say feel free to respond with a bad request, or unimplemented, if it's not, just to be explicit about this assumption in the code, until we work on #4346 .

That's my understanding. Is that what you would expect @antgamdia ?

[...]

PR #4456, opened 2022-03-17
fix for [fluxv2] Add defaultUpgradePolicy in the Fluxv2 #4424 

gfichtenholt:
Also DRY'ed up the code and put it in pkg so it can be shared between plugins


absoludity (approved):
Excellent - and thank you Greg for taking the extra time to update the kapp-controller plugin with the re-usable pkgutils :)

gfichtenholt:
> Excellent - and thank you Greg for taking the extra time to update the kapp-controller plugin with the re-usable pkgutils :)

Thanks, Michael

PR #4458, opened 2022-03-17
Add repo restriction to Github action

castelblanque:
### Description of the change

Restrict the execution of Custodian rule github action to the `kubeapps/kubeapps` repo only.

### Benefits

When forking Kubeapps repo, there will be no failing runs for the Github action.

### DIFF
--- a/.github/workflows/ci-custodian-rules.yaml
+++ b/.github/workflows/ci-custodian-rules.yaml
@@ -8,6 +8,7 @@ env:
   CLOUDSDK_CORE_PROJECT: ${{ secrets.GOOGLE_CLOUD_PROJECT }}
 jobs:
   remove-unattached-disks:
+    if: github.repository == 'kubeapps/kubeapps'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2


.github/workflows/ci-custodian-rules.yaml
```diff
@@ -8,6 +8,7 @@ env:
   CLOUDSDK_CORE_PROJECT: ${{ secrets.GOOGLE_CLOUD_PROJECT }}
 jobs:
   remove-unattached-disks:
+    if: github.repository == 'kubeapps/kubeapps'
```
antgamdia, line 11: Just curious, don't github actions have something like "this project" instead of having to write the project name?

antgamdia (approved):
Thanks!

.github/workflows/ci-custodian-rules.yaml
castelblanque: In this specific case it wouldn't work for us. When someone forks the repo, "this" would be the forked repo and thus the workflow would run 😄 
I got it from Github documentation [here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-only-run-job-for-specific-repository).

antgamdia: haha, you're right, of course (need more coffee to work). That's the purpose of the PR indeed, limiting forks 🤦

PR #4459, opened 2022-03-17
Create add-issues-to-project.yaml

ppbaena:
### Description of the change

For the management of the project to track issues it is necessary an automatic process to add issues to projects, so every new issue can be included in the triage process and then moved to the Backlog.

For that, a github action that assign issues to projects is included. At this moment, it is configured to assign NEW issues to the [old project](https://github.com/kubeapps/kubeapps/projects/11) and the [new project](https://github.com/kubeapps/projects/4).

### Benefits

New issues are automatically added to the project.

### Possible drawbacks

It is a public github action from the marketplace, so it is something to be considered for the maintenance.

### Additional information

GitHub Action: [Assign to One Project](https://github.com/marketplace/actions/assign-to-one-project)

### DIFF
--- a/.github/workflows/add-issues-to-project.yaml
+++ b/.github/workflows/add-issues-to-project.yaml
@@ -1,0 +1,23 @@
+name: Auto Assign to Project(s)
+
+on:
+  issues:
+    types: [opened]
+env:
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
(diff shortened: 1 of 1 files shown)


castelblanque:
Great, we are fully into GitHub actions! 
With this change we will not lost track of issues.
Thanks!

castelblanque (approved):

PR #4514, opened 2022-03-28
3rd step toward  Investigate and propose package repositories API with similar core interface to packages API. #3496 

gfichtenholt:
added core GetPackageRepositorySummaries() according to spec and flux implementation of it


cmd/kubeapps-apis/core/packages/v1alpha1/repositories.go
```diff
@@ -7,6 +7,8 @@ import (
 	"context"
 	"fmt"
 
+	. "github.com/ahmetb/go-linq/v3"
```
absoludity, line 10: You like controversy Greg :P https://github.com/golang/go/issues/29326 . I see you're using it for the LINQ DSL here. Fine either way (I don't expect support for dot imports will get removed, and if it does, it'll be easy to update).

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/common/utils.go
```diff
@@ -124,6 +111,17 @@ func NamespacedName(obj ctrlclient.Object) (*types.NamespacedName, error) {
 	}
 }
 
+// "Local" in the sense of no namespace is specified
+func NewLocalOpaqueSecret(name string) *apiv1.Secret {
+	return &apiv1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: name,
```
absoludity, line 118: Oh - GenerateName looks useful. I'd seen StringData for secrets before, but not GenerateName for metadata.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo.go
```diff
@@ -431,6 +431,57 @@ func (s *Server) repoDetail(ctx context.Context, repoRef *corev1.PackageReposito
...
+	summaries := []*corev1.PackageRepositorySummary{}
+	var repos []sourcev1.HelmRepository
+	var err error
+	if namespace == apiv1.NamespaceAll {
+		if repos, err = s.listReposInAllNamespaces(ctx); err != nil {
+			return nil, err
+		}
+	} else {
+		// here I think the right semantics are different than that of availablePackageSummaries
+		// namely, if a specific namespace is specified we need to list repos in that namespace
```
absoludity, line 444: Yes. So for example, even if there are repos in Carvel's `kapp-controller-packaging-global` namespace, we wouldn't want to be returning those when asking for the repos in the "default" namespace, even though for available packages we would. Agree.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo.go
```diff
@@ -431,6 +431,57 @@ func (s *Server) repoDetail(ctx context.Context, repoRef *corev1.PackageReposito
...
+		// and if the caller happens not to have 'read' access to that namespace, a PermissionDenied
+		// error should be raised, as opposed to returning an empty list with no error
```
absoludity, line 446: Yes.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo.go
```diff
@@ -431,6 +431,57 @@ func (s *Server) repoDetail(ctx context.Context, repoRef *corev1.PackageReposito
...
+
+		// also note that, at least theoretically, it is not clear what we'd do if we ever
+		// had a flavor of API that allowed 2 namespaces (as opposed to all/one), and the caller
+		// only had access to 1 of them. Raise the error (same as with one namespace) or ignore
+		// inaccessible namespace (same as with all namespaces)?
```
absoludity, line 451: Not sure we need this note, but yeah, I'd expect a 404 to b raised in that case (that is, explicitly requesting a namespace, whether it's one of many or alone, that I cannot access should return an explicit error).

absoludity (approved):
As always lately, thanks for the nice targeted PRs! So much easier to review :)

cmd/kubeapps-apis/core/packages/v1alpha1/repositories.go
gfichtenholt: not particularly. Do you? https://github.com/kubeapps/kubeapps/blob/c6b77631c917a6eed5afc138a17f95f9fcd75618/cmd/kubeapps-apis/core/packages/v1alpha1/packages.go#L10
Just copied your code 😛

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/common/utils.go
gfichtenholt: 👍

gfichtenholt:
> As always lately, thanks for the nice targeted PRs! So much easier to review :)

Thank you for review!

cmd/kubeapps-apis/core/packages/v1alpha1/repositories.go
absoludity: Haha - not me, but I did review it there too it seems :)

PR #4515, opened 2022-03-28
Bumped dev chart version to 8.0.4-dev1

castelblanque:

### Description of the change

Simply increase the dev chart version to `8.0.4-dev1` as noticed that the latest sync from Bitnami (#4511) didn't have it.
@absoludity Is it correct like this?


absoludity (commented):
Yep, that would be correct, but we can just do it as part of the next merge... which has already arrived in #4518 .

castelblanque:
Closing in favor of #4518

PR #4529, opened 2022-03-30
Update deps for the release process.

absoludity:

### Description of the change

Updates a bunch of deps following our release instructions. Also removes some notes in those instructions that I *think* are no longer necessary.

### Applicable issues

- fixes #


docs/reference/developer/release-process.md
```diff
@@ -66,62 +70,29 @@ yarn upgrade
...
-
-```bash
-# It requires Helm >=3.7.0, otherwise, just run `helm dep update ./chart/kubeapps`
-helm repo update bitnami
-helm dep update --skip-refresh ./chart/kubeapps
-```
-
-> As part of this release process, the chart dependencies _must_ be updated to the latest versions. In case of a major version, the change _should_ be tracked in a separate PR.
-
-### 0.4 - Protobuf dependencies and autogenerated code
```
absoludity, line 102: I've removed this section because we get this automatically from the merges of the bitnami chart now, don't we?

castelblanque: I think so. In the last 2-3 days we have received from Bitnami almost 1 chart update per day.

castelblanque:
This PR covers the whole step **0** of [release process document](https://github.com/kubeapps/kubeapps/blob/main/docs/reference/developer/release-process.md), right?
Thanks for the work!

castelblanque (approved):

absoludity:
> This PR covers the whole step **0** of [release process document](https://github.com/kubeapps/kubeapps/blob/main/docs/reference/developer/release-process.md), right? Thanks for the work!

Yep, except that I used `yarn upgrade-interactive` as per the updated docs.

PR #4530, opened 2022-03-30
Re-enable upgrade test now that 8.0.0 chart landed.

absoludity:

### Description of the change

Just re-enables the upgrade test now that the backwards-incompatible chart change landed.

### Applicable issues

- fixes #


antgamdia (approved):
Thanks!

PR #4548, opened 2022-04-01
Fix the unbound env var README_GENERATOR_REPO

absoludity:
CI error: /script/chart_sync.sh: line 38: README_GENERATOR_REPO: unbound variable

Signed-off-by: Michael Nelson <minelson@vmware.com>

### Description of the change

The release job failed with: /script/chart_sync.sh: line 38: README_GENERATOR_REPO: unbound variable

### Applicable issues

- fixes #

PR #4549, opened 2022-04-01
Remove readme-generator call for upstream PR

absoludity:
Since upstream now run the readme generator as an action when PR is
created.

### Applicable issues

- fixes #

### Additional information

@antgamdia This PR reverts the fix I did in #4548 to ensure the env var for the readme generator repo was bound, since the run with that change successfully ran the readme generator, but then `commitAndSendExternalPR` failed with "Wrong repo path. You should provide the root of the repository", as though after the readme generator ran, the chart.yaml wasn't in the expected place.

Although I've not verified the above, I also realised that for the upstream PR against bitnami/charts, we actually don't need to run the readme generator, as it's now run automatically as an action for each PR. So this just removes that step. I'm not confident that it'll fix the issue, but needs to be done anyway. Let me know if you've any other thoughts.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -773,7 +773,7 @@ jobs:
           # This token is passed as a GITHUB_TOKEN env var via CircleCI
           name: Run the chart_sync script
           command: |
-            ./script/chart_sync.sh << pipeline.parameters.CI_BOT_USERNAME >> << pipeline.parameters.CI_BOT_EMAIL >> << pipeline.parameters.CI_BOT_GPG >> << pipeline.parameters.CHARTS_REPO_ORIGINAL >> << pipeline.parameters.BRANCH_CHARTS_REPO_ORIGINAL >> << pipeline.parameters.CHARTS_REPO_FORKED >> << pipeline.parameters.BRANCH_CHARTS_REPO_FORKED >> << pipeline.parameters.README_GENERATOR_REPO >>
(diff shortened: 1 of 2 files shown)


antgamdia:
> Although I've not verified the above, I also realised that for the upstream PR against bitnami/charts, we actually don't need to run the readme generator, as it's now run automatically as an action for each PR. So this just removes that step. I'm not confident that it'll fix the issue, but needs to be done anyway. Let me know if you've any other thoughts.

No problem, thanks for unblocking the release. However, I still think this step is useful, I mean: let's consider we add a new parameter in the chart and we forget to document it. If this step is not being performed, we will be pushing a "wrong chart" to the bitnami/charts. 
It is not really an issue, since bitnami (now) will check the PR and determine if there exists any undocumented param, but, I mean, we'd be relying on 3rd party tools to fix/detect -our- issue.

Perhaps, we don't need to run this command at this point, but in every _on_main_ ?

absoludity:
> No problem, thanks for unblocking the release. However, I still think this step is useful, I mean: let's consider we add a new parameter in the chart and we forget to document it. If this step is not being performed, we will be pushing a "wrong chart" to the bitnami/charts. It is not really an issue, since bitnami (now) will check the PR and determine if there exists any undocumented param,

Right - I'm not sure why we should be doing the check here when generating the PR when it's going to be done when the PR is created anyway.

> but, I mean, we'd be relying on 3rd party tools to fix/detect -our- issue.

We're relying on 3rd party tools either way (readme-generator is used in both places), but I think you mean we're depending on their github action rather than our own explicit step. I don't see the issue since this PR is created explicitly for the bitnami/charts repo, not our own repo, so responsibility should be there imo?

> 
> Perhaps, we don't need to run this command at this point, but in every _on_main_ ?

Yep, running it for our own development chart would have some value for changes we make to our development chart, for sure.

PR #4550, opened 2022-04-01
Add Kubeapps logo in SVG format

castelblanque:

### Description of the change

Simply add Kubeapps standalone logo in SVG format.
There is already [this file in the repo](https://github.com/vmware-tanzu/kubeapps/blob/main/dashboard/src/logo.svg), but it contains both logo and name.

### Benefits

Kubeapps standalone logo is available.


dashboard/src/icons/kubeapps.svg
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="200" viewBox="55 45 520 520">
```
antgamdia, line 1: I'd say the rest of the icons are defaulted to 100x100. Have a look at the rest of the svg files (I should have wrote down the tool names I used)

antgamdia (approved):

dashboard/src/icons/kubeapps.svg
castelblanque: The `width` parameter is not really relevant. It is just the defaulting size used by rendering tool (i.e. browser).
As it is a vectorial image, whatever size it is set by CSS will prevail, and the image will be beautifully rescaled without glitches.

castelblanque: Have you used any tool that rewrites the vectorial path in a smaller view box?

antgamdia: Yep, I used the same view box, with the actual icon centered and a fixed height/width.

Even if the visualization can be later changed with css easily, it is useful when placing several svg and applying the same style to all of them at once.

For instance, I'd recall the Carvel and helm icons being misplaced (due to the canvas size and the icon not being centered) when rendering the catalog cards.

antgamdia: Also, I've noticed this is the b&w image, not the colored one. Shouldn't we use (or add it in addition to this one) the other?

castelblanque: Couldn't find the coloured SVG of the logo. Happy to add that one in addition.

castelblanque:
Added Kubeapps SVG icon in three different color schemes, all with transparent background:
- White
- Black
- Color

castelblanque:
One thing I noticed is that the compass needle is rotated 90º in some places.
@ppbaena could you please confirm that the needle pointing to NW is the right one?

Logos coming from AI file:
image

Current website:
image

castelblanque:
> the needle should be pointing to NE

Ooops, we have the `.ai` files with the needle pointing to NW. Should we try to make it rotate in SVG files (the ones we actually use in app/website), or should we wait for a new redesign of the logo?

ppbaena:
Let's wait for the redesign. Thanks @castelblanque !

PR #4561, opened 2022-04-06
Initialise serveOpts after arg parsing and update test to verify.

### LINKED ISSUE #4470: AppRepo sync jobs do not receive required imagePullSecrets
### Description:

This is a re-hash of #2144 which was fixed with #2156 but then, afaict, broken again with #3449 .

Similar to #2144, when deploying a helm chart from a private repo, with the global image pull secrets set correctly, the sync jobs sit in ImagePullBackoff due to an unauthorized error while pulling the image.

The issue appears to be that as part of the refactor in #3449 , `conf.ImagePullSecretsRefs` went from being set in the `main.go` after the command-line args had been parsed, to being moved to the `init()` function of the `root.go` command, which is executed when the module loads... *before* the command-line args are parsed. Since it relies on the value of one command-line arg in particular, `RepoSyncImagePullSecrets`, it is now empty always, even when the arg is set on the command-line.

This can be seen in the initial log of the command (formatted) where the resulting config has `RepoSyncImagePullSecrets` set from the cmd-line but `ImagePullSecretsRefs` is nil:
```
I0320 19:08:03.558376       1 root.go:48] apprepository-controller has been configured with: server.Config{
  ...
  RepoSyncImage:"registry.pivotal.io/tac-for-tanzu-advanced/containers/kubeapps-asset-syncer@sha256:3b21158f689cc8c297d44e5debb13a7e3661acc9ff3f773e648e80ce03917bd0",
  RepoSyncImagePullSecrets:[]string{"my-secret"},
  ImagePullSecretsRefs:[]v1.LocalObjectReference(nil),
  RepoSyncCommand:"/asset-syncer",
  ...
}

```

Since we don't have any unit-tests for this functionality currently (nor integration - we test private repos in integration, but not private images in private repos), we didn't notice the regression. We should ensure unit tests are added to avoid future regression.

See #2144

See #2144

See #2144

I haven't checked, but suspect the other values set in the same place will be the same, ie #3400.

absoludity:

### Description of the change

See #4470 for a complete description.

### Benefits

Secrets and custom annotations/labels are set correctly.

### Applicable issues

- fixes #4470

### DIFF
--- a/cmd/apprepository-controller/cmd/root.go
+++ b/cmd/apprepository-controller/cmd/root.go
@@ -32,6 +32,7 @@ func newRootCmd() *cobra.Command {
 		Use:   "apprepository-controller",
 		Short: "Apprepository-controller is a Kubernetes controller for managing package repositories added to Kubeapps.",
 		PreRun: func(cmd *cobra.Command, args []string) {
+			initServerOpts()
 			log.Infof("apprepository-controller has been configured with: %#v", serveOpts)
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
@@ -40,6 +41,14 @@ func newRootCmd() *cobra.Command {
 		},
 		Version: "devel",
 	}
+}
+
+// initServerOpts initialises the server options which are dependent
+// on the parsed arguments.
+func initServerOpts() {
+	serveOpts.ImagePullSecretsRefs = getImagePullSecretsRefs(serveOpts.RepoSyncImagePullSecrets)
+	serveOpts.ParsedCustomAnnotations = parseLabelsAnnotations(serveOpts.CustomAnnotations)
+	serveOpts.ParsedCustomLabels = parseLabelsAnnotations(serveOpts.CustomLabels)
 }
 
 // Execute adds all child commands to the root command and sets flags appropriately.
@@ -60,9 +69,6 @@ func init() {
 		log.Errorf("Error parsing verbosity: %v", viper.ConfigFileUsed())
 	}
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
-	serveOpts.ImagePullSecretsRefs = getImagePullSecretsRefs(serveOpts.RepoSyncImagePullSecrets)
-	serveOpts.ParsedCustomAnnotations = parseLabelsAnnotations(serveOpts.CustomAnnotations)
-	serveOpts.ParsedCustomLabels = parseLabelsAnnotations(serveOpts.CustomLabels)
 }
 
 func setFlags(c *cobra.Command) {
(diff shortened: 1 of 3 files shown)


cmd/apprepository-controller/cmd/root_test.go
```diff
@@ -134,9 +134,6 @@ func TestParseFlagsCorrect(t *testing.T) {
 			setFlags(cmd)
 			cmd.SetArgs(tt.args)
 			cmd.Execute()
-			serveOpts.ImagePullSecretsRefs = getImagePullSecretsRefs(serveOpts.RepoSyncImagePullSecrets)
-			serveOpts.ParsedCustomAnnotations = parseLabelsAnnotations(serveOpts.CustomAnnotations)
-			serveOpts.ParsedCustomLabels = parseLabelsAnnotations(serveOpts.CustomLabels)
```
absoludity, line 139: So it turned out that the test was already ensuring that the options were set correctly, but the test body was explicitly setting them (since the `init` function hadn't done so). Removing these causes the test to fail without the fix.

cmd/apprepository-controller/server/server.go
```diff
@@ -12,7 +12,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	kubeinformers "k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/tools/clientcmd" // Uncomment the following line to load the gcp plugin (only required to authenticate against GKE clusters).
+	"k8s.io/client-go/tools/clientcmd"
```
absoludity, line 15: Didn't see a following line, so it is now a dangling comment, hence removing.

cmd/apprepository-controller/cmd/root.go
```diff
@@ -32,6 +32,7 @@ func newRootCmd() *cobra.Command {
 		Use:   "apprepository-controller",
 		Short: "Apprepository-controller is a Kubernetes controller for managing package repositories added to Kubeapps.",
 		PreRun: func(cmd *cobra.Command, args []string) {
+			initServerOpts()
```
antgamdia, line 35: Great! Thanks for the fix! Haven't checked, but I wonder whether this same issue also appears in the rest of the cobra cmd we have in the project (kubeops, kubeapps-apis, etc...)

antgamdia (approved):
Thanks!

cmd/apprepository-controller/cmd/root.go
absoludity: Yep, at least in Kubeops. I'll check the others and do a separate diff.

PR #4569, opened 2022-04-07
Update asset-syncer and kubeops to ensure userAgent set correctly.

absoludity:

### Description of the change

Following on from the fix for 4470, it looks like with #3449 we may have also broke the userAgent setting for kubeops and the asset-syncer.

This PR updates the tests to actually verify that the userAgent is set and fixes accordingly.

### Applicable issues

- fixes #

### DIFF
--- a/cmd/asset-syncer/cmd/root.go
+++ b/cmd/asset-syncer/cmd/root.go
@@ -23,17 +23,12 @@ var (
 	version = "devel"
 )
 
-// rootCmd represents the base command when called without any subcommands
-var rootCmd *cobra.Command
-var syncCmd *cobra.Command
-var deleteCmd *cobra.Command
-var invalidateCacheCmd *cobra.Command
-
 func newRootCmd() *cobra.Command {
 	return &cobra.Command{
 		Use:   "asset-syncer",
 		Short: "Asset Synchronization utility",
-		PreRun: func(cmd *cobra.Command, args []string) {
+		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+			serveOpts.UserAgent = server.GetUserAgent(version, serveOpts.UserAgentComment)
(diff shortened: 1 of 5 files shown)


cmd/asset-syncer/cmd/root.go
```diff
@@ -23,17 +23,12 @@ var (
 	version = "devel"
 )
 
-// rootCmd represents the base command when called without any subcommands
-var rootCmd *cobra.Command
-var syncCmd *cobra.Command
-var deleteCmd *cobra.Command
-var invalidateCacheCmd *cobra.Command
-
```
absoludity, line 31: Wasn't sure why we needed all these global vars, so removed them.

cmd/asset-syncer/cmd/root.go
```diff
@@ -23,17 +23,12 @@ var (
...
 func newRootCmd() *cobra.Command {
 	return &cobra.Command{
 		Use:   "asset-syncer",
 		Short: "Asset Synchronization utility",
-		PreRun: func(cmd *cobra.Command, args []string) {
+		PersistentPreRun: func(cmd *cobra.Command, args []string) {
```
absoludity, line 30: PersistentPreRun is needed so that child commands still run it.

cmd/asset-syncer/cmd/root_test.go
```diff
@@ -18,8 +18,9 @@ func TestParseFlagsCorrect(t *testing.T) {
 		conf server.Config
 	}{
 		{
-			"all arguments are captured (root command)",
```
absoludity, line 21: The root command can only take a sub command, so testing the invalidate instead.

castelblanque:
Great, thanks for the fix.

castelblanque (approved):

PR #4573, opened 2022-04-08
4th step for  Investigate and propose package repositories API with similar core interface to packages API. #3496 

gfichtenholt:
4th step for Investigate and propose package repositories API with similar core interface to packages API. #3496

Changed AddPackageRepository and GetPackageRepositoryDetail to reflect latest design spec with user-managed vs kubeapps-managed secrets. For now, store the flag in plugin config, because it is easier to change on demand for unit tests than say, a kubeapps global setting. This is temporary, just until we settle on if/where to hang it.

+ added stub for UpdatePackageRepository
+ bumped flux version
+ added check that desired flux version is installed

Most of the changes are either in generated files or in unit test code


cmd/kubeapps-apis/proto/kubeappsapis/plugins/fluxv2/packages/v1alpha1/fluxv2.proto
```diff
@@ -100,4 +101,23 @@ service FluxV2RepositoriesService {
...
   }
+
+  rpc UpdatePackageRepository(kubeappsapis.core.packages.v1alpha1.UpdatePackageRepositoryRequest) returns (kubeappsapis.core.packages.v1alpha1.UpdatePackageRepositoryResponse) {
+    option (google.api.http) = {
+      put: "/plugins/fluxv2/packages/v1alpha1/repositories/plugin/{package_repo_ref.plugin.name}/{package_repo_ref.plugin.version}/c/{package_repo_ref.context.cluster}/ns/{package_repo_ref.context.namespace}/{package_repo_ref.identifier=**}"
+      body: "*"
+     };
+  }
+
+  // this endpoint only exists for the purpose of integration tests
```
absoludity, line 112: Not sure about having an actual extra endpoint for the integration tests?

absoludity (approved):
I'm struggling to review this one Greg. I've +1'd, but it'd be easier in the future if this was done as 2 or 3 PRs that each focused on just one thing (seemed to be a mix of a few things in this one, including general refactoring).

gfichtenholt:
> I'm struggling to review this one Greg. I've +1'd, but it'd be easier in the future if this was done as 2 or 3 PRs that each focused on just one thing (seemed to be a mix of a few things in this one, including general refactoring).

sorry about that Michael, I didn't think it would be that difficult to review. Really the there are 2 things in there: 1) separate the functionality of user-managed vs kubeapps-managed secrets based on a flag like we discussed and 2) just added empty stub for UpdateRepository (no implementation for plug in, just the core). 

Once again, I will try to  do better next time

cmd/kubeapps-apis/proto/kubeappsapis/plugins/fluxv2/packages/v1alpha1/fluxv2.proto
gfichtenholt: Well, I had a problem. I have a bunch of integration tests for AddRepository and GetRepositoryDetail. Once I split the server-side functionality of user-managed secrets and kubeapps-managed secrets a number of tests became invalid. For example, I had a test that would GetRepositoryDetail and return detail with secret refs. That only works with user-managed secrets. kubeapps-managed secrets would return Username/Password, etc. So I wanted the flexibility of switching from user-managed to kubeapps-managed and back while the server is running so that I can have a new test that returns Username/Password and the existing test that returns secretRefs. All while the server is running. My only way to do that is via a gRPC call. That's why the endpoint. The endpoint is only in the flux-specific API, BTW, not in the core API. In the future we may change how the flag works, so I may get rid of the endpoint

absoludity:
> > I'm struggling to review this one Greg. I've +1'd, but it'd be easier in the future if this was done as 2 or 3 PRs that each focused on just one thing (seemed to be a mix of a few things in this one, including general refactoring).
> 
> sorry about that Michael, I didn't think it would be that difficult to review. Really the there are 2 things in there: 1) separate the functionality of user-managed vs kubeapps-managed secrets based on a flag like we discussed and 2) just added empty stub for UpdateRepository (no implementation for plug in, just the core).

Oh - I didn't isolate it into those two when going through - seemed to be other refactoring too, which I wasn't sure whether it was changing code (tracking where it's deleted and added etc.)

> 
> Once again, I will try to do better next time

Heh - it's not that big a problem, no need for penance :P,  I'm just keen to be able to review things properly and find it difficult when it's not clear to me what has changed. I should have re-read the PR description after reading through, as you'd identified what had changed pretty clearly there.

cmd/kubeapps-apis/proto/kubeappsapis/plugins/fluxv2/packages/v1alpha1/fluxv2.proto
absoludity: Yeah, fine for now. If we need this kind of thing longer term, we could probably have a test plugin for example.

gfichtenholt: 👍

PR #4575, opened 2022-04-08
Add getting-started.md at old location with link.

absoludity:

### Description of the change

Apparently GH follows symlinks in your docs. We've had a number of reports for our old getting started being a 404 (from the old website, for eg.)

This should fix that.

Turns out the symlinks are only supported for a GH main README (afaict, can't find authoritatively, just doesn't work for the getting-started).

Updated now so that people following old links (eg. from kubeapps.com until we can get it updated) don't get a 404 but can follow through to the getting-started.

### Applicable issues

- fixes #

### DIFF
--- a/docs/user/getting-started.md
+++ b/docs/user/getting-started.md
@@ -1,0 +1,1 @@
+../tutorials/getting-started.md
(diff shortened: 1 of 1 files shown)


absoludity:
Hrmm, not working - will check on Monday.

PR #4587, opened 2022-04-13
5th step for  Investigate and propose package repositories API with similar core interface to packages API. #3496 

gfichtenholt:
* implemented UpdatePackageRepository() in flux plugin
* added unit and integration tests

### DIFF
--- a/cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo.go
+++ b/cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo.go
@@ -213,14 +213,14 @@ func (s *Server) newRepo(ctx context.Context, targetName types.NamespacedName, u
 			return nil, err
 		}
 	} else {
-		if secretRef, err = s.setupRepoKubeappsManagedSecrets(ctx, targetName, tlsConfig, auth); err != nil {
+		if secretRef, err = s.createRepoKubeappsManagedSecrets(ctx, targetName, tlsConfig, auth); err != nil {
 			return nil, err
 		}
 	}
 
 	passCredentials := auth != nil && auth.PassCredentials
 
-	if fluxRepo, err := s.newFluxHelmRepo(targetName, url, interval, secretRef, passCredentials); err != nil {
+	if fluxRepo, err := newFluxHelmRepo(targetName, url, interval, secretRef, passCredentials); err != nil {
 		return nil, err
 	} else if client, err := s.getClient(ctx, targetName.Namespace); err != nil {
 		return nil, err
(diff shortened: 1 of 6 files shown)


cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo.go
```diff
@@ -506,98 +429,129 @@ func (s *Server) setupRepoKubeappsManagedSecrets(
...
-			SecretRef: &corev1.SecretKeyReference{Name: secret.Name},
+	} else if existingSecretRef != nil {
+		if err = secretInterface.Delete(ctx, existingSecretRef.Name, metav1.DeleteOptions{}); err != nil {
+			log.Errorf("Error deleting existing secret: [%s] due to %v", err)
 		}
-	} else {
-		log.Warning("Unrecognized type of secret [%s]", secret.Name)
 	}
-	return tlsConfig, auth, nil
+	return secretRef, nil
```
absoludity, line 476: Confused at first because I thought a secret reference was an object (which wouldn't match the function return type), but see above it's just the name of the secret. No change needed, just noting.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo.go
```diff
@@ -506,98 +429,129 @@ func (s *Server) setupRepoKubeappsManagedSecrets(
...
 	}
-	return tlsConfig, auth, nil
+
+	repo.Spec.PassCredentials = auth != nil && auth.PassCredentials
+
+	// get rid of the status field, since now there will be a new reconciliation
+	// process and the current status no longer applies. metadata and spec I want
+	// to keep, as they may have had added labels and/or annotations and/or
+	// even other changes made by the user.
+	repo.Status = sourcev1.HelmRepositoryStatus{}
```
absoludity, line 535: Oh interesting... is that necessary? Won't it get overwritten when the new reconciliation process begins anyway? Interested to know if this is what the kubectl client does when updating something, or why it's otherwise needed here?

absoludity (approved):
Thanks Greg!

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo.go
gfichtenholt: > Oh interesting... is that necessary? Won't it get overwritten when the new reconciliation process begins anyway? Interested to know if this is what the kubectl client does when updating something, or why it's otherwise needed here?

eeeh, its been too long since I wrote this code (similar done for helmreleases) that I am struggling to remember the exact reason I put that in. I do remember same reasoning that you mentioned but something (I don't remember what) made me add this line in. It may have had something to do with unit tests (i.e. no real flux reconciler involved). Don't know off the top of my head if there is a solid reason for it to be in production. I know that its one line and its not hurting anything. If you insist, I will remove it and see what's what.

PR #4589, opened 2022-04-13
Auto add issues to the new project board

antgamdia:
### Description of the change

This PR simply adds a new GitHub action, extracted from [here](https://github.com/github/feedback/discussions/5378#discussioncomment-2137318), that automatically adds new issues to the board.

I've already set the proper environment variables in the GitHub repo.

### Benefits

Since we moved to the new projects (beta) boards, we need to add new issues there, so a change was required anyway. Besides, the current gh action we had, wasn't working at all (see https://github.com/vmware-tanzu/kubeapps/actions/runs/2114620529)

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

Added copyright headers that were missing.

### DIFF
--- a/.github/workflows/add-issues-to-project.yaml
+++ b/.github/workflows/add-issues-to-project.yaml
@@ -1,21 +1,20 @@
 # Copyright 2022 VMware, Inc.
 # SPDX-License-Identifier: Apache-2.0
 
-name: Auto Assign to Project(s)
+# From https://github.com/github/feedback/discussions/5378#discussioncomment-2137318
+name: Project automation
 
 on:
   issues:
-    types: [opened]
-env:
-  MY_GITHUB_TOKEN: ${{ secrets.MY_GITHUB_TOKEN }}
+    types:
+      - opened
 
 jobs:
-  assign_one_project:
+  add_issue_to_project:
+    name: Add new issues to project board
     runs-on: ubuntu-latest
-    name: Assign to One Project
     steps:
-    - name: Assign NEW issues to project
-      uses: srggrs/assign-one-project-github-action@1.3.1
-      if: github.event.action == 'opened'
-      with:
-        project: 'https://github.com/orgs/kubeapps/projects/4'
+      - run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/add-issues-to-project.yaml
```diff
@@ -1,18 +1,20 @@
...
+  add_issue_to_project:
+    name: Add new issues to project board
     runs-on: ubuntu-latest
-    name: Assign to One Project
     steps:
-    - name: Assign NEW issues to project
-      uses: srggrs/assign-one-project-github-action@1.3.1
-      if: github.event.action == 'opened'
-      with:
-        project: 'https://github.com/orgs/kubeapps/projects/4'
```
absoludity, line 18: Could the reason this was failing be related to it being for the wrong project? (haven't checked, but this is the old project under the kubeapps org)

.github/workflows/add-issues-to-project.yaml
```diff
@@ -1,18 +1,20 @@
...
+      - run: |
+          gh api graphql -f query='mutation($project:ID!, $issue:ID!) { addProjectNextItem(input: {projectId: $project, contentId: $issue}) { projectNextItem { id }  } }' -f project=${{ secrets.PROJECT_BOARD_ID }} -f issue=${{ github.event.issue.node_id }}
```
absoludity, line 18: I can't see what value you've used for `secrets.PROJECT_BOARD_ID`, but assuming you've set the correct one for project 38 under the vmware-tanzu org.

absoludity (approved):

.github/workflows/add-issues-to-project.yaml
antgamdia: No, I checked it at some point, but this action only works with standard projects, not the beta ones :(

.github/workflows/add-issues-to-project.yaml
antgamdia: It's in our internal system to share secrets if you want to have a look at it.

Unfortunately, It is not as easy as getting the value from the url haha, a graphql query is required to get the API node id of the project object.
I've documented everything here https://github.com/github/feedback/discussions/5378#discussioncomment-2558542

PR #4590, opened 2022-04-13
Fix custodian job

antgamdia:
### Description of the change

I've noticed the custodian job is failing since we transferred the repo to the `vmware-tanzu` organization. This PR is to change the hardcoded value with the correct one.

image

### Benefits

Jobs will continue to work again

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

@castelblanque , I wonder why we set `if: github.repository == 'vmware-tanzu/kubeapps'` instead of `if: github.repository == '$GITHUB_REPOSITORY'` or even removing this check. Perhaps there's a reason, but I don't recall it now haha.

### DIFF
--- a/.github/workflows/ci-custodian-rules.yaml
+++ b/.github/workflows/ci-custodian-rules.yaml
@@ -8,7 +8,7 @@ env:
   CLOUDSDK_CORE_PROJECT: ${{ secrets.GOOGLE_CLOUD_PROJECT }}
 jobs:
   remove-unattached-disks:
-    if: github.repository == 'kubeapps/kubeapps'
+    if: github.repository == 'vmware-tanzu/kubeapps'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3


castelblanque:
Good, thanks!

> I wonder why we set if: github.repository == 'vmware-tanzu/kubeapps' instead of if: github.repository == '$GITHUB_REPOSITORY' or even removing this check. 
We only want this logic to run on this very same repo, not on forks. If that assertion is removed, custodian rules will be run on all forks and thus fail.
If it was set to `$GITHUB_REPOSITORY`, it would always evaluate to true 😄 

[See](https://github.com/vmware-tanzu/kubeapps/pull/4458#discussion_r828938495)

antgamdia:
> Good, thanks!
> 
> > I wonder why we set if: github.repository == 'vmware-tanzu/kubeapps' instead of if: github.repository == '$GITHUB_REPOSITORY' or even removing this check.
> 
> We only want this logic to run on this very same repo, not on forks. If that assertion is removed, custodian rules will be run on all forks and thus fail. If it was set to `$GITHUB_REPOSITORY`, it would always evaluate to true 😄
> 
> [See](https://github.com/vmware-tanzu/kubeapps/pull/4458#discussion_r828938495)

🤦 You're right... forks, I always forget about'em, haha, sorry for the noise!

castelblanque (approved):

Issue #4564, opened 2022-04-06
[Regression] Allow package details to fetch from config.kubeappsCluster


aanthonyrizzo:
### Description:
The kubeapps dashboard no longer supports `kubeappsCluster` or `kubeappsNamespace`.

As of version 2.4.2 the kubeapps dashbord no longer respects the `kubeappsCluster` or kubeappsNamespace` configurations. 
as of v2.4.2, dashboard components no longer read the config for the kubeappCluster:
https://github.com/vmware-tanzu/kubeapps/blob/a1f95a580807e90ada19b8011407f2839213de15/dashboard/src/components/DeploymentForm/DeploymentForm.tsx#L44-L47

v2.4.1 does:
https://github.com/vmware-tanzu/kubeapps/blob/b8702bf1770ee887f83e7a4c7a1b0df6a91a84bc/dashboard/src/components/DeploymentForm/DeploymentForm.tsx#L44-L48

Previously these values were taken into account when fetching the package details from the UI. This is no longer the case. `GetAvailablePackageDetail` will not use the `kubeappsCluster` value when fetching the details.

Related to https://github.com/vmware-tanzu/kubeapps/issues/4563

Corresponding slack thread: https://kubernetes.slack.com/archives/C9D3TSUG4/p1649178273135249

### Repro Instructions:
1. Ensure that [kubeappsCluster](https://github.com/vmware-tanzu/kubeapps/blob/06774a4476219b6e031b5b2fb436f1c12a8048bd/chart/kubeapps/templates/dashboard/configmap.yaml#L70) is not included in [clusters](https://github.com/vmware-tanzu/kubeapps/blob/06774a4476219b6e031b5b2fb436f1c12a8048bd/chart/kubeapps/templates/dashboard/configmap.yaml#L79) in the configmap, this will make sure that the subsequet `GetAvailablePackageDetail` request will fail.
2. Load kubeapps as normal and try to deply a helm chart
3. Be greeted with: ![image](https://user-images.githubusercontent.com/36711320/161879048-58f3cb36-1f65-434a-977d-5fecd4786066.png)
4.  Verify it's fetching `GetAvailablePackageDetail` with a value in the clusters[] and not the kubeappsCluster. Ex: `+core GetAvailablePackageDetail (cluster="default", namespace="kubeapps")`

aanthonyrizzo:
[...]

PR #4591, opened 2022-04-13
Use an explicit token to identify kubeapps cluster when required.

absoludity:

### Description of the change

For discussion related to #4564 .

This PR aims to ensure that the context for an available package can always be resolved, even when there is no configured cluster name corresponding to the cluster on which Kubeapps is installed.

We need to test this as I suspect, if installing a package still requires fetching the AppRepository, that it would mean that the user credential must be valid to authenticate with the cluster on which Kubeapps is installed.
The user doesn't need any RBAC, only the ability to authenticate, since Kubeapps already includes RBAC to ensure that global repositories are readable by any authenticated user.

Also, if pinniped-proxy is being used, it won't currently handle this scenario (where the new token is used for the cluster on which Kubeapps is installed), but that will be easy to update. I'll add that depending on the discussion here.

### Applicable issues

- fixes #


antgamdia:
I'm noticing several random failures related to those tests since we upgraded some cargo deps :(
Don't know what's happening, though... but we might want to merge this PR 😅

antgamdia:
Once https://github.com/vmware-tanzu/kubeapps/pull/5566 was merged, can we close this one, @castelblanque ?

castelblanque:
Indeed.
Closing as this PR has been superseeded by #5566.

PR #4594, opened 2022-04-15
Step 6 for  Investigate and propose package repositories API with similar core interface to packages API. #3496 

gfichtenholt:
Changes in this PR:
1. added a check in UpdateInstalledPackage to raise an error if the package is in pending state
2. added several unit and integration tests for various scenarios. Done somy DRYing up of test code
3. per Dimitri's feedback removed 'type' from UpdatePackageRequest and added 'customDetail'


cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/release.go
```diff
@@ -386,6 +386,14 @@ func (s *Server) updateRelease(ctx context.Context, packageRef *corev1.Installed
 		return nil, err
 	}
 
+	// per discussion will Michael 4/12/2022
+	// for now: we disallow updates to pending releases and allow them for non-pending ones
+	// (i.e. success or failed status)
```
absoludity, line 391: Rather than "per discussion with Michael ...", it'd be more useful IMO to have the reason that you presented for this change, such as "Initially we'll disallow updates to pending releases to simplify the initial case, though we may implement support later."? (I have no issue if you want to include "As Michael and I agreed, initially we'll..." if that's important for you)

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo_integration_test.go
```diff
@@ -711,6 +714,42 @@ func TestKindClusterUpdatePackageRepository(t *testing.T) {
...
+		{
+			name:               "update repository change from TLS cert/key to basic auth (kubeapps-managed secrets)",
+			request:            update_repo_req_15,
+			repoName:           "my-podinfo-5",
+			repoUrl:            podinfo_tls_repo_url,
+			oldSecret:          newTlsSecret("secret-1", "TBD", pub, priv, ca),
+			expectedStatusCode: codes.OK,
+			expectedResponse:   update_repo_resp_5,
+			expectedDetail:     update_repo_detail_14,
+		},
```
absoludity, line 752: Out of interest: how is the test knowing which use case is user-managed secrets and which is kubeapps-managed secrets? I can't see anything in the use-case here that differentiates (other than the name).

absoludity: Wondering if there's a test showing that the existing secret won't be updated if it's not changed (ie. that the request does not need to contain the actual secret data when not updating the secret - though see my questions in the design doc, I'd understood we were always avoiding sending the secret data back to the UI or requiring it to be sent on update etc., as per your diagram in the appendix, but may have misunderstood that you'd not intended to apply that when Kubeapps is managing secrets, not sure.).

absoludity (approved):

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo_integration_test.go
gfichtenholt: > Out of interest: how is the test knowing which use case is user-managed secrets and which is kubeapps-managed secrets? I can't see anything in the use-case here that differentiates (other than the name).

There is a boolean userManagedSecrets in test struct definition

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/release.go
gfichtenholt: 👍

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo.go
```diff
@@ -483,7 +483,7 @@ func (s *Server) updateRepo(ctx context.Context, repoRef *corev1.PackageReposito
 		return nil, err
 	}
 
-	// (gfichtenholt) per discussion will Michael 4/12/2022
+	// per discussion will Michael 4/12/2022
```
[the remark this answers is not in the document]
gfichtenholt: 👍

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo_integration_test.go
gfichtenholt: > Wondering if there's a test showing that the existing secret won't be updated if it's not changed (ie. that the request does not need to contain the actual secret data when not updating the secret - though see my questions in the design doc, I'd understood we were always avoiding sending the secret data back to the UI or requiring it to be sent on update etc., as per your diagram in the appendix, but may have misunderstood that you'd not intended to apply that when Kubeapps is managing secrets, not sure.).

remember, this is a whole update, not a patch update semantics. No secret data in the update request (regardless of your intention not to update any secret data) => remove any secret data, if present, when handing the update on the backend.

Sounds like what you want is Patch() operation

Issue #4595, opened 2022-04-15
Kubeapps-Sync failing during Getting Started with `dial tcp 127.0.0.1:443: connect: connection refused`


browser-bug:
### Description:

Doing a basic deployment on a Minikube environment I'm getting the following error on the `apprepo-kubeapps-sync-bitnami` pod creation:
```
Error: Get "https://charts.bitnami.com/bitnami/index.yaml": dial tcp 127.0.0.1:443: connect: connection refused
```

### Steps to reproduce the issue:
Deploying a single node cluster with minikube `minikube version: v1.24.0`:
```
minikube start
```

Following the exact steps described in the Getting Started section:

```
helm repo add bitnami https://charts.bitnami.com/bitnami
kubectl create namespace kubeapps
helm install kubeapps --namespace kubeapps bitnami/kubeapps
```

### Describe the results you received:

`apprepo-kubeapps-sync-bitnami` going in `CrashLoopBackOff` with error:
```
kubectl -n kubeapps logs apprepo-kubeapps-sync-bitnami-27500250--1-pgnl7
Error: Error: Get "https://charts.bitnami.com/bitnami/index.yaml": dial tcp 127.0.0.1:443: connect: connection refused
```

### Version of Helm, Kubeapps and Kubernetes:

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.8.2", GitCommit:"6e3701edea09e5d55a8ca2aae03a68917630e91b", GitTreeState:"clean", GoVersion:"go1.17.5"}
```

- Output of `helm list -n <kubeapps-namespace> <kubeapps-release-name>`:

```
kubeapps	kubeapps 	1       	2022-04-15 11:02:39.603810736 +0200 CEST	deployed	kubeapps-8.0.10	2.4.4
```

- Output of `kubectl version`:

```
Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.5", GitCommit:"c285e781331a3785a7f436042c65c5641ce8a9e9", GitTreeState:"clean", BuildDate:"2022-03-16T15:58:47Z", GoVersion:"go1.17.8", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #4613, opened 2022-04-25
fix(dashboard): Add support for "" kubeappsCluster

aanthonyrizzo:
### Description of the change

Related frontend changes for #4591. These changes are required so that the route params are correctly translated. 

### Todo
- Test coverage for "" cluster for routing and fetching

@absoludity when I was testing locally I also ran into something new. When trying to deploy a helm chart I run into the following:
```
An error occurred: Unable to create the installed package for the package "coreweave/argo-cd" using the plugin "helm.packages": rpc error: code = Internal desc = Unable to create kubernetes clientset: rpc error: code = FailedPrecondition desc = unable to get client : rpc error: code = FailedPrecondition desc = unable to get controller runtime client due to: Unauthorized
```

### DIFF
--- a/dashboard/src/components/DeploymentForm/DeploymentForm.tsx
+++ b/dashboard/src/components/DeploymentForm/DeploymentForm.tsx
@@ -54,8 +54,10 @@ export default function DeploymentForm() {
   const {
     packages: { isFetching: packagesIsFetching, selected: selectedPackage },
     apps,
+    config: { kubeappsCluster },
   } = useSelector((state: IStoreState) => state);
 
+  const packageFetchCluster = packageCluster === "-" ? kubeappsCluster : packageCluster;
   const [isDeploying, setDeploying] = useState(false);
   const [releaseName, setReleaseName] = useState("");
   const [appValues, setAppValues] = useState(selectedPackage.values || "");
@@ -76,7 +78,7 @@ export default function DeploymentForm() {
 
   const [packageReference] = useState({
     context: {
-      cluster: packageCluster,
+      cluster: packageFetchCluster,
       namespace: packageNamespace,
     },
     plugin: pluginObj,
--- a/dashboard/src/components/PackageHeader/PackageView.tsx
+++ b/dashboard/src/components/PackageHeader/PackageView.tsx
@@ -49,13 +49,15 @@ export default function PackageView() {
   } = ReactRouter.useParams() as IRouteParams;
   const {
     packages: { isFetching, selected: selectedPackage },
-    config: { skipAvailablePackageDetails },
+    config: { skipAvailablePackageDetails, kubeappsCluster },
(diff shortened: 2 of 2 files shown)


absoludity:
> 
> Related frontend changes for #4591. These changes are required so that the route params are correctly translated.

See inline, wondering why we can't use this token `-` in this case, rather than an empty string (if that's what's happening).

> ### Todo
> 
>     * Test coverage for "" cluster for routing and fetching
> 
> 
> @absoludity when I was testing locally I also ran into something new. When trying to deploy a helm chart I run into the following:
> 
> ```
> An error occurred: Unable to create the installed package for the package "coreweave/argo-cd" using the plugin "helm.packages": rpc error: code = Internal desc = Unable to create kubernetes clientset: rpc error: code = FailedPrecondition desc = unable to get client : rpc error: code = FailedPrecondition desc = unable to get controller runtime client due to: Unauthorized
> ```

Hmm, is this possibly related to the comment I left in the description of my #4595 : 

> We need to test this as I suspect, if installing a package still requires fetching the AppRepository, that it would mean that the user credential must be valid to authenticate with the cluster on which Kubeapps is installed. The user doesn't need any RBAC, only the ability to authenticate, since Kubeapps already includes RBAC to ensure that global repositories are readable by any authenticated user.

Does your user have at least the ability to authenticate with the cluster on which Kubeapps is installed? Currently this would be required (even if they have no permissions), until we remove the need for Kubeapps to fetch the app repository using the users credentials (or possibly update the plugin to use its own service account in this instance only, when the kubeapps cluster is `-`).

Thanks for taking the changes in my PR for a spin!

aanthonyrizzo:
> See inline, wondering why we can't use this token - in this case, rather than an empty string (if that's what's happening).

Sorry the title is misleading. I needed to add some logic that correctly translates the `-` to the kubeappsCluster (which in my case is `""`).

> Hmm, is this possibly related to the comment I left in the description of my https://github.com/vmware-tanzu/kubeapps/issues/4595 

I don't think so. I've run into those errors before when I had issues with my app repositories. When I was testing the UI I didn't have any issues fetching or reading the helm charts.

> Does your user have at least the ability to authenticate with the cluster on which Kubeapps is installed? Currently this would be required (even if they have no permissions), until we remove the need for Kubeapps to fetch the app repository using the users credentials (or possibly update the plugin to use its own service account in this instance only, when the kubeapps cluster is -).

Yup! The account I was testing with has a cluster-admin role for the cluster with kubeapps so I dont think there should be any issues authenticating. I will keep try to debug and then circle back later this week if I cant get it.

I will also add the tests for this in a day or so. Sorry for the long delays.

antgamdia:
Let me close the PR for now, but please free to reopen it when once you have a chance to get to it again. Thanks!

PR #4618, opened 2022-04-26
step 7 for  Investigate and propose package repositories API with similar core interface to packages API. #3496 

gfichtenholt:
added .proto for DeletePackageRepository and flux implementation with basic unit and integration tests


absoludity (approved):
Looks good. Are you following up with the code to delete the related secrets? (kubeapps managed)?

gfichtenholt:
> Looks good. Are you following up with the code to delete the related secrets? (kubeapps managed)?

yes indeed. Thanks for the review

PR #4623, opened 2022-04-27
Return NotFound immediately when kapp configmap not found.

### LINKED ISSUE #4614: [carvel] `GetInstalledPackageResourceRefs` fails behind contour
### Description:

When using Contour (h2c, `application/grpc-web+proto`, just token-based login) `GetInstalledPackageResourceRefs` seems to fail for Carvel packages.
Specifically, I've been able to identify two scenarios: i) an OK package; ii) an unreconciled package.
While for i) the error really visible, in ii) it is navigation-breaking.

#### Case i)

In this scenario, there's no visible error, but the console shows:

image

Note the error:

```console
Failed to load resource: net::ERR_INCOMPLETE_CHUNKED_ENCODING
```

#### Case ii)

image

This is consistent with an error on the server-side:

````console
I0425 17:12:04.722787       1 server.go:59] OK 413.7089ms /kubeappsapis.core.packages.v1alpha1.PackagesService/GetAvailablePackageDetail
I0425 17:12:18.095622       1 server.go:59] Internal 14.9762018s /kubeappsapis.core.packages.v1alpha1.PackagesService/GetInstalledPackageResourceRefs
````

Besides, it is a concerning error because the user actually see a "working" package (where they get notified about the non-reconcialtion cause):
image

But after the get resources timer get executed (in about 2 seconds or so), immediately the screen becomes:

image

1.Install a Carvel package, ie. Contour
2. Wait until it fails to reconcile
3. Access the package details view
4. Wait for a couple of seconds

See above.

Everything working fine or, if there is an actual error, at least do not wipe down the entire page already loaded.

N/A

N/A - "Kubeapps running behind Contour"

### PR DESCRIPTION
**Pull request by absoludity**

### Description of the change

Updates the `GetInstalledPackageResourceRefs()` call so that it will not wait indefinitely (or until a timeout) for the kapp config map to have been created.

Instead, the code now returns a NotFound error immediately, which is already handled by the dashboard code with an explanatory comment at:

https://github.com/vmware-tanzu/kubeapps/blob/06774a4476219b6e031b5b2fb436f1c12a8048bd/dashboard/src/components/AppView/AppView.tsx#L180-L187

### Benefits

We neither wait for a timeout, nor do we switch to an error after the timeout, fixing case 2 of #4614 .

### Applicable issues

- fixes #4614

### Additional information

I'll now check the case 1 situation, which may be harder (optimizing the response when all the data is there). EDIT: Nothing to do there, even for a kubeapps carvel install, `GetInstallPackageResourceRefs` is returning in under 500ms.

### DIFF
--- a/cmd/kubeapps-apis/Dockerfile
+++ b/cmd/kubeapps-apis/Dockerfile
@@ -6,8 +6,6 @@
 FROM bitnami/golang:1.17.6 as builder
 WORKDIR /go/src/github.com/vmware-tanzu/kubeapps
 COPY go.mod go.sum ./
-COPY pkg pkg
-COPY cmd cmd
 ARG VERSION="devel"
 
 ARG BUF_VERSION="1.0.0-rc11"
@@ -18,6 +16,11 @@ RUN curl -sSL "https://github.com/bufbuild/buf/releases/download/v$BUF_VERSION/b
 RUN --mount=type=cache,target=/go/pkg/mod  \
     --mount=type=cache,target=/root/.cache/go-build \
     go mod download
+
+# We don't copy the pkg and cmd directories until here so the above layers can
+# be reused.
+COPY pkg pkg
+COPY cmd cmd
 
 # Lint the proto files to detect errors at build time
 RUN /tmp/buf lint ./cmd/kubeapps-apis
--- a/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_resources.go
+++ b/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_resources.go
@@ -6,14 +6,13 @@ package main
 import (
 	"context"
 	"fmt"
-	"time"
+	"strings"
 
 	ctlres "github.com/k14s/kapp/pkg/kapp/resources"
 	kappctrlv1alpha1 "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/kappctrl/v1alpha1"
 	packagingv1alpha1 "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/packaging/v1alpha1"
 	datapackagingv1alpha1 "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apiserver/apis/datapackaging/v1alpha1"
 	corev1 "github.com/vmware-tanzu/kubeapps/cmd/kubeapps-apis/gen/core/packages/v1alpha1"
-	"github.com/vmware-tanzu/kubeapps/cmd/kubeapps-apis/plugins/pkg/k8sutils"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_resources.go [hunk 1]
```diff
@@ -425,22 +424,6 @@ func (s *Server) inspectKappK8sResources(ctx context.Context, cluster, namespace
...
-	if err != nil {
-		return nil, status.Errorf(codes.Internal, "unable to get the k8s client: '%v'", err)
-	}
-
-	gvr := schema.GroupVersionResource{Group: "", Version: "v1", Resource: "configmaps"}
-	resource := dynClient.Resource(gvr).Namespace(namespace)
-
-	// In order to be able to fetch the resources created by the kapp-controller, we need to fetch a ConfigMap
-	// that contains the label (the application id) used for query the k8s resources.
-	// We actively wait for this ConfigMap to be present in the cluster before returning the list of resources
```
antgamdia, line 438: I wonder if should still wait for a couple of seconds (not infinitely), for the CM to be created in those cases everything is fine and it is just a matter of giving some time to k8s to actually create the resource.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_resources.go
```diff
@@ -458,6 +441,15 @@ func (s *Server) inspectKappK8sResources(ctx context.Context, cluster, namespace
...
 	usedGVs, err := app.UsedGVs()
 	if err != nil {
+		// TODO(minelson): We can't currently use `errors.IsNotFound(err)` here.
+		// See https://github.com/vmware-tanzu/carvel-kapp/issues/498
+		// Instead we need to match on the error string :/
+		if strings.Contains(err.Error(), fmt.Sprintf("configmaps %q not found", appName)) {
+			// We want to return a NotFound here because the dashboard already
+			// handles this case, knowing that the references may not be
+			// available immediately.
+			return nil, status.Errorf(codes.NotFound, "App not found: %+v", err)
```
antgamdia, line 451: Besides, perhaps a `NotFound`error perhaps is not the best fit, I'd rather use the `UNAVAILABLE` code, whose semantics is more aligned to sth like "ey client, right now I don't have it, but you can wait with some backoff if you want"

antgamdia (approved):
+1 as it solves the issue, but I wonder if it would lead to undesired behaviors in the frontend (like a 404 error right after creating the app, which is what we tried to address when adding the active wait). 
Perhaps a follow-up PR adding the retry in the package details view?

absoludity (commented):
> +1 as it solves the issue, but I wonder if it would lead to undesired behaviors in the frontend (like a 404 error right after creating the app, which is what we tried to address when adding the active wait).

Nope, it does not (at least in my testing). 

Initially we would get 404s in the dashboard, so the carvel plugin was updated with the active wait, but the flux one wasn't. We updated the dashboard instead with an exponential backoff which worked fine for most cases, but not when the app would never be created due to the service account RBAC. In this case, the user would eventually see a 404.

So in #4338 we updated the dashboard to continue to poll (not exponentially) so that the flux plugin can simply return not found in that case, much simpler, we just hadn't updated the carvel plugin afaict.  Longer term we want to have an extra state or something so clients no that it's not even ready to be queried for the resource refs or other data, that's #4337 which I created when implementing #4338 . Hope that's a bit clearer.

> Perhaps a follow-up PR adding the retry in the package details view?

That's exactly the code which the linked comment in the description points to (you can't see from the preview, but follow the link).

cmd/kubeapps-apis/Dockerfile
```diff
@@ -19,6 +17,11 @@ RUN --mount=type=cache,target=/go/pkg/mod  \
     --mount=type=cache,target=/root/.cache/go-build \
     go mod download
 
+# We don't copy the pkg and cmd directories until here so the above layers can
```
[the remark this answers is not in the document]
absoludity: Checked today - none of the other Dockerfiles have a large download (like buf) between copying the source files and building.

PR #4648, opened 2022-05-03
Rename `additionalFlags` to `extraFlags`

antgamdia:
### Description of the change

I've noticed we left unchanged some of our docs during the chart standardization. Specifically, with respect to this change:
>- `authProxy.additionalFlags` renamed as `authProxy.extraFlags`

This PR is to update our docs to to use `extraFlags` instead of `additionalFlags`.

### Benefits

Our docs will be up to date and will be consistent with our current chart values.

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

There is still another file using `additionalFlags`, but it is the chart readme, so we better change it upstream, so that the change get published sooner.


absoludity (approved):
Thanks Antonio. I added the change to the chart PR upstream too.

PR #4652, opened 2022-05-05
hide sensitive data for GetPackageRepoDetail and UpdatePackageRepository with kubeapps managed secrets

gfichtenholt:
Documented behavior in the spec (under GetPackageRepositoryDetail 
https://docs.google.com/document/d/1z03msZRGsRvcRaom-yrvEwIWcEfNy6fSA5Zg28gjYvA/edit#heading=h.ou8mya83exsw and UpdatePackageRepository https://docs.google.com/document/d/1z03msZRGsRvcRaom-yrvEwIWcEfNy6fSA5Zg28gjYvA/edit#heading=h.gd8t5sfz32li) as per agreement with team
Implemented according to spec in the flux plugin

### DIFF
--- a/cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo.go
+++ b/cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo.go
@@ -45,6 +45,7 @@ const (
 var (
 	// default poll interval is 10 min
 	defaultPollInterval = metav1.Duration{Duration: 10 * time.Minute}
+	redactedString      = "REDACTED"
 )
 
 // returns a list of HelmRepositories from all namespaces (cluster-wide), excluding
@@ -207,10 +208,10 @@ func (s *Server) newRepo(ctx context.Context, targetName types.NamespacedName, u
 		return nil, status.Errorf(codes.InvalidArgument, "TLS flag insecureSkipVerify is not supported")
 	}
 
-	var secretRef string
+	var secret *apiv1.Secret
 	var err error
 	if s.pluginConfig.UserManagedSecrets {
-		if secretRef, err = s.validateRepoUserManagedSecret(ctx, targetName, tlsConfig, auth); err != nil {
+		if secret, err = s.validateUserManagedRepoSecret(ctx, targetName, tlsConfig, auth); err != nil {
 			return nil, err
 		}
 	} else {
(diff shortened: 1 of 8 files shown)


cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo.go
```diff
@@ -45,6 +45,7 @@ const (
 var (
 	// default poll interval is 10 min
 	defaultPollInterval = metav1.Duration{Duration: 10 * time.Minute}
+	redactedString      = "REDACTED"
```
absoludity, line 48: This should be a `const` rather than a package-global variable shouldn't it?

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo.go
```diff
@@ -469,78 +464,83 @@ func (s *Server) createRepoKubeappsManagedSecret(
...
+				*metav1.NewControllerRef(
+					repo,
+					schema.GroupVersionKind{
+						Group:   sourcev1.GroupVersion.Group,
+						Version: sourcev1.GroupVersion.Version,
+						Kind:    sourcev1.HelmRepositoryKind,
+					}),
+			}
+			if _, err := secretsInterface.Update(ctx, secret, metav1.UpdateOptions{}); err != nil {
+				return statuserror.FromK8sError("update", "secrets", secret.Name, err)
```
absoludity, line 485: Nice - so the main change to this file is just the return type of `validateUserManagedRepoSecret` and so that the arg to `setOwnerReferencesForRepoSecret` can include the secret (and avoid an extra lookup?)

absoludity (approved):
Thanks Greg

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo.go
gfichtenholt: 👍

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo.go
gfichtenholt: > Nice - so the main change to this file is just the return type of `validateUserManagedRepoSecret` and so that the arg to `setOwnerReferencesForRepoSecret` can include the secret (and avoid an extra lookup?)

Well, I would say the 'redacted' string handling, is the main change to this file. What you mention is just a small unrelated performance improvement, an icing on the cake, so to speak.

Thanks for the review
Greg

Issue #3399, opened 2021-09-13
Implement pagination in the aggregated core APIs


antgamdia:
It will require some changes in the Helm plugin, the core API and the UI.

From https://github.com/kubeapps/kubeapps/pull/3376#discussion_r706959608
> I think we already have one in mind, we just need to update the existing pagination (throughout) to use the number of items as the offset, rather than the number of pages, to then be able to update the next page tokens in the plugins.

absoludity:
And the current asset svc helpers used by the core api, I think. In the sense that the current `GetPaginatedChartListWithFilters()` takes the `pageNumber` for an offset, even though it's converted to an item offset in the SQL. We just need to update the arg to be an item offset itself (should be trivial) and update the call-sites (or even delete the old call-sites if we don't need them :) ).

absoludity:
The actual change (after the prep work) for this is done in #4694 and is working nicely, other than one issue with the UX creating multiple requests with different pagination options. I'll work on that next, which should be simple. Once done, the only other remaining work will be to apply the same change to the `GetInstalledPackageSummaries`.

ppbaena:
This issue can be closed then.

absoludity:
Hah - indeed. With these issues with many PR's, I avoid adding a `fixes #` to the PRs so they don't automatically close after each PR lands and need re-opening... but need to remember to close at the end :P

PR #4653, opened 2022-05-05
Remove knowledge of pagination internals from UX/Client

absoludity:
### Description of the change

This is the first step towards fixing #3399 . When we initially updated the dashboard to use the new UX, we'd encoded knowledge of the pagination token in the frontend, which we really need to avoid. The plugin should simply be able to set the `NextPageToken` in the response and the client sends that back with the next request.

The main change is to remove the `page` (int) state and replace it with the token verbatim which is returned by the API. There was one existing point where we need to track which page number to avoid jitter (? not sure, didn't verify, just maintained the same behaviour) so the page state is still counted within the catalog view but not used for anything other than calculating `isFirstPage` for the CatalogItems component.

### Benefits

I'd like to encode more info into the next page token so that the core packages plugin is able to track pagination across plugins.

### Possible drawbacks

Potential issues paginating with the current plugins. I've tested the helm plugin in real life and it works fine, paginating the results until finished.

### Applicable issues

- ref #3399 

### Additional information

One thing I've noticed, which isn't new, is that we continue paginating regardless of the scroll position at the moment. Each time a response comes back, the state changes (used to be page, now it's nextPageToken) which triggers the effect again to request the next results, until finished.

### DIFF
--- a/dashboard/src/actions/availablepackages.test.tsx
+++ b/dashboard/src/actions/availablepackages.test.tsx
@@ -25,7 +25,7 @@ let store: any;
 const namespace = "package-namespace";
 const cluster = "default";
 const repos = "foo";
-const defaultPage = 1;
+const defaultPaginationToken = "defaultPaginationToken";
 const defaultSize = 0;
 const plugin = { name: "my.plugin", version: "0.0.1" } as Plugin;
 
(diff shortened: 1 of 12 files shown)


absoludity:
Closing this in favour of #4654 (only difference being the branch is on our vmware-tanzu org so I can stack other branches there).

PR #4654, opened 2022-05-05
Remove knowledge of pagination internals from UX/Client

absoludity:
### Description of the change

This is the first step towards fixing #3399 . When we initially updated the dashboard to use the new UX, we'd encoded knowledge of the pagination token in the frontend, which we really need to avoid. The plugin should simply be able to set the `NextPageToken` in the response and the client sends that back with the next request.

The main change is to remove the `page` (int) state and replace it with the token verbatim which is returned by the API. There was one existing point where we need to track which page number to avoid jitter (? not sure, didn't verify, just maintained the same behaviour) so the page state is still counted within the catalog view but not used for anything other than calculating `isFirstPage` for the CatalogItems component.

### Benefits

I'd like to encode more info into the next page token so that the core packages plugin is able to track pagination across plugins.

### Possible drawbacks

Potential issues paginating with the current plugins. I've tested the helm plugin in real life and it works fine, paginating the results until finished.

### Applicable issues

- ref #3399 

### Additional information

One thing I've noticed, which isn't new, is that we continue paginating regardless of the scroll position at the moment. Each time a response comes back, the state changes (used to be page, now it's nextPageToken) which triggers the effect again to request the next results, until finished.

### DIFF
--- a/dashboard/src/components/Catalog/Catalog.tsx
+++ b/dashboard/src/components/Catalog/Catalog.tsx
@@ -134,7 +134,16 @@ export default function Catalog() {
         searchFilter,
       ),
     );
-  }, [dispatch, nextPageToken, size, cluster, namespace, reposFilter, searchFilter]);
+  }, [
+    dispatch,
+    nextPageToken,
+    size,
+    cluster,
+    namespace,
+    reposFilter,
+    searchFilter,
+    hasFinishedFetching,
+  ]);
 
   // hasLoadedFirstPage is used to not bump the current page until the first page is fully
   // requested first
--- a/dashboard/src/components/Catalog/Catalog.tsx
+++ b/dashboard/src/components/Catalog/Catalog.tsx
@@ -266,7 +266,7 @@ export default function Catalog() {
     .filter(
       c =>
         filters[filterNames.CATEGORY].length === 0 ||
-        intersection(filters[filterNames.CATEGORY], getOperatorCategories(c)).length > 0,
+        intersection(filters[filterNames.CATEGORY], getOperatorCategories(c)).length,
     );
 
   // Required to have the latest value of page
--- a/dashboard/src/components/Catalog/Catalog.tsx
+++ b/dashboard/src/components/Catalog/Catalog.tsx
@@ -121,6 +121,9 @@ export default function Catalog() {
   const searchFilter = filters[filterNames.SEARCH]?.toString().replace(tmpStrRegex, ",") || "";
   const reposFilter = filters[filterNames.REPO]?.join(",") || "";
   useEffect(() => {
+    if (hasFinishedFetching) {
+      return;
+    }
     dispatch(
       actions.availablepackages.fetchAvailablePackageSummaries(
         cluster,
--- a/dashboard/src/components/Catalog/Catalog.tsx
+++ b/dashboard/src/components/Catalog/Catalog.tsx
@@ -79,6 +79,7 @@ export default function Catalog() {
   const {
     packages: {
       hasFinishedFetching,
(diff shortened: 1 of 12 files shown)


dashboard/src/reducers/availablepackages.ts
```diff
@@ -67,13 +68,19 @@ const packageReducer = (
...
+      // the first page and *after* we've fetched the last page. In both
+      // cases, the nextPageToken is empty. Since we have just fetched a
+      // page, only one option is possible.
+      const isLastPage = action.payload.response.nextPageToken === "";
       return {
         ...state,
         isFetching: false,
+        // TODO(minelson): Something is triggering a new request even after
+        // hasFinishedFetching is true, the response for which resets it.
+        // Or it could be that double requests are sent (seen in server logs)
```
antgamdia, line 81: Haven't checked locally, but this is tricky... just a minimum changing altering this pagination state machine starts triggering unnecessary requests. It is a pain, but let's resolve it in a second iteration once we have everything working, so +1 as is.

dashboard/src/components/Catalog/Catalog.tsx
```diff
@@ -250,7 +266,7 @@ export default function Catalog() {
     .filter(
       c =>
         filters[filterNames.CATEGORY].length === 0 ||
-        intersection(filters[filterNames.CATEGORY], getOperatorCategories(c)).length,
+        intersection(filters[filterNames.CATEGORY], getOperatorCategories(c)).length > 0,
```
antgamdia, line 269: Out of curiosity: any special reason for this change? wouldn't just `...X.length` check every falsy value, which subsumes the length===0 case ?

dashboard/src/components/Catalog/Catalog.tsx
```diff
@@ -94,7 +95,10 @@ export default function Catalog() {
   const dispatch = useDispatch();
 
   const [filters, setFilters] = React.useState(initialFilterState());
-  const [page, setPage] = React.useState(0);
+  // pageNum is only used to avoid flicker in the CatalogItems.
+  // It is no longer used for pagination which is handled by the server's
+  // opaque nextPageToken etc.
+  const [pageNum, setPageNum] = React.useState(0);
```
antgamdia, line 101: We might want to consider a map of triggered requests and their status? This way, when unmounting the component, we could cancel any unnecessary ongoing requests. Besides, it could help us avoid duplicated requests being fired at the same time?

antgamdia (approved):
Awesome! A step closer to the all-plugin pagination! Thanks

dashboard/src/components/Catalog/Catalog.tsx
absoludity: Other than a tendency to prefer explicit over implicit, I don't remember. Vague memory of a test failing and having to update code, but can't see how that could have affected this line.

dashboard/src/reducers/availablepackages.ts
absoludity: Oh - I actually meant to remove this comment as I solved that issue by adding the extra guard in the effect of Catalog.tsx which triggered the request:
```
    if (hasFinishedFetching) {
      return;
    }
```

The issue was that if the last page token changed, it would always initiate a new request, even if it'd finished fetching the whole series. So when it got to the last page, and the page token came back empty, it caused the extra request. Let me delete that comment.

dashboard/src/components/Catalog/Catalog.tsx
absoludity: Also reverting this change, since I need to update the branch anyway.

PR #4657, opened 2022-05-06
Use pagination options with plugin when only one plugin configured.

absoludity:

### Description of the change

Updates the core packages plugin so that if only a single plugin is used, it will pass through the pagination options and return the next page token of the plugin.

This small change (on top of #4654) will allow me to start updating the individual plugins without affecting the combined behaviour for now.

### Benefits

I can continue by next updating the helm plugin from a page number to an item number.

### Applicable issues

- ref #3399


antgamdia (approved):
Great, good idea. Deferring the pagination back to the plugin avoids dealing with the pagination multiple times. Thanks!

PR #4659, opened 2022-05-09
Update Helm plugin to use an item offset.

absoludity:

### Description of the change

Follows on from #4657 , updating the Helm plugin (and associated assetsvc helpers) so that an item offset is used for pagination (within the page token), rather than a page offset.

Verified in real life as well:
pagination-by-item-offset

### Benefits

Step closer to encoding item offsets for each plugin in the core pagination.

### Applicable issues

- ref #3399 

### Additional information

I also tested the pagination of the installed packages on the server only to find that we don't actually use this from our UX client at all (it currently does not send nor handle pagination options). We've had one report recently of someone with hundreds of installed carvel packages, so may want to update the UX to do this. I'll create an issue separately.


cmd/assetsvc/pkg/utils/postgres_db_test.go
```diff
@@ -374,14 +374,11 @@ func TestGetPaginatedChartList(t *testing.T) {
 				}
 			}
 
-			charts, numPages, err := pam.GetPaginatedChartListWithFilters(ChartQuery{Namespace: tc.namespace, Repos: []string{tc.repo}}, 1, 10)
+			charts, err := pam.GetPaginatedChartListWithFilters(ChartQuery{Namespace: tc.namespace, Repos: []string{tc.repo}}, 0, 10)
```
absoludity, line 377: I've removed the `numPages` part of the return for the function since it isn't currently used in real life anyway (I assume it was to be able to present `page 1 of 10` or similar type pagination, which we do not do.

cmd/assetsvc/server/handler.go
```diff
@@ -297,30 +276,6 @@ func getChartVersionSchema(w http.ResponseWriter, req *http.Request, params Para
...
-// listChartsWithFilters returns the list of repos that contains the given chart and the latest version found
-func listChartsWithFilters(w http.ResponseWriter, req *http.Request, params Params) {
-	namespace, repo, _, paramErr, err := extractDecodedNamespaceAndRepoAndVersionParams(params)
-	if err != nil {
-		handleDecodeError(paramErr, w, err)
-		return
-	}
-	cq := extractChartQueryFromRequest(namespace, repo, req)
-
-	pageNumber, pageSize := getPageAndSizeParams(req)
```
absoludity, line 309: This was the only non-test usage of the page number, and it's a handler for an API that we no longer used (replaced by the helm plugin), so just removed it.

cmd/kubeapps-apis/plugins/pkg/paginate/paginate.go
```diff
@@ -41,9 +41,16 @@ func PageOffsetFromInstalledRequest(request *corev1.GetInstalledPackageSummaries
...
 	offset, err := PageOffsetFromPageToken(request.GetPaginationOptions().GetPageToken())
 	if err != nil {
-		return 0, status.Errorf(codes.InvalidArgument, "unable to intepret page token %q: %v",
-			request.GetPaginationOptions().GetPageToken(), err)
+		return 0, err
 	} else {
 		return offset, nil
 	}
 }
+
```
absoludity, line 49: Note: it also wasn't obvious to me why we have the two separate functions above that do exactly the same thing, just for different types of requests. We can instead just have call-sites that pass in the pagination options rather than the request, I'd think? (which is what I'm doing below with the similar ItemOffset helper). Anyway, the above will be removed once the other two plugins are similarly updated.

castelblanque: Yep, maybe there is a reason for that, but indeed makes more sense having a single function with the `request.GetPaginationOptions()` type as param.

cmd/assetsvc/pkg/utils/postgres_db_test.go
antgamdia: Yep, that's precisely the reason why I added it. Right now, our UI is not displaying it, but, I still think letting the user know how many pages they is expected to fetch (thinking of a progress bar, or a more sophisticated scrolling like google photos ([interesting read here](https://medium.com/google-design/google-photos-45b714dfbed1)).
Without this value, we would have to calculate that number on the client's side, no?

Anyway, I'm ok with it as we don't have any real use case right now.

cmd/assetsvc/pkg/utils/postgresql_utils.go
```diff
@@ -44,34 +43,22 @@ func (m *PostgresAssetManager) GetAllChartCategories(cq ChartQuery) ([]*models.C
...
-	if pageNumber <= 0 {
-		pageNumber = 1
-	}
-
+func (m *PostgresAssetManager) GetPaginatedChartList(whereQuery string, whereQueryParams []interface{}, startItemNumber, pageSize int) ([]*models.Chart, error) {
 	paginationClause := ""
 	if pageSize > 0 {
-		offset := (pageNumber - 1) * pageSize
-		paginationClause = fmt.Sprintf("LIMIT %d OFFSET %d", pageSize, offset)
+		paginationClause = fmt.Sprintf("LIMIT %d", pageSize)
```
antgamdia, line 49: Each time I come across this file I wish we had [an ORM](https://pg.uptrace.dev/queries/#writing-queries)... 😢 or even moving towards a single-database model across plugins.

Anyway, for the scope of this PR, +1 to the change of course!

cmd/kubeapps-apis/plugins/pkg/paginate/paginate.go
antgamdia: ooor, we can start using [generics](https://go.dev/doc/tutorial/generics) :P 
Sometimes I wish go supported overloading...

antgamdia (approved):
That's great! Thanks for the changes here!

PR #4671, opened 2022-05-09
Update deps (except go ones)

antgamdia:
### Description of the change

This PR updates several dependencies in our codebase (frontend, rust and just k8s-related ones in go).

Minor changes have been performed for fixing a type-related issue in the frontend.

### Benefits

Up to date deps.

### Possible drawbacks

N/A

### Applicable issues

- related#4661

### Additional information

N/A


dashboard/src/components/ErrorAlert/ErrorAlert.tsx
```diff
@@ -45,8 +45,10 @@ export default function ErrorAlert({ error, children }: IErrorAlert) {
   }
   return (
     <Alert theme="danger">
-      {messages}
-      {children}
+      <>
+        {messages}
+        {children}
+      </>
```
absoludity, line 51: Is this the type related issue? Don't understand why we need an extra nesting here given it's already nested inside the `<Alert />`?

absoludity (approved):

dashboard/src/components/ErrorAlert/ErrorAlert.tsx
antgamdia: I was getting this error:

````
This JSX tag's 'children' prop expects a single child of type 'string | number | boolean | ReactElementLike | ReactNodeArray', but multiple children were provided.  TS2746

    46 |   }
    47 |   return (
  > 48 |     <Alert theme="danger">
       |      ^
    49 |       {messages}
    50 |       {children}
    51 |     </Alert>
````

I don't know why the `PropTypes.node` type is not the same as the `React.ReactNode`... however, I've just changed the type definition to `children: React.ReactNode || PropTypes.node` and removed the highlighted workaround and it works now.

Issue #4661, opened 2022-05-09
Bump up deps and versions


antgamdia:
I've noticed we have some stuck dependabot's PRs requiring changes, also GKE has updated their _stable_ k8s default version and we haven't updated the CI versions var yet, etc.
Specifically, we should follow the release process with regard to the upgrading process, namely:

- Pending PRs:
    - [x] https://github.com/vmware-tanzu/kubeapps/pull/4642
        - Trivial rename in a type name, no user-faced effect whatsoever.
    - [x] https://github.com/vmware-tanzu/kubeapps/pull/4656
        -  Add some pkgs in the `transformIgnorePatterns` when running jest.
    - [ ] https://github.com/vmware-tanzu/kubeapps/pull/4605
        - Use a new options object and fix tests. 
- Frontend-related
    - [x] https://github.com/vmware-tanzu/kubeapps/pull/4664 
    - [x] https://github.com/vmware-tanzu/kubeapps/pull/4665
- Project-related
    - [x] https://github.com/vmware-tanzu/kubeapps/pull/4666
    - [x] https://github.com/vmware-tanzu/kubeapps/pull/4670
- CI-related
    - [x] https://github.com/vmware-tanzu/kubeapps/pull/4667
    - [x] https://github.com/vmware-tanzu/kubeapps/pull/4668
- Backend-related
    - [x] https://github.com/vmware-tanzu/kubeapps/pull/4669
    - [x] https://github.com/vmware-tanzu/kubeapps/pull/4671
    - [x] https://github.com/vmware-tanzu/kubeapps/pull/4674
    - [x] https://github.com/vmware-tanzu/kubeapps/pull/4687

absoludity:
I had added some to the Kubeapps project - (for eg, if you look at the first bump PR #4642 above, you'll see it has the vmware-tanzu/kubeapps project. I had expected that would make it visible on the project page, but looks like that currently only shows issues. Worth checking as it would be nice to add such bump PRs requiring work to the board without needing a separate issue to track them (haven't checked, but will).

> Specifically, we should follow the release process with regard to the upgrading process

I did specifically mention in the standup before I did the last release that I'd skip the upgrading process to just do a quick release, knowing that we'd need to do them afterwards. You don't need to pick them up, but feel free :)

antgamdia:
[...]

PR #4674, opened 2022-05-09
Update go deps

antgamdia:
### Description of the change

To better trace any error, this is a follow-up of #4671 just updating the transitive go dependencies (the main ones, I'll defer them to dependabot, so that we can be testing each one separately)

### Benefits

Up to date deps.

### Possible drawbacks

N/A

### Applicable issues

- related #4661

### Additional information

Let's see if the CI thinks otherwise :P


absoludity (approved):
+1 when conflicts resolved.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_adapters.go
```diff
@@ -8,7 +8,6 @@ import (
 	"strings"
 	"time"
 
-	kappctrlinstalled "github.com/vmware-tanzu/carvel-kapp-controller/cli/pkg/kctrl/cmd/package/installed"
```
antgamdia, line 11: This dependency (the only one occurrence) was blocking the upgrade path for the kapp-controller itself (it is still pointing to v0.27). I've replaced it with the actual value `%s-%s-values"` of the template + a comment.

PR #4678, opened 2022-05-10
Update Carvel plugin to use an item offset rather than page offset.

absoludity:

### Description of the change

See #3399 for the background. This PR follows #4659 which updated the Helm plugin to use an item offset rather than a page offset, and makes the same change for the kapp_controller plugin (which just works seamlessly now that the Dashboard UX client does not introspect the page token at all).

Checked in real life:

carvel-item-based-offset

### Benefits

2 down, 1 to go (Flux). This will enable to core plugin to interleave paginated results from plugins.

### Applicable issues

- Ref #3399


castelblanque:
LGTM

castelblanque (approved):

PR #4679, opened 2022-05-10
Update flux plugin to use item offset rather than page offset.

absoludity:

### Description of the change

Follows on from #4678 with the similar change for the flux plugin.

Verified it works in real life:
flux-item-offset

The existing unit tests for pagination still pass because they use a pageSize of 1, so that an itemOffset is the same as a pageOffset.

Haven't checked @gfichtenholt 's integration tests. Might give them a spin after lunch and see (but they shouldn't be affected unless they're introspecting the page token).

~~Hmm, also didn't update the repo endpoints which I assume use the same/similar pagination. Marking this as a draft until I've done that~~ Checked, they don't use pagination.

### Benefits

Now I can do the core pagination :)

### Applicable issues

- ref #3399


gfichtenholt (approved):
LGTM. Noticed you didn't have to update any of the unit tests. Was that as intended?

absoludity:
> LGTM. Noticed you didn't have to update any of the unit tests. Was that as intended?

I expected pagination unit tests to fail after my change, but they didn't, which baffled me until I checked why - from above:

> The existing unit tests for pagination still pass because they use a pageSize of 1, so that an itemOffset is the same as a pageOffset.

Thanks Greg.

PR #4683, opened 2022-05-10
Skip some tests when running on Windows

antgamdia:
### Description of the change

Since I am temporarily running a Windows system, I've noticed that some of our tests are not ready to run there (asserting non-existing error messages, dealing with files and env vars, etc...), so, even we decrease the test coverage when running tests on windows, at least, we won't get consistently failing tests anymore.

### Benefits

We will be able to run our test suite on Windows without false positives.

### Possible drawbacks

Less coverage on Windows.

### Applicable issues

N/A

### Additional information

N/A

### DIFF
--- a/cmd/kubeapps-apis/cmd/root_test.go
+++ b/cmd/kubeapps-apis/cmd/root_test.go
@@ -5,6 +5,7 @@ package cmd
 
 import (
 	"bytes"
+	"runtime"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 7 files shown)


castelblanque:
Is it a problem that we could tackle to make the tests pass in both OSs? I would expect the env variables to be handled OOB by Go, and being maybe the file paths being problematic. Is that something we could "fix" from our code base? e.g. Don't know about Go, but in Java there are variables for path separator (`/` vs `\`) that are automatically set by the JVM.

castelblanque:
Ok, then there's too much to investigate here 😃 
Agree leaving it with the skip for now.

castelblanque (approved):

Issue #4692, opened 2022-05-12
make GetAvailablePackageSummaries more robust to potential plugin failures


dlaloue-vmware:
there are cases where one plugin can fail when calling GetAvailablePackageSummaries (for example postgres issue for helm plugin, bug in carvel plugin for empty result...), and in that case, the GetAvailablePackageSummaries call itself fails.

it should be possible for the core layer to catch those errors and report them along with the results from other plugins, rather than fail altogether.

absoludity:
Thanks Dimitri. Yeah, I think Greg and I have discussed this at one point, and given the options between silently ignoring that plugin and returning results from the others, vs exposing the error directly, I'm keen to expose the error directly.

But what you're proposing is I think trying to get the best of both worlds, where we return the results from the plugins without errors and *also* somehow notify that there was an issue with some plugins? It's tricky to see what the status would be for such a call (did it succeed? what if the plugin error was temporary, so the call for the next page then includes some results while the first didn't  etc.)

I need to think about it more, but I still think I'd prefer to return an error if there was an error retrieving the aggregated results. What do others think?

dlaloue-vmware:
yes, the goal is just to provide more robustness.
the GetAvailablePackageSummariesResponse already has several fields. it would be easy to add errors and from which plugin they are, and let the UI display both the summary and a error banner.

antgamdia:
> I need to think about it more, but I still think I'd prefer to return an error if there was an error retrieving the aggregated results. What do others think?

I have also brought this stuff up in the past, even if I'm 100% for returning any error back to the user I still think everything should fail just because of a tiny error in a single plugin.
Perhaps returning an `ABORTED` [status code](https://developers.google.com/maps-booking/reference/grpc-api-v2/status_codes) (_if the client should retry at a higher-level, such as when a client-specified test-and-set fails which indicates that the client should restart a read-modify-write sequence_) instead of an OK, but at least return something.

For example, a real escanrio I faced recently: I installed Kubeapps and then upgraded it via helm plugin but I wanted to manage carvel packages. Due to some incompatibilities with my chart values, the helm plugin was failing. Well, so as a result, I wouldn't be able to neither manage the installed packages via helm nor installing new packages with carvel.
I mean, of course, there certainly was something wrong with my kubeapps installation... but I would have expected to have been able to use kubeapps (even if in a _degradated mode_).

absoludity:
[...]

PR #4694, opened 2022-05-12
Enable fan-in of results for summaries from plugins.

absoludity:

### Description of the change

Updates the core plugin's GetAvailablePackageSummaries so that it outputs ordered results from multiple configured plugins.

It does this by deferring to a new helper function `fanInAvailablePackageSummaries()`. See inline comments for details.

Example of paginated responses coming back with combined, ordered results of Helm and Carvel packages:

combined-pagination

### Benefits

We can paginate aggregated results on the core packages `GetAvailablePackageSummaries` end point. 

### Possible drawbacks

Currently two issues I need to fix:

1. ~~Currently, it returns the combined results correctly for the first 70 results (helm + carvel), until, it seems, when it hits a streak where the whole page is from one plugin only, it then does not include the other plugins offset in the result, so that plugin begins again~~ (fixed)

2. When navigating away from the catalog, we're not resetting the next page token in the state so when returning, it fetches initially from the previous offsets, but then the dashboard also resets the state, and so there are multiple responses coming back, confusing the count.

I'll fix 2 in a separate PR so this can be reviewed as is (but not yet landed).

### Applicable issues

- ref #3399

### DIFF
--- a/cmd/kubeapps-apis/core/packages/v1alpha1/packages.go
+++ b/cmd/kubeapps-apis/core/packages/v1alpha1/packages.go
@@ -5,21 +5,21 @@ package v1alpha1
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 
 	. "github.com/ahmetb/go-linq/v3"
 	pluginsv1alpha1 "github.com/vmware-tanzu/kubeapps/cmd/kubeapps-apis/core/plugins/v1alpha1"
 	packages "github.com/vmware-tanzu/kubeapps/cmd/kubeapps-apis/gen/core/packages/v1alpha1"
 	"github.com/vmware-tanzu/kubeapps/cmd/kubeapps-apis/gen/core/plugins/v1alpha1"
-	"github.com/vmware-tanzu/kubeapps/cmd/kubeapps-apis/plugins/pkg/paginate"
 
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	log "k8s.io/klog/v2"
 )
 
-// pkgPluginsWithServer stores the plugin detail together with its implementation.
-type pkgPluginsWithServer struct {
+// pkgPluginWithServer stores the plugin detail together with its implementation.
+type pkgPluginWithServer struct {
 	plugin *v1alpha1.Plugin
 	server packages.PackagesServiceServer
 }
@@ -30,19 +30,19 @@ type packagesServer struct {
 
 	// pluginsWithServers is a slice of all registered pluginsWithServers which satisfy the core.packages.v1alpha1
 	// interface.
-	pluginsWithServers []pkgPluginsWithServer
+	pluginsWithServers []pkgPluginWithServer
 }
 
 func NewPackagesServer(pkgingPlugins []pluginsv1alpha1.PluginWithServer) (*packagesServer, error) {
 	// Verify that each plugin is indeed a packaging plugin while
 	// casting.
-	pluginsWithServer := make([]pkgPluginsWithServer, len(pkgingPlugins))
+	pluginsWithServer := make([]pkgPluginWithServer, len(pkgingPlugins))
 	for i, p := range pkgingPlugins {
 		pkgsSrv, ok := p.Server.(packages.PackagesServiceServer)
 		if !ok {
 			return nil, fmt.Errorf("Unable to convert plugin %v to core PackagesServicesServer", p)
 		}
-		pluginsWithServer[i] = pkgPluginsWithServer{
+		pluginsWithServer[i] = pkgPluginWithServer{
 			plugin: p.Plugin,
 			server: pkgsSrv,
 		}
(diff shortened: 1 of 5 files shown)


cmd/kubeapps-apis/plugin_test/mock_plugin.go
```diff
@@ -37,10 +39,21 @@ func (s TestPackagingPluginServer) GetAvailablePackageSummaries(ctx context.Cont
...
+	if err != nil {
+		return nil, err
+	}
+	summaries := s.AvailablePackageSummaries[itemOffset:]
+	pageSize := int(request.PaginationOptions.GetPageSize())
+	nextPageToken := ""
+	if pageSize > 0 && pageSize < len(summaries) {
+		summaries = summaries[:pageSize]
+		nextPageToken = fmt.Sprintf("%d", itemOffset+pageSize)
+	}
```
absoludity, line 52: This just updates the mock plugin to be a little more realistic for the `GetAvailablePackageSummaries` call, returning the correct (calculated) summaries based on the pagination options, and a calculated `nextPageToken` (rather than the stock one on the struct, which we can remove once we update the installed package summaries call as well).

cmd/kubeapps-apis/core/packages/v1alpha1/packages.go
```diff
@@ -58,79 +58,44 @@ func (s packagesServer) GetAvailablePackageSummaries(ctx context.Context, reques
...
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "Unable to request results from registered plugins: %v", err)
 	}
 
 	pkgs := []*packages.AvailablePackageSummary{}
 	categories := []string{}
+	var pkgWithOffsets summaryWithOffsets
+	for pkgWithOffsets = range summariesWithOffsets {
+		if pkgWithOffsets.err != nil {
+			return nil, pkgWithOffsets.err
```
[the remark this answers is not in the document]
absoludity: Yeah, let's keep discussing that on #4692

cmd/kubeapps-apis/core/packages/v1alpha1/packages.go
```diff
@@ -58,79 +58,44 @@ func (s packagesServer) GetAvailablePackageSummaries(ctx context.Context, reques
...
-
-			if len(pkgs) == int(pageSize) {
-				nextPageToken = fmt.Sprintf("%d", pageOffset+1)
-			}
-		} else {
-			From(pkgs).
-				// Order by package name, regardless of the plugin
-				OrderBy(func(pkg interface{}) interface{} {
-					return pkg.(*packages.AvailablePackageSummary).Name + pkg.(*packages.AvailablePackageSummary).AvailablePackageRef.Plugin.Name
-				}).ToSlice(&pkgs)
```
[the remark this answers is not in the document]
absoludity: :) Check the code in the fan in (`packages_fan_in.go`). Right here (in `packages.go`) we are assuming that the results are returned in name order (just like they would be from a k8s endpoint). In `packages_fan_in.go` you'll see that it is ensuring this is the case (and testing it). It's not a simple union, it's always grabbing the next minimum result (when comparing the next results of all plugins). I'll point out where...

cmd/kubeapps-apis/core/packages/v1alpha1/packages_fan_in.go
```diff
@@ -0,0 +1,217 @@
...
+// requests.
+func fanInAvailablePackageSummaries(ctx context.Context, pkgPlugins []pkgPluginWithServer, request *packages.GetAvailablePackageSummariesRequest) (<-chan summaryWithOffsets, error) {
+	summariesCh := make(chan summaryWithOffsets)
+
+	corePageSize := int(request.GetPaginationOptions().GetPageSize())
+	// We'll request a bit more than pageSize / n from each plugin. If the page
+	// size is 10 and we have 3 plugins, request 5 items from each to start.
+	pluginPageSize := corePageSize
+	if len(pkgPlugins) > 1 {
+		pluginPageSize = pluginPageSize / (len(pkgPlugins) - 1)
```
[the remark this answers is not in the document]
absoludity: What error propagation? Not sure which error you mean.

If `pluginPageSize` is an `int32` here (rather than `int`), then I need to explicitly cast the denominator as `int32`. Not sure what a ceiling function would achieve here? It's ints all the way down :P 

Looks like maybe you thought it was being preconverted *from a float* to an int to get integer division, rather than converting from `int32` to `int` to be able to use the types together? Not sure.

cmd/kubeapps-apis/core/packages/v1alpha1/packages_fan_in.go
```diff
@@ -0,0 +1,217 @@
...
+					}
+
+					if nextItems[i] != nil && nextItems[i].err != nil {
+						summariesCh <- summaryWithOffsets{
+							err: nextItems[i].err,
+						}
+						close(summariesCh)
+						return
+					}
+				}
```
absoludity, line 107: So this part ensures that `nextItems` is populated for each plugin (that hasn't yet been exhausted)...

cmd/kubeapps-apis/core/packages/v1alpha1/packages_fan_in.go
```diff
@@ -0,0 +1,217 @@
...
+				close(summariesCh)
+				return
+			}
+
+			// Otherwise, we find the minimum item of the next items from each channel.
+			for i, s := range nextItems {
+				if s != nil && s.availablePackageSummary.Name < nextItems[minIndex].availablePackageSummary.Name {
+					minIndex = i
+				}
+			}
```
absoludity, line 131: And this part find which item, out of the next items from each plugin, should be sent down the funnel.

cmd/kubeapps-apis/core/packages/v1alpha1/packages_test.go
```diff
@@ -154,24 +156,23 @@ func TestGetAvailablePackageSummaries(t *testing.T) {
...
 				},
-				PaginationOptions: &corev1.PaginationOptions{PageToken: "0", PageSize: 4},
+				PaginationOptions: &corev1.PaginationOptions{PageToken: "", PageSize: 3},
 			},
 
 			expectedResponse: &corev1.GetAvailablePackageSummariesResponse{
 				AvailablePackageSummaries: []*corev1.AvailablePackageSummary{
 					plugin_test.MakeAvailablePackageSummary("pkg-1", mockedPackagingPlugin1.plugin),
 					plugin_test.MakeAvailablePackageSummary("pkg-1", mockedPackagingPlugin2.plugin),
 					plugin_test.MakeAvailablePackageSummary("pkg-2", mockedPackagingPlugin1.plugin),
```
absoludity, line 166: And this test is ensuring that the actual call returns interleaved, ordered results from the plugins.

absoludity (commented):
Thanks for the review Antonio. Comments inline.

cmd/kubeapps-apis/core/packages/v1alpha1/packages_fan_in.go
antgamdia: > In packages_fan_in.go you'll see that it is ensuring this is the case (and testing it). It's not a simple union, it's always grabbing the next minimum result (when comparing the next results of all plugins). I'll point out where...

AHHH, got it! Thanks for the explanation. I have missed that part; this is awesome! So each call will return an ordered collection of results (by name) as a result of the aggregation of each plugin's response.

Follow-up question: it works because we are assuming each plugin is returning its data with an order, no?  (I mean, if the response wasn't deterministic, this pagination approach wouldn't work at all, I guess)

cmd/kubeapps-apis/core/packages/v1alpha1/packages_fan_in.go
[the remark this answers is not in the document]
antgamdia: Ahhhh, got it. Yep, I thought the division was returning a float instead of just performing the integer division. Thanks for pointing this out. Ignore the comment then :P 
PS: With error propagation I meant [sth like this](https://en.wikipedia.org/wiki/Propagation_of_uncertainty).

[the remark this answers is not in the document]
absoludity: Oh... that kind of error propagation (I'm so used to thinking about propagating errors in go code, in this PR through the channel, that my brain was looking for something along those lines that you might mean here, and I couldn't see an error :) )

cmd/kubeapps-apis/core/packages/v1alpha1/packages_fan_in.go
absoludity: Correct. In particular, we're assuming that like the k8s end points, the plugins return results in `metadata.name` alpha order only. Unfortunately we can't provide any sorting (due to the restrictions on the k8s endpoints themselves) without reading the whole result and sorting or caching that, which I'm not planning to do for now :)

PR #4708, opened 2022-05-14
remove un-needed .../plugin/... from URL paths in flux2.proto

gfichtenholt:
@dlaloue-vmware pointed this inconsistency out. `GetAvailablePackageDetail`() does not use .../plugin/... in the rest API path, while `GetPackageRepositoryDetail`() does. I think this was my mistake caused by copy-and-paste from core repositories.proto.

Separate note:
I am not exactly sure why there's so many generated files updates in this PR. All I did was literally change 3 lines in flux2.proto followed by 'buf generate', 'buf lint', and 'yarn prettier'. In my local environment:
```
[dashboard]$ buf --version
1.4.0
[dashboard]$ yarn --version
1.22.18
```
It looks like @antgamdia in https://github.com/vmware-tanzu/kubeapps/pull/4669 changed a bunch of files and my PR appears to revert some of these changes. If I need to do something on my end, please let me know

Update: manually removed a bunch of files from the PR. Would still like to know what's going on though


antgamdia (approved):
Thanks for the changes! When it comes to committing the autogenerated files, there's always some issue 😅 .
I ran these `buf generate` and `yarn prettier` commands, but I only get these files modified:

````
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
        modified:   cmd/kubeapps-apis/docs/kubeapps-apis.swagger.json
        modified:   cmd/kubeapps-apis/gen/apidocs.pb.go
        modified:   cmd/kubeapps-apis/gen/plugins/fluxv2/packages/v1alpha1/fluxv2.pb.go
        modified:   cmd/kubeapps-apis/gen/plugins/fluxv2/packages/v1alpha1/fluxv2.pb.gw.go
        modified:   cmd/kubeapps-apis/gen/plugins/helm/packages/v1alpha1/helm.pb.gw.go
        modified:   cmd/kubeapps-apis/proto/kubeappsapis/plugins/fluxv2/packages/v1alpha1/fluxv2.proto
````
The PR now looks fine, however, before merging, please try to get rid of the changes on `Submodule [code-generator](https://github.com/kubernetes/code-generator) updated 1 files.`

Have you tried running `buf mod update` and `go mod tidy` to ensure you have locally the latest deps? 
However, I don't really know what's happening. I have the same `buf` version that you also have. Perhaps worths checking if the `protoc-gen-buf-lint` and `protoc-gen-buf-breaking` binaries are also the same as the ones in https://github.com/bufbuild/buf/releases.


Finally, another step I use to do is manually updating the openapi.yaml file. It is a bit tedious, as it involves several manual activities, but we just do it before releasing, usually.

After replicating it, I've pushed my changes to this [upstream branch](https://github.com/vmware-tanzu/kubeapps/tree/4078-fixFluxProto); feel free to cherry-pick any changes.

gfichtenholt:
> The PR now looks fine, however, before merging, please try to get rid of the changes on `Submodule [code-generator](https://github.com/kubernetes/code-generator) updated 1 files.`

Anonio, I need help with this. There is some mis-configuration going on that is preventing me from doing this. Its been going on for a while (several months, this has come up before). Note, that in main, in our [.gitmodules ](https://github.com/vmware-tanzu/kubeapps/blob/main/.gitmodules)file we have this :
[submodule "code-generator"]
	path = cmd/apprepository-controller/vendor/k8s.io/code-generator
	url = https://github.com/kubernetes/code-generator
	branch = release-1.22
	
Notice that release-1.22 corresponds to SHA 9c63990c847dce9e6dca44ad39f7cc4e547bd55f
and the contents of corresponding go.sum have the following:
https://github.com/kubernetes/code-generator/blob/9c63990c847dce9e6dca44ad39f7cc4e547bd55f/examples/go.sum#L443
k8s.io/utils v0.0.0-20211116205334-6203023598ed h1:ck1fRPWPJWsMd8ZRFsWc6mh/zHp5fZ/shhbrgPUxDAE=
which is exactly the changes that I was trying to push.

Now when I look at directory kubeapps/cmd/apprepository-controller/vendor/k8s.io/code-generator
https://github.com/vmware-tanzu/kubeapps/tree/main/cmd/apprepository-controller/vendor/k8s.io
I see that is it set to point to remote branch with SHA b45d7156bcecd2b8c31486673c27e78dd36a8cb6 and the corresponding line is https://github.com/kubernetes/code-generator/blob/b45d7156bcecd2b8c31486673c27e78dd36a8cb6/examples/go.sum#L443 
k8s.io/utils v0.0.0-20210819203725-bdf08cb9a70a h1:8dYfu/Fc9Gz2rNJKB9IQRGgQOh2clmRzNIPPY1xLY5g=

So something is already broken regardless of my commit. 
Please advise


 https://github.com/kubernetes/code-generator/blob/release-1.22/examples/go.sum#L443 I see 
k8s.io/utils v0.0.0-20211116205334-6203023598ed h1:ck1fRPWPJWsMd8ZRFsWc6mh/zHp5fZ/shhbrgPUxDAE= 
which is different to

antgamdia:
> So something is already broken regardless of my commit: our .gitmodules links to a label that doesn't match the contents of the directory on remote branch Please advise

Thanks for the thorough analysis; you are right, there was something wrong in the upstream Kubeapps. So, please merge the PR this that change to get the submodule in sync again.

image

However... I'm starting to wonder... what if we just entirely remove the submodule ? I mean, it seems that using a state-of-the-art AppRepo controller is coming. Why not get rid of this submodule and later decide whether or not we should add it back to our codebase?

gfichtenholt:
> > So something is already broken regardless of my commit: our .gitmodules links to a label that doesn't match the contents of the directory on remote branch Please advise
> 
> 
> However... I'm starting to wonder... what if we just entirely remove the submodule ? I mean, it seems that using a state-of-the-art AppRepo controller is coming. Why not get rid of this submodule and later decide whether or not we should add it back to our codebase?

Separate task. I don't have any objections

Issue #4284, opened 2022-02-15
Unable to add OCI based GCP's Artifact Registry to Kubeapps


jeunii:
### Description:

I am trying to add an OCI Registry to the App Repository. The OCI registry is [GCP's Artifact Registry](https://cloud.google.com/artifact-registry/docs/helm/quickstart)

OCI registry type is [supported](https://cloud.google.com/artifact-registry/docs/supported-formats). 

In GCP I have setup a test repo and can successfully deploy a helm release

```
helm install hello-chart oci://europe-west4-docker.pkg.dev/sap-cx-pipeline-prod/quickstart-helm-repo/hello-chart --version 0.1.0

NAME: hello-chart
LAST DEPLOYED: Tue Feb 15 14:31:46 2022
NAMESPACE: default
STATUS: deployed
...
```

I have followed the [guide](https://github.com/kubeapps/kubeapps/blob/main/docs/user/private-app-repository.md#oci-registry) and my config looks like this

![Screen Shot 2022-02-15 at 4 30 31 PM](https://user-images.githubusercontent.com/46089687/154152648-d4c3d18c-efa0-477e-af91-67db8836b952.png)

I can create the repo in kubeapps successfully and its displayed on the UI

![Screen Shot 2022-02-15 at 4 05 06 PM](https://user-images.githubusercontent.com/46089687/154149020-f6076c89-7d95-4047-8a0e-aec5e433b3f4.png)

But when I click on it I get the error

```
Unable to fetch package: Unable to get the available package versions for the package "gcp-artifactory-repo%2Fquickstart-helm-repo%2Fhello-chart" using the plugin "helm.packages": rpc error: code = InvalidArgument desc = Incorrect request.AvailablePackageRef.Identifier, currently just 'foo/bar' patters are supported: gcp-artifactory-repo%2Fquickstart-helm-repo%2Fhello-chart
```

### Steps to reproduce the issue:

1. Create an helm repo of type docker in GCP's artifact registry
2. Create a Service Account in GCP, give it read rights to artifact registry and generate a token
3. Add the repo in App Repository in Kubeapps 

### Describe the results you received:

I cannot view the helm chart in Catalogue section in Kubeapps. Refreshing the repo creates a cronjob which runs succesfully

```
kl apprepo-apps-sync-sap-cx-pipeline-prod-vwrzr-5bdnx -f

[...]

PR #4716, opened 2022-05-16
Return the repo name in a carvel a pkg summary

### LINKED ISSUE #4543: [carvel] Add pkg "repository" pkg sour source: kappctrl 0.33.1 has annotations for it
Since kapp-controller 0.33.1 (see https://github.com/vmware-tanzu/carvel-kapp-controller/pull/532) there is a new annotation to know which `PackageRepository` a package comes from.

This is useful for setting the field "repository" the same way we were doing for Helm/Flux packages.

Apart from the backend-related changes, this line should be also modified: https://github.com/vmware-tanzu/kubeapps/blob/main/dashboard/src/components/Catalog/PackageCatalogItem.tsx#L30

antgamdia:
### Description of the change

As per https://github.com/vmware-tanzu/kubeapps/issues/4543, it is now possible to populate the chip with the repo name of a carvel package. This PR just replicates the current approach for the Helm/Flux packages, that is, it does not introduces any big change with regards to how repos are referenced.
IMO, it should be carried out in a separate PR and leveraging the new Repo API.

### Benefits

Carvel packages will also have a repo name, like the packages from Helm. Example:

image

### Possible drawbacks

It has been implemented just returning `Name: repo/pkgName`, for instance: `Name: tce-repo/grafana` in the AvailablePackageSummary response, which is not consistent with the rest of the packages.
However, since we are going to refactor it as soon as we start integrating the new Repos APIs, I don't think it is such a big deal, but willing to hear your opinion.

### Applicable issues

- fixes #4543

### Additional information

Still a draft as I haven't written/adapted the tests yet.


absoludity:
> This PR just replicates the current approach for the Helm/Flux packages, that is, it does not introduces any big change with regards to how repos are referenced. IMO, it should be carried out in a separate PR and leveraging the new Repo API.

+1 for postponing a longer term solution using the new Repo API, but why not kill two birds with one stone and use the current approach in the Helm/Flux packages to encode it in the `AvailablePackageReference.identifier`, rather than the `AvailablePackageSummary.Name` field?

It was done that way in the other plugins for two reasons:
1. We needed a way to ensure the data was included in the response to the client (same purpose as here), and importantly,
2. It moves closer towards an `AvailablePackageReference.identifier` that is unique for a plugin (I say moves closer, because with Helm's and Carvel`s support of per-namespace repositories, even `my-repo/apache` may not uniquely identify the available package).

Let me know if there's a reason you chose to encode it into the `AvailablePackageSummary.Name` field instead (also, wouldn't that make the name inconsistent with the same name in the `AvailablePackageDetail.Name`?). Otherwise, I'd be keen that we encode the repo name in the `AvailablePackageReference.identifier`, improving that field too?

> It has been implemented just returning `Name: repo/pkgName`, for instance: `Name: tce-repo/grafana` in the AvailablePackageSummary response, which is not consistent with the rest of the packages.

Just keen to understand why not do it consistently with the other plugins (using the `AvailablePackageReference.Identifier`) - it's just as simple, but also improves that field as above.

Thanks either way!

antgamdia:
> Let me know if there's a reason you chose to encode it into the AvailablePackageSummary.Name field instead
> (...)
>Just keen to understand why not do it consistently with the other plugins (using the AvailablePackageReference.Identifier) - it's just as simple, but also improves that field as above.

In fact, it was my first option, but then I noticed it would change the `id` we are requesting the pkg details with. Honestly, I wasn't aware (meaning I had forgotten it 😅 ) we wanted to move towards an FQDN with _plugin_ and _repo_ as the identifier, but the opposite (I thought we wanted to get rid of the _repo_ for the helm/flux pkgs).

However, I'm 100% for using this FQDN as the id, so I have just pushed those changes (still missing tests, will address them soon, though)

cmd/kubeapps-apis/plugins/pkg/pkgutils/pkgutils.go
```diff
@@ -182,29 +182,29 @@ func AvailablePackageSummaryFromChart(chart *models.Chart, plugin *plugins.Plugi
...
 	if err != nil {
-		return "", status.Errorf(codes.InvalidArgument, "Unable to decode chart ID chart: %v", chartID)
+		return "", status.Errorf(codes.InvalidArgument, "Unable to decode package ID: %v", packageID)
 	}
 	// TODO(agamez): support ID with multiple slashes, eg: aaa/bbb/ccc
-	chartIDParts := strings.Split(unescapedChartID, "/")
-	if len(chartIDParts) != 2 {
-		return "", status.Errorf(codes.InvalidArgument, "Incorrect package ref dentifier, currently just 'foo/bar' patterns are supported: %s", chartID)
+	packageIDParts := strings.Split(unescapedPackageID, "/")
+	if len(packageIDParts) != 2 {
```
absoludity, line 194: FWIW (not for this PR), but this is the code that is very much related to #4284

PR #4719, opened 2022-05-17
Updates core installed pkg summary handler for pagination.

absoludity:

### Description of the change

Follows on from #4694 , applying the same functionality to the core `GetInstalledPackageSummaries` handler.

~~It does not yet work, because the dashboard client is currently assuming knowledge of the pagination token, so I need to do a separate PR for the dashboard to remove that assumption for `GetInstalledPackageSummaries`, as I did earlier for the available packages in the dashboard (hence being a draft).~~ EDIT: it ended up only requiring a 2-line change to the dashboard so I've just included it in this PR.

Works to display correlated installed package summaries:
installed-apps

### Benefits

Core pagination for multiple plugins will display installed packages in order efficiently.

### Applicable issues

- ref #3399 

### Additional information

I did play a little with generics at first but couldn't yet see a useful way to re-use the functionality due to (a) small differences between the responses - available package summaries including categories, for example, and (b) the inability to define a shared interface because go doesn't let you add functions to a struct from another package (and I'm not sure whether we want to add them to the generated packages). Either way, I've left a TODO for us to do this in the future if we want.

### DIFF
--- a/cmd/kubeapps-apis/core/packages/v1alpha1/packages.go
+++ b/cmd/kubeapps-apis/core/packages/v1alpha1/packages.go
@@ -67,7 +67,7 @@ func (s packagesServer) GetAvailablePackageSummaries(ctx context.Context, reques
 
 	pkgs := []*packages.AvailablePackageSummary{}
 	categories := []string{}
-	var pkgWithOffsets summaryWithOffsets
+	var pkgWithOffsets availableSummaryWithOffsets
 	for pkgWithOffsets = range summariesWithOffsets {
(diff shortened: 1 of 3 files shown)


antgamdia (approved):
Awesome! That's great we finally have pagination in both get X summaries requests! Thanks for the good work here!

I've noticed the watermark behind some packages is the default one (kubeapps mark), instead of using the plugin.  Perhaps it gets fixed with https://github.com/vmware-tanzu/kubeapps/pull/4728:

image

Issue #3953, opened 2021-12-14
New Kubeapps web


ppbaena:

antgamdia:
Some actions I quickly foresee:

- [ ] Investigate which is the proper corporate style to follow.
- [ ] Build an initial website with the existing information in kubeapps.com
- [ ] Move the existing docs to the proper place (perhaps the `content` folder in Hugo?)
- [ ] Configure Netlify to build and deploy the website
- [ ] Set the appropriate DNS records
- [ ] Think about staging environments? 
- [ ] Document how to update the website and run the companion tools (style check, link check, etc...)

ppbaena:
Retake the revamping of Kubeapps website:

- Last template for website identified and available.
- Logo revamp feedback provided to the designer

antgamdia:
A brand new website has been proposed at https://github.com/vmware-tanzu/kubeapps/pull/4705. I've searched for any broken links, but, fortunately, it seems we have successfully gotten rid of them!
Once merged, #4715 should be also merged.

When everything gets merged, there are still some pending items:

- [ ] Change the default website's branch to `main` in Netlify.
- [ ] Add the proper DNS records to the Kubeapps.[com,io] zones.

ppbaena:
@antgamdia I'm asking for an ID for the corporate cookies manager and checking GA configuration. Keep you posted when available.

ppbaena:
New Google Tag Manager container for kubeapps.dev created. 
Pending to submit a PR to add these changes once we have the new OneTrust ID (already requested).

ppbaena:
All tests for Cookie compliance have been launched and the site has passed all of them.

ppbaena:
Google Analytics configured and linked to GTM:

- Universal GA tag for old data from kubeapps.com
- GA4 ID for new website at kubeapps.dev

ppbaena:
Requested to add TXT in DNS to verify ownership with Search Console. Once we finish this process we can close this task.

antgamdia:
As stated in an internal email, we have to ask for an "accessibility audit", which can take up to 5 weeks to complete, and then we will have to fix any issues detected by the team. Nevertheless, we already conducted a quick audit (see #4844 ), so I don't foresee plenty of issues.

ppbaena:
Accessibility audit was requested last Monday, so let's wait for the results and according to the previous work performed we should not have many issues.

ppbaena:
Search console identity verified.

[...]

PR #4720, opened 2022-05-17
Replace `’` with `'` and run `buf`

antgamdia:
### Description of the change

After https://github.com/vmware-tanzu/kubeapps/pull/4708 and #3953 I noticed we left unchanged some characters that were replaced when making the docs more uniform. This PR is just to fix it and re-run the `buf generate` command again.
There are some minor changes since the protoc generator has been updated (the `err` is initialized like `var err` instead of using  `:=`)
Also, I've updated the openapi.yaml file with the new links to the docs files.

### Benefits

Up to date generate code and API docs.

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

N/A


dashboard/public/openapi.yaml
```diff
@@ -3170,7 +3307,50 @@ paths:
             type: string
             pattern: .+
       requestBody:
-        $ref: "#/components/requestBodies/v1alpha1UpdatePackageRepositoryRequest"
```
antgamdia, line 3173: Don't know why this is being removed, but after the change in #4708, it isn't generating the associated requestBody separately :S

absoludity (approved):

dashboard/public/openapi.yaml
absoludity: Hmm, maybe the length of the URL being shorter? No idea.

Issue #4662, opened 2022-05-09
Implement Package Repository API for Helm plugin


castelblanque:
**Summary**
Implementation of the new package repository API for Helm plugin

**Description**
Greg designed the API definition, and implemented it for Flux plugin in #3496.
Similar work regarding implementation needs to be done for Helm plugin.

Repositories API definition can be found [here](https://github.com/vmware-tanzu/kubeapps/blob/main/cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/repositories.proto).

There is the current source code in charge of managing the direct Helm repos that could be reused partially.

**Acceptance criteria**
Following methods must be implemented in the Helm plugin:

- [ ] AddPackageRepository
- [ ] GetPackageRepositoryDetail
- [ ] GetPackageRepositorySummaries
- [ ] UpdatePackageRepository
- [ ] DeletePackageRepository

absoludity:
If possible, we'd want to be verifying the API satisfies the needs of the UX and finding the gaps as we go. Whether that means updating the UX to use the new API as it's implemented, not sure.

PR #4723, opened 2022-05-17
First step of AddPackageRepository - Package Repository API for Helm

castelblanque:
### Description of the change

Adds a basic implementation of `AddPackageRepository` method from package repository API for the Helm plugin (#3496). With basic I mean:

- No Docker credentials supported yet
- No TLS config with secret ref (it is, though, the TLS CA)
- No custom fields for Helm repos (e.g. template for cronjobs)

Package repositories for Helm plugin can be created now with the common repository API definition, e.g.

```bash
{
  "context": { "namespace": "example-ns" },
  "name": "example-repo",
  "type": "helm",
  "namespace_scoped": true,
}
```

The above example will create a repository in the namespace `example-ns`.

### Benefits

Repositories can be created meeting the new Package Repository API definition.

### Possible drawbacks

None that I know. Outcome of this implementation should be the same `AppRepository` CRD created by the already existing API.
Don't see a problem with both APIs coexisting.

### Applicable issues

Partially implements #4662.

### Additional information

Following the example above, the one below will have the repo created in the global namespace, even if there is a namespace specified:

```bash
{
  "context": { "namespace": "example-ns" },
  "name": "example-repo",
  "type": "helm",
  "namespace_scoped": false,
}
```

That does not seem very consistent to me, should we return failure in this case?


absoludity:
Hey @castelblanque . Did you check with @dlaloue-vmware about reviewing each other work? I can jump in and review if needed, but I think since you're both working on similar aspects for different plugins, it'd be a great chance to review each others and learn together?

castelblanque:
You are totally right @absoludity. Will do!

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -1024,3 +996,75 @@ func (s *Server) GetInstalledPackageResourceRefs(ctx context.Context, request *c
 		}, nil
 	}
 }
+
+func (s *Server) AddPackageRepository(ctx context.Context, request *corev1.AddPackageRepositoryRequest) (*corev1.AddPackageRepositoryResponse, error) {
+	log.Infof("+helm AddPackageRepository [%v]", request)
```
dlaloue-vmware, line 1001: logging the whole request may leak credentials. it would be better to abstract only what is necessary (see how it is done in the method CreateInstalledPackage)

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -1024,3 +996,75 @@ func (s *Server) GetInstalledPackageResourceRefs(ctx context.Context, request *c
...
+			codes.Unimplemented,
+			"not supported yet: request.Context.Cluster: [%v]",
+			request.Context.Cluster)
+	}
+
+	if request.Name == "" {
+		return nil, status.Errorf(codes.InvalidArgument, "no package repository Name provided")
+	}
+	name := types.NamespacedName{Name: request.Name}
+	if request.GetNamespaceScoped() {
```
dlaloue-vmware, line 1020: i think we have agreed to validate that the namespace and scope were consistent (e.g. if scope is global but namespace is not the global namespace, then it should be a validation error)

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -1024,3 +996,75 @@ func (s *Server) GetInstalledPackageResourceRefs(ctx context.Context, request *c
 		}, nil
 	}
 }
+
+func (s *Server) AddPackageRepository(ctx context.Context, request *corev1.AddPackageRepositoryRequest) (*corev1.AddPackageRepositoryResponse, error) {
+	log.Infof("+helm AddPackageRepository [%v]", request)
+	if request == nil {
+		return nil, status.Errorf(codes.InvalidArgument, "no request provided")
+	}
+	if request.Context == nil || request.Context.Namespace == "" {
```
dlaloue-vmware, line 1005: this is just a suggestion, but if the namespace is not provided, we could default it to the global namespace.
this makes it easier for users as the global namespace is not required to be known

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/repositories.go
```diff
@@ -0,0 +1,146 @@
...
+				Namespace: helmRepoCrd.Namespace,
+				Cluster:   s.kubeappsCluster,
+			},
+			Identifier: helmRepoCrd.Name,
+			Plugin:     GetPluginDetail(),
+		}, nil
+	}
+}
+
+func getValidRepoTypes() []string {
```
dlaloue-vmware, line 80: could that be a constant?

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
castelblanque: It is a line copied from Flux plugin [here](https://github.com/vmware-tanzu/kubeapps/blob/4662-addpackagerepoHelm-2/cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/server.go#L502).
If it is just to locate the operation in the logs, I would say that name and URL should suffice.
@gfichtenholt what do you think?

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/repositories.go
castelblanque: It seems that arrays in Go can't be constants. [See here.](https://stackoverflow.com/a/13137568)

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
castelblanque: Yep, it makes sense. Will add it.

castelblanque:
@dlaloue-vmware Thanks for the review!

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/repositories.go
dlaloue-vmware: good to know 👍

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
gfichtenholt: I needed more than name and URL when I was working on flux as some point. I debugged and found and fixed the problem, whatever it was, then and this just kind of remained. I think @dlaloue-vmware is right, could you fix it and also fix one in flux as well if you have the bandwidth. Thanks in advance

cmd/kubeapps-apis/docs/kubeapps-apis.swagger.json
```diff
@@ -29,6 +29,9 @@
     {
       "name": "HelmPackagesService"
     },
+    {
+      "name": "HelmRepositoriesService"
+    },
```
antgamdia, line 34: No action required in this PR, but just mentioning it for all us to be aware: the `openapi.yaml` file (which is the doc used for the interactive REST API portal) will get out of sync.

Also, but unrelated: I don't know why my generator is different (meaning worse) than yours haha.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/repositories_auth.go
```diff
@@ -0,0 +1,244 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: secretNameForRepo(ownerRepo.Name),
+		},
+		Type: k8scorev1.SecretTypeOpaque,
+		Data: map[string][]byte{},
+	}
+}
+
+// this func is only used with kubeapps-managed secrets
+func newSecretFromTlsConfigAndAuth(repoName types.NamespacedName,
```
antgamdia, line 41: Didn't we have a shared `/pkg/` folder share code across plugins? 
I'm not sure, but I'd rather remove any dup code if possible, no?
_(mentioning this bc I saw a similar fn in the flux plg)_

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/globals_vars_test.go
```diff
@@ -0,0 +1,381 @@
...
+package main
+
+import (
+	"github.com/vmware-tanzu/kubeapps/cmd/apprepository-controller/pkg/apis/apprepository/v1alpha1"
+	corev1 "github.com/vmware-tanzu/kubeapps/cmd/kubeapps-apis/gen/core/packages/v1alpha1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
```
antgamdia, line 13: Correct me if I'm wrong here, @absoludity , but I'd rather say we weren't so fans of globals in the tests, preferring instead each test case being defining (and customizing with a given set of test-case properties) the object.

So, what about replacing those globals with just functions creating each struct based upon a set of params (where you foresee the variability, I mean).

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/repositories_test.go
```diff
@@ -0,0 +1,346 @@
...
+		},
+		{
+			name:       "returns error if no name is provided",
+			request:    &corev1.AddPackageRepositoryRequest{Context: &corev1.Context{Namespace: "foo"}},
+			statusCode: codes.InvalidArgument,
+		},
+		{
+			name:       "returns error if invalid cluster",
+			request:    &corev1.AddPackageRepositoryRequest{Context: &corev1.Context{Cluster: "wrongCluster", Namespace: "foo"}},
+			statusCode: codes.Unimplemented,
```
antgamdia, line 52: Why unimplemented if invalid cluster?

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/testdata/cert/ca-csr.json
```diff
@@ -0,0 +1,8 @@
+{
+  "CN": "fluxv2plugin-testdata-ssl-svc.default.svc.cluster.local CA",
```
antgamdia, line 2: nitpick  (same in the rest of ocurrences)

```suggestion
  "CN": "helmplugin-testdata-ssl-svc.default.svc.cluster.local CA",
```

antgamdia (approved):
Thanks for the changes!  LGTM 
I wanted to get into the UI revamp to refresh my knowledge regarding repos before looking at the PR.

Minor side coment: 

- Have a look at https://github.com/vmware-tanzu/kubeapps/pull/4708#issuecomment-1127229140 where Greg commented about the file: `cmd/apprepository-controller/vendor/k8s.io/code-generator`

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/repositories_auth.go
castelblanque: Some code here is copied from Greg's Flux.
Most of the times due to Flux-specific structs it isn't possible to make it generic. I tried at the beginning but had to undo down the road.
We could do some refactoring in a later iteration if needed.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/globals_vars_test.go
castelblanque: Ok, I didn't know about that decision. I've used functions to generate structs at some points, but didn't want to convert tests in another set of complex code 😄

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/repositories_test.go
castelblanque: Mmm yes, probably InvalidArgument makes more sense. Thanks!

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/testdata/cert/ca-csr.json
castelblanque: Will change it in the following PR that is based in this branch, if you don't mind.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/globals_vars_test.go
absoludity: Not so much a decision as a recommendation - global vars create shared state that is then used by different tests. Can lead to one test updating that shared state causing hard to debug failures in other tests etc. See https://dave.cheney.net/2017/06/11/go-without-package-scoped-variables for a good discussion.

PR #4728, opened 2022-05-18
Ensure Helm plugin includes plugin detail in installed ref.

absoludity:

### Description of the change

Small prequel to #4719 which is failing CI with:

test-failed-1

The reason is that, in that PR, when updating the `GetInstalledPackageSummaries` core handler, I did not replace the code which explicitly set the plugin detail for each result if it wasn't set by the plugin, because I thought surely the plugins do this.

Turns out the helm plugin wasn't while the flux and carvel plugins are. So made sense to update the Helm plugiin to be the consistent.

### Benefits

CI will pass for #4719, core plugin doesn't need to add data and is simpler.

### Applicable issues

- ref #3399


antgamdia (approved):
Nice! Great to see our CI, apart from the headaches it gives us, is catching some bugs :P haha

PR #4745, opened 2022-05-19
set up AppRepository type

gfichtenholt:
You will notice a number of .go files with Flux copyright in them, such as everything in internal/ folder so far. Those files I copied AS IS from https://github.com/fluxcd/source-controller/, and made sure to keep the original copyright to adhere to to terms of the Apache 2.0 License agreement. I will separately check with VMware open source compliance office to make sure everything I am doing is "kosher" 

I expect more files (in the end total < 50) to copy from flux source controller project in order to save me weeks, if not months of work


cmd/apprepository-controller-new/main.go
```diff
@@ -36,10 +48,16 @@ func init() {
...
 
 func main() {
-	var metricsAddr string
-	var enableLeaderElection bool
-	var probeAddr string
-	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	var (
+		metricsAddr          string
+		eventsAddr           string
+		enableLeaderElection bool
```
antgamdia, line 54: Just out of curiosity: is this for enabling this controller to be scaled up/down ? If so, that's great. Such an improvement from our first CRD controller 😅

PR #4753, opened 2022-05-20
Truncate cronjobs' names

### LINKED ISSUE #4061: Combination of repository name and namespace greater than 52 characters makes Cronjobs silently fail
### Description:

When creating a repository which combination of name and namespace is > 52 characters, sync cronjob fails to be created.
Apprepository controller fails to create the cronjob due to K8s limitation to names of 52 chars, but error is not shown in Dashboard, only logs similar to:

```
E0111 14:41:46.734567       1 controller.go:253] error syncing 'test-long-namespace-name/my-long-named-repository-for-test': CronJob.batch "apprepo-test-long-namespace-name-sync-my-long-named-repository-for-test" is invalid: metadata.name: Invalid value: "apprepo-test-long-namespace-name-sync-my-long-named-repository-for-test": must be no more than 52 characters
```

User sees repo successfully created in Dashboard, but packages are never synced.

Problem can only be worked around by removing the repo and creating it again with a shorter name or in a shorter named namespace.

1. Log into Kubeapps
2. Create a new namespace with a long name (e.g. `test-long-namespace-name`)
3. Add a new repo with a long name (e.g. `my-long-named-repository-for-test`)
4. Check cronjobs created in K8s with `kubectl -n kubeapps get cronjobs`

Sync cronjob is never created. Packages from repo are never shown in catalog.

Sync cronjob being created, or at least a message to the user preventing the creation of the repo.

Different options:

1. Limit the length of repository names. Given that K8s namespaces can be up to 63 characters, even with this limitation we could have the issue.
2. Limit the length of both namespaces and repositories names. This at least would reduce the probability of the issue happening.
3. Truncate the combined string of namespace+repo name. Due to the mentioned rule of 63 chars for ns, not adviced. 

From [Kubernetes documentation](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/):
> The name must be no longer than 52 characters.

antgamdia:
### Description of the change

This PR addresses #4061 by truncating (+hashing) a cronjob name up to its 52-chars limit. The rationale has been:

Our cronjob's names have: 1) a fixed part `apprepo--cleanup--` and `apprepo--sync-`; 2) a namespace name; 3) a job name.
Whereas we don't want to change anything from 1), 2) and 3) can, on their own, have up to 63 characters. So, in theory, until now, we can exceed the 126 characters... far distant from the 52-chars limit.

So, in this PR I'm truncating both 2) and 3), but, in order to avoid name collisions, a hashing function is applied if possible. After having a look at some hashing functions producing short checksums, two main options were `fletcher` and `adler` in their 32-bit version. Having compared the chances of a collision [1] and given they both produces similar results, I went with `adler-32`, which is integrated in the go standard libs.

The hashes are up to 32 bits ints (unsigned), that is 2^32: the max number has up to 10 digits.

Considering all of this, each cronjob's name and namespace get truncated if they exceed the MAX_LENGTH/2 (that is 52/2 chars) and hashed so that we can still avoid (best effort, not guaranteed) collisions.

[1] https://www.zlib.net/maxino06_fletcher-adler.pdf

### Benefits

No more jobs silently failing anymore!

### Possible drawbacks

N/A

### Applicable issues

- fixes #4061 

### Additional information

Currently, we are giving the same amount of characters to both the namespace and job name. We could change it or use another approach to distribute the available remaining characters, but, I don't really think it is worthwhile.

IRL test:

```console
k get cronjob -n kubeapps 
NAME                                       SCHEDULE       SUSPEND   ACTIVE   LAST SCHEDULE   AGE
apprepo-default-sync-bitnami               */10 * * * *   False     1        6s              5m8s
apprepo-default-sync-thisisab-2213025484-27556480   0/1           11s        11s
apprepo-default-sync-thisisab-2213025484-29cbg      0/1           57s        57s

k get cronjob -n kubeapps 
NAME                                       SCHEDULE       SUSPEND   ACTIVE   LAST SCHEDULE   AGE
apprepo-default-sync-bitnami */10 * * * *   False     1        28s             5m30s
apprepo-default-sync-thisisab-2213025484   */10 * * * *   False     1        28s             73s
````

**Edit**: marked as a draft, as CI failed and I still have to debug what's the reason.

### DIFF
--- a/cmd/apprepository-controller/server/controller.go
+++ b/cmd/apprepository-controller/server/controller.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"hash/adler32"
 	"strconv"
 	"strings"
 	"time"
@@ -37,6 +38,11 @@ import (
 )
 
 const controllerAgentName = "apprepository-controller"
+
+// Although a k8s typical length is 63, some characters are appended from the cronjob
+// to its spawned jobs therefore restricting this limit up to 52
+// https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/
+const MAX_CRONJOB_CHARS = 52
 
 const (
 	// SuccessSynced is used as part of the Event 'reason' when an AppRepository
@@ -271,7 +277,7 @@ func (c *Controller) syncHandler(key string) error {
 
 			// TODO: Workaround until the sync jobs are moved to the repoNamespace (#1647)
 			// Delete the cronjob in the Kubeapps namespace to avoid re-syncing the repository
-			err = c.kubeclientset.BatchV1beta1().CronJobs(c.conf.KubeappsNamespace).Delete(context.TODO(), cronJobName(namespace, name), metav1.DeleteOptions{})
+			err = c.kubeclientset.BatchV1beta1().CronJobs(c.conf.KubeappsNamespace).Delete(context.TODO(), cronJobName(namespace, name, false), metav1.DeleteOptions{})
 			if err != nil && !errors.IsNotFound(err) {
 				log.Errorf("Unable to delete sync cronjob: %v", err)
 				return err
@@ -282,7 +288,7 @@ func (c *Controller) syncHandler(key string) error {
 	}
 
 	// Get the cronjob with the same name as AppRepository
-	cronjobName := cronJobName(namespace, name)
+	cronjobName := cronJobName(namespace, name, false)
 	cronjob, err := c.cronjobsLister.CronJobs(c.conf.KubeappsNamespace).Get(cronjobName)
 	// If the resource doesn't exist, we'll create it
 	if errors.IsNotFound(err) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: cmd/apprepository-controller/server/controller.go [hunk 1]
```diff
@@ -580,13 +586,49 @@ func ttlLifetimeJobs(config Config) *int32 {
...
+	// the "apprepo--cleanup--" string has 18 chars, which leaves us 52-18=34 chars for the final name
+	maxNamesapceLength, rem := (MAX_CRONJOB_CHARS-18)/2, (MAX_CRONJOB_CHARS-18)%2
+	maxNameLength := maxNamesapceLength
+	if rem > 0 {
+		maxNameLength++
+	}
+
+	truncatedName := fmt.Sprintf("apprepo-%s-cleanup-%s-", truncateAndHashString(namespace, maxNamesapceLength), truncateAndHashString(name, maxNameLength))
+	return truncatedName
+}
```
absoludity, line 614: Looks like these two functions could easily be a single one? `cronJobName(namepsace, name, action string, addDash bool)` ? (where `action` is passed in as "cleanup" or "sync"). You could still have the separate `deleteJobName` which would just call with the correct action? Not sure, maybe not worth it.

absoludity (approved):
Great, looks good.

PR #4754, opened 2022-05-20
Add subtitle for showing the available pkg in the app view

### LINKED ISSUE #4448: [carvel] Package name or description not displayed on install details page
### Description:

While writing a CI test, I noticed that when viewing the package details, we're not currently displaying the name of the (available) package anywhere, only the name given to the installed package (eg. test-carvel here):

gatekeeper-install-after-refresh

I don't see anything there identifying that `test-carvel` is an installation of the gatekeeper package. 

See screenshot

I expected to see the package name "Gatekeeper" as well as its tagline "policy management".

antgamdia:
### Description of the change

This PR just adds a `subtitle` in the current `PageHeader` so that the `AppView` can set a value. 
I have used the following structure: `from package X` (where X is the ~identifier~ display name of the available package). Open to change it if you come up with a better copy.

### Benefits

CI (and users) will know which available package an installed pkg belongs to.

### Possible drawbacks

N/A

### Applicable issues

- fixes #4448

### Additional information

image


dashboard/src/components/AppView/AppView.tsx
```diff
@@ -292,6 +294,20 @@ export default function AppView() {
...
             title={releaseName}
             titleSize="md"
+            subtitle={
+              appDetails?.availablePackageRef ? (
+                <span>
+                  from package{" "}
+                  <Link
+                    to={url.app.packages.get(cluster, namespace, appDetails.availablePackageRef)}
+                  >
+                    {appDetails.availablePackageRef.identifier}
```
absoludity, line 304: Wondering why we're using the identifier here. I mean, right now it looks useful for the plugin shown, but the identifier used by a plugin may not be useful to users at all. We're getting the identifier here through the `appDetails`, which is an `AvailablePackageDetail`, so why not use `appDetails.displayName` ?

Maybe because that doesn't show the repository? I don't think that's a reason not to use it (and users can see that by clicking your new link if they want too), but not sure.

Anyway, I'll +1 but keen to understand why we're not using the display name of the available package ref, since that's what it's for.

absoludity (approved):
Nice helpful change Antonio! One question about what we actually display, but +1 either way.

dashboard/src/components/AppView/AppView.tsx
antgamdia: > Maybe because that doesn't show the repository? I don't think that's a reason not to use it (and users can see that by clicking your new link if they want too), but not sure.

Yep, that's precisely the reason why I used the identifier. Perhaps it is just me being so biased to seeing the `repo/foo` pattern, but perhaps: 1) the repo is not that useful for users; 2) if it happens to be, we can always added it to the display name.

So, 100% agree with the suggestion. Let's use the `displayName` to actually display the name of the package 😅

antgamdia: image

absoludity: Yep, exactly (RE we can update the display name if we want). Thanks Antonio.

PR #4756, opened 2022-05-20
Update issue templates

ppbaena:
Remove feature request template to keep feature proposal template.


ppbaena:
I've noticed that every time I suggest a proposal to change issue templates directly from github, it changes " for ' '.

antgamdia (approved):

PR #4767, opened 2022-05-23
Allow new Kapp >= 0.47 ConfigMap suffix

### LINKED ISSUE #4672: [carvel] Address upcoming change in the `-ctrl` ConfigMap suffix (kapp >=0.47)
We were told (thanks Scott!!) that Kapp is going to change the naming convention of the ConfigMap from which we fetch the generated k8s resources by an App.

We have to:

- Analyze the implications of the change in ~https://github.com/vmware-tanzu/carvel-kapp/issues/146~ https://github.com/vmware-tanzu/carvel-kapp-controller/issues/376
    - Slack thread discussing this change: https://kubernetes.slack.com/archives/CH8KCCKA5/p1652136784883199 
- Come up with a solution ensuring backward compatibility

antgamdia:
### Description of the change

As pointed out by some community (thanks Scott) and Carvel people (thanks Neil, Soumik and Praveen) the `App` (and therefore its associated ConfigMap, which is what we are interested in) [suffix has changed](https://github.com/vmware-tanzu/kubeapps/issues/4672#issuecomment-1122628302).

This PR is to support both suffixes (`-ctrl` and `.apps.k14s.io`) when looking up the ConfigMap, trying first the new one and falling back to the old one if the first attepts does not succeed.

### Benefits

Kubeapps will work again with apps generated with Kapp >= 0.47

### Possible drawbacks

N/A (as it will be still continue to work on pre 0.47 kapp scenarios)
### Applicable issues

- fixes #4672

### Additional information

> **Note**
> _(testing this new note/warning gfm feature, hehe)_
> I've added the changes on top of the dependabot's PR updating Kapp as we are using the suffix constant from their code.

> **Warning**
> IRL testing still pending (will do near soon), but I don't foresee any surprise, I rely on our unit tests :P


cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_resources.go
```diff
@@ -444,12 +443,38 @@ func (s *Server) inspectKappK8sResources(ctx context.Context, cluster, namespace
 		// TODO(minelson): We can't currently use `errors.IsNotFound(err)` here.
 		// See https://github.com/vmware-tanzu/carvel-kapp/issues/498
 		// Instead we need to match on the error string :/
-		if strings.Contains(err.Error(), fmt.Sprintf("configmaps %q not found", appName)) {
+		cmErrPattern := "configmaps %q not found"
+		appErrPattern := "App '%s' (namespace: %s) does not exist"
+		if strings.Contains(err.Error(), fmt.Sprintf(cmErrPattern, appName)) || strings.Contains(err.Error(), fmt.Sprintf(appErrPattern, appName, namespace)) || strings.Contains(err.Error(), fmt.Sprintf(appErrPattern, packageId, namespace)) {
```
antgamdia, line 448: The messages differ from pre and post kapp 0.47:

- pre 0.47: just `configmaps...`
- post 0.47: `"App ...` using `packageId` for detecting the fallback mode and  `"App ...` using `appName` when the CM is certainly not found

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_test.go
```diff
@@ -6067,7 +6153,7 @@ func TestGetInstalledPackageResourceRefs(t *testing.T) {
 					},
 					ObjectMeta: metav1.ObjectMeta{
 						Namespace: "default",
-						Name:      "my-installation-ctrl",
+						Name:      "my-installation.apps.k14s.io",
```
antgamdia, line 6156: This is the actual change in Kapp 0.47. However, it is still an opt-in feature, so even if using kapp 0.47, users will still have the `-ctrl` thing.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_resources.go
```diff
@@ -444,12 +443,38 @@ func (s *Server) inspectKappK8sResources(ctx context.Context, cluster, namespace
...
+	refs := []*corev1.ResourceRef{}
+
+	// Get the Kapp different clients
+	appsClient, resourcesClient, failingAPIServicesPolicy, _, err := s.GetKappClients(ctx, cluster, namespace)
+	if err != nil {
+		return nil, err
+	}
+
+	// Get the App and its used GVs considering the pre/post Kapp 0.47 suffixes
+	usedGVs, app, err := getAppUsedGVs(appsClient, packageId, namespace, false)
```
absoludity, line 476: Not sure if I've missed something, but it seems a little confusing to me to call this function with the arg `useDeprecatedCtrlAppSuffix` set to false, and yet it is also checking the deprecated (pre 0.47) suffix?

Reading the code, it looks like you've added this arg so that you can call the function recursively (with `true`) when it fails initially? Why not instead just have the `getAppUsedGVs()` function re-call `appsClient.Find(appName)` with the deprecated suffix when it detects the NotFound, rather than calling the function recursively?

absoludity (approved):

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_resources.go
antgamdia: Sorry if it seems confusing, I've just tried not to repeat code as much as possible. Let me explain the rationale, but happy to change if still confusing:

Quick note: even if `appsClient.Find(appName)` would intuitively fail if the app is not found... it will not. In fact, it always succeed.  Why? Because, besides its name, `Find` is not finding but creating a new struct `return NewRecordedApp(name....` 

Therefore, we can't rely on the output of this `Find` but in a subsequent call to `app.XXXX`. For instance  `app.find(name)`  or `app.UsedGVs()`.

So, it turns out that we would require 1) to create the RecordedApp and 2) to execute an operation that actually interacts with the k8s client.

In that scenario, we could do 1) and 2) and fallback accordingly and start over 1) and 2) again. Or, what I did: use a flag to choose between scenarios invoking recursiverly the function (but just once, when it actually falls back to the legacy behavior).

Note that the recursive case is only exceuted when called with `useDeprecatedCtrlAppSuffix = false`:

````go
if !useDeprecatedCtrlAppSuffix {
    return getAppUsedGVs(appsClient, packageId, namespace, true)
}
````

This way we get rid of any in-line fallback mechanism.
I mean, if only `.Find(...)` returned a nil when looking up the application, we would be able to just change the name at the very beginning.
However, happy to move to another alternative if you suggest so!

absoludity: > Sorry if it seems confusing, I've just tried not to repeat code as much as possible. Let me explain the rationale, but happy to change if still confusing:
> 
> Quick note: even if `appsClient.Find(appName)` would intuitively fail if the app is not found... it will not. In fact, it always succeed. Why? Because, besides its name, `Find` is not finding but creating a new struct `return NewRecordedApp(name....`
> 
> Therefore, we can't rely on the output of this `Find` but in a subsequent call to `app.XXXX`. For instance `app.find(name)` or `app.UsedGVs()`.
> 
> So, it turns out that we would require 1) to create the RecordedApp and 2) to execute an operation that actually interacts with the k8s client.
> 
> In that scenario, we could do 1) and 2) and fallback accordingly and start over 1) and 2) again. Or, what I did: use a flag to choose between scenarios invoking recursiverly the function (but just once, when it actually falls back to the legacy behavior).

Yep, I don't have an issue with using recursion to do the fallback (though I do wonder if (1) and (2) could be pulled out to a function so that it can be done without recursion - but not important). The bit that reads as confusing to me is that:

> 
> Note that the recursive case is only exceuted when called with `useDeprecatedCtrlAppSuffix = false`:
> 
> ```go
> if !useDeprecatedCtrlAppSuffix {
>     return getAppUsedGVs(appsClient, packageId, namespace, true)
> }
> ```

worded another way, it seems that if you call with `useDeprecatedCtrlAppSuffix = true`, it *won't* fall back to try using the deprecated suffix, whereas as you've said, if you set `useDeprecatedCtrlAppSuffix = false`, it *will* fall back to try using the deprecated suffix. That has my brain confused :P

At least for me, if it was instead called `fallbackToDeprecatedCtrlAppSuffix` and reverse the condition on the recurse, it would make more sense as a call-site. See what you think, but fine either way. Still +1 :)

antgamdia: > At least for me, if it was instead called fallbackToDeprecatedCtrlAppSuffix and reverse the condition on the recurse, it would make more sense as a call-site. See what you think, but fine either way. Still +1 :)

I'm not sure what you really meant here. I'd say `useDeprecatedCtrlAppSuffix` and `fallbackToDeprecatedCtrlAppSuffix` are mostly semantically equivalent, so reversing the condition is something I still don't see.

I have pushed a change replacing `useDeprecatedCtrlAppSuffix` with `useNewCtrlAppSuffix` so that we can actually reverse the conditon + clarifying the comments.

absoludity: Heh, let's chat during the standup and I'll explain what I mean... not sure if I'm misunderstanding the code we're not understanding each other. It's not an important point, I've already +1'd, but I'm keen to understand where we're missing each other here.

BTW, looks like you might have updated your branch with new changes that aren't in the target PR.

PR #4772, opened 2022-05-24
Update the token-auth example for getting-started.

### LINKED ISSUE #4763: Verify kubeapps token instructions after SA change in k8s >=1.24
Converted into an issue since it requires further investigation.
Updating the issue content accordingly:

----

Hi,

In Kubernetes 1.24 tokens are not created for service accounts. So the authorization procedure in the documentation no longer works:

```
To create a Service Account for a user "example" in the "default" namespace, run the following:
kubectl create -n default serviceaccount example
To get the API token for this Service Account, run the following:
kubectl get -n default secret $(kubectl get -n default serviceaccount example -o jsonpath='{.secrets[].name}') -o go-template='{{.data.token | base64decode}}' && echo
```

So I created service account "example" like above, and then created a token with  
`kubectl create -f token.yml`

token.yml:
```
apiVersion: v1
kind: Secret`
metadata:`
   namespace: default` 
   name: example-secret`
   annotations:`
       kubernetes.io/service-account.name: "example"`
type: kubernetes.io/service-account-token`
```
I then apply this yaml:

```
kind: ClusterRoleBinding
metadata:
  name: admin-user-binding
roleRef:
  kind: ClusterRole
  name: cluster-admin
subjects:
- kind: ServiceAccount
  name: example
  namespace: default
```

I can now see the token with
`kubectl describe secret example-secret`

However, this token does not get me into the kubeapps dashboard.  It does get me into the standard kubernetes dasboard

Any ideas?

_Originally posted by @Wouter053 in https://github.com/vmware-tanzu/kubeapps/discussions/4746_

absoludity:

### Description of the change

With 1.24, token secrets are not created automatically for a service account. Updated our getting started to create the secret with the required annotation to be populated with a valid token, and update the instructions to just get that token (more simple anyway).

Verified this works both with k8s 1.24 and 1.21 .

### Benefits

Avoid confusion with token auth moving forward as more people use 1.24

### Applicable issues

- fixes #4763

### DIFF
--- a/script/makefiles/cluster-kind.mk
+++ b/script/makefiles/cluster-kind.mk
@@ -6,7 +6,7 @@
 KUBE ?= ${HOME}/.kube
 CLUSTER_NAME ?= kubeapps
 ADDITIONAL_CLUSTER_NAME ?= kubeapps-additional
-IMAGE ?= kindest/node:v1.24.0
+IMAGE ?= kindest/node:v1.24.0@sha256:0866296e693efe1fed79d5e6c7af8df71fc73ae45e3679af05342239cdc5bc8e
 
 CLUSTER_CONFIG = ${KUBE}/kind-config-${CLUSTER_NAME}
 ADDITIONAL_CLUSTER_CONFIG = ${KUBE}/kind-config-${ADDITIONAL_CLUSTER_NAME}
--- a/script/makefiles/cluster-kind.mk
+++ b/script/makefiles/cluster-kind.mk
@@ -6,6 +6,7 @@
 KUBE ?= ${HOME}/.kube
 CLUSTER_NAME ?= kubeapps
 ADDITIONAL_CLUSTER_NAME ?= kubeapps-additional
+IMAGE ?= kindest/node:v1.24.0
 
 CLUSTER_CONFIG = ${KUBE}/kind-config-${CLUSTER_NAME}
 ADDITIONAL_CLUSTER_CONFIG = ${KUBE}/kind-config-${ADDITIONAL_CLUSTER_NAME}
@@ -14,6 +15,7 @@ ADDITIONAL_CLUSTER_CONFIG = ${KUBE}/kind-config-${ADDITIONAL_CLUSTER_NAME}
 # but is sufficient for the pod to be created so that we can copy the certs below.
 ${CLUSTER_CONFIG}:
 	kind create cluster \
+		--image ${IMAGE} \
 		--kubeconfig ${CLUSTER_CONFIG} \
 		--name ${CLUSTER_NAME} \
 		--config=./site/content/docs/latest/reference/manifests/kubeapps-local-dev-apiserver-config.yaml \
--- a/site/content/docs/latest/tutorials/getting-started.md
+++ b/site/content/docs/latest/tutorials/getting-started.md
@@ -29,6 +29,16 @@ For any user-facing installation you should [configure an OAuth2/OIDC provider](
 ```bash
 kubectl create --namespace default serviceaccount kubeapps-operator
 kubectl create clusterrolebinding kubeapps-operator --clusterrole=cluster-admin --serviceaccount=default:kubeapps-operator
+cat <<EOF | kubectl apply -f -
+apiVersion: v1
+kind: Secret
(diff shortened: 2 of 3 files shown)


Wouter053:
@absoludity when I try the instructions I see in the preview they are identical to what they were, 
image
and do not work on 1.24. 
image

 There is no token being created. You have to create one explicitly it in a separate command:

`kubectl create -f token.yml`

token.yml:
```

apiVersion: v1
kind: Secret
metadata:
   namespace: default
   name: operator-secret
   annotations:
       kubernetes.io/service-account.name: "kubeapps-operator"
type: kubernetes.io/service-account-token
```

site/content/docs/latest/tutorials/getting-started.md
```diff
@@ -29,6 +29,16 @@ For any user-facing installation you should [configure an OAuth2/OIDC provider](
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: kubeapps-operator-token
+  namespace: default
+  annotations:
+    kubernetes.io/service-account.name: kubeapps-operator
+type: kubernetes.io/service-account-token
+EOF
 ```
```
antgamdia, line 42: Given that we also have these instructions replictated in the website, I guess we would need to update this file as well: https://github.com/vmware-tanzu/kubeapps/blob/main/site/themes/template/layouts/partials/use-cases.html

script/makefiles/cluster-kind.mk
```diff
@@ -6,6 +6,7 @@
 KUBE ?= ${HOME}/.kube
 CLUSTER_NAME ?= kubeapps
 ADDITIONAL_CLUSTER_NAME ?= kubeapps-additional
+IMAGE ?= kindest/node:v1.24.0
```
antgamdia, line 9: Kind recommends using the very exact image tag for each Kind version.  However, as we can change it just setting `IMAGE` is not a big deal.  Perhaps this image also works out of the box.

https://github.com/kubernetes-sigs/kind/releases

> New Node images have been built for kind v0.12.0, please use these exact images (IE like ....

site/content/docs/latest/tutorials/getting-started.md
```diff
@@ -48,7 +58,7 @@ kubectl get --namespace default secret $(kubectl get --namespace default service
 Open a Powershell terminal and run:
 
 ```powershell
-[Text.Encoding]::UTF8.GetString([System.Convert]::FromBase64String($(kubectl get --namespace default secret $(kubectl get --namespace default serviceaccount kubeapps-operator -o jsonpath='{.secrets[].name}') -o jsonpath='{.data.token}')))
+[Text.Encoding]::UTF8.GetString([System.Convert]::FromBase64String($(kubectl get --namespace default secret kubeapps-operator-token -o jsonpath='{.data.token}')))
```
antgamdia, line 61: I'll give it a try soon just to double-check it works

antgamdia (approved):

site/content/docs/latest/tutorials/getting-started.md
absoludity: Thanks for checking the powershell :)

script/makefiles/cluster-kind.mk
absoludity: Yep, I'll update to the exact image - easy to set in the Makefile and we can do what we like when we override (as I did to test out 1.21).

site/content/docs/latest/tutorials/getting-started.md
absoludity: Done, I'll check the demo website once updated.

absoludity:
> @absoludity when I try the instructions I see in the preview they are identical to what they were, image and do not work on 1.24. 

Hi @Wouter053 . The instructions that I'd changed in my first commit were in the [getting-started.md](https://github.com/absoludity/kubeapps/blob/k8s1.24-dev/site/content/docs/latest/tutorials/getting-started.md#step-2-create-a-demo-credential-with-which-to-access-kubeapps-and-kubernetes), but yes, I'd missed updating the similar section presented on the website (hence it not changing). I've done that in the second commit :)

Issue #3846, opened 2021-11-29
Select CronJob API group version dynamically


antgamdia:
### Description:

As per #3832, we had to revert back some changes in the k8s API group we were using for creating the CronJobs and Jobs. 
Kubernetes 1.20 does not have the v1 version available yet... so it made some CI tests fail.

A solution could be to automatically discover the API group resource we use in the k8s client. This way, we would query "CronJobs" and we got back either "v1beta1" or "v1" depending on the target cluster we use.

Note that other k8s resources that are created in installation-time via Helm are already aware of the API changes (by means of helpers maintained by Bitnami).

antgamdia:
Just a quick update: on April 2022, v1.22 [becomes `stable` in GKE](https://cloud.google.com/kubernetes-engine/docs/release-notes), meaning that we'll default our tests to match this version.

We can either hardcode it to `v1` or use the discovery client like in https://github.com/kubeapps/kubeapps/blob/main/cmd/kubeapps-apis/plugins/resources/v1alpha1/server.go#L76, to discover which version (grv) should be using for talking to the API server.

antgamdia:
Perhaps it is also relevant for #4695 ?

antgamdia:
Whereas other Kubeapps components are actively using the Discovery API client, the AppRepository controller is not. However, currently, this is not that important anymore given two reasons:

- The CronJob API just was promoted to `v1`, so it is supposed to be stable from now on.
- This APIversion is nowadays widespread and, more importantly, Kubeapps is explicitly requiring a Kubernetes version > 1.21, where the `v1` version is available.

https://github.com/vmware-tanzu/kubeapps/blob/838df9102dd3c542a512a3587d767e1c6c807575/chart/kubeapps/Chart.yaml#L27

Therefore, I think the effort in dynamically selecting the API group is not worth it. Let's reopen it in case it becomes an issue for someone.

mecampbellsoup:
Ah sorry @antgamdia I just saw this (coming from my comment [here](https://github.com/vmware-tanzu/kubeapps/pull/4775#issuecomment-1430448870))... how much work is it to copy the pattern of "dynamic versioning" like other parts of kubeapps does?

cc @aanthonyrizzo

mecampbellsoup:
> 

By the way, was ` kubeVersion: ">=1.21.0-0"` set solely because of this change? Just curious.

antgamdia:
[...]

PR #4775, opened 2022-05-24
Use `BatchV1` instead of `BatchV1Beta1`

antgamdia:
### Description of the change

As I described at https://github.com/vmware-tanzu/kubeapps/pull/4743#issuecomment-1135769326, we do support >= k8s v1.21, so CronJob v1 API is there. GIven that k8s 1.25 will remove that version, we might want to upgrade the API version we use.

### Benefits

We will be able to run Kubeapps on v1.25.

### Possible drawbacks

Kubeapps will stop working on <= v1.20

### Applicable issues

- related #3846 

### Additional information

Let's see if CI thinks otherwise.


absoludity (approved):

mecampbellsoup:
> Kubeapps will stop working on <= v1.20

This is a big drawback for us. We are not upgrading to 1.21.X until March so as a result of this change we are a bit stuck.

I would suggest an alternative could be allowing `batch/v1` vs. `batch/v1beta1` to be configured via a value instead, or by introspecting the cluster directly akin to `kubectl api-resources` or `kubectl explain cronjob`:

```
(⎈ default:kubeapps)mcampbell-1 :: coreweave/k8s-services/dev ‹master*› » k explain cronjob | head -2
KIND:     CronJob
VERSION:  batch/v1beta1
(⎈ default:kubeapps)mcampbell-1 :: coreweave/k8s-services/dev ‹master*› » k api-resources | grep batch
cronjobs                             cj                   batch/v1beta1                               true         CronJob
jobs                                                      batch/v1                                    true         Job
```
You should trust that people using kubeapps know what they're doing enough to not enforce things like this - if we want to shoot ourselves in the foot, let us! 😄 

cc @aanthonyrizzo

PR #3450, opened 2021-09-20
Add versions - asset-syncer

antgamdia:
### Description of the change

This PR adds the `--version` command to `asset-syncer`

### Benefits

Our binaries will properly display their version

### Possible drawbacks

N/A

### Applicable issues

- related #3429

### Additional information

```
DOCKER_BUILDKIT=1 docker build -t kubeapps/asset-syncer:dev --build-arg "VERSION=$(git rev-parse HEAD)" -f cmd/asset-syncer/Dockerfile .

docker run  docker.io/kubeapps/asset-syncer:dev ./asset-syncer --version
```

PR #4777, opened 2022-05-25
Redact db password when logging configuration.

absoludity:

### Description of the change

Just noticed this while working on https://github.com/bitnami/charts/pull/10392 . Looks like it crept in when we added the extra logging message of the config, not realising it included the config (#3450)

### Benefits

Don't log db password.

### Applicable issues

- fixes #

### DIFF
--- a/cmd/asset-syncer/cmd/root.go
+++ b/cmd/asset-syncer/cmd/root.go
@@ -29,7 +29,9 @@ func newRootCmd() *cobra.Command {
 		Short: "Asset Synchronization utility",
 		PersistentPreRun: func(cmd *cobra.Command, args []string) {
 			serveOpts.UserAgent = server.GetUserAgent(version, serveOpts.UserAgentComment)
-			log.Infof("asset-syncer has been configured with: %#v", serveOpts)
+			serveOptsCopy := serveOpts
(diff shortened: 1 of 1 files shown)


castelblanque:
Good catch! Even if the `asset-syncer` might be deprecated soon, it is always good not to have this kind of leaks.
Thanks!

castelblanque (approved):

antgamdia (approved):
Great, thanks!

I'd say this issue also happens at `assetsvc`  (https://github.com/vmware-tanzu/kubeapps/blob/3b5eadc75e7a4ef4e5823fbda1721c1ba66d1a75/cmd/assetsvc/server/server.go#L24), but it is now legacy code, I guess.

absoludity:
> Great, thanks!
> 
> I'd say this issue also happens at `assetsvc` (
> 
> https://github.com/vmware-tanzu/kubeapps/blob/3b5eadc75e7a4ef4e5823fbda1721c1ba66d1a75/cmd/assetsvc/server/server.go#L24
> ), but it is now legacy code, I guess.

Does it? I checked and couldn't see any logging of the password there (or the parent `ServeOptions`, only the url? But yeah, we're not running it either way.

Issue #4764, opened 2022-05-23
Use the new Repositories API in the UI


antgamdia:
As the new API for managing Packages Repositories is becoming more stable, the current UI needs to be properly adapted in consequence.

Currently, we can test it with the Flux plugin, but there is ongoing work on both the Helm (#4662) and Carvel (#4681) plugins.

IMO, the new design should be carried out in multiple stages:

- Phase 0: https://github.com/vmware-tanzu/kubeapps/pull/4817
    - [x]  Start implementing towards that direction (autogenerated apprepo grpc clients, start using new API calls, report issues back, improve the API if needed, etc)
    - [x] Ensure feature parity between the new API and what Kubeapps already supports.
    - [x] Ensure the implementation is consistent across plugins.
    - PRs:
      - https://github.com/vmware-tanzu/kubeapps/pull/4930
      - https://github.com/vmware-tanzu/kubeapps/pull/4939
      - https://github.com/vmware-tanzu/kubeapps/pull/4954
      - https://github.com/vmware-tanzu/kubeapps/pull/4955
- Phase 1:  
   - Blocking issues:
   -  [x] Allow private global repos in the Helm plugin
      - Tracked at: https://github.com/vmware-tanzu/kubeapps/issues/5129
   -  [x] Re-enable e2e tests: https://github.com/vmware-tanzu/kubeapps/issues/4876
      - Done at https://github.com/vmware-tanzu/kubeapps/pull/5115
      - Blocked by https://github.com/vmware-tanzu/kubeapps/issues/5129
    - [x] Add a way to specify img pull secrets in the helm plugin for kubeapps-managed secrets. Currently, the UI supports both approaches, but the API only allows setting the secret manually. This has led to disabling the e2e test nº 03/10.
      - Tracked at: https://github.com/vmware-tanzu/kubeapps/issues/5015
      - Done at:
        - https://github.com/vmware-tanzu/kubeapps/pull/5081
        - https://github.com/vmware-tanzu/kubeapps/pull/5109
    - [x] Add the `accessLevel` (= type of auth) to the  PackageRepositorySummary API + use it in the UI
      - Trivial, done at: https://github.com/vmware-tanzu/kubeapps/pull/5003
    - [x]  The helm plugin doesn't have a status.ready
      - Trivial, done at: https://github.com/vmware-tanzu/kubeapps/pull/5006
    - [x] The auth type "Bearer" implementation is inconsistent across the plugins
[...]

PR #4795, opened 2022-05-27
Rename `App Repositories` to `Package Repositories`

antgamdia:
### Description of the change

Trivial PR just renaming the old "App repo" thing to something more aligned to our concept of "package".
This is a prequel PR, but can be safely merged into main.

### Benefits

More consistent naming.

### Possible drawbacks

N/A

### Applicable issues

- related #4764

### Additional information

Let's see if the CI agrees :P

### DIFF
--- a/dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
+++ b/dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
@@ -614,8 +614,8 @@ export function AppRepoForm(props: IAppRepoFormProps) {
 
       {namespace === kubeappsNamespace && (
         <p>
-          <strong>NOTE:</strong> This App Repository will be created in the "{kubeappsNamespace}"
-          namespace and packages will be available in all namespaces for installation.
+          <strong>NOTE:</strong> This Package Repository will be created in the "{kubeappsNamespace}
+          " namespace and packages will be available in all namespaces for installation.
         </p>
       )}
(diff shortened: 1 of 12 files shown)


dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
```diff
@@ -614,8 +614,8 @@ export function AppRepoForm(props: IAppRepoFormProps) {
 
       {namespace === kubeappsNamespace && (
         <p>
-          <strong>NOTE:</strong> This App Repository will be created in the "{kubeappsNamespace}"
-          namespace and packages will be available in all namespaces for installation.
+          <strong>NOTE:</strong> This Package Repository will be created in the "{kubeappsNamespace}
+          " namespace and packages will be available in all namespaces for installation.
```
absoludity, line 618: Won't HTML assume that new-line there is a word-break? (Not what's wanted in that situation, with the quote, I don't think?)

absoludity (approved):
Thanks for the other drive-by fixes (doc links and wording).

dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
antgamdia: image

No, IRL it seems it works as expected. It is a breakline introduced by the linter.

PR #4796, opened 2022-05-27
Second step of package repos API - Package Repository API for Helm

castelblanque:
### Description of the change

Second step on top of first one in #4723.
Added custom details to Helm AppRepositories for AddPackageRepository.
Added Docker JSON authentication handling.
Added more tests.
Added validation flag `perform_validation` in the custom details that allows to perform the repo URL check that is being done currently as a separate API call when creating a repo.

### Applicable issues

Partially implements #4662


absoludity:
Looks like you want a `git rebase -i main` (after pulling a recent main) and drop the commits that already landed as part of your part 1?

castelblanque:
> Looks like you want a git rebase -i main (after pulling a recent main) and drop the commits that already landed as part of your part 1?

Yes, that is why it is still a Draft PR. Today I will raise the 3rd and final PR for the packages repo API, and get this one groomed too.

pkg/kube/kube_handler.go
```diff
@@ -506,6 +506,7 @@ func (a *userHandler) applyAppRepositorySecret(repoSecret *corev1.Secret, reques
 	return err
 }
 
+// Deprecated: Remove when the new Package Repository API implementation is completed
```
antgamdia, line 509: Great! I'm really looking forward to start removing old code from the repo and, especially, moving stuff from the root `/pkg` folder to the `/kubeapps-apis` directory when possible.

antgamdia (approved):
Thanks for the changes! Can't wait to use them in the UI :)

pkg/kube/kube_handler.go
castelblanque: Yes! When the Package Repository API is at full speed we can get rid of a lot of legacy code.

PR #4802, opened 2022-05-30
Update to CRA 5

### LINKED ISSUE #4018: Update to CRA react-scripts 5.x.x
### Description:

PRs intended to reduce some security issues (like #4011 ) are being rejected due to the non-trivial required effort. The main problem is react-scrips upgrading `webpack` to the major release 5. There are two types of problems:

* Webpack 5 not including the polyfills it used to include (https://github.com/facebook/create-react-app/issues/11756). Consequently, all our deps relying on non-browser deps (like `url`, `crypto`. `buffer`, `stream`, etc) are not being properly ported.
    * There is an ongoing PR aimed at simplifying the webpack config w/o ejecting (https://github.com/facebook/create-react-app/pull/11764).
    * Alternatively, `react-app-rewired` can be used.
* Webpack 5 now enforces the rule "a package.json top-level field type with a value of module should make modules load as ES modules", that is, they have to define an extension. Currently, `clarity` deps are not doing that (an issue should be eventually open in their repo).
    * Some related issues in CRA: https://github.com/facebook/create-react-app/issues/11865 and https://github.com/facebook/create-react-app/issues/11769
    * Alternatively, we can "patch" the affected deps as mentioned in https://github.com/facebook/create-react-app/issues/11769#issuecomment-996389278.

My current recommendation is to wait until the CRA team come up with a recommendation for both sort of problems.

1. Update to CRA >= 5.0.0

The app cannot be started or built.

Everything up and running.

antgamdia:
### Description of the change

Today, I performed a `yarn install` to update my local deps with the latest dependabot updates. However, I got an error related to webpack 4 not being able to parse some new(ish) JS features like nullish coalescing or optional chaining. The offending dependency was the latest version of `yaml` dependency that we recently updated.

Instead of work around this issue, I decided to invest some time to look into #4018.

This PR updates to CRA >=5 adding the usual workaround I mentioned in the aforementioned issue: adding the polyfills manually.
However, instead of adding them independently, I've added `node-polyfill-webpack-plugin`, which should add the required ones more transparently.

The webpack tweak is performed using `craco` a tool to inject some webpack configuration without ejecting the `react-scripts` way.

### Benefits

We will get rid of many security notices related to CRA < 5. 

### Possible drawbacks

Not "pure" react-scripts way, but... using `craco` is way less intrusive than ejecting. 

### Applicable issues

- fixes #4018

### Additional information

Working locally, but let's see if our CI thinks otherwise :P

Edit: it seems some eslint rules have been added, I'll fix the offending lines in the tests and will push again. DONE

### DIFF
--- a/dashboard/craco.config.js
+++ b/dashboard/craco.config.js
@@ -1,0 +1,17 @@
+const webpack = require("webpack");
+const NodePolyfillPlugin = require('node-polyfill-webpack-plugin');
+
+module.exports = {
+  webpack: {
+    configure: {
+      plugins: [
+        new NodePolyfillPlugin(), // add the required polyfills (not included in webpack 5)
+        new webpack.ProvidePlugin({
+          process: 'process/browser.js',
+          Buffer: ['buffer', 'Buffer'],
+        }),
+      ],
+      ignoreWarnings: [/Failed to parse source map/], // ignore source map warnings
+    },
+  },
+};
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -7,10 +7,10 @@
   "scripts": {
     "build": "npm-run-all build-css compile-lang build-js",
     "build-css": "sass src/ --no-source-map && yarn run copy-clr",
-    "build-js": "react-scripts build",
+    "build-js": "craco build",
     "compile-lang": "formatjs compile-folder lang src/locales/ --ast --format simple",
     "copy-clr": "shx cp ./node_modules/@clr/ui/clr-ui-dark.min.css public/clr-ui-dark.min.css && shx cp node_modules/@clr/ui/clr-ui.min.css public/clr-ui.min.css && shx cp ./node_modules/@clr/ui/clr-ui-dark.min.css.map public/clr-ui-dark.min.css.map && shx cp ./node_modules/@clr/ui/clr-ui.min.css.map public/clr-ui.min.css.map",
-    "eject": "react-scripts eject",
+    "eject": "craco eject",
     "eslint-check": "eslint --config ./.eslintrc.json 'src/**/*.{js,ts,tsx}' --max-warnings=0",
     "eslint": "eslint --config ./.eslintrc.json 'src/**/*.{js,ts,tsx}' --fix --max-warnings=0",
     "extract-lang": "formatjs extract 'src/**/*.ts*' --out-file lang/en.json --id-interpolation-pattern '[sha512:contenthash:base64:6]' --format simple",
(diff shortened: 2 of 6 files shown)


dashboard/craco.config.js
```diff
@@ -0,0 +1,17 @@
+const webpack = require("webpack");
+const NodePolyfillPlugin = require('node-polyfill-webpack-plugin');
+
+module.exports = {
+  webpack: {
+    configure: {
+      plugins: [
+        new NodePolyfillPlugin(), // add the required polyfills (not included in webpack 5)
```
antgamdia, line 8: This is preventing us from doing something like:

````
        fallback: {
          process: require.resolve("process/browser"),
          stream: require.resolve("stream-browserify"),
          buffer: require.resolve("buffer"),
          timers: require.resolve("timers-browserify")
          ... // a line per each polyfill we (not really "we" but any random transitive dependency we use) need
        },
````

File: dashboard/src/components/Config/AppRepoList/AppRepoForm.test.tsx [hunk 2]
```diff
@@ -699,13 +780,16 @@ describe("when the repository info is already populated", () => {
...
       } as any;
-      const wrapper = mountWrapper(defaultStore, <AppRepoForm {...defaultProps} repo={repo} />);
-
+      let wrapper: any;
+      await act(async () => {
+        wrapper = mountWrapper(defaultStore, <AppRepoForm {...defaultProps} repo={repo} />);
+      });
       await waitFor(() => {
+        wrapper.update();
         expect(wrapper.find("textarea").at(0).prop("value")).toBe("nginx");
```
antgamdia, line 789: Even if the tests were passing, they were throwing a message like

````
When testing, code that causes React state updates should be wrapped into act(...):
     
     act(() => {
       /* fire events that update state */
     });
     /* assert on the output */
     
This ensures that you're testing the behavior the user would see in the browser. Learn more at https://fb.me/react-wrap-tests-with-act
````

Since mounting `AppRepoForm` triggers an state update via `useEffect` in an async fetch function, it has to be wrapped into an `act`. Then, a `.update()` call is often required.

I've performed this change in the remaining test cases.

File: dashboard/package.json [hunk 3]
```diff
@@ -122,7 +124,7 @@
     "typescript": "^4.7.2"
   },
   "resolutions": {
-    "@babel/parser": "^7.15.3",
+    "@babel/parser": "^7.18.4",
```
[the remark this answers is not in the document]
absoludity: Thanks for the explanation here - sounds like it could be worth a comment in the file at this point to our future selves?

dashboard/src/components/Config/AppRepoList/AppRepoForm.test.tsx
absoludity: Excellent thanks. Yeah, we'd started using `act` for async tests in other places, but only as errors were raised in tests. Thanks for doing them all here.

absoludity (approved):
Excellent, thanks for finding a better alternative. cra-configuration override makes it much simpler for us.

PR #4807, opened 2022-05-31
Third and final step of package repos API - Package Repository API for Helm

### LINKED ISSUE #4662: Implement Package Repository API for Helm plugin

castelblanque:
### Description of the change

Third and final step on top of the second PR #4796.

This adds the following implementations and tests:

- GetPackageRepositoryDetail
- GetPackageRepositorySummaries
- UpdatePackageRepository
- DeletePackageRepository

I had to change the original `AppRepository` CR definition for the field `ResyncRequests` from type `uint` to `int`.

### Benefits

Package repository API is available for direct Helm plugin.

### Possible drawbacks

I haven't seen any, but there might be misalignments in the `AppRepository` CRDs data if both the old and this new API are used together over the same CRD. Due to the management of secrets, etc. 
However, this scenario could only happen during the transition time and should not happen in normal usage at all.

### Applicable issues

- resolves #4662

### Additional information

The reason for changing the data type in the original CRD was due to the k8s machinery `DeepCopyJSONValue` being unable to handle `uint` (see [here](https://github.com/kubernetes/apimachinery/blob/master/pkg/runtime/converter.go#L639)).


cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
```diff
@@ -1060,32 +1057,124 @@ func (s *Server) AddPackageRepository(ctx context.Context, request *corev1.AddPa
...
+	var customDetails *helmv1.RepositoryCustomDetails
+	if request.CustomDetail != nil {
+		customDetails = &helmv1.RepositoryCustomDetails{}
+		if err := request.CustomDetail.UnmarshalTo(customDetails); err != nil {
+			return nil, status.Errorf(codes.InvalidArgument, "customDetails could not be parsed: [%v]", request.CustomDetail)
+		}
+		log.V(4).Infof("+helm upgrade repo %s customDetails [%v]", repoRef.Identifier, customDetails)
+	}
+
+	helmRepo := &HelmRepository{
```
castelblanque, line 1139: Using this custom `HelmRepository` to avoid having 8 parameters in functions signatures :)

antgamdia (approved):
Haven't tested it IRL, but looking forward to it while I'm implementing the corresponding changes in the UI.

PR #4812, opened 2022-06-01
Add full-text search to the website

antgamdia:
### Description of the change

This PR (plus the configuration already performed in the Algolia side) brings full-text search in our website.

+ squeezing some required changes now that we have the DNS poiting to https://kubeapps.dev (if redirected to .com, clear your cache)

### Benefits

Users will be able to perform full text queries to look up the docs 

### Possible drawbacks

The index is not being properly crawled/polished up, therefore the results are not the best ones.

### Applicable issues

- related #3953

### Additional information

search

### DIFF
--- a/site/config.yaml
+++ b/site/config.yaml
@@ -1,9 +1,9 @@
-baseURL: "https://kubeapps-dev.netlify.app/"
+baseURL: "https://kubeapps.dev/"
 languageCode: "en-us"
 title: "Kubeapps"
 theme: "template"
 outputs:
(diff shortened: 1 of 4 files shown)


antgamdia:
Test it here: https://deploy-preview-4812--kubeapps-dev.netlify.app/docs/latest/

absoludity (commented):
Other than the config, it looks great! RE the config, looks like we're exposing our algolia private id/key here, rather than adding it during the build step... let me know if I'm missing something?

PR #4813, opened 2022-06-01
Add `userManagedSecrets` to the flux plugin config in the chart

antgamdia:
### Description of the change

I've noticed there were two missing Flux plugin-wide config items not being present in the chart config options. Besides, one of them wasn't really that configurable, since the value was hardcoded.
This PR is to add the those fields in the chart + regenerate the readme.

### Benefits

More customizable Flux plugin config.

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

N/A


gfichtenholt (commented):
Antonio, I feel this is might be little misguided. The fact that the flag is in plug-in config at all was just a temporary hack, until we find the right scope for it. We still haven't settled it. https://vmware.slack.com/archives/C032Y8QL77C/p1653592281210039 Better settle the question first, and then implement accordingly, IMO

gfichtenholt (approved):
Having said, feel free to merge this. Just be aware this might have to be re-done soon. Thanks

antgamdia:
> Antonio, I feel this is might be little misguided. The fact that the flag is in plug-in config at all was just a temporary hack, until we find the right scope for it. We still haven't settled it. https://vmware.slack.com/archives/C032Y8QL77C/p1653592281210039 Better settle the question first, and then implement accordingly, IMO

Yep, I get your point, but, on the other hand, there was another field (eg. `defaultUpgradePolicy`) that wasn't added to the chart, so I added it as well.

Regarding the `userManagedSecrets`, I added it just because I was playing around with the values for the UI revamp and I wanted to change it more easily.

> Having said, feel free to merge this. Just be aware this might have to be re-done soon. Thanks

Great, I'd rather add it in the meantime until we come up with another alternative. Fortunately, it is pretty easy to remove it.

PR #4815, opened 2022-06-01
Carvelrepos types

dlaloue-vmware:
This change adds support for other repository types than just imgpkg bundle. It supports all types for read operations, and all types (but inline) for write operations.
The change also uses the custom details field for read/write of fields not captured by the common schema (e.g. tag or ref version selection).

note that in this change, the custom details are mirrored from the kapp-controller proto types, they do not reuse the existing proto types (this is left to a later change if possible to do).


castelblanque (approved):
LGTM

absoludity:
@dlaloue-vmware is a member of vmware-tanzu org. Removing cla-rejected label.

PR #4817, opened 2022-06-01
Use the new Repositories API in the current UI

antgamdia:
> **Warning**
> This is an ongoing PR, created as a draft to start sharing the code and progress.
> ~I have some questions regarding the auth mapping, especially in flux, I'll write something up tomorrow~
> ~Now dealing with Any serialization...~
> Continue with an IRL test
> Splitting the PR into smaller ones
> Removing the secret management ui
> --> For doing so, I'm adapting the code in this combined PR first, then, I'll extract as much as possible as I start fixing the test cases. 

```console
Test Suites: 6 failed, 128 passed, 134 total
Tests:       55 failed, 1069 passed, 1124 total
```

### Description of the change

TBD

### Benefits

TBD

### Possible drawbacks

This PR does not implement the proposed flow for a

### Applicable issues

- related #4764 (phase 0)

##### For follow-up PRs
- [ ]  `getPluginByName` is a hardcoded list of plugin objects for getting the object with the version set. Ideally, it should fetch it from the actual configured plugins.
- [ ] Delete `url.apprepositories.get` once we don't need it anymore. It also means we will be able to remove the endpoints from kubeops!
- [ ] Add the `accessLevel` (= type of auth) to the  PackageRepositorySummary API + use it in the UI

##### Other comments
- The existing `initialRepos` in the chart is not working anymore. Check
- Minor UI tweak if Flux were to support OCI

#####  To be done in this PR:
- [x] Support more TLS/Auth options in the UI
    - [x] `auth.dockerCreds` for creating the secret on a user's behalf
    - [x] `tlsConfig.secretRef` for passing CA certificate inside a secret
    - [x] `auth.tlsCertKey` for handling certificate-based auth - though it isn't supported by every plugin
- [X] Support the `resync/resync all` option or get rid of it entirely.
    - Outcome: remove it, we are not supporting it anymore
- [X] Support `syncJobPodTemplate` in the helm plugin + use it in the UI
    - Outcome: remove it, we are not supporting it anymore
- [x] The form, when editing, is not populating every field as expected
    - Cause: the API is not setting some customDetail (like `performValidation`). Other fields do require further investigation.
- [X] Question: Flux doesn't support `namespaceScoped` to be true, so every repo will become "global". The new UI should avoid the "global==in kubeapps ns" assumption I guess, but for now, I think it is good enough. 
- [X] Question: Helm is creating the repo with the validate option set to true. Clarify it with Rafa. If it's ok, change the logic in the UI
    - Outcome: an isolated endpoint for validation no longer make sense. Let's remove it entirely, so the "add" logic subsumes the validation one.

##### Demo: adding both Flux and Carvel repos (w/o authentication)

addRepos

##### Demo: handling user-managed secrets and kubeapps-managed secrets
addReposSecrets


antgamdia:
Closing in favor of https://github.com/vmware-tanzu/kubeapps/pull/4954, as it contains the proper test cases and it is based on top of incremental PRs to better communicate the changes.

Issue #4832, opened 2022-06-03
Operators e2e test randomly failing


castelblanque:
**Describe the bug**
Lately some e2e tests runs fail due to failure in

`tests/operators/06-operator-deployment.spec.js:8:1 › Deploys an Operator`

**To Reproduce**

Hard to reproduce, it happens some times e.g. [this CI run](https://app.circleci.com/pipelines/github/vmware-tanzu/kubeapps/7398/workflows/e1a32fa9-7d4f-4b70-80a5-2bd3c78e1648/jobs/85652).

**Expected behavior**
E2E tests run smoothly

**Screenshots**
![image](https://user-images.githubusercontent.com/67455978/171872144-a163cec2-4f85-4077-9428-9694ac6c4015.png)

dlaloue-vmware:
The issue is not fixed, as i had more than 5 failed run in a row on 2 different changes: e.g. https://app.circleci.com/pipelines/github/vmware-tanzu/kubeapps/7484/workflows/53ebeea6-5d27-4712-b645-5d2fb3185842/jobs/86399

The code that makes the test fail is the following line of code that was changed:
      await page.waitForTimeout(utils.getDeploymentTimeout());
The waitForTimeout is in fact a sleep, so it will wait for the entire duration of the deployment timeout value.

The reason why this change made the test fail consistently instead of intermittently is that in all the failure cases, the "deployment timeout" ends up being "86" instead of the expected "2". And so the waitFortimeout waits for 86mn but times out at the test timeout (6mn).

I was able to trace this value "86" as being half of 172, which is the beginning of the additional cluster ip address 172.18.0.3.

so for some unknown reason yet, when CI executes the test, there are times where the environment is corrupted and not all variables are properly passed in.
The TEST_TIMEOUT_MINUTES variable ends being 172.18.03, which will lead to a deployment timeout of 86.

I am still trying to trace back to why the environment is not properly set

dlaloue-vmware:
failed test showing the deployment timeout:
<img width="959" alt="e2e fail" src="https://user-images.githubusercontent.com/74025858/172540743-c2c19cf9-8381-4e44-bf72-7672b1ef41b5.png">

successful test, deployment timeout at 2mn:
<img width="953" alt="e2e success" src="https://user-images.githubusercontent.com/74025858/172540836-27fc5e00-5e49-495b-ab47-fd487357c4ce.png">

screenshot showing the environment for failed case:
[...]

PR #4833, opened 2022-06-03
Fix for carvel e2e random failure

### LINKED ISSUE #4627: Carvel e2e test occasionally fails with pending install
### Description:

Occasionally we're seeing failures for the carvel e2e test which end with:

```
  1) [chromium] › tests/carvel/04-default-deployment.spec.js:8:1 › Deploys package with default values in main cluster 

    page.waitForSelector: Timeout 120000ms exceeded.
    =========================== logs ===========================
    waiting for selector "css=.application-status-pie-chart-number >> text=3" to be visible
    ============================================================

      36 |
      37 |   // Assertions
    > 38 |   await page.waitForSelector("css=.application-status-pie-chart-number >> text=3", {
         |              ^
      39 |     timeout: utils.getDeploymentTimeout(),
      40 |   });
      41 |   await page.waitForSelector("css=.application-status-pie-chart-title >> text=Ready", {

        at /app/tests/carvel/04-default-deployment.spec.js:38:14

    attachment #1: video (video/webm) --------------------------------------------------------------
    reports/tests-carvel-04-default-deployment-Deploys-package-with-default-values-in-main-cluster-chromium/b52d1dd878a4c79dd9e733a3a71abd22.webm
    ------------------------------------------------------------------------------------------------

    attachment #2: trace (application/zip) ---------------------------------------------------------
    reports/tests-carvel-04-default-deployment-Deploys-package-with-default-values-in-main-cluster-chromium/trace.zip
    Usage:

        npx playwright show-trace reports/tests-carvel-04-default-deployment-Deploys-package-with-default-values-in-main-cluster-chromium/trace.zip

    ------------------------------------------------------------------------------------------------

    attachment #3: screenshot (image/png) ----------------------------------------------------------
    reports/tests-carvel-04-default-deployment-Deploys-package-with-default-values-in-main-cluster-chromium/test-failed-1.png

```

and the screenshot showing:
test-failed-1

We need to investigate why this happens only occasionally, as though the new deployment isn't watched.

castelblanque:

### Description of the change

Tries to fix #4627. Given the nature of E2E tests, it is hard to know if this will fix the random error.

I did 8 runs of the E2E tests with this change: no failures.
See [CircleCI branch here](https://app.circleci.com/pipelines/github/vmware-tanzu/kubeapps?branch=4627-carvel-e2e-error&filter=all). The error seen in run 7398 is unrelated to this and caused by another bug #4832.

The proposed fix is based in the investigation done comparing the trace data provided by Playwright and a stable installation of Kubeapps.

### Benefits

Carvel error in E2E hopefully is gone.

### Possible drawbacks

I was unable to reproduce locally the error, so not 100% sure this will fix it.

### Applicable issues

- fixes #4627


absoludity (commented):
Thanks for looking into this @castelblanque . I reckon your suspicion about a race-condition there is correct, and I think there might be a way to reproduce this locally - more below. But first, wondering why you've added this change here to the core packages endpoint - so that it makes this decision for all plugins?

That is, it seems like the carvel plugin should be returning a NotFound if it receives a request for the resource refs while the related app is not yet reconciled? The spot in the code where this could happen, is, I think, at:

https://github.com/vmware-tanzu/kubeapps/blob/6f886f453221cc77069e270a5f668bdb44b4c89d/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_resources.go#L543

I'd be happy to +1 this change if we move the notfound there (and a quick unit test if possible to document it / avoid regression).

Now, *if* you want to reproduce this locally, I reckon it would be as simple as temporarily disabling (replicas to zero?) the controller that watches `App` CRs and creates the related k8s resources, then creating the app in Kubeapps. That would mean that Kubeapps would go ahead and create the InstalledPackage, kapp-controller would create the `App`, but it wouldn't be reconciled so no resources generated, giving you plenty of time to witness the issue.... I *think*. But only if you're keen to understand exactly what's going on, not necessary.

The other thought here is: you've potentially found a race-condition here for carvel packages that ends up returning an Ok response with zero refs. But does the UX need to work in this scenario for cases where there are validly no resource refs that we'd track? (eg. helm chart that just creates config - weird, but they are around). I'm not asking that we deal with that here, just wondering.

castelblanque:
> wondering why you've added this change here to the core packages endpoint - so that it makes this decision for all plugins?

Exactly. Don't we want to respond "NotFound" when there are 0 resources for a given ref? Does it make sense that an app is considered "existing" if it doesn't have any resource created at all? That assumes we are able to track until the last resource created by an app 😄 

> I'd be happy to +1 this change if we move the notfound there (and a quick unit test if possible to document it / avoid regression).

+1 to the unit test, of course. But continuing with the topic above, why having it only for Carvel plugin?

> But does the UX need to work in this scenario for cases where there are validly no resource refs that we'd track? (eg. helm chart that just creates config - weird, but they are around)

If that is the case, why not considering configmaps just as another resource instead? Is it that they don't have the same labels for tracking them? I mean, why keeping resources hidden to the viewer?

Just my point of view (maybe I'm missing something):
- if an app returns 0 resources -> continue polling (Haven't checked but I believe you added exponential backoff?).
- If an app returns > 0 resources -> stop polling. 
In any case, regardless of the plugin.

Many thanks for the hint for reproducing it locally. It makes absolute sense, will try!

castelblanque:
> it would be as simple as temporarily disabling (replicas to zero?) the controller that watches App CRs  ... , kapp-controller would create the App, but it wouldn't be reconciled 

Just checking I noticed that the only controller existing is `kapp-controller`. Running `k scale deploy -n kapp-controller kapp-controller --replicas=0` makes the whole Carvel functionality to stop working: applications can't be installed from Kubeapps, `PackageInstall` are not reconciled, etc.
Could you please elaborate a bit more on how to do it?

absoludity:
> > wondering why you've added this change here to the core packages endpoint - so that it makes this decision for all plugins?
> 
> Exactly. Don't we want to respond "NotFound" when there are 0 resources for a given ref? Does it make sense that an app is considered "existing" if it doesn't have any resource created at all? That assumes we are able to track until the last resource created by an app smile

I don't think so, no - but I'm happy to defer to our in-house API expert :) My understanding is that a 404 NotFound is for a response where "the origin server did not find a current representation for the target resource or is not willing to disclose that one exists." (from the [http 1.1 spec](https://datatracker.ietf.org/doc/html/rfc7231#section-6.5.4)). So if this URI was for a resource which did not exist (like a particular resource or available package), then 404 is what we want. Even if it was for a list of resource refs for a *package* or kapp App that didn't exist (as in, the package itself didn't exist), then a 404 would also be appropriate, since there is no representation for the target resource.

But if we're requesting a collection of resources for a package or app that exists, then whether the collection is empty or has multiple results, the collection has a valid endpoint and has a current representation and the representation for that collection should be returned,

Now, our case here is kind of in between, in that a specific plugin has a case where (I think?) the App for which the resource refs collection is being requested doesn't exist (or hasn't been reconciled?) at the time of the request - so a 404 may be appropriate there (that's similar to the flux case where a 404 is returned and handled in the UX for the same reason, I think). Ideally it'd be something non-cacheable like a not-ready-yet-try-again-later for correct semantics but I can't see something for that.

So yes, this case is tricky (it's why I created #4337), but I don't think it means that we should convert every empty response returned by our (aggregated) API to a 404. But again, feel free to chat with @antgamdia about it - I'd trust his expertise in API semantics over mine :)

> 
> > I'd be happy to +1 this change if we move the notfound there (and a quick unit test if possible to document it / avoid regression).
> 
> +1 to the unit test, of course. But continuing with the topic above, why having it only for Carvel plugin?
> 
> > But does the UX need to work in this scenario for cases where there are validly no resource refs that we'd track? (eg. helm chart that just creates config - weird, but they are around)
> 
> If that is the case, why not considering configmaps just as another resource instead? Is it that they don't have the same labels for tracking them? I mean, why keeping resources hidden to the viewer?

Yeah, you're absolutely correct there - that was a bad example as we would return all the resource refs including config maps (I think I was thinking about how we then filter the resource refs and only stream the related resources that change, like pods etc.). So yes, there will always be resource refs for an install package or app, eventually.


> 
> Just my point of view (maybe I'm missing something):
> 
>     * if an app returns 0 resources -> continue polling (Haven't checked but I believe you added exponential backoff?).
> 
>     * If an app returns > 0 resources -> stop polling.
>       In any case, regardless of the plugin.

Yep, we agree, I just think semantically it should be (eventually, not in this PR):

* if an app is not finishing reconciling (see #4337 ) -> continue polling

* if an app has finished reconciling -> stop polling

and because we've not yet done that work (#4337 ), I'd currently gone with "if the request returns a 404, keep polling". I'm not strongly against "if an app returns 0 resources", it just seems like a step away from #4337 , let's see.

> 
> 
> Many thanks for the hint for reproducing it locally. It makes absolute sense, will try!

Hmm, sounded like a good hint, but as you've said:

> > it would be as simple as temporarily disabling (replicas to zero?) the controller that watches App CRs  ... , kapp-controller would create the App, but it wouldn't be reconciled
> 
> Just checking I noticed that the only controller existing is `kapp-controller`. Running `k scale deploy -n kapp-controller kapp-controller --replicas=0` makes the whole Carvel functionality to stop working: applications can't be installed from Kubeapps, `PackageInstall` are not reconciled, etc. Could you please elaborate a bit more on how to do it?

I'd assumed they were different controllers managing the different CRs there. No more thoughts there, other than running a custom kapp-controller that delayed the reconciliation, but doesn't sound worthwhile.

castelblanque:
I believe GRPC is lacking some granularity here. In the case of HTTP, maybe a good fit would be `202` status (see [here](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.2.3)). Not the perfect one, but closer than the statuses offered by GRPC.
Although I don't think there is a GRPC status code that would [would suit us](https://grpc.github.io/grpc/core/md_doc_statuscodes.html), maybe 'UNAVAILABLE (14)' could serve, `The service is currently unavailable. This is most likely a transient condition, which can be corrected by retrying with a backoff.`.

If what we want is a way to tell the consumer client "_There is an app, but its resources are pending, they might or might not be created. Try again._", I see two possibilities that don't involve modifying the returned status as we have it now:
- Add a custom header to the response with the information we are lacking in the status (e.g. [here](https://github.com/vmware-tanzu/kubeapps/blob/main/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go#L994)). 
  ```go
  headers := metadata.Pairs("x-resource-status", "PENDING")
  grpc.SendHeader(ctx, headers)
  ```
  Personally, I don't like this option as the real response status will have to be composed by the consumer (GRPC status code + custom header).

- Extend the `GetInstalledPackageResourceRefsResponse` struct with a status for the resources (not for the app). So that the response would be status=OK and content:
  ```json
  {
    "context": "...",
    "resourceRefs": [],
    "status": "PENDING"
  }
  ```

Status returned could be "PENDING" or any other custom one.
With either of these responses, a consumer (like UI), could decide if continue polling or not.
WDYT?

absoludity:
> Status returned could be "PENDING" or any other custom one.
> With either of these responses, a consumer (like UI), could decide if continue polling or not.
> WDYT?

I think that sounds like a good proposal - thanks for adding it to the other issue too. But right now, let's not block landing this change maybe? Currently the dashboard client is polling as long as it gets a 404. This PR can update the carvel plugin to do so and land as the current work-around used in the client. Then we can tackle the #4337 and update the client separately?

To avoid any more waiting for timezones, I'm going to +1 this based on what seems to be the mutual understanding?

absoludity (approved):
+1 (I'm keen for us to move the snippet that returns a 404 to the carvel plugin, as the current work-around, then deal investigate and fix #4337 with Rafa's suggested path forward there). But if you prefer to do that alternative straight away - that's fine too.

castelblanque:
Agree, thanks!
Moved check for 0 resources to the Kapp plugin and added test.

PR #4841, opened 2022-06-06
cache bitnami index.yaml during build and use cached copy at runtime

gfichtenholt:
1)  cache bitnami index.yaml during build and use cached copy at runtime to avoid intermittent errors from bitnami.com

2) changed signature of test utils from `(name, namespace string)` to `(name types.NamespacedName)` to avoid inadvertent  errors where function is invoked and input params are in the wrong order

ALL integration+unit flux tests are passing

This PR does not have any changes to production code.


antgamdia (approved):
Awesome! Given the recent changes on the bitnami `index.yaml`, it's great you found a way to isolate our test cases from those transient network errors.
Thanks!

gfichtenholt:
Thank you Antonio

ppbaena:
Thank you Greg. It's always good news to see tests are passing

Issue #2875, opened 2021-05-24
pinniped-proxy should cache credentials for lifetime of credential request


absoludity:
### Description:

When initially creating the pinniped-proxy service, we had planned to cache the response of the token credential request and re-use it for subsequent requests.

As it is, it worked OK without this initially, but it is still overkill to be doing so for every request, and may be too slow in the future when pinniped tries to use the [certificate signing requests API](https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/).

viktor-haag:
Continuing the discussion we had in #2363, I have one more observation which might help. From what I could see, it seem that requests to **/api/clusters/default/api/v1** and **/api/clusters/default/api/v1/namespaces/<mynamespace>** always run into the 60s timeout. Aside from that, the response time of the other api calls is pretty inconsistant. If I directly hit one of the apis via the browser I get in extreme cases an instant response but am also as likely to run into the 60s timeout.

Aside from that, is there a combination of Kubeapps and Pinniped versions which you could recommend? I'd try a fresh install and check if the problem still persists there.

antgamdia:
Thanks for your comments. It is something we have to investigate.

wrt your question: 
Kubeapps 2.3.2 is compatible with Pinniped >= v6.0.0 (latest is now 0.9.2)
Kubeapps 2.3.3 is compatuble with any Pinniped version (including < v6.0.0)

Apart from that, there are no extra requirements on the versions.

PR #4860, opened 2022-06-07
[pinniped-proxy] Create identity cert directly from x509.

absoludity:

### Description of the change

Just a fly-by while getting setup with pinniped-proxy locally again, this PR does 3 things:

* Updates to use the latest pinniped version (0.17.0) in our dev environment
* Updates the dev environment targets to allow creating a single-cluster environment
* Simplifies the `identity_for_exchange` now that the underlying `native_tls::Identity` supports creating an identity cert directly from an x509.

### Benefits

Simpler code, easier dev env.

### Applicable issues

- ref #2875

### Additional information

Although I've tested in real life, I'm keen to see if it's trivial to add a unit test for this also, so leaving in draft until I've tried.

### DIFF
--- a/cmd/pinniped-proxy/Cargo.lock
+++ b/cmd/pinniped-proxy/Cargo.lock
@@ -132,6 +132,7 @@ dependencies = [
  "num-integer",
  "num-traits",
  "serde",
+ "time",
  "winapi",
 ]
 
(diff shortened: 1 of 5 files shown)


absoludity:
> Great! Nice to see the logic can be simplified now! Thanks

Thanks. Added the unit tests.

Issue #4869, opened 2022-06-08
[Repository] change interval type from uint32 to time+unit type


dlaloue-vmware:
In the PackageRepository struct, the interval field is of type uint32.
It is more common to find such interval be of a type that allows users to enter the value using unit formats - e.g. 2m30s

dlaloue-vmware:
Update:

I started looking into this but ran into several issues:

The first issue is that 'buf' does support dependencies, but only if they are published in the buf registry (BSR). Unfortunately the k8s modules (especially k8s apimachinery) are not available in the registry. This forces us to follow the "vendor" model.
I was able to workaround this by creating a build vendor with symlinks to modules declared in the go.mod files.

The vendor directory must then be added as a root in buf.yaml, but when calling 'buf generate' the path must be adjusted to only point to the kubeapps's proto directory (otherwise buf tries to generate everything instead of using them as imports).

The second main issue is that protobuf has made a backward incompatible change where now go_package cannot just be a simple path element (e.g. "v1alpha1") as is used in k8s.io protos and other protos. With the current versions we have of the k8s.io modules, the proto files do not comply with the latest validation rule and buf generate fails.
To workaround this issue, we have to downgrade by several releases the generators. It is required that the generator and runtime version match, which means the versions have to also be downgraded in go.mod, resulting in too many changes to the versions.

After the workarounds for the previous 2 issues, the next issue i ran into is that the custom json (un)marshaller that is used for Duration is somehow not recognized by the grpc gateway. It is correctly recognized and called when converting between the CRD and go code, but not when going through the gateway.
So far, I found very little documentation reagarding custom marshallers for the gateway.

Due to all those issues, and the time required to research/find workarounds, i think we should wait for our versions to catch up to the latest proto support (i believe kubernetes v0.24, we are on v0.22.9).

In the meantime, the type can be changed from int32 to string so at least the usability is improved and the wire format is set to string.
conversion from string to Duration can be done in the backend (similar to the conversion from int32 to Duration we do today)

PR #4879, opened 2022-06-08
Add check for interval in Helm Repository API

castelblanque:

### Description of the change

New repositories API adds field `interval` when adding/updating a repo.
It is not possible to map that field to any field in the existing CR `AppRepository`, and `syncJobPodTemplate` is not used anymore.

This PR adds a check for the Helm plugin to return error when `interval` has been specified.

### Benefits

Interval data posted to the API will not be swallowed and will produce an error.

### Possible drawbacks

N/A

### Applicable issues

None


antgamdia (approved):
Thanks! I've already adapted the code on the UI side :P 
BTW, I guess we were using unsigned ints, so negative numbers are not processed whatsoever, no?

castelblanque:
> I guess we were using unsigned ints, so negative numbers are not processed whatsoever, no?

I think so. `interval` field is set to `uint32`  right now, until #4869 is implemented.

Issue #4847, opened 2022-06-06
Remove `asset-svc` from our codebase + bitnami builds


antgamdia:
**Summary**

As we are shifting towards the `kubeapps-apis` component, some old components are no longer useful. Specifically, the `assetsvc` component was scaled down to 0 replicas in the latest releases.

**Background and rationale**
Now it's time to remove the old `assetsvc` code in our codebase, so that we can start pruning some dead-code, etc.

**Description**
This task requires some coordination (so that the release process doesn't get stuck) with the Bitnami team, as they have to remove this component from their system as well and follow their deprecation guidelines.

**Acceptance criteria**

- [x] The `assetsvc` component has been removed from the Kubeapps codebase: https://github.com/vmware-tanzu/kubeapps/pull/4880
- [x] There is no dead-code as a result of this removal (especially under the `/pkg` folder) https://github.com/vmware-tanzu/kubeapps/pull/4883
  - [x] Move "no longer common packages" to the directory they belong (eg. some db management to the kubeapps-apis plugins)
- [x] Bitnami has been notified about our deprecation intentions and we have acknowledged their internal task.

**Additional context**

antgamdia:
Our intentions to remove this component have been properly communicated internally. They will work independently to deprecate on their side the component.

I'll start removing the related code from our codebase soon.

antgamdia:
Quick report on the `pkg/*` usage (details below):
_The image depicts the future situation, after merging PR XXXX (TBD)_

![graphviz](https://user-images.githubusercontent.com/11535726/172735000-a80fee46-e530-4f5c-bdf2-dcf105d87103.png)

If a vertex is (or becomes) connected by a single edge, then the pkg is no longer "global" and should be placed on the call-site directory. For instance `pkg/agent` and `pkg/handlerutil` can be moved to `kubeapps-apis` once kubeops is removed.

- pkg/agent
  - cmd/kubeapps-apis/plugins/helm/packages
  - cmd/kubeapps-apis/plugins/pkg/clientgetter
  - cmd/kubeops
- pkg/chart
  - cmd/asset-syncer
  - cmd/kubeapps-apis/plugins/fluxv2/packages
  - cmd/kubeapps-apis/plugins/helm/packages
  - cmd/kubeapps-apis/plugins/pkg/pkgutils
  - cmd/kubeops
  - pkg/agent
  - pkg/chart
  - pkg/dbutils
  - pkg/handlerutil
  - pkg/helm
[...]

Issue #4785, opened 2022-05-25
Remove/migrate UI calls to Kubeops


antgamdia:
**Summary**

As part of the analysis performed for #4764, I noticed we still have several calls to the kubeops endpoints that should be properly migrated to the Resources plugin or the PackageRepositories plugin.

**Background and rationale**

Once we get rid of the Kubeops calls for managing repositories, there is no reason for us to maintain Kubeops. Therefore, any remaining call should be migrated.

**Description**

First, further analysis of what we have to delete or migrate has to be carried out.

**Acceptance criteria**

- [x] We have identified which are the calls to the Kubeops service.
  - Done at https://github.com/vmware-tanzu/kubeapps/pull/5023
- [ ] The current calls to Kubeops have been properly migrated to the corresponding services.
- [ ] Kubeapps continues to work after scaling Kubeops's deployment down to 0 replicas.

**Additional context**

Note that, first, we have to finish the implementation of the Helm-plugin Repositories API.

absoludity:
Kubeops currently has the `kubeops-ns-discovery` `ClusterRole` which gives access to list the namespaces of a cluster (before they are filtered by user RBAC). We'll need to move this cluster role and apply it to the kubeapps-apis service account so that we can use it from the resources plugin for similar functionality.

antgamdia:
For now, I have indentified: 

- https://github.com/vmware-tanzu/kubeapps/blob/main/dashboard/src/shared/Kube.ts
  - [ ] getResource
  - [ ] getAPIGroups
  - [ ] getResourceKinds
  - [ ] canI
- https://github.com/vmware-tanzu/kubeapps/blob/main/dashboard/src/shared/Namespace.ts
  - [ ] list
- https://github.com/vmware-tanzu/kubeapps/blob/main/dashboard/src/shared/Operators.ts
  - All of them, but it should be carried out at #4920

castelblanque:
> For now, I have indentified

After some investigation:
- `getResource` is not used and thus has been removed (pending PR)
- `getAPIGroups`/`getResourceKinds` are direct K8s APIs calls unrelated to Kubeops. They are only used when operators are enabled.
- `canI` will be moved into resources plugin (pending PR)

[...]

PR #4883, opened 2022-06-08
Clean and tidy up the `pkg/` dir

antgamdia:
### Description of the change

Having removed the `assetsvc` makes possible to clean up the `pkg/*` directory, moving those no longer "common" dependencies to the proper module or even deleting unused code.

### Benefits

Less unused and/or misplaced code.

### Possible drawbacks

The changes mainly consist of moving and deleting code... but who knows if my IDE search function wasn't working properly.... let's our CI decide :P

### Applicable issues

- related #4847

### Additional information

See a post-PR analysis in my comment: https://github.com/vmware-tanzu/kubeapps/issues/4847#issuecomment-1150520731


absoludity (approved):
Odd that git didn't pick up some of the moved files (instead sees them as deleted and added).

castelblanque:
Great work Antonio!
Just one question: why still using `cmd/kubeops` path name if we will get rid of it?

> Odd that git didn't pick up some of the moved files (instead sees them as deleted and added).

For Git, a file is identified not only by the name, but by the content (hashed). Content in files seems to be modified too, so for Git is a different file.

antgamdia:
> Just one question: why still using cmd/kubeops path name if we will get rid of it?

Currently, `kubeops` is a necessary service as it is running the -current- API for repositories, as well as performing some requests to k8s (which should be migrated to the Resources plugin). There is an open issue for that: #4785 
This removal, consequently, is still assuming `kubeops` is still there but, yep, another removal like this one ought to be performed once we start sunsetting kubeops.

absoludity:
> > Odd that git didn't pick up some of the moved files (instead sees them as deleted and added).
> 
> For Git, a file is identified not only by the name, but by the content (hashed). Content in files seems to be modified too, so for Git is a different file.

I'm pretty sure Git does more than just compare the hash to see if an added and removed file match - it applies some other heuristics which are usually pretty good at determining moved files. But if there are a lot of moved files with lots of changes in the files (more than 50%) the heuristic is less successful, from what I understand. See https://vjeko.com/2020/11/24/understanding-renaming-moving-files-with-git/ for some examples. Interestingly, their recommendation of renaming/content change in different commits doesn't work when we squash commits (because the separate commits won't exist in the main branch).

PR #4852, opened 2022-06-06
Update triage and prioritization processes

ppbaena:
### Description of the change

Update documentation and issue templates to update and standardize the issue creation, triage and prioritization processes.

### Benefits

Standard way to manage issues from maintainers. Information clear and updated for contributors.

### Possible drawbacks

Need for update the stale process.

PR #4890, opened 2022-06-09
Update triage doc and stale rules to be consistent

ppbaena:
### Description of the change

After updating triage document in #4852, it was necessary to update some labels and stalebot rules to be consistent with the final process described in that document.

### Benefits

Stale bot configured according to Kubeapps triage and prioritization rules. Keep maintainers team alert from proposals and questions without activity, to avoid leaving them without a reply.

### Possible drawbacks

Need to be agile in some decisions about the proposals we receive. Depending on this, we should improve the process to define new features in Kubeapps, to avoid having a big bunch of them in the backlog without activity.


castelblanque:
Great, thanks Pepe.
There is a conflict to be solved with `issue-triage-process.md`.

castelblanque (approved):

PR #4895, opened 2022-06-09
Add h8lio to ADOPTERS.md file

ppbaena:
### Description of the change

Update ADOPTERS.md file with a new success story, this time from [h8l.io](https://h8l.io/)

### Benefits

New Kubeapps adopters and success story to share with our community

### Possible drawbacks

N/A


absoludity (approved):
Great, thanks Pepe. Might be worth explicitly specifying a background for the icons with transparent backgrounds. This is what I see if I switch GH to use darkmode:

adopters-darkmode

ADOPTERS.md
```diff
@@ -4,15 +4,15 @@ If you're using Kubeapps and want to add your organization to this list, follow
 
 ## Organizations using Kubeapps
 
-<a href="https://www.coreweave.com/" border="0" target="_blank"><img alt="CoreWeave" src="./site/content/docs/latest/img/adopters/coreweave.svg" height="90"></a>
+<a href="https://www.coreweave.com/" border="0" target="_blank"><img alt="CoreWeave" src="./site/content/docs/latest/img/adopters/coreweave.svg" height="90" style="background-color: #FFF;"></a>
```
antgamdia, line 7: Just FTR, Github recently supported conditional images based on the selected theme; for example: https://www.stefanjudis.com/notes/how-to-define-dark-light-mode-images-in-github-markdown/

Source: https://github.blog/changelog/2021-11-24-specify-theme-context-for-images-in-markdown/

ppbaena: Thanks for the heads-up @antgamdia. I think we can keep it simple by adding the background-color to the image, so we can improve visualization in dark mode (instead of maintaining two files for each adopter logo). 

Anyway, open to listen to alternatives that makes the contribution and maintenance as simple as possible.

absoludity: fwiw, GH does not allow you to add custom style to an image (or any element).

If I check the result now, although I see the new `background-color` in a style attribute if I check the raw file, this is what I see (note the debugger showing the element):

background-not-fff

You can read a [relevant answer on stackoverflow](https://stackoverflow.com/a/60340266), but it comes down to stripping such attributes due to security issues. So we might need the theme-based icons (might be best to update to request both light and dark versions if available, but we can do them if necessary).

On another note, I expected that we'd not be specifying the `<a>` and `` tags here, but rather than the markdown equiv of `[name](link to your URL)`, but maybe that looks more complicated :)

ppbaena:
I have updated the logos in #5158 replacing the .svg file for a .jpg one with withe background so we avoid problems in dark theme.

PR #4899, opened 2022-06-10
Add expired value cache.

absoludity:

### Description of the change

Adds an implementation of the [cached trait](https://docs.rs/cached/latest/cached/trait.Cached.html) that allows a cached value to determine if it has expired or not. If it has expired when fetched from the cache, it is removed.

### Benefits

We can use this for caching the pinniped `ClusterCredential` rather than fetching it for every user request.

### Applicable issues

- ref #2875

### DIFF
--- a/cmd/pinniped-proxy/src/expired_value_cache.rs
+++ b/cmd/pinniped-proxy/src/expired_value_cache.rs
@@ -5,19 +5,19 @@ use std::hash::Hash;
 
 use cached::Cached;
 
-pub trait Expired {
-    fn has_expired(&self) -> bool;
+pub trait CanExpire {
+    fn is_expired(&self) -> bool;
 }
 
 // Implement our ExpiredValueCache which uses the returned values own
 // trait to determine whether it has expired, rather than a set timestamp.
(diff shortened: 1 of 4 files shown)


cmd/pinniped-proxy/src/expired_value_cache.rs
```diff
@@ -0,0 +1,216 @@
...
+                    return None;
+                }
+            },
+            None => {
+                self.misses += 1;
+                return None
+            }
+        }
+        self.hits += 1;
+        self.store.get_mut(k)
```
castelblanque, line 77: It seems that the only difference between this function and the `cache_get` above is this line.
Why not refactoring the common part out so that both `cache_get` and `cache_get_mut` use it?

castelblanque: I'm not into Rust (for now!).
If there is no straight forward way of refactoring it, I don't think it is worth investing the effort.
Thanks!

castelblanque (approved):

antgamdia (approved):
Great, thanks for these laser-focused PRs when adding code here!

PR #4900, opened 2022-06-10
Ignore hash tests until I sort out why they occasionally fail with a set other hash

absoludity:

### Description of the change

While running other tests, I'm occasionally seeing these hash tests fail, generating one specific other hash - which is quite odd. I have not yet determined what is causing the difference in input, though it is obviously always the same difference.

I'm just creating this and leaving in draft in case people start seeing these fail in CI. I've not seen any, but if so, land this :)

### Applicable issues

- fixes #


antgamdia:
It seems that, fortunately, the tests aren't failing, are they? Can we close this PR then?

absoludity:
Yeah - it was odd. I saw them fail multiple times locally without a change (that I was aware of), but they always failed with a specific other hash. Odd, but yes, closing.

antgamdia:
I'm noticing several random failures related to those tests since we upgraded some cargo deps :(
Don't know what's happening, though... but we might want to merge this PR 😅

castelblanque:
> Don't know what's happening, though... but we might want to merge this PR

+1 100%!!
The Rust hash test is failing randomly. Let's open an issue to tackle the disabled tests in the future?

antgamdia (approved):
Approving and merging

PR #4909, opened 2022-06-13
Disable E2E tests for forked PRs

### LINKED ISSUE #4832: Operators e2e test randomly failing

castelblanque:
### Description of the change

PRs from forked Git repositories don't have access to environment variables/secrets from CI.
That would be leaking information to whoever created a forked PR.

In order to avoid having tests failing, E2E have been disabled for forked PRs with this change.

An alternative path to take for solving this issue would be to selectively choose which E2E tests to run depending on the available variables (e.g. if DOCKER_PASSWORD is available -> run the E2E test that makes use of it).
By now we would be able to do so, as the two secret variables (`DOCKER_*`) are only used in `03-create-private-package-repository.spec.js`.

### Benefits

Developers from forked repos won't get E2E tests failing.

### Possible drawbacks

Developers from forked repos won't get E2E tests running.

### Applicable issues

- fixes #4832

### Additional information

This is the part that the forked PRs don't get when running CI:
image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -150,6 +150,10 @@ workflows:
       - build_pinniped_proxy:
           <<: *build_always
       - local_e2e_tests:
+          filters:
+            branches:
+              # Forked pull requests have CIRCLE_BRANCH set to pull/XXX
+              ignore: /pull\/[0-9]+/
           <<: *build_always
           matrix:
             parameters:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -712,6 +712,10 @@ jobs:
     # Reference for available machines
     # https://circleci.com/docs/2.0/configuration-reference/#machine-executor-linux
     resource_class: large
+    filters:
+      branches:
+        # Forked pull requests have CIRCLE_BRANCH set to pull/XXX
+        ignore: /pull\/[0-9]+/
     environment:
       DEFAULT_DEX_IP: "172.18.0.2"
       TEST_UPGRADE: "1"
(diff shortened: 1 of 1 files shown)


castelblanque:
Happy to take the alternative path if you think that disabling E2E tests for forked repos is too detrimental.

antgamdia (commented):
I'm not sure we should disable the e2e tests in forked repos... it would simply stop us from accepting contributions confidently (ie, ack-ing the changes don't break anything).

The question, IMHO, is why the e2e test requires such a private variable. As far as I'm concerned, the Dockerhub password/username is required for:

````
  await page.fill("input#kubeapps-docker-cred-username", process.env.DOCKER_USERNAME);
  await page.fill("input#kubeapps-docker-cred-password", process.env.DOCKER_PASSWORD);
````

However, this is for creating the img pull secret for the chart we are pushing (in CI) to the Chartmusseum instance... so if we do control the publication of the chart (it's part of the CI process)... why would we need such a Docker (production) credentials ??

dlaloue-vmware:
I am not sure either we should disable the e2e tests for forked branches.
I am not sure when they started failing for me, but i think we used to runt he e2e tests before with no issues.

if the only issues we have is the timeout issue, we can fix it another way.

Issue #4910, opened 2022-06-13
Add E2E test checking Docker credentials against a local Docker registry.


castelblanque:
**Summary**
Create a dedicated E2E test that checks Docker credentials against a local Docker registry when creating a new package repository.

**Background and rationale**
Current E2E test `03-create-private-package-repository.spec.js` makes use of `DOCKER_USERNAME` and `DOCKER_PASSWORD` from environment variables. The repo that is added in the test points to ChartMuseum with a chart that does not actually require Docker credentials for pulling the images.

```js
  await page.click('.docker-creds-subform-button button:has-text("Add new credentials")');
  await page.fill("input#kubeapps-docker-cred-secret-name", secretName);
  await page.fill("input#kubeapps-docker-cred-server", "https://index.docker.io/v1/");
  await page.fill("input#kubeapps-docker-cred-username", process.env.DOCKER_USERNAME);
  await page.fill("input#kubeapps-docker-cred-password", process.env.DOCKER_PASSWORD);
  await page.click('.docker-creds-subform button:has-text("Submit")');
```

**Description**
Create a new E2E test that adds a repo with a chart using a container image that requires Docker credentials from a local registry.
As a positive effect, we will be able to succesfully run E2E tests in forked PRs.

**Acceptance criteria**
- New E2E test that checks Docker credentials against local registry.
- Remove the above piece of code from the `03-create-private-package-repository.spec.js` test.

PR #4927, opened 2022-06-15
E2E checking Docker registry credentials in package repo

### LINKED ISSUE #4832: Operators e2e test randomly failing

castelblanque:
### Description of the change

Until now, E2E test '03-create-private-package-repository.spec.js' was testing two things:
- Deploying a chart hosted in a repo with authentication
- Adding Docker credentials and making an API request to check that `imagePullSecrets` was being filled in the CRD

Apart from that, E2E tests where failing when running from forked PRs. This was partially solved by @dlaloue-vmware with reorganizing the E2E script parameters.

This PR separates the 03 test into two separate ones, introducing the `10-deploy-package-with-private-image.spec.js`.
The test makes use of a cluster local installed Docker registry with authentication.

### Benefits

- Better granularity of tests: one for app repo auth, one for Docker credentials
- Docker credentials for a repo checked with a real deployment
- No more `DOCKER_USERNAME` and `DOCKER_PASSWORD` needed from the CI environment
- The above brings the ability for forked PRs to run E2E without needing CI context data

### Possible drawbacks

Creation of local authenticated Docker registry brings some scripts using port-forward, which could introduce some instability when running.

### Applicable issues

- implements #4910
- fixes #4832


.circleci/config.yml
```diff
@@ -501,7 +502,7 @@ run_e2e_tests: &run_e2e_tests
         DEV_TAG=${latest/v/}
         IMG_MODIFIER=""
       fi
-      if ./script/e2e-test.sh $USE_MULTICLUSTER_OIDC_ENV $OLM_VERSION $DEV_TAG $IMG_MODIFIER $TEST_TIMEOUT_MINUTES $DEFAULT_DEX_IP $ADDITIONAL_CLUSTER_IP $DOCKER_USERNAME $DOCKER_PASSWORD $KAPP_CONTROLLER_VERSION; then
```
castelblanque, line 504: Getting rid of this CI context variables

integration/charts/simplechart/values.yaml
```diff
@@ -0,0 +1,16 @@
+image:
+  repository: local-docker-registry:5600/nginx
```
castelblanque, line 2: Adding a test chart that makes use of an authenticated container image

integration/registry/local-registry.yaml
```diff
@@ -0,0 +1,70 @@
+---
+apiVersion: v1
+kind: Secret
+type: Opaque
+metadata:
+  name: registry-auth
+  namespace: ci
+data:
+  htpasswd: dGVzdHVzZXI6JDJ5JDA1JHVLUnAvdUV1aEMxRXpmbzQ1a1NCbk96eUlnV2tKNmk3R1VKTzNhLjVvY0JrZUNBNTZjRWxXCgo=
```
castelblanque, line 9: Credentials for Docker registry

integration/tests/main/03-create-private-package-repository.spec.js
```diff
@@ -85,31 +74,6 @@ test("Create a new private package repository successfully", async ({ page }) =>
...
-      // Axios doesn't provide streaming responses, so splitting on new line works
-      // but gives us a string, not JSON, and may leave a blank line at the end.
-      const response = JSON.parse(r);
-      const resourceRef = response.result?.resourceRef;
-      if (resourceRef.kind === "Deployment" && resourceRef.name.match(appName)) {
-        deployment = JSON.parse(response.result?.manifest);
-      }
-    });
-
-  expect(deployment?.spec?.template?.spec?.imagePullSecrets).toEqual([{ name: secretName }]);
```
castelblanque, line 111: We no longer need this. Test '10' actually checks that the chart is correctly deployed using a container image from the authenticated Docker registry.

script/install-local-registry.sh
```diff
@@ -0,0 +1,84 @@
...
+
+pushContainerToLocalRegistry() {
+    DOCKER_REGISTRY="$DOCKER_REGISTRY_HOST:$DOCKER_REGISTRY_PORT"
+
+    echo "127.0.0.1  $DOCKER_REGISTRY_HOST" | sudo tee -a /etc/hosts
+
+    docker pull nginx
+    docker tag nginx $DOCKER_REGISTRY/nginx
+
+    /bin/sh -c "kubectl -n ${REGISTRY_NS} port-forward service/docker-registry 5600:5600 &"
```
castelblanque, line 50: Solution far from elegant, but it does the trick for pushing from CI host into Docker registry inside cluster.

script/install-local-registry.sh
```diff
@@ -0,0 +1,84 @@
...
+
+    kubectl create namespace $REGISTRY_NS
+
+    # Add our CA to the node
+    docker cp "$(mkcert -CAROOT)"/rootCA.pem "$CONTROL_PLANE_CONTAINER:/usr/share/ca-certificates/local-rootCA.pem"
+    docker exec --user root $CONTROL_PLANE_CONTAINER sh -c "echo 'local-rootCA.pem' >> /etc/ca-certificates.conf"
+    docker exec --user root $CONTROL_PLANE_CONTAINER update-ca-certificates -f
+
+    # Restart containerd to make the CA addition effective
+    docker exec --user root $CONTROL_PLANE_CONTAINER systemctl restart containerd
```
castelblanque, line 22: Key step in order to get CA certificates refreshed for Kubelet.

integration/registry/local-registry.yaml
```diff
@@ -0,0 +1,70 @@
...
+    spec:
+      volumes:
+      - name: auth-vol
+        secret:
+          secretName: registry-auth
+      - name: certs-vol
+        secret:
+          secretName: registry-tls
+      containers:
+        - image: registry:2
```
castelblanque, line 36: @antgamdia I guess this new dependency needs to be added to OSM.

script/install-local-registry.sh
absoludity: Nice - I bet that one line cost a bit... sometimes I wish it were possible to show how much one line costs after feeling like I've been banging my head against a wall :)

integration/registry/local-registry.yaml
absoludity: Or another thought: I see below that you create the `registry-tls` secret as part of the script to setup the registry: why not also create this simple opaque secret there so that we only have the one location where the secret is defined and used? Just a thought, fine either way.

script/install-local-registry.sh
```diff
@@ -0,0 +1,84 @@
...
+
+    # Generate new certificate for registry
+    mkcert -key-file $PROJECT_PATH/devel/localhost-key.pem -cert-file $PROJECT_PATH/devel/docker-registry-cert.pem $DOCKER_REGISTRY_HOST "docker-registry.$REGISTRY_NS.svc.cluster.local"
+    kubectl -n $REGISTRY_NS delete secret registry-tls --ignore-not-found=true
+    kubectl -n $REGISTRY_NS create secret tls registry-tls --key $PROJECT_PATH/devel/localhost-key.pem --cert $PROJECT_PATH/devel/docker-registry-cert.pem
```
absoludity, line 27: As above, I wonder if it's worth creating the `registry-auth` here as well so you don't need to repeat the encoded secret in the yaml.

script/install-local-registry.sh
absoludity: setting up a port-forward temporarily is OK I think, we do the same for chartmuseum. But the sleeps are more worrying in my opinion (as the time required is variable). It'd be nice if we could instead wait until the service is available (with a simple curl or something).

absoludity (approved):
Great, thanks for this Rafa... much nicer than just checking the actual secret is included in the yaml.

script/install-local-registry.sh
```diff
@@ -0,0 +1,84 @@
+#!/bin/bash
```
antgamdia, line 1: +1 to Michael's comment, plus adding the copyright header :P

```suggestion
#!/bin/bash

# Copyright 2022 the Kubeapps contributors.
# SPDX-License-Identifier: Apache-2.0
```

script/install-local-registry.sh
antgamdia: Woah, that's totally true... it seems a tricky one.

I thought this change in the CA could be done as a kubeadm config yaml during the cluster creation in kind (below), but you'll probably have explored many paths and this one is the simplest one for our use case, I guess, so +1 and thanks for coming up with it :P 

````
apiVersion: kubeadm.k8s.io/v1beta3
kind: ClusterConfiguration
controllerManager:
  extraArgs:
    cluster-signing-cert-file: /etc/kubernetes/pki/ca.crt
    cluster-signing-key-file: /etc/kubernetes/pki/ca.key
````

**Review by antgamdia — APPROVED** [review]
Awesome, thanks! That's great we are finally testing repos with auth , it's gonna be a key aspect for properly testing the new repos API!

As I said in a comment, please do not merge the PR until we trigger the release, so that it does not interfere with the already reported deps to OSM.

File: integration/registry/local-registry.yaml (hunk 3 above)
Comment (castelblanque) [reply integration/registry/local-registry.yaml:9]: `registry-tls` is created from the script as it needs some externally generated input (the key file). I prefer having everything possible put in a static yaml file, that is why `registry-auth` is there, it does not depend on anything.
But agree with your suggestion, I'll add a comment.

File: integration/charts/simplechart/templates/tests/test-connection.yaml [hunk 10]
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: "{{ include "simplechart.fullname" . }}-test-connection"
+  annotations:
+    "helm.sh/hook": test
```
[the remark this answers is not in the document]
castelblanque: Yep, fine with me.

integration/charts/simplechart/values.yaml
```diff
@@ -0,0 +1,16 @@
+image:
```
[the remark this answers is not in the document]
castelblanque: Adding the header, but wondering why in the Kubeapps chart Yaml files there are no headers like this?
E.g. [values file](https://github.com/vmware-tanzu/kubeapps/blob/main/chart/kubeapps/values.yaml)

script/install-local-registry.sh
castelblanque: | I bet that one line cost a bit.

Finding that out unlocked the whole thing to me after banging my head against the wall.

| you'll probably have explored many paths

Indeed, tried many ways to get it working, including the use of `/etc/kubernetes/pki/ca*` for generating the new certificate. Didn't work.
It seemed to me that the issue is when the "kubelet" establishes a TLS connection to contact the image repository. Acting as a client, it uses a set of CAs available form the underlying OS in which the `/etc/kubernetes/pki/ca.crt`, for some reason, is not included. Therefore, I needed to add a new CA to sign the certificate.

Issue #3889, opened 2021-12-02
Add "filter by plugin" in the catalog view


antgamdia:
As new plugins are starting to become available, there is an increasing need of having a way to filter them through the Catalog view.
I can think of two different approaches:

1. Doing the same thing as we already do for retrieving the Operator providers, that is, push each new provider to an array of filters we can select.

2. Or following the same 'static' approach for retrieving the list of categories (ie, the backend provides the list of categories and we just use it). Perhaps a single call to `GetConfiguredPlugins` is enough. 

As usual, the decision is based on how we want the filters to work: they show everything available, although a combination of filters yields an empty result, or we only display filters that make sense for the current view.

To keep the alignment with the current design, I'd go with 1)

PR #4930, opened 2022-06-16
Remove no longer used elements from AppRepoList 

antgamdia:
### Description of the change

This PR is one of the "repos API revamp" series. In this one, we simply remove what is no longer used in the upcoming move toward the new API.
More details inline.

Note this PR is targeting a temporary branch for the UI, so merging it won't make it be included in main's.

### Benefits

No actual feature-wise benefits, but less code to review later on in follow-up PRs.

### Possible drawbacks

We lose the "force resync" feature, but it is no longer available in the new Repos API.

### Applicable issues

- related #4764

### Additional information

image


dashboard/src/components/Config/AppRepoList/AppRepoButton.tsx
```diff
@@ -48,9 +51,9 @@ export function AppRepoAddButton({
     ociRepositories: string[],
     skipTLS: boolean,
     passCredentials: boolean,
-    filter?: IAppRepositoryFilter,
+    filter?: IPkgRepositoryFilter,
   ) => {
-    if (repo) {
+    if (packageRepoRef) {
```
antgamdia, line 56: Just a couple of renames

dashboard/src/components/Config/AppRepoList/AppRepoList.tsx
```diff
@@ -94,27 +95,56 @@ function AppRepoList() {
...
+      : namespacedRepos.push(repo);
   });
 
   const tableColumns = [
     { accessor: "name", Header: "Name" },
     { accessor: "url", Header: "URL" },
+    { accessor: "plugin", Header: "Package Type" },
     { accessor: "accessLevel", Header: "Access Level" },
     { accessor: "namespace", Header: "Namespace" },
+    { accessor: "status", Header: "Status" },
```
antgamdia, line 111: Adding more columns to the table but with hardcoded values for now.

dashboard/src/components/Config/AppRepoList/AppRepoList.tsx
```diff
@@ -94,27 +95,56 @@ function AppRepoList() {
...
         name: getRepoNameLinkAndTooltip(cluster, repo),
         url: repo.spec?.url,
-        accessLevel: repo.spec?.auth?.header ? "Private" : "Public",
+        // TODO(agamez): the PackageRepositorySummary API doesn't expose this field. It will be added in upcoming PRs; in the meantime, set to "unknown"
+        // accessLevel: repo.type?.auth?.header ? "Private" : "Public",
+        accessLevel: "unknown",
         namespace: repo.metadata.namespace,
+        plugin: "Helm",
+        // eslint-disable-next-line no-constant-condition
+        status: true ? (
```
antgamdia, line 125: Leaving this change here, but will be modified and tested in the next PRs

dashboard/src/components/Config/AppRepoList/AppRepoList.tsx
```diff
@@ -155,80 +184,83 @@ function AppRepoList() {
...
-                ) : (
-                  <p>
-                    There are no <i>global</i> Package Repositories yet. Click on the "Add Package
-                    Repository" button to create one.
-                  </p>
-                )}
-                {namespace !== globalReposNamespace && (
-                  <>
-                    <h3>Namespaced Repositories: {namespace}</h3>
+      <div className="catalog-container">
```
antgamdia, line 187: Fixing a css issue: leaving more space at the bottom.

image

dashboard/src/shared/url.ts
```diff
@@ -130,9 +131,6 @@ export const api = {
   // access the k8s api server directly).
   k8s: {
     base: (cluster: string) => `api/clusters/${cluster}`,
-    namespaces: (cluster: string) => `${api.k8s.base(cluster)}/api/v1/namespaces`,
-    namespace: (cluster: string, namespace: string) =>
-      namespace ? `${api.k8s.namespaces(cluster)}/${namespace}` : `${api.k8s.base(cluster)}/api/v1`,
```
antgamdia, line 135: Unused, but looking forward to removing kubeops and operators to clean up the whole file 😝

dashboard/src/shared/utils.test.ts
```diff
@@ -0,0 +1,158 @@
+// Copyright 2022 the Kubeapps contributors.
```
antgamdia, line 1: Adding a test for the whole utils.ts file as we didn't have any yet.

dashboard/src/shared/utils.ts
```diff
@@ -137,6 +136,32 @@ export function getPluginPackageName(plugin?: Plugin | PluginNames | string) {
   }
 }
 
+// FIXME(agamez): replace with a proper call to the plugins server (see getPluginsServiceClientImpl)
```
antgamdia, line 139: This is a workaround for setting the `Plugin` object later on. However, we still need a way to get (from the UI) the current plugins (closely related to #3889 )

dashboard/src/components/Config/AppRepoList/AppRepoList.tsx
absoludity: Sometimes what I do in this situation, where I'm not targetting the main branch, is that I actually write the test to fail so I can't forget to fix it in the follow-up branch.

dashboard/src/components/Config/AppRepoList/AppRepoList.tsx
absoludity: Great, thanks.

dashboard/src/shared/utils.test.ts
absoludity: Excellent, that'll encourage more tests :)

absoludity (approved):

PR #4939, opened 2022-06-20
Revamp `AppRepoForm`

antgamdia:
### Description of the change

This incremental PR brings the revamp on the AppRepoForm but still uses the old API. This means that the current PR does not allow using Flux or Carvel repos yet. This way, the effort related to the tests is scoped just to ensure the existing test suites continue to work as well as adding new test cases eventually for the new UI behavior.

### Benefits

The UI is more consistent with what we want to achieve mid-term and adding/fixing test cases is easier.

### Possible drawbacks

This PR hasn't gotten properly tested IRL, but it doesn't really matter, as it is just a transient change that will get superseded in the upcoming PRs (using the new Repos API).

### Applicable issues

- related #4764

### Additional information

image

image


dashboard/src/components/Config/AppRepoList/AppRepoForm.tsx
```diff
@@ -9,155 +15,142 @@ import { CdsRadio, CdsRadioGroup } from "@cds/react/radio";
...
         setFilterNames(names);
       }
-
-      if (repo.spec?.auth?.customCA || repo.spec?.auth?.header) {
-        fetchRepoSecret(currentCluster, repo.metadata.namespace, repo.metadata.name);
-      }
     }
   }, [repo, namespace, currentCluster, dispatch]);
 
-  async function fetchRepoSecret(cluster: string, repoNamespace: string, repoName: string) {
```
antgamdia, line 132: Removing any secret-related logic.

castelblanque:
Awesome step towards the new repositories UI.
Many thanks!

E2E tests will be fixed in the target branch before creating PR for `main`, right?

PR #4951, opened 2022-06-24
Add TLS support to the pinniped-proxy directly.

absoludity:

### Description of the change

After playing around with different options, this appears to be the simplest way to add tls support to pinniped-proxy without breaking the existing non-tls setup.

I'm leaving this as a draft until I figure out how TLS on the related service will interact with this. After that, it should just be updating the api server to be able to trust the cert, then the related chart changes.

### Applicable issues

- fixes #

### DIFF
--- a/cmd/pinniped-proxy/src/main.rs
+++ b/cmd/pinniped-proxy/src/main.rs
@@ -39,6 +39,7 @@ async fn main() -> Result<()> {
     // Run the server for ever. If it returns with an error, return the
     // result, otherwise, if it completes, we return Ok.
     if with_tls {
+        info!("Configuring with TLS cert {} and key {}", opt.proxy_tls_cert, opt.proxy_tls_cert_key);
         // For every incoming connection, we make a new hyper `Service` to handle
         // all incoming HTTP requests on that connection. This is done by passing a
         // closure to the hyper `make_service_fn` which returns our custom `make_svc`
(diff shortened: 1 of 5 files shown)


cmd/pinniped-proxy/src/main.rs
```diff
@@ -28,34 +32,58 @@ async fn main() -> Result<()> {
...
-            }))
-        }
-    });
-
     let addr = ([0, 0, 0, 0], opt.port).into();
 
-    let server = Server::bind(&addr).serve(make_svc);
-
-    info!("Listening on http://{}", addr);
+    let with_tls = opt.proxy_tls_cert != "" && opt.proxy_tls_cert_key != "";
```
antgamdia, line 37: Maybe worth adding a warning if just one of them is set ?

antgamdia (approved):
Great! Thank for the explanation!

cmd/pinniped-proxy/src/main.rs
```diff
@@ -28,34 +32,58 @@ async fn main() -> Result<()> {
...
 
     // Run the server for ever. If it returns with an error, return the
     // result, otherwise, if it completes, we return Ok.
-    server.await?;
+    if with_tls {
+        info!("Configuring with TLS cert {} and key {}", opt.proxy_tls_cert, opt.proxy_tls_cert_key);
```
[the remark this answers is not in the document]
absoludity: Nope, they're just file paths. Not so clear from the option names...

cmd/pinniped-proxy/src/main.rs
absoludity: Even erroring and stopping - we can't run with just one.

PR #4954, opened 2022-06-24
Switch to the new Repos API

antgamdia:
### Description of the change

This PR finally starts using the new Repos API backend to perform the actions in the new repos UI.
Besides, this PR is pretty similar to https://github.com/vmware-tanzu/kubeapps/pull/4817, which I'll close soon, but in this one, all the test suites have been properly modified (mainly fixing the existing test cases, but eventually adding some extra ones).

### Benefits

We will finally use the new API for managing repos!

### Possible drawbacks

As it is a quite big UI change, eventual bugs are expected.

### Applicable issues

- related #4764 (phase 0)

### Additional information

Pending here: 

- Fix the E2E tests after merging other PRs.

Still pending, but for yet another PR:

- Cleanup + several renames: https://github.com/vmware-tanzu/kubeapps/pull/4955


dashboard/src/components/Config/PkgRepoList/PkgRepoList.tsx
```diff
@@ -131,24 +130,24 @@ function AppRepoList() {
...
         ),
         actions: disableControls ? (
-          <AppRepoDisabledControl />
+          <PkgRepoDisabledControl />
         ) : (
-          <AppRepoControl
+          <PkgRepoControl
             repo={repo}
             refetchRepos={refetchRepos}
             kubeappsNamespace={globalReposNamespace}
```
castelblanque, line 153:
```suggestion
            globalReposNamespace={globalReposNamespace}
```

dashboard/src/components/Config/PkgRepoList/PkgRepoList.tsx
```diff
@@ -94,35 +91,37 @@ function AppRepoList() {
...
     { accessor: "url", Header: "URL" },
-    { accessor: "plugin", Header: "Package Type" },
+    { accessor: "packageFormat", Header: "Package Format" },
     { accessor: "accessLevel", Header: "Access Level" },
     { accessor: "namespace", Header: "Namespace" },
     { accessor: "status", Header: "Status" },
     { accessor: "actions", Header: "Actions" },
   ];
-  const getTableData = (targetRepos: IAppRepository[], disableControls: boolean) => {
+  const getTableData = (targetRepos: PackageRepositorySummary[], disableControls: boolean) => {
```
castelblanque, line 114: At some point we discussed the alternative of having a single table for all repositories, differentiating the global ones with some distinctive icon (e.g. world icon). Is this something that could be done with the new repos API? (not for this PR/issue of course).

dashboard/src/components/Catalog/Catalog.tsx
```diff
@@ -212,9 +216,9 @@ export default function Catalog() {
   // We do not currently support package repositories on additional clusters.
   const supportedCluster = cluster === kubeappsCluster;
   useEffect(() => {
-    if (!supportedCluster || namespace === globalReposNamespace) {
-      // Global namespace or other cluster, show global repos only
-      dispatch(actions.repos.fetchRepos(globalReposNamespace));
+    if (!namespace || !supportedCluster || namespace === globalReposNamespace) {
+      // All Namespaces. Global namespace or other cluster, show global repos only
+      dispatch(actions.repos.fetchRepos(""));
```
castelblanque, line 221:
```suggestion
      dispatch(actions.repos.fetchRepos(globalReposNamespace));
```

If I'm not mistaken the result will be the same, but isn't it easier to read that the action is retrieving global repos?

dashboard/src/components/Config/PkgRepoList/PkgRepoForm.tsx
```diff
@@ -0,0 +1,1701 @@
...
+                        target="_blank"
+                        rel="noopener noreferrer"
+                      >
+                        Golang duration
+                      </a>
+                      ) to wait between synchronizing the repository.
+                    </CdsControlMessage>
+                  </CdsInput>
+                )}
+                {plugin?.name === PluginNames.PACKAGES_HELM && (
```
castelblanque, line 1563: Looking forward to the Plugins UI customization to be implemented to avoid this kind of condition blocks!

dashboard/src/components/Config/PkgRepoList/PkgRepoForm.tsx
```diff
@@ -0,0 +1,1701 @@
...
+                    />
+                  </CdsCheckbox>
+                )}
+
+                <CdsToggleGroup>
+                  <CdsToggle>
+                    <label>
+                      {isUserManagedCASecret
+                        ? "Use user-managed secrets"
+                        : "Use Kubeapps-managed secrets"}
```
castelblanque, line 1584: Just throwing the idea: what about using more friendly label `I will manage the secrets`, `Let Kubeapps manage the secrets`, or similar?

dashboard/src/components/Config/PkgRepoList/PkgRepoForm.tsx
```diff
@@ -0,0 +1,1701 @@
...
+          type="text"
+          placeholder="my-secret-name"
+          value={secretAuthName}
+          onChange={handleSecretAuthNameChange}
+          required={
+            isUserManagedSecret &&
+            authMethod !==
+              PackageRepositoryAuth_PackageRepositoryAuthType.PACKAGE_REPOSITORY_AUTH_TYPE_UNSPECIFIED
+          }
+          pattern="[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*"
```
castelblanque, line 482: This regex is used in four places, maybe having a constant for it?

dashboard/src/components/Config/PkgRepoList/PkgRepoForm.tsx
```diff
@@ -0,0 +1,1701 @@
...
+                        name="auth"
+                        value={
+                          PackageRepositoryAuth_PackageRepositoryAuthType[
+                            PackageRepositoryAuth_PackageRepositoryAuthType
+                              .PACKAGE_REPOSITORY_AUTH_TYPE_BEARER
+                          ]
+                        }
+                        checked={
+                          authMethod ===
+                          PackageRepositoryAuth_PackageRepositoryAuthType.PACKAGE_REPOSITORY_AUTH_TYPE_BEARER
```
castelblanque, line 785: Not all backend implementations support all auth types defined in the API.
For example, Flux [does not implement some](https://github.com/vmware-tanzu/kubeapps/blob/main/cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo.go#L1124), or Helm does not implement the SSH for GIt like Carvel.
Maybe we should add some conditions for that in the frontend? Otherwise the user will get an `Unimplemented` error.
This is one of the things that the Plugins customizable UI feature should solve in the future.

dashboard/src/components/Config/PkgRepoList/PkgRepoForm.tsx
```diff
@@ -0,0 +1,1701 @@
...
+                    <div cds-layout="col@xs:8">
+                      {/* Begin docker creds authentication */}
+                      <div
+                        id="kubeapps-repo-imagePullSecrets-details-docker"
+                        hidden={
+                          helmPSAuthMethod !==
+                          PackageRepositoryAuth_PackageRepositoryAuthType.PACKAGE_REPOSITORY_AUTH_TYPE_DOCKER_CONFIG_JSON
+                        }
+                      >
+                        {/* TODO(agamez): enable the selection once the API supports it */}
```
castelblanque, line 1324:
```suggestion
                        {/* TODO(agamez): enable the selection once the API supports it (#5015) */}
```

dashboard/src/components/Config/PkgRepoList/PkgRepoForm.tsx
```diff
@@ -0,0 +1,1701 @@
...
+                              checked={isUserManagedPSSecret}
+                              disabled={
+                                !!(repo?.customDetail as Partial<RepositoryCustomDetails>)
+                                  ?.dockerRegistrySecrets?.length
+                              }
+                            />
+                          </CdsToggle>
+                        </CdsToggleGroup>
+                        <br /> */}
+                        {isUserManagedPSSecret ? (
```
castelblanque, line 1344: This involves having a separate toggle for the user to choose? Maybe we are adding to the cognitive load of users (two choices instead of one).
Why not checking the general `isUserManagedSecret` toggle/flag?

dashboard/src/components/Config/PkgRepoList/PkgRepoForm.tsx
```diff
@@ -0,0 +1,1701 @@
...
+            <CdsAccordionHeader onClick={() => toggleAccordion(3)}>Advanced</CdsAccordionHeader>
+            <CdsAccordionContent>
+              <CdsFormGroup layout="vertical">
+                {plugin?.name !== PluginNames.PACKAGES_HELM && (
+                  <CdsInput>
+                    <label htmlFor="kubeapps-repo-interval">Synchronization Interval</label>
+                    <input
+                      id="kubeapps-repo-interval"
+                      type="text"
+                      placeholder="10m"
```
castelblanque, line 1546: Is this the same value defined in line 126?

dashboard/src/shared/PackageRepositoriesService.ts
```diff
@@ -0,0 +1,252 @@
...
+      case PluginNames.PACKAGES_HELM:
+        return await this.helmRepositoriesClient().AddPackageRepository(
+          addPackageRepositoryRequest,
+        );
+      default:
+        return await this.coreRepositoriesClient().AddPackageRepository(
+          addPackageRepositoryRequest,
+        );
+    }
+  }
```
castelblanque, line 81: Why having separate calls if the request object has the field `customDetail` is present in `AddPackageRepositoryRequest` and should be processed by the core accordingly?

dashboard/src/components/Config/PkgRepoList/PkgRepoForm.tsx
```diff
@@ -0,0 +1,1701 @@
...
+    isInstallingRef.current = true;
+
+    // send the proper header depending on the auth method
+    let finalHeader = "";
+    switch (authMethod) {
+      case PackageRepositoryAuth_PackageRepositoryAuthType.PACKAGE_REPOSITORY_AUTH_TYPE_AUTHORIZATION_HEADER:
+        finalHeader = authCustomHeader;
+        break;
+      case PackageRepositoryAuth_PackageRepositoryAuthType.PACKAGE_REPOSITORY_AUTH_TYPE_BEARER:
+        finalHeader = `Bearer ${bearerToken}`;
```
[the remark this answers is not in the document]
antgamdia: Good catch! I've added to the main issue (https://github.com/vmware-tanzu/kubeapps/issues/4764) as a blocker thing.
I have mixed opinions on this problem: on the one hand, I don't see why the API should care about something else than an "HTTP header authentication" with two alternatives: either an RFC-compliant _Authorization_ header or a custom header.
However, on the other hand, we would be delegating that logic to the clients, but... why not? 

I'll post it in the issue as well, so that we can continue the discussion there.

dashboard/src/components/Config/PkgRepoList/PkgRepoForm.tsx
antgamdia: I've implemented this way as I wasn't really sure of the existence of repos with, let's say: basic auth for accessing the index.yaml + image pull secrets for the images.
In this case, what if a user wants to set the values directly for the basic auth but they already have a Secret in their k8s cluster for the pull secrets? 

I mean, even if we'd be reducing the points of decisions, we would be removing flexibility, wouldn't we?

dashboard/src/components/Config/PkgRepoList/PkgRepoForm.tsx
antgamdia: Good idea! I'll add to this PR tomorrow! 

> This is one of the things that the Plugins customizable UI feature should solve in the future.

Big +1 to this

dashboard/src/shared/PackageRepositoriesService.ts
antgamdia: Mmm, I have to re-check. I would say I faced some problems when encoding the customDetails in the javascript client as the core API client didn't know how to encode this message. I'll have a look at it, but, of course, if the core client is able to handle it, this duplication wouldn't be required.

dashboard/src/components/Config/PkgRepoList/PkgRepoForm.tsx
castelblanque: Yeah, those "Provide the secret values" and "Use existing secrets" sound good too!

dashboard/src/components/Catalog/Catalog.tsx
antgamdia: I've checked it and, in the current impl, fetching `globalReposNamespace` instead of just `""` only gives us the global repos, not every repo the user can access to:

image

dashboard/src/components/Config/PkgRepoList/PkgRepoForm.tsx
antgamdia: > My rationale for the question was: if the user already selected "I will manage the secrets", then how likely is that he/she already created the secrets for the pullSecrets?

Yep, I agree, it's really unlikely wanting to select something different. What if, if the field hasn't been populated yet, set the pullSecret's toggle to the same value. This way we will be suggesting sth to the user, but not forcing to.

> Yes, they are. I feel more natural to write the actual values for the secrets than the secret names from k8s.
Also, note that the secret-selector (re)implementation is still pending, so the UX for this is a little bit worse.

dashboard/src/shared/PackageRepositoriesService.ts
antgamdia: It seems it works! (see https://github.com/vmware-tanzu/kubeapps/pull/4954/commits/991c115b22e1c194644e5e04a6b368c4e685a50d). 
However, a more thorough testing is required, but let's see if we can keep it this (simpler) way

antgamdia:
Ready for a 2nd review :P

dashboard/src/shared/PackageRepositoriesService.ts
```diff
@@ -0,0 +1,233 @@
...
+    switch (request.plugin?.name) {
+      case PluginNames.PACKAGES_HELM:
+        return {
+          typeUrl: `${helmProtobufPackage}.RepositoryCustomDetails`,
+          value: RepositoryCustomDetails.encode({
+            dockerRegistrySecrets: request.customDetails.dockerRegistrySecrets,
+            ociRepositories: request.customDetails.ociRepositories,
+            filterRule: request.customDetails.filterRule,
+            performValidation: request.customDetails.performValidation,
+          } as RepositoryCustomDetails).finish(),
```
castelblanque, line 227: Just a suggestion, haven't tested it, but maybe we could do this?

```suggestion
          value: RepositoryCustomDetails.encode(request.customDetails as RepositoryCustomDetails).finish(),
```

castelblanque: Having `any` as type instead of `Partial<RepositoryCustomDetails>` [here](https://github.com/vmware-tanzu/kubeapps/blob/c0598eb0411183723e927b23c8c2d6b6dcb19cb2/dashboard/src/shared/types.ts#L446) would be more generic, not depending on the Helm implementation. But again, havent't tested it. Perhaps for a future PR 😄

dashboard/src/components/Config/PkgRepoList/PkgRepoForm.tsx
castelblanque: Cool!

dashboard/src/shared/PackageRepositoriesService.ts
antgamdia: Good idea, it should work. Testing it straight away, thanks!

PR #4959, opened 2022-06-27
Update chart to support TLS for pinniped-proxy

absoludity:
### Description of the change

Following on from #4958, this PR updates the kubeapps chart to:
 - supply the pinniped-proxy service with the tls cert and key, when the secret is specified
 - supply the clients of pinniped-proxy (kubeappsapis, and currently still kubeops) with the related CA cert.

### Benefits

The pinniped-proxy service can run as a TLS web service, when configured so.

### Applicable issues

- ref #2268

### DIFF
--- a/chart/kubeapps/values.yaml
+++ b/chart/kubeapps/values.yaml
@@ -1581,6 +1581,10 @@ pinnipedProxy:
   ## @param pinnipedProxy.defaultPinnipedAPISuffix Specify the (default) API suffix
   ##
   defaultPinnipedAPISuffix: pinniped.dev
+  ## @param pinnipedProxy.TLSSecret Specify an optional TLS secret with which to proxy requests
+  TLSSecret: ""
+  ## @param pinnipedProxy.CACert Specify the TLS CA cert which clients of pinniped proxy should use with tls requests
+  CACert: ""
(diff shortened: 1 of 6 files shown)


castelblanque (approved):
LGTM

chart/kubeapps/values.yaml
```diff
@@ -1581,6 +1581,14 @@ pinnipedProxy:
   ## @param pinnipedProxy.defaultPinnipedAPISuffix Specify the (default) API suffix
   ##
   defaultPinnipedAPISuffix: pinniped.dev
+  ## @param pinnipedProxy.TLSSecret Specify an optional TLS secret with which to proxy requests
```
antgamdia, line 1584: I guess we have to run the readmenator to add those params to the chart docs.

antgamdia (approved):
Thanks!

chart/kubeapps/values.yaml
absoludity: Yeah - I'll do that. It does now get run automatically upstream, but no need to wait.

antgamdia: Yes, fortunately, there's no need anymore, but it's nice to have our chart up to date, especially if we, want to give a try to the brand new TLS config. Thanks!

PR #4964, opened 2022-06-28
Updated Pinniped installation instructions

castelblanque:

### Description of the change

This PR fixes wrong instruction in the docs to install Pinniped in a cluster.

In [Using and OIDC provider with Pinniped](https://kubeapps.dev/docs/latest/howto/oidc/using-an-oidc-provider-with-pinniped/#installing-pinniped), the instructions to install Pinniped are wrong.

For example, the impersonation proxy did not exist with previous instructions, so:
`kubectl get credentialissuer -o json`  would return no resources (although the CRD did exist).

### Benefits

User following documentation will have a working Pinniped installation.

### Possible drawbacks

N/A

### DIFF
--- a/site/content/docs/latest/howto/OIDC/using-an-OIDC-provider-with-pinniped.md
+++ b/site/content/docs/latest/howto/OIDC/using-an-OIDC-provider-with-pinniped.md
@@ -4,10 +4,12 @@ The [Pinniped project](https://pinniped.dev/) exists to "Simplify user authentic
 
 ## Installing Pinniped
 
-Install Pinniped into a `pinniped-concierge` namespace on your cluster with:
+Install Pinniped on your cluster with:
 
 ```bash
-kubectl apply -f https://get.pinniped.dev/latest/install-pinniped-concierge.yaml
+kubectl apply -f https://github.com/vmware-tanzu/pinniped/releases/download/v0.18.0/install-pinniped-concierge-crds.yaml
+
+kubectl apply -f  https://github.com/vmware-tanzu/pinniped/releases/download/v0.18.0/install-pinniped-concierge.yaml
 ```
 
 ## Configure Pinniped to trust your OIDC identity provider


site/content/docs/latest/howto/OIDC/using-an-OIDC-provider-with-pinniped.md
```diff
@@ -4,10 +4,12 @@ The [Pinniped project](https://pinniped.dev/) exists to "Simplify user authentic
...
 ## Installing Pinniped
 
-Install Pinniped into a `pinniped-concierge` namespace on your cluster with:
+Install Pinniped on your cluster with:
 
 ```bash
-kubectl apply -f https://get.pinniped.dev/latest/install-pinniped-concierge.yaml
+kubectl apply -f https://github.com/vmware-tanzu/pinniped/releases/download/v0.18.0/install-pinniped-concierge-crds.yaml
+
+kubectl apply -f  https://github.com/vmware-tanzu/pinniped/releases/download/v0.18.0/install-pinniped-concierge.yaml
```
antgamdia, line 12: Do you know if they also offer the `latest` version through their website? Otherwise, we are gonna have our docs outdated soon :(

Also, could you double-check if the rest of our scripts are consistent with this new approach?

castelblanque: > Do you know if they also offer the latest version through their website? 

`latest` is not mentioned in the official documentation [here](https://pinniped.dev/docs/howto/install-concierge/).
Substituting `v0.18.0` for `latest` in Github URL don't work.
It seems that `https://get.pinniped.dev/latest/install-pinniped-concierge-crds.yaml` and `https://get.pinniped.dev/latest/install-pinniped-concierge.yaml` do work...
Anyway, we have versions specified for everything, so I would keep consistency and keep the specific version, so that we know which version we are using. WDYT?

> Also, could you double-check if the rest of our scripts are consistent with this new approach?

Yes, I checked that, but

antgamdia: > It seems that (...) do work...
> Anyway, we have versions specified for everything, so I would keep consistency and keep the specific version, so that we know which version we are using. WDYT?

For the docs, I'd rather keep the `latest` in the docs, otherwise we would be saying "ey, this is the only pinniped version we can recommend", and, imagine there is a security issue solved on their side, unless we send a PR for updating the version, our users would be being recommended a wrong version.

ppbaena: > Anyway, we have versions specified for everything, so I would keep consistency and keep the specific version, so that we know which version we are using. WDYT?

IMO we should suggest to use the `latest` version, any other way we will always have our docs outdated. 

But, for those cases the `latest` version is not provided, we should include an example by using a specific version (like `v.0.18.0`), but it must be clear it is just an example (so they should replace the version in the url if a newer version is available).

Agree with Rafa, we should be consistent in the approach in all our docs.

antgamdia: > Agree with Rafa, we should be consistent in the approach in all our docs.

Yes, I also agree, however, I would distinguish between docs and our internal scripts:

image

While we do need our scripts to be fully replicable (the version has to be fixed, so that we know what we are installing), the docs can just point to `latest`, as it is just a user-faced text. 

In the example above, we only have a docs-related match, whereas the rest are just internal scripts.

castelblanque: That is not my concern, I totally agree that the best approach is having the `latest` in the documentation.
But it is the fact that Pinniped official documentation does not mention any URL with `latest` anywhere. In every place they use `v0.18.0`.
This could be because the URLs with `latest` might be discontinued, or not up-to-date...who knows.
Asked the question at the [Pinniped slack](https://kubernetes.slack.com/archives/C01BW364RJA/p1656495081001239).

castelblanque: Response from Pinniped team:
```
I would not recommend using latest

We change the requirements of the manifest all the time

So it would be trivial for kubeapps to break

For example recently we recently changed the logging config in a way that won't work on older versions

We purposely avoid using latest in our docs as well to make sure the docs are consistent with the version
```

I would leave v0.18.0 and update it accordingly. For example, if in v0.19.0 they make a breaking change, our documentation wouldn't match if we point to latest.

antgamdia: Great, thanks for asking. Got it; let's avoid using `latest` then

antgamdia (approved):

PR #4966, opened 2022-06-28
Specifiy the mininum k8s version in the chart

antgamdia:
### Description of the change

I wasn't aware that Helm allows setting the minimum k8s version until people from Content team just told me. In our case, we are relying on certain resource versions (especially, cronjobs `v1` instead of `v1beta`, see #3846) and, when we update them, we are not backward compatible. 

### Benefits

We will be clear about the version we do support.

### Possible drawbacks

Users won't be able to install Kubeapps on unsupported versions, which might be not ok in some cases (people working on old k8s versions not using those old resource versions, for instance, just using the flux or carvel plugin instead of the helm one).

### Applicable issues

N/A

### Additional information

Happy to close the PR if we don't really like the outcome as described above.

### DIFF
--- a/chart/kubeapps/Chart.yaml
+++ b/chart/kubeapps/Chart.yaml
@@ -2,6 +2,7 @@ annotations:
   category: Infrastructure
 apiVersion: v2
 appVersion: DEVEL
+kubeVersion: ">=1.21"
 dependencies:
   - name: redis
     repository: https://charts.bitnami.com/bitnami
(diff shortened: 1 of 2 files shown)


chart/kubeapps/Chart.yaml
```diff
@@ -2,6 +2,7 @@ annotations:
   category: Infrastructure
 apiVersion: v2
 appVersion: DEVEL
+kubeVersion: ">=1.21"
```
castelblanque, line 5: Cool, didn't know about that parameter!

ppbaena:
I agree with this approach. From https://kubernetes.io/releases/ :

> The Kubernetes project maintains release branches for the most recent three minor releases (1.24, 1.23, 1.22). Kubernetes 1.19 and newer receive [approximately 1 year of patch support](https://kubernetes.io/releases/patch-releases/#support-period). Kubernetes 1.18 and older received approximately 9 months of patch support.

So in our case, setting that parameter for 1 version behind the officially version maintained by Kubernetes should be enough.

PR #4977, opened 2022-06-29
Add a message to troubleshoot an aks oidc problem

antgamdia:
### Description of the change

When upgrading our pre-prod server, I noticed the AKS-issued login wasn't working. After digging into it, I discovered it is due to [an open issue in oauth2-proxy](https://github.com/oauth2-proxy/oauth2-proxy/issues/1680), whose version has been recently updated to 7.3.0 (and so has it in our chart).

This PR adds a quick message with the workaround I've used to get it to work again.

### Benefits

AKS users will still be able to follow our guide and get it configured properly.

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

N/A


absoludity (approved):

ppbaena:
Thanks @antgamdia for the update.

Issue #2792, opened 2021-05-12
Build and push the integration image during the CI


antgamdia:
Currently, the `kubeapps/integration-tests:v1.0.0` image is being built manually. After https://github.com/kubeapps/kubeapps/pull/2783, we enabled dependabot in the `/integration` folder. As a result, the dependency versions appearing in the code won't match the actual ones in the container image.

Note that this image is usually built when releasing a new Kubeapps version.

It was suggested that we perform an automatic building and versioning directly from our CI system, so we get rid of any manual step.

More information in https://github.com/kubeapps/kubeapps/pull/2783#issuecomment-838038384

PR #5004, opened 2022-07-01
Change integration img base image

### LINKED ISSUE #4206: Clean up integration container image
### Description:

Integration [container image](https://github.com/kubeapps/kubeapps/blob/main/integration/Dockerfile) is still tailored to Puppeteer needs. Now that Kubeapps will be using Playwright, this image should get a review, mainly for checking dependencies used.

### Describe the results you expect:

Simple and lightweight image to run Playwright tests.

### Additional information:

See [this comment](https://github.com/kubeapps/kubeapps/pull/4185#discussion_r796665253) from #4185

antgamdia:
### Description of the change

This PR replaces the old custom image for running the e2e test with a simpler one. It is just a just a `FROM mcr.microsoft.com/playwright` dockerfile, plus installing the browsers with `npx playwright install`, as stated in [the Playwright docs](https://playwright.dev/docs/ci#introduction). 

Besides, this PR also renames `executor.yaml` to `e2e-runner.yaml`.

### Benefits

We will rely on an official (and supported) image of Playwright. Otherwise, we would be stuck to a certain base image, as already discussed at #4206.

### Possible drawbacks

N/A

### Applicable issues

- fixes #4206 

### Additional information

Now that the image has been simplified, we should start thinking of #2792. However, the base image is pretty large (~1GB), so perhaps we should skip the `push` phase.


integration/manifests/e2e-runner.yaml
```diff
@@ -21,7 +21,7 @@ spec:
             - tail
             - -f
             - /dev/null
-          image: kubeapps/integration-tests:v1.1.1
+          image: kubeapps/integration-tests:v2.0.0
```
castelblanque, line 24: Small detail unrelated to this line, but to L8,L9,L13 and L17.
Given that the file has been renamed to `e2e-runner`, shouldn't we rename the deployment accordingly? `integration` seems too generic.

castelblanque (approved):
Thanks!

integration/manifests/e2e-runner.yaml
antgamdia: Ah good idea, thanks!

Issue #153, opened 2018-03-06
See status of AppRepository syncs (especially failures) in UI


prydonius:
See also https://github.com/kubeapps/dashboard/issues/7

Since AppRepository syncing is done asynchronously, we can't really report back a status immediately after adding a repo. But the App Repositories configuration page should report the status:

- See when syncs are happening
- See if a sync has failed, check logs of failing sync
- See a healthy status if syncs are succeeding

This will require a read-only list/get/logs for pods in the `kubeapps` namespace.

m1small:
It would be nice to complement this with the number of charts retrieved from that source.

absoludity:
Related, it'd be great to validate the user input when adding a repository (ie. verifying the repo can be contacted with the entered details).

dlaloue-vmware:
+ to improve on repository feedback.
I am facing an issue where all sync jobs are failing, and the only log message available is:

sync
time="2021-09-08T02:00:51Z" level=fatal msg="GET request to [https://releases.rancher.com/server-charts/stable/index.yaml] failed due to status [401]: <?xml version='1.0' encoding='UTF-8'?><Error><Code>AuthenticationRequired</Code><Message>Authentication required.</Message></Error>"

when the index.yaml file is actually not requiring any authentication (can get access it from any browser, curl).

absoludity:
The request for the rancher index.yaml does 401 if an authorization header is provided:

```
curl -Is -H "Authorization: foo" https://releases.rancher.com/server-charts/stable/index.yaml | head -n1
HTTP/2 401 
```

Are you certain you didn't define an auth header for the repo? If I do the same request with an empty Authorization header, it's fine:

```
curl -Is -H "Authorization: " https://releases.rancher.com/server-charts/stable/index.yaml | head -n1 
HTTP/2 200 
```

Not sure what more we could log in that specific case, since we can't include the headers. But it would be helpful if the log message indicated whether an authorization header was sent (filled).

dlaloue-vmware:
here is the apprepository CRD generated by the UI (global repository):

apiVersion: kubeapps.com/v1alpha1
kind: AppRepository
metadata:
  creationTimestamp: "2021-09-08T01:48:31Z"
  generation: 2
  name: rancher
  namespace: system-kubeapps
  resourceVersion: "74157222"
  selfLink: /apis/kubeapps.com/v1alpha1/namespaces/system-kubeapps/apprepositories/rancher
[...]

PR #5006, opened 2022-07-01
Add status helm repos API

antgamdia:
### Description of the change

I've noticed the Helm plugin is not setting the `status.ready` field. As it defaults to `false`, the UI shows misleading information to the users. This PR sets it to `true`, but ideally, we would want actually to query the status of the repo sync.

### Benefits

No more misleading "not ready" info to the users.

### Possible drawbacks

A failed sync job will still be marked as "ready" when it is not. However, this is the initial behavior and there is another (old) issue tracking it: #153. 

### Applicable issues

- related https://github.com/vmware-tanzu/kubeapps/issues/4764

### Additional information

This is just a workaround, but we should be able to query, somehow, the status of an AppRepository job's execution.


castelblanque (approved):
Thanks!
Agree, we should check if we can have a proper status up-to-date.

PR #5065, opened 2022-07-12
Add ignore clause to netlify builds

### LINKED ISSUE #4878: Netlify reporting "failed" status main's pushes when hitting a cached version
_Converting into an issue since I haven't found yet a quick way to do it._

Currently, the Netlify builds often fail because there is no new content added to the website. Since it has a cache, it compares the cached version with the PR's (to main) content and given there are no changes, no deploy action is performed.
That is a great feature since we save some building cycles (in the free plan they are limited), but, as a result, it reports a failed status.
The only real downside of it is that we have the "failed" badge on top of our readme.

Alternatives:

1. Find a solution to force builds without cache (+ no failed status; - more building time consumed)
2. Remove the badge and ignore the building status (+ non-visible failed status; - no badge)
3. Stay happy with the failed badge as is
4. Release website changes in a separate branch (I wouldn't go with this option)

Personally, I'd go with 2, but see what you think.

ppbaena:
### Description of the change

Netlify builds often fail because there is no new content added to the website. It compares the cached version with the PR's (to main) content and given there are no changes, no deploy action is performed. Adding an ignore clause to netlify.toml file to prevent this situation, so builds only proceed when changes in the base directory (/site).

### Benefits

- Save some building cycles for netlify.
- Avoid the "failed" badge for netlify on top of our readme.

### Possible drawbacks

N/A

### Applicable issues

- fixes #4878

### DIFF
--- a/site/netlify.toml
+++ b/site/netlify.toml
@@ -1,6 +1,7 @@
 [build]
 publish = "public"
 command = "hugo --gc --minify"
+ignore = "git diff --quiet $CACHED_COMMIT_REF $COMMIT_REF ."
 
 [context.production.environment]
 HUGO_VERSION = "0.98.0"


ppbaena:
See [netlify docs](https://docs.netlify.com/configure-builds/ignore-builds/) and example 5 in [this support question](https://answers.netlify.com/t/support-guide-how-to-use-the-ignore-command/37517) mentioning the same use case we're looking for (meaning deploy only changes inside /site folder).

antgamdia (approved):

PR #5067, opened 2022-07-12
Avoid overriding categories on each chunk

### LINKED ISSUE #5066: Pkg categories get deleted
**Describe the bug**
The category list saves the response of the response, which is now paginated. As a result, we are losing those categories not present in the current chunk.

**To Reproduce**

1. Go to the catalog.
2. See how categories start showing/disapearing

**Expected behavior**
The categories should remain regardless of the ones sent in each chunk

**Screenshots**

loadingcats

- Version DEVEL
- Kubernetes version N/A
- Package version N/A

**Additional context**

N/A

antgamdia:
### Description of the change

Since the API (after pagination) is returning the categories for the current chunk of pkg repos, replacing the whole `categories: xxxxx` leads to end up with an incomplete set of categories, as reported in the issue.
This PR aggregates the categories like we are doing for the packages.

### Benefits

Categories won't get overridden on each chunk.

### Possible drawbacks

N/A

### Applicable issues

- fixes #5066

### Additional information

~Draft as I have to write a test to prevent regressions.~

### DIFF
--- a/dashboard/src/reducers/availablepackages.test.ts
+++ b/dashboard/src/reducers/availablepackages.test.ts
@@ -273,6 +273,50 @@ describe("packageReducer", () => {
     expect(state2.items.length).toBe(2);
   });
 
+  it("two receiveAvailablePackageSummaries should add categories (no dups)", () => {
+    const state1 = packageReducer(
+      {
+        ...initialState,
+        isFetching: true,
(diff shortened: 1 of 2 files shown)


castelblanque (approved):
Great, thanks!

PR #5070, opened 2022-07-13
flux OCI support part 5

gfichtenholt:
Uptake oras.land/oras-go/v2 v2.0.0-rc.1 which fixes https://github.com/oras-project/oras-go/issues/196


antgamdia (approved):
Great, thanks for updating it!

PR #5072, opened 2022-07-13
Fix app fetch error

antgamdia:
### Description of the change

This PR hides the LoadingWrapper when it has finished the fetching and there is an error (it was leading to a bug reported at https://github.com/vmware-tanzu/kubeapps/issues/4960). Besides, I've updated the test suite to set the state types so that we can get earlier feedback at build (aka transpile) time.

### Benefits

No more endless spinner when loading an installed package that doesn't exist. Instead, an error message will be thrown, and the user will be able to go to the apps view.

### Possible drawbacks

N/A

### Applicable issues

- fixes https://github.com/vmware-tanzu/kubeapps/issues/4960

### Additional information

fetch


castelblanque (approved):
Great, thanks!

PR #5081, opened 2022-07-15
Add imagePullSecrets option to repos API

### LINKED ISSUE #5015: Specify imgPullSecrets in Helm's Repos API by its values, not only by secret name
_Extracted from https://github.com/vmware-tanzu/kubeapps/issues/4764_

**Summary**
 Add a way to specify img pull secrets in the helm plugin for kubeapps-managed secrets. 

**Background and rationale**
Currently, the former UI supports both approaches, but the API only allows setting the secret manually. This has led to disabling the e2e test nº 03/10.

**Description**
Add a way to set the imgPullSecret (a Helm plugin's custom detail) by its value, not only by secret name.

**Acceptance criteria**

- The Helm repos API is able to be passed imgPullSecret values and it creates the secret under the hood.
- The UI has enabled back the functionality to pass the values (a toggle button has been disabled to prevent users from using it)

**Additional context**

N/A

castelblanque:

### Description of the change

This PR changes the new package repositories API with regards to handling the images pull secret credentials in the direct Helm plugin.
Until now, the new API allowed to set such credentials by specifying the secret name only. With this PR it is now possible to do it in one of this mutually exclusive ways:
- Specifying the secret name
- Providing the credentials (server, user, password and email)

Using one way or the other still depends on the global configuration parameter `userManagedSecrets` (defaults to false).

### Benefits

Images pull secret management is more in line with the authentication and other secrets management in the API.

### Possible drawbacks

Backwards incompatibility with previous state of package repositories API, but it wasn't being used.

### Applicable issues

- resolves #5015


cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/repositories.go
```diff
@@ -358,6 +395,53 @@ func (s *Server) updateRepo(ctx context.Context, repo *HelmRepository) (*corev1.
 	}, nil
 }
 
+func handleImagesPullSecret(ctx context.Context, typedClient kubernetes.Interface,
+	isUserManagedSecrets bool, repo *HelmRepository, newRepo bool) (imgPullSecret *k8scorev1.Secret, updateImgPullSecret bool, err error) {
+	if repo.customDetails == nil || repo.customDetails.ImagesPullSecret == nil {
+		if !isUserManagedSecrets {
+			managedSecretName := imagesPullSecretName(repo.name.Name)
+			secretsInterface := typedClient.CoreV1().Secrets(repo.name.Namespace)
+			if deleteErr := deleteSecret(ctx, secretsInterface, managedSecretName); deleteErr != nil {
```
antgamdia, line 404: If not found, it will yield an error and we won't be able to install the repo, AFAIK

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/repositories_auth.go
```diff
@@ -192,19 +194,128 @@ func newAppRepositoryAuth(secret *k8scorev1.Secret,
...
+		}
+		secret.Data[k8scorev1.DockerConfigJsonKey] = dockerConfigJson
+	}
+
+	return secret, isSameSecret, nil
+}
+
+func deleteSecret(ctx context.Context, secretsInterface v1.SecretInterface, secretName string) error {
+	// Ignore action if secret didn't exist
+	if secret, _ := secretsInterface.Get(ctx, secretName, metav1.GetOptions{}); secret != nil {
```
antgamdia, line 275: When performing a `.Get` operation, if not found, the struct being returned is an empty one, like &corev1.Secret{}, but it is not `nil`. 

Instead, I'd do:
```suggestion
	if _, err := secretsInterface.Get(ctx, secretName, metav1.GetOptions{}); err == nil {
```

antgamdia (approved):
Apart from the suggestion for the bug I found (when performing the integration of these changes with my current UI work), LGTM. Thanks!

PR #5083, opened 2022-07-16
Make CI build and use the e2e image

### LINKED ISSUE #2792: Build and push the integration image during the CI

antgamdia:
### Description of the change

This PR simply adds a new step in the CI so that the e2e runner image is built on each run.

### Benefits

The CI will always use the latest integration image as it is being built on each CI run.

### Possible drawbacks

If a dependency happens to break the e2e tests, the CI runs will also fail.

### Applicable issues

- fixes #2792

### Additional information

N/A


castelblanque:
Not sure that we want to build the E2E image every CI run. That adds a lot of time to the (already) slow CI pipeline.
On top of that, it is an image that rarely changes, so why having it built every time?

castelblanque:
> Perhaps we could improve it and just build the image for pushes to main if you want to save some time

+1 to building the image less times.

Another option would be to have the task that builds it run only if e.g. integration's`Dockerfile` has changed using GH actions like [this](https://github.com/RasaHQ/pr-changed-files-filter): `enables conditional execution of workflow steps and jobs, based on the files modified by pull request, feature branch or in pushed commits.`

ppbaena:
> Perhaps we could improve it and just build the image for pushes to main

+1. In general, we should try to decrease CI times, but it makes sense for pushes to main, so we rely on what we're building is going to work later when releasing (do not wait to the release process to identify a dependency problem).

castelblanque (approved):
Approving as discussed in Slack.

PR #5097, opened 2022-07-19
Make repos' bearer auth prefix-safe

### LINKED ISSUE #5028: [Package repositories] Review and fix backend and frontend handling of auth headers
**Summary**
Review and fix combined backend and frontend handling of package repositories auth headers.

**Background and rationale**
While reviewing PR #4954, I noticed that the handling of auth header values is different by plugin.

For bearer authentication, Helm plugin assumes that the content of the header in the repo request is only the token itself without `Bearer`, whereas implementation for Kapp assumes that it contains both. 
See this [LOC in the backend of Helm plugin](https://github.com/vmware-tanzu/kubeapps/blob/main/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/repositories_auth.go#L89) and this [LOC in the backend of Kapp plugin](https://github.com/vmware-tanzu/kubeapps/blob/main/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_adapters.go#L1063). Flux does not implement this auth type.

UI must send the right data to the backend in order to avoid repeating the `Bearer` string many times when storing the repo. This is: if the Helm plugin only expects the token itself (without `Bearer`), but UI is sending the whole header, then the final header value stored might be `Bearer Bearer <token>`.

It might happen with other types of authentications.

**Description**
This task is for reviewing that all three plugins handle the repo authentication values in the same way.

Also make sure that vocabulary used is consistent. Avoid referring to token when using a header, e.g.:

```go
token := auth.GetHeader()
```

**Acceptance criteria**
Repositories authentication data is correctly handled for all plugins between backend and frontend.
Objects/Structs are correctly named when referring to headers and tokens.

antgamdia:
### Description of the change

This PR addresses the Bearer authentication in the two plugins supporting it (Helm and Carvel) and makes the implementation "prefix-safe". Regardless of what the API client send (either `Bearer foo` or `foo`) the prefix will be trimmed out and the secret's content will become `Bearer foo`.

### Benefits

The UI will be able to send the bearer token regardless of the plugin's implementation.

### Possible drawbacks

N/A

### Applicable issues

- fixes #5028

### Additional information

I haven't changed anything related to the terminology as, in supporting both (prefixed/non-prefixed) the "token" can actually be the token and not the whole header. Anyway, happy to rename if you still think so.

Edit: squeezing some golang-ci-lint-related code; the tool (used in the gh action) got updated and new checks were being triggered. Ignoring them as it is just code used for testing.

### DIFF
--- a/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
+++ b/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
@@ -524,8 +524,7 @@ func (s *Server) GetInstalledPackageDetail(ctx context.Context, request *corev1.
 		}
 	}
 	// trim the new doc separator in the last element
-	//nolint:staticcheck
-	valuesApplied = strings.Trim(valuesApplied, "---")
+	valuesApplied = strings.TrimSuffix(valuesApplied, "---")
 
 	installedPackageDetail, err := s.buildInstalledPackageDetail(pkgInstall, pkgMetadata, pkgVersionsMap, app, valuesApplied, cluster)
 	if err != nil {
--- a/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/globals_vars_test.go
+++ b/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/globals_vars_test.go
@@ -341,7 +341,12 @@ var (
 		},
 	}
 
-	addRepoReqBearerToken = func(token string) *corev1.AddPackageRepositoryRequest {
+	addRepoReqBearerToken = func(token string, withPrefix bool) *corev1.AddPackageRepositoryRequest {
+		prefixedToken := token
+		if withPrefix {
+			prefixedToken = "Bearer " + token
+		}
+
 		return &corev1.AddPackageRepositoryRequest{
 			Name:            "bar",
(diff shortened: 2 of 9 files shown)


cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go
```diff
@@ -524,8 +524,7 @@ func (s *Server) GetInstalledPackageDetail(ctx context.Context, request *corev1.
 		}
 	}
 	// trim the new doc separator in the last element
-	//nolint:staticcheck
-	valuesApplied = strings.Trim(valuesApplied, "---")
+	valuesApplied = strings.TrimSuffix(valuesApplied, "---")
```
antgamdia, line 527: The golint gh action said we were using `.Trim` incorrectly: it trims chars (unicode runes), so using `---` was -partially- wrong, as it wasn't removing the occurrences of the string `---`  but rather removing every occurrence of `-`.

To trim strings, we should use TrimSuffix or TrimPrefix instead (or Replace even). See https://topic.alibabacloud.com/a/the-difference-between-the-trimright-and-trimsuffix-of-the-golang-strings-package_1_38_30927804.html

Squeezing this change in this PR.

castelblanque (approved):
LGTM

Issue #5107, opened 2022-07-21
Add kapp plugin add "fetch" details in the repos UI


antgamdia:
**Summary**
The kapp plugin's custom details include several fields for customizing the version constraints and many other options. Unfortunately, we are not supporting them in the UI yet.

**Background and rationale**
For the sake of simplicity, we just migrated the existing UI to ensure feature parity when we switched to the new repos API.

**Description**
The UI should display those plugin-specific options (besides the helm plugin's).

**Acceptance criteria**

- The UI allows setting kapp plugin's custom details.

**Additional context**
N/A

PR #5108, opened 2022-07-21
Minor fixes in repos UI 

antgamdia:
### Description of the change

While adding the pullsecrets to the UI, I've come across minor bugs/improvements to be performed. This PR is for sending those simple changes upstream. Specifically:

- Add kapp's customdetails handling: in the initial impl I just assumed Helm's custom details, so here I'm paving the way for supporting kapp's details
- Improvements on repo list page: renamed to `packaging formats` and added the `type`. For instance: `Helm (OCI)`. Moved the tooltip position.
- Skip building customdetail if any: if a repo happens to support  customDetails but they are not passed (like what we are doing now for the kapp's details - as it isn't yet implemented), ignore the building and encoding procress.
- Avoid infinite loading after editing a repo: due to a missing `isFetching` reset in the state, the spinner was being rendered forever after a repo edit.

### Benefits

See above

### Possible drawbacks

N/A

### Applicable issues

- related #4764
- related #5107

### DIFF
--- a/dashboard/src/shared/PackageRepositoriesService.ts
+++ b/dashboard/src/shared/PackageRepositoriesService.ts
@@ -235,7 +235,7 @@ export class PackageRepositoriesService {
         return {
           typeUrl: `${kappControllerProtobufPackage}.KappControllerPackageRepositoryCustomDetail`,
           value: KappControllerPackageRepositoryCustomDetail.encode(
-            request.customDetail as KappControllerPackageRepositoryCustomDetail,
+            request.customDetails as KappControllerPackageRepositoryCustomDetail,
           ).finish(),
         } as Any;
       default:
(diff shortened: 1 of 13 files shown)


dashboard/src/shared/PackageRepositoriesService.ts
```diff
@@ -223,8 +231,15 @@ export class PackageRepositoriesService {
             request.customDetail as HelmPackageRepositoryCustomDetail,
           ).finish(),
         } as Any;
+      case PluginNames.PACKAGES_KAPP:
+        return {
+          typeUrl: `${kappControllerProtobufPackage}.KappControllerPackageRepositoryCustomDetail`,
+          value: KappControllerPackageRepositoryCustomDetail.encode(
+            request.customDetail as KappControllerPackageRepositoryCustomDetail,
+          ).finish(),
+        } as Any;
```
castelblanque, line 240: Can we add a test to be sure that Helm and Kapp details are handled correctly?

castelblanque (approved):

ppbaena (approved):

PR #5106, opened 2022-07-21
Rename to `customDetail`

antgamdia:
### Description of the change

This PR simply performs a consistent rename from `customDetails` (note the final `s`) to `customDetail` (as defined in the proto file).

### Benefits

Today, the custom details weren't being processed simply due to this extra `s`.

### Possible drawbacks

N/A

### Applicable issues

- related #4764 

### Additional information

N/A

PR #5109, opened 2022-07-21
Add pullSecret management in repos UI

antgamdia:
> **Warning**
> ~This PR is currently targeting a temporary branch assuming both https://github.com/vmware-tanzu/kubeapps/pull/5081 and https://github.com/vmware-tanzu/kubeapps/pull/5106 have been merged. This way, I will just re-target the PR once these PRs get actually merged. Main benefit: this PR will include fewer changes and will be easier to be reviewed.~

### Description of the change

After the revamping of the pullsecret handling https://github.com/vmware-tanzu/kubeapps/pull/5081, the UI ought to be able to handle this information. This should also unblock the currently commented out e2e test, but it will be performed in a follow-up PR.

This PR is mainly to allow passing the pullsecret directly, instead of just referencing the secret name. Besides, this PR also allows copying the information from the auth credentials (as we were doing in the previous UI) as, usually, both values are the same.

### Benefits

Users will be able to specify img pull secrets for the helm plugin's repos.

### Possible drawbacks

N/A

### Applicable issues

- fixes https://github.com/vmware-tanzu/kubeapps/issues/5015

### Additional information

image

image


dashboard/src/components/Config/PkgRepoList/PkgRepoForm.tsx
```diff
@@ -260,48 +288,80 @@ export function PkgRepoForm(props: IPkgRepoFormProps) {
       filter = toFilterRule(filterNames, filterRegex, filterExclude);
     }
 
-    const success = await onSubmit({
-      authHeader: isUserManagedSecret ? "" : finalHeader,
+    // build the custum details object based for each plugin
```
castelblanque, line 291:
```suggestion
    // build the custom details object based for each plugin
```

castelblanque:
> This PR is currently targeting a temporary branch assuming both https://github.com/vmware-tanzu/kubeapps/pull/5081 and https://github.com/vmware-tanzu/kubeapps/pull/5106 have been merged. This way, I will just re-target the PR once these PRs get actually merged. Main benefit: this PR will include fewer changes and will be easier to be reviewed.

@antgamdia Both #5106 and #5081 are merged now, so you can proceed.

dashboard/src/shared/PackageRepositoriesService.ts
```diff
@@ -223,14 +223,30 @@ export class PackageRepositoriesService {
     // An "Any" object has "typeUrl" with the FQN of the type and a "value",
     // which is the result of the encoding (+finish(), to get the Uint8Array)
     // of the actual custom object
+    let detail, helmCustomDetail: HelmPackageRepositoryCustomDetail;
```
castelblanque, line 226: Both `detail` and `helmCustomDetail` seem to be scoped to the `PluginNames.PACKAGES_HELM` case block. Why not declaring it inside, instead of outside the `switch` statement?

castelblanque:
Good job! Happy to see that integration test number 10 is finally passing successfully 😄

castelblanque (approved):

dashboard/src/shared/PackageRepositoriesService.ts
antgamdia: I was simply getting a linter error (https://eslint.org/docs/latest/rules/no-case-declarations), but after a more thorough reading, I can just add brackets around the `case:  { ........ }` to avoid the linter issue. Will do it (and check similar issues in the code) straight away! Thanks

PR #5113, opened 2022-07-22
Flux oci support 6: implement Update OCI repo and integration tests

gfichtenholt:
The goal of this PR was to implement Update() for OCI helm repositories. I expected 
flux source-controller to notify my code if the remote OCI repository contents changes, 
like it does with a regular HTTP helm repository.

Like most of my PRs, this was test-driven. The plan was:
1) To write an integration test that is using flux source-controller 
  that was failing due to production code I had yet to write
   (i.e. not implemented Update)
2) Implement Update operaiton
3) Finish coding integration test

That was the plan. Unfortunately it did not work out due to reasons beyond my 
control. For the integration test to be reproducible, I needed to have "my own" 
OCI registry the contents of which I can modify (i.e add or delete chart versions) 
and I don't have to worry about others making changes to it. In addition to serving this integration test having "my own" registry would be a big stop toward resolving an outstanding dependency to GitHub registry which is not in my ownership, so any chart version hard-coded in the test are subject to change  

The PR reflects all my attempts to do so:
1) I tried to set up a local helm OCI registry over HTTP the way it is advised in 
https://helm.sh/docs/topics/registries/ and have the integration test use that. 
That effort failed due to https://github.com/fluxcd/source-controller/issues/805
2) I tried to set up a local helm OCI registry over HTTPS. That failed due to 
https://github.com/fluxcd/source-controller/issues/807 and https://github.com/helm/helm/issues/6324
3) Since I could not make progress with local registry, I then set up a remote public registry on 
GitHub. I set up access control so others can read, but not modify the repo contents. 
The setup worked but the integration test was ultimately blocked due to flux not supporting 
the use case I had in mind. https://github.com/fluxcd/source-controller/issues/839. 
This discussion is ongoing. I did not want to discard my work so far, because all the scenarios 
should be valid once the dependent issues are resolved

Update 7/25: added an integration test for what is arguably more important scenario that works fine:
1)  remote OCI repo contains only podinfo version 6.1.5
2) create flux HelmRepository for (1)
3) create flux HelmRelease for (2) results in podinfo version 6.1.5 deployed
4) push podinfo version 6.1.6 to (1)
5) wait for specified repo interval + a bit
6) now (3) is automatically upgraded to 6.1.6
i.e. flux server-side is auto-upgrading to latest chart when it is pushed to remote


cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/oci_repo.go
```diff
@@ -261,7 +261,14 @@ func (r *OCIRegistry) downloadChart(chart *repo.ChartVersion) (*bytes.Buffer, er
 
 	// trim the oci scheme prefix if needed
 	getThis := strings.TrimPrefix(u.String(), fmt.Sprintf("%s://", registry.OCIScheme))
-	return r.helmGetter.Get(getThis, clientOpts...)
+	log.Infof("about to call helmGetter.Get(%s)", getThis)
```
antgamdia, line 264: Not sure if this log should be a debug-level one ?

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/testdata/kind-cluster-setup.sh
```diff
@@ -5,88 +5,283 @@
 
 # this is used to build an image that can be used to stand-up a pod that serves static test-data in 
 # local kind cluster. Used by the integration tests. This script needs to be run once before the running
-# the test(s) 
+# the test(s). This script requires GitHub CLI (gh) to be installed locally. On MacOS you can
+# install it via 'brew install gh'
```
antgamdia, line 9: I'd also add the link to the releases page: 
`https://github.com/cli/cli/releases`

antgamdia (approved):
Awesome, thanks for the integration tests ensuring it works as expected!

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/oci_repo.go
gfichtenholt: all these pesky log messages will be gone by end of next week. They're here only while I was actively working on this feature. See my 2nd action item in https://github.com/vmware-tanzu/kubeapps/issues/5007#issuecomment-1186031237. I just haven't got to it yet

PR #5115, opened 2022-07-22
Add pending e2e tests for the repos UI

### LINKED ISSUE #4876: Build E2E tests for the new package repository API UI
**Summary**
Build E2E tests for the new package repository API UI when work in #4764 is finished.

**Description**
Build tests in Playwright that test each feature offered by the new UI built for the package repository API that makes use of all three plugins existing.

**Acceptance criteria**
E2E tests that cover the new UI for the package repository UI.

antgamdia:
### Description of the change

In previous PRs we have been extensively working on the new repos API and UI. However, the integration wasn't fully working due to several blocking issues.
This PR is to re-enable some commented-out test cases we had disabled.

### Benefits

 Once passing, we should be able to start creating the PR against `main`, as, in theory, it wouldn't break the current UX.

### Possible drawbacks

If they don't work, we might discover additional issues we have to fix 😝 

### Applicable issues

- fixes #4876

### Additional information

N/A

### DIFF
--- a/integration/tests/main/03-create-private-package-repository.spec.js
+++ b/integration/tests/main/03-create-private-package-repository.spec.js
@@ -28,7 +28,10 @@ test("Create a new private package repository successfully", async ({ page }) =>
 
   await page.locator("text=Add Package Repository >> div").click();
   await page.fill("input#kubeapps-repo-name", repoName);
-  await page.fill("input#kubeapps-repo-url", "http://chartmuseum.chart-museum.svc.cluster.local:8080");
+  await page.fill(
+    "input#kubeapps-repo-url",
+    "http://chartmuseum.chart-museum.svc.cluster.local:8080",
+  );
   await page.locator("text=Helm Charts").first().click();
   await page.locator("text=Helm Repository").click();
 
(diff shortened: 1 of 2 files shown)


antgamdia:
More updates here... it's odd: I've tunneled the CircleCI deployments, so that I could interact with the UI directly. After setting the images manually via:

````
helm upgrade -n kubeapps kubeapps-ci /home/circleci/project/chart/kubeapps  --set authProxy.enabled=false --set dashboard.image.repository=kubeapps/dashboard-ci  --set dashboard.image.tag=build-f298db6ea6a0df3b08fed307dd581840be68e7d7 --set kubeappsapis.image.repository=antgamdia/kubeapps-apis --set kubeappsapis.image.tag=dev
````
(note that the handcrafted image is just adding log traces)

 it seems to work, the main issue (copying the secrets to the global ns) is gone and the packages are actually fetched [meaning my both previous fixes were useless for the sake of this issue 😛]. It's like the `kubeapps-apis` is not getting properly built or uploaded to the cluster somehow.

For instance, if switching manually the image and we use the one built by the CI, namely `kubeapps/kubeapps-apis-ci:build-f298db6ea6a0df3b08fed307dd581840be68e7d7`, it does perform the secret copy! 

So this is gonna be my focus the next day: get any differences between the images loaded in the CI cluster when running the e2e test suite vs the expected ones. At least we are tracking down the problem.

castelblanque:
> For instance, if switching manually the image and we use the one built by the CI, namely kubeapps/kubeapps-apis-ci:build-f298db6ea6a0df3b08fed307dd581840be68e7d7, it does perform the secret copy!

That is weird! Could you please share link to a CircleCI run failing with the problem, so that I can take a look to see if it helps?

castelblanque:
@antgamdia Found the source of the problem.
Scenario in which it happens:
- Global repos namespace is set different from the Kubeapps installation namespace. E.g. `kubeapps-global-repos` vs `kubeapps`
- Repo created is of type direct Helm
- Repo created is namespaced (not global)
- Repo is private (needs a secret)

With the introduction of #5140 , the secret for namespaced repos is copied to the *global namespace*, which is incorrect.
The right one is the *kubeapps namespace*. Both are equal by default, but when `apprepository.globalReposNamespaceSuffix` is specified, syncer job runs in kubeapps namespace, and secrets are being created in the global namespace. This leads the syncer job creation to fail with error: `Error: secret "test-apprepo-example-repo" not found`.

I will create a PR that will copy the secrets not to the global namespace, but to the kubeapps one.

antgamdia:
> I will create a PR that will copy the secrets not to the global namespace, but to the kubeapps one.

THANKS! I totally missed it. So... the CI was working properly and detecting an actual error indeed. We always blame the poor CI, haha. 😅

castelblanque:
@antgamdia Problem should be fixed with #5183.

antgamdia:
> @antgamdia Problem should be fixed with #5183.

indeed; tests 03 and 10 are now passing! I'm retriggering the CI again because of an operators-related issue (hoping it is just a transient thing 🙏 ).

antgamdia:
Finally PASSING ! Merging and preparing the PR against `main` soon.

PR #5117, opened 2022-07-24
Switch to use upstream ExpiringValueCache.

### LINKED ISSUE #4934: Move kubeapps' cache of expiry-aware values upstream
**Summary**

During the last iteration, part of the work for #2875 involved creating a cache implementation for values which know themselves when they expire (see #4899). This code could exist in the upstream library that we used, and given that the [author of the library agrees it would be valuable for others there](https://github.com/jaemk/cached/issues/115), we should move it there instead.

**Background and rationale**

We benefited from the upstream library and can give some benefit back.

**Acceptance criteria**

- Kubeapps does not contain its own cache implementation but rather uses the upstream one.

absoludity:

### Description of the change

Just circling back to remove our custom `ExpiredCacheValue` cache and instead use the one we landed upstream.

### Benefits

Less code to maintain here.

### Applicable issues

- fixes #4934


absoludity:
Sorry, my editor's auto format-on-save also made some changes (just rust-fmt). If that's an issue, let me know.

ppbaena:
Thanks Michael!

Issue #5082, opened 2022-07-15
Release new Kubeapps version (2.5.0)


ppbaena:
**Summary**
Release a new version of Kubeapps

**Background and rationale**
Kubeapps monthly release including [Use the new Repositories API in the UI](https://github.com/vmware-tanzu/kubeapps/issues/4764).

**Description**
Release a new version of Kubeapps including all features, enhancements and bug fixing merged once the phase 1 for [Use the new Repositories API in the UI](https://github.com/vmware-tanzu/kubeapps/issues/4764) is ready.

This new version must comply [semver](https://semver.org/), though it means change a little bit Kubeapps versioning, but to be adapted to standards.

**Acceptance criteria**
 - New release is tagged and available in github
 - New release is available in Bitnami Application Catalog
 - New release is available in Tanzu Package for Kubeapps

**Additional context**
In the past Kubeapps versioning rules did not follow any criteria, so to help our community users Kubeapps maintainers team will start using [semver](https://semver.org/) for Kubeapps releases.

antgamdia:
Dependencies have been bumped up (https://github.com/vmware-tanzu/kubeapps/pull/5123). The next step is to start reporting scanning and reporting the licenses internally, as well as to declare a dependency-freeze period.

ppbaena:
Release 2.5.0 reported in BOSS Director

antgamdia:
Starting a manual review of the release candidate, some bugs were found, but we are working to send a fix ASAP. Once this step is done, we will be able to trigger a release soon.

antgamdia:
Tests are failing on GKE somehow:

![Image](https://user-images.githubusercontent.com/11535726/184234236-547e2c5c-6f24-473a-ad04-cc528b008fc1.png)

antgamdia:
The release is finally out: https://github.com/vmware-tanzu/kubeapps/releases/tag/v2.5.0

PR #4216, opened 2022-02-02
Use OCI code from Helm as a dependency

antgamdia:
### Description of the change

_This PR is still a draft since it requires some changes (albeit really minor) on the helm side to work. I'll update the status here once I have more info_

In this PR we moved from the OCI chart puller code (that came from Helm), to an OCI registry client using Helm as a dependency.

Some refactor has been required in order to properly mock the relevant functions for testing. More comments inline.

### Benefits

We no longer have directly "imported" code from Helm, but we will use it as a dependency now the OCI support moved out from experimental.

### Possible drawbacks

Perhaps some unnoticed incompatibility?

### Applicable issues

- fixes #4194

### Additional information

Current status:

- [x] Changes in our code: this PR
- [x] Changes in Helm: done in my personal Helm fork
- [x] PR in the Helm repo: https://github.com/helm/helm/pull/10638
- [ ] Helm releasing a new version
- [x] Final IRL test to ensure everything works as expected

PR #5123, opened 2022-07-25
Prerelease updates

antgamdia:
### Description of the change

This PR is just performing the usual update on our dependencies as described in our release process.

### Benefits

Up-to-date deps

### Possible drawbacks

Perhaps an incompatibility? Let's our CI bespeak.

### Applicable issues

- related #5082

### Additional information

N/A


dashboard/Dockerfile
```diff
@@ -12,5 +12,5 @@ COPY . /app
 RUN yarn run prettier-check && yarn run ts-compile-check
 RUN yarn run build
 
-FROM bitnami/nginx:1.22.0-debian-11-r3
+FROM bitnami/nginx:1.23.1-debian-11-r1
```
antgamdia, line 15: I have confirmed that the update to the branch 1.23.x is also going to be carried out by the Bitnami team near soon.

antgamdia:
I haven't pushed any CI image as I'd rather wait for https://github.com/vmware-tanzu/kubeapps/pull/5083 to be merged.

antgamdia:
I've been looking into upgrading also the Helm version to 3.9.X (as required by #4216, eventually). In this version, they have also bumped the k8s versions up. As long as we also update our k8s dependencies we should be fine... well, we're not:

Currently, our code is not compatible with k8s >=1.24 (including upgrading the submodule) due to the following deprecations:

- [ ] Implement a `ShutDownWithDrain` function in the `Type` type (in [this file](https://github.com/vmware-tanzu/kubeapps/blob/main/cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/cache/rate_limiting_queue.go)). This is easy, and even we might replace the `reset` function we implemented (not sure), cc @gfichtenholt, [no action required now, simply just for you to know]
- [ ] Use an updated version of the `kapp` dependency. It relies on a `coreClient.Discovery().ServerResources()` that has been finally deprecated in favor of `ServerGroupsAndResources()` in k8s 1.24.

In short, even if we perform the change on the Flux queue, we would still be blocked by the kapp dependency; in particular, [this line](https://github.com/vmware-tanzu/carvel-kapp/blob/develop/go.mod#L23). If we manually update client-go in our code, we will get several compilation errors due to a dependency rename from `googleapis/gnostic` to `google/gnostic`.

So, in short, as of today, we cannot update to Helm 3.9.X nor k8s >= 1.24 unless kapp is using a `k8s.io/client-go` >= 0.23.X

gfichtenholt:
I can take care of ShutDownWithDrain in the next PR

antgamdia:
> I can take care of ShutDownWithDrain in the next PR

Unfortunately, this function is not yet available in the [workqueue interface](https://pkg.go.dev/k8s.io/client-go@v0.23.9/util/workqueue) on client-go v0.22.x, as it was introduced in v0.23. That said, I don't see any problem with implementing this additional function now, this way we will have done that change once we bump it up from 0.22.x, so thanks for taking care of it!

gfichtenholt:
> > I can take care of ShutDownWithDrain in the next PR
> 
> Unfortunately, this function is not yet available in the [workqueue interface](https://pkg.go.dev/k8s.io/client-go@v0.23.9/util/workqueue) on client-go v0.22.x, as it was introduced in v0.23. That said, I don't see any problem with implementing this additional function now, this way we will have done that change once we bump it up from 0.22.x, so thanks for taking care of it!

No problem. Will take care of it

castelblanque (approved):
Great, thanks!

antgamdia:
Kapp has just released a new version, https://github.com/vmware-tanzu/carvel-kapp/releases/tag/v0.52.0, updating the k8s deps!

PR #5140, opened 2022-07-28
Added Helm global secret for namespaced repos

### LINKED ISSUE #5129: Repos API/Helm - Support global private repos
**Summary**
We aren't copying the namespace repos' secrets to the global namespace in the new implementation, as opposed to the former implementation.

**Description**
I've been working on polishing up the UI leftovers but,  when I finished fixing the e2e tests, I noticed they weren't working with the new API.
After some debugging time, I noticed the error thrown by the job (asset-syncer) is sth like:  Error: secret `default-apprepo-my-repo-88078` not found
Note that the secrets' name for namespace repos are like `apprepo-%s` . However, there is [another function](https://github.com/vmware-tanzu/kubeapps/blob/f030403973c748a15940ba0a535832370839cd0d/pkg/kube/kube_handler.go#L1085) that adds the namespace as a prefix, like `%s-apprepo-%s`
This function is referenced [here](https://github.com/vmware-tanzu/kubeapps/blob/f030403973c748a15940ba0a535832370839cd0d/pkg/kube/kube_handler.go#L524)... pointing out an [actual issue](https://github.com/vmware-tanzu/kubeapps/issues/1647) (that we, maybe mistakenly, closed).

To the best of my knowledge, it seems the asset-syncer is spinning up the sync jobs in the global namespace and therefore it is requiring every namespace repo's secret to be copied to that location. Otherwise, private repos won't sync.

So, if my assumptions are right, we could either:

1. Adapt the current Helm plugin's code to also copy the secret as we were already doing in the kubeops-based API.
2. Adapt the asset-syncer's code to create the jobs in each repo's namespace (aka fixing #1647 )
3. Continue to work on the apprepo controller full revamp (no more postgres, etc...) that Greg was exploring.

Given the current bandwidth and priorities, I'd simply go with 1) and reopen #1647.

**Acceptance criteria**

- Global private repos can be installed and used in Kubeapps
- PR #5115 works

**Additional context**

As @absoludity commented, we could give cluster-wide access to read secrets to the controller. Although it's not great that our controller would have that access, I think it's the most sensible option right now to support private repos while they're indexed by the controller explicitly. That's also an option for us, and would be better than replicating the ugly copying of the secret.
Perhaps that can even be refined these days so that our controller only has access to the specific secret object that it needs. Or we could only enable that RBAC if support for private repos is enabled or something so that it's not there by default

castelblanque:

### Description of the change

This PR adds the handling of global secrets that affects to namespaced package repositories with secrets related.
Asset-syncer spings up the sync jobs **in the global namespace** and therefore it is requiring every namespace repo's secret to be copied to that location. Otherwise, private repos won't sync. 

This management wasn't implemented yet in the new package repos API logic.

This wasn't, and isn't the best solution (cloning secrets into the global ns) but for now we have followed the same approach as in the old implementation.

### Benefits

Repos controller can find now the secret for namespaced repositories in the global namespace.

### Possible drawbacks

Maybe some edge cases not yet detected where repository controller needs the global secret? Can't think of any now. 

### Applicable issues

- fixes #5129

### DIFF
--- a/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/repositories.go
+++ b/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/repositories.go
@@ -5,6 +5,7 @@ package main
 
 import (
 	"context"
+
 	apprepov1alpha1 "github.com/vmware-tanzu/kubeapps/cmd/apprepository-controller/pkg/apis/apprepository/v1alpha1"
 	corev1 "github.com/vmware-tanzu/kubeapps/cmd/kubeapps-apis/gen/core/packages/v1alpha1"
 	"github.com/vmware-tanzu/kubeapps/cmd/kubeapps-apis/gen/plugins/helm/packages/v1alpha1"
@@ -72,7 +73,7 @@ func (s *Server) newRepo(ctx context.Context, repo *HelmRepository) (*corev1.Pac
 			return nil, err
 		}
 	}
-	// Copy secret to global namespace if needed
+	// Copy secret to global namespace if needed. See issue #5129.
 	if repo.name.Namespace != s.globalPackagingNamespace && secret != nil {
 		if err = s.copyRepositorySecret(typedClient, secret, repo.name); err != nil {
 			return nil, err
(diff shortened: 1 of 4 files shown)


cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/repositories.go
```diff
@@ -72,6 +72,12 @@ func (s *Server) newRepo(ctx context.Context, repo *HelmRepository) (*corev1.Pac
 			return nil, err
 		}
 	}
+	// Copy secret to global namespace if needed
```
antgamdia, line 75: I would add here a reference to the issue, maybe #5129 or the previous one, in order to have the full context once we get back to this line in the future.

antgamdia (approved):
Great, thanks! Looking forward to integrating these changes into my branch to see if the e2e tests finally pass!

PR #5141, opened 2022-07-28
Fixed E2E tests reporting output

castelblanque:

### Description of the change

E2E tests output in CI shows a red error message saying

```
Configuration Error: HTML reporter output folder clashes with the tests output folder:

    html reporter folder: /app/reports-html
    test results folder: /app/reports

HTML reporter will clear its output directory prior to being generated, which will lead to the artifact loss.
```

image

This might be due to some path pattern usage that Playwright does.
Both paths have been made different with this PR.

### Benefits

E2E CI tests not display the red message about wrong report path anymore.

### Possible drawbacks

N/A

### Applicable issues

N/A

### DIFF
--- a/integration/playwright.config.js
+++ b/integration/playwright.config.js
@@ -27,8 +27,8 @@ const config = {
       use: { ...devices["Desktop Chrome"] },
     },
   ],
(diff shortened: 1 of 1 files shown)


antgamdia (approved):
Thanks!

PR #5153, opened 2022-08-01
Update kubeapps.dev site

beni0888:
### Description of the change

This PR contains the following changes:
- Updates the contributors' section.
- Adds support for running the site locally via `docker` and `make`.

### Benefits

- Developers no longer need to install `Hugo` in their local machines to render `kubeapps.dev` site.
- The list of contributors reflects the current state of the team


script/makefiles/site.mk
```diff
@@ -0,0 +1,12 @@
+# Copyright 2021-2022 the Kubeapps contributors.
+# SPDX-License-Identifier: Apache-2.0
+
+WORKDIR = $(shell pwd)
+HUGO_VERSION = "0.101.0"
```
castelblanque, line 5: Why not "latest"?

script/makefiles/site.mk
```diff
@@ -0,0 +1,12 @@
+# Copyright 2021-2022 the Kubeapps contributors.
+# SPDX-License-Identifier: Apache-2.0
+
+WORKDIR = $(shell pwd)
```
castelblanque, line 4:
```suggestion
WORKDIR = $(pwd)
```

script/makefiles/site.mk
beni0888: Well, I was thinking about that and wasn't sure about the fact of using a non-fixed version that could break the compatibility at any point... Do you think it would be better to use latest here?

site/README.md
```diff
@@ -27,6 +27,17 @@ Access the site at [http://localhost:1313](http://localhost:1313). Press `Ctrl-C
 
 The [site/content/docs/latest](./content/docs/latest) directory holds the project documentation whereas [site/themes/template/static../img/docs](./themes/template/static../img/docs) contains the images used in the documentation. Note they have to be under that folder to be properly served.
 
+#### Run Hugo with Docker
+
+To ease the local development and prevent you from polluting your local environment with tools that rarely use, 
+we have added the option of running the `Hugo` server via `Docker` through a `Make` target.
```
castelblanque, line 33: For development there is a new style guide that contributors must start following ([see here](https://github.com/vmware-tanzu/kubeapps/blob/main/site/content/docs/latest/reference/developer/style-guide.md#avoid-using-we)). It recommends avoiding "we".

castelblanque (approved):
Awesome! Thanks for the contribution.

ppbaena (approved):

script/makefiles/site.mk
castelblanque: This piece of code is meant for checking that website is OK when rendered by Netlify. Maybe we should check which version is used by Netlify? I mean, this should be in sync with that.

beni0888:
Already applied all suggestions. Could you please review it again? 🙏🏻

castelblanque (approved):
Great, many thanks!!

script/makefiles/site.mk
```diff
@@ -0,0 +1,13 @@
+# Copyright 2021-2022 the Kubeapps contributors.
```
antgamdia, line 1:
```suggestion
# Copyright 2022 the Kubeapps contributors.
```

antgamdia (approved):
Thanks for the improvements here!

script/makefiles/site.mk
beni0888: This is a copy&paste from another script in the same file, I think we should be consistent and, in case we agreed on changing this, we should do it for every file in the repo.

@antgamdia Do you agree to merge it as is and handle this task in another issue?

antgamdia: No, no, it is just a one-time change for this specific file. The policy we are following (after checking with the legal counsel office) is using the date on which the file was created and the date of the more recent (significant) change.
In case of fresh new files, like this one, both dates match, so, to avoid redundancies, we shorten `2022-2022` with just `2022`, wherefore my suggestion.

beni0888: Ok, I understand. Let's apply the suggestion then. Thanks for the explanation!

PR #5163, opened 2022-08-04
Fix DeploymentForm app name regular expression

youjeongsue:
### Description of the change
Fix first word in regular expressions of DeploymentForm release name not to contain number

### Benefits
If the release name starts with number, it seems to be created in kubeapps(-we can click the create button-), but the actual creation fails because it does not match the naming convention of Kubernetes. This change can prevent this kind of creation failure.

### Possible drawbacks
I think there's no drawbacks

### Applicable issues

I cannot find related issue. (Do I need to register as an issue first?)

### Additional information

N/A

### DIFF
--- a/dashboard/src/components/DeploymentForm/DeploymentForm.tsx
+++ b/dashboard/src/components/DeploymentForm/DeploymentForm.tsx
@@ -217,7 +217,7 @@ export default function DeploymentForm() {
                   <label>Name</label>
                   <input
                     id="releaseName"
-                    pattern="[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*"
+                    pattern="[a-z]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*"
                     title="Use lowercase alphanumeric characters, '-' or '.'"
                     onChange={handleReleaseNameChange}
                     value={releaseName}


antgamdia:
First of all, thanks for the contribution, we really appreciate it!

Just some additional information here: 

The release names with leading numbers are indeed valid, as the existing regex is DNS1136-compliant and it indeed matches the one in Kuberentes (see [this file](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L178)) 
However, the `Service`s objects have more strict names, as they are DNS1035 (see https://github.com/kubernetes/kubernetes/issues/3752). The main difference is the leading digit, as you noticed.

So the question is: should Kubeapps be more strict than k8s and disallow leading numbers (DNS1035) from every installed package (regardless of this package is yielding `Service`s objects or not), or should it just remains with DNS1136-compliant names and let the server complain about the error ? 

Even if the user experience is a bit worse, I'm keen for leaving the DNS1136 validation (as currently is) as Kubeapps shouldn't be more restrictive than other package managers are (like helm CLI, for instance).


> (Do I need to register as an issue first?)

Yes, it would be great; this way we can track and reference the issues more quickly. And, in this case, we can bring the discussion there.

youjeongsue:
Thank you for quick response! First of all, I agree with your opinion. Especially this part.
> Even if the user experience is a bit worse, I'm keen for leaving the DNS1136 validation (as currently is) as Kubeapps shouldn't be more restrictive than other package managers are (like helm CLI, for instance).

Since I mainly deployed the service objects through Kubeapps, I don't seem to have thought of any other cases. I also agree with the opinion that it would be nice if kubeapps wasn't stricter than kubernetes or helm cli etc.

So I will close this pr. Thanks to you, I got to know the various naming cases in kubernetes and understand more about Kubeapps. Thank you so much for your kind explanation! :) (And sorry for posting it as a pr first, not an issue)

PR #5183, opened 2022-08-09
[Helm plugin] Changed repo secrets copied now to Kubeapps namespace

### LINKED ISSUE #5129: Repos API/Helm - Support global private repos

castelblanque:

### Description of the change

With the introduction of #5140, the secret for namespaced repos was being copied to the global namespace, which is incorrect.
The right one is the **kubeapps** namespace. Both are equal by default, but when `apprepository.globalReposNamespaceSuffix` is specified, syncer job runs in kubeapps namespace, and secrets are being created in the global namespace. This leads the syncer job creation to fail with error: `Error: secret "test-apprepo-example-repo" not found`.

This PR changes logic when creating and updating package repos so that secrets are copied to the `kubeapps` namespace if repo wasn't created in that namespace. In this way, we make sure that asset syncer has the secret available.

### Benefits

Namespaced repos can be synced correctly now.

### Possible drawbacks

We need to differentiate between *global* repos (in the Kubeapps global namespace, e.g. `kubeapps-repos-global`) and repos in `kubeapps` namespace (which might not be global).

### Applicable issues

- fixes #5140
- fixes #5129

### Additional information

This bug was found to be the origin of the problem mentioned in [this comment](https://github.com/vmware-tanzu/kubeapps/pull/5115#issuecomment-1208243856) of 5115.

### DIFF
--- a/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/globals_vars_test.go
+++ b/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/globals_vars_test.go
@@ -235,7 +235,7 @@ var (
 		},
 		ObjectMeta: metav1.ObjectMeta{
 			Name:            "bar",
-			Namespace:       "kubeapps",
+			Namespace:       globalPackagingNamespace,
 			ResourceVersion: "1",
 		},
 		Spec: v1alpha1.AppRepositorySpec{
@@ -259,7 +259,7 @@ var (
 
 	addRepoReqGlobal = &corev1.AddPackageRepositoryRequest{
 		Name:            "bar",
-		Context:         &corev1.Context{Namespace: "kubeapps", Cluster: KubeappsCluster},
+		Context:         &corev1.Context{Namespace: globalPackagingNamespace, Cluster: KubeappsCluster},
 		Type:            "helm",
 		Url:             "http://example.com",
 		NamespaceScoped: false,
--- a/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/repositories.go
+++ b/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/repositories.go
@@ -73,9 +73,11 @@ func (s *Server) newRepo(ctx context.Context, repo *HelmRepository) (*corev1.Pac
 			return nil, err
 		}
 	}
-	// Copy secret to global namespace if needed. See issue #5129.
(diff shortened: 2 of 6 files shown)


antgamdia (approved):
Great!! Thanks for catching the error and finding a solution. Just a question: what would happen when updating a repo? Is the secret also being updated in the kubeapps ns?

Issue #5165, opened 2022-08-04
Filter by repo not showing pkgs if exceeding pagination


antgamdia:
**Describe the bug**

When debugging why my PR https://github.com/vmware-tanzu/kubeapps/pull/5115 was failing I noticed there is an actual issue when several repos have been added and we try to filter out by repos: the filter yields a `No application matches the current filter.` even if when removing some repos the package pops up.

**To Reproduce**

1. Add the bitnami repo (as it contains several packages)
2. Add another repo with at least 1 package (like the chartmusseum one that we install when e2e-testing)
3. Do not scroll down to the bottom (to prevent the whole packages to be fetched)
4. Click on the repo filter and select the additional one you added (so the URL becomes sth like `/catalog?Repository=my-repo1`)
5. Notice the `No application matches the current filter.`  message.

**Expected behavior**
The packages from the additional repo should appear.

**Screenshots**
N/A

**Desktop (please complete the following information):**

N/A

**Additional context**
This is blocking https://github.com/vmware-tanzu/kubeapps/pull/5115 (and therefore the release)

antgamdia:
While debugging this issue I noticed it can be somehow related to #5168. At some point the react state gets reset (because of a new repos filter item) and the `isFetching` param is wrongly set (because the carvel plugin is returning data when shouldn't, I guess).
I'm still looking into it, but probably will send a fix for #5168 on top of #5115 to see if this works around the problem.

antgamdia:
Well... after some investigation here I think I've discovered some problems:

- The pagination wasn't working as designed, that is, on the very first load we were retrieving the whole list of packages regardless of the scrolling. 
  - Proposed solution: make the UI decide when to paginate, but leave the nextPageToken selection to the backend.

Example of my local's vs what is currently on `main` just accessing `/#/c/default/ns/default/catalog`:
![image](https://user-images.githubusercontent.com/11535726/183074495-c3a44197-f7f0-4a4d-aa8b-94fc2aebceae.png)

- Because of the above (well, not because, but above behavior was increasing the frecency) a filter checkbox could be clicked arbitrarily (which may be OK) leading, in some cases, to a component re-render with in-flight state updates. A solution was given at https://github.com/vmware-tanzu/kubeapps/pull/4714, but the conditional ignore was leading to this issue.
  - Proposed solution: disable the checkboxes when loading to prevent multiple clicks.

<details>
  <summary>Show demo video</summary>
[...]

PR #5185, opened 2022-08-09
Make pagination UI-driven, but based upon server's tokens

antgamdia:
### Description of the change

- The pagination wasn't working as designed, that is, on the very first load we were retrieving the whole list of packages regardless of the scrolling. 
  - Proposed solution: make the UI decide when to paginate, but leave the nextPageToken selection to the backend.

Example of my local's vs what is currently on `main` just accessing `/#/c/default/ns/default/catalog`:
image

### Benefits

Fewer initial requests and better UX, as endless pagination will work as it is expected to.

### Possible drawbacks

N/A

### Applicable issues

- related #5165

### Additional information

Squeezing a minor linter-related change.


dashboard/src/components/Catalog/Catalog.tsx
```diff
@@ -149,30 +151,32 @@ export default function Catalog() {
         cluster,
         namespace,
         reposFilter,
-        nextPageToken,
+        localNextPageToken,
```
castelblanque, line 154: Why renaming the variable? Why is important the nuance of "local"?

castelblanque (approved):
LGTM

dashboard/src/components/Catalog/Catalog.tsx
antgamdia: Becasue `nextPageToken` is coming from the global packages state and `localNextPageToken` is a copy of this value, but in the component's state. This way, we have fine-grained control of the lifecycle of the state. 
In fact, this is the crux of the PR: `nextPageToken` gets updated each time new data arrives (that's why it was failing).
 So, technically we are not renaming but storing a local copy of `nextPageToken` and managing how it should be updated (that is, on each pagination event).

PR #5187, opened 2022-08-09
Disable catalog filters when `isFetching`

antgamdia:
### Description of the change

A filter checkbox could be clicked arbitrarily (which may be OK) leading, in some cases, to a component re-render with in-flight state updates. A solution was given at https://github.com/vmware-tanzu/kubeapps/pull/4714, but the conditional ignore was leading to this issue.
  - Proposed solution: disable the checkboxes when loading to prevent multiple clicks.

https://user-images.githubusercontent.com/11535726/183107750-2c39f44a-0b44-4a7c-984b-30da7e8cb8a5.gif

### Benefits

Multiple clicks on the filters won't generate unnecessary requests, at least, we have to wait until the ongoing request finishes to prevent inconsistencies.

### Possible drawbacks

A user can perceive the filtergroup as unresponsive if clicking like a bot... but I think that's fair enough. 

### Applicable issues

- related #5165

### Additional information

[Demo gif](https://user-images.githubusercontent.com/11535726/183107750-2c39f44a-0b44-4a7c-984b-30da7e8cb8a5.gif)

*test are still pending*


castelblanque:
> A user can perceive the filtergroup as unresponsive if clicking like a bot... but I think that's fair enough.

Agree, although usually in those "disabled" situations the controls are greyed out for the user to notice.
It would be a matter of adding some CSS rules.

castelblanque (approved):

antgamdia:
> Agree, although usually in those "disabled" situations the controls are greyed out for the user to notice.
> It would be a matter of adding some CSS rules.

Well... I ended up refactoring the whole component to use `cds` instead of the old `clr-xxx` CSS rules. This brings us several improvements in both a11y and richer UX. Example:

disabledclick

Working on the tests, but will push my changes soon.

castelblanque:
> Well... I ended up refactoring the whole component to use `cds` instead of the old `clr-xxx` CSS rules. This brings us several improvements in both a11y and richer UX. Example:

That looks very good!

PR #5190, opened 2022-08-10
Delete git submodule

### LINKED ISSUE #5184: Delete the git submodule
**Summary**
From time to time, we got some mismatches between the local/upstream commit HEAD pointer to the `codegen` submodule. It was added 5 years ago (see https://github.com/vmware-tanzu/kubeapps/pull/181/files)... but it is seldom used. I don't see any reason for us to keep having it.

**Background and rationale**
Since we are not using the checkout code, it is an unnecessary extra element to maintain., IMO.

**Description**
Remove the git submodule and the unnecessary code.

**Acceptance criteria**
- Remove the git submodule 
- Remove the unnecessary related code.
- Inform the team members to sync locally this change as well

**Additional context**
N/A

antgamdia:
### Description of the change

As stated in #5184, we are not really using the `code-generator` code and keeping it in sync is sometimes a pain when upgrading deps. Therefore, as agreed with the team, this PR is getting rid of it.

### Benefits

Less code to maintain.

### Possible drawbacks

N/A

### Applicable issues

- fixes #5184

### Additional information

Perhaps you may need to remove it locally as well:

```bash
git config  --edit 
# Manually remove the section under "[submodule "code-generator"]"

git rm -r --cached .\cmd\apprepository-controller\vendor
````


castelblanque (approved):
LGTM

PR #5199, opened 2022-08-11
Move favicons outside a subfolder

antgamdia:
### Description of the change

The favicons icons can't be accessed from a subfolder. Probably it would require some nginx rules, but we can just extract them to the root folder as they were before. This PR is doing so.

### Benefits

Favicons will work again.

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

N/A


castelblanque (approved):
LGTM

PR #5203, opened 2022-08-11
Fix for Carvel global namespaces

### LINKED ISSUE #5200: Global repositories handling is inconsistent - Carvel issue
**Describe the bug**
Creating global repositories is not correctly handled now that plugins came into the picture.

- Helm direct plugin assumes that the global namespace is made up of whatever comes from the chart values in `apprepository.globalReposNamespaceSuffix` as a suffix. E.g. `kubeapps-repos-global`
- Carvel plugin assumes `kapp-controller-packaging-global`. See [the backend definition](https://github.com/vmware-tanzu/kubeapps/blob/main/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server.go#L34).
- Flux plugin does not have global namespaces

This bug report focuses on Carvel.

**To Reproduce**

For the Carvel problem:

1. Enable Carvel plugin
2. Switch context in Kubeapps to the global namespace, e.g. `kubeapps-repos-global`
3. Add a Carvel package repository
4. Get error `Namespace Scope is inconsistent with the provided Namespace`

Furthermore, if the repository is instead added to the namespace `kapp-controller-packaging-global`, it fails with the same error.
For the rest of namespaces it works.

**Expected behavior**
Package repo is created.

**Screenshots**
image

castelblanque:

### Description of the change

This PR partially addresses bug #5200 about global namespaces handling for Carvel repositories.

### Benefits

It is possible now to have Carvel package repositories correctly created in the global namespace, where global repository depends on the plugin.
Listing of repositories still aggregates both global from Helm and from Carvel.

### Possible drawbacks

There is still a lot of frontend code that assumes global namespace is only the one for Helm.
We need to work in the future to refactor the global namespaces part.

### Applicable issues

- fixes #5200 

### Additional information

Follow-up work tracked in https://github.com/vmware-tanzu/kubeapps/issues/5202


dashboard/craco.config.js
```diff
@@ -8,7 +8,11 @@ module.exports = {
   webpack: {
     configure: {
       plugins: [
-        new NodePolyfillPlugin(), // add the required polyfills (not included in webpack 5)
+        // add the required polyfills (not included in webpack 5)
+        new NodePolyfillPlugin({
+          // Allow using console.log
+          excludeAliases: ['console'],
+        }),
```
antgamdia, line 15: Interesting... don't know how I haven't faced this problem in my local :S

dashboard/src/shared/utils.ts
```diff
@@ -212,3 +213,23 @@ export function getSupportedPackageRepositoryAuthTypes(
...
+
+/**
+ * Check if namespace is global depending on the plugin
+ *
+ * @param namespace Namespace to check
+ * @param pluginName Plugin name
+ * @param kubeappsConfig Kubeapps configuration
+ * @returns true if namespace is global
+ */
+export function isGlobalNamespace(namespace: string, pluginName: string, kubeappsConfig: IConfig) {
```
antgamdia, line 225: Great, thanks for the refactor here!

antgamdia (approved):
Awesome, thanks for giving a last-minute workaround for this problem!

dashboard/craco.config.js
castelblanque: Try removing the modules folder, lock file and `yarn install` maybe.

PR #5204, opened 2022-08-11
Skip re-fetching deleted installedPkgs status

### LINKED ISSUE #5205: Deleted applications showing an error after deletion
**Describe the bug**
When a package gets uninstalled, after a few seconds, the UI shows:

image

_Originally reported by @beni0888_

**To Reproduce**

1. Delete a package
2. Wait some seconds

**Expected behavior**
No error

**Screenshots**
Above

Devel

**Additional context**
N/A

antgamdia:
### Description of the change

This PR adds a check before getting the installed pkg status: "if there is not an active subscription, skip the process". This way, when unmounting components, like when we are deleting a package (we get redirected to the apps view), the fetch won't happen. 

### Benefits

If the app had been deleted, the fetch would have failed, showing an unpleasant error back to the user. 

### Possible drawbacks

N/A

### Applicable issues

- fixes #5205

### Additional information

N/A

### DIFF
--- a/dashboard/src/actions/installedpackages.test.tsx
+++ b/dashboard/src/actions/installedpackages.test.tsx
@@ -3,41 +3,41 @@
 
 import {
   AvailablePackageDetail,
-  InstalledPackageDetail,
-  InstalledPackageSummary,
   GetInstalledPackageSummariesResponse,
+  InstalledPackageDetail,
   InstalledPackageReference,
   InstalledPackageStatus,
-  VersionReference,
-  ReconciliationOptions,
   InstalledPackageStatus_StatusReason,
+  InstalledPackageSummary,
+  ReconciliationOptions,
+  VersionReference,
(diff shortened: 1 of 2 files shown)


castelblanque:
Is this really blocking the release?
I'm wondering if we could leave it for the next one (patch release?).
In this way, there will be less risk for breaking new stuff. Plus the UI manual testing has already been done, and we should do another one to check regressions. WDYT?

antgamdia:
> Is this really blocking the release?

For me, this is pretty bad, enough to block the release. I mean, see the UX right after deleting...

errordelete

Note that the function `getInstalledPkgStatus` is only being called at `getResources`, which is something scoped to the installed packages view. Worst case regression is an installed package not changing its status from "reconciling" to "ready"... but that's all. However, leaving it as is would have implications for each pkg we delete...

castelblanque (approved):

PR #5207, opened 2022-08-11
Fix test environment on GKE

antgamdia:
### Description of the change

The local registry setup for the e2e test is not working on GKE. While finding a way to create a local registry there (main issue is dealing with certs), it's ok with disabling this test (and its setup) for now. This PR is moving the `10-xxx` test into the `multicluster` folder (which is not run when on GKE).

Plus: using a *cluster*rolebinding for allowing listing repos from `ns=""` in the EDIT user in CI

### Benefits

GKE tests (and therefore the release) will pass.

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

N/A


antgamdia:
Tested and working in pre-release:

image

script/e2e-test.sh
```diff
@@ -309,9 +309,11 @@ installChartMuseum "${CHARTMUSEUM_VERSION}"
...
 pushChart apache 8.6.3 admin password
 
-# Setting up local Docker registry
-setupLocalDockerRegistry
-pushLocalChart
+# Setting up local Docker registry if not in GKE
+if [[ -z "${GKE_BRANCH-}" ]]; then
+  setupLocalDockerRegistry
+  pushLocalChart
+fi
```
antgamdia, line 316: The local docker registry doesn't work right now on GKE as the current configuration requires the k8s cert to be copied for issuing the certs. The quickest solution: ignore those tests when in GKE.

script/e2e-test.sh
```diff
@@ -396,7 +398,7 @@ kubectl create serviceaccount kubeapps-edit -n kubeapps
 # default). See https://github.com/vmware-tanzu/kubeapps/issues/4435
 kubectl create rolebinding kubeapps-edit -n kubeapps --clusterrole=edit --serviceaccount kubeapps:kubeapps-edit
 kubectl create rolebinding kubeapps-edit -n default --clusterrole=edit --serviceaccount kubeapps:kubeapps-edit
-kubectl create rolebinding kubeapps-repositories-read -n kubeapps --clusterrole kubeapps:kubeapps:apprepositories-read --serviceaccount kubeapps:kubeapps-edit
+kubectl create clusterrolebinding kubeapps-repositories-read --clusterrole kubeapps:kubeapps:apprepositories-read --serviceaccount kubeapps:kubeapps-edit
```
antgamdia, line 401: Changing to clusterrolebinding instead for allowing listing repos when using `namespace=""`
Issue tracked at https://github.com/vmware-tanzu/kubeapps/issues/5215

ppbaena (approved):

PR #5209, opened 2022-08-11
Add support for setting target architecture for images

### LINKED ISSUE #5208: Support building docker images for ARM architectures
**Summary**
Add/improve the support for developing the project using a computer with ARM architecture.

**Background and rationale**
Currently, the docker images build scripts don't specify the platform for the images, and they are built for `linux/amd64` (although I guess this could depend on the version of docker in use), so they cannot be run on computers with ARM processors.

**Description**
It should be possible to build the images for different architectures, including at least `linux/amd64` and `linux/arm64`.

**Acceptance criteria**
- Docker images can be built locally for a target platform by providing a parameter to the corresponding Make target.

**Additional context**
N/A

beni0888:
### Description of the change

This PR adds support for specifying the target CPU architecture for the docker images through a parameter in the call to the corresponding make target. This PR doesn't modify the CI, it just adds support for building the docker images for different platforms through `Make`.

### Benefits

Docker images can be built for different CPU architectures (others than AMD64), so developers using computers with those architectures can take advantage of it.

### Possible drawbacks

N/A

### Applicable issues

- Helps fix #5208 

### Additional information

N/A

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,14 @@ GOFMT = /usr/bin/env gofmt
 IMAGE_TAG ?= dev-$(shell date +%FT%H-%M-%S-%Z)
 VERSION ?= $$(git rev-parse HEAD)
 TARGET_ARCHITECTURE ?= amd64
+SUPPORTED_ARCHITECTURES := amd64 arm64 riscv64 ppc64le s390x 386 arm/v7 arm/v6
+
+ifeq ($(filter $(TARGET_ARCHITECTURE),$(SUPPORTED_ARCHITECTURES)),)
+COMMA:=,
+EMPTY:=
+WHITESPACE:=$(EMPTY) $(EMPTY)
+$(error The provided TARGET_ARCHITECTURE '$(TARGET_ARCHITECTURE)' is not supported, provide one of '$(subst $(WHITESPACE),$(COMMA)$(WHITESPACE),$(SUPPORTED_ARCHITECTURES))')
+endif
 
 default: all
 
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ GO = /usr/bin/env go
 GOFMT = /usr/bin/env gofmt
 IMAGE_TAG ?= dev-$(shell date +%FT%H-%M-%S-%Z)
 VERSION ?= $$(git rev-parse HEAD)
+TARGET_ARCHITECTURE ?= amd64
 
 default: all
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -27,10 +28,10 @@ all: kubeapps/dashboard kubeapps/apprepository-controller kubeapps/kubeops kubea
 # Currently the go projects include the whole repository as the docker context
 # only because the shared pkg/ directories?
 kubeapps/%:
-	DOCKER_BUILDKIT=1 docker build -t kubeapps/$*$(IMG_MODIFIER):$(IMAGE_TAG) --build-arg "VERSION=${VERSION}" -f cmd/$*/Dockerfile .
+	DOCKER_BUILDKIT=1 docker build --platform "linux/$(TARGET_ARCHITECTURE)" -t kubeapps/$*$(IMG_MODIFIER):$(IMAGE_TAG) --build-arg "VERSION=${VERSION}" -f cmd/$*/Dockerfile .
```
antgamdia, line 31: I'd do `--platform "$(TARGET_ARCHITECTURE)"` instead. If Windows containers are ever supported in buildkit, we'd like to also have them.

Besides, I'd rather add some notes commenting the following:

The target platforms supported by buildkit are `linux/amd64, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/arm/v7, linux/arm/v6`

However, if a user happens to use Windows Containers (not WSL or VM), it won't work as buildkit doesn't support Windows (see https://github.com/microsoft/Windows-Containers/issues/34)

Note that it refers to the container architecture. For instance, a Docker for Windows or Docker for Mac usually runs Linux containers,so we only need to build these two target platforms: `linux/amd64` and `linux/arm64`

antgamdia (approved):
See the comment about generalizing the target arch and adding a comment. Otherwise, +1

Makefile
beni0888: After discussing this topic, we've agreed on just providing the `TARGET_ARCHITECTURE` and keeping hardcoded the operating system, as we won't likely support Windows containers in the near future and have to provide the whole platform string (eg. `linux/arm64`) would add more complexity. 

In addition, instead of adding a not regarding the supported platforms, I've decided to add a validation in the code, so `Make` will fail with a descriptive error if an unsupported architecture is provided.

castelblanque (approved):
Excellent, thanks for the improvement!

PR #5212, opened 2022-08-14
Flux oci support 8 

gfichtenholt:
In this PR I added integration tests for flux plugin OCI support hosted on harbor CR. 
My testing showed that harbor supports Docker Registry HTTP API V2, like github, so I don't have to add any new code


cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/testdata/ghcr-util.sh
```diff
@@ -0,0 +1,191 @@
+#!/bin/bash
+
+# Copyright 2021-2022 the Kubeapps contributors.
```
antgamdia, line 3: Nitpick: technically, since the file has been created now, this is the proper header:

```suggestion
# Copyright 2022 the Kubeapps contributors.
```

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/testdata/harbor-util.sh
```diff
@@ -0,0 +1,91 @@
+#!/bin/bash
+
+# Copyright 2021-2022 the Kubeapps contributors.
```
antgamdia, line 3:
```suggestion
# Copyright 2022 the Kubeapps contributors.
```

antgamdia (approved):
Awesome, thanks for extracting some logic to the util files. It's great to hear that supporting ghcr also means supporting Harbor. Looking forward to giving it a try!

gfichtenholt:
Thanks Antonio

gfichtenholt:
> This PR is still in progress with lots of changes coming in, so adding this as comment here rather than creating a new issue: The current redis cache for helm OCI repos is broken. It loads the initial state into the cache just fine, but when the state on the remote changes, such as:
> 
>     * a new chart is pushed to repo
> 
>     * an existing chart is removed from repo
> 
>     * a new chart version (aka tag) is pushed
> 
>     * an existing tag is removed from a chart
> 
> 
> will result in out-of-sync cache. Basically any change on remote server, the cache will get out of sync with the remote. The reason is that fluxcd chose not to implement a feature where the caller gets notified when source on remote changes (ref [fluxcd/source-controller#839](https://github.com/fluxcd/source-controller/issues/839)). I understand the seriousness of the issue and I am currently considering different fixes for this, such as:
> 
>     * disable cache for helm OCI repositories
> 
>     * implement logic to poll remote for changes and react (i.e. what flux is already doing)
> 
>     * manual resync, including delete/create of helm OCI repo
> 
> 
> will update this PR as I narrow the solution down

The fix I ended up implementing is as follows:
There are 2 API calls that were possibly affected by the scenarios I described: `GetAvailablePackageSummaries()` and `GetAvailablePackageVersions()`. For those calls I have added code to compute the checksum from remote and compare it with what's cached on every call. If cached checksum does not match what's on remote, the cache entry will be recomputed at this time and an updated info will be cached. For a given repo checksum is composed of available chart names (packages) and versions for each chart (package). So, in total it adds 1 (list chart names) + N round-trips to remote, where N is the number of charts in a given repo. This is much cheaper that having to re-populate cache from scratch with all the information needed to compute chart model, so is better than having no cache

ppbaena:
Great solution Greg!

Issue #5121, opened 2022-07-25
Improve the UX of new Repositories API in the UI


antgamdia:
Once the new Repos API is being used in the UI (https://github.com/vmware-tanzu/kubeapps/issues/4764), there are several improvements and pending items to be performed.

This issue is intended to hold and sum up those pending items/open issues, so that we can track the overall status easily:

### Phase 2

- Issues already solved with the new UI
  - [x] https://github.com/vmware-tanzu/kubeapps/issues/153
  - [x] https://github.com/vmware-tanzu/kubeapps/issues/5200

- Issues to solve in a timely manner:
  - [x] Minor UI tweak if Flux were to support OCI
    - Tracked at: https://github.com/vmware-tanzu/kubeapps/issues/5007
  - [x] https://github.com/vmware-tanzu/kubeapps/issues/5201
  - [x] Improve docs adding info about the repos
    - Tracked at: https://github.com/vmware-tanzu/kubeapps/issues/5220  
  -  [x] Ensure the proper usage of the global namespace, move to `isGlobalNamespace(...` function when possible.
    - Tracked at: https://github.com/vmware-tanzu/kubeapps/issues/5202
  - [x]  `getPluginByName` is a hardcoded list of plugin objects for getting the object with the version set. Ideally, it should fetch it from the actually configured plugins.
    - Tracked at: https://github.com/vmware-tanzu/kubeapps/issues/5017
  - [ ] (**blocked** by the backend) Ensure the UI works as expected after stopping using config-level flags
    - Example: https://github.com/vmware-tanzu/kubeapps/issues/4870

- Not critical issues (rbac/multicluster) as of today:
  - [x] (multicluster) https://github.com/vmware-tanzu/kubeapps/issues/2717
  - [x] (rbac) https://github.com/vmware-tanzu/kubeapps/issues/4220
  - [x] (rbac) https://github.com/vmware-tanzu/kubeapps/issues/3679
  - [x] (rbac) https://github.com/vmware-tanzu/kubeapps/issues/5215

Further improvements will be tracked by issue #5515

PR #5217, opened 2022-08-17
Add types to the state used in tests

antgamdia:
### Description of the change

This PR simply adds types to the state variables used along the UI test suite. Sometimes, we add some properties (or we miss others) that don't really belong to the state type definition. Without a type declaration, we don't get any warning in the IDE. This PR is to fix this. 

### Benefits

Type-related warnings when adding wrong fields to the state vars used in the tests

### Possible drawbacks

N/A

### Applicable issues

- slightly related #5121 

### Additional information

N/A


castelblanque (approved):
Good, this will make building tests less error prone. Thanks!

Issue #5007, opened 2022-07-01
Support for OCI registries in Flux plugin


gfichtenholt:
This is a long-running issue I started a little over a month ago.
Flux CD added support for OCI registries recently
https://github.com/fluxcd/flux2/pull/2597

My work is to expose this functionality to kubeapps end user via flux plugin.

Areas that I worked on so far:
- Basic CRUD support for OCI registries, including caching of repos and charts in redis
  (like an index for regular helm repositories)
  U(pdate) implementation still pending.
- Introduce an API for repository lister (catalog). Implementation so far is
  docker api v2 compatible registry
  https://github.com/opencontainers/distribution-spec/blob/main/spec.md#api using
  [ORAS v2 libraries](https://github.com/oras-project/oras-go). Works for GitHub (ghcr.io), others TBD
  Full discussion here: https://github.com/vmware-tanzu/kubeapps/pull/4932. I stopped
  when I got blocked by https://github.com/oras-project/oras-go/issues/196 below. Once I verify it is fixed and will pick up
  where I left off

Areas which I have not started yet:
- Support other kinds of repositories, such as Harbor, Azure CR, AWS CR, etc.
- Support for k8s docker registry credentials (OCI registries may use them)
- Support for InsecureTLS flag
- unit tests. I wrote one integration test so far which is testing (back)end-2-end,
  because I wanted to see what flux engine really does on the server-side
  I know I need to do the client-side tests.

Currently outstanding issues that will require verification and/or code changes when fixed:
-  Add support for insecure registries in HelmRepositories of Type oci https://github.com/fluxcd/source-controller/issues/807
-  Flux fails to login to HTTP OCI registry where helm CLI logs in https://github.com/fluxcd/source-controller/issues/805
-  Support last parameter for listing catalog and tags  https://github.com/oras-project/oras-go/issues/196

I plan to be updating this issue as I make further progress with OCI support in flux plugin

ppbaena:
Hi Greg, could you update the current status of this issue and next steps planned?

gfichtenholt:
Current status: in progress.
Items left to do before I can mark this Closed:
- Update operation implementation for OCI repos + integration tests. 
  Time estimate: 2 days
- Production code cleanup (removing instrumentation via log.Info()). 
  Other tidying up: 2 days
- Integration tests: use cases involving TLS: 2 days
- Regular unit tests (i.e. when k8s isn't running). 
[...]

PR #5218, opened 2022-08-17
Changes to support OCI with flux in the UI

antgamdia:
### Description of the change

Since the OCI support in Flux was something in progress, we didn't allow adding OCI repos in the past release. As we are implementing this support, the UI should also support it.

In this PR I'm also squeezing a fix for a regression we've had: Flux repos couldn't be created as they need to be global. This was broken in the https://github.com/vmware-tanzu/kubeapps/pull/5203 hotfix.

Besides, this PR fixes several minor problems, namely:

- Rename to `setInterval` to `setSyncInterval` since there is a js function called setInterval and the IDE was triggering some warnings.
- Avoid re-auto-selecting storage type if the user already has selected one.
- Avoid overriding the scheme (`https://`, `oci://`, etc) with the default one except for the helm plugin. It was causing some issues with the Flux OCI repos.

### Benefits

See above.

### Possible drawbacks

N/A

### Applicable issues

- related #5007 

### Additional information

I haven't been able to fetch packages from an OCI repo yet IRL, but the repo CR is being properly created and used, according to the logs.

### DIFF
--- a/dashboard/src/actions/repos.ts
+++ b/dashboard/src/actions/repos.ts
@@ -139,12 +139,7 @@ export const addRepo = (
     } = getState();
     try {
       dispatch(addOrUpdateRepo());
-      // TODO(agamez): improve the global/namespaced repos UX
-      // to avoid hacks like this (flux repos are always global)
-      // https://github.com/vmware-tanzu/kubeapps/issues/2995
-      const namespaceScoped =
-        !isGlobalNamespace(namespace, request.plugin?.name, config) &&
-        request.plugin?.name !== PluginNames.PACKAGES_FLUX;
+      const namespaceScoped = !isGlobalNamespace(namespace, request.plugin?.name, config);
       const addPackageRepositoryResponse = await PackageRepositoriesService.addPackageRepository(
         currentCluster,
         namespace,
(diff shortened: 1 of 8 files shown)


dashboard/src/actions/repos.ts
```diff
@@ -139,7 +139,12 @@ export const addRepo = (
     } = getState();
     try {
       dispatch(addOrUpdateRepo());
-      const namespaceScoped = !isGlobalNamespace(namespace, request.plugin?.name, config);
+      // TODO(agamez): improve the global/namespaced repos UX
+      // to avoid hacks like this (flux repos are always global)
+      // https://github.com/vmware-tanzu/kubeapps/issues/2995
+      const namespaceScoped =
+        !isGlobalNamespace(namespace, request.plugin?.name, config) &&
+        request.plugin?.name !== PluginNames.PACKAGES_FLUX;
```
castelblanque, line 147: `isGlobalNamespace` has a switch per plugin.
Why not just returning `true` for the Flux case inside that function?

antgamdia: Yep, good point. I guess it makes sense for the `isGlobalNamespace` function to always return `true` for flux packages. I'll do it.

antgamdia: Done at https://github.com/vmware-tanzu/kubeapps/pull/5218/commits/3f8b65c8551ec387bcfdd9ee9fe7386d2813fe28

castelblanque (approved):
LGTM

PR #5222, opened 2022-08-19
Flux oci support 9: fix helm repository cache out-of-sync when remote contents changes

gfichtenholt:
see https://github.com/vmware-tanzu/kubeapps/pull/5212#issuecomment-1220342823 for description of the problem and the fix

Also renamed cache funcs for simplicity and to save time on typing:

- GetForOne => Get
- GetForMultiple => GetMultiple
- fetchForOne => fetch
- fetchForMultiple => fetchMultiple

All tests (integration+unit) passing


cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/oci_repo.go
```diff
@@ -380,23 +384,17 @@ func newRegistryClient(isLogin bool, tlsConfig *tls.Config, getterOpts []getter.
 // OCI Helm repository, which defines a source, does not produce an Artifact
 // ref https://fluxcd.io/docs/components/source/helmrepositories/#helm-oci-repository
 
-// TODO: this function is way too long. Break it up
 func (s *repoEventSink) onAddOciRepo(repo sourcev1.HelmRepository) ([]byte, bool, error) {
 	log.Infof("+onAddOciRepo(%s)", common.PrettyPrint(repo))
 	defer log.Info("-onAddOciRepo")
 
+	// I am disabling repo cache due to
```
antgamdia, line 391: Are you or is it just a leftover comment? It seems you are still using the `cacheEntryValue`, no?

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/oci_repo.go
```diff
@@ -526,29 +462,56 @@ func (s *repoEventSink) onModifyOciRepo(key string, oldValue interface{}, repo s
...
+		// use gob encoding instead of json, it peforms much better
+		var buf bytes.Buffer
+		enc := gob.NewEncoder(&buf)
+		if err = enc.Encode(cacheEntryValue); err != nil {
+			return nil, false, err
+		}
+
+		if s.chartCache != nil {
+			fn := downloadOCIChartFn(ociChartRepo)
+			if err = s.chartCache.SyncCharts(charts, fn); err != nil {
```
antgamdia, line 506: I guess this is still syncing by just adding any new versions, no? I mean, in the event of a chart deprecation, the package won't get removed from the cache, no?
If so, wouldn't the upstream checksum always differ from the cached one which may lead to a permanent cache-miss state? 

Not sure, just asking, though.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/testdata/ghcr-util.sh
```diff
@@ -59,7 +59,7 @@ function pushChartToMyGitHubRegistry() {
   -H "Accept: application/vnd.github+json" \
   /user/packages/container/helm-charts%2Fpodinfo/versions | jq -rc '.[].metadata.container.tags[]')
   echo
-  echo Remote Repository aka Package [${L_YELLOW}$GITHUB_OCI_REGISTRY_URL/podinfo${NC}] / All Versions 
+  echo -e Remote Repository aka Package [${L_YELLOW}$GITHUB_OCI_REGISTRY_URL/podinfo${NC}] / All Versions
```
antgamdia, line 62: Nitpick: `echo -e` [is not POSIX-compliant](https://www.unix.com/man-page/posix/1posix/echo/) (although we are using it a couple of times in the CI and it works). Just for the record, I'm ok with it, thoguh.

antgamdia (approved):
Great, thanks for adding this solution. Way better than just getting rid of the cache!
I just have a question (see above), but +1 anyway. Thanks!

antgamdia:
The CI fix is now out, once you merge latest `main`, the CI checks should be passing.
FYI, you can just click on the new "update branch" button in github:
image

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/oci_repo.go
gfichtenholt: just left over comment. WIll fix

gfichtenholt:
> Great, thanks for adding this solution. Way better than just getting rid of the cache! I just have a question (see above), but +1 anyway. Thanks!

Thank you for the review

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/oci_repo.go
antgamdia: Ah, ok, ok, got it. If we aren't looking at the charts in the local cache for computing it, then it doesn't seem an issue. Thanks for the clarification!

PR #5227, opened 2022-08-22
Fix olm test case

### LINKED ISSUE #5225: CI jobs failin because of an operator-related test case
**Describe the bug**
Since some days ago, every PR is failing due to a failing test case in the operator test suite.

**To Reproduce**
Run any CI job.

**Expected behavior**
No failing operator-related tests.

**Screenshots**
N/A

N/A

**Additional context**
N/A

antgamdia:
### Description of the change

It seems a new operator has been added and we are installing this new one instead of the expected one. See screenshot:
image.

The new operator is indeed failing, therefore, the test case is failing too.

```
 install failed: deployment ack-prometheusservice-controller not ready before timeout: deployment "ack-prometheusservice-controller" exceeded its progress deadline
```

We need to make the selector more specific so that we really select the one we are expecting.

### Benefits

CI will pass again.

### Possible drawbacks

N/A

### Applicable issues

- fixes #5225 

### Additional information

N/A

### DIFF
--- a/integration/tests/operators/06-operator-deployment.spec.js
+++ b/integration/tests/operators/06-operator-deployment.spec.js
@@ -19,7 +19,7 @@ test("Deploys an Operator", async ({ page }) => {
   // Select operator to deploy
   await page.locator("input#search").fill("prometheus");
   await page.waitForTimeout(3000);
-  await page.click('a:has-text("prometheus")');
+  await page.click('a:has-text("Manage the full lifecycle of configuring and managing Prometheus")');
   await page.click('cds-button:has-text("Deploy") >> nth=0');
   await page.click('cds-button:has-text("Deploy")');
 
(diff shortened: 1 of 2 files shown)


integration/tests/operators/06-operator-deployment.spec.js
```diff
@@ -19,7 +19,7 @@ test("Deploys an Operator", async ({ page }) => {
   // Select operator to deploy
   await page.locator("input#search").fill("prometheus");
```
antgamdia, line 20: Alternatively, we could just navigate to `/#/c/default/ns/default/operators?Search=prometheus&Provider=Red%20Hat`, however I didn't want to make more changes in the tests as I hope we move the operators to a plugin soon 😅

castelblanque: I would rather go with adding an instruction before to click the checkbox "Red Hat" in the providers filter.
With that and the search term, maybe it is less error prone? I assume it is less likely to happen that they change the Provider than the description, though.

castelblanque (approved):
LGTM

integration/tests/operators/06-operator-deployment.spec.js
antgamdia: Yep, fair point. However, it seems we also have multiple matchings with the provider itself:

image

But, sure, let's give it a try, I'll push a commit soon. CI passed with the current fix, so clearly, this was the reason behind the CI failure.

Issue #5214, opened 2022-08-16
Start/Stop feature


SohrabTa:
Implement a Start/Stop button in the app view that does the following:

### Functionality

When stopping: 
Save the current `values.yaml` of an app and in the background delete that app, but keep it in the app list with the status `stopped`. Gray out the `Stop` button when status is `stopped`.

When starting: 
Restore those values while continuing that app again and re-deploy it in the background. Gray out the `Start` button when status is `started`.

### To-dos

- Save the values.yaml of an app to a database outside of the K8s cluster when stopping an app (so it can be recovered, after the pod has been down)
- Keep the stopped app in the app list and switch its status to `stopped`
- Pair every values.yaml to the installed apps (is there a unique identifier of an installation?)
- Recover the values.yaml from the database and deploy the stopped app with those values again, switch its status to `started`

### Why
Empower users to adjust relevant parameters and stop and start the apps with their own configuration on-demand to save energy and costs.

castelblanque:
hi @SohrabTa , thanks for the proposal. It is really appreciated that the community takes part in the development of Kubeapps.
It is an interesting feature, but sadly we can not include it in the roadmap for this quarter.
Let's use this issue to discuss the proposal meanwhile.

Have you considered other alternatives? For example KNative's `KnativePodAutoscaler` that allows [configuring scale to zero](https://knative.dev/docs/serving/autoscaling/scale-to-zero/), or Keda's feature for [scaling deployments and statefulsets](https://keda.sh/docs/2.8/concepts/scaling-deployments/).

ppbaena:
The maintainers' team discussed this proposal during the planning meeting today. It was agreed not to include this feature in Kubeapps. I hope you can keep sharing such interesting ideas with the maintainers' team in the future to improve Kubeapps. Thanks for your contribution!

PR #5229, opened 2022-08-22
Start/Stop feature #5214 (proposal)

SohrabTa:
Implement a Start/Stop button in the app view that does the following:

### Functionality

When stopping: 
Save the current `values.yaml` of an app and in the background delete that app, but keep it in the app list with the status `stopped`. Gray out the `Stop` button when status is `stopped`.

When starting: 
Restore those values while continuing that app again and re-deploy it in the background. Gray out the `Start` button when status is `started`.

### To-dos

- Save the values.yaml of an app to a database outside of the K8s cluster when stopping an app (so it can be recovered, after the pod has been down)
- Keep the stopped app in the app list and switch its status to `stopped`
- Pair every values.yaml to the installed apps (is there a unique identifier of an installation?)
- Recover the values.yaml from the database and deploy the stopped app with those values again, switch its status to `started`

### Why
Empower users to adjust relevant parameters and stop and start the apps with their own configuration on-demand to save energy and costs.


antgamdia:
Please note that some members of the team are still on PTO 🌴 , so please expect some delays in an official response. Anyway, thank you for the contribution and the proposal. Please also feel free to join [our bi-weekly iteration planning sessions](https://kubernetes.slack.com/files/U91HAPE5R/F01EGHESH4L/kubeapps-iteration-planning.ics)!

antgamdia:
I have transferred this PR to a GitHub discussion, which fits better with the actual purpose: https://github.com/vmware-tanzu/kubeapps/discussions/5316

_We are still polishing up the best way to discuss proposals and ideas, sorry for the inconvenience_ 😅

PR #5239, opened 2022-08-26
Migrate namespaces retrieval from Kubeops to Kubeapps APIs

### LINKED ISSUE #4785: Remove/migrate UI calls to Kubeops

castelblanque:

### Description of the change

Migration of the namespaces list retrieval from Kubeops to Kubeapps APIs. It resides now in the resources plugin, under the endpoint `GetNamespaceNames`. Frontend has been changed to use the new endpoint too.

image

One affected functionality is the one for trusted namespaces (#2671/#2506). Chart values for specifying header name and pattern have been moved into the resources plugin configuration section.

### Benefits

We are one step closer to deprecating Kubeops.

### Possible drawbacks

Backwards compatibilty issue for adopters using the trusted namespaces functionality from headers.

### Applicable issues

- partially resolves #4785


cmd/kubeops/cmd/root.go
```diff
@@ -52,8 +52,6 @@ func setFlags(c *cobra.Command) {
 	c.Flags().StringVar(&serveOpts.PinnipedProxyCACert, "pinniped-proxy-ca-cert", "", "Path to certificate authority to use with requests to pinniped-proxy service")
 	c.Flags().IntVar(&serveOpts.Burst, "burst", 15, "internal burst capacity")
 	c.Flags().Float32Var(&serveOpts.Qps, "qps", 10, "internal QPS rate")
-	c.Flags().StringVar(&serveOpts.NamespaceHeaderName, "namespace-header-name", "", "name of the header field, e.g. namespace-header-name=X-Consumer-Groups")
-	c.Flags().StringVar(&serveOpts.NamespaceHeaderPattern, "namespace-header-pattern", "", "regular expression that matches only single group, e.g. namespace-header-pattern=^namespace:([\\w]+):\\w+$, to match namespace:ns:read")
```
antgamdia, line 56: In cobra, when a non-recognized flag is passed, it errors. I'd rather deprecate the flags for now instead of deleting them.

See lines below.

pkg/kube/kube_handler.go
```diff
@@ -312,7 +307,6 @@ type handler interface {
 	UpdateAppRepository(appRepoBody io.ReadCloser, requestNamespace string) (*v1alpha1.AppRepository, error)
 	RefreshAppRepository(repoName string, requestNamespace string) (*v1alpha1.AppRepository, error)
 	DeleteAppRepository(name, namespace string) error
```
antgamdia, line 309: just curious: do you know if we can remove more logic from this file? If so, do you plan to do so as part of the kubeops deprecation? Or should we file another issue for it?

antgamdia (approved):
Really happy to see we are removing old code from the `pkg/kube` thing and moving to the place it really belongs now: a plugin in the kubeapps-apis server! Thanks

pkg/kube/kube_handler.go
castelblanque: Exactly, from the `pkg/kube` we can remove a lot of stuff. That will be done after removing Kubeops in a subsequent step.

cmd/kubeapps-apis/plugins/resources/v1alpha1/common/plugin.go
```diff
@@ -0,0 +1,61 @@
...
+						HeaderName    string `json:"headerName"`
+						HeaderPattern string `json:"headerPattern"`
+					} `json:"trustedNamespaces"`
+				} `json:"v1alpha1"`
+			} `json:"packages"`
+		} `json:"resources"`
+	}
+	var config resourcesConfig
+
+	// #nosec G304
```
beni0888, line 44: Hey @castelblanque, just for the sake of learning, I've never seen this kind of comment/annotation before, and I guess it should be processed by some kind of security linter/checker. Could you tell me what's the tool in charge of this and where is its configuration?

cmd/kubeapps-apis/plugins/resources/v1alpha1/common/plugin_test.go
```diff
@@ -0,0 +1,97 @@
...
+			// TODO(agamez): env vars and file paths should be handled properly for Windows operating system
+			if runtime.GOOS == "windows" {
+				t.Skip("Skipping in a Windows OS")
+			}
+			filename := ""
+			if tc.pluginYAMLConf != nil {
+				pluginJSONConf, err := yaml.YAMLToJSON(tc.pluginYAMLConf)
+				if err != nil {
+					log.Fatalf("%s", err)
+				}
```
beni0888, line 73: Haven't you considered the use of `github.com/stretchr/testify` library (which I've checked is already included in the project) to ease the writing and reading of tests? This way you could replace these three lines with just a oneliner like `require.NotNil(t, err)`

cmd/kubeapps-apis/plugins/resources/v1alpha1/namespaces_filtering.go
```diff
@@ -0,0 +1,184 @@
...
+		namespaceList = append(namespaceList, trustedNamespaces...)
+	} else {
+
+		typedClient, _, err := s.clientGetter(ctx, cluster)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "unable to get the k8s client: '%v'", err)
+		}
+
+		// Try to list namespaces with the user token, for backward compatibility
+		namespaces, err := typedClient.CoreV1().Namespaces().List(context.TODO(), metav1.ListOptions{})
```
beni0888, line 50: I don't understand why `context.TODO()` is used here. Shouldn't we pass the parent function's context `ctx` or `context.Background()` in case we want a totally new context?

cmd/kubeapps-apis/plugins/resources/v1alpha1/namespaces_filtering.go
```diff
@@ -0,0 +1,184 @@
...
+		if err != nil {
+			if k8sErrors.IsForbidden(err) {
+				backgroundCtx := context.Background()
```
beni0888, line 53: Also, here we're not being consistent with previous uses and are using `context.Background()`, and below we resort to `context.TODO()` again. Also, I guess you could get rid of `backgroundCtx` variable as it's only used in one single place, so it could be inlined.

cmd/kubeapps-apis/plugins/resources/v1alpha1/common/plugin.go
antgamdia: This is a `gosec` annotation we use for ignoring false positives and wontfixes. See https://github.com/securego/gosec#annotating-code

cmd/kubeapps-apis/plugins/resources/v1alpha1/common/plugin_test.go
castelblanque: Definitely, I'll try to remember next time. In many cases, this block of code is just copy-pasted.
Thanks!

cmd/kubeapps-apis/plugins/resources/v1alpha1/namespaces_filtering.go
castelblanque: The auth token of the request is actually handled in the `s.clientGetter` func of line 44. It is, in turn, handled [here](https://github.com/vmware-tanzu/kubeapps/blob/main/cmd/kubeapps-apis/core/plugins/v1alpha1/plugins.go#L319) based on the request's context.
Commenting here your post below: parts of the code are legacy from Kubeops so we inherit some stuff, but agree with you that we should try to harmonize the usage of `context.Background()` and `context.TODO()`. The two are mainly the same (see the [contexts definition here](https://go.dev/src/context/context.go#L199)), it is regarding semantics.

Go docs say that `Code should use context.TODO when it's unclear which Context to use or it is not yet available (because the surrounding function has not yet been extended to accept a Context parameter).`

[Documentation here](https://pkg.go.dev/context#Background).

beni0888: I see, thanks for the explanation 🙂

PR #5240, opened 2022-08-29
Add explicit scope to the new repo form

### LINKED ISSUE #2995: No obvious way to create global app repository
### Description:

Currently if you click on the option to add an app repository, you are presented with:

create-repos

showing both global app repositories and those for the current namespace (if any). It is completely non-obvious how you create a global repository. The message displayed there used to encourage the user to select "All Namespaces" for the old namespace selector, but the concept of "All namespaces" was removed when we moved to the context selector, so now you need to know to select the "kubeapps" (or namespace in which Kubeapps is installed) to then add a global namespace.

Originally reported by Scott from customer reports.

antgamdia:
### Description of the change

Adding an explicit scope for selecting whether the repo should be namespace or global + some info messages to the user.

### Benefits

Adding a global repo will no longer require switching the context, which wasn't something trivial to figure out as a user.

### Possible drawbacks

N/A

### Applicable issues

- fixes #2995 

### Additional information

reposWithScope

Squeezing two changes:
- Disable the description field if using carvel repos (as it currently errors if any is passed)
- Fix some <label> tags to comply with a11y recommendations.

### DIFF
--- a/dashboard/src/components/Config/PkgRepoList/PkgRepoForm.tsx
+++ b/dashboard/src/components/Config/PkgRepoList/PkgRepoForm.tsx
@@ -32,6 +32,7 @@ import { IConfig } from "shared/Config";
 import { toFilterRule, toParams } from "shared/jq";
 import { IPkgRepoFormData, IPkgRepositoryFilter, IStoreState } from "shared/types";
 import {
+  getGlobalNamespaceOrNamespace,
   getPluginByName,
   getPluginPackageName,
   getSupportedPackageRepositoryAuthTypes,
@@ -82,6 +83,11 @@ export function PkgRepoForm(props: IPkgRepoFormProps) {
     clusters: { currentCluster },
   } = useSelector((state: IStoreState) => state);
 
+  const currentNsConfig = {
+    globalReposNamespace: globalReposNamespace,
+    carvelGlobalNamespace: carvelGlobalNamespace,
+  } as IConfig;
+
   const initialInterval = "10m";
 
   // -- Auth-related variables --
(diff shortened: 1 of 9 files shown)


ppbaena:
Nice to see this UX improvement when configuring repositories. Thanks @antgamdia!

dashboard/src/components/Config/PkgRepoList/PkgRepoForm.tsx
```diff
@@ -637,6 +676,7 @@ export function PkgRepoForm(props: IPkgRepoFormProps) {
                     placeholder="Description of the repository"
                     value={description || ""}
                     onChange={handleDescriptionChange}
+                    disabled={plugin?.name === (PluginNames.PACKAGES_KAPP as string)}
```
antgamdia, line 679: Disabling description if the plugin is kapp

dashboard/src/components/Config/PkgRepoList/PkgRepoForm.tsx
```diff
@@ -686,143 +732,210 @@ export function PkgRepoForm(props: IPkgRepoFormProps) {
...
+                      {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
+                      <label>Scope:</label>
+                      <CdsControlMessage>
+                        If the repository is global, its packages will be availabe in every
+                        namespace of the cluster. If it is namespaced-scoped, its packages will be
+                        available only in the namespace where the repository has been installed.
+                      </CdsControlMessage>
+
+                      {repo.namespaceScoped ||
+                        (isGlobalNamespace(namespace, plugin?.name, currentNsConfig) && (
```
antgamdia, line 746: Using either the field from the repo object (editing the existing repo) or calculating if the ns is global or not.

dashboard/src/shared/utils.ts
```diff
@@ -235,3 +235,21 @@ export function isGlobalNamespace(namespace: string, pluginName: string, kubeapp
       return false;
   }
 }
+
+export function getGlobalNamespaceOrNamespace(
```
antgamdia, line 239: This should belong to each plugin's config

castelblanque (approved):
LGTM

PR #5243, opened 2022-08-30
Disable scope field when editing a repo

antgamdia:
### Description of the change

A quick follow-up PR of #5240 just for disabling the field when editing a repo.
The rest of the changes are just trivial: using `includes` instead of chaining `||`  when possible and avoid using the `?.` operator if the field is not optional.

### Benefits

image

### Possible drawbacks

N/A

### Applicable issues

- related #5121 

### Additional information

N/A

### DIFF
--- a/dashboard/src/components/Config/PkgRepoList/PkgRepoForm.tsx
+++ b/dashboard/src/components/Config/PkgRepoList/PkgRepoForm.tsx
@@ -654,7 +654,7 @@ export function PkgRepoForm(props: IPkgRepoFormProps) {
                     required={true}
                     pattern={k8sObjectNameRegex}
                     title="Use lower case alphanumeric characters, '-' or '.'"
-                    disabled={!!repo?.name}
+                    disabled={!!repo.name}
                   />
                 </CdsInput>
                 <CdsInput>
@@ -676,7 +676,9 @@ export function PkgRepoForm(props: IPkgRepoFormProps) {
                     placeholder="Description of the repository"
                     value={description || ""}
                     onChange={handleDescriptionChange}
-                    disabled={plugin?.name === (PluginNames.PACKAGES_KAPP as string)}
+                    disabled={(
+                      [PluginNames.PACKAGES_FLUX, PluginNames.PACKAGES_KAPP] as string[]
+                    ).includes(plugin?.name)}
                   />
                 </CdsInput>
                 {/* TODO(agamez): these plugin selectors should be loaded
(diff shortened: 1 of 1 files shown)


dashboard/src/components/Config/PkgRepoList/PkgRepoForm.tsx
```diff
@@ -767,7 +769,8 @@ export function PkgRepoForm(props: IPkgRepoFormProps) {
                           checked={repo.namespaceScoped || isNamespaceScoped}
                           onChange={handleRepoScopeChange}
                           disabled={
-                            !!repo && isGlobalNamespace(namespace, plugin?.name, currentNsConfig)
+                            !!repo.name ||
+                            isGlobalNamespace(namespace, plugin?.name, currentNsConfig)
```
antgamdia, line 773: This is the actual change, the condition was wrong 🤦 . We want to disable the field either if the current namespace is already global (so we don't have much to choose) OR if there is a repo being edited.

dashboard/src/components/Config/PkgRepoList/PkgRepoForm.tsx
```diff
@@ -797,8 +801,9 @@ export function PkgRepoForm(props: IPkgRepoFormProps) {
                       {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
                       <label>Package Storage Type</label>
                       <CdsControlMessage>Select the package storage type.</CdsControlMessage>
-                      {(plugin?.name === (PluginNames.PACKAGES_HELM as string) ||
-                        plugin?.name === (PluginNames.PACKAGES_FLUX as string)) && (
+                      {(
+                        [PluginNames.PACKAGES_HELM, PluginNames.PACKAGES_FLUX] as string[]
+                      ).includes(plugin?.name) && (
```
castelblanque, line 806: [Not related to this PR] If in the future we want a really decoupled frontend, plugins themselves should provide this kind of information (e.g. some endpoint informing that Kapp allows [A,B,C] as package storage types).

castelblanque: ..or implementing the plugins customizable UI proposal 😄

castelblanque (approved):
LGTM, thanks!

dashboard/src/components/Config/PkgRepoList/PkgRepoForm.tsx
castelblanque: Right. If the context namespace is already global, the user does not have a way to select a namespace. That is done through the context dialog.

dashboard/src/components/Config/PkgRepoList/PkgRepoForm.tsx
antgamdia: 100% agree with the comment. At least, we could start extracting each param to a separate fn like a `getPluginsSupporting(foo)` that, given an input id, returns the list of plugins in which this `id` should be rendered.

PR #5248, opened 2022-08-31
Fixes for carvel repo UI

### LINKED ISSUE #5244: carvel repository UI: several issues with carvel support in creation dialog
I did some tests of the new repository creation dialog UI, and i found a couple issues for carvel support:
 - the "inline" type is not supported and should not be listed (or greyed out) for creation
 - the auth options are enabled regardless of the type when they are type specific (for example SSH is only supported by GIT type, HTTP only supports basic auth...)
 - in the advanced section, only the duration is supported, the other fields should be removed

for duration, the description says it is "goland" format. I don't thin this should be described that way, instead it is the time/unit duration format supported by the k8s api spec (it does not really matter whether they modeled it after goland or not).

antgamdia:
### Description of the change

As described at #5244 there were several remaining issues with the new repos UI form for the Carvel plugin. This PR includes several fixes, namely:

- Disable `inline` format 
- Disable the rest of the advanced params
- Make getSupportedPackageRepositoryAuthTypes type-aware + selecting the appropriate ones for each carvel type
- Reword golang duration msg referring to Kubernetes

Additionally:
- Fix some typos 
- Move exported enums to `./shared/types`
- Rename css class 
- Rename some conditions 

### Benefits

Better UX when adding repos with the Carvel plugin

### Possible drawbacks

N/A

### Applicable issues

- fixes #5244

### Additional information

auths

image


castelblanque (approved):
Great! LGTM

Issue #5253, opened 2022-08-31
Remove Kubeops


castelblanque:
**Summary**
After #5239 and #5249, Kubeops is effectively disabled.
No endpoints are served from it, or resources used.
It can be safely removed without affecting anybody.

**Description**
Given that Kubeops is now useless and a burden, let's completely remove it.

**Acceptance criteria**
Kubeops is completely removed from Kubeapps

**Additional context**
Strongly related to #4785

There might be additional work to be done on the Bitnami side to remove the generation of Kubeops assets, tests, etc.

antgamdia:
Also, we have to remove the old API endpoints from the `openapi.yaml` file. This is still a manual task because we needed to add th kubeops, assetsvc endpoints to the automatically generated swagger doc from the grpc. Now, it should be way easier to do it once we remove all the legacy stuff!

PR #5254, opened 2022-08-31
Disable Kubeops by default

castelblanque:
### Description of the change

Disables Kubeops by default in Kubeapps chart.
Effectively, Kubeops is completely disabled, even in E2E.
A subsequent PR will completely remove it from the project.

### Benefits

Kubeops is not deployed by default with Kubeapps. Not even with Operators enabled.

### Possible drawbacks

Applicable ones from #5249

### Applicable issues

- related to PR #5239
- related to PR #5249
- related to #4785
- related to #5253


antgamdia (approved):
Awesome!! That's really good news we can finally disable it!

PR #5255, opened 2022-08-31
Add filter by plugin in the catalog

### LINKED ISSUE #4097: Add a "filter by plugin" in the Catalog view

antgamdia:
### Description of the change

This is a replacement of the old https://github.com/vmware-tanzu/kubeapps/pull/4100, but using the getConfiguredPlugins instead.
Basically, this is adding a new "filter by plugin" option in the catalog.

### Benefits

It will become easier to look up for pkgs in the catalog.

### Possible drawbacks

More complexity when it comes to filtering perhaps? 

### Applicable issues

- fixes #4097

### Additional information

image


dashboard/src/components/Catalog/Catalog.tsx
```diff
@@ -235,7 +237,10 @@ export default function Catalog() {
     repos
       .filter(r => !r.namespaceScoped || r.packageRepoRef?.context?.namespace === namespace)
       .map(r => r.name),
-  );
+  ).sort();
+  const allPlugins = uniq(
+    configuredPlugins.filter(p => p.name.includes(".packages")).map(p => getPluginName(p) || ""),
```
castelblanque, line 242: Is this `.packages` plugin name a convention?
Perhaps in the future the getConfiguredPlugins API could return a specific field determining the type of plugin.

dashboard/src/components/Catalog/Catalog.tsx
```diff
@@ -483,6 +493,19 @@ export default function Catalog() {
                   />
                 </div>
               )}
+              {allPlugins.length > 0 && (
+                <div className="filter-section">
+                  <label className="filter-label">Package Type</label>
+                  <FilterGroup
+                    name={filterNames.PKG_TYPE}
+                    options={allPlugins}
```
castelblanque, line 501: Perhaps we should differentiate the plugin name from the package type label. With this, we are listing a friendly name of active plugins, e.g.: 

- Helm
- Kapp
- Flux

But the control is for `Package Type`. There is no "Flux" packaging format per se, it is a Helm chart.
So packages types would be slightly different from plugin names:
- Helm
- Kapp (or Carvel)
- Helm via Flux

WDYT?

castelblanque (approved):
LGTM, my only concern is about the "Package types" labels.
Thanks!

dashboard/src/components/Catalog/Catalog.tsx
antgamdia: Yes, currently all the plugins are using this `packaging` prefix, so it's safe to use it here (but it's an ugly workaround, though).
As we are adding new plugins (thinking now of the RBAC one we discussed), we may want to add a `type` field, good point; we should do it.

dashboard/src/components/Catalog/Catalog.tsx
antgamdia: Good idea. Initially I used "Packaging Plugin" as the filter name, hence the names I chose. But, 100% agree, now we're using "Package Type", we better use this existing function:

https://github.com/vmware-tanzu/kubeapps/blob/88030f61605dab0fd176ae14d4ef22b2b29438a6/dashboard/src/shared/utils.ts#L125-L133

See the result:

image

PR #5256, opened 2022-08-31
Completely remove Kubeops

### LINKED ISSUE #5253: Remove Kubeops

castelblanque:
### Description of the change

Completely remove Kubeops from the repo.
Only to be done after #5254

### Benefits

Kubeapps repo is lighter, and less confusion on Kubeapps architecture.

### Possible drawbacks

N/A

### Applicable issues

- resolves #5253


dashboard/src/shared/Auth.test.ts
```diff
@@ -279,7 +279,7 @@ describe("is403FromAuthProxy", () => {
     expect(
       Auth.is403FromAuthProxy({
         status: 403,
-        data: 'namespaces is forbidden: User "system:serviceaccount:kubeapps:kubeapps-internal-kubeops" cannot list resource "namespaces" in API group "" at the cluster scope',
+        data: 'namespaces is forbidden: User "system:serviceaccount:kubeapps:kubeapps-internal-kubeapps-apis" cannot list resource "namespaces" in API group "" at the cluster scope',
```
beni0888, line 282: It seems there's a typo in either this or the following editions in this file. Note that we are using `kubeapps-apis` (with a hyphen) here, while we're using `kubeappsapis` in the rest of the occurrences.

beni0888 (approved):
LGTM! I've just pointed out a tiny detail it would be nice to fix. Great work!!

dashboard/src/shared/Auth.test.ts
castelblanque: Good catch! It is a frontend test that actually checks if [data points to a `system:serviceaccount`](https://github.com/vmware-tanzu/kubeapps/blob/03eb5924f4b82a7c1eb4521b550f5d269bf5ed40/dashboard/src/shared/Auth.ts#L118), so what comes after that is not used. But I'll update it to be consistent. Thanks.

PR #5273, opened 2022-09-01
Fix for E2E Carvel test 04 selectOption

### LINKED ISSUE #5267: E2E test 04 for Carvel fails many times across different branches
**Describe the bug**
In many CI runs, all retries for E2E test 04 (Carvel deployment) fail with:

```
1 failed
    [chromium] › tests/carvel/04-default-deployment.spec.js:8:1 › Deploys package with default values in main cluster
```

Tests steps involved from [Carvel's 04 test file](https://github.com/vmware-tanzu/kubeapps/blob/main/integration/tests/carvel/04-default-deployment.spec.js) are:

```
  await page.waitForSelector('select[name="package-versions"]');
  await page.selectOption('select[name="package-versions"]', "2.0.0");
  const releaseNameLocator = page.locator("#releaseName");
  await releaseNameLocator.waitFor();
  await expect(releaseNameLocator).toHaveText("");
  const releaseName = utils.getRandomName("test-04-release");
  await releaseNameLocator.fill(releaseName);
```

So version 2.0.0 should be set before setting the release name and submitting the deployment.
But it is not the case, as shown in the video attached to CircleCI:

image

Therefore value set by the version's `selectOption` is not updated and backend receives the wrong version value.

Error message shown in the Kubeapps UI is:

```
An error occurred: Unable to create the installed package for the package "unknown/pkg.test.carvel.dev" using the plugin "kapp_controller.packages": rpc error: code = InvalidArgument desc = Unable to create the PackageInstall 'test-04-release-60601' due to 'rpc error: code = InvalidArgument desc = The selected version "3.0.0-rc.1" is not elegible to be installed: Expected to find at least one version, but did not (details: all=1 -> after-prereleases-filter=0 -> after-constraints-filter=0)'
```

**To Reproduce**

Run the E2E tests in CI just pushing a branch or re-running an existing workflow in CircleCI.

**Expected behavior**
All E2E tests pass.

**Screenshots**
image

**Additional context**

It might be related to [Playwright issue #4921](https://github.com/microsoft/playwright/issues/4921)

castelblanque:

### Description of the change

Trivial PR trying to fix the error appearing with E2E test 04 for Carvel.
It is the typical error on E2E CI that fails from time to time, but I noticed that the frequency has been increasing lately.

### Benefits

CI pipelines should pass successfully more often.

### Possible drawbacks

That this PR does not really fix the issue.

### Applicable issues

- fixes #5267

### DIFF
--- a/integration/tests/carvel/04-default-deployment.spec.js
+++ b/integration/tests/carvel/04-default-deployment.spec.js
@@ -24,7 +24,8 @@ test("Deploys package with default values in main cluster", async ({ page }) => 
 
   // Deploy package
   await page.waitForSelector('select[name="package-versions"]');
-  await page.selectOption('select[name="package-versions"]', "2.0.0");
+  const versionSelector = await page.$('select[name="package-versions"]')
+  await versionSelector?.selectOption("2.0.0")
   const releaseNameLocator = page.locator("#releaseName");
   await releaseNameLocator.waitFor();
   await expect(releaseNameLocator).toHaveText("");


integration/tests/carvel/04-default-deployment.spec.js
```diff
@@ -24,7 +24,8 @@ test("Deploys package with default values in main cluster", async ({ page }) =>
 
   // Deploy package
   await page.waitForSelector('select[name="package-versions"]');
-  await page.selectOption('select[name="package-versions"]', "2.0.0");
+  const versionSelector = await page.$('select[name="package-versions"]')
+  await versionSelector?.selectOption("2.0.0")
```
antgamdia, line 28: I think (not sure though) you can use the page.locator (which should auto-wait) and then use the `selectOption`, like

````js
  await page.locator("select[name="package-versions"]").selectOption("2.0.0");
````

But, if your fix is proven to work, +1. Thanks for fixing it, it was generating some headaches recently!

castelblanque: The underlying problem is not that the `select` control is not found, but that the option selected with `selectOption` was not being applied. Analyzing the failing tests dump shows that Kubeapps backend was receiving version `3.0.0-rc.1` (default one).

Let's see if this consistently fixes the problem, otherwise we'll have to try other approaches.

PR #5294, opened 2022-09-02
Remove io/ioutils imports

antgamdia:
### Description of the change

The latest golangci-lint version warned us about a deprecation in go: the `io/ioutils`  modules have been moved to `io` and `os` packages. This PR is removing this dependency in favor of each `io` and `os` functions.

### Benefits

No deprecated code in our codebase.

### Possible drawbacks

N/A (some fn signatures might have changed, I'll perform the requires changes as well)

### Applicable issues

-  related https://github.com/vmware-tanzu/kubeapps/pull/5290 (in fact, this PR has to be merged first, otherwise the linter will fail)

### Additional information
N/A

### DIFF
--- a/cmd/asset-syncer/server/utils.go
+++ b/cmd/asset-syncer/server/utils.go
@@ -12,7 +12,6 @@ import (
 	"fmt"
 	"image"
 	"io"
-	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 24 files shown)


castelblanque (approved):
Thanks!

PR #5297, opened 2022-09-02
Handle grpc authentication errors

### LINKED ISSUE #5295: Handle authentication errors in grpc calls
**Describe the bug**
The removal of `kubeops` brought an unintended side effect: we are no longer able to redirect users back to the login page in the event of an authentication error.

Even if the vast majority of our calls were already migrated to kubeapps-apis, we still had some of them handled by kubeops. Those calls (+the error handling) were being performed by the `axios` interceptor in the frontend. Now, every call comes from the grpc client.

We have to intercept the auth-related errors to force a programmatic logout. Otherwise, invalid (or even outdated) tokens will remain forever unless a user manually deletes their browser's localstorage.

**To Reproduce**

Assuming a token-based installation, for the sake of simplicity:

1. Use a valid token for login into kubeapps
2. Replace (via browser dev tools) the token in the localstorage
3. Refresh the view, or navigate to another view, etc
4. The error will appear and NO redirection is being performed.

**Expected behavior**
Forced logout when an auth error happens.

**Screenshots**

The UI shows:

image

DEVEL

**Additional context**

N/A

antgamdia:
### Description of the change

As reported at #5295, the axios interceptor handling the logout does not work anymore (bc all the requests are handled by the grpc client).
This PR is to intercept the auth-related errors to force a programmatic logout.

### Benefits

If an auth error happens, we will log the user out.

### Possible drawbacks

Maybe we don't want to kick the user out in some points, but we can fine-tune it later.

### Applicable issues

- fixes #5295 

### Additional information

Problem: I've found some grpc calls NOT returning the proper status code... this forced me to add some ugly workarounds looking for certain keywords to appear to detect a true auth error. 

Demo: 

errorsAuth

WIP as there are still several test cases to fix


dashboard/src/actions/installedpackages.ts
```diff
@@ -128,17 +130,25 @@ export function getInstalledPackage(
...
-              "this package has missing information, some actions might not be available.",
+        if (e.constructor === UnauthorizedError) {
+          dispatch(actions.auth.logoutByAuthenticationError());
+        } else {
+          dispatch(
+            errorInstalledPackage(
+              new FetchWarning(
+                "this package has missing information, some actions might not be available.",
+              ),
             ),
```
castelblanque, line 141: There is `actions.auth.logoutByAuthenticationError()` repeated in many places.
I was wondering if we could refactor that to be generic with something like e.g.:

```
export function handledErrorAction(err: ActionType<any>, e: any){
  if (e.constructor === UnauthorizedError) {
    return actions.auth.logoutByAuthenticationError();
  } else {
     return err;
  }
}
```

So that invocations can be almost a one-liner like e.g.:
```
dispatch(
      handledErrorAction(errorInstalledPackage(new FetchError("Unable to list apps", [e])), e)
)
```

antgamdia: Good idea, thanks! Will do! 
Also, I'd like to explore the grpc interceptors at some point. Don't know if they can fit our main goal here, though. https://grpc.io/blog/grpc-web-interceptor/

dashboard/src/shared/utils.ts
```diff
@@ -276,3 +286,36 @@ export function getGlobalNamespaceOrNamespace(
...
+      } else {
+        return new InternalServerError(msg);
+      }
+    case grpc.Code.PermissionDenied:
+      return new ForbiddenError(msg);
+    case grpc.Code.NotFound:
+      return new NotFoundError(msg);
+    case grpc.Code.AlreadyExists:
+      return new ConflictError(msg);
+    case grpc.Code.InvalidArgument:
```
castelblanque, line 309: `InvalidArgument` does not like an internal server error? Maybe an error like e.g. `UnprocessableEntity` fits better?

dashboard/src/shared/utils.ts
```diff
@@ -276,3 +286,36 @@ export function getGlobalNamespaceOrNamespace(
...
+
+export function convertGrpcAuthError(e: any): CustomError | any {
+  const msg = e?.metadata?.headersMap?.["grpc-message"].toString();
+  switch (e?.code) {
+    case grpc.Code.Unauthenticated:
+      return new UnauthorizedError(msg);
+    case grpc.Code.FailedPrecondition:
+    case grpc.Code.Internal:
+      //TODO(agamez): this code shouldn't be returned by the API, but it is
+      if (["credentials", "unauthorized"].some(p => msg?.toLowerCase()?.includes(p))) {
```
castelblanque, line 298: A bit weak, but I guess it is the only thing we can do until all error codes from backend are harmonized.

dashboard/src/shared/utils.ts
antgamdia: It is not, of course. This is tricky, as we aren't really actually using those error objects... so I just defaulted them to "server error" with no specific criteria.
Let's create error objects grouping some of them, as we may want to have them in the future. See what you think

castelblanque: Agree with creating error objects that extend `CustomError` so that we can have some groups.

castelblanque (approved):
LGTM, thanks!

antgamdia:
To investigate: using goharbor.com chart repo+ credentials + kicks me out, verify which grpc code we are returning when adding repos.

castelblanque:
Tested the case, and it is returning a GRPC status 9 (Failed precondition).

```
grpc-message: Unable to add package repository "asdasda" using the plugin "helm.packages": rpc error: code = FailedPrecondition desc = Failed repository validation: &{401 {"errors":[{"code":"UNAUTHORIZED","message":"UnAuthorized"}]}%0A}
grpc-status: 9
```

That wraps the HTTP 401 returned by Harbor, indeed.

Maybe we need to process the 401 error in the backend and map it correctly. If it is a 401 that Harbor sends, I would return a grpc status 7 (Permission denied) or 3 (Invalid argument).
However, the `PERMISSION_DENIED` status could be confused with the user not having permissions to perform the operation itself in Kubeapps APIs, which is not the case. We might want to return instead the grpc status 3 (Invalid argument) and specify a more readable message text.

Issue #5219, opened 2022-08-18
Unable to use Flux OCI with Harbor robot accounts


antgamdia:
**Summary**
I've been adding minor changes for the UI to also support the OCI registries with Flux (PR #5218), in doing so, I'm kind of unable to add one (harbor) and get the packages.

**Background and rationale**

It's important as in many Harbor deployments we don't have full admin access, but maybe admin permissions over a single project (not the whole harbor instance). Using robot accounts seems to be the preferred way to consume TAC/VAC, therefore, I guess it's important that we also support this way to enable VAC users to deploy their apps using kubeapps+flux.

**Description**

In the logs I see:

```
ORAS .Repositories() returned err: GET "https://harbor-repo.vmware.com/v2/_catalog?last=agamez": unexpected status code 401: unauthorized: unauthorized to list catalog: unauthorized to list catalog
```

I assume that this is because the robot accounts are not privileged enough (see https://github.com/goharbor/harbor/issues/8723#issuecomment-730248740).

However, the following endpoint seems to work with robot accounts and does give me the list of artifacts inside the project I have access to: `https://harbor-repo.vmware.com/api/v2.0/projects/agamez/repositories`

**Acceptance criteria**

- The Flux OCI support works with robot accounts.

**Additional context**

N/A

gfichtenholt:
I am able to use robot accounts with demo.goharbor.io without any issues. A PR with an integration test to that effect is pending https://github.com/vmware-tanzu/kubeapps/pull/5246/files#

gfichtenholt:
please feel free to re-open as you see fit

gfichtenholt:
Reopening due to:
If I run a test that uses admin account followed by a test that uses a robot account everything runs fine. If I then bounce the kubeapps apis pod and run only the test that uses a robot account that fails. So I think there are a couple of issues:

1. there is some stale in-memory cached state (maybe in ORAS libs) that is set up by one test and somehow affects other tests.
2. There is some likelihood the robot account never worked with kubeapps. To fix it:
  a. We can pursue ppl behind harbor to see if it is possible to use standard API with robot accounts via some server-side config settings OR
  b.  might have to introduce a vendor-specific repository-lister

PR #5305, opened 2022-09-06
Flux oci support 12: fix for  Unable to use Flux OCI with Harbor robot accounts #5219 

gfichtenholt:
implemented fix as described in the issue 
https://github.com/vmware-tanzu/kubeapps/issues/5219#issuecomment-1233738309
fixed 1 and implemented 2b
verified it works with respect to robot account of VMware corporate harbor instance


cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/harbor_api_v2_repo_lister.go
```diff
@@ -0,0 +1,298 @@
...
+
+	httpclient "github.com/vmware-tanzu/kubeapps/pkg/http-client"
+	"helm.sh/helm/v3/pkg/registry"
+	log "k8s.io/klog/v2"
+	"oras.land/oras-go/v2/errdef"
+	orasregistryv2 "oras.land/oras-go/v2/registry"
+	orasregistryauthv2 "oras.land/oras-go/v2/registry/remote/auth"
+)
+
+// This flavor of OCI repsitory ister works with respect to to VMware Harbor
```
antgamdia, line 25:
```suggestion
// This flavor of OCI repository lister works with respect to VMware Harbor
```

Also, I don't think it is just for the VMware harbor instance, no? Ah, perhaps you meant the VMware Harbor product? If so, I think Harbor is not -just-a VMware project, but a Harbor is a [Cloud Native Computing Foundation](https://cncf.io/) one, as it appears as a Graduated project on their website.

Anyway, just being nitpicky, feel free to ignore the comment :P

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/harbor_api_v2_repo_lister.go
```diff
@@ -0,0 +1,298 @@
...
+// container registry. The Swagger for the API can be found on an running server
```
antgamdia, line 26: Another nitpick: the `Swagger` name is discouraged in favor of "OpenAPI"

```suggestion
// container registry. The OpenAPI description of the API can be found on a running server
```

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/harbor_api_v2_repo_lister.go
```diff
@@ -0,0 +1,298 @@
...
+
+	switch resp.StatusCode {
+	case http.StatusOK:
+		return nil
+
+	case http.StatusNotFound:
+		return errdef.ErrNotFound
+
+	case http.StatusForbidden:
+		return errors.New("forbidden")
```
antgamdia, line 158: I don't know where both errors are being handled, but it seems we are mixing up the error builder from oras (`errdef`) and the builtin `errors` packages. Should we use the same for both cases? Or perhaps it doesn't really matter due to the error hanlding being performed on the call site?

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/harbor_api_v2_repo_lister.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+}
+
+// buildHarborRegistryBaseURL builds the URL for accessing the base API.
+func buildHarborRegistryBaseURL(plainHTTP bool, ref orasregistryv2.Reference) string {
+	scheme := "https"
+	if plainHTTP {
+		scheme = "http"
+	}
+	return fmt.Sprintf("%s://%s/api/v2.0", scheme, ref.Host())
```
antgamdia, line 196: Just thinking:  would be a good idea to extract all the templates into `const` being defined at the bottom? This way we would have all the paths in the same place (in case the change their API.
Anyway, not really important, as I don't foresee that breaking changes for the major (v2) this lister is intended to.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/harbor_api_v2_repo_lister.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+}
+
+// parseHarborErrorResponse parses the error returned by the remote registry.
+func parseHarborErrorResponse(resp *http.Response) error {
+	var errmsg string
+	var body struct {
+		Errors requestErrors `json:"errors"`
+	}
+	lr := io.LimitReader(resp.Body, maxErrorBytes)
```
antgamdia, line 291: Just curious as it seems a good idea for preventing some buffer overflow attacks (and we aren't doing it in the rest of the codebase, AFAIK). 
Do you think we should limit the read operations in other places?

antgamdia (approved):
Great to see we are finally supporting Harbor! 🎉 , thanks for making it possible!

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/harbor_api_v2_repo_lister.go
gfichtenholt: will fix comment

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/harbor_api_v2_repo_lister.go
gfichtenholt: link to source please? 
Harbor certainly did not get the message
Screen Shot 2022-09-06 at 8 18 02 AM

antgamdia: Sorry, here it is: https://www.openapis.org/faq/style-guide.
Besides, my former company was a member of the OpenAPI Initiative and I was somehow also involved in.

But... the reality is that there are plenty of people still using the "Swagger" word. In fact, it can be a synonym, but only if the spec version is `2.0`. 

In fact, the according to the style guide, the proper term would be `OpenAPI Specification document` or just `OAS document`

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/harbor_api_v2_repo_lister.go
gfichtenholt: I will get rid of lines 157-158. They are not needed.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/harbor_api_v2_repo_lister.go
gfichtenholt: thanks. I will re-word comment

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/harbor_api_v2_repo_lister.go
gfichtenholt: yeah, I don't see them changing v2 API in a backwards incompatible way anywhere soon. Will keep your comment in mind for the next PR(s)

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/harbor_api_v2_repo_lister.go
gfichtenholt: I think in general it is a good idea, provided you can make this assumption. Normally, you can't really assume anything about the size of the payload you're dealing with. So whatever limit/boundary X you pick, I will probably find a valid use case where receiving X+1 bytes is valid

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/chart_integration_test.go
```diff
@@ -681,8 +767,21 @@ func TestKindClusterAvailablePackageEndpointsForOCI(t *testing.T) {
 				t.Fatal(err)
 			}
 			// wait until this repo reaches 'Ready'
-			if err = kubeWaitUntilHelmRepositoryIsReady(t, repoName); err != nil {
-				t.Fatal(err)
+			err = kubeWaitUntilHelmRepositoryIsReady(t, repoName)
+			if !tc.unauthorized {
+				if err != nil {
+					t.Fatal(err)
+				}
```
beni0888, line 774: I'm seeing this pattern repeated a lot through our code base, have you considered using `stretchr/testify` library, which is already included as a project dependency, and would make tests (IMHO) easier to write, read and maintain? In this case, we'd replace the three lines with just `require.NotNil(err)`.

beni0888 (commented):
LGTM!

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/chart_integration_test.go
gfichtenholt: I haven't considered it due to I did not know it existed. I don't have any objection to using it

PR #5322, opened 2022-09-08
Remove unused `testImage` from chart

antgamdia:
### Description of the change

We noticed we had an unused `.testImage` in our chart. It was being used by a former `helm test` command we used to have.
Today, it is no longer used and can be safely deleted.

### Benefits

No old code in the chart.

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

Also squeezing a couple of auth-proxy bypasses for the favicon files.


castelblanque (approved):
Great, thanks for the clean up!

PR #5336, opened 2022-09-12
Make carvel `globalPackagingNamespace` configurable

### LINKED ISSUE #5334: Carvel plugin - Make `globalPackagingNamespace` configurable via chart
**Summary**
Add a `globalPackagingNamespace` in the Carvel plugin config and use it in the code.

**Background and rationale**
Currently we have `kapp-controller-packaging-global` hardcoded, but, depending on the environment this value can change. This value should be modifiable from the chart.

**Description**
Create a `globalPackagingNamespace` in the chart and use it in the code.

**Acceptance criteria**
The `globalPackagingNamespace` field is configured through the chart in both the plugin backend and the UI.

**Additional context**
N/A

antgamdia:
### Description of the change

This PR mainly extracts the global ns constant into a configurable param in the chart, so that it can be used in both the backend and UI code.

### Benefits

Users will be able to configure the global ns for Carvel packages.

### Possible drawbacks

N/A

### Applicable issues

- fixes #5334

### Additional information

image


chart/kubeapps/templates/dashboard/configmap.yaml
```diff
@@ -70,7 +70,7 @@ data:
       "kubeappsCluster": {{ include "kubeapps.kubeappsCluster" . | quote }},
       "kubeappsNamespace": {{ .Release.Namespace | quote }},
       "globalReposNamespace": {{ include "kubeapps.globalReposNamespace" . | quote }},
```
castelblanque, line 72: This `globalReposNamespace` flag relates only to direct Helm plugin. Shouldn't we also move it to value e.g. `kubeappsapis.pluginConfig.core.packages.v1alpha1.globalPackagingNamespace` ?

castelblanque, line 72: This `globalReposNamespace` flag relates only to direct Helm plugin. Shouldn't we also move it to value e.g. `kubeappsapis.pluginConfig.core.packages.v1alpha1.globalPackagingNamespace` ?

castelblanque: So that it is in line with the one added for Carvel.

castelblanque: This comment should have been added to the `values.yaml` file, but the referenced flag is not available 😄

castelblanque (approved):

chart/kubeapps/templates/dashboard/configmap.yaml
antgamdia: 100 % agree, we should move this config to a plugin-aware one.

> kubeappsapis.pluginConfig.core.packages.v1alpha1.globalPackagingNamespace

Did you really mean `core` ? Or `helm` and `flux` separately?  I don't think it should be a core config, as in each plugin it can be different.

castelblanque: We don't yet have the specific `helm` config path, that is why mentioned the core. But totally +1 to creating the specific `directhelm`/`helm` config like we have for the other plugins, aggregating the ones from `core`.
So it would be `kubeappsapis.pluginConfig.helm.packages.v1alpha1.globalPackagingNamespace`.

antgamdia: I've been looking into it... but modifying it would imply backward incompatible changes. Currently, we are defaulting to "kubeapps namespace + suffix" in `apprepos ctrl`, `asset-syncer` and `kubeappsapis`, however, it's being done by defining  a template constant in helm.

So my approach is: if the user selects a `helm....globalPackagingNamespace`, then use it. Otherwise, fall back to the old behavior. This way, we can avoid releasing a new major chart version with an explicit upgrade path.

dlaloue-vmware:
what will be the new chart version for this change? I think there is a possible backward incompatibility and a regression.

If a user does an upgrade of kubeapps, the new code will expect the global namespace to be in the configmap, but that will not be the case and the plugin will fail to create global repositories (I just spent a couple hours to track why i could not create carvel global repos anymore).

i think to make the code bacward compatible, we should have a default value in the code, that can be overriden in the values.yaml file for later chart releases.

antgamdia:
> what will be the new chart version for this change? I think there is a possible backward incompatibility and a regression.

This change will be released along with the Kubeapps 2.6.0 version, I guess. The change in the chart will be synced upstream at the same time of the actual release.

> i think to make the code bacward compatible, we should have a default value in the code, that can be overriden in the values.yaml file for later chart releases.

Mmm, I assumed it already was, as we are setting some defaults in case the user doesn't pass any value for a param, see:

https://github.com/vmware-tanzu/kubeapps/blob/394cb46a31c82c0af70b2bd696ca271df3f96f37/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server.go#L33-L38

But, given what you have noticed, it seems it is not working as expected? What was your setup: using bitnami chart + `kubeapps/*:latest` images? Happy to fine-tune the code to make it backwards compatible, of course.

dlaloue-vmware:
you are correct, my setup is that i installed Kubeapps from the bitnami chart, then built kubeapps to have the latest UI and upgraded the images to the latest bits. This meant no change in values.
I assume this is similar to what would happen if a user were to upgrade from 2.5.x to 2.6.0 (but i could be wrong).

also, while we are defining defaults, the defaults are not merged, it is either / or. That is, if the plugin config is provided, then we use it otherwise we use the defaults. But if the plugin config is missing values, then they are blank.
maybe the plugin config should be loaded and then merged with the defaults to ensure we are not missing any values.
this would solve any backward incompatibility issues.

PR #5337, opened 2022-09-12
Fix for CI: adding chart museum ingress

castelblanque:
### Description of the change

CI E2E runs are constantly failing due to problems with the `port-forward` done in the chart-museum part.
Whenever a chart is needed to be pushed to chart-museum, there is a port-forward done to the service.

That solution is weak and error prone, this PR removes that and adds an ingress setup for Chart museum so that there is no port-forward used anymore.
Ingress is only used from the scripts to push charts. From the E2E tests, the same internal service locator is used (`http://chartmuseum.${CHARTMUSEUM_NS}.svc.cluster.local:8080/`).

### Benefits

It should bring more reliability.

### Possible drawbacks

N/A

### Applicable issues

N/A

### DIFF
--- a/script/chart-museum.sh
+++ b/script/chart-museum.sh
@@ -3,7 +3,6 @@
 # Copyright 2022 the Kubeapps contributors.
 # SPDX-License-Identifier: Apache-2.0
 
-CHARTMUSEUM_PORT=${CHARTMUSEUM_PORT:-8090}
 CHARTMUSEUM_USER=${CHARTMUSEUM_USER:-"admin"}
 CHARTMUSEUM_PWD=${CHARTMUSEUM_PWD:-"password"}
 CHARTMUSEUM_NS=${CHARTMUSEUM_NS:-"chart-museum"}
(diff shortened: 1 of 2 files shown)


script/chart-museum.sh
```diff
@@ -107,9 +109,6 @@ spec:
         pathType: ImplementationSpecific
 EOF
 
-  # Add Ingress IP to hosts file
-  sudo echo "${DEX_IP}  ${CHARTMUSEUM_HOSTNAME}" >> /etc/hosts
```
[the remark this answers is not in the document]
castelblanque: Not really. That line was a test I did, but I didn't remember that the whole script must be run with `sudo`, which is not a good option. Finally went with sending the `Host` header and going to the IP directly.
However, it still fails and I can't replicate it locally. Whenever I have time I'll continue debugging in CircleCI directly.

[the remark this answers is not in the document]
castelblanque: Feel free to jump in if you see the cause of the current error.

script/chart-museum.sh
```diff
@@ -88,17 +79,47 @@ installChartMuseum() {
...
+  echo "Installing Ingress for ChartMuseum with access through host ${CHARTMUSEUM_HOSTNAME}"
+  kubectl create -n $CHARTMUSEUM_NS -f - -o yaml << EOF
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  annotations:
+    nginx.ingress.kubernetes.io/connection-proxy-header: keep-alive
+    nginx.ingress.kubernetes.io/proxy-read-timeout: "600"
+    nginx.ingress.kubernetes.io/proxy-buffer-size: "8k"
+    nginx.ingress.kubernetes.io/proxy-buffers: "4.0"
```
antgamdia, line 92: not sure we really need these annotations here, no?

script/chart-museum.sh
```diff
@@ -88,17 +79,47 @@ installChartMuseum() {
...
+      paths:
+      - backend:
+          service:
+            name: chartmuseum
+            port:
+              number: 8080
+        path: /
+        pathType: ImplementationSpecific
+EOF
+  sleep 20
```
antgamdia, line 108: Why this manual sleep here? Maybe it's for giving some time to the ingress controller to become available? If so, why not just sth like: `        kubectl wait --namespace ingress-nginx --for=condition=ready pod --selector=app.kubernetes.io/component=controller --timeout=120s
`

antgamdia (approved):
Great, thanks! Let's see if it brings more reliability to the CI runs 🤞

script/chart-museum.sh
castelblanque: That `sleep` is the key that missing. Before that, `Ingress` wasn't ready by the time it was used some instructions after. I thought about adding the `kubectl wait`, but that is for the nginx controller itself which is ready way before ([when the cluster is installed](https://github.com/vmware-tanzu/kubeapps/blob/bfec7d10a1d87e2b7808e3244bada03e7fa40a99/.circleci/config.yml#L364)).
It seems that it takes some time for the route to be processed by the Nginx controller?

script/chart-museum.sh
castelblanque: We are using those annotations in the other Ingress definition we use, so I just copy-pasted. Happy to remove them.

PR #5368, opened 2022-09-21
Flux oci support 13: support HelmRepository.spec.provider + document use case with GCP workload identity

gfichtenholt:
Implemented as discussed in https://github.com/vmware-tanzu/kubeapps/issues/5007#issuecomment-1233691352

+some un-related clean up of scripts in testdata that help set up the integration test env.
+bump flux version
+some DRYing up code

regarding adding an integration test for the use case, I am still thinking about how to do it and whether it is worth the effort. To run a use case like that it would require a dedicated GKE cluster, which costs our business unit money, so to stand one up and keep it running long term just so we can run this one test every once in a while seems very wasteful.


cmd/kubeapps-apis/proto/kubeappsapis/plugins/fluxv2/packages/v1alpha1/fluxv2.proto
```diff
@@ -127,3 +127,19 @@ message SetUserManagedSecretsRequest {
...
+  // optional field that allows specifying an OIDC provider used for authentication purposes
+  // Supported options are:
+  //  - generic
+  //  - aws
+  //  - azure
+  //  - gcp
+  // The provider field is supported only for Helm OCI repositories. The repository type must 
+  // be set to "oci"
+  // ref https://fluxcd.io/flux/components/source/helmrepositories/#provider
+  string provider = 1;
```
castelblanque, line 144: Given that it is a fixed set of options, why not using [a protobuf `enum`](https://developers.google.com/protocol-buffers/docs/proto3#enum)?

gfichtenholt: these come from flux. I don't want to repeat their definitions and I want to let them validate the string values

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo.go
```diff
@@ -252,7 +254,19 @@ func (s *Server) newRepo(ctx context.Context, request *corev1.AddPackageReposito
...
+	// Get Flux-specific values
+	provider := ""
+	var customDetail *v1alpha1.FluxPackageRepositoryCustomDetail
+	if request.CustomDetail != nil {
+		customDetail = &v1alpha1.FluxPackageRepositoryCustomDetail{}
+		if err := request.CustomDetail.UnmarshalTo(customDetail); err != nil {
+			return nil, status.Errorf(codes.InvalidArgument, "customDetail could not be parsed due to: %v", err)
+		}
+		log.Infof("fluxv2 customDetail: [%v]", customDetail)
+		provider = customDetail.Provider
```
castelblanque, line 266: Shouldn't we fail at this point if there is a provider and repo type is not `HelmRepositoryTypeOCI`?

gfichtenholt: again, I prefer to let flux do this validation, which they already do. See the note in https://fluxcd.io/flux/components/source/helmrepositories/#provider

castelblanque (approved):

gfichtenholt:
Thanks for the review Rafa. I would like to resolve this CodeQL failure before I push these changes

PR #5358, opened 2022-09-19
New K8s clients provider for Helm plugin

castelblanque:
Signed-off-by: Rafa Castelblanque <rcastelblanq@vmware.com>

### Description of the change

This PR adds a new logic for obtaining the clients for K8s APIs. It resides in the `/pkg` package, but it is only applied to the Helm plugin by now.
It should speed up the response times for Helm plugin: a lot when using Pinniped, not so much without Pinniped.

Problem solved here was found to be the root cause for #4919.

In the business logic of Helm plugin (at least), usually only one type of client is required per scenario (e.g. the typed client when retrieving available packages summary).
When using the old client getter, all types of clients were being retrieved due to [the invocation of `clientGetterHelper`](https://github.com/vmware-tanzu/kubeapps/blob/cf858134a3b67e9f6ee2c53ce7b5071407b3ff7a/cmd/kubeapps-apis/plugins/pkg/clientgetter/clientgetter.go#L228), even if only one was needed.

It turns out that when creating the dynamic client, many requests are done at the same time to the K8s APIs to get the resources, which probably is triggering the throttling mechanism. There is no throttling when `pinniped-proxy` is not in the picture, but when it is, all requests go through it, with throttling. List of URLs invoked:

```
/api/v1?timeout=32s
/apis?timeout=32s
/apis/admissionregistration.k8s.io/v1?timeout=32s
/apis/apiextensions.k8s.io/v1?timeout=32s
/apis/apiregistration.k8s.io/v1?timeout=32s
/apis/apps/v1?timeout=32s
/apis/authentication.concierge.pinniped.dev/v1alpha
/apis/authentication.k8s.io/v1?timeout=32s
/apis/authorization.k8s.io/v1?timeout=32s
/apis/autoscaling/v1?timeout=32s
/apis/autoscaling/v2?timeout=32s
/apis/autoscaling/v2beta2?timeout=32s
/apis/batch/v1?timeout=32s
/apis/certificates.k8s.io/v1?timeout=32s
/apis/config.concierge.pinniped.dev/v1alpha1?timeou
/apis/coordination.k8s.io/v1?timeout=32s
/apis/discovery.k8s.io/v1?timeout=32s
/apis/events.k8s.io/v1?timeout=32s
/apis/flowcontrol.apiserver.k8s.io/v1beta1?timeout=
/apis/flowcontrol.apiserver.k8s.io/v1beta2?timeout=
/apis/identity.concierge.pinniped.dev/v1alpha1?time
[...]

Issue #4919, opened 2022-06-14
Requests for catalog delayed


absoludity:
**Describe the bug**
While testing #4918 , I noticed an oddity that is causing requests for `GetAvailablePackageSummaries` to be delayed significantly (up to 16s sometimes:

![GetAvailablePackageSummaries-long](https://user-images.githubusercontent.com/497518/173587565-ba98e4ad-dbc5-40b2-be99-8db3b6c13ea5.png)

While watching the logs for pinniped proxy, I see the delay is because every request is creating a new REST API mapper and I suspect it's being throttled. What I see is that each paginated request results in a bunch of requests to the APi server:

```
INFO  pinniped_proxy::expired_value_cache > cache hit
 INFO  pinniped_proxy::expired_value_cache > cache hit
 INFO  pinniped_proxy::expired_value_cache > cache hit
 INFO  pinniped_proxy::service             > GET https://kubernetes.default/apis/autoscaling/v2beta2?timeout=32s 200 OK
 INFO  pinniped_proxy::service             > GET https://kubernetes.default/apis/scheduling.k8s.io/v1?timeout=32s 200 OK
 INFO  pinniped_proxy::service             > GET https://kubernetes.default/apis/batch/v1?timeout=32s 200 OK
 INFO  pinniped_proxy::service             > GET https://kubernetes.default/apis/kubeapps.com/v1alpha1?timeout=32s 200 OK
 INFO  pinniped_proxy::service             > GET https://kubernetes.default/apis/authentication.k8s.io/v1?timeout=32s 200 OK
 INFO  pinniped_proxy::service             > GET https://kubernetes.default/apis/autoscaling/v1?timeout=32s 200 OK
 INFO  pinniped_proxy::service             > GET https://kubernetes.default/apis/apps/v1?timeout=32s 200 OK
 INFO  pinniped_proxy::service             > GET https://kubernetes.default/apis/discovery.k8s.io/v1?timeout=32s 200 OK
 INFO  pinniped_proxy::service             > GET https://kubernetes.default/apis/events.k8s.io/v1beta1?timeout=32s 200 OK
 INFO  pinniped_proxy::service             > GET https://kubernetes.default/apis/coordination.k8s.io/v1?timeout=32s 200 OK
 INFO  pinniped_proxy::service             > GET https://kubernetes.default/apis/batch/v1beta1?timeout=32s 200 OK
[...]

PR #5379, opened 2022-09-22
Use new K8s clients provider in Flux plugin

castelblanque:
### Description of the change

Follow-up PR to #5358. In this case for Flux plugin.
Using the new clients provider with independent functions, so not all clients are created at once.

Specifically, when creating the `Dynamic` and `ControllerRuntime` clients, many requests are done at the same time againts K8s APIs. In the scenario of Pinniped, those requests are done against `pinniped-proxy`, which proxies to the actual k8s API server. There is no throttling between `kubeapps-apis` and `pinniped-proxy`, as all clients count with `qps`/`burst` configuration. So if there is a throttling problem, it must be at `pinniped-proxy` against k8s API server. Maybe a connection pool limitation?

### Benefits

Should speed up a bit requests to Kubeapps APIs in the pinniped scenario.

### Possible drawbacks

N/A

### Applicable issues

- related #4919 

### Additional information

Added client `QPS` and `burst` configuration to in-cluster clients too.


cmd/kubeapps-apis/plugins/pkg/clientgetter/clients_provider.go
```diff
@@ -157,3 +215,29 @@ func NewFixedClientProvider(clientsGetter *ClientGetter) ClientProviderInterface
...
+		if config, err := rest.InClusterConfig(); err != nil {
+			code := codes.FailedPrecondition
+			if status.Code(err) == codes.Unauthenticated {
+				// want to make sure we return same status in this case
+				code = codes.Unauthenticated
+			}
+			return nil, status.Errorf(code, "unable to get in cluster config due to: %v", err)
+		} else {
+			config.QPS = clientQPS
+			config.Burst = clientBurst
```
castelblanque, line 239: Added `qps` and `burst` params to in-cluster/background clients provider.

antgamdia (approved):
Thanks!! Forgot to +1ing before, sorry!

Issue #4436, opened 2022-03-15
Migrate CI to Github actions


castelblanque:
### Description:

Move CI to Github actions

- Benefit: Be aligned with most of the Tanzu projects using it

Related to #4096

antgamdia:
In the meantime, we could start splitting some logic into some container images, as Rafa suggested (see https://github.com/vmware-tanzu/kubeapps/pull/5177#pullrequestreview-1083477458)

beni0888:
The first iteration of the migration has been completed with the merge of the first version of the new workflow, which includes those jobs that are run for every single push of a branch: test, build images, push images, run e2e tests, and sync from/to Bitnami chart (all but those that belong to the `pre-release` flow, and `report_srp`).

castelblanque:
Awesome @beni0888!
Are we storing artifacts from E2E in case they fail? (e.g. screenshots, videos, logs, etc.)

antgamdia:
I think so, under the "artifacts" section in the github action:
However, they are stored in a zip, so we have to download and uncompress it to view the files. An extra step, unfortunately, but not a big deal.

![image](https://user-images.githubusercontent.com/11535726/194036898-8062c53f-a77e-40a1-b8db-c08e415f6498.png)

beni0888:
Yeah, as Antonio said, in case E2E tests fail, the reports are stored under the `Artifacts` section in the workflow run.

castelblanque:
There is no way to get the artifacts downloaded independently like in CircleCI, but only as a zip file.
Citing [GH action documentation](https://github.com/actions/upload-artifact#zipped-artifact-downloads): `There is currently no way to download artifacts after a workflow run finishes in a format other than a zip or to download artifact contents individually.`
This will make debugging errors in CI slower, e.g. a huge file will have to be downloaded just to see one picture of failing E2E.

beni0888:
Lately, I've been fighting against the issues that were making the jobs `sync_from_bitnami` and `sync_to_bitnami` fail. A thorough explanation of the issues faced and the solutions applied can be found in #5524.

beni0888:
I noticed that `dependabot`'s PRs are failing at `push_dev_images`, I guess it should be because GH is considering dependabot as an outsider and secrets aren't available in the triggered workflow. I need to dig deeper into that issue and find out the best approach to fix it. Adding a new task to the list.

beni0888:
[...]

PR #5384, opened 2022-09-23
Add new CICD workflow in Github Actions

beni0888:
### Description of the change

This PR consists of a first iteration in the migration of the CICD pipeline from CircleCI to GitHub actions (well at this moment, it just consists of the creation of a new pipeline in GHA containing a subset of the jobs in the existent pipeline). 

### Benefits

* We will be compliant with the standard way of handling CICD in the majority of VMware's open-source projects.
* The new workflow in GHA seems to be faster than the one from CircleCI.
* It will be possible to generate multiplatform images.

### Possible drawbacks

In Github Actions, there isn't a one-to-one match for some of the existing features in CircleCI (eg. yaml snippets reuse), so we can end up with a more complex setup.

### Applicable issues

- related to #4436

### Additional information

Once merged, the new workflow in GHA will coexist with the existent CircleCI pipeline for some time, until we have fully tested it and decide to move forward and remove/deactivate the old pipeline in favor of the new one.


.github/workflows/kubeapps.yaml
```diff
@@ -0,0 +1,778 @@
...
+            kubectl --context kind-kubeapps-ci-additional --kubeconfig ${HOME}/.kube/kind-config-kubeapps-ci-additional apply --kubeconfig=${HOME}/.kube/kind-config-kubeapps-ci-additional -f ./site/content/docs/latest/reference/manifests/kubeapps-local-dev-users-rbac.yaml &&
+            kubectl --context kind-kubeapps-ci-additional --kubeconfig ${HOME}/.kube/kind-config-kubeapps-ci-additional apply --kubeconfig=${HOME}/.kube/kind-config-kubeapps-ci-additional -f ./site/content/docs/latest/reference/manifests/kubeapps-local-dev-namespace-discovery-rbac.yaml &&
+
+            kubectl --context kind-kubeapps-ci-additional --kubeconfig ${HOME}/.kube/kind-config-kubeapps-ci-additional create rolebinding kubeapps-view-secret-oidc --role view-secrets --user oidc:kubeapps-user@example.com &&
+            kubectl --context kind-kubeapps-ci-additional --kubeconfig ${HOME}/.kube/kind-config-kubeapps-ci-additional create clusterrolebinding kubeapps-view-oidc  --clusterrole=view --user oidc:kubeapps-user@example.com &&
+            echo "Additional cluster created"
```
castelblanque, line 463: Same as above, this is repetead, only used for retrying.

.github/workflows/kubeapps.yaml
```diff
@@ -0,0 +1,778 @@
...
+
+          echo DEFAULT_DEX_IP=$DEFAULT_DEX_IP
+          echo DEX_IP=$DEX_IP
+          echo ADDITIONAL_CLUSTER_IP=$ADDITIONAL_CLUSTER_IP
+
+          # If running kubectl without args, use the default "kubeapps-ci" cluster
+          cp ${HOME}/.kube/kind-config-kubeapps-ci ${HOME}/.kube/config
+          kubectl config set-context kind-kubeapps-ci
+
+          # If the default IP the proper one, the multicluster setup will fail
```
castelblanque, line 480:
```suggestion
          # If the default IP is not the proper one, the multicluster setup will fail
```

.github/workflows/kubeapps.yaml
```diff
@@ -0,0 +1,778 @@
...
+          if [ "$DEFAULT_DEX_IP" != "$DEX_IP" ]; then echo "Default IP does not match with current IP used in Kind"; exit 1; fi
```
castelblanque, line 481:
```suggestion
          if [ "$DEFAULT_DEX_IP" != "$DEX_IP" ]; then echo "Default IP does not match with current IP"; exit 1; fi
```

.github/workflows/kubeapps.yaml
```diff
@@ -0,0 +1,778 @@
...
+            kind load image-archive "${path}/${image}.tar" --name kubeapps-ci; 
+          done
+      - name: "Install multicluster deps"
+        run: |
+          sed -i -e "s/172.18.0.2/$DEFAULT_DEX_IP/g;s/localhost/kubeapps-ci.kubeapps/g" ./site/content/docs/latest/reference/manifests/kubeapps-local-dev-dex-values.yaml
+          helm repo add dex https://charts.dexidp.io
+
+          # Install dex
+          kubectl create namespace dex
+          helm install dex dex/dex --version 0.5.0 --namespace dex --values ./site/content/docs/latest/reference/manifests/kubeapps-local-dev-dex-values.yaml
```
castelblanque, line 525: Maybe we could simplify this with `make deploy-dex`?

.github/workflows/kubeapps.yaml
```diff
@@ -0,0 +1,778 @@
...
+
+          # Install openldap
+          helm repo add stable https://charts.helm.sh/stable
+          kubectl create namespace ldap
+          helm install ldap stable/openldap --namespace ldap
```
castelblanque, line 530: Same as above, with `make deploy-openldap`.

.github/workflows/kubeapps.yaml
```diff
@@ -0,0 +1,778 @@
...
+        with:
+          key: ${{ secrets.SSH_KEY_FORKED_CHARTS_DEPLOY }}
+          name: ${{ needs.setup.outputs.ssh_key_forked_charts_deploy_filename }}
+          known_hosts: |
+            |1|2YkQ4jjACcc/1rgSBszyeEuKxW4=|hO4GB0XMwQj1gYQDmaS304aU8Tc= ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEAq2A7hRGmdnm9tUDbO9IDSwBK6TbQa+PXYPCPy6rbTrTtw7PHkccKrpp0yVhp5HdEIcKr6pLlVDBfOLX9QUsyCOV0wzfjIJNlGEYsdlLJizHhbn2mUjvSAHQqZETYP81eFzLQNnPHt4EVVUh7VfDESU84KezmD5QlWpXLmvU31/yMf+Se8xhHTvKSCZIFImWwoG6mbUoWf9nzpIoaSjB+weqqUUmpaaasXVal72J+UX2B+2RPW3RcT0eOzQgqlJL3RKrTJvdsjE3JEAvGq3lGHSZXy28G3skua2SmVi/w4yCE6gbODqnTWlg7+wC604ydGXA8VJiS5ap43JXiUFFAaQ==
+          if_key_exists: ignore
+      -
+        # This is a key pair: https://circleci.com/docs/2.0/gh-bb-integration/
+        # public key uploaded to GitHub as a deploy key with write permissions
+        # private key uploaded to CircleCI with hostname "github.com"
```
castelblanque, line 679: Does this comment need updating?

.github/workflows/kubeapps.yaml
```diff
@@ -0,0 +1,778 @@
...
+            kubectl --context kind-kubeapps-ci --kubeconfig ${HOME}/.kube/kind-config-kubeapps-ci apply -f ./site/content/docs/latest/reference/manifests/kubeapps-local-dev-users-rbac.yaml &&
+            kubectl --context kind-kubeapps-ci --kubeconfig ${HOME}/.kube/kind-config-kubeapps-ci apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/main/deploy/static/provider/kind/deploy.yaml &&
+            sleep 5 &&
+            kubectl wait --namespace ingress-nginx --for=condition=ready pod --selector=app.kubernetes.io/component=controller --timeout=120s &&
+
+            kubectl --context kind-kubeapps-ci --kubeconfig ${HOME}/.kube/kind-config-kubeapps-ci create rolebinding kubeapps-view-secret-oidc --role view-secrets --user oidc:kubeapps-user@example.com &&
+            kubectl --context kind-kubeapps-ci --kubeconfig ${HOME}/.kube/kind-config-kubeapps-ci create clusterrolebinding kubeapps-view-oidc  --clusterrole=view --user oidc:kubeapps-user@example.com &&
+            echo "Cluster created"
```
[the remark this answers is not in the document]
beni0888: Well, this is actually and copy&paster from what there is in the CircleCI flavor of the pipeline, but yes, you're right, we can take advantage of this PR to remove that duplication

.github/workflows/kubeapps.yaml
beni0888: What about just creating a script `install-multicluster-deps.sh`? After taking a look, I can see that this block of code is just used once in the whole CI (in the CircleCI one, I mean), so I don't think it would be really useful to have `deploy-dex` and `deploy-openldap` make targets, and I would prefer not to bloat the makefile. WDYT?

castelblanque: Both `deploy-dex` and `deploy-opendldap` already exist in the [dev makefile](https://github.com/vmware-tanzu/kubeapps/blob/main/script/makefiles/deploy-dev.mk#L10-L22). My comment was towards reusing them in the CI, if possible. Otherwise, fine with leaving them in the GH Yaml.

beni0888: Ok, I see... For `deploy-dex` I think we could reuse it as is. However, for `deploy-openldap`, the make target is calling Helm providing the values file in `./site/content/docs/latest/reference/manifests/kubeapps-local-dev-openldap-values.yaml`, while the scripting version in the CI is installing it without providing custom values, so I don't know what would imply that change...

castelblanque (approved):
LGTM, awesome work!
Both CircleCI and GH pipelines will be running simultaneously from now?

beni0888:
> LGTM, awesome work! Both CircleCI and GH pipelines will be running simultaneously from now?

yeah, they will be running simultaneously until we decide that GH pipeline has been tested enough and works properly, and we decide to switch off the CircleCI pipeline.

castelblanque:
Ok, I guess we can switch off CircleCI pipeline when we have completed a full release cycle.

Issue #2881, opened 2021-05-25
add end-2-end tests that test core api server and plug-ins features


gfichtenholt:
In addition to unit-tests we should also have end-2-end tests, where we stand up a cluster deploy carvel/flux on it test out our core and plug-in apis against that.

gfichtenholt:
I think ideally, everybody contributing to core api server and/or plug-ins, including me, should part-own it, but I can take it if no-one else will

absoludity:
> I think ideally, everybody contributing to core api server and/or plug-ins, including me, should part-own it, but I can take it if no-one else will

+1, with the caveat that I'd be keen to focus this iteration on getting the unit tests there (and actually running those in CI - we're currently not running them in CI nor pushing an image to dockerhub), since we've committed to that. I'll move this to next iteration discussion, but no problem if we start it early.

gfichtenholt:
Sounds good, will work on this. I have never run end-2-end tests locally. So, as a first step, I wanted to run the end-2-end tests on my laptop. So I tried to follow end-to-end-tests.md to see what all I need to do. I found out that this document is out-of-date. There are references to puppeteer, e2e-runner.yaml does not exist, etc. I am just trying to find the easiest way to run this stuff locally so I can add a test of my own. I would like to propose that the prerequisite  to my doing that is the up-to-date and easy to follow doc. Otherwise I am looking at spending a while just to set up the environment by reading scripts, etc.
I am volunteering to verify the contents of the doc, as I go along. Thanks

PR #5410, opened 2022-09-29
Open add end-2-end tests that test core api server and plug-ins features #2881 

gfichtenholt:
Also added some logic to make flux integration tests that rely on external repos (harbor, gcp, github) to handle infrequent intermittent errors due to long latencies and network failures by using grpc_retry package

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -165,6 +165,7 @@ require (
 	github.com/gorilla/mux v1.8.0 // indirect
 	github.com/gosuri/uitable v0.0.4 // indirect
 	github.com/gregjones/httpcache v0.0.0-20190611155906-901d90724c79 // indirect
+	github.com/grpc-ecosystem/go-grpc-middleware v1.3.0 // indirect
 	github.com/hashicorp/go-version v1.6.0 // indirect
 	github.com/hashicorp/hcl v1.0.0 // indirect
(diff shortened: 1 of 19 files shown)


go.mod
```diff
@@ -165,6 +165,7 @@ require (
 	github.com/gorilla/mux v1.8.0 // indirect
 	github.com/gosuri/uitable v0.0.4 // indirect
 	github.com/gregjones/httpcache v0.0.0-20190611155906-901d90724c79 // indirect
+	github.com/grpc-ecosystem/go-grpc-middleware v1.3.0 // indirect
```
castelblanque, line 168: We are in the frozen dependencies period. As this is introducing a new dependency, we won't be able to merge this PR by now.

gfichtenholt: it's a change in test code, not production file. Surely that qualifies as an exception

antgamdia: Technically, whereas we should avoid adding new deps in the code, we postponed the deadline to October, so if this PR is really needed for the release, we could re-trigger the process if we all agree.

gfichtenholt: Understood. This PR is not urgent and it can wait. I have other things I will work on in the meantime

antgamdia: @gfichtenholt , please feel free to merge the PR whenever you want. The release is finally out!

Issue #4396, opened 2022-03-08
Component for editing package installation values in table mode


castelblanque:
### Description:

Currently Kubeapps is capable of dynamically generating a form with chart values metadata so that users can use it to adjust installation values.
Sometimes this kind of dynamic forms are a bit sluggish, specially when the form is very large. Also, finding the right control that corresponds to the right installation value is sometimes complicated.

It could be nice being able to edit installation values in table mode. One row -> one installation value.
Something like TMC colleagues have done:

| Key                | Type    | Description                         | Value |
|--------------------|---------|-------------------------------------|-------|
| global             |         | Global configuration                |       |
| ![image](https://user-images.githubusercontent.com/67455978/157249735-a4c107e0-9218-4173-86bc-2780f28af014.png)imageRegistry | string  | Global image registry               |  docker.io/...     |
| ![image](https://user-images.githubusercontent.com/67455978/157249735-a4c107e0-9218-4173-86bc-2780f28af014.png)imagePullSecrets | array   |  Secrets....                                   |   ["secret1"] *    |
| commonAnnotations  | object   | Annotations common to all resources |  {}     |
| startupProbe       |         |                                     |       |
| ![image](https://user-images.githubusercontent.com/67455978/157249735-a4c107e0-9218-4173-86bc-2780f28af014.png)timeoutSeconds   | integer | Timeout seconds for startup probe   |  20 *     |

Information presented on screen would be the same, but smaller in size and quicker to edit than the form edition. It would be straight forward to know the corresponding package value key.
The only editable part would be the `Value` column. Edited values would have some kind of visual sign (e.g. asterisk, different color, etc.)

absoludity:
This might also make it simpler for us to have a standard way to render forms both for simple form components as well as schema items for extra fields required by packaging plugins.

antgamdia:
More updates here:

- As the schema increases, dealing with a real-time-in-yaml-param-setting process is unbearable. I have switched to a new model where we store the modifications and then apply them altogether as user's request.
- Unfortunately, this is still not solving the rendering time when the schema is long. I think I'll go with an external Table solution like https://tanstack.com/table. However, I wanted to have something working before going deeper into the table think.
- Some good news: even being in the table, I think we can reuse some logic for setting the <input>, <toggle> and other components to ease the UX when entering data.
- I haven't thought yet of the "custom forms" feature. I think we have to preserve this functionality as it is being used by our community. However, this is for a 2nd stage.

[...]

PR #5411, opened 2022-09-29
New deployment forms: changes in operator-related views

antgamdia:
### Description of the change

Before start adding deep changes to the deployment form, this PR is extracting those views related to operators (which will get refactored once we encapsulate the operators as a plugin) and aligning the views to the new "in-line diff editor". 

Therefore, the old (and pretty unmaintained) `Differential` component, has been removed in these views (but they have not in the non-operators ones for now).

Also, adding a couple of minor changes in the utility functions.

### Benefits

The non-operator-related views can be now safely modified in subsequent PRs.

### Possible drawbacks

N/A

### Applicable issues

- required by #4396

### Additional information

I'll be stacking more PRs on top of this one.

image

### DIFF
--- a/dashboard/src/shared/schema.ts
+++ b/dashboard/src/shared/schema.ts
@@ -4,9 +4,9 @@
 import Ajv, { ErrorObject, JSONSchemaType } from "ajv";
 import * as jsonpatch from "fast-json-patch";
 import * as yaml from "js-yaml";
-import { isEmpty, set } from "lodash";
+import _ from "lodash";
 // TODO(agamez): check if we can replace this package by js-yaml or vice-versa
-import YAML, { ToStringOptions, Scalar } from "yaml";
+import YAML, { Scalar, ToStringOptions } from "yaml";
 import { IBasicFormParam } from "./types";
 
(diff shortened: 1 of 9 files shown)


dashboard/src/shared/schema.ts
```diff
@@ -4,9 +4,9 @@
 import Ajv, { ErrorObject, JSONSchemaType } from "ajv";
 import * as jsonpatch from "fast-json-patch";
 import * as yaml from "js-yaml";
-import { isEmpty, set } from "lodash";
+import _ from "lodash";
```
castelblanque, line 7: Why using `_` instead of importing the needed functions?

antgamdia: You're right, we should import them explicitly (see [this benchmark](https://www.blazemeter.com/blog/import-lodash-libraries)). I just added some ongoing changes in my working branch, where I just used the usual `_`  import for lodash, but we should decrease the bundle size when possible. Thanks!

castelblanque (approved):
LGTM, thanks!

Issue #4917, opened 2022-06-14
Adapt "basic forms" to a JSONSchema-compliant spec


antgamdia:
**Summary**
Currently, the "basic form" feature is using a custom format to represent the data that should be rendered (pics below). This format is not standard nor it is being properly updated by the Bitnami team.
Besides, the current implementation becomes unusable when the number of params in the form is more than ~10 ?

**Background and rationale**

If Bitnami (the primary producer of content with this schema format) stops creating this file and moves towards a usual JSONSchema (see https://github.com/bitnami-labs/readme-generator-for-helm/issues/15), the "basic form" functionality will be useless.

**Description**

First, we should define which `x-whicheverCustomFormProperty` we are going to use (https://github.com/bitnami-labs/readme-generator-for-helm/issues/15).
Once agreed, we have to implement a UI form somehow. For instance, using `jsonschema-forms` (as in https://github.com/vmware-tanzu/kubeapps/pull/3527/files) or a tabular mode (as suggested in https://github.com/vmware-tanzu/kubeapps/issues/4396).

Or even we can have both: the forms-based approach for the "basic form" and the tabular one for the "advanced view" or something like that.

Doing so would solve, IMO both https://github.com/vmware-tanzu/kubeapps/issues/2540 and https://github.com/vmware-tanzu/kubeapps/issues/1924.

Additionally, we could revisit https://github.com/vmware-tanzu/kubeapps/issues/3535 to see how we can include custom schemas coming from 3rd party places.

**Acceptance criteria**

- We have designed an alternative for rendering the forms, preserving (when possible) the current functionality.
- There is a coordinated plan with Bitnami for them to stop publishing the current schema and start serving the new one.
- We have implemented the "basic form" alternative.

**Additional context**

![image](https://user-images.githubusercontent.com/11535726/173559143-43df83ea-03b1-4aef-ab9e-980c433f498d.png)
![image](https://user-images.githubusercontent.com/11535726/173559696-dc38cf0a-b488-41f0-ab5f-dbc6ceea6111.png)

antgamdia:
Some random thoughts before digging into this issue:

- If we happen to replace today's basic forms with a new solution, all the historical series of released charts would be incompatible with this new schema-based form.
  - We can support both, however, we would be maintaining old and potentially unnecessary code. 
  - This is a kind of a blind decision as we don't have data for actually deciding with certainty. 
  - I'd opt for dropping support unless we got a request explicitly asking for it. 
[...]

PR #5412, opened 2022-09-29
New deployment forms: add json-schema-based table

### LINKED ISSUE #4396: Component for editing package installation values in table mode

antgamdia:
### Description of the change

This PR is finally adding the changes related to the new JSON Schema-based table that aims to replace the old "basic form". This way, we can render a large amount of params with no lag and, what's more important: we get rid of the hacky ways to display a form using the `form: true` property in the schema.

I'll comment on some changes in-line to make the review process easier once the PR is ready.

### Benefits

Kubeapps will be able to render any JSONSchema-compliant schema. The overall performance of the form view gets boosted as well.

### Possible drawbacks

Custom components: we need to re-check with our adopters the current situation. 

### Applicable issues

- fixes #4396
- related #4917

Additionally,
- closes https://github.com/vmware-tanzu/kubeapps/issues/5114 (as we are just using native HTML `step` in both the slider and the input here)
- closes https://github.com/vmware-tanzu/kubeapps/issues/5025 (as the restore button now works as expected)
- closes https://github.com/vmware-tanzu/kubeapps/issues/2540 (as we have used CDS components in this new approach)
- closes https://github.com/vmware-tanzu/kubeapps/issues/1924 (as the `required` prop is now being added if present in the schema)
- closes https://github.com/vmware-tanzu/kubeapps/issues/4803 (as the issue seems to be gone in this PR)

### Additional information

Still a draft as some tests are still pending.

image

image


antgamdia:
CI passed 🎉  Marking as ready for review + adding in-line comments to the code.

dashboard/src/components/DeploymentForm/DeploymentFormBody/BasicDeploymentForm/TabularSchemaEditorTable/Params/SliderParam.tsx
```diff
@@ -0,0 +1,90 @@
...
+        value={currentValue}
+      />
+      <CdsControlMessage>{unsavedMessage}</CdsControlMessage>
+    </CdsRange>
+  );
+
+  const inputText = (
+    <div className="self-center">
+      <CdsInput className={isModified ? "bolder" : ""}>
+        <input
```
antgamdia, line 72: Since we are not validating, I'm not adding max/min restrictions here yet. This way users can edit the slider value manually; happy to add it if you think otherwise.

dashboard/src/components/DeploymentForm/DeploymentFormBody/BasicDeploymentForm/TabularSchemaEditorTable/Params/TextParam.tsx
```diff
@@ -0,0 +1,144 @@
...
+  };
+
+  const unsavedMessage = isValueModified ? "Unsaved" : "";
+  const isDiffCurrentVsDefault =
+    toStringValue(param.currentValue) !== toStringValue(param.defaultValue);
+  const isDiffCurrentVsDeployed =
+    toStringValue(param.currentValue) !== toStringValue(param.defaultValue);
+  const isModified =
+    isValueModified ||
+    (isDiffCurrentVsDefault && (!param.deployedValue || isDiffCurrentVsDeployed));
```
antgamdia, line 83: Display "modified" (aka text in bold) as soon as it diverges from the deployed value (if any) or from the default one.

dashboard/src/components/DeploymentForm/DeploymentFormBody/BasicDeploymentForm/TabularSchemaEditorTable/TabularSchemaEditorTable.tsx
```diff
@@ -0,0 +1,251 @@
...
+}
+
+export default function TabularSchemaEditorTable(props: TabularSchemaEditorTableProps) {
+  const { columns, data, globalFilter, setGlobalFilter, isLoading, saveAllChanges } = props;
+
+  // Component state
+  const [columnFilters, setColumnFilters] = useState<ColumnFiltersState>([]);
+  const [globalExpanded, setGlobalExpanded] = useState<ExpandedState>({});
+
+  const table = useReactTable({
```
antgamdia, line 45: This is the actual configuration of the Table. More info about the capabilities at https://tanstack.com/table/v8/docs/adapters/react-table

dashboard/src/components/DeploymentForm/DeploymentFormBody/BasicDeploymentForm/TabularSchemaEditorTable/TabularSchemaEditorTableHelpers.ts
```diff
@@ -0,0 +1,27 @@
+// Copyright 2022 the Kubeapps contributors.
+// SPDX-License-Identifier: Apache-2.0
+
+import { compareItems, rankItem } from "@tanstack/match-sorter-utils";
+import { FilterFn, SortingFn, sortingFns } from "@tanstack/react-table";
```
antgamdia, line 5: In this file, we are just exposing the filtering functions as per the docs: https://tanstack.com/table/v8/docs/examples/react/filters

dashboard/src/components/DeploymentForm/DeploymentFormBody/BasicDeploymentForm/TabularSchemaEditorTable/TabularSchemaEditorTableRenderer.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+function renderCellWithTooltip(
+  value: IBasicFormParam,
+  property: string,
+  className = "",
+  trimFromBeginning = false,
+  maxLength = MAX_LENGTH,
+) {
+  // If the value is an object/array, we need to stringify it
+  const stringValue = ["string", "number"].includes(typeof value?.[property])
```
antgamdia, line 24: If the cell is an object, show the JSON-based textual representation

dashboard/src/components/DeploymentForm/DeploymentFormBody/BasicDeploymentForm/TabularSchemaEditorTable/TabularSchemaEditorTableRenderer.tsx
```diff
@@ -0,0 +1,222 @@
...
+    ? value?.[property] || ""
+    : JSON.stringify(value?.[property]);
+
+  if (stringValue?.length > maxLength) {
```
antgamdia, line 28: Trimming the string to prevent the cell from overflowing its size: we just show a tooltip with the whole content.

dashboard/src/components/DeploymentForm/DeploymentFormBody/BasicDeploymentForm/TabularSchemaEditorTable/TabularSchemaEditorTableRenderer.tsx
```diff
@@ -0,0 +1,222 @@
...
+export function renderConfigCurrentValuePro(
+  param: IBasicFormParam,
+  handleBasicFormParamChange: (
+    p: IBasicFormParam,
+  ) => (e: React.FormEvent<HTMLInputElement | HTMLTextAreaElement | HTMLSelectElement>) => void,
+) {
+  // early return if the value is marked as a custom form component
+  if (param.isCustomComponent) {
+    // TODO(agamez): consider using a modal window to display the full value
+    return (
```
antgamdia, line 137: This is something we have to double-check with our adopters.

dashboard/src/components/DeploymentForm/DeploymentFormBody/BasicDeploymentForm/TabularSchemaEditorTable/TabularSchemaEditorTableRenderer.tsx
```diff
@@ -0,0 +1,222 @@
...
+          <ArrayParam
+            id={param.key}
+            label={param.title || param.path}
+            param={param}
+            handleBasicFormParamChange={handleBasicFormParamChange}
+            type={param?.schema?.items?.type ?? "string"}
+          />
+        );
+      } else {
+        // TODO(agamez): render the object properties
```
antgamdia, line 200: Leaving this case explicit as we might want to render a more user-friendly editor for objects (like "+ Add property" button or sth like this).

dashboard/src/components/DeploymentForm/DeploymentFormBody/DeploymentFormBody.tsx
```diff
@@ -0,0 +1,326 @@
...
+  appValues: valuesFromTheParentContainer,
+  setValues: setValuesFromTheParentContainer,
+  setValuesModified,
+  formRef,
+}: IDeploymentFormBodyProps) {
+  const {
+    availablePackageDetail,
+    versions,
+    schema: schemaFromTheAvailablePackage,
+    values: valuesFromTheAvailablePackage,
```
antgamdia, line 50: I've renamed the params to make it clearer what they really mean while developing. Happy to revert the change back, though.

dashboard/src/actions/installedpackages.ts
```diff
@@ -234,11 +234,20 @@ export function installPackage(
...
-        const validation = validate(values, schema);
+        const schemaValidation = validateSchema(schema);
+        if (!schemaValidation.valid) {
+          const errorText = schemaValidation?.errors
+            ?.map(e => `  - ${e.instancePath}: ${e.message}`)
+            .join("\n");
+          throw new UnprocessableEntityError(
+            `The schema for this package is not valid. Please contact the package author. The following errors were found:\n${errorText}`,
+          );
+        }
```
antgamdia, line 245: Validating the schema before actually creating the InstalledPackage. I have considered adding a "force button", however, Helm is also failing if a wrong schema is added (as it is validating against the `chart.values.schema.json` if present in the tar.gz). A "force" button would imply some extra backend work.

Currently, for us, the problem is mainly just https://github.com/bitnami-labs/readme-generator-for-helm/issues/34.

dashboard/src/components/DeploymentForm/DeploymentFormBody/BasicDeploymentForm/TabularSchemaEditorTable/Params/ArrayParam.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+  const renderInput = (type: string, index: number) => {
+    switch (type) {
+      case "number":
+      case "integer":
+        return (
+          <>
+            <CdsInput className="self-center">
+              <input
+                required={param.required}
```
antgamdia, line 60: Adding this basic HTML-based validation for solving this issue: https://github.com/vmware-tanzu/kubeapps/issues/1924

dashboard/src/components/DeploymentForm/DeploymentForm.tsx
```diff
@@ -267,6 +267,7 @@ export default function DeploymentForm() {
                 setValues={handleValuesChange}
                 appValues={appValues}
                 setValuesModified={setValuesModifiedTrue}
+                formRef={formRef}
```
castelblanque, line 270: It seems that `formRef` is only used for performing a `requestSubmit()` in the parent.
Why not passing a plain callback function as parameter instead of this reference?

antgamdia: It was introduced in one of my previous attempts when working on this. The idea is to trigger a plain normal HTML form "submit" event from a child, this way we leverage the built-in HTML validations (like `required`, `max`, `min`, `pattern`).

Anyway, just to double-check, I'm trying locally to replace the `ref` with the `handleDeploy()` function directly to ensure the `ref` is really needed or not.

antgamdia: Exactly, that's why I left the hacky `ref`, otherwise the HTML are not taken into account :(

nohtmlval

dashboard/src/shared/types.ts
```diff
@@ -398,37 +398,24 @@ export interface IKubeState {
...
-  sliderStep?: number;
-  sliderUnit?: string;
+  defaultValue: any;
+  deployedValue: any;
+  currentValue: any;
+  schema: JSONSchemaType<any>;
+  isCustomComponent?: boolean;
+};
+
+export interface IAjvValidateResult {
```
castelblanque, line 416: Why using this "criptic"" name? :D

antgamdia: No strong opinion on the naming, but since it is the result of performing the `validate` function on an`ajv` (the library for handling schemas) object, this name is the best I could think of. Suggestions? :P

antgamdia: Well, `ErrorObject`, one of the used types, is indeed imported from the Ajv library, so we would be kinda coupled to `ajv` anyway? But yep, at least we should add a comment explaining what is this interface for.

integration/tests/main/07-upgrade.spec.js
```diff
@@ -33,9 +33,13 @@ test("Upgrades an application", async ({ page }) => {
   // Deploy package
   await page.click('cds-button:has-text("Deploy") >> nth=0');
 
-  // Set replicas
+  // Increase replicas
   await page.locator("input[type='number']").fill("2");
-  await page.click('li:has-text("Changes")');
+
+  // Wait until changes are applied (due to the debounce in the input)
+  await page.waitForTimeout(1000);
```
castelblanque, line 40: Using `waitForTimeout` is always a potential source of problems.
When having problems or debugging, it gives a bit more information to use  something like 

```
const locator = page.locator("input[type='number']")
await expect(locator).toHaveValue('2', {timeout: 1000})
```

BTW is there only one input of type number for this chart? Selector is very generic.

castelblanque (approved):
Awesome work. Big step in making Kubeapps more generic and sustainable, thanks!

dashboard/src/shared/types.ts
castelblanque: You are right. Fine leaving it as is with a comment.

integration/tests/main/07-upgrade.spec.js
castelblanque: Ok, so it is the event that we are waiting here.
Could we add more info in the comment to specify that it is the event please?
Not sure if [Playwright's `waitForEvent` + predicate func](https://playwright.dev/docs/api/class-page#page-wait-for-event) will help here.

antgamdia: Thanks for the tip, I've been exploring it, but unfortunately, the events it listens to are not HTML events, but some playwright-specific ones: `backgroundpage`, `close`, `page`, `request`, `requestfailed`, `requestfinished`, `response`, `serviceworfker`

In SO, some responses point to more sophisticated alternatives (ideally for more denounced inputs). Example: https://stackoverflow.com/questions/73739729/how-to-test-expected-delay-time-on-page-element-fixed-wait-debounce-with-play

So, for now, I think we should keep it simple, and leave the wait :(

PR #5415, opened 2022-09-30
Fix for Kapp app suffix

castelblanque:
### Description of the change

Kapp controller has introduced a [new suffix for apps identifier since v0.39.0](https://github.com/vmware-tanzu/carvel-kapp-controller/blob/v0.39.0/pkg/deploy/factory.go#L46).
In our CircleCI we were actually using and old version of Kapp Controller (v0.32.0) due to some problem in passing variables to `e2e-test.sh`.

We discovered that changes done in #4767 were not being correctly tested in E2E. That PR adds a fix for handling the app identifier suffix in Kapp Controller versions >0.47, with a fallback to previous versions (using suffix `-ctrl`). E2E was only testing the fallback scenario. Suffix used was [`.apps.k14s.io`](https://github.com/vmware-tanzu/carvel-kapp/blob/develop/pkg/kapp/app/recorded_app.go#L28). Actual suffix for `ConfigMap` used by Carvel has suffix `.app`.

This PR fixes the current suffix usage.

### Benefits

CircleCI should use the latest version of Kapp Controller by now. Although the underlying problem with variables not being passed to `e2e-test.sh` remains.

### Possible drawbacks

We are not covering the right set of suffixes (`.app`/`.pkgr` or `.apps.k14s.io`)

### Applicable issues

- related #4672
- related #4767

### DIFF
--- a/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_resources.go
+++ b/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_resources.go
@@ -506,7 +506,7 @@ func (s *Server) updatePkgInstall(ctx context.Context, cluster, namespace string
 // getAppUsedGVs returns the list of GVs used by the given app, falling back to pre 0.47 Kapp version behavior with regards to suffixes
 func getAppUsedGVs(appsClient ctlapp.Apps, packageId string, namespace string, useNewCtrlAppSuffix bool) ([]schema.GroupVersion, ctlapp.App, error) {
 	// We first try to fetch the app using the suffixed name (kapp >= 0.47)
-	appName := fmt.Sprintf("%s%s", packageId, ctlapp.AppSuffix)
+	appName := fmt.Sprintf("%s%s", packageId, ".app")
 
 	// Workaround to also support pre-0.47 kapp versions, whose ConfigMap were suffixed with "-ctrl" instead of ".apps.k14s.io"
 	if !useNewCtrlAppSuffix {
--- a/script/e2e-test.sh
+++ b/script/e2e-test.sh
@@ -16,7 +16,7 @@ IMG_MODIFIER=${4:-""}
 TEST_TIMEOUT_MINUTES=${5:-"4"}
 DEX_IP=${6:-"172.18.0.2"}
 ADDITIONAL_CLUSTER_IP=${7:-"172.18.0.3"}
-KAPP_CONTROLLER_VERSION=${8:-"v0.32.0"}
+KAPP_CONTROLLER_VERSION=${8:-"v0.41.2"}
 CHARTMUSEUM_VERSION=${9:-"3.9.0"}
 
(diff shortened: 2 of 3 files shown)


cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_resources.go
```diff
@@ -506,7 +506,7 @@ func (s *Server) updatePkgInstall(ctx context.Context, cluster, namespace string
 // getAppUsedGVs returns the list of GVs used by the given app, falling back to pre 0.47 Kapp version behavior with regards to suffixes
 func getAppUsedGVs(appsClient ctlapp.Apps, packageId string, namespace string, useNewCtrlAppSuffix bool) ([]schema.GroupVersion, ctlapp.App, error) {
 	// We first try to fetch the app using the suffixed name (kapp >= 0.47)
-	appName := fmt.Sprintf("%s%s", packageId, ctlapp.AppSuffix)
```
antgamdia, line 509: Why changing this? It should the one available in the kappctrl dependency at compile time, no?

script/e2e-test.sh
```diff
@@ -16,7 +16,7 @@ IMG_MODIFIER=${4:-""}
 TEST_TIMEOUT_MINUTES=${5:-"4"}
 DEX_IP=${6:-"172.18.0.2"}
 ADDITIONAL_CLUSTER_IP=${7:-"172.18.0.3"}
-KAPP_CONTROLLER_VERSION=${8:-"v0.32.0"}
+KAPP_CONTROLLER_VERSION=${8:-"v0.41.2"}
```
antgamdia, line 19: So the version is not being passed through and instead we are using this hardcoded, fallback version, no?

castelblanque: Currently yes, we should work in providing the right values from CircleCI to `e2e-test.sh`.

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_resources.go
castelblanque: It turns out that this looks for a name ending in `.apps.k14s.io` (value in constant), but the actual identifier ends in `.app`.

antgamdia: Ahh, ok, ok... what a pity we don't have any constant to rely on in the kappctrl pkg :(

antgamdia (approved):

cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_resources.go
castelblanque: More info in [Rename KC owned apps from `x-ctrl` to `x.app` or `x.pkgr`](https://github.com/vmware-tanzu/carvel-kapp-controller/pull/665).

PR #5419, opened 2022-09-30
Fix circleci pipeline

### LINKED ISSUE #5416: CircleCI workflow is not properly handling config values
**Describe the bug**
Some config values are being defined as parameters (under the global `parameter` object) in our CircleCI workflow definition, and later they are fetched as if they were environment variables at the time of passing them to jobs or scripts (eg. `KAPP_CONTROLLER_VERSION` in `local_e2e_tests` job), so those empty values are passed in the calls to scripts, ending in positional parameters not being provided and (in the best case) scripts resourcing to the default values defined for those params. That has led to a subtle error in our CI pipeline that has been masking a bug in the Carvel plugin due to a breaking change introduced by the Carvel team in a recent release.

**To Reproduce**

1. Change the version of `KAPP_CONTROLLER_VERSION` param in CircleCI config file `/.circle/config.yml`.
2. Commit and push your changes, so a new build is triggered in CircleCI.
3. After the build has finished, go and check the logs of the execution of the `Run e2e tests script` step of the `local_e2e_tests` job of the build.
4. Look for the string `Installing kapp-controller` and check that version installed is the one defined as default in the script and not the one you've just set up in the `KAPP_CONTROLLER_VERSION` parameter of the `config.yml` file.

**Expected behavior**
Every configuration value defined as a parameter in CircleCI config file has to be taken into account when passing its value to jobs and scripts. To make sure this is the current behavior, apart from checking every used of every defined parameter, we could define a counterpart environment variable for each of them, as we are currently doing for just some of them. 

**Screenshots**

Screenshot 2022-09-30 at 13 28 23
Screenshot 2022-09-30 at 13 27 35
Screenshot 2022-09-30 at 13 26 49

beni0888:
### Fix CircleCI pipeline environment variables handling

This PR fixes the problem described in #5416. It defines global environment variables for `KAPP_CONTROLLER_VERSION` and `CHART_MUSEUM_VERSION`, so those values are properly passed to the `e2e_tests.sh` script.

### Benefits

It fixes an issue that had been causing a big headache to our team due to a subtle bug in our CircleCI config.

### Possible drawbacks

None

### Applicable issues

- fixes #5416 

### Additional information

N/A

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -258,11 +258,13 @@ workflows:
 
 ## Definitions
 common_envars: &common_envars
+  CHARTMUSEUM_VERSION: << pipeline.parameters.CHARTMUSEUM_VERSION >>
   DOCKER_VERSION: << pipeline.parameters.DOCKER_VERSION >>
   DOCKER_REGISTRY_VERSION: << pipeline.parameters.DOCKER_REGISTRY_VERSION >>
   GOLANG_VERSION: << pipeline.parameters.GOLANG_VERSION >>
   HELM_VERSION_MIN: << pipeline.parameters.HELM_VERSION_MIN >>
   HELM_VERSION_STABLE: << pipeline.parameters.HELM_VERSION_STABLE >>
+  KAPP_CONTROLLER_VERSION: << pipeline.parameters.KAPP_CONTROLLER_VERSION >>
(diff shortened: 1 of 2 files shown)


castelblanque (approved):
Great, thank you!

PR #5430, opened 2022-10-05
Change IMG_MODIFIER suffix for the GHA workflow

beni0888:
### Description of the change

Currently, both GHA and CircleCI workflows (aka. pipelines), are running in parallel and configured to generate the development docker images using the same suffix `-ci` to avoid polluting the canonical images repository. However, this leads us to a situation where both pipelines overlap when it comes to pushing the images to Dockerhub, and cannot make sure which was the pipeline that generated each published image (it will depend on which finishes last).

### Benefits

We can clearly identify which workflow generated what image and avoid overlap by having the images generated by both repositories publicly available in Dockerhub, under different suffixes.

### Possible drawbacks

Nothing, as we're in the development phase of the GHA workflow, it should have any negative consequences in case we break anything on it.

### Applicable issues

- related to #4436

### DIFF
--- a/.github/workflows/kubeapps.yaml
+++ b/.github/workflows/kubeapps.yaml
@@ -6,7 +6,7 @@ name: Kubeapps general pipeline
 on:
   push:
     branches:
-      - migrate-ci
+      - main
   pull_request:
     branches:
       - main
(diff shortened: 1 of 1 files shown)


antgamdia (approved):
Good idea, this way we avoid the overlap between different CI runs. Later on, we can just remove the `-gha` images from DockerHub manually, as they are just temporary.

PR #5431, opened 2022-10-05
Fix the job push_dev_images from GHA workflow

beni0888:

### Description of the change

Currently, we have a bug in the job `push_dev_images` from GitHub Actions workflow, that fails for two reasons:
* The env var `IMAGES_TO_PUSH` redefined at the job level is not taking the right value.
* The path used to load the images previously generated, uploaded and downloaded, is wrong.

### Benefits

The job `push_dev_images` will work as it is expected to do.

### Possible drawbacks

Nothing.

### Applicable issues

- related to #4436

### DIFF
--- a/.github/workflows/kubeapps.yaml
+++ b/.github/workflows/kubeapps.yaml
@@ -339,9 +339,10 @@ jobs:
       - build_dashboard_image
       - build_e2e_runner_image
     env:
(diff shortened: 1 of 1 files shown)


antgamdia (approved):
Thanks!

castelblanque (approved):
LGTM

PR #5444, opened 2022-10-08
Search table nested params

### LINKED ISSUE #5437: Allow searching params in nested values of the JSON Schema table
**Summary**
After #5412 is done, the param search box requires some improvements, as it only searches top-level params (not the nested ones).

**Background and rationale**
We just replicated the example of the new Table component we added, but we need to improve it.

**Description**
Allow searching nested params in the table search box.

**Acceptance criteria**
- The search box is able to find nested params.

**Additional context**
Related to https://github.com/vmware-tanzu/kubeapps/issues/4917

antgamdia:
### Description of the change

This PR adds a workaround while the TabStack table bug is being fixed (adding the `filterFromLeafRows` should be enough to get what we want). We just mark a row as "passing" if it matches the search query (existing behavior) OR it has a child that matches the query (workaround)

### Benefits

Users will be able to search by any nested value in the table.

### Possible drawbacks

I was a little reluctant to add this workaround for the performance implications I was foreseeing... but I have tested IRL and it is performing pretty well (see gif below), so I decided to to ahead.

### Applicable issues

- fixes #5437

### Additional information

(unedited speed)

filterFromLeafRows


castelblanque (approved):
Let's hope they fix the issue soon.
LGTM, thanks.

PR #5453, opened 2022-10-11
[Helm plugin] Fix for listing repos without cluster-wide permissions

### LINKED ISSUE #5215: Listing all repositories an unpriviledged users can access to
**Summary**
When using non-admin RBACs, the Apprepos `List` over the whole cluster action requires a ClusterRole (since we are passing `""` namespace). However, we should still be able to identify which ns a users has access to and then, list repos for each one.

**Background and rationale**
Otherwise, the current behavior is returning an empty set of repos, which could be misleading for the users.

**Description**

See above. 

**Acceptance criteria**

- Unprivileged users are able to fetch all the repositories from the namespaces they have access to.

**Additional context**
An issue related to this problem was solved at https://github.com/vmware-tanzu/kubeapps/pull/5207

castelblanque:

### Description of the change

Before the PR, when a user with non-admin invoked:

1. `GetPackageRepositorySummaries` passing as context e.g. the namespace: `kubeapps-user-namespace` -> Returned a valid list of accesible repositories in that namespace
2. `GetPackageRepositorySummaries` passing as context no namespace (or namespace: "") -> Returned an error `Forbidden to get the AppRepository 'all' due to...`. This happened even when the user had permissions for `AppRepository` and `get/Secret` in some namespaces.

This is because when using non-admin RBACs, the Apprepos List over the whole cluster action requires a ClusterRole (since we are passing "" namespace)

After the PR:

1. Is unchanged.
2. When user performs `GetPackageRepositorySummaries` passing as context no namespace (or namespace: "") -> Gets a valid list of repositories from the namespaces in which the user has access.

Warning: This logic is not applying the `kubeappsapis.pluginConfig.resources.packages.v1alpha1.trustedNamespaces.headerName` and `headerPattern`, so it could be time-consuming for scenarios with many namespaces.

### Benefits

No need for a ClusterRoleBinding in order to get the list of repositories cluster-wide. Logic now checks namespace by namespace.

### Possible drawbacks

Slow with high number of namespaces.

### Applicable issues

- fixes #5215 for Helm plugin


cmd/kubeapps-apis/plugins/pkg/resources/namespaces.go
```diff
@@ -0,0 +1,144 @@
...
+				allowedNamespaces = append(allowedNamespaces, res.ns)
+			}
+		} else {
+			log.Errorf("failed to check namespace permissions. Got %v", res.Error)
+		}
+	}
+	return allowedNamespaces, nil
+}
+
+func FilterActiveNamespaces(namespaces []corev1.Namespace) []corev1.Namespace {
```
antgamdia, line 136: I guess you're moving this logic as it will be re-used by the kapp plugin, great then.

antgamdia (approved):
Thanks! I guess you haven't had a chance to actually measure the overhead added by this solution, but I assume it is worth it (much better than an error 😅 )
Perhaps we can post it on slack, to double-check with some of our adopters using a large number of namespaces.

cmd/kubeapps-apis/plugins/resources/v1alpha1/namespaces_filtering.go
```diff
@@ -40,44 +25,22 @@ func (s *Server) GetAccessibleNamespaces(ctx context.Context, cluster string, tr
...
 	} else {
-
-		typedClient, err := s.clusterServiceAccountClientGetter.Typed(ctx, cluster)
-		if err != nil {
-			return nil, status.Errorf(codes.Internal, "unable to get the k8s client: '%v'", err)
+		clusterTypedClientFunc := func() (kubernetes.Interface, error) {
+			return s.clusterServiceAccountClientGetter.Typed(ctx, cluster)
+		}
+		inClusterTypedClientFunc := func() (kubernetes.Interface, error) {
+			return s.localServiceAccountClientGetter.Typed(context.Background())
```
beni0888, line 32: Just for the sake of knowledge, I don't understand why in this line you use `context.Background()` while above you're passing the existing `ctx` variable. Could you please explain the reason? 🙏🏻

beni0888 (approved):
LGTM! I've just added some minor suggestions

cmd/kubeapps-apis/plugins/pkg/resources/namespaces.go
castelblanque: Indeed, it will be used by Kapp, Helm and Resources plugins.

cmd/kubeapps-apis/plugins/pkg/resources/namespaces.go
```diff
@@ -0,0 +1,144 @@
...
+		}
+	} else {
+		// If the user can list namespaces, do not filter them
+		return namespaces.Items, nil
+	}
+}
+
+func nsCheckerWorker(client kubernetes.Interface, nsJobs <-chan checkNSJob, resultChan chan checkNSResult) {
+	for j := range nsJobs {
+		res, err := client.AuthorizationV1().SelfSubjectAccessReviews().Create(context.TODO(), &authorizationapi.SelfSubjectAccessReview{
```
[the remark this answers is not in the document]
castelblanque: You are right. This is a copied-pasted piece of code, so the whole block was left as-is.
Changed to `context.Background()`.

cmd/kubeapps-apis/plugins/resources/v1alpha1/namespaces_filtering.go
castelblanque: Most of the times, `clusterServiceAccountClientGetter` will use the context from the incoming request (`ctx` in this case), whereas `localServiceAccountClientGetter` does not need it, as it will provide a so called "in-cluster" client that will use the pod's service account. This local getter will have then different RBAC applied than the incoming context.

PR #5454, opened 2022-10-12
Small changes to check Carvel multicluster support.

absoludity:

### Description of the change

I was keen to see exactly what was required to use the Carvel plugin with the existing multicluster support. It turns out, nothing (maybe @antgamdia already checked this and got it working a while ago? Not sure). The only changes I needed were:

* Updating the version of kapp-controller that we run in the cluster (since v0.35 was still requiring secrets attached to service-accounts, which stopped with k8s 1.24, and updated in kapp-controller with v0.37, but I've tested with the latest which works).
* Adding a makefile target to add kapp-controller to the second dev cluster
* Fixing an issue in our carvel docs for creating the reconciliation service account.

With those changes, I can deploy kubeapps on the main cluster, with the patched config:

```diff
diff --git a/site/content/docs/latest/reference/manifests/kubeapps-local-dev-values.yaml b/site/content/docs/latest/reference/manifests/kubeapps-local-dev-values.yaml
index 3cebfad41..56004b56b 100644
--- a/site/content/docs/latest/reference/manifests/kubeapps-local-dev-values.yaml
+++ b/site/content/docs/latest/reference/manifests/kubeapps-local-dev-values.yaml
@@ -32,3 +32,8 @@ ingress:
     nginx.ingress.kubernetes.io/proxy-buffers: "4"
     # Required for ingress-nginx 1.0.0 for a valid ingress.
     kubernetes.io/ingress.class: nginx
+packaging:
+  helm:
+    enabled: true
+  carvel:
+    enabled: true
```

as well as the reconciliation service account as per the carvel docs modified here. I can then login, switch to the second cluster, then browse the catalog to see available apps including the carvel ones available on that cluster:

carvel-catalog-cluster-2

I can then select one to deploy

carvel-deploy-cluster-2

and successfully deploy it

carvel-deployed-second-cluster

as well as delete it.

### Benefits

Demonstrates that the carvel plugin can be used with the current multicluster support.

### Possible drawbacks

None

### Applicable issues

- fixes #

### Additional information

Interestingly, we had code that may have been intended to say that this was not supported, when creating a carvel app on a second cluster:

https://github.com/vmware-tanzu/kubeapps/blob/c4d25a4151fd1b5c9c86fe4ecd2b41abf33e9317/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_ctrl_packages.go#L588-L590

but what it actually does is ensure that the package being installed is from the same cluster as that being targeted, which is the case here anyway (since we're browsing available packages on the second cluster). So it allows deploying the package  from the second cluster on the second cluster, but will correctly stop a situation where the API is used to deploy a package from the first cluster on the second cluster.

### DIFF
--- a/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_adapters.go
+++ b/cmd/kubeapps-apis/plugins/kapp_controller/packages/v1alpha1/server_data_adapters.go
@@ -364,7 +364,7 @@ func (s *Server) buildPkgInstall(installedPackageName, targetCluster, targetName
 	// Ensure the selected version can be, actually installed to let the user know before installing
 	elegibleVersion, err := versions.HighestConstrainedVersion([]string{pkgVersion}, vendirversions.VersionSelection{Semver: versionSelection})
 	if elegibleVersion == "" || err != nil {
-		return nil, status.Errorf(codes.InvalidArgument, "The selected version %q is not elegible to be installed: %v", pkgVersion, err)
+		return nil, status.Errorf(codes.InvalidArgument, "The selected version %q is not eligible to be installed: %v", pkgVersion, err)
 	}
 
 	pkgInstall := &packagingv1alpha1.PackageInstall{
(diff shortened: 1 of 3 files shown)


antgamdia (approved):
Thanks!

PR #5461, opened 2022-10-13
[Dashboard] Fix for error flashing when deleting repo

### LINKED ISSUE #4220: UI error flashes when trying to delete a repository without permissions
### Description:

When a user that does not have permissions to delete repos, tries to do so in a non-global repo, the UI error appears for milliseconds and then disappears. Not possible to read it.

1. Login as a user with permissions to create a repo in a non-global namespace (e.g. `default`)
2. Create a non-global repo in that namespace
3. Login as a user without permissions to delete a repo but with view permissions on the namespace
4. Try to delete the repo, and confirm when prompted in the pop-up

- Repo is not deleted -> That is correct, because user does not have permissions.
- In practice there is no UI error message. The error message flashes but for milliseconds.

User is shown an error message that can be seen and read.

### Additional information

Perhaps the ideal situation would be that a user like this should not even be able to click on the "Delete" button if it doesn't have permissions? I mean, seeing the button disabled like the global repos buttons.
image

castelblanque:

### Description of the change

This PR fixes a problem in the package repositories deletion action. When the user didn't have enough permissions to delete the repo (or any other error at deletion), the error message was shown flashing and disappearing.
This was confusing: user was seeing an error for a fraction of a second, and the repository not deleted.

The user now can see the error message, after the confirmation popup is automatically closed:

image

Repos are now re-fetched only when there is no error in the deletion action.

### Benefits

User is able to see the actual error message and repos are not unnecessarily re-fetched.

### Possible drawbacks

N/A

### Applicable issues

- fixes #4220

### DIFF
--- a/dashboard/src/components/Config/PkgRepoList/PkgRepoControl.test.tsx
+++ b/dashboard/src/components/Config/PkgRepoList/PkgRepoControl.test.tsx
@@ -39,6 +39,10 @@ const defaultProps = {
 
 it("deletes the repo and refreshes list", async () => {
   const deleteRepo = jest.fn();
+  // Mock the response for delete repo action
+  const mockDispatch = jest.fn();
+  mockDispatch.mockReturnValue(true);
+  spyOnUseDispatch = jest.spyOn(ReactRedux, "useDispatch").mockReturnValue(mockDispatch);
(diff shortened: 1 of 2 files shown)


antgamdia (approved):
Thanks!

PR #5466, opened 2022-10-17
Repos global list on additional clusters

### LINKED ISSUE #2717: Inconsistent message if viewing app repositories on non-kubeapps cluster
### Description:

When viewing AppRepositories when on a cluster other than the one Kubeapps is installed on, a warning message is correctly displayed warning that app repositories are not currently supported on other clusters:

kubeapps-repos-on-default-cluster

The problem is that this is also displayed when on a Kubeapps install where the user cannot target the cluster on which Kubeapps is installed.

1. Configure Kubeapps so that users can target a second cluster and not target the cluster on which Kubeapps is installed,
2. Click on the AppRepositories menu item

See above screenshot.

Either we:
* should not display the AppRepositories menu (or have it disabled) if the user is unable to view them, or
* consider updating the UX so that users can always see the global app repositories

I think the second option would be the best UX.

devel Kubeapps

castelblanque:

### Description of the change

This PR adds the list of global repos when context is set to a cluster other than the default one.
List is set on read only mode.
Same UI warning message is kept, as it still describes correctly the situation.

image

Targeted at the case stated in #2717, users that do not have access to the main/default cluster where Kubeapps is installed.

### Benefits

- Users in non-default cluster can see the list of global repositories, at least in read-only mode.

### Possible drawbacks

N/A

### Applicable issues

- fixes #2717

### DIFF
--- a/dashboard/src/components/Config/PkgRepoList/PkgRepoList.tsx
+++ b/dashboard/src/components/Config/PkgRepoList/PkgRepoList.tsx
@@ -160,6 +160,30 @@ function PkgRepoList() {
     });
   };
 
+  const getGlobalReposTable = (
+    globalRepos: PackageRepositorySummary[],
+    disableControls: boolean,
+  ) => {
+    return (
+      <>
+        <h3>Global Repositories:</h3>
+        <p>Global Package Repositories are available for all Kubeapps users.</p>
+        {globalRepos.length ? (
+          <Table
+            valign="center"
+            columns={tableColumns}
+            data={getTableData(globalRepos, disableControls)}
+          />
+        ) : (
+          <p>
+            There are no <i>global</i> Package Repositories yet. Click on the "Add Package
+            Repository" button to create one.
+          </p>
+        )}
+      </>
+    );
+  };
+
   /* eslint-disable jsx-a11y/label-has-associated-control */
   return (
     <>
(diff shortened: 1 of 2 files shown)


dashboard/src/components/Config/PkgRepoList/PkgRepoList.tsx
```diff
@@ -172,6 +196,7 @@ function PkgRepoList() {
             namespace={currentNamespace}
             helmGlobalNamespace={helmGlobalNamespace}
             carvelGlobalNamespace={carvelGlobalNamespace}
+            disabled={!supportedCluster}
```
castelblanque, line 199: Also disabled the "Add package repository" button if the cluster is not the default one.

antgamdia (approved):
Great, having a read-only view of the global repos is a good idea. 

What about adding a button to "manage repos in the main cluster" (or sth like that), to let users access to the management view more quickly ?

castelblanque:
> What about adding a button to "manage repos in the main cluster" (or sth like that), to let users access to the management view more quickly ?

Do you mean adding a button like that close the current "Add package repository" (top right)?
I would do it as part of a separate issue, because we should also check upfront which access level the user has. e.g. not show it if the Kubeapps cluster is not accessible to the user, or any needed RBAC is missing for the user.

PR #5480, opened 2022-10-18
Package repository management UI for users with restricted permissions

### LINKED ISSUE #3679: App Repository UI does not properly handle cases where the user does not have read/write to app repos in ns
I created a user that only has access to his/her own namespace, and that does not have the right to view/create private repositories, but only use the global repositories (e.g. because of company policy).

The user can login to Kubeapps, view applications and view the catalog fine.

But when looking at the repository page, the UI shows an error message:
Screen Shot 2021-10-27 at 10 58 29 AM

I would expect that instead, the page should render the global repositories and then have a message in the namespaced section that indicates the user does not have the priviledges to manage repositories.
something like this:
Screen Shot 2021-10-27 at 10 55 36 AM

The screenshot above was taken after granting "read" to repositories in the user's namespace.
We can see that the UI can properly handle disabling the edit/refresh/delete buttons, so it should also be able to handle when the user does not even have read access.

Last, though the repositories edit/refresh/delete are disabled, the "add app repository" and "refresh all" buttons are enabled instead of being disabled.

castelblanque:
### Description of the change

This PR improves the repositories management screen for the cases of users with restricted permissions.
Before, there was a single `disableControls` flag for all repositories, regardless of the plugin.

Now this PR introduces a function `PackageRepositoriesService.getRepositoriesPermissions` that checks the permissions for repos CRDs per plugin. It does so by invoking multiple `can-i` requests, which could be improved by adding a new, on purpose, backend API.

This permissions check allows to enable/disable edit controls at the repo level and not globally as before.
For example, a user that:
- Does not have permissions to global repos
- Has read only permission to Helm repos in a namespace
- Has update permission to Carvel repos in a namespace
- Has no creation permissions for any repo type
will see a screen like this:
image

Also, after #5453 and #5460, users with no cluster-wide permissions but with permissions in a specific ns, can see the repositories list without an alert error.

### Benefits

- Users will have a better feedback on the actions that they can do in UI when it comes to managing repositories

### Possible drawbacks

N/A

### Applicable issues

- fixes #3679


ppbaena:
Thanks for this enhancement to have better feedback from the UI according to your permission.

dashboard/src/components/Config/PkgRepoList/PkgRepoList.tsx
```diff
@@ -87,13 +99,57 @@ function PkgRepoList() {
...
+      configuredPlugins
+        .map(plugin =>
+          PackageRepositoriesService.getRepositoriesPermissions(
+            currentNamespace,
+            pluginsConfig,
+            plugin,
+          ),
+        )
+        .filter(i => i !== undefined),
+    ).then(results => {
```
antgamdia, line 118: what about using the `await` keyword instead like in other places? 

also, we should have a `.catch` to handle any promise error, shouldn't we?

dashboard/src/components/Config/PkgRepoList/PkgRepoList.tsx
```diff
@@ -87,13 +99,57 @@ function PkgRepoList() {
...
+      results.forEach(r => {
+        if (r !== undefined) {
+          rbac.set(JSON.stringify(r.plugin), r);
+        }
+      });
+      setReposRBAC(rbac);
+    });
+
+    // Cluster-wide check
     Kube.canI(cluster, "kubeapps.com", "apprepositories", "list", "")
```
antgamdia, line 129: unrelated to this PR, but just curious... what this `canSetAllNS` is for?  It seems it's coupled with the helm plugin, no?

dashboard/src/shared/PackageRepositoriesService.ts
```diff
@@ -282,4 +292,84 @@ export class PackageRepositoriesService {
         throw convertGrpcAuthError(e);
       });
   }
+
+  public static getRepositoriesPermissions = (
+    namespace: string,
+    config: IConfig,
+    plugin: Plugin,
+  ): IPackageRepositoryPermission | undefined => {
+    switch (plugin.name) {
```
antgamdia, line 301: I think the answer is no... but: can we obtain these strings from the plugin backend somehow?  Like, what if `packaging.carvel.dev` get eventually renamed?

antgamdia (approved):
Woah, interesting change! Fine-grained permissions finally, excellent!

dashboard/src/components/Config/PkgRepoList/PkgRepoList.tsx
castelblanque: It was a previously existing flag. It controls if the toggle to show repos from all namespaces is hidden.
image
Making this feature cross-plugin could be a task for the new UI #5515.

dashboard/src/shared/PackageRepositoriesService.ts
castelblanque: There is a comment exactly about that a little bit below.
We could even add a new API endpoint to provide the permissions of the user for e.g. repos.
In this way the UI would not need to know anything about resource group, name, verb, etc.

PR #5504, opened 2022-10-19
[Helm plugin] Remove unnecessary context check

castelblanque:

### Description of the change

Trivial PR removing unneeded checks for provided context cluster.
Cluster is not strictly required, as it will fall back to the Kubeapps cluster if not specified. See [code in charge here](https://github.com/vmware-tanzu/kubeapps/blob/main/cmd/kubeapps-apis/core/plugins/v1alpha1/plugins.go#L328-L330).

Otherwise, an error like this was returned:

```bash
ERROR:
  Code: InvalidArgument
  Message: Invalid GetPackageRepositorySummaries response from the plugin helm.packages: rpc error: code = InvalidArgument desc = no valid context provided
```

### Benefits

Specify cluster context for operations in Helm plugin is not required.

### Possible drawbacks

N/A

### Applicable issues

N/A


antgamdia (approved):
Thanks!

PR #5524, opened 2022-10-20
Fix the jobs sync from/to bitnami in the GHA's "kubeapps general" workflow

beni0888:
### Description of the change

The changes in this PR are aimed at fixing the issues with the GHA's jobs `sync_from_bitnami` and `sync_to_bitnami` that have been failing so far. The main problem was that the GHA runner wasn't able to connect to the required GitHub repositories `kubeapps-bot/charts` and `bitnami/charts`, mainly because in our scripts we were using the `HTTPS` protocol for defining the remotes for those repositories, and that made the `git` commands to ask for credentials instead of using the corresponding SSH keys. The reason why this wasn't failing in GHA workflow and not in the CircleCI one, is that for some unknown reason (at least for me, although I suspect it has to do with the use of the [checkout step](https://circleci.com/docs/configuration-reference/?utm_source=google&utm_medium=sem&utm_campaign=sem-google-dg--emea-en-dsa-maxConv-auth-nb&utm_term=g_-_c__dsa_&utm_content=&gclid=Cj0KCQjw48OaBhDWARIsAMd966AI5kBSd6PAWCeyIe3UrgXtD9tDgmHwF1Q8aB3AFdYZV7AeVCweAMoaAid9EALw_wcB#checkout)), the CircleCI runner has a specific `~/.gitconfig` setup as follows: 
```
...
[url "ssh://git@github.com"]
	insteadOf = https://github.com
... 
```
So it was transparently using `git over ssh` protocol and taking advantage of the configured keys in the runner.

Besides that, in this PR I have taken advantage to document some of the functions used by the  `bash` scripts that support those jobs and refactor them (mainly renaming variables) to try to better convey the actual intention and responsibility of each script and function (although I cannot be 100% sure I got my goal).

Also, during the development/debugging/head-hitting-against-the-wall process, I discovered that we have added some unnecessary/unused SSH keys to the jobs, so I have removed the associated code and, hopefully, some complexity from the workflow.

Finally, while working on this, I detected a bug in the code of the `setup` job in charge of setting the value for the `triggered_from_fork` output variable, so it's been fixed and a previous step has been added to that job, sending to the output the value of the `github.event` that triggered the workflow, just with debugging purposes.

### Benefits

* The jobs `sync_from_bitnami` and `sync_to_bitnami` from GHA's `kubeapps general workflow` work as expected.
* Some pieces of the code have been refactored to make their intention and functionality clearer 🤞🏻 .
* Some complexity has been removed.
* Our CI scripts are now more robust, not relying on accidental/hidden Git configuration.
* The `setup` job's output variable `triggered_from_fork` is now set as expected.

### Possible drawbacks

* Even though I think I have thoroughly tested the changes, some unforeseen side-effects/bugs could appear.

### Applicable issues

- related to #4436


castelblanque (approved):
Excellent. Thanks for the investigation, the fix and the improvements!

PR #5528, opened 2022-10-20
Avoid docker production images overlapping between GHA and CircleCI

beni0888:

### Description of the change

Right now, when Docker production images are generated from the GHA kubeapps general workflow, those are pushed to Docker Hub in the `kubeapps` workspace (eg. `kubeapps/kubeapps-apis`, `kubeapps/dashboard`, etc) and tagged with `latest` or the release version `x.y.z`,  so they overlap with the ones generated in the CircleCI workflow that runs in parallel (until we finish the migration to GHA and decommission the CircleCI workflow). To avoid images from both workflows to overwrite each other, this PR adds a `-gha` suffix to the tag for the images generated from the GHA workflow (eg. `kubeapps/dashboard:latest-gha)`.

### Benefits

Production images generated from each of the different workflows are clearly identifiable and don't overwrite each other.

### Possible drawbacks

In case you want to use the images generated from GHA, you'll have to use a tag that don't follows the standard convention.

### Applicable issues

- related to #4436

### DIFF
--- a/.github/workflows/kubeapps.yaml
+++ b/.github/workflows/kubeapps.yaml
@@ -508,6 +508,10 @@ jobs:
           password: ${{ secrets.DOCKER_PASSWORD }}
       - uses: actions/download-artifact@v3
       - run: |
+          if [[ "${DEV_MODE}" == true && "IMG_PROD_TAG" == latest ]]; then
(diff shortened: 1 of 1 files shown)


castelblanque (approved):
Great, thanks!

Issue #5468, opened 2022-10-17
Introduce multicluster proxy mode for Kubeapps APIs


castelblanque:
**Summary**
Introduce a new setup for installing Kubeapps in which Kubeapps APIs can act as a proxy for all incoming requests to another Kubeapps APIs.

**Background and rationale**
As decided in the [Multicluster redesign document](https://docs.google.com/document/d/1qYHIa5dd9qwxTdzyNUn7yXacwmYjH7R-KnnWxXkPItE/edit), Kubeapps will be required to be present in all workload clusters of a multicluster setup. At least Kubeapps APIs with plugins.
The Kubeapps installation in charge of managing the multicluster (acting as a control plane) will have a "proxy" mode in which al business-related requests (e.g. create repo, list available packages, etc.) will be proxied as-is to the selected workload/target cluster. Target cluster will expose a Kubeapps APIs that will receive those proxied requests, not needed to expose the K8s APIs endpoint.

**Acceptance criteria**
- Clusters set up for multicluster (`clusters:` value in chart) allow to provide, per cluster, the Kubeapps endpoint and not the K8s API endpoint.
- Chart has a flag value to start Kubeapps APIs with proxy mode enabled.
- Some selected requests to Kubeapps APIs are proxied to the target cluster, instead of being processed internally.
- All possible operations are computed by the target Kubeapps APIs and not the source Kubeapps APIs.

PR #5529, opened 2022-10-20
Add Kubeapps API proxy mode

castelblanque:
### Description of the change

This PR introduces a selective proxy mode for Kubeapps APIs.
With the chart value `kubeappsapis.proxyMode.enabled: true` or arg flag `--proxy-mode=true` the proxy mode will be enabled. Defaults to false, so nothing will change in operations.

When enabled, not all requests will be proxied. Only those with header `x-kubeapps-proxy`. The value of the header will be the cluster name to which the requests will be proxied.
Internally, starting with that header value, cluster details are retrieved and request is sent to the ingress of that cluster.
Reason behind this approach is to avoid having to unnecessarily parse the content of the messages to find out the `context`.

Requests will be proxied to a GRPC endpoint, not to the K8s API server anymore.

New optional values added for chart (per cluster):
- `.clusters[x].ingress.endpoint` must point to the target cluster ingress. Ingress with GRPC capabilities will have to be enabled. (PR #5543 to enable Kubeapps in API-only mode)
- `.clusters[x].ingress.certificateAuthorityData` is the optional CA data to perform the TLS handshake. It won't be the one from the K8s API service, but the one set at the ingress.

For example, requests will continue to be processed locally with e.g.

`grpcurl -H "Authorization: $TOKEN" $APIHOST ....RepositoriesService.GetPackageRepositorySummaries`

The above will return the repos in the cluster at `APIHOST`.

But requests will be proxied if Kubeapps is set up with a cluster named `remote-cluster` and the header is present:

`grpcurl -H "x-kubeapps-proxy: remote-cluster" -H "Authorization: $TOKEN" $APIHOST ....RepositoriesService.GetPackageRepositorySummaries`

will return the repos in the target cluster (`remote-cluster`).

### Benefits

GRPC requests will be proxied to another cluster where Kubeapps is installed.

### Possible drawbacks

- Operators might get confused when setting up the target workloads, as it needs to have a valid ingress for GRPC.
- Target cluster is based on the content of the header `x-kubeapps-proxy`, so if a cluster is specified in the `context.cluster` payload, it is different from the one specified in the header, and that cluster name does not exist in the target cluster, request will fail.

### Applicable issues

- implements #5468

### Additional information

In order to get a valid Kubeapps workload setup, changes introduced in #5543 are helpful.


absoludity:
Before looking at the code too closely, a couple of questions:

> When enabled, not all requests will be proxied. Only those with header `x-kubeapps-proxy`. The value of the header will be the cluster name to which the requests will be proxied.
> Internally, starting with that header value, cluster details are retrieved and request is sent to the ingress of that cluster.
> Reason behind this approach is to avoid having to unnecessarily parse the content of the messages to find out the `context`.

Who then *writes* the header value? Wouldn't this mean that the dashboard (or client requesting the info) now needs to be aware of the architecture decision? They should not need to be aware IMO... they just make the request using the normal grpc (or http) client and Kubeapps handles it either in proxy mode or in non-proxy mode? Keen to understand why you think this short-cut is worth making the decision a client responsibility?

> * `.clusters[x].ingress.endpoint` must point to the target cluster ingress. Ingress with GRPC capabilities will have to be enabled.

Do you mean the Kubeapps ingress on the target cluster? Or otherwise, what is the cluster ingress that helps you find Kubeapps? And if that is what's meant here, does that now mean that Kubeapps ingress will also now be required on workload clusters? (current multicluster doesn't require any ingress other than access to k8s API server - quite an extra overhead). I don't see any mention of ingress in the doc, but maybe it was chatted about in the meeting. It'd be worth including the explicit requirements such as ingress on workload clusters, clients setting headers for the decision to proxy (I think I'd be a strong -1 on that one, for eg., but again, happy to step back).

castelblanque:
> Keen to understand why you think this short-cut is worth making the decision a client responsibility?

It was done like this mainly to avoid having to analyze each payload in order to check what is the target cluster. It would mean parsing the payloads assuming that the data is always under `context.cluster`.
If backend doesn't have the header nor parses the payload there is no way to know what is the target cluster in an GRPC call. There is at least one case in which requests must not be proxied: requests that target the "manager" cluster itself. Therefore we can not enable just a "proxy all" mode, we need to check what is the target cluster and decide where to forward the request.

With this approach, we can also allow for a hybrid scenario, where some worker clusters could have Kubeapps while others don't. This is a question that has come up recently and we still need to investigate and decide.

I agree that we should avoid the header, do you agree on trying to parse the payload and decide based on that?

> Do you mean the Kubeapps ingress on the target cluster?

Yes, Kubeapps ingress. But not the default one we ship. I realized while developing the feature, that current ingress setup does not allow for GRPC calls, just REST and GRPC-web. That is why I created #5543, and it could be simplified to just GRPC and no REST. This is needed until ingresses for K8s sort out the GRPC handling properly.
If we removed the need to expose direct access to K8s API in Kubeapps (except for operators), I think it is consistent to remove the need also here. With just the API ingress deployed by the Kubeapps chart, the workload can be set up and no access to K8s API is required.
However, we still have the scenario of workloads in which operators want to have the Dashboard. That we don't yet support. We should then come up with a full-ingress setup that allows for GRPC, GRPC-web and REST. GRPC requires to hold the `/` path so maybe using `pathType: Exact` or serving the `dashboard` in another path?

castelblanque:
> using a header means the client decides, which we'd want to avoid.

Yes, but the drawback of not having a header is that we need to inspect every single incoming request to check for a target cluster. Even if 100% of requests target the local cluster.
Requests coming in might have two destinations:
- Local cluster (not proxied)
- Proxied cluster (different cmux handler)
We need a marker to decide where to forward the request.

I'll make the changes to remove the header, look for the cluster in the payload and build a new cmux `Matcher` for that purpose.

> Can gRPC-Web be used when proxying the call? 

We could give it a try. Although I thought that grpc-web is for communication of web UI clients. All I can find are clients for React/TS, and just an [abandonded Go client](https://github.com/ktr0731/grpc-web-go-client). I doubt that building machinery for the grpc-web inter-backend communications is something that we want to do?
Regarding this, a question comes to my mind: why are we supporting plain GRPC in Kubeapps? Are we using that protocol anywhere if there is no ingress for it? We could just stick to grpc-web and remove the rest. Not even REST if we follow the same logic as for GRPC.

> Let us know when you've updated removing the dependence on a header, to be able to review without reviewing unnecessarily?

Yep

absoludity:
> Yes, but the drawback of not having a header is that we need to inspect every single incoming request to check for a target cluster.

Sure - but how can we have a header without tying the implementation detail of how requests are proxied  into the client request? If you've got an idea or a way for the header to be somehow inserted enroute without being a client decision - maybe? But I don't yet see how that can be (without parsing the content at some point).

> Even if 100% of requests target the local cluster.

The server already parses the whole request when handling the query... but yes, I think you mean that you're intending to do the proxy much earlier before that's happened.

> Requests coming in might have two destinations:
> 
>     * Local cluster (not proxied)
> 
>     * Proxied cluster (different cmux handler)
>       We need a marker to decide where to forward the request.

Yes - that's worth thinking about where that would be. If we could hook in after the gRPC framework has parsed the request but before it gets to our handler, then there'd be no extra cost for the parsing as it's already been done, no extra cost if it's local (other than a single condition).
 
> I'll make the changes to remove the header, look for the cluster in the payload and build a new cmux `Matcher` for that purpose.

If that's the best way - that's where I think more thought or prototyping might be useful. It may be that a new cmux `Matcher` could be worth the extra expense of parsing the request, or it may be that there's somewhere else we can hook in to the gRPC handling after the request is parsed but before it gets to the handler. I just think it's worth avoiding jumping straight to a solution before evaluating the possibilities (though maybe you already have - but I didn't see it in the doc).

> 
> > Can gRPC-Web be used when proxying the call?
> 
> We could give it a try. Although I thought that grpc-web is for communication of web UI clients. All I can find are clients for React/TS, and just an [abandonded Go client](https://github.com/ktr0731/grpc-web-go-client). I doubt that building machinery for the grpc-web inter-backend communications is something that we want to do?

I wouldn't think so either, no. It was just a thought if we had issues getting the gRPC ingress. If it works fine, then no need to pursue this.

> Regarding this, a question comes to my mind: why are we supporting plain GRPC in Kubeapps?

We don't need to do anything to support it - it just comes supported because the grpc-web requests are translated to that, but right, we've not exposed it externally. I did use it briefly while looking for ways for envoy's ingress to work.

> Are we using that protocol anywhere if there is no ingress for it? We could just stick to grpc-web and remove the rest.

The grpc-web requests are translated to gRPC for kubeapps-apis to handle.

castelblanque:
Update on the implementation of payload inspection:
- After some investigation, realized that `cmux.MatchWriter` can't be used as it is evaluated too early in the process. Payload body is parsed and available later.
- Trying now with a chained `...ServerInterceptor`: request is correctly proxied when the response type is known upfront. Need to make it generic.
- Still need to check how it will behave with the streamed `GetResources` method

castelblanque:
Tried different approaches, but there are some factors that make the proxying very tricky when having to take the payload as deciding piece for proxying.
- One by one inspection of request objects (using UnaryInterceptor in this case) leads to code hard to maintain:
   ```go
    // Context cluster might be hidden in any of these
    type MessageWithContext interface {
	    GetContext() *packagesGRPCv1alpha1.Context
    }
    type MessageWithTargetContext interface {
	    GetTargetContext() *packagesGRPCv1alpha1.Context
    }
    type MessageWithInstalledPackageRef interface {
	    GetInstalledPackageRef() *packagesGRPCv1alpha1.InstalledPackageReference
    }
    type MessageWithAvailablePackageRef interface {
	    GetAvailablePackageRef() *packagesGRPCv1alpha1.AvailablePackageReference
    }
    type MessageWithPackageRepoRef interface {
	    GetPackageRepoRef() *packagesGRPCv1alpha1.PackageRepositoryReference
    }
   ```
- Some operations do not have context at all (e.g. `GetConfiguredPlugins`), which totally make the payload body decision approach a bit inconsistent.
- Cmux `Matcher` is evaluated too early in the process, before the payload is accessed. It only provides a pure `io.Reader` for HTTP2 data. Fine for using headers to decide.
- Streaming. For Unary calls we might make it, but for Streaming, it is much complicated to tackle through interceptors. Available input is the `grpc.ServerStream`, and messages are retrieved trough a goroutine that continuously checks for `RecvMsg`. [See here](https://github.com/mwitkow/grpc-proxy/blob/master/proxy/handler.go#L144). The point is that the target stream to proxy to (if needed) must be based on the message delivered.
- GRPC server interceptors: as mentioned above, interceptors are at the `ServerStream` level, but we want proxying based on the content of the incoming messages. Unclear to me how to redirect the message at the point of receiving a message inside the go routine.
- Maybe we can build a valid `grpc.ClientConn` out of a received message at some point. Then by analyzing the method name we might be able to map it to a valid `*ServiceClient` e.g. [RepositoriesServiceClient](https://github.com/vmware-tanzu/kubeapps/blob/main/cmd/kubeapps-apis/gen/core/packages/v1alpha1/repositories_grpc.pb.go#L36).

I think that the original approach of proxying based on a header might not be so wrong after all.
However, given that I am stuck in exploring, I will stop the task by now. New ideas are welcome :)

absoludity:
Thanks for the overview Rafa. Starting with your last comment first:

> I think that the original approach of proxying based on a header might not be so wrong after all.

Just want to be clear that I'm not against proxying based on a header at all, but the related decision of *who* adds those headers does concern me. That is, if you have a way for clients to use our API without knowledge of how the multicluster support is implemented (such as proxying), so they just issue a request as normal and when the request gets to kubeapps-apis it has the header you want - then great!

But if you mean that clients themselves have to use knowledge about the multi-cluster implementation (ie. a client/user must know that the request will be proxied and that they have to set a header in the request to ensure that happens), then I do have some concern because we're mixing responsibilities in a way that ties our client to a specific backend implementation. But I don't want to get in the way here given the current context etc.. If @antgamdia is on board with this direction, feel free to move forward.

> However, given that I am stuck in exploring, I will stop the task by now. New ideas are welcome :)

Perhaps we could modify the grpc-web client used in the dashboard so that it adds the header... still not great, since it'd need to be done manually for an http request, but at least would mean the user is not involved in the decision. And it'd still have some of the difficulty you mention below (finding the non-standard context).

Another (imperfect) option could be to (somehow) use the grpc-web URL to extract the target and insert into a header. This would only be present if grpc-web was used (which it always is for the dashboard).

Some other thoughts to your points below.

> Tried different approaches, but there are some factors that make the proxying very tricky when having to take the payload as deciding piece for proxying.
> 
>     * One by one inspection of request objects (using UnaryInterceptor in this case) leads to code hard to maintain:
>       ```go
>        // Context cluster might be hidden in any of these
>        type MessageWithContext interface {
>            GetContext() *packagesGRPCv1alpha1.Context
>        }
>        type MessageWithTargetContext interface {
>            GetTargetContext() *packagesGRPCv1alpha1.Context
>        }
>        type MessageWithInstalledPackageRef interface {
>            GetInstalledPackageRef() *packagesGRPCv1alpha1.InstalledPackageReference
>        }
>        type MessageWithAvailablePackageRef interface {
>            GetAvailablePackageRef() *packagesGRPCv1alpha1.AvailablePackageReference
>        }
>        type MessageWithPackageRepoRef interface {
>            GetPackageRepoRef() *packagesGRPCv1alpha1.PackageRepositoryReference
>        }
>       ```

Yeah, there's no simple way to know where the context is within a protoc-generated request message (it's not as simple as having a standard location in the proto def, given some requests have a target context differing from the request context).  As you found, we can't write an interface implementation for the request structs because the structs are defined in a different generated module. We could write a protoc plugin (see the [validators plugin](https://medium.com/swlh/know-your-inputs-or-grpc-request-validation-8eb29a0ebc31)) that adds an extra function to each generated request (based on annotations in the proto file), but I'm not sure if that's worthwhile - it may be worth a look. A simpler way to encapsulate the decision in the one place might be to have a `fn GetContextForRequestType(t interface{}) v1alpha1.Context` and use reflection or similar - not sure.

Note: isn't this same decision present if the client/user has to put the context/cluster in the header, we've just deferred the responsibility to the client or user? (but yes, it seems simpler for us if the user to decides the proxy destination for us, just not sure they should be as above).

> 
>     * Some operations do not have context at all (e.g. `GetConfiguredPlugins`), which totally make the payload body decision approach a bit inconsistent.

Not sure why you see this as inconsistent? If there's no context, there's no proxying. I assume we'd do the same if using the header method and the header wasn't present for the call to `GetConfiguredPlugins`? (Or I guess you might decide instead to do a BadRequest too - not sure)

> 
>     * Streaming. For Unary calls we might make it, but for Streaming, it is much complicated to tackle through interceptors. Available input is the `grpc.ServerStream`, and messages are retrieved trough a goroutine that continuously checks for `RecvMsg`. [See here](https://github.com/mwitkow/grpc-proxy/blob/master/proxy/handler.go#L144). The point is that the target stream to proxy to (if needed) must be based on the message delivered.

I don't think that last sentence is accurate? I think what you're looking at in that loop of the code you linked to is the streamed messages for a single client request being forwarded back to the client by the proxy. That function would be called *for each client request* separately, and so is only forwarding the streamed messages for that one request back - there's no decision of which "target stream to proxy to", I don't think?

> 
>     * GRPC server interceptors: as mentioned above, interceptors are at the `ServerStream` level, but we want proxying based on the content of the incoming messages. Unclear to me how to redirect the message at the point of receiving a message inside the go routine.

I don't think you'd need to do any redirection within the go routine within an equivalent of `forwardServerToClient`, as above.

> 
>     * Maybe we can build a valid `grpc.ClientConn` out of a received message at some point. Then by analyzing the method name we might be able to map it to a valid `*ServiceClient` e.g. [RepositoriesServiceClient](https://github.com/vmware-tanzu/kubeapps/blob/main/cmd/kubeapps-apis/gen/core/packages/v1alpha1/repositories_grpc.pb.go#L36).

This is for the same not-a-problem I think? Also, just remember we (well, grpc-web) only support server streaming. We don't do any client streaming (nor do we have a need to).

Hope the above helps. Again, I've no problem with using a header to decide on the proxy target, my only concern is with who is responsible for adding that header (ie. I think if we have users/clients being responsible for setting that it may make future changes more difficult - the same concern I have for proxying these requests to kubeapps instances on other clusters in the first place). And even then, if you and @antgamdia are in agreement, I won't be blocking.

absoludity:
Just closing for now until this work gets picked up / prioritised again.

PR #5530, opened 2022-10-20
Add JSON Schema editor view

### LINKED ISSUE #3535: Easy way to provide custom values.schema.json files for Helm Charts
### Description:
This is not an issue, rather than a nice-to-have.

Currently, Kubeapps reads the `values.schema.json` file, from the helm chart. There are some public charts, without such file, or one that I (as a final user) want to modify according to my needs. 

Workaround for this is to customize the helm chart only to add my custom file, publish it to a chart repo like Harbor (and deploy it if needed), and add this custom Harbor to my kubeapps.

Would be very nice, if we could overwrite this file, either during `deploy` or by providing a repo, dedicated only for such files, with some predefined structure.

antgamdia:
### Description of the change

The basic forms view has been recently revamped with plain JSON schema support. However, most of the charts (and maybe other packages) do not bring any schema file as part of their bundle.

#3535 brings some ideas on how to solve it, but, this PR is just paving the way with a very simplistic approach: let users edit/add a schema from a text editor.

Currently, there is no way to upload a file externally, but if this happens to be helpful, could be added in a follow-up PR.

### Benefits

Users will be able to edit/add json schemas for improving the "basic form" experience (validations, defaults, etc...).

### Possible drawbacks

Important: this "edited (or added)" schema is NOT being used on the backend side. Meaning you can't expect a proper validation or any further schema enforcement. Using the schema in the backend as well would imply some extra changes that could be performed in a follow-up PR if really needed.

### Applicable issues

- fixes #3535

### Additional information

> **Note**
> This PR is part of a series of PRs aimed at closing [this milestone](https://github.com/vmware-tanzu/kubeapps/milestone/27). I have split the changes to ease the review process, but as there are many interrelated changes, the tests will be performed in a separate PR (on top of the branch containing all the changes).
>  PR 1 out of 6

PR-1


castelblanque:
Excellent, looks really great.
Just a couple of questions:
- Maybe seeing yet another tab named "Schema editor (advanced)" is a bit intimidating for some users. Also there might be admins that do not want Kubeapps users to mess around with the schema. Why not showing it based in a chart feature flag? (e.g. `dashboard.advancedMode.enabled` or similar).
- Why the user editing the schema has to proactively push the "Update schema" button? Can't we do it so that the new schema is automatically applied e.g. when textarea or tab loses focus?

absoludity:
> #3535 brings some ideas on how to solve it, but, this PR is just paving the way with a very simplistic approach: let users edit/add a schema from a text editor.

Hi Antonio. What's the use case for this change? The PR seems to address a user scenario something like:

```
As a Kubeapps user
I want to edit a chart schema each time that I deploy an app
So that I can use a simplified form to deploy each app
```

I'm not sure, but that doesn't seem too realistic? I would have thought the most useful scenario is

```
As a Kubeapps operator
I want to edit a chart schema once
So that every time a Kubeapps user installs that chart the updated schema is used for an improved simplified form.
```

or at least

```
As a Kubeapps user
I want to edit a chart schema once when deploying an app
So that every future deployment of that app uses my updated schema.
```

I know you said this is just a simplistic approach, but I don't really see much use in having this UI unless it somehow addresses at least one of those two use-cases (and that could be tricky - I mean, we can easily store a schema in a config map, but there's no concept of a per-user place to store such a config map etc.)

> 
> Currently, there is no way to upload a file externally, but if this happens to be helpful, could be added in a follow-up PR.

I'd be keen to only land this PR when that follow-up is designed, tested and ready, otherwise we're cluttering the UX just to demo part of a feature that's not yet beneficial.

> ### Benefits
> 
> Users will be able to edit/add json schemas for improving the "basic form" experience (validations, defaults, etc...).

but they'll need to do this *every time* that they deploy a chart, right? Not sure that's much of a benefit... if I know how to do that every time, it'd be simpler for me to use the json I think? Sorry if I missed something where this was discussed earlier (or code that actually is saving the schema changes)!

antgamdia:
Thank you both for the feedback here! 

> Maybe seeing yet another tab named "Schema editor (advanced)" is a bit intimidating for some users. Also there might be admins that do not want Kubeapps users to mess around with the schema. Why not showing it based on a chart feature flag? (e.g. dashboard.advancedMode.enabled or similar).

100% agree. This feature is not useful for everybody, just for those users (or package authors) wanting to play around with the schema... but once we're done with the schema, there's no point in allowing its modification.

Perhaps adding a `dashboard.schemaEditor.enabled` defaulted to `false` makes the trick. Will do it in a follow-up PR.

> Why the user editing the schema has to proactively push the "Update schema" button? Can't we do it so that the new schema is automatically applied e.g. when textarea or tab loses focus?

Good point; in fact, I did implement it that way. However, the experience is not as good as expected. First of all, rendering the schema is time-expensive, so rendering it on every `onChange` is not viable. 
Let's suppose we do it on every tab focus, as you suggested: that, indeed, works, but while playing around with it, I felt it was kind of misleading: sometimes, when the schema is invalid, it wasn't sure whether or not it was being applied. Also, I wanted to have a look at the former schema before actually loading the new one.

To make the edit/save process as easy as possible, I added both the "update schema" button, a context menu button and a shortcut (crtl/cmd+s) to save it. I'd rather keep it that way instead of auto-saving it, but no strong opinion though.

> What's the use case for this change? The PR seems to address a user scenario something like (...)

Totally right, the current use case is not realistic at all. We just went implementing that way (and right now) as it was really useful for us when developing the new basic from view (adding new validations, new fields, etc with kind of a "hot reload").

> I know you said this is just a simplistic approach, but I don't really see much use in having this UI unless it somehow addresses at least one of those two use-cases

I also see little benefit in deploying this change for every user, as Rafa also pointed out. Putting this behind a feature flag could be a good tradeoff: we ship the feature for those users or package authors wanting to create/edit a schema but we don't for a default kubeapps installation.

That said, I believe this PR could still be really beneficial in some cases: imagine an organization using charts/packages, whose schema is really basic or even non-existent, from a repo X . Currently, unless they mirror the repo X, there is no way to "override" the schema.
With this PR, operators from this organization can just provide the schema, manually, to the users, so that they can just paste it onto Kubeapps and start filling up the basic form.
This is far from the very ideal situation, but yet solves a potential problem.


To sum up, I'd rahter move forward with this PR, but:
  1. putting the code behind  `dashboard.schemaEditor.enabled=true` flag
  2. considering the work herein performed for the design in #5560

WDYT?

absoludity:
> With this PR, operators from this organization can just provide the schema, manually, to the users, so that they can just paste it onto Kubeapps and start filling up the basic form. This is far from the very ideal situation, but yet solves a potential problem.

Yeah - I don't think the above is realistic, but agree with the other points so...

> 
> To sum up, I'd rahter move forward with this PR, but:
> 
>     1. putting the code behind  `dashboard.schemaEditor.enabled=true` flag
> 
>     2. considering the work herein performed for the design in

+1 . I'll plan to go through the code later today before finishing.

dashboard/src/components/DeploymentForm/DeploymentFormBody/BasicDeploymentForm/TabularSchemaEditorTable/Params/SliderParam.tsx
```diff
@@ -6,77 +6,106 @@ import { CdsInput } from "@cds/react/input";
...
+    param?.multipleOf || (param.schema?.type === "number" || param?.type === "number" ? 0.5 : 1);
+
   const onChange = (e: React.FormEvent<HTMLInputElement>) => {
     setCurrentValue(toNumber(e.currentTarget.value));
-    setIsValueModified(toNumber(e.currentTarget.value) !== param.currentValue);
+    setIsValueModified(toNumber(e.currentTarget.value) !== toNumber(param.currentValue));
+
+    // twofold validation: using the json schema (with ajv) and the html5 validation
+    setValidated(validateValuesSchema(e.currentTarget.value, param.schema));
+    e.currentTarget?.reportValidity();
```
absoludity, line 49: Hopefully these can't oppose each other... :P

absoludity (approved):
I was going to say +1 after you add the feature flag, but I see you've done that separately in #5564 . Great, thanks Antonio.

dashboard/src/components/DeploymentForm/DeploymentFormBody/BasicDeploymentForm/TabularSchemaEditorTable/Params/SliderParam.tsx
antgamdia: haha, right. In theory, the html5 validation should be a subset of the schema validation, since we are manually adding stuff like `max`, `min`, etc but we are not considering more complex schemas using `anyOf`, `allOf`, etc. So the browser validation should be able to provide quick feedback in a native way, whereas the ajv validation will prompt the schema validation error, which could be more difficult to understand for non-experienced users.

Issue #5538, opened 2022-10-21
Unable to delete Flux package in "pending" state


gfichtenholt:
I tried to deploy a flux package via UX. 
The app failed to deploy for whatever reason (separate issue). 
I would like to be able to delete this package but the UX won't let me. See attached screenshot

Also, I would like to know why it says "pending". There is nothing "pending" about it. The install is complete and was a failure.

Let me know if step-by-step reproduction instructions are needed I will be happy to provide. 
Thanks

<img width="1440" alt="Screen Shot 2022-10-20 at 7 01 50 PM" src="https://user-images.githubusercontent.com/74032303/197093703-5832f4af-9a4b-48ca-8ba7-7a889410ceb0.png">

gfichtenholt:
Workaround is to use 
```
$ kubectl delete helmrelease/podinfo --now
helmrelease.helm.toolkit.fluxcd.io "podinfo" deleted
```

gfichtenholt:
yes to question 1, we should allow deleting apps in "pending" or "failed" state
question 2: is flux plugin issue, i verified. feel free to pass it to me when (1) is fixed. Thanks

gfichtenholt:
I've merged a fix for issue 2. re-opening to keep track of issue 1

gfichtenholt:
the remaining issue is purely a UX change. Please re-assign accordingly. Thanks

Issue #5523, opened 2022-10-20
Optimize OCI repository lister lookups in flux plugin


gfichtenholt:
Noticed an inefficiency while debugging a scenario with flux plugin. 
Currently, given a single request from the UX to say, `GetAvailablePackageSummaries()`, the code will execute a number of RPC calls for an OCI HelmRepository (Registry). A few have to do with figuring out what kind of OCI provider (github/GCR/harbor, etc) an OCI repository is. The back-end goes through a number of built-in listers and checks whether any of them apply to a given OCI repo. On every call from the UI to the kubeapps-api server that involves this repo, such as `GetAvailablePackageSummaries()` or `GetAvailablePackageDetail()`. This works, but is expensive. It would be good to be able to execute the call once and somehow store the result to avoid further RPC calls on subsequent requests. While the actual list of repos obtained from the lister may change on request-to-request basis, the lister itself is immutable for a given OCI registry and can be cached somehow. I haven't done it yet, because I have yet to find a good way of saving state from one request to the next that doesn't involve the redis cache

gfichtenholt:
my own bookkeeping:

```
I1021 06:27:31.751389       1 docker_reg_v2_repo_lister.go:56] ORAS v2 Registry [oci://ghcr.io/gfichtenholt/stefanprodan-podinfo-clone PlainHTTP=false] PING: OK, elapsed time: [378] ms
I1021 06:27:34.635317       1 docker_reg_v2_repo_lister.go:71] ORAS v2 Registry [oci://ghcr.io/gfichtenholt/stefanprodan-podinfo-clone .Repositories()] elapsed time: [2883] ms
```
i.e. single call to https://github.com/oras-project/oras-go/blob/1969551cc3c77b46f09090f00360e8465871fe8f/registry/remote/registry.go#L111 can take ~3s

PR #5549, opened 2022-10-22
Fixes for Optimize OCI repository lister lookups in flux plugin #5523 and  Unable to delete Flux package in "pending" state #5538 

gfichtenholt:
performance optimizations in flux plugin for OCI HelmRepositories and a fix to mark the release failed instead of pending


cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/cache/watcher_cache.go
```diff
@@ -1004,6 +1006,30 @@ func (c *NamespacedResourceWatcherCache) ForceAndFetch(key string) (interface{},
...
 }
 
+// force a particular key to be processed
+func (c *NamespacedResourceWatcherCache) ForceIfNotProcessingAndFetch(key string) (interface{}, error) {
+	// There is one use case when the client needs to be able to do an Add(), regardless of whether
+	// the item is being processed, e.g. when the corresponding value goes through several quick changes.
+	// Then there is a separate use case when the client doesn't want to do an Add
+	// if the item is currently being processed, such as a .Get() operation that leads to a lengthy
+	// .Add() cuncurrently with another .Get() immediately. Executing two 2 .Add() operations does
+	// not solve any problems just slows the whole thing down.
```
absoludity, line 1016: I'm missing a lot of context, so this could be totally irrelevant, but wouldn't the described issue be avoided automatically if the `.Get()` and `.Add()` operations were in [transactions](https://redis.io/docs/manual/transactions/)? (ie. the second `.Get()` operation should not result in another `.Add()`?)

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/cache/rate_limiting_queue.go
```diff
@@ -317,6 +322,30 @@ func (q *Type) isProcessing(item string) bool {
 	return q.processing.Has(item)
 }
 
+// Atomic check if not already processing then Add.
```
absoludity, line 325: The docstring sounds nice and simple, and I can map the `Atomic` to the lock and unlock (great), the `if not already processing` to a single line, but the `then Add` appears to do a bunch of things that I don't understand without context (eg. why `expected.Delete()` or why an early return if `q.dirty.Has()` etc.). Maybe a few more comments inside that conditional would help people who read this in the future?

absoludity (approved):

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/oci_repo.go
```diff
@@ -385,7 +389,7 @@ func newRegistryClient(isLogin bool, tlsConfig *tls.Config, getterOpts []getter.
 	}, file, nil
 }
 
-// OCI Helm repository, which defines a source, does not produce an Artifact
+// OCI Helm repository, xwhich defines a source, does not produce an Artifact
```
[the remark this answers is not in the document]
gfichtenholt: a typo

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/cache/watcher_cache.go
gfichtenholt: We aren't using Redis transactions. Whether or not we should is a different topic. If we do its going to take some of design/work. Currently, redis is used with a (rate-limiting) [queue](https://github.com/kubernetes/client-go/blob/master/util/workqueue/queue.go) which doesn't grok transactions. The `.Get() `about redis cache and the queue, while `.Add()` is all about the queue. I don't see how redis transactions would be relevant yet. Sorry it wasn't clear from the comment.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/cache/rate_limiting_queue.go
gfichtenholt: two things:
1) the explanation for` expected.Delete() ` or the whole reason behind `expected sets.String` in the queue is to avoid a race condition in unit tests, where an item is added to the queue and then the test code needs to [wait until its been processed](https://github.com/vmware-tanzu/kubeapps/blob/fe00fbb366b78ae21b413b944e44b76a6f7d69e5/cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/cache/rate_limiting_queue.go#L321) by the queue before taking further action. 
2) q.dirty, etc comes from the original [queue](https://github.com/kubernetes/client-go/blob/62756eea6f0e65cf1bda27ac7f96e754f2f90951/util/workqueue/queue.go#L78) which was the inspiration for this code

PR #5565, opened 2022-10-25
Allow deleting pending pkgs

### LINKED ISSUE #5538: Unable to delete Flux package in "pending" state

antgamdia:
### Description of the change

As reported in #5538, sometimes, packages get stuck in a pending state. Currently, Kubeapps UI was preventing the deletion of pending packages, but this might not be a good solution if we really want to delete a package. 
It should be the backend that manages the deletion (or not) of a pending package.

However, the tooltip is still being shown, this way users, at least, will notice something is happening with this package.

### Benefits

No more pending packages stuck

### Possible drawbacks

N/A

### Applicable issues

- fixes #5538

### Additional information

image

image

### DIFF
--- a/dashboard/src/components/AppView/AppControls/DeleteButton/DeleteButton.test.tsx
+++ b/dashboard/src/components/AppView/AppControls/DeleteButton/DeleteButton.test.tsx
@@ -76,7 +76,7 @@ it("renders an error", async () => {
   expect(wrapper.find(Alert)).toIncludeText("Boom!");
 });
 
-it("should render a deactivated button if when passing an in-progress status", async () => {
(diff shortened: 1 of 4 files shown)


castelblanque (approved):
LGTM

Issue #4563, opened 2022-04-06
Add CI coverage for environment where kubeappsCluster is empty.


absoludity:
### Description:

We had a [regression after v2.4.1](https://kubernetes.slack.com/archives/C9D3TSUG4/p1649178273135249) for the environment where Kubeapps is installed on one cluster, but configured so that users operate only on other clusters (ie. they can't install or even select the cluster on which Kubeapps is installed). In this scenario, the `kubeappsCluster` config option to the dashboard is empty (and in the past, the kubeappsapis service would assume that if it's empty, it'd use the `kubeappsCluster` where as it's now correctly required by the apisserver and we missed a place where the dashboard is not sending it). 

This is a useful scenario that we've explicitly designed for in the past, so it'd be great to add a CI test for this environment.

Thanks @aanthonyrizzo for debugging and finding a solution. Anthony plans to push his fix for the dashboard itself separately, this issue is just to ensure that longer term we cover that use-case in CI.

PR #5566, opened 2022-10-25
Add explicit token for Kubeapps cluster reference

### LINKED ISSUE #4564: [Regression] Allow package details to fetch from config.kubeappsCluster

castelblanque:

### Description of the change

Adapted version of the outdated PR #4591. After removing kubeops, there has been a refactor of places changes by the old PR.
It fixes the scenario in which the cluster where Kubeapps is installed is not part of the configured list of clusters.
For example:

```yaml
clusters:
  - name: second-cluster
    domain: ...
    apiServiceURL: ....
    certificateAuthorityData: ...
    serviceToken: eyJhbGciOiJS...
```

For doing so, it uses a token to reference the Kubeapps cluster: `-`.
Also empty string `''` is accepted as cluster reference, for allowing payloads in which there is no cluster specified in context.

Therefore, if a cluster string is empty or `-`, backend logic assumes that the Kubeapps cluster will be used.

### Benefits

User can work in a multicluster setups in which the Kubeapps cluster is not part of the allowed clusters.

### Possible drawbacks

N/A

### Applicable issues

- fixes #4564

### Additional information

Manual tests have been done with Helm plugin (only one working in multicluster so far).
CI E2E test will be done in #4563

### DIFF
--- a/cmd/kubeapps-apis/core/plugins/v1alpha1/plugins.go
+++ b/cmd/kubeapps-apis/core/plugins/v1alpha1/plugins.go
@@ -323,9 +323,8 @@ func createConfigGetterWithParams(inClusterConfig *rest.Config, serveOpts core.S
 		var config *rest.Config
 
 		// Enable existing plugins to pass an empty cluster name to get the
-		// kubeapps cluster for now, until we support (or otherwise decide)
-		// multicluster configuration of all plugins.
-		if cluster == "" {
+		// kubeapps cluster
+		if kube.IsKubeappsClusterRef(cluster) {
 			cluster = clustersConfig.KubeappsClusterName
 		}
 
--- a/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
+++ b/cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/server.go
@@ -23,6 +23,7 @@ import (
 	"github.com/vmware-tanzu/kubeapps/pkg/chart/models"
 	"github.com/vmware-tanzu/kubeapps/pkg/dbutils"
 	httpclient "github.com/vmware-tanzu/kubeapps/pkg/http-client"
+	"github.com/vmware-tanzu/kubeapps/pkg/kube"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/anypb"
@@ -150,7 +151,7 @@ func NewServer(configGetter core.KubernetesConfigGetter, globalPackagingCluster 
 			cluster := pkgContext.GetCluster()
 			// Don't force clients to send a cluster unless we are sure all use-cases
 			// of kubeapps-api are multicluster.
-			if cluster == "" {
+			if kube.IsKubeappsClusterRef(cluster) {
 				cluster = globalPackagingCluster
 			}
 			fn := helm.NewHelmActionConfigGetter(configGetter, cluster)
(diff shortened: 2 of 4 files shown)


castelblanque:
> but Anthony's PR (https://github.com/vmware-tanzu/kubeapps/pull/4613) would still be required, wouldn't it?

Not really. It is enough with the backend handling the Kubeapps cluster name correctly, even when the cluster is not available in the list. I have tested it and are preparing a CI E2E test to avoid regression on this again.

> don't we also need to apply these changes in other packages?

Not needed either in this case:

```bash
if cluster != "" && cluster != s.kubeappsCluster {
```

because:

- If cluster is empty -> OK. It will fall back to the kubeapps cluster later, inside the client getter
- If `cluster` is not empty and equals the configured kubeapps cluster name (or `-` if none)` -> OK
- Otherwise it will fail, which is the desired behavior

mecampbellsoup:
Possible to get this fix backported? As of https://github.com/vmware-tanzu/kubeapps/releases/tag/v2.4.6 it sounds like you need k8s version > 1.21, but we're on 1.20.X and can't easily upgrade for a while.

cc @aanthonyrizzo

absoludity:
> Possible to get this fix backported? As of https://github.com/vmware-tanzu/kubeapps/releases/tag/v2.4.6 it sounds like you need k8s version > 1.21, but we're on 1.20.X and can't easily upgrade for a while.
> 
> cc @aanthonyrizzo

We don't currently do branched releases (where we support previous releases with bug fixes etc.), so your best bet, if you were unable to upgrade your cluster (and have verified that the latest Kubeapps does not work on the older cluster - often we specify that based on the internal k8s client that we import, not because it's not compatible), would be to see if you can backport the fix to your own `kubeapps-apis` image to run. It's a pretty small change.

PR #5567, opened 2022-10-25
Update deps

antgamdia:
### Description of the change

This PR just performs an update of several dependencies prior to the license reporting process we are used to before releasing a new version.

### Benefits

Deps will be up to date

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

I haven't been able to upgrade Helm to the latest version, as it also upgrades k8s deps, which brings some incompatible changes. Not a big deal, just a function signature used in an integration test of the flux plugin. 
However, neither we need to update Helm (it doesn't include our PR yet) nor do we want to update k8s deps to 1.25 right now.


castelblanque (approved):
LGTM, thanks for the update!

PR #5573, opened 2022-10-26
E2E test to cover scenario of Kubeapps cluster not part of clusters list

### LINKED ISSUE #4563: Add CI coverage for environment where kubeappsCluster is empty.

castelblanque:
### Description of the change

This PR adds an E2E test in CI to avoid regression on #5566 and hence to avoid that #4564 happens again.

### Benefits

Kubeapps is usable even if Kubeapps cluster is not among the clusters configured.

### Possible drawbacks

N/A

### Applicable issues

- fixes #4563


antgamdia (approved):
Thanks! Great way to avoid regressions like the one you mentioned
+1 once the merge conflicts are solved.

PR #5518, opened 2022-10-20
Remove sync_writes optino for pinniped-proxy cred cache.

absoludity:
Signed-off-by: Michael Nelson <minelson@vmware.com>

### Description of the change

This PR is the result of the investigation described on #5407 , the end result of which is a single line removal, though I've left some other changes that were helpful during debugging (logging related).

### Benefits

Pinniped-proxy is faster when hit with lots of simultaneous requests.

NOTE: Leaving in draft as although the requests are serviced more quickly in pinniped-proxy, it's still taking 5s for the kubeapps-apis to respond. Investigating why...

### Possible drawbacks

None.

### Applicable issues

- fixes #5407

### Additional information

After

```
2022-10-20T02:22:19.383560 [TRACE] - registering event source with poller: token=Token(1), interests=READABLE | WRITABLE
2022-10-20T02:22:19.383685 [INFO] - Listening on http://0.0.0.0:3333
2022-10-20T02:22:37.624364 [TRACE] - registering event source with poller: token=Token(2), interests=READABLE | WRITABLE
2022-10-20T02:22:37.626434 [TRACE] - registering event source with poller: token=Token(3), interests=READABLE | WRITABLE
2022-10-20T02:22:37.627386 [TRACE] - registering event source with poller: token=Token(4), interests=READABLE | WRITABLE
2022-10-20T02:22:37.627472 [TRACE] - registering event source with poller: token=Token(5), interests=READABLE | WRITABLE
2022-10-20T02:22:37.643862 [INFO] - prep_and_call took 17ms, 0 of which was call_pinniped
2022-10-20T02:22:37.649102 [TRACE] - deregistering event source from poller
2022-10-20T02:22:37.649281 [INFO] - prep_and_call took 22ms, 0 of which was call_pinniped
2022-10-20T02:22:37.654546 [TRACE] - deregistering event source from poller
2022-10-20T02:22:37.654692 [TRACE] - registering event source with poller: token=Token(16777220), interests=READABLE | WRITABLE
2022-10-20T02:22:37.655035 [TRACE] - registering event source with poller: token=Token(16777221), interests=READABLE | WRITABLE
2022-10-20T02:22:37.659207 [INFO] - POST https://kubernetes.default/apis/authorization.k8s.io/v1/selfsubjectaccessreviews 201 Created
2022-10-20T02:22:37.659305 [TRACE] - deregistering event source from poller
[...]

Issue #5407, opened 2022-09-28
Improve response times in Kubeapps APIs when using Pinniped-proxy


castelblanque:
**Summary**
Some requests (e.g. `core.packages.v1alpha1.PackagesService.GetAvailablePackageSummaries`) are slow when using `pinniped-proxy`. Investigate and improve those times.

**Description**
After work done for #5362 and #4919, response times have improved due to not creating unused multiple K8s Go clients.
However, there is still what seems to be a throttling problem when using `pinniped-proxy`. Multiple requests are done from `kubeapps-apis` to K8s API (with correct QPS and Burst values), but requests go through `pinniped-proxy`.
Log from the proxy:

```
2022-09-28T13:48:38.784778 [INFO] - GET https://kubernetes.default/apis/flowcontrol.apiserver.k8s.io/v1beta2?timeout=32s 200 OK
2022-09-28T13:48:38.787947 [INFO] - GET https://kubernetes.default/apis/flowcontrol.apiserver.k8s.io/v1beta1?timeout=32s 200 OK
2022-09-28T13:48:38.788917 [INFO] - GET https://kubernetes.default/apis/storage.k8s.io/v1?timeout=32s 200 OK
2022-09-28T13:48:38.789350 [INFO] - GET https://kubernetes.default/apis/kubeapps.com/v1alpha1?timeout=32s 200 OK
2022-09-28T13:48:38.880150 [INFO] - GET https://kubernetes.default/apis/authorization.k8s.io/v1?timeout=32s 200 OK
2022-09-28T13:48:38.880827 [INFO] - GET https://kubernetes.default/apis/apiregistration.k8s.io/v1?timeout=32s 200 OK
2022-09-28T13:48:38.885347 [INFO] - GET https://kubernetes.default/apis/source.toolkit.fluxcd.io/v1beta2?timeout=32s 200 OK
2022-09-28T13:48:38.885365 [INFO] - GET https://kubernetes.default/apis/identity.concierge.pinniped.dev/v1alpha1?timeout=32s 200 OK
2022-09-28T13:48:38.886346 [INFO] - GET https://kubernetes.default/apis/image.toolkit.fluxcd.io/v1beta1?timeout=32s 200 OK
2022-09-28T13:48:38.887521 [INFO] - GET https://kubernetes.default/apis/kustomize.toolkit.fluxcd.io/v1beta1?timeout=32s 200 OK
[...]

PR #5576, opened 2022-10-27
Switch cache to custom without write-lock for reads.

absoludity:

Follows on from #5518, this time replacing the `cached` package with a custom credential cache.

### Description of the change

After further digging, I found that one cause of the slow handling of 50 concurrent requests going through the pinniped-proxy was that:
 - We were caching a function with an async/await signature which means that even the cached version must have that signature as well - which means a blocking i/o call (which switches the task), and
 - The `Cached` trait specifies that even a `cache_get` operation mutates the cache (in our case, just for statistics of hits/misses), which, as a result, requires acquiring a *write* lock to the cache to read a cached value.

For more details, please see the [discussion with the `Cached` author](https://github.com/jaemk/cached/issues/133).

To avoid both of those issues, this PR:
1. Adds a `cache` module that provides a generic read/write `LockableCache` (for multiple readers, single writer) and builds on that with a `PruningCache` that will prune entries (given a test function) when they should no longer be cached,
2. Uses (1) to create a single `CredentialCache` on startup (in `main.rs`) specifically for caching `TokenCredentialRequest` objects and pruning expired entries, and then passes this through for use in different threads concurrently.
3. Uses the cache to fetch the credentials.

### Benefits

Fetching from the cache is now non-blocking (generally, except when an entry is being added) and so leads to less task switching, improving the total query time by ~2s (down to 3-4). 

There is still something else using significant CPU when creating the client itself (cert-related), which I'm investigating now in a separate PR.

### Applicable issues

- Ref #5407 

### Additional information

Example log when using `RUST_LOG=info,pinniped_proxy::pinniped=debug` which shows the cache being used after the first request. I've not included it in the output generally, but the cache get is now always under a millisecond. As above, the significant delays (some calls to prepare_and_call_pinniped_exchange only 4ms, others 98ms) are what I'll look at next.

```
2022-10-27T01:42:47.820245 [INFO] - Listening on http://0.0.0.0:3333
2022-10-27T01:43:05.077116 [DEBUG] - prepare_and_call_pinniped_exchange took 17ms. Used cache?: false
2022-10-27T01:43:05.085273 [INFO] - GET https://kubernetes.default/api?timeout=32s 200 OK
2022-10-27T01:43:05.091663 [DEBUG] - prepare_and_call_pinniped_exchange took 5ms. Used cache?: true
2022-10-27T01:43:05.100437 [INFO] - GET https://kubernetes.default/apis?timeout=32s 200 OK
2022-10-27T01:43:05.106005 [DEBUG] - prepare_and_call_pinniped_exchange took 4ms. Used cache?: true
2022-10-27T01:43:05.209952 [DEBUG] - prepare_and_call_pinniped_exchange took 21ms. Used cache?: true
2022-10-27T01:43:05.299424 [DEBUG] - prepare_and_call_pinniped_exchange took 5ms. Used cache?: true
2022-10-27T01:43:05.311599 [DEBUG] - prepare_and_call_pinniped_exchange took 5ms. Used cache?: true
2022-10-27T01:43:05.493269 [DEBUG] - prepare_and_call_pinniped_exchange took 98ms. Used cache?: true
2022-10-27T01:43:05.593683 [DEBUG] - prepare_and_call_pinniped_exchange took 4ms. Used cache?: true
2022-10-27T01:43:05.604348 [DEBUG] - prepare_and_call_pinniped_exchange took 4ms. Used cache?: true
2022-10-27T01:43:05.697828 [DEBUG] - prepare_and_call_pinniped_exchange took 87ms. Used cache?: true
2022-10-27T01:43:05.811590 [DEBUG] - prepare_and_call_pinniped_exchange took 20ms. Used cache?: true
2022-10-27T01:43:06.004358 [DEBUG] - prepare_and_call_pinniped_exchange took 94ms. Used cache?: true
2022-10-27T01:43:06.098603 [DEBUG] - prepare_and_call_pinniped_exchange took 5ms. Used cache?: true
2022-10-27T01:43:06.108756 [DEBUG] - prepare_and_call_pinniped_exchange took 4ms. Used cache?: true

```

### DIFF
--- a/cmd/pinniped-proxy/src/cache.rs
+++ b/cmd/pinniped-proxy/src/cache.rs
@@ -1,0 +1,134 @@
+// Copyright 2020-2022 the Kubeapps contributors.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::{collections::HashMap, hash::Hash, sync::RwLock};
+
+/// An unlimited cache that can be RwLocked across threads.
+///
+/// Importantly, checking the cache does not require a write-lock
+/// (unlike the [`Cached` trait's `cache_get`](https://github.com/jaemk/cached/blob/f5911dc3fbc03e1db9f87192eb854fac2ee6ac98/src/lib.rs#L203))
+#[derive(Default)]
+struct LockableCache<K, V>(RwLock<HashMap<K, V>>);
+
+impl<K, V> LockableCache<K, V> {
+    fn get(&self, k: &K) -> Option<V>
+    where
+        K: Eq + Hash,
+        V: Eq + Hash + Clone,
+    {
+        (*self.0.read().unwrap())
+            .get(k)
+            .and_then(|v| Some((*v).clone()))
+    }
+
+    // insert is called in PrunableCache via a write lock guard, but compiler
+    // doesn't see this, apparently.
+    #[allow(dead_code)]
+    fn insert(&self, k: K, v: V)
+    where
+        K: Eq + Hash,
+        V: Eq + Hash,
+    {
+        self.0.write().unwrap().insert(k, v);
+    }
+
+    #[cfg(test)]
+    fn len(&self) -> usize {
+        (*self.0.read().unwrap()).len()
+    }
+}
+
+/// A cache that additionally prunes itself whenever it holds the write-lock.
+pub struct PruningCache<K, V> {
+    cache: LockableCache<K, V>,
+    prune_fn: fn(&(K, V)) -> bool,
+}
+
+impl<K, V> PruningCache<K, V> {
+    pub fn new(f: fn(&(K, V)) -> bool) -> PruningCache<K, V>
+    where
+        K: Default,
+        V: Default,
+    {
+        PruningCache {
+            cache: LockableCache::default(),
+            prune_fn: f,
+        }
+    }
+
+    pub fn get(&self, k: &K) -> Option<V>
+    where
+        K: Eq + Hash + Clone,
+        V: Eq + Hash + Clone,
+    {
+        // Only return the value from the cache if it should not have been
+        // pruned.
+        self.cache
+            .get(k)
+            .and_then(|v| (self.prune_fn)(&(k.clone(), v.clone())).then_some(v))
+    }
+
+    // Prunes the cache while holding the write-lock during an insert.
+    pub fn insert(&self, k: K, v: V)
+    where
+        K: Eq + Hash + Clone,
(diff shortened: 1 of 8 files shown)


cmd/pinniped-proxy/src/cache.rs
```diff
@@ -0,0 +1,134 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use std::{collections::HashMap, hash::Hash, sync::RwLock};
+
+/// An unlimited cache that can be RwLocked across threads.
+///
+/// Importantly, checking the cache does not require a write-lock
+/// (unlike the [`Cached` trait's `cache_get`](https://github.com/jaemk/cached/blob/f5911dc3fbc03e1db9f87192eb854fac2ee6ac98/src/lib.rs#L203))
+#[derive(Default)]
+struct LockableCache<K, V>(RwLock<HashMap<K, V>>);
```
absoludity, line 11: This struct is a [NewType pattern](https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#using-the-newtype-pattern-to-implement-external-traits-on-external-types) which is a zero-cost abstraction (ie. no penalty at run-time) that defines a new type as a thin wrapper - a 1-tuple of another type, so that we can add our caching functions (`get`, `insert`) to this type without needing an extra struct. So in this case, a LockableCache is really just a read-write lock wrapping a hash, but one which behaves like a cache.

cmd/pinniped-proxy/src/main.rs
```diff
@@ -26,15 +27,17 @@ mod tls_config;
...
+        .format(|buf, record| {
+            writeln!(
+                buf,
+                "{} [{}] - {}",
+                Local::now().format("%Y-%m-%dT%H:%M:%S%.6f"),
+                record.level(),
+                record.args()
+            )
+        })
+        .init();
```
absoludity, line 40: Woops - I'd left the `.init()` off when I re-enabled logging in my last PR, but didn't check it (so nothing will be logged without this change).

antgamdia (approved):
Thanks for digging into the root cause; it's a pity we had to drop the `cached` crait and implement it on our own. I didn't expect a read operation to be blocking :S

absoludity:
> Thanks for digging into the root cause; it's a pity we had to drop the `cached` crait and implement it on our own. I didn't expect a read operation to be blocking :S

Yeah - either did I, though I did afterwards remember that the `cached` trait keeps statistics of misses etc.

Issue #5397, opened 2022-09-27
Accessibility compliance for kubeapps.dev


ppbaena:
**Summary**
The goal is to ensure [kubeapps.dev](https://kubeapps.dev/) website is accessible

**Background and rationale**
All VMware’s OSS products must meet Web Content Accessibility Guidelines (WCAG) 2.1 Level AA standards.
Devices, Assistive Technology combinations used: Chrome + NVDA on Windows, Safari + Voiceover on Mac

**Description**
Review the results of the accessibility audit for [kubeapps.dev](https://kubeapps.dev/) and fix all [issues reported](https://jira.eng.vmware.com/issues/?jql=labels%20%3D%20A11y-kubeapps-Sept22) (JIRA - internal tool)

**Acceptance criteria**

- [x] All issues identified in the accessibility audit are fixed
- [ ] For every major UI/ Feature update fixed, a request for Accessibility Testing has been performed and succeeded.

**Additional context**
An Accessibility Testing has been completed for [kubeapps.dev](https://kubeapps.dev/) with several issues identified to comply with Web Content Accessibility Guidelines (WCAG) 2.1 Level AA standards (issues: 7 critical, 20 High, 2 Medium, 5 Low)

antgamdia:
#### Current status: 
- 4 out of 34 a11y issues are pending (awaiting an official response)
- Still pending: submit this evidence in JIRA and close the tickets.

#### Summary:
- [X] [14258](https://jira.eng.vmware.com/browse/VPAT-14258): 🟢 
- [X] [14230](https://jira.eng.vmware.com/browse/VPAT-14230): 🔴  BLOCKED (awaiting a11y team's response)
- [X] [14229](https://jira.eng.vmware.com/browse/VPAT-14229): 🟢
- [X] [14228](https://jira.eng.vmware.com/browse/VPAT-14228): 🟢
- [X] [14227](https://jira.eng.vmware.com/browse/VPAT-14227): 🟢
- [X] [14226](https://jira.eng.vmware.com/browse/VPAT-14226): 🟢
- [X] [14225](https://jira.eng.vmware.com/browse/VPAT-14225): 🟢 
- [ ] [14224](https://jira.eng.vmware.com/browse/VPAT-14224): 🔴 BLOCKED (awaiting a11y team's response)
- [X] [14223](https://jira.eng.vmware.com/browse/VPAT-14223): 🟢 
[...]

PR #5578, opened 2022-10-27
Fix website meta tags issues

antgamdia:
### Description of the change

While working on #5397, I discovered some minor issues with the meta tags we were using. Mainly, the title and description of each page were always the same, hindering the crawler process we run via Algolia.

### Benefits

Title and other meta tags will have more meaningful content.

### Possible drawbacks

N/A

### Applicable issues

- related #5397

### Additional information

I'm removing some unused files as well.

image


ppbaena (approved):
LGTM

PR #5582, opened 2022-10-28
Migrate GKE jobs to GitHub Actions

beni0888:
### Description of the change

This PR contains mainly the migration of the GKE jobs from CircleCI to GitHub Actions. During the development of this PR we noticed that we have useless jobs (those testing the latest version, provided with the variable `TEST_LATEST_RELEASE=1`) that were being always skipped by the CI because they are checking some conditions based on the result of a non-existent bash function `changedVersion`, so we have remove them from the new GHA workflow.

### Benefits

Those jobs testing Kubeapps in a GKE cluster are run in the GHA workflow when needed.

### Possible drawbacks

None

### Applicable issues

- related to #4436


beni0888:
GKE jobs aren't going to be run for this PR, as they run on tag and prerelease (still not migrated) flows. So I add here the link to a previous commit where some temporary hacks were in place to make then run and test its behavior: https://github.com/vmware-tanzu/kubeapps/actions/runs/3342960475

.circleci/config.yml
```diff
@@ -539,7 +539,32 @@ run_e2e_tests: &run_e2e_tests
...
+      info "USE_MULTICLUSTER_OIDC_ENV: $USE_MULTICLUSTER_OIDC_ENV"
+      info "OLM_VERSION: $OLM_VERSION"
+      info "DEV_TAG: $DEV_TAG"
+      info "IMG_MODIFIER: $IMG_MODIFIER"
+      info "TEST_TIMEOUT_MINUTES: $TEST_TIMEOUT_MINUTES"
+      info "DEFAULT_DEX_IP: $DEFAULT_DEX_IP"
+      info "ADDITIONAL_CLUSTER_IP: $ADDITIONAL_CLUSTER_IP"
+      info "KAPP_CONTROLLER_VERSION: $KAPP_CONTROLLER_VERSION"
+      info "CHARTMUSEUM_VERSION: $CHARTMUSEUM_VERSION"
+      info "###############################################################################################"
```
antgamdia, line 554: This debug info is gonna help identify some future problems and mismatches. Thanks!

.circleci/config.yml
```diff
@@ -592,7 +617,17 @@ gke_test: &gke_test
...
         name: Export escaped GKE cluster name
         command: |
-          echo "export ESCAPED_GKE_CLUSTER=$(echo ${GKE_CLUSTER}-${CIRCLE_BRANCH:-$CIRCLE_TAG}-${TEST_LATEST_RELEASE}-${GKE_BRANCH}-ci | sed 's/[^a-z0-9-]//g')" >> $BASH_ENV
+          ESCAPED_GKE_CLUSTER=$(echo ${GKE_CLUSTER}-${CIRCLE_BRANCH:-$CIRCLE_TAG}-${TEST_LATEST_RELEASE}-${GKE_BRANCH}-ci | sed 's/[^a-z0-9-]//g')
+          
+          # Trim the cluster name in case it exceeds the max length allowed by GKE (currently 40 chars)
+          MAX_LENGTH=40 
+          len=${#ESCAPED_GKE_CLUSTER}
+          if ((len > MAX_LENGTH)); then
+            ESCAPED_GKE_CLUSTER="${ESCAPED_GKE_CLUSTER:0:MAX_LENGTH}"
```
antgamdia, line 626: Given that we might be trimming the name, perhaps we should move the `GKE_BRANCH` close to the beginning of the name?  I mean, to avoid name collisions if the name is trimmed out, we may want to have the disambiguating part first.

.github/workflows/gke_e2e_tests.yaml
```diff
@@ -0,0 +1,126 @@
...
+      #TODO: right now it contains just the version number (eg. 1.22), rename to GKE_VERSION once CircleCI is decommissioned
+      GKE_BRANCH:
+        type: string
+        description: "The branch of the GKE environment in which we want to test."
+      GKE_CLUSTER:
+        type: string
+      GKE_PROJECT:
+        type: string
+      GKE_RELEASE_CHANNEL:
+        type: string
```
antgamdia, line 27: I'm not getting this part, I think. `GKE_BRANCH` would be either `stable` or `regular`, wouldn't it? If so, what's the difference with `GKE_RELEASE_CHANNEL` ?
Why can't we just use the `GKE_VERSION`  now? I mean, once the name collision is fixed, why can't we have two 1.22 clusters?  (besides, in https://github.com/vmware-tanzu/kubeapps/pull/5567 I bumped it up to 1.23)

.github/workflows/gke_e2e_tests.yaml
```diff
@@ -0,0 +1,126 @@
...
+        default: false
+    secrets:
+      GKE_ADMIN:
+        required: true
+      GCLOUD_KEY:
+        required: true
+
+jobs:
+  e2e_tests:
+    runs-on: ubuntu-latest
```
antgamdia, line 66: Shouldn't we be sticking to a certain version instead?

.github/workflows/kubeapps.yaml
```diff
@@ -487,9 +500,8 @@ jobs:
         with:
           name: e2e_tests_${{ matrix.tests_group }}_reports
           path: integration/reports
-      - name: Setup upterm session
-        uses: lhotari/action-upterm@v1
-        if: failure() && env.DEBUG_MODE == 'true'
+      - uses: lhotari/action-upterm@v1
+        if: false
```
antgamdia, line 508: Leftover maybe?

script/chart-museum.sh
```diff
@@ -123,28 +125,30 @@ EOF
...
+      pushChartToChartMuseum $2 $3 $4
+      ;;
 
-  *)
-    ;;
-esac
+    *)
+      ;;
+  esac
+fi
```
antgamdia, line 154: Missing newline at the end

script/start-gke-env.sh
```diff
@@ -27,12 +30,15 @@ if [[ $(gcloud container clusters list --filter="name:${CLUSTER}") ]]; then
...
         echo "GKE cluster already exits. Deleting it"
-        gcloud container clusters delete "${CLUSTER}" --zone "${ZONE}"
+        gcloud container clusters delete "${CLUSTER}" --zone "${ZONE}" --quiet
     fi
 fi
 
 echo "Creating cluster ${CLUSTER} in ${ZONE} (v$BRANCH)"
-gcloud container clusters create --cluster-version="${BRANCH}" --zone "${ZONE}" "${CLUSTER}" --num-nodes 2 --machine-type=n1-standard-2 --preemptible --labels=team=kubeapps
+
+# FIXME Removing the use of --labels=team=kubeapps due to a bug in gcloud cli: https://issuetracker.google.com/issues/255708239
```
antgamdia, line 39: I'd try to stick to the `#TODO(author): foo` format for the TODOs, this way it's easier to identify all the TODOs and who created them.

antgamdia (approved):
Thanks for the effort in migrating and improving everything. I'm sure this is gonna be really helpful for us!

.circleci/config.yml
beni0888: yeah, you're right, although for the CircleCI flow I haven't added the `gke_release_channel` param, so in case `stable` and `regular` versions are the same, as they are right now, they will collide anyway...

.github/workflows/gke_e2e_tests.yaml
beni0888: `GKE_BRANCH` is how it's been called through the code so far, it takes the value from `GKE_STABLE_VERSION` and `GKE_REGULAR_VERSION` (the numeric values, eg `1.22`). Taking into account that those values could be the same (as they've been lately), we're adding the new variable `GKE_RELEASE_CHANNEL` that can take the values `STABLE` and `REGULAR`, to disambiguate them.

.github/workflows/kubeapps.yaml
beni0888: yeah, leftover a reminder for when I need it 😅 But yeah, it would be better removed

PR #5579, opened 2022-10-27
Fix reopened website a11y issues

antgamdia:
### Description of the change

This PR aims at fixing the remaining a11y issues that were reopened. Details at https://github.com/vmware-tanzu/kubeapps/issues/5397#issuecomment-1293284797

Summary of the changes: 

1. Change the border color of the searchbox
2. Display a menu behind "documentation" if the screen is small. If you don't like it, we can always remove it and try to convince  a11y folks.
3. Fix the overflow of the left menu
4. Add a "space" and "enter" keyboard events handler for the OneTrust cookie manager. 
5. Remove the focus on code blocks (every code block, indeed)

### Benefits

a11y compliance

### Possible drawbacks

N/A

### Applicable issues

- fixes #5397

### Additional information

This PR is based on top of https://github.com/vmware-tanzu/kubeapps/pull/5578, as I assume it will get merged soon.

Also, I've removed some unused files (images mainly) and unified the icons we are using (ie, using the SVG ones, and resizing them, via CSS, instead of having several logos for the same purpose)

~EDIT: there are more issues to solve I've just seen, will add the fixes to this PR~

PR #5587, opened 2022-10-28
Fix website search issue

antgamdia:
### Description of the change

This PR is a hotfix for a bug introduced at #5579: the searchbox init code was moved to a `static` folder, which is unable to retrieve the values from the site params. Therefore, the API key is  literally `.Site.Params.docs_search_api_key` instead of the replaced value.

Also, squeezing a minor fix to prevent `Kubeapps Kubeapps ...`  to appear in the descriptions.

### Benefits

 Search will work again

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

N/A


antgamdia:
Rubber stamping as the fix works in the preview:

image

PR #5599, opened 2022-11-02
Add custom per-repo interval in AppRepo

antgamdia:
TBD: just to trigger CI rignt now.

### Applicable issues

- fixes #

### DIFF
--- a/cmd/apprepository-controller/server/controller.go
+++ b/cmd/apprepository-controller/server/controller.go
@@ -420,7 +420,7 @@ func ownerReferencesForAppRepo(apprepo *apprepov1alpha1.AppRepository, childName
 
 // intervalToCron transforms string durations like "1m" or "1h" to cron expressions
 // Even if valid time units are "ns", "us", "ms", "s", "m", "h",
-// the result will get rounded up to seconds.
+// the result will get rounded up to minutes.
 func intervalToCron(duration string) (string, error) {
 	if duration == "" {
 		return "", fmt.Errorf("duration cannot be empty")
@@ -428,8 +428,9 @@ func intervalToCron(duration string) (string, error) {
 		if d, err := time.ParseDuration(duration); err != nil {
 			return "", err
 		} else {
-			cronSecs := math.Ceil(d.Seconds())                  // round up to nearest second
-			return fmt.Sprintf("*/%v * * * * *", cronSecs), nil // every cronSecs seconds
+			cronMins := math.Ceil(d.Minutes()) // round up to nearest minutes
+			// https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#cron-schedule-syntax
+			return fmt.Sprintf("*/%v * * * *", cronMins), nil // every cronMins minutes
 		}
 	}
 }
--- a/cmd/apprepository-controller/server/controller.go
+++ b/cmd/apprepository-controller/server/controller.go
@@ -433,7 +433,7 @@ func newCronJob(apprepo *apprepov1alpha1.AppRepository, config Config) *batchv1.
 			// Set to replace as short-circuit in k8s <1.12
 			// TODO re-evaluate ConcurrentPolicy when 1.12+ is mainstream (i.e 1.14)
 			// https://github.com/kubernetes/kubernetes/issues/54870
-			ConcurrencyPolicy: "Replace",
+			ConcurrencyPolicy: batchv1.ReplaceConcurrent,
 			JobTemplate: batchv1.JobTemplateSpec{
 				Spec: syncJobSpec(apprepo, config),
 			},
--- a/cmd/apprepository-controller/server/controller.go
+++ b/cmd/apprepository-controller/server/controller.go
@@ -421,15 +421,15 @@ func ownerReferencesForAppRepo(apprepo *apprepov1alpha1.AppRepository, childName
 // intervalToCron transforms string durations like "1m" or "1h" to cron expressions
 // Even if valid time units are "ns", "us", "ms", "s", "m", "h",
 // the result will get rounded up to seconds.
-func intervalToCron(duration string) string {
+func intervalToCron(duration string) (string, error) {
 	if duration == "" {
-		return ""
+		return "", fmt.Errorf("duration cannot be empty")
 	} else {
 		if d, err := time.ParseDuration(duration); err != nil {
-			return ""
+			return "", err
 		} else {
-			cronSecs := math.Ceil(d.Seconds())             // round up to nearest second
-			return fmt.Sprintf("*/%v * * * * *", cronSecs) // every cronSecs seconds
+			cronSecs := math.Ceil(d.Seconds())                  // round up to nearest second
+			return fmt.Sprintf("*/%v * * * * *", cronSecs), nil // every cronSecs seconds
 		}
 	}
 }
@@ -438,10 +438,18 @@ func intervalToCron(duration string) string {
 // the appropriate OwnerReferences on the resource so handleObject can discover
 // the AppRepository resource that 'owns' it.
 func newCronJob(apprepo *apprepov1alpha1.AppRepository, config Config) *batchv1.CronJob {
-	// If the apprepo has its own Interval, use that instead of the default global crontab.
+	var err error
+
 	cronTime := config.Crontab
+
+	// If the apprepo has its own interval,
+	// use that instead of the default global crontab.
 	if apprepo.Spec.Interval != "" {
-		cronTime = intervalToCron(apprepo.Spec.Interval)
+		cronTime, err = intervalToCron(apprepo.Spec.Interval)
+	}
+	// If the interval is invalid, use the default global crontab
+	if err != nil {
+		cronTime = config.Crontab
 	}
 
 	return &batchv1.CronJob{
--- a/cmd/apprepository-controller/server/controller.go
+++ b/cmd/apprepository-controller/server/controller.go
@@ -272,7 +272,7 @@ func (c *Controller) syncHandler(key string) error {
 		// processing.
 		if errors.IsNotFound(err) {
 			log.Infof("AppRepository '%s' no longer exists so performing cleanup of charts from the DB", key)
-			// Trigger a Job to perfrom the cleanup of the charts in the DB corresponding to deleted AppRepository
+			// Trigger a Job to perform the cleanup of the charts in the DB corresponding to deleted AppRepository
 			_, err = c.kubeclientset.BatchV1().Jobs(c.conf.KubeappsNamespace).Create(context.TODO(), newCleanupJob(c.conf.KubeappsNamespace, namespace, name, c.conf), metav1.CreateOptions{})
 			if err != nil {
 				log.Errorf("Unable to create cleanup job: %v", err)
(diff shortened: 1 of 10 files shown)


cmd/apprepository-controller/server/controller.go
```diff
@@ -417,10 +418,41 @@ func ownerReferencesForAppRepo(apprepo *apprepov1alpha1.AppRepository, childName
...
+
+	cronTime := config.Crontab
+
+	// If the apprepo has its own interval,
+	// use that instead of the default global crontab.
+	if apprepo.Spec.Interval != "" {
+		cronTime, err = intervalToCron(apprepo.Spec.Interval)
+	}
+	// If the interval is invalid, use the default global crontab
+	if err != nil {
```
absoludity, line 452: Hmm - I know we don't yet have any way to report this back to the user (other than failing the operation, which isn't great either), so I reckon it'd be useful to at least log this error so that we can help debug if there's an issue?

cmd/apprepository-controller/server/controller.go
```diff
@@ -429,11 +461,11 @@ func newCronJob(apprepo *apprepov1alpha1.AppRepository, config Config) *batchv1.
 			Annotations:     config.ParsedCustomAnnotations,
 		},
 		Spec: batchv1.CronJobSpec{
-			Schedule: config.Crontab,
+			Schedule: cronTime,
 			// Set to replace as short-circuit in k8s <1.12
 			// TODO re-evaluate ConcurrentPolicy when 1.12+ is mainstream (i.e 1.14)
 			// https://github.com/kubernetes/kubernetes/issues/54870
-			ConcurrencyPolicy: "Replace",
+			ConcurrencyPolicy: batchv1.ReplaceConcurrent,
```
absoludity, line 468: Dunno, but does this change to the concurrency policy mean we can remove the above comment? EDIT: hah, or it's not a change at all, just switch to constant :)

cmd/apprepository-controller/server/controller_test.go
```diff
@@ -1883,6 +1975,59 @@ func TestGenerateJobName(t *testing.T) {
...
+		},
+		{
+			name:         "good interval, every two minutes",
+			interval:     "2m",
+			expectedCron: "*/2 * * * *",
+		},
+		{
+			name:         "good interval, every two hours",
+			interval:     "2h",
+			expectedCron: "*/120 * * * *",
```
absoludity, line 2012: Is that actually valid? I thought Cron was stateless (and so effectively reset at the turn of each hour for a minutes range here). I'm not 100%, but worth checking (see some [notes on the wikipedia article](https://en.wikipedia.org/wiki/Cron#Non-standard_characters) - not a source of truth obviously).

Could be worth using https://pkg.go.dev/github.com/robfig/cron#Every , not sure.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/globals_vars_test.go
```diff
@@ -485,6 +502,15 @@ var (
 		}),
 	}
 
+	addRepoReqWithInterval = &corev1.AddPackageRepositoryRequest{
+		Name:            "bar",
+		Context:         &corev1.Context{Namespace: "foo", Cluster: KubeappsCluster},
+		Type:            "helm",
+		Url:             "http://example.com",
+		NamespaceScoped: true,
+		Interval:        "99m",
```
absoludity, line 511: Again, not sure that's valid. We could even simplify things by just using a drop-down selection in the UX (and validating valid values of 1, 2, ..., 15, 20, 30 etc.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/repositories_test.go
```diff
@@ -1246,12 +1241,22 @@ func TestUpdatePackageRepository(t *testing.T) {
 			expectedStatusCode: codes.InvalidArgument,
 		},
 		{
-			name: "check that interval is not used",
+			name: "check that interval is used",
 			requestCustomizer: func(request *corev1.UpdatePackageRepositoryRequest) *corev1.UpdatePackageRepositoryRequest {
-				request.Interval = "1s"
+				request.Interval = "99m"
```
absoludity, line 1246: Again, maybe use 15m as a valid interval.

absoludity (approved):
Thanks Antonio, nice clean change :)

> One caveat is that the upgrade process: since we are creating a CronJob and passing the crontab value on creation via the apprepository controller, from the Repos API we don't have a clear way to _edit_ the cronjob once generated.
> 
>     * In this PR we've addressed it by just disabling the field when upgrading

I think this should be explained in a bit more detail in the code, or in an issue, that can be referenced in the code. I'd currently have to dig a bit to understand why we can't delete and re-create the `CronJob` when updating the `AppRepository` CRD.

beni0888 (approved):

cmd/apprepository-controller/server/controller.go
antgamdia: Exactly, I just replaced the `Replace` string with the constant coming from the k8s package. Below the different options:

````go

const (
	// AllowConcurrent allows CronJobs to run concurrently.
	AllowConcurrent ConcurrencyPolicy = "Allow"

	// ForbidConcurrent forbids concurrent runs, skipping next run if previous
	// hasn't finished yet.
	ForbidConcurrent ConcurrencyPolicy = "Forbid"

	// ReplaceConcurrent cancels currently running job and replaces it with a new one.
	ReplaceConcurrent ConcurrencyPolicy = "Replace"
)
````

File: cmd/apprepository-controller/server/controller_test.go (hunk 3 above)
Comment (antgamdia) [reply cmd/apprepository-controller/server/controller_test.go:2012]: Thanks for your comments and pointers, I've had to revisit my knowledge about how `cron` works internally. 

The expression I was building is syntactically valid, that's why I thought it was valid (=meaningful). However, it is semantically pointless given how the cron works.
Kubernetes is using the `vixie-cron` implementation, the POSIX-compliant implementation developed for BSD (back in 1988). 

NOTE: There are some other implementations, like `dcron`, `fcron`, `bcron` and `cronie`. The latter, is a `vixie-cron` fork maintained by Fedora where, for instance, [does validate the `step` field](https://github.com/cronie-crond/cronie/blob/master/readme.md) and therefore makes the discussed expression invalid syntactically.

Well, so coming back to the vixie implementation, the step can be any number:

https://github.com/vixie/cron/blob/2e298fc87bf6f9220d09e60162ebf909a607c4fb/entry.c#L518-L521

But, it is going to, silently, drop the invalid values:

https://github.com/vixie/cron/blob/2e298fc87bf6f9220d09e60162ebf909a607c4fb/entry.c#L580-L584

In short, as you pointed out, in an interval `0-59`, the only possible values for step are `[1-59]`.

Unfortunately, this is something that k8s is NOT validating:

````
k get cronjob        
NAME                  SCHEDULE       SUSPEND   ACTIVE   LAST SCHEDULE   AGE
kubernetes-cron-job   */61 * * * *   False     0        <none>          20m
````

> Note the "last schedule" equals to "none"

The library you pointed out seems to be a scheduler itself, as it does not generate the crontab thing, but actually runs a timer to calculate the next executions.

So, it seems we might need to make the function a bit smarter :S and limit the values + perform the proper conversions between time units.

Comment (absoludity) [reply cmd/apprepository-controller/server/controller_test.go:2012]: > The library you pointed out seems to be a scheduler itself, as it does not generate the crontab thing, but actually runs a timer to calculate the next executions.

Yes, sorry - I realised the library was for a scheduler, but had thought that function I linked to was for converting a duration into a crontab format, but seems it is not.

**Comment by antgamdia** [comment]
> I think this should be explained in a bit more detail in the code, or in an issue, that can be referenced in the code. I'd currently have to dig a bit to understand why we can't delete and re-create the `CronJob` when updating the `AppRepository` CRD.

After re-reading it, you're right, there's something wrong. The CR controller should be smart enough to detect a change in the definition to recreate the cronjon... the problem is in how we consider two AppRepos are equals. Below the excerpt of the code:

````go
if oldApp.Spec.URL != newApp.Spec.URL || oldApp.Spec.ResyncRequests != newApp.Spec.ResyncRequests {
  controller.enqueueAppRepo(newApp)
}
```` 

That's why the "update" branch of the controller wasn't being triggered. Fixing this in the next commit. Thansk for the heads-up!

File: cmd/apprepository-controller/server/controller.go [hunk 6]
```diff
@@ -136,7 +139,7 @@ func NewController(
 		UpdateFunc: func(oldObj, newObj interface{}) {
 			oldApp := oldObj.(*apprepov1alpha1.AppRepository)
 			newApp := newObj.(*apprepov1alpha1.AppRepository)
-			if oldApp.Spec.URL != newApp.Spec.URL || oldApp.Spec.ResyncRequests != newApp.Spec.ResyncRequests {
+			if !reflect.DeepEqual(oldApp.Spec, newApp.Spec) {
```
antgamdia, line 142: @absoludity, do you know if there's any reason for checking the equality manually instead of just using `reflect.DeepEqual` as I'm proposing herein?  I guess that a change in the AppRepo spec should trigger an update in the controller, doesn't it?

cmd/apprepository-controller/server/controller.go
```diff
@@ -417,10 +420,77 @@ func ownerReferencesForAppRepo(apprepo *apprepov1alpha1.AppRepository, childName
...
+
+	cronDays := math.Ceil(cronHours / 24) // get the days
+	if cronDays < 32 {
+		return fmt.Sprintf("0 0 */%v * *", cronDays), nil // every cronHoursDays days
+	}
+
+	cronMonths := math.Ceil(cronDays / 31) // get the months
+	if cronMonths < 13 {
+		return fmt.Sprintf("0 0 1 */%v *", cronMonths), nil // every cronHoursMonths months
+	}
```
antgamdia, line 457: New logic for roughly estimating the cron line from a duration

cmd/apprepository-controller/server/controller.go
```diff
@@ -417,10 +420,77 @@ func ownerReferencesForAppRepo(apprepo *apprepov1alpha1.AppRepository, childName
...
+	// If the apprepo has its own interval,
+	// use that instead of the default global crontab.
+	if apprepo.Spec.Interval != "" {
+		// if the passed interval is indeed a cron expression, use it straight
+		if gron.IsValid(apprepo.Spec.Interval) {
+			cronTime = apprepo.Spec.Interval
+		} else {
+			// otherwise, convert it
+			cronTime, err = intervalToCron(apprepo.Spec.Interval)
+		}
```
antgamdia, line 486: Since people might want to specify their own cron lines, this Helm plugin is accepting a valid cron expression and, if invalid, it will try to pasre it as a go duration. This brings more flexibility, but it differs from the API spec a little bit. WDYT?

cmd/apprepository-controller/server/controller.go
absoludity: Nope. I can't see anything on that [AppRepo Spec struct ](https://pkg.go.dev/github.com/vmware-tanzu/kubeapps/cmd/apprepository-controller/pkg/apis/apprepository/v1alpha1#AppRepositorySpec) that, if changed, should not trigger an update. Agree 100%.

cmd/apprepository-controller/server/controller.go
absoludity: Nice - thanks for the update. Just note, small typo in the inline comments on lines 451 and 456 (s/Hours//)

cmd/apprepository-controller/server/controller.go
absoludity: Hmmm, handy I think. I'd be ok with that (and updating the API spec to match, if we want - still backwards compatible so a nice addition, imo). Nice one!

absoludity (approved):

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/repositories.proto
```diff
@@ -85,6 +85,9 @@ message AddPackageRepositoryRequest {
   string url = 6;
 
   // The interval at which to check the upstream for updates (in time+unit)
+  // Some plugins may, addionally, support other formats, for instance,
+  // a cron expression.
+  // e.g., "*/10 * * * *" will be equivalent to "10m"
   // Optional. Defaults to 10m if not specified
   string interval = 7;
```
antgamdia, line 92: I've added this note to document the fact that other plugins might support other formats while also supporting the "time+unit" one.

PR #5604, opened 2022-11-03
Package repository permissions endpoint

### LINKED ISSUE #5521: New package repository API endpoint to provide repositories permissions
**Summary**
New package repository API endpoint to provide a summary of user's permission on repositories CRDs

**Background and rationale**
Currently UI is making multiple 'canI' calls to retrieve which permissions the user has related to repositories.
Everything is grouped in the TS function [`getRepositoriesPermissions`](https://github.com/vmware-tanzu/kubeapps/blob/dbb26e26bf3a4e545df7b4fbb3034250d98e4c84/dashboard/src/shared/PackageRepositoriesService.ts#L296).
The resources involved are different per plugin: Helm is about `kubeapps.com` and `apprepositories`, Carvel is about `packaging.carvel.dev` and `packagerepositories`, etc.

The current setup is inefficient, as multiple `canI` calls are performed at the same time. And also UI has to know the resource names, groups, verbs...from the underlying K8s.

**Description**
Add a new API endpoint in package repository API that will provide permissions for the current user:

- Per plugin
  - For the specified namespace
  - For the global namespace
- For verbs
  - list
  - update
  - create
  - delete
  - get

Input for the request:
- Namespace
- Auth token in header

Response of the request:
Similar to the [structure in TS](https://github.com/vmware-tanzu/kubeapps/blob/dbb26e26bf3a4e545df7b4fbb3034250d98e4c84/dashboard/src/shared/types.ts#L481).
```typescript
interface IResourcePermission {
  list: boolean;
  create: boolean;
  update: boolean;
  delete: boolean;
  get: boolean;
  namespace: string;
}

interface IPackageRepositoryPermission {
  global: IResourcePermission;
  namespaced: IResourcePermission;
  plugin: Plugin;
}

```

**Acceptance criteria**
- Implemented new API endpoint in core and plugins for repositories to return permissions of the user to operate with package repositories resources.

**Additional context**
See comment in the [originating PR #5480.](https://github.com/vmware-tanzu/kubeapps/pull/5480#discussion_r1000353135)

castelblanque:
### Description of the change

Adds a new procedure `GetPackageRepositoryPermissions` in the repositories API that allows to get the permissions for the current user with regards to package repositories.

Returned structure of data is (per plugin):
- Global permissions (no namespace). Verbs allowed (`get`, `list`, etc.)
- Same as above but specific for a namespace, if specified

### Benefits

This endpoint will allow to:
- Perform in a single call what is being done now with multiple `canI` requests from frontend.
- Abstract away the frontend from knowing about the CRDs used by plugins. See [here](https://github.com/vmware-tanzu/kubeapps/blob/main/dashboard/src/shared/PackageRepositoriesService.ts#L302-L343).

### Possible drawbacks

N/A

### Applicable issues

- fixes #5521


cmd/kubeapps-apis/plugins/pkg/resources/resources_access.go
```diff
@@ -0,0 +1,78 @@
...
+	type accessReviewResult struct {
+		verb    string
+		allowed bool
+	}
+	accessReviewChan := make(chan accessReviewResult, len(AccessVerbs))
+
+	// Each access review is requested in a go routine
+	for _, v := range AccessVerbs {
+		wg.Add(1)
+		go func(verb string) {
```
castelblanque, line 38: Each `SelfSubjectAccessReview` is created on a go routine.

cmd/kubeapps-apis/docs/kubeapps-apis.swagger.json
```diff
@@ -720,6 +720,48 @@
         ]
       }
     },
+    "/core/packages/v1alpha1/repositories/c/{context.cluster}/permissions": {
```
antgamdia, line 723: Just FTR, we should also update the `openapi.yaml` file when upgrading the API. 
See https://github.com/vmware-tanzu/kubeapps/pull/5610

cmd/kubeapps-apis/proto/kubeappsapis/core/packages/v1alpha1/repositories.proto
```diff
@@ -499,3 +505,20 @@ message UpdatePackageRepositoryResponse {
...
+   // The context (cluster/namespace) for the request
+   Context context = 1;
+ }
+
+ message PackageRepositoriesPermissions {
+   kubeappsapis.core.plugins.v1alpha1.Plugin plugin = 1;
+
+   map<string, bool> global = 2;
+
+   map<string, bool> namespace = 3;
```
antgamdia, line 519: Please, add a short sentence above each param for the autogenerated docs

antgamdia (approved):
Neat! It's great we can finally free the UI up from performing the low-level `canI` for the repo auth.

beni0888 (approved):
LGTM! Great job @castelblanque! 👏🏻

dlaloue-vmware:
From the code, it seems that "global" is considered equivalent to not providing a namespace when doing the checks, which i assume is doing a check cluster wide (similar to the --all-namespace option?)

This is technically not accurate, and aside from cluster admins, this may not always return the correct results.

In the case of Carvel and Helm plugins, being able to create/update/delete global repositories is equivalent to having the permission on the corresponding global namespace - which is different than having permissions at the cluster level.
For example, i may be a user with privileges to create global repos and yet not be a cluster admin. I would not be able to manage global repositories via the UI.

For a user scoped to a namespace, "list" and "get" will return false for the global checks, even though the user can list and view global repositories.

Flux does not really have a notion of global repositories, I think the checks should always be false for the global map (i.e. we can skip the checks).

castelblanque:
Thanks all for the review!

@dlaloue-vmware I agree, switched now to the global namespace (in Helm and Kapp). For Flux, no global info is returned.

PR #5605, opened 2022-11-03
Remove "isLoaded" flag from deployment form

### LINKED ISSUE #5603: Wrong initial values in the form when upgrading a package
**Describe the bug**
During the initial load of the package's values and deployed values, it seems the former takes precedence over the latter. 

**To Reproduce**

1. Deploy a package with values != default AND version != latest
2. Upgrade the package
3. Notice the values are wrong, but they get updated if switching tabs

**Expected behavior**
The displayed values should be the deployed ones, not the pkg defaults'

**Screenshots**
deployedValues

N/A

**Additional context**
N/A

antgamdia:
### Description of the change

This PR solves a bug found during a manual test prior to the release: when selecting a different version, the `isLoaded` flag was preventing a proper execution of the useEffect flow. The flag has been replaced with `isLoading`, which just avoid re-renders when an expensive operation is running (like parsing the schema).

### Benefits

Users upgrading package will see the deployed values (correct) instead of the pkg defaults (incorrect). 

### Possible drawbacks

In those cases, the component will re-render twice, but that's not a big deal (no expensive backend calls, the table is now performing well, etc).

### Applicable issues

- fixes #5603

### Additional information

deployedValuesFix

### DIFF
--- a/dashboard/src/components/DeploymentForm/DeploymentFormBody/DeploymentFormBody.tsx
+++ b/dashboard/src/components/DeploymentForm/DeploymentFormBody/DeploymentFormBody.tsx
@@ -83,7 +83,6 @@ function DeploymentFormBody({
   );
 
   const [restoreModalIsOpen, setRestoreModalOpen] = useState(false);
-  const [isLoaded, setIsLoaded] = useState(false);
   const [isLoading, setIsLoading] = useState(true);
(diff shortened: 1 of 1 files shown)


ppbaena (approved):
Nice catch during testing the release. Thanks!

beni0888 (approved):

PR #5591, opened 2022-10-28
Issue4870 - removed use of UserManagedSecrets flag for helm plugin repos

dlaloue-vmware:
refactored helm plugin to not depend on a system flag "UserManagedSecrets" but allow secrets to be managed by user or by kubeapps per repository (issue #4870)

PR #5610, opened 2022-11-04
Regenerate `openapi.yaml` file

### LINKED ISSUE #5609: Incorrect swagger spec - reconciliationOptions.interval - invalid value for string type
**Describe the bug**
The `POST` request for `/apis/core/packages/v1alpha1/installedpackages` throws the following error because the `reconciliationOptions.interval` property in the request body is an integer and not a string, but in the swagger spec, this property is specified to be an integer and this was the case in older chart versions.
 
```js
{"code":3,"message":"proto: (line 1:322): invalid value for string type: 0"}
```

**To Reproduce**

1. Install latest chart
2. Install a helm package using `/apis/core/packages/v1alpha1/installedpackages` and the following body
```js
let body = {
  "param": {
    "availablePackageRef": {
      "context": {
        "cluster": "default",
        "namespace": "kubeapps"
      },
      "identifier": "bitnami/argo-cd",
      "plugin": {
        "name": "helm.packages",
        "version": "v1alpha1"
      }
    },
    "targetContext": {
      "cluster": "default",
      "namespace": "default"
    },
    "name": "argo-cd",
    "pkgVersionReference": {
      "version": "4.3.1"
    },
    "reconciliationOptions": {
      "interval": 0,
      "suspend": true,
      "serviceAccountName": "kubeapps-admin"
    }
  },
  "url": "<base_path>/apis/core/packages/v1alpha1/installedpackages",
  "token": "",
  "kubeAppToken": ""
}
```
4. See the error 
```js
{"code":3,"message":"proto: (line 1:322): invalid value for string type: 0"}
```

**Expected behavior**
This error shouldn't appear as the swagger spec was followed and in the swagger spec `reconciliationOptions.interval` is of value int32 and not string.

**Resolution**
Instead of integer pass string value in `reconciliationOptions.interval`. So in the above body `0` changes to `"0"`.

**Version Info**
- Version latest [2.6.0] from a week back
- Kubernetes version [1.22]

Let me know if I am misunderstanding something or there is anymore context I can provide. For now I will keep it as string and pass but I believe somewhere the code broke that is why it now only accepts string when in the swagger spec it is specified as int32.

antgamdia:
### Description of the change

This PR updates the public `openapi.yaml` file we are bundling along with the dashboard component. We forgot to do it (we were deprecating components like kubeops and assetsvc and we just missed it).
As reported, there were some mismatches in our API docs, so a sync was required by all means.

Additionally, I've squeezed a few words in the release process docs not to forget about it in the next releases.

### Benefits

Public OpenAPI file will get in sync again.

### Possible drawbacks

N/A  (we will have to repeat the process once #5591 gets merged)

### Applicable issues

- fixes #5609 

### Additional information

image


ppbaena (approved):
LGTM

PR #5618, opened 2022-11-08
Adjust concurrency in gha workflow

beni0888:
### Description of the change

With these changes we achieve the following:
- Do not trigger the `Kubeapps Main Pipeline` workflow for every push, only for pushes to the master branch.
- Use a different concurrency group for the `Release Pipeline`, so a later workflow triggered from a push to the main branch won't cancel the release pipeline.
- Do not trigger the `Kubeapps Main Pipeline` workflow on tags creation. We trigger the release pipeline for this event and don't want a second pipeline running on the same version of the code, that could interfere with the former (eg. canceling the release pipeline because they both run in the same concurrency group).

### Benefits
- Less number of pipelines triggered, and the triggered ones are more meaningful.
- A more resilient release flow.

### Possible drawbacks
- The main pipeline isn't triggered until a PR is triggered so the feedback loop could be larger.

### Applicable issues

- related to #4436 

### Additional information
N/A

### DIFF
--- a/.github/workflows/kubeapps-main.yaml
+++ b/.github/workflows/kubeapps-main.yaml
@@ -6,6 +6,8 @@ name: Main Pipeline
 
 on:
   push:
+    branches:
+      - main
     tags-ignore:
       - '*'
   pull_request:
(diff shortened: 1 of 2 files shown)


antgamdia (approved):
Thanks!

PR #5620, opened 2022-11-08
Add status badges for GHA's workflows

beni0888:

### Description of the change

This PR adds several status badges to the README.md file for some of the GitHub Action's workflows (main, release and codeql).

### Benefits
The latest status of the GitHub actions workflows is shown in the README.md file.

### Possible drawbacks
N/A

### Applicable issues

- related to #4436 

### Additional information
N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,9 @@
 # <img src="site/content/docs/latest/img/logo.svg" width="40" align="left"/> Kubeapps
 
 [![CircleCI](https://circleci.com/gh/vmware-tanzu/kubeapps/tree/main.svg?style=svg)](https://circleci.com/gh/vmware-tanzu/kubeapps/tree/main)
+[![Main Pipeline](https://github.com/vmware-tanzu/kubeapps/actions/workflows/kubeapps-main.yaml/badge.svg)](https://github.com/vmware-tanzu/kubeapps/actions/workflows/kubeapps-main.yaml)
+[![Release Pipeline](https://github.com/vmware-tanzu/kubeapps/actions/workflows/kubeapps-release.yml/badge.svg)](https://github.com/vmware-tanzu/kubeapps/actions/workflows/kubeapps-release.yml)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,9 @@
 # <img src="site/content/docs/latest/img/logo.svg" width="40" align="left"/> Kubeapps
 
 [![CircleCI](https://circleci.com/gh/vmware-tanzu/kubeapps/tree/main.svg?style=svg)](https://circleci.com/gh/vmware-tanzu/kubeapps/tree/main)
+[![Main Pipeline](https://github.com/vmware-tanzu/kubeapps/actions/workflows/kubeapps-main.yaml/badge.svg)](https://github.com/vmware-tanzu/kubeapps/actions/workflows/kubeapps-main.yaml)
+[![Release Pipeline](https://github.com/vmware-tanzu/kubeapps/actions/workflows/kubeapps-release.yml/badge.svg)](https://github.com/vmware-tanzu/kubeapps/actions/workflows/kubeapps-release.yml)
+[![CodeQL](https://github.com/vmware-tanzu/kubeapps/actions/workflows/codeql-scheduled.yml/badge.svg)](https://github.com/vmware-tanzu/kubeapps/actions/workflows/codeql-scheduled.yml)
```
antgamdia, line 6: I'm not sure we really need these two ones. I mean, given that the release is outsourced to Bitnami, does really make sense for us to show the status here? 
Regarding the CodeQL, well... it's ok, but, what's really the point (for any user coming to the project, not for us, I mean)

Anyway, +1ing, we can always change it back :P

antgamdia (approved):

README.md
beni0888: Besides that, we commented that would be interesting to add a new workflow in charge of running the `Full Integration Workflow` on schedule. Do you think it would be useful to add its badge to the `README.md` file?

beni0888: @antgamdia Taking into account that we already have the `Full Integration Pipeline` and to run it on schedule we don't need to create a new workflow, but just add a new `on.schedule` event condition, I've added a badge for it and remove the one for the `Release Pipeline` because, as you pointed out, it maybe doesn't add much value. WDYT?

antgamdia: Ok, great! Makes sense. Let's keep all the badges for now, at least, it's great to see everything green :P

PR #5622, opened 2022-11-08
Add more helm custom fields

### LINKED ISSUE #5128: Repos API/Helm - Analyze and migrate the former `SyncJobTemplate` options to the API
**Summary**
Some of the options that can be set from the chart inside the syncJobPodTemplate are no longer available via API.
For instance, the proxy options and custom labels/annotations.

**Background and rationale**
After the new Repos API, the majority of the options under the `syncJobPodTemplate` weren't migrated. We 

**Description**
For example:

- Add `http_proxy`, `https_proxy` and `no_proxy` to the helm apprepo via API
- Add annotation and labels to the helm apprepo's pods via API

**Acceptance criteria**

- [ ] Get a list of features that are/were available vs the ones supported by the repos API.
- [ ] For each one, decide whether to ignore it or accept it.

**Additional context**
IMHO, this is not urgent as, at least, users can still apply these options via chart.

antgamdia:
### Description of the change

This PR brings more custom fields to the Helm plugin (repos API). Specifically, I've mimicked the currently customizable values through the `values.yaml` file. That is: `tolerations`, `nodeSelector`, `securityContext` and `proxyOptions`. 
Notwithstanding, users were able to set their custom `PodTemplateSpec` for the sync job in the old fashion way, but there's too much information to be included as a custom value in the repos API. But why?

Well, since we are generating code from the grpc messages, if we define a `pod_template_spec` field of type `PodTemplateSpec` in the grpc proto file.... we need to have the corresponding type defined.
In doing so, there's two alternatives: 1) using an existing definition and importing it using Buf (via BSR) or 2) maintaining ourselves the grpc messages definition.

- Option 1: almost not viable. There is no official definition in BSR, just one coming from a private company with no guarantees whatsoever. I've explored that way, but I don't feel we should go in this direction. Too risky.
- Option 2: if we are to maintain the messages, we have two options:  2.1) using the one from Kubernetes or 2.2) defining our own model.
  - Option 2.1: the autogenerated file is huge (=every k8s API object model...) and we would have to ensure we're in sync when a new k8s version is out. 
  - Option 2.2: it would require copying a subset of messages from the k8s API, but keeping just those ones we want.
 
I've gone with option 2.2, but simplifying the fields we support. Just 1 or 2 objects having simple types like string or bool.
I'm adding some PR review messages to highlight what I mean.

That said, I'm open to any other better ideas you may have.

On the other hand, there's a separate issue: what should we display in the Repos UI? 
Well, in this case, I've opted for just showing the proxy options. The rest of the fields are just related to the PodTemplateSpec and I feel they are too complex for even advanced users. If someone really wants to tune them up, they can just go an invoke the API.

### Benefits

Users will be able to customize the proxy options (along with a few other options) via the repos API instead of just using the chart `initialRepos` field.

### Possible drawbacks

Not every custom field is available through the UI, but I don't really think we want to.

### Applicable issues

- fixes #5128

### Additional information

N/A


cmd/kubeapps-apis/proto/kubeappsapis/plugins/helm/packages/v1alpha1/helm.proto
```diff
@@ -195,6 +195,18 @@ message HelmPackageRepositoryCustomDetail {
...
+  ProxyOptions proxy_options = 5;
+
+  // TODO
+  map<string, string> node_selector = 6;
+
+  // TODO
+  repeated Toleration tolerations = 7;
+
+  // TODO
+  PodSecurityContext security_context = 8;
```
antgamdia, line 209: New options for the Helm plugin (repos API)
[will add the description soon, I just wanted to share my approach with you ASAP]

cmd/kubeapps-apis/proto/kubeappsapis/plugins/helm/packages/v1alpha1/helm.proto
```diff
@@ -207,3 +219,54 @@ message RepositoryFilterRule {
...
+  repeated int64 supplemental_groups = 4;
+  optional int64 f_s_group = 5;
+
+  // TODO(agamez): more complex fields are not supported yet
+  // optional SELinuxOptions seLinuxOptions = 1;
+  // optional WindowsSecurityContextOptions windowsOptions = 8;
+  // repeated Sysctl sysctls = 7;
+  // optional string fsGroupChangePolicy = 9;
+  // optional SeccompProfile seccompProfile = 10;
+}
```
antgamdia, line 272: As I said in the PR description, I only kept the most simple fields. This is preventing us from importing the whole proto document from k8s.

cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/repositories.go
```diff
@@ -152,6 +157,11 @@ func newHelmRepoCrd(repo *HelmRepository, secret *k8scorev1.Secret, imagePullSec
 			Description:           repo.description,
 			PassCredentials:       repo.auth != nil && repo.auth.PassCredentials,
 			Interval:              repo.interval,
+			// TODO(agamez): are more fields here if they're requested
```
absoludity, line 160: Is this meant to be:
```suggestion
			// TODO(agamez): add more fields here if they're requested
```

absoludity (approved):
LGTM, after a rebase/resolve. Thanks Antonio!

absoludity:
I've just merged main again and fixed the conflict. Should be good to land once tests pass.

PR #5636, opened 2022-11-15
Add makefile lint target

beni0888:

### Description of the change

Currently, we have several linters running in our GHA pipeline, but we do not have an easy way of running those linters locally. This PR adds a new target to the makefile that allows executing those linters locally, so we can easily run them as git hooks or whatever.

### Benefits

We are able to execute the linters locally.

### Possible drawbacks
None

### Applicable issues

- related to #5628


absoludity (approved):
Excellent, thanks for making that work locally easily @beni0888 !

Issue #4115, opened 2022-01-19
flux plugin: clean up old charts from chart cache after repo update


gfichtenholt:
Documenting leftover items for flux plugin per Pepe's request so we can prioritize it against other work

This is a leftover item from #3899
@antgamdia said: 
"Think about an already cached repo, if the repo authors eventually sunset a given version and it gets removed from the index (case 1), what would happen then? What if a whole chart got removed (case 2)?
Let's suppose:

    (case 1): my-chart: v1, v2 --> then my-chart:v2
    (case 2): my-chart: v1, foo-chart: v1 --> then foo-chart: v1
"
case 1 is no problem but case 2 is an issue today. In both cases there will be an event 'repo XYZ' was modified. In handling this event, I will call SyncCharts(repo XYZ). Currently I just go through the most recent versions for every chart and add that. I don't delete any charts prior to that. So in case 1: you'll end up with my-chart:v2, same as before the update. In case 2, you'll end up with my-chart: v1, foo-chart: v1. The problem is my-chart: v1 does not belong in the cache anymore: it takes up precious space and the user might get details on a package which no longer exists. 

There is another thing that kind of complicates case 2: my-chart: v1 might have been added explictly to to chart cache by the user in a scenario where the entry was first evicted due to lack of space and then added due to user actually performing getAvailableChartDetail(my-chart: v1). So in that case, assuming my-chart: v1 is still a valid chart, do I delete it or leave it alone because its in the cache due to user's request?

antgamdia:
FWIW, we have had this (case 2) problem for quite a long time. In fact, we had this old issue tracking it: https://github.com/kubeapps/kubeapps/issues/164

I'd personally opt for clearing this cache entry in case of doubt. If we do so, the worst-case scenario is a user waiting some seconds. Otherwise, a user would see an item that does no longer exists, which is a worse UX, IMHO.

PR #5644, opened 2022-11-16
fix for flux plugin: clean up old charts from chart cache after repo update #4115 

gfichtenholt:
fix for long-standing issue https://github.com/vmware-tanzu/kubeapps/issues/4115

plus DRYing up a bunch of existing test code

Not 100% sure who to ask for review. Its not urgent. If no-one materializes in the next few days, I will ask

The only thing of interest here is really a new func call 
```
chartCache.PurgeObsoleteChartVersions(charts)
```
executed before an old call to 
```
SyncCharts()
```

There is a new integration test I added for this scenario.
Most everything else is test code cleanup and doesn't need any review


gfichtenholt:
could one of you guys review this when you get a chance? Thanks

dlaloue-vmware (approved):

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/cache/chart_cache.go
```diff
@@ -329,6 +329,47 @@ func (c *ChartCache) DeleteChartsForRepo(repo *types.NamespacedName) error {
...
 }
 
+func (c *ChartCache) DeleteChartsForRepo(repo *types.NamespacedName) error {
+	log.Infof("+DeleteChartsForRepo(%s)", repo)
+	defer log.Infof("-DeleteChartsForRepo(%s)", repo)
+
+	return c.deleteChartsHelper(repo, sets.String{})
+}
+
+func (c *ChartCache) PurgeObsoleteChartVersions(keepThese []models.Chart) error {
```
absoludity, line 339: Just reading the code below, I think `keepThese` is the slice of charts for which you're wanting to keep all versions? (as in, below, each chart is iterated with the redis key for each version added to the set of keys to keep for that repo, which is then passed as `keep` to `deleteChartsHelper`).

I must be missing something here, because it looks like it only ever calls `deleteChartsHelper` with repos of charts that were in the `keepThese` slice? I expected to find a function that purges all *other* repositories, other than the ones you wanted to keep (maybe it does and I'm just mis-reading).

EDIT: Oh - reading further down, I think I understand: this function is called when re-importing charts after an update to the repo, for example, so `keepThese` is actually populated from the new data, meaning that if the new data no longer includes a certain version, it'll get purged here. OK, I think I was confused initially because it wasn't clear to me that `keepThese` was the new data that hadn't been synced yet. (?)

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/cache/chart_cache.go
```diff
@@ -329,6 +329,47 @@ func (c *ChartCache) DeleteChartsForRepo(repo *types.NamespacedName) error {
...
+	log.Infof("+PurgeObsoleteChartVersions()")
+	defer log.Infof("-PurgeObsoleteChartVersions")
+
+	repos := map[types.NamespacedName]sets.String{}
+	for _, ch := range keepThese {
+		if ch.Repo == nil {
+			continue
```
absoludity, line 346: Should that be an error if this function is called with a chart with a nil repo? Or at least, logging the issue? Wondering why we'd not want to report that somehow?

absoludity (approved):
Thanks for adding the (very helpful) PR description Greg. I just looked at the `PurgeObsoleteChartVersions` function.

gfichtenholt:
thank you for reviewing

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/cache/chart_cache.go
gfichtenholt: yes, you got it

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/cache/chart_cache.go
gfichtenholt: shouldn't ever happen, but you're right. Let me at least log it

PR #5647, opened 2022-11-16
Fix srp_report GHA's job

beni0888:

### Description of the change

It seems that there is a bug in the current version of the GHA's `report_srp` job we've just merged, and for some reason, the SRP_CLI_VERSION is not being properly passed in the call to the `srp-source-provenance` action.

### Benefits

Fix the bug in the GHA's `report_srp` job, so it works as expected.

### Possible drawbacks

None

### Applicable issues

- related to #4436


ppbaena (approved):
Thanks @beni0888!

PR #5649, opened 2022-11-16
Fix typo in srp command

beni0888:

### Description of the change

This PR fixes a typo in the call to `srp provenance source` command in the GHA's `srp-source-provenance` action.

### Benefits

A typo in the call to a command is removed, so it should work as expected now.

### Possible drawbacks

None.

### Applicable issues

- related to #4436

### DIFF
--- a/.github/actions/srp-source-provenance/action.yml
+++ b/.github/actions/srp-source-provenance/action.yml
@@ -58,7 +58,7 @@ runs:
         echo "COMP_UID=$COMP_UID"
         mkdir -p /tmp/provenance
         sudo srp provenance source \
-        --verbose\
+        --verbose \
         --scm-type git \
         --name "kubeapps" \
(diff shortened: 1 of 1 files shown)


gfichtenholt (approved):
lgtm

absoludity (approved):
Did you end up checking if you can run these locally? I'll try again with the tool I mentioned before, as it'd be great to be able to check these locally and avoid stuff. Thanks for all your efforts to see this through! It'll make it easy for us to switch and enjoy the benefits :)

beni0888:
> Did you end up checking if you can run these locally? I'll try again with the tool I mentioned before, as it'd be great to be able to check these locally and avoid stuff. Thanks for all your efforts to see this through! It'll make it easy for us to switch and enjoy the benefits :)

To be honest, I didn't check... 😓

PR #5657, opened 2022-11-17
Update deprecated parameter in the call to srp tool

beni0888:

### Description of the change

Even though the `report_srp` job of the GHA's pipeline is working properly, it's showing a warning due to the use of a deprecated argument in the call to the `srp` tool. Here we replace that argument by the new recommended one.

### Benefits

We get rid of a deprecated argument in the call to the `srp` CLI tool and remove the risk of a future failure when updating to a new version that no longer supports the deprecated argument.

### Possible drawbacks

None

### Applicable issues

- related to #4436 

### Additional information

* https://github.com/vmware-tanzu/kubeapps/actions/runs/3485334417/jobs/5830985505#step:3:701


absoludity (approved):
Thanks :)

PR #5660, opened 2022-11-17
Move gke_e2e_tests from reusable workflow to custom action

beni0888:

### Description of the change

Currently, we have a reusable workflow for running the e2e tests in a GKE cluster. This approach, although valid is probably not the best one for this problem as we have to explicitly declare every single env var the workflow use as an input, pass those env vars in every call to the workflow, and finally, inside the reusable workflow turn the inputs into the env vars the scripts are relying on. A better approach for this is to define a custom action using the `composite` approach, that will automatically inherit all the env vars available in the calling workflow.

### Benefits

Make our GHA setup simpler a more easily maintainable.

### Possible drawbacks

None

### Applicable issues

- related to #4436


beni0888:
Closing this PR as it is not feasible to continue with it due to a bug in GHA actions by which secrets containing JSON docs as values don't get masked when shown in the build logs. That added to the fact that custom actions don't inherit the secrets from the caller workflow and they have to be passed as inputs, makes it impossible to continue with this task. We'll have to stick to a reusable workflow, at least until the GitHub Actions guys fix the bug.

PR #5664, opened 2022-11-18
Issue5542 - Unable to edit or delete newly created flux package repository

### LINKED ISSUE #5542: Unable to edit or delete newly created flux package repository
Scenario:
* user A has cluster admin access to namespace A
* user A logs in to portal, switch context to namespace A and creates a new flux repo there
* the repo is created successfully but EDIT, REFRESH and DELETE button are disabled. A bit ironic since REFRESH button was enabled in previous screen

Detailed repro steps:
See attached .txt file, some of steps 1-22, having to do with setting up ns A and user A 

screen 1
Screen Shot 2022-10-20 at 9 12 39 PM

screen 2
Screen Shot 2022-10-20 at 9 05 37 PM

[multi-tenant-setup.txt](https://github.com/vmware-tanzu/kubeapps/files/9835624/multi-tenant-setup.txt)

dlaloue-vmware:
### Description of the change

This change completes the changes previously made in an earlier change regarding flux repository edit/delete buttons enablement issues.
The main switch is to change the scope of a flux repository from global to namespaced. A flux repository behaves more similar to a namespaced repository than a global repository, making its handling easier in the UI.

The change also includes a fix where carvel global repositories were not included, even if the user has list/view access to them.

Finally, the create/update repository dialog in the UI has been updated to reflect that the default scope of a flux repository is namespaced, and not global, and that it cannot be changed.

### Benefits

Users that have namespace-only access can now properly manage the flux repositories they created.

### Applicable issues

- fixes #5542

### Additional information

This change does not include a fix for the page and table buttons when the "show in all namespaces" toggle is on.
Users can use the workaround to toggle off the switch. For this reason, a proper fix is best delayed to when the repository page is revamped.


gfichtenholt:
just adding an expclit note here so we can link to it in the future: this change will make all repositories in flux plugin "namespaced". Previously they were NOT namespaced.  
per slack: "...change is to expand the definition of namespaced a bit :slightly_smiling_face:
“if you create a repo in one namespace, it is available to be used in that namespace, and may be available to be used in other namespaces you have access to,  if supported by the plugin”.
and only flux (without no-cross enabled) supports such extended capability

gfichtenholt (approved):
My only ask is in flux repo.go, put an explicit comment mentioning that we chose to change the way flux repos are returning the namespacedScoped flag on purpose with maybe a link ref to the discussion. This kind of thing tends to be forgotten quickly. Thanks

dlaloue-vmware:
yes, will do.

PR #5673, opened 2022-11-20
issue4870  - refactor helm to use managed-by annotation

### LINKED ISSUE #4870: [Repository] refactor helm to use managed-by annotation instead of system flag
For repositories, secrets used for authentication can be either user managed or kubeapps managed.
The current mechanism is based on a system flag to enable one more or the other mode. This limits the flexibility to have some users use one way (e.g. cli) or the other (e.g. UI) in a single Kubeapps installation.

this issue is to refactor the helm plugin.

dlaloue-vmware:
### Description of the change

The previous fix for issue 4870 did not introduce the managed-by annotation as per the spec, but relied on naming pattern for backward compatibility.
This presents the issue that new secrets will still depend on the old naming pattern.

This change ensures that new or updated secrets have the managed-by annotation, and that the code looks for this annotation.

The code stills falls back to the naming pattern for backward compatibility, but now with a deprecation path. We can remove the naming pattern support in a future version.

### Applicable issues

- fixes #4870


cmd/kubeapps-apis/plugins/helm/packages/v1alpha1/repositories_auth.go
```diff
@@ -31,12 +31,16 @@ const (
 	SecretCaKey         = "ca.crt"
 	SecretAuthHeaderKey = "authorizationHeader"
 	DockerConfigJsonKey = ".dockerconfigjson"
+
```
gfichtenholt, line 34: may I suggest  "app.kubernetes.io/managed-by" for the key (ref https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels).  Seemed fitting to me so I was planning to use that in flux plugin

dlaloue-vmware: i do not remember what we discussion we previously had, but this is following what is in the carvel plugin.
if we decide to change, we would either have different implementations, or need what we have in carvel.

gfichtenholt (approved):
OK then, no objections

PR #5685, opened 2022-11-21
Add some minor improvements in the GHA's kubeapps-general pipeline

beni0888:

### Description of the change

This commit adds just some minor improvements to the GHA's kubeapps-general workflow. The changes consist mainly of applying more restrictive bash options (`set -eu`) so the scripts fail out loud when a required env var is not defined and finish their execution when an intermediate command returns a non-zero code.

### Benefits

We follow shell script's best practices to try to avoid subtle bugs.

### Possible drawbacks
More restrictive bash options could result in breaking some currently running scripts, or even lead us to discover some unnoticed misbehaviors.

### Applicable issues

- related to #4436 

### Additional information

None

### DIFF
--- a/.github/workflows/kubeapps-general.yaml
+++ b/.github/workflows/kubeapps-general.yaml
@@ -258,6 +258,7 @@ jobs:
       - name: "Install helm (minimum and stable)"
         run: |
           set -eu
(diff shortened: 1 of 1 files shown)


ppbaena (approved):
LGTM

PR #5694, opened 2022-11-22
Fix GHA's release workflow

beni0888:

### Description of the change

This commit fixes the GHA's Release Pipeline. Currently, the workflow is not working properly because the `sync_chart_from_bitnami` only runs when the pipeline is triggered from the main branch, and it should run when it's triggered from a version tag also.

### Benefits

The GHA's Release Pipeline works as expected.

### Possible drawbacks

None

### Applicable issues
- related to #4436

### DIFF
--- a/.github/workflows/kubeapps-general.yaml
+++ b/.github/workflows/kubeapps-general.yaml
@@ -617,7 +617,7 @@ jobs:
   sync_chart_from_bitnami:
     needs:
       - setup
-    if: needs.setup.outputs.running_on_main == 'true'
+    if: needs.setup.outputs.running_on_main == 'true' || needs.setup.outputs.running_on_tag == 'true'
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


antgamdia (approved):

.github/workflows/kubeapps-general.yaml
```diff
@@ -617,7 +617,7 @@ jobs:
   sync_chart_from_bitnami:
     needs:
       - setup
-    if: needs.setup.outputs.running_on_main == 'true'
+    if: needs.setup.outputs.running_on_main == 'true' || needs.setup.outputs.running_on_tag == 'true'
```
absoludity, line 620: I'm still not clear in my mind why we want to sync the bitnami chart each time a PR targets main? Wouldn't we only want to do so if there isn't a PR already open to sync the bitnami chart into our main? (given that the one PR can be recreated or just have latest changes merged etc?)

Thanks either way!

Issue #5671, opened 2022-11-18
POST requests to /apis/kubeappsapis.core.packages.v1alpha1.PackagesService/CreateInstalledPackage and /apis/kubeappsapis.core.packages.v1alpha1.PackagesService/DeleteInstalledPackage fail with a 502


RGPosadas:
**1. Describe the bug**

POST requests to create/delete a helm release sometimes fail with a 502 due to a timeout being reached (30s). This timeout is always reached when deploying our large charts (~30+ k8s resources). Our small charts (~3 resources) do not encounter this timeout. However, the **resources eventually get created/deleted**, and the release is seen on/removed from the `Applications` tab after a browser refresh. Tested on both Chrome and Firefox.

This issue only appeared when transitioning from kubeapps helm chart version `10.2.2` to `12.1.0`.

**2. To Reproduce**

**2.a Helm Chart Setup**
- `kubeapps` helm chart version `12.1.0`, for dependencies:
   - `common` chart version `2.1.2`
   - `postgresql` chart version `12.1.0`
   - `redis` is fully disabled as we do not use `flux` packages
- Since we run on GKE and have GCP Artifact Registry as a package repo, we have the flag `frontend.proxypassAccessTokenAsBearer: true` enabled
- `kubeappsapis.pluginConfig.core.packages.v1alpha1.timeoutSeconds: 300` as default and is untouched
- `dex` as our OIDC provider. Relevant config:
```yaml
  authProxy:
    clientID: "kubeapps"
    extraFlags:
      - --skip-provider-button
      - --oidc-issuer-url=https://<dex-domain>
      - --provider-ca-file=/certs/ca.crt
      - --cookie-secure=true
      - --cookie-expire=24h
      - --scope=openid profile email groups
      - --proxy-prefix=/oauth2
      - --redirect-url=/oauth2/callback
```
- We offer kubeapps to internal customers who do not have direct access to our k8s cluster, therefore we have `pinnipedProxy.enabled: true`. Relevant config:
```yaml
    clusters:
    - name: "default"
      isKubeappsCluster: true
      apiServiceURL: "<provided by pinniped-concierge>"
      certificateAuthorityData: "<provided by pinniped-concierge>"
      pinnipedConfig:
        enabled: true
```
- `pinniped-concierge` is running on version `0.20.0`

**2.b Kubeapps Dashboard - Actions Taken**
1. Login successfully
2. Deploy a chart that will create 30+ k8s resources

[...]

PR #5695, opened 2022-11-22
Improve wording to avoid implying pinniped is used either way.

absoludity:

### Description of the change

While chatting about [OIDC configuration](https://github.com/vmware-tanzu/kubeapps/issues/5671#issuecomment-1324168711) with @RGPosadas I was looking through the docs and found that it reads currently as if you should use Pinniped either way.

This change just clears that up a bit, I hope, so that people use the cluster's own OIDC support when they can, and pinniped when they can't.

### Applicable issues

- ref #5671


ppbaena (approved):
Thanks for improving Kubeapps documentation!

Issue #5692, opened 2022-11-21
Support additional default values from a package


absoludity:
The first chunk of work that we can implement from the [Kubeapps Custom Package Data](https://docs.google.com/document/d/1e29bHJ2O4NqdFmOHcYdczQ74QnOEwpQZo7UJObK7yXQ/edit?usp=sharing) is to support custom default value files when they are included within a package.

## Use case 1: A single customized default values file

```
As a Kubeapps operator,
I want to repackage my packageA with a custom default values file (e.g. production-values.yaml)
So that any user who installs package A will be presented with my custom default values.
```

## Use case 2: Multiple customized default values files

```
As a Kubeapps operator,
I want to provide my package A with multiple default values files (production, staging etc.),
So that any user who installs package A will be able to select one of these default value files as a base for the deployment.
```

Or, from the user point-of-view:

```
As a Kubeapps user,
I want to install my company's custom packageA,
So that I can choose from the defaults provided by my company for my situation.
```

## Task breakdown

This task will focus enable the functionality for the Helm plugin only, though the 
- [x] Define a format for including data in the packages themselves. This will be plugin-specific, so for Helm (and flux) it can be just matching on  `values-*.yaml` file in the root. If we later decide to support this functionality with Carvel, it will similarly be an extra file in the bundle's root.
- [x] Design and implement support for the in-package custom defaults in the Kubeapps-APIs service, so that the Helm plugin is able to populate additionally a custom_defaults_values map, which, together with the existing “default_values”, allows the package’s original defaults to be overlaid with custom values.
- [x] Implement the dashboard changes so that, when present, custom defaults can be selected by the user. By default, the first custom defaults will be selected (sorted alpha-numerically). When only one custom default is present, it will be used without giving the user a selection at all.

absoludity:
So the first case is now done (well, in review). I'm looking at the case 2. The installation of a package seems straight forward enough (small UX change), but where I'm uncertain with case 2 is when viewing an installed appication: Kubeapps will (currently) not have a way to know which default value was used by the user during install.

Fundamentally, there's no way for Kubeapps to know without us storing some indication with the release itself, which seems like the obvious solution: a custom values.yaml can include a `custom-defaults-name` field, for example:

```
# Copyright 2022 the Kubeapps contributors.
# SPDX-License-Identifier: Apache-2.0

image:
  repository: bitnami/nginx
  tag: "1.23.3-debian-11"
  pullPolicy: IfNotPresent

[...]

PR #5702, opened 2022-11-23
Extract custom default values from package tarball

absoludity:
### Description of the change

This PR begins the work specified in #5692 by updating the `AvailablePackageDetail` message to include custom default values and updating the tar utils so that they are fetched and returned (if present) when the other files are fetched from a tarball.

### Applicable issues

- ref #5692


cmd/kubeapps-apis/docs/kubeapps-apis.swagger.json
```diff
@@ -4582,6 +4574,14 @@
           "description": "An example of default values used during package templating that can serve\nas documentation or a starting point for user customization.",
           "title": "Available package default values"
         },
+        "customDefaultValues": {
```
dlaloue-vmware, line 4577: this is just a suggestion, but maybe we can use a more neutral word? The description is using the term extra for example, another term could be "additional" or "alternate"?

dlaloue-vmware (approved):
code looks good.

cmd/kubeapps-apis/docs/kubeapps-apis.swagger.json
absoludity: I'd go for "additional" (as they are layered on top) rather than "alternate" (as they are not instead of), but yep, "additional" is good for me (better than "custom"). I'll update to that. Thanks.

PR #5718, opened 2022-11-25
Simplify release-related logic in GHA's kubeapps general pipeline

beni0888:

### Description of the change

This commit aims to simplify the logic in the GHA's Kubeapps General workflow. Currently, we have some logic that checks if the pipeline has been triggered by a tag pushed to the repo, and if it is, we run some jobs: push_images, sync_charts, release, etc. As we already have a top-level workflow `kubeapps-release.yml` that internally uses the general workflow, and only runs when a tag following the release pattern is pushed to the repo, we can just introduce an input parameter in the general workflow that controls whether those jobs should be run or not, instead of relying on the logic residing in the general pipeline.

### Benefits

GHA's workflows `kubeapps-general.yaml` and `kubeapps-release.yml` are simpler to understand and maintain.

### Possible drawbacks

None.

### Applicable issues

- related to #4436

### DIFF
--- a/.github/workflows/kubeapps-general.yaml
+++ b/.github/workflows/kubeapps-general.yaml
@@ -740,6 +740,8 @@ jobs:
       - # Assuming there is a personal access token created in GitHub granted with the scopes
         # "repo:status", "public_repo" and "read:org"
         name: Run the create_release script
+        env:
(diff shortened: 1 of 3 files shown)


antgamdia (approved):
Ah great, so this way the condition is only written once and then just passed as a param to the workflow. Neat!

PR #5730, opened 2022-11-25
Decommission circleci

beni0888:
### Description of the change

This commit decommissions the CircleCI pipeline and removes the temporary logic we added for development purposes from the GHA one.

### Benefits

We get rid of the CircleCI pipeline and stick to the GHA one.

### Possible drawbacks

None.

### Applicable issues

- related to #4436


.github/workflows/kubeapps-general.yaml
```diff
@@ -54,11 +46,11 @@ env:
   # as an output of the `setup` job.
   IMG_DEV_TAG: "build-${{ github.sha }}"
   # Apart from using a dev tag we use a different image ID to avoid polluting the tag history of the production tag
-  IMG_MODIFIER: "-ci-gha"
+  IMG_MODIFIER: "-ci"
   IMG_PREFIX: "kubeapps/"
   # We use IMG_PREFIX_FOR_FORKS for development purposes, it's used when the workflow is run from a fork of the kubeapps repo
-  IMG_PREFIX_FOR_FORKS: "beni0888/"
-  # IMG_PLATFORMS: "linux/amd64, linux/arm64"
+  IMG_PREFIX_FOR_FORKS: "johndoe/"
```
antgamdia, line 52: I'd use the kubeapps-bot account instead, just in case:

```suggestion
  IMG_PREFIX_FOR_FORKS: "kubeapps-bot/"
```

antgamdia (approved):
Whoa!! Finally, this moment arrived!  Thanks!

.github/workflows/kubeapps-general.yaml
beni0888: I chose `johndoe` on purpose to make clear that it is something external to the Kubeapps organization. IMHO using `kubeapps-bot` could lead to confusion.

absoludity: `your-dockerhub-username`?

absoludity (approved):

.github/workflows/kubeapps-general.yaml
antgamdia: Just asking: what if someone creates a `your-dockerhub-username` or a `johndoe` username in GitHub? Can they perform some undesired actions against our CI ? Or is it 100% safe ? 

If there is a minimum risk of an undesired action, I'd rather go with something we have control over, like the bot user. If not, anything you suggest seems ok to me.

absoludity:
Hi @beni0888 . The checks here just have one error, which I assume is related to removing the circleCI pipeline? If so, is there anything stopping this from landing? I'd like to start the next release soon, but am keen for this to land before I do so. Thanks either way!

.github/workflows/kubeapps-general.yaml
absoludity: I didn't understand exactly what you meant at first @antgamdia, as I thought the prefix is just used to push images when running in a forked repo, in which case it would fail anyway, but I think what you're suggesting is that maybe there is some way that us configuring a valid name that we don't own could lead to a privilege of that account? I don't see how, but I guess that's the point. Anyway, if so, why not give it an invalid name, like "<your-dockerhub-username>" (assuming the `<>` are invalid).

Anyway, I think this point we're discussing here is probably not that important compared to the actual change, so still +1 from me, whatever you decide @beni0888 .

beni0888: @antgamdia I think you're not really understanding the meaning of that variable 🙂 Its only purpose is to allow you to push the images to a different workspace in DockerHub other than `kubeapps`, on which you (the user running the pipeline on its own fork) won't have permissions to push. It was added mainly for developing purposes, as I didn't want to push images from my fork to DockerHub at the early stages of the migration. But I cannot see any chance of performing an undesired action thanks to the value of this variable. BTW, I think that @absoludity suggestion of using `your-dockerhub-username` instead of `johndoe` is a good idea 🙂

antgamdia: Ahhh, got it, got it. Thanks for the clarification! Sorry for the confusion :P
 +1 to `your-dockerhub-username` or even  `your-container-registry-username`

beni0888:
> Hi @beni0888 . The checks here just have one error, which I assume is related to removing the circleCI pipeline? If so, is there anything stopping this from landing? I'd like to start the next release soon, but am keen for this to land before I do so. Thanks either way!

No, there's nothing preventing us from landing this. I will apply your suggestion of using `you-dockerhub-username` in `IMG_PREFIX_FOR_FORKS` and tune the branch protection rules to remove those for CircleCI and add the GHA ones, and then will merge this.

Issue #5745, opened 2022-11-28
Repository UI - the edit dialog does not allow one to update credentials


dlaloue-vmware:
**Describe the bug**
After creating a repository with authentication, click edit to change the configuration. The UI does not allow for the authentication information to be updated.
This is an issue if users must update the password for example.

**Expected behavior**
Users should be able to update credentials information, especially in case the backend password has been updated (e.g. password rotation etc...)

**Desktop (please complete the following information):**
Kubeapps version  12.1.1/2.6.1

Issue #5747, opened 2022-11-28
flux repository - updated credentials are ignored


dlaloue-vmware:
**Describe the bug**

When updating a repository's auth credentials, the secret is not properly updated, and changes are ignored.

**To Reproduce**

Create a flux repository and configure auth credentials and a tls CA.
Edit the flux repository, update the tls CA and save. Notice that the auth secret is not updated with the new CA.
In fact, as long as one of the credentials remains "redacted", the other updates are ignored.

**Expected behavior**

Users should be able to update only some of the credentials info (e.g. only the password) and expect the secret to be updated with what changed, retaining what was not changed

**Desktop (please complete the following information):**

kubeapps  12.1.1 / 2.6.1

Issue #5746, opened 2022-11-28
helm repository - some auth updates are ignored or lost


dlaloue-vmware:
**Describe the bug**
There are 2 bugs happening during a repository update with auth changes:
 - the pull image secrets are wrongly deleted
 - the changes are ignored.

**To Reproduce**

Create a helm repository and configure auth credentials, pull image credentials and a tls CA.

issue 1:
Edit the helm repository, update the tls CA and save. Notice that the pull image secret is deleted, making the repository not readable anymore.

issue 2:
Edit the helm repository, update the tls CA and save. Notice that the auth secret is not updated with the new CA.
In fact, as long as one of the credentials remains "redacted", the other updates are ignored.

**Expected behavior**

issue 1:  the pull image secret is not expected to be deleted when modifying the tls CA.
issue 2:  users should be able to update only some of the credentials info (e.g. only the password) and expect the secret to be updated with what changed, retaining what was not changed

**Desktop (please complete the following information):**

kubeapps  12.1.1 / 2.6.1

PR #5763, opened 2022-11-30
[repository] fixes and enhancements for repository update support

dlaloue-vmware:
### Description of the change

This PR contains changes for two purposes:
 - improve feature parity between plugins
 - enable update support in UI

Feature parity:
 - flux: description support
 - flux: support changing to/from auth provider 
 - carvel: description support
 - carvel: support changing auth type 

Bug fixes to support update support in UI:
 - auth updates were not possible
 - fixed conditions for fields enablement
 - form fields data were sent incorrectly as a group, instead of only sending what applied
 - errors handling 'undefined' values
 - fixed which auth types are valid per plugin+type combination
 - fixed issue with helm plugins image pull secrets

helm/flux bug fixes:
 - fixed handling of redacted fields leading to secrets not being updated as expected or secret data being overwritten with 'redacted' as value.

(note: the PR has many small changes, but i was fixing issues as i was testing, hence the list)

### Benefits

With the changes, users should see more commonality between plugin features, as well as a more robust UI experience regarding repository creation/updates.

### Applicable issues

fixes issues #5745, #5746 and #5747

### Additional information

The issue 1 in # 5746 could not be reproduce with the latest bits, so no change were made for it.


gfichtenholt:
just curious, is there a similar issue in flux plugin?

dlaloue-vmware:
> just curious, is there a similar issue in flux plugin?
yes, issue #5747

gfichtenholt:
> > just curious, is there a similar issue in flux plugin?
> > yes, issue #5747

would it be ok to ask you to fix both as long as you're at it? Code should be similar

dlaloue-vmware:
> would it be ok to ask you to fix both as long as you're at it? Code should be similar
I don't mind.
but can it be 2 separate PRs, or you prefer me to do it in the same PR?

gfichtenholt:
> > would it be ok to ask you to fix both as long as you're at it? Code should be similar
> 
> I don't mind. but can it be 2 separate PRs, or you prefer me to do it in the same PR?

just do it here, its like 2 more files for a grand total of 5 or 6

gfichtenholt (approved):
I am ok with this. I think there are just some linter error(s) to clean up, then you're set

dlaloue-vmware:
moved to draft, as additional changes will be added

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo.go
```diff
@@ -462,7 +468,29 @@ func (s *Server) updateRepo(ctx context.Context, repoRef *corev1.PackageReposito
...
+			if request.Auth != nil && request.Auth.Type != corev1.PackageRepositoryAuth_PACKAGE_REPOSITORY_AUTH_TYPE_UNSPECIFIED {
+				return nil, status.Errorf(codes.InvalidArgument, "Auth provider cannot be configured in combination with another auth method")
+			}
+			if repo.Spec.Provider != "" && repo.Spec.Provider != "generic" && repo.Spec.Provider != provider {
+				return nil, status.Errorf(codes.InvalidArgument, "Auth provider cannot be changed.")
+			}
+			repo.Spec.Provider = provider
+		} else {
+			repo.Spec.Provider = ""
+		}
```
absoludity, line 492: I'm just trying to understand some of the logic here: on lines 487 we explicitly disallow the provider from being changed if it has been set, but on line 491 we allow it to be set to empty. So, unless I'm missing something, users can change the auth provider, they just need to first set it to empty then set it?

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo_auth.go
```diff
@@ -113,33 +114,44 @@ func (s *Server) handleRepoSecretForUpdate(
...
+		} else if isSameSecret {
+			// Do nothing if repo auth data came fully redacted
+			return nil, true, false, nil
+		} else {
+			// secret has changed, we try to delete any existing secret
+			if existingSecret != nil {
+				if err = secretInterface.Delete(ctx, existingSecret.Name, metav1.DeleteOptions{}); err != nil {
+					log.Errorf("Error deleting existing secret: [%s] due to %v", err)
+				}
+			}
```
absoludity, line 137: I don't remember how the discussions ended, but I thought in the past we'd talked about the fact that we should not delete a secret unless it was created by kubeapps? That is, if I create a repo in Kubeapps (perhaps via the CLI, not relevant) and associate an existing secret with that repo, what should happen if I then update the repo credentials via the Kubeapps UI.

Also wondering, in what case should the error when deleting the secret here not be raised? Why is it not raised here?

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo_auth.go
```diff
@@ -318,96 +319,149 @@ func (s *Server) getRepoTlsConfigAndAuth(ctx context.Context, repo sourcev1.Helm
...
-				}
 			}
 
+			ck := auth.GetTlsCertKey()
+			if ck == nil || ck.Cert == "" || ck.Key == "" {
+				return nil, false, status.Errorf(codes.InvalidArgument, "TLS Cert/Key configuration is missing")
+			}
+			if (ck.Cert == redactedString && !hadSecretTlsCert) || (ck.Key == redactedString && !hadSecretTlsKey) {
+				return nil, false, status.Errorf(codes.InvalidArgument, "Invalid configuration, unexpected REDACTED content")
+			}
```
absoludity, line 394: Thanks for all these checks - ensuring that we'll know if the UX sends the incorrect redacted content (I assume?).

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo_auth.go
```diff
@@ -533,16 +587,32 @@ func newLocalOpaqueSecret(ownerRepo types.NamespacedName) *apiv1.Secret {
...
+	}
+	for server, entry := range config.Auths {
+		docker := &corev1.DockerCredentials{
+			Server:   server,
+			Username: entry.Username,
+			Password: entry.Password,
+			Email:    entry.Email,
+		}
+		return docker, nil
+	}
```
absoludity, line 616: I don't understand why you're using a for loop over the range of `config.Auths` here? It reads as though you're just returning after using the first of `config.Auths` always? But even then, there is no concept of "first" for a map in go (`config.Auths` is a `map[string]DockerConfigEntry` I think?) - the [order when you iterate a map in go is *explicitly* random](https://go.dev/doc/go1#iteration).

dashboard/src/components/Config/PkgRepoList/PkgRepoForm.tsx
```diff
@@ -320,7 +321,7 @@ export function PkgRepoForm(props: IPkgRepoFormProps) {
         finalHeader = authCustomHeader;
         break;
       case PackageRepositoryAuth_PackageRepositoryAuthType.PACKAGE_REPOSITORY_AUTH_TYPE_BEARER:
-        finalHeader = `Bearer ${bearerToken}`;
+        finalHeader = bearerToken;
```
absoludity, line 324: This change is surprising? Was it a bug currently, or why is the `Bearer ` not included now?

absoludity (commented):
Hi Dimitri

> (note: the PR has many small changes, but i was fixing issues as i was testing, hence the list)

Yeah - I don't think I did a good job of reviewing this past the first 400 lines or so. I think it's pretty well established that code reviews degrade as the diff gets bigger beyond a certain point... certainly true for me. If possible in the future, try to push a PR whenever you have a small unit that can be reviewed (ie. 200-400 lines).

Anyway, I've left a few thoughts, but none of it blocking.

Please don't land this until we've cut the release - it's the kind of larger change that could have unseen impact that I'd prefer to have us dogfood for a while in dev before releasing.

Thanks!

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo.go
```diff
@@ -40,6 +40,9 @@ const (
 	// see docs at https://fluxcd.io/docs/components/source/ and
 	// https://fluxcd.io/docs/components/helm/api/
 	fluxHelmRepositories = "helmrepositories"
+
+	// description support
+	Annotation_Description_Key = "kubeapps.dev/description"
```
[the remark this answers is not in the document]
dlaloue-vmware: no particular reason.
this was copied from the carvel plugin, where i introduced them that way - i think the reason is that the code already had constants with underscore though all caps.
i can easily fix that.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo.go
dlaloue-vmware: The idea was that changing the provider did not really make sense (the original code was completely ignoring the new value) so we are explicitly rejecting such change.
The main reason to allow to unset the provider is that the user may have tried with the provider and it did not work and as such is switching back to using basic auth for example.
But yes, a side effect would be that you could unset the provider and switch to another one in the next update.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo.go
```diff
@@ -927,3 +959,19 @@ func newFluxHelmRepo(
...
+		}
+		fluxRepo.Annotations[Annotation_Description_Key] = description
+	} else {
+		delete(fluxRepo.Annotations, Annotation_Description_Key)
+	}
+}
+
+func getDescription(fluxRepo *sourcev1.HelmRepository) string {
+	return fluxRepo.Annotations[Annotation_Description_Key]
+}
```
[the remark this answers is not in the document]
dlaloue-vmware: i don't think we have those functions in a utils file.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo_auth.go
dlaloue-vmware: It is only deleting it because this is a secret created by kubeapps. we are not deleting secrets that kubeapps has not created.
As per the spec, there is a validation that ensures that we are not switching 'mode' between user managed and kubeapps managed secrets. So in this code path, we know this is a kubeapps managed secret.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo_auth.go
dlaloue-vmware: I reused existing code. i.e. that is how Kubeapps has been handling AppRepository image pull secrets and docker config, assuming there was only one even though the struct is a collection.
is there a better way to take one item from a map than this kind of loop?

dashboard/src/components/Config/PkgRepoList/PkgRepoForm.tsx
dlaloue-vmware: i think this was a left-over bug from when the UI was switched to the new API.
The UI code was not changed to comply with what the spec says. The UI was sending the token prefixed with Bearer, and the backend would strip the prefix if present, only (for helm) to add it back before putting it in the secret.
On a read, the backend would always send the token only, without the Bearer prefix (as per the spec). So the UI was inconsistent between its read and write operations.
this is double defensive and inconsistent code that makes things so hard to read and understand, so i cleaned it up.

cmd/kubeapps-apis/plugins/fluxv2/packages/v1alpha1/repo.go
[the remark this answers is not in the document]
absoludity: It's in the kapp_controller/server_utils.go that you've added below (I hadn't noticed earlier that that was specific to kapp_controller). Just wondering if it should be DRY'd up and used just once... perhaps in `plugins/pkg/k8sutils/k8sutils.go`? Fine either way.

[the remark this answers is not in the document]
dlaloue-vmware: I will make the code common, that is the right thing to do.

dlaloue-vmware:
to avoid making even more changes to this PR, is it ok to push the cleanups to another PR?

PR #5771, opened 2022-12-03
Add boilerplate scaffolding for new operators plugin

gfichtenholt:


gfichtenholt:
> > Looks like a great start Greg. My only comment is that I don't reckon we should yet include the `packaging.operators.enabled` option in the README/values until it can be used by users. Can we, for now, put it in the `featureFlags.operatorPluginDev.enabled` or similar? Or at the very least, we should document it as such - perhaps "Enable support for under-development Operators packaging"
> 
> Agree with Michael on this approach.

👍

bayareagreg:
Closed per 'hacking together' discussion on 12/20/2022 due to reassignment to another project

Issue #4920, opened 2022-06-14
Pluggable support for Operators in Kubeapps


ppbaena:
**Summary**
Create a plugin to provide support for operators in Kubeapps

**Background and rationale**
We've been working on plugin-based support for different packaging systems, moving across our initial Helm support as well as adding fluxv2 and Carvel support.

As Kubeapps provides support for operators (not by default), it would be interesting to add this support in a standard, pluggable way.

**Description**
Create a plugin to provide support for operators in Kubeapps, according to the new pluggable architecture and kubeapps-api.

**Acceptance criteria**
A new plugin including the following is available in Kubeapps:

- [ ] Design for the new plugin

- [ ] List available packages
  - [ ]  GetAvailablePackageSummaries
  - [ ]  GetAvailablePackageDetail
  - [ ]  GetAvailablePackageVersions
- [ ]  List installed packages
  - [ ] GetInstalledPackageSummaries
  - [ ]  GetInstalledPackageDetail
- [ ]  Action "buttons"
  - [ ]  CreateInstalledPackage
  - [ ]  UpdateInstalledPackage
  - [ ]  DeleteInstalledPackage
- [ ]  GetPackageRepositories

- [ ] Operators are listed together with other package formats in Kubeapps

- [] End-to-end tests are available for operators

**Additional context**
This is the parent issue with the proposal to provide pluggable support for operators. Any other issue/task created should be linked to this one.

absoludity:
> As Kubeapps provides support for operators (not by default), it would be interesting to add this support in a standard, pluggable way.

Not only interesting: the reason we don't enable the operator support by default is because to support the existing operator functionality, the dashboard relies on the insecure proxying of the k8s API, talking directly with the k8s API server. So enabling it currently exposes the k8s API server, possibly unintentionally by the installer.

ppbaena:
Proposal for the design of the operators plugin added. Moved to backlog to re-prioritize it.

PR #5815, opened 2022-12-20
add link to operator plugin proposal to README

gfichtenholt:
related to #4920


absoludity (approved):
Just note, the other google docs listed there have their sharing access so that any anonymous user can comment, whereas, if you open your new doc in a private browser tab, it has zero access to view.

ppbaena (approved):
Thanks Greg!

gfichtenholt:
> Just note, the other google docs listed there have their sharing access so that any anonymous user can comment, whereas, if you open your new doc in a private browser tab, it has zero access to view.

access updated

PR #5841, opened 2023-01-09
Update k8s version in CI to unblock CI tests

### LINKED ISSUE #5827: CI tests failing for flux integration
**Summary**
A few commits ago, the CI tests for our main branch [began failing](https://github.com/vmware-tanzu/kubeapps/actions/runs/3754612567/jobs/6379065615). The last successful CI test of main was when #5798 landed, syncing the upstream chart changes for the 12.1.3 chart release.

At the time of writing, there are only three commits on main since, all of which failed once landed (though each PR passed before landing).

The flux test fails while waiting for the deployed package to show pods (for the pie chart to be populate):
```
  1) [chromium] › tests/flux/04-default-deployment.spec.js:8:1 › Deploys podinfo package with default values in main cluster 

    page.waitForSelector: Timeout 180000ms exceeded.
    =========================== logs ===========================
    waiting for locator('.application-status-pie-chart-number').locator('text=1') to be visible
    ============================================================

      34 |
      35 |   // Assertions
    > 36 |   await page.waitForSelector("css=.application-status-pie-chart-number >> text=1", {
         |              ^
      37 |     timeout: utils.getDeploymentTimeout(),
      38 |   });
      39 |   await page.waitForSelector("css=.application-status-pie-chart-title >> text=Ready", {

        at /app/tests/flux/04-default-deployment.spec.js:36:14

```

but the screenshot shows that there's nothing available by the timeout:

test-failed-1

(creating the issue to document what's known right now as I won't get to this before the break).

absoludity:

### Description of the change

The main update is to switch to k8s 1.24.7 in our CI (and dev) environments, because CI was running with 1.22 and the podinfo chart that the flux tests use to test is no longer compatible with 1.22.

As a result, we also needed to create specific service token secrets and update the way CI gets those.

There's also some extra logging to help trace issues in the future.

### Benefits

CI passes again.

### Applicable issues

- fixes #5827


absoludity:
Landing this to unblock CI on main (so we can then land all the security bumps).

PR #5842, opened 2023-01-09
Add some fixes for the markdown linter

antgamdia:
### Description of the change

This PR is fixing some pending issues highlighted by the `markdownlint` linter. A first batch of them have been fixed by just using the `--fix` arg, but a bunch of them still required some manual process.

### Benefits

Most of the true positive warnings will be gone; besides, the copyright year of the website will get updated to `2023` :P  (it is generated based upon the last deployment date).

### Possible drawbacks

Don't think so, but let's double-check with the site preview just in case.

### Applicable issues

N/A

### Additional information

N/A


ppbaena (approved):
Thanks for the deep revision @antgamdia!

antgamdia:
Thanks! Merging as it is solely a change in the website and the failing test is just the Flux one (see https://github.com/vmware-tanzu/kubeapps/issues/5827).

PR #5893, opened 2023-01-19
[Bug] Parse the yaml before dumping it back to string

### LINKED ISSUE #5894: [Bug] Installation values formatted incorrectly
**Describe the bug**
The installation values  are not properly formatted in the yaml format in the installation values section

**To Reproduce**

1. Go to Kubeapps UI and open the details page of any release
3. Scroll down to Installation Values

**Expected behavior**
The installation values should be properly formatted in yaml format

**Screenshots**
Before
Before

Created a PR for resolution #5893 

After
After

- Version 2.6.2
- Kubernetes version 1.24.3
- Package version default in kubeapps helm chart

**Additional context**

mittalvaibhav1:
Signed-off-by: Vaibhav Mittal <53215737+mittalvaibhav1@users.noreply.github.com>

### Description of the change
Fix the incorrect formatting of the installation values on the Kubeapps UI

### Benefits
- Cleaner UI 😀 

### Possible drawbacks
- None as far as ik

### Applicable issues

- fixes #5894

### DIFF
--- a/dashboard/src/components/AppView/AppView.tsx
+++ b/dashboard/src/components/AppView/AppView.tsx
@@ -32,7 +32,7 @@ import {
   NotFoundNetworkError,
 } from "shared/types";
 import { getPluginsSupportingRollback } from "shared/utils";
-import { parseToString } from "shared/yamlUtils";
+import { parseToString, parseToJS } from "shared/yamlUtils";
 import ApplicationStatus from "../../containers/ApplicationStatusContainer";
 import * as url from "../../shared/url";
 import LoadingWrapper from "../LoadingWrapper/LoadingWrapper";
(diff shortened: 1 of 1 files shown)


absoludity (approved):
As long as CI is happy, +1... thanks again for jumping in with the report and fix.

mittalvaibhav1:
Good to go :)

absoludity:
Woo!

absoludity:
Just want to double-check: @antgamdia : I'm assuming there wasn't a reason you didn't use `parseToJS` here was there? (I'm 99% sure you would have left a comment if that were the case, so it's probably just something that was missed, but want to be sure).

antgamdia:
> if that were the case, so it's probably just something that was missed, but want to be sure).

From memory, I can't recall why I didn't use it; I mean, when performing IRL tests we should have noticed it, but we didn't. So, either I actually missed it (most likely, I guess 😆 ) or the latest changes on the custom values have had a sort of side effect here.
Anyway, let's keep an eye on the whole process of displaying/editing values just in case.

PR #5913, opened 2023-01-25
Update the cache used by pinniped proxy to include the k8s api server url.

### LINKED ISSUE #5912: Pinniped Proxy is caching TokenCredentialRequests across clusters
**Describe the bug**
While investigating #5755 I've found another issue which contributes to the problem, in that our pinniped-proxy service is caching a little too heavily:

I'd setup the cache to cache the `TokenCredentialRequest` resources - so when a request is received, we construct the `TokenCredentialRequest` ready to exchange for a cert with pinniped, and if an identical `TokenCredentialRequest` exists in the cache, that will be returned immediately instead.

What I had not noticed is that the `TokenCredentialRequest` does not mention or have any info about which cluster, so if the OIDC token is the same, the TCR will be the same for different clusters.

This leads to the user being logged out when attempting to switch clusters.

**To Reproduce**

Exactly what I had [described for #5755](https://github.com/vmware-tanzu/kubeapps/issues/5755#issuecomment-1401297272). I'll update the steps for that separate issue once this fix is in. 

**Expected behavior**
The cache should not be used when switching clusters so that a cert for that cluster is fetched.

absoludity:

### Description of the change

Fixes #5912 by including the k8s server url in the cache key.

### Benefits

Fixes an issue that was causing logouts when switching clusters.

### Applicable issues

- fixes #5912

### DIFF
--- a/cmd/pinniped-proxy/src/pinniped.rs
+++ b/cmd/pinniped-proxy/src/pinniped.rs
@@ -115,7 +115,10 @@ pub struct TokenCredentialRequestStatus {
 ///
 /// The cache takes care of pruning expired tokens whenever a write lock is
 /// acquired to set a new value.
-pub type CredentialCache = Arc<PruningCache<TokenCredentialRequest, TokenCredentialRequest>>;
+/// The Cache key is a 2-tuple of the api endpoint url and the
+/// TokenCredentialRequest.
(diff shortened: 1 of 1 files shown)


absoludity:
Verified locally that this fixes one of the errors when authenticating with another cluster.

PR #5943, opened 2023-02-01
Add test action to reproduce issue in normal PR.

absoludity:

### Description of the change

I used this PR to debug why the installation of the GPG key was failing. The process I used was:
1. list the files written to disk in the `installGPGKeys` command (`ls -al /tmp/*.key`) which showed that they had the incorrect size,
2. print out the `head -n1` of each (which is just the `-----BEGIN PGP PRIVATE KEY BLOCK-----`) (had to also base64 it to avoid it being correctly redacted as it should).

This showed clearly that, since I updated the secrets, the GPG data was in fact not being written to disk. After some experimenting, and reading the [GH documentation for secrets in actions](https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow) I realised that the issue could be because these secrets are being passed unquoted to the `installGPGKeys` bash function.

Rather than correctly quoting them, I instead removed these cmd-line args from the function so that the env vars can be used more safely (as mentioned in the linked GH doc above, including secrets as cmdline args should be avoided anyway).

This fixed the problem in my tests. Now to see if it's the only issue in CI :)

### Applicable issues

- fixes #

### DIFF
--- a/.github/workflows/kubeapps-general.yaml
+++ b/.github/workflows/kubeapps-general.yaml
@@ -605,7 +605,7 @@ jobs:
 
           installGithubCLI ${GITHUB_VERSION}
           installSemver ${SEMVER_VERSION}
-          installGPGKey ${GPG_KEY_PUBLIC} ${GPG_KEY_PRIVATE} ${CI_BOT_GPG} ${CI_BOT_EMAIL}
+          installGPGKey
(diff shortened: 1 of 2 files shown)


absoludity:
Landing to test if it makes main green (could still be an issue with the actual cred).

PR #5950, opened 2023-02-02
Part one of updates for release.

absoludity:

### Description of the change

Just following up to Dashboard Dependencies in the release process.

I've a question about one spot in the release process (which I've removed because it does not appear to be true, but would like a second pair of eyes, even though I'll land this for now).

### Applicable issues

- fixes #


site/content/docs/latest/reference/developer/release-process.md
```diff
@@ -58,7 +58,6 @@ The versions used there _must_ match the ones used for building the container im
 
 Besides, the `GKE_STABLE_VERSION` and the `GKE_REGULAR_VERSION` might have to be updated if the _Stable_ and _Regular_ Kubernetes versions in GKE have changed. Check this information on [this GKE release notes website](https://cloud.google.com/kubernetes-engine/docs/release-notes).
 
-> **NOTE**: at least one of those `GKE_STABLE_VERSION` or `GKE_REGULAR_VERSION` versions _must_ match the Kubernetes-related dependencies in [Go](https://github.com/vmware-tanzu/kubeapps/blob/main/go.mod) and [Rust](https://github.com/vmware-tanzu/kubeapps/blob/main/cmd/pinniped-proxy/Cargo.toml).
```
absoludity, line 61: So I removed this line in the instructions, only because I saw that the existing values (before this PR) for `GKE_STABLE_VERSION` and `GKE_REGULAR_VERSION` were 1.22 and 1.23 respectively, even though the k8s related dependencies were already on 1.25 before @antgamdia [updated them to 1.26 recently](https://github.com/vmware-tanzu/kubeapps/pull/5938/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6). So I assume this statement is not true (as in, from that commit, it seems we don't want to be held to the GKE regular version for our go/rust deps). Or I could update it to a `_should_` rather than `_must_` statement :P

antgamdia: Yep... we wrote that at the time the shift from `v1beta` -> `v1` was happening for `CronJobs`. Updating to a version without it resulting in a compile error in our code, as we were still supporting the old one.
Given that for the rest of the API groups are selected from the Discovery client and CronJobs have reached a stable version, it's safe to no longer observe that rule.

Besides, some packages are usually bumping their k8s deps and, on some occasions (albeit unlikely), they ran into a conflict with ours.

Thanks for noticing and removing the statement!

Issue #6013, opened 2023-02-22
Replace improbable-eng/grpc-web with grpc/web


absoludity:
**Summary**
Improbable stopped supporting grpc-web a while back since the official repo was launched.

We should update to the official [grpc-web](https://github.com/grpc/grpc-web) version. Unfortunately it's not that simple though, since the official grpc-web version has the client-side implementation, but does not have the server-side in-process grpc-Web -> grpc translation (at least for go), instead utilising envoy to do the translation.

We may need to investigate depending on envoy for the translation.

absoludity:
So bufbuild's [`connect` library](https://github.com/bufbuild/connect-es) is now at a 0.8.1 release, over 6 months after their [initial announcement](https://buf.build/blog/connect-web-protobuf-grpc-in-the-browser) and has some features that would be very nice for us:
- they fully support the gRPC-Web protocol (and have an [extended version of gRPC's own compatibility](https://github.com/bufbuild/connect-crosstest) suite to ensure it remains that way)
- Switching your frontend (typescript) code between the gRPC-Web protocol and their own connect-web (now connect-es) protocol is a single line (different constructor - no change to the client usage), which gives you
- a go library - connect-go which does what improbable-web's go library gave us: translation from the web protocol to native gRPC, so no need for envoy or something in between

This would allow us to transition the code as follows:
1. Generate the typescript clients using buf and connect
2. Update our dashboard code to use the new clients but still with gRPC-Web protocol, which should continue to work with the existing improbable-eng backend translation
3. Switch the backend to use `connect-go` and update the dashboard client code (just changing the client constructor)

If we ever change our minds and want to use envoy + gRPC-Web, it's a single-line change to switch back, so no lock-in.

I'm investigating this as the way forward, looks much nicer to use as well, appears to be much more active, better documented and has the same Apache 2.0 license.

absoludity:
I've been investigating switching the backend over and realised that our use of grpc-gateway, to provide a ReST-ish version of the API (in addition to the grpc and grpc-web), won't be possible, since it is supported by the generated code of the  tied with the standard gRPC registration of services.

We don't use this outside of a liveness check (which we should replace with a grpc check anyway), but we have advertised it as an optional API (which could change at any time), and do currently use it to generate the openapi spec.

This has been discussed on the connect-go project: https://github.com/bufbuild/connect-go/discussions/256 . We could potentially continue to generate the openapi spec by updating the http annotations to match the connect http.

[...]

PR #6044, opened 2023-03-03
Switch Dashboard to use new gRPC client throughout.

absoludity:
### Description of the change

This PR switches the dashboard over to use the new connect generated grpc-web client and associated generated classes.

It was pretty tedious, with some complexity due to the different generated classes and client.

## Non-tedious actual changes
- Updating the way the proto one-of (single choice for different values) work throughout to match the new library (which actually ensures there is only a single value, actually using a `OneOf` concept, whereas the old improbable library just allowed all the options at once). This was quite confusing to change in the actual code functionality as the code is written as if all options may be specified, but I didn't want to change that in this PR, just switch.
- Updating the `shared/ResourceRef` class to be a child of the generated `ResourceRef` class so that we can continue to use it in place of the generated one as we're currently doing (compiler won't accept it otherwise now). Created #6062 to track removing our shared one if we can.
- Updating the method used in the frontend to stream resources (and changes) for an installed package, since improbable used subscriptions while the connect one uses async iterators (need to verify this works, since I don't think we test it in CI, but it should just fall back to polling if it doesn't).
- Update the `KubeappsGRPCClient` to use the new client (and the different way of setting headers/metadata).

## Tedious parts

- Any object that is generated from the proto messages now needs to be instantiated with, for example, `new AvailablePackageSummary({...})` to pass the compilation, rather than the previous `{...} as AvailablePackageSummary` as this type assertion fails due to the missing methods.
- Updating all the generated class imports with `_pb` to import the new versions (without deleting the old in this PR, as that blows it out by 16k lines).
- Updating all the enums to the more sensible names used by the new lib
- Update a lot of tests that used jest to spy on the old implementation.
- Update all API method names since the connect one uses snake-case with an initial lower case :/

### Benefits

Let's see what CI says, but possibly removing 16k lines of the old generated code (I'll do that in a follow-up PR). The main benefit is that we'll be using a supported grpc-web typescript client and can remove the old improbable generation in the kubeapps apis service.

### Possible drawbacks

It's a lot of changes, there maybe unexpected side-effects.

### Applicable issues

- ref #6013


absoludity:
I've got the image compiling now, but just need to sort out the slightly different behavior for our login checks, which is the cause of the e2e failures.

absoludity:
> I've got the image compiling now, but haven't yet sorted out the slightly different behavior for our login checks, which is the cause of the e2e failures.

Turned out to be a simple difference in the way headers are extracted from a grpc-web response. Locally I checked everything works, let's see what CI thinks.

absoludity:
> > I've got the image compiling now, but haven't yet sorted out the slightly different behavior for our login checks, which is the cause of the e2e failures.
> 
> Turned out to be a simple difference in the way headers are extracted from a grpc-web response. Locally I checked everything works, including the bits that had changed (long-running requests), let's see what CI thinks.

This is where I miss team-mates to celebrate with :P. After a lot of debugging, found what is hopefully the final CI issue - simply that the incoming error was not being converted correctly because the new grpc error does not have a `headerMap` like the old. Should have grepped for `headerMap` much earlier after changing the code in some places.

PR #6097, opened 2023-03-16
Transition core plugin and grpc health status as example.

absoludity:
### Description of the change

Transitions the core plugins gRPC service to use connect's grpc implementation. This does so in a way that I can transition services one by one without breaking main.

### Benefits

Enables transitioning gRPC services (and their gRPC-Web support) to the new library, given that the improbable one is no longer maintained.

### Possible drawbacks

This doesn't work with the grpc-gateway which provided a ReSTish version of the API which was useful for testing, but we don't actually use anywhere except for the readiness/liveness checks (and we provide no API guarantees for kubeapps-apis).

I've updated the readiness/liveness checks to use the standard grpc-health check instead (we can't yet use the built-in beta grpc health check, as it's only available for 1.23+).

### Applicable issues

- fixes #


absoludity:
Finally was able to get this working, but am going to split it out into reasonable chunks and test properly.

absoludity:
Closing as change was refactored and added in other PRs.

PR #6312, opened 2023-06-15
Ensure Kubeapps-APIs sends correct response codes. Fixes #6269

### LINKED ISSUE #6269: Auth loop when no admin permissions set on default namespace
Hi,

to resolve issue #6223 , we updated Kubeapps from version 12.2.10 to version 12.4.2.

We use ADFS to authenticate on Kubeapps and unfortunately after the update, we were unable to access the portal listing the applications already deployed. We systematically return to the page which asks us to connect via OIDC.

We found a workaround by creating a RoleBinding, we tried to use the cluster role **reader** but without success
```
kind: RoleBinding
metadata:
  name: ns-admin
  namespace: default
subjects:
      kind: User
      name: <user@email.com>
roleRef:
  kind: ClusterRole
  name: admin
```
I specify we confine users to prevent them from deploying applications elsewhere than in their namespace. We find that this behavior will generate security issues and lead to misunderstandings on the user side.

Is there a solution to solve this problem appeared with the last versions?

Regards

absoludity:
### Description of the change

When we switched to connect-grpc, we continued sending standard gRPC error codes from Kubeapps-APIs, and without more information, the [connect-gRPC machinery was converting these to code unknown](https://connect.build/docs/go/errors/#working-with-errors).

When we check that a user has authenticated successfully by getting the default namespace after the login dance, either an OK or a forbidden means the request was authenticated, unauthenticated means it was not. But instead the error was coming back with an unknown code and so Kubeapps was not recognising this as authenticated. This was affecting logins for users who do not have permission to get the default namespace (see #6269 for the workaround).

I reproduced this locally (though with token auth) and so before this change, the gRPC request to CheckNamespaceExists came bace with what *looks* like a permission denied response - but the [grpc-status code is 2 - which is unknown](https://grpc.github.io/grpc/core/md_doc_statuscodes.html):

```
grpc-message | rpc  error: code = PermissionDenied desc = Forbidden to get the Namespace  'default' due to 'namespaces "default" is forbidden: User  "system:serviceaccount:kubeapps:kubeapps-view" cannot get resource  "namespaces" in API group "" in the namespace "default"'
grpc-status | 2 <----- this is "Unknown"
```

After this change, the error arrives as a PermissionDenied as expected:
```
grpc-message | Forbidden  to get the Namespace 'default' due to 'namespaces "default" is  forbidden: User "system:serviceaccount:kubeapps:kubeapps-view" cannot  get resource "namespaces" in API group "" in the namespace "default"'
grpc-status | 7 <----- this is the "PermissionDenied" code
```
which Kubeapps happily recognises and assumes the user is now authenticated.

### Benefits

People can login without requiring RBAC to get the default namespace.

### Possible drawbacks
None

### Applicable issues

- fixes #6269 

### Additional information

Although I was glad to find the fix here, I'm still uncertain why our CI integration tests did not pick this up, given that the intention is that we check login with an unprivileged user. I need to investigate that further.

### DIFF
--- a/cmd/kubeapps-apis/plugins/pkg/statuserror/statuserror.go
+++ b/cmd/kubeapps-apis/plugins/pkg/statuserror/statuserror.go
@@ -4,8 +4,9 @@
 package statuserror
 
 import (
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
+	"fmt"
+
+	"github.com/bufbuild/connect-go"
 	"k8s.io/apimachinery/pkg/api/errors"
 )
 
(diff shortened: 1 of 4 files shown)


absoludity:
OK, it turns out I'll need to change a bunch more endpoints so that they work similarly, even though they don't affect the authn.

absoludity:
I'm closing this one to take a different approach, as changing `statuserror.FromK8sError` affects too many call-sites that I can't go through and fix easily all at once. Instead, I'll create a new fn and then move different parts of the code over then remove the old.

PR #6319, opened 2023-06-19
Switch CheckNamespaceExist to return a connect error as expected by client

### LINKED ISSUE #6269: Auth loop when no admin permissions set on default namespace

absoludity:
### Description of the change

When we switched to connect-grpc, we continued sending standard gRPC error codes from Kubeapps-APIs, and without more information, the [connect-gRPC machinery was converting these to code unknown](https://connect.build/docs/go/errors/#working-with-errors).

When we check that a user has authenticated successfully by getting the default namespace after the login dance, either an OK or a forbidden means the request was authenticated, unauthenticated means it was not. But instead the error was coming back with an unknown code and so Kubeapps was not recognising this as authenticated. This was affecting logins for users who do not have permission to get the default namespace (see #6269 for the workaround).

I reproduced this locally (though with token auth) and so before this change, the gRPC request to CheckNamespaceExists came bace with what *looks* like a permission denied response - but the [grpc-status code is 2 - which is unknown](https://grpc.github.io/grpc/core/md_doc_statuscodes.html):

```
grpc-message | rpc  error: code = PermissionDenied desc = Forbidden to get the Namespace  'default' due to 'namespaces "default" is forbidden: User  "system:serviceaccount:kubeapps:kubeapps-view" cannot get resource  "namespaces" in API group "" in the namespace "default"'
grpc-status | 2 <----- this is "Unknown"
```

After this change, the error arrives as a PermissionDenied as expected:
```
grpc-message | Forbidden  to get the Namespace 'default' due to 'namespaces "default" is  forbidden: User "system:serviceaccount:kubeapps:kubeapps-view" cannot  get resource "namespaces" in API group "" in the namespace "default"'
grpc-status | 7 <----- this is the "PermissionDenied" code
```
which Kubeapps happily recognises and assumes the user is now authenticated.

### Benefits

People can login without requiring RBAC to get the default namespace.

### Possible drawbacks
None

### Applicable issues

- fixes #6269 

### Additional information

Although I was glad to find the fix here, I'm still uncertain why our CI integration tests did not pick this up, given that the intention is that we check login with an unprivileged user. I need to investigate that further.

Also, this is a second attempt at fixing this after #6312 ended up touching too many call-sites. I'll then followup switching other call-sites and removing the old code.

### DIFF
--- a/cmd/kubeapps-apis/plugins/pkg/connecterror/connecterror.go
+++ b/cmd/kubeapps-apis/plugins/pkg/connecterror/connecterror.go
@@ -1,0 +1,29 @@
+// Copyright 2023 the Kubeapps contributors.
+// SPDX-License-Identifier: Apache-2.0
+
+package connecterror
+
+import (
+	"fmt"
+
+	"github.com/bufbuild/connect-go"
+	"k8s.io/apimachinery/pkg/api/errors"
+)
+
+// FromK8sResourceError generates a grpc status error from a Kubernetes error
+// when querying a resource.
+func FromK8sError(verb, resource, identifier string, err error) error {
+	if identifier == "" {
+		identifier = "all"
+	}
+	if errors.IsNotFound(err) {
+		return connect.NewError(connect.CodeNotFound, fmt.Errorf("unable to %s the %s '%s' due to '%w'", verb, resource, identifier, err))
+	} else if errors.IsForbidden(err) {
+		return connect.NewError(connect.CodePermissionDenied, fmt.Errorf("Forbidden to %s the %s '%s' due to '%w'", verb, resource, identifier, err))
+	} else if errors.IsUnauthorized(err) {
+		return connect.NewError(connect.CodeUnauthenticated, fmt.Errorf("Authorization required to %s the %s '%s' due to '%w'", verb, resource, identifier, err))
+	} else if errors.IsAlreadyExists(err) {
+		return connect.NewError(connect.CodeAlreadyExists, fmt.Errorf("Cannot %s the %s '%s' due to '%w' as it already exists", verb, resource, identifier, err))
+	}
+	return connect.NewError(connect.CodeInternal, fmt.Errorf("unable to %s the %s '%s' due to '%w'", verb, resource, identifier, err))
+}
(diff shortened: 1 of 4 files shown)


RGPosadas:
Hi @absoludity , thanks for the quick work on this! 🚀 
When would you expect the fix to become available as a helm chart?

absoludity:
No problem - it's great to be fixing things people need :) I'm waiting for the development image to be published so you can verify the fix, then after than @ppbaena and the team will chat about the next release.

absoludity:
@RGPosadas Can you please test the fix by running with this [kubeapps-apis image sha](https://hub.docker.com/layers/kubeapps/kubeapps-apis/latest/images/sha256-b823598c7c780f902b26bcd5fcbaf3393b8f65aa469ad900f6d5b55da1c85311?context=explore), and verify whether the issue is fixed or continues? Thanks.

RGPosadas:
@absoludity Just confirmed and the auth loop is no longer blocking authenticated users. I am still getting correct 403s on the default namespace (and other non-authorized resources), and getting valid 2XX for the correct resources

absoludity:
Great, thanks for the confirmation. I'll chat with Pepe about a release soon :)

RGPosadas:
Sounds good!

absoludity:
Hi @RGPosadas , we chatted last night about this. I'll aim to finish the related work (switching over all the other call-sites that are sending the old encoded errors, even though they don't cause problems like this call-site did) and release once that work is done, hopefully later next week.

RGPosadas:
Gotcha, thanks for the update. Will keep an eye out for the next 2 weeks 🚀

PR #6322, opened 2023-06-19
Add T-system as a Kubeapps adopter

ppbaena:
### Description of the change

New use case by T-system added to the list of Kubeapps adopters

### Benefits

Information to community users about corporations using Kubeapps and their use cases.

### Possible drawbacks

N/A


absoludity (approved):
Excellent, thanks @ppbaena

Issue #6263, opened 2023-05-18
Implementation of OCI-Catalog gRPC micro-service


absoludity:
**Summary**

Enable Kubeapps to display a catalog of apps for an OCI registry or a namespace of an OCI registry. The proposed implementation is for a stateless gRPC micro-service that can be run (though is not restricted to run) as a side-car of existing the asset-syncer job to provide lists of repositories for a (namespaced) registry, regardless of the registry provider.

**Background and rationale**

See #6179 for the problem/investigation as well as the[ Kubeapps OCI-Catalog proposal](https://docs.google.com/document/d/1uP4cLpdWCQaYbfuIMS8z1DRd1EK7qvXeMXWgGE2oCYA/edit#) for more information.

**Description**
This issue tracks the implementation of the functionality. I'm listing the tasks here, which we may expand into subtasks as needed, but initially having them here is fine.

#### Implement OCI Catalog service
- [x] Create gRPC service with stubbed/unimplemented that serves the two RPC calls `ListRepositoriesForRegistry` and `ListTagsForRepository` 
- [x] Create Dockerfile and update Makefile with required infrastructure changes to include the container in dev.
- [x] Start request for Bitnami official image early (can take time)
- [x] Setup strategy pattern/abstraction to enable different provider implementations to be used, based on request.
- [x] Implement `ListRepositoriesForRegistry` and `ListTagsForRepository` for dockerhub (at this point, the integration tasks below can begin too)

#### Integrate OCI Catalog service
- [x] PR adding OCI-catalog side-car for existing sync job in development
- [x] Update asset-syncer so that AppRepositories for OCI registries *without* list of repositories will use the sidecar gRPC call to obtain the charts for the cache (we may not need a feature flag in this case).
-  ~~Switch existing code for obtaining tags to use the gRPC call instead.~~
-  ~~Additional changes to the add AppRepository form functionality (we may want to adjust the filter so that it operates as a filter only rather than specification of available repos). I suspect we may require other small changes too (to hint which provider should be used for ambiguous cases, or provide extra endpoint).~~
-  ~~Ensure integration for single Kubeapps installation with one dockerhub OCI registry (bitnamicharts namespace).~~

**Acceptance criteria**

See the last integration task. When the catalog lists the available apps correctly from all three OCI registries then this task is considered finished.

absoludity:
[...]

PR #6344, opened 2023-06-29
Update the asset-syncer to fetch OCI catalog when present.

absoludity:
### Description of the change

Part 3/3 for experimental support of VAC indexes.

### Benefits

Potentially quick win to display VAC OCI repos.

### Applicable issues

- ref #6263

### Additional information

Still need to test locally.


absoludity:
In the end I do need to update a dashboard form requirement, so will land this and do that separately, testing the whole process.

PR #6525, opened 2023-07-26
Changed baseUrl to use relative path to support URL subpath

### LINKED ISSUE #6314: Path hosted kubeapps behind Ingress provided by nginx on AKS
**Summary**
My aim is to set up a demonstrator taking Kubeapps, deploying behind an nginx ingress controller on AKS with OIDC provided by Keycloak which is running on the same cluster in its own Namespace.  The intention will be to show authentication and authorisation based on Keycloak.  However, after Authentication in Keycloak and redirection back to kubeapps I am faced with:-
"Unable to load kubeapps configuration".

**Background and rationale**
I am providing a single Ingress Proxy for the cluster and given kubeapps Chart and the Keycloak charts I have two paths the route to the backend services:-
https://MyDomainName/Apps/  this routes to cubes
https://MyDomainName/Auth/ this routes to the Keycloak instance

**Description**
I have created a TLS end point and have used your chart (v 12.4.*) to define a path.  When I launch KubeApps at this defined path I am correctly directed to the OAuth2-Proxy which in turn correctly sends me to my KeyCloak authorisation end point.  On return I get "Unable to load Kubeapps configuration:" .  When I look at the network trace in Chrome I see the last call with a non-200 response is a 404.  The client is attempting to POST to https://MydomainName/apis/kubeappsapis.core.plugins.v1alpha1.PluginsService/GetConfiguredPlugins.

Given the Path I have set in the chart it is no surprise that this URL is returning a 404,  I would have thought I would be expecting the API call made to be:-
https://MyDomainName**/apps/**apis/kubeappsapis.core.plugins.v1alpha1.PluginsService/GetConfiguredPlugins 

This feels like I am missing something fundamental, or perhaps this 404 is not related to the Error.

**Additional context**
I have spent quite a few days looking at this and have of course searched through your documentation but have failed to come up with a match.  Any suggestions would be appreciated and of course I can provide more information at your request.

jcatana:
### Description of the change
Fixes static `baseUrl: /...` to be relative `baseUrl: ./...`

### Benefits
Restores previous functionality of working behind url subpath aka ingress

### Possible drawbacks
Unknown

### Applicable issues

- fixes #6314

### DIFF
--- a/dashboard/src/shared/KubeappsGrpcClient.ts
+++ b/dashboard/src/shared/KubeappsGrpcClient.ts
@@ -36,7 +36,7 @@ export class KubeappsGrpcClient {
       return await next(req);
     };
     this.transport = createGrpcWebTransport({
-      baseUrl: `/${URL.api.kubeappsapis}`,
+      baseUrl: `./${URL.api.kubeappsapis}`,
       interceptors: [auth],
     });
   }


absoludity:
Thanks @jcatana . I'm assuming you tested this in real life and verified that it fixes the issue without breaking anything else. Looks like we just switched to an absolute path when upgrading as [mentioned on the issue](https://github.com/vmware-tanzu/kubeapps/issues/6314#issuecomment-1596296558). I'll let the CI tooling run for now, while you get to do the joys of signing the CLA (apologies for the friction to contribute :) )

absoludity:
Looks good @jcatana , once the CLA is done I'll approve and land. Thanks!

jcatana:
Updated. Autocomplete only puts street & number in "address" fields. Maybe a fix to the CLA if you require all that info. Can I make a PR on your CLA form? :rofl:

absoludity:
> Updated. Autocomplete only puts street & number in "address" fields. Maybe a fix to the CLA if you require all that info. Can I make a PR on your CLA form? 🤣

I wish - not something we control. Thanks again!

absoludity (approved):

Issue #6187, opened 2023-04-14
Upgrade to react-router-6


absoludity:
**Summary**
For a while now we've been avoiding the upgrade to react-router 6 (eg #6153), as it will contain a significant chunk of work.

There are tools to help with the migration (so that you can update the code piecemeal (see [upgrading from v5](https://reactrouter.com/en/main/upgrading/v5)), but I'm sure there will also be unknowns along the way.

I've not yet found any clear info about when react router react-router v5 will cease support.

absoludity:
I've done step 1 here, which is [refactoring custom routes](https://reactrouter.com/en/main/upgrading/v5#refactor-custom-routes).

absoludity:
> I've done step 1 here, which is [refactoring custom routes](https://reactrouter.com/en/main/upgrading/v5#refactor-custom-routes).

Or not... turns out we'll need to convert the Operator container+component classes to function components to be able to use the new route system... so today's task is to do that prequel PR then be able to update the PR from yesterday to use it :/

PR #6557, opened 2023-08-03
Ensure token auth can be configured while oauthLoginURI is defined.

absoludity:
### Description of the change

When switching from a container to a component with hooks, I missed that the prop was passed from the container as:

```
   oauthLoginURI: authProxyEnabled ? oauthLoginURI : "",
```

but with the component the oauthLoginURI was always defined.

### Benefits

Can configure token auth again

### Applicable issues

- ref #6187

### DIFF
--- a/dashboard/src/shared/specs/mountWrapper.tsx
+++ b/dashboard/src/shared/specs/mountWrapper.tsx
@@ -1,7 +1,6 @@
 // Copyright 2020-2023 the Kubeapps contributors.
 // SPDX-License-Identifier: Apache-2.0
 
-import { RouterState } from "connected-react-router";
 import { mount } from "enzyme";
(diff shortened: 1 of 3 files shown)


dashboard/src/shared/specs/mountWrapper.tsx
```diff
@@ -39,6 +38,7 @@ export const initialState = {
     kubeappsNamespace: "kubeapps",
     helmGlobalNamespace: "kubeapps-repos-global",
     carvelGlobalNamespace: "kapp-controller-packaging-global",
+    oauthLoginURI: "/oauth/login",
```
absoludity, line 41: I set this in our default state so to trigger the failures in the tests that should have caught it. Lesson: always use default settings in tests that match your prod default settings.

PR #6554, opened 2023-08-02
Remove old js components: Spinner, Alert

antgamdia:
### Description of the change

This PR is following up #6550 (and requires it to get merged first). In this one, the focus is on:

- Remove the `Spinner` old js component in favor of `CdsProgressCircle`.
- Replace the old js `Alert` with the `AlertGroup` component we already had. It is just a wrapper with some handy configuration of the clarity components.
  - By default, they are now closable. Sometimes we have some transient errors and I think we can just defer to the user the decision on whether or not to close them.
- Add a "Cancel" button in a modal (new ns) to keep UX consistency.
- Replace the remaining plain `<button>` (we weren't able to back in time due to an upstream bug).

Again, this PR is pretty extensive (sorry!)

### Benefits

Again, let me show some visual examples :)  (the error messages like "error.message" are just placeholders while developing, they are removed in the PR).

<details><summary>Initial emptyKubeapps alert</summary>
<p>

alert1_BA

</p>
</details> 
<details><summary>Typical error alert</summary>
<p>

alert2_BA

</p>
</details> 
<details><summary>Error with action (in accordance with the clarity guidelines)</summary>
<p>

alert3_(layout)_BA

</p>
</details> 
<details><summary>Error with action (another example)</summary>
<p>

alert4_(catalog)_BA

</p>
</details> 
<details><summary>New update available</summary>
<p>

alert5_(pkgUpdate)_BA

</p>
</details> 
<details><summary>Operators warning</summary>
<p>

alert6_(operators)_BA

</p>
</details> 
<details><summary>General error (note the proper margins now)</summary>
<p>

alert7_(cloader)_BA

</p>
</details> 
<details><summary>Repos in !kubeappsCluster</summary>
<p>

alert8_(repos)_BA

</p>
</details> 
<details><summary>Add repo error</summary>
<p>

alert9_(newRepo)_BA

</p>
</details> 
<details><summary>Error in a modal (note the margins again)</summary>
<p>

alert10_(rollback)_BA

</p>
</details> 

### Possible drawbacks

N/A (hope any!)

### Applicable issues

Related to https://github.com/vmware-tanzu/kubeapps/issues/6563

### Additional information

After this PR there are just two remaining js components: `Table` and `Card`. I will work on them soon!

PR #6560, opened 2023-08-03
Remove old js components: Card, Table

antgamdia:
## Description of the change

This PR is following up #6554 (and requires it to get merged first). In this final one, the focus is on:

- Transform the `Card` and its sub-components to tsx.
- Transform the `Table` to tsx
- Remove the remaining test snapshots 

### Benefits

No cool before/after pics this time :P  Here we have just migrated js-based code to typescript.
However, after this PR we will able to say we finally got rid of the legacy js-components!! 

### Possible drawbacks

N/A - hopefully.

### Applicable issues

Related to https://github.com/vmware-tanzu/kubeapps/issues/6563

### Additional information

While transforming the `Card` to tsx I also had a look at the `CdsCard` from Clarity. Unfortunately, there is no drop-in replacement; a migration would require plenty of style-related changes. Besides, the "native" style differs a bit from the general style we have in Kubeapps. Below are some examples:

<details><summary>Example from the Clarity website</summary>

image

</details>

<details><summary>Quick/dirty replacement, it does not look good</summary>

image

</details>

<details><summary>Our current custom card</summary>

image

</details>


dashboard/src/components/InfoCard/Card/CardImage/CardImage.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+const CardImage = ({ src, alt }: ICardImageProps) => {
+  return (
+    <div className="card-img">
+      <img src={src} alt={alt} />
+    </div>
+  );
+};
+
+export default CardImage;
```
absoludity, line 17: Almost doesn't seem worth a separate component?

dashboard/src/components/InfoCard/Card/CardText/CardText.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+export interface ICardTextProps {
+  children: React.ReactNode;
+}
+
+const CardText = ({ children }: ICardTextProps) => {
+  return <div className="card-text">{children}</div>;
+};
+
+export default CardText;
```
absoludity, line 14: Same here. Is it a react thing to break the elements up like this? Seems like a `/InfoCard/Card` could just have the text and image without splitting those out into separate components?

absoludity (approved):
Sorry, I'd missed that this one was still waiting for a review.

dashboard/src/components/InfoCard/Card/CardText/CardText.tsx
antgamdia: 100% agree, no actual benefits, especially if we ever plan to use the card from clarity at some point in the future. I thought the same while moving the code to tsx. I'll make a follow-up PR with the changes shortly.

PR #6568, opened 2023-08-04
Ensure only single request for installed apps.

absoludity:
### Description of the change

Fixes an issue I'd noticed while upgrading react-router, in that sometimes the installed apps for the namespace would display, then seconds later, be updated with installed apps across all namespaces.

### Benefits

No longer have this confusing UI behaviour.

### Possible drawbacks

None.

### Applicable issues

- ref #6187

### DIFF
--- a/dashboard/src/components/AppList/AppList.tsx
+++ b/dashboard/src/components/AppList/AppList.tsx
@@ -37,9 +37,8 @@ function AppList() {
   const { currentNamespace } = clusters[cluster];
 
   const [searchFilter, setSearchFilter] = useState("");
-  const [allNS, setAllNS] = useState(false);
+  const [allNS, setAllNS] = useState(allNSQuery === "yes");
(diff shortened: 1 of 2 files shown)


dashboard/src/components/AppList/AppList.tsx
```diff
@@ -111,7 +103,7 @@ function AppList() {
             />
             {canSetAllNS && (
               <CdsToggleGroup className="flex-v-center">
-                <CdsToggle>
+                <CdsToggle data-testid="cds-all-ns">
```
antgamdia, line 106: Why this `testid` ?

antgamdia (approved):
Thanks! Good catch!

dashboard/src/components/AppList/AppList.tsx
absoludity: Argh - I'd added it when struggling to get the test to work, but forgot to remove. Thanks.

PR #6580, opened 2023-08-07
Add DockerHub implementation for send_tags

absoludity:
### Description of the change

Adds the implementation for retrieving and sending tags for dockerhub repositories.

### Benefits

Enables an initial integration to be tested.

### Applicable issues

- ref #6263 

### Additional information

IRL test:

```console
grpcurl -plaintext -proto ./proto/ocicatalog.proto -d '{ "repository": {"registry": "registry-1.docker.io", "namespace": "bitnamicharts", "name": "zookeeper" }}' 0.0.0.0:50001 ocicatalog.OCICatalog.ListTagsForRepository
{
  "name": "11.4.10"
}
{
  "name": "11.4.9"
}
{
  "name": "11.4.8"
}
{
  "name": "11.4.7"
}
{
  "name": "11.4.6"
}
{
  "name": "11.4.5"
}
{
  "name": "11.4.4"
}
{
  "name": "11.4.3"
}
{
  "name": "11.4.2"
}
{
  "name": "11.4.1"
}
{
  "name": "11.3.2"
}
{
  "name": "11.3.1"
}
{
  "name": "11.2.1"
}
{
  "name": "11.1.6"
}
{
  "name": "11.1.5"
}
{
  "name": "11.1.4"
}
{
  "name": "11.1.3"
}
{
  "name": "11.1.2"
}
{
  "name": "11.1.1"
}
{
  "name": "11.0.3"
}
{
  "name": "11.0.2"
}
{
  "name": "11.0.1"
}
{
  "name": "11.0.0"
}
{
  "name": "10.2.5"
}
{
  "name": "10.2.4"
}
{
  "name": "10.2.3"
}
```

### DIFF
--- a/cmd/oci-catalog/src/providers/dockerhub.rs
+++ b/cmd/oci-catalog/src/providers/dockerhub.rs
@@ -12,6 +12,7 @@ use tonic::Status;
 /// The default page size with which requests are sent to docker hub.
 const DEFAULT_PAGE_SIZE: u8 = 100;
 pub const PROVIDER_NAME: &str = "DockerHubAPI";
+pub const DOCKERHUB_URI: &str = "https://hub.docker.com";
 
 #[derive(Serialize, Deserialize)]
 struct DockerHubV2Repository {
@@ -29,6 +30,21 @@ struct DockerHubV2RepositoriesResult {
     results: Vec<DockerHubV2Repository>,
 }
 
+#[derive(Serialize, Deserialize)]
+struct DockerHubV2Tag {
+    name: String,
+    repository_type: Option<String>,
+    content_type: String,
+}
+
+#[derive(Serialize, Deserialize)]
+struct DockerHubV2TagsResult {
+    count: u16,
+    next: Option<String>,
+    previous: Option<String>,
+    results: Vec<DockerHubV2Tag>,
+}
+
 #[derive(Debug, Default)]
 pub struct DockerHubAPI {}
 
@@ -44,7 +60,7 @@ impl OCICatalogSender for DockerHubAPI {
         tx: mpsc::Sender<Result<Repository, Status>>,
         request: &ListRepositoriesRequest,
     ) {
-        let mut url = url_for_request(request);
+        let mut url = url_for_request_repositories(request);
 
         let client = reqwest::Client::builder().build().unwrap();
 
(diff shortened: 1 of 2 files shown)


cmd/oci-catalog/src/providers/dockerhub.rs
```diff
@@ -137,6 +207,30 @@ fn url_for_request(request: &ListRepositoriesRequest) -> Url {
...
+    };
+
+    url.set_path(&format!(
+        "/v2/repositories/{}/{}/tags",
+        repo.namespace, repo.name
+    ));
+
+    // For now we use a default page size.
+    url.query_pairs_mut()
+        .append_pair("page_size", &format!("{}", DEFAULT_PAGE_SIZE));
```
antgamdia, line 229: Just to confirm, are we only fetching (`DEFAULT_PAGE_SIZE=100`) elements for now, aren't we? I mean, there I don't see any pagination logic here. It seems Docker is using a `Link` header to let the client know about the next page ([source](https://docs.docker.com/registry/spec/api/#pagination)), but we are not using it at the moment. No?

cmd/oci-catalog/src/providers/dockerhub.rs
```diff
@@ -105,25 +121,79 @@ impl OCICatalogSender for DockerHubAPI {
...
     }
 
-    async fn send_tags(&self, tx: mpsc::Sender<Result<Tag, Status>>, _request: &ListTagsRequest) {
-        for count in 0..10 {
-            tx.send(Ok(Tag {
-                name: format!("tag-{}", count),
-            }))
-            .await
-            .unwrap();
+    async fn send_tags(&self, tx: mpsc::Sender<Result<Tag, Status>>, request: &ListTagsRequest) {
```
antgamdia, line 124: For what is worth, (I guess you already know it, but just in case), here is the "official" (albeit tagged as experimental) DockerHub client: https://github.com/docker/hub-tool/blob/main/pkg/hub/tags.go#L61

cmd/oci-catalog/src/providers/dockerhub.rs
```diff
@@ -105,25 +121,79 @@ impl OCICatalogSender for DockerHubAPI {
...
+            log::trace!("response body: {}", body);
+
+            let response: DockerHubV2TagsResult = serde_json::from_str(&body).unwrap();
+
+            for tag in response.results {
+                tx.send(Ok(Tag { name: tag.name })).await.unwrap();
+            }
+
+            if response.next.is_some() {
+                url = reqwest::Url::parse(&response.next.unwrap()).unwrap();
```
antgamdia, line 179: Ah - reading it twice, perhaps this is the pagination logic? If so, I would add some comments just for our future ourselves to understand 😅

cmd/oci-catalog/src/providers/dockerhub.rs
```diff
@@ -44,7 +60,7 @@ impl OCICatalogSender for DockerHubAPI {
         tx: mpsc::Sender<Result<Repository, Status>>,
         request: &ListRepositoriesRequest,
     ) {
-        let mut url = url_for_request(request);
+        let mut url = url_for_request_repositories(request);
 
         let client = reqwest::Client::builder().build().unwrap();
```
antgamdia, line 65: Not now, but given the DockerHub rate limits, perhaps it is worthwhile adding sth like: https://crates.io/crates/governor for better handling it on the client side.

antgamdia (approved):
Thanks!

cmd/oci-catalog/src/providers/dockerhub.rs
absoludity: This is us using the pagination of the dockerhub API, yes - requesting the next page if `response.next` is defined. I'll add a comment to that effect.

cmd/oci-catalog/src/providers/dockerhub.rs
absoludity: That's just the current default pagesize for the requests we make to dockerhub (ie. how big are the batches that we request), so will affect performance only (number of requests required to exhaust the result). It's not limiting how many results we get at all (ie. if I set this to 10, then the loop will fetch the results in batches of 10).

The gRPC API that we provide streams the result back, so no pagination required.

cmd/oci-catalog/src/providers/dockerhub.rs
absoludity: Yeah, as you can see in their (go) client, they need to collect all the page results (appending to the array) before returning the result. In our case, we're wanting to stream the results back (via gRPC stream) while they're collected, which is why I'm using these `send_tags` and `send_repositories`, to show the results down a channel (which we can buffer as needed) and continue fetching.

BTW: interesting to see they use the same [default page size](https://github.com/docker/hub-tool/blob/main/pkg/hub/client.go#L43) :)

cmd/oci-catalog/src/providers/dockerhub.rs
absoludity: Yeah, that'd be great to add in the future if we find we're hitting rate limits (I'll be testing this using the bitnami catalog, which is probably on the larger side of OCI helm registries).

cmd/oci-catalog/src/providers/dockerhub.rs
antgamdia: Neat! Thanks for the explanation!

PR #6589, opened 2023-08-08
Improve kubeapps-apis dockerfile and docs

antgamdia:
### Description of the change

Just a minor PR mainly sorting the `ARG`s and adding some comments to make it easier to find the latest release. Also, making the download of golangci-lint conditional.
Finally, adding `GRPC_HEALTH_PROBE_VERSION` to the docs, so that we can find it easily.

### Benefits

More clear dockerfile

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

N/A

### DIFF
--- a/cmd/kubeapps-apis/Dockerfile
+++ b/cmd/kubeapps-apis/Dockerfile
@@ -8,12 +8,27 @@ WORKDIR /go/src/github.com/vmware-tanzu/kubeapps
 COPY go.mod go.sum ./
 ARG VERSION="devel"
 
+# If true, run golangci-lint to detect issues
+ARG lint
+
+# https://github.com/bufbuild/buf/releases/
 ARG BUF_VERSION="1.25.1"
+
+# https://github.com/golangci/golangci-lint/releases
 ARG GOLANGCILINT_VERSION="1.53.3"
 
+# https://github.com/grpc-ecosystem/grpc-health-probe/releases/
+ARG GRPC_HEALTH_PROBE_VERSION="0.4.19"
+
 # Install lint tools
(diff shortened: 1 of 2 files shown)


cmd/kubeapps-apis/Dockerfile
```diff
@@ -8,12 +8,27 @@ WORKDIR /go/src/github.com/vmware-tanzu/kubeapps
...
+
 # Install lint tools
+RUN if [ ! -z "$lint" ]; then \
+        go install github.com/golangci/golangci-lint/cmd/golangci-lint@v$GOLANGCILINT_VERSION; \
+    fi
+
 RUN curl -sSL "https://github.com/bufbuild/buf/releases/download/v$BUF_VERSION/buf-Linux-x86_64" -o "/tmp/buf" && chmod +x "/tmp/buf"
-RUN go install github.com/golangci/golangci-lint/cmd/golangci-lint@v$GOLANGCILINT_VERSION
+
+# TODO: Remove and instead use built-in gRPC container probes once we're supporting >= 1.24 only. https://kubernetes.io/blog/2022/05/13/grpc-probes-now-in-beta/
```
absoludity, line 30: Just noticed GKE_STABLE_VERSION is now 1.25, so we can remove this and update the chart too (not here).

absoludity (approved):

cmd/kubeapps-apis/Dockerfile
antgamdia: Yeah, we could potentially do it, but it can have some unintended side effects if we sync our chart with Bitnami's. AFAIK, there is still some k8s 1.23 testing in place :(

PR #6590, opened 2023-08-08
Avoid endless spinner when no readme is provided

antgamdia:
### Description of the change

When packages don't have any readme file, currently the spinner is never hidden. It gives the impression of something still unloaded, so the user should remain in that page because sth else will be displayed soon.
However, this is not true: once `! isFetching`, if there is no readme... it will never be available.

This PR passes the `isFetching` prop downstream so that the readme renderer can decide what to display.

### Benefits

No more endless spinner when no readme is provided.

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

image

### DIFF
--- a/dashboard/src/components/PackageHeader/PackageReadme.test.tsx
+++ b/dashboard/src/components/PackageHeader/PackageReadme.test.tsx
@@ -6,11 +6,12 @@ import LoadingWrapper from "components/LoadingWrapper/LoadingWrapper";
 import ReactMarkdown from "react-markdown";
 import { HashLink as Link } from "react-router-hash-link";
 import { defaultStore, mountWrapper } from "shared/specs/mountWrapper";
-import PackageReadme from "./PackageReadme";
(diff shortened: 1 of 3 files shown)


absoludity (approved):
Great, thanks Antonio. 

One thing that I was trying to do while doing the recent removal of conatiners, is to actually remove the passing down of state props from our components (except where it makes sense for simple components that aren't using state per-se). In the past, the container would get the state and pass all the props down through the chain to the endpoints. You'll see in the PackageView component that you modified here that it doesn't have any props, it just selects the required state. Anyway, point was not to change this, just be aware than where you think it makes sense, we can do so.

antgamdia:
> One thing that I was trying to do while doing the recent removal of conatiners, is to actually remove the passing down of state props from our components (except where it makes sense for simple components that aren't using state per-se).

Ah, great to know. Thanks for the context here; I'll take it into account for future components

PR #6618, opened 2023-08-14
Adds generated go client for oci-catalog that we'll use from syncer.

absoludity:
### Description of the change

Follows on from #6595, updates the proto service files according to buf's lint, and generates the go client for use in the asset-syncer.

Note: I added a separate `buf.yaml` and `buf.gen.yaml` for the oci-catalog service as buf doesn't let you refer to a proto outside of the `buf.yaml` root, so the other option would have been to move our existing `buf.yaml` and `buf.generate` into the Kubeapps root directory. I don't mind either way, but it made more sense to me for the oci-catalog service to have its own so that it is independent (it may be useful outside of Kubeapps, theoretically).

### Benefits

The next PR can use the client without being cluttered by the extra diff.

### Applicable issues

- ref: #6263


antgamdia (approved):
+1 to the independent `buf.yaml` files. However, I would add this to the release docs (under the _Protobuf dependencies and autogenerated code_ section), mainly just not to forget about this component when upgrading our deps.

absoludity:
> +1 to the independent `buf.yaml` files. However, I would add this to the release docs (under the _Protobuf dependencies and autogenerated code_ section), mainly just not to forget about this component when upgrading our deps.

Thanks. I added two small makefile targets to handle both projects for us and updated the release docs to use that.

PR #6635, opened 2023-08-16
Use OCI catalog service when available for OCI validation

absoludity:
### Description of the change

Follows on from #6620, uses the grpc Client to check if we can find repos for an OCI registry. While there I improved a couple of other things regarding the validation:
- a context.Context can (and must) now be passed through to the validation (we should always pass the request context through subsequent requests); and
- the getValidator function is now a method so that the different validators can have custom data set (such as the OCI catalog address, or the repo client getter).

I've added a basic test double grpc service for the OCI catalog to use here in tests, and will undoubtedly improve this test double when adding the actual sync functionality to use the OCI catalog service next.

### Benefits

OCI Catalog service can be used to check an OCI Registry, without impacting current behaviour.

### Possible drawbacks

None that I'm aware of.

### Applicable issues

- ref #6263

### Additional information

I'll test IRL locally and update here:

Logs showing the fallback when the OCI catalog is not running:
```
E0816 03:49:15.487354       1 repositories_validation.go:354] unable to query OCI Catalog service at ":50061": error querying OCI Catalog for repos: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp :50061: connect: connection refused"
E0816 03:49:16.462758       1 utils.go:400] unable to get catalog manifest: GET request to [https://registry-1.docker.io/v2/bitnamicharts/charts-index/manifests/latest] failed due to status [401]: {"errors":[{"code":"UNAUTHORIZED","message":"authentication required","detail":[{"Type":"repository","Class":"","Name":"bitnamicharts/charts-index","Action":"pull"}]}]}
E0816 03:49:16.462894       1 repositories_validation.go:52] Failed repository validation validation: &{Code:400 Message:unable to determine the OCI catalog, you need to specify at least one repository}
```

and verified I can add the Bitnami catalog using the OCI `https://registry-1.docker.io/bitnamicharts` . That is, it passes validation and gets added. It doesn't yet sync - that'll be the next PR.

### DIFF
--- a/chart/kubeapps/templates/kubeappsapis/deployment.yaml
+++ b/chart/kubeapps/templates/kubeappsapis/deployment.yaml
@@ -166,6 +166,8 @@ spec:
                 secretKeyRef:
                   key: postgres-password
                   name: {{ include "kubeapps.postgresql.secretName" . }}
+            - name: OCI_CATALOG_URL
+              value: {{ printf ":%d" .Values.ociCatalog.containerPorts.http . }}
             {{- end }}
             {{- if .Values.kubeappsapis.extraEnvVars }}
             {{- include "common.tplvalues.render" (dict "value" .Values.kubeappsapis.extraEnvVars "context" $) | nindent 12 }}
--- a/chart/kubeapps/templates/kubeappsapis/deployment.yaml
+++ b/chart/kubeapps/templates/kubeappsapis/deployment.yaml
@@ -167,7 +167,7 @@ spec:
                   key: postgres-password
                   name: {{ include "kubeapps.postgresql.secretName" . }}
             - name: OCI_CATALOG_URL
-              value: {{ printf ":%d" .Values.ociCatalog.containerPorts.http . }}
+              value: {{ printf ":%d" (int .Values.ociCatalog.containerPorts.http) | quote }}
             {{- end }}
             {{- if .Values.kubeappsapis.extraEnvVars }}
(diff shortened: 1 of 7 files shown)


chart/kubeapps/templates/kubeappsapis/deployment.yaml
```diff
@@ -166,6 +166,8 @@ spec:
                 secretKeyRef:
                   key: postgres-password
                   name: {{ include "kubeapps.postgresql.secretName" . }}
+            - name: OCI_CATALOG_URL
+              value: {{ printf ":%d" (int .Values.ociCatalog.containerPorts.http) | quote }}
```
antgamdia, line 170: Haven't checked, but I assume `containerPorts.http` is actually pointing to the grpc endpoint, no?

antgamdia (approved):
Thanks!

chart/kubeapps/templates/kubeappsapis/deployment.yaml
absoludity: Yes, there is only one endpoint. Perhaps we should name it grpc instead... (I think the container template I used to create the ocicatalog helm section used http).

absoludity: Done.

antgamdia: Thanks! Much more descriptive this way

PR #6645, opened 2023-08-17
Allow `oci://` protocol to be specified for OCI AppRepositories

### LINKED ISSUE #6427: Support `oci://` as the protocol when adding an OCI registry/namespace source
**Summary**

Support specifying an OCI registry using the `oci://` protocol.

**Background and rationale**

Currently when adding an OCI registry+namespace as a source for Helm charts, you need to add it with the http protocol, such as "https://my.registry.example.com/namespace" (see [Kubeapps with VAC](https://docs.vmware.com/en/VMware-Application-Catalog/services/main/GUID-using-consume-tanzu-advanced-kubeapps.html#step-1-add-the-vmware-application-catalog-for-tanzu-advanced-to-kubeapps-1) docs) - which as a developer *might* make sense since you're really asking Kubeapps to talk to the http API, but would be much more sensible to support "oci://my.registry.example.com/namespace". Multiple people (myself included) have tried that first, given their exposure to using OCI in Helm I guess.

**Description**

Functionality should be otherwise the same when adding the registry.

**Acceptance criteria**

An OCI registry + namespace can be added using the `oci://` protocol (and still works with `https://`).

**Additional context**

absoludity:
### Description of the change

When OCI support was added originally, we required people to specify an http/https URL for the registry, even though Helm itself allows oci://, because we're (Kubeapps) actually interacting with the OCI distribution spec in that code.

In addition, there was an issue in the dashboard that even if you did specify `oci://registry-1.docker.io/foo` the dashboard would assume the URI was missing a protocol since it doesn't start with `http` and add it, resulting in `https://oci://registry-1.docker.io/foo`.

This PR fixes the dashboard issue (so you can specify an oci protocol), updates the validation in the kubeapps-apis to use https in that case, and the sync code to do the same.

- [x] TODO: check that the oci catalog handles it as well. (all good - the grpc API only takes the host and namespace etc.

### Benefits

Users can specify the OCI registry the way they would with Helm or other tools (this actually tripped up someone trying it out recently).

### Possible drawbacks

Bugs?

### Applicable issues

- fixes #6427

### DIFF
--- a/cmd/asset-syncer/server/utils.go
+++ b/cmd/asset-syncer/server/utils.go
@@ -763,6 +763,13 @@ func getOCIRepo(namespace, name, repoURL, authorizationHeader string, filter *ap
 		log.Errorf("failed to parse URL, url=%s: %v", repoURL, err)
 		return nil, err
 	}
+
+	// If the AppRepo has the URL specified as `oci://` then replace it with
+	// https for talking with the API. If people are using non-https OCI
+	// registries (?!) then they can specify the URL with http.
+	if url.Scheme == "oci" {
+		url.Scheme = "https"
+	}
 	headers := http.Header{}
 	if authorizationHeader != "" {
 		headers["Authorization"] = []string{authorizationHeader}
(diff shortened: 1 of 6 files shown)


antgamdia (approved):
Thanks for the fix!

I just remembered we had an open issue partially related to this code (the repo validation, in general). I'm referencing it just in case we wanted to have a look at: https://github.com/vmware-tanzu/kubeapps/issues/5781

PR #6647, opened 2023-08-17
Use finalizers to spin up AppRepo clean-up jobs

### LINKED ISSUE #6545: Cannot configure securityContext of sync and cleanup jobs

antgamdia:
### Description of the change

Even if the sync jobs were added a security context (by means of each AppRepo CRD), this information was not available for Cleanup jobs. This is mainly due to the fact that those jobs are spun up once a NotFound error is thrown when fetching an AppRepo. 
However, Kubernetes does have a native approach for dealing with these scenarios: finalizers.

In https://github.com/vmware-tanzu/kubeapps/pull/6605 we proposed a simplistic workaround based on adding more params to the controller... but as suggested in https://github.com/vmware-tanzu/kubeapps/pull/6605#issuecomment-1678268807, moving to finalizers is a better long-term solution.

### Benefits

Cleanup jobs are now handled within an existing AppRepo context... meaning we have all the syncJobTemplate available to be used (ie, securityPolicies and so on).

### Possible drawbacks

When dealing with finalizers in the past I often found it really annoying when they get stuck and prevent the resource to get deleted. I wonder if we should add some info in the FAQ on how to manually remove the finalizers.

Additionally, and this might be something important: for the AppRepo controller to be able to `update` AppRepos in other namespaces != kubeapps.... (to add the finalizer) it now needs extra RBAC. Before we were just granting `...-appprepos-read`.. but now we would need to grant `...-write` as well...and I'm not sure we really want to do so. 
WDYT, @absoludity ?
Another idea is using an admission policy... but not sure again if we want to deal with that...

~(I haven't modified the RBAC yet in this PR)~ Changes have been performed finally

### Applicable issues

- fixes #6545

### Additional information

This PR is based on top of https://github.com/vmware-tanzu/kubeapps/pull/6646, but the main change to review is https://github.com/vmware-tanzu/kubeapps/commit/6e7091015f9a6c3a289224d05dab5f60736489a0
The rest is just moving code into separate files, mostly.

Also, I have been taking a look at `kubebuilder` to create a new controller, based on the `sigs.k8s.io/controller-runtime` rather than on the workqueues we currently have. While it is pretty easy to start with ([see quickstart](https://book.kubebuilder.io/quick-start)), I think it is adding too much complexity (using kustomize, adding rbac proxies, prometheus metrics, etc...
I also quickly tried the k8s codegen scripts, but ran into some issues with my setup... but perhaps it's the best option.

IMO, at some point we should start thinking about moving towards a new state-of-the-art k8s controller boilerplate.


cmd/apprepository-controller/server/controller.go
```diff
@@ -386,504 +434,100 @@ func (c *Controller) ensureCronJob(apprepo *apprepov1alpha1.AppRepository) (meta
...
-	if apprepo.Spec.FilterRule.JQ != "" {
-		rulesJSON, err := json.Marshal(apprepo.Spec.FilterRule)
-		if err != nil {
-			log.Errorf("Unable to parse filter rules for %s: %v", apprepo.Name, err)
-		} else {
-			args = append(args, "--filter-rules", string(rulesJSON))
-		}
-	}
+			// start the clean-up
+			err = c.cleanUpAppRepo(apprepo, apprepo.GetNamespace(), apprepo.GetName())
```
antgamdia, line 506: I know it sounds a bit hacky/tricky... but when testing IRL (and then confirmed in the k8s repo), a deletion is triggering also update events... so I ended up handling it that way. 
I did my best at explaining the code, but if you can think of a better solution, let me know :P

cmd/apprepository-controller/cmd/root.go
```diff
@@ -85,12 +85,12 @@ func setFlags(c *cobra.Command) {
...
-	c.Flags().StringSliceVar(&serveOpts.CustomAnnotations, "custom-annotations", []string{""}, "optional annotations to be passed to the generated CronJobs, Jobs and Pods objects. For example: my/annotation=foo")
-	c.Flags().StringSliceVar(&serveOpts.CustomLabels, "custom-labels", []string{""}, "optional labels to be passed to the generated CronJobs, Jobs and Pods objects. For example: my/label=foo")
+	c.Flags().StringVar(&serveOpts.TTLSecondsAfterFinished, "ttl-lifetime-afterfinished-job", "3600", "Lifetime limit after which the resource Jobs are deleted expressed in seconds by default is 3600 (1h)")
+	c.Flags().Int32Var(&serveOpts.SuccessfulJobsHistoryLimit, "successful-jobs-history-limit", 3, "Number of successful finished jobs to retain")
+	c.Flags().Int32Var(&serveOpts.FailedJobsHistoryLimit, "failed-jobs-history-limit", 1, "Number of failed finished jobs to retain")
+	c.Flags().StringVar(&serveOpts.ConcurrencyPolicy, "concurrency-policy", "Replace", "How to treat concurrent executions of a Job. Valid values are: 'Allow', 'Forbid' and 'Replace'")
```
antgamdia, line 91: I have also added these extra flags just in case we wanted to configure the cronjob. For now, I left the existing default values, so as to the behavior doesn't actually change.

antgamdia (commented):
Adding some in-line comments. Most changes are just refactoring and separating the code into different files... it was becoming too messy to deal with 😅

absoludity:
> When dealing with finalizers in the past I often found it really annoying when they get stuck and prevent the resource to get deleted. I wonder if we should add some info in the FAQ on how to manually remove the finalizers.

Yeah - might be worth while. I think our finalizer shouldn't cause issues like others do (in that, it's not doing anything complex), but right, we all think that then something unexpected will stop it (DB connection or something). Updating the FAQ sounds like a good idea (once this is out).

> 
> Additionally, and this might be something important: for the AppRepo controller to be able to `update` AppRepos in other namespaces != kubeapps.... (to add the finalizer) it now needs extra RBAC. Before we were just granting `...-appprepos-read`.. but now we would need to grant `...-write` as well...and I'm not sure we really want to do so. WDYT, @absoludity ? Another idea is using an admission policy... but not sure again if we want to deal with that...

Hmm. I don't have a problem with our controller having write access to app repos in all namespaces? It's our CR, I don't think people generally have an issue with the controller having write access across all namespaces. Or did you mean something else?

cmd/apprepository-controller/server/controller.go
```diff
@@ -261,18 +262,18 @@ func (c *Controller) processNextWorkItem() bool {
 			// Forget here else we'd go into a loop of attempting to
 			// process a work item that is invalid.
 			c.workqueue.Forget(obj)
-			runtime.HandleError(fmt.Errorf("expected string in workqueue but got %#v", obj))
+			runtime.HandleError(fmt.Errorf("Expected string in workqueue but got %#v", obj))
```
absoludity, line 265: Any reason for capitalizing these? I know our codebase is not consistent, but if we were making it consistent, I'd go the other way. See https://google.github.io/styleguide/go/decisions.html#error-strings

cmd/apprepository-controller/server/controller.go
```diff
@@ -354,541 +403,167 @@ func (c *Controller) syncHandler(key string) error {
...
-	}, dbFlags(config)...)
-}
-
-func dbFlags(config Config) []string {
-	return []string{
-		"--database-url=" + config.DBURL,
-		"--database-user=" + config.DBUser,
-		"--database-name=" + config.DBName,
-	}
 }
```
absoludity, line 569: Just noting that there's a lot of deletions in this file - guessing they'll be re-added below. EDIT: yep, see them in the job_utils. Mild preference for keeping moves of code like this to separate PRs, but realise it's not always possible when the code is changing too. Sometimes I do the code changes and refrain from moving/reorganising until a follow-up PR, just so it's easier to follow what has actually changed.

cmd/apprepository-controller/server/utils_test.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			if got, want := objectBelongsTo(tc.object, tc.parent), tc.expect; got != want {
+				t.Errorf("got: %t, want: %t", got, want)
+			}
+		})
+	}
+}
```
absoludity, line 219: Thanks for all the new tests for the new functionality too.

site/content/docs/latest/tutorials/managing-package-repositories.md
```diff
@@ -258,3 +258,51 @@ apprepository:
...
+          env:
+            - name: DB_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  key: postgres-password
+                  name: kubeapps-postgresql
+          image: docker.io/kubeapps/asset-syncer:latest
+          name: delete
+EOF
+```
```
absoludity, line 308: And thanks for the doc update!

absoludity (approved):
Unreal - thanks Antonio for going the extra mile to do this properly so we don't accumulate more tech-debt!

chart/kubeapps/templates/apprepository/rbac.yaml
```diff
@@ -43,6 +44,23 @@ rules:
...
     verbs:
       - create
+---
+# ClusterRole for managing AppRepository objects in every namespace,
+# so that we can update the finalizers on AppRepository objects
+apiVersion: {{ include "common.capabilities.rbac.apiVersion" . }}
+kind: ClusterRole
+metadata:
+  name: {{ template "kubeapps.apprepository.fullname" . }}
+  namespace: {{ .Release.Namespace | quote }}
```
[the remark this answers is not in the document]
antgamdia: Ooops, I thought I had actually removed them. They don't have any effect, as you pointed out.
Ah, I know what happened... I copied and pasted the metadata so that all the objects had the same... and I forgot to remove the `namespace` again! Thanks for catching it!

cmd/apprepository-controller/server/controller.go
antgamdia: No reason apart from internal consistency... having mixed-up capitalized log traces was killing me... so I just normalized them using the most common choice (in this piece of code). But I agree, log traces should be lower-cased instead.
If we start changing little by little, at some point we might end up with a consistent codebase :P, which will be helpful if we ever think of exporting metrics with OpenTelemetry or any other stack.

For the record, the regex `(debug|warn|Error|Info)(f?)\(("|')\p{Lu}` yields 549 upper-cased occurrences across 66 `.go` files. For now, I'm going to lower-case the ones in apprepo controller.

antgamdia: Mmm, you're right... I was mixing up the error strings (fmt.Error....) and the log error traces (log.Errorf...), so:

- log.* --> uppercased
- fmt.Errorf --> lowercased

Reverting changes in https://github.com/vmware-tanzu/kubeapps/pull/6647/commits/1696b831364b342887b03c582a045e30c0a9c334 then  :P

Issue #6706, opened 2023-08-25
Support public OCI registries without Auth


absoludity:
**Summary**

Following on from #6263 , where we've added support for listing repositories and tags for a registry (DockerHub) without credentials via the OCI-Catalog service, we'd like to ensure that the complete process of adding a public OCI catalog, such as the Bitnami Catalog, can be done without requiring credentials - even if it is rate limited.

The issue is that the existing code, which pulls chart data using the OCI Distribution Spec API, requires a credential to interact with that API. It may well be that we can't pull the chart tarball without any credential, so another solution to the same end would be to ensure Kubeapps itself has a docker credential (or similar for other public registries) - not sure yet.

**Background and rationale**

We want to demonstrate the OCI catalog enabling adding the Bitnami OCI chart catalog.

**Acceptance criteria**

This will be complete when a user can add the `oci://registry-1.docker.io/bitnamicharts` as the URL for an `AppRepository`, without credentials, and have Kubeapps populate the catalog (up until any rate limit is hit - that can't be worked around without a credential).

absoludity:
> The issue is that the existing code, which pulls chart data using the OCI Distribution Spec API, requires a credential to interact with that API. It may well be that we can't pull the chart tarball without any credential, so another solution to the same end would be to ensure Kubeapps itself has a docker credential (or similar for other public registries) - not sure yet.

In today's happy news: it turns out that this is only partly correct - yes, the OCI Distribution Spec API requires a credential to interact with the API, but an anonymous credential can be obtained for public repositories using the `WWW-Authenticate` header from the original 401 request (tested both with dockerhub and harbor). So we can use the existing code as originally planned just with a small update to request the anon token if a 401 is received!

With Harbor, verified with:

```console
token=$(curl -s "https://demo.goharbor.io/service/token?service=harbor-registry&scope=repository:testpublic/kubeapps/oci-catalog:pull" \          
        | jq -r '.token')

curl -v -H "Accept: application/vnd.oci.image.manifest.v1+json" -H "Authorization: Bearer $token" https://demo.goharbor.io/v2/testpublic/kubeapps/oci-catalog/manifests/dev6
```

and similarly with dockerhub:

```
token=$(curl -s "https://auth.docker.io/token?service=registry.docker.io&scope=repository:${repo}:pull" \                                       | jq -r '.token')

[...]

PR #6804, opened 2023-09-13
Less obnoxious OCI pulling on sync

absoludity:
### Description of the change

This PR update the way we pull OCI repos, so that we:
- Don't pull all manifests of all tags to verify that they are helm charts before syncing
- Don't do an all-or-nothing sync (previous code synced all tags from the repo if the checksum differed), instead only syncing those tags that don't exist in the cache.
- Reduces the number of workers to 5 for now, but I want to put this back to 10 in a subsequent PR with other changes.

### Benefits

We can pull large OCI catalogs from public repos (in particular, can pull the Bitnami catalog from Dockerhub) without hitting public pull limits.

### Possible drawbacks

The initial time until the catalog can be used is slightly longer. But I have two things I want to improve in a subsequent PR:

- Update the cache after each app/repo is synced so they can be viewed in the catalog more quickly (more responsive). Currently it doesn't update the cache until the sync completes.
- Only sync a maximum of 5 missing tags per app per sync. This means that we can run the initial sync with 10 workers, and will get the latest 5 versions for each app, then subsequent syncs will add more each time. We should make this adjustable too, perhaps.

This combination of changes will also remove another issue we have where a subsequent sync, running within 10m of the original (by default) will do the same work currently, rather than just doing the work that hasn't been done yet.

### Applicable issues

- ref #6706

### DIFF
--- a/cmd/asset-syncer/cmd/root.go
+++ b/cmd/asset-syncer/cmd/root.go
@@ -104,7 +104,7 @@ func init() {
 	log.InitFlags(nil)
 	cobra.OnInitialize(initConfig)
 	//set initial value of verbosity
-	err := flag.Set("v", "3")
+	err := flag.Set("v", "4")
 	if err != nil {
 		log.Errorf("Error parsing verbosity: %v", viper.ConfigFileUsed())
 	}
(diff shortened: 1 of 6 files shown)


ppbaena (approved):
Thanks Michael!

absoludity:
> Thanks Michael!

Thanks for the review. Just note, I'm holding of landing this until I have the follow-up PRs also done and can check it works as expected locally. First one is at #6825 but there will be one more following that also. I'll land them once tested locally.

PR #6826, opened 2023-09-15
Ensure already synced versions are included in the result.

### LINKED ISSUE #6706: Support public OCI registries without Auth

absoludity:
### Description of the change

Currently ensures that the existing (already synced) chart versions are included when syncing new chart versions, so that the full set of data is retained in the DB (since the chart versions aren't actually rows in the db, but rather within a json blob of the chart).

This PR refactored the way the syncing happens, specifically ensuring that data is synced to the db as each chart is processed, rather than when all charts are synced (so that we get better feedback in the UI when syincing OCI repos).

### Benefits

Before the change:
- a shallow + deep sync of the non-OCI Helm repo takes around 7s
- a shallow sync of the Bitnami OCI repo would see us hitting the rate limit and getting 429s.

With the changes,
- a shallow + deep sync of a non-OCI Helm repo takes around 10s (extra overhead of sycing to the DB for each chart)
- a shallow sync of the Bitnami OCI repo takes 2 minutes and a syncing the next 5 versions for each chart takes a further 2 minutes (and does not excessively hammer the dockerhub distribution spec API to get banned). (should remove the shallow and just always do 5, also should test 10).

### Possible drawbacks

Currently:
- ~~Something not correct with the db info being synced as getting `Unable to get the available package detail for the package "testoci2/airflow" using the plugin "helm.packages": internal: Unable to retrieve chart files: sql: no rows in result set`~~ Fixed - it was a synchronization issue that left the sync process finishing after charts were synced but before files were synced.
- ~~OCI catalog is taking a *long* time to display (shouldn't be affected, getting info from the DB).~~ This was a resource issue locally.

~~So still more investigation here needed.~~

### Applicable issues

- fixes #6706

### DIFF
--- a/cmd/apprepository-controller/Dockerfile
+++ b/cmd/apprepository-controller/Dockerfile
@@ -3,7 +3,7 @@
 
 # syntax = docker/dockerfile:1
 
-FROM bitnami/golang:1.20.7 as builder
+FROM bitnami/golang:1.21.1 as builder
 WORKDIR /go/src/github.com/vmware-tanzu/kubeapps
 COPY go.mod go.sum ./
 COPY pkg pkg
--- a/cmd/asset-syncer/Dockerfile
+++ b/cmd/asset-syncer/Dockerfile
@@ -3,7 +3,7 @@
 
 # syntax = docker/dockerfile:1
 
-FROM bitnami/golang:1.20.7 as builder
+FROM bitnami/golang:1.21.1 as builder
 WORKDIR /go/src/github.com/vmware-tanzu/kubeapps
 COPY go.mod go.sum ./
 COPY pkg pkg
--- a/cmd/asset-syncer/server/postgresql_db_test.go
+++ b/cmd/asset-syncer/server/postgresql_db_test.go
@@ -366,7 +366,7 @@ func TestRemoveMissingCharts(t *testing.T) {
 				ensureFilesExist(t, pam, chartID, files)
 			}
 
-			err := pam.removeMissingCharts(repo, tc.remainingCharts)
+			err := pam.RemoveMissingCharts(repo, []string{"TODO", "fixme"})
 			if err != nil {
 				t.Fatalf("%+v", err)
 			}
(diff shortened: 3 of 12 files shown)


absoludity:
Updated today so that we track the charts that can be deleted and refactored to ensure we don't simply remove charts that aren't synced. Still need to test locally, but once I confirm that's working (or fix to ensure it works), I'm happy to trigger a release.

absoludity:
Just ran this locally to test it out, a couple of things affected for the Helm (non-OCI) import to check tomorrow:
- n workers (go-routines) being spawned for each call to fetchFiles (fix to ensure we either request the file fetching via a channel too, so that the 10 workers can operate on all requsets) 
```
I0918 11:44:53.472293       1 utils.go:955] Starting 10 workers
I0918 11:44:53.472368       1 utils.go:997] Importing icon, name=vault
I0918 11:44:53.473160       1 utils.go:1003] Importing readme and values, ID=bitnami/vault, version=0.3.3
I0918 11:44:53.476385       1 utils.go:1075] Fetching files, id=bitnami/vault, version=0.3.3
I0918 11:44:53.794933       1 sync.go:142] Repository synced, shallow=true
```
- The remove missing charts fails with the following error (even though the pg test passed), investigate and fix:
```
I0918 11:44:55.407894       1 sync.go:142] Repository synced, shallow=true
I0918 11:44:55.407949       1 postgresql_utils.go:92] Removing the following charts that are no longer present in the repo: 
Error: error while removing missing charts: pq: syntax error at or near ")"

```

PR #6827, opened 2023-09-15
Identify/fix outdated docs

antgamdia:
### Description of the change

Even if we have changed several things during recent months, our docs didn't get updated at the same pace. This PR (still a WIP draft) is trying to go through the docs and add TODOs where required and/or to add easy fixes.

### Benefits

We will know which doc pages need a revamp/update/etc (screenshots, wording, etc...)

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

Once finished, an issue summarizing all the TODOs should be created


absoludity (approved):
Thanks Antonio for the attention to detail as always. Approving in case you don't get to put any time into this, as it's still worthwhile as is.

